PR #1902, opened 2022-05-24
drt: calculate and report guide coverage

osamahammad21:
[OR-1151](https://openroad.atlassian.net/browse/OR-1151)


src/drt/src/dr/FlexDR.cpp
```diff
@@ -2328,6 +2328,148 @@ static std::vector<FlexDR::SearchRepairArgs> strategy()
...
+    }
+
+    for (frLayerNum lNum = 0; lNum < getTech()->getLayers().size(); lNum++) {
+      if (getTech()->getLayer(lNum)->getType() != dbTechLayerType::ROUTING
+          || lNum > TOP_ROUTING_LAYER)
+        continue;
+      if (routeSetByLayerNum[lNum].empty()) {
+        #pragma omp critical
+        netsCoverage[net.get()].push_back(-1.0);
+        continue;
```
maliberty, line 2426: Two different critical sections can run concurrently so this isn't thread-safe.  The section below is accessing netsCoverage too in a different critical section.

src/drt/src/dr/FlexDR.cpp
```diff
@@ -2328,6 +2328,148 @@ static std::vector<FlexDR::SearchRepairArgs> strategy()
...
+      }
+      auto routingArea = gtl::area(routeSetByLayerNum[lNum]);
+      auto coveredArea
+          = gtl::area(routeSetByLayerNum[lNum] & guideSetByLayerNum[lNum]);
+      auto coveredPercentage = (coveredArea / (routingArea * 1.0)) * 100;
```
maliberty, line 2431: Though the compiler will probably optimize away '* 1.0' it would be better to just cast to double rather than multiply.

src/drt/src/dr/FlexDR.cpp
```diff
@@ -2328,6 +2328,148 @@ static std::vector<FlexDR::SearchRepairArgs> strategy()
           /* 63 */ {7, -6, 64, shapeCost * 64, MARKERCOST * 16, 0, false}};
 }
 
+void FlexDR::reportGuideCoverage()
+{
+  using namespace boost::polygon::operators;
+
+  ofstream file(GUIDE_REPORT_FILE);
```
maliberty, line 2335: Its odd to split the writing of the file around the computation.  Just move this after the computation is done.

src/drt/src/dr/FlexDR.cpp
```diff
@@ -2328,6 +2328,148 @@ static std::vector<FlexDR::SearchRepairArgs> strategy()
...
+      else
+        file << fmt::format("{:.2f}%,", coveredPercentage);
+    file << std::endl;
+  }
+  file << "Total,";
+  unsigned long long totalArea = 0;
+  unsigned long long totalCoveredArea = 0;
+  for (const auto& layer : getTech()->getLayers()) {
+    if (layer->getType() == dbTechLayerType::ROUTING
+        && layer->getLayerNum() <= TOP_ROUTING_LAYER) {
```
maliberty, line 2454: It would be nice to have a getRoutingLayers() rather than repeating this filtering everywhere.

maliberty (changes requested):

maliberty (approved):

PR #1915, opened 2022-05-26
Grt cleanup

jjcherry56:


src/grt/src/GlobalRouter.cpp
```diff
@@ -282,8 +282,11 @@ NetRouteMap GlobalRouter::findRouting(std::vector<Net*>& nets,
                                       int max_routing_layer)
 {
   NetRouteMap routes;
-  if (getNetCount() > 0) {
-    routes = fastroute_->run();
+  if (!nets.empty()) {
+    std::vector<odb::dbNet*> db_nets;
```
QuantamHD, line 286: Since we already know the size of nets could we reserve the correct amount of space for db_nets?

PR #1917, opened 2022-05-26
Odb group

osamahammad21:


src/odb/src/db/dbRegion.cpp
```diff
@@ -349,6 +380,26 @@ dbBlock* dbRegion::getBlock()
   return (dbBlock*) getImpl()->getOwner();
 }
 
+void dbRegion::addGroup(dbGroup* group)
+{
+  _dbRegion* _region = (_dbRegion*) this;
+  _dbGroup* _group = (_dbGroup*) group;
+  if(_group->region_)
+    return;
```
maliberty, line 388: clang-format

src/odb/src/db/dbRegion.cpp
```diff
@@ -349,6 +380,26 @@ dbBlock* dbRegion::getBlock()
...
+  _group->region_ = _region->getOID();
+  _group->region_next_ = _region->groups_;
```
maliberty, line 390: You need to set the region_prev on the old head to the new group

src/odb/src/db/dbRegion.cpp
```diff
@@ -308,6 +310,35 @@ void dbRegion::removeInst(dbInst* inst_)
 
   inst->_region = 0;
 }
+void dbRegion::removeGroup(dbGroup* group)
```
maliberty, line 313: Verify that the group does belong to this region

src/ifp/src/InitFloorplan.cc
```diff
@@ -227,16 +227,19 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
   // checks if a group is defined as a voltage domain, if so it creates a region
   for (dbGroup* group : block_->getGroups()) {
     if (group->getType() == dbGroupType::VOLTAGE_DOMAIN) {
-      dbRegion* domain_region = dbRegion::create(block_, group->getName());
-
+      dbRegion* domain_region = group->getRegion();
```
maliberty, line 230: Ok for now as we only create single boxes but this will need to deal with rectilinear regions eventually.

src/odb/test/cpp/TestGroup.cpp
```diff
@@ -27,7 +27,8 @@ struct F_DETAILED {
     n2 = block->findNet("n2");
     n3 = block->findNet("n3");
     group  = dbGroup::create(block,"group");
-    domain = dbGroup::create(block,"domain",0,0,100,100);
+    domain = dbGroup::create(block,"domain");
```
maliberty, line 30: Do we have tests for the group/region relationship?

maliberty (changes requested):
Please update defin to map DEF GROUPS to dbGroups

maliberty:
You'll also need to look at Opendp::makeGroups

maliberty:
Drop dbRegions as parent's of dbRegions since dbGroups replace that usage.

osamahammad21:
> Remove dbInsts from dbRegion as those should be handled by dbGroup.

@maliberty according to the lefdefref, in the COMPONENTS section, the inst could have a REGION option.

maliberty:
> > Remove dbInsts from dbRegion as those should be handled by dbGroup.
> 
> @maliberty according to the lefdefref, in the COMPONENTS section, the inst could have a REGION option.

Ok but what a horrible model DEF has.

maliberty:
Please ping me for review again once the CI issues are resolved.

osamahammad21:
@maliberty it's just some warnings I'll remove from the ok file. However, while I was editing dpl/test/fence01.ok, I found that this ok file has differences generated in the end between the resulting def and the defok. I think someone just saved the ok files with these generated differences. Should I fix it myself or does this need more investigation by whomever working on dpl?

maliberty (approved):

PR #1930, opened 2022-05-31
Optimizes LayoutViewer performance

QuantamHD:
This PR optimizes the LayoutViewer performance by
removing anti aliasing routines from axis aligned boxes and
replaces our Pixmap target with a QImage instead.

I experimented with threaded rendering, but found the code
unsavory in its current iteration. We should explore using
an OpenGL paint device instead since a lot of the rendering
we're doing is very suboptimal.

Potential Areas of Improvement
* 50% of render loop is spend in drawing the PinLabels. This is being
  eaten up by both the layout code and all the calls to getBoundingBox
  for the text. We should make sure that we are caching these layouts
  and not rendering text if it falls off screen.

* Using OpenGL instead of our own custom rasterizer. A lot of time is
  spent drawing rectangles that are occuluded by other rectangles. On a
  CPU renderer this is killing our performance. Filling pixels is the
  single most expensive thing we could be doing. By moving to OpenGL
  (even a software implementation) we would benefit from its occlusion
  culling algorithms which would be faster than our implementation.

* Distance based culling, possibly with QuadTree. When rendering the
  power grid at low resolution our framerate dips significantly. We
  should try to throw away geometry if it's not going to visibile. I can
  think of a few methods to doing this, but it's not exactly a simple
  optimization.

* MipMapping/Caching different tile based renders of the layout. As a
  opposed to dynamically rendering them.

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -40,6 +40,7 @@
 #include <QGridLayout>
 #include <QHBoxLayout>
 #include <QImageWriter>
+#include <QtGlobal>
 #include <QLabel>
 #include <QLineEdit>
 #include <QPaintEvent>
@@ -226,7 +227,7 @@ class GuiPainter : public Painter
     painter_->drawLine(x - o, y + o, x + o, y - o);
   }
 
-  const odb::Point determineStringOrigin(int x, int y, Anchor anchor, const QString& text, bool rotate_90 = false, QRect *bounding_rect = nullptr)
+  const odb::Point determineStringOrigin(int x, int y, Anchor anchor, const QString& text, bool rotate_90 = false, const QRect *bounding_rect = nullptr)
   {
     
     QRect text_bbox;
@@ -310,7 +311,7 @@ class GuiPainter : public Painter
   {
     const QString text = QString::fromStdString(s);
     const QRect text_bbox = painter_->fontMetrics().boundingRect(text);
-    const odb::Point origin = determineStringOrigin(x, y, anchor, text, &text_bbox);
+    const odb::Point origin = determineStringOrigin(x, y, anchor, text, false, &text_bbox);
 
     const qreal scale_adjust = 1.0 / getPixelsPerDBU();
 
@@ -2247,7 +2248,11 @@ void LayoutViewer::drawPinMarkers(Painter& painter,
   QPainter* qpainter = static_cast<GuiPainter&>(painter).getPainter();
   const QFont initial_font = qpainter->font();
   QFont marker_font = options_->pinMarkersFont();
+  
+  #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
   marker_font.setStyleStrategy(QFont::StyleStrategy::PreferNoShaping);
+  #endif
+
   qpainter->setFont(marker_font);
 
   const QFontMetrics font_metrics(marker_font);
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -33,8 +33,8 @@
 #include "layoutViewer.h"
 
 #include <QApplication>
-#include <QDebug>
 #include <QDateTime>
+#include <QDebug>
 #include <QFileDialog>
 #include <QFont>
 #include <QGridLayout>
@@ -49,12 +49,15 @@
 #include <QScrollBar>
 #include <QSizePolicy>
 #include <QStaticText>
+#include <QTextOption>
 #include <QToolButton>
 #include <QToolTip>
 #include <QTranslator>
 #include <deque>
 #include <iostream>
+#include <ostream>
 #include <tuple>
+#include <utility>
 #include <vector>
 
 #include "db.h"
@@ -62,11 +65,10 @@
 #include "gui/gui.h"
 #include "highlightGroupDialog.h"
 #include "mainWindow.h"
-#include "search.h"
-#include "scriptWidget.h"
-#include "utl/Logger.h"
-
 #include "ruler.h"
+#include "scriptWidget.h"
+#include "search.h"
+#include "utl/Logger.h"
 
 // Qt's coordinate system is defined with the origin at the UPPER-left
 // and y values increase as you move DOWN the screen.  All EDA tools
@@ -224,9 +226,16 @@ class GuiPainter : public Painter
     painter_->drawLine(x - o, y + o, x + o, y - o);
   }
 
-  const odb::Point determineStringOrigin(int x, int y, Anchor anchor, const QString& text, bool rotate_90 = false)
+  const odb::Point determineStringOrigin(int x, int y, Anchor anchor, const QString& text, bool rotate_90 = false, QRect *bounding_rect = nullptr)
   {
-    const QRect text_bbox = painter_->fontMetrics().boundingRect(text);
+    
+    QRect text_bbox;
+    if (bounding_rect) {
+      text_bbox = *bounding_rect;
+    } else {
+      text_bbox = painter_->fontMetrics().boundingRect(text);
+    }
+    
     const QPoint text_bbox_center = text_bbox.center();
 
     const qreal scale_adjust = 1.0 / getPixelsPerDBU();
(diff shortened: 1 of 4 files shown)


maliberty:
We already drop small objects when rendering unless you have 'detailed view' turned on.   I guess you have it on if you see pin labels taking that long.

The layout is alpha blended so nothing is fully occluded.

You will find some people don't have opengl available and will still need a fallback (I did a GL enabled gui before and it happens with thin-client workstations or companies whose IT won't support a GL enabled vnc no matter how much you ask).   I think it is better to treat this as a slippy map like google maps with more caching.

QuantamHD:
https://github.com/rswinkle/PortableGL is the one I'm familiar with

QuantamHD:
I think O(600) pins

maliberty:
Something seems out of whack - the bbox of 600 pieces of text should be very fast compared to rendering all the layout shapes.  Is this a fully routed design you are viewing?

QuantamHD:
Nope it's just that text shaping with harfbuzz is very very slow. Text rendering is some of the most complicated code you'll run across. 

image

QuantamHD:
Harfbuzz routines 
image

maliberty:
Do you have a count of how many times these APIs are being called?  Does it match you pin count?

QuantamHD:
I haven't Instrumented it to that level, but that shouldn't impact the goodness of this patch set.

QuantamHD:
I'll also point out that time is relative here. I think the pin grids are eating 20 fps of time or roughly 50ms of frame time.

If we want to hit 60fps we need to complete each frame in less than 16ms. That's going to require some very smart planning on our side with graphics accelerators.

src/gui/src/layoutViewer.cpp
```diff
@@ -2049,6 +2064,7 @@ void LayoutViewer::drawBlock(QPainter* painter,
                                      bounds.yMax(),
                                      instance_limit);
 
+    painter->setRenderHint(QPainter::RenderHint::Antialiasing, false);
```
maliberty, line 2067: Shapes on the RDL layer often contain 45 degree segments and would not be anti-aliased.

src/gui/src/layoutViewer.cpp
```diff
@@ -2468,7 +2486,8 @@ void LayoutViewer::updateBlockPainting(const QRect& area)
   repaint_requested_ = false;
 
   // build new drawing of layout
-  auto* block_drawing = new QPixmap(area.width(), area.height());
+  auto* block_drawing = new QImage(area.width(), area.height(),
```
maliberty, line 2489: Have you measured that this is faster?  
"QImage is designed and optimized for I/O, and for direct pixel access and manipulation, while [QPixmap](https://doc.qt.io/qt-5/qpixmap.html) is designed and optimized for showing images on screen."

src/gui/src/layoutViewer.cpp
```diff
@@ -1746,6 +1759,7 @@ void LayoutViewer::drawInstanceOutlines(QPainter* painter,
 
   painter->setPen(QPen(Qt::gray, 0));
   painter->setBrush(QBrush());
+  painter->setRenderHint(QPainter::RenderHint::Antialiasing, false);
```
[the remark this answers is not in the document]
QuantamHD: I question the value of the anti-aliasing at all. This feels like a good tradeoff for a close to 4x improvement in FPS.

src/gui/src/layoutViewer.cpp
QuantamHD: 20% reduction in frametime.

src/gui/src/layoutViewer.cpp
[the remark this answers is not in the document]
arlpetergadfort: I would say that if the drawing quality suffers, we should keep the anti-aliasing, since I'm sure there are other improvements that will have bigger impact without reducing the image quality.

src/gui/src/layoutViewer.cpp
```diff
@@ -2235,6 +2248,11 @@ void LayoutViewer::drawPinMarkers(Painter& painter,
   QPainter* qpainter = static_cast<GuiPainter&>(painter).getPainter();
   const QFont initial_font = qpainter->font();
   QFont marker_font = options_->pinMarkersFont();
+  
+  #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+  marker_font.setStyleStrategy(QFont::StyleStrategy::PreferNoShaping);
```
arlpetergadfort, line 2253: Is there an option available before 5.10?

src/gui/src/layoutViewer.cpp
```diff
@@ -2485,8 +2504,24 @@ void LayoutViewer::updateBlockPainting(const QRect& area)
...
+    QFont fps_font;
+    fps_font.setPixelSize(25);
+    fps_font.setBold(true);
+    block_painter.resetTransform();
+    block_painter.setFont(fps_font);
+    block_painter.setPen(Qt::yellow);
+    block_painter.translate(10, 40);
+    block_painter.drawText(0, 0,
+                          "fps: " + QString::number(1000.0 / frame_time_ms));
+  }
```
arlpetergadfort, line 2521: FPS is not really what you are measuring, since we do not force repaints on every possible attempt. I like the idea of having this available, but might be good to instrument the code with a "debug" mode instead that can provide an overlay with the paint times for major blocks of code (pins, instances, layers, etc.) so we can better isolate what improvements can be made. I would just use the paint time instead of FPS.

src/gui/src/layoutViewer.h
```diff
@@ -126,6 +127,7 @@ class LayoutViewer : public QWidget
                const std::vector<std::unique_ptr<Ruler>>& rulers,
                std::function<Selected(const std::any&)> makeSelected,
                std::function<bool(void)> usingDBU,
+               std::function<bool(void)> showFps,
```
arlpetergadfort, line 130: Maybe this can be accessed via the tcl interface instead?

src/gui/src/layoutViewer.cpp
```diff
@@ -224,9 +227,16 @@ class GuiPainter : public Painter
     painter_->drawLine(x - o, y + o, x + o, y - o);
   }
 
-  const odb::Point determineStringOrigin(int x, int y, Anchor anchor, const QString& text, bool rotate_90 = false)
+  const odb::Point determineStringOrigin(int x, int y, Anchor anchor, const QString& text, bool rotate_90 = false, const QRect *bounding_rect = nullptr)
```
arlpetergadfort, line 230: Since we use the boundingRect (alot) maybe it would be possible to cache it centrally, so we can look up the bounding box if its computed? This is just one of many places we use this and if this adds a lot of time, there would be a lot of value in correcting that.

src/gui/src/layoutViewer.cpp
```diff
@@ -33,13 +33,14 @@
...
 #include <QApplication>
-#include <QDebug>
 #include <QDateTime>
+#include <QDebug>
 #include <QFileDialog>
 #include <QFont>
 #include <QGridLayout>
 #include <QHBoxLayout>
 #include <QImageWriter>
+#include <QtGlobal>
```
arlpetergadfort, line 43: why is this needed?

src/gui/src/layoutViewer.cpp
```diff
@@ -49,25 +50,27 @@
 #include <QScrollBar>
 #include <QSizePolicy>
 #include <QStaticText>
+#include <QTextOption>
```
arlpetergadfort, line 53: why is this needed?

src/gui/src/layoutViewer.cpp
```diff
@@ -49,25 +50,27 @@
...
 #include <QToolButton>
 #include <QToolTip>
 #include <QTranslator>
 #include <deque>
 #include <iostream>
+#include <ostream>
 #include <tuple>
+#include <utility>
```
arlpetergadfort, line 61: why is this needed?

src/gui/src/layoutViewer.cpp
QuantamHD: This is for the version check I used in the NoShaping ifdef.

src/gui/src/layoutViewer.h
QuantamHD: I think we're trying to move away from using the TCL interface inside the application itself. Maybe we could package these into an object instead?

arlpetergadfort: hmm, maybe the solution would be to have an access to through the GUI to "trigger" these from swig so it would extend easier for python and be extensible as we add more layer. I'll make a note for future fixes.  Either way this is fine. If we choose to make it more of an debugging overlay, then the name would change but thats about it.

src/gui/src/layoutViewer.cpp
QuantamHD: We should be able to  create a QCache of them. I'll add this.

src/gui/src/layoutViewer.cpp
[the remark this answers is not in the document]
QuantamHD: Anti-Aliasing isn't cheap though, it nearly doubles the cost of draw calls for a what I think is a pretty small benefit in the grand scheme of things. If we had GPU acceleration I wouldn't mention it at all, but CPU renderers just do not have the memory bandwidth to drive interactive experiences without significant tradeoffs. 

With high resolution monitors becoming so commonplace it feels like this is a luxury feature that isn't adding as much value as it used to, while detracting from interactivity which is very annoying for the user. I'm amenable to make it a view option if you think that's better.

maliberty:
This looks to be stale and will have a lot of conflicts with the threading changes.  Reopen if you are going to work on this.

PR #1936, opened 2022-06-05
K8s drt cpu headless

Mahfouz-z:
Adding support for balancer auto-discovery of workers using k8s headless service.


src/dst/src/Distributed.cc
```diff
@@ -104,6 +106,32 @@ void Distributed::runLoadBalancer(const char* ip, unsigned short port)
   }
 }
 
+bool Distributed::lookUpWorkers(const char* DNS, unsigned short port)
+{
+  asio::io_service ios;
+  asio::ip::udp::resolver::query resolver_query(DNS, std::to_string(port), asio::ip::udp::resolver::query::numeric_service);
```
osamahammad21, line 112: too many namespaces. replace with using asio::ip::udp and using asio::ip::udp::resolver

src/dst/src/Distributed.cc
```diff
@@ -104,6 +106,32 @@ void Distributed::runLoadBalancer(const char* ip, unsigned short port)
...
+  asio::ip::udp::resolver::iterator it_end;
+
+  for (; it != it_end; ++it) {
+      asio::ip::udp::endpoint ep = it->endpoint();
+      addWorkerAddress(ep.address().to_string().c_str(), port);
+      workers_count += 1;
+    }
+
+  if(workers_count == 0){
+    logger_->error(utl::DST, 201, "Discovered zero workers with the given DNS.");
```
osamahammad21, line 126: firing an error might be an overkill. There might not be workers up for now but they might be up later. That's why I think there should be a discovery thread that's constantly looking up workers (every n seconds). What do you think?

osamahammad21 (changes requested):

src/dst/src/Distributed.cc
```diff
@@ -91,11 +91,13 @@ void Distributed::runWorker(const char* ip,
   }
 }
 
-void Distributed::runLoadBalancer(const char* ip, unsigned short port)
+void Distributed::runLoadBalancer(const char* ip, unsigned short port, const char* workersDNS)
 {
   try {
     asio::io_service io_service;
     LoadBalancer balancer(this, io_service, logger_, ip, port);
+    if(!lookUpWorkers(workersDNS, port))
```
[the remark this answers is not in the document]
Mahfouz-z: Do you already identify the first message? Is it the message of type "JobMessage::BALANCER"? Maybe instead of starting workers discovery later, we can start the discovery thread that you mention later with the start of the balancer?

src/dst/src/Distributed.cc
Mahfouz-z: will reduce it to warn. Agreed with the discovery thread.

src/dst/src/Distributed.cc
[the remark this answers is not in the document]
osamahammad21: There is no first message identification for now. But I agree with the discovery thread.

src/dst/src/LoadBalancer.cc
```diff
@@ -103,6 +105,46 @@ void LoadBalancer::getNextWorker(ip::address& ip, unsigned short& port)
...
+    if(ec)
+      logger_->error(utl::DST, 203, "Workers domain resolution failed with error code = {}. Message = {}.", 
+        ec.value(),
+        ec.message());
+    int new_workers_count = 0;
+    udp::resolver::iterator it_end;
+    for (; it != it_end; ++it) {
+        auto discovered_worker = worker(it->endpoint().address(), port, 0);
+        if(std::find(workers_set.begin(), workers_set.end(), discovered_worker) == workers_set.end()){
+          workers_set.push_back(discovered_worker);
```
osamahammad21, line 126: I think workers_set could be local variable to the function only since it'not used anywhere else and lookupWorkers doesn't return.

osamahammad21 (changes requested):

Mahfouz-z:
@osamahammad21 kind reminder to approve or request further reviews.

src/dst/src/LoadBalancer.cc
```diff
@@ -103,6 +119,64 @@ void LoadBalancer::getNextWorker(ip::address& ip, unsigned short& port)
...
+    std::vector<worker> new_workers;
+    boost::system::error_code ec;
+    auto it = resolver.resolve(resolver_query, ec);
+    if (ec)
+      logger_->error(utl::DST,
+                     203,
+                     "Workers domain resolution failed with error code = {}. "
+                     "Message = {}.",
+                     ec.value(),
+                     ec.message());
```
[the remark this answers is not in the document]
osamahammad21: a warn would be sufficient. If you want to handle it correctly, you could have a counter for this type of error. if it happens over a max number of times, fire an error.

osamahammad21 (changes requested):

src/dst/src/LoadBalancer.cc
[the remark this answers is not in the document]
osamahammad21: You don't have to add the counter in this PR. It could be handled in another PR that handles failures.

src/dst/src/Distributed.cc
```diff
@@ -91,13 +91,16 @@ void Distributed::runWorker(const char* ip,
...
+                                  unsigned short port,
+                                  const char* workers_domain)
 {
   try {
     asio::io_service io_service;
-    LoadBalancer balancer(this, io_service, logger_, ip, port);
-    for (auto worker : end_points_)
-      balancer.addWorker(worker.ip, worker.port);
+    LoadBalancer balancer(this, io_service, logger_, ip, workers_domain, port);
+    if (std::strcmp(workers_domain, "") != 0)
```
osamahammad21, line 101: ==

src/dst/src/LoadBalancer.cc
```diff
@@ -103,6 +119,64 @@ void LoadBalancer::getNextWorker(ip::address& ip, unsigned short& port)
...
+                 "load_balancer",
+                 1,
+                 "Discovered {} new workers with the given domain. Total "
+                 "workers = {}.",
+                 new_workers_count,
+                 workers_set.size());
+    }
+
+    for (auto worker : new_workers)
+      this->addWorker(worker.ip.to_string(), worker.port);
```
osamahammad21, line 174: why this

src/dst/src/LoadBalancer.cc
```diff
@@ -103,6 +119,64 @@ void LoadBalancer::getNextWorker(ip::address& ip, unsigned short& port)
   }
 }
 
+void LoadBalancer::lookUpWorkers(
+    const char* domain,
+    unsigned short port)  // LoadBalancer & balancer)
```
osamahammad21, line 124: remove unneeded comment

osamahammad21 (approved):
you can merge after applying the final changes requested in this review

src/dst/src/LoadBalancer.cc
Mahfouz-z: If new workers are discovered on the network, they should be added to the list of workers available for the load balancer. For instance, if the load balancer is launched before some of the workers were up.

osamahammad21: No I mean why "this->" 
just call addWorker...

src/dst/src/Distributed.cc
osamahammad21: the equality holds here but not in LoadBalancer.cc.
Here you are checking if workers_domain == "" to add workers manually from the list of workers. However, in LoadBalancer.cc you are checking that workers_domain != "" to lookUpWorkers. It's always better to use strings than cstrings to avoid such problems

osamahammad21 (changes requested):

PR #1942, opened 2022-06-07
fix coverity warnings

osamahammad21:


src/drt/src/db/obj/frInstBlockage.h
```diff
@@ -51,13 +51,13 @@ class frInstBlockage : public frBlockObject
   // setters
   // others
   frBlockObjectEnum typeId() const override { return frcInstBlockage; }
-  void setOrderId(int in) { order_id_ = in; }
-  int getOrderId() const { return order_id_; }
+  void setOrderId(int in) { index_in_owner_ = in; }
```
maliberty, line 54: Since you changed the member name the method name should update too

maliberty (approved):
I'll approve assuming you'll fix my comment

PR #1952, opened 2022-06-11
Fix SegFault in CTS

QuantamHD:
My PDK lists clock buffer cell inputs as signal type clock rather than generic signal. What ends up happening is this code fails to set charBufIn leading to a null returned BTerm down the line and segfault.

This triggers an error if they are not found, and makes sure to include the clock type as a valid input port.


maliberty (approved):

QuantamHD:
@maliberty I can't merge things in so you need to click the button whenever you're ready. I'm pretty sure the Mac failure is unrelated

PR #1968, opened 2022-06-17
grt: sorting nets by name before starting FastRouter

luis201420:
Signed-off-by: luis201420 <tauro_luisito_20@hotmail.com>


maliberty:
There are CI issues to resolve.  Does this affect results in any way?

luis201420:
There are errors when executing the grt tests (regression), I think because they expect a result with the nets in a different order.

maliberty:
In that case you should be updating the ok files in the PR

maliberty (approved):

PR #1972, opened 2022-06-20
Adds support for the basic LEF58_CLASS endcap types

QuantamHD:
My PDK includes these fancy LEF properties that define
the master type. I added support for these, but not all
class properties which includes a much larger grammar.

I added a warning, and we should add others as they appear.


src/odb/src/lefin/lefMacroPropParser.h
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2002, Google LLC
```
maliberty, line 2: Time travel? 2022

src/odb/src/lefin/lefMacroClassParser.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "db.h"
+#include "lefMacroPropParser.h"
+#include "lefin.h"
+
+namespace lefMacroClassType {
+
+template <typename Iterator>
+bool parse(Iterator first, Iterator last, odb::dbMaster* master) {
+  qi::rule<std::string::iterator, space_type> TypeRule =
+      (lit("CLASS") >>
```
maliberty, line 40: You need all the ordinary cases too - COVER, RING, BLOCK, PAD, CORE and their subtypes.

src/odb/src/lefin/lefMacroClassParser.cpp
```diff
@@ -0,0 +1,91 @@
...
+       (lit("ENDCAP") >>
+        (lit("BOTTOMEDGE")[boost::bind(
+             &odb::dbMaster::setType, master,
```
maliberty, line 43: When I started on this I did:

  auto make_lit = [master](const char* name, dbMasterType::Value type) {
    return lit(name)[boost::bind(&odb::dbMaster::setType, master, type)];
  };

to save on save on repetition

maliberty (changes requested):
As noted

src/odb/src/lefin/lefMacroClassParser.cpp
QuantamHD: I think I'm trying to avoid these cases for the moment. I don't think we have any PDKs with these types available yet. It's not an error and the user will be shown a warning. 

I think for now keeping it this way has the least chance of introducing hard to debug problems. i.e. we should only parse what we actually use.

src/odb/src/lefin/lefMacroPropParser.h
QuantamHD: 8 year old me really knew what was up.

src/odb/src/lefin/lefMacroClassParser.cpp
maliberty: I thinking of anything we already support through MACRO CLASS <> we should support through PROPERTY LEF58_CLASS "CLASS <>" as they are equivalent.

QuantamHD:
Done.

maliberty (approved):

PR #1980, opened 2022-06-22
grt: save_guides option

osamahammad21:


osamahammad21:
@maliberty @vvbandeira there seems to be a problem with the macos build. It has been unstable for a while with my pull requests.

PR #1983, opened 2022-06-22
Move to self-hosted-mac

vvbandeira:
Signed-off-by: Vitor Bandeira <vitor.vbandeira@gmail.com>


vvbandeira:
@maliberty FYI, after this merge (and people update their branches) I expect to not have issues with macOS. Also, faster turn-around with reduced build times.

PR #1998, opened 2022-06-26
drt: remove destroyed filler instances

osamahammad21:
[[OR-1499]](https://openroad.atlassian.net/browse/OR-1499)

### DIFF
--- a/src/drt/src/db/obj/frBlock.h
+++ b/src/drt/src/db/obj/frBlock.h
@@ -314,6 +314,21 @@ class frBlock : public frBlockObject
     name2inst_[in->getName()] = in.get();
     insts_.push_back(std::move(in));
   }
+  void removeInst(frInst* inst)
+  {
+    for (const auto& iterm : inst->getInstTerms()) {
+      auto net = iterm->getNet();
+      if (net != nullptr)
+        net->removeInstTerm(iterm.get());
+    }
+    name2inst_.erase(inst->getName());
+    for (auto itr = insts_.begin(); itr != insts_.end(); itr++) {
+      if (itr->get() == inst) {
+        insts_.erase(itr);
+        return;
+      }
+    }
+  }
   void addNet(std::unique_ptr<frNet> in)
   {
     in->setId(nets_.size());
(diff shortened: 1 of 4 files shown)


src/drt/src/db/obj/frBlock.h
```diff
@@ -314,6 +314,21 @@ class frBlock : public frBlockObject
...
+      if (net != nullptr)
+        net->removeInstTerm(iterm.get());
+    }
+    name2inst_.erase(inst->getName());
+    for (auto itr = insts_.begin(); itr != insts_.end(); itr++) {
+      if (itr->get() == inst) {
+        insts_.erase(itr);
+        return;
+      }
+    }
```
maliberty, line 330: This makes removal O(n) which can be a performance issue.  I don't see a good reason why we keep name2inst_ and insts_ separate.  If we fused them then you could look up the inst by name and then remove it in O(1).  

Also there is no good reason to copy the instance name - that just doubles the memory.

jjcherry56: It is actually N^2 because finding the iterator is O(N) and deleting the element is O(N).
But since all of the fillers are removed it is called N times so it is N^3.
It never makes no sense to use vectors when deletes are required.

jjcherry56: And this loop is just std::remove so there is no point in writing it out.

osamahammad21: @jjcherry56 I think the cost for deleting one element is O(n) not O(n^2) because finding the iterator is O(n) and performing the erase; which is done only once per deletion, is O(n). so it's O(n) + O(n) not O(n) * O(n).
Also check the updates as I now call erase with the inst index and I don't lookup the iterator. But I have to loop over the next elements to update their IDs

maliberty: TritonRoute was never written with netlist modification in mind so the use of vectors is pretty pervasive.  This is a good opportunity to upgrade it to a more flexible model.

osamahammad21: Actually Matt have suggested doing this a long time ago but there was never good time. Most of Tritonroute's db is a duplicate of odb's because it used to be a standalone tool.
I can take your suggestion for setting the ids before serialization as a fast solution for getting this PR merged soon. However, changes requested for the db itself could take more time if it's not high priority issue.

maliberty:
@jjcherry56 though we are fixing this it makes no sense to insert and then remove the filler cells.  You should just delay insertion until after all placement relate optimizations are done.

PR #2010, opened 2022-07-01
Ant repair

jjcherry56:


jjcherry56:
repair_antennas do not remove or allow fillers
check_antennas fix antenna ratio pwl interpolation
detailed_placement do not allow fillers
update Nangate45.lef to include TAPCELL_X1 (heavy okfile fallout)

PR #2014, opened 2022-07-01
Implements CORE_SPACER auto-detection

QuantamHD:
LEF5.8 page 116 specifies

```
SPACER—Sometimes called a filler cell, this cell is used to fill in space
between regular core cells. The SPACER sub-class needs to be cells with
no logic-pins. Thus even with the sub-class defined, a cell will not be
considered SPACER (also called FILLER) unless it has no logic/signal
pins. A filler can only have Power and Ground pins. The instances of
these cells will be marked by the insertion command to be of type
'Physical'.
```

This went unimplemented in OpenROAD. This PR implements the the
detection algorithm.

### DIFF
--- a/src/odb/include/odb/lefin.h
+++ b/src/odb/include/odb/lefin.h
@@ -124,7 +124,7 @@ class lefin
     else
       return (int) (value + 0.5);
   }
-
+  bool isSpacer(dbMaster* _master);
   bool readLef(const char* lef_file);
   bool addGeoms(dbObject* object, bool is_pin, lefiGeometries* geometry);
   void createLibrary();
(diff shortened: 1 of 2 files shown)


maliberty:
The LEF doesn't say any cell with only power/ground is a filler - only that those with other pins can't be fillers.

QuantamHD:
Then ASAP7 definitely needs a PR to add the correct subtypes to their types. Would you agree or is there a detection that's safe?

maliberty:
https://github.com/The-OpenROAD-Project/asap7/issues/10

I fixed the copies in ORFS

QuantamHD:
I'd rather we just fix it upstream.

PR #2033, opened 2022-07-06
drt: fix off-grid violations on first iterations

eder-matheus:

### DIFF
--- a/src/drt/src/dr/FlexDR_init.cpp
+++ b/src/drt/src/dr/FlexDR_init.cpp
@@ -3413,3 +3413,17 @@ void FlexDRWorker::init(const frDesign* design)
   setGCWorker(std::move(gcWorker));
   initMazeCost(design);
 }
+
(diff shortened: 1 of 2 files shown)


src/drt/src/dr/FlexDR_init.cpp
```diff
@@ -1019,12 +1019,26 @@ void FlexDRWorker::initNet_termGenAp_new(const frDesign* design, drPin* dPin)
             xLoc = *(xLocs.begin());
           } else {
             xLoc = (xl(pinRect) + xh(pinRect)) / 2;
+            frCoord manuGrid = getTech()->getManufacturingGrid();
```
maliberty, line 1022: There look to be two similar blocks in the frcBTerm section below.

maliberty: If you are going to have four copies of this it might be time to factor it into a helper function.

PR #2046, opened 2022-07-09
odb: support LEF58_METALWIDTHVIAMAP

osamahammad21:
https://openroad.atlassian.net/browse/OR-1520


src/odb/src/lefin/MetalWidthViaMapParser.cpp
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright (c) 2019, The Regents of the University of California
```
maliberty, line 2: 2022

src/odb/src/lefin/MetalWidthViaMapParser.cpp
```diff
@@ -0,0 +1,115 @@
...
+                      "parse mismatch in propery LEF58_METALWIDTHVIAMAP"
+                      ":\"{}\"",
+                      rule);
+  }
+}
+
+bool MetalWidthViaMapParser::parseSubRule(std::string s)
+{
+  cut_class_ = false;
+  qi::rule<std::string::iterator, std::string(), ascii::space_type> _string;
```
maliberty, line 94: Don't use leading (or trailing) underscore on local variable names.

maliberty (approved):

maliberty:
You need to update the schema index (that's the cause of your CI failure).  I've approved as the changes requested don't need rereview.

maliberty:
You also need
```
WRAP_DB_CONTAINER(odb::dbMetalWidthViaMap)
````
src/odb/src/swig/common/containers.i

src/odb/src/lefin/MetalWidthViaMapParser.cpp
osamahammad21: I am trying to match the boost spirit terminals (e.x. double_ int_). Is that ok? if not then it needs to be changed in all lefin.

osamahammad21: @maliberty I am gonna merge and we can address this in a separate PR if it's an issue.

maliberty: It can be fixed in a separate PR.  I see in some cases you used this style and other you used appended "Rule" to the name which I prefer (eg lefTechLayerCornerSpacingParser.cpp)

PR #2049, opened 2022-07-11
DRC cmd

osamahammad21:
This is a separate drc check command to get routing violations.

### DIFF
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -988,7 +988,7 @@ void TritonRoute::reportDRC(const string& file_name,
     for (const auto& marker : markers) {
       // get violation bbox
       Rect bbox = marker->getBBox();
-      if (drcBox != Rect(0, 0, 0, 0) && !drcBox.intersects(bbox))
+      if (drcBox != Rect() && !drcBox.intersects(bbox))
         continue;
       auto tech = getDesign()->getTech();
       auto layer = tech->getLayer(marker->getLayerNum());
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -245,7 +245,8 @@ void TritonRoute::debugSingleWorker(const std::string& dumpDir,
              worker->getBestNumMarkers(),
              updated);
   if (updated)
-    reportDRC(drcRpt, worker.get());
+    reportDRC(
+        drcRpt, design_->getTopBlock()->getMarkers(), worker->getDrcBox());
 }
 
 void TritonRoute::updateGlobals(const char* file_name)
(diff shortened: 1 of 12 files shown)


src/drt/src/TritonRoute.cpp
```diff
@@ -960,18 +983,17 @@ void TritonRoute::reportDRC(const string& file_name, FlexDRWorker* worker)
...
   }
-
   ofstream drcRpt(file_name.c_str());
   if (drcRpt.is_open()) {
-    for (auto& marker : getDesign()->getTopBlock()->getMarkers()) {
+    for (const auto& marker : markers) {
       // get violation bbox
       Rect bbox = marker->getBBox();
-      if (worker != nullptr && !worker->getDrcBox().intersects(bbox))
+      if (drcBox != Rect(0, 0, 0, 0) && !drcBox.intersects(bbox))
```
maliberty, line 991: drcBox !=Rect() is more concise

src/drt/src/TritonRoute.tcl
```diff
@@ -310,6 +310,30 @@ proc pin_access { args } {
   }
   drt::pin_access_cmd $db_process_node $bottom_routing_layer $top_routing_layer $verbose $min_access_points
 }
+sta::define_cmd_args "check_drc" {
+    [-drc_box box]
+    [-output_drc filename]
```
maliberty, line 315: You don't need to repeat drc in the option names.  -box & -output_file are fine.

Put the command in the drt namespace to make it private.

src/drt/src/gc/FlexGC_init.cpp
```diff
@@ -830,8 +926,11 @@ void FlexGCWorker::Impl::init(const frDesign* design)
   // ProfileTask profile("GC:init");
   addNet(design->getTopBlock()->getFakeVSSNet());  //[0] floating VSS
   addNet(design->getTopBlock()->getFakeVDDNet());  //[1] floating VDD
-  initDesign(design);
+  initDesign(design, true);
```
maliberty, line 929: Why is skipDR=true here?

src/drt/src/gc/FlexGC_init.cpp
```diff
@@ -355,6 +412,46 @@ void FlexGCWorker::Impl::initDRWorker()
...
 }
+void FlexGCWorker::Impl::initNetsFromDesign(const frDesign* design)
+{
+  int cnt = 0;
+  auto block = design->getTopBlock();
+  for (auto& net : block->getNets()) {
+    // always first generate gcnet in case owner does not have any object
+    auto it = owner2nets_.find(net.get());
+    if (it == owner2nets_.end()) {
+      addNet(net.get());
```
maliberty, line 423: Why add every net even if it has no shapes in the checking region?  Should this be delayed to initRouteObj?

osamahammad21: I don't think it has high cost anyway. In main, if the net doesn't have any shapes it will be skipped.

src/drt/src/gc/FlexGC_init.cpp
osamahammad21: This init function is only used by the DRWorker and the drc command. In both you shouldn't query dr objects in initDesign because:
- if it's used by FlexDRWorker, then the dr shapes should be handled in initDRWorker
- if it's used by drc command, then we shouldn't assume that there are any dr objects in the regionquery, and we handle the route objects in initNetsFromDesign
In case you are wondering why we need skipDR flag, it's because initDesign is used in initPA which is used in pin access.

src/drt/src/TritonRoute.tcl
osamahammad21: Why do you want it private?

maliberty: I don't want to create false expectations that this is a general purpose drc engine.  It is a sufficient engine to drive the router and no more (ie it doesn't check violations between fixed shapes).

osamahammad21: @maliberty  ok. done

maliberty (approved):

PR #2060, opened 2022-07-13
add layer range for pin_access and detailed_route

eder-matheus:

### DIFF
--- a/test/flow.tcl
+++ b/test/flow.tcl
@@ -190,7 +190,9 @@ write_db $grt_db
 ################################################################
 # Global routing
 
-pin_access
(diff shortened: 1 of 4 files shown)


test/flow.tcl
```diff
@@ -190,7 +190,9 @@ write_db $grt_db
 ################################################################
 # Global routing
 
-pin_access
+pin_access -bottom_routing_layer $min_routing_layer \
```
maliberty, line 193: You should also pass the range to the detailed_route command

maliberty (approved):

PR #2063, opened 2022-07-14
grt: update reductions calculation

eder-matheus:


src/grt/include/grt/GlobalRouter.h
```diff
@@ -48,6 +48,9 @@
 #include "odb/db.h"
 #include "odb/dbBlockCallBackObj.h"
 #include "sta/Liberty.hh"
+#include "boost/icl/interval.hpp"
+
+using boost::icl::interval;
```
maliberty, line 53: This doesn't need to be in the public header.

src/grt/src/Grid.cpp
```diff
@@ -136,6 +136,40 @@ void Grid::getBlockedTiles(const odb::Rect& obstruction,
...
+  if (direction == odb::dbTechLayerDir::VERTICAL) {
+    if (obs.xMin() >= tile.xMin() && obs.xMax() <= tile.xMax()) {
+      start_point = obs.xMin();
+      end_point = obs.xMax();
+    } else if (first) {
+      start_point = obs.xMin();
+      end_point = tile.xMax();
+    } else {
+      start_point = tile.xMin();
+      end_point = obs.xMax();
```
maliberty, line 155: In the first tile wouldn't you want to start from the tile.xMin?

src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -545,6 +556,15 @@ class FastRouteCore
 
   FastRouteRenderer* fastrouteRender_;
   std::unique_ptr<DebugSetting> debug_;
+
+  std::unordered_map<std::tuple<int, int, int>,
```
maliberty, line 560: Prefer a struct to a tuple as it is clear what the three ints represent from this declaration

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -538,11 +538,93 @@ void FastRouteCore::applyHorizontalAdjustments(const odb::Point& first_tile,
...
+    for (int y = first_tile.getY(); y < last_tile.getY(); y++) {
+      if (x == first_tile.getX()) {
+        vertical_blocked_intervals_[std::make_tuple(x, y, layer)]
+            += first_tile_reduce_interval;
+      } else if (x == last_tile.getX()) {
+        vertical_blocked_intervals_[std::make_tuple(x, y, layer)]
+            += last_tile_reduce_interval;
+      } else {
+        addAdjustment(x, y, x, y + 1, layer, 0, true);
+      }
```
maliberty, line 559: why are we only tracking intervals in the first & last tiles?

maliberty (changes requested):

src/grt/src/Grid.cpp
eder-matheus: Not really. If I understand correctly what Luis did, we want the interval blocked by the obs. The figure below shows why using the obs.xMin is correct:
image

src/grt/src/fastroute/src/FastRoute.cpp
eder-matheus: @luis201420 could you comment here?

luis201420: Because the other (middle) tiles I am reducing to 0 because they are crossed by the block completely.

src/grt/src/Grid.cpp
maliberty: got it

PR #2069, opened 2022-07-18
Rsz cleanup

jjcherry56:
cleanup mess left by 659e8fec1 Don't resize fixed drivers (e.g. IO pads)


maliberty:
Do you think rsz should resize std cells that are marked fixed?  I don't.

jjcherry56:
resizing has nothing to do with placement status. it is the wrong criteria.

maliberty:
> resizing has nothing to do with placement status. it is the wrong criteria.

What is the right criteria?

PR #2071, opened 2022-07-18
odb: typemap dbTechLayerCutSpacingTableDefRule::LOOKUP_STRATEGY

osamahammad21:


maliberty (approved):

arlpetergadfort:
> @arlpetergadfort is this the API you want?

I think. I won't have a chance to check if I can get the correct value from the table, but it will make it easier to test

PR #2076, opened 2022-07-20
CUDA version of GPL

RuijiaoSun:


QuantamHD:
@cdleary @hzeller would you mind reviewing Rachel's first PR?

src/gpl/CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
 
 find_package(Eigen3 REQUIRED)
 
+#enable_language(CUDA)
+
```
QuantamHD, line 48: This should be removed.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+  // Parameters that don't change with iteration
+  int m = pb_->placeInsts().size(); // number of rows of matrix A
+  float tol = 1e-6; // Threshold to decide if the 
+  int reorder = 0;  // "0" for common matrix meaning no extra processing
+  int singularity = -1; // Output. Showing if the sparse matrix is signular or not. -1 = A is invertible
+
   for(int i=1; i<=ipVars_.maxIter; i++) {
     updatePinInfo();
     createSparseMatrix();
+    if (GPU == 1){
```
QuantamHD, line 155: We should move the CUDA code into a seperate method that just takes the sparse values and returns a solution std::vector. Currently the CUDA code is very verbose in what was otherwise a short function.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+      CUDA_ERROR(cudaFree(d_fixedInstForceVecY_));
+      CUSPARSE_ERROR(cusparseDestroyMatDescr( descrX ) );
+      CUSPARSE_ERROR(cusparseDestroy(handleCusparseX));
+      CUSOLVER_ERROR(cusolverSpDestroy(handleCusolverX));  
+      CUSPARSE_ERROR(cusparseDestroyMatDescr( descrY ) );
+      CUSPARSE_ERROR(cusparseDestroy(handleCusparseY));
+      CUSOLVER_ERROR(cusolverSpDestroy(handleCusolverY));   
+
+      // std::cout << "GPU X: " << instLocVecX_[0] << ", " << instLocVecX_[1] << ", " << instLocVecX_[200] << std:: endl;
+      // std::cout << "GPU Y: " << instLocVecY_[0] << ", " << instLocVecY_[1] << ", " << instLocVecY_[200] << std:: endl;
```
QuantamHD, line 275: Remove the comments.

src/gpl/CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
 
 find_package(Eigen3 REQUIRED)
 
+#enable_language(CUDA)
```
hzeller, line 47: is this meant to be commented out here, or is this some leftover debugging change ?

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
 
   // set ExtId for idx reference // easy recovery
   setPlaceInstExtId();
+
+  // Parameters that don't change with iteration
+  int m = pb_->placeInsts().size(); // number of rows of matrix A
+  float tol = 1e-6; // Threshold to decide if the
```
cdleary, line 148: Comment not finished here?

src/gpl/src/initialPlace.cpp
```diff
@@ -41,6 +43,40 @@
...
 
 #include "utl/Logger.h"
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include "cusparse.h"
+#include "cusolverDn.h"
+#include "cusolverSp.h"
+
+#define CUDA_ERROR(value) do { \
+   cudaError_t m_cudaStat = value; \
```
hzeller, line 53: What does the `m` prefix signify here ?

Usually, `m_xxx` is often used as an indication for 'member variable', but looks like this is more a temporary value here.

src/gpl/src/initialPlace.cpp
```diff
@@ -313,6 +482,26 @@ void InitialPlace::createSparseMatrix() {
...
+            // cooColIndexX.push_back(inst2);
+            // cooValX.push_back(weightX);
+
+            // cooRowIndexX.push_back(inst1);
+            // cooColIndexX.push_back(inst2);
+            // cooValX.push_back(-weightX);
+            // cooRowIndexX.push_back(inst2);
+            // cooColIndexX.push_back(inst1);
+            // cooValX.push_back(-weightX);
+
```
cdleary, line 504: Comments left in?

src/gpl/src/initialPlace.cpp
```diff
@@ -41,6 +43,40 @@
...
+#define CUDA_ERROR(value) do { \
+   cudaError_t m_cudaStat = value; \
+   if(m_cudaStat != cudaSuccess){ \
+       fprintf(stderr, "Error %s at line %d in file %s\n", cudaGetErrorString(m_cudaStat), __LINE__, __FILE__); \
+       exit(-1); \
+   } \
+} while(0)
+#define CUSPARSE_ERROR(value) do { \
+    cusparseStatus_t _m_status = value;\
```
hzeller, line 60: consistency: above, you use `m_xxx`, here you use `_m_xxx`. Picking one consistent naming scheme makes it easier for future readers of the code.

src/gpl/src/initialPlace.cpp
```diff
@@ -405,9 +643,22 @@ void InitialPlace::createSparseMatrix() {
...
+  if (outFile){
+    for(int i = 0; i < placeInstForceMatrixX_.rows(); i++){
+      outFile << placeInstForceMatrixX_.row(i) << std::endl;
+      outFile << ", " << std::endl;
+      // outFile << listX[i] << std::endl;
+      // outFile << listX[i] << std::endl;
+
+    }
+  }
+  outFile.close();
```
cdleary, line 661: Debugging code left in? I imagine we'd want to remove this before merging.

src/gpl/src/initialPlace.h
```diff
@@ -42,6 +42,11 @@ namespace utl {
 class Logger;
 }
 
+struct tri{
```
cdleary, line 45: Good to have a solid comment on the struct definition for what it's representing / how it's generally used. Looks like maybe the convention in this code base is to use a leading cap, e.g. `Tri` ?

src/gpl/src/initialPlace.h
```diff
@@ -102,6 +106,9 @@ class InitialPlace {
     Eigen::VectorXf instLocVecX_, fixedInstForceVecX_;
     Eigen::VectorXf instLocVecY_, fixedInstForceVecY_;
     SMatrix placeInstForceMatrixX_, placeInstForceMatrixY_;
+    bool GPU = 1;
```
cdleary, line 109: Style nit: in C++ we generally prefer "true" / "false" keywords to 1/0 when the value is boolean, and avoid explicit comparisons to `== 1` preferring just `if (myBoolean)` and such.

src/gpl/src/initialPlace.cpp
```diff
@@ -41,6 +43,40 @@
...
 #include "utl/Logger.h"
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include "cusparse.h"
+#include "cusolverDn.h"
+#include "cusolverSp.h"
+
+#define CUDA_ERROR(value) do { \
+   cudaError_t m_cudaStat = value; \
+   if(m_cudaStat != cudaSuccess){ \
```
hzeller, line 54: Style: below, you use a space between `if` and the opening parenthesis, here not. Best is to adapt it to the style that is already in the file and use that consistently.

test/flow.tcl
```diff
@@ -83,6 +83,8 @@ set_routing_layers -signal $global_routing_layers \
   -clock $global_routing_clock_layers
 set_macro_extension 2
 
+write_db /tmp/gcd_sky130hd.db
```
cdleary, line 86: Do we intend to leave this in?

src/gpl/src/initialPlace.cpp
cdleary, line 155: Perhaps for a followup PR: it'll be good to turn this into its own function or method, because the GPU code makes the method grow quite large (and it gets hard to track the control flow with big methods). Having the function separated out would also potentially let us put the implementation in a separate "GPU routines" file.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
 
   // set ExtId for idx reference // easy recovery
   setPlaceInstExtId();
+
+  // Parameters that don't change with iteration
+  int m = pb_->placeInsts().size(); // number of rows of matrix A
```
hzeller, line 147: If things dont' change, suggst to make them `const`. This makes it easier for a reader to follow the code as it means one thing less to keep track of.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+      CUSPARSE_ERROR(cusparseXcoo2csr(handleCusparseX, d_cooRowIndexX, nnz_x, m, d_csrRowIndX, CUSPARSE_INDEX_BASE_ZERO));
+      CUSPARSE_ERROR(cusparseXcoo2csr(handleCusparseY, d_cooRowIndexY, nnz_y, m, d_csrRowIndY, CUSPARSE_INDEX_BASE_ZERO));
+      CUSOLVER_ERROR(cusolverSpScsrlsvqr(handleCusolverX, m, nnz_x, descrX, d_cooValX, d_csrRowIndX, d_cooColIndexX, d_fixedInstForceVecX_, tol, reorder, d_instLocVecX, &singularity));
+      CUSOLVER_ERROR(cusolverSpScsrlsvqr(handleCusolverY, m, nnz_y, descrY, d_cooValY, d_csrRowIndY, d_cooColIndexY, d_fixedInstForceVecY_, tol, reorder, d_instLocVecY, &singularity));
+      
+      // Sync and Copy data to host
+      CUDA_ERROR(cudaMemcpyAsync(instLocVecX_.data(), d_instLocVecX, sizeof(float)*m, cudaMemcpyDeviceToHost, streamX));
+      CUDA_ERROR(cudaMemcpyAsync(instLocVecY_.data(), d_instLocVecY, sizeof(float)*m, cudaMemcpyDeviceToHost, streamY));
+
+      // Destroy what is not needed in both of device and host
```
cdleary, line 254: Just want to check -- this mentions destroying what is not needed, all of the on-device constructs will be allocated and freed every iteration, correct?

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+      
+      solver.compute(placeInstForceMatrixY_);
+      instLocVecY_ = solver.solveWithGuess(fixedInstForceVecY_, instLocVecY_);
+      errorY = solver.error();
 
-    solver.compute(placeInstForceMatrixY_);
-    instLocVecY_ = solver.solveWithGuess(fixedInstForceVecY_, instLocVecY_);
-    errorY = solver.error();
+      log_->report("[InitialPlace]  Iter: {} CG residual: {:0.8f} HPWL: {}",
+        i, max(errorX, errorY), pb_->hpwl());
```
cdleary, line 291: This looks like possibly useful status info, is it possible for us to log it using the output of the GPU routine so somebody could e.g. run with GPU and without GPU and compare the status output side-by-side in the logs?

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
   // set ExtId for idx reference // easy recovery
   setPlaceInstExtId();
+
+  // Parameters that don't change with iteration
+  int m = pb_->placeInsts().size(); // number of rows of matrix A
+  float tol = 1e-6; // Threshold to decide if the 
+  int reorder = 0;  // "0" for common matrix meaning no extra processing
+  int singularity = -1; // Output. Showing if the sparse matrix is signular or not. -1 = A is invertible
+
   for(int i=1; i<=ipVars_.maxIter; i++) {
```
hzeller, line 152: Here is an `i` loop variable and then below another one. Whlie it will work (the other one will shadow this `i`), it might be confusing.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
     createSparseMatrix();
+    if (GPU == 1){
+      // Set sparse matrices from dense matrices.
+      cooRowIndexX.clear();
+      cooColIndexX.clear();
+      cooValX.clear();
+      cooRowIndexY.clear();
+      cooColIndexY.clear();
+      cooValY.clear();
+      for (size_t i = 0; i < placeInstForceMatrixX_.rows(); i ++){
```
hzeller, line 163: See above, we already have an `i` here.

How about we call these index variables `row` and `column` ? That way it is easier to follow.

CMakeLists.txt
```diff
@@ -13,7 +13,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-cmake_minimum_required (VERSION 3.9)
+cmake_minimum_required (VERSION 3.12)
```
cdleary, line 16: Will it be ok for the OpenROAD project that we increase the CMake required version?

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+        for (size_t j = 0; j < placeInstForceMatrixX_.cols(); j ++){
+          if (placeInstForceMatrixX_.coeffRef(i,j) != 0){
```
hzeller, line 165: I'd always keep a space after a comma, so `coeffRef(i, j)` ... or with the new naming `coeffRef(row, column)`

src/gpl/src/initialPlace.cpp
```diff
@@ -41,6 +43,40 @@
...
+#include <cuda_runtime.h>
+#include "cusparse.h"
+#include "cusolverDn.h"
+#include "cusolverSp.h"
+
+#define CUDA_ERROR(value) do { \
+   cudaError_t m_cudaStat = value; \
+   if(m_cudaStat != cudaSuccess){ \
+       fprintf(stderr, "Error %s at line %d in file %s\n", cudaGetErrorString(m_cudaStat), __LINE__, __FILE__); \
+       exit(-1); \
```
cdleary, line 56: Just to check we're following the expected convention: does the C++ code in this part of the code base generally exit on failures that we can't recover from? Are particular status codes expected? (e.g. how did we choose -5 below vs -1 here)?

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+        }
+      }
 
-    // BiCGSTAB solver for initial place
-    BiCGSTAB< SMatrix, IdentityPreconditioner > solver;
-    solver.setMaxIterations(ipVars_.maxSolverIter);
-    solver.compute(placeInstForceMatrixX_);
-    instLocVecX_ = solver.solveWithGuess(fixedInstForceVecX_, instLocVecX_);
-    errorX = solver.error();
+      int nnz_x = cooRowIndexX.size(); // number of non-zeros for matrix X
```
hzeller, line 178: Do they change ? If not, make them `const`.

hzeller (commented):
A first set of comments, mostly stylistic.

cdleary (commented):
Whoohoo, accelerate all the things! :-) Made a few comments, I know it's still a draft, but wanted to give a few things that popped out at a first glance, will try to do a deeper look at the CUDA APIs we're calling in the next pass.

CMakeLists.txt
maliberty: What is this change is driving the need to upgrade cmake?

CMakeLists.txt
```diff
@@ -13,7 +13,9 @@
...
+
+enable_language(CUDA)
```
maliberty, line 18: I don't see anything in this PR that makes CUDA optional as we discussed.

src/gpl/src/initialPlace.cpp
maliberty, line 56: All messages should go through the logger.  These should be error() calls which will generate an exception.  exit() is never used as a form of error handling.

src/gpl/src/initialPlace.cpp
```diff
@@ -41,6 +43,40 @@
...
+    if (_m_status != CUSPARSE_STATUS_SUCCESS){\
+        fprintf(stderr, "Error %d at line %d in file %s\n", (int)_m_status, __LINE__, __FILE__);\
+    exit(-5); \
+    } \
+} while(0)
+
+#define CUSOLVER_ERROR(value) do { \
+    cusolverStatus_t _m_status = value; \
+    if (_m_status != CUSOLVER_STATUS_SUCCESS){ \
+        fprintf(stderr, "Error %d at line %d in file %s\n", (int)_m_status, __LINE__, __FILE__);\
```
maliberty, line 70: Error messages should be meaningful to end users which this is not.

src/gpl/src/initialPlace.cpp
```diff
@@ -41,6 +43,40 @@
...
+#include "cusolverSp.h"
+
+#define CUDA_ERROR(value) do { \
+   cudaError_t m_cudaStat = value; \
+   if(m_cudaStat != cudaSuccess){ \
+       fprintf(stderr, "Error %s at line %d in file %s\n", cudaGetErrorString(m_cudaStat), __LINE__, __FILE__); \
+       exit(-1); \
+   } \
+} while(0)
+#define CUSPARSE_ERROR(value) do { \
```
maliberty, line 59: Avoid macros

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+      CUDA_ERROR(cudaMalloc((void**)&d_cooRowIndexX, nnz_x * sizeof(int)));
+      CUDA_ERROR(cudaMalloc((void**)&d_cooColIndexX, nnz_x * sizeof(int)));
+      CUDA_ERROR(cudaMalloc((void**)&d_cooValX, nnz_x * sizeof(float)));
+      CUDA_ERROR(cudaMalloc((void**)&d_fixedInstForceVecX_, m * sizeof(float)));
+      CUDA_ERROR(cudaMalloc((void**)&d_instLocVecX, m * sizeof(float)));
+
+      CUDA_ERROR(cudaMalloc((void**)&d_cooRowIndexY, nnz_y * sizeof(int)));
+      CUDA_ERROR(cudaMalloc((void**)&d_cooColIndexY, nnz_y * sizeof(int)));
+      CUDA_ERROR(cudaMalloc((void**)&d_cooValY, nnz_y * sizeof(float)));
+      CUDA_ERROR(cudaMalloc((void**)&d_fixedInstForceVecY_, m * sizeof(float)));
```
maliberty, line 193: Any failures in this will lead to memory leaks if not inconsistent state.  This is very old C style programming.  If the cuda library forces this then you should be safe abstractions on top of it or find a better library.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
-    solver.compute(placeInstForceMatrixY_);
-    instLocVecY_ = solver.solveWithGuess(fixedInstForceVecY_, instLocVecY_);
-    errorY = solver.error();
+      log_->report("[InitialPlace]  Iter: {} CG residual: {:0.8f} HPWL: {}",
+        i, max(errorX, errorY), pb_->hpwl());
 
-    log_->report("[InitialPlace]  Iter: {} CG residual: {:0.8f} HPWL: {}",
-       i, max(errorX, errorY), pb_->hpwl());
+      // std::cout << "CPU X: " << instLocVecX_[0] << ", " << instLocVecX_[1] << ", " << instLocVecX_[200] << std:: endl;
+      // std::cout << "CPU Y: " << instLocVecY_[0] << ", " << instLocVecY_[1] << ", " << instLocVecY_[200] << std:: endl;
```
maliberty, line 294: Remove dead code

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +142,157 @@ void InitialPlace::doBicgstabPlace() {
...
+      errorY = solver.error();
 
-    solver.compute(placeInstForceMatrixY_);
-    instLocVecY_ = solver.solveWithGuess(fixedInstForceVecY_, instLocVecY_);
-    errorY = solver.error();
+      log_->report("[InitialPlace]  Iter: {} CG residual: {:0.8f} HPWL: {}",
+        i, max(errorX, errorY), pb_->hpwl());
 
-    log_->report("[InitialPlace]  Iter: {} CG residual: {:0.8f} HPWL: {}",
-       i, max(errorX, errorY), pb_->hpwl());
```
maliberty, line 125: Why are you removing the existing messages?

src/gpl/src/initialPlace.h
maliberty: Use meaningful names

maliberty (changes requested):

maliberty:
You should build your code on a machine without cuda installed and make sure there is no issue

src/gpl/CMakeLists.txt
RuijiaoSun: This is the leftover change, I have deleted this line.

src/gpl/src/initialPlace.cpp
RuijiaoSun: Good suggestion! Helped me with the future coding.

src/gpl/src/initialPlace.cpp
RuijiaoSun: Yeah... Finished now.

src/gpl/src/initialPlace.cpp
RuijiaoSun: Have deleted all the leftover changes.

src/gpl/src/initialPlace.h
RuijiaoSun: I can delete this structure actually .... I didn't use it in the final result.

test/flow.tcl
RuijiaoSun: For me, this line is super important. But I didn't go over the whole process and don't know if it is necessary for the whole project. I will delete it. But I will take care of this line and add it again if I find it useful later.

src/gpl/src/initialPlace.cpp
RuijiaoSun: Yes.

src/gpl/src/initialPlace.cpp
RuijiaoSun: Ha You found it. It should be outside. I just didn't find a good function to record the errorX and errorY for the GPU. I will look for a good method to record it for GPU version.

CMakeLists.txt
RuijiaoSun: Cmake 3.9 should be able to support CUDA libraries well. I changed it because the lowest version I have used is 3.12 not 3.9. I will change it back.

src/gpl/src/initialPlace.cpp
RuijiaoSun: If the status is not "Success", the code will exit. But the number I picked here is random. I have changed them all to "-1" so that people know it is just a general failure.

src/gpl/src/initialPlace.cpp
RuijiaoSun: So right. Chris mentioned this as well. I can add a small function outside storing CUDA related stuff. I will use a few lines to call it in this function.

CMakeLists.txt
RuijiaoSun: Yeah. I haven't accomplished this function yet. This version is just hard coded the bool value in the header file as GPU = 1. Will change it soon!

src/gpl/src/initialPlace.cpp
RuijiaoSun: I didn't remove this. It is in line 287 and line 288. 
But it is not perfect now. I am trying to get a similar errorX and errorY calculation method for the GPU. Then we will see the  difference between CPU and GPU.

src/gpl/src/initialPlace.cpp
maliberty: Do not use exit with any code - please see my comment about the logger.

src/gpl/src/initialPlace.cpp
RuijiaoSun: I see. From my knowledge, this is the only way. If I want to secure the command, what I can think of is to "Free all CUDA variables" once the program gets an unexpected problem and exits. I will also discuss with Chris about possible solution for it.

src/gpl/src/initialPlace.cpp
RuijiaoSun: This is what I always do in CUDA coding. Without macros, it is hard to detect the status of CUDA or CUSPARSE. CUDA's principle is: The code never stopped even if it has an error. It just returns a wrong status.  Maybe I have limited knowledge. I will bring this point and talk with Chris about this.

src/gpl/src/initialPlace.cpp
RuijiaoSun: This macro just helps the code to stop and show where the error is. Or the code will continue to run and return a bunch of zeros. I think the better way is to keep the macro and print where the error is. Meantime, I will put more detailed error message into log file. Is this a good solution?

src/gpl/src/initialPlace.cpp
QuantamHD: @RuijiaoSun the alternative to the macro is turning them into c++ functions which would be more well received.

src/gpl/src/initialPlace.cpp
RuijiaoSun: I use the relative residual to output error information, the same information as the CPU gave.

src/gpl/src/initialPlace.cpp
RuijiaoSun: I also add another function to check the error for the GPU version.

QuantamHD:
@maliberty You need to approve the scan workflow.

maliberty:
> @maliberty You need to approve the scan workflow since Rachel's a first time contributor.

Unfortunately I have to approve it on every push as she is still a first time contributor.  I just did so again.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +109,56 @@ void InitialPlace::doBicgstabPlace() {
...
-    errorX = solver.error();
+  for(size_t iter=1; iter<=ipVars_.maxIter; iter++) {
 
-    solver.compute(placeInstForceMatrixY_);
-    instLocVecY_ = solver.solveWithGuess(fixedInstForceVecY_, instLocVecY_);
-    errorY = solver.error();
+    updatePinInfo();
+    createSparseMatrix();
+    #if CUDA
+      // Set sparse matrices from dense matrices.
```
QuantamHD, line 118: You might want to break these two #ifs into two different methods so it looks more like this

#if CUDA
  cudaSparseSolve(...);
#else
 cpuSparseSolve(...)
#endif

src/gpl/src/cudasplibs.cu
```diff
@@ -0,0 +1,119 @@
...
+        cudaDeviceReset();
+    }
+}
+
+cudasplibs::cudasplibs(std::vector<int>& cooRowIndex, std::vector<int>& cooColIndex, std::vector<float>& cooVal, Eigen::VectorXf& fixedInstForceVec, utl::Logger* logger){
+    m = fixedInstForceVec.size();
+    nnz = cooVal.size();
+    log_ = logger;
+    // Allocate device memeory and copy data to device
+    cudaerror(cudaMalloc((void**)&d_cooRowIndex, nnz * sizeof(int)));
```
maliberty, line 29: Same comment as last time - you should be using RAII classes not low level management.

src/gpl/src/cudasplibs.cu
```diff
@@ -0,0 +1,119 @@
...
+        sum += (d_fixedInstForceVec[row] > 0) ? d_fixedInstForceVec[row] : -d_fixedInstForceVec[row];
+        if (d_fixedInstForceVec[row] > d_Ax[row])
+            error += d_fixedInstForceVec[row] - d_Ax[row];
+        else
+            error -= d_fixedInstForceVec[row]- d_Ax[row];
+    }
+    if (sum != 0) error = error / sum;
+}
+
+float cudasplibs::error_cal(){
```
maliberty, line 98: Can .cu files be clang-formatted?

src/gpl/src/cudasplibs.cu
```diff
@@ -0,0 +1,119 @@
...
+    cudaerror(cudaMemcpy(d_cooRowIndex, cooRowIndex.data(), sizeof(int)*nnz, cudaMemcpyHostToDevice));
+    cudaerror(cudaMemcpy(d_cooColIndex, cooColIndex.data(), sizeof(int)*nnz, cudaMemcpyHostToDevice));
+    cudaerror(cudaMemcpy(d_cooVal, cooVal.data(), sizeof(float)*nnz, cudaMemcpyHostToDevice));
+    cudaerror(cudaMemcpy(d_fixedInstForceVec, fixedInstForceVec.data(), sizeof(float)*m, cudaMemcpyHostToDevice));
+}
+
+void cudasplibs::cusolverSpQR(Eigen::VectorXf& instLocVec){
+    // Parameters that don't change with iteration and used in the CUDA code
+    float tol = 1e-6; // 	Tolerance to decide if singular or not.
+    int reorder = 0;  // "0" for common matrix without ordering
```
maliberty, line 45: const?

src/gpl/src/cudasplibs.h
```diff
@@ -0,0 +1,42 @@
...
+#include <thrust/sequence.h>
+ #include "device_launch_parameters.h"
+#include <vector>
+#include <Eigen/SparseCore>
+
+namespace utl {
+class Logger;
+}
+
+class cudasplibs{
```
maliberty, line 21: Improper class naming https://google.github.io/styleguide/cppguide.html#Type_Names

Unclear class name

src/gpl/src/cudasplibs.h
```diff
@@ -0,0 +1,42 @@
...
+private:
+int m;      // Rows of the SP matrix
```
maliberty, line 23: clang-format

src/gpl/src/cudasplibs.h
```diff
@@ -0,0 +1,42 @@
...
+ #include "device_launch_parameters.h"
+#include <vector>
+#include <Eigen/SparseCore>
+
+namespace utl {
+class Logger;
+}
+
+class cudasplibs{
+private:
```
maliberty, line 22: Public before private https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/gpl/src/cudasplibs.h
```diff
@@ -0,0 +1,42 @@
...
+utl::Logger* log_;
+
+void cudaerror(cudaError_t code);
+void cusparseerror(cusparseStatus_t code);
+void cusolvererror(cusolverStatus_t code);
+
+public:
+
+int *d_cooRowIndex, *d_cooColIndex;
+float *d_cooVal, *d_instLocVec, *d_fixedInstForceVec;
```
maliberty, line 34: Do not have public data members

Members need a trailing _

src/gpl/src/initialPlace.cpp
```diff
@@ -106,23 +109,56 @@ void InitialPlace::doBicgstabPlace() {
...
-    instLocVecX_ = solver.solveWithGuess(fixedInstForceVecX_, instLocVecX_);
-    errorX = solver.error();
+  for(size_t iter=1; iter<=ipVars_.maxIter; iter++) {
 
-    solver.compute(placeInstForceMatrixY_);
-    instLocVecY_ = solver.solveWithGuess(fixedInstForceVecY_, instLocVecY_);
-    errorY = solver.error();
+    updatePinInfo();
+    createSparseMatrix();
+    #if CUDA
```
maliberty, line 117: Even if you compile with CUDA the user should still have a runtime option to use the cpu instead (at a minimum for testing).

maliberty (changes requested):

src/gpl/src/initialPlace.h
```diff
@@ -102,6 +101,8 @@ class InitialPlace {
     Eigen::VectorXf instLocVecX_, fixedInstForceVecX_;
     Eigen::VectorXf instLocVecY_, fixedInstForceVecY_;
     SMatrix placeInstForceMatrixX_, placeInstForceMatrixY_;
+    std::vector<int> cooRowIndexX, cooColIndexX, cooRowIndexY, cooColIndexY;
+    std::vector<float> cooValX, cooValY;
```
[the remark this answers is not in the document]
RuijiaoSun: Added. coo is a storage method for sparse matrix. (https://en.wikipedia.org/wiki/Sparse_matrix#Coordinate_list_(COO) )

src/gpl/src/cudasplibs.h
RuijiaoSun: This is so easy and useful!

src/gpl/src/initialPlace.cpp
RuijiaoSun: Is the current version okay or is it better to clear them more to one or two lines like "cuda**" and "cpu**"?
(I sent the screenshot to the chat)

maliberty: Include the screenshot here so we can all see

RuijiaoSun: Screenshot 2022-07-27 3 52 15 PM

RuijiaoSun: Sorry I didn't know I can send pictures here.

src/gpl/src/initialPlace.h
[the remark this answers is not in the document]
RuijiaoSun: Did!

src/gpl/src/initialPlace.cpp
maliberty: You'll have to get rid of the #if to make it runtime switchable (see my other feedback).  So I think you should split it into two different functions.

RuijiaoSun: Thanks for clarifying it. I think you are mentioning a different problem with Ethan. For your suggestion, I will need to set an option in .tcl file so that user can input either "% global_place -density ***" or "% global_place -gpu -density ***" to control if they can choose to use CUDA or not. Am I understanding right?

maliberty: #if is a compile time choice so how will you implement the tcl option?

RuijiaoSun: Oh! I am not sure if I understood your meaning this time. I can wrap the two code blocks in the pictures as two functions. Then I can use tcl option to control if I will GPU code or not. This means that, if the user wants to use the CPU version with GPU enabled, they just enable GPU in CMake and choose to use CPU code from the .tcl option. Right?

RuijiaoSun: image
The structur would be like this?

maliberty: Yes the user should be able to choose gpu or cpu in any case.  If they choose gpu and it wasn't enabled then  give a warning and automatically switch to CPU.

RuijiaoSun: This is a good point. I should add this as well. Thanks!

RuijiaoSun: Hi Matt, I submitted a new version, in which I modified the CMakelists and added a flag "-force_cpu" via the .tcl command line. 
In this way, if GPU is enabled by CMake, the program uses CUDA solver by default and uses CPU solver if the "force_cpu" is set. If there is no available GPU, it turns into the CPU solver. 
If GPU is not enabled, the program uses CPU solver all the time.
image

src/gpl/src/cudasplibs.cu
RuijiaoSun: They will be used as "float" and "int".  If I define them as const here, I will still need to force them to be "float" and "int" again. So Which is better, define "float" directly or define "const float" first and change it to "float" later?

src/gpl/src/cudasplibs.cu
RuijiaoSun: I used clang-format for all files I added.

src/gpl/src/cudasplibs.cu
RuijiaoSun: CUDA thrust is being used instead of the Malloc. The vectors allocated by thrust can be released automatically when it is out of scope.

src/gpl/src/cudasplibs.cu
maliberty: clearly you didn't here: "cudasplibs::error_cal(){" is not formatted.

src/gpl/src/cudasplibs.cu
maliberty: why will you need to force them to be non-const later?  I don't see any non-const usage

RuijiaoSun: The CUSOLVER function doesn't accept const.

image

src/gpl/src/cudasplibs.cu
RuijiaoSun: Sorry I will do it and check every place else.

src/gpl/src/cudasplibs.cu
maliberty: primitive types are pass by value so it will accept a const as it is just going to copy it

src/gpl/src/cudasplibs.cu
RuijiaoSun: Oh. I did the wrong explanation before. Are you checking with the newest version? I have deleted "cudasplibs.cu". I am using "solver.h" and "gpuSolver.h" currently.

src/gpl/CMakeLists.txt
```diff
@@ -85,6 +85,32 @@ target_link_libraries(gpl
     rsz
     grt
 )
+# If CUDA is not installed there will not be cmake
+# support for the package so this needs to be "quiet".
+find_package(CUDA QUIET COMPONENTS Core Widgets)
```
maliberty, line 90: This is marked deprecated (see https://cmake.org/cmake/help/latest/module/FindCUDA.html). 

See also https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html

src/gpl/src/gpuSolver.h
```diff
@@ -0,0 +1,63 @@
+#ifndef __GPUSOLVER__
+#define __GPUSOLVER__
+
+#include <cuda_runtime.h>
```
maliberty, line 4: Does this compile if cuda isn't installled?

src/gpl/src/gpuSolver.h
```diff
@@ -0,0 +1,63 @@
...
+#include "odb/db.h"
+#include "utl/Logger.h"
+
+namespace utl {
+class Logger;
+}
+
+namespace gpl {
+
+using namespace std;
```
maliberty, line 27: Don't put this in a header which affects all users.  I don't see any std usage here at all.

src/gpl/src/replace.tcl
```diff
@@ -37,6 +37,7 @@ sta::define_cmd_args "global_placement" {\
     [-timing_driven]\
     [-routability_driven]\
     [-incremental]\
+    [-force_gpu_acceleration]\
```
maliberty, line 40: Why do we have this and -gpu ?

src/gpl/src/replace.tcl
```diff
@@ -83,7 +85,8 @@ proc global_placement { args } {
       -disable_timing_driven \
       -disable_routability_driven \
       -skip_io \
-      -incremental}
+      -incremental\
+      -force_cpu}
```
maliberty, line 89: This doesn't match define_cmd_args above

src/gpl/src/solver.h
```diff
@@ -0,0 +1,85 @@
...
+                     Eigen::VectorXf& instLocVecX,
+                     SMatrix& placeInstForceMatrixY,
+                     Eigen::VectorXf& fixedInstForceVecY,
+                     Eigen::VectorXf& instLocVecY,
+                     float errorX,
+                     float errorY,
+                     utl::Logger* logger,
+                     int hpwl)
+{
+  class GpuSolver SP1(placeInstForceMatrixX, fixedInstForceVecX, logger);
```
maliberty, line 43: Why is the keyword class here?  same for SP2

src/gpl/src/gpuSolver.h
RuijiaoSun: Yes. If there is no cuda installed, gpuSolver.h won't be compiled. The control sentence is written in solver.h.

src/gpl/src/replace.tcl
RuijiaoSun: Sorry. I didn't clear it completely. We have changed this to "force_cpu".

src/gpl/src/solver.h
RuijiaoSun: It is a bad typo... I didn't see any errors here. So didn't notice there exists a weird "class". Deleted it now.

src/gpl/src/gpuSolver.h
```diff
@@ -0,0 +1,63 @@
...
+  // {d_cooRowIndex_, d_cooColIndex_, d_cooVal_} are the device vectors used to
+  // store the COO formatted triplets.
+  // https://en.wikipedia.org/wiki/Sparse_matrix#Coordinate_list_(COO)
+  // d_instLocVec_ and d_fixedInstForceVec_ are the device lists corresponding
+  // to instLocVec and fixedInstForceVec.
+  thrust::device_vector<int> d_cooRowIndex_, d_cooColIndex_;
+  thrust::device_vector<float> d_cooVal_, d_fixedInstForceVec_, d_instLocVec_;
+
+  // {r_cooRowIndex_, r_cooColIndex_, r_cooVal_} are the raw pointing to the
+  // device ectors above.
```
maliberty, line 54: typo ectors

src/gpl/src/gpuSolver.h
```diff
@@ -0,0 +1,63 @@
...
+
+  // {d_cooRowIndex_, d_cooColIndex_, d_cooVal_} are the device vectors used to
+  // store the COO formatted triplets.
+  // https://en.wikipedia.org/wiki/Sparse_matrix#Coordinate_list_(COO)
+  // d_instLocVec_ and d_fixedInstForceVec_ are the device lists corresponding
+  // to instLocVec and fixedInstForceVec.
+  thrust::device_vector<int> d_cooRowIndex_, d_cooColIndex_;
+  thrust::device_vector<float> d_cooVal_, d_fixedInstForceVec_, d_instLocVec_;
+
+  // {r_cooRowIndex_, r_cooColIndex_, r_cooVal_} are the raw pointing to the
```
maliberty, line 53: pointing -> pointers

src/gpl/CMakeLists.txt
RuijiaoSun: Hi Matt, this is a bit weird. I deleted find_package() then there appeared many many CUDA errors. Once I added "find_pakage()" back, it compiles successfully again. Some websites said some third-parties still depend on "find_package()". I will research more on this problem.

maliberty:
Do you have any measurements of the benefit of this change?

RuijiaoSun:
>

RuijiaoSun:
> Do you have any measurements of the benefit of this change?

Am doing it.

RuijiaoSun:
Hi Matt, I tested the code on my working computer and another cloudtop without a GPU. It can work on both of the computers. But here are some not successful checks. Do you have any ideas on this?https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-Public/detail/PR-2076-head/14/pipeline 
image

maliberty:
Have you looked at gpl incremental02?  You can click the details link to see the Jenkins results.

src/gpl/src/solver.h
```diff
@@ -0,0 +1,84 @@
...
+  errorY = std::max(errorX, SP2.error_cal());
+  logger->report("[InitialPlace]  Iter: {} CG residual: {:0.8f} HPWL: {}",
+                 iter,
+                 errorY,
+                 pb->hpwl());
+  return errorY;
+}
+#endif
+
+float cpuSparseSolve(int maxSolverIter,
```
QuantamHD, line 56: Instead of a float let's return a struct with both errors

src/gpl/src/solver.h
```diff
@@ -0,0 +1,84 @@
...
+  float errorX, errorY;
+  BiCGSTAB<SMatrix, IdentityPreconditioner> solver;
+  solver.setMaxIterations(maxSolverIter);
+  solver.compute(placeInstForceMatrixX);
+  instLocVecX = solver.solveWithGuess(fixedInstForceVecX, instLocVecX);
+  errorX = solver.error();
+
+  solver.compute(placeInstForceMatrixY);
+  instLocVecY = solver.solveWithGuess(fixedInstForceVecY, instLocVecY);
+  errorY = std::max(errorX, solver.error());
```
QuantamHD, line 76: Let's move the max calculation back to the placer class.

src/gpl/src/solver.h
```diff
@@ -0,0 +1,84 @@
...
+  logger->report("[InitialPlace]  Iter: {} CG residual: {:0.8f} HPWL: {}",
```
QuantamHD, line 77: Let's make this log message a function call for both the cpu and gpu

QuantamHD (changes requested):

src/gpl/src/solver.h
RuijiaoSun: I put the log message in the initialplace.cpp. It is not a function call but also appears once.

src/gpl/src/solver.h
```diff
@@ -0,0 +1,79 @@
...
+#include "plot.h"
+#include "utl/Logger.h"
+
+namespace utl {
+class Logger;
+}
+
+namespace gpl {
+
+struct err{
```
QuantamHD, line 23: Let's name this something like ResidualError instead of just error

QuantamHD (changes requested):

RuijiaoSun:
Hello Matt @maliberty , I found a library CUSP, which is used to solve linear systems based on GPU. It is very fast. I have tested it successfully on GPL. Here is the performance:
image

I think the current direct CUDA solver is done. It is slow but correct. Can you please check it once more and let me know if I have any more places to modify?  If there is no problem, my thought is that we close this PR as soon as possible and open a new PR to upload the CUSP version.

etc/Build.sh
```diff
@@ -37,6 +37,7 @@ OPTIONS:
                                                   compile. Default: \`nproc\`
   -keep-log                                     Keep a compile log in build dir
   -help                                         Shows this message
+  -GPU                                          Enable GPU to accelerate the process
```
maliberty, line 40: all the other options are lower case so please use -gpu

src/gpl/CMakeLists.txt
```diff
@@ -85,6 +85,32 @@ target_link_libraries(gpl
...
+
+# Allow users to use GPU or not
+option(GPU "Enable GPU" OFF)
+
+if (CUDA_FOUND AND GPU)
```
maliberty, line 95: If GPU and !CUDA_FOUND warrants a warning beyond the status "GPU is not enabled"

src/gpl/src/solver.h
```diff
@@ -0,0 +1,79 @@
+#ifndef __SOLVER__
+#define __SOLVER__
```
maliberty, line 2: use "#pragma once" instead

Copyright notice needed.

src/gpl/src/solver.h
```diff
@@ -0,0 +1,79 @@
...
+
+namespace utl {
+class Logger;
+}
+
+namespace gpl {
+
+struct err{
+  float errorX; // The relative residual error for X
+  float errorY; // The relative residual error for Y
```
maliberty, line 25: error_x per the Google naming style.   Since the class name includes error it is a bit redundant to repeat it here.

src/gpl/src/solver.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+using Eigen::BiCGSTAB;
+using Eigen::IdentityPreconditioner;
+using utl::GPL;
+
+typedef Eigen::SparseMatrix<float, Eigen::RowMajor> SMatrix;
+
+#ifdef ENABLE_GPU
+ err cudaSparseSolve(int iter,
```
maliberty, line 35: This is too long to inline in a header.  Same below.

src/gpl/src/initialPlace.cpp
```diff
@@ -106,318 +96,356 @@ void InitialPlace::doBicgstabPlace() {
...
+              fixedInstForceVecX_,
+              instLocVecX_,
+              placeInstForceMatrixY_,
+              fixedInstForceVecY_,
+              instLocVecY_,
+              log_);
+    }
+  }
+  else{
+      error = cpuSparseSolve(ipVars_.maxSolverIter, iter,
```
maliberty, line 131: clang-format

src/gpl/src/initialPlace.cpp
```diff
@@ -106,318 +96,356 @@ void InitialPlace::doBicgstabPlace() {
...
+                      instLocVecX_,
+                      placeInstForceMatrixY_,
+                      fixedInstForceVecY_,
+                      instLocVecY_,
+                      log_);
+    }
+    else{
+      log_->warn(GPL, 250, "GPU is not available. CPU solver is automatically used.");
+      // BiCGSTAB solver for initial place
+      error = cpuSparseSolve(ipVars_.maxSolverIter, iter,
```
maliberty, line 120: you have this same block of code three times - refactor

src/gpl/src/gpuSolver.h
```diff
@@ -0,0 +1,64 @@
+#ifndef __GPUSOLVER__
+#define __GPUSOLVER__
```
maliberty, line 2: #pragma once

src/gpl/src/gpuSolver.h
```diff
@@ -0,0 +1,64 @@
+#ifndef __GPUSOLVER__
```
maliberty, line 1: copyright header

src/gpl/src/gpuSolver.cu
```diff
@@ -0,0 +1,208 @@
+#include "gpuSolver.h"
```
maliberty, line 1: copyright header

RuijiaoSun:
> Have you looked at gpl incremental02? You can click the details link to see the Jenkins results.

src/gpl/src/gpuSolver.cu
```diff
@@ -0,0 +1,208 @@
...
+float GpuSolver::error()
+{
+  float* r_Ax;
+  thrust::device_vector<float> t_Ax(m_);
+  thrust::fill(t_Ax.begin(), t_Ax.end(), 0);
+  r_Ax = thrust::raw_pointer_cast(t_Ax.data());
+
+  unsigned int threads = 512;
+  unsigned int blocks = (m_ + threads - 1) / threads;
+  Multi_MatVec<<<blocks, threads>>>(error_,
```
maliberty, line 193: are all the <<< needed?

maliberty (changes requested):

maliberty:
re "my thought is that we close this PR as soon as possible and open a new PR to upload the CUSP version."
Is there a reason not to integrate that change into this PR?

src/gpl/src/gpuSolver.cu
RuijiaoSun: Yes.

QuantamHD: That's the CUDA kernel launch syntax. It is needed

src/gpl/src/solver.h
RuijiaoSun: Do you suggest that I have a solver.cpp correspondingly?

src/gpl/src/initialPlace.cpp
RuijiaoSun: I think the three blocks are necessary. The program gives different warnings under different circumstances. Do you want me to wrap it and write it as a shorter sentence?

RuijiaoSun:
> re "my thought is that we close this PR as soon as possible and open a new PR to upload the CUSP version." Is there a reason not to integrate that change into this PR?

Yes. I think CUSP gives a big improvement on the performance. It's worth using it as the final version for the GPU solver. But the last update of CUSP is years ago and I am not sure if it will be maintained in the future. The direct CUDA solver is a version not dependent on high-level libraries. One day we can continue to develop our iterative GPU solver on the basis of it.

src/gpl/src/initialPlace.cpp
maliberty: Just use a simple boolean and an if statement (if (run_cpu) ....)

src/gpl/src/solver.h
maliberty: Yes

maliberty:
> > re "my thought is that we close this PR as soon as possible and open a new PR to upload the CUSP version." Is there a reason not to integrate that change into this PR?
> 
> Yes. I think CUSP gives a big improvement on the performance. It's worth using it as the final version for the GPU solver. But the last update of CUSP is years ago and I am not sure if it will be maintained in the future. The direct CUDA solver is a version not dependent on high-level libraries. One day we can continue to develop our iterative GPU solver on the basis of it.

What is the performance of the current version if we don't integrate CUSP? " It is slow but correct."

src/gpl/src/initialPlace.cpp
RuijiaoSun: Get it. Thanks.

RuijiaoSun:
> 

Much slower. 20 times slower for some experiments. Its existance just provides a frame and some records for the CUDA direct solver. Not useful for users but can be used as a reference for developers.

src/gpl/CMakeLists.txt
maliberty: This is not resolved

RuijiaoSun: find_package() is the simplest command to check if any of CUDA_capable GPUs exists. 
I have deleted this line, then here we just check if GPU is enabled. If GPU is enabled but CUDA is not found, initialplace.cpp will turn the program to CPU solver and give a warning.

RuijiaoSun: find_package() is the simplest command to check if any of CUDA_capable GPUs exists. The true value of (CUDA_found) allows the .cu file and CUDA libraries to be linked. I think we should keep this sentence.
image

maliberty: FindCUDAToolkit requires cmake 3.17.  We currently only require 3.9 but could easily update.  However Ubuntu 20.04 LTS ships with 3.16.3 which means updating to 3.17 would require a special install.  In light of that we can leave this for now.  Hopefully people will migrate to 22.04 LTS with 3.22.1 and this can be updated.

QuantamHD: @maliberty @RuijiaoSun Let's add a comment with the requirements to remove this, and what the replacement is.

i.e. OpenROAD only supports Ubuntu 22.04+ distros, and Cmake 3.17+

RuijiaoSun: image
Is this okay?

QuantamHD: That seems good.

README.md
```diff
@@ -218,6 +219,10 @@ about 1 minute to compile times and improves the runtime by about 11%. If
 you would like to disable LTO pass `-DLINK_TIME_OPTIMIZATION=OFF` when
 generating a build.
 
+### GPU acceleration
+The default solver for initial placement is based on CPU. If you would like
```
QuantamHD, line 223: Suggest the following rewordings

The default solver for initial placement is based on CPU -> The default solver for initial placement is single threaded.

set `-DGPU=true` at the compiling time -> set `-DGPU=true` at compile time.

QuantamHD:
@maliberty I think this is ready for another review.

maliberty:
Are the runtime from the CUSP solver column now what this PR will produce?

etc/DependencyInstaller.sh
```diff
@@ -69,6 +69,16 @@ _installCommonDev() {
         echo "Eigen already installed."
     fi
 
+    # CUSP
+    if [[ ! -d /usr/local/include/cusp/ ]]; then
```
maliberty, line 73: What about the cuda dependencies?

etc/DependencyInstaller.sh
```diff
@@ -69,6 +69,16 @@ _installCommonDev() {
...
+        cd "${baseDir}"
+        git clone -b cuda9 https://github.com/cusplibrary/cusplibrary.git
+        cd cusplibrary
+        sudo cp -r ./cusp /usr/local/include/
```
maliberty, line 77: @vvbandeira any concern about the use of sudo?

README.md
```diff
@@ -218,6 +219,10 @@ about 1 minute to compile times and improves the runtime by about 11%. If
 you would like to disable LTO pass `-DLINK_TIME_OPTIMIZATION=OFF` when
 generating a build.
 
+### GPU acceleration
+The default solver for initial placement is single threaded. If you would like 
+to enable GPU and use the CUDA solver, set `-DGPU=true` at compiling time.
```
maliberty, line 224: "at cmake time" not "compiling time".

src/gpl/README.md
```diff
@@ -67,6 +68,7 @@ global_placement
 - `-initial_place_max_fanout`: set net escape condition in initial place when 'fanout >= initial_place_max_fanout'. Default value is 200. Allowed values are `[1-MAX_INT, int]`.
 - `-timing_driven_net_reweight_overflow`: set overflow threshold for timing-driven net reweighting. Allowed values are `tcl list of [0-100, int]`.
 - `-verbose_level`: set verbose level for RePlAce. Default value is 1. Allowed values are `[0-5, int]`.
+- `-force_cpu`: Force to use CPU solver no matter if GPU is enabled during the compiling time.
```
maliberty, line 71: Force the use of the CPU solver even if the GPU is available.

src/gpl/src/gpuSolver.cu
```diff
@@ -0,0 +1,206 @@
...
+                "[CUSOLVER ERROR] {} at line {} in file {}\n",
+                cudaGetErrorString(*(cudaError_t*) &code),
+                __LINE__,
+                __FILE__);
+  }
+}
+
+GpuSolver::GpuSolver()
+{
+}
```
maliberty, line 83: There is no need to write the default ctor

maliberty (changes requested):

etc/DependencyInstaller.sh
QuantamHD: I'd advocate that we don't try to install CUDA it's super difficult to lock down the right version for a particular device.

You also run a high risk of screwing something up on the user's machine.

arlpetergadfort: Maybe provide a link to where the users should go get the right CUDA downloads?

maliberty: Ok but at least a message would be good as people think the dependencies will all be installed.

RuijiaoSun:
> 

This is the updated one, CUSP solver with preconditioner. The direct CUSP solver is very fast but can't converge well for big chips. The current version can provide a correct result and bring a 30% acceleration.
image

etc/DependencyInstaller.sh
RuijiaoSun: Is it okay to mention the message here?
image

etc/DependencyInstaller.sh
vvbandeira: Is not necessary. The script requires to be run as root or with sudo, otherwise it will fail on apt/yum commands.

RuijiaoSun:
> Are the runtime from the CUSP solver column now what this PR will produce?

Hi Matt, do you have more suggestions or questions for this PR?

maliberty (approved):

PR #2084, opened 2022-07-23
Fixes boost spirit parsing issues.

QuantamHD:
It turns out that boost spirit is very sensitive to being encapsulated
in functions since it heavily relies on template meta-programming.

Meaning that the lambda and functions in the old version made the parser
not actually work even though it compiled. This is the best we can
do.


maliberty:
Would you explain why the encapsulation didn't work?

QuantamHD:
Boost qi relies on both parser references and template specializations. When you put a parser in a lambda or function, it no longer allows the compiler to correctly figure out what the parser structure is. It also creates memory issues since the final construction in a given function context ref is no longer considered valid.

All examples of boost qi construct the parser in a single function.

QuantamHD:
Also added a basic unit test.

maliberty:
I like the current structure better and don't really understand why it doesn't work.  I'd like to study this a bit but will approve if you are in a rush.

QuantamHD:
I would say I need it Wednesday ish. I've tried a bunch of different things, but no matter what i did I could not return a rule from a function and still have the parser work.

maliberty (approved):

maliberty:
It looks like there is a better solution based on the feedback in https://github.com/boostorg/spirit/issues/729

maliberty (changes requested):
use qi::copy

maliberty:
Closed in favor of https://github.com/The-OpenROAD-Project/OpenROAD/pull/2093

PR #2088, opened 2022-07-25
Update new tool patch

Aidan-McNay:
I updated ./docs/misc/AddTool.patch. The previous version was outdated, and resulted in lots of hunks being lost when patching. This one should be revised so that the new tool works right after patching.

### DIFF
--- a/docs/misc/AddTool.patch
+++ b/docs/misc/AddTool.patch
@@ -1,8 +1,10 @@
-diff --git a/include/openroad/OpenRoad.hh b/include/openroad/OpenRoad.hh
-index 5026a61e..10fd7594 100755
---- a/include/openroad/OpenRoad.hh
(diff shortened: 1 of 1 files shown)


maliberty (approved):
Just curious - are you trying to add a new tool?

Aidan-McNay:
Yes! Working with @rovinski to develop a tool to visualize the clock tree with the GUI, allowing users to possibly better visualize timing errors

maliberty:
Great!  I think it would be good to have a sense of how you plan to implement this so we have up front agreement on the architecture.  It isn't clear to me that it needs to be a new tool.

rovinski:
Yep, totally fair. I was just giving @Aidan-McNay a good starting point to get new code running in OpenROAD. As we get closer to getting some of the base functionality in, we can talk more about architecture.

PR #2097, opened 2022-07-29
Verbose SPDLOG_FMT_EXTERNAL during cmake detection.

cbalint13:
Small PR to ~~fix~~ avoid ```fmt``` errors like below.

```
/usr/include/fmt/core.h:1733:7: error: static assertion failed: Cannot format an argument. 
To make type T formattable provide a formatter<T> specialization: 
https://fmt.dev/latest/api.html#udt
```

Output now signals SPDLOG mode:

```
{...}
-- OpenROAD version: 2.0-20220728.gitbd841a8c.fc37
-- System name: Linux
-- Compiler: GNU 12.1.1
{...}
-- spdlog: 1.10.0
-- spdlog: SPDLOG_FMT_EXTERNAL=OFF
{...}
```
----

More system details:

```
$ rpm -q fmt-devel
fmt-devel-9.0.0-2.fc37.x86_64
$ rpm -q spdlog-devel
spdlog-devel-1.10.0-3.fc37.x86_64
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -188,6 +188,9 @@ endif()
 
 find_package(spdlog REQUIRED)
 message(STATUS "spdlog: ${spdlog_VERSION}")
+if (${SPDLOG_FMT_EXTERNAL})
+  message(WARNING "spdlog: SPDLOG_FMT_EXTERNAL=${SPDLOG_FMT_EXTERNAL}")
+endif()
 
 find_package(Threads REQUIRED)
 set(THREADS_PREFER_PTHREAD_FLAG ON)


maliberty:
What version of fmt are you using?

cbalint13:
> What version of fmt are you using?

@maliberty ,


(wrote in details)
**stock** ones that comes with distro: https://koji.fedoraproject.org/koji/packageinfo?packageID=22982

> $ rpm -q fmt-devel
> fmt-devel-9.0.0-2.fc37.x86_64

maliberty:
Are you using the version bundled with spdlog?  What version of spdlog are you using?

cbalint13:
@maliberty ,

> Are you using the version bundled with spdlog? What version of spdlog are you using?

(described in inital details)
These are the stock distro versions (fc37 in this case):

> $ rpm -q fmt-devel
> fmt-devel-9.0.0-2.fc37.x86_64
> $ rpm -q spdlog-devel
> spdlog-devel-1.10.0-3.fc37.x86_64

cbalint13:
@maliberty ,
> 
> > Are you using the version bundled with spdlog? What version of spdlog are you using?

Seems a confusing distro related issue:

* Will check difference betweend the "two" variants of bundled vs unbundled, wasn't aware of these.
* However openroad gets confused since plain ```fmt``` is dragged in by ```spd``` (enforced as own dependence)

```
fmt-devel(x86-64) is needed by (installed) spdlog-devel-1.10.0-3.fc37.x86_64
```
```
$ rpm -ql fmt-devel | grep fmt
/usr/include/fmt
/usr/include/fmt/args.h
/usr/include/fmt/chrono.h
/usr/include/fmt/color.h
/usr/include/fmt/compile.h
/usr/include/fmt/core.h
/usr/include/fmt/format-inl.h
/usr/include/fmt/format.h
/usr/include/fmt/os.h
/usr/include/fmt/ostream.h
/usr/include/fmt/printf.h
/usr/include/fmt/ranges.h
/usr/include/fmt/std.h
/usr/include/fmt/xchar.h

$ rpm -ql spdlog-devel | grep fmt
/usr/include/spdlog/details/fmt_helper.h
/usr/include/spdlog/fmt
/usr/include/spdlog/fmt/bin_to_hex.h
/usr/include/spdlog/fmt/chrono.h
/usr/include/spdlog/fmt/compile.h
/usr/include/spdlog/fmt/fmt.h
/usr/include/spdlog/fmt/ostr.h
/usr/include/spdlog/fmt/ranges.h
/usr/include/spdlog/fmt/xchar.h
```

cbalint13:
> @maliberty ,
> 
> > > Are you using the version bundled with spdlog? What version of spdlog are you using?

Ok, figured out, updated this PR:

 * if ```fmt => 9.0.0``` then ```spd``` should be ```SPD_LOG_FMT_EXTERNAL=OFF```.
 * best we can do is to display SPD_LOG_FMT_EXTERNAL state at cmake configuration.

@maliberty , please review this cmake change instead, now it should display for us the state of flag:

```
{...}
-- OpenROAD version: 2.0-20220728.gitbd841a8c.fc37
-- System name: Linux
-- Compiler: GNU 12.1.1
{...}
-- spdlog: 1.10.0
-- spdlog: SPDLOG_FMT_EXTERNAL=OFF
{...}
```

QuantamHD:
IMO this is the wrong fix for the problem. Spdlog stopped providing automated ways to convert enums to ints.

OpenROAD should static cast the enums to ints or provide enum to string conversions. Google does the former since we pull in nightly versions of fmtlib.

cbalint13:
> IMO this is the wrong fix for the problem. Spdlog stopped providing automated ways to convert enums to ints.
> 
> OpenROAD should static cast the enums to ints or provide enum to string conversions. Google does the former since we pull in nightly versions of fmtlib.

@QuantamHD ,

I found that bunded fmt within spdlog always works fine.
With fmt-9.0 (non-bundled) gets more complicated than just static casts see: [here](https://github.com/The-OpenROAD-Project/OpenROAD/commit/7e32351f1d1b407a3831d4fcbe0c374278624673)

Best for now is only to track "SPDLOG_FMT_EXTERNAL" within cmake and issue a warning.

maliberty:
I'm not clear why we need this message.  If we do need it why is it a warning if set?

cbalint13:
> I'm not clear why we need this message. If we do need it why is it a warning if set?

It warns user that ```fmt``` is not a bundled variant within ```spd```:

   *  If ```fmt``` is external there is a high chance that openroad will **not compile** (e.g. latest stock fc37).
   *  If ```fmt``` is bundled version inside ```spd``` then openroad will **always** compile just fine.

We can make it to work in **booth cases** but **then such a patch** and future maintanace is needed: [here](https://github.com/The-OpenROAD-Project/OpenROAD/commit/7e32351f1d1b407a3831d4fcbe0c374278624673) . But in such case it is easier to just drop out the whole ```fmt+spd``` saga and think of some smarter lib (cc @QuantamHD).

---

I remind that everything started here, having issues with formatters of fmt:
```
 /usr/include/fmt/core.h:1733:7: error: static assertion failed: Cannot format an argument. 
 To make type T formattable provide a formatter<T> specialization: 
 https://fmt.dev/latest/api.html#udt
 ```

File: src/CMakeLists.txt [hunk 1]
```diff
@@ -188,6 +188,11 @@ endif()
 
 find_package(spdlog REQUIRED)
 message(STATUS "spdlog: ${spdlog_VERSION}")
+if(${SPDLOG_FMT_EXTERNAL})
```
maliberty, line 191: space after if

src/CMakeLists.txt
```diff
@@ -188,6 +188,11 @@ endif()
...
+  message(WARNING "spdlog: SPDLOG_FMT_EXTERNAL=${SPDLOG_FMT_EXTERNAL}")
+else()
+  message(STATUS "spdlog: SPDLOG_FMT_EXTERNAL=${SPDLOG_FMT_EXTERNAL}")
```
maliberty, line 194: why are we printing the value when the if failed and it is unset?  This won't be very informative.

maliberty: I would just drop the else clause

cbalint13: > I would just drop the else clause

done.

cbalint13: > why are we printing the value when the if failed and it is unset? This won't be very informative.
* Previous ```find_package(spdlog REQUIRED)``` would always set the ```SPDLOG_FMT_EXTERNAL``` otherwise would stop/error since it is enforced with ```REQUIRED``` flag.
* But let's remove the ```else``` flag anyway.

src/CMakeLists.txt
cbalint13: > space after if

linted, done.

maliberty (approved):

PR #2109, opened 2022-08-02
Drt saving memory by removing unnecessary setters and variables from…

habibayassin:
… FrLayer


rovinski:
I'm guessing the reason for the fake cut/layer is to create an `frLayer` without an underlying `dbTechLayer`, but under what circumstance would one want to do that?

osamahammad21:
> I'm guessing the reason for the fake cut/layer is to create an `frLayer` without an underlying `dbTechLayer`, but under what circumstance would one want to do that?

@rovinski The ispd test cases start with the routing layer directly without defining a masterslice layer and a cut layer. For consistency, we create a fake masterslice layer and cut layer before the defined layer stack.

jjcherry56:
This doesn't really save any memory because there are only about 15 instances of frLayer.
If you want to really have an impact do the same for frInst and frNet.

osamahammad21:
> This doesn't really save any memory because there are only about 15 instances of frLayer.
> 
> If you want to really have an impact do the same for frInst and frNet.

@jjcherry56 Hi james. This is the first step of removing duplication in tritonroute's database as we discussed before. Sure we would do the same with the instances and the netlist, but the layer seemed a good place to start. This is not solely for saving memory, but also to have a unified database.

habibayassin:
@maliberty

rovinski:
`getWidth()` and `getMinWidth()` can also be replaced.

habibayassin:
> `getWidth()` and `getMinWidth()` can also be replaced.

@rovinski I actually tried to replace them, but they are being set with values other than the ones from dbtechlayer->getWidth() or dbtechlayer->getMinWidth()

src/drt/src/db/tech/frLayer.h
```diff
@@ -127,30 +124,29 @@ class frLayer
...
   void setDefaultViaDef(frViaDef* in) { defaultViaDef = in; }
   void addConstraint(frConstraint* consIn) { constraints.push_back(consIn); }
-  void setType(dbTechLayerType typeIn) { type = typeIn; }
   void addViaDef(frViaDef* viaDefIn) { viaDefs.insert(viaDefIn); }
   void setHasVia2ViaMinStepViol(bool in) { hasMinStepViol = in; }
   // getters
   odb::dbTechLayer* getDbLayer() const { return db_layer_; }
-  frUInt4 getNumMasks() const { return numMasks; }
+  bool getFakeCut() const { return fakeCut;}
+  bool getFakeMs() const { return fakeMs;}
```
maliberty, line 139: These should be 

isFakeCut()
isFakeMasterslice()

M is too much of an abbreviation.  Please change it in the setter and member as well.

src/drt/src/db/tech/frLayer.h
```diff
@@ -170,7 +166,7 @@ class frLayer
   frViaDef* getDefaultViaDef() const { return defaultViaDef; }
   bool hasVia2ViaMinStepViol() { return hasMinStepViol; }
   std::set<frViaDef*> getViaDefs() const { return viaDefs; }
-  dbTechLayerType getType() const { return type; }
+  dbTechLayerType getType() const { return (fakeCut)? dbTechLayerType("cut") : (fakeMs)? dbTechLayerType("masterslice") : db_layer_->getType(); }
```
maliberty, line 169: Use the enum values rather than string lookup, e.g.

dbTechLayerType("masterslice") -> dbTechLayerType::MASTERSLICE

rovinski:
How so? It looks like they are:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/a2b9a98b5917fddaa9e99d801328f3f8d0e9d039/src/drt/src/io/io.cpp#L1521-L1528

src/drt/src/io/io.h
```diff
@@ -137,7 +137,7 @@ class Parser
   std::unique_ptr<frBlock> tmpBlock;
   // temporary variables
   int readLayerCnt;
-  std::string masterSliceLayerName;
+  odb::dbTechLayer* masterSliceLayerPtr;
```
maliberty, line 140: Drop Ptr from the name.

maliberty (changes requested):

maliberty:
> How so? It looks like they are:
> 
> https://github.com/The-OpenROAD-Project/OpenROAD/blob/a2b9a98b5917fddaa9e99d801328f3f8d0e9d039/src/drt/src/io/io.cpp#L1521-L1528

Line 1528 is updating minWidth.   io::Parser::initCutLayerWidth() updates width if unset from LEF

rovinski:
> Line 1528 is updating minWidth. io::Parser::initCutLayerWidth() updates width if unset from LEF

That raises an interesting question. If DRT is working around deficiencies in the loaded database, should drt be making its own modified copy or should it be updating the db? Considering DRT was already copying everything, the answer was obviously yes, but if we merge data with odb, I think it becomes a real question.

Maybe something to be addressed in a different PR.

maliberty:
@rovinski I think it is something for another PR.  I feel messing with the minWidth reflects a shortcoming in drt itself that should be fixed rather than hacking the constraints.   The other case is more interesting.

habibayassin:
> How so? It looks like they are:
> 
> https://github.com/The-OpenROAD-Project/OpenROAD/blob/a2b9a98b5917fddaa9e99d801328f3f8d0e9d039/src/drt/src/io/io.cpp#L1521-L1528

it is being set here as well:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/a2b9a98b5917fddaa9e99d801328f3f8d0e9d039/src/drt/src/io/io_parser_helper.cpp#L286

src/drt/src/db/tech/frLayer.h
```diff
@@ -127,30 +124,29 @@ class frLayer
...
-  frUInt4 getPitch() const { return pitch; }
+  void getName(frString& nameIn) const { nameIn = (fakeCut)? "FR_VIA": (fakeMasterslice)? "FR_MASTERSLICE": db_layer_->getName(); }
+  frString getName() const { return (fakeCut)? "Fr_VIA": (fakeMasterslice)? "FR_MASTERSLICE": db_layer_->getName(); }
+  frUInt4 getPitch() const { return (fakeCut || fakeMasterslice)? 0 : db_layer_->getPitch(); }
   frUInt4 getWidth() const { return width; }
   frUInt4 getMinWidth() const { return minWidth; }
-  dbTechLayerDir getDir() const { return dir; }
-  bool isVertical() { return dir == dbTechLayerDir::VERTICAL; }
-  bool isHorizontal() { return dir == dbTechLayerDir::HORIZONTAL; }
+  dbTechLayerDir getDir() const { return (fakeCut || fakeMasterslice)? dbTechLayerDir("none") : db_layer_->getDirection(); }
```
maliberty, line 147: Don't use the string constructor, just use the enum

src/drt/src/db/tech/frLayer.h
```diff
@@ -170,7 +166,16 @@ class frLayer
   frViaDef* getDefaultViaDef() const { return defaultViaDef; }
   bool hasVia2ViaMinStepViol() { return hasMinStepViol; }
   std::set<frViaDef*> getViaDefs() const { return viaDefs; }
-  dbTechLayerType getType() const { return type; }
+  dbTechLayerType getType() const
+  { 
+    if(fakeCut)
```
maliberty, line 171: clang-format

maliberty (changes requested):

maliberty (approved):

maliberty:
You'll have to fix the build errors but otherwise it looks ok

PR #2132, opened 2022-08-09
Ant fix area and length bugs

ramezmoussa:


maliberty:
When this is ready for review you can click 'request' next to my name in the 'Reviewers'.   Please address 'repair_antennas1' first.

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -68,6 +68,25 @@ struct Violation
...
+{
+    dbNet* net;
+    int routing_level;
+ 
+    // constructor
+    NetLayerPair(dbNet* net, int routing_level): net(net), routing_level(routing_level) {}
+ 
+    // overload `<` operator to use this struct as a key for an std::map
+    bool operator<(const NetLayerPair &ob) const {
+        return net->getId() < ob.net->getId() ||  (!(ob.net->getId()  < net->getId()) && routing_level < ob.routing_level);
```
rovinski, line 84:
```suggestion
        return net->getId() < ob.net->getId() ||  (ob.net->getId() == net->getId()) && routing_level < ob.routing_level);
```

src/ant/src/AntennaChecker.cc
```diff
@@ -1613,18 +1757,33 @@ AntennaChecker::parMaxWireLength(dbNet* net, int layer)
...
-  if (!net->isSpecial()) {
-    vector<std::pair<double, vector<dbITerm*>>> par_max_length_wires
-        = parMaxWireLength(net, layer);
-    for (auto par_wire : par_max_length_wires) {
-      logger_->report("Net {}: Routing Level: {}, Max Length for PAR: {:3.2f}",
-                      net_name,
-                      layer,
-                      par_wire.first);
-    }
+  if(net == NULL)
```
rovinski, line 1763:
```suggestion
  if(net == nullptr)
```

src/ant/src/AntennaChecker.cc
```diff
@@ -1613,18 +1757,33 @@ AntennaChecker::parMaxWireLength(dbNet* net, int layer)
...
+  {
+      logger_->error(ANT, 15, "No net found with the name: {}", net_name);
+  }
+
+  dbTechLayer* layer = db_->getTech()->findLayer(layer_name);
+  if(layer == NULL)
```
rovinski, line 1769:
```suggestion
  if(layer == nullptr)
```

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -68,6 +68,25 @@ struct Violation
...
+{
+    dbNet* net;
+    int routing_level;
+ 
+    // constructor
+    NetLayerPair(dbNet* net, int routing_level): net(net), routing_level(routing_level) {}
+
+    // overload `<` operator to use this struct as a key for an std::map
+    bool operator<(const NetLayerPair &ob) const {
+        return net->getId() < ob.net->getId() ||  ((ob.net->getId()  ==  net->getId()) && routing_level < ob.routing_level);
```
maliberty, line 84: This is correct but I tend to use

  return std::tie(net->getId(), routing_level) < std::tie(ob.net->getId(), ob.routing_level);

which easily generalizes to many fields.

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -193,6 +212,12 @@ class AntennaChecker
   int net_violation_count_;
 
   static constexpr int max_diode_count_per_gate = 10;
+
+  // A map indexed by: pair(net, routing_level)
+  // It stores the maximum length allowed to be held by that wire in the given layer without violating the antenna rules as
+  // well as the current wire length.
+  std::map<NetLayerPair, std::pair<double, double>> allowed_wire_length;
```
maliberty, line 219: We are 1/2 way there - convert the other pair to a simple struct too.

src/ant/src/AntennaChecker.cc
```diff
@@ -746,7 +758,50 @@ void AntennaChecker::calculateParInfo(PARinfo& par_info)
   if (par_info.iterm_gate_area == 0)
     return;
 
+
+
+  // Find the theoretical limits for PAR and its variants
+  dbTechLayerAntennaRule* antenna_rule = tech_layer->getDefaultAntennaRule();
+
+  double PAR_ratio = antenna_rule->getPAR();
```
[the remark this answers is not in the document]
maliberty: Why can't PAR_ratio be const?

maliberty (changes requested):

maliberty (approved):

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -68,6 +68,48 @@ struct Violation
...
+
+// A struct that contains 3 fields:
+//          1- net_layer: a NetLayerPair pointer
+//          2- max_length: it represents the maximum allowed length for net_layer
+//          3- cur_length: it represents the current length of the wire for the given net_layer
+// This struct is used as a value to the  the map, allowed_wire_length which is created 
+// in the class below.
+struct MaxLength
+{
+    NetLayerPair net_layer;
```
rovinski, line 102: Why does the struct have `net_layer`? It doesn't appear to be used anywhere.

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -68,6 +68,48 @@ struct Violation
...
+    double max_length;
+    double cur_length;
+
+    // constructor
+    MaxLength(NetLayerPair net_layer, double max_length, double cur_length): net_layer(net_layer), max_length(max_length),
+                                      cur_length(cur_length) {}
+
+    MaxLength()  {}
```
rovinski, line 110: Why not default constructor?

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -68,6 +68,48 @@ struct Violation
...
+// in the class below.
+struct NetLayerPair
+{
+    dbNet* net;
+    int routing_level;
+ 
+    // default constructor
+    NetLayerPair() {}
+    // constructor
+    NetLayerPair(dbNet* net, int routing_level): net(net), routing_level(routing_level) {}
```
rovinski, line 82: Why not default constructor?

src/ant/README.md
```diff
@@ -33,6 +33,18 @@ check_antennas [-net net] [-verbose]
...
+```
+find_max_allowed_length net_name layer
+```
+
+For example, you can run:
+```
+find_max_allowed_length "net5" "met2"
+```
+This will search for "net5" in your design and if found, it will find the maximum allowed length for that net on the layer `met2` such that no antenna violations occur
+
```
jjcherry56, line 47: missing period

src/ant/README.md
```diff
@@ -33,6 +33,18 @@ check_antennas [-net net] [-verbose]
 - `-verbose` report all antenna calculations for violating nets
 - 'net' check antennas on one net
 
+### `find_max_allowed_length`
+Find the maximum allowed length for a given net and layer
```
jjcherry56, line 37: this command name seems overly generic. It says nothing about the fact that it is related to antennas.

src/ant/include/ant/AntennaChecker.hh
jjcherry56: Why not typedef std::pair<dbNet*, int>?

src/ant/src/AntennaChecker.tcl
```diff
@@ -58,3 +58,13 @@ proc check_antennas { args } {
   }
   return [ant::check_antennas $net_name $verbose]
 }
+
+
+
+sta::define_cmd_args "find_max_allowed_length" {net_name \
+                                                 layer}
+proc find_max_allowed_length { args } {
+  set net_name [lindex $args 0]
```
jjcherry56, line 67: Why is this a "command"? It seems like it is just a function that should be defined with arguments. Using "args" does not test the number of args passed or the type of the arguments. A command should take a layer name, not a layer number, which is ambiguous (routing layer? layer number? which is it?). Nothing checks that the layer is actually a number either.

src/ant/include/ant/AntennaChecker.hh
ramezmoussa: It is not required in any computations at the moment but I added it for clarity; to show what this MaxLength is for. Should it be removed?

src/ant/include/ant/AntennaChecker.hh
ramezmoussa: Can you please clarify what you mean?

src/ant/include/ant/AntennaChecker.hh
ramezmoussa: @jjcherry56 initially I had the NetLayerPaid and the MaxLength as std::pair<dbNet*, int> and std::pair<double, double> respectively but @maliberty recommended using Structs instead so that they have meaningful names. Should I convert them back to Pairs and use typedef to provide the names?

src/ant/src/AntennaChecker.tcl
ramezmoussa: Would you recommend changing it a command and properly parse its argument or do I just make it a function with defined arguments? Regarding the 'layer' field, it represents the layer name so I will rename it to layer_name instead for clarity. Regarding checking, I am not making any checks here beause both 'net_name' and 'layer' are expected to be strings and checks are handled within the ANT module.

src/ant/include/ant/AntennaChecker.hh
jjcherry56, line 102: There is no point in repeating the net_layer in the struct because the map key already has it

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -68,6 +68,48 @@ struct Violation
...
+        int lhs_id = net->getId();
+        int rhs_id = ob.net->getId();
+        return std::tie(lhs_id, routing_level) < std::tie(rhs_id, ob.routing_level);
+
+    }
+};
+
+
+// A struct that contains 3 fields:
+//          1- net_layer: a NetLayerPair pointer
```
jjcherry56, line 95: It make more sense to comment the member vars in the class than add a lengthy header that repeats their names

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -193,6 +235,13 @@ class AntennaChecker
   int net_violation_count_;
 
   static constexpr int max_diode_count_per_gate = 10;
+
+  // A map indexed by: NetLayerPair
+  // It stores a MaxLength object, which holds the maximum length allowed to be held by that wire 
+  // in the given layer without violating the antenna rules.
+  // It also holds the current wire length.
+  std::map<NetLayerPair, MaxLength> allowed_wire_length;
```
jjcherry56, line 243: member var missing _ after name

src/ant/include/ant/AntennaChecker.hh
jjcherry56: Since it is only used as a key for allowed_wire_length it is overkill IMHO

src/ant/src/AntennaChecker.tcl
jjcherry56: What is the use model for this "command"? Is is going to be called as an API to the global router? Used in tests? I can't imagine it being very useful to a "user", which is why I don't like to see it pollute the command namespace and think it should just be a function in the ant namespace.

src/ant/include/ant/AntennaChecker.hh
maliberty: I never suggest using pair as first/second are meaningless member names.

src/ant/src/AntennaChecker.tcl
maliberty: @msaligane would you comment on the use model for this?  It is from the original UMich development.

msaligane: If it is not useful anymore, then please remove it. It was made initially for fastroute's antenna avoidance flow (~2y ago). I think we assigned numbers to the layers (I don't remember precisely but I think that's how layers were defined in ORFS).

src/ant/include/ant/AntennaChecker.hh
jjcherry56: https://en.cppreference.com/w/cpp/language/default_constructor

src/ant/src/AntennaChecker.cc
```diff
@@ -444,11 +456,11 @@ std::pair<double, double> AntennaChecker::calculateWireArea(
           level_nodes.insert(node);
           edge->target()->xy(end_x, end_y);
           wire_area += dbuToMicrons(abs(end_x - start_x) + abs(end_y - start_y))
-                       * wire_width;
+                       * wire_width + wire_width * wire_width;
           side_wire_area
               += (dbuToMicrons(abs(end_x - start_x) + abs(end_y - start_y))
                   + wire_width)
-                 * dbuToMicrons(wire_thickness_dbu) * 2;
+                 * dbuToMicrons(wire_thickness_dbu) * 2  + 2 * dbuToMicrons(wire_thickness_dbu) * wire_width;
```
jjcherry56, line 463: only need one call to dbuToMicrons

src/ant/src/AntennaChecker.cc
```diff
@@ -1012,19 +1092,52 @@ std::pair<bool, bool> AntennaChecker::checkWirePar(ARinfo AntennaRatio,
...
+    const double diffPAR_PWL_ratio = getPwlFactor(diffPAR, diff_area, 0);
+
+    // A struct to be used to index the 'allowed_wire_length' map.
+    // For reference, this is a map that stores the maximum allowed wire length for each layer and net along with the current wire length.
+    NetLayerPair wire_pair = NetLayerPair(net, layer->getRoutingLevel());
 
     if (PAR_ratio != 0) {
+      // Check if no entry exists for the given net-layer pair in our map
+      if(par >= 0)
+      {
```
jjcherry56, line 1108: inconsistent paren style

src/ant/src/AntennaChecker.cc
```diff
@@ -1012,19 +1092,52 @@ std::pair<bool, bool> AntennaChecker::checkWirePar(ARinfo AntennaRatio,
...
+        }
+      }
+
       if (par > PAR_ratio) {
         par_violation = true;
         violated = true;
       }
     } else {
       if (diffPAR_PWL_ratio != 0) {
+        if(diff_par > 0)
```
jjcherry56, line 1127: space after if; use clang-format

src/ant/src/AntennaChecker.tcl
jjcherry56: proc find_max_allowed_length { net_name layer_name}

maliberty:
Let's just remove find_max_allowed_length  and stop commenting on it.

Issue #2135, opened 2022-08-10
Crash during `repair_antennas`


donn:
* OpenROAD 273a9ac3a

```
OpenROAD v2.0-4585-g273a9ac3a 
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
[INFO]: Setting signal min routing layer to: met1 and clock min routing layer to met1. 
[INFO]: Setting signal max routing layer to: met5 and clock max routing layer to met5. 
-congestion_iterations 50 -verbose
[INFO GRT-0020] Min routing layer: met1
[INFO GRT-0021] Max routing layer: met5
[INFO GRT-0022] Global adjustment: 30%
[INFO GRT-0023] Grid origin: (0, 0)
[INFO GRT-0043] No OR_DEFAULT vias defined.
[INFO GRT-0088] Layer li1     Track-Pitch = 0.4600  line-2-Via Pitch: 0.3400
[INFO GRT-0088] Layer met1    Track-Pitch = 0.3400  line-2-Via Pitch: 0.3400
[INFO GRT-0088] Layer met2    Track-Pitch = 0.4600  line-2-Via Pitch: 0.3500
[INFO GRT-0088] Layer met3    Track-Pitch = 0.6800  line-2-Via Pitch: 0.6150
[INFO GRT-0088] Layer met4    Track-Pitch = 0.9200  line-2-Via Pitch: 1.0400
[INFO GRT-0088] Layer met5    Track-Pitch = 3.4000  line-2-Via Pitch: 3.1100
[INFO GRT-0019] Found 0 clock nets.
[INFO GRT-0001] Minimum degree: 2
[INFO GRT-0002] Maximum degree: 65
[INFO GRT-0003] Macros: 0
[INFO GRT-0004] Blockages: 896

[INFO GRT-0053] Routing resources analysis:
          Routing      Original      Derated      Resource
Layer     Direction    Resources     Resources    Reduction (%)
---------------------------------------------------------------
li1        Vertical            0             0          0.00%
met1       Horizontal       4480          2626          41.38%
met2       Vertical         3360          2308          31.31%
met3       Horizontal       2240          1521          32.10%
met4       Vertical         1344           949          29.39%
met5       Horizontal        448           208          53.57%
---------------------------------------------------------------

[INFO GRT-0197] Via related to pin nodes: 1363
[INFO GRT-0198] Via related Steiner nodes: 18
[...]

PR #2136, opened 2022-08-10
grt: fix bug when diode cell has definitions in multiple liberty files

eder-matheus:

Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/2135

### DIFF
--- a/src/grt/src/GlobalRouter.tcl
+++ b/src/grt/src/GlobalRouter.tcl
@@ -282,12 +282,12 @@ proc repair_antennas { args } {
       lassign [split $diode_port_name "/"] master mterm
 
       set diode_master [$db findMaster $master]
-      if { $diode_master == "" } {
+      if { $diode_master == "NULL" } {
         utl::error GRT 69 "Diode cell $master not found."
       }
 
       set diode_mterm [$diode_master findMTerm $mterm]
-      if { $diode_mterm == "" } {
+      if { $diode_mterm == "NULL" } {
         utl::error GRT 73 "Diode cell pin $mterm not found."
       }
     } else {
--- a/src/grt/src/GlobalRouter.tcl
+++ b/src/grt/src/GlobalRouter.tcl
@@ -277,15 +277,19 @@ proc repair_antennas { args } {
       # repairAntennas locates diode
       set diode_mterm "NULL"
     } elseif { [llength $args] == 1 } {
+      set db [ord::get_db]
       set diode_port_name [lindex $args 0]
-      set diode_port [sta::get_lib_pins -quiet $diode_port_name]
-      if { $diode_port == "" } {
-        utl::error GRT 69 "Diode $diode_port_name not found."
+      lassign [split $diode_port_name "/"] master mterm
+
+      set diode_master [$db findMaster $master]
+      if { $diode_master == "" } {
+        utl::error GRT 69 "Diode cell $master not found."
       }
-      if { [llength $diode_port] > 1 } {
-        set diode_port [lindex $diode_port 0]
+
+      set diode_mterm [$diode_master findMTerm $mterm]
+      if { $diode_mterm == "" } {
+        utl::error GRT 73 "Diode cell pin $mterm not found."
       }
-      set diode_mterm [sta::sta_to_db_mterm $diode_port]
     } else {
       utl::error GRT 245 "Too arguments to repair_antennas."
     }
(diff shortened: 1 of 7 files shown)


src/grt/src/GlobalRouter.tcl
```diff
@@ -282,6 +282,9 @@ proc repair_antennas { args } {
       if { $diode_port == "" } {
         utl::error GRT 69 "Diode $diode_port_name not found."
       }
+      if { [llength $diode_port] > 1 } {
+        set diode_port [lindex $diode_port 0]
+      }
       set diode_mterm [sta::sta_to_db_mterm $diode_port]
```
maliberty, line 288: Why use sta at all if you are just trying to find a dbMTerm?

eder-matheus: Do we have something like `findBTerm` to MTerms?

maliberty: findMaster() and then findMTerm() on that master

src/grt/test/repair_antennas_error1.ok
```diff
@@ -14,5 +14,5 @@
 [INFO ODB-0134] Finished DEF file: gcd_sky130.def
 [ERROR GRT-0045] Run global_route before repair_antennas.
 GRT-0045
-[ERROR GRT-0069] Diode sky130_fd_sc_hs__diode_2/xxx not found.
-GRT-0069
+[ERROR GRT-0073] Diode cell pin xxx not found.
```
maliberty, line 17: This is confusing as you don't know what master xxx belongs to.

rovinski:
Is a user-specified port for a diode even necessary? According to the LEF/DEF manual, `ANTENNACELL` should only have one input (which isn't power/ground).

eder-matheus:
> Is a user-specified port for a diode even necessary? According to the LEF/DEF manual, `ANTENNACELL` should only have one input (which isn't power/ground).

repair_antenna already detects a ANTENNACELL if no input is given for the diode name. If that is the case for every possible case, I can just remove this parameter. But notice that a few technologies don't have the ANTENNACELL defined on their LEF (which is the case in the testcase from https://github.com/The-OpenROAD-Project/OpenROAD/issues/2135).

maliberty:
> repair_antenna already detects a ANTENNACELL if no input is given for the diode name. If that is the case for every possible case, I can just remove this parameter. But notice that a few technologies don't have the ANTENNACELL defined on their LEF (which is the case in the testcase from #2135). A bunch of things can start to break.

I don't think you need to remove the option for cases like this.  However I think we don't need to require the pin name as I don't expect any cell used as an antenna to have extra non-pg pins.

Cherry fixed the LEF in ORFS but it wasn't upstreamed to open-pdks.  I'll open an issue there.

rovinski:
I agree with Matt. Specifying the port name should always be a fallback option. If the cell has more than one non-pg port, the user is using an invalid antenna cell. I would produce a warning that the user is trying to use an improperly defined cell.

src/grt/src/GlobalRouter.tcl
```diff
@@ -277,12 +277,26 @@ proc repair_antennas { args } {
...
+      set non_pg_count 0
+      foreach mterm $diode_mterms {
+        if { [$mterm getSigType] != "POWER" && [$mterm getSigType] != "GROUND" } {
+          set diode_mterm $mterm
+          incr non_pg_count
+        }
+      }
+
+      if { $non_pg_count > 1 } {
+        utl::warn "Diode cell has more than one non-pg port."
```
maliberty, line 298: I think it should be an error since we don't know which mterm to use in this case.

maliberty (approved):
I'll approve so you can merge after the change to error

donn (approved):
Moving from lib to lef is a good call for this. I was going to ask you to ensure that you use the default corner for the cell definition but you sidestepped that entirely.

eder-matheus:
> Moving from lib to lef is a good call for this. I was going to ask you to ensure that you use the default corner for the cell definition but you sidestepped that entirely.

@donn Notice that after this is merged, the OpenLane scripts need to be updated by removing the pin name. Do you want me to make this change? I'm not sure if it's ok to bump OpenROAD submodule in any PR

maliberty:
@eder-matheus does ORFS need updating?

eder-matheus:
> @eder-matheus does ORFS need updating?

No, repair_antennas isn't enabled in the ORFS

PR #2144, opened 2022-08-12
Fixes nullptr segfault when building without GUI support.

QuantamHD:
I should note that gui::Gui::get() would be safer if it was
wrapped in std::optional<>...


maliberty (approved):

QuantamHD:
@maliberty I don't have merge access, so you'll need to hit the button.

Issue #2089, opened 2022-07-26
GRT: SEGV when running STA after read_guides


antonblanchard:
I'm trying to make changes to the global routing guides and then run STA to see what effect it has. First up - I'm not sure if this is expected to work, so feel free to close if it is invalid.

My test case looks like:

```
read_liberty asap7sc7p5t_AO_RVT_FF_nldm_201020.lib      
read_liberty asap7sc7p5t_OA_RVT_FF_nldm_201020.lib   
read_liberty asap7sc7p5t_SIMPLE_RVT_FF_nldm_201020.lib
read_liberty asap7sc7p5t_INVBUF_RVT_FF_nldm_201020.lib  
read_liberty asap7sc7p5t_SEQ_RVT_FF_nldm_201020.lib  

read_lef asap7_tech_1x_201209.lef
read_lef asap7sc7p5t_27_R_1x_201211.lef  

read_def before.def

read_sdc 4_cts.sdc

source setRC.tcl

set_global_routing_layer_adjustment M2-M7 0.3
set_routing_layers -signal M2-M7

#global_route
read_guides route.guide

set_propagated_clock [all_clocks]
estimate_parasitics -global_routing

report_checks -path_delay max -fields {slew cap input nets fanout} -format full_clock_expanded
```

If I uncomment `global_route` this works (presumably because we have set up the required state), but otherwise it fails with:

```
Stack trace:
 0# 0x0000000000BB1B4E in openroad
 1# 0x00007F3F56DBD350 in /lib64/libc.so.6
 2# grt::MakeWireParasitics::estimateParasitcs(odb::dbNet*, std::vector<grt::Pin, std::allocator<grt::Pin> >&, std::vector<grt::GSegment, std::allocator<grt::GSegment> >&) in openroad
 3# grt::GlobalRouter::estimateRC() in openroad
 4# rsz::Resizer::estimateParasitics(rsz::ParasiticsSrc) in openroad
```

I added this to `readGuides`, which fixed the SEGV but I get a bunch of errors about missing routes to pins:

```
    nets = initNetlist();
    initNets(nets);
```

At that point I wondered if we need to call `addRemainingGuides`, `connectPadPins` etc.

Test case:  [read_guides.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/9185821/read_guides.tar.gz)

eder-matheus:
[...]

PR #2145, opened 2022-08-12
Grt: fix box to global routing

luis201420:
Fix issue #2089


maliberty:
There are CI failures to be fixed.

What issue motivated this change?

eder-matheus:
> There are CI failures to be fixed.
> 
> What issue motivated this change?

Luis and I identified the reasons for the issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/2089. When grt reads the guides, it was generating GSegments that:
1) Were not aligned with the routing grid. It happens to all the pins in the upper boundary and in the right boundary, where we may have GCells slightly larger than the regular ones. The first fix was to ensure that these points were assigned to the routing grid properly

2) Some pins are in the middle of a segment that the global router wasn't identifying as connected to the routing. In the figure below, this is what we have when first reading the guides:
Screen Shot 2022-08-10 at 18 16 52
The pins in the guide in the center were not connected because we're not generating a GSegment that touches it. So we changed the readGuides function to create GSegments at each grid point, like the figure below:
Screen Shot 2022-08-10 at 18 18 49

The failure in the drt unit test happens because we're trying to ensure the GSegments are aligned with the routing grid. The problem with this unit test is that the guides (from ISPD18) were not created in the grt routing grid. When we create the GSegments for it, they are not corresponding with the guides.

My suggestion to @luis201420 was to create a function that stores the guides in the DB directly from the file, not from the GRT data structures. But perhaps we should build the grt grid based on the guides file, but I'm not sure about that. What do you think, @maliberty?

maliberty:
Going directly to odb seems fine.

maliberty (approved):

PR #2152, opened 2022-08-14
OR-1550 drt: set unidirectional routing layer

osamahammad21:
https://openroad.atlassian.net/browse/OR-1550

### DIFF
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -911,6 +911,19 @@ void TritonRoute::addUserSelectedVia(const std::string& viaName)
   }
 }
 
+void TritonRoute::setUnidirectionalLayer(const std::string& layerName)
+{
+  if (db_->getTech() == nullptr) {
+    logger_->error(DRT, 615, "Load tech before setting unidirectional layers");
+  }
+  auto tech = db_->getTech();
+  if (tech->findLayer(layerName.c_str()) == nullptr) {
+    logger_->error(utl::DRT, 616, "Layer {} not found", layerName);
+  } else {
+    design_->getTech()->setUnidirectionalLayer(layerName);
+  }
+}
+
 void TritonRoute::setParams(const ParamStruct& params)
 {
   OUT_MAZE_FILE = params.outputMazeFile;
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -917,10 +917,11 @@ void TritonRoute::setUnidirectionalLayer(const std::string& layerName)
     logger_->error(DRT, 615, "Load tech before setting unidirectional layers");
   }
   auto tech = db_->getTech();
-  if (tech->findLayer(layerName.c_str()) == nullptr) {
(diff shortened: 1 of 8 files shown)


src/drt/src/TritonRoute.cpp
```diff
@@ -911,6 +911,19 @@ void TritonRoute::addUserSelectedVia(const std::string& viaName)
...
+void TritonRoute::setUnidirectionalLayer(const std::string& layerName)
+{
+  if (db_->getTech() == nullptr) {
+    logger_->error(DRT, 615, "Load tech before setting unidirectional layers");
+  }
+  auto tech = db_->getTech();
+  if (tech->findLayer(layerName.c_str()) == nullptr) {
+    logger_->error(utl::DRT, 616, "Layer {} not found", layerName);
+  } else {
+    design_->getTech()->setUnidirectionalLayer(layerName);
```
maliberty, line 923: Keep the return value of findLayer and use the pointer rather than the name from this point forward.

maliberty: Never mind I see this is before the frLayer is generated

src/drt/src/io/io.cpp
```diff
@@ -1217,7 +1217,7 @@ void io::Parser::setRoutingLayerProperties(odb::dbTechLayer* layer,
     tmpLayer->addLef58SpacingEndOfLineConstraint(con.get());
     tech->addUConstraint(std::move(con));
   }
-  if (layer->isRectOnly()) {
+  if (layer->isRectOnly() || tech->hasUnidirectionalLayer(layer->getName())) {
     auto rectOnlyConstraint = make_unique<frLef58RectOnlyConstraint>(
         layer->isRectOnlyExceptNonCorePins());
```
maliberty, line 1222: @arlpetergadfort is this what you have in mind?  This generates a rect only constraint.

arlpetergadfort: No, this will cause issues when the via enclosures are wider than the routing width.
It seems like the better place to do this would be in:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/4a99e88667b0840531ca0096c4fa0da8f80d6cb1/src/drt/src/db/tech/frLayer.h#L154

arlpetergadfort: When I had to do this, I modified the number of masks, since that didn't introduce new DRCs

osamahammad21: I get your point. However, my only problem with editing isUnidirectional is that pin access does not check for it when generating access points which pushed me to make it rectOnly. So without rectOnly, pin access would allow for planar access in the wrong direction(if I understand correctly). @maliberty is there a good reason pin access does not use isUnidrectional instead of rectOnly? https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/drt/src/pa/FlexPA_prep.cpp#L203-L218

src/drt/src/TritonRoute.cpp
osamahammad21: still it makes more sense to keep the pointer instead, right?

maliberty (approved):

PR #2155, opened 2022-08-16
DRT Report Violations at end of each iterations

osamahammad21:
https://openroad.atlassian.net/browse/OR-1552

### DIFF
--- a/src/drt/src/db/tech/frConstraint.h
+++ b/src/drt/src/db/tech/frConstraint.h
@@ -80,12 +80,152 @@ class frConstraint
   void setLayer(frLayer* layer) { layer_ = layer; }
   void setId(int in) { id_ = in; }
   int getId() const { return id_; }
+  std::string getViolName() const
+  {
+    switch (typeId()) {
+      case frConstraintTypeEnum::frcShortConstraint:
+        return "Short";
+      case frConstraintTypeEnum::frcMinWidthConstraint:
+        return "MinWid";
+
+      case frConstraintTypeEnum::frcSpacingConstraint:
+        return "MetSpc";
+
+      case frConstraintTypeEnum::frcSpacingEndOfLineConstraint:
+        return "EOLSpc";
+
+      case frConstraintTypeEnum::frcSpacingTablePrlConstraint:
+        return "MetSpc";
+
+      case frConstraintTypeEnum::frcCutSpacingConstraint:
+        return "CutSpc";
+
+      case frConstraintTypeEnum::frcMinStepConstraint:
+        return "MinStp";
+
+      case frConstraintTypeEnum::frcNonSufficientMetalConstraint:
+        return "NSMet";
+
+      case frConstraintTypeEnum::frcSpacingSamenetConstraint:
+        return "MetSpc";
+
+      case frConstraintTypeEnum::frcOffGridConstraint:
+        return "OffGrid";
+
+      case frConstraintTypeEnum::frcMinEnclosedAreaConstraint:
+        return "MinHole";
+
+      case frConstraintTypeEnum::frcAreaConstraint:
+        return "MinArea";
+
+      case frConstraintTypeEnum::frcLef58CornerSpacingConstraint:
+        return "CornerSpc";
+
+      case frConstraintTypeEnum::frcLef58CutSpacingConstraint:
+        return "CutSpc";
+
+      case frConstraintTypeEnum::frcLef58RectOnlyConstraint:
+        return "RectOnly";
+
+      case frConstraintTypeEnum::frcLef58RightWayOnGridOnlyConstraint:
+        return "RightWayOnGridOnly";
+
+      case frConstraintTypeEnum::frcLef58MinStepConstraint:
+        return "MinStp";
+
+      case frConstraintTypeEnum::frcSpacingTableInfluenceConstraint:
+        return "MetSpcInf";
+
+      case frConstraintTypeEnum::frcSpacingEndOfLineParallelEdgeConstraint:
+        return "SpacingEndOfLineParallelEdge";
+
+      case frConstraintTypeEnum::frcSpacingTableConstraint:
+        return "SpacingTable";
+
+      case frConstraintTypeEnum::frcSpacingTableTwConstraint:
+        return "SpacingTableTw";
+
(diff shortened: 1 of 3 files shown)


src/drt/src/dr/FlexDR.cpp
```diff
@@ -1823,6 +1824,43 @@ void FlexDR::searchRepair(const SearchRepairArgs& args)
...
                   getDesign()->getTopBlock()->getNumMarkers());
+    if (getDesign()->getTopBlock()->getNumMarkers() > 0) {
+      // report violations
+      std::map<std::string, std::map<frLayerNum, uint>> violations;
+      std::set<frLayerNum> layers;
+      const std::map<std::string, std::string> relabel
+          = {{"Lef58SpacingEndOfLine", "EOL"},
+             {"Lef58CutSpacingTable", "CutSpcTbl"},
+             {"Lef58EolKeepOut", "eolKeepOut"}};
+      for (const auto& marker : getDesign()->getTopBlock()->getMarkers()) {
```
maliberty, line 1835: I don't think users will care about recheck violations as they are really internal to the router.  I would limit them to a developer debug level.

src/drt/src/dr/FlexDR.cpp
```diff
@@ -1823,6 +1824,43 @@ void FlexDR::searchRepair(const SearchRepairArgs& args)
...
                   "  Number of violations = {}.",
                   getDesign()->getTopBlock()->getNumMarkers());
+    if (getDesign()->getTopBlock()->getNumMarkers() > 0) {
+      // report violations
+      std::map<std::string, std::map<frLayerNum, uint>> violations;
+      std::set<frLayerNum> layers;
+      const std::map<std::string, std::string> relabel
+          = {{"Lef58SpacingEndOfLine", "EOL"},
+             {"Lef58CutSpacingTable", "CutSpcTbl"},
+             {"Lef58EolKeepOut", "eolKeepOut"}};
```
maliberty, line 1834: There is no need to remap here since you can just change the name in getViolName()  [my script didn't have the choice]

src/drt/src/dr/FlexDR.cpp
```diff
@@ -1823,6 +1824,43 @@ void FlexDR::searchRepair(const SearchRepairArgs& args)
...
+      logger_->report(line);
+      for (auto [type, typeViolations] : violations) {
+        std::string typeName = type;
+        if (relabel.find(typeName) != relabel.end())
+          typeName = relabel.at(typeName);
+        if (typeName.size() >= 15)
+          typeName = typeName.substr(0, 12) + "..";
+        line = fmt::format("{:<15}", typeName);
+        for (auto lNum : layers) {
+          line += fmt::format("{:<7}", typeViolations[lNum]);
```
maliberty, line 1859: Right align the numeric values and their headers

src/drt/src/db/tech/frConstraint.h
```diff
@@ -80,12 +80,152 @@ class frConstraint
   void setLayer(frLayer* layer) { layer_ = layer; }
   void setId(int in) { id_ = in; }
   int getId() const { return id_; }
+  std::string getViolName() const
+  {
+    switch (typeId()) {
+      case frConstraintTypeEnum::frcShortConstraint:
+        return "Short";
+      case frConstraintTypeEnum::frcMinWidthConstraint:
+        return "MinWid";
```
maliberty, line 89: Since we are afforded 15 characters lets pick some better names.  Would you take a first pass through and pick nicer names.  For example

MinWid -> Min Width 
CutSpc -> Cut Spacing

src/drt/src/db/tech/frConstraint.h
```diff
@@ -80,12 +80,152 @@ class frConstraint
...
+      case frConstraintTypeEnum::frcRecheckConstraint:
+        return "Recheck";
+
+      case frConstraintTypeEnum::frcLef58EolExtensionConstraint:
+        return "Lef58EolExtension";
+
+      case frConstraintTypeEnum::frcLef58EolKeepOutConstraint:
+        return "Lef58EolKeepOut";
+      default:
+        return "UNKNOWN";
```
maliberty, line 221: Remove the default case and handle all values (this will also error if we add a new type and forget to update this).

maliberty (changes requested):

src/drt/src/dr/FlexDR.cpp
osamahammad21: The same getViolName function is used in the drcReport. I don't want to change the names there.

maliberty: I don't like having two different sets of names in different reports.   I think we should have a compact name for reporting and in the drc report we should have the lef file/line# if someone wants the full details of the constraint (it wouldn't be hard to keep that in odb).

osamahammad21: I am not sure if it's easy to get the line number of the rule from the lef in odb, but I think it would require me to edit the "lef" module in odb to keep track of the line numbers for the properties. Even if it's easy, we'd have to wait for the frConstraint's to be linked to the odb Rules.

src/drt/src/dr/FlexDR.cpp
osamahammad21: Wouldn't it be wierd that the number of markers reported by info message "199" doesn't match the violations reported here if we remove the recheck violations? Or should we remove the recheck violations from the info message as well?

maliberty: That's a good point.  Removing them from 199 could make it appear that there are zero violations and yet the router keeps running which will seem odd too.  I guess we can just leave it the way it is and see if people are confused or not.

src/drt/src/dr/FlexDR.cpp
maliberty: We can take this up in a separate PR.

src/drt/src/dr/FlexDR.cpp
osamahammad21: Should the violation names still be lef alligned?

maliberty: Yes

maliberty (approved):

Issue #2187, opened 2022-08-23
ant: OpenROAD and magic disagree on antenna violations


antonblanchard:
Using the following test case OpenROAD reports no antenna violations, but magic reports some quite large ones. An example:

```
Cell: fanout1050
Antenna violation detected at plane metal3
Effective antenna ratio 1584.67 > limit 400
Gate rect (15531 76759) to (15561 76887)
Antenna rect (36813 76739) to (36879 76742)
```

I'm not sure if OpenROAD or magic is wrong.

Test case: [antenna.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/9398600/antenna.tar.gz)

maliberty:
Anton noticed a drop in antenna violations with your recent changes. Please investigate

antonblanchard:
I don't see any `THICKNESS` properties in the TLEF file. Could be a bug in open_pdks. Investigating.

maliberty:
They are there in sky130A.

antonblanchard:
Recent shuttles are using sky130B, and that's where the issue is:

https://github.com/RTimothyEdwards/open_pdks/pull/272

ramezmoussa:
@maliberty is there a defualt `THICKNESS` that should be used in cases like this? The current implementation uses 0 for the thickness and I can confirm that this is why no violations were reported. I experimented with various values for the `THICKNESS` and multiple violations arose.

antonblanchard:
Since this was a problem in the LEF file, one option might be to just return an error. A default value could mask the issue.

antonblanchard:
After fixing the tech LEF `THICKNESS` issues, I still see some rather large differences. In this example magic calculates the antenna violation to be about twice as much as OpenROAD:

```
magic:
Cell: final_adder.U$$49
Antenna violation detected at plane metal2
Effective antenna ratio 1479.6 > limit 400
```

```
OpenROAD:
  final_adder.U$$49/A (sky130_fd_sc_hd__ha_1)
    met2
    PAR:  759.75* Ratio:  400.00 (S.Area)
```

@rtimothyedwards I think there may be an error in how magic calculates antenna violations, at least if we want to match
the tech LEF. The antenna rule is:

```
  ANTENNADIFFSIDEAREARATIO PWL ( ( 0 400 ) ( 0.0125 400 ) ( 0.0225 2609 ) ( 22.5 11600 ) ) ;
```

[...]

PR #2194, opened 2022-08-23
ant: error when thickness is missing

ramezmoussa:
This PR addresses Issue #2187.

Originally, when thickness ismissing from any routing layer, a value of 0 was used. In this PR, I changed this so that an error is reported if the thickness is missing.

### DIFF
--- a/src/ant/src/AntennaChecker.cc
+++ b/src/ant/src/AntennaChecker.cc
@@ -30,31 +30,30 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 #include <stdio.h>
+#include <tcl.h>
 
 #include <cstring>
 #include <iostream>
 #include <unordered_set>
 
-#include <tcl.h>
-
 #include "ant/AntennaChecker.hh"
+#include "grt/GlobalRouter.h"
 #include "odb/db.h"
 #include "odb/dbTypes.h"
 #include "odb/dbWireGraph.h"
 #include "odb/wOrder.h"
 #include "sta/StaMain.hh"
 #include "utl/Logger.h"
-#include "grt/GlobalRouter.h"
 
 namespace ant {
 
 using odb::dbBox;
-using odb::dbLib;
 using odb::dbBTerm;
 using odb::dbInst;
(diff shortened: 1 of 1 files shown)


src/ant/src/AntennaChecker.cc
```diff
@@ -218,10 +220,22 @@ void AntennaChecker::initAntennaRules()
                                   plus_diff_factor,
                                   diff_metal_reduce_factor};
     layer_info_[tech_layer] = layer_antenna;
-  }
 
-  block_ = db_->getChip()->getBlock();
-  dbu_per_micron_ = block_->getDbUnitsPerMicron();
+    double wire_width = dbuToMicrons(tech_layer->getWidth());
```
maliberty, line 224: Why both with dbuToMicrons if you are just going to == 0?  Same with thickness

make const  (I'm quite pedantic about const).

src/ant/src/AntennaChecker.cc
```diff
@@ -218,10 +220,22 @@ void AntennaChecker::initAntennaRules()
...
+    uint wire_thickness_dbu = 0;
+    tech_layer->getThickness(wire_thickness_dbu);
+    double wire_thickness = dbuToMicrons(wire_thickness_dbu);
+
+    if(tech_layer->getType() == dbTechLayerType::ROUTING && wire_thickness == 0)
```
maliberty, line 229: Just do the getType comparison once.

clang-format

src/ant/src/AntennaChecker.cc
```diff
@@ -218,10 +220,22 @@ void AntennaChecker::initAntennaRules()
...
+    {
+      logger_->error(ANT, 13, "No THICKNESS is provided for {}.", tech_layer->getConstName());
+    }
+
+    if(tech_layer->getType() == dbTechLayerType::ROUTING && wire_width == 0)
```
maliberty, line 234: Do we need to check width?  Any real wire will have a width determined by its geometry.

maliberty (approved):

PR #2195, opened 2022-08-24
Set dont touch

jjcherry56:
set_dont_touch command

### DIFF
--- a/src/rsz/src/Resizer.tcl
+++ b/src/rsz/src/Resizer.tcl
@@ -247,7 +247,7 @@ proc set_dont_use { args } {
 sta::define_cmd_args "unset_dont_use" {lib_cells}
 
 proc unset_dont_use { args } {
-  set_dont_use_cmd "set_dont_use" $args 0
+  set_dont_use_cmd "unset_dont_use" $args 0
 }
 
 proc set_dont_use_cmd { cmd cmd_args dont_use } {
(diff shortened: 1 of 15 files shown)


QuantamHD:
Perhaps adding a write constraint on the db from specific tools would be an interesting addition.

So something like when entering resizer we set a flag in the db that it's resizer doing the modifications. Then we assert that no write operations on instances occur when in resizer when the don't touch flag is set.

maliberty:
> Perhaps adding a write constraint on the db from specific tools would be an interesting addition.
> 
> So something like when entering resizer we set a flag in the db that it's resizer doing the modifications. Then we assert that no write operations on instances occur when in resizer when the don't touch flag is set.

dont-touch is already a flag in the db and we could make it an error to attempt to modify such an object regardless of the client (eg setLocation is an error on a dont-touch instance).  That would be a pure odb enhancement.

src/rsz/src/Resizer.tcl
```diff
@@ -241,8 +241,43 @@ proc estimate_parasitics { args } {
...
 proc set_dont_use { args } {
-  sta::check_argc_eq1 "set_dont_use" $args
-  rsz::set_dont_use_cmd [sta::get_lib_cells_arg "-dont_use" [lindex $args 0] utl::warn]
+  set_dont_use_cmd "set_dont_use" $args 1
+}
+
+sta::define_cmd_args "unset_dont_use" {lib_cells}
+
+proc unset_dont_use { args } {
+  set_dont_use_cmd "set_dont_use" $args 0
```
arlpetergadfort, line 250: Should this be "unset_dont_use"

PR #2208, opened 2022-08-27
ant: THICKNESS error only when side area is checked

ramezmoussa:

### DIFF
--- a/src/ant/src/AntennaChecker.cc
+++ b/src/ant/src/AntennaChecker.cc
@@ -218,18 +218,6 @@ void AntennaChecker::initAntennaRules()
                                   plus_diff_factor,
                                   diff_metal_reduce_factor};
     layer_info_[tech_layer] = layer_antenna;
-
-    uint wire_thickness_dbu = 0;
-    tech_layer->getThickness(wire_thickness_dbu);
-
-    const dbTechLayerType layerType = tech_layer->getType();
-
-    if (layerType == dbTechLayerType::ROUTING && wire_thickness_dbu == 0) {
-      logger_->warn(ANT,
-                     13,
-                     "No THICKNESS is provided for {}.",
-                     tech_layer->getConstName());
-    }
   }
 }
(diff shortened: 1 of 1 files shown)


maliberty:
Why can't this be checked in initAntennaRules once rather than delaying it?

ramezmoussa:
> Why can't this be checked in initAntennaRules once rather than delaying it?

I think the checks can be moved into the `initAntennaRules()` function. At first,  I thought it might be problematic because to get the ratios, we might need to know the value of the diffusion connected. However, I think we only need to check that a rule exists; no need to find the exact value so moving it to the `initAntennaRules()` should be possible.

src/ant/src/AntennaChecker.cc
```diff
@@ -1744,12 +1750,14 @@ bool AntennaChecker::checkViolation(const PARinfo& par_info, dbTechLayer* layer)
...
+    const dbTechLayerAntennaRule::pwl_pair diffPSR = antenna_rule->getDiffPSR();
+    const double diffPSR_ratio = getPwlFactor(diffPSR, diff_area, 0.0);
+
+
     if (PSR_ratio != 0) {
       if (psr > PSR_ratio)
         return true;
     } else {
-      dbTechLayerAntennaRule::pwl_pair diffPSR = antenna_rule->getDiffPSR();
-      const double diffPSR_ratio = getPwlFactor(diffPSR, diff_area, 0.0);
```
maliberty, line 1752: Is this change needed?

ramezmoussa: Not really. It was introduced when the checks were handled here instead of the init function but it is not required anymore. It should not have any impact though.

maliberty: There would be a small runtime cost and a bit of readability.  If there is no reason then let's return it to the previous form.

maliberty (approved):

PR #2216, opened 2022-08-29
updating CTS units and capacitance/slew intervals.

Mahfouz-z:
Disambiguating cts units and removing unecssary extra units dimensions. Editing capacitance and slew steps.

### DIFF
--- a/src/cts/src/CtsOptions.h
+++ b/src/cts/src/CtsOptions.h
@@ -72,13 +72,11 @@ class CtsOptions
   void setWireSegmentUnitDbu(unsigned wireSegmentUnit)
   {
     wireSegmentUnitDbu_ = wireSegmentUnit;
-    setWireSegmentUnitMicron(wireSegmentUnit * 1.0 / getDbUnits());
   }
   unsigned getWireSegmentUnitDbu() const { return wireSegmentUnitDbu_; }
   void setWireSegmentUnitMicron(unsigned wireSegmentUnit)
   {
     wireSegmentUnitMicron_ = wireSegmentUnit;
-    setWireSegmentUnitDbu(wireSegmentUnit * getDbUnits());
   }
   unsigned getWireSegmentUnitMicron() const { return wireSegmentUnitMicron_; }
   void setPlotSolution(bool plot) { plotSolution_ = plot; }
@@ -217,7 +215,7 @@ class CtsOptions
   std::string metricFile_ = "";
   int dbUnits_ = -1;
   unsigned wireSegmentUnitDbu_ = 0;
-  double wireSegmentUnitMicron_ = 0;
+  unsigned wireSegmentUnitMicron_ = 0;
   bool plotSolution_ = false;
   bool simpleCts_ = false;
   bool sinkClusteringEnable_ = true;
--- a/src/cts/src/CtsOptions.h
+++ b/src/cts/src/CtsOptions.h
@@ -67,8 +67,8 @@ class CtsOptions
     bufferList_ = buffers;
   }
   std::vector<std::string> getBufferList() const { return bufferList_; }
-  void setDbUnits(float units) { dbUnits_ = units; }
-  float getDbUnits() const { return dbUnits_; }
+  void setDbUnits(int units) { dbUnits_ = units; }
+  int getDbUnits() const { return dbUnits_; }
   void setWireSegmentUnitDbu(unsigned wireSegmentUnit)
   {
     wireSegmentUnitDbu_ = wireSegmentUnit;
@@ -89,8 +89,6 @@ class CtsOptions
   bool getSinkClusteringUseMaxCap() const { return sinkClusteringUseMaxCap_; }
   void setNumMaxLeafSinks(unsigned numSinks) { numMaxLeafSinks_ = numSinks; }
   unsigned getNumMaxLeafSinks() const { return numMaxLeafSinks_; }
-  void setMaxSlew(unsigned slew) { maxSlew_ = slew; }
-  unsigned getMaxSlew() const { return maxSlew_; }
   void setMaxCharSlew(double slew) { maxCharSlew_ = slew; }
   double getMaxCharSlew() const { return maxCharSlew_; }
   void setMaxCharCap(double cap) { maxCharCap_ = cap; }
@@ -208,7 +206,7 @@ class CtsOptions
   std::string sinkBuffer_ = "";
   std::string treeBuffer_ = "";
   std::string metricFile_ = "";
-  float dbUnits_ = -1;
+  int dbUnits_ = -1;
   unsigned wireSegmentUnitDbu_ = 0;
   bool plotSolution_ = false;
   bool simpleCts_ = false;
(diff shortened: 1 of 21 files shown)


Mahfouz-z:
@maliberty the tests are still not passing as I did not update the .ok files until we make sure the logic is correct.

src/cts/src/HTreeBuilder.cpp
```diff
@@ -157,7 +157,7 @@ void HTreeBuilder::preSinkClustering(
 
 void HTreeBuilder::initSinkRegion()
 {
-  unsigned wireSegmentUnitInMicron = techChar_->getLengthUnit();
+  unsigned wireSegmentUnitInMicron = techChar_->getLengthUnitMicron();
```
maliberty, line 160: Yikes - microns in unsigned will truncate.

Mahfouz-z:
@maliberty In this commit, I tried to eliminate the use of Microns in TechChar and HTreeBuilder. I also added your note that the user should input the count of capacitance and slew intervals instead of their deltas.

src/cts/src/CtsOptions.h
```diff
@@ -67,13 +67,13 @@ class CtsOptions
     bufferList_ = buffers;
   }
   std::vector<std::string> getBufferList() const { return bufferList_; }
-  void setDbUnits(int units) { dbUnits_ = units; }
-  int getDbUnits() const { return dbUnits_; }
-  void setWireSegmentUnit(unsigned wireSegmentUnit)
+  void setDbUnits(float units) { dbUnits_ = units; }
```
maliberty, line 70: Why is this a float now?

src/cts/src/HTreeBuilder.cpp
```diff
@@ -367,7 +375,7 @@ void HTreeBuilder::computeLevelTopology(unsigned level,
     remainingLength = previousLevel.getRemainingLength();
   }
 
-  const unsigned SLEW_THRESHOLD = options_->getMaxSlew();
+  const unsigned SLEW_THRESHOLD = techChar_->getMaxSlew(); // The normalized value of the slew of the characterization buffer (max slew)
```
maliberty, line 378: Why is this stored in two different places?  Can we consolidate this to just one?

src/cts/src/TechChar.cpp
```diff
@@ -551,28 +536,32 @@ void TechChar::initCharacterization()
 
   // Defines the different wirelengths to test and the characterization unit.
   unsigned wirelengthIterations = options_->getCharWirelengthIterations();
-  unsigned maxWirelength = (charBuf_->getHeight() * 10)
-                           * wirelengthIterations;  // Hard-coded limit
-  if (options_->getWireSegmentUnit() == 0) {
-    unsigned charaunit = charBuf_->getHeight() * 10;
-    options_->setWireSegmentUnit(charaunit);
+  if (options_->getWireSegmentUnitDbu() == 0) {
+    unsigned charaunitDbu = charBuf_->getHeight() * 10;
```
maliberty, line 540: charaunitDbu is an odd name - maybe wireSegmentUnitDbu

src/cts/src/TechChar.cpp
```diff
@@ -551,28 +536,32 @@ void TechChar::initCharacterization()
...
+    options_->setWireSegmentUnitDbu(charaunitDbu);
   } else {
-    // Updates the units to DBU.
-    unsigned segmentDistance = options_->getWireSegmentUnit();
-    options_->setWireSegmentUnit(segmentDistance * dbUnitsPerMicron);
+    // user has passed the segment unit in Micron, so we need to convert
+    // it to DBU to keep correctness and uniformity.
+    options_->setWireSegmentUnitDbu(options_->getWireSegmentUnitDbu()
+                                    * dbUnitsPerMicron);
```
maliberty, line 546: The conversion from microns to dbu should be done on the way in from tcl rather the delayed to here.  I want to remove microns as much as possible from the core code.

maliberty (changes requested):

src/cts/src/CtsOptions.h
Mahfouz-z: TechChar.cpp line 447 confused me that the value is actually a float but yes, it should be an int.

src/cts/src/HTreeBuilder.cpp
Mahfouz-z: Yeah makes since to make a single source for that, actually, PostCtsOpt was still using the max from options_ too.

src/cts/src/CtsOptions.h
Mahfouz-z: Sorry, I confused it.

src/cts/src/CtsOptions.h
```diff
@@ -209,20 +207,19 @@ class CtsOptions
...
   bool simpleCts_ = false;
   bool sinkClusteringEnable_ = true;
   bool sinkClusteringUseMaxCap_ = true;
   bool simpleSegmentsEnable_ = false;
   bool vertexBuffersEnable_ = false;
   bool gui_debug_ = false;
-  double vertexBufDistance_ = 240;
-  double bufDistance_ = 100;
+  double vertexBufDistance_ = 0; // in DBU 
+  double bufDistance_ = 0; // in DBU
```
rovinski, line 219: Comment is DBU but type is double? Also why change value to 0?

src/cts/src/HTreeBuilder.cpp
```diff
@@ -33,14 +33,15 @@
 //
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "HTreeBuilder.h"
```
rovinski, line 36: Why is this moved?

src/cts/src/TechChar.cpp
```diff
@@ -33,13 +33,14 @@
 //
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "TechChar.h"
```
rovinski, line 36: Same

src/cts/src/TechChar.cpp
```diff
@@ -451,18 +444,10 @@ void TechChar::initCharacterization()
 {
   odb::dbChip* chip = db_->getChip();
   odb::dbBlock* block = chip->getBlock();
-  float dbUnitsPerMicron = block->getDbUnitsPerMicron();
-
+  int dbUnitsPerMicron = block->getDbUnitsPerMicron();
```
rovinski, line 447: Need to be careful because `getClockLayerResCap()` now takes in `int`, not `float`

src/cts/src/TritonCTS.i
```diff
@@ -141,14 +144,14 @@ void
 set_distance_between_buffers(double distance)
 {
   getTritonCts()->getParms()->setSimpleSegmentsEnabled(true);
-  getTritonCts()->getParms()->setBufferDistance(distance);
+  getTritonCts()->getParms()->setBufferDistance(distance * getDb()->getChip()->getBlock()->getDbUnitsPerMicron());
```
rovinski, line 147: When converting from um to DBU I would recommend using `std::round()` because the implicit cast from float to int will truncate.

rovinski (commented):
Needs another clang format I believe

src/cts/src/CtsOptions.h
Mahfouz-z: These two entries used to be in Microns; therefore, they were double, will convert them to ints. I changed the value to 0 because it was set to a value that could be considered as default; however, it was not really used at all as these values are only used when a boolean is activated that depends on these values being passed in the tcl.

src/cts/src/HTreeBuilder.cpp
Mahfouz-z: It was moved after running clang-format.

src/cts/src/TechChar.cpp
Mahfouz-z: same here

src/cts/src/TechChar.cpp
Mahfouz-z: static cast to float will do the trick, right?

src/cts/src/TritonCTS.i
Mahfouz-z: Done.

src/cts/src/HTreeBuilder.cpp
rovinski: clang-format moved it from the correct order to out-of-order? That doesn't seem right...

maliberty:
obsolete

PR #2230, opened 2022-09-01
Improve drt performance by inlining dbTechLayerDir constructors

antonblanchard:
getIdx() continually calls into odb code to determine the layer direction.
Create a helper isZDirHorizontal() which caches this.

Detailed routing of the Microwatt register file is 8.5% faster with this
applied.

### DIFF
--- a/src/odb/include/odb/dbTypes.h
+++ b/src/odb/include/odb/dbTypes.h
@@ -504,17 +504,17 @@ class dbTechLayerType
   ///
   /// Create a dbTechLayerType instance with an explicit value.
   ///
-  dbTechLayerType(Value value);
+  dbTechLayerType(Value value) { _value = value; }
 
   ///
   /// Create a dbTechLayerType instance with value = "none".
   ///
-  dbTechLayerType();
+  dbTechLayerType() { _value = ROUTING; }
 
   ///
   /// Copy constructor.
   ///
-  dbTechLayerType(const dbTechLayerType& value);
+  dbTechLayerType(const dbTechLayerType& value) { _value = value._value; }
 
   ///
   /// Returns the layer-value.
@@ -558,17 +558,17 @@ class dbTechLayerDir
   ///
   /// Create a dbTechLayerDir instance with an explicit direction.
   ///
-  dbTechLayerDir(Value value);
+  dbTechLayerDir(Value value) { _value = value; }
 
   ///
   /// Create a dbTechLayerDir instance with direction = "none".
   ///
-  dbTechLayerDir();
+  dbTechLayerDir() { _value = NONE; }
 
   ///
   /// Copy constructor.
   ///
-  dbTechLayerDir(const dbTechLayerDir& value);
+  dbTechLayerDir(const dbTechLayerDir& value) { _value = value._value; }
 
   ///
   /// Returns the layer-direction.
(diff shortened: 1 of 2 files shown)


antonblanchard:
FlexGRGridGraph does something similar already.

The other users of getZDir() error out if the direction is neither horizontal nor vertical. Perhaps we could move that error into the initialisation code and make getZDir() match FlexGRGridGraph and return direction == HORIZONTAL

src/drt/src/dr/FlexGridGraph.h
```diff
@@ -1009,6 +1011,7 @@ class FlexGridGraph
   frVector<frLayerNum> zCoords_;
   frVector<frCoord> zHeights_;  // accumulated Z diff
   std::vector<dbTechLayerDir> layerRouteDirections_;
+  std::vector<bool> layerRouteDirectionsHorizontal_;
```
maliberty, line 1014: vector<bool> does bit packing which I suspect would make it slower than vector<char>.  Would you try that out and see?

maliberty, line 1014: What's the point in having both?  Why not just change layerRouteDirections_?

src/drt/src/dr/FlexGridGraph.h
```diff
@@ -1009,6 +1011,7 @@ class FlexGridGraph
   frVector<frLayerNum> zCoords_;
   frVector<frCoord> zHeights_;  // accumulated Z diff
   std::vector<dbTechLayerDir> layerRouteDirections_;
```
maliberty, line 1013: I suspect the only reason that dbTechLayerDir is slower is because the ctor is out of line and so the == comparison requires an out of line call.   I suspect inlining dbTechLayerType(Value value) would solve the problem just as well.

antonblanchard: Good idea. It turns out the copy constructor is the one we needed to inline. I ended up inlining all the simple constructors in both dbTechLayerDir and dbTechLayerType.

maliberty (approved):

rovinski:
Is this something that would be fixed by LTO? Would it be possible to add LTO only to DRT? e.g.

```
set_property(TARGET drt
             PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)
```

maliberty:
I don't think you can LTO only one part of the program afaik but I'm not certain.

rovinski:
This stack overflow post indicates it can be done: https://stackoverflow.com/questions/31355692/how-do-i-enable-link-time-optimization-lto-with-cmake

It might be worth a quick check. I might have time later this evening.

maliberty:
I suspect the target has to be an executable not a library.  Fromhttps://gcc.gnu.org/wiki/LinkTimeOptimization

This expands the scope of inter-procedural optimizations to encompass the **whole program** (or, rather, everything that is visible at link time).

QuantamHD:
LTO just keeps all the .o files in LLVM IR. This should be more than possible.

maliberty:
Perhaps someone wants to try this out.  In general I see no real reason to limit LTO to drt.  The main concern is the compile time cost of LTO.

antonblanchard:
@rovinski great idea - there are a lot of small functions at the top of the drt profile, so LTO should help quite a lot.

Issue #2220, opened 2022-08-30
compile error in spdlog 1.10.0 on M1 MacBook Air


stefanottili:
Hi Folks,

I've been successfully compiling/running openroad on a M1 MacBook Air using brew for the dependencies for a while now.
A recent "brew update" seems to have broken the compile with spdlog: 1.10.0

`[INFO FLW-0018] Compiling OpenROAD.
-- OpenROAD version: v2.0-4818-g4174c3ad8
-- System name: Darwin
-- Compiler: AppleClang 13.1.6.13160021
-- Build type: RELEASE
-- Install prefix: /Users/user/OpenROAD-flow-scripts/tools/install/OpenROAD
-- TCL library: /opt/homebrew/opt/tcl-tk/lib/libtcl8.6.dylib
-- TCL header: /opt/homebrew/opt/tcl-tk/include/tcl.h
-- Found Boost: /opt/homebrew/lib/cmake/Boost-1.79.0/BoostConfig.cmake (found version "1.79.0")  
-- spdlog: 1.10.0

[ 70%] Built target abc
In file included from /Users/user/OpenROAD-flow-scripts/tools/OpenROAD/src/dst/src/LoadBalancer.cc:34:
In file included from /Users/user/OpenROAD-flow-scripts/tools/OpenROAD/src/utl/include/utl/Logger.h:50:
In file included from /opt/homebrew/include/spdlog/spdlog.h:12:
In file included from /opt/homebrew/include/spdlog/common.h:45:
In file included from /opt/homebrew/include/spdlog/fmt/fmt.h:27:
/opt/homebrew/include/fmt/core.h:1756:3: error: static_assert failed due to requirement 'formattable' "Cannot format an argument. To make type T formattable provide a formatter<T> specialization: https://fmt.dev/latest/api.html#udt"
  static_assert(
  ^
/opt/homebrew/include/fmt/core.h:1777:10: note: in instantiation of function template specialization 'fmt::detail::make_value<fmt::basic_format_context<fmt::appender, char>, const boost::asio::ip::address &>' requested here
  return make_value<Context>(val);
         ^
/opt/homebrew/include/fmt/core.h:1899:23: note: in instantiation of function template specialization 'fmt::detail::make_arg<true, fmt::basic_format_context<fmt::appender, char>, fmt::detail::type::custom_type, const boost::asio::ip::address &, 0>' requested here
        data_{detail::make_arg<
                      ^
[...]

PR #2240, opened 2022-09-03
Update DependencyInstaller.sh to support MacOS

### LINKED ISSUE #2173: Add MacOS to dependency installer
It looks like we have the steps to install dependencies on MacOS in a GitHub action:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/f0685da2f7304ba8d0871377086486522990485a/.github/workflows/github-actions-macos.yml#L8-L45

Is it possible to move this to [`etc/DependencyInstaller.sh`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/f0685da2f7304ba8d0871377086486522990485a/etc/DependencyInstaller.sh)?

I don't have immediate access to MacOS, but I can probably do so soon for a PR.

rovinski:
Draft PR to support MacOS. Closes #2173.

This isn't a perfect solution to the problem, but I think it's a good start. There are several issues I worked around while making this.

* Problem: Homebrew's `spdlog` does not work out of the box as of right now.
* Cause: This has been mentioned a few times in issues such as #2173, #2220, and #2097, but the main issue seems to be that `spdlog` is dependent on `fmt`. While `fmt 8` works with OpenROAD's code, `fmt 9` has more strict standards for conversions of objects into strings for logging, thus resulting in several compiler errors.
* Solution: I think that longer term, we simply have to fix all the object to string conversions so that we can upgrade to `spdlog 1.10` which depends on `fmt 9`. This is what Homebrew wants to do, and it makes installing old package versions as difficult as possible. This PR works around this by forcibly installing `fmt 8.1.1_1` and `spdlog 1.9.2`, which I have found to work.

---

* Problem: Some dependencies don't take precedence in the `PATH` over system-installed libraries, which can cause compilation errors (such as bison)
* ~~Cause: `/usr/bin` tends to be in front of `/usr/local`~~ Homebrew refuses to link libs which are already provided by MacOS
* Solution: For right now, I just have the script print commands for the user to copy+paste to update the `PATH`. This is a dependency script, so it's not expected to build `openroad` for the user.

---

* Problem: Xcode Command Line Tools and the Xcode app seem to conflict on my system when both are installed.
* Cause: Not sure. I have consistently found that openroad has std lib reference issues *if* the Command Line Tools are installed (e.g. `xcode-select --install`), regardless of whether the Xcode app is installed. I can only compile if *only* the Xcode app is installed. I have checked `CMakeCache.txt` and cmake appears to be pointing to all the right directories (no mixing between the CLT and the app), so I don't know why this fails. Annoyingly, homebrew refuses to install Python if it doesn't detect CLT (even if the app is installed).
* My current workaround is to install CLT to install the dependencies, then remove CLT and install the Xcode app to actually build openroad. I sincerely hope this is just a problem on my side and not typical.

The script was tested on an x86_64 macbook running `Monterey 12.5.1` with `Apple clang version 13.1.6 (clang-1316.0.21.2.5)`

TODO items
- [x] Separate packages into runtime and development
- [x] See if someone can replicate my issues
- [ ] Add package(s) to enable GUI

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -200,6 +200,57 @@ _installCentosRuntime() {
     yum update -y
 }
 
+_installHomebrewPackage() {
+    package=$1
+    commit=$2
+    url=https://raw.githubusercontent.com/Homebrew/homebrew-core/${commit}/Formula/${package}.rb
+    curl -L ${url} > ${package}.rb
+
+    if brew list $package &> /dev/null
+        then
+        # Homebrew is awful at letting you use the version you want if a newer
+        # version is installed. The package must be completely removed to ensure
+        # only the correct version is installed
+        brew remove --force --ignore-dependencies $package
+    fi
+
+    # Must ignore dependencies to avoid automatic upgrade
+    export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+    brew install --ignore-dependencies --formula ./${package}.rb
+    brew pin ${package}
+
+    # Cleanup
+    rm ./${package}.rb
+}
+
+_installDarwin() {
+    if ! command -v brew &> /dev/null
+      then
+      echo "Homebrew is not found. Please install homebrew before continuing."
+      exit 1
+      fi
+    if ! xcode-select -p &> /dev/null
+      then
+      # xcode-select does not pause execution, so the user must handle it
+      cat <<EOF
+Xcode command line tools not installed.
+Run the following command to install them:
+  xcode-select --install
+Then, rerun this script.
+EOF
+      exit 1
+    fi
+    brew install bison boost cmake eigen flex libomp pyqt5 python swig tcl-tk zlib
+
+    # Lemon is not in the homebrew-core repo
+    brew install The-OpenROAD-Project/lemon-graph/lemon-graph
+
+    # Install fmt 8.1.1 because fmt 9 causes compile errors
+    _installHomebrewPackage "fmt" "8643c850826702923f02d289e0f93a3b4433741b"
+    # Install spdlog 1.9.2
+    _installHomebrewPackage "spdlog" "0974b8721f2f349ed4a47a403323237e46f95ca0"
+}
+
 _help() {
     cat <<EOF
 
@@ -242,6 +293,9 @@ case "${platform}" in
             os="Unidentified OS, could not find /etc/os-release."
         fi
         ;;
+    "Darwin" )
+        os="Darwin"
+        ;;
     *)
         echo "${platform} is not supported" >&2
         echo "We only officially support Linux at the moment." >&2
@@ -274,6 +328,16 @@ EOF
             _installCommonDev
         fi
         _installUbuntuCleanUp
+        ;;
+    "Darwin" )
+        _installDarwin
+        cat <<EOF
+
+To install or run openroad, update your path with:
+    export PATH="\$(brew --prefix bison)/bin:\$(brew --prefix flex)/bin:\$(brew --prefix tcl-tk)/bin:\$PATH"
+
+You may wish to add this line to your .bashrc file
+EOF
         ;;
     *)
         echo "unsupported system: ${os}" >&2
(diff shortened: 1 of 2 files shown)


maliberty:
is this ready for review?

rovinski:
Not quite, I'm hoping to get someone else with MacOS to test it out to make sure it works as expected. Also I need to address the TODO items.

If you have feedback, you are welcome to comment though.

maliberty:
I checked out spdlog 1.10.0 on my usual linux machine and OR builds without error.  Are you using an external fmt library?

rovinski:
Yes, Homebrew packages `fmt` separately from `spdlog`. There isn't a good way to get `spdlog` to build with the bundled `fmt` unless it's built from source.

maliberty:
I was able to reproduce it.  I find the changes to fmt to be horrible...

rovinski:
I was able to confirm that at least 1 other person with a macbook was able to build using the script.

Maybe I can look into `FMT_DEPRECATED_OSTREAM` from https://github.com/fmtlib/fmt/issues/3081#issuecomment-1238789389 and resolve the spdlog version issue in a separate PR.

maliberty:
lgtm, I'll leave it to @vvbandeira to approve.

stefanottili:
Having the right fmt/spdlog combination installed with HomeBrew on a MacBook Air M1 Darwing gets OpenRoad to compile again, thanks.

etc/DependencyInstaller.sh
```diff
@@ -200,6 +200,59 @@ _installCentosRuntime() {
     yum update -y
 }
 
+_installDarwinCleanUp() {
+    echo "";
+}
+
+_installDarwinDevelopment() {
+    echo "";
+}
```
vvbandeira, line 209: Let's just remove

etc/DependencyInstaller.sh
```diff
@@ -200,6 +200,59 @@ _installCentosRuntime() {
...
+    # Must ignore dependencies to avoid automatic upgrade
+    export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+    brew install --ignore-dependencies --formula ./${package}.rb
+    brew pin ${package}
+
+    # Cleanup
+    rm ./${package}.rb
+}
+
+_installDarwinRuntime() {
```
vvbandeira, line 234: We can have just a `_installDarwin` for macOS for sake of simplicity.

etc/DependencyInstaller.sh
```diff
@@ -200,6 +200,59 @@ _installCentosRuntime() {
...
+      echo "Xcode command line tools not installed."
+      echo "Run the following command to install them:"
+      echo "  xcode-select --install"
+      echo "Then, rerun this script."
+      exit 1
+    fi
+    brew install bison boost cmake eigen flex libomp pyqt5 python swig tcl-tk zlib
+
+    # TODO fork to openroad organization
+    brew install donn/lemon-graph/lemon-graph
```
vvbandeira, line 247: Tom already forked, could you update to https://github.com/The-OpenROAD-Project/homebrew-lemon-graph

etc/DependencyInstaller.sh
```diff
@@ -242,6 +295,10 @@ case "${platform}" in
             os="Unidentified OS, could not find /etc/os-release."
         fi
         ;;
+    "Darwin" )
+        #os=$(sw_vers -productVersion)
```
vvbandeira, line 299: remove commented out code

etc/DependencyInstaller.sh
```diff
@@ -275,6 +332,23 @@ EOF
         fi
         _installUbuntuCleanUp
         ;;
+    "Darwin" )
+        _installDarwinRuntime
+        # TODO what is needed in dev vs. runtime?
+        #if [[ "${option}" == "dev" ]]; then
+        #    _installDarwinDev
+        #fi
+        #_installDarwinCleanUp
```
vvbandeira, line 341: remove, see above about only having `_installDarwin`

vvbandeira (changes requested):

rovinski:
@vvbandeira can you check on the CI? There seems to be some instability. I was failing CTS tests, then I reran, and now I'm failing read_db even though this PR doesn't touch any source code.

I will work on the change requests.

rovinski:
I addressed the comments and I also updated `FindTCL.cmake` so that it is no longer required to use `-DTCL_LIB_PATHS` or `-DTCL_NO_DEFAULT_PATH`. The issue was that FindTCL.cmake was only looking for the default path on Apple M1 silicon but not for Intel silicon.

vvbandeira (approved):
Thanks for the changes.

PR #2245, opened 2022-09-06
Multi power domain ODB Support

ahmadelrouby:
- Added limited support for reading UPF commands in OpenROAD.
- ODB Structures to hold information regarding multi power domains
- TCL interface allowing querying of that information.
- Test case provided for reading upf into odb and querying info.

This is still in the early phases of development and should go through multiple cycles of development.


src/odb/include/odb/db.h
```diff
@@ -9223,6 +9271,116 @@ class dbAccessPoint : public dbObject
...
 };
 
+class dbPowerDomain : public dbObject
+{
+ public:
+  // User Code Begin dbPowerDomainEnums
+  // User Code End dbPowerDomainEnums
+  const char* getName() const;
+
+  void setPowerSwitch(dbPowerSwitch* power_switch);
```
Colin-Holehouse, line 9281: A power domain is not limited to only having one power switch associated with it. It's better to say that there are 0 or more power switches for a given power domain.

Colin-Holehouse (changes requested):
A power domain has 0 or more power switches defined for it, and has 0 or more isolation rules defined for it.

src/odb/include/odb/db.h
```diff
@@ -9223,6 +9271,116 @@ class dbAccessPoint : public dbObject
...
+
+  dbIsolation* getIsolation() const;
+
+  // User Code Begin dbPowerDomain
+  static dbPowerDomain* create(dbBlock* block,
+                               const char* name);
+  static void destroy(dbPowerDomain* pd);
+
+  void addElement(const char* element);
+  std::vector<std::string> getElements();
```
arlpetergadfort, line 9295: Shouldn't this return the list of matching elements?

src/OpenRoad.cc
```diff
@@ -237,6 +239,8 @@ OpenRoad::init(Tcl_Interp *tcl_interp)
   initLogger(logger_, tcl_interp);
   initGui(this); // first so we can register our sink with the logger
   Odbtcl_Init(tcl_interp);
+  Upf_Init(tcl_interp);
+  evalTclInit(tcl_interp, sta::upf_tcl_inits);
```
arlpetergadfort, line 243: Should this be a tool on it's own? ie. with a `MakeUPF()`

src/odb/src/db/dbIsolation.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+void dbIsolation::setIsolationSense(const char* isolation_sense){
+  _dbIsolation* obj = (_dbIsolation*) this;
+  obj->_isolation_sense = strdup(isolation_sense);
+}
+
+void dbIsolation::setLocation(const char* location){
+  _dbIsolation* obj = (_dbIsolation*) this;
+  obj->_location = strdup(location);
+}
```
arlpetergadfort, line 282: the use of `strdu` seems like it would leak if these values get set again

src/OpenRoad.cc
maliberty: I told Ahmad he could fix this in another PR as we already have several cases here that need fixing.

src/odb/src/db/dbIsolation.cpp
maliberty: Using std::string would be better

maliberty:
Please don't add copies of pdk files.  They already exist in the repo and a symlink can be created.  See how rsz/test is setup for an example.

src/odb/src/upf/upf.tcl
```diff
@@ -0,0 +1,189 @@
+############################################################################
+##
+## Copyright (c) 2019, The Regents of the University of California
```
maliberty, line 3: 2022 here and any new files elsewhere

src/odb/src/upf/upf.tcl
```diff
@@ -0,0 +1,189 @@
...
+sta::define_cmd_args "read_upf" { [-file file] }
+proc read_upf { args } {
+    sta::parse_key_args "read_upf" args \
+        keys {-file} flags {}
+
+    source $keys(-file)
+
+}
+
+sta::define_cmd_args "create_power_domain" { name [-elements elements] }
```
maliberty, line 45: Flags should come before positional args (here and elsewhere)

src/odb/src/upf/upf.tcl
```diff
@@ -0,0 +1,189 @@
...
+proc create_power_domain { args } {
+    sta::parse_key_args "create_power_domain" args \
+        keys {-elements} flags {}
+
+    set domain_name [lindex $args 0]
```
maliberty, line 50: Add 
```
sta::check_argc_eq1 "create_power_domain" $args
```
to make sure there is an argument otherwise lindex will generate an error.  Same elsewhere.

src/odb/src/upf/upf.cpp
```diff
@@ -0,0 +1,160 @@
...
+  return dbPowerDomain::create(block, name) != nullptr;
+}
+
+bool upf::update_power_domain(utl::Logger* logger,
+                              dbBlock* block,
+                              const char* name,
+                              const char* elements)
+{
+  dbPowerDomain* pd = block->findPowerDomain(name);
+  if(pd != nullptr) {
```
maliberty, line 57: clang-format

src/odb/src/upf/upf.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+bool upf::update_power_switch_control(utl::Logger* logger,
+                                      dbBlock* block,
+                                      const char* name,
+                                      const char* control_port)
+{
+  dbPowerSwitch* ps = block->findPowerSwitch(name);
+  if(ps == nullptr) return false;
+  ps->addControlPort(control_port);
+  return true;
```
maliberty, line 100: The return values in update functions are ignored in the .i.  I think it would be better to issue a warning message here than put the handling in the caller.  Same with the other updates

src/odb/include/odb/upf.h
```diff
@@ -0,0 +1,96 @@
...
+#include <list>
+#include <string>
+#include <vector>
+
+#include "db.h"
+#include "odb.h"
+
+namespace odb {
+
+class upf
```
maliberty, line 50: With only static functions this is a namespace not a class.

src/odb/src/codeGenerator/schema.json
```diff
@@ -143,6 +143,181 @@
...
     {
       "src":"dbAccessPoint.json"
+    },
+    {
+      "name":"dbPowerDomain",
+      "type":"dbObject",
+      "fields":[
+        {
+          "name":"_name",
+          "type":"char *",
```
maliberty, line 153: We should use std::string rather than char*

maliberty (changes requested):

src/odb/src/codeGenerator/schema.json
maliberty: _name needs to be a char* due to the dbHashTable but the others were converted to std::string

src/odb/include/odb/db.h
maliberty: I'm not sure what you mean here?  Do you mean the associated dbInst?  I think that will come in the future as we flesh out the usage of UPF.

maliberty (approved):
I'm sure there will be a lot of revision to this but hopefully it enables us to start implementing and learning how we need to refine

Colin-Holehouse (approved):

PR #2254, opened 2022-09-08
Grt improve repair antennas

luis201420:


src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -309,6 +309,11 @@ void FastRouteCore::clearRoute(const int netID)
   newRipupNet(netID);
 }
 
+void FastRouteCore::clearAllRoute(const int netID)
+{
+  releaseResourcesNet(netID);
```
maliberty, line 314: Why the indirection?

src/grt/src/fastroute/src/RipUp.cpp
```diff
@@ -443,6 +443,53 @@ bool FastRouteCore::newRipup3DType3(const int netID, const int edgeID)
...
+
+    for (int i = 0; i < routeLen; i++) {
+
+      if(gridsL[i] != gridsL[i + 1])
+        continue;
+
+      else if (gridsX[i] == gridsX[i + 1]) {  // a vertical edge
+
+        const int ymin = std::min(gridsY[i], gridsY[i + 1]);
+        if (v_edges_[ymin][gridsX[i]].usage >= edgeCost)
```
maliberty, line 468: Get v_edges_[ymin][gridsX[i]] and store it in a reference variable instead of repeatedly looking it up.  Same in the horiz case.  I think you could collapse the H/V cases by doing so

maliberty, line 468: Is it an error if the usage doesn't exceed the edgeCost?

luis201420: When the usage does not exceed the edgeCost and we subtract it, it becomes (~2^16), this is because the usage variable is unsigned.

maliberty: When is usage less than edgeCost?

maliberty: Please answer this question as this seems like a bug that usage would be less than the cost of a routed edge.

maliberty:
Do you have any measurement of the runtime improvements for small changes?

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -309,6 +325,16 @@ void FastRouteCore::clearRoute(const int netID)
   newRipupNet(netID);
 }
 
+void FastRouteCore::clearAllRoute(const int netID)
```
eder-matheus, line 328: Perhaps `clearNetRoute` is a better name.

src/grt/src/fastroute/src/RipUp.cpp
```diff
@@ -443,6 +445,47 @@ bool FastRouteCore::newRipup3DType3(const int netID, const int edgeID)
   return true;
 }
 
+void FastRouteCore::releaseResourcesNet(const int netID)
```
eder-matheus, line 448: `releaseNetResources`?

src/grt/src/fastroute/src/RipUp.cpp
```diff
@@ -443,6 +445,47 @@ bool FastRouteCore::newRipup3DType3(const int netID, const int edgeID)
...
+    Edge* edge;
+    Edge3D* edge_3D;
+
+    for (int i = 0; i < routeLen; i++) {
+
+      if(gridsL[i] != gridsL[i + 1])
+        continue;
+
+      else if (gridsX[i] == gridsX[i + 1]) {  // a vertical edge
+
```
eder-matheus, line 470: clang-format here. You don't need the new line.

eder-matheus (changes requested):

src/grt/src/fastroute/src/RipUp.cpp
luis201420: There were parts of the code that were not updating the usage value, I updated and fixed the bug.

maliberty:
@vvbandeira there is something very odd going on in the CI here.  Please take a look when you can.

maliberty:
@luis201420 In the build I see the error:
```
 > git rev-parse --verify HEAD # timeout=10
No valid HEAD. Skipping the resetting
```
https://confluence.atlassian.com/fishkb/setting-a-valid-head-on-your-git-repository-260772024.html talks about how to fix the HEAD

PR #2258, opened 2022-09-09
Adding obstructions shapes to write_abstract_lef

niv280:
The write_abstract_left is dumping the shapes rects obstruction instead of the whole layer rect.
The new steps:

1. collect all the shapes per layer
2. Bloat all the shapes
3. Merged all the shapes if overlap ( per layer)
4. Shrink back the merged polygon
5. Decomposed the polygon into rects

( The process is in the image below)

The size of the bloat can be determine by the user ( with the -bloat flag), adding it also in the tcl procs, while the default amount is 2*pitch.

This is my first PR, so if there is any comment, suggest for improvement please let me know and I will work to improve it.

image


QuantamHD:
Super cool stuff! I wrote the original version of this code, and it's so cool to see it improved

arlpetergadfort (changes requested):
I'll take a first stab at a review. This was just a cursory check. You will definitely need to include a unit test to ensure this works.
Is there a particular benefit derived from merging the shapes?

maliberty:
I wrote an abtrace for ng45/gcd and I see
```
	 LAYER metal4 ;  
	   RECT 0.0 0.0 70.11 70.0 ;  
	 LAYER metal5 ;  
	   RECT 0.0 0.0 70.11 70.0 ;  
	 LAYER metal6 ;  
	   RECT 0.0 0.0 70.11 70.0 ;  
	 LAYER metal7 ;  
	   RECT 0.0 0.0 70.11 70.0 ;  
	 LAYER metal8 ;  
	   RECT 0.0 0.0 70.11 70.0 ;  
	 LAYER metal9 ;  
	   RECT 0.0 0.0 70.11 70.0 ;  
	 LAYER metal10 ;  
	   RECT 0.0 0.0 70.11 70.0 ;  
```
which makes no sense since those layer are mostly empty but here they are fully blocked.  How have you tested this?

src/odb/src/lefout/lefout.cpp
```diff
@@ -135,32 +137,104 @@ void lefout::writeHeader(dbBlock* db_block)
...
+    } else if (strcasecmp(tech_layer->getDirection().getString(),"VERTICAL")) {
+      polySet.get_rectangles(rects, boost::polygon::HORIZONTAL); 
+    } else if (strcasecmp(tech_layer->getDirection().getString(),"NONE")) {
+      polySet.get_rectangles(rects); 
+    }
+
+    // Add the new rects to the obstructions map
+    obstructions.erase(tech_layer); 
+    for (const auto& rect : rects) {
+      dbObstruction* new_obs = dbObstruction::create(block,
```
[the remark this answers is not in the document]
maliberty: This seems to be a mistake and a local type should be used as these are just transient to the construction of the final OBS

src/odb/src/lefout/lefout.cpp
```diff
@@ -135,32 +137,104 @@ void lefout::writeHeader(dbBlock* db_block)
...
+
+    int pitch = tech_layer->getPitch();
+
+    // Compose all the rects in the layer to merged polygon with bloat
+    boost::polygon::polygon_90_set_data<int> polySet;
+    for (dbObstruction* obs : obstructions_per_layer) {
+      Rect rect = obs->getBBox()->getBox();
+      boost::polygon::polygon_90_set_data<int> poly;
+      poly = boost::polygon::rectangle_data<int>{rect.xMin(), rect.yMin(), rect.xMax(), rect.yMax()};
+      polySet += poly.bloat(bloat*pitch,bloat*pitch,bloat*pitch,bloat*pitch);
```
maliberty, line 180: compute bloat*pitch once and put it in a const variable

maliberty:
You are pretty close to having it right.  Thanks for working on this.

src/odb/src/lefout/lefout.cpp
```diff
@@ -36,6 +36,8 @@
 #include <stdio.h>
 
 #include <algorithm>
+#include <boost/polygon/polygon.hpp>
+#include <iostream>
```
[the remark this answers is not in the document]
niv280: No need, that was for debug, I will remove it.

src/odb/src/lefout/lefout.cpp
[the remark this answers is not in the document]
niv280: Yes, It was easy to started with the dbObstruction object just because it contain dbBox object and so I can use the writeBox without adding new write function.

I changed the values of the obstructions map to contain Rect objects and I modify the writeBox function to work with dbBox and Rect, I hoping it's ok in that way

niv280:
Hi, I have fixed the issues that you mention.

src/odb/src/lefout/lefout.cpp
```diff
@@ -135,32 +138,81 @@ void lefout::writeHeader(dbBlock* db_block)
   writeUnits(/*database_units = */db_block->getDbUnitsPerMicron());
 }
 
-void lefout::writeObstructions(dbBlock* db_block)
+void lefout::writeObstructions(dbBlock* db_block, int bloat_factor)
 {
-  std::set<dbTechLayer*> obstruction_layers;
-  getTechLayerObstructions(db_block, obstruction_layers);
+  std::map<dbTechLayer*,std::set<Rect*>> obstructions;
```
maliberty, line 143: Just use a std::set<Rect> and drop the indirection.  As written the code will leak all the Rect objects.

src/OpenRoad.tcl
```diff
@@ -111,14 +111,19 @@ proc write_def { args } {
...
 }
 
-sta::define_cmd_args "write_abstract_lef" {filename}
+sta::define_cmd_args "write_abstract_lef" {[-bloat amount] filename}
 
 proc write_abstract_lef { args } {
-  sta::parse_key_args "write_abstract_lef" args keys {} flags {}
+  sta::parse_key_args "write_abstract_lef" args keys {-bloat} flags {}
+
+  set bloat 2
```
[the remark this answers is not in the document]
maliberty: I disagree on this point.  The current behavior is not equivalent to zero and it isn't a good default as it will produce overly detailed abstract (really no abstraction at all).
@arlpetergadfort would you comment further on your thoughts?

maliberty:
for ng45/gcd the blockage for metal3 under various bloat choices:
0: 
image

2: 
image

5:
image

15:
image

The last looks like a reasonable model.  I think the others are too detailed.  I'm curious what others think...

src/odb/src/lefout/lefout.cpp
```diff
@@ -135,32 +138,81 @@ void lefout::writeHeader(dbBlock* db_block)
...
+    }
+  }
+}
+
+void lefout::getObstructions(
     dbBlock* db_block,
-    std::set<dbTechLayer*>& obstruction_layers) const
+    std::map<dbTechLayer*, std::set<Rect*>>& obstructions) const
 {
   for (dbNet* net : db_block->getNets()) {
```
maliberty, line 208: There is a pre-existing issue here that becomes more important with this change.  You are not picking up and shapes from instances that acts as obs as well.  You can see it by looking at m1 which only shows the power straps.

You should add a loop over db_block->getInsts() and add the pins and obs shapes to obstructions

maliberty:
Rather than gather all the obs and then merge them it would be better to merge them as you go.  The set of objects to be merged could be huge on a large design.

src/odb/src/lefout/lefout.cpp
niv280: I don't sure that I understand, can you elaborate?

maliberty: You calling 'new Rect' but never delete on those objects. 

However I think it would be best to drop Rect altogether and use polygon_90_set_data<int> here and directly add the shapes as I mentioned elsewhere.

niv280: Got it, yes in that way I could merge the rects on the way and not in the end as you wrote below.

maliberty: Note that you could bloat the rects as you add them rather than bloating the whole set at the end.  The shapes will merge earlier that way potentially reducing memory usage.

rovinski:
This is an interesting abstract strategy but generally it seems undesirable. For most macros being abstracted, you want to mark any internal layers as "cover" blockages - the entire layer is covered with a blockage if any shapes exist on that layer. The reasons for this are that:

1, Computation is easier for the router
2. The LEF file size is reduced
3. It disallows routing near the internal nets of the macro, which could cause signal integrity violations that won't be caught by almost any other means.
4. Most macros should be metal filled before abstracting to ensure that the extracted parasitics and therefore timing models are accurate. In this case, the layer will be entirely blocked anyways.

It seems this PR changes the default abstraction method to bloating. I definitely think that if there is no bloating, the old method (which I believe is "all layers are cover layers") should be used.

src/OpenRoad.tcl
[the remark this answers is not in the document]
arlpetergadfort: @maliberty i was wrong. I agree with @rovinski that the 0 bloat behavior should revert to the current implementation. That is what I meant.

arlpetergadfort:
> for ng45/gcd the blockage for metal3 under various bloat choices: 0: image
> 
> 2: image
> 
> 5: image
> 
> 15: image
> 
> The last looks like a reasonable model. I think the others are too detailed. I'm curious what others think...

15 looks better. Is there a good way to determine the largest bloat factor that would basically create a single sheet for each layer?

maliberty:
I think it is reasonable to have an option to just block the entire area as we do currently.  I think that is a different option than -bloat_factor and a different 'algorithm'.

maliberty:
1, Computation is easier for the router
> the point of the bloating is to trade off simplicity versus accuracy.  Its why I objected to bloat 0.
2. The LEF file size is reduced
> Same as above
3. It disallows routing near the internal nets of the macro, which could cause signal integrity violations that won't be caught by almost any other means.
> We can easily add a final bloat option to increase the spacing as desired.
4. Most macros should be metal filled before abstracting to ensure that the extracted parasitics and therefore timing models are accurate. In this case, the layer will be entirely blocked anyways.
> Not if you want to be able to route through the area in question which is a goal here.  Additional spacing / shielding removes most of the RC extraction concerns (inter-layer effects are still possible but can be conservatively modelled as they are smaller).

rovinski:
I'm not objecting to the functionality as long as the default behavior of `write_lef_abstract` is to generate obstructions for the whole layer. I think bloating is very useful in perhaps a few circumstances.

An ideal interface to me would look like
```
write_lef_abstract [-bloat_factor bloat_factor [-layers {layers}]]
```
Examples:
```
# Write an abstract LEF with cover obstructions on each layer
# There are still cutouts for pins
write_lef_abstract

# Write an abstract LEF with bloat factor of 15 on each layer
write_lef_abstract -bloat 15

# Write an abstract LEF with bloat factor of 15 on M4, M5
# All other layers use cover obstructions
write_lef_abstract -bloat 15 -layers {M4 M5}

# Write an abstract LEF with bloat factor of 15 on M4, M5
# Bloat factor of 0 on M6, M7
# All other layers use cover obstructions
write_lef_abstract -bloat 15 -layers {M4 M5} -bloat 0 -layers {M6 M7}
```

maliberty:
There is no need for pin cutouts unless they are not on the macro edge

rovinski:
> There is no need for pin cutouts unless they are not on the macro edge

It would be nice for consistency/compatibility with other abstracting/P&R tools. It is standard for IP to have pin cutouts.

maliberty:
> > There is no need for pin cutouts unless they are not on the macro edge
> 
> It would be nice for consistency/compatibility with other abstracting/P&R tools. It is standard for IP to have pin cutouts.

It is perhaps common but it isn't standard - you can see plenty of cases where that isn't true.  lef/def specifically allow you to access pins covered by an obs.

maliberty:
Do we really need layer specific bloating?  I'd rather not add complexity unless someone will actually use it.

rovinski:
What if someone doesn't want to allow routing on lower layers but is fine with routing on upper layers?

An example: top-level power mesh is on M6-M7. Signal routing is on M1-M5. Designer is fine with allowing external signals to pass through on M6-M7 but doesn't want signals passing through M1-M5.

Doing all-or-nothing bloating for the whole block seems like kind of extreme.

rovinski:
Just speaking personally, I would use this feature if I could free up routing resources on upper layers. I would probably not use it if I couldn't block all of the lower layers.

maliberty:
Ok if there is value we should add it.  I am also ok to push it into a follow on PR.

QuantamHD:
You should update the src/README as this change would make the warning there obsolete.

src/odb/src/lefout/lefout.cpp
niv280: I should block the entire inst or only his obs with inst->getBBox()->getObstructions?

niv280: The obstructions of the insts is the nets/wires of the insts? same as the nets of the block?

maliberty: The instances are in addition to the nets.  pins & obs of instances also block routing just like nets do.

maliberty: For example in this area these are the shapes belonging to nets:
image
but there are also these shapes belonging to the instance:
image

niv280:
Hi,
I did some changes:

1.  Merge the polygon as we go, which supposed to easier for large design.
2. I added insts ITerms to the obstructions
3. Added the insts obstructions to the lef ( I am not sure about this one, please let me know if that what you meant )

There was other discussion about the this, but I don't understand what the conclusion, If you rather to do some change of the bloat_factor default please let me know.

maliberty:
There is a request to have an option for the old behavior.  I don't like calling that '-bloat 0' and suggest we just add a different option like -block_occupied_layers.  In a later PR we can allow users to choose the new or old style on a per layer basis.

niv280:
So what is the next step here?
Should I add this option to this PR and the next one?
Any change needed regarding the previous requests?

maliberty:
I think you should add that change here as otherwise it would represent a loss of functionality for those who prefer the older style.   Also you need to fix the CI errors.

niv280:
I added the bloat_occupied_layers option, to cover the whole layer same as the previous functionality.

I would like to get some help with the CI errors, it seems that the build of the project are ok but it failed on certain test ( TestGuide) with undefine reference for dbITermShapeItr.
I don't understand why it's failing since all the correct include are there and the build don't throw error on this.

maliberty:
The dependencies in odb are off.  I'll fix that up and you can merge afterwards.

maliberty:
If you merge master to your branch this problem should be fixed (after you resolve the merge conflicts).

niv280:
I merge with master and it looks like it passing now, except the lef test that failed as expected.

maliberty:
For the LEF tests - please verify that the new results are correct and update the test.  You can use the save_ok script to do so.

niv280:
Added the new LEF for test

src/odb/test/gcd_abstract_lef.ok
```diff
@@ -10,4 +10,6 @@
 [INFO ODB-0132]     Created 2 special nets and 3754 connections.
 [INFO ODB-0133]     Created 439 nets and 1193 connections.
 [INFO ODB-0134] Finished DEF file: data/gcd/gcd_nangate45_route.def
-No differences found.
+Differences found at line 808.
+     RECT  10.07 34.915 20.56 35.085 ;
+    RECT  0 0 100.13 100.8 ;
```
rovinski, line 15: You need to update the lefok file, The log should not have any differences found.

src/odb/test/gcd_abstract_lef_with_power.ok
```diff
@@ -10,4 +10,6 @@
 [INFO ODB-0132]     Created 2 special nets and 3754 connections.
 [INFO ODB-0133]     Created 439 nets and 1193 connections.
 [INFO ODB-0134] Finished DEF file: data/gcd/gcd_nangate45_route_with_power_pins.def
-No differences found.
+Differences found at line 462.
+     RECT  10.07 34.915 20.56 35.085 ;
+    RECT  0 0 100.13 100.8 ;
```
rovinski, line 15: Same

niv280:
I used the save_ok script to change the log too.
But the DCO check always failed even when I use git commit -s, any idea why?

maliberty:
Most likely you have a commit that was unsigned.  DCO checks all commits.

niv280:
Thank, it worked.
The commit pass the tests.

src/odb/test/cpp/TestGuide.cpp
```diff
@@ -11,6 +11,7 @@ BOOST_AUTO_TEST_SUITE(test_suite)
 
 BOOST_AUTO_TEST_CASE(test_default)
 {
+  cout << "here" << endl;
```
maliberty, line 14: Remove

maliberty:
remove src/odb/src/swig/python/.CMakeLists.txt.swo & src/odb/src/swig/common/.odb.i.swp

src/odb/include/odb/db.h
```diff
@@ -1009,8 +1009,8 @@ class dbBlock : public dbObject
   //
   // Utility to save_lef
   //
-  // void dbBlock::saveLef(char *filename);
-  void saveLef(char* filename);
+  // void dbBlock::saveLef(char *filename, int bloat, bool bloat_occupied_layers);
```
maliberty, line 1012: remove commented line

maliberty (changes requested):

maliberty:
re: the discussion of having an option for the old behavior - is that implemented?

niv280:
Done.

> re: the discussion of having an option for the old behavior - is that implemented?

Yes, I added the option to use the bloat_occupied_layers flag, and in that way the whole Layer will be cover in the OBS.
I tested it and the results is the same as the previous lef.

maliberty (approved):

src/OpenRoad.tcl
```diff
@@ -121,14 +122,23 @@ proc write_lef { args } {
   ord::write_lef_cmd $filename
 }
 
-sta::define_cmd_args "write_abstract_lef" {filename}
-
 proc write_abstract_lef { args } {
-  sta::parse_key_args "write_abstract_lef" args keys {} flags {}
+  sta::parse_key_args "write_abstract_lef" args keys {-bloat_factor} flags {-bloat_occupied_layers}
+
+  set bloat_factor 0
```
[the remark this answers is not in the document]
niv280: I changed it to 0 as someone else suggested here, but that was before the additional option for the old behavior.
I changed it back to two tracks, do you want it to be bigger value?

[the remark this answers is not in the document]
maliberty: I think that's ok for now.  I expect we will want to make some post-processing passes in the future that tries to optimize further  (remove holes, square up the edges, etc).

maliberty:
Changing the default makes the tests fail so you'll have to update them again

niv280:
> Changing the default makes the tests fail so you'll have to update them again

Yes, forgot about it.
I'm running it now.

maliberty:
@arlpetergadfort I believe your concern about the old mode is addressed by -bloat_occupied_layers.  Since you are out I'm going to merge and if you have further concerns we can address them when you are back.

niv280:
Thanks!
Sure I will be happy to keep contributing.
Can you suggest on what I could working on next?
I rather to work on something new other then the write_lef for now.

maliberty:
Some ideas that come to mind
- GUI performance (opengl, caching, etc)
- odb code generator could be improved / used more broadly (more automated generation of database code)
- dpo (detailed placement optimization) has a simple hard coded recipe; explore the possibilities looking for improvements
- General software improvements - Coverity issues, logger usage, valgrind
If you have specific interests there are always more things that need doing.

niv280:
dpo is sound very intersting but I hope that is something that I will able to handle with.
Also after the write_lef adventure I was starting to familiar the odb db types, so that also be an option.

BTW, there is another place that we can discuss about it?

maliberty:
I'm on the skywater slack (Matt Liberty)

rovinski:
This PR added tests for the new behavior but removed the tests for the old behavior. Should we have tests for that?

rovinski:
Also requires updates to the README about how to use the new flags.

maliberty:
> Also requires updates to the README about how to use the new flags.

@niv280 would you be willing to do a quick PR with updates to the README about the new capability?

niv280:
Yes, sorry I started it before but didn't succeed to PR in here.
I create new PR with the update README

niv280:
Sure, sorry I started to add it before but didn't understand how to add another commit after I deleted the branch.
I added it in new PR now.

PR #2267, opened 2022-09-13
odb: add support for LEF58_AREA property

eder-matheus:


src/odb/src/db/dbTechLayerAreaRule.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+bool _dbTechLayerAreaRule::operator==(const _dbTechLayerAreaRule& rhs) const
+{
+  if (flags_.except_rectangle != rhs.flags_.except_rectangle)
+    return false;
+
+  if (area != rhs.area)
+    return false;
+
+  if (except_min_witdh != rhs.except_min_witdh)
```
arlpetergadfort, line 56: Width

src/odb/src/db/dbTechLayerAreaRule.cpp
```diff
@@ -0,0 +1,357 @@
...
+  obj->area = area;
+}
+
+int dbTechLayerAreaRule::getArea() const
+{
+  _dbTechLayerAreaRule* obj = (_dbTechLayerAreaRule*) this;
+  return obj->area;
+}
+
+void dbTechLayerAreaRule::setExceptMinWitdh(int except_min_witdh)
```
arlpetergadfort, line 206: Width, here and several other functions

arlpetergadfort (changes requested):

src/odb/src/lefin/lefTechLayerAreaRuleParser.cpp
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright (c) 2019, The Regents of the University of California
```
osamahammad21, line 2: update year

src/odb/src/lefin/lefTechLayerAreaRuleParser.cpp
```diff
@@ -0,0 +1,203 @@
...
+  qi::rule<std::string::iterator, space_type> LAYER
+      = ((lit("LAYER")
+          >> _string[boost::bind(&lefTechLayerAreaRuleParser::setTrimLayer,
+                                 this,
+                                 _1,
+                                 rule,
+                                 layer,
+                                 boost::ref(incomplete_props))])
+         >> lit("OVERLAP")
+         >> double_[boost::bind(&lefTechLayerAreaRuleParser::setNumber,
```
osamahammad21, line 157: Why double_? I think this should be int_
Also I don't think there is a need for setNumber as it's just passing  the number to the function without any processing. you could use setOverlap directly.

src/odb/src/lefin/lefTechLayerAreaRuleParser.cpp
```diff
@@ -0,0 +1,203 @@
...
+                                &odb::dbTechLayerAreaRule::setOverlap)]);
+
+  qi::rule<std::string::iterator, space_type> AREA
+      = (lit("AREA") >> double_[boost::bind(&lefTechLayerAreaRuleParser::setInt,
+                                            this,
+                                            _1,
+                                            rule,
+                                            &odb::dbTechLayerAreaRule::setArea)]
+         >> -(lit("MASK")
+              >> double_[boost::bind(&lefTechLayerAreaRuleParser::setNumber,
```
osamahammad21, line 170: Ditto

src/odb/src/codeGenerator/dbTechLayerAreaRule.json
```diff
@@ -0,0 +1,67 @@
...
+    {
+      "type": "dbId<_dbTechLayer>",
+      "name": "trim_layer_",
+      "flags": [
+        "private"
+      ]
+    },
+    {
+      "type": "int",
+      "name": "overlap"
```
osamahammad21, line 51: I think this could be of 2 bits only as it's either 1 or 2

osamahammad21 (changes requested):

maliberty:
Has this been through secure-CI?

maliberty:
Is src/odb/test/data/design.odb necessary?  The next time the schema changes this test will break.  It would be better to use lef/def for stability of the test.

eder-matheus:
> Has this been through secure-CI?

I've started a run yesterday, and I saw a few metrics failures. I've started another run now.

src/odb/src/lefin/lefTechLayerAreaRuleParser.cpp
```diff
@@ -0,0 +1,191 @@
...
+bool lefTechLayerAreaRuleParser::parseSubRule(
+    std::string s,
+    odb::dbTechLayer* layer,
+    std::vector<std::pair<odb::dbObject*, std::string>>& incomplete_props)
+{
+  qi::rule<std::string::iterator, std::string(), ascii::space_type> _string;
+  _string %= lexeme[+(char_ - ' ')];
+  odb::dbTechLayerAreaRule* rule = odb::dbTechLayerAreaRule::create(layer);
+
+  qi::rule<std::string::iterator, space_type> EXCEPTEDGELENGTH
```
maliberty, line 122: Use standard naming for variables (ie snake case).  Here and below

eder-matheus:
> Is src/odb/test/data/design.odb necessary? The next time the schema changes this test will break. It would be better to use lef/def for stability of the test.

We already have lef/def tests in the unit tests. To update the read_db test, should I read lef/def, write an odb file and then read the generated odb?  Is there a way to "unload" a design using the openroad cmds?

maliberty:
> > Is src/odb/test/data/design.odb necessary? The next time the schema changes this test will break. It would be better to use lef/def for stability of the test.
> 
> We already have lef/def tests in the unit tests. To update the read_db test, should I read lef/def, write an odb file and then read the generated odb? Is there a way to "unload" a design using the openroad cmds?

Never mind I see this is the schema version test which is the only one that uses an odb.

eder-matheus:
@maliberty My secure-ci run finished with two metric errors in the private designs (gf12 ariane and gf12 swerv wrapper). Not sure if my changes cause it.

src/odb/src/lefin/lefTechLayerAreaRuleParser.cpp
osamahammad21: @eder-matheus I think Matt meant to use snake case like: exceptEdgeLength. I personally don't mind the all caps because it matches the rule option name. We can fix this in a separate PR though because it's the same case for all parsers.

osamahammad21:
>My secure-ci run finished with two metric errors in the private designs (gf12 ariane and gf12 swerv wrapper). Not sure if my changes cause it.

@vvbandeira are those tests failing on master?

vvbandeira:
> > My secure-ci run finished with two metric errors in the private designs (gf12 ariane and gf12 swerv wrapper). Not sure if my changes cause it.
> 
> @vvbandeira are those tests failing on master?

Yes, you can consider secure ci as passing.

osamahammad21 (approved):

osamahammad21:
@arlpetergadfort I've confirmed your requested changes are addressed.

arlpetergadfort (approved):

PR #2271, opened 2022-09-14
gpl: add random noise to initial locations

ramezmoussa:
During global placement, all cells are initially placed at the center. This turned out to be problematic for identical cells connected in parallel because such cells would experience the exact same wirelength and density forces and consequently they will always remain on top of one another.

This was discovered while investigating an [issue](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/576) in OpenROAD-flow-scripts where a small design (<200 cells) failed to converge because 30+ identical buffers were connected in parallel. These buffers would always end up at the same location, causing large overflow in their bin, preventing convergence.

To avoid this, we add a random noise between -1 and 1 microns to both the x and y coordinates of each cell to prevent cells from starting at the exact same position.


src/gpl/src/nesterovBase.cpp
```diff
@@ -1004,13 +1004,25 @@ NesterovBase::reset() {
...
   gCellStor_.reserve(pb_->placeInsts().size());
+
+  int dbu_per_micron = pb_->db()->getChip()->getBlock()->getDbUnitsPerMicron();
+
   for(auto& inst: pb_->placeInsts()) {
+
+    // For any cell, add a random noise between -1 and 1 microns to each of its x and y components.
+    // This is added to make it very unlikely that identical cells connected in parallel do not start at the 
+    // exact same position and consequently shadow each other throughout the entire placement process
+    inst->setLocation(inst->lx() + rand() % (2 * dbu_per_micron) - 1 * dbu_per_micron, inst->ly() +  rand() % (2 * dbu_per_micron) - 1000);
```
maliberty, line 1023: "- 1 * dbu_per_micron" vs "- 1000" is incorrect they should both be the former.  There is no need to multiple by 1.

The order of argument evaluation is not fixed by the standard and vary by platform or compiler.  It would be better to do the calls first
```
int x_offset = rand() % (2 * dbu_per_micron) - dbu_per_micron;
int y_offset = rand() % (2 * dbu_per_micron) - dbu_per_micron;
inst->setLocation(inst->lx() + x_offset, inst->ly() + y_offset);
```

uniform_int_distribution might be easier

QuantamHD:
rand calls need to use a fixed seed.

maliberty:
This looks good.  Have you tested with secure CI?

ramezmoussa:
> This looks good. Have you tested with secure CI?
Yes I ran it and I believe you had a look at the results last week and they were fine.

maliberty (approved):

PR #2292, opened 2022-09-21
drt: initial support for LEF58_AREA rule

eder-matheus:


eder-matheus:
Currently testing with some private designs and making cpp tests, but created this draft PR for early comments on the code.

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -1692,6 +1649,181 @@ void FlexGCWorker::Impl::checkMetalShape_minEnclosedArea(gcPin* pin)
   }
 }
 
+void FlexGCWorker::Impl::checkMetalShape_lef58Area(gcPin* pin)
```
maliberty, line 1652: Where is the caller?

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -1692,6 +1649,181 @@ void FlexGCWorker::Impl::checkMetalShape_minEnclosedArea(gcPin* pin)
...
+    for (auto& edges : pin->getPolygonEdges()) {
+      for (auto& edge : edges) {
+        if (edge->isFixed())
+          continue;
+      }
+    }
+
+    // add patch only when the poly is not a rect
+    if (checkMetalShape_lef58Area_exceptRectangle(poly, db_rule)) {
+      checkMetalShape_addPatch(pin, min_area);
```
maliberty, line 1705: checkMetalShape_lef58Area_exceptRectangle returns true if db_rule->isExceptRectangle() is false.  Is it right to patch if you didn't even check?  Should it return false in that case?

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -1692,6 +1649,181 @@ void FlexGCWorker::Impl::checkMetalShape_minEnclosedArea(gcPin* pin)
...
+    }
+
+    // add patch only if poly is rect and its width is less than or equal to
+    // rectWidth value on constraint
+    if (check_rect_width
```
maliberty, line 1710: What happens if someone defines a LEF58_AREA without setting RECTWIDTH or EXCEPTRECTANGLE?  Is it still checked?

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -1692,6 +1649,181 @@ void FlexGCWorker::Impl::checkMetalShape_minEnclosedArea(gcPin* pin)
...
+      auto rect = rects.back();
+      auto xLen = gtl::delta(rect, gtl::HORIZONTAL);
+      auto yLen = gtl::delta(rect, gtl::VERTICAL);
+      // Horizontal rect
+      bool apply_rect_width_area = false;
+      if (xLen > yLen) {
+        apply_rect_width_area = yLen <= min_width;
+      } else {
+        apply_rect_width_area = xLen <= min_width;
+      }
```
maliberty, line 1763: apply_rect_width_area = std::min(xLen, yLen) <= min_width

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -1692,6 +1649,181 @@ void FlexGCWorker::Impl::checkMetalShape_minEnclosedArea(gcPin* pin)
...
+    if (checkMetalShape_lef58Area_exceptRectangle(poly, db_rule)) {
+      checkMetalShape_addPatch(pin, min_area);
+    }
+
+    // add patch only if poly is rect and its width is less than or equal to
+    // rectWidth value on constraint
+    if (check_rect_width
+        && checkMetalShape_lef58Area_rectWidth(
+            poly, db_rule, check_rect_width)) {
+      checkMetalShape_addPatch(pin, min_area);
```
maliberty, line 1713: Could you patch more than once since this is inside the constraint loop?

eder-matheus: The `checkMetalShape_lef58Area_rectWidth` function the `check_rect_width` as false when the first patch for min rect width is applied, so in the later iterations, this clause will always be false.

src/drt/src/gc/FlexGC_main.cpp
eder-matheus: I have a question about this. My interpretation of the lefdef manual is that if we have a LEF58_AREA with RECTWIDTH or with EXCEPTRECTANGLE, we'll not have anything else in this specific property. Is this correct?

E.g.: this property declaration would be invalid:
PROPERTY LEF58_AREA
       "AREA 0.08 RECTWIDTH 0.05 MASK 2 ;"

In any case, the curr code is incorrect, I'll push a temp fix to call the patch when only the area is defined.

image

eder-matheus: Btw, the min area value is always checked at the beginning of the function. With my latest commit, I'm now applying the patch when only AREA is defined.

maliberty: Knowing what the LEF manual means is always half the problem.  Usually the 58 properties are a superset of the older syntax.  I don't know what it means to have and '|' with a required field.  I suspect they just forgot the [].

I can say that in one of our private pdks I see 
```
  PROPERTY LEF58_AREA "
  AREA X.XXX ; " ;
```
Its not a layer we route on but it suggests that it is allowed.

eder-matheus: Yes, I think they forgot the `[]` too. My doubt is more related to the `|` symbols. Is it implying that we can only have one of the keywords? Like, is `AREA 0.8 RECTWIDTH 0.1 EXCEPTEDGELENGTH 0.5` valid? From what I understand, it is not valid.

maliberty: I don't think it makes sense to combine those two and I don't see any examples of it in our PDKs.  I could imagine combining EXCEPTRECTANGLE with LAYER so the semantics of the '|' is unclear.  Don't stress to much over weird cases and just give an error if they happen - we can always address them if they come up.

eder-matheus: Got it. I'm finishing testing with our PDKs and will let you know when I finish.

eder-matheus:
@maliberty I've tested with our PDKs and this PR looks good, drt is catching the values properly, and the checks seem correct. I've started a secure-ci run now.

I plan to create a new PR soon for the remaining keywords of this property.

eder-matheus:
@maliberty Secure-ci is green for this update. I think this PR is ready to merge.

maliberty (approved):

PR #2295, opened 2022-09-22
Fixing Bison and Flex detection on MacOS

QuantamHD:
By default CMake cannot use the brew installed flex and bison. This change prefrences user installed bison and flex over the system provided libs.

### DIFF
--- a/.github/workflows/github-actions-macos.yml
+++ b/.github/workflows/github-actions-macos.yml
@@ -31,8 +31,6 @@ jobs:
           submodules: 'recursive'
       - name: Build OpenROAD
         run: |
-          export PATH="/usr/local/opt/bison/bin:$PATH"
-          export PATH="/usr/local/opt/flex/bin:$PATH"
           export PATH="/usr/local/opt/tcl-tk/bin:$PATH"
           export PKG_CONFIG_PATH="/usr/local/opt/tcl-tk/lib/pkgconfig"
           export MY_QT_PATH=$(find /usr/local/Cellar/qt@5/ -type d -iwholename "*cmake/Qt5")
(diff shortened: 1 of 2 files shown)


QuantamHD:
@maliberty The self hosted mac seems a bit weird with the self hosted mac at least compared to mine. By default my machine is picking up the XCode Bison. Do you know how the machine is configured?

jjcherry56:
It isn't that cmake "cannot use the brew installed flex and bison", it "does not" because the typical PATH does not include it. I have worked around this since openroad's inception by setting PATH in .bashrc
    export PATH="/opt/homebrew/opt/bison/bin:$PATH"
    export PATH="/opt/hbomebrew/opt/flex/bin:$PATH"
Just setting the BISON/FLEX executable vars would be a lot simpler and more consistent with the rest of the
build, which uses etc/DependencyInstaller.sh to install dependencies instead of cmake and already supports darwin.

Vitor setup the mac host (he is not around right now). I believe it is some docker based approximation of a mac so the config is not very realistic.

maliberty:
> Vitor setup the mac host (he is not around right now). I believe it is some docker based approximation of a mac so the config is not very realistic.

I believe we purchased a real mac and it is not a docker image.

vvbandeira:
> @maliberty The self hosted mac seems a bit weird with the self hosted mac at least compared to mine. By default my machine is picking up the XCode Bison. Do you know how the machine is configured?

@QuantamHD, there are two mutually exclusive jobs in this action.

1. The "build" which you are modifying is a GitHub-hosted machine -- we are locked with the image provided by GitHub. This machine is also ephemeral, so changes to bashrc and other configs are not propagated between builds. Thus, the best way I could find is to just export PATH and other variables in the action before building the binary.
2. The "build-self-hosted" is a physical machine we own and control. In this machine, I have setup the environment (similar to what Cherry described above) and the only commands needed are the `cmake -B build && cmake --build build`

maliberty:
@QuantamHD is there any next step here?

QuantamHD:
Don't really remember. I can close it.

PR #2303, opened 2022-09-26
Support WRONGDIRECTION Width Routing

osamahammad21:


src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2622,6 +2623,18 @@ void FlexDRWorker::processPathSeg(frMIdx startX,
                   z,
                   i - 1 >= 0 ? &points[i - 1] : nullptr,
                   i + 2 < (int) points.size() ? &points[i + 2] : nullptr);
+  } else {
```
QuantamHD, line 2626: You could make this a little bit cleaner if you made this else block into an else if, else. It would remove one level of nesting which is generally good

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2787,6 +2800,33 @@ void FlexDRWorker::setNDRStyle(drNet* net,
...
+{
+  auto layer = getTech()->getLayer(gridGraph_.getLayerNum(z));
+  if (layer->getWrongDirWidth() >= layer->getWidth())
+    return;
+  bool vertical = startX == endX;
+  bool nextVertical = next && endX == next->x();
+  bool prevVertical = prev && startX == prev->x();
+  if (layer->isHorizontal() == vertical)
+    return;
+  if (next && next->z() == z && vertical != nextVertical
```
QuantamHD, line 2820: I'd suggest turning these ifs into helper functions. They have identical structure so you could have a small function that you could call twice that takes a FlexMazeIdx* and a frMIdx and call it twice in this function

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2622,6 +2623,18 @@ void FlexDRWorker::processPathSeg(frMIdx startX,
                   z,
                   i - 1 >= 0 ? &points[i - 1] : nullptr,
                   i + 2 < (int) points.size() ? &points[i + 2] : nullptr);
+  } else {
+    if (layer->isHorizontal() == vertical)
```
rovinski, line 2627: I think something needs to get changed in the code because this line is super unintuitive. Maybe use an [`Orientation2D`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/ba39bf1a326bf1d492e7ebe11049916247f9a6a5/src/odb/include/odb/isotropy.h#L98) instead of `bool vertical`?

src/drt/src/dr/FlexDR_maze.cpp
rovinski, line 2626: Same thing as @QuantamHD mentioned.

src/drt/src/dr/FlexDR_maze.cpp
maliberty: Orient2D is good.  At least
```
 if (layer->isVertical() != seg_is_vertical)  // wrong way seg
```
[rename vertical to seg_is_vertical]

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2787,6 +2800,33 @@ void FlexDRWorker::setNDRStyle(drNet* net,
...
+                                frMIdx startY,
+                                frMIdx endY,
+                                frMIdx z,
+                                FlexMazeIdx* prev,
+                                FlexMazeIdx* next)
+{
+  auto layer = getTech()->getLayer(gridGraph_.getLayerNum(z));
+  if (layer->getWrongDirWidth() >= layer->getWidth())
+    return;
+  bool vertical = startX == endX;
```
maliberty, line 2815: is_vertical

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2787,6 +2800,33 @@ void FlexDRWorker::setNDRStyle(drNet* net,
...
+  bool nextVertical = next && endX == next->x();
+  bool prevVertical = prev && startX == prev->x();
```
maliberty, line 2817: next_is_vertical
prev_is_vertical

(variables are snake case in new code)

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2787,6 +2800,33 @@ void FlexDRWorker::setNDRStyle(drNet* net,
...
+                                frMIdx endX,
+                                frMIdx startY,
+                                frMIdx endY,
+                                frMIdx z,
+                                FlexMazeIdx* prev,
+                                FlexMazeIdx* next)
+{
+  auto layer = getTech()->getLayer(gridGraph_.getLayerNum(z));
+  if (layer->getWrongDirWidth() >= layer->getWidth())
+    return;
```
maliberty, line 2814: Why?

src/drt/src/io/io.cpp
```diff
@@ -2784,23 +2785,31 @@ void io::Writer::updateDbConn(odb::dbBlock* block, odb::dbTech* db_tech)
...
+                _wire_encoder.addPoint(begin.x(), begin.y());
             } else if (segStyle.getBeginStyle()
                        == frEndStyle(frcTruncateEndStyle)) {
-              _wire_encoder.addPoint(begin.x(), begin.y(), 0);
+              _wire_encoder.addPoint(begin.x(), begin.y(), 0, 0);
             } else if (segStyle.getBeginStyle()
                        == frEndStyle(frcVariableEndStyle)) {
               _wire_encoder.addPoint(
-                  begin.x(), begin.y(), segStyle.getBeginExt());
+                  begin.x(), begin.y(), segStyle.getBeginExt(), 0);
```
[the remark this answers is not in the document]
osamahammad21: I was planning to, but added this temporary fix in drt.

src/drt/src/dr/FlexDR_maze.cpp
osamahammad21: this function clips the right way segment extension if the wrongway segment widht is smaller to avoid minstep violations. if wrongway width is larger than default width, then the wrongway segment if present is already covering the extension of the rightway segment. so no need for clipping

maliberty:
You've got a merge conflict to resolve

src/drt/src/dr/FlexDR_maze.cpp
maliberty: if the wrongway width is large then couldn't it's 1/2 width extension go beyond the narrower wire?

osamahammad21: the wrongway segment has the same extension as the rightway segment, half of the default width. So it wouldn't be a problem.

maliberty: I'm thinking of:
image

osamahammad21: That would never happen because the wrongway segment would never go beyond the rightway segment. The wrongway segment's extension is half of the default width not the wrongway width.
wrongway drawio (1)

maliberty: No the default extension is 1/2 the wire width not the default width:
image

maliberty:
This needs finishing.  @antonblanchard would like to use this

osamahammad21:
@maliberty secure ci passes
Ready to merge

maliberty (approved):

maliberty:
@antonblanchard FYI merged

PR #2309, opened 2022-09-27
drt netlist refactor for memory optimization

habibayassin:


src/drt/src/db/obj/frInst.h
```diff
@@ -45,12 +46,13 @@ class frInst : public frRef
...
-  frInst(const frString& name, frMaster* master)
-      : name_(name), master_(master), pinAccessIdx_(0), toBeDeleted_(false)
+  frInst(frMaster* master, odb::dbInst* inst)
+      : master_(master), inst_(inst), pinAccessIdx_(0), toBeDeleted_(false)
   {
   }
   // getters
-  const frString& getName() const { return name_; }
+  odb::dbInst* getDbInst() const { return inst_; }
+  const frString getName() const { return inst_->getName(); }
```
rovinski, line 55: Passed by value so does not need to be `const`

src/drt/src/db/obj/frNet.h
```diff
@@ -50,9 +51,10 @@ class frNet : public frBlockObject
 {
  public:
   // constructors
-  frNet(const frString& in)
+  frNet()
```
rovinski, line 54: If you can construct an `frNet` with no name and no `dbNet*` then how can you disambiguate two empty `frNet`s?

If this is to support fake nets then I would suggest removing the default constructor and instead having a `frNet(bool, bool)` constructor to specify `isFakeVSSNet_` and `isFakeVDDNet_`.

src/drt/src/db/obj/frNet.h
```diff
@@ -231,8 +267,11 @@ class frNet : public frBlockObject
       max = std::max(max, CLOCK_NETS_ABS_PRIORITY);
     absPriorityLvl = max;
   }
-  bool isSpecial() const { return isSpecial_; }
-  void setIsSpecial(bool s) { isSpecial_ = s; }
+  bool isSpecial() const
+  {
+    return db_net_->isSpecial();
+    ;
```
rovinski, line 273: remove extra `;`

src/drt/src/io/io.cpp
```diff
@@ -176,7 +156,7 @@ void io::Parser::setObstructions(odb::dbBlock* block)
       continue;
     }
     frLayerNum layerNum = tech_->name2layer[layerName]->getLayerNum();
-    auto blkIn = make_unique<frBlockage>();
+    auto blkIn = make_unique<frBlockage>(blockage->getBBox());
```
rovinski, line 159: Does it make more sense to store the `dbObstruction` than the bbox of the obstruction?

src/drt/src/io/io.cpp
```diff
@@ -821,7 +801,7 @@ void io::Parser::setNets(odb::dbBlock* block)
         }
       }
     }
-    netIn->setType(net->getSigType());
+    // netIn->setType(net->getSigType());
```
rovinski, line 804: Remove commented code

src/drt/test/fixture.cpp
```diff
@@ -83,8 +86,10 @@ void Fixture::addLayer(frTechObject* tech,
 
 void Fixture::setupTech(frTechObject* tech)
 {
-  auto db = odb::dbDatabase::create();
+  // auto db = odb::dbDatabase::create();
```
rovinski, line 89: Remove

src/drt/test/fixture.cpp
```diff
@@ -83,8 +86,10 @@ void Fixture::addLayer(frTechObject* tech,
...
   db_tech = odb::dbTech::create(db);
+  db->getTech()->setDbUnitsPerMicron(1000);
+  db->getTech()->setManufacturingGrid(10);
```
rovinski, line 92: Is this duplicate?

src/drt/src/io/io.cpp
```diff
@@ -1810,9 +1771,8 @@ void io::Parser::setMacros(odb::dbDatabase* db)
             }
           }
         }
-        auto blkIn = make_unique<frBlockage>();
+        auto blkIn = make_unique<frBlockage>(obs);
```
habibayassin, line 1774: > Does it make more sense to store the `dbObstruction` than the bbox of the obstruction?
I am trying to be consistent with the object used in this code block which is of type bbox

src/drt/src/db/obj/frMaster.h
```diff
@@ -78,13 +76,49 @@ class frMaster : public frBlockObject
...
+    frBoundary bound;
+    vector<Point> points;
+    points.push_back(Point(originX, originY));
+    points.push_back(Point(sizeX, originY));
+    points.push_back(Point(sizeX, sizeY));
+    points.push_back(Point(originX, sizeY));
+    bound.setPoints(points);
+    bounds.push_back(bound);
+    return bounds;
+  }
```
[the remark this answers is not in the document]
habibayassin: I think it is a tradeoff between mem consumption and runtime cost...

PR #2310, opened 2022-09-27
drt: fix check for second layer

eder-matheus:


habibayassin:
> @habibayassin do you see any issue here?

no it seems okay

Issue #2308, opened 2022-09-27
package swig binding as shared libraries


proppy:
In order to be able to import them from `.py` script, the swig binding needs to be built as shared libraries.

Currently the `swig_library` helper forces static linking, see:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/cmake/swig_lib.cmake#L74

@maliberty is the `.a` actually needed by something else in the build? do we need to build both or can we get away with only building the `.so`?

/cc @QuantamHD

QuantamHD:
@proppy I have a demo working for one of them

maliberty:
The .a are used to build a statically linked binary which is much simpler for users as it doesn't require setting up an LD_LIBRARY_PATH entry.

proppy:
@maliberty understood, but curious if there a usecase for having the swig bindings `openroad_swig_py.a` as a static library? Is it meant for users that'd like to link and extend them?

maliberty:
openroad -python needs the swig bindings and is statically linked.

QuantamHD:
@proppy We need to set up the infrastructure to build and package the so's for pip.

QuantamHD:
https://github.com/The-OpenROAD-Project/OpenROAD/pull/2311

maliberty:
@macd FYI

macd:
It would be nice to be able to import into an unmodified Python. I will take a look at @QuantamHD changes (I have PoC wrappers for about 10 of the tools/libs at present). How many versions of Python do you envision building against?

Issue #1424, opened 2021-12-14
odb should be distributed as a regular python package


proppy:
Currently the documented way to interact with the openroad python package is thru `openroad -python` (thru https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/src/swig/python/main.cpp?), it would be great if it was also distributed as a regular python package on [PyPI](https://pypi.org/) and/or [conda-eda](https://github.com/hdl/conda-eda) so that developers can import it from a regular python interpreter without relying on the `openroad` binary, ex:
```
$ python
>>> import openroad
```

maliberty:
We used to support odb in python as you describe but have never tried to support all of openroad.   I suspect it is possible but a fair amount of work to do and support (e.g. we don't build shared libraries currently).  Would you describe how this would be advantageous over "openroad -python"?

proppy:
>  Would you describe how this would be advantageous over "openroad -python"?

It does make it easier to use openroad from other python projects or from a jupyter notebook.

> but have never tried to support all of openroad

I wonder if the remaining part could be bridged with https://docs.python.org/3/library/tkinter.html, it does seems to allow TCL from python using: https://docs.python.org/3/library/tkinter.html#tkinter.Tcl

maliberty:
I don't think that will help as we need to have the tcl interpreter object in c++ in order to register our swig wrapper on it.  We could make a command to forward to tcl though that would be the ugliest interface possible.

proppy:
> I don't think that will help as we need to have the tcl interpreter object in c++ in order to register our swig wrapper on it.

Ah yes, you're right those must be two different Tcl context. 

> that would be the ugliest interface possible.

Ack, out of the command listed under https://openroad.readthedocs.io/en/latest/main/README.html#run, what are the ones that are heavily dependant on a non-trivial amount of TCL to be usable, and therefore not trivially expose-able to python thru the existing SWIG interface?

maliberty:
> Ack, out of the command listed under https://openroad.readthedocs.io/en/latest/main/README.html#run, are there ones that are heavily dependent on a non-trivial amount of TCL to be usable? and therefore not trivially expose-able to python thru the existing SWIG interface?

[...]

PR #2311, opened 2022-09-27
Draft: Adds python pip support

QuantamHD:


src/cmake/swig_lib.cmake
```diff
@@ -62,16 +62,29 @@ function(swig_lib)
...
+                 PROPERTY SWIG_MODULE_NAME ${ARG_NAME})
+  endif()
+  
   set_property(SOURCE ${ARG_I_FILE}
                PROPERTY USE_SWIG_DEPENDENCIES TRUE)
   set_property(SOURCE ${ARG_I_FILE}
                PROPERTY USE_TARGET_INCLUDE_DIRECTORIES true)
 
+
+  if (ARG_SHARED_LIBRARY)
```
proppy, line 79: https://cmake.org/cmake/help/latest/module/UseSWIG.html seems to have a `USE_BUILD_SHARED_LIBS` option to achieve something similar.

src/cmake/swig_lib.cmake
```diff
@@ -62,16 +62,29 @@ function(swig_lib)
                                         -Werror
                                         -w317,325,378,401,402,467,472,503,509)
 
-  set_property(SOURCE ${ARG_I_FILE}
```
proppy, line 65: should swig be able to infer this from the `%module` decl

python/setup.py
```diff
@@ -0,0 +1,61 @@
+import os
+import pathlib
+
+from setuptools import setup, Extension
+from setuptools.command.build_ext import build_ext as build_ext_orig
+
+
+class CMakeExtension(Extension):
```
proppy, line 8: what about reusing something like https://github.com/diegoferigo/cmake-build-extension?

src/CMakeLists.txt
```diff
@@ -293,6 +297,24 @@ target_link_libraries(openroad
...
+add_library(libopenroad
+  ${OPENROAD_LIB_SRCS}
+  OpenRoad-py.cc
+)
+
+target_link_libraries(libopenroad
+  ${OPENROAD_TOOL_LIBS}
+)
+
+target_link_libraries(openroad
```
proppy, line 309: should this be `libopenroad`? (or removed)

proppy: ah my bad, that's for the executable.

src/CMakeLists.txt
```diff
@@ -293,6 +297,24 @@ target_link_libraries(openroad
   ${CMAKE_THREAD_LIBS_INIT}
 )
 
+add_library(libopenroad
```
proppy, line 300: should this also be linked against the optional import below (`zlib`, `cudd`, etc)

src/CMakeLists.txt
```diff
@@ -293,6 +297,24 @@ target_link_libraries(openroad
...
+  ${OPENROAD_LIB_SRCS}
+  OpenRoad-py.cc
```
proppy, line 302: curious if that could be a regular c++ library (i.e: drop the `OpenRoad-py.cc` dummy file) and have the main binary link against it?

proppy:
Thanks for working on this @QuantamHD! Made a few comments, let me know if you'd prefer me to do propose the change directly against your branch.

python/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "openroad"
+version = "0.0.1"
+authors = [
+    { name = "Matt Liberty", email = "mailiberty@openroad.tools"}
+]
+description = "A production ready opensource physical design EDA toolkit"
```
rovinski, line 11: "production-ready open-source"

Also maybe drop "production-ready" to not convey a bug-free experience :)

python/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "openroad"
+version = "0.0.1"
```
rovinski, line 7: What parses the version number and is this going to break things when it doesn't match the source code version? Can this be generated by cmake?

QuantamHD:
We need to complete the librarification of the rest of the openroad modules.

maliberty:
> We need to complete the librarification of the rest of the openroad modules.

It seems this could proceed in parallel.

maliberty:
@QuantamHD @proppy any further plans or close?

maliberty:
@proppy is there a timeline?  As I said earlier "It seems this could proceed in parallel." so I think you are unblocked from progress.

proppy:
Another pattern/reference here is https://github.com/Mizux/cmake-swig (which I believe is what @Mizux uses for https://github.com/google/or-tools).

maliberty:
@proppy we already have swig setup with cmake.  The main point here is about pip packaging

Mizux:
> Another pattern/reference here that we could reuse/take inspiration from is https://github.com/Mizux/cmake-swig (which I believe is what @Mizux uses for packaging https://github.com/google/or-tools).

For python only I would look at:
* https://github.com/mizux/python-native (C++ + SWIG python)
* https://github.com/Mizux/cmake-pybind11 (C++ + pybind11 for wrapping)
* https://github.com/Mizux/bazel-pybind11 (Using bazel based build) (also provide cmake based build for comparison)

proppy:
@maliberty but my understanding is that what this PR (despite its title) is trying to address is more building the swig binding as a shared library importable from python #2308 rather than addressing the distribution on pypi #1424 (ultimately we need both, but one need to happen before the other)

proppy:
@Mizux thanks for sharing, that's an awesome reference!

python/pyproject.toml
Mizux: using a `pyproject.toml.in` with some `@CMAKE_VARIABLES@` or generator expression should be easy
e.g.
A "template" file (here setup.py.in but could be a toml.in file ;))
https://github.com/Mizux/python-native/blob/14648387a267b8ce65db0c0cdd47c1200c8cd7fb/python/setup.py.in#L16-L18
And the CMake stuff to auto-replace...
https://github.com/Mizux/python-native/blob/14648387a267b8ce65db0c0cdd47c1200c8cd7fb/cmake/python.cmake#L173-L181

Last step a custom_command to generate the python package (1 create a the package layout layout with all files in the build dir, 2) ask python the package wheel our stuff, 3) profit !)
https://github.com/Mizux/python-native/blob/14648387a267b8ce65db0c0cdd47c1200c8cd7fb/cmake/python.cmake#L197-L237

maliberty:
> @maliberty but my understanding is that what this PR (despite its title) is trying to address is more building the swig binding as a shared library importable from python #2308 rather than addressing the distribution on pypi #1424 (ultimately we need both, but one need to happen before the other)

There is no question about how to build the python bindings and we've already done a lot of work in that direction and don't plan to change course.  The only question is about building shared libraries.

This PR contains python/pyproject.toml which seems related to pip distribution.

If the title is wrong, please fix it.

QuantamHD:
Yeah let me signal boost @maliberty. The bindings part is already solved we have a defined C++ public API. The only question is build the shared libraries such that they don't require the giant OpenROAD object that forces you to link the entire app for the library.

@cdleary and I did a bunch of work to separate libraries (gpl, grt, etc.) into library components which should make this work easier. There's probably still challenges around logging, and cmake config to be solved.

maliberty:
Thanks @QuantamHD 
Even if we have to "link the entire app for the library" it would be better than waiting for the perfect architecture in OR.  This is >1year old so something non-ideal is better than nothing at all.

QuantamHD:
There's actually a problem building it with the whole app. It won't link because it's pulling in main symbols which python will reject.

maliberty:
> There's actually a problem building it with the whole app. It won't link because it's pulling in main symbols which python will reject.

I didn't mean that quite so literally.  I just meant making a single .so for everything but Main.cc is tolerable.

QuantamHD:
A single .so is still the goal it's just a question of actually getting one of those that actually works with pip that's the hard part. I was just pointing out one problem I ran into which is what stopped this PR. Pulling out Main.cc wasn't as trivial as I expected it to be.

maliberty:
@QuantamHD I'm missing something here.  The work you and Chris did was so you could use a single library in near isolation for unit testing.  How does that relate to building a single .so?  What are the issues with pulling out Main?

QuantamHD:
It's all related @maliberty. I don't recall the exact details, but there's something Main that the OpenROAD object depends, and without it none of the logging builds.

proppy:
> which seems related to pip distribution

Sorry for the miss-understanding: I do agree all step to get there are inter-related, I just thought it would be beneficial to talk about them in isolation.
a/ writing the swig binding (done)
b/ building a python importable shared library (this PR)
c/ packaging this shared library as a platform wheel that works on most linux distro (a bit of this PR, but more is needed)
d/ automating the distribution on PyPi

maliberty:
It would be helpful if this PR were brought up to date.

Please give a clear statement of the specific issue to be solved is given.  All I know is that "something in Main that the OpenROAD object depends on..." but that isn't enough to take any action.

maliberty:
Its also confusing that the title says Draft but this isn't actually a draft PR.  Is this seeking help or just in progress?

QuantamHD:
It was draft because it was broken. I don't really have time to work on it at the moment. So this is basically scaffolding that some can pick up, or if no one is interested then it can be closed, and I'll pick it up again when I have more time to focus on this.

PR #2327, opened 2022-10-04
etc: DependencyInstaller no need to use sudo

vvbandeira:
The script should be run with sudo or with root permissions, using sudo inside it is redundant and might break on systems that do not have sudo installed.

Signed-off-by: Vitor Bandeira <vitor.vbandeira@gmail.com>


maliberty:
Mention the use of sudo be mentioned in https://github.com/The-OpenROAD-Project/OpenROAD#install-dependencies

maliberty (approved):

PR #2331, opened 2022-10-05
adding dependencies support for ubuntu 22

habibayassin:

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -5,15 +5,35 @@ set -euo pipefail
 _installCommonDev() {
     lastDir="$(pwd)"
     # tools versions
-    cmakeVersionBig=3.14
-    cmakeVersionSmall=${cmakeVersionBig}.0
-    swigVersion=4.0.1
-    boostVersionBig=1.76
-    boostVersionSmall=${boostVersionBig}.0
-    eigenVersion=3.3
-    lemonVersion=1.3.1
-    spdlogVersion=1.8.1
-
+    if [[ $1  == 22.04 ]]; then
+        dist="linux"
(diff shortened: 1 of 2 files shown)


etc/DependencyInstaller.sh
```diff
@@ -5,26 +5,46 @@ set -euo pipefail
...
+        cmakeVersionSmall=${cmakeVersionBig}.0
+        swigVersion=4.0.1
+        swigChecksum="ef6a6d1dec755d867e7f5e860dc961f7"
+        boostVersionBig=1.76
+        boostVersionSmall=${boostVersionBig}.0
+        boostChecksum="e425bf1f1d8c36a3cd464884e74f007a"
+        eigenVersion=3.3
+        lemonVersion=1.3.1
+        lemonChecksum="e89f887559113b68657eca67cf3329b5"
+        spdlogVersion=1.8.1
```
vvbandeira, line 35: why the slightly different versions? If the newest versions work for both versions of ubuntu we should keep them. @maliberty any reservations on changing the version of boost, eigen and spdlog?

habibayassin: I will check the new versions compatibility with ubuntu 20.04

vvbandeira: please also test with Centos7. you can use a docker image for quicker tests. thanks

maliberty: I would prefer to have one version of dependencies on all platforms if possible.  My only concern is that spdlog 1.10 doesn't work for us.  If 1.9.2 compiles cleanly then I'm ok with it.

vvbandeira (approved):

PR #2345, opened 2022-10-08
Adding Python wrapper for antenna checking (ant) and Python regressio…

macd:
…n tests


src/ant/test/check_api1.py
```diff
@@ -0,0 +1,21 @@
...
+ack.checkAntennas("", False)
+count = ack.antennaViolationCount()
+print(f"violation count = {count}", flush=True)
+
+# note that the tcl function check_net_violations is a boolean but
+# the tcl reg test treats it as an int. So we print an int here
+# so we can have the same log file
+net = "net50"
+viol = 1 if ant.anyViolations(net) else 0
+print(f"Net {net} violations: {viol}")
```
maliberty, line 21: Can you just do
```
print(f"Net {net} violations: {int(viol)}")
```

src/ant/src/AntennaChecker-py.i
```diff
@@ -0,0 +1,75 @@
...
+{ 
+  odb::dbNet* net = ord::getDb()->getChip()->getBlock()->findNet(net_name);
+  if (net) {
+    auto antChecker = ord::OpenRoad::openRoad()->getAntennaChecker();
+    auto vios = antChecker->getAntennaViolations(net, nullptr);
+    return !vios.empty();
+  }
+  else
+    return false;
+}
```
maliberty, line 69: Do we need this or could you just call 
```
antChecker->getAntennaViolations(net, nullptr);
```
directly from python

maliberty: I added
```
%template(ViolationVector) std::vector<ant::Violation>;
```
to AntennaChecker-py.i and in the test used
```
net = design.getBlock().findNet('net50')
ack.getAntennaViolations(net, None);
```
without a problem.

macd:
looks good.  I'll do that.
Don

maliberty (approved):

PR #2353, opened 2022-10-10
ifp: fix make_tracks to avoid unusable routing tracks

eder-matheus:


src/ifp/src/InitFloorplan.cc
```diff
@@ -502,14 +506,42 @@ void InitFloorplan::makeTracks(odb::dbTechLayer* layer,
         layer->getName());
     return;
   }
+
+  int layer_min_width = layer->getMinWidth();
+
   auto x_track_count = int((die_area.dx() - x_offset) / x_pitch) + 1;
-  grid->addGridPatternX(die_area.xMin() + x_offset, x_track_count, x_pitch);
+  int origin_x = die_area.xMin() + x_offset;
+  // Check if track origin is usable during router
```
maliberty, line 514: "Check if the track origin is usable during routing"
similarly below

src/ifp/src/InitFloorplan.cc
```diff
@@ -502,14 +506,42 @@ void InitFloorplan::makeTracks(odb::dbTechLayer* layer,
...
+  if (layer_min_width / 2 > origin_x - die_area.xMin()) {
```
maliberty, line 515: >=
it can touch the edge, just not go beyond.  Similarly in the other checks

maliberty (changes requested):

src/ifp/src/InitFloorplan.cc
eder-matheus: I think my comments are not clear about these checks. If the if statement is true, I'm ignoring the first (or the last) x/y coordinate, because the wire would go out of the die area. If I use `>=`, I'll be removing the tracks when the wire is just touching the edge.

Perhaps a better comment is "Check if the track is **not** usable during routing".

maliberty: The code is correct but the comment is confusing.  Also I would have written
```
 if (origin_x - layer_min_width / 2 <  die_area.xMin()) {
```
its the same mathematically but in mind it is more natural to compare the two x coordinates than two widths.

maliberty:
Why did so many new diffs appear in the ok files with your last change?  I would have thought it was just for readability and not functionally different
https://github.com/The-OpenROAD-Project/OpenROAD/pull/2353/files/898503b4f7def62bf026d6c4e45937e1a6f77a00..1820fd021260346d66ec32a18acb00f36831e9b9

eder-matheus:
> Why did so many new diffs appear in the ok files with your last change? I would have thought it was just for readability and not functionally different https://github.com/The-OpenROAD-Project/OpenROAD/pull/2353/files/898503b4f7def62bf026d6c4e45937e1a6f77a00..1820fd021260346d66ec32a18acb00f36831e9b9

These were failing before the readability changes. You can see the failures here: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-Public/detail/PR-2353-merge/1/pipeline/123

maliberty (approved):

PR #2356, opened 2022-10-11
gpl: Add -timing_driven_nets_percentage

antonblanchard:
This allows us to tune the percentage of nets that are reweighted
in timing driven placement.

### DIFF
--- a/src/gpl/README.md
+++ b/src/gpl/README.md
@@ -44,6 +44,7 @@ global_placement
     [-routability_rc_coefficients routability_rc_coefficients]
     [-timing_driven_net_reweight_overflow]
     [-timing_driven_net_weight_max]
+    [-timing_driven_nets_percentage]
     [-pad_left pad_left]
     [-pad_right pad_right]
     [-verbose_level level]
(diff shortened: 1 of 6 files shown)


antonblanchard:
I created the `-timing_driven_nets_percentage` option to mirror `-critical_nets_percentage` recently added to grt, but I just realised it might be more consistent to name it `-timing_driven_net_percentage`:

```
    [-timing_driven_net_reweight_overflow]
    [-timing_driven_net_weight_max]
```

src/gpl/README.md
```diff
@@ -68,15 +68,17 @@ global_placement
...
 `-timing_driven` does a virtual `repair_design` to find slacks and
-weight nets with low slack. It adjusts the 10% worst slacks using
-a multiplier (1.9 by default, modified with `-timing_driven_net_weight_max`).
-The multiplier is scaled from the full value for the worst slack, to 1.0 for
-the 10% worst slack. Use the `set_wire_rc` command to set resistance and
+weight nets with low slack. It adjusts the worst slacks (10% by default,
+modified with -timing_driven_nets_percentage) using a multiplier (1.9 by
+default, modified with `-timing_driven_net_weight_max`). The multiplier
+is scaled from the full value for the worst slack, to 1.0 for the 10%
+worst slack. Use the `set_wire_rc` command to set resistance and
```
maliberty, line 81: is " to 1.0 for the 10% worst slack" still accurate or is 1.0 at the timing_driven_nets_percentage point?

src/gpl/src/timingBase.cpp
```diff
@@ -186,7 +186,12 @@ TimingBase::updateGNetWeights(float overflow) {
         gNet->setTimingWeight(weight);
         weighted_net_count++;
       }
-      debugPrint(log_, GPL, "replace", 3, "timing: net:{} slack:{} weight:{}", 
+
+      int level = 3;
+      if (sta::fuzzyEqual(gNet->totalWeight(), 1.0))
+        level = 4;
+
+      debugPrint(log_, GPL, "replace", level, "timing: net:{} slack:{} weight:{}",
```
maliberty, line 194: What is the level conditional?

src/gpl/README.md
antonblanchard: Yes this does need updating. Will do that.

src/gpl/src/timingBase.cpp
antonblanchard: The idea was to keep the previous behaviour of printing every net but at a higher debug level. I'm not sure there is a lot of point though.

maliberty: Its just a debug so let's pick one level.

antonblanchard: I've removed this part of the patch.

antonblanchard:
Still to be resolved:

> I created the `-timing_driven_nets_percentage` option to mirror `-critical_nets_percentage` recently added to grt, but I just realised it might be more consistent to name it `-timing_driven_net_percentage`:
> 
> ```
>     [-timing_driven_net_reweight_overflow]
>     [-timing_driven_net_weight_max]
>     [-timing_driven_nets_percentage]
> ```

maliberty (approved):

PR #2358, opened 2022-10-11
odb: version checking fixes

antonblanchard:
It looks like the minor revision check is reversed, because before I added the second patch I was getting:

```
incompatible database schema minor revision (database requires 56 or greater but OpenROAD is 57)
```

I presume the idea of a minor revision is an update that retains compatibility with older minor revisions (is that correct?). After making the fix I tested that database (minor rev 56, OpenROAD minor rev 57) and it promptly used all of my memory and crashed. That suggests we are misusing the minor rev number.

### DIFF
--- a/src/odb/src/db/dbDatabase.cpp
+++ b/src/odb/src/db/dbDatabase.cpp
@@ -306,12 +306,12 @@ dbIStream& operator>>(dbIStream& stream, _dbDatabase& db)
   stream >> db._schema_major;
 
   if (db._schema_major != db_schema_major)
-    throw ZException("Incompatible database schema revision");
+    throw ZException("incompatible database schema major revision (OpenROAD requires %d but database is %d)", db_schema_major, db._schema_major);
 
   stream >> db._schema_minor;
 
   if (db._schema_minor < db_schema_initial)
-    throw ZException("incompatible database schema revision");
+    throw ZException("incompatible database schema minor revision (database requires %d or greater but OpenROAD is %d)", db._schema_minor, db_schema_initial);
 
   stream >> db._master_id;
 
--- a/src/odb/src/db/dbDatabase.cpp
+++ b/src/odb/src/db/dbDatabase.cpp
@@ -310,7 +310,7 @@ dbIStream& operator>>(dbIStream& stream, _dbDatabase& db)
 
   stream >> db._schema_minor;
 
-  if (db._schema_minor < db_schema_initial)
+  if (db_schema_initial < db._schema_minor)
     throw ZException("incompatible database schema minor revision (database requires %d or greater but OpenROAD is %d)", db._schema_minor, db_schema_initial);
 
   stream >> db._master_id;


maliberty:
> I presume the idea of a minor revision is an update that retains compatibility with older minor revisions (is that correct?). 

Yes, in the sense that a new rev can read the older rev but not visa-versa.

> After making the fix I tested that database (minor rev 56, OpenROAD minor rev 57) and it promptly used all of my memory and crashed. That suggests we are misusing the minor rev number.

Do you have an example of that?  57 should read 56.

antonblanchard:
> > After making the fix I tested that database (minor rev 56, OpenROAD minor rev 57) and it promptly used all of my memory and crashed. That suggests we are misusing the minor rev number.
> 
> Do you have an example of that? 57 should read 56.

I was trying the test case from https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/576

If I patch OpenROAD with the first of my commits, which just adds revision numbers to the incompatibility errors:
```
Error: global_place_skip_io.tcl, 2 incompatible database schema minor revision (database requires 56 or greater but OpenROAD is 57)
```

If I add the patch to reverse the minor revision check, it died after consuming 43GB of RAM.

antonblanchard:
This seems to be the problem:

```
Thread 1 (Thread 0x7fb3fa57d8c0 (LWP 1765885) "openroad"):
#0  0x00007fb3f9558ac0 in __memset_avx2_unaligned_erms () from /lib64/libc.so.6
#1  0x0000000000cd6db6 in std::vector<unsigned int, std::allocator<unsigned int> >::_M_default_append(unsigned long) ()
#2  0x0000000002af0da0 in odb::dbIStream& odb::operator>><unsigned int>(odb::dbIStream&, odb::dbMatrix<unsigned int>&) [clone .isra.0] ()
#3  0x0000000002af3408 in odb::operator>>(odb::dbIStream&, odb::_dbTechLayer&) ()
#4  0x0000000002a31cd0 in odb::operator>>(odb::dbIStream&, odb::_dbTech&) ()
#5  0x00000000029d323a in odb::operator>>(odb::dbIStream&, odb::_dbDatabase&) ()
#6  0x00000000029d3ca5 in odb::dbDatabase::read(_IO_FILE*) ()
#7  0x0000000000c33ecf in ord::OpenRoad::readDb(char const*) ()
#8  0x0000000000c41c46 in _wrap_read_db_cmd ()
#9  0x00007fb3fb722970 in TclNRRunCallbacks () from /lib64/libtcl8.6.so
```

antonblanchard:
One of these two patches is making an incompatible change to the database:

commit 2f0cb8875fa3f79d80db2c546fbad65d4452ee42 ('odb: use 2 bits for overlap instead of int')

commit fc7d8a9508785f8b6ad3282434d4dcd4ff1b36ab ('odb: use int for mask and overlap and remove setNumber')

@eder-matheus FYI

maliberty:
its the first of those two

maliberty:
You've found a real issue where we failed to update the schema.  This PR however is wrong.  db_schema_initial is the older version of the schema that this code can load (it can load any later version).  If you have an even older schema it is an error.  The code is correct to me.

maliberty:
It looks like the schema version has been updated since @eder-matheus change so there is no easy way to fix this omission.  We'll just have to be careful for a while until such dbs become obsolete.

antonblanchard:
Perhaps this change:

```
-      "type": "int",
-      "name": "overlap_"
+      "type": "uint",
+      "name": "overlap_",
+      "bits": 2
```

Seems like we should have bumped the major revision to the database.

maliberty:
You can solve that in the io operators and remain backwardly compatible but its too late now.

antonblanchard:
I misunderstood how the versioning worked and thought it was a bad check because many of the test cases I've downloaded recently (odb based, no longer def based) wont load in OpenROAD. It would be nice to have some older databases in our test suite so things don't break over time.

maliberty:
We do have one test but apparently it didn't catch this

PR #2359, opened 2022-10-11
Added getTech and evalTclString to Design

macd:


maliberty (approved):
Remember to request a review

macd:
Yes, I was looking at that but there was no indication on how to do that. Looking at the github docs it looks like only owners and collaborators can request a review.

include/ord/Design.h
```diff
@@ -73,6 +73,10 @@ class Design
 
   int micronToDBU(double coord);
 
+  int evalTclString(const std::string& cmd);
```
maliberty, line 76: Please add a comment that this is a temporary backdoor and not a permanent feature.

maliberty (approved):

include/ord/Design.h
macd: Will do. (CI in progress)

PR #2360, opened 2022-10-12
Minor cleanup for odb regression paths (python/TCL) + small fix in dependency install

openroadie:
Fixed issues with dependency installation (missing graphviz/doxygen)
Fixed problems that most likely arose from renaming opendb to odb and restructuring the build directory.

### DIFF
--- a/src/grt/src/fastroute/src/FastRoute.cpp
+++ b/src/grt/src/fastroute/src/FastRoute.cpp
@@ -968,7 +968,7 @@ NetRouteMap FastRouteCore::run()
   // debug mode Rectilinear Steiner Tree before overflow iterations
   if (debug_->isOn_ && debug_->rectilinearSTree_) {
     for (int netID = 0; netID < netCount(); netID++) {
-      if (nets_[netID]->db_net == debug_->net_) {
+      if (nets_[netID]->db_net == debug_->net_ && !nets_[netID]->is_routed) {
         StTreeVisualization(sttrees_[netID], nets_[netID], false);
       }
     }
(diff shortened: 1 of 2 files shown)


etc/DependencyInstaller.sh
```diff
@@ -120,11 +120,13 @@ _installUbuntuDev() {
         automake \
         autotools-dev \
         bison \
+	doxygen \
         flex \
         clang \
         g++ \
         gcc \
         git \
+	graphviz \
```
vvbandeira, line 129: Why add these packages? Also, if we add for ubuntu we should add for other all the OS.

vvbandeira (commented):
You are missing DCO, click on "Details" to get instructions on how to fix.

image

etc/DependencyInstaller.sh
openroadie: These packages were missing and needed to be installed by hand to get the docker container to work. Will fix for other OSes.

openroadie:
@maliberty I saw the following issues:
- If I build a docker container using the base ubuntu image, I had to manually install doxygen and graphviz
- The odb README file was written back when odb was probably called opendb
- The python and tcl regressions were failing to start because the build structure has changed. 

Is it possible that the base docker image you are using already has doxygen/graphviz which are not standard ?

maliberty:
I'm not aware that we have any dependency on those tools in or out of docker.  What specifically goes wrong when they are not installed?

rovinski:
doxygen is optional as far as I know. It was added by request of a user in https://github.com/The-OpenROAD-Project/OpenDB/pull/124 (before merging odb into OR) because they wanted to get a hierarchy tree generated for odb (there isn't any doxygen-specific documentation in odb).

I'm guessing graphviz is a dependency for doxygen? If so then the same applies.

maliberty:
graphviz will be for doxygen.  There is an optional setup for doxygen in odb but it isn't required and should be skipped if not found.

openroadie:
@maliberty @rovinski Thank you for the reviews. I am still figuring quite a few things out. I will remove the doxygen/graphviz changes and see if I can bypass the build section that required this.

Do the rest of the changes look ok (mostly related to opendb being renamed to odb and changes in the way cmake is generating the build hierarchy).

@vvbandeira I will fix the DCO stuff as well.

PR #2365, opened 2022-10-13
ppl: check pin min area

eder-matheus:


src/ppl/src/Netlist.cpp
```diff
@@ -174,4 +174,11 @@ void Netlist::clear()
   _db_pin_idx_map.clear();
 }
 
+int IOPin::getArea() const
+{
+  int area = std::abs(upper_bound_.getX() - lower_bound_.getX())
+             * std::abs(upper_bound_.getY() - lower_bound_.getY());
```
rovinski, line 180: This is a nitpick but you could do a single `std::abs` after multiplying.

eder-matheus: Thanks, makes more sense to have only one `std::abs`.

maliberty (approved):

PR #2370, opened 2022-10-14
gpl: Fix for trivial designs (RePlAce diverged at initial iteration)

antonblanchard:
I have a trivial design (a few wires between input and output ports) that fails in gpl with the following error:

[ERROR GPL-0304] RePlAce diverged at initial iteration. Re-run with a smaller init_density_penalty value.

The issue is gradDistance is 0, and getStepLength divides by 0. Catch this and return 0 instead.

### DIFF
--- a/src/gpl/src/nesterovPlace.cpp
+++ b/src/gpl/src/nesterovPlace.cpp
@@ -842,7 +842,10 @@ NesterovPlace::getStepLength(
   debugPrint(log_, GPL, "replace", 3, "getStepLength:  CoordinateDistance: {:g}", coordiDistance);
   debugPrint(log_, GPL, "replace", 3, "getStepLength:  GradientDistance: {:g}", gradDistance);
 
-  return coordiDistance / gradDistance;
+  if (gradDistance == 0.0)
+    return 0;
+  else
+    return coordiDistance / gradDistance;
 }
 
 float


antonblanchard:
This fixes the fail, but I wonder if returning 0 here is correct, or it should be something non zero.

QuantamHD:
Math wise this should probably return a really big number, but my gut reaction is that if there's no std cells to place we probably shouldn't be calling this function.

It feels like there should be a higher level check that prevents 0 gradients.

antonblanchard:
My test case has a few (6 standard cells), so it does need some global placement.

antonblanchard:
Here is the test case (`doit.tcl`): [gpl-diverge.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/9784390/gpl-diverge.tar.gz)

antonblanchard:
Sorry about that, it requires `-skip_io` to reproduce, and I somehow dropped it from the test case.

[gpl-diverge-2.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/9835146/gpl-diverge-2.tar.gz)

maliberty:
I'm ok with the change but you need to address the test failures

maliberty:
Minor nit - use braces rule for if-else statement.  https://google.github.io/styleguide/cppguide.html#Conditionals

antonblanchard:
I've taken another look at this. The issue is with `-skip_io`, in combination with a design that only has a single cells between input and output pins. This happens if input pins are directly connected to output pins, because abc adds a buffer and OpenROAD `remove_buffers` keeps them around (as discussed recently).

Since `-skip_io` doesn't include the module pins in the wire calculation, total remaining wire length is 0, and gpl falls apart.

So perhaps a bit of a corner case, but gpl should be able to handle it.

vijayank88:
@antonblanchard 
Can you please rebase the code and update again? Will check the PR status.

vijayank88:
@antonblanchard 
Based on your code, I just rebased and created new PR that mentioned in thread.
Can you please review once, is that OK?

maliberty:
Replaced / fixed by #3170

PR #2376, opened 2022-10-15
adding github action to check format and readability of changed files

habibayassin:


vvbandeira:
LGTM, @maliberty see example of failures with the latest commit (errors were introduced intentionally to test the action). When the PR is merged into master I can set the action as non-blocking for future PRs -- note that this will still show as a failure.

.github/workflows/github-actions-format-on-push.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_dispatch:
+
+jobs:
+  format:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: setup
+        run: |
+          sudo apt update 
+          sudo apt install -y clang clang-format
```
maliberty, line 16: Will this need to happen every time?  It might be better to make an image that has these pre-installed.

maliberty:
I would like to roll this out gradually.  Let's start with format and save readability for a second step.  We can make it non-blocking until we can fully format the code base.  I don't want to disrupt developers with pending changes with lots of merge conflicts so we'll have to work through this carefully.  I am looking forward to no longer having to point this out in code reviews.

This should only apply to OR and not to any other repos (eg sta).

.github/workflows/github-actions-format-on-push.yml
vvbandeira: we depend on GH images for this action. also I think GH has optimized this (or clang is already installed) since the step only takes ~10s.

maliberty:
clang-format does get changes/improvements over time so we should lock down the version we are using and only upgrade intentionally.

vvbandeira:
@habibayassin LGTM, please revert the `drt` change so we can merge

vvbandeira (approved):

Issue #2377, opened 2022-10-16
grt: parasitic estimates don't use pin access locations


antonblanchard:
I'm running tests with `pin_access` called before `global_routing`. I notice a number of short nets have significantly higher resistances. In this case (`_03345_`) grt estimates 88 ohms but after detailed routing it is 41 ohms. Capacitance is also off by an order of magnitude.

Looking at the debug:

```
[DEBUG GRT-0001] net _03345_
(33.21, 64.53)  1 -> (33.21, 64.53)  2 l=0.00
[DEBUG GRT-0001] _03345_:0 -> _25665_/B (33.44, 64.59) 0.30u layer=2 r=0.009 via_res=0.017 c=0.043
[DEBUG GRT-0001] pin _25665_/B -> to grid 0u layer=2 r=0.009 via_res=0.017 c=0.043
[DEBUG GRT-0001] _03345_:0 -> _25664_/Y (32.20, 64.56) 1.03u layer=2 r=0.030 via_res=0.017 c=0.151
[DEBUG GRT-0001] pin _25664_/Y -> to grid 1u layer=2 r=0.030 via_res=0.017 c=0.151
 Pi model c2=8.68e-17 rpi=72.1 c1=2.29e-15 max_r=73.6
 Pi model c2=8.68e-17 rpi=72.1 c1=2.29e-15 max_r=73.6
```

The yellow dots show the points grt is using. It seems it always selects the bottom left position (I guess the first valid position?):

![grt-pin-access-2](https://user-images.githubusercontent.com/2029971/196056067-fe1b5302-7e2e-4d5c-a7c1-3de4c7c5eb54.png)

Would it be possible to choose a better access point so grt can make better estimates (eg the green point)? I presume drt is doing that pretty early on.

antonblanchard:
The issue is with `_25664_/Y`, and if I'm reading this right, pin access did choose a location on the right of the cell:

```
 gcclean2via _25664_ A VIA23 189 81 MX
 gcclean2via _25664_ C VIA12 549 117 MX
 gcclean2via _25664_ B VIA12 729 117 MX
 gcclean2via _25664_ Y VIA12 1053 45 MX
```

It looks like the issue is GRT is using pin access, but the GRT parasitics are not.

maliberty:
[...]

PR #2381, opened 2022-10-18
grt: Use pin access in parasitic estimates

antonblanchard:
This is a first attempt at fixing https://github.com/The-OpenROAD-Project/OpenROAD/issues/2377

`pinAccessPointPositions()` creates pairs of `{ap_position, ap_position_on_grid}`, does that look reasonable?

### DIFF
--- a/src/grt/src/MakeWireParasitics.cpp
+++ b/src/grt/src/MakeWireParasitics.cpp
@@ -189,7 +189,18 @@ void MakeWireParasitics::makeParasiticsToPin(Pin& pin)
 {
   sta::Pin* sta_pin = staPin(pin);
   sta::ParasiticNode* pin_node = parasitics_->ensureParasiticNode(parasitic_, sta_pin);
-  const odb::Point& grid_pt = pin.getOnGridPosition();
+
+  odb::Point pt = pin.getPosition();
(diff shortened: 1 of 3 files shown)


src/grt/src/MakeWireParasitics.cpp
```diff
@@ -331,8 +341,17 @@ MakeWireParasitics::routeLayerLengths(odb::dbNet* db_net)
...
       int layer = pin.getConnectionLayer() + 1;
-      const odb::Point& grid_pt = pin.getOnGridPosition();
-      const odb::Point& pt = pin.getPosition();
+      odb::Point grid_pt = pin.getOnGridPosition();
+      odb::Point pt = pin.getPosition();
+
+      std::vector<std::pair<odb::Point, odb::Point>> ap_positions;
+      bool has_access_points = grouter_->pinAccessPointPositions(pin, ap_positions);
+      if (has_access_points) {
+        auto ap_position = ap_positions.front();
```
maliberty, line 350: It would be better to use the preferred access point rather than just the first one.

maliberty: This is ok I just missed it.

maliberty:
Note that we don't run pin access in grt currently.  There is pending work to enable that.

PR #2385, opened 2022-10-19
Fixes tab parsing in LEF files

QuantamHD:


QuantamHD:
Talked to Matt about this and the difference is `- blank - '\n'`. This should support both use cases.

maliberty:
Evolution
char_ - ' ' - '\n'   [doesn't handle tabs]
to
char_ - blank  [doesn't handle new lines; got reverted]
to
char_ - blank - '\n'  [handles both tabs and new lines]

(why people are still formatting with tabs is beyond me)

maliberty (approved):

PR #2387, opened 2022-10-20
drt: sort insts and nets to ensure stable results

antonblanchard:
I have a test case for global and detailed routing where the results of detailed routing are different if an odb or def is used. Both the odb and def are created at the same time (before global routing), and there should be nothing in the odb at that stage that cannot be represented by the def.

One difference (pointed out by Matt) is that write_def sorts insts and nets, so we iterate through them in a different order. Sorting them before using them in drt fixes the issue.

### DIFF
--- a/src/drt/src/io/io.cpp
+++ b/src/drt/src/io/io.cpp
@@ -26,6 +26,8 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "io/io.h"
+
 #include <exception>
 #include <fstream>
 #include <iostream>
@@ -35,7 +37,6 @@
 #include "frProfileTask.h"
 #include "frRTree.h"
 #include "global.h"
-#include "io/io.h"
 #include "odb/db.h"
 #include "odb/dbWireCodec.h"
 #include "utl/Logger.h"
--- a/src/drt/src/io/io.cpp
+++ b/src/drt/src/io/io.cpp
@@ -122,9 +122,21 @@ void io::Parser::setTracks(odb::dbBlock* block)
   }
 }
 
+template <typename T>
+static std::vector<T*> sortedSet(odb::dbSet<T>& to_sort)
+{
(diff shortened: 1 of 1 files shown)


antonblanchard:
I pulled the template straight out of odb. Is it overkill? Should odb be sorting these in their iterators? Should we be sorting other things in drt (eg obstructions)?

QuantamHD:
IMO we should probably add a set of API methods to ODB to support sort iterators, or potentially just store them in sorted order.

antonblanchard:
> IMO we should probably add a set of API methods to ODB to support sorted iterators, or potentially just store them in sorted order.

Sounds good. In this specific case we should also experiment with sorting by the worst n% of slacks, similar to what we do in gpl and now grt. That might help route critical timing nets better.

maliberty:
Sorry, I put this aside and then lost track of it.  This only works because drt copies odb and once that is resolved this wouldn't be possible anymore.  It also doesn't solve the similar problems that can occur in any module.

I'm not sure we need to guarantee that we get the same result no matter how the netlist is reordered.  If we do then we need to look at it everywhere.  Most algorithms should be order independent or should have some more meaningful ordering than names. 

Would you say more about why this is an issue?

antonblanchard:
> Sorry, I put this aside and then lost track of it. This only works because drt copies odb and once that is resolved this wouldn't be possible anymore. It also doesn't solve the similar problems that can occur in any module.
> 
> I'm not sure we need to guarantee that we get the same result no matter how the netlist is reordered. If we do then we need to look at it everywhere. Most algorithms should be order independent or should have some more meaningful ordering than names.
> 
> Would you say more about why this is an issue?

This started because I built a test case using lib/lef/def/sdc and it didn't reproduce the issue. It turns out I needed the odb because the sorting between the two formats was different.

This is less of an issue now that the odb format is more stable and we can use it for test cases.

maliberty:
@antonblanchard what do you want to have happen here - merge or close?  Your last comment suggests this is no longer a problem.

PR #2396, opened 2022-10-21
grt: Fix crash when calling read_guides before loading design

antonblanchard:


antonblanchard:
An example of the issue:

```
# echo read_guides src/grt/test/report_wire_length6.guide | openroad

Stack trace:
 0# 0x0000000000C3D61E in openroad
 1# 0x00007EFC6523EA70 in /lib64/libc.so.6
 2# odb::dbChip::getBlock() in openroad
 3# grt::GlobalRouter::readGuides(char const*) in openroad
```

maliberty:
Note that the clang-format check is currently not blocking but will be in the future.

Issue #2395, opened 2022-10-21
grt: read_guides doesn't clear existing guides from database


antonblanchard:
**Describe the bug**
`read_guides` partially appends to the previous guides, such that RC parasitic estimates change.

**Expected Behavior**
I would expect either the previous guides to be completely cleared, or the old guides and new guides should be successfully merged.

**Environment**
```
Git commit: a01312143a07839e13ed723264d8e34b87b4d750
kernel: Linux 5.19.14-200.fc36.x86_64
os: Fedora Linux 36 (Workstation Edition)
cmake version 3.22.2
gcc (GCC) 12.2.1 20220819 (Red Hat 12.2.1-2)
clang version 14.0.5 (Fedora 14.0.5-1.fc36)
```

**To Reproduce**

With the attached test case, `doit_0.tcl` uses the data directly in the database after calling `global_route`:

```
openroad doit_0.tcl | grep -A 20 'net final_adder.o.*11'
```

```
[DEBUG GRT-0001] net final_adder.o\[11\]
(5.67, 1.35)  1 -> (5.67, 1.35)  2 l=0.00
(5.67, 1.35)  2 -> (21.87, 1.35)  2 l=16.20
(21.87, 1.35)  2 -> (21.87, 1.35)  3 l=0.00
(21.87, 1.35)  3 -> (21.87, 5.13)  3 l=3.78
(21.87, 5.13)  3 -> (21.87, 5.13)  2 l=0.00
(21.87, 5.13)  2 -> (22.41, 5.13)  2 l=0.54
(22.41, 5.13)  2 -> (22.41, 5.13)  1 l=0.00
[DEBUG GRT-0001] final_adder.o\[11\]:0 -> final_adder.o\[11\]:1 16.20u layer=2 r=0.392 c=2.175
[DEBUG GRT-0001] final_adder.o\[11\]:1 -> final_adder.o\[11\]:2 via 2-3 r=0.017
[DEBUG GRT-0001] final_adder.o\[11\]:2 -> final_adder.o\[11\]:3 3.78u layer=3 r=0.092 c=0.488
[DEBUG GRT-0001] final_adder.o\[11\]:3 -> final_adder.o\[11\]:4 via 3-2 r=0.017
[DEBUG GRT-0001] final_adder.o\[11\]:4 -> final_adder.o\[11\]:5 0.54u layer=2 r=0.013 c=0.073
[...]

PR #2397, opened 2022-10-21
grt: fixes for read_guides cmd

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/2395

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -1399,6 +1399,7 @@ void GlobalRouter::perturbCapacities()
 void GlobalRouter::readGuides(const char* file_name)
 {
   block_ = db_->getChip()->getBlock();
+  routes_.clear();
   if (max_routing_layer_ == -1 || routing_layers_.empty()) {
     max_routing_layer_ = computeMaxRoutingLayer();
     int min_layer = min_layer_for_clock_ > 0
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -33,6 +33,8 @@
 //
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "grt/GlobalRouter.h"
+
 #include <algorithm>
 #include <cmath>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


maliberty (approved):

src/grt/src/GlobalRouter.cpp
```diff
@@ -33,6 +33,8 @@
 //
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "grt/GlobalRouter.h"
+
```
rovinski, line 37: This issue with moving header lines keeps popping up.  Is this due to a bad version of clang-format? This passes the format check.

eder-matheus: Not sure about that. In this PR, the format check complained about the header being down in the list, so I've run clang-format for this file. I', using `clang-format version 14.0.6` on a macos

rovinski: @vvbandeira I don't think this is desired behavior

maliberty: What is the concern?

rovinski: Different versions of clang-format keep changing the position of the header files.

* Here the header is moved to the top for some reason.
* 72b7afd moves `#include "FastRoute.h"` to alphabetical order.
* dc8d7ec moves `#include "sta/Parasitics.hh"` to alphabetical order.
* c78920b moves `#include "ppl/IOPlacer.h"` to the top

There's been more instances but those are the ones I found in the last 50 PRs or so.

I would think keeping the headers in alphabetical order makes more sense, no?

maliberty: clang-format splits includes into groups and sorts within groups.  In google style, which we use, <> and "" are separate groups.   The first two are as expected but the third seems odd.   @eder-matheus what version of clang-format are you using?

Once this check becomes mandatory it will be an error to use a non-conforming version.  The action uses 14.0.6

rovinski: The weirdness you see in c78920b is the same weirdness in this commit, and it's what the format CI wants (see [the CI failure](https://github.com/The-OpenROAD-Project/OpenROAD/actions/runs/3298813598/jobs/5441411670)). So this problem exists in the current format CI.

maliberty: Its a good practice that foo.cc include foo.h first to validate that the header is correct without relying on any further headers to be included first.  I didn't realize clang-format does that but I am in favor of it.  Other headers are grouped separately and sorted.

rovinski: gotcha

PR #2401, opened 2022-10-22
Grt python

macd:
Not all the tcl tests have a Python version yet, but there is a good variety of Python regressions.

### DIFF
--- a/src/grt/README.md
+++ b/src/grt/README.md
@@ -228,6 +228,60 @@ The `global_route_debug` command allows you to start a debug mode to view the st
 
 ## Limitations
 
+## Using the Python interface to grt
+
+**NOTE:** The Python interface is currently in development and may
+change.
+
+This api tries to stay close to the api defined in C++ class
+GlobalRouter that is located in grt/include/grt/GlobalRouter.h. 
+
+When initializing a design, a sequence of Python commands might look like
+the following:
+
+    from openroad import Design, Tech
+    tech = Tech()
+    tech.readLef(...)
+    design = Design(tech)
+    design.readDef(...)
+    gr = design.getGlobalRouter()
+    
+Here are some options / configurations to the globalRoute
+command. (See GlobalRouter.h for a complete list)
+
+    gr.setGridOrigin(x, y)                     # int, default 0,0
+    gr.setCongestionReportFile(file_name)      # string
+    gr.setOverflowIterations(n)                # int, default 50
+    gr.setAllowCongestion(allowCongestion)     # boolean, default False
+    gr.setCriticalNetsPercentage(percentage)   # float
+    gr.setMinRoutingLayer(minLayer)            # int
+    gr.setMaxRoutingLayer(maxLayer)            # int
+    gr.setMinLayerForClock(minLayer)           # int
+    gr.setMaxLayerForClock(maxLayer)           # int
+    gr.setVerbose(v)                           # boolean, default False
+
+and when ready to actually do the global route:
+
+    gr.globalRoute(save_guides)                # boolean, default False
+    
+If you have set `save_guides` to True, you can then save the guides in `file_name` with:
+
+    design.getBlock().writeGuides(file_name)
+
+You can find the index of a named layer with
+
+    lindex = tech.getDB().getTech().findLayer(layer_name)
+
+or, if you only have the Python design object
+
+    lindex = design.getTech().getDB().getTech().findLayer(layer_name)
+    
+Be aware that much of the error checking is done in TCL, so that with
+the current C++ / Python api, that might be an issue to deal
+with. There are also some useful Python functions located in the file
+grt/test/grt_aux.py but these are not considered a part of the (final)
+api and they may change.
+
 ## FAQs
 
 Check out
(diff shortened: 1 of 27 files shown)


macd:
The failure in format code looks to be a failure running on a .i file. That is probably not what we want to do.

maliberty:
@vvbandeira @habibayassin .i files should be excluded from clang-fomat as they are not c++

maliberty:
I've excluded .i in https://github.com/The-OpenROAD-Project/OpenROAD/pull/2403 and you'll need to merge master after that PR is merged.

macd:
Sounds good.

src/grt/README.md
```diff
@@ -228,6 +228,60 @@ The `global_route_debug` command allows you to start a debug mode to view the st
...
+
+    gr.setGridOrigin(x, y)                     # int, default 0,0
+    gr.setCongestionReportFile(file_name)      # string
+    gr.setOverflowIterations(n)                # int, default 50
+    gr.setAllowCongestion(allowCongestion)     # boolean, default False
+    gr.setCriticalNetsPercentage(percentage)   # float
+    gr.setMinRoutingLayer(minLayer)            # int
+    gr.setMaxRoutingLayer(maxLayer)            # int
+    gr.setMinLayerForClock(minLayer)           # int
+    gr.setMaxLayerForClock(maxLayer)           # int
```
maliberty, line 260: It would be better if these took a dbLayer* rather than an integer.

src/grt/include/grt/GlobalRouter.h
```diff
@@ -153,7 +153,7 @@ class GlobalRouter
   void setMinLayerForClock(const int min_layer);
   void setMaxLayerForClock(const int max_layer);
   void setCriticalNetsPercentage(float critical_nets_percentage);
-  unsigned getDbId();
+  // unsigned getDbId();  // UNUSED
```
maliberty, line 156: Just remove them - same below

src/grt/test/grt_aux.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def set_global_routing_layer_adjustment(design, layer, adj):
+    """In design "design", set the global layer adjustment for "layer".
+       layer can be a named layer or range specified as layer1-layer2,
+       where layer1 < layer2. The adjustment "adj" must be > 0.0"""
+    gr = design.getGlobalRouter()
+    
+    if adj <= 0.0:
+        utl.error(utl.GRT, 507, "Layer adjustment must be positive")
```
maliberty, line 62: should be checked in c++

src/grt/test/grt_aux.py
```diff
@@ -0,0 +1,157 @@
...
+    design.getGlobalRouter().setMinRoutingLayer(lmin)
+    design.getGlobalRouter().setMaxRoutingLayer(lmax)
+
+    rtech = design.getTech().getDB().getTech()
+    for layer in range(lmin, lmax+1):
+        db_layer = rtech.findRoutingLayer(layer)
+        if (not layer_has_tracks(design, db_layer, orient=Orient.HORZ) and
+            not layer_has_tracks(design, db_layer, orient=Orient.VERT)):
+           layer_name = db_layer.getName()
+           utl.error(utl.GRT, 509, f"Missing track structure for layer {layer_name}.")
```
maliberty, line 118: check in c++

maliberty (changes requested):

maliberty:
In general it feels like there is too much support code in the unit tests that should be either in c++ or a better api.

src/grt/README.md
macd: So these are defined by the current C++ api (in GlobalRouter.h) that is then used by GlobalRouter.i and then further wrapped by GlobalRouter.tcl. So if we change this in C++ we will have to change the tcl interface too (or at least the tcl code to maintain a similar interface), or we can have a parallel interface (not my preference).

src/grt/include/grt/GlobalRouter.h
macd: OK

src/grt/test/grt_aux.py
macd: So effectively, that means moving much of grt_aux.py into the C++ api. You comment below that "In general it feels like there is too much support code in the unit tests..." I can understand that but the "support code" for the tcl interface lies in both the GlobalRouter.i and the GlobalRouter.tcl files and totals about 800 lines (let's de-rate to 700 for boilerplate). The GlobalRouter-py.i file and grt_aux.py total about 200 loc, so that is still quite a step down. It will be better in C++ for sure, but it is a redesign of the api.

maliberty:
Are you wanting to tackle the API redesign or do you want to just merge this and leave it to me to handle later?

macd:
Good question Matt. I think I'd like to merge this and have you do the api redesign later. I don't think I know enough about the whole flow to be able to put down a consistent api at the moment. I will do a few of the simple changes first. Should I leave grt_aux.py in grt/test or should we move it to grt/src/GlobalRouter.py to mirror grt/src/GlobalRouter.tcl? I chose the former to signal "this api might not be stable".

maliberty:
Leave grt_aux.py in test

maliberty (approved):

PR #2404, opened 2022-10-23
gpl: cells get placed on top of macros

ramezmoussa:
One problem gpl had is that sometimes cells would get placed on top of macros. This is not an issue if the placement is towards the edges of the macro but it becomes problematic when cells get placed towards the center of the macro because these cells will have to be moved long distances to find a legal place.

An example of such case can be found below (the microwatt design):

image

In this PR, I implement scaling within the macro area such that it is more dense towards the center of the macro and consequently, cells will tend to be placed away from the center of the macro. Scaling is represented as the integration of the density function of a bivariate normal distribution with no correlation. For efficiency, I am using the closed-form solution for the integration instead of integrating numerically.

Such scaling affects the overflow significantly (because the regions at the center of the macro would cause huge unavoidable overflows) and that can make it impossible to indicate convergence in some cases. Consequently, I created an unscaled overflow variables which will be used along the scaled overflow ones. Essentially, the unscaled overflow will be used to indicate convergence while the scaled overflow will be the one actually used to determine the densities and the gradients.

The new implementation has been tested on a couple of designs and an example of using the new implementation can be found below (the microwatt design):

image

As seen in the image above, no cells got placed within the macro at all. In some cases, cells would get placed on top of the macro but it will always be towards the edges.

Once approved, I will format the files to meet the coding style check.


maliberty:
Have you run a secure CI?

ramezmoussa:
> Have you run a secure CI?

Not yet. I assumed I should do that during the review process.

src/gpl/src/nesterovBase.cpp
```diff
@@ -1899,6 +2038,24 @@ getOverlapArea(const Bin* bin, const Instance* inst) {
...
 
+
+// A function that does 2D integration to the density function of a
+// bivariate normal distribution with 0 correlation.
+// Essentially, the function being integrated is the product
+// of 2 1D probability density functions (for x and y). The means and standard deviation of the 
+// probablity density functions are parametarized.
+// In this function, I am using the closed-form solution of the integration.
+// The limits of integration are lx->ux and ly->uy
+static float integrate(integrationParams i)
```
QuantamHD, line 2049: integrate is a bit too general a name for this function. I would probably name it something more like calculateBiVariateNormalCDF(...).

I think renaming the struct from integrationParams to biNormalParameters might be helpful as well. That way it could potentially be used for more than just the integration.

src/gpl/src/nesterovBase.cpp
```diff
@@ -808,9 +824,11 @@ BinGrid::updateBinsGCellDensityArea(
         for(int i = pairX.first; i < pairX.second; i++) {
           for(int j = pairY.first; j < pairY.second; j++) {
             Bin* bin = bins_[ j * binCntX_ + i ];
-            bin->addInstPlacedArea( 
-                getOverlapDensityArea(bin, cell) 
-                * cell->densityScale() * bin->targetDensity() ); 
+
+            float overlap = getOverlapDensityArea(bin, cell);
```
maliberty, line 828: const

src/gpl/src/nesterovBase.cpp
```diff
@@ -808,9 +824,11 @@ BinGrid::updateBinsGCellDensityArea(
...
+            bin->addInstPlacedArea(  overlap * cell->densityScale() * bin->targetDensity() ); 
+            bin->addInstPlacedAreaUnscaled( overlap * cell->densityScale() * bin->targetDensity() );
```
maliberty, line 830: factor the common calculation

src/gpl/src/nesterovBase.cpp
```diff
@@ -819,9 +837,9 @@ BinGrid::updateBinsGCellDensityArea(
         for(int i = pairX.first; i < pairX.second; i++) {
           for(int j = pairY.first; j < pairY.second; j++) {
             Bin* bin = bins_[ j * binCntX_ + i ];
-            bin->addInstPlacedArea( 
-                getOverlapDensityArea(bin, cell) 
-                * cell->densityScale() ); 
+            int64_t overlapArea = getOverlapDensityArea(bin,cell);
+            bin->addInstPlacedArea( overlapArea * cell->densityScale() ); 
+            bin->addInstPlacedAreaUnscaled( overlapArea * cell->densityScale() );
```
maliberty, line 842: const and factor the common calculation

src/gpl/src/nesterovBase.cpp
```diff
@@ -855,6 +873,12 @@ BinGrid::updateBinsGCellDensityArea(
           static_cast<float>(bin->instPlacedArea()) 
           + static_cast<float>(bin->nonPlaceArea())
           - (binArea * bin->targetDensity()));
+    
+    overflowAreaUnscaled_ 
+      += std::max(0.0f, 
+          static_cast<float>(bin->instPlacedAreaUnscaled()) 
+          + static_cast<float>(bin->nonPlaceAreaUnscaled())
+          - (binArea * bin->targetDensity()));
```
maliberty, line 881: const and factor the common calculation (shared with overflowArea_ above)

src/gpl/src/nesterovBase.cpp
```diff
@@ -1876,15 +1905,70 @@ getOverlapDensityArea(const Bin* bin, const GCell* cell) {
   if( rectLx >= rectUx || rectLy >= rectUy ) {
     return 0;
   }
-  else {
+  else
+  {
```
maliberty, line 1909: clang-format

src/gpl/src/nesterovBase.cpp
```diff
@@ -1876,15 +1905,70 @@ getOverlapDensityArea(const Bin* bin, const GCell* cell) {
...
+getOverlapArea(const Bin* bin, const Instance* inst, int dbu_per_micron) {
+  int rectLx = max(bin->lx(), inst->lx()), 
+      rectLy = max(bin->ly(), inst->ly()),
+      rectUx = min(bin->ux(), inst->ux()), 
+      rectUy = min(bin->uy(), inst->uy());
+
+  if( rectLx >= rectUx || rectLy >= rectUy ) {
+    return 0;
+  }
+  else {
```
maliberty, line 1927: no need for an else after the return 0

src/gpl/src/nesterovBase.cpp
```diff
@@ -1876,15 +1905,70 @@ getOverlapDensityArea(const Bin* bin, const GCell* cell) {
...
+    if(inst->isMacro())
+    {
+      float meanX = (inst->cx()-inst->lx())/(float)dbu_per_micron;
+      float meanY = (inst->cy()-inst->ly())/(float)dbu_per_micron;
```
maliberty, line 1931: const throughout

src/gpl/src/nesterovBase.cpp
```diff
@@ -1876,15 +1905,70 @@ getOverlapDensityArea(const Bin* bin, const GCell* cell) {
...
+                            (rectLy - inst->ly())/(float)dbu_per_micron, (rectUx - inst->lx())/(float)dbu_per_micron,
+                            (rectUy - inst->ly())/(float)dbu_per_micron};
+
+      float original = static_cast<float>(rectUx - rectLx) 
+        * static_cast<float>(rectUy - rectLy);
+      float scaled = calculateBiVariateNormalCDF(i) * static_cast<float>(inst->ux() - inst->lx()) 
+        * static_cast<float>(inst->uy() - inst->ly());
+      
+      // For heavily dense regions towards the center of the macro,
+      // we are using an upper limit of 2*(overlap) between the macro
```
maliberty, line 1946: "2*" is out of date

src/gpl/src/nesterovBase.cpp
```diff
@@ -1899,6 +1983,24 @@ getOverlapArea(const Bin* bin, const Instance* inst) {
...
+// A function that does 2D integration to the density function of a
+// bivariate normal distribution with 0 correlation.
+// Essentially, the function being integrated is the product
+// of 2 1D probability density functions (for x and y). The means and standard deviation of the 
+// probablity density functions are parametarized.
+// In this function, I am using the closed-form solution of the integration.
+// The limits of integration are lx->ux and ly->uy
+static float calculateBiVariateNormalCDF(biNormalParameters  i)
+{
+  float x1 = (i.meanX-i.lx)/(sqrt(2)*i.sigmaX);
```
maliberty, line 1996: all const

src/gpl/src/nesterovBase.cpp
```diff
@@ -1899,6 +1983,24 @@ getOverlapArea(const Bin* bin, const Instance* inst) {
...
 
+
+// A function that does 2D integration to the density function of a
+// bivariate normal distribution with 0 correlation.
+// Essentially, the function being integrated is the product
+// of 2 1D probability density functions (for x and y). The means and standard deviation of the 
+// probablity density functions are parametarized.
+// In this function, I am using the closed-form solution of the integration.
+// The limits of integration are lx->ux and ly->uy
+static float calculateBiVariateNormalCDF(biNormalParameters  i)
```
maliberty, line 1994: Is there a reference for this formula?  It's nice to provide one for non-obvious math.

maliberty (changes requested):

src/gpl/src/nesterovBase.cpp
ramezmoussa: I added the equation being integrated as a comment.

maliberty (approved):

PR #2405, opened 2022-10-24
support non-root user common dev install

habibayassin:


vvbandeira (changes requested):
I think we should add a GHA to test this script. Could you add an action that runs this on Ubuntu 20 / Ubuntu 22 / CentOS with and without `-prefix`?

etc/DependencyInstaller.sh
```diff
@@ -20,7 +20,7 @@ _installCommonDev() {
     spdlogVersion=1.8.1
 
     # temp dir to download and compile
-    baseDir=/tmp/installers
+    baseDir=${PREFIX}/tmp/installers
```
rovinski, line 23: I don't think you need to prefix `/tmp` because it is accessible by all users by default.

If you want to be extra careful, you can look at using `mktemp` which is  a shell utility that ensures that temporary filenames are unique (in the case of multi-user systems)

rovinski (commented):
As a convenience feature, maybe you could add a `-local` flag which will set the prefix to `$HOME/.local`

habibayassin:
> I think we should add a GHA to test this script. Could you add an action that runs this on Ubuntu 20 / Ubuntu 22 / CentOS with and without `-prefix`?

@vvbandeira is there a github hosted CentOS runner?

vvbandeira:
> > I think we should add a GHA to test this script. Could you add an action that runs this on Ubuntu 20 / Ubuntu 22 / CentOS with and without `-prefix`?
> 
> @vvbandeira is there a github hosted CentOS runner?

I don't think so, but we can use a docker container for this.

rovinski:
@habibayassin here is how you can properly use `mktemp`:

```
baseDir=$(mktemp -d openroad_installer.XXXX)

# install things here

# Cleanup
rm -r $baseDir
```

.github/workflows/github-actions-test-installer.yml
```diff
@@ -0,0 +1,39 @@
...
+          else
+            yum -y install git
+          fi
+      - name: Checkout repository code with submodules
+        run: |
+          git clone --recursive https://github.com/${{ github.repository }}.git .
+          PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
+          git config --global --add safe.directory '*'
+          git fetch origin pull/${PR_NUMBER}/head:test
+          git checkout test
```
vvbandeira, line 33: is this temporary? if not, why not use `uses: actions/checkout@v2`?

.github/workflows/github-actions-test-installer.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: runs installer
+        run: |
+          echo "test" | useradd -m testUser
+          mkdir -p /home/testUser/openroad-deps
```
vvbandeira, line 37: do we need to create a user?

etc/DependencyInstaller.sh
```diff
@@ -332,6 +352,10 @@ case "${os}" in
 To enable GCC-8 or Clang-7 you need to run:
     source /opt/rh/devtoolset-8/enable
     source /opt/rh/llvm-toolset-7.0/enable
+if you have provided a prefix, set cmake as an alias:
+    alias cmake='<YOUR-PREFIX>/bin/cmake' 
+    or  run
+    source ~/.bash_aliases
```
vvbandeira, line 358: we know if the user has used the argument, or at least we can create a flag for that and print a conditional message here. also, we can use the actual prefix in the message instead of the `<YOUR_PREFIX>` -> `$PREFIX`.

etc/DependencyInstaller.sh
```diff
@@ -344,6 +368,10 @@ EOF
             _installCommonDev "${version}"
         fi
         _installUbuntuCleanUp
+        cat <<EOF
+if you have provided a prefix, set cmake as an alias:
+    alias cmake='<YOUR-PREFIX>/bin/cmake'
```
vvbandeira, line 373: same as above, and provide a command the user can copy-paste

.github/workflows/github-actions-test-installer.yml
habibayassin: because it doesn't work properly inside a container in case of recursively pulling submodules

.github/workflows/github-actions-test-installer.yml
habibayassin: not really just used it to create a home directory for the user to kind of simulate the scenario

habibayassin:
commits need to be squashed before merge

.github/workflows/github-actions-test-installer.yml
```diff
@@ -0,0 +1,46 @@
...
+          if [ "${{ matrix.os }}" != "centos:centos7" ]; then 
+            apt-get update
+            apt-get -y install git
+          else
+            yum -y install git
+          fi
+      - name: Checkout repository code with submodules
+        run: |
+          git clone --recursive https://github.com/${{ github.repository }}.git .
+          PR_NUMBER=$(echo $GITHUB_REF | awk 'BEGIN { FS = "/" } ; { print $3 }')
```
vvbandeira, line 30: Will this only work for PRs? Or will it also get a branch name, e.g., `master`?

.github/workflows/github-actions-test-installer.yml
```diff
@@ -0,0 +1,46 @@
+name: Test DependencyInstaller script
+on:
+  # Triggers the workflow on push or pull request events
+  push:
+  pull_request:
```
vvbandeira, line 5: You can keep pull_request as a trigger for tests, but I think we will run this action on a cron job weekly since the dependencies are stable and change rarely.

etc/DependencyInstaller.sh
```diff
@@ -105,12 +111,17 @@ _installCommonDev() {
...
+        ${cmakePrefix}/bin/cmake -B build .
+        ${cmakePrefix}/bin/cmake --build build -j $(nproc) --target install
     else
         echo "spdlog already installed."
     fi
 
+    if [[ ! -z ${PREFIX} ]]; then 
+        export PATH=${PREFIX}/bin:$PATH
+        echo "export PATH=${PREFIX}/bin:$PATH" >> ~/.bash_profile
+    fi
```
vvbandeira, line 123: I don't know if we should modify the users env. Maybe just a suggestion at the end of the script is good enough. If a  user runs this multiple times it can pollute their `bash_profile`. @rovinski @maliberty thoughts? I think this is the policy of homebrew on mac, after installing a package it recommends:
```
Run:
   echo export PATH=${PREFIX}/bin:'$PATH' >> ~/.bash_profile
```
Also note the use of single quotes to let the value of PATH be resolved when sourcing the bash_profile and not when running this script.

etc/DependencyInstaller.sh
```diff
@@ -309,6 +323,12 @@ while [ "$#" -gt 0 ]; do
         -dev|-development)
             option="dev"
             ;;
+        -local)
+            export PREFIX="$HOME/.local"
+            ;;
+        -prefix=*)
+            export PREFIX=`echo $1 | sed -e 's/^[^=]*=//g'`
```
vvbandeira, line 330:
```suggestion
            export PREFIX="$(echo $1 | sed -e 's/^[^=]*=//g')"
```
surround with quotes to consider whitespace? shell can be tricky with spaces, but at least here it should work.

etc/DependencyInstaller.sh
```diff
@@ -347,6 +367,14 @@ case "${os}" in
...
         _installOrTools "centos" "7" "amd64"
         _installCentosCleanUp
+        if [[ ! -z ${PREFIX} ]]; then
+            cat <<EOF
+To use cmake, set cmake as an alias:
+    alias cmake='${PREFIX}/bin/cmake' 
+    or  run
+    source ~/.bash_profile
+EOF
+        fi
```
vvbandeira, line 377: can we do this once at the very end?

.github/workflows/github-actions-test-installer.yml
```diff
@@ -0,0 +1,46 @@
...
+          mkdir -p /home/testUser/openroad-deps
+          ./etc/DependencyInstaller.sh -dev ${{ matrix.args }}
+          if [ "${{ matrix.os }}" == "centos:centos7" ]; then
+            source /opt/rh/devtoolset-8/enable
+            source /opt/rh/llvm-toolset-7.0/enable
+          fi
+          if [ -f ~/.bash_profile ]; then
+            . ~/.bash_profile
+          fi
+          ./etc/Build.sh -clean
```
vvbandeira, line 46: Move this to a different step so we can better see where the failure happened.

.github/workflows/github-actions-test-installer.yml
habibayassin: yes I believe so, I added something more generic in case if it isn't a pull request

etc/DependencyInstaller.sh
```diff
@@ -382,3 +397,19 @@ EOF
         _help
         ;;
 esac
+
+cat <<EOF
+Make sure that CMake find_package can find or-tools. Example, add to your ~/.bashrc:
+    export CMAKE_PREFIX_PATH=/opt/or-tools/lib/cmake
+EOF
+
```
vvbandeira, line 405: this is no longer needed

maliberty:
Is there a reason to have this PR open?  It seems to be under development rather than in review.

habibayassin:
> Is there a reason to have this PR open? It seems to be under development rather than in review.

@maliberty I am working on requested changes and fixes. I can convert it to draft for now

habibayassin:
@vvbandeira I tested the changes done to the GHA and it succeeded here https://github.com/The-OpenROAD-Project/OpenROAD/actions/runs/3493151301/jobs/5847668623 

So I converted it to run as a cron job

.github/workflows/github-actions-test-installer.yml
```diff
@@ -0,0 +1,53 @@
...
+            source /opt/rh/devtoolset-8/enable
+            source /opt/rh/llvm-toolset-7.0/enable
+          fi
+          if [ ! -z "${{ matrix.args }}" ]; then
+            PREFIX=$(echo ${{ matrix.args }}|awk -F"=" '{split($2,a," ");gsub(/"/, "", a[1]);print a[1]}')
+            echo export PATH=${PREFIX}/bin:'$PATH' >> ~/.bash_profile
+            . ~/.bash_profile
+          fi
+          chmod +x ./etc/Build.sh
+          ./etc/Build.sh -clean
```
vvbandeira, line 53:
```suggestion
        os: ["ubuntu20", "ubuntu22", "centos7"]
        args: ["", "-local"]
        compilers: ["gcc", "clang"]
    runs-on: ubuntu-latest
    steps:
      - name: Check out repository code
        uses: actions/checkout@v3
        with:
          fetch-depth: 1
      - name: Runs installer
        run: |
          ./etc/DockerHelper.sh create -target=dev -os=${{ matrix.os }}
      - name: Build project
        run: |
          ./etc/DockerHelper.sh create -target=builder -os=${{ matrix.os }} -compiler=${{ matrix.compiler }}
```

You would also need to add the `-local` argument to `./etc/DockerHelper.sh` to test the functionality of rootless install.

etc/DockerHelper.sh
```diff
@@ -33,6 +33,7 @@ usage: $0 [CMD] [OPTIONS]
   -sha                          Use git commit sha as the tag image. Default is
                                   'latest'.
   -h -help                      Show this message and exits
+  -local                        installs with prefix /home/openroad-deps
```
vvbandeira, line 36:
```suggestion
  -local                        Installs with prefix /home/openroad-deps
```

etc/DockerHelper.sh
```diff
@@ -204,7 +212,10 @@ os="centos7"
 target="dev"
 compiler="gcc"
 useCommitSha="no"
+isLocal="no"
 numThreads="$(nproc)"
+PREFIX="-prefix=/home/openroad-deps"
+LOCAL_PATH="/home/openroad-deps/bin:"
```
vvbandeira, line 218:
```suggestion
LOCAL_PATH="/home/openroad-deps/bin"
```

docker/Dockerfile.builder
```diff
@@ -10,8 +10,11 @@ FROM $fromImage
 
 ARG compiler=gcc
 ARG numThreads=$(nproc)
+ARG LOCAL_PATH=""
 
 COPY . /OpenROAD
 WORKDIR /OpenROAD
 
+ENV PATH=${LOCAL_PATH}${PATH}
```
vvbandeira, line 18:
```suggestion
ENV PATH=${LOCAL_PATH}:${PATH}
```

etc/DockerHelper.sh
```diff
@@ -75,12 +76,19 @@ _setup() {
...
+                buildArgs="${buildArgs} --build-arg LOCAL_PATH=${LOCAL_PATH}"
+            fi
             imageName="${IMAGE_NAME_OVERRIDE:-"${imageName}-${compiler}"}"
             ;;
         "dev" )
             fromImage="${FROM_IMAGE_OVERRIDE:-$osBaseImage}"
             context="etc"
-            buildArgs=""
+            if [[ "${isLocal}" == "yes" ]]; then
+                buildArgs="--build-arg PREFIX=${PREFIX}"
```
vvbandeira, line 88:
```suggestion
                buildArgs="--build-arg INSTALLER_ARGS=-prefix=${LOCAL_PATH}"
```

docker/Dockerfile.dev
```diff
@@ -7,5 +7,8 @@ ARG fromImage=centos:centos7
 
 FROM $fromImage
 
+ARG PREFIX=""
```
vvbandeira, line 10:
```suggestion
ARG INSTALLER_ARGS=""
```

vvbandeira (changes requested):
This looks almost done. Just a couple of suggestions to improve readability.

etc/DockerHelper.sh
habibayassin: LOCAL_PATH has "/home/openroad-deps/bin" not just "/home/openroad-deps"

vvbandeira (approved):

PR #2408, opened 2022-10-24
dpo:edit commands of improvePlacement for better HPWL.

niv280:
Signed-off-by: niv280 <niv280@gmail.com>

Hi,
I change the set of command in improvePlacmentto to give better results.
Here is comparision between the new set commands ( from the left ) to the old command ( in the right):
```
 [INFO ODB-0128] Design: aes_cipher_top	
Original HPWL           372201.2 u	Original HPWL           372201.2 u
Final HPWL              354025.3 u	Final HPWL              360418.2 u
Delta HPWL                  -4.9 %	Delta HPWL                  -3.2 %

[INFO ODB-0128] Design: gcd
Original HPWL             7709.2 u	Original HPWL             7709.2 u
Final HPWL                7154.5 u	Final HPWL                7322.4 u
Delta HPWL                  -7.2 %**	Delta HPWL                  -5.0 %

[INFO ODB-0128] Design: ibex_core
Original HPWL           652460.9 u	Original HPWL           652460.9 u
Final HPWL              636120.8 u	Final HPWL              639841.3 u
Delta HPWL                  -2.5 %	Delta HPWL                  -1.9 %
```

The most affect gain from increase the movePerCandidate from 20 originally to 100, even though the run time degraded I think it worth it, but that results is on small design so please advise if you think that is too much in bigger design.

Other change is change the order of the commands that gave better results.
Also I notice that the **mis** and **reorder** method don't make much different at least on the tested design so I moved them to the end, but I also can comment them to reduce some run time.


src/dpo/test/ibex.ok
```diff
@@ -59,7 +64,9 @@ Importing netlist into detailed improver.
...
 Original HPWL           652460.9 u
-Final HPWL              639841.3 u
-Delta HPWL                  -1.9 %
+Final HPWL              636304.9 u
+Delta HPWL                  -2.5 %
 
-No differences found.
+Differences found at line 5426.
+    - _27724_ AND3_X1 + PLACED ( 975340 826000 ) N ;
+    - _27724_ AND3_X1 + PLACED ( 970020 834400 ) FS ;
```
maliberty, line 72: You need to update the defok file so that no differences are report.  Similarly with the other tests.

maliberty:
@akennings FYI for any comments

src/dpo/test/ibex.ok
niv280: I am getting different def file each time, can it be because of the change of parameters so the def came out a bit different each run?

In the old set of command the run time was ~10sec and in the new it is ~40sec so it is big diff.
I can also reduce the -f flag to 50 it will still give better results but reduce the run time to ~15-20 sec instead.

I didn't try routing it later, I will try it on some design and update.

akennings:
Increasing the moves per candidate seems fine if the run time does not get out of control for larger designs.  Changing the order of commands is also fine.

maliberty:
If the def files varies run to run with the same parameters then that is a problem that needs investigating.  We expect deterministic results.

niv280:
The problem was on my side, the results are deterministic.

The format code check are failing me, it's looks like it expecting change only in c++ files, this is true?

I still need to try to route the design with the new command, I will update about it tomorrow.

maliberty:
The format code check is not blocking for now.  However it is helpful if you can run clang-format on the code to keep it up to date.

niv280:
Hi,
I fixed the clang format error.

Also, I tried to route ibex design with the new command, here what it looked like:
image

Here is with the old set of command:
image


Unfortunately, the short and metal spacing violation was degraded with the new command from 4 to 21 but the overall wire length reduced.
I will try to reduce the moves per candidate and see if I getting better results in the shorts violations.

I think it is also important to compare the congestion but I'm don't sure were I can check it, can you guide me?

antonblanchard:
> Unfortunately, the short and metal spacing violation was degraded with the new command from 4 to 21 but the overall wire length reduced.

Interesting. I was wondering how we prevent DPO from undoing both timing and congestion based global placement. Should it incorporate the STA and grt data (like gpl currently does)?

maliberty:
To check congestion you can look at the congestion map in the GUI or dump it out with "gui::dump_heatmap name filename"

niv280:
The routing congestion looks pretty similar to me.
The old commands:
image

with new commands:
image

What do you suggest for the next steps?

maliberty:
Perhaps it is something more local.  The m3 shorts are new so perhaps take a look and see what is happening there.  Is there increased local congestion that might explain the problem?

QuantamHD:
@maliberty Maybe we should add a feature to compare the congestion values between two runs?

maliberty:
> @maliberty Maybe we should add a feature to compare the congestion values between two runs?

Sure though you can dump the heat map to a csv and do it easily externally.

niv280:
I think that I misunderstood the short in the design.
I used that command for routing:
detailed_route -output_drc dpo_tests/ibex_drc -output_maze dpo_tests/ibex.output.maze.log -output_guide_coverage dpo_tests/ibex.coverage.csv

and the output_drc file is empty, but in the last message in the tool I see the short that I upload above.
Maybe the last drc report in the detailed_route is not the final one?

I tired also the check_drc command but it keep failed to me.

niv280:
I checked it again the drc file is empty so as I understand it there is no drc in both of the command and the route succeed with the new command also there is small improvement in the power ( in aes design) .

Please advise for the next steps here.

QuantamHD:
@maliberty Do you have some advice?

maliberty:
Would you include more of the router log.  I wonder if you are looking at the right message or not.

niv280:
I attached log of 2 runs of ibex_sky130hd test, one with old command and one with new commands.
To the test  ( in flow.tcl) I added the improve_placement and check_placement proc to review the diff.
Let me know if that not enough and you need me to add other command to the test.

Withe new commands:
[ibex_sky130hd.log](https://github.com/The-OpenROAD-Project/OpenROAD/files/10096948/ibex_sky130hd.log)

With old commands:
[ibex_sky130hd_old_commands.log](https://github.com/The-OpenROAD-Project/OpenROAD/files/10096986/ibex_sky130hd_old_commands.log)

rovinski:
@niv280 I don't see actual detailed routing in either of the logs. It looks like pin access is running, but not detailed routing itself. To get the violation count, you should see something in the form of
```
[INFO DRT-0199]   Number of violations = ###.
```

niv280:
This is the detail route command in flow.tcl:
``` TCL
detailed_route -output_drc [make_result_file "${design}_${platform}_route_drc.rpt"] \
 221   │                -output_maze [make_result_file "${design}_${platform}_maze.log"] \
 222   │                -no_pin_access \
 223   │                -save_guide_updates \
 224   │                -bottom_routing_layer $min_routing_layer \
 225   │                -top_routing_layer $max_routing_layer \
 226   │                -verbose  
```

Maybe it is because the verbose 0?

niv280:
I think it was the verbose=0, here is the update logs:

With new commands:
[ibex_sky130hd.log](https://github.com/The-OpenROAD-Project/OpenROAD/files/10099104/ibex_sky130hd.log)

With old commands:
[ibex_sky130hd_old_commands.log](https://github.com/The-OpenROAD-Project/OpenROAD/files/10099106/ibex_sky130hd_old_commands.log)

Sorry for the multiple comments.

rovinski:
From the logs you posted, both versions converge to 0 violations, but the new one takes 11 iterations to do so compared to 10 previously.

niv280:
Yes, this is the downside of the new commands change but we are getting better HPWL 

@maliberty  what do you think?

maliberty:
So long as we can still get to 0 drvs that's ok.  It is expected that denser designs will take longer to route.

niv280:
So, it required anything else to be approved?
I'm just won't be available for the next 2 months from next week, so I want to close it.

rovinski:
@niv280 One thing for sure is that you need to adhere to the formatting rules. Please run clang-format on your code to pass the required formatting check.

src/dpo/test/aes.ok
```diff
@@ -24,42 +24,42 @@ Importing netlist into detailed improver.
...
 [INFO DPO-0311] Found 0 overlaps between adjacent cells.
 [INFO DPO-0312] Found 0 edge spacing violations and 0 padding violations.
 Detailed Improvement Results
 ------------------------------------------
-Original HPWL           372201.2 u
-Final HPWL              360418.2 u
-Delta HPWL                  -3.2 %
+Original HPWL           351168.7 u
+Final HPWL              350977.5 u
+Delta HPWL                  -0.1 %
```
maliberty, line 63: Why is the Original HPWL reduced?  I would expect on the Final & Delta to change.

niv280: I change the also the design def so that will be the original for the regression tests later on.
You can see the comparison between the two set of commands in the image the I upload above.

maliberty: What was the motivation to change the input def?  How was it modified?

niv280: Maybe I misunderstood the flow.
As I understand it, is that every test should be identical in every regression run.
So if the new set of commands approved it need to change also the original def in dpo test since when there will another test running there will be def between the new def to the original because the original will not got the def with the new improve placement commands.

Anyway I can change it again, I just need to understand the correct flow

maliberty: All the unit tests are standalone and don't depend on other tests.  Generally you would update the *ok files not the input files.

maliberty:
I ran ORFS with this change and I see some degradation of results that need investigating (there are some more minor changes that we could just accept):
nangate45/bp_multi - fails in grt due to congestion
nangate45/jpeg - [ERROR] finish__timing__setup__ws fail test: -0.09 >= 0.0

rovinski:
Maybe there should be less ambitious optimization e.g. 50 moves per candidate instead of 100? This is a good candidate for a sensitivity study / possibly throw at the autotuner.

PR #2410, opened 2022-10-24
Don't call atexit() functions from signal handler

antonblanchard:
I hit a SEGV:

  Stack trace:
   0# sta::ClkInfo::refsFilter(sta::StaState const*) const in openroad
   1# sta::TagGroupBldr::setMatchArrival(sta::Tag*, sta::Tag*, float const&, int, sta::PathVertexRep*) in openroad

Which then proceeded to abort and dump the stack a second time:

  terminate called after throwing an instance of 'std::system_error'
    what():  Resource deadlock avoided
  Stack trace:
   0# raise in /lib64/libc.so.6
   1# abort in /lib64/libc.so.6
  ...
  16# sta::Sta::~Sta() in openroad
  17# ord::deleteDbSta(sta::dbSta*) in openroad
  18# ord::OpenRoad::~OpenRoad() in openroad
  20# on_exit in /lib64/libc.so.6

The issue is we are calling exit() which is not async safe and so cannot be used in a signal handler. Considering the application is in a bad state, it also doesn't make sense to try and gracefully shutdown.

Call _exit() instead.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -260,11 +260,22 @@ static void initPython()
 }
 #endif
 
-static void handler(int)
+static volatile sig_atomic_t fatal_error_in_progress = 0;
+
+static void handler(int sig)
 {
+  if (fatal_error_in_progress) {
+    raise(sig);
+  }
+  fatal_error_in_progress = 1;
+
+  std::cerr << "Signal " << sig << " received\n";
+
   std::cerr << "Stack trace:\n";
   std::cerr << boost::stacktrace::stacktrace();
-  exit(1);
+
+  signal(sig, SIG_DFL);
+  raise(sig);
 }
 
 int main(int argc, char* argv[])


maliberty:
https://www.gnu.org/software/libc/manual/html_node/Termination-in-Handler.html suggests a somewhat different method with the comment:
```
     We could just call exit or abort,
     but reraising the signal sets the return status
     from the process correctly.
```

maliberty (approved):
I don't know that it is too important either way so I'll approve and let you decide.

antonblanchard:
> https://www.gnu.org/software/libc/manual/html_node/Termination-in-Handler.html suggests a somewhat different method with the comment:
> 
>      We could just call exit or abort,
>      but reraising the signal sets the return status
>      from the process correctly.

This is a better idea than using `_exit()`, because it means we'll create coredumps. I've modified the PR, and I also added a line to better explain what went wrong (eg `Signal 11 received`).

PR #2411, opened 2022-10-25
drt: Add net name to genGuides_final error messages

antonblanchard:


maliberty (approved):
The pin name would be nice too but at least this is an improvement.

PR #2420, opened 2022-10-26
Add spdlog as submodule

### LINKED ISSUE #2418: suggestion: include fmt and spdlog as submodules?
Building OR is something of a problem now on some operating systems given that fmt recently had a new major version release which package managers such as brew and pacman have upgraded to.

As you already include abc in the build process which is much larger than either of those packages, I'm wondering why not simply build fmt and spdlog? You get the ability to pin both versions to avoid this kind of chaos down the line, and its two less dependencies to install.

donn:
Adds `spdlog` as an OpenROAD submodule just like `abc` to sidestep the whole situation with fmt 9.0 breaking APIs and system package managers.

Also adds `USE_SYSTEM_SPDLOG` as a Cmake option.

---
Resolves #2418


QuantamHD:
I'm personally not in favor of this solution since I think it's avoiding dealing with a rather annoying breakage upstream.

donn:
I respectfully disagree. The library OpenROAD depends on is spdlog, not fmt. OpenROAD is bound to whichever version of fmt spdlog supports, and that version is not fmt 9. Mismatching library versions has always been a bad idea, this situation is just one example of which.

donn:
Misclick-- apologies.

Furthermore, this argument is moot. As it stands, OpenROAD makes the unreasonable demand of expecting users to lock their system package managers to a certain fmt and spdlog. This makes OpenROAD finicky to compile on macOS and Arch derivatives.

This at least mitigates this particular problem.

rovinski:
Has anyone tested if the solution suggested in https://github.com/fmtlib/fmt/issues/3081#issuecomment-1238789389 works? It would be substantially cleaner than any of the current solutions.

maliberty:
I think FMT_DEPRECATED_OSTREAM  was agreed to an one the numerous previous discussions.  I would prefer a PR with that solution to this one.

PR #2421, opened 2022-10-26
Adding Python wrappers for gpl

macd:


src/Main.cc
```diff
@@ -131,6 +133,11 @@ static void initPython()
     exit(1);
   }
 
+  if (PyImport_AppendInittab("_gpl_py", PyInit__gpl_py) == -1) {
+    fprintf(stderr, "Error: could not add module _gpl_py\n");
+    exit(1);
+  }
+
```
rovinski, line 140: I wonder if this could be a function or preprocessor define considering it's going to be made for every tool, e.g.

```
void pyInit(a, b) {
  if (PyImport_AppendInittab(a, b) == -1) {
    fprintf(stderr, "Error: could not add module %s\n", a);
    exit(1);
  }
}
```
or
```
define PYINIT(X) \
  if (PyImport_AppendInittab(X, PyInit__##X) == -1) { \
    fprintf(stderr, "Error: could not add module X\n"); \
    exit(1); \
  }
```
Same for any repeated code.

macd: Yes, I did consider that, but it's always easier to do "just one more". It does seem like the perfect example for metaprogramming, though. My heartache in general is that people get too frisky with it, but I don't think that will be the case here. In any event, that should be in a separate PR. I'll take a look after this one is merged.

rovinski: No worries, just a suggestion.

src/gpl/test/gpl_aux.py
```diff
@@ -0,0 +1,195 @@
...
+            utl.error(utl.GPL, 502, "No liberty libraries found.")
+
+        if skip_io:
+            utl.warn(utl.GPL, 503, "-skip_io will disable timing driven mode.")
+            gpl.setTimingDrivenMode(False)
+
+        if timing_driven_net_reweight_overflow != None:
+            overflow_list = timing_driven_net_reweight_overflow
+        else:
+            overflow_list = [79, 64, 49, 29, 21, 15]
```
maliberty, line 68: Why is this repeated here?  Can't we let the c++ defaults be used instead.

src/gpl/test/gpl_aux.py
```diff
@@ -0,0 +1,195 @@
...
+        gpl.setInitialPlaceMaxFanout(initial_place_max_fanout)
+    elif initial_place_max_fanout != None:
+        utl.error(utl.GPL, 505, "initial_place_max_fanout must be a positive integer")
+
+    uniform_mode = False
+
+    if density != None:
+        target_density = density
+    else: 
+        target_density = DEFAULT_TARGET_DENSITY
```
maliberty, line 95: again c++ default?

src/gpl/test/gpl_aux.py
```diff
@@ -0,0 +1,195 @@
...
+            if not skip_nesterov_place:
+                gpl.doNesterovPlace()
+
+        gpl.reset()
+        gpl.setDb(design.getTech().getDB())
+        gpl.setLogger(design.getLogger())
+        gpl.setGlobalRouter(design.getGlobalRouter())
+        # This will have to wait until rsz is Python wrapped
+        #rz = design.getResizer()
+        #gpl.setResizer(rz)
```
maliberty, line 173: Does this mean that something doesn't currently work in Python?

src/gpl/test/gpl_aux.py
```diff
@@ -0,0 +1,195 @@
+import utl
+
+# Definitions
+DEFAULT_TARGET_DENSITY = 0.7
+
+
+# Besides design, there are no positional args here. General strategy is that 
+# when an arg is None, we just skip setting it, otherwise we will set the
+# parameter after a quick type check.
+def global_placement(design, *,
```
maliberty, line 10: Most test cases pass 1-4 flags.  Couldn't we just have them called in-situ rather than using this big wrapper?

src/gpl/test/simple01-td.py
```diff
@@ -0,0 +1,46 @@
...
+
+# global_placement -timing_driven
+
+# # check reported wns
+# estimate_parasitics -placement
+# report_worst_slack
+
+# set def_file [make_result_file $test_name.def]
+# write_def $def_file
+# diff_file $def_file $test_name.defok
```
maliberty, line 46: dead code

src/gpl/test/gpl_aux.py
macd: Didn't know these were set in C++ and just ported the tcl to Python. I think it is best, as you are suggesting, to take them out of the Python code so that changed defaults in C++ will propagate to the Python api. But, tcl users will be insulated from those changes.

src/gpl/test/gpl_aux.py
macd: No, this just means we haven't yet wrapped rsz in Python. To get the resizer before we wrap it in Python, we would have to update design.evalTclString("...") to return objects. I did update it in this PR to return strings, because that is needed for some conditional tests that are specific to STA, but I did not want to start getting objects back from tcl to Python.

src/gpl/test/simple01-td.py
macd: Ooops. My mistake. I generally delete these sections when the test is successful. Missed this one.

src/gpl/test/gpl_aux.py
macd: We can certainly do that. One of the reasons that I ported the global_place command because I thought that the defaults set there were not done in the C++ code (well that and the error checking). I'll take a look.

src/gpl/test/gpl_aux.py
maliberty: Annoyingly tcl set the defaults in the command to different values than defaults in c++.  I guess the c++ default values are unused in the current setup.  I guess this matches and is therefore ok for now.  Really we need both to stop setting the default and migrate the value to c++

src/gpl/test/gpl_aux.py
maliberty: Sharing the error checking makes sense though again that should move to c++.

src/gpl/test/gpl_aux.py
macd: This doesn't appear to be in the C++ code, it is only in the replace.tcl code.

src/gpl/test/gpl_aux.py
macd: I did look at it more. It appears that there are a handful of defaults that are only set in the tcl (and by extension the gpl_aux.py code). So if we got rid of gpl_aux we would need on the order of 8-10 lines of Python in each of the tests. Moreover, it would be inconvenient to update ~20 tests with new defaults. So I say for now, stay with the gpl_aux wrapper.

src/gpl/test/gpl_aux.py
maliberty: I thought it was, perhaps it moved or I mis-remember.

maliberty (approved):

src/Main.cc
macd: So I have done this now in 2427. Take a look and let me now what you think.

src/gpl/test/erro01.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+# source helpers.tcl
+# set test_name error01 
+# read_lef ./nangate45.lef
+# read_def ./$test_name.def
+
+# catch {global_placement -init_density_penalty 0.01 \
+#          -skip_initial_place -density 0.001} error
+# puts $error
```
maliberty, line 24: dead code

maliberty (approved):
I'll approve.  Please remove the dead code before merging.

macd:
crap, I thought I did! will do.
Don

PR #2427, opened 2022-10-28
Added macros to Main.cc to make adding a Python module now a one line…

macd:
… change. Also changed the name odbpy to odb_py to matching the naming convention of the other Python modules.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -79,20 +79,20 @@ using sta::stringEq;
 using std::string;
 
 #ifdef ENABLE_PYTHON3
-#define PY_MODULES_WITHOUT_OPENROAD \
-  X(ifp)                            \
-  X(utl)                            \
-  X(ant)                            \
-  X(grt)                            \
+#define FOREACH_TOOL_WITHOUT_OPENROAD(X) \
+  X(ifp)                                 \
+  X(utl)                                 \
+  X(ant)                                 \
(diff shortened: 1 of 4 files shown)


src/Main.cc
```diff
@@ -79,13 +79,21 @@ using sta::stringEq;
 using std::string;
 
 #ifdef ENABLE_PYTHON3
+#define PY_MODULES_WITHOUT_OPENROAD \
```
maliberty, line 82: This should take X as an arg

src/Main.cc
```diff
@@ -79,13 +79,21 @@ using sta::stringEq;
...
+  X(ifp)                            \
+  X(utl)                            \
+  X(ant)                            \
+  X(grt)                            \
+  X(odb)
+
+#define PY_MODULES            \
```
maliberty, line 89: This too

src/Main.cc
maliberty: eg https://github.com/The-OpenROAD-Project/OpenROAD/blob/846c27706460b3bbce5d584eaab8c2bec8a3d49b/src/utl/include/utl/Logger.h#L58

macd: Yes, that is less obscure. Will do.

maliberty (approved):

PR #2435, opened 2022-10-30
exclude dbsta and rsz

habibayassin:

### DIFF
--- a/src/dbSta/src/dbSdcNetwork.cc
+++ b/src/dbSta/src/dbSdcNetwork.cc
@@ -48,13 +48,15 @@ dbSdcNetwork::dbSdcNetwork(Network* network) : SdcNetwork(network)
 }
 
 // Override SdcNetwork to NetworkNameAdapter.
-Instance* dbSdcNetwork::findInstance(const char* path_name) const{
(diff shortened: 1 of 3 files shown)


src/dbSta/src/dbSdcNetwork.cc
```diff
@@ -48,15 +48,13 @@ dbSdcNetwork::dbSdcNetwork(Network* network) : SdcNetwork(network)
...
-{
+Instance* dbSdcNetwork::findInstance(const char* path_name) const{
   Instance* inst = network_->findInstance(path_name);
   if (inst == nullptr)
     inst = network_->findInstance(escapeDividers(path_name, this));
   if (inst == nullptr)
     inst = network_->findInstance(escapeBrackets(path_name, this));
-  return inst;
-}
+  return inst;}
```
maliberty, line 57: I guess this is a test change but it should be removed from the PR.  Same with src/rsz/src/EstimateWireParasitics.cc

habibayassin: I was waiting for the review to revert it

vvbandeira (approved):

PR #2438, opened 2022-10-31
src/Main: simplify python flag handling

### LINKED ISSUE #2419: `openroad -python` API breaks
**Describe the bug**
There are a number of API breaks when it comes to `openroad -python` as of late: First, `-exit` is now required as otherwise OpenROAD will drop into a command prompt, which was previously not the case (but overall easy to work around); more importantly, `openroad -exit -python -m` to load a Python module no longer works *at all*. This impedes [DFFRAM](https://github.com/AUCOHL/DFFRAM)'s placeram module.

The behavior used to be that whenever the interpreter saw `-python` it appeared to leave processing the rest of the arguments to the Python executable.

**Expected Behavior**
`openroad -exit -python -m` should work.

**Environment**
```
Git commit: f7df40058721c701313c67009df1e25ba41365cf
kernel: Linux 5.15.65-1-MANJARO
os: Manjaro Linux 
cmake version 3.24.1
gcc (GCC) 12.2.0
clang version 14.0.6
```

**To Reproduce**
```sh
mkdir -p my_module
echo "print('hello, world!')" > my_module/__main__.py
openroad -exit -python -m my_module
```

**Logs**
```
OpenROAD v2.0-5406-gf7df40058 
[WARNING ORD-0039] .openroad ignored with -python
Usage: openroad [-help] [-version] [-no_init] [-exit] [-gui] [-threads count|max] [-log file_name] [-metrics file_name] cmd_file
  -help                 show help and exit
  -version              show version and exit
  -no_init              do not read .openroad init file
  -threads count|max    use count threads
  -no_splash            do not show the license splash at startup
  -exit                 exit after reading cmd_file
  -gui                  start in gui mode
  -python               start with python interpreter [limited to db operations]
  -log <file_name>      write a log in <file_name>
  -metrics <file_name>  write metrics in <file_name> in JSON format
  cmd_file              source cmd_file
Python 3.10.6 (main, Aug  3 2022, 17:39:45) [GCC 12.1.1 20220730] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
```

**Screenshots**

N/A

**Additional context**

The new `[WARNING ORD-0039] .openroad ignored with -python` is thoughtful, but mostly irrelevant to my use-cases. Is there a way to suppress it?

proppy:
- set interpreter config according to command line flags
- forward arguments to python interpreter
- only init python when flag is passed (remove signal save/restore)

Fixes #2419

# Tests

I'd love to add something to the test suite, but https://wiki.tcl-lang.org/page/Expect doesn't seems to be supported by the `test/regression` (and would be needed in order to test proper interaction with the REPL).

```
OpenROAD 🍡 cat foo.py 
import odb
print(odb)
import sys
a = sys.argv
print(a)
```

```
OpenROAD 🍧 build/src/openroad -python foo.py 
OpenROAD v2.0-5434-gb348ed433 
[WARNING ORD-0039] .openroad ignored with -python
<module 'odb' from '/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/odb_py.py'>
['foo.py']
>>> print(a)
['foo.py']
>>> 
```

```
OpenROAD 🍉 build/src/openroad -python -m foo 
OpenROAD v2.0-5434-gb348ed433 
[WARNING ORD-0039] .openroad ignored with -python
<module 'odb' from '/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/odb_py.py'>
['/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/foo.py']
>>> print(a)
['/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/foo.py']
>>> 
```

```
OpenROAD 🍫 build/src/openroad -python -exit foo.py 
OpenROAD v2.0-5434-gb348ed433 
[WARNING ORD-0039] .openroad ignored with -python
<module 'odb' from '/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/odb_py.py'>
['foo.py']
OpenROAD 🍫 build/src/openroad -exit -python foo.py 
OpenROAD v2.0-5434-gb348ed433 
[WARNING ORD-0039] .openroad ignored with -python
<module 'odb' from '/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/odb_py.py'>
['foo.py']
```

```
OpenROAD 💀 build/src/openroad -python -exit -m foo a b c 
OpenROAD v2.0-5434-gb348ed433 
[WARNING ORD-0039] .openroad ignored with -python
<module 'odb' from '/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/odb_py.py'>
['/usr/local/google/home/proppy/src/github.com/The-OpenROAD-Project/OpenROAD/foo.py', 'a', 'b', 'c']
```

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -84,6 +84,7 @@ using std::string;
   X(utl)                                 \
   X(ant)                                 \
   X(grt)                                 \
+  X(gpl)                                 \
   X(odb)
 
 #define FOREACH_TOOL(X)            \
--- a/include/ord/Design.h
+++ b/include/ord/Design.h
@@ -57,6 +57,10 @@ namespace grt {
 class GlobalRouter;
 }
 
+namespace gpl {
+class Replace;
+}
+
 namespace ord {
 
 class Tech;
@@ -81,7 +85,7 @@ class Design
   int micronToDBU(double coord);
 
   // This is intended as a temporary back door to tcl from Python
-  int evalTclString(const std::string& cmd);
+  const std::string evalTclString(const std::string& cmd);
 
   Tech* getTech();
 
@@ -89,6 +93,7 @@ class Design
   ifp::InitFloorplan* getFloorplan();
   ant::AntennaChecker* getAntennaChecker();
   grt::GlobalRouter* getGlobalRouter();
+  gpl::Replace* getReplace();
 
  private:
   Tech* tech_;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -366,6 +366,7 @@ if (Python3_FOUND AND BUILD_PYTHON)
     utl_py
     ant_py
     grt_py
+    gpl_py
   )
 else()
   message(STATUS "Python3 disabled")
--- a/src/Design.cc
+++ b/src/Design.cc
@@ -132,10 +132,11 @@ ant::AntennaChecker* Design::getAntennaChecker()
   return app->getAntennaChecker();
 }
 
-int Design::evalTclString(const std::string& cmd)
+const std::string Design::evalTclString(const std::string& cmd)
 {
   Tcl_Interp* tcl_interp = OpenRoad::openRoad()->tclInterp();
-  return Tcl_Eval(tcl_interp, cmd.c_str());
+  Tcl_Eval(tcl_interp, cmd.c_str());
+  return std::string(Tcl_GetStringResult(tcl_interp));
 }
 
 Tech* Design::getTech()
@@ -147,6 +148,12 @@ grt::GlobalRouter* Design::getGlobalRouter()
 {
   auto app = OpenRoad::openRoad();
   return app->getGlobalRouter();
+}
+
+gpl::Replace* Design::getReplace()
+{
+  auto app = OpenRoad::openRoad();
+  return app->getReplace();
 }
 
 }  // namespace ord
--- a/src/gpl/CMakeLists.txt
+++ b/src/gpl/CMakeLists.txt
@@ -113,3 +113,30 @@ if (GPU)
 else()
   message(STATUS "GPU is not enabled")
 endif()
+
+if (Python3_FOUND AND BUILD_PYTHON)
+  swig_lib(NAME          gpl_py
+           NAMESPACE     gpl
+           LANGUAGE      python
+           I_FILE        src/replace-py.i
+           SWIG_INCLUDES ${PROJECT_SOURCE_DIR}/include/gpl
+           SCRIPTS       ${CMAKE_CURRENT_BINARY_DIR}/gpl_py.py
+  )
+
+  target_include_directories(gpl_py
+    PUBLIC
+      include
+  )
+
+  target_link_libraries(gpl_py
+    PUBLIC
+      utl
+      Eigen3::Eigen
+      gui
+      odb
+      OpenSTA
+      rsz
+      grt
+  )
+
+endif()
--- a/src/gpl/README.md
+++ b/src/gpl/README.md
@@ -87,6 +87,33 @@ resistance and capacitance of estimated wires used for timing.
 
 ## Limitations
 
+## Using the Python interface to gpl
+
+This api tries to stay close to the api defined in C++ class `Replace`
+that is located in gpl/include/gpl/Replace.h
+
+When initializing a design, a sequence of Python commands might look like
+the following:
+
+    from openroad import Design, Tech
+    tech = Tech()
+    tech.readLef(...)
+    design = Design(tech)
+    design.readDef(...)
+    gpl = design.getReplace()
+    
+Here is an example of some options / configurations to the global placer.
+(See Replace.h for a complete list)
+
+    gpl.setInitialPlaceMaxIter(iter)
+    gpl.setSkipIoMode(skip_io)
+    gpl.setTimingDrivenMode(timing_driven)
+    gpl.setTimingNetWeightMax(weight)
+
+There are some useful Python functions located in the file
+grt/test/grt_aux.py but these are not considered a part of the (final)
+api and they may change.
+
 ## FAQs
 
 Check out [GitHub discussion](https://github.com/The-OpenROAD-Project/OpenROAD/discussions/categories/q-a?discussions_q=category%3AQ%26A+replace+in%3Atitle)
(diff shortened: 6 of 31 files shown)


maliberty:
@macd is looking at pytest so perhaps Expect can be handled after that.

src/Main.cc
```diff
@@ -124,7 +125,12 @@ static void initPython()
   FOREACH_TOOL(X)
 #undef X
 
-  Py_Initialize();
+    PyConfig config;
+    PyConfig_InitPythonConfig(&config);
+    PyConfig_SetBytesArgv(&config, argc, argv); 
+    config.inspect = inspect;
+    Py_InitializeFromConfig(&config);
+    PyConfig_Clear(&config);
```
maliberty, line 133: What does this do that is different from Py_Initialize?

src/Main.cc
```diff
@@ -225,12 +231,6 @@ int main(int argc, char* argv[])
   cmd_argc = argc;
   cmd_argv = argv;
 #ifdef ENABLE_PYTHON3
-  // Capture the current SIGINT handler before python changes it.
-  struct sigaction orig_sigint_handler;
-  sigaction(SIGINT, NULL, &orig_sigint_handler);
```
maliberty, line 230: Why is this no longer needed?

src/Main.cc
proppy: it initialize the interpreter with a config (see https://docs.python.org/3/c-api/init_config.html#c.PyConfig.PyConfig_InitPythonConfig) which allow us to set the `inspect` flag to drop into the interpreter after the script execution (instead of exiting) as documented in https://docs.python.org/3/c-api/init_config.html#c.PyConfig.inspect

src/Main.cc
proppy: tried to clarify this in the commit message (but I now realize that wasn't really explicit, sorry!).

The python initialization logic (which is the code path altering the signal) is now only triggered when the `-python` flag is passed, and since the python code path return from `main` there shouldn't be anything to restore in either cases.

proppy:
re-pushed to fix DCO and formatting.

proppy:
oops, it looks like `PyConfig` is >=3.8 only https://docs.python.org/3/c-api/init_config.html#python-initialization-configuration (and the CI runs 3.6 for some of the distribution tested in the matrix).

@maliberty can you confirm if you care about backward compat up to 3.6? (note that it is already EOL'ed: https://devguide.python.org/versions/)

maliberty:
I don't think we need to worry about EOL python versions.

src/Main.cc
maliberty: I'm not sure what you mean by "the python code path return from main".  What happens in the event of a SEGV?

proppy: > I'm not sure what you mean by "the python code path return from main

I agree this was not very precise, what I meant is that the code path that handle the case where the `-python` flag is present returns from the `main` function, effectively existing the `openroad` program.

> What happens in the event of a SEGV?

The default Python Interpreter code path doesnt' seems to trap `SIGSEGV`:
- https://github.com/python/cpython/blob/342b1151ae7e6ae849c1ed7c8a2cbfdb4edcf51c/Modules/signalmodule.c#L1948 
- https://github.com/python/cpython/blob/d291a82df33cd8c917a374fef2a2373beda78b77/Modules/main.c#L653

proppy:
> I don't think we need to worry about EOL python versions.

But the CI environment does seems to build and test openroad against 3.6; is that something we can safely drop from the Jenkins configuration?

src/Main.cc
maliberty: Please test that assumption.  In the past I added that code because python does install its own handler which we don't want to override ours.

proppy: Yes, but the looking at the previous code it seemed that you only restored those handler if the `-python` was not passed (and I assumed this was because `initPython` was inconditionally called).

Was there a good reason we called `initPython` even if `-python` was not passed?

maliberty: python_cmd allows you to call python from tcl.  It was never documented but was more for testing.  We could drop it.

proppy: ah that all make sense now! happy to restore the old code path if needed.

maliberty: My only care is that you get a proper stack trace on error whether in tcl or python.  Feel free to remove python_cmd if that makes things simpler.

arlpetergadfort:
> I don't think we need to worry about EOL python versions.

@maliberty we are still on 3.6 and I do know know when that will change. So it would be nice to preserve the backwards compatibility if possible

maliberty:
> > I don't think we need to worry about EOL python versions.
> 
> @maliberty we are still on 3.6 and I do know know when that will change. So it would be nice to preserve the backwards compatibility if possible

@proppy can you make this work?

maliberty:
Our [CI ](https://jenkins.openroad.tools/job/OpenROAD-Public/job/master/1438/execution/node/89/log/?consoleFull)is also using
-- Found Python3: /usr/lib64/libpython3.6m.so (found version "**3.6.8**") found components:  Development

proppy:
@maliberty not sure what happened?

maliberty:
@proppy did you intend to close this
image

You should be able to reopen if needed

proppy:
@maliberty didn't intent to close it (I don't remember being in front of my computer at that time). I'll reopen another PR I can't seem to be able to reopen this one.
image

proppy:
ah, I think it cames from that automation that I had in my fork to sync the main branch with upstream: https://github.com/The-OpenROAD-Project/OpenROAD/actions/workflows/github-actions-cron-sync-fork-from-upstream.yml.

Because it reset the branch to the current upstream version, it confused GitHub into thinking the content of the PR branch (that happened to be main) was merged.
image

maliberty:
It sounds like that action will overwrite your changes on master so you need to do your changes in a branch in your fork.

proppy:
> 
It sounds like that action will overwrite your changes on master so you need to do your changes in a branch in your fork.

Yep, that/s usually what I do but for some reason I did not for that one :)

PR #2472, opened 2022-11-06
drt/ta: Remove duplicate single threaded and multi threaded code

antonblanchard:


maliberty:
This PR seems to be based on a misunderstanding of the batching.  The batch is not about limiting parallelism (which openmp does) but about limiting memory usage.  We write the results of a batch back to the db and then release the batch before starting a new one.   The workers consume more memory while they are working than the final results they produce.

dr has the same strategy.

antonblanchard:
> This PR seems to be based on a misunderstanding of the batching. The batch is not about limiting parallelism (which openmp does) but about limiting memory usage. We write the results of a batch back to the db and then release the batch before starting a new one. The workers consume more memory while they are working than the final results they produce.

Makes sense, I missed that aspect. Most of the PR is removing the duplication of the `MAX_THREADS=1` and `MAX_THREADS!=1` cases, is it ok to resubmit with the batching restored?

maliberty:
The current setup will be slightly more efficient for threads=1 due to avoiding locking & omp but I think simplicity is better here.  This is likely for having a reference for testing and reporting threading speedup over the base case.  Updating is fine.

antonblanchard:
I've taken out the patch that removes the batching. There is another patch in the updated series that removes some unused code that hasn't been reviewed yet.

maliberty (approved):

PR #2473, opened 2022-11-07
Swig update

habibayassin:


maliberty (approved):

maliberty:
@donn FYI we are updating our swig version

PR #2474, opened 2022-11-08
Rhel support in dependency installer

habibayassin:
semi support for RHEL 8 because
- qt5 is broken on  rhel 8
- there isn't a specific or-tools release for rhel 8


habibayassin:
@vvbandeira 
Still need to install:

- [ ] OR-tools 
- [ ] QT5

.github/workflows/github-actions-test-rhel-dep-install.yml
```diff
@@ -0,0 +1,40 @@
+name: Test RHEL Dependency Installation script
+on:
+  # Allows you to run this workflow manually from the Actions tab
```
vvbandeira, line 3: This file is no longer needed right?

.github/workflows/github-actions-test-installer.yml
```diff
@@ -2,14 +2,16 @@ name: Test DependencyInstaller script
 on:
   schedule:
   - cron: "0 8 * * MON"
+  push:
+  pull_request:
```
vvbandeira, line 6: Remove as we don't want to trigger this on every push/PR. I suggested on other PR to trigger only when the relevant files change.

vvbandeira (changes requested):

.github/workflows/github-actions-test-rhel-dep-install.yml
habibayassin: yes will remove it

vvbandeira (approved):

PR #2481, opened 2022-11-09
dpl python wrapper and tests

macd:


src/dpl/CMakeLists.txt
```diff
@@ -81,3 +81,29 @@ add_custom_target(opendp_tags etags -o TAGS
...
+      include
+  )
+
+  target_link_libraries(dpl_py
+    PUBLIC
+      gui
+      odb
+      OpenSTA
+      utl
+  )
```
maliberty, line 107: Can you just say
```
target_link_libraries(dpl_py
    PUBLIC
      dpl
  )
```

src/dpl/test/dpl_aux.py
```diff
@@ -0,0 +1,91 @@
+import utl
+import re
+
+def detailed_placement(design, *, max_displacement=None, suppress=False):
```
maliberty, line 4: What is '*' for?

src/dpl/test/fence01.ok
```diff
@@ -16,6 +16,4 @@ original HPWL              79.2 u
 legalized HPWL             89.9 u
 delta HPWL                   14 %
 
-Differences found at line 37.
-    - f0/_284_ NOR2_X1 + PLACED ( 30280 28000 ) FS ;
-    - f0/_284_ NOR2_X1 + PLACED ( 29140 28000 ) FS ;
+No differences found.
```
maliberty, line 19: good fix

src/dpl/CMakeLists.txt
macd: So, as written, this doesn't work. You need to add odb, ie
```
target_link_libraries(dpl_py
    PUBLIC
      dpl
      odb
  )
```
The include file Opendp.h is used as the %include file in the -py.i file and this file heavily depends on odb. Generally, my strategy here has been to just mirror what is done for the tcl wrapping. Is there a better way than just trial and error (and on my part that is mostly error) to find out what is superfluous? Anyway, this does work and simplifies the CMakeLists.txt file. Will upload that shortly

src/dpl/test/dpl_aux.py
macd: In Python, all args can be by position or by keyword. This * is a relatively new syntax feature. It forces all args after the * to be keyword only, ie no positional, if you want to change the default values. I think this is a lot safer and it kind of matches the tcl interface.

src/dpl/test/fence01.ok
macd: Thanks!

src/dpl/CMakeLists.txt
maliberty: That's annoying.  Swig gets its include path by a different mechanism than c++ so I guess this is hard to avoid.  I suspect odb is the main case where we include a .i from another module.

src/dpl/test/dpl_aux.py
maliberty: Thanks I learned something :)

maliberty:
The CI failure looks to be false and related to a full disk.  @vvbandeira to address

PR #2485, opened 2022-11-11
Mpl python: Python wrapper and regression tests for mpl

macd:


macd:
The failure on Mac looks to be related to def_parser.cpp and lefr_Reader.cpp, which I've got to believe is spurious to the changes here.

maliberty (approved):

PR #2486, opened 2022-11-11
ant: turn -report_file keyword useful

eder-matheus:


macd:
Maybe think about adding ant_report.py?

src/ant/src/AntennaChecker.cc
```diff
@@ -1157,31 +1159,57 @@ std::pair<bool, bool> AntennaChecker::checkWirePar(const ARinfo& AntennaRatio,
...
+        if (par_violation || verbose) {
+          if (report_file.is_open()) {
+            report_file << "    PAR: " << par << (par_violation ? "*" : " ")
+                        << " Ratio: " << PAR_ratio << " (Area)\n";
+          }
+
           logger_->report("    PAR: {:7.2f}{} Ratio: {:7.2f} (Area)",
                           par,
                           par_violation ? "*" : " ",
                           PAR_ratio);
```
maliberty, line 1171: Rather than duplicating every message it would be better to build the string once using fmt and then output it once or twice depending on the options.

src/ant/test/ant_report.rptok
```diff
@@ -1,18 +1,5 @@
...
-    PAR: 43.1344* Ratio: 15.1501 (S.Area)
-
-    met2
-    PAR: 208.381* Ratio: 10 (S.Area)
-
-    met1
-    PAR: 273.942* Ratio: 10 (S.Area)
-
+Net net50  _264_/B2 (sky130_fd_sc_ms__a222oi_1)    met3    PAR:   43.13* Ratio:   15.15 (S.Area)
+    met2    PAR:  208.38* Ratio:   10.00 (S.Area)
```
maliberty, line 2: It is intentional the met3 ... is on the same line with Net... ?

rovinski:
As long as we are messing with the format can I request a change to something more readable and user-friendly? Like:

```
Net: net1
  Pin: output50/A (sky130_fd_sc_ms__buf_1)
    Layer: met3
      Partial area ratio: 43.13
      Required ratio:     15.15 (side area) (VIOLATED)
    Layer: met2
      Partial area ratio: 208.38
      Required ratio:     10.00 (side area) (VIOLATED)
    Layer: met1
      Partial area ratio: 273.94
      Required ratio:     10.00 (side area) (VIOLATED)
```

Or maybe even as a table
```
Net                   Pin                  Layer    PAR           Required
--------------------------------------------------------------------------------------
net1  output50/A (sky130_fd_sc_ms__buf_1)  met3    43.13  15.15 (side area) (VIOLATED)
net1  output50/A (sky130_fd_sc_ms__buf_1)  met2   208.38  10.00 (side area) (VIOLATED)
net1  output50/A (sky130_fd_sc_ms__buf_1)  met1   273.94  10.00 (side area) (VIOLATED)
```

maliberty:
I expect a table will turn out badly once you have full verilog names which tend to be long.  The former suggestion looks good

maliberty (approved):

PR #2487, opened 2022-11-11
ppl: fix compile warning messages

eder-matheus:

### DIFF
--- a/src/ppl/src/Slots.h
+++ b/src/ppl/src/Slots.h
@@ -50,6 +50,7 @@ namespace ppl {
 class Interval
 {
  public:
+  Interval() = default;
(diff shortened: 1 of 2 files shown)


src/ppl/src/Slots.h
```diff
@@ -50,6 +50,7 @@ namespace ppl {
 class Interval
 {
  public:
+  Interval() = default;
```
maliberty, line 53: What was the warning that required this?

eder-matheus: The default constructor for the `Constraint` struct required this, but I noticed we don't need either of these default actors. I've updated the code in my last commit.

maliberty (approved):

PR #2494, opened 2022-11-14
Ppl python

macd:
This adds the Python wrapper for ppl and the Python regression tests. Note that since IOPlacer.tcl outputs a fair amount of "info" messages, we need to suppress those messages for the regression tests and also update the .ok files.


src/ppl/src/IOPlacer-py.i
```diff
@@ -0,0 +1,77 @@
...
+#include "odb/odb.h"
+#include "odb/db.h"
+
+using std::set;
+using ppl::PinList;    
+using namespace ppl;
+
+template <class TYPE>
+set<TYPE> *
+PyListSet(PyObject *const source,
```
maliberty, line 48: Is this neceesary as swig for Python provides support for std_set.i

macd: I did try (multiple ways, too) with std_set.i but I was not successful. So this just (essentially) mirrors what was done on the tcl side.

maliberty (approved):

PR #2495, opened 2022-11-14
floorplan: add error message when liberty for cell or port is missing

oharboe:


src/ifp/src/InitFloorplan.cc
```diff
@@ -407,6 +407,14 @@ void InitFloorplan::insertTiecells(odb::dbMTerm* tie_term,
 
   auto* port = network_->dbToSta(tie_term);
   auto* lib_port = network_->libertyPort(port);
+  if (!lib_port) {
+    logger_->error(utl::IFP,
+                   39,
+                   "liberty for this cell or the port is missing in the liberty {}/{}.",
```
maliberty, line 413: "in the liberty" seems redundant

No need for a return after this as it will throw an exception.

src/ifp/src/InitFloorplan.cc
```diff
@@ -407,6 +407,14 @@ void InitFloorplan::insertTiecells(odb::dbMTerm* tie_term,
...
+                   master->getName(),
+                   tie_term->getName());
+    return;
```
rovinski, line 416:
```suggestion
                   "Liberty cell or port {}/{} not found.",
                   master->getName(),
                   tie_term->getName());
```

maliberty (approved):

maliberty:
Please note that you are missing the DCO.  Follow the 'details' link for info on fixing that.  In the future, 'git commit -s' takes care of it.

maliberty:
Thanks for the contribution!

PR #2498, opened 2022-11-15
Tap python

macd:
Python wrapper for tap.  Since this is a very small interface, I took a different approach and did not recreate a minimal tap_aux.py file so all the tests are using the native python from wrapping the tapcell.h file. The tests then tend to be more verbose and error checking suffers somewhat as well.


src/tap/include/tap/tapcell.h
```diff
@@ -55,7 +55,7 @@ enum LocationType
 {
   AboveMacro,
   BelowMacro,
-  None,
+  TapNone,
```
maliberty, line 58: How about NoMacro?

maliberty (approved):

PR #2519, opened 2022-11-17
Cts python

macd:
Python wrapper and regression tests for cts. Updated helpers to suppress tcl generated info messages.


macd:
So, I am really puzzled. The failing build on macOS is complaining about tcl things not being declared in a Python wrapper file, such as
```
/Users/openroad/github-action-runners/actions-runner-0/_work/OpenROAD/OpenROAD/build/src/cts/src/CMakeFiles/cts_py.dir/TritonCTS-pyPYTHON_wrap.cxx:4760:23: error: use of undeclared identifier 'interp'
      Tcl_ResetResult(interp);
```
None of these error message with line numbers correspond to the line numbers in my local copy of `TritonCTS-pyPYTHON_wrap.cxx`. @maliberty do you have any tips that can help me find out what is going wrong?

maliberty:
The first error is in _wrap_new_CtsOptions where the exception wrapper
```
    catch (std::exception &excp) {
      Tcl_ResetResult(interp);
```
triggers
```
error: use of undeclared identifier 'interp'
```

maliberty:
I don't really see what is wrong here.  The only odd thing is that you are wrapping CtsOptions.h from src.  Really it should be moved to include if it is part of the public API.  I don't know why that would cause this problem though.  It seems like swig is using Exception.i instead of Exception-py.i but I don't see why.

maliberty:
@vvbandeira any ideas?

src/cts/src/CMakeLists.txt
```diff
@@ -80,3 +80,28 @@ messages(
   TARGET cts
   OUTPUT_DIR ..
 )
+
+if (Python3_FOUND AND BUILD_PYTHON)
+  swig_lib(NAME          cts_py
+           NAMESPACE     cts
+           LANGUAGE      python
+           I_FILE        TritonCTS-py.i
+           SWIG_INCLUDES ${PROJECT_SOURCE_DIR}/include/cts
```
vvbandeira, line 89:
```suggestion
           SWIG_INCLUDES ${PROJECT_SOURCE_DIR}/../include/cts
```
This works on macOS. Why Linux is more forgiving I do not know.

macd: Good catch Vitor, thanks! And you say it is strange that the Linux ones work, but it is also strange that the Mac one just uses the wrong file.  In both cases it should error out with a file not found. I guess there must be some default that is different on the two platforms?

maliberty: Its totally mysterious but I don't want to spend the time to figure it out.

macd: Agreed.

test/helpers.tcl
```diff
@@ -77,3 +77,7 @@ suppress_message PPL 60
 # suppress tap info messages
 suppress_message TAP 100
 suppress_message TAP 101
+
+# suppress cts info messages
+suppress_message CTS 36
+suppress_message CTS 37
```
maliberty, line 83: Why are these hidden?  They are generated in c++ and not language dependent.

macd: That  was my thought too, but it turns out that they seem to be suppressed in some (but not all) tcl log files and also in some (but not all) Python log files. If we remove these directives, then 9 of the regressions will differ, where one will have the messages and the other will not. It is not consistent, so I just disabled for both log files.

maliberty: That because you have a bug:

parms.setRunPostCtsOpt(post_cts_disable)

macd: The default value of `post_cts_disable` is `False`, so this should be exactly equivalent with `  cts::set_disable_post_cts [info exists flags(-post_cts_disable)]` in the tcl version of `clock_tree_synthesis` so will look at it a little more later.

maliberty: You are setting whether to run it to the the value of whether to disable it.

macd:
Is there anything blocking this?

maliberty (commented):
Sorry I had comment but just noticed it was pending

maliberty (approved):

PR #2520, opened 2022-11-18
drt/ta: Fix BTerm coordinate calculation

antonblanchard:
It looks like this code was lifted from FlexDRWorker::initNet_term_new_helper in which trueTerm is a template parameter and can either be an frMTerm or frBTerm.

Fix this by separating out the InstTerm and BTerm cases.


antonblanchard:
@rovinski @maliberty this seems to be what was intended (get the offset of either a InstTerm or BTerm pin).

Iterating access points always confuses me so it would be good to confirm I'm doing the right thing. In particular pinIdx isn't incremented if there is no access point for a pin. This was in the previous code and I see similar behaviour elsewhere in drt:

```
        int pinIdx = 0;
        for (auto& pin : mterm->getPins()) {
          if (!pin->hasPinAccess()) {
            continue;
          }
          frAccessPoint* ap = (static_cast<frInstTerm*>(term)->getAccessPoints())[pinIdx];
          if (ap == nullptr) {
            continue;
          }
...
          pinIdx++
}
```

It looks like we should increment idx if the access point is null, at least looking at this code setting access points:

```
      for (auto mpin : db_term->getMTerm()->getMPins()) {
        if (db_aps_map.find(mpin) == db_aps_map.end()) {
          term->setAccessPoint(idx, nullptr);
        } else {
          term->setAccessPoint(idx, ap_map[db_aps_map[mpin]]);
        }
        ++idx;
      }
```

rovinski:
Glad someone is looking at this. This is a relic from migrating an old part of the DRT database model from a singular `frTerm` to the ODB model of `iTerm` and `bTerm`. I found that in the case of `bTerm`s where there is no instance pointer (`inst == nullptr` in the old model), the code really didn't do anything. I was only trying to keep feature parity during the db migration and not change the output, so it got left as is.

src/drt/src/ta/FlexTA_init.cpp
```diff
@@ -242,70 +245,59 @@ void FlexTAWorker::initIroute_helper_generic_helper(frGuide* guide,
...
-        Point apBp = ap->getPoint();
-        shiftXform.apply(apBp);
-        if (getRouteBox().intersects(apBp)) {
-          wlen2 = isH ? apBp.y() : apBp.x();
-          return;
-        }
-        pinIdx++;
-      }
-    };  // to do @@@@@
-    wlen2 = 0;
```
rovinski, line 308: I remember when I was hacking at this code, removing this `wlen2 = 0` changed the results. Are you sure you want to remove it?

antonblanchard: Good catch, let me restore that for now. I struggled to understand the wlen vs wlen2 code but we can clarify and simplify it in subsequent patches.

antonblanchard: I took another look at this, and I'm not sure how setting wlen2 to zero is ever correct. wlen2 is the x or y coordinate of the pin or access point. If we set it to zero then `assignIroute_getPinCost` will be huge:

```
frUInt4 FlexTAWorker::assignIroute_getPinCost(taPin* iroute, frCoord trackLoc)
{ 
  frUInt4 sol = 0;
  if (iroute->hasWlenHelper2()) {
    sol = abs(trackLoc - iroute->getWlenHelper2());
```

I do plan to clean this up (and create better names, wlen, wlen2 are just confusing).

antonblanchard:
@maliberty noticed a drt issue today that is fixed by this issue. I've added a test case.

antonblanchard:
I've updated the tests. I now have both a pin and access point alignment test, and test in both the horizontal and vertical direction.

src/drt/src/ta/FlexTA_init.cpp
maliberty: Given this change this will need to go through a full secure CI testing just to be safe.  Is this otherwise ready to go?

antonblanchard: The PR is ready. I've tried to decipher the logic here and I'm pretty sure it's a mistake to set wlen2 to 0.

maliberty: I've started the run.

maliberty: The run outcome is good

QuantamHD:
@maliberty looks like this is ready for another review

maliberty:
I think "It looks like we should increment idx if the access point is null" is correct.  Do you want to try that change in this PR?

antonblanchard:
> I think "It looks like we should increment idx if the access point is null" is correct. Do you want to try that change in this PR?

I was able to exercise this path by add a PORT that will have no pin access to a PIN:

```
 MACRO sky130_fd_sc_hd__a21oi_4
   CLASS CORE ;
   FOREIGN sky130_fd_sc_hd__a21oi_4 ;
   ORIGIN  0.000000  0.000000 ;
   SIZE  5.980000 BY  2.720000 ;
   SYMMETRY X Y R90 ;
   SITE unithd ;
   PIN A1
     ANTENNAGATEAREA  0.990000 ;
     DIRECTION INPUT ;
     USE SIGNAL ;
+    PORT
+      LAYER li1 ;
+        RECT 0.0 0.0 0.0 0.0 ;
+    END
     PORT
       LAYER li1 ;
         RECT 2.565000 1.065000 4.000000 1.310000 ;
     END
   END A1
```

I've pushed a patch to this PR which fixes it, by incrementing idx as suggested.

maliberty:
I've started a new secure CI run with your update

maliberty:
asap7/ethmac shows
```
[ERROR] finish__timing__drv__max_slew_limit fail test: -1.5569 >= -1.55
```
which I'm ok to accept.

Issue #2516, opened 2022-11-17
unable to repair antennas


kareefardi:
### Describe the bug

openroad reports antenna violations but it is unable to repair them

### Expected Behavior

repair antenna violations

### OpenROAD Environment

```shell
Git commit: 7f00621cb612fd94e15b35790afe744c89d433a7
kernel: Linux 5.14.0-1054-oem
os: Ubuntu 20.04.3 LTS (Focal Fossa)
cmake version 3.14.0
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
clang version 10.0.0-4ubuntu1
```

### OpenLane Environment

```shell
Kernel: Linux v5.14.0-1054-oem
Distribution: ubuntu 20.04
Python: v3.8.10 (OK)
Container Engine: docker v20.10.11 (OK)
OpenLane Git Version: 1da2eec3d9b5a3612cb026cf9060cb7954e4c24a
pip: INSTALLED
python-venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

1da2eec3 2022-11-16T01:37:27+02:00 [BOT] Update netgen (#1452) - Openlane Bot -  (HEAD -> master, tag: 2022.11.16, origin/master, origin/HEAD)
77fd531e 2022-11-15T13:32:13+02:00 Revert #1485 (#1492) - Mohamed Gaber -  ()
dbb24870 2022-11-15T13:05:26+02:00 [BOT] Update PDK (#1455) - Openlane Bot -  ()
---
Git Remotes

donn	git@github.com:donn/openlane.git (fetch)
donn	git@github.com:donn/openlane.git (push)
fork	git@github.com:kareefardi/OpenLane.git (fetch)
fork	git@github.com:kareefardi/OpenLane.git (push)
origin	git@github.com:The-OpenROAD-Project/OpenLane.git (fetch)
origin	git@github.com:The-OpenROAD-Project/OpenLane.git (push)
```

### To Reproduce

`./run.sh` or `./run.tcl` [issue_reproducible.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10032235/issue_reproducible.tar.gz)

### Relevant log output

```shell
[INFO GRT-0018] Total wirelength: 1570186 um
[INFO GRT-0014] Routed nets: 12882
[INFO GRT-0006] Repairing antennas, iteration 1.
[...]

PR #2524, opened 2022-11-18
Fixes GF180MCU antenna violations

QuantamHD:
The issue looks to be down to GF180MCU being simpler than we expect.

Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/2516

In SKY130 we have antenna rules that look like this.
```
  ANTENNAMODEL OXIDE1 ;
  ANTENNADIFFSIDEAREARATIO PWL ( ( 0 400 ) ( 0.0125 400 ) ( 0.0225 2609 ) ( 22.5 11600 ) ) ;

```

However, in GF180MCU we have rules that look like.
```
    ANTENNAMODEL OXIDE1 ;
    ANTENNADIFFSIDEAREARATIO 400 ;
    ANTENNAGATEPLUSDIFF 2 ;
```

There is only a single value instead of a linearly interpolated list of values.

```c++
bool AntennaChecker::antennaRatioDiffDependent(dbTechLayer* layer)
{
  if (layer->hasDefaultAntennaRule()) {
    const dbTechLayerAntennaRule* antenna_rule = layer->getDefaultAntennaRule();
    dbTechLayerAntennaRule::pwl_pair diffPAR = antenna_rule->getDiffPAR();
    dbTechLayerAntennaRule::pwl_pair diffPSR = antenna_rule->getDiffPSR();
    return diffPAR.indices.size() > 1 || diffPSR.indices.size() > 1;
  }
  return false;
}
```

The following code asserts that there must be at least two values for the rule to be diff dependent which is incorrect. When I changed the code to

```c++
bool AntennaChecker::antennaRatioDiffDependent(dbTechLayer* layer)
{
  if (layer->hasDefaultAntennaRule()) {
    const dbTechLayerAntennaRule* antenna_rule = layer->getDefaultAntennaRule();
    dbTechLayerAntennaRule::pwl_pair diffPAR = antenna_rule->getDiffPAR();
    dbTechLayerAntennaRule::pwl_pair diffPSR = antenna_rule->getDiffPSR();
    return diffPAR.indices.size() >= 1 || diffPSR.indices.size() >= 1;
  }
  return false;
}
```

OpenROAD fixed nearly all of the antennas in this design.

### DIFF
--- a/src/ant/src/AntennaChecker.cc
+++ b/src/ant/src/AntennaChecker.cc
@@ -1997,7 +1997,7 @@ bool AntennaChecker::antennaRatioDiffDependent(dbTechLayer* layer)
     const dbTechLayerAntennaRule* antenna_rule = layer->getDefaultAntennaRule();
     dbTechLayerAntennaRule::pwl_pair diffPAR = antenna_rule->getDiffPAR();
     dbTechLayerAntennaRule::pwl_pair diffPSR = antenna_rule->getDiffPSR();
-    return diffPAR.indices.size() > 1 || diffPSR.indices.size() > 1;
+    return diffPAR.indices.size() >= 1 || diffPSR.indices.size() >= 1;
   }
   return false;
 }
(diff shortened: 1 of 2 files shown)


maliberty (approved):

maliberty:
Thanks

PR #2527, opened 2022-11-20
CMake or-tools for ubuntu and macos

vvbandeira:


maliberty (approved):

maliberty:
@vvbandeira changing CMAKE_INSTALL_RPATH to CMAKE_PREFIX_PATH does not work for or-rack.   I'm going to revert this variable change but leave the extra path.

PR #2528, opened 2022-11-20
drt/ta: Add gui debug

antonblanchard:

### DIFF
--- a/src/drt/src/ta/FlexTA.cpp
+++ b/src/drt/src/ta/FlexTA.cpp
@@ -35,6 +35,7 @@
 #include <iostream>
 #include <sstream>
 
+#include "FlexTA_graphics.h"
 #include "db/infra/frTime.h"
 #include "frProfileTask.h"
 #include "global.h"
@@ -153,6 +154,13 @@ int FlexTAWorker::main_mt()
   }
   return 0;
 }
+
+FlexTA::FlexTA(frDesign* in, Logger* logger)
+    : tech_(in->getTech()), design_(in), logger_(logger)
+{
+}
+
+FlexTA::~FlexTA() = default;
 
 int FlexTA::initTA_helper(int iter,
                           int size,
@@ -412,6 +420,15 @@ void FlexTA::searchRepair(int iter, int size, int offset)
   }
 }
 
+void FlexTA::setDebug(frDebugSettings* settings, odb::dbDatabase* db)
+{
+  bool on = settings->debugTA;
+  graphics_
+      = on && FlexTAGraphics::guiActive()
+            ? std::make_unique<FlexTAGraphics>(settings, design_, db, logger_)
+            : nullptr;
+}
+
 int FlexTA::main()
 {
   ProfileTask profile("TA:main");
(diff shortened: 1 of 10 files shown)


src/drt/src/ta/FlexTA_graphics.cpp
```diff
@@ -0,0 +1,115 @@
...
+      gui_(gui::Gui::get()),
+      top_block_(design->getTopBlock())
+{
+  // Build the layer map between opendb & tr
+  auto odb_tech = db->getTech();
+
+  layer_map_.resize(odb_tech->getLayerCount(), -1);
+
+  for (auto& tr_layer : design->getTech()->getLayers()) {
+    auto odb_layer = odb_tech->findLayer(tr_layer->getName().c_str());
```
maliberty, line 49: Why not use getDbLayer() ?

src/drt/src/ta/FlexTA_graphics.cpp
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  if (settings_->netName.empty()) {
+    return;
+  }
+
+  for (auto& net : top_block_->getNets()) {
+    if (net->getName() != settings_->netName) {
+      continue;
+    }
```
maliberty, line 74: You could look up the net once rather than on every layer draw.  On a big design it might help.

src/drt/src/ta/FlexTA.cpp
```diff
@@ -154,6 +155,16 @@ int FlexTAWorker::main_mt()
...
 
+FlexTA::FlexTA(frDesign* in, Logger* logger)
+    : tech_(in->getTech()), design_(in), logger_(logger)
+{
+}
+
+FlexTA::~FlexTA()
+{
+  // must be out-of-line due to unique_ptr
+}
```
maliberty, line 166: = default;

src/drt/src/ta/FlexTA_graphics.cpp
antonblanchard: Copied from PA graphics. I'll fix this, and submit a PR to fix the other cases.

src/drt/src/ta/FlexTA.cpp
antonblanchard: Thanks, fixed

src/drt/src/ta/FlexTA_graphics.cpp
antonblanchard: I'm wondering how best to cache the `FrNet *`. Would we have to handle updates, eg
```
set_detailed_route_debug -net net1
detailed_route
set_detailed_route_debug -net net2
detailed_route
```

I'm not sure if the detailed routing instance is reused or deleted and recreated.

maliberty: Just like you have a endIter callback you could have a start callback where you do the lookup.

antonblanchard: Makes sense. I just did it in endIter, since the data only gets written back to the database at the end of an iteration.

I also changed it to dump all iroutes if `-net` isn't specified, since I thought it might be useful to visualize.

maliberty (approved):

PR #2548, opened 2022-11-29
GHA macOS uses installer

vvbandeira:


etc/DependencyInstaller.sh
```diff
@@ -389,8 +396,9 @@ EOF
 
 To install or run openroad, update your path with:
     export PATH="\$(brew --prefix bison)/bin:\$(brew --prefix flex)/bin:\$(brew --prefix tcl-tk)/bin:\$PATH"
+    export CMAKE_PREFIX_PATH=\$(brew --prefix)/opt/or-tools
```
rovinski, line 399: Are you sure this is the correct path? Typically the path on M1 silicon is different from x86 silicon. I believe `$(brew --prefix)` is supposed to handle whether the package is installed in `/opt` or not?

rovinski (commented):
Seems ok to me

PR #2550, opened 2022-11-30
ant/grt: updates for repair_antennas

eder-matheus:


QuantamHD:
Does this need to be upstreamed to OpenLane?

eder-matheus:
> Does this need to be upstreamed to OpenLane?

Not yet. This WIP requires a few more updates before going to production. I'm just creating this PR to avoid a PR with lots of updates in the future.

src/grt/src/GlobalRouter.tcl
```diff
@@ -274,11 +274,12 @@ proc global_route { args } {
...
 
 sta::define_cmd_args "repair_antennas" { [diode_cell/diode_port] \
-                                         [-iterations iterations]}
+                                         [-iterations iterations] \
+                                         [-par_margin par_margin]}
 
 proc repair_antennas { args } {
   sta::parse_key_args "repair_antennas" args \
-                 keys {-iterations}
+                 keys {-iterations -par_margin}
```
jjcherry56: I suggest renaming this to ratio_margin, since it applied to PAR and PSR.

src/grt/src/GlobalRouter.tcl
```diff
@@ -314,7 +315,13 @@ proc repair_antennas { args } {
       sta::check_positive_integer "-iterations" $iterations
     }
 
-    grt::repair_antennas $diode_mterm $iterations
+    set par_margin 0
+    if { [info exists keys(-par_margin)] } {
+      set par_margin $keys(-par_margin)
+      sta::check_positive_float "-par_margin" $par_margin
```
jjcherry56: I am guessing the parameter is a percentage, which for consistency with the resizer margins should be between 0 and 100. There is no check on the range here. See src/rsz/src/Resizer.tcl repair_design.
Also the README is not updated to doc the arg.

src/grt/README.md
```diff
@@ -125,15 +125,18 @@ Example:
...
 ```
 
 The repair_antenna command checks the global routing for antenna
 violations and repairs the violations by inserting diodes near the
 gates of the violating nets.  By default the command runs only one
 iteration to repair antennas. Filler instances added by the
 `filler_placement` command should NOT be in the database when
-`repair_antennas` is called.
+`repair_antennas` is called. Use `-ratio_margin` to add a margin
+to the antenna ratios.
```
jjcherry56, line 139: you should explain it is a percentage between 0 and 100

jjcherry56 (approved):

PR #2555, opened 2022-12-01
rcx: Use ADS_MAX_CORNER

antonblanchard:
Instead of hard coding 5, 10 or 16, use the ADS_MAX_CORNER define which is used elsewhere in the rcx source.


QuantamHD:
@maliberty should we merge?

maliberty:
@QuantamHD thanks for the ping

PR #2556, opened 2022-12-01
drt/ta: Use getMinSpacingValue()

antonblanchard:


maliberty:
Are there more of these to do still?

PR #2569, opened 2022-12-03
Support Debian 10 & Debian 11

habibayassin:


.github/workflows/github-actions-test-installer.yml
```diff
@@ -2,14 +2,16 @@ name: Test DependencyInstaller script
 on:
   schedule:
   - cron: "0 8 * * MON"
+  push:
+  pull_request:
```
vvbandeira, line 6: The following might be a good improvement, it should be triggered on etc/DependencyInstaller and docker/* at least. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files

etc/DependencyInstaller.sh
```diff
@@ -292,6 +292,65 @@ EOF
...
+            qt5-qmake \
+            qtbase5-dev-tools
+    else
+        apt-get install -y \
+            libpython3.7 \
+            qt5-default
+    fi
+
+    # need the strip "hack" above to run on docker
+    strip --remove-section=.note.ABI-tag /usr/lib/x86_64-linux-gnu/libQt5Core.so
```
vvbandeira, line 351: Is this something we want to move to Dockerfile or does this has no effect on installs outside docker?

vvbandeira (changes requested):

etc/DependencyInstaller.sh
habibayassin: I copied it from ubuntu's installation because Debian is so similar to it, however, I can try to push the code without it to notice its effect.

vvbandeira (approved):
Looks good, just need to fix the conflicts due to the RHEL PR.

PR #2574, opened 2022-12-05
Drt python

macd:
Python wrapper and Python regression for TritonRoute

### DIFF
--- a/include/ord/Tech.h
+++ b/include/ord/Tech.h
@@ -54,6 +54,8 @@ class Tech
   void readLef(const std::string& file_name);
   void readLiberty(const std::string& file_name);
   odb::dbDatabase* getDB();
+  void set_thread_count(const int threads);
+  int thread_count();
 
  private:
   odb::dbDatabase* db_;
--- a/include/ord/Tech.h
+++ b/include/ord/Tech.h
@@ -54,8 +54,6 @@ class Tech
   void readLef(const std::string& file_name);
   void readLiberty(const std::string& file_name);
   odb::dbDatabase* getDB();
-  void set_thread_count(const int threads);
-  int thread_count();
 
  private:
   odb::dbDatabase* db_;
--- a/src/drt/CMakeLists.txt
+++ b/src/drt/CMakeLists.txt
@@ -240,16 +240,7 @@ if (Python3_FOUND AND BUILD_PYTHON)
 
   target_link_libraries(drt_py
     PUBLIC
-      odb
-      stt
-      OpenSTA
-      utl
-      dst
-      dbSta
-      Threads::Threads
-      OpenMP::OpenMP_CXX
-      ${Boost_LIBRARIES}
-      ZLIB::ZLIB
+      drt
   )
 
 endif()
--- a/src/drt/CMakeLists.txt
+++ b/src/drt/CMakeLists.txt
@@ -223,3 +223,33 @@ if (VTune_FOUND)
   )
 
 endif(VTune_FOUND)
+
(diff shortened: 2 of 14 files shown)


include/ord/Tech.h
```diff
@@ -54,6 +54,8 @@ class Tech
   void readLef(const std::string& file_name);
   void readLiberty(const std::string& file_name);
   odb::dbDatabase* getDB();
+  void set_thread_count(const int threads);
+  int thread_count();
```
maliberty, line 58: Since these are application level APIs I think they could be free functions in the ord namespace rather than Tech members, e.g.
```
ord::set_thread_count(40)
```

src/drt/CMakeLists.txt
```diff
@@ -223,3 +223,33 @@ if (VTune_FOUND)
...
+           SWIG_INCLUDES ${PROJECT_SOURCE_DIR}/include/drt
+           SCRIPTS       ${CMAKE_CURRENT_BINARY_DIR}/drt_py.py
+  )
+
+  target_include_directories(drt_py
+    PUBLIC
+      include
+  )
+
+  target_link_libraries(drt_py
```
maliberty, line 241: I thought we could have drt_py depend on drt rather than duplicating the dependencies.

src/drt/test/drt_aux.py
```diff
@@ -0,0 +1,76 @@
...
+                   min_access_points=-1,
+                   save_guide_updates=False):
+
+    router = design.getTritonRoute()
+
+    if param != "":
+        utl.warn(utl.DRT, 8001, "param file overrides all other arguments")
+        router.readParams(param)
+        router.main()
+        return
```
maliberty, line 40: The param file is for legacy support, we can drop it from Python

src/drt/test/ispd18_sample.py
```diff
@@ -0,0 +1,38 @@
...
+# read_def testcase/ispd18_sample/ispd18_sample.input.def
+# read_guides testcase/ispd18_sample/ispd18_sample.input.guide
+# detailed_route -output_drc results/ispd18_sample.output.drc.rpt \
+#                -output_maze results/ispd18_sample.output.maze.log \
+#                -output_guide_coverage results/ispd18_sample.coverage.csv \
+#                -verbose 0
+
+# set def_file [make_result_file ispd18_sample.def]
+# write_def $def_file
+# diff_files ispd18_sample.defok $def_file
```
maliberty, line 38: remove

maliberty (changes requested):

src/drt/CMakeLists.txt
macd: So I am not certain how this would work  in cmake. Can you give me an example?

src/drt/test/drt_aux.py
macd: Will do.

src/drt/test/ispd18_sample.py
macd: Ooops. Will do.

src/drt/CMakeLists.txt
macd: Actually, my first guess was correct. So this is done.

maliberty: You did this previously in https://github.com/The-OpenROAD-Project/OpenROAD/blob/6052d6a75bfbb6308f4f70820388afd34123407f/src/dpl/CMakeLists.txt#L96

maliberty (approved):

PR #2578, opened 2022-12-06
Dpo python

macd:
Python wrapper and regression tests for dpo

### DIFF
--- a/src/dpo/test/aes.py
+++ b/src/dpo/test/aes.py
@@ -1,0 +1,15 @@
+from openroad import Design, Tech
+import helpers
+
+tech = Tech()
+tech.readLef("Nangate45/Nangate45.lef")
+
+design = Design(tech)
+design.readDef("aes.def")
+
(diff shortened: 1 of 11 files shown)


src/dpo/test/aes.py
```diff
@@ -0,0 +1,15 @@
...
+import helpers
+
+tech = Tech()
+tech.readLef("Nangate45/Nangate45.lef")
+
+design = Design(tech)
+design.readDef("aes.def")
+
+design.getOptdp().improvePlacement(1, 0, 0)
+design.evalTclString("check_placement")
```
maliberty, line 11: Couldn't you call this through dpl?

macd: Good catch. I think I wrote these before dpl was wrapped. There are others that can probably be removed as we finish the Python wrapping. Will need to make another pass when most of the wrapping is done, but I have fixed these.

maliberty (approved):

PR #2581, opened 2022-12-06
drt: avoid creating unnecessary patch metal

eder-matheus:
Fixes OR-1638

### DIFF
--- a/src/drt/src/dr/FlexDR_maze.cpp
+++ b/src/drt/src/dr/FlexDR_maze.cpp
@@ -3079,13 +3079,15 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio(
       }
       // init for next path
       if (currIdx.z() < prevIdx.z()) {
+        // get the bottom layer box of the current via to initialize the area
+        // for the next shape
         currArea = getHalfViaEncArea(
-            prevIdx.z() - 1, true, net->getFrNet()->getNondefaultRule());
+            prevIdx.z(), true, net->getFrNet()->getNondefaultRule());
         startViaHalfEncArea = getHalfViaEncArea(
-            prevIdx.z() - 1, true, net->getFrNet()->getNondefaultRule());
+            prevIdx.z(), true, net->getFrNet()->getNondefaultRule());
       } else {
-        currArea = getHalfViaEncArea(
-            prevIdx.z(), false, net->getFrNet()->getNondefaultRule());
+        // get the top layer box of the current via to initialize the area
+        // for the next shape
         currArea = getHalfViaEncArea(
             prevIdx.z(), false, net->getFrNet()->getNondefaultRule());
         startViaHalfEncArea = gridGraph_.getHalfViaEncArea(prevIdx.z(), false);
--- a/src/drt/src/dr/FlexDR_maze.cpp
+++ b/src/drt/src/dr/FlexDR_maze.cpp
@@ -3128,7 +3128,9 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio(
       gridGraph_.getPoint(ep, nextIdx.x(), nextIdx.y());
       frCoord pathLength = abs(bp.x() - ep.x()) + abs(bp.y() - ep.y());
       if (currArea < reqArea) {
-        currArea /= 2;
+        if (!prev_is_wire) {
+          currArea /= 2;
+        }
         currArea += pathLength * pathWidth;
       }
       prev_is_wire = true;
--- a/src/drt/src/dr/FlexDR_maze.cpp
+++ b/src/drt/src/dr/FlexDR_maze.cpp
@@ -2998,10 +2998,13 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio(
       if (nextIdx.z() < currIdx.z()) {
         if (prev_is_wire) {
           currArea += getHalfViaEncArea(
-            currIdx.z() - 1, false, net->getFrNet()->getNondefaultRule());
+              currIdx.z() - 1, false, net->getFrNet()->getNondefaultRule());
         } else {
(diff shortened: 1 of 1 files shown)


src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2849,7 +2849,7 @@ void FlexDRWorker::routeNet_prepAreaMap(drNet* net,
       FlexMazeIdx mIdx = ap->getMazeIdx();
       auto it = areaMap.find(mIdx);
       if (it != areaMap.end()) {
-        it->second = max(it->second, ap->getBeginArea());
+        areaMap[mIdx] = max(it->second, ap->getBeginArea());
```
maliberty, line 2852: why the change?

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -3088,44 +3105,53 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio(
...
-      if (currIdx.z() < prevIdx.z()) {
-        currArea = getHalfViaEncArea(
-            prevIdx.z() - 1, true, net->getFrNet()->getNondefaultRule());
+      if (nextIdx.z() < currIdx.z()) {
+        // get the bottom layer box of the current via to initialize the area
+        // for the next shape
+        currArea
+            = getHalfViaEncArea(
+                  currIdx.z() - 1, true, net->getFrNet()->getNondefaultRule())
+              * 2;
```
maliberty, line 3114: Is this correct if the next shape is a wire?

eder-matheus: I've tried to compensate for it in this line:
https://github.com/The-OpenROAD-Project/OpenROAD/pull/2581/files/b9d5dfea7a109560dcdc3f6625afec2c338280ee#diff-3b0ee5af511f4838323555d94ad93cc7da9df71a64bf049fa6b634da77978e64R3141

eder-matheus:
@maliberty Secure-ci is failing for ca53, with the same 2 DRCs we have on master. It also fails for gf12/swerv_wrapper, but on a CTS metric that is not related to my change. I think it's safe to merge this PR.

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2982,28 +2982,45 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio(
     currArea = (minAreaConstraint) ? minAreaConstraint->getMinArea() : 0;
   }
   frCoord startViaHalfEncArea = 0, endViaHalfEncArea = 0;
-  FlexMazeIdx prevIdx = points[0], currIdx;
+  FlexMazeIdx currIdx = points[0], nextIdx;
   int i;
   int prev_i = 0;  // path start point
+  bool prev_is_wire = true;
```
maliberty, line 2988: The terminology is getting mixed.  You changed prev/curr to curr/next but kept prev_i/prev_is_wire.

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -2982,28 +2982,45 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio(
...
+      layerNum = gridGraph_.getLayerNum(currIdx.z());
       minAreaConstraint = getTech()->getLayer(layerNum)->getAreaConstraint();
       frArea reqArea
           = (minAreaConstraint) ? minAreaConstraint->getMinArea() : 0;
-      // add next via enclosure
-      if (currIdx.z() < prevIdx.z()) {
-        currArea += getHalfViaEncArea(
-            prevIdx.z() - 1, false, net->getFrNet()->getNondefaultRule());
+      // add curr via enclosure
+      if (nextIdx.z() < currIdx.z()) {
```
maliberty, line 2998: The if and else blocks are nearly identical.  Rather than having two copies, just assign two variables and use a common block.  "currIdx.z() - 1, false" vs "currIdx.z(), true" appears to be the only difference.

src/drt/src/dr/FlexDR_maze.cpp
maliberty: What happens when there are multiple wire segments in series.  It seems "currArea /= 2;" would reduce the area of the previous segment not only the via.

src/drt/src/dr/FlexDR_maze.cpp
eder-matheus: I've kept prev_i and prev_is_wire because they're actually referring to the previous objects. If I understand the code correctly, the vars that I've changed to curr/next didn't have a good naming with prev/curr.

src/drt/src/dr/FlexDR_maze.cpp
eder-matheus: I've updated the code to perform `currArea /= 2;` only when the last object is not a wire. I believe this will prevent reducing the area of all the previous segments.

maliberty: My guess is this won't hurt anything.  Do you want to roll the dice or do another secure CI?

maliberty:
Please avoid a force push after a review.  You can see changes since the last review afterwards.

eder-matheus:
> Please avoid a force push after a review. You can see changes since the last review afterwards.

Didn't know about it. My bad! I'll avoid rebasing and force-pushing in the future.

maliberty:
Prior to review that's fine.

src/drt/src/dr/FlexDR_maze.cpp
eder-matheus: We've waited a long time for this PR, and waiting a few more hours is fine. I'll start the secure-ci now.

PR #2582, opened 2022-12-06
gui: add inverter shape to CTS viewer

gadfort:
Adds:
- inverter graphics item for the CTS viewer incase the output port is inverting.
inverter


maliberty (approved):
Would it have been simpler to just making inverting an option on the existing buffer class?  It seems like the delta is just an ellipse.

gadfort:
> Would it have been simpler to just making inverting an option on the existing buffer class? It seems like the delta is just an ellipse.

Yeah, that would be easier. I'll make the switch.

PR #2583, opened 2022-12-06
gui: add qtmessagehandler and save images without transparency

gadfort:
Adds:
- qtmessagehandler to redirect qt messages to the logger and suppress warnings about plugins when compiled with NDEBUG.

Changes:
- Converts images before saving to avoid transparency in the image. @maliberty could you try this and see if this solves the background issue you are having?

### DIFF
--- a/src/gui/src/clockWidget.cpp
+++ b/src/gui/src/clockWidget.cpp
@@ -564,8 +564,6 @@ ClockTreeScene::ClockTreeScene(QWidget* parent)
   connect(menu_->addAction("Save"), SIGNAL(triggered()), this, SIGNAL(save()));
 
   clear_path_->setEnabled(false);
-
-  setBackgroundBrush(Qt::white);
 }
 
 void ClockTreeScene::contextMenuEvent(QGraphicsSceneContextMenuEvent* event)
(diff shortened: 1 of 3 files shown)


maliberty:
It doesn't help with the color.  I loaded the webp into gimp and I see the alpha is 255 in the background area:
image

gadfort:
> With the DISPLAY set: image

grrrrrrrr, okay, I'll keep poking around

maliberty:
Even in master alpha=255

maliberty (approved):
The message handler piece is good.

gadfort:
> The message handler piece is good.

@maliberty I've tried setting the background to white instead of nobrush (when you have a chance could you see if that fixes it?) If it doesn't I'll revert is out and then I'm out of ideas at the moment for how to solve it.

maliberty:
No change, still blackish

gadfort:
> No change, still blackish

alright, well worth a shot, so for now I'm out of ideas.

QuantamHD:
It might help to drop into gdb and step through the image writing function. Eventually you'll get to a place where the background color is being chosen, and it'll become clear why

maliberty:
Unfortunately Peter can't reproduce my problem so I'll have to do that myself

PR #2585, opened 2022-12-07
gui: ensure renderImage does not request an impossibly large QImage

gadfort:
Fixes:
- automatic scaling of renderImage limited to 7200px, preventing max image size violations.

### DIFF
--- a/src/gui/src/gui_utils.cpp
+++ b/src/gui/src/gui_utils.cpp
@@ -78,16 +78,11 @@ QString Utils::fixImagePath(const QString& path, utl::Logger* logger)
   return fixed_path;
 }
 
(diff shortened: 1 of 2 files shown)


gadfort:
@maliberty this should fix the issue you ran into with the max size.

maliberty:
Not so good, its just cut off:
image

gadfort:
> Not so good, its just cut off: image

@maliberty I've adjusted rendering pixel ratio, which should fix it. I have not been able to replicate your exact issue, so I have no great way to test this unfortunately.

maliberty:
replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/2587

Issue #1197, opened 2021-10-13
[INFO DRT-1000] Pin u_sram_2kb/din0[15] not in any guide. Attempting to patch guides to cover (at least part of) the pin.


dineshannayya:
Is this information is saying any specific issue here ?  I see u_sram_2kb/din0[15]  connected to mem_din_b\[15\] and this signal information is available inside guide file : 16-fastroute.guide

openroad version: OpenROAD v2.0-1285-g92dd19146 

Attached the setup
[exp3.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/7336972/exp3.tar.gz)

Run command: run_cmd

In the MPW-3 Efabless openroad version: OpenROAD 1 fb8ae93b6c7a5eb0e6fac83360a8a48d76c41885
This  reported as ERROR below:
<pre>[WARNING DRT-0215] Pin u_sram_2kb/din0[15] not covered by guide.
[WARNING DRT-0215] Pin u_sram_2kb/din0[15] not covered by guide.
[WARNING DRT-0225] mem_din_b\[15\] 2 pin not visited, fall back to feedthrough mode.
[WARNING DRT-0224] mem_din_b\[15\] 2 pin not visited, number of guides = 7.
[ERROR DRT-0218] Guide is not connected to design.
Error: or_droute.tcl, 34 DRT-0218
<font color="#CC0000">[ERROR]: during executing: &quot;openroad -exit /openLANE_flow/scripts/openroad/or_droute.tcl |&amp; tee &gt;&amp;@stdout /project/openlane/user_project_wrapper/runs/13-10_09-31/logs/routing/18-tritonRoute.log&quot;</font>
</pre>

eder-matheus:
@dineshannayya I've run the attached testcase on a CentOS 7 machine and I see the `[INFO DRT-1000] Pin u_sram_2kb/din0[15] not in any guide. Attempting to patch guides to cover (at least part of) the pin.` message, but it is not failing with the latest OpenROAD. Could you try it with the latest version of OpenROAD?

dralabeing:
This sounds like an Openlane bug as well  based on the initial version reported on MPW-3.
@maliberty @eder-matheus Should this also be submitted against OpenLane to fix against that OR version too since that is the latest shuttle version?

eder-matheus:
[...]

PR #2600, opened 2022-12-09
grt: snap pin position on grid only when track is not at the edge of the shape

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/1197


rovinski:
Maybe a small test case would be good?

PR #2604, opened 2022-12-11
test build in test installer GHA

habibayassin:


vvbandeira:
The goal should be an easy pass/fail test to avoid things like #2608, I think we can have just an invocation of `openroad -h` to start with since running unit tests will likely result in slightly different results as we have seen in the past.

vvbandeira (approved):

PR #2606, opened 2022-12-11
Unify options for installing dependencies 

habibayassin:


etc/DependencyInstaller.sh
```diff
@@ -494,21 +489,17 @@ EOF
...
         ;;
     "Red Hat Enterprise Linux")
         spdlogFolder="/usr/local/lib64/cmake/spdlog/spdlogConfigVersion.cmake"
         export spdlogFolder
         _installRHELRuntime
-        if [[ "${option}" == "dev" ]]; then
-            _installRHELDev
-            _installCommonDev
-        fi
+        _installRHELDev
```
vvbandeira, line 501: The goal of this PR should also include replacing both`_install*Runtime` and `_install*Dev` with a single function. Bonus if we can remove any unnecessary packages.

vvbandeira (approved):

PR #2607, opened 2022-12-12
grt: prevent usage from wrapping

antonblanchard:
The following test case fails in global routing with:

```
[ERROR GRT-0229] Vertical edge usage exceeds the maximum allowed. (88, 22) usage=65535 limit=1240
```

I'm not sure this is the correct fix, but it gets me past the fail. 

Test case: [grt-usage-wraps.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10204601/grt-usage-wraps.tar.gz)

### DIFF
--- a/src/grt/src/fastroute/src/utility.cpp
+++ b/src/grt/src/fastroute/src/utility.cpp
@@ -1210,7 +1210,7 @@ void FastRouteCore::removeLoops()
 
     const auto& treeedges = sttrees_[netID].edges;
 
-    const int edgeCost = nets_[netID]->getEdgeCost();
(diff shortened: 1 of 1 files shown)


arthurjolo:
Hi, this error occurs indeed because of this method removeLoops(), however this is not the correct fix,it is just avoiding the problem.
This should fix it: [#2640](https://github.com/The-OpenROAD-Project/OpenROAD/pull/2640)
Over there is a better explanation on the error

eder-matheus:
@antonblanchard we've merged an update that should fix the underflow in the usage values. We've tried with your testcase and it fixes the issue. I'll close your PR, but let me know if you still have issues with usage values.

PR #2613, opened 2022-12-13
drt: parallelize the standalone drc check

osamahammad21:

### DIFF
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -802,13 +802,9 @@ void TritonRoute::pinAccess(std::vector<odb::dbInst*> target_insts)
   writer.updateDb(db_, true);
 }
 
-void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
+void TritonRoute::getDRCMarkers(frList<std::unique_ptr<frMarker>>& markers,
+                                const Rect& requiredDrcBox)
 {
-  initDesign();
-  Rect box(x1, y1, x2, y2);
-  if (box.area() == 0) {
-    box = design_->getTopBlock()->getBBox();
-  }
   MAX_THREADS = ord::OpenRoad::openRoad()->getThreadCount();
   std::vector<std::unique_ptr<FlexGCWorker>> workers;
   auto size = 7;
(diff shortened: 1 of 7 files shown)


src/drt/src/TritonRoute.cpp
```diff
@@ -809,16 +809,77 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
   frList<std::unique_ptr<frMarker>> markers;
-  for (auto& marker : gcWorker->getMarkers())
-    markers.push_back(std::make_unique<frMarker>(*marker));
+  std::map<std::tuple<Rect,
+                      frLayerNum,
+                      frConstraint*,
+                      frBlockObject*,
+                      frBlockObject*>,
+           frMarker*>
+      mapMarkers_;
```
maliberty, line 849: Local variables should not have an _ suffix.

src/drt/src/TritonRoute.cpp
```diff
@@ -809,16 +809,77 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+    }
+  }
   frList<std::unique_ptr<frMarker>> markers;
-  for (auto& marker : gcWorker->getMarkers())
-    markers.push_back(std::make_unique<frMarker>(*marker));
+  std::map<std::tuple<Rect,
+                      frLayerNum,
+                      frConstraint*,
+                      frBlockObject*,
+                      frBlockObject*>,
```
maliberty, line 847: A struct would be better than a tuple.  I can't guess what  this holds and names would make it clearer.

src/drt/src/TritonRoute.cpp
```diff
@@ -809,16 +809,77 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+      mapMarkers_;
+  omp_set_num_threads(MAX_THREADS);
+#pragma omp parallel for schedule(dynamic)
+  for (int i = 0; i < workers.size(); i++) {
+    workers[i]->init(design_.get());
+    workers[i]->main();
+#pragma omp critical
+    {
+      for (auto& marker : workers[i]->getMarkers()) {
+        Rect bbox = marker->getBBox();
```
maliberty, line 858: The markers might be inside a worker but outside the area being checked ('box').  Do you want to filter them?

Surely there is a better name than box

maliberty (changes requested):

src/drt/src/TritonRoute.cpp
osamahammad21: They are filtered out by reportDRC

maliberty (approved):

PR #2616, opened 2022-12-14
grt: Correctly round up when subtracting obstructions

antonblanchard:


maliberty (approved):

maliberty:
Btw, it is handy to put "Fixes #<issue_number>" in the commit (assuming the issue and the commit are in the same repo).  Then when the issue is merged to master the issue is automatically closed.

antonblanchard:
Quite a lot of test case failures which I'll look into tomorrow morning.

antonblanchard:
Quite a few tests had to be updated but this is passing the CI now.

src/grt/test/gcd.ok
```diff
@@ -64,8 +64,10 @@ metal8            4512             0            0.00%             0 /  0 /  0
...
 ---------------------------------------------------------------------------------------
-Total           149816          3343            2.23%             0 /  0 /  0
+Total           148811          3343            2.25%             0 /  0 /  0
 
 [INFO GRT-0018] Total wirelength: 10258 um
 [INFO GRT-0014] Routed nets: 563
-No differences found.
+Differences found at line 914.
+105000 67200 113400 71400 metal1
+105000 67200 109200 71400 metal1
```
maliberty, line 73: There should be no differences in the .ok files

maliberty (changes requested):

src/grt/test/gcd.ok
antonblanchard: Ouch, let me fix that

antonblanchard:
I think all the test case failures are fixed now

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -533,7 +533,8 @@ void FastRouteCore::initBlockedIntervals(std::vector<int>& track_space)
     if (edge_cap > 0) {
       int reduce = 0;
       for (auto interval_it : intervals) {
-        reduce += ceil(std::abs(interval_it.upper() - interval_it.lower())
+        reduce += ceil(static_cast<float>(
+                           std::abs(interval_it.upper() - interval_it.lower()))
                        / track_space[layer - 1]);
```
maliberty, line 538: @luis201420 why is "track_space[layer - 1]" the divisor and not "track_space[layer]"?  It looks like we are using the wrong layer.

maliberty: Never mind, I got tripped up by dbTechLayer::getRoutingLevel() being 1-based which is where the -1 comes from.

src/grt/test/ndr_1w_3s.ok
```diff
@@ -31,31 +31,30 @@
...
-met1              5439           122            2.24%             1 /  0 /  2
+met1              5388           120            2.23%             0 /  0 /  0
 met2              6475            57            0.88%             0 /  0 /  0
-met3              6142            42            0.68%             0 /  0 /  0
+met3              6142            44            0.72%             0 /  0 /  0
 met4              4625            12            0.26%             0 /  0 /  0
 met5              1480             0            0.00%             0 /  0 /  0
 ---------------------------------------------------------------------------------------
-Total            24161           233            0.96%             1 /  0 /  2
+Total            24110           233            0.97%             0 /  0 /  0
```
maliberty, line 57: nice

maliberty (approved):

QuantamHD:
@maliberty should we merge?

maliberty:
Re-running secure CI shows that sky130hd/aes is still failing.  Anton is taking a look.

antonblanchard:
Here is a test case from the sky130hd aes fail: [grt-diverge-2.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10297669/grt-diverge-2.tar.gz)

With the patch in this PR applied we fail with:

```
[ERROR GPL-0307] RePlAce divergence detected. Re-run with a smaller max_phi_cof value.
```

Working with Matt, it seems that gpl routability is continually overestimating the congestion.

antonblanchard:
As an experiment, I disabled divergence detection in gpl and ran with routability off and on:

routability disabled:
grt-no-routability

routability enabled:
grt-routability

So we might be overestimating the routing congestion, but it looks we ended up with a reasonable solution. I wonder if we should make the divergence detection less sensitive.

We might be circling around at least three issues:

- gpl routability mode is overestimating congestion
- The sky130hd PDN is too dense, especially on met5
- divergence detection is too sensitive and errors out even when we would have made it to a reasonable solution

antonblanchard:
This is now passing sky130hd aes as of: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/a947b24256d894aa12cb61f8e5267821c695c711

Likely it's just luck, but it would be great if we could work the other issues in fresh PRs.

maliberty:
I tried with your latest version and I see in uW: [ERROR] finish__timing__setup__ws fail test: -3.68 >= -1.32
along with some more minor max cap/slew changes.  Is this ok with you?

maliberty:
@vvbandeira I see some secure CI failures as well but they seem infrastructure related.  Would you take over helping Anton get this PR through next week.

antonblanchard:
> I tried with your latest version and I see in uW: [ERROR] finish__timing__setup__ws fail test: -3.68 >= -1.32 along with some more minor max cap/slew changes. Is this ok with you?

Ok with me. met4 and met5 have a lot of signal routing on the microwatt design and this fix in tandem with the very dense PDN is going to significantly reduce the routing resources available. I'm hoping once we agree on PDN settings that things will improve again.

antonblanchard:
I had to rebase due to conflicts with a recent merge. No other changes.

PR #2619, opened 2022-12-14
gui: README: add example of how to read files

xobs:
The GUI is very powerful, and is shown off a lot. However, the documentation of the GUI itself never actually spells out how to open files, nor specifies what sorts of files it reads. This is compounded by the lack of an "Open" dialogue box.

Add a note to the top of the GUI documentation to explicitly state how to open files.

### DIFF
--- a/src/gui/README.md
+++ b/src/gui/README.md
@@ -2,6 +2,18 @@
 
 The graphical user interface can be access by launching OpenROAD with ``-gui`` or by opening it from the command-line with ``gui::show``.
 
+## Reading files
+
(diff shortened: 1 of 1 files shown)


gadfort:
This seems to me like it belongs in OpenROAD or ODB documentation. If an Open file dialog box would be useful to interface for read_db and read_def I think that would be relatively easy to implement.

xobs:
Thanks! The addition of a GUI entry somewhat obviates the need for this, and makes the GUI much more approachable to newcomers. I still think there should be explicit documentation about the fact that `read_db` is used to open the DB files, however I agree that it's less urgent to get this documentation in since there is now a more easily-discoverable method.

PR #2634, opened 2022-12-15
Add `or-tools` to cmake and remove hard coded paths

joennlae:
This fixes the non-docker build. It includes `or-tools` as described by the devs here: https://github.com/or-tools/cmake_or-tools/tree/main/LocalInstall

I tried to get the `local` build running due to not having access to docker (not my decision), and it always crashed because of the hard-coded paths in `cmake`.

Here:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/df19865e51b0c192caf4e8ee7ecf7b88ee593194/src/mpl2/CMakeLists.txt#L37-L40

And here:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/df19865e51b0c192caf4e8ee7ecf7b88ee593194/src/CMakeLists.txt#L39-L42

This works nicely in a docker env, and I tried to bypass it by setting `CMAKE_INSTALL_RPATH` and `CMAKE_INSTALL_PREFIX`, but I still had some linking issues.

I know my proposed solution will lead to longer build times, but I think it will mitigate issues in the future. 

I tagged the latest release here:

https://github.com/joennlae/OpenROAD/blob/7c3c5cdef94cb29434b418999434fc5f62db2e4c/src/cmake/ortools.CMakeLists.txt#L8

Tell me what you think about this solution.

It could maybe also fix the macOS problem:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/df19865e51b0c192caf4e8ee7ecf7b88ee593194/src/CMakeLists.txt#L328-L331
but I did not run any tests.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -122,18 +122,6 @@ _installCommonDev() {
     rm -rf "${baseDir}"
 }
 
-_installOrTools() {
-    os=$1
-    version=$2
-    arch=$3
-    orToolsVersionBig=9.4
-    orToolsVersionSmall=${orToolsVersionBig}.1874
-    orToolsFile=or-tools_${arch}_${os}-${version}_cpp_v${orToolsVersionSmall}.tar.gz
-    wget https://github.com/google/or-tools/releases/download/v${orToolsVersionBig}/${orToolsFile}
-    mkdir -p /opt/or-tools
-    tar --strip 1 --dir /opt/or-tools -xf ${orToolsFile}
-}
-
 _installUbuntuCleanUp() {
     apt-get autoclean -y
     apt-get autoremove -y
@@ -481,7 +469,6 @@ case "${os}" in
             _installCentosDev
             _installCommonDev
         fi
-        _installOrTools "centos" "7" "amd64"
         _installCentosCleanUp
         cat <<EOF
 To enable GCC-8 or Clang-7 you need to run:
@@ -498,7 +485,6 @@ EOF
             _installUbuntuDev
             _installCommonDev
         fi
-        _installOrTools "ubuntu" "${version}" "amd64"
         _installUbuntuCleanUp
         ;;
     "Red Hat Enterprise Linux")
@@ -509,12 +495,10 @@ EOF
             _installRHELDev
             _installCommonDev
         fi
-        _installOrTools "centos" "8" "amd64"
         _installRHELCleanUp
         ;;
     "Darwin" )
         _installDarwin
-        _installOrTools "MacOsX" "12.5" $(uname -m)
         cat <<EOF
 
 To install or run openroad, update your path with:
@@ -532,7 +516,6 @@ EOF
             _installDebianDev
             _installCommonDev
         fi
-        _installOrTools "debian" "${version}" "amd64"
         _installDebianCleanUp
         ;;
     *)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -41,6 +41,33 @@ list(APPEND CMAKE_INSTALL_RPATH
   "/opt/or-tools/lib"
 )
 
+################
+##  OR_TOOLS  ##
+################
+message(STATUS "Installing OR-Tools...")
+# Download and install or-tools at configure time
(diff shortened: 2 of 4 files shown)


maliberty:
CMAKE_INSTALL_RPATH just sets a default and isn't a requirement.  It avoids the need to set LD_LIBRARY_PATH if you choose to use that path but shouldn't cause a crash or any error.

Would you explain further your crash?  I use a non-docker local build daily without issue.  

I'm not keen on the increased build time this solution represents.  We also have seen that such solutions lead to random build failures when the remote site is unavailable/unreachable for any reason.

joennlae:
Yes, so I try to build an independent toolchain. Also, building other eda tools for our environment here. I tried to make it work, but I always get a linker error when not using the defined `/opt/or-tools/`. And I can not use that folder. The best-case scenario for me would be not to patch it (openroad). That is why I proposed this solution.

I obviously understand that introducing this build dependency is not the best case, especially for the devs. Yeah, so maybe I have to figure out another solution.

joennlae:
I also tried with LD_LIBRARY_PATH and CMAKE_INSTALL_RPATH, but somehow it doesn't help.

maliberty:
What is the link error?

joennlae:
LD_error:
```
mpl2/mpl2.a(bus_synthesis.cpp.o): In function `mpl2::calNetPaths(std::vector<mpl2::SoftMacro, std::allocator<mpl2::SoftMacro> >&, std::vector<int, std::allocator<int> >&, std::vector<mpl2::Edge, std::allocator<mpl2::Edge> >&, std::vector<mpl2::Vertex, std::allocator<mpl2::Vertex> >&, std::vector<mpl2::BundledNet, std::allocator<mpl2::BundledNet> >&, float, utl::Logger*)':                                                                                                                                                                                                                                                                                                                                    
bus_synthesis.cpp:(.text+0x97af): undefined reference to `operations_research::MPSolver::CreateSolver(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
bus_synthesis.cpp:(.text+0x98b3): undefined reference to `operations_research::MPSolver::MakeIntVar(double, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
bus_synthesis.cpp:(.text+0x99fe): undefined reference to `operations_research::MPSolver::MakeIntVar(double, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
bus_synthesis.cpp:(.text+0xa0b4): undefined reference to `operations_research::MPSolver::MakeRowConstraint(double, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'                                                                                                                                                                                                                                                                                                             
bus_synthesis.cpp:(.text+0xa18c): undefined reference to `operations_research::MPSolver::MakeRowConstraint(double, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
bus_synthesis.cpp:(.text+0xa3c0): undefined reference to `operations_research::MPSolver::MakeRowConstraint(double, double, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
collect2: error: ld returned 1 exit status   
```

joennlae:
What compiler are you using? I am using `gcc-11.2.0`

maliberty:
Does closing this mean you have resolved it?

joennlae:
@maliberty It is resolved now. 
So I think it is a local issue. Because when I build `or-tools` locally and link it the same way when using the downloaded centos7 release, it works. So I believe there must be some ABI compatibility issues between the official centos7 release build image and our local centos7 installation. Our setup is non-standard. So I believe this is an isolated bug. Sorry for the inconvenience.

Issue #2633, opened 2022-12-15
[ERROR GRT-0228] Horizontal edge usage exceeds the maximum allowed


kareefardi:
### Describe the bug

A design in openlane ci, which was previously passing, is failing now.

### Expected Behavior

It might be an error at OpenROAD end. If it isn't, some advice on how to fix such an error would be appreciated 

### OpenROAD Environment

```shell
Git commit: 0720c9223313835d34033c44816213cfd02a9861
kernel: Linux 5.14.0-1054-oem
os: Ubuntu 20.04.3 LTS (Focal Fossa)
cmake version 3.14.0
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
clang version 10.0.0-4ubuntu1

This is a ci run but I was able to reproduce the problem with the version of openroad pasted
```

### OpenLane Environment

```shell
This is a run done by the CI. So assume the latest openlane/pdk
```

### To Reproduce

[issue_reproducible.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10237705/issue_reproducible.tar.gz)

### Relevant log output
```
OpenROAD v2.0-6086-g0720c9223 
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
[INFO]: Setting signal min routing layer to: met1 and clock min routing layer to met1. 
[INFO]: Setting signal max routing layer to: met5 and clock max routing layer to met5. 
-congestion_iterations 50 -verbose
[INFO GRT-0020] Min routing layer: met1
[INFO GRT-0021] Max routing layer: met5
[INFO GRT-0022] Global adjustment: 30%
[INFO GRT-0023] Grid origin: (0, 0)
[INFO GRT-0043] No OR_DEFAULT vias defined.
[INFO GRT-0088] Layer li1     Track-Pitch = 0.4600  line-2-Via Pitch: 0.3400
[INFO GRT-0088] Layer met1    Track-Pitch = 0.3400  line-2-Via Pitch: 0.3400
[INFO GRT-0088] Layer met2    Track-Pitch = 0.4600  line-2-Via Pitch: 0.3500
[INFO GRT-0088] Layer met3    Track-Pitch = 0.6800  line-2-Via Pitch: 0.6150
[INFO GRT-0088] Layer met4    Track-Pitch = 0.9200  line-2-Via Pitch: 1.0400
[...]

PR #2640, opened 2022-12-16
grt: fix usage from underflow after removing loops

arthurjolo:
Signed-off-by: Arthur <arthurjolo@gmail.com>

The method FastCore::removeLoops() was choosing the wrong edges to decrease the usage, this fix changes it to choose the correct edge to decrease the usage.
Before the method was choosing the one with the minimum coordinates, and in some cases the same edge was decreased 2 times in a row, making the usage be updated to wrong value, some times that generated an underflow.
With the change the method chooses every edge on the loop one at a time, updating the usage only once for each edge.

### DIFF
--- a/src/grt/src/fastroute/src/maze.cpp
+++ b/src/grt/src/fastroute/src/maze.cpp
@@ -174,6 +174,30 @@ void FastRouteCore::fixOverlappingEdge(
     routeLShape(
         startpoint, endpoint, blocked_positions, new_route_x, new_route_y);
 
+    // Updates the usage of the altered edge
+    const int edgeCost = nets_[net_id]->getEdgeCost();
+    for (int k = 0; k < treeedge->route.routelen; k++) { // Remove the usages of the old edges
+      if (treeedge->route.gridsX[k] == treeedge->route.gridsX[k + 1]) {
+        if (treeedge->route.gridsY[k] != treeedge->route.gridsY[k + 1]) {
+          const int min_y = std::min(treeedge->route.gridsY[k], treeedge->route.gridsY[k + 1]);
+          v_edges_[min_y][treeedge->route.gridsX[k]].usage -= edgeCost;
+        }
+      } else {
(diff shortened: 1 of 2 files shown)


maliberty:
This is very subtle.  Could you provide an example of what the edge looks like where the current goes wrong?

arthurjolo:
Yes, the following edge for exmple:

error_edge

Edge N1(88, 22) -> N2(88, 25): Loop 1->2->3 
algorithm: k = 1…2

1st iteration: vertical edge (88,22)->(89,22)
        k=1
	gridX[k] = 22, gridY[k] = 88,
	gridX[k+1] = 22, gridY[k+1] = 89
the minimal point is (88,22), and the vertical edge on this point is decreased

2nd iteration: vertical edge (89,22)->(88,22)
k=2
	gridX[k] = 22, gridY[k] = 89,
	gridX[k+1] = 22, gridY[k+1] = 88
the minimal point is (88,22), and the vertical edge on this point is decreased again

So in this case the vertical edge on the point (88,22) is decreased twice when it should be decrease only once, and the vertical edge on the point(89,22) is never decreased when it should be decreased once.

A edge similar to this onde can be found on the following test case at the net with ID:15270 edge: 1
[grt-usage-wraps.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10259135/grt-usage-wraps.tar.gz)

eder-matheus:
@arthurjolo Did you try your update with the testcase from Issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/2633?

arthurjolo:
@maliberty I just didn't understand the abreveations E & N

maliberty:
> @maliberty I just didn't understand the abreveations E & N

East & North

maliberty:
Please get to the root of why there is on a usage of one where there are two edges at 88,22

(I'm also curious how the loop formed in the first place).

arthurjolo:
I will close this PR due to the signoff error and Open a new one

PR #2644, opened 2022-12-17
Added Python wrapper for fin and also regression test

macd:
There were no regression tests for this so I added one tcl and one Python test.


src/fin/test/gcd_fill.py
```diff
@@ -0,0 +1,13 @@
+from openroad import Design, Tech
+import odb
+
+tech = Tech()
+design = Design(tech)
+
+design.evalTclString("read_db gcd_fill.odb; read_sdc gcd_fill.sdc")
```
maliberty, line 7: Why read_db in tcl eval and not through the design object?

Why do we need timing for metal fill?

src/fin/test/gcd_fill.tcl
```diff
@@ -0,0 +1,10 @@
+read_db gcd_fill.odb
+read_sdc gcd_fill.sdc
+
+set_propagated_clock [all_clocks]
```
maliberty, line 4: same - timing not needed

maliberty (commented):
Thanks for adding a test.  It was really needed.

src/fin/test/gcd_fill.tcl
macd: You are right, we don't need the timing. I was just doing copy paste from flow scripts to make a test case. Will remove.

macd: And therefore we  don't need the sdc either.

src/fin/test/gcd_fill.py
macd: So readDb is not yet on the design object, which I guess, it should be. Will do

maliberty (approved):

maliberty (changes requested):
You need to add fin to test/regression so this test is run with the others.

macd:
Ah, yes, of course. Will do.

maliberty (approved):

PR #2646, opened 2022-12-20
Par python

macd:
This adds the Python wrapper and Python regression tests for par. I've seen some differences in tcl regression results on Ubuntu 22.04, but matched the Python output to the relevant tcl output.

### DIFF
--- a/src/par/test/cluster.py
+++ b/src/par/test/cluster.py
@@ -1,0 +1,16 @@
+from openroad import Design, Tech
+import helpers
+import par_aux
+
+tech = Tech()
+tech.readLef("Nangate45/Nangate45.lef")
+tech.readLiberty("Nangate45/Nangate45_typ.lib")
+design = Design(tech)
+
+design.readVerilog("gcd.v")
+design.evalTclString("link_design gcd")
+
+cid = par_aux.cluster_netlist(design, tool="mlpart", coarsening_ratio=0.6)
+clus_file = helpers.make_result_file("cluster.par")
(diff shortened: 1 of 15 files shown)


src/par/test/cluster.py
```diff
@@ -0,0 +1,16 @@
...
+import helpers
+import par_aux
+
+tech = Tech()
+tech.readLef("Nangate45/Nangate45.lef")
+tech.readLiberty("Nangate45/Nangate45_typ.lib")
+design = Design(tech)
+
+design.readVerilog("gcd.v")
+design.evalTclString("link_design gcd")
```
maliberty, line 11: We should add this to the python api as it is a common operation

maliberty (approved):

src/par/test/cluster.py
macd: You mean "link_design" I take it? I think that is in sta, and I have avoided that for that reason. Is there a low overhead way to do that?

maliberty: Yes.  You could put a python API that still does a tcl eval internally but at least it would be more encapsulated in the right API.

macd: No, it;s in dbSta. What is your perspective on wrapping that?

macd: Looking at it in dbSta, it looks easy to pull out and put it on the Design object. Let me do that.

macd: Actually, it's already done, just need to update the regressions.

PR #2651, opened 2022-12-21
grt: fix usage from underflow

arthurjolo:
Fixes: [#2633](https://github.com/The-OpenROAD-Project/OpenROAD/issues/2633)

At the fastroute after calling the Flute algorthm, the function calls fixEmbeddedTrees(). This method changes the embadded trees, but it doesen't updates the usage for the edges that were removed and the edges that were added. This fix updates this usages after changing the trees.

An exemple can be found on the following test case at the net with ID:15270 edge: 1
[grt-usage-wraps.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10278138/grt-usage-wraps.tar.gz)

Before fixing the embadded tree this edge is:

image

after fixing the embadded tree this edge is:

image

But the usage is never added for the new edges, or is decreased for the old edges.


eder-matheus (approved):
The update looks fine, and it also fixes a failure in a private design. But let's wait for @maliberty review before merging it.

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -174,6 +174,34 @@ void FastRouteCore::fixOverlappingEdge(
     routeLShape(
         startpoint, endpoint, blocked_positions, new_route_x, new_route_y);
 
+    // Updates the usage of the altered edge
+    const int edgeCost = nets_[net_id]->getEdgeCost();
+    for (int k = 0; k < treeedge->route.routelen;
+         k++) {  // remove the usages of the old edges
+      if (treeedge->route.gridsX[k] == treeedge->route.gridsX[k + 1]) {
+        if (treeedge->route.gridsY[k] != treeedge->route.gridsY[k + 1]) {
```
eder-matheus, line 182: Why this extra check here? Same for the code below.

arthurjolo: I made this based on another code, but i belive it is not necessary.

maliberty:
Why does the loop appear when "fixing"?  It seems like another bug.  It looks like it is just flipping the L segment.

src/grt/src/fastroute/src/maze.cpp
maliberty: what was the other code you based it on?

arthurjolo: It is based  on the code from the function removeLoops()

arthurjolo:
I belive there is a bug that creates the loop, according to the code it was supose to create the inverse L route, but this loop is generated. I was looking into it today but still didn't find why is this happening.

src/grt/src/fastroute/src/maze.cpp
maliberty: Since loops should never happen in the first place I guess someone was just being extra careful and considering that you might also have zero length edges (which also shouldn't happen).  It isn't harmful so I we can leave it.

arthurjolo: Ok

eder-matheus:
Secure-ci is green, so I'll merge this PR.

PR #2650, opened 2022-12-20
pdn: informative warning on pin not connected to layer

oharboe:
Signed-off-by: Øyvind Harboe <oyvind.harboe@zylin.com>

PR #2652, opened 2022-12-21
pdn: account for vias in macro pins

gadfort:
This should address #2650

### DIFF
--- a/src/pdn/src/grid.cpp
+++ b/src/pdn/src/grid.cpp
@@ -1260,12 +1260,28 @@ ShapeTreeMap InstanceGrid::getInstancePins(odb::dbInst* inst)
     if (net != nullptr) {
       for (auto* mpin : iterm->getMTerm()->getMPins()) {
         for (auto* box : mpin->getGeometry()) {
-          odb::Rect box_rect = box->getBox();
-          transform.apply(box_rect);
-          auto shape
-              = std::make_shared<Shape>(box->getTechLayer(), net, box_rect);
-          shape->setShapeType(Shape::FIXED);
-          pins.push_back(shape);
+          if (box->isVia()) {
+            odb::dbTechVia* tech_via = box->getTechVia();
+            if (tech_via == nullptr) {
+              continue;
+            }
+
+            for (auto* via_box : tech_via->getBoxes()) {
+              odb::Rect box_rect = via_box->getBox();
+              transform.apply(box_rect);
+              auto shape = std::make_shared<Shape>(
+                  via_box->getTechLayer(), net, box_rect);
+              shape->setShapeType(Shape::FIXED);
+              pins.push_back(shape);
+            }
+          } else {
+            odb::Rect box_rect = box->getBox();
+            transform.apply(box_rect);
+            auto shape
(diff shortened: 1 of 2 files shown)


gadfort:
@oharboe FYI, this should solve the issue in PDN, if you have time, please check

oharboe:
> @oharboe FYI, this should solve the issue in PDN, if you have time, please check

What about all the other cases I discovered in my pr, including the openroad gui?

gadfort:
> > @oharboe FYI, this should solve the issue in PDN, if you have time, please check
> 
> What about all the other cases I discovered in my pr, including the openroad gui?

@oharboe I think you should consider opening separate issues for each one you encountered. Feel free to assign me to the GUI related one.

src/pdn/src/grid.cpp
```diff
@@ -1260,12 +1260,28 @@ ShapeTreeMap InstanceGrid::getInstancePins(odb::dbInst* inst)
...
-          transform.apply(box_rect);
-          auto shape
-              = std::make_shared<Shape>(box->getTechLayer(), net, box_rect);
-          shape->setShapeType(Shape::FIXED);
-          pins.push_back(shape);
+          if (box->isVia()) {
+            odb::dbTechVia* tech_via = box->getTechVia();
+            if (tech_via == nullptr) {
+              continue;
+            }
```
maliberty, line 1267: What is the have a VIA from the LEF itself and not the tech?

gadfort: Would that manifest as a block via? I think there only those two, correct?

QuantamHD:
@maliberty @gadfort Should we consider modifying the ODB API to force users to deal with this? It seems like the current API is ripe with "shooting yourself in the foot".

maliberty:
> @maliberty @gadfort Should we consider modifying the ODB API to force users to deal with this? It seems like the current API is ripe with "shooting yourself in the foot".

What change do you suggest?

oharboe:
@gadfort I've created a pull request that you can use to reproduce the problem: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/709

QuantamHD:
> > @maliberty @gadfort Should we consider modifying the ODB API to force users to deal with this? It seems like the current API is ripe with "shooting yourself in the foot".
> 
> What change do you suggest?

Perhaps `auto* box : mpin->getGeometry(); box->getBox()` should be removed in favor of a more general getBoxes function that returned a list of shapes in that box. Which could automatically do the kind of thing Peter did in this PR. 

The other thing that might work is having a function that returns std::variant<dbTechVia*, db::Rect*, ...> that would force users to check on the type.

src/pdn/src/grid.cpp
maliberty: I see that in lefin any via would become a tech via.  So I guess it isn't possible to have a block via in a macro pin.

Yes there are only block & tech vias.

maliberty (approved):

maliberty:
Based on the comments in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/709 it seems this does resolve this issue reported by @oharboe

PR #2656, opened 2022-12-22
Rcx python

macd:
Python wrapper for rcx. A slightly different approach was used here. The ext.h header file makes heavy use of various Options structures. But these are defined inside of the Ext class and SWIG will not wrap them there. We can try to think about duplicating them in global scope by .i file hacking, but then we run afoul of "const char *" declarations.  So the easier path is to just use the api that is defined in ext.i

### DIFF
--- a/include/ord/Design.h
+++ b/include/ord/Design.h
@@ -97,6 +97,10 @@ namespace par {
 class PartitionMgr;
 }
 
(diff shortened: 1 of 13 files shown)


maliberty (approved):
RCX has a terrible C++ API so this seems reasonable for now.

PR #2657, opened 2022-12-22
Multi power domain

ahmadelrouby:
Added the following to support multi power domains in ifp
- Logic Ports Instantiation 
- Create groups + regions for each power domain
- Associate instances with their respective groups except for the top domain
- Add Isolation cells for each net crossing power domains 
- Add inverters to enable and output pins based on isolation_sense and clamp_value in upf
- Specify halos for the domains


msaligane:
@minghungumich would be great if you could check this out.

src/odb/test/data/dummy/mpd_top.upf
```diff
@@ -0,0 +1,108 @@
...
+  -isolation_sense low
+
+set_isolation iso_d_4 \
+  -domain PD_D4 \
+  -update \
+  -isolation_signal d4_iso_control \
+  -isolation_sense high
+
+
+set_isolation iso_d_1 \
```
minghungumich, line 66: Why don't merge the command with above set_isolation iso_d_1

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +221,631 @@ bool set_isolation(utl::Logger* logger,
...
+  // 1. Create a dbGroup associated to it
+  // 2. Create a dbRegion associated to it
+  // 3. map defined path to the power domain
+  if (associate_groups(logger, block, path_to_domain, top_domain) == false) {
+    return false;
+  }
+
+  if (top_domain == nullptr) {
+    // A TOP DOMAIN should always exist
+    logger->warn(utl::ODB, 10022, "No TOP DOMAIN found, aborting");
```
maliberty, line 598: error

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +221,631 @@ bool set_isolation(utl::Logger* logger,
...
+
+    // For now we're only using the first isolation
+    // TODO: determine what needs to be done in case of multiple strategies
+    // for same domain
+    auto isos = domain->getIsolations();
+
+    if (isos.size() < 1)
+      continue;
+
+    odb::dbIsolation* iso = isos[0];
```
maliberty, line 630: Warn if more than one isolation strategy?

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +221,631 @@ bool set_isolation(utl::Logger* logger,
...
+  // location is parent, it should be placed in the parent's power domain
+  // For each domain, determine for each net, if there's an input to another
+  // domain, create two nets that split 'data' into 'data_i' + 'data_o' 'data_i'
+  // goes into the isolation cell from the original cell 'data_o' got outside of
+  // the isolation cell into the end cell then connect isolation signal to the
+  // enable (or reversed depending on iso sense + cell type) place the isolation
+  // cell into either first or second domain?? (Not sure exactly)
+
+  auto network_ = ord::OpenRoad::openRoad()->getDbNetwork();
+  for (auto&& domain : pds) {
```
maliberty, line 618: Too long, extract this into its own function (or more than one)

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +221,631 @@ bool set_isolation(utl::Logger* logger,
...
+          continue;
+
+        // Find ITERMS that belong to instances outside of this power domain
+        std::vector<odb::dbITerm*> external_iterms;
+        for (auto&& connectedIterm : connectedIterms) {
+          auto connectedInst = connectedIterm->getInst();
+
+          if (!connectedInst->getGroup()
+              || connectedInst->getGroup()->getName()
+                     != curr_group->getName()) {
```
maliberty, line 757: Can you compare the groups rather than their names?  Would two different groups ever have the same name?

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +221,631 @@ bool set_isolation(utl::Logger* logger,
...
+        } else if (iso->getLocation() == "parent") {
+          auto ppd = domain->getParent();
+          if (ppd && ppd->getGroup()) {  // if the parent domain is the top
+                                         // domain, don't add to any group
+            ppd->getGroup()->addInst(isolation_inst);
+          }
+        } else {
+          logger->warn(utl::ODB,
+                       15022,
+                       "Isolation %s has location %s, but only self|parent "
```
maliberty, line 781: Isn't there an option to put it in the sink's domain (possibly more than one)?

src/odb/src/upf/upf.tcl
```diff
@@ -195,3 +195,64 @@ proc set_isolation { args } {
 
 
 
+sta::define_cmd_args "use_interface_cell" { \
```
maliberty, line 198: The new commands need documentation.

src/odb/test/upf_test.tcl
```diff
@@ -3,7 +3,8 @@ source "helpers.tcl"
 read_liberty data/sky130hd/sky130_fd_sc_hd__tt_025C_1v80.lib
 read_lef data/sky130hd/sky130_fd_sc_hd.tlef
 read_lef data/sky130hd/sky130_fd_sc_hd_merged.lef
-read_verilog data/mpd_top/mpd_top.v
+# read_verilog data/mpd_top/mpd_top.v
+read_verilog data/mpd_top/1_synth.v
```
maliberty, line 7: mod_top.v is clearer than 1_synth.v (its not obvious what test it belongs to)

src/odb/test/upf_test1.tcl
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+    }
+}
+
+
+initialize_floorplan -utilization 20 \
+                       -aspect_ratio 1 \
+                       -core_space 2 \
+                       -site unithd
```
maliberty, line 54: This test belongs in ifp not odb

maliberty (changes requested):
Is there code to cut the rows or does the already happen automatically when you create the regions?  It would be good to include the floorplan def in your test (once moved)

rovinski:
is synth.v really necessary? Can you get away with a smaller test case instead of a huge 23k line file?

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +221,631 @@ bool set_isolation(utl::Logger* logger,
...
+  pd->setArea(x1, y1, x2, y2);
+
+  return true;
+}
+
+bool findLongestPrefix(
+    utl::Logger* logger,
+    odb::dbBlock* block,
+    std::map<std::string, std::vector<odb::dbPowerDomain*>>& module_to_domain,
+    std::map<std::string, std::vector<odb::dbPowerDomain*>>& path_to_domain,
```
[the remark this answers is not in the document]
ahmadelrouby: which one are you referring to here? because at least one of the maps is being updated.

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +221,631 @@ bool set_isolation(utl::Logger* logger,
...
+{
+  // Create a reverse lookup table
+  // Module Path (Including empty i.e top module) -> Power Domain
+  // If nothing matches then there's no power domain matching
+  std::map<std::string, std::vector<odb::dbPowerDomain*>> module_to_domain;
+  odb::dbModule* top_module = block->getTopModule();
+
+  if (path_to_domain.find(".") != path_to_domain.end()) {
+    module_to_domain[""] = path_to_domain["."];
+  }
```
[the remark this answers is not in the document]
ahmadelrouby: removed

src/odb/src/upf/upf.cpp
ahmadelrouby: Not really. I was working with names to avoid working with pointers

src/odb/src/upf/upf.cpp
ahmadelrouby: there's a fanout option which I also added support for this commit

src/odb/src/upf/upf.tcl
ahmadelrouby: the first one "use_interface_cell" is already part of upf. the area one however is ours.

src/odb/test/data/dummy/mpd_top.upf
Colin-Holehouse: It can be, but this srves to demonstrate the "successive refinement" model of writing UPF, where the command can be used in one place, then using the -update switch the object can later be modified.

src/odb/src/upf/upf.cpp
```diff
@@ -217,4 +222,732 @@ bool set_isolation(utl::Logger* logger,
...
+    case sta::FuncExpr::Operator::op_or:
+      invert_output = !clamp_val;
+      invert_control = !new_enable_sense;
+      break;
+    case sta::FuncExpr::Operator::op_and:
+      invert_output = clamp_val;
+      invert_control = new_enable_sense;
+      break;
+    default:
+      logger->warn(utl::ODB, 921011, "unknown isolation cell function");
```
maliberty, line 353: Please use sequential message numbers (here and elsewhere)

src/odb/src/upf/upf.cpp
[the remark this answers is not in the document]
maliberty: The code has changed but use constant wherever possible.  I think current_path still is an example of an arg that could be const.

src/odb/src/upf/upf.tcl
ahmadelrouby: added

ahmadelrouby:
@rovinski so far this is the test we've been provided with to be able to test the functionality of multi-power domains.

rovinski:
Maybe others disagree, but I would prefer small, targeted unit tests in this repo. Large tests could go in ORFS or elsewhere.

maliberty:
> Maybe others disagree, but I would prefer small, targeted unit tests in this repo. Large tests could go in ORFS or elsewhere.

We have several flow tests in the top level test directory.  I envision this will become a flow test once we have UPF support throughout.  Does that seem reasonable?  If not perhaps you would be willing to craft a smaller test - this is the one Colin provided.

rovinski:
> We have several flow tests in the top level test directory.

Aside from the flow tests

> I envision this will become a flow test once we have UPF support throughout. Does that seem reasonable? If not perhaps you would be willing to craft a smaller test - this is the one Colin provided.

I guess this isn't out of line in terms of size for a flow test. The thing I don't like about large flow tests is that it's really hard to spot if something is wrong because it can easily get lost in the noise. Maybe a large flow test is necessary here because we can't test all the features with a smaller test, but if it _is_ possible to do so, I'd recommend forgoing the large test.

Are there features/conditions exercised by the large test that we don't expect to be exercised by the smaller test?

maliberty:
I believe there are some smaller tests as well.  I don't think it is xor.

maliberty (approved):

Issue #2637, opened 2022-12-15
cts viewer highlight cofusing for mulitple clocks


maliberty:
### Describe the bug

If I open sky130hd/microwatt I see:

![image](https://user-images.githubusercontent.com/761514/207980434-7b14b513-4ecc-4f49-9705-666564a8ec02.png)

It seems both clocks are being highlighted at once.  

### Expected Behavior

I think we should only highlight for the clock in the active tab.

### Environment

```shell
head of master
```

### To Reproduce

as above

gadfort:
@maliberty would you still want to be able to turn off the clock rendering all together? Ie. so you can view the tree without the rendering or do you think that would not be a useful control?

maliberty:
I think that would be useful too

gadfort:
@maliberty when I run the skyhd microwatt I get
```
[INFO][FLOW-xxxx] Using manual macro placement file ./designs/sky130hd/microwatt/microwatt.macro_placement.cfg
Error: macro_place.tcl, 125 Cannot find instance soc0.bram.bram0.ram_0.memory_0
```

maliberty:
It passed in last night's CI.  Something odd is happening in your setup.

gadfort:
> It passed in last night's CI.  Something odd is happening in your setup.

Yes, I had turned off flattening its working again

PR #2658, opened 2022-12-22
gui: change CTS viewer to default to only showing the currently active clock

gadfort:
https://github.com/The-OpenROAD-Project/OpenROAD/issues/2637


maliberty (approved):
Thanks!

PR #2661, opened 2022-12-23
Python wrapper and regression tests for rmp

macd:
rmp makes significant use of sta, so there was some things to address that. The constructor for Blif needs the pointer to a sta object, so getSta() was added to the Tech object (seemed better here than on the Design Object).  For the time being, this is an opaque pointer that is only passed to the Blif constructor. Also, there were some graph consistency checking / repair that needs to happen before writeBlif. Moved them from C++ code in an interface file into the body of the C++ method writeBlif (where they belong). Finally, since we needed sta::get_lib_pins (which gets pins from a Liberty library) and there was no obvious way to use odb for that, had to do some moderately ugly Python string Template hacking in order to use evalTclString

### DIFF
--- a/include/ord/Design.h
+++ b/include/ord/Design.h
@@ -101,6 +101,10 @@ namespace rcx {
 class Ext;
 }
 
+namespace rmp {
+class Restructure;
+}
+
 namespace ord {
 
 class Tech;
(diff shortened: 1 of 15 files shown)


maliberty (approved):
lohitemp is pretty ugly but as this module isn't even in use yet we can leave it for now.  rsz will have a similar issue though

PR #2663, opened 2022-12-23
grt: Route non leaf clock nets first

antonblanchard:
initNetlist() inserts non leaf clock nets at the start of the list, but undoes this by sorting the list.


eder-matheus:
@antonblanchard The updates look fine, but I think we could run it in secure-ci before merging. Can you start a run? If you don't have access, I can start it for you.

maliberty:
@eder-matheus he doesn't have access

maliberty (approved):
I ran the secure CI and everything passes but gf12/bp_single which picks up 6 detailedroute__route__drc_errors.  That design is prone to issues so I'll accept this change but also ask @osamahammad21 to look at the errors.

PR #2672, opened 2022-12-28
gui: decouple script widget and tcl input widget

gadfort:
Changes:
- Removes all command execution from script widget and moves them to tcl command widget

Adds:
- Adds base class to help with eventual python input

### DIFF
--- a/src/gui/src/cmdInputWidget.h
+++ b/src/gui/src/cmdInputWidget.h
@@ -79,8 +79,11 @@ class CmdInputWidget : public QPlainTextEdit
   void dragEnterEvent(QDragEnterEvent* event) override;
   void dropEvent(QDropEvent* event) override;
   bool handleHistoryKeyPress(QKeyEvent* event);
+  bool handleEnterKeyPress(QKeyEvent* event);
 
   void addCommandToHistory(const QString& command);
+
+  virtual bool isCommandComplete(const std::string& cmd) const = 0;
 
  private:
   // back in history
--- a/src/gui/src/cmdInputWidget.h
+++ b/src/gui/src/cmdInputWidget.h
@@ -46,12 +46,12 @@ class CmdInputWidget : public QPlainTextEdit
   CmdInputWidget(const QString& interp_name, QWidget* parent = nullptr);
   virtual ~CmdInputWidget();
 
-  void setFont(const QFont& font);
+  void setWidgetFont(const QFont& font);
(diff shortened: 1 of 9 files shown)


src/gui/src/cmdInputWidget.h
```diff
@@ -0,0 +1,103 @@
...
+ public:
+  CmdInputWidget(const QString& interp_name, QWidget* parent = nullptr);
+  virtual ~CmdInputWidget();
+
+  void setFont(const QFont& font);
+
+  QString text() const;
+  void setText(const QString& text);
+
+  void setMaximumHeight(int height);
```
maliberty, line 54: override keyword - here and other methods as needed

maliberty (approved):

src/gui/src/cmdInputWidget.h
gadfort: when I add the override it stops compiling with the error : `error: ‘void gui::CmdInputWidget::setMaximumHeight(int)’ marked ‘override’, but does not override`
any idea?

maliberty: It isn't a virtual method.  Are you intending to scope it rather than override it?  It usually isn't a good idea to do that.

gadfort: Not 100% what you mean by scope it or override it.

maliberty:
```
#include <cstdio>

class B
{
public:
    void foo() { printf("B::foo()\n"); }
    virtual void bar() { printf("B::bar()\n"); }
};

class D : public B
{
public:
    void foo() { printf("D::foo()\n"); }
    virtual void bar() override { printf("D::bar()\n"); }
};

int main()
{
    D* d = new D;
    B* b = d;

    b->foo();
    b->bar();

    d->foo();
    d->bar();
    return 0;
}
```

when run produces
```
B::foo()
D::bar()
D::foo()
D::bar()
```

bar is overriden but foo is scoped (it hides the base class definition from the derived class)

gadfort: Ahhh, thanks for clarifying. This was a copy and paste of the code already in use, but I see the issue, I'll take a look and see what can be done for these.

gadfort: renamed functions and their calls to correct this

PR #2682, opened 2022-12-31
metrics: add group_count to setup/hold violation metrics to return more than 1 or 0 for a more accurate result

gadfort:
I noticed that the metrics for hold and setup violation count is basically binary since `find_timing_paths` returns just one by default. By adding `group_count` the result will include more than 1 and will default to `sta::endpoint_count`.

### DIFF
--- a/src/Metrics.tcl
+++ b/src/Metrics.tcl
@@ -87,7 +87,7 @@ proc report_worst_slack_metric { args } {
 }
 
 define_cmd_args "report_erc_metrics" {}
-proc report_erc_metrics { args } {
+proc report_erc_metrics { } {
 
     set max_slew_limit [sta::max_slew_check_slack_limit]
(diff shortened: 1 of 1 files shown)


antonblanchard:
We've been discussing this over on https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/591#discussion_r1053859746. OpenSTA now has `sta::endpoint_violation_count` and we just need OpenROAD to update OpenSTA.

gadfort:
@antonblanchard that's way better :-) I'll fix the metric here to use that command

maliberty:
@vvbandeira this will require updating all the metrics when it goes through.  Please help on this one.

gadfort:
@maliberty I rebased this PR based on the updates to STA so this PR does not need to change it. This makes the code changes pretty small.

vvbandeira (approved):

PR #2686, opened 2023-01-01
dbSta: enable string redirection in openroad to function with with_output_to_variable

gadfort:
Fixes:
- `with_output_to_variable` does not work in OpenROAD like it does in OpenSTA because the functions that enable this behavior is overridden for the logger. This enables this redirection to have the same behavior as OpenSTA.

### DIFF
--- a/src/dbSta/src/dbSta.cc
+++ b/src/dbSta/src/dbSta.cc
@@ -428,6 +428,12 @@ dbStaReport::setLogger(Logger* logger)
 void
 dbStaReport::printLine(const char* buffer, size_t length)
 {
+  if (redirect_to_string_) {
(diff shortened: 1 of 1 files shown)


maliberty:
What command would show the effect of this change?

gadfort:
```
with_output_to_variable test {check_setup}
```

In this case, instead of printing the output the text is stored in $test

A quick example:
```
with_output_to_variable test {puts "1234"}
puts "Text in variable: $test"
```
Should print: 
```
Text in variable: 1234
```

Currently the redirect doesn't work, so it outputs:
```
1234
Text in variable: 
```

maliberty (approved):

PR #2694, opened 2023-01-03
cts: fix a warning

oharboe:
we're iterating over floats, but comparing doubles, so we do want to copy the std::pair<float,float> into std::pair<double,double>

### DIFF
--- a/src/cts/src/Clock.cpp
+++ b/src/cts/src/Clock.cpp
@@ -90,16 +90,16 @@ Box<int> Clock::computeSinkRegion()
 Box<double> Clock::computeSinkRegionClustered(
     std::vector<std::pair<float, float>> sinks)
 {
-  double xMin = std::numeric_limits<double>::max();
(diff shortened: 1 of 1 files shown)


src/cts/src/Clock.cpp
```diff
@@ -95,7 +95,7 @@ Box<double> Clock::computeSinkRegionClustered(
   double yMin = std::numeric_limits<double>::max();
   double yMax = std::numeric_limits<double>::lowest();
 
-  for (const std::pair<double, double>& sinkLocation : sinks) {
+  for (const std::pair<double, double> sinkLocation : sinks) {
```
gadfort, line 98: could this be `const auto& [sink_x, sink_y]` instead of a `std::pair`?

rovinski: I think you may want just `const auto [sink_x, sink_y]` to avoid references to primitive types?

maliberty (approved):

Issue #2386, opened 2022-10-19
FMTLib OpenROAD Compile


QuantamHD:
**Describe the bug**
New versions of FMTLib no longer allow formatting of user defined types. 

**Expected Behavior**
OpenROAD should compile with newer versions of fmtlib. We should add a CI check to make sure that OpenROAD compiles with new versions of fmtlib so that we don't introduce more cases like this.

**Environment**

Debian Testing

maliberty:
I think fmt has been horrible broken by their new changes.  See https://github.com/fmtlib/fmt/issues/3088 (and https://github.com/fmtlib/fmt/issues/3081).

ekliptik:
I got it to build with new versions through the provided workaround. I added `add_compile_definitions(FMT_DEPRECATED_OSTREAM)` to CMakeLists.txt of fmtlib, built and `make install`ed it, then I also built spdgen and OpenROAD from source.

maliberty:
Is it necessary to recompile fmtlib in order to use FMT_DEPRECATED_OSTREAM?  Can you just add it to OR?

ekliptik:
Yes, it is necessary. It's behind an `#ifdef` directive, so it needs to be set at fmtlib compilation. It only applies to new versions of fmtlib, which CentOS and Debian might not have yet, I don't know how packaging there really works. A patched fmtlib fork could be maintained alongside OpenROAD, used as a submodule, and CMake somehow instructed to ignore local fmtlib instances and always use the forked one to build OpenROAD. Or we can ask the maintainers to expose the flag in their CMakeLists so an unpatched version can be built with the flag passed to CMake.

fgaray:
Confirmed that this was fixed by #4351, I can compile in debian testing but keep in mind that currently it is shipping with fmt [9.1.0](https://packages.debian.org/search?keywords=libfmt&searchon=names&suite=bookworm&section=all) so you may want to manually compile version >=10 or wait until it is available in the repos.

I think you may want to close this @maliberty or @QuantamHD

PR #2696, opened 2023-01-03
add missing ostream formatters

kammoh:
Fixes compile on [at least] the following platforms:
- Fedora 37 (spdlog:1.10.0, fmt:9.1.0)
- macOS Ventura (homebrew spdlog:1.11.0 fmt:9.1.0)

https://fmt.dev/latest/api.html#format-api
https://fmt.dev/latest/api.html#ostream-api

### DIFF
--- a/src/cts/src/Util.h
+++ b/src/cts/src/Util.h
@@ -36,6 +36,7 @@
 #pragma once
 
 #include <ostream>
+#include <fmt/ostream.h>
 
 namespace cts {
 
@@ -140,3 +141,5 @@ class Box
 };
 
 }  // namespace cts
+
+template <typename T> struct fmt::formatter<cts::Box<T>> : fmt::ostream_formatter {};
--- a/src/cts/src/Util.h
+++ b/src/cts/src/Util.h
@@ -36,7 +36,7 @@
 #pragma once
 
 #include <ostream>
-#include <fmt/ostream.h>
+#include <spdlog/fmt/fmt.h>
 
 namespace cts {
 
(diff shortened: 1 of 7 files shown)


QuantamHD:
You need to sign off your git commits with the -s flag. 

https://github.com/The-OpenROAD-Project/OpenROAD/pull/2696/checks?check_run_id=10416684588

vvbandeira:
Looks like this change is not compatible with the current version we support (and use on our CI):
```
In file included from /OpenROAD/src/dst/src/LoadBalancer.cc:29:
/OpenROAD/src/dst/src/LoadBalancer.h:36:10: fatal error: fmt/ostream.h: No such file or directory
   36 | #include <fmt/ostream.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

@kammoh  After signing your commit and passing DCO, I would suggest adding some conditional for the proposed fix. Maybe an `ifdef` depending on the version of `fmt` as to not break all other installs.

maliberty:
I am opposed to the changes in fmt 9 and don't want to accept these changes (see https://github.com/fmtlib/fmt/issues/3088).  This has been discussed in https://github.com/The-OpenROAD-Project/OpenROAD/issues/2386.  As I said there, we should use FMT_DEPRECATED_OSTREAM instead.

maliberty:
This change is not sufficient to handle all the missing formatting fmt9 would require (which would be very extensive).

maliberty:
I believe FMT_DEPRECATED_OSTREAM  is a better solution.  Is there a reason we can't use it?

maliberty:
> I believe FMT_DEPRECATED_OSTREAM is a better solution. Is there a reason we can't use it?

any response?

maliberty:
stale

PR #2697, opened 2023-01-04
gui: add option to select just one path per endpoint in staGui

gadfort:
Adds to gui sta interface:
- Option to just get one path per endpoint
- Total number of endpoints


gadfort:
@maliberty this should help with #2697 proving the interface they will need

maliberty:
It looks fine but why not make it an integer?

maliberty (approved):

gadfort:
> It looks fine but why not make it an integer? 

I thought about it, but decided to keep the interface a simplified. We could always change it later if desired.

PR #2698, opened 2023-01-04
drt: Reduce searchRepair batch size from 4 to 2

antonblanchard:
drt searchRepair creates 4 batches which means we have to wait on the slowest worker 4 times per searchRepair iteration.

Reduce the number of batches to 2, with each batch containing regions that have no neighbouring regions on any of their edges.

This improves performance of sky130hd aes detailed routing by 39%.

### DIFF
--- a/src/drt/src/dr/FlexDR.cpp
+++ b/src/drt/src/dr/FlexDR.cpp
@@ -1506,12 +1506,6 @@ FlexDR::initDR_mergeBoundaryPin(int startX,
   return bp;
 }
 
-void FlexDR::getBatchInfo(int& batchStepX, int& batchStepY)
-{
-  batchStepX = 2;
-  batchStepY = 2;
-}
-
 void FlexDR::searchRepair(const SearchRepairArgs& args)
 {
   const int iter = iter_++;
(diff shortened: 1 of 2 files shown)


antonblanchard:
I tried removing the batching entirely (thinking it was used to reduce memory consumption), but that resulted in drt failures. I'm therefore guessing that batching is used here to avoid operating on wires that go to neighbouring cells.

This needs help from someone who understands drt, but the gains are promising.

antonblanchard:
A couple of SEGVs in the CI (eg sky130hd aes) which I can't reproduce.

vvbandeira:
@osamahammad21 can you take a look and see if you can reproduce the segv?

antonblanchard:
Thanks @osamahammad21. The TritonRoute paper talks about standard, DRC and extended bounding boxes, and each worker is only allowed to modify wires with center lines within the non overlapping standard boxes. Also, with Manhattan routing no wire can directly cross diagonally, it will have to intersect one of the edges first.

As such, I'm not sure why it should SEGV. Taking it further, we should be able to remove all batching and still not SEGV. I can see how we'd have some false positives/negatives with DRC checking, since the DRC and extended boxes may change, but that is a different issue.

rovinski:
I think there might also be the case where one worker reads a wire while another worker is modifying it, which I could also see leading to a SEGV. I don't think that any of the data structures have locks - it relies on joining the threads to synchronize.

antonblanchard:
> I think there might also be the case where one worker reads a wire while another worker is modifying it, which I could also see leading to a SEGV. I don't think that any of the data structures have locks - it relies on joining the threads to synchronize.

I ran drt using the clang thread sanitizer and it found a couple of small issues, but nothing serious:  https://github.com/The-OpenROAD-Project/OpenROAD/pull/2718

osamahammad21:
I am not sure exactly where the segv could happen but I can suspect that it happens in the connectivity checker.
I know that we don't have diagonal routes, but there are other ways a segment could be shared between two workers in the same batch. For example, a segment could be in the bottom border of the top worker and on the top border of the lower worker. Since currently the router follows the shared border concept, this could happen and lead to false routes. And any false routes blow up the connectivity checker.

QuantamHD:
@antonblanchard You could try building with asan which would show up as an error in the CI.

maliberty:
This PR won't get merged due to the above discussion.  You can take the segv into a separate issue if desired.

PR #2702, opened 2023-01-04
Psm python

macd:
This was another easy one. Still had to use evalTclString for the read_sdc command, though.

### DIFF
--- a/src/psm/test/pdnsim_aux.py
+++ b/src/psm/test/pdnsim_aux.py
@@ -1,0 +1,153 @@
+############################################################################
+##
+## Copyright (c) 2023, The Regents of the University of Minnesota
+## All rights reserved.
+##
+## BSD 3-Clause License
+##
+## Redistribution and use in source and binary forms, with or without
+## modification, are permitted provided that the following conditions are met:
+##
+## * Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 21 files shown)


src/psm/test/pdnsim_aux.py
```diff
@@ -0,0 +1,153 @@
...
+## INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+## CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+## POSSIBILITY OF SUCH DAMAGE.
+##
+############################################################################
+from os.path import isfile
+
+
+def analyze_power_grid(design, *,
```
maliberty, line 38: What is the * for?

maliberty (approved):

src/psm/test/pdnsim_aux.py
macd: The * makes all the following keyword args keyword only and does not allow any of them to be positional. So iit avoids confusion.

PR #2708, opened 2023-01-05
ant: add violating nets/pins to metrics

gadfort:


maliberty (approved):
@ravi-varadarajan any concerns about naming re: metrics 2.1?

gadfort:
@ravi-varadarajan if the naming is not consistent, let me know and I'll correct them

PR #2705, opened 2023-01-05
Add Ubuntu 22.10 support

nekronos:
This makes it build on Ubuntu 22.10

PR #2710, opened 2023-01-06
depedencyinstaller: Ubuntu 22.10 support

oharboe:


vvbandeira:
I have approved #2705, so I think we can close this PR as a duplicate. Apologies for the delay.

PR #2712, opened 2023-01-06
metrics: avoid double formatting metrics and adding core_area and die_area metrics

gadfort:
Fixes:
- by calling `format_time` and recording floats, the value "inf" appears in the json format, which is not supported.
```
	"sc__metric__timing__setup__ws": inf,
```
- This fixes this to record the value and letting the metrics logger format the information.

Adds:
- Core and die area metrics

### DIFF
--- a/src/Metrics.tcl
+++ b/src/Metrics.tcl
@@ -138,7 +138,7 @@ proc report_units_metric { args } {
 define_cmd_args "report_design_area_metrics" {}
 proc report_design_area_metrics {args} {
   set db [::ord::get_db]
-  set dbu_per_uu [[$db getTech] getDbUnitsPerMicron]
+  set dbu_per_uu [expr double([[$db getTech] getDbUnitsPerMicron])]
   set block [[$db getChip] getBlock]
   set die_bbox [$block getDieArea]
   set die_area [expr [$die_bbox dx] * [$die_bbox dy]]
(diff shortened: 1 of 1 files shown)


maliberty:
What is reported now instead of 'inf'?

gadfort:
It reports whatever OpenSTA has, usually 1e30 (it seems for +Inf, and -1e30 for -Inf).

gadfort:
it might be possible to use std::numeric_limits to check if the value is Inf and record it as a string instead, if that is better, since this doesn't fix @donn issue as far as I can tell.

maliberty (approved):

maliberty:
@vvbandeira FYI

gadfort:
@maliberty I see your PR will fix @donn utilization thing. Thanks.

maliberty:
@gadfort - @ravi-varadarajan noted that removing the call to format_time changed the units of the metrics being reported to seconds instead of the Liberty time units.  You need to either scale the units or revert back and handle inf directly.

gadfort:
I think it's just missing a call to get it into lib units scaling from sta, I really don't want to use format_time because of the Inf issue

ravi-varadarajan:
Hi Peter

What was the Inf issue?  format_time is currently used in log file for
report sta timing slacks.  Does the Inf issue show in the log file too?

-ravi

PR #2714, opened 2023-01-07
Python wrapper for pdn and core set of regression tests

macd:
pdn exhibits a lot of the problems that can happen when the C++ header file, in this case PdnGen.h, is not designed with the objective of wrapping it in Python. The first easy case is that does a `using std::regex`, which SWIG does not, by design, support. So that is handled by `#ifndef SWIG`. The second problem is that the PdnGen.h interface uses many 'private' classes defined in header files in pdn/src directory. While we do not need or want these wrapped, the generated C++ must have these headers included so that the size of the objects can be determined. Also, there was a deep issue with SWIG, where a single line in odb/src/swig/python/dbtypes.i, would cause SWIG to generate incorrect (un-compilable) C++ code. This made any `%template(stuff) std::vector<int>` and `%template(split_map)    std::map<odb::dbTechLayer *, int>;` unusable. Moreover this line in dbtypes.i serves no useful purpose, probably added as a cut and paste, so it has been deleted. This problem seems to be on the tcl side as well, as I have seen multiple places in pdn where vectors are de-structured to send over the interface only to be reconstituted on the C++ side.


src/pdn/include/pdn/PdnGen.hh
```diff
@@ -57,7 +57,9 @@ using odb::dbRegion;
 
 using utl::Logger;
 
+#ifndef SWIG
```
gadfort, line 60: I don't where `std::regex` is used in the header, I think it might be left over from when global connect was here. You can probably just remove it along with the header import

src/pdn/src/PdnGen-py.i
```diff
@@ -0,0 +1,101 @@
...
+namespace std {
+  %template(split_cuts_stuff) std::vector<int>;
+  %template(stuff)        std::array<int, 4>;
+  %template(split_map)    std::map<odb::dbTechLayer *, int>;  
+  %template(grid_list)    std::vector<pdn::Grid *>;
+  %template(domain_list)  std::vector<pdn::VoltageDomain *>;
+  %template(net_list)     std::vector<odb::dbNet *>;
+  %template(viagen_list)  std::vector<odb::dbTechViaGenerateRule *>;
+  %template(techvia_list) std::vector<odb::dbTechVia *>;
+  %template(layer_list)   std::vector<odb::dbTechLayer *>;
```
gadfort, line 85: shouldn't these come from the ODB wrapper?

src/pdn/test/pdn_aux.py
```diff
@@ -0,0 +1,676 @@
...
+        net.setSpecial()
+        net.setSigType("GROUND")
+
+    # region = None
+    if region is not None:
+        region = design.getBlock().findRegion(region)
+        if region is None:
+            utl.error(utl.PDN, 1504, f"Region {region} not defined")
+
+    design.getBlock().addGlobalConnect(region, inst_pattern, pin_pattern, net, True)
```
gadfort, line 47: Lets not co-mingle things more. If this lived in OpenROAD land for tcl, I think it should live there as well for python.

src/pdn/test/pdn_aux.py
```diff
@@ -0,0 +1,676 @@
...
+
+
+# Only to change the case? Seems weird.
```
gadfort, line 50: This was a backwards compatibility issue with the old PDNGEN

src/pdn/test/pdn_aux.py
```diff
@@ -0,0 +1,676 @@
...
+    split_cuts_dict = dict(zip(split_cuts_layers, split_cuts_pitches))
+    
+    for g in pdngen.findGrid(grid):
+        pdngen.makeConnect(g, l0, l1, cut_pitch[0], cut_pitch[1],
+                           fixed_generate_vias, fixed_tech_vias,
+                           max_rows, max_columns, ongrid_list,
+                           split_cuts_dict, dont_use_vias)
+
+
+def pdngen_db(design, failed_via_report="", skip_trim=False, dont_add_pins=False):
```
gadfort, line 480: I'm not sure what the thought is on this, but it seems like these functions should either get exposed to the users or we need to go through the tcl and python setups to move a lot of the error checking here moved into C

src/pdn/test/power_switch.py
```diff
@@ -0,0 +1,58 @@
...
+
+# set_voltage_domain -power VDD -ground VSS -switched_power VDD_SW
+# define_power_switch_cell -name POWER_SWITCH -control SLEEP -acknowledge SLEEP_OUT -power_switchable VPWR -power VDDG -ground VGND
+# define_pdn_grid -name "Core" -power_switch_cell POWER_SWITCH -power_control nPWRUP -power_control_network DAISY
+
+# add_pdn_connect -layers {met1 met4} 
+# add_pdn_connect -layers {met2 met4} 
+# add_pdn_connect -layers {met4 met5}
+
+# pdngen -report_only
```
gadfort, line 58: I think you can remove all of this

src/pdn/test/widthtable.py
```diff
@@ -0,0 +1,24 @@
...
+import pdn_aux
+import helpers
+
+tech = Tech()
+tech.readLef("asap7_vias/asap7_tech_1x_noviarules.lef")
+tech.readLef("asap7_vias/asap7sc7p5t_27_R_1x.lef")
+
+design = Design(tech)
+design.readDef("asap7_vias/floorplan.def")
+pdngen = design.getPdnGen()
```
gadfort, line 11: this isn't used.

src/pdn/include/pdn/PdnGen.hh
macd: will do.

src/pdn/src/PdnGen-py.i
macd: You would think so, but using '%import "odb.i"' or '%include "odb.i"' and friends, and dropping all the %templates, does get everything to compile and link, the recognition of these vectors does not work in the generated C++. I found these templates to be necessary.

src/pdn/test/pdn_aux.py
macd: OK

src/pdn/test/pdn_aux.py
macd: pdngen_db is just the Python version of the tcl proc pdngen.  I just added the _db to avoid name clash / shadowing issues with the name pdngen. Moving the error checking into C++ is a good goal, but that is going to have to be done in stages.

src/pdn/test/power_switch.py
macd: So remove power_switch.py...   Will do.

src/pdn/test/widthtable.py
macd: By that do you mean the entirety of widthtable.py or just the pdngen = design.getPdnGen()? I will assume the latter and update.

src/pdn/test/pdn_aux.py
macd: There are a couple of reasons why we might not want that. First is the fact that these ***_aux.py files are not really considered (yet) part of a defined Python API. I think @maliberty would like to have as much of the final API defined in C++ as possible, with the wrappers being very thin. I have mostly (but not always) been able to do that, but I have translated varying parts of the support tcl into Python (in the ***_aux.py files) in order to expediently port the regression tests to Python. The other reason to keep it here, for now, is avoid the "what namespace is that in?" problem.

src/pdn/test/power_switch.py
gadfort: Well, just the parts that are completely commented out. the rest of the file is ok

macd: Ooops. Was a little too quick here.  Will fix.

src/odb/src/swig/python/dbtypes.i
```diff
@@ -217,7 +217,6 @@
 
 // Handle return by ref.
 %apply int &OUTPUT { int & overhang1, int & overhang2 };
-%apply int &OUTPUT { int & x, int & y };
```
maliberty, line 220: Why is this removed?  Are you certain it doesn't affect any db APIs?

src/pdn/test/pdn_aux.py
maliberty: Most modules have and *_aux.py.  I view them all as temporary scaffolding until we refine the c++/python APIs.  My goal is that c++ & python have the same API and simple swig'ing is sufficient.  To make that a reality we'll have to refactor the APIs as well as move error checking into C++.  utl/include/utl/validation.h is a step in that direction.

In terms of process I want Don to get everything working and then I'll come through for API improvements.  He didn't feel familiar enough to refactor all the APIs involved.  @gadfort - you are welcome to handle this for pdn if you wish.

In general the python api will be more programmatic/oo than the command oriented tcl api.

src/pdn/test/pdn_aux.py
maliberty: I think we can drop backward compatibility now as everyone has moved on afaik.  If they haven't they are going to have bigger problems...

maliberty (commented):
@gadfort I'll hold the approval pending any further discussion.  I don't have a lot more code feedback.

src/odb/src/swig/python/dbtypes.i
macd: It was removed, after a very laborious binary search, because it caused SWIG to generate incorrect code when using relatively simple %templates for integer vector recognition in the PdnGen-py.i file. I ran the entire test suite with only this change and without this change and got exactly the same results.

macd: Also, note, that this is only for the Python api, not the tcl api.

src/pdn/test/widthtable.py
gadfort: yes, just the pdngen variable

gadfort (approved):
LGTM

PR #2717, opened 2023-01-07
pdngen: check if net is in domain net list before looking in net_map

gadfort:


maliberty:
Is this ready for review?

gadfort:
@maliberty yes

maliberty (approved):
It looks fine but when does this problem actually arise?

gadfort:
It really shouldn't, but it's possible (possibly by defining overlapping domains, I'll need to track down the fundamental issue, hopefully I can figure it out this week). But during regular use, it doesn't come up.

PR #2718, opened 2023-01-08
Fix some threading data races

antonblanchard:
These were found with the clang thread sanitizer. The concurrent update of getFrNet()->setModified(true) is probably OK, but it really should be protected.


osamahammad21 (approved):

rovinski:
Does this serialization have a performance impact? It might be good to have a secure run to check for that. I wonder if this bug doesn't actually manifest when batch size == 4?

maliberty:
Can you just move it to FlexDRWorker::end which is already run single-threaded?  The only user looks to be the connectivity checker which runs later.

maliberty:
stale

PR #2720, opened 2023-01-09
drt: Move FIXEDSHAPECOST and MARKERDECAY into SearchRepairArgs 

antonblanchard:
FIXEDSHAPECOST and MARKERDECAY change based on the iteration, so move
them into SearchRepairArgs where all other per iteration variables are
set.

### DIFF
--- a/src/drt/include/triton_route/TritonRoute.h
+++ b/src/drt/include/triton_route/TritonRoute.h
@@ -136,6 +136,8 @@ class TritonRoute
   void setDebugWorkerParams(int mazeEndIter,
                             int drcCost,
                             int markerCost,
+                            int fixedShapeCost,
+                            float markerDecay,
                             int ripupMode,
                             int followGuide);
   void setDistributed(bool on = true);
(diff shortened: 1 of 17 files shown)


antonblanchard:
I noticed this when scanning through drt. It seems preferable to me that all iteration dependent settings are in the one place.

antonblanchard:
SearchRepairArgs is pretty horribly formatted by clang-format

maliberty:
> SearchRepairArgs is pretty horribly formatted by clang-format

I agree.  Perhaps the /* 8 */ could be moved to the end of line // 8 to save a few characters (or dropped).  Another option is to turn off formatting locally and manually format the lines allowing them to be extra long.

antonblanchard:
> > SearchRepairArgs is pretty horribly formatted by clang-format
> 
> I agree. Perhaps the /* 8 */ could be moved to the end of line // 8 to save a few characters (or dropped). Another option is to turn off formatting locally and manually format the lines allowing them to be extra long.

I did both of these, but also aligned the columns so hopefully it makes it easier to read.

PR #2722, opened 2023-01-09
ppl: add option to mirror pairs of pins

eder-matheus:


src/ppl/src/HungarianMatching.cpp
```diff
@@ -119,10 +124,33 @@ void HungarianMatching::getFinalAssignment(std::vector<IOPin>& assigment) const
...
+            || io_pin.isPlaced()) {
+          continue;
+        }
         io_pin.setPos(slots_[slot_index].pos);
         io_pin.setLayer(slots_[slot_index].layer);
+        io_pin.setPlaced();
         assigment.push_back(io_pin);
         slots_[slot_index].used = true;
+
+        if (assign_mirrored) {
```
maliberty, line 141: What happens if the mirrored slot is already in use?

src/ppl/src/IOPlacer.tcl
```diff
@@ -115,79 +115,93 @@ proc define_pin_shape_pattern { args } {
...
-  }
-
   set dbTech [ord::get_db_tech]
   set dbBlock [ord::get_db_block]
   set lef_units [$dbTech getLefUnits]
 
-  if [regexp -all {(top|bottom|left|right):(.+)} $region - edge interval] {
-    set edge_ [ppl::parse_edge "-region" $edge]
+  if [info exists keys(-region)] {
+    set region $keys(-region)
```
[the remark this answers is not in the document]
eder-matheus: `set_io_pin_constraint` was failing if a `-region` was not defined. So I update it to evaluate the `-region` option only when it exists, allowing the user to just set a list of mirrored pins.

src/ppl/src/HungarianMatching.cpp
eder-matheus: I thought this scenario could be avoided by first assigning the mirrored pins and then the remaining pins. This way, all the slots are free when assigning the mirrored slots.

But I just realized that when we have pin groups, we can have used slots when assigning the mirrored pins. We are currently given priority for groups over the remaining pins. I think prioritizing the mirrored pins over the groups should work, but I wonder how often we'll have pin groups and mirrored pins together.

maliberty: @oharboe do you plan to have mirrored busses?

maliberty (approved):
We can take up groups & mirroring into another PR

src/ppl/src/HungarianMatching.cpp
oharboe: I dont know what a mirrored bus is.

maliberty: Will you have a group of pins that you also want to mirror?

oharboe: I am trying to do "routing by abutment", i.e. that all the routing happens inside the macros and that I can just place the macros next to eachother with no spacing and have them connect correctly. So all the pins on one edge.must match exactly all the pins on the neighbouring edge.

oharboe: I have no opinion on the order of the pins, but it is probably worth ordering them. Easier to view it in the GUI. Faster to debug.

I am doing design exploration, not tapeout, so I am more concerned about implementation effort and the pedagogic element than sqeueezing out the last drops of PPA.

PR #2726, opened 2023-01-10
ppl: fix coverity issue

eder-matheus:


src/ppl/src/HungarianMatching.cpp
```diff
@@ -266,11 +280,12 @@ void HungarianMatching::getAssignmentForGroups(std::vector<IOPin>& assigment)
...
 
-int HungarianMatching::getSlotIdxByPosition(const odb::Point& position) const
+int HungarianMatching::getSlotIdxByPosition(const odb::Point& position,
+                                            int layer) const
 {
   int slot_idx = -1;
   for (int i = 0; i < slots_.size(); i++) {
-    if (slots_[i].pos == position) {
+    if (slots_[i].pos == position && slots_[i].layer == layer) {
       slot_idx = i;
```
maliberty, line 289: Just curious why we need to check the layer now and not previously?

maliberty (approved):

src/ppl/src/HungarianMatching.cpp
eder-matheus: Previously it was a bug. In the designs I was testing, this check wasn't necessary because I wasn't using multiple layers in the same direction. But in these cases, checking the layer is required.

PR #2727, opened 2023-01-10
ppl: fix pin counting in mirrored pins to fix infinite loop

oharboe:


eder-matheus (approved):

oharboe:
@yupferris FYI

PR #2730, opened 2023-01-11
Undo change to dbtypes.i and add %clear directive to PdnGen-py.i

macd:
So this got through the regression tests because I completely forgot that the odb Python regressions do not run with the standard OR regression system. We have to fix that.


maliberty:
I'm hoping that's the royal we ("We have to fix that.")

maliberty (approved):

macd:
Yes, that is the "royal" we. I'm so ticked off by this (and myself), I just might do it.

PR #2731, opened 2023-01-11
Pre-empt user from running -local if root

habibayassin:


vvbandeira:
@habibayassin could you also modify the docs to include the limitation of the script? e.g., mention which packages can be installed with `-local` and which will require `sudo ./DependencyInstaller.sh`?

habibayassin:
@vvbandeira please let me know if this is what you have in mind

etc/DependencyInstaller.sh
```diff
@@ -456,15 +456,23 @@ _installDebianPackages() {
...
-Usage: $0 -prefix=DIR
-       $0 -local
+Usage: $0                       Installs all of OpenROAD's dependencies no need to run -base or -common.
+                                                           Requires priviledged access. 
+       $0 -base                 Installs OpenROAD's dependencies using package managers 
+                                                           (-common must be executed in another command)
+       $0 -common               Installs OpenROAD's common dependencies (-base must be executed in another command)
+       $0 -prefix=DIR           Installs dependencies in an existing user specified directory      
+       $0 -local                Installs common dependencies in "$HOME/.local". Only used with -common. Root access 
+                                                           is still needed for other dependencies. Flag cannot be used with sudo or with root access
```
vvbandeira, line 466:
```suggestion
       $0 -prefix=DIR           Installs common dependencies in an existing user-specified directory. Only used with -common. This flag cannot be used with sudo or with root access.
       $0 -local                Installs common dependencies in "$HOME/.local". Only used with -common. This flag cannot be used with sudo or with root access.
```

vvbandeira (changes requested):

etc/DependencyInstaller.sh
```diff
@@ -478,7 +486,17 @@ while [ "$#" -gt 0 ]; do
...
+            option="common"
+            ;;
         -local)
+            if [[ $(id -u) == 0 ]]; then
+                echo "Error: cannot install locally if you are root or using sudo" >&2
+                exit 1
+            fi
             export PREFIX="${HOME}/.local"
             ;;
         -prefix=*)
```
vvbandeira, line 502: Should `-prefix` also check for root/sudo?

habibayassin: I don't think so, the problem with using sudo and local was that the $HOME value changed from ~ to /root but I believe if the user specified a defined prefix from the start that won't be determined in runtime then it shouldn't be a problem. 
For example, running `sudo ./etc/DependencyInstaller.sh -prefix=$HOME/local` the prefix will be set to ~/local.

If you agree, I can allow it to work normally for -prefix.

habibayassin: Honestly, I believe if we only have -prefix indicating that it can be set with $HOME/local we can get rid of -local. Then we wouldn't need all of the new changes (new flags).

vvbandeira: Let's allow in prefix then and keep the local as it was a suggestion/request from Austin https://github.com/The-OpenROAD-Project/OpenROAD/pull/2405#pullrequestreview-1153810442

docs/user/BuildLocally.md
```diff
@@ -10,6 +10,10 @@ During initial setup or if you have a new machine, run this script:
 ``` shell
 # either run as root or use sudo
 ./etc/DependencyInstaller.sh
+
+Options:    ./etc/DependencyInstaller.sh -prefix=DIR    # installs dependencies in an existing user specified directory
+            ./etc/DependencyInstaller.sh -local         # installs common dependencies in "$HOME/.local". Root access is still needed for other
+                                                        # dependencies. Flag cannot be used with sudo or with root access
```
vvbandeira, line 16: Remove to avoid duplicated info. It is ok to have only line 12 as it will likely be the default path. If a user has special requirements such as using `-local` and `-prefix` options they can access their description via the top level docs `Build.md` where we have all the options listed.

etc/DependencyInstaller.sh
```diff
@@ -456,15 +456,22 @@ _installDebianPackages() {
...
 
-Usage: $0 -prefix=DIR
-       $0 -local
+Usage: $0                       Installs all of OpenROAD's dependencies no need to run -base or -common.
+                                                           Requires priviledged access. 
+       $0 -base                 Installs OpenROAD's dependencies using package managers 
+                                                           (-common must be executed in another command)
+       $0 -common               Installs OpenROAD's common dependencies (-base must be executed in another command)
+       $0 -prefix=DIR           Installs common dependencies in an existing user-specified directory. Only used with -common. This flag cannot be used with sudo or with root access.
+       $0 -local                Installs common dependencies in "$HOME/.local". Only used with -common. This flag cannot be used with sudo or with root access.
```
vvbandeira, line 465:
```suggestion
Usage: $0
                                # Installs all of OpenROAD's dependencies no
                                #     need to run -base or -common. Requires
                                #     priviledged access.
                                #
       $0 -base
                                # Installs OpenROAD's dependencies using
                                #     package managers (-common must be
                                #     executed in another command).

       $0 -common
                                # Installs OpenROAD's common dependencies
                                #     (-base must be executed in another
                                #     command).

       $0 -prefix=DIR
                                # Installs common dependencies in an existing
                                #     user-specified directory. Only used
                                #     with -common. This flag cannot be used
                                #     with sudo or with root access.

       $0 -local
                                # Installs common dependencies in
                                #    "$HOME/.local". Only used with
                                #    -common. This flag cannot be used with
                                #    sudo or with root access.
```
Slightly formatted, this way we can just copy the output of the `-help` and put in the docs.

README.md
```diff
@@ -145,8 +145,14 @@ script you need to run as root or prepend `sudo` to the commands below.
...
-Usage: ./etc/DependencyInstaller.sh -prefix=DIR    # installs dependencies in an existing user specified directory
-       ./etc/DependencyInstaller.sh -local         # installs dependencies in "$HOME/.local"
+Usage: ./etc/DependencyInstaller.sh                # Installs all of OpenROAD's dependencies no need to run -base or -common.
+                                                       # Requires priviledged access.
+       ./etc/DependencyInstaller.sh -base          # Installs OpenROAD's dependencies using package managers 
+                                                       # (-common must be executed in another command)
+       ./etc/DependencyInstaller.sh -common        # Installs OpenROAD's common dependencies (-base must be executed in another command)
+       ./etc/DependencyInstaller.sh -prefix=DIR    # Installs dependencies in an existing user specified directory
+       ./etc/DependencyInstaller.sh -local         # Installs common dependencies in "/home/habibayassin/.local". Only used with 
+                                                       # -common. Root access is still needed for other dependencies. Flag cannot be used # with sudo or with root access
```
vvbandeira, line 155:
```suggestion
Usage: etc/DependencyInstaller.sh
                                # Installs all of OpenROAD's dependencies no
                                #     need to run -base or -common. Requires
                                #     priviledged access.
                                #
       etc/DependencyInstaller.sh -base
                                # Installs OpenROAD's dependencies using
                                #     package managers (-common must be
                                #     executed in another command).

       etc/DependencyInstaller.sh -common
                                # Installs OpenROAD's common dependencies
                                #     (-base must be executed in another
                                #     command).

       etc/DependencyInstaller.sh -prefix=DIR
                                # Installs common dependencies in an existing
                                #     user-specified directory. Only used
                                #     with -common. This flag cannot be used
                                #     with sudo or with root access.

       etc/DependencyInstaller.sh -local
                                # Installs common dependencies in
                                #    "$HOME/.local". Only used with
                                #    -common. This flag cannot be used with
                                #    sudo or with root access.
```

etc/DependencyInstaller.sh
```diff
@@ -478,7 +485,17 @@ while [ "$#" -gt 0 ]; do
         -dev|-development)
             echo "The use of this flag is deprecated and will be removed soon"
             ;;
+        -base)
+            option="base"
+            ;;
+        -common)
+            option="common"
+            ;;
```
vvbandeira, line 493: Check if `option` is not `all` and give a warning about overwriting previous `-common` or `-base` argument.

etc/DependencyInstaller.sh
```diff
@@ -515,10 +532,18 @@ case "${os}" in
     "CentOS Linux" )
         spdlogFolder="/usr/local/lib64/cmake/spdlog/spdlogConfigVersion.cmake"
         export spdlogFolder
-        _installCentosPackages
-        _installCommonDev
-        _installOrTools "centos" "7" "amd64"
-        _installCentosCleanUp
+        if [[ "${option}" == "base" || "${option}" == "all" ]]; then
+            if [[ ! -z "${PREFIX}" ]]; then
+                echo "Error: cannot install base packages locally, you need priviledged access" >&2
```
vvbandeira, line 537:
```suggestion
                echo "Error: cannot install base packages locally; you need privileged access." >&2
```

etc/DependencyInstaller.sh
```diff
@@ -478,7 +485,17 @@ while [ "$#" -gt 0 ]; do
...
             ;;
+        -base)
+            option="base"
+            ;;
+        -common)
+            option="common"
+            ;;
         -local)
+            if [[ $(id -u) == 0 ]]; then
+                echo "Error: cannot install locally if you are root or using sudo" >&2
```
vvbandeira, line 496:
```suggestion
                echo "Error: cannot install locally (i.e., use -local) if you are root or using sudo." >&2
```

etc/DependencyInstaller.sh
```diff
@@ -529,18 +554,34 @@ EOF
...
         export spdlogFolder
-        _installUbuntuPackages "${version}"
-        _installCommonDev
-        _installOrTools "ubuntu" "${version}" "amd64"
-        _installUbuntuCleanUp
+        if [[ "${option}" == "base" || "${option}" == "all" ]]; then
+            if [[ ! -z "${PREFIX}" ]]; then
+                echo "Error: cannot install base packages locally, you need priviledged access" >&2
+                exit 1
+            fi
```
vvbandeira, line 561: Let's move this to a function to avoid copy and paste. Please note that the suggestion above should be addressed.

etc/DockerHelper.sh
```diff
@@ -56,9 +56,12 @@ _setup() {
         "ubuntu20")
             osBaseImage="ubuntu:20.04"
             ;;
-        "ubuntu22")
+        "ubuntu22.04")
```
vvbandeira, line 59: Let's use the same convention for all ubuntu versions and have the minor release as well on lie 56 and where else it is used.

etc/DockerHelper.sh
```diff
@@ -157,10 +160,14 @@ _push() {
...
-                ./etc/DockerHelper.sh create -target=dev -os=ubuntu22 -sha \
-                    2>&1 | tee build/create-ubuntu22-${commitSha}.log
+                ./etc/DockerHelper.sh create -target=dev -os=ubuntu22.04 \
+                    2>&1 | tee build/create-ubuntu22.04-latest.log
+                ./etc/DockerHelper.sh create -target=dev -os=ubuntu22.04 -sha \
+                    2>&1 | tee build/create-ubuntu22.04-${commitSha}.log
+                ./etc/DockerHelper.sh create -target=dev -os=ubuntu22.10 \
+                    2>&1 | tee build/create-ubuntu22.10-latest.log
+                ./etc/DockerHelper.sh create -target=dev -os=ubuntu22.10 -sha \
+                    2>&1 | tee build/create-ubuntu22.10-${commitSha}.log
```
vvbandeira, line 170: Could you change these repetitive commands to be in a for loop? Below as well for the `test` and dry-run push.

etc/DockerHelper.sh
```diff
@@ -260,7 +272,17 @@ target="dev"
...
 useCommitSha="no"
 isLocal="no"
-numThreads="$(nproc)"
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+  numThreads=$(nproc --all)
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+  numThreads=$(sysctl -n hw.ncpu)
+else
+  cat << EOF
+[WARNING FLW-0025] Unsupported OSTYPE: cannot determine number of host CPUs"
```
vvbandeira, line 281: Let`s not use `FLW` outside of ORFS.

vvbandeira (changes requested):

habibayassin:
@vvbandeira in order for the pipeline to succeed, an image with the new ubuntu tag needs to be pushed (i.e  openroad/ubuntu20.04-dev)

vvbandeira (approved):

PR #2732, opened 2023-01-11
ppl: add assert when pin assignment goes off the rails

oharboe:

### DIFF
--- a/src/ppl/src/IOPlacer.cpp
+++ b/src/ppl/src/IOPlacer.cpp
@@ -755,6 +755,15 @@ bool IOPlacer::assignPinsToSections(int assigned_pins_count)
 
   total_pins_assigned += assigned_pins_count;
 
+  if (total_pins_assigned > net->numIOPins()) {
+    logger_->error(
+        PPL,
+        13,
+        "Internal error, placed more pins than exist ({} out of {}).",
+        total_pins_assigned,
+        net->numIOPins());
+  }
+
   if (total_pins_assigned == net->numIOPins()) {
     logger_->info(PPL, 8, "Successfully assigned pins to sections.");
     return true;


src/ppl/src/IOPlacer.cpp
```diff
@@ -755,6 +755,8 @@ bool IOPlacer::assignPinsToSections(int assigned_pins_count)
 
   total_pins_assigned += assigned_pins_count;
 
+  assert(net->numIOPins() >= total_pins_assigned);
```
gadfort, line 758: Not sure an `assert` is the right thing. If it's an error then we should generate an error with a human readable string, or a warning if a warning is appropriate.

oharboe: This is simply a bug in the application, so I think an assert() is appropriate.

I think about errors as appropriate when the  application is correctly implemented, but cant create a result for some reason.

maliberty: Asserts can be compiled out but errors can't.  Also after an assert you can't inspect the application state.

maliberty: On a large design it may be too slow to run in a debugger conveniently.  If it happens in a release build it is still bad so why wouldn't you want the check?

oharboe: All good. I will change it to an error.

src/ppl/src/IOPlacer.cpp
```diff
@@ -758,6 +758,14 @@ bool IOPlacer::assignPinsToSections(int assigned_pins_count)
   if (total_pins_assigned == net->numIOPins()) {
     logger_->info(PPL, 8, "Successfully assigned pins to sections.");
     return true;
+  } else if (total_pins_assigned > net->numIOPins()) {
+    logger_->error(
+        PPL,
+        13,
+        "Internal error, placed more pins than exist ({} out of {}).",
+        total_pins_assigned,
+        net->numIOPins());
```
QuantamHD, line 767: I would suggest moving this to it's own if block above the first check 758. That way we conceptually guarantee that code after that point "isn't off the rails"

src/ppl/src/IOPlacer.cpp
```diff
@@ -755,7 +755,15 @@ bool IOPlacer::assignPinsToSections(int assigned_pins_count)
...
 
-  if (total_pins_assigned == net->numIOPins()) {
+  if (total_pins_assigned > net->numIOPins()) {
+    logger_->error(
+        PPL,
+        13,
+        "Internal error, placed more pins than exist ({} out of {}).",
+        total_pins_assigned,
+        net->numIOPins());
+    return false;
```
maliberty, line 765: error will throw an exception so there is no need for a return statement

src/ppl/src/IOPlacer.cpp
```diff
@@ -755,7 +755,14 @@ bool IOPlacer::assignPinsToSections(int assigned_pins_count)
...
 
-  if (total_pins_assigned == net->numIOPins()) {
+  if (total_pins_assigned > net->numIOPins()) {
+    logger_->error(
+        PPL,
+        13,
+        "Internal error, placed more pins than exist ({} out of {}).",
+        total_pins_assigned,
+        net->numIOPins());
+  } else if (total_pins_assigned == net->numIOPins()) {
```
QuantamHD, line 765: Minor nit: I think it makes sense to keep this block and the following as two seperate ifs. Totally up to you if you want to change it.

```c++
  if (total_pins_assigned > net->numIOPins()) {
    logger_->error(
        PPL,
        13,
        "Internal error, placed more pins than exist ({} out of {}).",
        total_pins_assigned,
        net->numIOPins());
  } 

  if (total_pins_assigned == net->numIOPins()) {
  ...
```

maliberty (approved):
I agree w/ @QuantamHD but will approve.

PR #2737, opened 2023-01-11
A few code cleanups in RCX as a first OpenROAD PR.

cdleary:
- Use unique_ptr in two spots where unique lifetime ownership was immediately clear.
- Delete some unreachable code flagged by static analysis tooling.
- Add Vim swp files to the ignored file extension list in .gitignore (similar to the existing ignorance of vscode files).

### DIFF
--- a/src/rcx/src/ext.cpp
+++ b/src/rcx/src/ext.cpp
@@ -51,10 +51,9 @@ extern "C" {
 extern int Rcx_Init(Tcl_Interp* interp);
 }
 
-Ext::Ext()
-  : odb::ZInterface(),
-    _ext(std::make_unique<extMain>())
-{}
+Ext::Ext() : odb::ZInterface(), _ext(std::make_unique<extMain>())
+{
+}
 
 void Ext::init(Tcl_Interp* tcl_interp, odb::dbDatabase* db, Logger* logger)
 {
--- a/src/rcx/src/ext.cpp
+++ b/src/rcx/src/ext.cpp
@@ -51,17 +51,10 @@ extern "C" {
 extern int Rcx_Init(Tcl_Interp* interp);
 }
 
-Ext::Ext() : odb::ZInterface()
-{
-  _ext = new extMain(5);
-  _tree = NULL;
-  logger_ = nullptr;
-}
-
-Ext::~Ext()
-{
-  delete _ext;
-}
+Ext::Ext()
+  : odb::ZInterface(),
+    _ext(std::make_unique<extMain>(5))
+{}
 
 void Ext::init(Tcl_Interp* tcl_interp, odb::dbDatabase* db, Logger* logger)
 {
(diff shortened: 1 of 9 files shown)


src/rcx/src/ext.cpp
```diff
@@ -51,17 +51,10 @@ extern "C" {
...
-  logger_ = nullptr;
-}
-
-Ext::~Ext()
-{
-  delete _ext;
-}
+Ext::Ext()
+  : odb::ZInterface(),
+    _ext(std::make_unique<extMain>(5))
```
QuantamHD, line 56: It looks like extMain's constructor accepts a param called menuId that's completely unused. I think it would be useful to remove it in the PR, and potentially remove a magic 5 from this file and others.

Other than that LGTM! Thanks for the contribution

cdleary: Done! I didn't see in the developer docs, are squashes preferred or is it ok to stack commits that address review comments?

maliberty:
In GH, once a review is given it is better to not force push as you lose the ability to see changes since last review.

cdleary:
> In GH, once a review is given it is better to not force push as you lose the ability to see changes since last review.

Understood. Normally I wouldn't (if OpenROAD doesn't require squashes), but I had forgotten to sign the commit, and I believe all the commits need to be done with the `-s` flag, correct?

src/rcx/src/ext.cpp
cdleary: Sounds like squashes are not necessary, resolving thread.

maliberty:
np and yes they all need to be signed

cdleary:
It's a little unclear what failed in the OS X build, I don't see anything saying "error" and the last command was just making an archive file and running ranlib on it which IIRC doesn't resolve symbols or anything (also don't see any unresolved symbol errors).

Maybe I'll delete some unused functions I found as a followon commit and see if it was just a flake for that build?

cdleary:
Looks like it was some sort of flake on OS X then, all green checks now.

src/rcx/src/netRC.cpp
```diff
@@ -1881,9 +1874,11 @@ bool extMain::setCorners(const char* rulesFileName, const char* cmp_file)
     logger_->info(
         RCX, 435, "Reading extraction model file {} ...", rulesFileName);
 
-    if (!fopen(rulesFileName, "r"))
+    FILE* rules_file = fopen(rulesFileName, "r");
+    if (rules_file == nullptr)
       logger_->error(
           RCX, 468, "Can't open extraction model file {}", rulesFileName);
+    fclose(rules_file);
```
maliberty, line 1881: error() will throw an exception and the file will be left open.  Probably an iostream would be simpler but otherwise you need to fclose before the error

cdleary: It's good to know that exceptions are enabled and the error routine throws, thanks for pointing that out.

I think this is ok in that circumstance though -- if the file handle is `nullptr` there's nothing to close, right? We only need to close in the case we successfully opened the file descriptor. (Really we should be using like `std::filesystem::exists` or a `stat` variant but I wanted to try to make minimal changes in this CL just to get my bearings :-)

maliberty: fair point

maliberty (approved):

PR #2741, opened 2023-01-12
Fix a number of compiler warnings

antonblanchard:


maliberty:
ready for review?

src/cts/src/Clock.cpp
```diff
@@ -95,7 +95,7 @@ Box<double> Clock::computeSinkRegionClustered(
   auto yMin = std::numeric_limits<float>::max();
   auto yMax = std::numeric_limits<float>::lowest();
 
-  for (const auto [sinkX, sinkY] : sinks) {
+  for (const auto& [sinkX, sinkY] : sinks) {
```
rovinski, line 98: What's the warning on this one? sinkX and sinkY are floats. They shouldn't need to be references.

antonblanchard:
> ready for review?

Yes. It doesn't remove all warnings with recent compilers, but it reduces it quite a lot.

maliberty (commented):
lgtm but please answer @rovinski 's question  (he seems correct)

src/cts/src/Clock.cpp
maliberty: See discussion in https://github.com/The-OpenROAD-Project/OpenROAD/pull/2751

maliberty (approved):

PR #2744, opened 2023-01-12
[rcx] ext2dBox translation unit: encapsulate/test

cdleary:
- Put a note on `AthPool` that the type parameter has to be default-constructible.
- Break `ext2dBox` into its own translation unit, remove friend access to members in favor of it being immutable-after-construction. We use placement-`new` since it needs to be default constructible to use `AthPool`.
- Use `std::array`s to indicate safe bounds instead of raw pointers / C arrays that degenerate to raw pointers.
- Make `dir` a boolean since it was always boolean-valued.
- Make a unit test (boost based).
- Factor out unit test running harness TCL code and put into common `helpers.tcl`

### DIFF
--- a/src/rcx/include/rcx/ext2dBox.h
+++ b/src/rcx/include/rcx/ext2dBox.h
@@ -54,7 +54,6 @@ class ext2dBox  // assume cross-section on the z-direction
            bool dir);
 
   void rotate();
-  bool matchCoords(int* ll, int* ur) const;
   void printGeoms3D(FILE* fp,
                     double h,
                     double t,
--- a/src/rcx/include/rcx/ext2dBox.h
+++ b/src/rcx/include/rcx/ext2dBox.h
@@ -1,0 +1,86 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2023, Nefelus Inc
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice, this
+//   list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+//   this list of conditions and the following disclaimer in the documentation
+//   and/or other materials provided with the distribution.
+//
+// * Neither the name of the copyright holder nor the names of its
+//   contributors may be used to endorse or promote products derived from
+//   this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <array>
+#include <cstdio>
+
+namespace rcx {
+
+class ext2dBox  // assume cross-section on the z-direction
+{
+ public:
+  // Implementation note: a default constructor is necessary because these are
+  // AthPool allocated, and it requires the type to be default constructible.
+  // We use placement new (with the full constructor below) for initialization
+  // after pool allocation.
+  ext2dBox() = default;
+
+  ext2dBox(std::array<int, 2> ll,
+           std::array<int, 2> ur,
+           unsigned int met,
+           unsigned int id,
+           unsigned int map,
+           bool dir);
+
+  void rotate();
+  bool matchCoords(int* ll, int* ur) const;
+  void printGeoms3D(FILE* fp,
+                    double h,
+                    double t,
+                    const std::array<int, 2>& orig) const;
+  unsigned int length() const;
+  unsigned int width() const;
+  int loX() const;
+  int loY() const;
+  unsigned int id() const;
+
+  unsigned int met() const { return _met; }
+  unsigned int map() const { return _map; }
+  bool dir() const { return _dir; }
+
+  int ur0() const { return _ur[0]; }
+  int ur1() const { return _ur[1]; }
+  int ll0() const { return _ll[0]; }
+  int ll1() const { return _ll[1]; }
+
+ private:
+  std::array<int, 2> _ll;
+  std::array<int, 2> _ur;
+  unsigned int _met;
+  unsigned int _id;
+  unsigned int _map;
+  bool _dir;
+};
+
+}  // namespace rcx
--- a/src/rcx/include/rcx/ext2dBox.h
+++ b/src/rcx/include/rcx/ext2dBox.h
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (c) 2023, Nefelus Inc
+// Copyright (c) 2019-2023, Nefelus Inc, Google LLC
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
--- a/src/rcx/src/ext2dBox.cpp
+++ b/src/rcx/src/ext2dBox.cpp
@@ -1,3 +1,35 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2023, Nefelus Inc
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice, this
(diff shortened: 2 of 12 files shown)


src/rcx/include/rcx/ext2dBox.h
```diff
@@ -0,0 +1,86 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2023, Nefelus Inc
```
QuantamHD, line 4: This should be "// Copyright (c) 2019-2023, Nefelus Inc, Google LLC"

src/rcx/src/ext2dBox.cpp
```diff
@@ -0,0 +1,107 @@
...
+          _id,
+          _map,
+          _met,
+          _dir,
+          0.001 * _ll[0],
+          0.001 * _ll[1],
+          0.001 * _ur[0],
+          0.001 * _ur[1],
+          0.001 * length(),
+          0.001 * width(),
```
QuantamHD, line 100: @maliberty This looks suspiciously like a database units translation. If so should this be using the ODB units value instead?

src/rcx/src/extmeasure.cpp
```diff
@@ -249,23 +249,19 @@ ext2dBox* extMeasure::addNew2dBox(dbNet* net,
 {
   ext2dBox* bb = _2dBoxPool->alloc();
 
+  std::array<int, 2> bb_ll;
+  std::array<int, 2> bb_ur;
```
QuantamHD, line 253: Now that you've refactored this I'm fairly certain that bb_ll and bb_ur stand for (bounding box lower left) and (bounding box upper right). Which means that these variable are likely storing rects. 

ODB has a class for storing these call odb::dbRect that we could potentially move these to. However, we could do that in another PR.

As another note, chip people's graphics origin is the lower left corner i.e. (0,0 is the lower left pixel on the screen) as opposed to most games and other graphics programming where 0,0 is the top left corner.

src/rcx/include/rcx/ext2dBox.h
cdleary: Updated copyright notices on touched files, PTAL.

QuantamHD: LGTM

src/rcx/src/ext2dBox.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+}
+bool ext2dBox::matchCoords(int* ll, int* ur) const
+{
+  if ((ur[0] < _ll[0]) || (ll[0] > _ur[0]) || (ur[1] < _ll[1])
+      || (ll[1] > _ur[1]))
+    return false;
+
+  return true;
+}
```
QuantamHD, line 55: This method is poorly named, but I'm 99% certain that this is checking if two boxes intersect 

```c++
// Does our X range intersect
(ur[0] < _ll[0]) || (ll[0] > _ur[0])

// Does our Y range intersect
(ur[1] < _ll[1]) || (ll[1] > _ur[1])
```

src/rcx/src/extmeasure.cpp
```diff
@@ -249,23 +249,19 @@ ext2dBox* extMeasure::addNew2dBox(dbNet* net,
...
-    bb->_ur[1] = ur[1];
+    bb_ll = {ll[0], ll[1]};
+    bb_ur = {ur[0], ur[1]};
   }
 
-  bb->_met = m;
-  bb->_dir = d;
-  bb->_id = id;
-  bb->_map = 0;
+  assert(d == 0 || d == 1);
```
QuantamHD, line 263: Should this be moved to a logger error instead of an assert?

src/rcx/src/extmeasure.cpp
maliberty: Mathematical convention is much older than either and aligns with EDA usage.

src/rcx/src/ext2dBox.cpp
maliberty: In short, who knows?  Maybe there is similar code elsewhere making the same assumption and so this is correct and maybe not.  It is the sort of thing that makes this code painful.  You would have to trace back to all the creators of these shapes that are being printed here.  Equally likely you'll find out that this function is itself dead code if you trace back far enough.

src/rcx/src/extmeasure.cpp
QuantamHD: Not arguing with the convention, but I find most non-EDA computer people are familiar with the upper left convention, and can cause confusion if not clarified.

maliberty: agreed

src/rcx/src/extmeasure.cpp
cdleary: Good review question -- is there OpenROAD developer guidance on when to prefer logging error (and having a corresponding exception thrown) vs assertion / why? Right now it basically seems like an invariant that the direction value was either 0 or 1. Since we're using `<cassert>` we don't see if there's an invariant violation in opt/release mode with the way it's done here.

QuantamHD: assert will crash the whole program, whereas error will only exit the current command. In most CI user configurations the result is the same. However, in interactive mode this may be less desirable.

src/rcx/src/ext2dBox.cpp
cdleary: I'm hoping "Maybe there is similar code elsewhere making the same assumption and so this is correct and maybe not" would show up in the regression testing / CI, but I'm also trying to avoid functional changes for now just to understand the development flow (e.g. I haven't mapped out how the RCX subsystem works yet, just trying to make basic changes).

src/rcx/src/extmeasure.cpp
cdleary: It does imply that invalid values may already exist and be floating around in in-memory objects, so the question is really whether "keep on trucking" is better in those circumstances. (Which is a value judgement.)

If so, it sounds a reasonable policy would be something like, "if there's a logger already accessible in your scope, prefer to assert using the logger error instead of a C assert". Will switch to that.

maliberty: Logger error is preferred.

cdleary: I'm also not sure how these error IDs are allocated, I grepped and eyeballed what the max one seemed to be and added one. Since I imagine this error is not very user-relevant I'm not sure if there's a good canonical way to flag that it's an internal error in the error string? LMK if so.

src/rcx/src/extmeasure.cpp
cdleary: Resolving, appreciate the clarification, could be a good addition to the developer intro docs.

rovinski:
Agree with @QuantamHD that now this is clear, there should be different types used.
Anything labeled `ll` or `ur` and has 2 `int` values is likely the lower-left and upper-right points of a rectangle. There is an `odb::Rect` class that could handle this.

`ll` and `ur` by themselves are (x, y) coordinates, and can be handled by the `odb::Point` class

Directions (where the only valid directions are horizontal or vertical) can be handled by the `odb::Orientation2D` enum class.

cdleary:
> Agree with @QuantamHD that now this is clear, there should be different types used. Anything labeled `ll` or `ur` and has 2 `int` values is likely the lower-left and upper-right points of a rectangle. There is an `odb::Rect` class that could handle this.

Makes sense -- do you agree with him that can be done in the subsequent PR? This PR breaks out the translation unit and adds unit testing so leaving "move to odb types" to the next PR seems sensible to me, "more smaller changes" usually better from an incremental refactoring standpoint.

src/rcx/src/ext2dBox.cpp
cdleary: Looks like this was dead, so I removed it.

rovinski:
> Makes sense -- do you agree with him that can be done in the subsequent PR?

Yes, fine by me.

src/rcx/src/extmeasure.cpp
cdleary: Ethan pointed out: https://openroad.readthedocs.io/en/latest/contrib/Logger.html#coding That resolves this question. Thanks!

cdleary:
Moving review questions from the PR description to a comment thread:

- I believe there's already a dependency on gtest via ortools dependency, that would be my preference but I wanted to ask if it'd be an acceptable thing to use in lieu of the boost unit testing constructs. I could do it as a followup.
- Going forward, what is a good place to either look-for-or-put a "read all the contents from this `FILE*` to a C++ string" utility (for testing in particular) if we want to make more tests of file-written strings, `utl/`?

src/rcx/src/ext2dBox.cpp
maliberty: The CI tests for rcx are far from comprehensive.  That's another area for improvement.

maliberty:
> * I believe there's already a dependency on gtest via ortools dependency, that would be my preference but I wanted to ask if it'd be an acceptable thing to use in lieu of the boost unit testing constructs. I could do it as a followup.

I don't think we have any gtest dependency via ortools today.  Why do you think so?

> * Going forward, what is a good place to either look-for-or-put a "read all the contents from this `FILE*` to a C++ string" utility (for testing in particular) if we want to make more tests of file-written strings, `utl/`?

utl seems fine

cdleary:
> > * I believe there's already a dependency on gtest via ortools dependency, that would be my preference but I wanted to ask if it'd be an acceptable thing to use in lieu of the boost unit testing constructs. I could do it as a followup.
> 
> I don't think we have any gtest dependency via ortools today. Why do you think so?

I saw them testing a bunch with gtest, but I'm not seeing a FindGtest in their CMakeLists, maybe those targets are only enabled on the Bazel build side. https://github.com/google/or-tools/blob/92eab40155d5566039c92662632f196174f8da47/ortools/pdlp/CMakeLists.txt#L20 Perils of a multi-build-system world.

> > * Going forward, what is a good place to either look-for-or-put a "read all the contents from this `FILE*` to a C++ string" utility (for testing in particular) if we want to make more tests of file-written strings, `utl/`?
> 
> utl seems fine

Thanks for the guidance!

src/rcx/src/ext2dBox.cpp
cdleary: Sounds good, as I get my bearings I'll ask how to generate more test collateral, would be happy to set up more testing so we can refactor with more confidence.

cdleary: Resolving since I don't think there's anything else outstanding in this thread.

maliberty:
I would guess or-tools developers use gtest but end users do not

cdleary:
> I would guess or-tools developers use gtest but end users do not

True -- in Bazel I believe the workspace brings in all deps necessary for all the targets to build (both mission mode and test targets), maybe CMake tends to have more of a "test deps vs mission mode deps" separation, or the CMake build system is somewhat secondary in the ortools project development flow.

QuantamHD:
I think this is ready to merge @maliberty

maliberty (approved):

PR #2751, opened 2023-01-13
Prevent warnings about unneeded copy in for loops

fgaray:
Just two small changes to prevent copy of std::pair in for loops

### DIFF
--- a/src/cts/src/Clock.cpp
+++ b/src/cts/src/Clock.cpp
@@ -95,7 +95,7 @@ Box<double> Clock::computeSinkRegionClustered(
   auto yMin = std::numeric_limits<float>::max();
   auto yMax = std::numeric_limits<float>::lowest();
 
-  for (const auto [sinkX, sinkY] : sinks) {
+  for (const auto& [sinkX, sinkY] : sinks) {
     xMin = std::min(xMin, sinkX);
     xMax = std::max(xMax, sinkX);
     yMin = std::min(yMin, sinkY);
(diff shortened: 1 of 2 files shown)


src/cts/src/Clock.cpp
```diff
@@ -95,7 +95,7 @@ Box<double> Clock::computeSinkRegionClustered(
   auto yMin = std::numeric_limits<float>::max();
   auto yMax = std::numeric_limits<float>::lowest();
 
-  for (const auto [sinkX, sinkY] : sinks) {
+  for (const auto& [sinkX, sinkY] : sinks) {
```
maliberty, line 98: These are just float so there is no copying

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -3034,7 +3034,7 @@ void HierRTLMP::multiLevelMacroPlacement(Cluster* parent)
   }
   // add the virtual connections (the weight related to IOs and macros belong to
   // the same cluster)
-  for (const auto [cluster1, cluster2] : parent->getVirtualConnections()) {
+  for (const auto& [cluster1, cluster2] : parent->getVirtualConnections()) {
```
maliberty, line 3037: These are just ints so there is no copying here either.

src/cts/src/Clock.cpp
gadfort: this does generate a warning though on my builds

fgaray: Sinks is a vector of std::pair, the unpack of the floats inside the vector is doing a copy of the std::pair before assignment to the floats, that's why we need the reference. It is a reference to the std::pair, not the floats.

fgaray: My gcc is "gcc (Debian 12.2.0-10) 12.2.0"

maliberty:
I looked in godbolt and the assembly is the same either way and the warning goes away so I'll go for it.  It seems a dumb warning for primitive types.

maliberty (approved):

PR #2755, opened 2023-01-14
lef: check if windows compile in lef.y and avoid referencing the lef_parser.hpp as it fails to compile

gadfort:
Fixes:
- Compiles for LEF parser only on windows.

### DIFF
--- a/src/odb/src/lef/lef/lef.y
+++ b/src/odb/src/lef/lef/lef.y
@@ -63,7 +63,10 @@
 #include "lefrCallBacks.hpp"
 #include "lefrSettings.hpp"
 
-#include "lef_parser.hpp"
+#ifndef WIN32
+  // Only include this on non-Windows platforms
+  #include "lef_parser.hpp"
+#endif
 
 BEGIN_LEFDEF_PARSER_NAMESPACE
 


maliberty:
What happens without this change on windows?  It looks strange to exclude that include.

gadfort:
> What happens without this change on windows? It looks strange to exclude that include.

A bunch of linker errors related to undefined symbols, it's not clear to me why it's in the .y to begin with, but this was the smallest change I needed to make to get windows to pass.

maliberty (approved):

rovinski:
...is there native Windows support in the making? Or is this for WSL?

maliberty:
> ...is there native Windows support in the making? Or is this for WSL?

There is no plan for native windows support.  It wouldn't be required for WSL.  I assume this is a one-off for some side purpose.

gadfort:
> > ...is there native Windows support in the making? Or is this for WSL?
> 
> There is no plan for native windows support. It wouldn't be required for WSL. I assume this is a one-off for some side purpose.

SC builds a lef parser and it's distributed on windows as well. It's a one of application, I doubt anyone else is doing that

PR #2761, opened 2023-01-15
grt: add load congestion report tip routing failure

oharboe:


src/grt/src/GlobalRouter.cpp
```diff
@@ -284,7 +284,7 @@ void GlobalRouter::globalRoute(bool save_guides)
     logger_->error(GRT,
                    118,
                    "Routing congestion too high. Check the congestion heatmap "
-                   "in the GUI.");
+                   "in the GUI and load the congestion.rpt in the DRC viewer.");
```
gadfort, line 287: You should reference the actual report if you are adding this `congestion_file_name_`. And if it's not set probably don't mention it and keep the message the same as before.

gadfort (changes requested):

maliberty:
@gadfort please review/approve

gadfort (approved):

PR #2762, opened 2023-01-15
gui: initial support to open GUI with python

gadfort:
This is a draft. It is meant to show a basic implementation of the GUI Python widget.

Adds:
- Python inputs

Changes needed:
- Need to move the stdout/stderr redirects to utl so it's consistent with Tcl and all printout go through the logger.
- Opening and closing of GUI is very closely tied to Tcl at the moment and therefore might not be possible to implement for Python until the restore code is updated
- Reading of gui settings is disabled until that can be fixed as well.


gadfort:
@maliberty it's still a work in progress, but if you have any thoughts let me know

maliberty:
I tried to compile this but get
```
                 from /workspaces/mliberty/w3/OpenROAD-flow-scripts/tools/OpenROAD/src/gui/src/pythonCmdInputWidget.cpp:36:
/usr/include/python3.6m/modsupport.h:41:24: error: redundant redeclaration of 'PyObject* _Py_BuildValue_SizeT(const char*, ...)' in same scope [-Werror=redundant-decls]
 PyAPI_FUNC(PyObject *) _Py_BuildValue_SizeT(const char *, ...);
                        ^~~~~~~~~~~~~~~~~~~~
/usr/include/python3.6m/modsupport.h:20:41: note: previous declaration of 'PyObject* _Py_BuildValue_SizeT(const char*, ...)'
 #define Py_BuildValue                   _Py_BuildValue_SizeT
                                         ^~~~~~~~~~~~~~~~~~~~
/usr/include/python3.6m/modsupport.h:40:24: note: in expansion of macro 'Py_BuildValue'
 PyAPI_FUNC(PyObject *) Py_BuildValue(const char *, ...);
                        ^~~~~~~~~~~~~
```

src/gui/src/gui.cpp
```diff
@@ -666,9 +666,9 @@ void Gui::saveImage(const std::string& filename,
     save_cmds += "unset ::gui::display_settings\n";
     // end with hide to return
     save_cmds += "gui::hide";
-    showGui(save_cmds, false);
+    showGui(Interpreter::Tcl, save_cmds, false);
   } else {
-    // save current display settings and apply new
+    // save current displexitingay settings and apply new
```
maliberty, line 671: typo

src/gui/src/scriptWidget.cpp
```diff
@@ -232,7 +253,7 @@ void ScriptWidget::addTextToOutput(const QString& text, const QColor& color)
 void ScriptWidget::readSettings(QSettings* settings)
 {
   settings->beginGroup(objectName());
-  input_->readSettings(settings);
+  // input_->readSettings(settings);
```
maliberty, line 256: Why?

maliberty (commented):
Generally looks good.  What is the next step?

gadfort:
@maliberty Python generates a lot of warnings, when included, I haven't taken time to figure out why yet.
The tasks listed as need to be completed, need to be fixed, I haven't had time to look into it yet.

src/gui/src/scriptWidget.cpp
gadfort: because input is created once the init is called for either python or tcl so it's a nullptr, I haven't figured out the best method to solve this yet.

Issue #2343, opened 2022-10-07
Par tool tests fail in Ubuntu 22


habibayassin:
Running `./test/regression par` in Ubuntu 22.04 with gcc version 11.2.0 causes the following tests to fail. Meanwhile, they pass on Ubuntu 20 with gcc version 9.4.0 and on Centos 7 with gcc version 4.8.5. 

```
partition (tcl) *FAIL* 
change_eval_method (tcl) *FAIL*
verilog_write (tcl) *FAIL*
report_netlist_partitions (tcl) *FAIL*
cluster (tcl) *FAIL*
```

diffs file
```
*** ~/OpenROAD/src/par/test/partition.ok	2022-10-04 08:42:59.098020854 +0000
--- ~/OpenROAD/src/par/test/results/partition-tcl.log	2022-10-07 21:24:00.706236225 +0000
***************
*** 8,11 ****
  [INFO PAR-0060] [MLPart] Run completed. Partition ID = 0. Total runs = 1.
  Writing partition id's to DB.
  Writing done.
! No differences found.
--- 8,13 ----
  [INFO PAR-0060] [MLPart] Run completed. Partition ID = 0. Total runs = 1.
  Writing partition id's to DB.
  Writing done.
! Differences found at line 1.
! _397_ 0
! _397_ 1
*** ~/OpenROAD/src/par/test/change_eval_method.ok	2022-10-04 08:42:59.098020854 +0000
--- ~/OpenROAD/src/par/test/results/change_eval_method-tcl.log	2022-10-07 21:24:01.322295143 +0000
***************
*** 20,35 ****
  [INFO PAR-0059] [MLPart] Partitioned graph for seed 1914837114.
  [INFO PAR-0060] [MLPart] Run completed. Partition ID = 1. Total runs = 2.
  [INFO PAR-0006] Partitioning Results for ID = 1 and Tool = mlpart.
! [INFO PAR-0007] Best results used seed 1914837114.
! [INFO PAR-0008] Number of Hyperedge Cuts = 53.
! [INFO PAR-0009] Number of Terminals = 116.
! [INFO PAR-0010] Cluster Size SD = 63.215405559088204.
! [INFO PAR-0011] Cluster Area SD = 756.9479423976262.
! [INFO PAR-0012] Total Hop Weight = 5300.
  [INFO PAR-0006] Partitioning Results for ID = 0 and Tool = mlpart.
! [INFO PAR-0007] Best results used seed -1731515372.
[...]

PR #2763, opened 2023-01-15
test multiple boost versions

habibayassin:


.github/workflows/github-actions-test-boost-version.yml
```diff
@@ -0,0 +1,37 @@
+name: Test Boost Version
+on:
```
vvbandeira, line 2: should also trigger on installer or yaml changes.

.github/workflows/github-actions-test-boost-version.yml
```diff
@@ -0,0 +1,54 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      boostVersion:
+        description: 'Boost Version'
+        required: true
+        default: '1.81.0'
```
vvbandeira, line 9: Can we have the dispatch receive a single version, but when it is triggered by the push or PR triggers then we run on multiple versions? If this cannot be one GHA that is fine, but it would be great if they were in the same.

habibayassin:
@vvbandeira regression tests are failing for all, do you want me to save the logs in the artifacts?

vvbandeira:
> @vvbandeira regression tests are failing for all, do you want me to save the logs in the artifacts?

Yes, please do save the logs and diffs files -- anything that would be helpful to understand the failures.

etc/DependencyInstaller.sh
```diff
@@ -12,9 +12,8 @@ _installCommonDev() {
     swigVersionType="tag"
     swigVersion=4.1.0
     swigChecksum="794433378154eb61270a3ac127d9c5f3"
-    boostVersionBig=1.80
-    boostVersionSmall=${boostVersionBig}.0
-    boostChecksum="077f074743ea7b0cb49c6ed43953ae95"
+    boostVersionBig=${boostVersion%.*}
+    boostVersionSmall=${boostVersion}
```
vvbandeira, line 16: Let's drop `boostVersionSmall` and replace with `boostVersion`. And move `boostVersionBig` to line ~58.

habibayassin:
> Yes, please do save the logs and diffs files -- anything that would be helpful to understand the failures.

@vvbandeira they're being saved now if you want to take a look. Latest run logs: https://github.com/The-OpenROAD-Project/OpenROAD/actions/runs/4020276840

vvbandeira:
Thanks, @habibayassin. This looks good, but I was expecting boost 1.80 + ubuntu 20.04 to work with ubuntu since we test that in the CI. Could you double-check these failures?

@maliberty FYI. There was talk on multiple boost versions before and maybe adding CMake minimum / supported versions. This PR is intended to add a GHA to allow us to test multiple versions of boost with ease.

habibayassin:
> Thanks, @habibayassin. This looks good, but I was expecting boost 1.80 + ubuntu 20.04 to work since we test that in the CI. Could you double-check these failures?

@vvbandeira some of the errors are Log errors either with "argument not found" or "invalid format string". There is an error because the number of threads in the ok file is not the same as the resulting file. Another error is how the numbers are printed, e.g. 6 not 6.0.

src/pdn/src/via.h
```diff
@@ -34,7 +34,7 @@
 
 #include <array>
 #include <boost/geometry.hpp>
-#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry.hpp>
```
rovinski, line 37: Duplicate line

etc/Build.sh
```diff
@@ -121,8 +121,14 @@ case "${compiler}" in
...
-        export CC="$(command -v gcc)"
-        export CXX="$(command -v g++)"
+        
+        if [[ $(awk -F= '/^NAME/{print $2}' /etc/os-release | sed 's/"//g') == "Ubuntu" &&  $(awk -F= '/^VERSION_ID/{print $2}' /etc/os-release | sed 's/"//g') == "22.10" ]]; then 
+            export CC="$(command -v gcc-11)"
+            export CXX="$(command -v g++-11)"
+        else
+            export CC="$(command -v gcc)"
+            export CXX="$(command -v g++)"
+        fi
```
vvbandeira, line 131: Do we still need the `gcc-11` code since we disabled the ubuntu 22.10?

test/regression
```diff
@@ -42,7 +42,6 @@ define_tool_script "tap" "src/tap/test/regression"
 define_tool_script "ppl" "src/ppl/test/regression"
 define_tool_script "pdn" "src/pdn/test/regression"
 define_tool_script "pad" "src/pad/test/regression"
-define_tool_script "mpl" "src/mpl/test/regression"
```
vvbandeira, line 45: are all the tests failing? we can also disable individual tests [here](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/mpl/test/regression_tests.tcl)

.github/workflows/github-actions-test-boost-version.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Build project
+        shell: bash
+        run: |
+          sudo ./etc/Build.sh -threads=3
+      - name: Test build
+        shell: bash
+        run: |
+          build/src/openroad -help
+          export OMP_NUM_THREADS=3
+          sudo ./test/regression
```
vvbandeira, line 88: did this had any effect in the end or can we remove? also could you give it a try again without sudo?

src/pdn/src/straps.cpp
```diff
@@ -134,7 +134,7 @@ bool Straps::checkLayerOffsetSpecification(bool error) const
           utl::PDN,
           185,
           "Insufficient width ({} um) to add straps on layer {} in grid \"{}\" "
-          "with total strap width {} um and offset {} um.",
+          "with total strap width {:.1f} um and offset {:.1f} um.",
```
vvbandeira, line 137: did this help?

habibayassin: yess for this kind of error "how the numbers are printed, e.g. 6 not 6.0."

test/regression
habibayassin: I removed the test files specified in this issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/2343

src/pdn/src/straps.cpp
vvbandeira: @maliberty is there a way to make the printing of floats uniform with the logger or maybe should we review the current messages and add the precision to each and start enforcing the use of precision on future messages?

etc/DependencyInstaller.sh
```diff
@@ -192,17 +188,23 @@ _installUbuntuPackages() {
...
-            libpython3.8
+            libpython3.8 \
+            gcc \
+            g++
     else
         apt-get install -y qt5-default \
-            libpython3.8
+            libpython3.8 \
+            gcc \
+            g++
```
vvbandeira, line 207: Missed this code? Should it be just `gcc g++` for all versions?

etc/DependencyInstaller.sh
```diff
@@ -192,17 +188,23 @@ _installUbuntuPackages() {
             qtchooser \
             qt5-qmake \
             qtbase5-dev-tools \
-            libpython3.11
+            libpython3.11 \
+            gcc-11 \
+            g++-11
```
vvbandeira, line 193: Missed this code? Should it be just gcc g++ for all versions?

src/pdn/src/straps.cpp
maliberty: I'm missing the context - what is varying?

vvbandeira: Depending on the OS/compiler version the output of floats with `{}` is undefined and varies between `6` and `6.0`, thus causing the test to fail when diffing the files.

maliberty:
@habibayassin @vvbandeira  stale?? close?

PR #2769, opened 2023-01-16
grt: estimate parasitics over corners based on rsz corner values if available

gadfort:
I noticed that the post routing parasitics didn't generate anything across for corners and didn't honor the rsz values when set for corners.
 
Changes:
- Generate parasitics across corners like resizer does (if only one corner is defined there shouldn't be any changes)
- Use the resizer parasitics values for corners, overwise default back to LEF

Notes:
If rsz values are not set and only one corner is defined, there should be no functional changes

### DIFF
--- a/src/grt/src/MakeWireParasitics.cpp
+++ b/src/grt/src/MakeWireParasitics.cpp
@@ -59,27 +59,20 @@ MakeWireParasitics::MakeWireParasitics(utl::Logger* logger,
                                        sta::dbSta* sta,
                                        odb::dbTech* tech,
                                        GlobalRouter* grouter)
+    : grouter_(grouter),
+      tech_(tech),
+      logger_(logger),
+      resizer_(resizer),
+      sta_(sta),
+      network_(sta_->getDbNetwork()),
+      parasitics_(sta_->parasitics()),
+      min_max_(sta::MinMax::max())
 {
-  grouter_ = grouter;
-  logger_ = logger;
-  resizer_ = resizer;
-  sta_ = sta;
-  tech_ = tech;
-  parasitics_ = sta_->parasitics();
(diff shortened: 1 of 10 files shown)


src/grt/src/MakeWireParasitics.cpp
```diff
@@ -93,13 +97,17 @@ void MakeWireParasitics::estimateParasitcs(odb::dbNet* net,
...
-  makeRouteParasitics(net, route);
-  makeParasiticsToPins(pins);
+  for (sta::Corner* corner : *sta_->corners()) {
+    node_id_ = 0;
+    net_corner_ = corner;
+    net_analysis_point_ = net_corner_->findParasiticAnalysisPt(min_max_);
+    net_parasitic_ = parasitics_->makeParasiticNetwork(
+        sta_net_, false, net_analysis_point_);
+    makeRouteParasitics(net, route);
+    makeParasiticsToPins(pins);
```
maliberty, line 109: Is there a reason to make these class members rather than arguments to the methods?  They seem like transient data.

gadfort: They were class members before, so I left it that way to try to minimize the code changes, but it would make better sense to pass them around in the function classes (which is how the placement estimator in rsz works)

maliberty: I'll approve now and you can fix that here or in another PR

maliberty (approved):

src/grt/src/MakeWireParasitics.cpp
gadfort: @maliberty I went ahead and made the switch in the PR, that removed all the member variables used during the parasitic estimation

PR #2776, opened 2023-01-17
Fixing metrics float handling

gadfort:
We ran into a new problem with the metrics handler and it was easier to fix the handling of `inf` and `nan` by converting them to a string metric and recording them so.
Additionally, Tcl doesn't really have floats they are all doubles, so large values that are legal in tcl will cause a SWIG error if `metric_float` takes a float.

Fixes:
- Change float to double in SWIG interface
- Add checking of values for `inf` and `nan` to ensure the metrics are recorded.
- Adds units tests to the metrics (along with ability to close the metrics file), so these issues can be tested for.

### DIFF
--- a/src/utl/include/utl/Logger.h
+++ b/src/utl/include/utl/Logger.h
@@ -38,6 +38,7 @@
 #include <array>
 #include <cstdlib>
 #include <iomanip>
+#include <limits>
 #include <map>
 #include <sstream>
 #include <stack>
@@ -46,7 +47,6 @@
 #include <type_traits>
 #include <unordered_map>
 #include <vector>
-#include <limits>
 
 #include "Metrics.h"
 #include "spdlog/fmt/fmt.h"
@@ -84,7 +84,8 @@ namespace utl {
   X(STA)                \
   X(STT)                \
   X(TAP)                \
-  X(UKN)
+  X(UKN)                \
+  X(UTL)
 
 #define GENERATE_ENUM(ENUM) ENUM,
 #define GENERATE_STRING(STRING) #STRING,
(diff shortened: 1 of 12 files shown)


src/utl/include/utl/Logger.h
```diff
@@ -174,11 +176,22 @@ class Logger
...
-    log_metric(std::string(metric), oss.str());
+    const std::string name = std::string(metric_name);
+    if (std::isinf(value)) {
+      if (value < 0) {
+        metric(name, "-inf");
+      } else {
+        metric(name, "inf");
+      }
+    } else if (std::isnan(value)) {
+      metric(name, "nan");
```
maliberty, line 189: In the [python json module](https://docs.python.org/3/library/json.html):
```
 If allow_nan is true, their JavaScript equivalents (NaN, Infinity, -Infinity) will be used.
```
Similarly with parse_constant in json.load.  We should match their naming.

src/utl/src/LoggerCommon.h
```diff
@@ -43,9 +43,10 @@ void warn(utl::ToolId tool, int id, const char* msg);
 void error(utl::ToolId tool, int id, const char* msg);
 void critical(utl::ToolId tool, int id, const char* msg);
 void open_metrics(const char* metrics_filename);
+void close_metrics(const char* metrics_filename);
```
maliberty, line 46: When will we use this?

maliberty (changes requested):

src/utl/include/utl/Logger.h
gadfort: Will do

src/utl/src/LoggerCommon.h
gadfort: Realistically, never, but I needed something to force the metrics to be flushed for CI, so I thought I could add it in case anyone wanted to record metrics in different files.
Without it the CI was not possible, so I think it's worth having something, but I'm open to suggestions

maliberty (approved):

PR #2785, opened 2023-01-19
Add support for C++20 Compilation

QuantamHD:
Couple of notes on C++20

* LEMON 1.3.1 has an issue that needs to be resolved related to allocators. (Included a patch from https://lemon.cs.elte.hu/trac/lemon/ticket/631#no1)
* fmt in versions greater than 8 added the ability to constexpr compile format lines. The downside is runtime strings need to be wrapped in fmt::runtime. Added FMT_RUNTIME macro to support both versions. The current version of spd_log is on version 7 of FMT
* template ids on constructors and destructors have been removed. Removed usage in abk libs.

Introduced a github action to build openroad on c++20. Should prevent any usage of non-supported c++20 features, but this does not mean the project is moving to c++20 yet.

**OpenROAD will stay on C++17**

### DIFF
--- a/.github/workflows/github-actions-c++.yml
+++ b/.github/workflows/github-actions-c++.yml
@@ -15,11 +15,6 @@ jobs:
           wget https://apt.llvm.org/llvm.sh
           chmod +x llvm.sh
           sudo ./llvm.sh 16
-      - name: Install Ninja
-        run: |
-          sudo wget -qO /usr/local/bin/ninja.gz https://github.com/ninja-build/ninja/releases/latest/download/ninja-linux.zip
-          sudo gunzip /usr/local/bin/ninja.gz
-          sudo chmod a+x /usr/local/bin/ninja
       - name: Check out repository code
         uses: actions/checkout@v3
         with:
@@ -42,4 +37,4 @@ jobs:
           cd ../..
       - name: Build OpenROAD
         run: |
-          ./etc/Build.sh -compiler='clang-16' -cmake='-GNinja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_STANDARD=20'
+          ./etc/Build.sh -compiler='clang-16' -cmake='-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_STANDARD=20'
--- a/.github/workflows/github-actions-c++.yml
+++ b/.github/workflows/github-actions-c++.yml
@@ -1,0 +1,27 @@
+name: Test C++20 Compile
+on:
+  # Triggers the workflow on push or pull request events
+  push:
+  pull_request:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install clang
+        run: |
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh 16
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      - name: Install dependencies
+        run: |
+          sudo ./etc/DependencyInstaller.sh
+      - name: Build OpenROAD
+        run: |
+          ./etc/Build.sh -compiler='clang-16' -cmake='-DCMAKE_CXX_STANDARD=20'
--- a/.github/workflows/github-actions-c++.yml
+++ b/.github/workflows/github-actions-c++.yml
@@ -15,6 +15,12 @@ jobs:
           wget https://apt.llvm.org/llvm.sh
           chmod +x llvm.sh
           sudo ./llvm.sh 16
+      - name: Install Ninja
+        run: |
+          sudo wget -qO /usr/local/bin/ninja.gz https://github.com/ninja-build/ninja/releases/latest/download/ninja-linux.zip
+          sudo gunzip /usr/local/bin/ninja.gz
+          sudo chmod a+x /usr/local/bin/ninja
+          sudo rm /usr/local/bin/ninja.gz
       - name: Check out repository code
         uses: actions/checkout@v3
         with:
@@ -37,4 +43,4 @@ jobs:
           cd ../..
       - name: Build OpenROAD
         run: |
-          ./etc/Build.sh -compiler='clang-16' -cmake='-DCMAKE_CXX_STANDARD=20'
+          ./etc/Build.sh -compiler='clang-16' -cmake='-GNinja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_STANDARD=20'
--- a/.github/workflows/github-actions-c++.yml
+++ b/.github/workflows/github-actions-c++.yml
@@ -27,7 +27,7 @@ jobs:
           sudo apt-get install mercurial
           hg clone http://lemon.cs.elte.hu/hg/lemon
           cd lemon
-          hg co ed2c21cb
+          hg co a278d16bd2d082aa3c52ff4a9b0e2224ddc0549a
           hg import ../third-party/lemon/patches/allocator-patch.patch
           mkdir build
           cd build
--- a/.github/workflows/github-actions-c++.yml
+++ b/.github/workflows/github-actions-c++.yml
@@ -20,7 +20,6 @@ jobs:
           sudo wget -qO /usr/local/bin/ninja.gz https://github.com/ninja-build/ninja/releases/latest/download/ninja-linux.zip
           sudo gunzip /usr/local/bin/ninja.gz
           sudo chmod a+x /usr/local/bin/ninja
-          sudo rm /usr/local/bin/ninja.gz
       - name: Check out repository code
         uses: actions/checkout@v3
         with:
--- a/.github/workflows/github-actions-c++.yml
+++ b/.github/workflows/github-actions-c++.yml
@@ -22,6 +22,19 @@ jobs:
       - name: Install dependencies
         run: |
           sudo ./etc/DependencyInstaller.sh
+      - name: Install Patched LEMON # Current LEMON does not compile with C++20 https://lemon.cs.elte.hu/trac/lemon/ticket/631#no1
+        run: |
+          sudo apt-get install mercurial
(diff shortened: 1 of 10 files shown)


QuantamHD:
ABK required a format see https://github.com/The-OpenROAD-Project/OpenROAD/pull/2785/commits/196ce1dad35a6985e9a23ea469f844b68db8675b#diff-f5e647735a7b1efd76bcb1f5f3a12f573e699220fa7bfbc3e68a1e9f339aa6d1

for C++20 changes only

maliberty:
@vvbandeira please comment on the action & build changes.

third-party/lemon/patches/allocator-patch.patch
```diff
@@ -0,0 +1,298 @@
+# HG changeset patch
```
maliberty, line 1: Is there a new lemon release we should upgrade to rather than apply one patch?

QuantamHD: The last lemon release was in 2014. There are commits after this point, but this patch was never merged in.

src/utl/include/utl/Logger.h
```diff
@@ -88,6 +88,13 @@ namespace utl {
 #define GENERATE_ENUM(ENUM) ENUM,
 #define GENERATE_STRING(STRING) #STRING,
 
+// backward compatibility with fmt versions older than 8
+#if FMT_VERSION >= 80000
+#define FMT_RUNTIME(format_string) fmt::runtime(format_string)
+#else
+#define FMT_RUNTIME(format_string) format_string
+#endif
```
maliberty, line 96: Kinda icky but hopefully the compile time checking is beneficial.

third-party/lemon/patches/allocator-patch.patch
QuantamHD: Also important to note that this patch is only applied in the action not the DependencyInstaller script which is intentional.

Really my goal was to get the CI action up and running and not to support c++20 in the project directly.

src/utl/include/utl/Logger.h
QuantamHD: You only need to add this macro when you are passing a non compile const string, or maybe said another way a dynamic template.

third-party/lemon/patches/allocator-patch.patch
gadfort: Does it make sense to fork that project and apply the patch there instead? It doesn't seem like they are maintaining that project and if it's a required dependency we need the ability to fix issues like this in a better way than a patch getting applied during an install step

src/utl/include/utl/Logger.h
gadfort: Can we add some developer documentation about this requirement in utl? Or could we provide a static function in utl to do this automatically? I worry developer not on c++20 will not know this and break things for you.

gadfort: Can we add some developer documentation about this requirement in utl? Or could we provide a static function in utl to do this automatically? I worry developer not on c++20 will not know this and break things for you.

gadfort: Can we add some developer documentation about this requirement in utl? Or could we provide a static function in utl to do this automatically? I worry developer not on c++20 will not know this and break things for you.

maliberty: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/docs/contrib/Logger.md would be a good place.  I think the added CI is to prevent such mistakes.

third-party/lemon/patches/allocator-patch.patch
maliberty: That sounds reasonable.  I'm sure there will be other users of c++20 - eventually even the project itself.

QuantamHD: Are we suggesting dumping the files in this repo, or adding a fork?

maliberty: I would make a new repo in GH.

QuantamHD: I'm a little bit torn on this particular issues the LEMON mercurial repo does have recent commits, but no new releases. 

I'm going to open a separate issue where we can decide what we want to do.

http://lemon.cs.elte.hu/hg/lemon

gadfort: I guess my point is this patch was proposed to them 3 years and it's still not in any release. I see the commits, but they don't appear to be that responsive. Please reference the issue when created here.

.github/workflows/github-actions-c++.yml
```diff
@@ -0,0 +1,40 @@
+name: Test C++20 Compile
+on:
+  # Triggers the workflow on push or pull request events
+  push:
+  pull_request:
```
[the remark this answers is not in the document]
maliberty: I suppose that time is mostly in setting up the dependencies.  It would be better to have a prepared container to skip that overhead per build.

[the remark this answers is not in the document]
QuantamHD: 5 minutes or so is in the dependencies. Which a container would help resolve. I could also look into switching the build from make to ninja.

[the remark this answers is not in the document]
maliberty: It should not take 40min to compile OR.  For comparison it is ~10min total to build in docker for our platforms in Jenkins.  The native builds at 5 minutes.

third-party/lemon/patches/allocator-patch.patch
QuantamHD: https://github.com/The-OpenROAD-Project/OpenROAD/issues/2791

.github/workflows/github-actions-c++.yml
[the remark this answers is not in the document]
QuantamHD: How many cores do those machines have? I think the Github runners are limited to 2.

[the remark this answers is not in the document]
QuantamHD: Got the compile time down to 28 minutes.

src/utl/include/utl/Logger.h
QuantamHD: Done.

etc/Build.sh
```diff
@@ -113,6 +113,16 @@ case "${compiler}" in
         export CC="$(command -v clang)"
         export CXX="$(command -v clang++)"
         ;;
+    clang-* )
```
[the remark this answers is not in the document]
QuantamHD: Done.

third-party/lemon/patches/allocator-patch.patch
QuantamHD: Discussion happening in https://github.com/The-OpenROAD-Project/OpenROAD/issues/2791

QuantamHD:
@vvbandeira I updated the spdlog version in the Dependency Installer script to v1.10.0. @maliberty signed off.

Letting you know so you can update the docker images in Jenkins.

maliberty:
@vvbandeira we'll need to update the CI images for the spdlog update

QuantamHD:
@maliberty This is ready to go on my side let me know if you need any more changes.

maliberty:
lgtm but I'll leave it to @vvbandeira to approve and coordinate the spdlog upgrade.

cdleary:
Just noting I was looking to use `std::span` which I believe is available as of C++20, in case the motivation is helpful. :-)

QuantamHD:
@vvbandeira Any objections to merging?

QuantamHD:
@cdleary std::span (gcc10) appears to not be supported by gcc8 on centos7. Which we need to continue to support. `boost::span` appears to be the next best option.

maliberty:
@cdleary we are only saying you can compile OR with c++, not that we are moving to c++20.  Any c++20 or later features should not be used in OR.

docs/contrib/Logger.md
```diff
@@ -33,6 +33,27 @@ critical, error, warning, information and debug. These are supported by
 automatic calls to the logger which will then prefix the appropriate
 severity type to the message.
 
+## C++20 Requirements
+
+In C++20 the logger messages are checked during compile time which introduces
+restrictions around rutime format strings. See [docs](https://fmt.dev/latest/api.html#compile-time-format-string-checks)
+
+OpenROAD uses SPDLog which usues fmt_lib under the hood below is an example of
+what is no longer allowed.
```
rovinski, line 42:
```suggestion
OpenROAD uses `spdlog` which uses `fmt_lib` under the hood. Below is an example of
what is no longer allowed.
```

docs/contrib/Logger.md
```diff
@@ -33,6 +33,27 @@ critical, error, warning, information and debug. These are supported by
...
+
+In order to make use of runtime format strings we have introduced a
```
rovinski, line 44:
```suggestion
In order to make use of runtime format strings, we have introduced a 
```

docs/contrib/Logger.md
```diff
@@ -33,6 +33,27 @@ critical, error, warning, information and debug. These are supported by
...
+
+In C++20 the logger messages are checked during compile time which introduces
+restrictions around rutime format strings. See [docs](https://fmt.dev/latest/api.html#compile-time-format-string-checks)
+
+OpenROAD uses SPDLog which usues fmt_lib under the hood below is an example of
+what is no longer allowed.
+
+In order to make use of runtime format strings we have introduced a 
+`FMT_RUNTIME` macro in Logger.h. You should use this any time that you are 
+passing a dynamic string as the format string
```
rovinski, line 46:
```suggestion
`FMT_RUNTIME` macro in Logger.h. You should use this macro any time you
pass a dynamic string as the format string 
```

QuantamHD:
@vvbandeira gentle ping on CI update for the C++20 patch. I think it's ready to go.

vvbandeira:
@QuantamHD, apologies for the delay. I have started to generate the new CI image, but since I am travelling today, it will be best to do the switch tomorrow when I have a more stable connection to address any issues.

QuantamHD:
No rush just wanted to make sure it was on your radar. Thanks for taking a look

vvbandeira (approved):

Issue #2770, opened 2023-01-17
grt: pin offset code doesn't mark gcells crossed as used


antonblanchard:
### Describe the bug

I have a test case that routes a bus around an obstruction:

![bus-route](https://user-images.githubusercontent.com/2029971/212798583-838cedd0-3606-4fca-9837-d198833f05b2.png)

When looking at the grt 3D grid usage, the first two cells next to the pins all have 0 usage (ie 20 available tracks):

```
20 20  1  1  1  1 11 26 30 30 30 30 30  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
 0  0  0  0  0  0 20 20 20 20 20 20 20  0 
20 20  1  1  1  1  1 16 20 20 20 20 20  0 
```

The issue looks to be the pin offset code, which moves the grid position for all pins 2 gcells in, but doesn't add any cost to the first 2 gcells. That seems bad, because any congestion would make grt want to route back through those 2 "empty" gcells.

I can guess at what the pin offset code is trying to achieve (avoid congestion in the gcell containing the pin), but I wonder how much tuning has been done of this (1 or 0 could easily be a better default).

### Expected Behavior

Correct grid usage on all gcells

### Environment

```shell
Git commit: 1ff101efbd3bf6ce4952e93c9c6b8db7e13c9346
os: Fedora Linux 37 (Workstation Edition)
cmake version 3.25.1
gcc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
clang version 15.0.6 (Fedora 15.0.6-2.fc37)
```

### To Reproduce

Run test case in https://github.com/The-OpenROAD-Project/OpenROAD/pull/2636

[...]

PR #2799, opened 2023-01-20
grt: decrease the capacity for fake pins segments

arthurjolo:
Signed-off-by: Arthur <arthurjolo@gmail.com>

Fixes: [issue#2770](https://github.com/The-OpenROAD-Project/OpenROAD/issues/2770#issuecomment-1396105370)

For each segment created to connect a 'fake pin' to its 'real pin', decreases the capacity of the edges that the segment goes throw, so that the fastroute considers this segments.

The test case in the mentioned issue shows an example. When running the test case with this changes, an underflow will occur in the capacity due to a bug in the function "computeTrackAdjustments()". But this bug is fixed by [this](https://github.com/The-OpenROAD-Project/OpenROAD/pull/2629)


src/grt/src/GlobalRouter.cpp
```diff
@@ -1880,6 +1882,56 @@ void GlobalRouter::connectPadPins(NetRouteMap& routes)
   }
 }
 
+void GlobalRouter::computePinOffsetAdjustments()
+{
+  for (auto net : block_->getNets()) {
```
eder-matheus, line 1887: I think you don't need to iterate over all nets. Just iterate directly into the `pad_pins_connections_` map. A reference here: https://stackoverflow.com/questions/26281979/c-loop-through-map

src/grt/src/GlobalRouter.cpp
```diff
@@ -1880,6 +1882,56 @@ void GlobalRouter::connectPadPins(NetRouteMap& routes)
...
+                                      gcell_id_y,
+                                      gcell_id_x + i + 1,
+                                      gcell_id_y,
+                                      segment.init_layer,
+                                      curr_cap,
+                                      true);
+          }
+        } else if (segment.init_x == segment.final_x) {
+          for (int i = 0; i < gcells_offset_; i++) {
+            int curr_cap = fastroute_->getEdgeCapacity(gcell_id_x + i,
```
eder-matheus, line 1915: Is this correct? If it follows the above code, the `+ i` should be on the `gcell_id_y`.

eder-matheus (changes requested):

eder-matheus:
@arthurjolo Did you run the unit tests? I expect some changes in the results, leading to errors in the grt unit tests. You may need to update the ok files.

src/grt/src/GlobalRouter.cpp
arthurjolo: yes it should be on the gcell_id_y, I will change that

src/grt/src/GlobalRouter.cpp
arthurjolo: Yes I wanted to improve that but I had never iterated over a map before and forgot to do it, thanks!

arthurjolo:
> @arthurjolo Did you run the unit tests? I expect some changes in the results, leading to errors in the grt unit tests. You may need to update the ok files.

No I didn't run the unity tests, will do it

src/grt/test/congestion1.ok
```diff
@@ -36,8 +36,8 @@ metal1     Horizontal          0             0          0.00%
 metal2     Vertical        24816          3008          87.88%
 metal3     Horizontal      33840          2806          91.71%
 metal4     Vertical        15792           564          96.43%
-metal5     Horizontal      15792           276          98.25%
-metal6     Vertical        15792           564          96.43%
+metal5     Horizontal      15792       3408096          -21481.16%
+metal6     Vertical        15792       3539452          -22312.94%
```
rovinski, line 40: 👀

eder-matheus:
The updates look good, but I've started a secure-ci run to ensure nothing will break.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2286,7 +2346,52 @@ GSegment GlobalRouter::createFakePin(Pin pin,
...
+
+    for (Pin& net_pin : nets_pins) {
+      if (!(net_pin.getITerm() == pin.getITerm())) {
+        auto net_pin_pos = net_pin.getOnGridPosition();
+        int net_pin_id_x
+            = floor((float) ((net_pin_pos.x() - die_area_min_x) / tile_size));
+        if ((net_pin_id_x >= init_id_x) && (net_pin_id_x <= final_id_x)) {
+          pin_position.setX(original_x);
+          pad_pins_connections_[net->getDbNet()].pop_back();
+          break;
```
eder-matheus, line 2370: Following my first comment about adding the `pin_connection` to the map, you can remove the `pop_back` line and use `return` instead of break. This way, you will still update the `pin_position`, but will not creat a connection when an instance pin is inside the pin offset.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2286,7 +2346,52 @@ GSegment GlobalRouter::createFakePin(Pin pin,
...
+
+    for (Pin& net_pin : net->getPins()) {
+      if (!(net_pin.getITerm() == pin.getITerm())) {
+        auto net_pin_pos = net_pin.getOnGridPosition();
+        int net_pin_id_y
+            = floor((float) ((net_pin_pos.y() - die_area_min_y) / tile_size));
+        if ((net_pin_id_y >= init_id_y) && (net_pin_id_y <= final_id_y)) {
+          pin_position.setY(original_y);
+          pad_pins_connections_[net->getDbNet()].pop_back();
+          break;
```
eder-matheus, line 2390: Same here.

eder-matheus (changes requested):

eder-matheus (approved):

PR #2795, opened 2023-01-20
pdngen: speed up PDN insertion

oharboe:
@maliberty @gadfort A stab in the dark... Runs MUCH faster, but does it still do the right thing?

Signed-off-by: Øyvind Harboe <oyvind.harboe@zylin.com>

PR #2803, opened 2023-01-21
pdn: fix repeated creation of rtrees in Shape::cut

gadfort:
Fixes:
- Filtering method to use boost filtering instead and leaving the obs rtree alone. This will likely be possible in several other places to help speed things up.

Replaces: #2795

### DIFF
--- a/src/pdn/src/shape.cpp
+++ b/src/pdn/src/shape.cpp
@@ -639,6 +639,9 @@ bool FollowPinShape::cut(const ShapeTree& obstructions,
   return Shape::cut(
       obstructions, replacements, [](const ShapeValue& other) -> bool {
         // followpins can ignore grid level obstructions
+        // grid level obstructions represent the other grids defined
+        // followpins should only get cut from real obstructions and
+        // not estimated obstructions
         return other.second->shapeType() != GRID_OBS;
       });
 }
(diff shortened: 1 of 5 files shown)


gadfort:
@oharboe FYI, I think this will have similar performance to your solution.

src/pdn/src/shape.cpp
```diff
@@ -627,17 +636,11 @@ const odb::Rect FollowPinShape::getMinimumRect() const
...
-      // followpins can ignore grid level obstructions
-      continue;
-    }
-    filtered_obstructions.insert({box, shape});
-  }
-
-  return Shape::cut(filtered_obstructions, replacements);
+  return Shape::cut(
+      obstructions, replacements, [](const ShapeValue& other) -> bool {
+        // followpins can ignore grid level obstructions
```
maliberty, line 641: Would you explain this a bit further?

maliberty (approved):

src/pdn/src/shape.cpp
gadfort: I added a little more of a comment

oharboe:
> @oharboe FYI, I think this will have similar performance to your solution.

Yes, if anything, it is marginally faster. I measured 9 seconds and with this change, I saw 7 seconds. Previously the same PDN insertion took 424 seconds. So 40x speedup, which is not nothing :-)

gadfort:
@oharboe thanks for pointing out the inefficiency.

PR #2815, opened 2023-01-25
Avoid evaluating floating point divide-by-zero in CTS

cdleary:
We observed there was a floating point divide by zero when running in a build with sanitizers on.

### DIFF
--- a/src/cts/src/HTreeBuilder.cpp
+++ b/src/cts/src/HTreeBuilder.cpp
@@ -348,17 +348,7 @@ void HTreeBuilder::computeLevelTopology(unsigned level,
 
   const unsigned minLength = minLengthSinkRegion_ / 2;
 
-  // Implementation note: this must be non-zero for the algorithm to work, or
-  // we end up dividing by a zero minLength.
-  if (minLength == 0) {
-    logger_->error(
-        CTS,
-        542,
-        "Minimum length must be non-zero; min length sink region: {}",
-        minLengthSinkRegion_);
-  }
-
-  unsigned segmentLength = std::round(width / (2.0 * minLength)) * minLength;
+  unsigned segmentLength = minLength == 0 ? 0 : std::round(width / (2.0 * minLength)) * minLength;
   if (isVertical(level)) {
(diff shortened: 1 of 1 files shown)


cdleary:
Sorry for the breakage, rookie mistake -- I should have run more tests but I thought to myself "oh this will make everything better defined where it wasn't before!" I think we just need a ternary. Will comment when updated.

maliberty:
I don't see any change - what am I missing?

cdleary:
It's very small, just avoiding running this expression: `std::round(width / (2.0 * minLength)) * minLength` when minLength is 0, since that requires `width / 0.0` to be evaluated. When we run with sanitizers on (in our Bazel build flow), this expression traps.

src/cts/src/HTreeBuilder.cpp
```diff
@@ -347,7 +347,9 @@ void HTreeBuilder::computeLevelTopology(unsigned level,
   logger_->report("    Sub-region size: {:.4f} X {:.4f}", width, height);
 
   const unsigned minLength = minLengthSinkRegion_ / 2;
-  unsigned segmentLength = std::round(width / (2.0 * minLength)) * minLength;
+
+  unsigned segmentLength
+      = minLength == 0 ? 0 : std::round(width / (2.0 * minLength)) * minLength;
   if (isVertical(level)) {
     segmentLength = std::round(height / (2.0 * minLength)) * minLength;
```
maliberty, line 354: It would be a error if it ever were zero and probably nothing else would work.  Why isn't it an error on 354 as well?

maliberty (approved):
It would be a error if it ever were zero and

src/cts/src/HTreeBuilder.cpp
cdleary: Agreed on both points -- I'll investigate more deeply than just working around the exception we saw, please don't merge yet. I'm hoping we can get to the bottom of why the logger error fails on the regression tests tomorrow. (Will try to convert this to "draft" mode to indicate we're going to investigate more.)

cdleary:
Closing this draft one as well in favor of unit tests we're going to write / workaround change that's in flight.

PR #2823, opened 2023-01-27
[cts] Fix base case handling for HTreeBuilder.

cdleary:
Previously when the region was too small for CTS HTree building we would set the minimum segment length to be `1` independent of the tech lef. This value of `1` would get divided in half when computing minimum length which would result in a value of `0`, which would become a divisor.

The correct fix seems to be to say that the minimum length comes directly from the characterization. This also simplifies the control flow -- we rip out the `enableFakeLutEntries` which was universally true and obfuscated the control flow.

Co-authored-by: Ethan Mahintorabi <ethanmoon@google.com>

### DIFF
--- a/src/cts/src/HTreeBuilder.cpp
+++ b/src/cts/src/HTreeBuilder.cpp
@@ -248,7 +248,6 @@ void HTreeBuilder::run()
   clockTreeMaxDepth_ = options_->getClockTreeMaxDepth();
   minInputCap_ = techChar_->getActualMinInputCap();
   numMaxLeafSinks_ = options_->getNumMaxLeafSinks();
-  minLengthSinkRegion_ = techChar_->getMinSegmentLength() * 2;
 
   initSinkRegion();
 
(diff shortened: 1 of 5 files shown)


QuantamHD:
I noticed that the wire length in the balance trees test is reduced with this change. I think this gets rid of a real problem.

maliberty:
I'm fine with removing enableFakeLutEntries but I don't understand your description of the problem you are fixing.  Where is the division by two that leads to 0?

src/cts/src/HTreeBuilder.cpp
```diff
@@ -346,7 +335,7 @@ void HTreeBuilder::computeLevelTopology(unsigned level,
   logger_->report("    Sinks per sub-region: {}", numSinksPerSubRegion);
   logger_->report("    Sub-region size: {:.4f} X {:.4f}", width, height);
 
-  const unsigned minLength = minLengthSinkRegion_ / 2;
```
QuantamHD, line 349: @maliberty if fake lutentries is triggered it sets minLengthSinkRegion_ = 1 on line 265 when you divide by 2 it's integer divided into 0

```
        unsigned minIndex = 1;
        techChar_->createFakeEntries(minLengthSinkRegion_, minIndex);
        minLengthSinkRegion_ = 1;
```

maliberty: I don't see any divide by two on the lines you mention.

QuantamHD: It's in the code block above my first comment?

const unsigned minLength = minLengthSinkRegion_ / 2;

maliberty:
This doesn't seem correct to me.  The fake lut entry is really creating an entry with a smaller length.  I think the problem is just that :
```
  const unsigned minLength = minLengthSinkRegion_ / 2;
  unsigned segmentLength = std::round(width / (2.0 * minLength)) * minLength;
```
should be
```
  const unsigned minLength = minLengthSinkRegion_;
  unsigned segmentLength = std::round(width / minLength) * minLength / 2;
```
That avoids the div 0 problem and segmentLength is protected by
```
  segmentLength = std::max<unsigned>(segmentLength, 1);
```

QuantamHD:
@maliberty We haven't had time to debug this yet, but a Google customer ran into this issue as well. It would be good to either get this merged in or another fix.

image

QuantamHD:
@grotival

maliberty:
If I were going to take a solution without any further investigation I'd be inclined to use the one I suggested.

cdleary:
Closing this PR in favor of unit tests we're going to write / the workaround which is in flight in other code.

Issue #2781, opened 2023-01-18
Add a GUI option to see all buffer connected nets


maliberty:
### Description

The inspector lets you see a single net easily.  However a single logical net may have been broken into many physical nets during buffering.  There should be some way to easily visualize all the buffer connected nets.

### Suggested Solution

I'm not positive of the best way to do this.  I am thinking of a inspector property on the net like "Aggregate Net" that would be a GUI object (not db) that contains a list of nets and buffers.  The highlight could draw the individual nets in different colors.  Here's an example I did by hand of tracing out such a aggregate (the root segment is green):
![image](https://user-images.githubusercontent.com/761514/213275786-7babb2df-e379-4010-afb9-ad8ef6aceb3d.png)

maliberty:
To be able to use the 'Route Guides' and focus buttons on the complex would be great too.

gadfort:
@maliberty Maybe calling it something like logical net might be a better term?
Also would you consider inverters and buffers the same?

maliberty:
I am hunting for a good name.  I'm concerned that logical net might be confused with nets in the hierarchical Verilog (which we will be storing in odb soon I hope).  "Net Complex" came to mind but I don't love that either.

Inverters are tricky.  I'm not interested in tracing through a user inverter but if resizer put in a pair instead of a buffer then I would be.  rsz doesn't do that today so its a bit theoretical.  Perhaps SOURCE NETLIST vs TIMING could be used to distinguish.

rovinski:
"Buffered Path"?

maliberty:
> "Buffered Path"?

Kinda. Its not a path like a timing path and is more of a tree of nets.  "Buffered Net" but that is also a bit odd.  Maybe there is no perfect, short name.

rovinski:
"Buffered (net) fanout" then?

maliberty:
> "Buffered (net) fanout" then?

Depending on which net you select you could be talking about fanout or fanin or both.  Perhaps "Buffer tree"?

gadfort:
@maliberty This PR adds the type, but does not implement the route guides

gadfort:
Competed in https://github.com/The-OpenROAD-Project/OpenROAD/pull/2824

PR #2824, opened 2023-01-29
gui: add aggregate net descriptor

gadfort:
https://github.com/The-OpenROAD-Project/OpenROAD/issues/2781

### DIFF
--- a/src/gui/src/aggregateNetDescriptor.cpp
+++ b/src/gui/src/aggregateNetDescriptor.cpp
@@ -124,9 +124,13 @@ bool AggregateNet::isAggregate(odb::dbInst* inst)
 
 ///////
 
-AggregateNetDescriptor::AggregateNetDescriptor(odb::dbDatabase* db,
-                                               sta::dbSta* sta)
-    : db_(db), net_descriptor_(Gui::get()->getDescriptor<odb::dbNet*>())
+AggregateNetDescriptor::AggregateNetDescriptor(
+    odb::dbDatabase* db,
+    sta::dbSta* sta,
+    const std::set<odb::dbNet*>& guide_nets)
+    : db_(db),
+      net_descriptor_(Gui::get()->getDescriptor<odb::dbNet*>()),
+      guide_nets_(guide_nets)
 {
   AggregateNet::setSTA(sta);
 }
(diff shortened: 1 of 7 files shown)


maliberty:
This looks like a bug as the aggregate is VDD:
image

src/gui/src/aggregateNetDescriptor.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+  for (auto* iterm : net->getITerms()) {
+    auto* inst = iterm->getInst();
+    if (!AggregateNet::isAggregate(inst)) {
+      continue;
+    }
+
+    for (auto* next_iterm : inst->getITerms()) {
+      populate(next_iterm->getNet());
+    }
```
maliberty, line 81: I expect the bug is here in not checking whether the iterm is power/gnd

gadfort: Yeah, you are correct, supplies shouldn't be in the list.

maliberty: if (!next_iterm->getSigType().isSupply()) {
        populate(next_iterm->getNet());
      }

seems to work

maliberty:
After the discussion in the issue I did come to like "Buffer tree" better than "Aggregate Net".   Perhaps it could include the buffer instances as well as the nets.

maliberty:
Also the non-buffer iterm/bterms would be nice (like a regular net) to see where the tree terminates.

gadfort:
@maliberty I renamed it to Buffer Tree, added iterms/bterms and instances to the descriptor (they are not part of the highlight though). Route guides and focus nets is also there.

src/gui/src/dbDescriptors.cpp
```diff
@@ -1332,6 +1333,10 @@ Descriptor::Properties DbNetDescriptor::getProperties(std::any object) const
     props.push_back({"Non-default rule", gui->makeSelected(ndr)});
   }
 
+  if (BufferTree::isAggregate(net)) {
+    props.push_back({"Aggregate net", gui->makeSelected(BufferTree(net))});
```
maliberty, line 1337: There is still some use of Aggregate in the code and visible in the GUI

src/gui/src/bufferTreeDescriptor.cpp
```diff
@@ -0,0 +1,310 @@
...
+BufferTree::BufferTree(odb::dbNet* net)
+{
+  populate(net);
+
+  std::sort(
+      nets_.begin(), nets_.end(), [](odb::dbNet* r, odb::dbNet* l) -> bool {
+        return r->getName() < l->getName();
+      });
+
+  name_ = nets_[0]->getName();
```
maliberty, line 58: std::min_element is more efficient than sort

src/gui/src/bufferTreeDescriptor.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+  for (auto* iterm : net->getITerms()) {
+    auto* inst = iterm->getInst();
+    if (!BufferTree::isAggregate(inst)) {
+      if (std::find(iterm_terms_.begin(), iterm_terms_.end(), iterm)
+          == iterm_terms_.end()) {
+        iterm_terms_.push_back(iterm);
+      }
+      continue;
+    }
```
maliberty, line 88: A std::set would be more efficient than linear search of a vector.

src/gui/src/bufferTreeDescriptor.cpp
```diff
@@ -0,0 +1,310 @@
...
+    return Selected(*anet, this);
+  }
+  return Selected();
+}
+
+bool BufferTreeDescriptor::lessThan(std::any l, std::any r) const
+{
+  BufferTree* l_anet = std::any_cast<BufferTree>(&l);
+  BufferTree* r_anet = std::any_cast<BufferTree>(&r);
+  return l_anet->getName() < r_anet->getName();
```
maliberty, line 238: anet needs updating

maliberty (changes requested):
I am really finding this helpful.  Thanks!

maliberty:
Free bonus, If you click a clock net you get the clock tree:
image

src/gui/src/bufferTreeDescriptor.cpp
gadfort: The sort is used to ensure a stable ordering of the nets.

gadfort:
Feedback should be addressed

maliberty (approved):

PR #2829, opened 2023-01-31
drt: detect patch owner on FlexGC

eder-matheus:
During FlexGC, DRT creates `drPatchWires` without assigning an owner. In a design from the Frankstein experiment, this led to the error `Error: pwire with no net`.

This PR updates the `checkMetalShape_addPatch` to detect the `drNet` that owns the `drPatchWire`.

Tested with the failing design, which ends successfully with this update. Currently running secure-ci.


src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -1849,11 +1849,67 @@ void FlexGCWorker::Impl::checkMetalShape_addPatch(gcPin* pin, int min_area)
...
+    }
+
+    for (auto&& conn_fig : dr_net->getRouteConnFigs()) {
+      if (patch_box.overlaps(conn_fig->getBBox())) {
+        patch->addToNet(dr_net);
+        return;
+      }
+    }
+
+    for (auto&& conn_fig : dr_net->getBestRouteConnFigs()) {
```
maliberty, line 1904: why both getRouteConnFigs & getBestRouteConnFigs ?

maliberty (approved):

PR #2839, opened 2023-02-02
dpl: filler_placement regexp is overly permissive

antonblanchard:
I see DRVs on ASAP7 designs because filler_placement is choosing TAPCELL_WITH_FILLER_ASAP7_75t_R cells, which when placed next to TAP cells results in nselect length issues. Even though we aren't requesting these cells, filler_placement is using them.

The issue is the regexp to filter cells is overly permissive and FILLER_ASAP7_75t_R is matching TAPCELL_WITH_FILLER_ASAP7_75t_R.

Fix this by adding start/end of string regexps. Also update the README as it claims we are doing wildcard matching but are in fact doing regexp matching.

### DIFF
--- a/src/dpl/README.md
+++ b/src/dpl/README.md
@@ -14,10 +14,11 @@ to legal locations after global placement.
 
 ```
 set_placement_padding -global|-instances insts|-masters masters
+                      [-regexp]
                       [-left pad_left] [-right pad_right]
 detailed_placement [-max_displacement disp|{disp_x disp_y}]
 check_placement [-verbose]
-filler_placement [-prefix prefix] filler_masters
+filler_placement [-prefix prefix] [-regexp] filler_masters
 remove_fillers
 optimize_mirroring
 ```
@@ -43,10 +44,10 @@ a weak attempt to reduce total wirelength (HPWL).
 
 The `filler_placement` command fills gaps between detail-placed instances
 to connect the power and ground rails in the rows. `filler_masters` is a
-list of master/macro names to use for filling the gaps. Wildcard matching
-is supported, so `FILL*` will match, e.g., `FILLCELL_X1 FILLCELL_X16 FILLCELL_X2
-FILLCELL_X32 FILLCELL_X4 FILLCELL_X8`.  To specify a different naming prefix
-from `FILLER_` use `-prefix <new prefix>`.
+list of master/macro names to use for filling the gaps. Both simple matching and
+regexp matching (via the `-regexp` flag) is supported, so `FILL*` and `-regexp FILL.*`
+will match `FILLCELL_X1 FILLCELL_X16 FILLCELL_X2 FILLCELL_X32 FILLCELL_X4 FILLCELL_X8`.
+To specify a different naming prefix from `FILLER_` use `-prefix <new prefix>`.
 
 ## Example scripts
 
--- a/src/dpl/src/Opendp.tcl
+++ b/src/dpl/src/Opendp.tcl
@@ -73,15 +73,17 @@ proc detailed_placement { args } {
 }
 
 sta::define_cmd_args "set_placement_padding" { -global|-masters masters|-instances insts\
+                                                 [-regexp]\
                                                  [-right site_count]\
-                                                 [-left site_count] \
+                                                 [-left site_count]\
                                                  [instances]\
                                                }
 
 proc set_placement_padding { args } {
   sta::parse_key_args "set_placement_padding" args \
-    keys {-masters -instances -right -left} flags {-global}
+    keys {-masters -instances -right -left} flags {-global -regexp}
 
+  set regexp [info exists flags(-regexp)]
   set left 0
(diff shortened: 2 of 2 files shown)


maliberty (approved):
I wonder if there are similar cases elsewhere

antonblanchard:
> I wonder if there are similar cases elsewhere

I was wondering the same (eg `set_dont_use`).

Another thing to note - this probably requires changes to the sky130 and gf180 pdks, because they are using wildcard matching vs regexps:

```
set ::env(FILL_CELL) "sky130_fd_sc_hd__fill*"

set ::env(FILL_CELL) "$::env(STD_CELL_LIBRARY)__fill_*"
```

I guess the question is, should we be doing wildcard matching and not regexps?

antonblanchard:
Similar issues for `set_placement_padding`:

```
set ::env(CELL_PAD_EXCLUDE) "sky130_fd_sc_hd__tap* sky130_fd_sc_hd__decap* sky130_ef_sc_hd__decap* sky130_fd_sc_hd__fill*"
```

```
set ::env(CELL_PAD_EXCLUDE) "$::env(STD_CELL_LIBRARY)__filltie_* $::env(STD_CELL_LIBRARY)__filldecap_* $::env(STD_CELL_LIBRARY)__fill_* $::env(STD_CELL_LIBRARY)__endcap_*"
```

QuantamHD:
I ran into this issue as well with a closed source PDK.

antonblanchard:
> As a note I added the begin and end line regexes to my string already, and I suspect this would break them. I'm fine with this change, but this is a breaking change.

It is, and I'm worried about how much breakage it will cause (including sky130 and gf180).

We need to decide if we want regexp matching or simpler globbing matching. I'd prefer regexps but they are likely to create more short term breakage than globbing.

maliberty:
In sdc the commands glob by default but have a -regexp option.  That's the most general way to handle it but would be more work.

antonblanchard:
> In sdc the commands glob by default but have a -regexp option. That's the most general way to handle it but would be more work.

That sounds good. I added -regexp to `set_placement_padding` and `filler_placement`. A couple of questions:

- `set_placement_padding -regexp` only applies to `-masters` and not `-instances`. To fix that we'd need `sta::get_instances_error` modified to take a regexp flag. Worth doing? FYI @jjcherry56  
- Would adding `-nocase` be useful?

gadfort:
I would like to support glob * over requiring .*

jjcherry56:
The command already supports regexp/nocase etc via using sdc get_cells in the arg which evals to an instance list that sta::get_instances_error already handles. This is how all STA commands work (like commercial tools) instead of adding -regexp/-nocase args to every command/arg combo. I think the problem is some platforms don't have the fillers in the liberty libraries so the model fails with get_lib_cells. From my perspective it is a bug in the libraries.

gadfort:
@maliberty I would certainly think having the same functionality in ODB (glob searching) would be extremely helpful.

jjcherry56:
It is called consistency. A library is a library is a library. Not some cells and not others. Not this command for some cells and that command for other cells. Yes, you can reinvent the wheel and make up commands that mirror the ones that are standardized by SDC (STANDARD design constraints) that tool designers already know but that doesn't make much sense to me. This is a problem that was solved in a very elegant way a long time ago.

maliberty:
which pdk has this issue of missing fill cells?

tspyrou:
We need to have a checker to check the pdk consistency. This is commonly done and ensures that the cells and their pins match up and are self consistent.

maliberty:
asap7, sky130*, and at least one private pdk would fail this 'requirement'.  

The global dependence on the timer is also not solved.  Fill cell placement has nothing to do with timing and only wants odb objects.

jjcherry56:
The issue has nothing to do with dependence on timing. It has to do with how command arguments work in the tool. Tools that support SDC for arguments (ie, commercial tools in the real world) implicitly call the corresponding get_cells/nets/pins function that allows glob style wildcards or a list of objects returned by the get function that has more complex arguments. 

set_filler_placement sky130_fd_sc_hd__fill*
set_filler_placement [get_lib_cells -regexp sky130_fd_sc_hd__fill[0-9]]

What matters is what the user sees. How you implement get_lib_cells under the hood by using existing opensta primitives or building odb based primitives for the get_* functions is a separate issue. My point is adding -regexp arguments to commands is a poor and inconsistent alternative.
Things get very messy once hierarchy and escaped dividers are added to the mix and what starts
out as a simple wildcard matching function for a flat db becomes complicated.

maliberty:
I'm all for using the SDC style commands.  I am concerned about how they are implemented and the internal implications.  Those SDC commands are implemented in sta and if we use them broadly that will affect the dependency structure.  Building odb based alternatives would generate two sets of commands which is worse.

maliberty:
The general issue remains but this is resolved for now by #3210

PR #2844, opened 2023-02-03
drt: Remove via2viaMinLenNew()

antonblanchard:
via2viaMinLenNew() is unused.


antonblanchard:
I presume this was meant to replace `via2viaMinLen()`, so perhaps we want to remove it instead. Regardless, there is so much code dealing with via2via distances that it would be nice to clean this up.

maliberty:
Is via2viaMinLen itself used?  I didn't see it with a quick grep but maybe I missed it

antonblanchard:
> Is via2viaMinLen itself used? I didn't see it with a quick grep but maybe I missed it

I had assumed it was, but it does appear unused as well. I added a patch to remove it too.

Does this mean we aren't applying any cut spacing rules when calculating route costs?

maliberty:
I believe what's happened here is that rp pre-computes these values and this is vestigial code.   getTech()->isVia2ViaForbiddenLen seems to be what is active now.

maliberty (approved):

PR #2850, opened 2023-02-05
Rename wlen2 to pinCoord

antonblanchard:
wlen2 is a cryptic name, and every time I look at TA I have to reacquaint myself with what it is.


maliberty (approved):
I've never spent the time to understand what it means so thanks for figuring it out.

PR #2855, opened 2023-02-06
Some rovinski cleanup rebased

oharboe:
Some commits that rebased cleanly from #1533

### DIFF
--- a/src/drt/src/dr/FlexDR.cpp
+++ b/src/drt/src/dr/FlexDR.cpp
@@ -322,18 +322,8 @@ void FlexDR::initGCell2BoundaryPin()
                   = getDesign()->getTopBlock()->getGCellBox(Point(x, y));
               frCoord leftBound = gcellBox.xMin();
               frCoord rightBound = gcellBox.xMax();
-              bool hasLeftBound = true;
-              bool hasRightBound = true;
-              if (bp.x() < leftBound) {
-                hasLeftBound = true;
-              } else {
-                hasLeftBound = false;
-              }
-              if (ep.x() >= rightBound) {
-                hasRightBound = true;
-              } else {
-                hasRightBound = false;
-              }
+              const bool hasLeftBound = bp.x() < leftBound;
+              const bool hasRightBound = ep.x() >= rightBound;
               if (hasLeftBound) {
                 Point boundaryPt(leftBound, bp.y());
                 gcell2BoundaryPin_[x][y][netPtr].insert(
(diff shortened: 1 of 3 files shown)


src/drt/src/dr/FlexDR.cpp
```diff
@@ -322,18 +322,8 @@ void FlexDR::initGCell2BoundaryPin()
...
-              } else {
-                hasLeftBound = false;
-              }
-              if (ep.x() >= rightBound) {
-                hasRightBound = true;
-              } else {
-                hasRightBound = false;
-              }
+              bool hasLeftBound = bp.x() < leftBound;
+              bool hasRightBound = ep.x() >= rightBound;
```
maliberty, line 326: these can be const now

src/drt/src/dr/FlexDR.cpp
```diff
@@ -354,18 +344,8 @@ void FlexDR::initGCell2BoundaryPin()
...
-              } else {
-                hasBottomBound = false;
-              }
-              if (ep.y() >= topBound) {
-                hasTopBound = true;
-              } else {
-                hasTopBound = false;
-              }
+              bool hasBottomBound = bp.y() < bottomBound;
+              bool hasTopBound = ep.y() >= topBound;
```
maliberty, line 348: const

src/drt/src/dr/FlexDR_init.cpp
```diff
@@ -2542,6 +2535,7 @@ void FlexDRWorker::route_queue_init_queue(queue<RouteQueueEntry>& rerouteQueue)
     // ripup all nets and clear objs here
     // nets are ripped up during initNets()
     vector<drNet*> ripupNets;
+    ripupNets.reserve(nets_.size());
     for (auto& net : nets_) {
       ripupNets.push_back(net.get());
     }
```
maliberty, line 2541: This can just be

vector<drNet*> ripupNets(nets_);

src/drt/src/dr/FlexGridGraph.cpp
```diff
@@ -66,20 +69,23 @@ void FlexGridGraph::initGrids(
   // initialize all grids
   frMIdx xDim, yDim, zDim;
   getDim(xDim, yDim, zDim);
+  int capacity = xDim * yDim * zDim;
```
maliberty, line 72: const

maliberty (approved):
After a review it is helpful if you don't force push, that way I can use the 'changes since last review' feature.

oharboe:
> After a review it is helpful if you don't force push, that way I can use the 'changes since last review' feature.

Ah. I'll leave the ```git commit --fixup```s in place next time.

maliberty:
It looks like these are subtly different in a way I didn't notice:
```
/OpenROAD/src/drt/src/dr/FlexDR_init.cpp:2537:35: error: no matching function for call to 'std::vector<fr::drNet*>::vector(std::vector<std::unique_ptr<fr::drNet> >&)'
```

I guess we'll have to go back to the prior version.  Sorry for the hassle.

PR #2857, opened 2023-02-06
Fix apparent typos in `InitFloorplan.tcl`

WRansohoff:
I could be wrong, but it looks like the `utl::logger` calls in `InitFloorplan.tcl` cause syntax errors if they are ever reached. Were these lines supposed to be `utl::error`?

Signed-off-by: WRansohoff <5217539+WRansohoff@users.noreply.github.com>


maliberty (approved):
Thanks

PR #2860, opened 2023-02-06
GHA for C++20 test uses Docker image

vvbandeira:


vvbandeira:
@QuantamHD fyi, using docker image to avoid instability. We have seen some PRs failing on this test due to a timeout while installing the dependencies.

QuantamHD:
Cool! Looks like a good change.

maliberty (approved):

PR #2862, opened 2023-02-06
drt: remove duplicated code, a workaround and unused function prototypes

antonblanchard:


maliberty (approved):

src/drt/src/dr/FlexDR_init.cpp
```diff
@@ -2957,14 +2939,8 @@ void FlexDRWorker::initMazeCost_fixedObj(const frDesign* design)
...
   for (auto& [net, objs] : frNet2Terms) {
-    // to remove once verify error will not be triggered
-    if (owner2nets_.find(net) == owner2nets_.end()) {
-      // cout << "Error: frNet with term(s) does not exist in owner2nets\n";
-      // continue;
-    } else {
-      for (auto dNet : owner2nets_[net]) {
-        dNet->setFrNetTerms(objs);
-      }
+    for (auto dNet : owner2nets_[net]) {
```
rovinski, line 2942: `auto&`?

maliberty: The type is drNet* so I don't think it will make much difference.

PR #2863, opened 2023-02-06
Adds Precompiled headers to build

QuantamHD:
This PR adds some commonly included headers as pre-compiled headers (PCH) it speeds up the build time pretty considerably on my laptop.

It only works on CMake 3.16 so added a few ifdef guards.

### DIFF
--- a/src/drt/CMakeLists.txt
+++ b/src/drt/CMakeLists.txt
@@ -138,6 +138,15 @@ target_include_directories(drt
     src
 )
 
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16.0") 
+  target_precompile_headers(drt
+    PRIVATE
+      src/serialization.h
+      src/io/io.h
+      src/dr/FlexDR.h
+  )
+endif()
+
 target_link_libraries(drt
   PUBLIC
     gui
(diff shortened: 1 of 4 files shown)


maliberty (approved):
Great.  I bet there are more places that would benefit.

I would be find to use 3.16 in cmake_minimum_required and drop the ifs.   Ubuntu 20, Centos8, FreeBSD12, and Debian 11 all are 3.16+

antonblanchard:
> I would be find to use 3.16 in cmake_minimum_required and drop the ifs. Ubuntu 20, Centos8, FreeBSD12, and Debian 11 all are 3.16+

Openlane is still CentOS7, but we really should move that and the OpenROAD CI to something newer.

maliberty:
centos7 defaults to cmake 2.8 which we don't support anyways.  You have to install a newer cmake on centos7 today.

QuantamHD:
It's super interesting to me that this fails CI.

maliberty:
> It's super interesting to me that this fails CI.

It is surprising.  It looks just like two tests on certain platforms.

maliberty:
@QuantamHD do you have any further insight on the CI?

maliberty:
@QuantamHD do you plan to look at the failures or close?

maliberty:
Reopen if you plan to work on this

PR #2866, opened 2023-02-07
grt: fix fake pin calculation

eder-matheus:
Update the checks for creating a fake pin position for IO and macro pins. Avoid creating a fake pin position only when another pin of the same net is inside the pin offset region.


eder-matheus:
@maliberty This fixes the DRC on the secure-ci. Currently running a whole secure-ci to ensure nothing else will break.

maliberty (approved):
I'll approve but please check secure CI before merging

src/grt/src/GlobalRouter.cpp
```diff
@@ -2243,41 +2243,26 @@ void GlobalRouter::createFakePin(Pin pin,
...
-            = floor((float) ((net_pin_pos.x() - die_area_min_x) / tile_size));
-        if ((net_pin_id_x >= init_id_x) && (net_pin_id_x <= final_id_x)) {
+  int pin_conn_init_x = pin_connection.init_x;
+  int pin_conn_init_y = pin_connection.init_y;
+
+  int pin_conn_final_x = pin_connection.final_x;
+  int pin_conn_final_y = pin_connection.final_y;
+
+  for (Pin& net_pin : net->getPins()) {
+    if (!(net_pin.getITerm() == pin.getITerm())) {
```
rovinski, line 2253:
```suggestion
    if (net_pin.getITerm() != pin.getITerm()) {
```

PR #2867, opened 2023-02-07
Breaks UPF Dependency Cycle

QuantamHD:
This change moves upf to the top level src folder, and makes it a proper module.

UPF dependency cycle

```
upf <- dbSta <- odb <- db <- upf
```

It also integrates the new app, lib structure.


QuantamHD:
@maliberty Can you fix the scan code?

maliberty:
Please update the table in docs/contrib/Logger.md

maliberty:
scan fixed

maliberty:
@ahmadelrouby will this affect any work you have in progress?

ahmadelrouby:
Currently it shouldn’t affect anything since I’m working on gpl module

QuantamHD:
Completed all requested changes.

src/dbSta/src/CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ target_sources(dbSta
 target_include_directories(dbSta
   PUBLIC
     ../include
+    ${PROJECT_SOURCE_DIR}/include
```
maliberty, line 75: Aren't these the same?  I'm not sure why the new line is needed.

src/upf/src/CMakeLists.txt
```diff
@@ -61,7 +73,7 @@ target_include_directories(upf
 
 target_link_libraries(upf 
   PRIVATE
+    upf_lib
     utl
-    dbSta
-    OpenSTA
+    odb
```
maliberty, line 78: upf_lib already depends on odb

src/upf/test/helpers.tcl
```diff
@@ -0,0 +1,89 @@
+# Helper functions common to multiple regressions.
```
maliberty, line 1: Should this be a symlink?

src/dbSta/src/CMakeLists.txt
QuantamHD: ${PROJECT_SOURCE_DIR} pull in the OpenRoad.hh files. Which were missing in dbSta

src/upf/src/CMakeLists.txt
QuantamHD: The SWIG also depends on odb headers which are not included by default. You need to call out odb explicitly.

src/upf/test/helpers.tcl
QuantamHD: Done.

src/upf/src/CMakeLists.txt
maliberty: Since upf.h includes db.h, odb should be a public dependency of upf_lib not a private one and this would be unnecessary.

QuantamHD: Hmm, I'm not a C++ guru, but at least at Google it would still be required to declare the dependency because a public header of rcx also imports odb headers.

We're not allowed to depend on the headers of our transitive dependencies.

QuantamHD: @cdleary do I have the above right?

maliberty (approved):

src/upf/src/CMakeLists.txt
maliberty: I don't agree in this case but don't care enough to debate it.

PR #2893, opened 2023-02-13
grt: update min routing layer calculation

eder-matheus:
Covers the discussion from https://github.com/The-OpenROAD-Project/OpenROAD/discussions/2870


maliberty (approved):

QuantamHD:
Same comment from the discussion. What happens if you can't actually route on this layer? It seems like there should be some way for the global router to lower it's min layer iteratively.

maliberty:
It depends if the min layer is a hard constraint or a soft one.  You usually have multiple layers available.

QuantamHD:
If we take @oharboe's case, but imagine there was a long blockage on M4 it make more sense to me to go to M3 rather than M5.

maliberty:
As a hypothetical I agree, and suggested a possible solution.  However it would be more work now and I'd rather get the problem at hand solved.

eder-matheus:
I have an idea of how to improve grt to make it prefer the specified layer instead of just limiting the solution space, but as Matt mentioned, it could take more time to implement. So I will work this week to enhance the layer assignment as we discussed here.

src/grt/src/GlobalRouter.cpp
```diff
@@ -828,9 +828,12 @@ bool GlobalRouter::makeFastrouteNet(Net* net)
                                        max_layer - 1,
                                        net->getSlack(),
                                        edge_cost_per_layer);
+    int min_pin_layer = std::numeric_limits<int>::max();
```
QuantamHD, line 831: @eder-matheus Would you mind throwing in a comment with a link to the various GitHub issues and discussion on why this particular code is here.

I'd like to make sure we don't lose the context if we want to change it in the future.

oharboe:
@maliberty @eder-matheus It seems to work a lot of the time, but there are lots of wires that could be horizontal/vertical that are not, e.g.:

First a wire from a pin is routed up a bit:

image

The down a bit:

image


So it works as intended, maybe 50% of the time?

image


Also, I see violations that are no longer resolved:

image

oharboe:
Trying with:

```
make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```

image

Here we can easily see that only some wires are horizontal, they go a bit up and down:

image

Vertically we can also see wires that could be vertical that go a bit left/right:

image

oharboe:
Trying a much bigger array (64x64 instead of 8x8, so 64x bigger):

```
MOCK_ARRAY_WIDTH=64 MOCK_ARRAY_HEIGHT=64 designs/asap7/mock-array-big/configure.sh
MOCK_ARRAY_WIDTH=64 MOCK_ARRAY_HEIGHT=64 make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```

eder-matheus:
@oharboe How do the guides look for these nets? It is possible that we have the detours from the global router, and it can be related to the number of nets being routed in the same region.

oharboe:
> @oharboe How do the guides look for these nets? It is possible that we have the detours from the global router, and it can be related to the number of nets being routed in the same region.

It takes a few minutes to run this locally, if you prefer to look directly:

```
make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```

eder-matheus:
> > @oharboe How do the guides look for these nets? It is possible that we have the detours from the global router, and it can be related to the number of nets being routed in the same region.
> 
> It takes a few minutes to run this locally, if you prefer to look directly:
> 
> ```
> make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
> ```

I've run it, and I don't see the DRC. Looking at the guides, they are behaving as expected, but I don't see an obvious reason for the detours created by the detailed router:
image image

In a lot of the nets with the detours, the pins are close to the edge of the guide, like in the figure below:
image
But the pins are completely covered by the guide. Perhaps drt has some costing for this situation, and I need to investigate. However, we still have cases where the pins are away from the edge of the guide, and the detour occurs.

Since the global router is behaving as expected, I would suggest merging this PR and keeping the discussion about the detailed router on https://github.com/The-OpenROAD-Project/OpenROAD/discussions/2870.

oharboe:
> I've run it, and I don't see the DRC.

@eder-matheus We don't have a robust way to ensure that we are both testing with the exact same code... I'm creating a mix locally and then testing, very slightly different from your version... Could that explain it?

@maliberty Told me that detailed routing is idempotent by design. So if I discover a case where it isn't I would report it.

eder-matheus:
> @eder-matheus We don't have a robust way to ensure that we are both testing with the exact same code... I'm creating a mix locally and then testing, very slightly different from your version... Could that explain it?

Yes, that's probably the case here.

maliberty:
You can always specify a commit id for OR & ORFS to make sure you are in sync.

oharboe:
Still there is the doubt about local or build problems.

I will create  a draft pull request to disambiguate next time it comes up.

tir. 14. feb. 2023, 17:30 skrev Matt Liberty ***@***.***>:

PR #2895, opened 2023-02-14
dpl: disallow one site gap implementation 

mina1460:
This is the implementation for the new option for the detailed placement engine. 
- usage:
` detailed_placement -disallow_one_site_gaps `

It currently checks for 1 site gaps to the left and right only of the cell being placed. 
I added a function similar to the visitCellPixels that can traverse boundary pixels only and apply a parameter function, and I use it in the check placement. 

Signed-off-by: Mina Ashraf <mina146@aucegypt.edu>


src/dpl/src/Opendp.tcl
```diff
@@ -32,11 +32,16 @@
...
 
 proc detailed_placement { args } {
   sta::parse_key_args "detailed_placement" args \
-    keys {-max_displacement} flags {}
+    keys {-max_displacement} flags {-disallow_one_site_gaps}
+
+  set disallow_one_site_gaps 0
+  if { [info exists flags(-disallow_one_site_gaps)] } {
+    set disallow_one_site_gaps 1
+  }
```
maliberty, line 44:
```suggestion
set disallow_one_site_gaps [info exists flags(-disallow_one_site_gaps)]
```

QuantamHD:
@maliberty Could we add the automatic inference based on fill cells in this PR? I think the option is still good to have, but theoretically we could iterate over all the masters, and check to see if there's a one site cell of type SPACER. Or really any cell since it doesn't theoretically matter what we put there.

src/dpl/include/dpl/Opendp.h
```diff
@@ -96,6 +96,15 @@ struct Master
   bool is_multi_row = false;
 };
 
+enum class Edge
+{
+  top,
+  bottom,
+  left,
+  right,
+  invalid
```
QuantamHD, line 105: Enums should be in all caps (TOP, BOTTOM), or if we're following the [Google style guide](https://google.github.io/styleguide/cppguide.html#Enumerator_Names) kTop, kBottom.

src/dpl/src/Opendp.cpp
```diff
@@ -182,7 +183,9 @@ void Opendp::setDebug(bool displacement,
   }
 }
 
-void Opendp::detailedPlacement(int max_displacement_x, int max_displacement_y)
+void Opendp::detailedPlacement(int max_displacement_x,
+                               int max_displacement_y,
+                               int disallow_one_site_gaps)
```
maliberty, line 188:
```suggestion
                               bool disallow_one_site_gaps)
```

src/dpl/src/Opendp.cpp
```diff
@@ -197,6 +200,12 @@ void Opendp::detailedPlacement(int max_displacement_x, int max_displacement_y)
     max_displacement_x_ = max_displacement_x;
     max_displacement_y_ = max_displacement_y;
   }
+  if (disallow_one_site_gaps == 0) {
+    // default
+    disallow_one_site_gaps_ = false;
+  } else {
+    disallow_one_site_gaps_ = disallow_one_site_gaps;
+  }
```
maliberty, line 208:
```suggestion
    disallow_one_site_gaps_ = disallow_one_site_gaps;
```

src/dpl/src/Place.cpp
```diff
@@ -859,6 +859,40 @@ bool Opendp::checkPixels(const Cell* cell,
         }
       }
     }
+    if (disallow_one_site_gaps_) {
+      // here we need to check for abutting first, if there is an abut cell
```
maliberty, line 863:
```suggestion
      // here we need to check for abutting first, if there is an abutting cell
```

src/dpl/src/Place.cpp
```diff
@@ -859,6 +859,40 @@ bool Opendp::checkPixels(const Cell* cell,
...
+      // then we continue as there is nothing wrong with it
+      // if there is no abut cell, we will then check cells at 1+ distances
```
maliberty, line 865:
```suggestion
      // if there is no abutting cell, we will then check cells at 1+ distances
```

src/dpl/src/Place.cpp
```diff
@@ -859,6 +859,40 @@ bool Opendp::checkPixels(const Cell* cell,
...
+      // we only need to loop on all 4 sides only
```
maliberty, line 866:
```suggestion
      // we only need to check on the left and right sides
```

src/dpl/src/Place.cpp
```diff
@@ -859,6 +859,40 @@ bool Opendp::checkPixels(const Cell* cell,
...
+      int x_begin = max(0, x - 1);
+      int y_begin = max(0, y - 1);
+      // inclusive search, so we don't add 1 to the end
+      int x_finish = min(x_end, row_site_count_ - 1);
+      int y_finish = min(y_end, row_count_ - 1);
+
+      // loop over left and right sides
```
maliberty, line 873: You are looping over the interior of the cell as well.  You only need to check the left and right edges (x-1, x_end) and not the points in-between.

src/dpl/include/dpl/Opendp.h
```diff
@@ -96,6 +96,15 @@ struct Master
   bool is_multi_row = false;
 };
 
+enum class Edge
+{
+  top,
+  bottom,
```
maliberty, line 102: No longer needed

src/dpl/include/dpl/Opendp.h
maliberty, line 105: rm - unused

src/dpl/include/dpl/Opendp.h
```diff
@@ -96,6 +96,15 @@ struct Master
   bool is_multi_row = false;
 };
 
+enum class Edge
```
maliberty, line 99: Can this be replaced with Direction1D from odb/include/odb/isotropy.h ?

src/dpl/src/CheckPlacement.cpp
```diff
@@ -203,6 +209,42 @@ bool Opendp::overlap(const Cell* cell1, const Cell* cell2) const
...
+
+        switch (edge) {
+          case Edge::left:
+            abut_x = -1;
+            break;
+          case Edge::right:
+            abut_x = 1;
+            break;
+          default:
+            break;
```
maliberty, line 229: It is best to avoid default cases.  If a new enum value is added the compiler can flag any switch statements that need updating.  In this case it reveals that the other values are unneeded.

src/dpl/src/Grid.cpp
```diff
@@ -188,6 +188,72 @@ void Opendp::visitCellPixels(
...
+          visitor(pixel, Edge::bottom, x, y_end - 1);
+      }
+      for (int y = y_start; y < y_end; y++) {
+        Pixel* pixel = gridPixel(x_start, y);
+        if (pixel)
+          visitor(pixel, Edge::left, x_start, y);
+        pixel = gridPixel(x_end - 1, y);
+        if (pixel)
+          visitor(pixel, Edge::right, x_end - 1, y);
+      }
```
maliberty, line 229: There is no need to visit the top & bottom edges

src/dpl/src/Grid.cpp
```diff
@@ -188,6 +188,72 @@ void Opendp::visitCellPixels(
...
+        visitor(pixel, Edge::bottom, x, y_end - 1);
+    }
+    for (int y = y_start; y < y_end; y++) {
+      Pixel* pixel = gridPixel(x_start, y);
+      if (pixel)
+        visitor(pixel, Edge::left, x_start, y);
+      pixel = gridPixel(x_end - 1, y);
+      if (pixel)
+        visitor(pixel, Edge::right, x_end - 1, y);
+    }
```
maliberty, line 253: same

maliberty (commented):
I wound up review it as I tried to understand the code.  Mostly there seems to be some vestiges of the old top/bottom check that was removed.  Otherwise it looks correct.

maliberty:
> @maliberty Could we add the automatic inference based on fill cells in this PR? I think the option is still good to have, but theoretically we could iterate over all the masters, and check to see if there's a one site cell of type SPACER. Or really any cell since it doesn't theoretically matter what we put there.

I think it is reasonable to infer the need for the check.  @mina1460 you would iterate over all the dbMasters of type SPACER and see if any has a width of 1 site or not.  If not then you would disable the 1-site gap.

src/dpl/src/Grid.cpp
mina1460: I am a bit confused here. I am looping over the y values, which I imagine would be the left and right edges. Is it the other way around?

I get the part that we don't need to check inner parts of the edge as the corners are enough.

src/dpl/src/Grid.cpp
mina1460: same concern. Is the x-axis and y-axis flipped? Wouldn't looping over y corresponds to vertical movement?

maliberty: You are right, it is the loop over x values that is obsolete

src/dpl/src/Grid.cpp
mina1460: Considering that this is a generic function that allows its users to iterate over the boundaries, wouldn't it be useful to leave both? the lambda (visitor) function can decide to ignore top and bottom edges. or we never need to visit top and bottom edges?

maliberty: I don't see a use case for it today but maybe it could be useful in the future.  In that case I would use Direction2D  to indicate the edge

QuantamHD:
@mina1460 Thanks for adding the inferencing.

I do have one concern (@maliberty) I've seen a few PDKs that don't properly mark their Filler cells as SPACER. With the current inferencing code if the foundry fails to mark the cells as SPACER then 1 site fillers get disabled silently which seems like a bad default. I would suggest one of two behaviors.

1. If the code detects no dbMasters of type SPACER do not disable 1 site fillers. There's not enough information to say whether or not this should happen.
2. Don't filter on type at all. Since most cells can be used as fillers perhaps we should just check to see if there are any masters with a 1 site width.

Thoughts?

maliberty:
filler_placement takes a list of cells to use.  We should just check that list and base our decision on that.  There is no use in considering cells that we won't fill with.

@mina1460 this is better than just checking all the SPACER cells as it explicitly follow the user's choice of cells.

maliberty:
@QuantamHD filling with non-fill cells would drive up leakage power for no purpose.  You can't have a 1-site cell of any sort in the relevant PDKs.

QuantamHD:
How is the command going to get the user regex for the filler cells? Will the user need to provide that to both the detailed_placement command and the filler_placement command? That seems like a poor user experience.

maliberty:
The detailed_placement command has nothing to do with filler placement

mina1460:
@maliberty I don't have access to that (dbMasterSeq* filler_masters) in the detailed placement. I can add it to the command similar to the filler_placement as you suggested, but that's the point @QuantamHD was making. 

The detailed_placement is called prior to the filler_placement so we don't have access to that list at that moment. So Ethan was asking how can we get that list at that point. Would the user write the regex in both commands? once in detailed_placement for me to use, and then another time for the filler_placement. That's why he thought it will be a bad user experience and probably a source of inconsistencies as the regex can be different in both commands.

On the other hand, if we checked all cells, even if they are indeed not used for filling as they cause power leakage, it might be better UX.

maliberty:
I see, you need to know about the possibility of 1-site filling when doing detailed_placement.  I think we should support auto-detection of the fill cells from the LEF SPACER type and if people want to use a custom list then they will have to specify it twice.  The two calls happen in two different flow steps so there is no good way to make them share data (I don't think it belongs in odb).

I expect most of the time the auto case should be correct and sufficient.

QuantamHD:
@maliberty I think that what I'm pointing out is that I have personally seen PDKs where the SPACER type is not used, and the fills are marked as CORE. The current code will disable 1 site fills unnecessarily which may never be uncovered by the user, they'll just get worse PPA.

I suggest we do one of two things where I prefer the first

1. Remove the type check all together. If there's a 1 site cell it's highly likely that there's a 1 site fill. If there's not it will be caught in check_placement command. (We should update the error message to let the user know about the existence of the disable-1-site-fill flag)
2.  Don't remove the type check, but if there are no SPACER cells detected then we bail out of inference. (i.e. not enough information to infer.)

rovinski:
I think I mostly agree with @QuantamHD. My preferred solution would be to:

* During DPL
  * If 1-site fill cell is detected: continue as normal
  * If 1-site fill cell is not detected && no `disallow_one_site_gaps` flag: warn the user then continue as normal
  * If `disallow_one_site_gaps` flag: only create 2-site gaps
* During fill cell insertion
  * Try to fill using user's fill cell list
  * If a gap exists which is smaller than the smallest fill cell, error and alert the user about the `disallow_one_site_gaps` flag

This will remove false positives, still error on problems, and not require the user to supply the fill cell list twice.

maliberty:
@QuantamHD I'm ok with what @rovinski suggests - are you?

One note on "only create 2-site gaps" should be "only create 0 or 2+-site gaps"

QuantamHD:
@rovinski That sounds good to me.

src/dpl/src/Opendp.cpp
```diff
@@ -200,13 +201,17 @@ void Opendp::detailedPlacement(int max_displacement_x,
...
-    disallow_one_site_gaps_ = disallow_one_site_gaps;
+  disallow_one_site_gaps_ = disallow_one_site_gaps;
+  if (!have_one_site_cells_) {
+    // If 1-site fill cell is not detected && no disallow_one_site_gaps flag:
+    // warn the user then continue as normal
+    if (!disallow_one_site_gaps_) {
+      logger_->warn(DPL,
+                    38,
+                    "No 1-site fill cells detected. to remove 1-site gaps, use "
+                    "-disallow_one_site_gaps flag");
```
maliberty, line 212:
```suggestion
                    "No 1-site fill cells detected.  To remove 1-site gaps use "
                    "the -disallow_one_site_gaps flag");
```

src/dpl/src/Place.cpp
```diff
@@ -860,37 +860,38 @@ bool Opendp::checkPixels(const Cell* cell,
...
-        pixel = gridPixel(x_finish, y1);
-        is_abut = (pixel == nullptr || pixel->cell);
-        if (!is_abut) {
-          // here we need to check the pixel at distance 1
-          pixel = gridPixel(x_finish + 1, y1);
-          bool cell_at_1 = (pixel != nullptr && pixel->cell);
-          if (cell_at_1)
-            return false;
-        }
+      auto isAbut = [this](int x, int y) {
```
maliberty, line 872: isAbutting or isAbutted

maliberty:
The code is close.  I'd like to see a unit test where a 1-site gap exists before legalization but is removed afterwards while an abutted case is kept.

maliberty:
I'm guessing your unit test failures need a merge of master to resolve.

mina1460:
The added unit test seems to be showing that the disallow_one_site_gaps option is working
Before:
Screenshot 2023-03-10 at 5 11 39 pm
Without the disallow:
A warning is issued: 
- [WARNING DPL-0038] No 1-site fill cells detected.  To remove 1-site gaps use the -disallow_one_site_gaps flag
Placement Analysis
Screenshot 2023-03-10 at 5 11 55 pm

and the results are: 
---------------------------------
total displacement          0.6 u
average displacement        0.1 u
max displacement            0.6 u
original HPWL              63.5 u
legalized HPWL             63.5 u
delta HPWL                    0 %


With the disallow flag:
Screenshot 2023-03-10 at 5 12 07 pm

No warning is issued in this case

results are:

Placement Analysis
---------------------------------
total displacement          0.8 u
average displacement        0.2 u
max displacement            0.6 u
original HPWL              63.5 u
legalized HPWL             63.5 u
delta HPWL                    0 %

mina1460:
All DPL failures are from log messages that I added. I will fix them once the code itself is reviewed so that I don't waste time editing log messages in the .ok files in case change was requested in the review.

maliberty:
@mina1460 what's happening with resolving the CI issues here?

mina1460:
@maliberty 
I am working on them. They are mostly from the warning we added when there are no filler cells of 1-site width detected.
so I was focusing on the IFP tests because they are more complex.

mina1460:
@maliberty this PR should be ready

maliberty:
Please resolve the merge conflicts from your other PR

mina1460:
@maliberty 
I fixed the conflicts and rebased to master. The tests pass with both changes together.

src/rsz/test/repair_wire10.ok
```diff
@@ -9,6 +9,7 @@
 [INFO ODB-0132]     Created 2 special nets and 6 connections.
 [INFO ODB-0133]     Created 4 nets and 6 connections.
 [INFO IFP-0001] Added 142 rows of 11052 site FreePDK45_38x28_10R_NP_162NW_34O with height 1.
+[WARNING DPL-0038] No 1-site fill cells detected.  To remove 1-site gaps use the -disallow_one_site_gaps flag.
```
maliberty, line 12: I don't see why we get this message for ng45.  FILLCELL_X1 is a one site filler afaik.

mina1460: That's a really good catch. I ended up discovering a bug that affected 10s of tests for this warning.

That was the logic of the checker, and FILLCELL_X1 is indeed a core spacer and of width equal to that of one site.

```Cpp
// Return true if cell is a single site Core Spacer.
bool Opendp::isOneSiteCell(odb::dbMaster* db_master) const
{
  return db_master->getType() == odb::dbMasterType::CORE_SPACER
         && db_master->getWidth() == site_width_;
}
```

The bug was that the site_width_ was 0.  I assumed it gets in the constructor or something in the initialisation, but it  was set in the  examineRows() method.
Because I was unaware of this, simply calling  checkOneSiteDbMaster() before examineRows() resulted in a lot of warnings. 

reordering them fixes the issue, but it makes me think if having site_width_ being set in the examineRows was a good idea or not.

Also, would it ever be the case that we have different site_width for different rows? for example as we did in multi-height rows but for width instead of height? If that might ever be the case, then the last row that examineRows iterate over would be the site_width_ used to check all one_site_cells

mina1460:
@maliberty 
Well, now PSM fails miserably.  
Noticeably, it is the only tool that is affected, apart from one test in rsz.

Reordering the two lines back again (examineRows and checkOneSiteCells) changes nothing and both still fails. Both were working today's morning, I will continue investigating this.

src/rsz/test/repair_wire10.ok
maliberty: In theory you could have different rows with different sites and they could have different widths (likely in different regions).  In practice we don't support that so I don't think we need to worry about it (many other things would also break too).

PR #2900, opened 2023-02-14
Adding scan replace to dft

fgaray:
This PR adds scan replace to the insert_dft command.

The main code is in ScanReplace.cpp where we look for all the available scan cells in the current lib and try to match them with the existing non-scan lib cells.  We iterate over the ports of each cell to see what are the best candidate to replace with scan cells.

Then, we iterate over the sequential cells' instances of the design and replace them keeping the same connectivity. We tie the scan pins (Scan In and Scan Enable) to ground making them functional equivalent as the original cells.

Basic regression testing is also added for sky130 and nandgate45 (3 new tests).

New message outputs:
- DFT-0003 INFO:  "Cell '{:s}' is already an scan cell, we will not replace it",
- DFT-0002 WARN: "Can't scan replace cell '{:s}', that has lib cell '{:s}'. No scan equivalent lib cell found"
- DFT-0001 WARN: "Lib cell '{:s}' doesn't have an scan equivalent"


maliberty:
When this is ready for review please unmark as a draft and request a review in GH.

fgaray:
> When this is ready for review please unmark as a draft and request a review in GH.

fgaray:
Thanks @maliberty !, Just fixed the regressions, now it is ready for review

src/dft/src/ScanReplace.hh
```diff
@@ -0,0 +1,108 @@
...
+#include "db_sta/dbSta.hh"
+#include "odb/db.h"
+#include "utl/Logger.h"
+
+namespace dft {
+namespace replace {
+
+// A scan cell that can be use to replace a non-scan cell
+// It contains the LibertyCell that we can use to instanciate a new cell and a
+// port map of how to connect the non-scan cell to the new one's equivalent
```
maliberty, line 46: Confusingly worded:  The port map is an equivalence relationship not a connection.

src/dft/src/ScanReplace.hh
```diff
@@ -0,0 +1,108 @@
...
+
+  // Returns the LibertyCell of the scan cell
+  sta::LibertyCell* getScanCell() const;
+
+  // How to connect the old non scan cell's ports to the new scan cell
+  const std::unordered_map<std::string, std::string>& getPortMapping() const;
+
+  // Prints to std::cout the port mapping to debug how we want to connect the
+  // old's cell ports to the new one
+  void debugPrintPortMapping() const;
```
maliberty, line 67: The logger should be used for printing not cout

src/dft/src/ScanReplace.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+namespace {
+
+// Checks if the given LibertyCell is really a Scan Cell with a Scan In and a
+// Scan Enable
+bool IsScanCell(const sta::LibertyCell* libertyCell)
+{
+  if (libertyCell->isBuffer() || libertyCell->isInverter()) {
+    return false;
+  }
```
maliberty, line 55: Is this for efficiency?  I don't expect either to have scan pins.

src/dft/src/ScanReplace.cpp
```diff
@@ -0,0 +1,426 @@
...
+  for (odb::dbBlock* next_block : block->getChildren()) {
+    scanReplace(next_block);
+  }
+}
+
+void ScanReplace::debugPrintScanEquivalents() const
+{
+  for (const auto& [liberty_cell, scan_candidate] :
+       non_scan_to_scan_lib_cells_) {
+    std::cout << liberty_cell->name() << " -> "
```
maliberty, line 420: Use the logger

src/dft/src/ScanReplace.cpp
```diff
@@ -0,0 +1,426 @@
...
+}
+
+// Recursive function that iterates over a block (and the blocks inside this
+// one) replacing the cells with scan equivalent
+void ScanReplace::scanReplace(odb::dbBlock* block)
+{
+  odb::dbNet* ground_net = utils::FindGroundNet(db_network_, block);
+
+  for (odb::dbInst* inst : block->getInsts()) {
+    if (inst->isDoNotTouch()) {
```
maliberty, line 359: You also need to look at whether the iterms connect to dont touch nets as you are going to modify them by disconnect/connect

src/dft/src/ScanReplace.cpp
```diff
@@ -0,0 +1,426 @@
...
+      }
+
+      // We skip clock gating lib cells
+      if (liberty_cell->isClockGate()) {
+        continue;
+      }
+
+      if (IsScanCell(liberty_cell)) {
+        available_scan_lib_cells_.insert(liberty_cell);
+      } else if (CanBeScanReplace(master, liberty_cell)) {
```
maliberty, line 300: Why is this necessary?  The cell has to be sequential to reach here and CanBeScanReplace only excludes buff/inv which are not sequential.

src/dft/src/ScanReplace.cpp
```diff
@@ -0,0 +1,426 @@
...
+  return diff;
+}
+
+// We select the scan_cell that is most similar to the non_scan_cell in
+// performance
+std::unique_ptr<ScanCandidate> SelectBestScanCell(
+    const sta::LibertyCell* non_scan_cell,
+    std::vector<std::unique_ptr<ScanCandidate>>& scan_candidates)
+{
+  std::sort(scan_candidates.begin(),
```
maliberty, line 223: Use std::max_element since you are only using at(0) from the result.

src/dft/src/ScanReplace.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+// We calculate the difference in performance between the cells iterating
+// through the ports
+double DifferencePerformanceCells(const sta::LibertyCell* non_scan_cell,
+                                  const sta::LibertyCell* scan_cell)
+{
+  double diff = 0;
+
+  // Area
+  diff += std::abs(non_scan_cell->area() - scan_cell->area());
```
maliberty, line 193: You are adding area to leakage to delay which makes no sense.  They are totally different units.

src/dft/src/ScanReplace.cpp
```diff
@@ -0,0 +1,426 @@
...
+  }
+
+  return true;
+}
+
+// We calculate the difference in performance of the ports given
+double DifferencePerformancePorts(const sta::LibertyPort* port1,
+                                  const sta::LibertyPort* port2)
+{
+  double diff = 0;
```
maliberty, line 167: adding incompatible units

maliberty (changes requested):

maliberty:
In general you don't tie signals directly to the power net physically.  A tie cell is used to avoid noise problems.  Nothing will route such a connection today.  Is this just some intermediate state while you are building out the code or an intended use model?

src/dft/src/dft.tcl
```diff
@@ -29,4 +29,6 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-
+proc insert_dft {} {
```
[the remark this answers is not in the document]
fgaray: Yes, In the future it is going to be all of dft: scan replacement, scan architect and scan stitching. At the end of the insert_dft command you will get a design with the cells scan replaced and connected to form scan chains.

Just that now it only do scan replace.

src/dft/include/dft/Dft.hh
```diff
@@ -31,10 +31,40 @@
...
+#include <unordered_set>
+
+#include "db_sta/dbSta.hh"
+#include "odb/db.h"
+#include "utl/Logger.h"
+
 namespace dft {
+namespace replace {
+class ScanReplace;
+}  // namespace replace
```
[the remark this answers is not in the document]
fgaray: The ScanReplace class is defined in ScanReplace.hh with another one. I guess the scope now is too small. I am fine with just defining this in only one namespace "dft", let me know if you would prefer that.

src/dft/src/dft.tcl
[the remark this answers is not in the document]
maliberty: In that case I would just leave the scan pins dangling as the code to ground them doesn't really make sense.  It is physically wrong and will have to be removed later anyways.

src/dft/src/Utils.hh
```diff
@@ -0,0 +1,71 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+#pragma once
+
+#include "db_sta/dbSta.hh"
+#include "odb/db.h"
+#include "sta/Liberty.hh"
+#include "utl/Logger.h"
+
+namespace dft {
+namespace utils {
```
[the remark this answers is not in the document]
fgaray: Ah, much better, thanks!

src/dft/src/Utils.hh
```diff
@@ -0,0 +1,71 @@
...
+
+// Creates a new cell in the given block using the master. Returns an instance
+// of the new cell with a default temporal name
```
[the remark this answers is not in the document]
fgaray: It is just a name for the cell that I am creating until I rename it with the original's cell name. I guess you are right that "temporal" is not 100% accurate here since I am changing the name outside the function. I will just leave it as "default name" if that sounds better.

src/dft/src/Utils.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+bool IsSequentialCell(sta::dbNetwork* db_network, odb::dbInst* instance)
+{
+  odb::dbMaster* master = instance->getMaster();
+  sta::Cell* master_cell = db_network->dbToSta(master);
+  sta::LibertyCell* liberty_cell = db_network->libertyCell(master_cell);
+  return liberty_cell->hasSequentials();
+}
+
+odb::dbInst* CreateCell(odb::dbBlock* top_block, odb::dbMaster* new_master)
```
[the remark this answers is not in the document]
fgaray: Yeah, you are right. I will just call the odb's function to create the instance instead of defining a new function. Thanks!

src/dft/src/MakeDft.cpp
```diff
@@ -32,19 +32,33 @@
...
 {
   return new dft::Dft();
 }
 
 void initDft(ord::OpenRoad* openroad)
 {
-  // TODO(#2784): implement initDft
+  Tcl_Interp* interp = openroad->tclInterp();
+  Dft_Init(interp);
+  sta::evalTclInit(interp, sta::dft_tcl_inits);
```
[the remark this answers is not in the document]
QuantamHD: Yeah as long as this code is in Make*.cc it works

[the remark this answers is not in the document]
maliberty: @QuantamHD  Is this preferable to passing in a std::function then?  It is simpler.

fgaray:
> In general you don't tie signals directly to the power net physically. A tie cell is used to avoid noise problems. Nothing will route such a connection today. Is this just some intermediate state while you are building out the code or an intended use model?

This is just for now while I implement the next step (scan stitching). I will just connect connect the scan in and scan enable to the corresponding pin/nets in a future PR.

Do you think I should just leave this cells without any connection?, or it is fine to have them connected to ground? Wdyt?

src/dft/src/dft.tcl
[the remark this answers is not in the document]
fgaray: Ok, this answer the other comment, I will just leave them dangling, thanks!

src/dft/src/ScanReplace.cpp
fgaray: Thanks!, I added an additional check

src/dft/src/ScanReplace.cpp
fgaray: You are right. I wonder if you have a better criteria in mind to select the best scan cell for a given cell. One alternative is to use the edit distance of the names of the cell vs the scan cell.

Usually in a lib, the scan cell equivalent will have a similar name to the non-scan one like:

- sky130_fd_sc_hd__dfstp_1
- sky130_fd_sc_hd__sdfsbp_1

Of course, this is just an heuristic and maybe there is a better way to select it. What do you think?

src/dft/src/MakeDft.cpp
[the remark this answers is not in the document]
QuantamHD: Yes. This wouldn't work for the graphics issue, but for the other modules this works, and is how some of the ones I converted already work.

src/dft/src/ScanReplace.cpp
QuantamHD: Here's an alternate way of measuring similarity that preserves units.

Treat attributes as vectors example, a = <area, leakage, port speed ,etc.>. Then norm the vector a/|a|, multiply vector by 1 over the magnitude.

Once you have a list of normed vectors for all the scan flops use the [Cosine Similarity]( https://en.wikipedia.org/wiki/Cosine_similarity) which ranges from 0 -> 1 where 1 is identical. Select max value.

maliberty: The elements of the vector are in very different scales so norming the vector won't improve that.  I think finding the nearest output strength is probably enough.

QuantamHD: You can subtract the mean vector to normalize scale.

maliberty: Even then it doesn't make sense to try to match area and leakage. Scan cells are going to be larger than their equivalent non scan versions. Output drive strength will matter to preserving timing

QuantamHD: Fair enough on leakage equivalent, but isn't the resizer going to change these cells anyways? It seems like the things that's most important is D -> Q latency

maliberty: There is no d->q path on an ff, I guess you mean clk->q.  I think the insertion should try to preserve the current timing regardless of what optimizations will happen later on.  The output drive strength is the main determiner of that.  Roughly you want an X2 gate to stay an X2 gate after swapping.

fgaray: Thanks @QuantamHD and @maliberty for the input here!

Matt, what is the current way of getting the output drive for a given LibertyPort? Thanks!

src/dft/src/Utils.cpp
```diff
@@ -0,0 +1,131 @@
...
+  // Rename as the old cell
+  new_instance->rename(old_cell_name.c_str());
+
+  return new_instance;
+}
+
+odb::dbNet* FindGroundNet(sta::dbNetwork* db_network, odb::dbBlock* block)
+{
+  for (odb::dbNet* net : block->getNets()) {
+    if (db_network->isGround(db_network->dbToSta(net))) {
```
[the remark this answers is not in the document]
fgaray: Deleted this function as we will just leave the scan pins dangling for now until we create the scan chains.

src/dft/src/Utils.cpp
```diff
@@ -0,0 +1,131 @@
...
+      return net;
+    }
+  }
+  return nullptr;
+}
+
+void TieScanPins(sta::dbNetwork* db_network,
```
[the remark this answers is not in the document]
fgaray: Yes, but I removed this function as it is no longer needed.

src/dft/src/ScanReplace.cpp
maliberty: LibertyPort::driveResistance() should work

src/dft/src/ScanReplace.cpp
fgaray: Removed. Solved in DifferencePerformanceCells

src/dft/src/ScanReplace.cpp
fgaray: Thanks!, I just changed it.

src/dft/test/caravel_sky130/chip_io.sigmap
```diff
@@ -0,0 +1,63 @@
+vssa_hvc_0 mgmt_vssa_hvclamp_pad
```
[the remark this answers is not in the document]
fgaray: Not really. I can remove it but what do you think if I just symlink to the caravel directory in ./src/pad/test/caravel_sky130?, or do you prefer to have the tests independent between them?, For example, Nangate45 is a symlink to another directory.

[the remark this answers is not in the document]
maliberty: I still don't understand why you need to test with caravel.  Pad does as it explicitly deals with IO rings.  Dft doesn't have any obvious need for caravel.  Smaller targeted unit tests seems better than one giant one.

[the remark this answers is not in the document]
fgaray: Yeah, I was using as reference the other test but it is not needed. I changed this and now the tests are much simpler. I am using just sky130hd now

fgaray:
Sorry @maliberty , I was missing resolving two comments!

fgaray:
Sorry @maliberty , I was missing resolving two comments!

maliberty:
Please avoid force pushing after a review as it blocks the function to see changes since last review.
image

src/dft/test/one_cell_sky130.ok
```diff
@@ -0,0 +1,9 @@
+[INFO ODB-0222] Reading LEF file: sky130hd/sky130hd.tlef
+[INFO ODB-0223]     Created 13 technology layers
+[INFO ODB-0224]     Created 25 technology vias
+[INFO ODB-0226] Finished LEF file:  sky130hd/sky130hd.tlef
+[WARNING ORD-1012] Liberty cell sky130_fd_sc_hd__dfstp_1 has no LEF master.
+[WARNING ORD-1013] instance ff1 LEF master sky130_fd_sc_hd__dfstp_1 not found.
+Differences found at line 1.
```
maliberty, line 7: There should be no differences in the .ok file

maliberty (changes requested):
almost

fgaray:
> Please avoid force pushing after a review as it blocks the function to see changes since last review. image

So sorry about that!

QuantamHD:
@maliberty I think @fgaray has addressed all the changes, but it still says one left. Is there anything missing?

maliberty (approved):

PR #2909, opened 2023-02-17
pad: rewrite ICeWall

gadfort:
Changes:
- Switches from a TCL based tool to a c++ based tool
- Uses an A*Star router to route the RDL, instead of using templates.

Notes:
- This leaves all the previous commands in place for backwards compatibility, this can be deprecated in a separate PR
- As a starting point for conversion, a helper command `pad::convert_tcl` can be used on a design built with the previous version to start the conversion.


src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+
+        odb::Rect shape = geom->getBox();
+        xform.apply(shape);
+        makeBTerm(net, layer, shape);
+      }
+    }
+  }
+}
+
+void ICeWall::makeFakeSite(const std::string& name, int width, int height)
```
maliberty, line 176: What about symmetry settings?

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+             vertical_site,
+             x_sites,
+             {nw->getBBox().xMax(),
+              outer_io.yMax() - static_cast<int>(vertical_site->getHeight())},
+             odb::dbOrientType::MX,
+             odb::dbRowDir::HORIZONTAL);
+  create_row("EAST",
+             horizontal_site,
+             y_sites,
+             {outer_io.xMax() - static_cast<int>(horizontal_site->getHeight()),
```
maliberty, line 307: - getWidth() ?

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+                               "IO_CORNER_SOUTH_WEST",
+                               "IO_CORNER_SOUTH_EAST"}) {
+    odb::dbRow* row;
+    if (ring_index >= 0) {
+      row = findRow(fmt::format("{}_{}", row_name, ring_index));
+    } else {
+      row = findRow(row_name);
+    }
+    if (row == nullptr) {
+      continue;
```
maliberty, line 359: error or warning?

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+      = std::max(horizontal_site->getHeight(), corner_site->getWidth())
+        / corner_site->getWidth();
+  auto create_corner = [block, corner_site, corner_sites, &name_format, &xform](
+                           const std::string& name,
+                           const odb::Point& origin,
+                           odb::dbOrientType orient) -> odb::dbRow* {
+    const std::string row_name = fmt::format(name_format, name);
+    odb::dbTransform rotation(orient);
+    rotation.concat(xform);
+    return odb::dbRow::create(block,
```
maliberty, line 258: dbBlock::getCoreArea() returns the bbox of the rows.  This will case the core area to expand out to the pads.  If you are going to use rows we'll need to rethink how the core area gets defined/stored.  The current icewall doesn't create rows.

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    inst = odb::dbInst::create(block, master, name.c_str());
+  }
+
+  if (row == nullptr) {
+    return;
+  }
+
+  odb::dbTransform orient(odb::dbOrientType::R0);
+  if (mirror) {
+    const odb::dbTransform mirror_transform(odb::dbOrientType::MY);
```
maliberty, line 398: why does mirror == MY?  Do we not mirror with MX on the other sides?

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+  const odb::Rect row_bbox = row->getBBox();
+  const std::string row_name = row->getName();
+
+  odb::dbTransform xform(base_orient);
+  xform.concat(row->getOrient());
+  inst->setOrient(xform.getOrient());
+  const odb::Rect inst_bbox = inst->getBBox()->getBox();
+
+  odb::Point index_pt;
+  if (row_name.find("NORTH") != std::string::npos) {
```
maliberty, line 452: Can we avoid using string lookup to find the side?

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+    int width;
+    int start;
+    if (row->getDirection() == odb::dbRowDir::HORIZONTAL) {
+      width = new_rect.dx();
+      start = new_rect.xMin();
+    } else {
+      width = new_rect.dy();
+      start = new_rect.yMin();
+    }
+    int sites = width / site_width;
```
maliberty, line 597: What happens if the width isn't an integral site_width?  The round down could leave a gap.  If the assumption is that is an integral multiple then we should have a check.

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+        2,
+        "Instance in {} -> {} ({:.3f}um, {:.3f}um) -> ({:.3f}um, {:.3f}um)",
+        row->getName(),
+        inst->getName(),
+        pts[0].x() / dbus,
+        pts[0].y() / dbus,
+        pts[2].x() / dbus,
+        pts[2].x() / dbus);
+
+    Polygon90 poly;
```
maliberty, line 560: I don't think you really need polygon here as this a one dimensional problem (the cells are all the same height).   https://www.boost.org/doc/libs/1_81_0/libs/icl/doc/html/index.html would suffice

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+        auto* pin_layer = geom->getTechLayer();
+        if (pin_layer == nullptr) {
+          continue;
+        }
+        if (pin_layer->getRoutingLevel() == 0) {
+          continue;
+        }
+        bond_layer = pin_layer;
+        bond_pin = mterm;
+        bond_rect = geom->getBox();
```
maliberty, line 698: The assumption seems to be that there is a single shape.  Is that always true?

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+            bond_inst->getITerm(bond_pin)->connect(net);
+            makeBTerm(net, pin_layer, pin_shape);
+          }
+        }
+      }
+    }
+  }
+}
+
+void ICeWall::connectByAbutment()
```
maliberty, line 754: This is rather inefficient but perhaps the number of objects is small enough to get away with it.

src/pad/src/RDLRouter.h
```diff
@@ -0,0 +1,201 @@
...
+  void setRouter(RDLRouter* router) { router_ = router; }
+
+  virtual void drawObjects(gui::Painter& painter) override;
+
+  virtual const char* getDisplayControlGroupName() { return "RDL Router"; }
+
+ private:
+  RDLRouter* router_;
+
+  static constexpr const char* draw_vertex_ = "Vertecies";
```
maliberty, line 80: Vertices

src/pad/src/RDLRouter.h
```diff
@@ -0,0 +1,201 @@
...
+  static constexpr const char* draw_obs_ = "Obstructions";
+};
+
+class RDLRouter
+{
+ public:
+  struct RouteTarget
+  {
+    odb::Point point;
+    odb::Rect shape;
```
maliberty, line 91: why both?

src/pad/src/RDLRouter.h
```diff
@@ -0,0 +1,201 @@
...
+
+  std::vector<Edge> insertTerminalVertex(const RouteTarget& target,
+                                         const RouteTarget& source);
+
+  std::vector<TargetPair> generateRoutingPairs(odb::dbNet* net) const;
+
+  int getBloatFactor() const;
+
+  std::vector<int> x_grid_;
+  std::vector<int> y_grid_;
```
maliberty, line 198: don't split the data members, put them all here after the methods

maliberty (changes requested):
I haven't gotten to the router yet but its a first pass

src/pad/include/pad/ICeWall.h
```diff
@@ -0,0 +1,161 @@
...
+                odb::dbTechVia* via,
+                const std::vector<odb::dbNet*>& nets,
+                int width = 0,
+                int spacing = 0,
+                bool allow45 = false);
+  void routeRDLDebugGUI(bool enable);
+
+  void connectByAbutment();
+
+  static void makeSpecial(odb::dbNet* net);
```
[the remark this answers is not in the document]
gadfort: No

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+{
+  auto* chip = db_->getChip();
+  if (chip == nullptr) {
+    return nullptr;
+  }
+
+  return chip->getBlock();
+}
+
+void ICeWall::makeBumpArray(odb::dbMaster* master,
```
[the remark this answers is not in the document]
gadfort: I think so, otherwise the functions could be removing all kinds of other things

src/pad/src/ICeWall.cpp
gadfort: it could be as simple as checking the row type before using it for the core area, but I think PDNgen and DPL will likely need to be updated to handle this correctly.

src/pad/src/ICeWall.cpp
gadfort: For the other sides, the MY is concatenated with the row rotation, so a R90 row will end up with a MX rotation.

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+                                 outer_io.yMin(),
+                                 outer_io.xMax() - cwidth,
+                                 outer_io.yMax() - cheight);
+
+  const odb::dbTransform xform(rotation);
+
+  // Create corners
+  const int corner_sites
+      = std::max(horizontal_site->getHeight(), corner_site->getWidth())
+        / corner_site->getWidth();
```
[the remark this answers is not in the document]
gadfort: Its technically possible if there is no corner site (ultimately only one instance is inserted, but I need it to have the right boundaries)

src/pad/src/ICeWall.cpp
gadfort: for the horizontal sites, the height of the site is how deep the row will be.

src/pad/src/ICeWall.cpp
gadfort: I added a check, but this ideally shouldn't happen since the pads a snapped to the site indexes.

src/pad/src/RDLRouter.h
gadfort: I just found it easier to have both than to compute the center every time I needed the point.

src/pad/src/ICeWall.cpp
gadfort: For the COVER cells, I think so (atleast I have yet to encounter that not being the case)

src/pad/src/ICeWall.cpp
gadfort: I haven't noticed a runtime issue yet, but I can probably look to see if there is a way to make it cleaner.

src/pad/src/ICeWall.cpp
gadfort: For the fake sites it's really just there to be able to make the rows, I could add the symmetry if needed.

src/pad/src/ICeWall.cpp
gadfort: If you have suggestions, I'm open. The reason for the string check was it made things easy to read and works.

src/pad/src/ICeWall.cpp
gadfort: okay

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+                     die.yMax() - north_offset);
+
+  const int cheight = corner_site->getHeight();
+  const int cwidth
+      = std::max(vertical_site->getHeight(), corner_site->getWidth());
+
+  const int x_sites = std::ceil(static_cast<double>(outer_io.dx() - 2 * cwidth)
+                                / vertical_site->getWidth());
+  outer_io.set_xhi(outer_io.xMin() + 2 * cwidth
+                   + x_sites * vertical_site->getWidth());
```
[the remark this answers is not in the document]
gadfort: cwidth is a standin for the fact that the corner IO site may just be the normal IO site. I fixed the ceil, you are correct that it would grow out of bounds.

src/pad/src/ICeWall.cpp
```diff
@@ -0,0 +1,1070 @@
...
+  const int y_sites = std::ceil(static_cast<double>(outer_io.dy() - 2 * cheight)
+                                / horizontal_site->getWidth());
+  outer_io.set_yhi(outer_io.yMin() + 2 * cheight
+                   + y_sites * horizontal_site->getWidth());
+
+  const odb::Rect corner_origins(outer_io.xMin(),
+                                 outer_io.yMin(),
+                                 outer_io.xMax() - cwidth,
```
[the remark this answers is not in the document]
gadfort: same

src/pad/src/ICeWall.cpp
gadfort: @maliberty I can go through the code to find the places `getRows()` is used and either (create a `getCoreRows()` function which preserves the current functionality and use that instead, or add the check for the site class type in the places where it's being used)? Which do you think makes sense?

src/pad/src/RDLRouter.h
gadfort: I could also add a `.center()` to the `odb::Rect` which would do the same thing?

src/pad/README.md
```diff
@@ -1,21 +1,282 @@
...
+          [-allow45] \
+          nets
+```
+
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-layer` | Layer to route on |
+| `-via` | Via to use to access terminals on routed net |
```
rovinski, line 272: Not clear to me if this is supposed to be for via up to bump or via down to IO pad (or both?). Some PDKs don't have the IO cell PAD pin on the RDL layer.

src/pad/README.md
```diff
@@ -1,21 +1,282 @@
...
+place_io_fill -row row_name \
+              masters
+```
+
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-row` | Name of the row to place the pad into, examples include: IO_NORTH, IO_SOUTH, IO_WEST, IO_EAST, IO_NORTH_0, IO_NORTH_1 |
+| `masters` | Names of the masters for the IO filler cells. |
```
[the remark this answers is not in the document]
gadfort: I can certainly add something to allow overlaps. It's usually not all filler cells that are overlapable, so it might need to be two different calls (one without allowing overlaps, but allowing gaps) followed by a second call with the overlapping ones? or an additional flag to indicate which cells allow for overlaps? (I'm partial to the second option), which do you think makes sense?

src/pad/README.md
```diff
@@ -1,21 +1,282 @@
...
+
+### Placing Wirebond Pads
+
+To place the wirebond pads over the IO cells
+```
+place_bondpad -bond master
+              [-offset {x y}] \
+              [-rotation rotation] \
+              io_instances
+```
```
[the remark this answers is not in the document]
gadfort: Wirebond pads cannot be placed before the io cells are placed. I assume you mean the `*bump_array` commands, which I decided to put below since not all designs will those and this keeps the cover cell handling together in the documentation

src/pad/README.md
```diff
@@ -1,21 +1,282 @@
...
+assign_io_bump -net DVDD BUMP_7_3
+assign_io_bump -net p_ddr_addr_7_o BUMP_7_1
+assign_io_bump -net p_ddr_addr_6_o BUMP_7_0
+```
+
+### RDL Routing
+
+To route the RDL for the bump arrays
+```
+route_rdl -layer layer \
```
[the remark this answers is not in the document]
gadfort: That makes sense, I'll rename it

src/pad/README.md
gadfort: It's for either end, I could provide a separate flag to specify vias for the bump or IO pad, incase you you are not routing on the layer with either side. I'll split it into two.

src/pad/src/ICeWall.cpp
maliberty: What do you mean by "row type"?  Do you mean the class of the site used by the row?  I think that would be fine to eliminate class=PAD from the core area calculation.

maliberty: You'll also need to make sure gpl, dpl, and dpo understand to ignore PAD rows.

src/pad/src/ICeWall.cpp
maliberty: Please add a warning if this is not true

src/pad/src/RDLRouter.h
maliberty: A comment would help clarify what is going on (perhaps center_point for a name)?

maliberty:
I think this is nearly ready.  I'm sure you'll be testing it more now that you'll rely on it.

src/pad/src/RDLRouter.cpp
```diff
@@ -0,0 +1,1285 @@
...
+template <class Vertex>
+class RDLRouterGoalVisitor : public boost::default_astar_visitor
+{
+ public:
+  RDLRouterGoalVisitor(RDLRouter::grid_vertex goal) : goal_(goal) {}
+  template <class Graph>
+  void examine_vertex(Vertex u, Graph& g)
+  {
+    if (u == goal_) {
+      throw RDLRouterGoalFound();
```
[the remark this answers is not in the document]
gadfort: Yeah, it's straight from their examples. and yes, seems odd.

gadfort:
> I think this is nearly ready. I'm sure you'll be testing it more now that you'll rely on it.

yeah, I'm sure as more designs are converted, I'll be finding the edge cases then, but that should be easy to debug and add test cases for in followon PRs.

gadfort:
@maliberty I searched for "getRows()" in the tools and made the changes needed to ignore the PAD rows

src/ifp/src/InitFloorplan.cc
```diff
@@ -172,9 +172,13 @@ void InitFloorplan::initFloorplan(const odb::Rect& die,
...
       // Destroy any existing rows.
       auto rows = block_->getRows();
-      for (dbSet<dbRow>::iterator row_itr = rows.begin(); row_itr != rows.end();
-           row_itr = dbRow::destroy(row_itr))
-        ;
+      for (dbSet<dbRow>::iterator row_itr = rows.begin();
+           row_itr != rows.end();) {
+        if (site != row_itr->getSite()) {
+          row_itr++;
+        }
```
maliberty, line 179: Shouldn't there be a continue here after row_iter++?  This will just delete whatever the next row is regardless of type (or fail if row_iter was just before end()).

gadfort: yes, thanks for catching that

maliberty (approved):

PR #2914, opened 2023-02-18
Removing using namespace std;

QuantamHD:
Remove all usages of using namespace std; in the entire codebase. :)


QuantamHD:
@maliberty @vvbandeira The clang format seems to be off. I can't replicate it on my side.

QuantamHD:
nvm I just didn't push correctly

src/dpl/include/dpl/Graphics.h
```diff
@@ -0,0 +1,72 @@
...
+#include "gui/gui.h"
+#include "odb/db.h"
+
+namespace dpl {
+
+class Opendp;
+struct Cell;
+
+// Decorates the gui::Renderer with DPL specific routines.
+class Graphics : public gui::Renderer
```
maliberty, line 39: Why has this been moved to a public header?

src/drt/src/dr/FlexDR.cpp
```diff
@@ -502,21 +506,22 @@ void FlexDR::searchRepair(const SearchRepairArgs& args)
...
 
-  vector<unique_ptr<FlexDRWorker>> uworkers;
+  std::vector<std::unique_ptr<FlexDRWorker>> uworkers;
   int batchStepX, batchStepY;
 
   getBatchInfo(batchStepX, batchStepY);
 
-  vector<vector<vector<unique_ptr<FlexDRWorker>>>> workers(batchStepX
-                                                           * batchStepY);
+  std::vector<std::vector<std::vector<std::unique_ptr<FlexDRWorker>>>> workers(
```
maliberty, line 514: I think a using std::vector in this cpp would help readability   [which isn't great to start with]

maliberty:
With a broad change like this it is best to coordinate with other developers.  I imagine this may give @osamahammad21  many merge conflicts.

QuantamHD:
You can merge his first. I don't mind rebasing.

maliberty:
@osamahammad21 what PRs do you have remaining that should be merged first?

src/dpl/include/dpl/Graphics.h
QuantamHD: Fixed. Left over from different branch.

src/drt/src/dr/FlexDR.cpp
QuantamHD: I'm going to say I agree, but that feels ill suited to this PR.

This particular case is a good example of the potential for making a large PR larger. The first std::vector in this type is being treated as a virtual 2D array. It's not clear to me that just adding a using declaration is enough to fix the readability issues. I think it's best to keep this PR focused on removing `using namespace std;`, and try evolving readability separately

QuantamHD:
@maliberty @osamahammad21 Can you give me an O(5 day) merge window. One of the `using namespace std` does cause a compilation error on GCC 12.

osamahammad21:
Merging this PR now will cause a lot of trouble as I have about 6 open PRs in the router. I would prefer if we delay it a week or two.

QuantamHD:
@osamahammad21 Is it looking like we can merge this soon?

maliberty:
Do you want to fix all the conflicts or abandon this?

QuantamHD:
Happy to fix the conflicts if you guarantee a merge.

maliberty:
> Happy to fix the conflicts if you guarantee a merge.

Ok, its as a good a time as any.

maliberty (approved):

PR #2918, opened 2023-02-19
drt/ta: Some simplifications

antonblanchard:


src/drt/src/ta/FlexTA.h
```diff
@@ -34,6 +34,7 @@
 #include "db/obj/frVia.h"
 #include "db/taObj/taPin.h"
 #include "frDesign.h"
+#include "frRTree.h"
```
maliberty, line 37: The goal of the impl pattern was to avoid including the rtree in the public header.  The boost headers are huge and slow down compilation.  If we move to precompiled headers this could make more sense. 

@QuantamHD thoughts?

antonblanchard: I figured there was a reason, but wasn't sure what exactly it was. Happy to drop this part.

maliberty:
stale

PR #2921, opened 2023-02-20
odb: log unsupported LEF properties

antonblanchard:
Don't silently ignore unsupported LEF properties.

### DIFF
--- a/src/psm/test/aes_asap7_vdd.ok
+++ b/src/psm/test/aes_asap7_vdd.ok
@@ -3,6 +3,10 @@
 The LEF parser will ignore this statement.
 To avoid this warning in the future, remove this statement from the LEF file with version 5.6 or later. See file asap7_data/asap7.lef at line 2.
 
+[INFO ODB-0388] unsupported LEF58_PITCH property for layer M2 :"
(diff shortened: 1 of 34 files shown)


maliberty (approved):

maliberty:
You'll have a lot of tests to update

src/psm/test/aes_asap7_vdd.ok
```diff
@@ -3,6 +3,10 @@
 The LEF parser will ignore this statement.
 To avoid this warning in the future, remove this statement from the LEF file with version 5.6 or later. See file asap7_data/asap7.lef at line 2.
 
+[INFO ODB-0388] unsupported LEF58_PITCH property for layer M2 :"
```
gadfort, line 6: @maliberty  is this a parameter that's something we should support? It seems like it handles the asap7 tracks

maliberty: Yes we should.

maliberty: https://github.com/The-OpenROAD-Project/OpenROAD/issues/2926

maliberty (approved):

PR #2933, opened 2023-02-23
par: Removes unused files

QuantamHD:
Also fixes a compile bug on GCC12 with using namespace std; causing an overlap with boost::point_xy


maliberty:
@ravi-varadarajan @ZhiangWang033 any concern about removing the autocluster code?  I think that is integrated into mpl2 now but want to confirm you have no plans for it.

ravi-varadarajan:
It can be removed from par

maliberty (approved):

Issue #2924, opened 2023-02-20
Routing congestion heat-map shows bogus heat congestion at top in a horizontal strip


oharboe:
### Describe the bug

Run:

```
make DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```

Look at routing congestion heat-map. There's bogus congestion at the top in a horizontal strip:

![image](https://user-images.githubusercontent.com/2798822/220089038-b0cb39b6-f40d-4ca1-95ac-01dba306bb34.png)

![image](https://user-images.githubusercontent.com/2798822/220089199-be043b5b-e3eb-425a-ae7d-8e0bf7f7931e.png)

![image](https://user-images.githubusercontent.com/2798822/220089296-163e22cf-31f5-416b-b153-3d3f5417fc98.png)

### Expected Behavior

There should only be routing congestion shown in the heat map if there is actually routing congestion.

### Environment

```shell
Git commit: 1af9a6cdefb5c627bb5ebd73148d7f217fc6d308
kernel: Linux 5.15.0-60-generic
os: Ubuntu 22.04.1 LTS (Jammy Jellyfish)
cmake version 3.24.2
-- The CXX compiler identification is GNU 11.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-6917-g1af9a6cde
-- System name: Linux
-- Compiler: GNU 11.3.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
[...]

PR #2937, opened 2023-02-23
grt: fix averageTrackPattern function

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/2924


maliberty:
This should only affect asap7 but it would be good to test with a test* branch in ORFS (like secure* but skips the private designs).

maliberty (approved):
@eder-matheus merge if the test pass at your discretion

PR #2938, opened 2023-02-24
Bump OpenSTA Version

QuantamHD:
@jjcherry56 @maliberty I have a couple of issues with this PR that I could use some help with. I got the app building, but many of the tests fail either with correctness or segfaulting issues. I need some help to fix, or need someone to take this as a base, and fix the remaining issues.

1. The dbNetwork creates an invalid top_instance_ that causes segfaults in some of the tests

image

```c++
// dbNetwork.cc

dbNetwork::dbNetwork() :
  ConcreteNetwork(),
  db_(nullptr),
  logger_(nullptr),
  block_(nullptr),
  top_instance_(reinterpret_cast<Instance*>(1)),
  top_cell_(nullptr)
{
}
```

Seems like this is caused by dbNetwork returning this 0x1 cast Instance * that gets cast to a ConcreteInstance in the visitMatches function. Any thoughts on this @jjcherry56?

```c++
NetSeq
SdcNetwork::findNetsMatching(const Instance *parent,
			     const PatternMatch *pattern) const
{
  NetSeq matches;
  visitMatches(parent, pattern,
	       [&](const Instance *instance,
		   const PatternMatch *tail)
	       {
		 size_t match_count = matches.size();
		 network_->findInstNetsMatching(instance, tail, matches);
		 return matches.size() != match_count;
	       });
  return matches;
}
```

2.  @jjcherry56 There's a number of areas where PinSeq is integrated and then the pin is deleted. These require `const_cast<T*>`'s is this how you intended users to use this API? If not what's the right approach?

image

### DIFF
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -424,6 +424,14 @@ dbNetwork::metersToDbu(double dist) const
 
 ////////////////////////////////////////////////////////////////
 
+ObjectId dbNetwork::id(const Instance* instance) const
+{
+  if (instance == top_instance_) {
+    return 0;
+  }
+  return staToDb(instance)->getId();
+}
+
 const char*
 dbNetwork::name(const Instance* instance) const
 {
@@ -604,6 +612,18 @@ dbNetwork::netIterator(const Instance* instance) const
 
 ////////////////////////////////////////////////////////////////
 
+ObjectId dbNetwork::id(const Pin* pin) const
+{
+  dbITerm* iterm;
+  dbBTerm* bterm;
+  staToDb(pin, iterm, bterm);
+
+  if (iterm) {
+    return iterm->getId() + iterm->getBlock()->getBTerms().size();
+  }
+  return bterm->getId();
+}
+
 Instance*
 dbNetwork::instance(const Pin* pin) const
 {
@@ -777,6 +797,11 @@ dbNetwork::isPlaced(const Pin* pin) const
 
 ////////////////////////////////////////////////////////////////
 
+ObjectId dbNetwork::id(const Net* net) const
+{
+  return staToDb(net)->getId();
+}
+
 const char*
 dbNetwork::name(const Net* net) const
 {
@@ -839,6 +864,11 @@ const Net* dbNetwork::highestConnectedNet(Net* net) const
 }
 
 ////////////////////////////////////////////////////////////////
+
+ObjectId dbNetwork::id(const Term* term) const
+{
+  return staToDb(term)->getId();
+}
 
 Pin*
 dbNetwork::pin(const Term* term) const
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -835,7 +835,7 @@ dbNetwork::visitConnectedPins(const Net* net,
   }
 }
 
-const Net*
+const Net* 
 dbNetwork::highestConnectedNet(Net* net) const
 {
   return net;
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -422,6 +422,15 @@ dbNetwork::metersToDbu(double dist) const
   return dist * dbu * 1e+6;
 }
 
+ObjectId dbNetwork::id(const Port *port) const
+{
+  if (!port) {
+    // should not match anything else
+    return std::numeric_limits<ObjectId>::max();
+  }
+  return ConcreteNetwork::id(port);
+}
+
 ////////////////////////////////////////////////////////////////
 
 ObjectId dbNetwork::id(const Instance* instance) const
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -823,7 +823,7 @@ dbNetwork::termIterator(const Net* net) const
 void
 dbNetwork::visitConnectedPins(const Net* net,
                               PinVisitor& visitor,
-                              ConstNetSet& visited_nets) const
+                              NetSet& visited_nets) const
 {
   dbNet* db_net = staToDb(net);
   for (dbITerm* iterm : db_net->getITerms()) {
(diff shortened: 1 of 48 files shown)


jjcherry56:
opendb netlists are not hierarchical so there is no top instance. top_instance_ is a unique value that is not a real instance (there is no class for sta::Instance), which is the intended design. I think the problem is that you didn't update dbNetwork::findInstNetsMatching and findInstancesHierMatching to match the expected signature. diff the Network.hh to see how dbNetwork has to be updated.

QuantamHD:
Thanks for the tips @jjcherry56 😄

QuantamHD:
@jjcherry56 any suggestions around the `const_cast`s? I'm not totally clear if that's how you intended the APIs to be used. Is there another pattern that we should follow to get a non const Pin* to call disconnect Pin?

image

jjcherry56:
Not really. Sometimes const just gets in the way because it is so contagious.

QuantamHD:
@jjcherry56 In one of the comments you mentioned that the 0x00...1 pointer didn't need to change. I'm not so sure.

With the updated changes in this PR I'm seeing a segfault in dbSta/test/read_verilog9.tcl related to trying to insert the 0x000..1 instance into a set. 

image

Which eventually tries to call id() which was a new addition in OpenSTA, which breaks. This seems like the instance actually needs to be valid in order to be placed into a set. 

image

Would you agree with my assessment?

jjcherry56:
I don't think you appreciate the implications of polluting the database with a fictitious instance. Are you going to change every reference in OR to instances to ignore it? Are you going to insert it every time the database is read (like your previous attempt did)? Are you going to add bterms to correspond to a top level instance as in a hierarchical netlist? Are you going to create a master that corresponds to the design that the instance references? The db netlist is not hierarchical. It is a square peg that has to be inserted into a round hole.

The change to fix the problem you are observing is trivial compared to what you are proposing: only leaf instances have functions that can propagate simulation values so this code should not even be running on the top level instance. And so you have to ask "why is deletePinBefore being called on a pin on the top level instance (ie, a bterm in odb speak)?

jjcherry56:
BTW, I will not approve this PR as long as it has hundreds of unrelated changes (ie, const and reformatting of code).

QuantamHD:
@jjcherry56 The const changes are required for compiling with the new OpenSTA version. I've removed reformatting from Resizer.cc

Everything else is a functional rather than cosmetic change.

src/dbSta/src/dbSdcNetwork.hh
```diff
@@ -43,26 +43,22 @@ class dbSdcNetwork : public SdcNetwork
...
-  virtual void findPinsMatching(const Instance* instance,
-                                const PatternMatch* pattern,
-                                PinSeq* pins) const;
+  Instance* findInstance(const char* path_name) const override;
+  InstanceSeq findInstancesMatching(const Instance* contex,
+                                            const PatternMatch* pattern) const override;
+  virtual NetSeq findNetsMatching(const Instance*,
+                                  const PatternMatch* pattern) const override;
+  virtual PinSeq findPinsMatching(const Instance* instance,
+                                  const PatternMatch* pattern) const override;
```
maliberty, line 52: virtual is superfluous when specifying overrride.

src/gpl/src/nesterovPlace.cpp
```diff
@@ -48,6 +48,7 @@
 #include "utl/Logger.h"
 
 namespace gpl {
+using namespace std;
 
 using namespace std;
```
maliberty, line 53: Not needed twice

maliberty:
There are new id() APIs that need to be implemented/overridden in dbNetwork.

maliberty:
```
  ObjectId id(const Cell *cell) const override;
  ObjectId id(const Port *port) const override;
  ObjectId id(const Instance *instance) const override;
  ObjectId id(const Pin *pin) const override;
  ObjectId id(const Term *term) const override;
  ObjectId id(const Net *net) const override;
```

```
ObjectId dbNetwork::id(const Cell* cell) const
{
  return staToDb(cell)->getId();
}

ObjectId dbNetwork::id(const Port* port) const
{
  return staToDb(port)->getId();
}

ObjectId dbNetwork::id(const Instance* instance) const
{
  if (instance == top_instance_) {
    return 0;
  }
  return staToDb(instance)->getId();
}

ObjectId dbNetwork::id(const Pin* pin) const
{
  dbITerm* iterm;
  dbBTerm* bterm;
  staToDb(pin, iterm, bterm);

  if (iterm) {
    return iterm->getId() + iterm->getBlock()->getBTerms().size();
  }
  return bterm->getId();
}

ObjectId dbNetwork::id(const Term* term) const
{
  return staToDb(term)->getId();
}

ObjectId dbNetwork::id(const Net* net) const
{
  return staToDb(net)->getId();
}
```

maliberty:
The changes to CellSet, PortSet, InstanceSet, PinSet, and NetSet are also API breaking.  Now if you default construct them you get a useless set that can't have anything added to it because the network is null.  You have to use the ctor that takes a network argument.  This means rsz is broken and likely anything else using these API.  You also can't update the set after creation so we will have to delay construction.

For example:
```
  InstanceSeq inserted_buffers_;
  InstanceSet inserted_buffer_set_;
```

QuantamHD:
I fixed one of the sets in rsz. I don't think we're using instance set in that many places.

src/dbSta/src/dbNetwork.cc
```diff
@@ -604,6 +622,18 @@ dbNetwork::netIterator(const Instance* instance) const
...
 ////////////////////////////////////////////////////////////////
 
+ObjectId dbNetwork::id(const Pin* pin) const
+{
+  dbITerm* iterm;
+  dbBTerm* bterm;
+  staToDb(pin, iterm, bterm);
+
+  if (iterm) {
+    return iterm->getId() + iterm->getBlock()->getBTerms().size();
```
gadfort, line 632: why add the bterm size?

maliberty: A sta pin maps to either a bterm or an iterm.  To give a unique id I can't just use the odb id for each as they would overlap.  I offset the iterms ids to make them unique.  I could have done the reverse and offset the bterms by the #iterms.

maliberty (approved):

PR #2942, opened 2023-02-25
rsz: Fix int overflow SIGILL

QuantamHD:
The result of `Resizer::findMaxWireLength1()` will eventually be
passed to this function, and can return `(double)sta::INF`. Which
when converted to an int in `metersToDbu` blows up.

```c++
double Resizer::findMaxWireLength1()
{
  double max_length = INF;
  for (const Corner* corner : *sta_->corners()) {
    if (wireSignalResistance(corner) > 0.0) {
      for (LibertyCell* buffer_cell : buffer_cells_) {
        double buffer_length = findMaxWireLength(buffer_cell, corner);
        max_length = min(max_length, buffer_length);
      }
    }
  }
  return max_length;
}
```

I added the code in this PR to protect against this.

### DIFF
--- a/src/rsz/src/Resizer.cc
+++ b/src/rsz/src/Resizer.cc
@@ -37,33 +37,32 @@
 
 #include <cmath>
 #include <limits>
+#include <optional>
 
 #include "BufferedNet.hh"
 #include "RepairDesign.hh"
-#include "RepairSetup.hh"
 #include "RepairHold.hh"
-
-#include "utl/Logger.h"
+#include "RepairSetup.hh"
 #include "db_sta/dbNetwork.hh"
-
-#include "sta/FuncExpr.hh"
-#include "sta/PortDirection.hh"
-#include "sta/Units.hh"
-#include "sta/Liberty.hh"
-#include "sta/TimingArc.hh"
-#include "sta/TimingModel.hh"
-#include "sta/Network.hh"
-#include "sta/Graph.hh"
 #include "sta/ArcDelayCalc.hh"
-#include "sta/GraphDelayCalc.hh"
-#include "sta/Parasitics.hh"
-#include "sta/Sdc.hh"
-#include "sta/InputDrive.hh"
-#include "sta/Corner.hh"
 #include "sta/Bfs.hh"
+#include "sta/Corner.hh"
+#include "sta/FuncExpr.hh"
+#include "sta/Fuzzy.hh"
+#include "sta/Graph.hh"
+#include "sta/GraphDelayCalc.hh"
+#include "sta/InputDrive.hh"
+#include "sta/Liberty.hh"
+#include "sta/Network.hh"
+#include "sta/Parasitics.hh"
+#include "sta/PortDirection.hh"
+#include "sta/Sdc.hh"
 #include "sta/Search.hh"
 #include "sta/StaMain.hh"
-#include "sta/Fuzzy.hh"
+#include "sta/TimingArc.hh"
+#include "sta/TimingModel.hh"
+#include "sta/Units.hh"
+#include "utl/Logger.h"
 
 // http://vlsicad.eecs.umich.edu/BK/Slots/cache/dropzone.tamu.edu/~zhuoli/GSRC/fast_buffer_insertion.html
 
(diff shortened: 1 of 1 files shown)


QuantamHD:
@maliberty Base commit is formatting 

Important changes are here https://github.com/The-OpenROAD-Project/OpenROAD/pull/2942/commits/07f45507898cc5c6f146cbec6a3c52c451b15381

maliberty:
I think you are addressing the problem at to low a level.  Either you have no resistance or no buffers both of which should be caught earlier and are more actionable by the user than this error.

QuantamHD:
Agreed, but I'll caveat that I do think we should still add this change. casting a double to an int is an inherently unsafe thing to do.

Should we error instead of returning `sta::INF`? @maliberty

maliberty:
> Agreed, but I'll caveat that I do think we should still add this change. casting a double to an int is an inherently unsafe thing to do.
> 
> Should we error instead of returning `sta::INF`? @maliberty

I think so as the function has failed.  Also you can say something more useful at that point in the code about the problem.

QuantamHD:
@maliberty Separate, but related question. 

I haven't dug into it, but I'm assuming this resistance value either comes from the LEF or the set_rc commands, but let's assume it does for the sake of the question.

Why do we rely on these values instead of going to RCX to estimate RC values from nets. Wouldn't it have a more accurate estimation?

maliberty:
Resistance is pretty simple so there isn't much benefit to using rcx.  If you haven't done routing it is even less useful as you don't have real layers.

src/rsz/src/Resizer.cc
```diff
@@ -1165,7 +1168,15 @@ Resizer::dbuToMeters(int dist) const
 int
 Resizer::metersToDbu(double dist) const
 {
-  return dist * dbu_ * 1e+6;
+  if (dist < 0) {
+    logger_->error(
+        RSZ, 86, "metersToDbu({}) cannot convert negative distances", dist);
+  }
+  // sta::INF is passed to this function in some cases. Protect against
+  // overflow conditions.
```
maliberty, line 1176: sta::INF is a 1E+30F so this check won't help.  In general I would put the comment before the check rather than after.

src/rsz/src/Resizer.cc
```diff
@@ -1901,16 +1912,33 @@ Resizer::findMaxWireLength()
...
-      }
+    if (wireSignalResistance(corner) <= 0.0) {
+      logger_->warn(RSZ,
+                    88,
+                    "Corner: {} has no wire signal resistance value.",
+                    corner->name());
+      continue;
     }
+
+    // buffer_cells_ is asserted to be non-empty else where in resizer.
```
maliberty, line 1925: I would say required or verified rather than asserted as that sounds like an assert() was used (which it isn't).

src/rsz/src/Resizer.cc
QuantamHD: The comment is unrelated to the check. The check at the top is to prevent people from putting in negative distances.

Whereas the `& INT_MAX` maxes sure that any values are correctly saturated to int max. Which is what the comment is referring to.

src/rsz/src/Resizer.cc
QuantamHD: Done.

src/rsz/src/Resizer.cc
```diff
@@ -1165,7 +1168,15 @@ Resizer::dbuToMeters(int dist) const
...
-  return dist * dbu_ * 1e+6;
+  if (dist < 0) {
+    logger_->error(
+        RSZ, 86, "metersToDbu({}) cannot convert negative distances", dist);
+  }
+  // sta::INF is passed to this function in some cases. Protect against
+  // overflow conditions.
+  double distance = dist * dbu_ * 1e+6;
+  return static_cast<int>(std::lround(distance)
+                          & std::numeric_limits<int>::max());
```
maliberty, line 1179: Nothing about unit conversion prevents a negative value from being converted.  Only underflow / overflow is an issue.  I think std::clamp 'ing to the integer min()/max() would be more correct and avoid the bit trickery.

QuantamHD: This function accepts a distance which conceptually shouldn't be negative.

I would rather have callers call abs on their code if it is. I'll change to std::clamp

maliberty: The argument name suggests distance but the actual operation is just metersToDBU which can operate on coordinates as well as distances.  I think the argument name should be changed.  If the incoming values shouldn't be negative that's on the caller.

maliberty (approved):

PR #2949, opened 2023-02-27
Adds clang-tidy to review workflow

QuantamHD:
This commit will add clang tidy comments to PRs.
See https://github.com/marketplace/actions/clang-tidy-review for more examples.

Once adopted we should move the action into the OpenROAD Project.


QuantamHD:
@maliberty This won't post any comments until it's merged in due to permission issues related to forks.

QuantamHD:
Importantly this clang-tidy only makes comments on modified lines not on the files in general

QuantamHD:
Example of the output once merged on Openroad.

@cdleary 

https://github.com/QuantamHD/OpenROAD/pull/2

maliberty:
Most of our code does not comply with the google naming style. This is going to produce tons of warnings on every change. I wouldn't want people to mix-and-match coding styles. We would have to convert modules and then enable it for that module. I don't think we want to just turn this on as is without it being part of a larger plan. Its a substantial effort and disruptive as well.

QuantamHD:
How about this we split the check into two actions one for all the static analyzers that gets run over the whole project, and one that runs style checking over all modules except sub folders we opt out of?

However, this is one of those workflow things that's important to us so we need to make a plan to enable style over the whole project.

.github/workflows/github-actions-clang-tidy.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+      - name: ClangTidy Gen
+        uses: QuantamHD/clang-tidy-review@v0.10.2
+        with:
+          build_dir: './build'
+          cmake_command: cmake . -B build '-DCMAKE_EXPORT_COMPILE_COMMANDS=on'
```
maliberty, line 17: We already have set(CMAKE_EXPORT_COMPILE_COMMANDS 1) in the top level CMakeLists.txt

maliberty:
I agree it is good to have.  Splitting it is fine but I don't see any static analyzers being run here.  Is that something you plan to add?  I assume that would be more of a nightly thing and not per PR.

I don't think we will ever convert abc or sta but the rest of the project is possible (rsz & dbSta are debatable).  However the exclude list right now would be nearly everything.

Does Google have resources to put behind making this conversion happen?

QuantamHD:
The following analyzers are running `-*,performance-*,readability-*,bugprone-*,clang-analyzer-*,cppcoreguidelines-*,mpi-*,misc-*`

This is definitely a PR thing. Eventually we want to make it a blocking thing. 

See uninit'd variable 
https://github.com/QuantamHD/OpenROAD/pull/2#discussion_r1118425540

maliberty:
This PR, in this form, can't be approved now.  We should work out a plan for this.

@vvbandeira FYI

QuantamHD:
@maliberty This should be good to go given what we discussed. I plan on adding the identifier naming one in a different PR.

maliberty:
Is it expected that this is failing on this PR 
image
 ?

@vvbandeira any comments?  We should put this in OpenROAD/actions rather than QuantamHD/clang-tidy-review

QuantamHD:
Yes. It's failing because it's running as a forked workflow which doesn't have permission at the moment. It'll work once merged.

QuantamHD:
> @vvbandeira any comments? We should put this in OpenROAD/actions rather than QuantamHD/clang-tidy-review

This is the repo in question. I assumed we just keep the repo together for simplicity rather than trying to merge it into a larger one.

https://github.com/QuantamHD/clang-tidy-review

vvbandeira:
I have forked the original repo at https://github.com/The-OpenROAD-Project/clang-tidy-review
@QuantamHD can you update the action target to be our fork? This should be enough to fix the permission problem, right?

.clang-tidy
```diff
@@ -1,27 +1,64 @@
 Checks: >
   -*,
-  readability-identifier-naming
+  performance-*,
+  readability-*,
```
hzeller, line 4: I would exclude
`-readability-braces-around-statements`
, because it will suggest to put braces into one-liners

```c++
 if (foo) return bar;
```

The `google-readability-braces-around-statements` is more useful. It allows the above, but will complain when things take more than a line
```c++
if (foo)
   return bar;
```

It emphasizes reducing vertical whitespace if possible.

Another example is `-readability-implicit-bool-conversion`. It complains when pointers are used in boolean expressions, but that is a very common and very readable technique. So

```c++
 if (thing) thing->foo();
```

... would need to become

```c++
if (thing != nullptr) thing->foo();
```

to satisfy the rule. Which is hiding intent and reduces readability.

.. there are a few others.

hzeller (commented):
Some clang-tidy warnings are a bit overzealous and might trigger on actually desirable code constructs or are sometimes confused.

I am using a fairly aggressive rules in Verible, but have filtered out some that felt not right. Might be a good starting point.

https://github.com/chipsalliance/verible/blob/master/.clang-tidy

.clang-tidy
maliberty: I agree about -readability-implicit-bool-conversion.  I prefer readability-braces-around-statements.  It avoids the classic
```
if (cond) 
   a;
   b;
```
mistake.  I know others will disagree.

hzeller: both, `readability-braces-around-statements` and `google-readability-braces-around-statements` will complain about the example you mention and suggest to 

```c++
  if (cond) {
     a;
  }
  b;
```

The difference is, that `google-readability-braces-around-statements` will allow, if and only if, the result is in the same line, to not put braces around it. Because if there are braces around it, it will then have the formatter kick in and format it in three lines.

But for readability, typical one-liners (typically used in quick-out  `return`, `break` or `continue`) as in my example will not require braces.

QuantamHD:
@maliberty This should be ready to go.

maliberty (approved):

maliberty:
lgtm @vvbandeira any final concerns?  If not go ahead and merge

vvbandeira (approved):

QuantamHD:
@maliberty can you merge now that @vvbandeira has approved

PR #2951, opened 2023-02-27
Fix RMP timing mode

osamahammad21:
- Write ABC log to the dedicated log files as they are needed for parsing the resulting delay values later
- Add print_stats cmd to the abc script
- Use the best abc result for timing instead of DELAY_4 only
- Add a test case for timing mode: worst slack reduced from 1.29 to 1.18
- Update the ok files


osamahammad21:
@maliberty The part for redirecting stdout to abc log files may seem ugly, but I don't know any other way to redirect the abc log to the required files. If you know any abc api that allows specifying the log file, it would be very helpful.

maliberty:
> @maliberty The part for redirecting stdout to abc log files may seem ugly, but I don't know any other way to redirect the abc log to the required files. If you know any abc api that allows specifying the log file, it would be very helpful.

I don't see anything in abc to control where the output goes.  Originally we forked abc which avoided this issue.  When abc was integrated I believe the thinking was that we would fork OR and in that process we could call abc + gpl to better evaluate the results.  However right now it doesn't fork anything which makes the whole process serial.  If you fix this problem the output will become secondary as it will be in the log file of the forked run.  Better than forking would be to use dst so we aren't limited to one machine.

src/rmp/src/Restructure.cpp
```diff
@@ -267,7 +281,7 @@ void Restructure::runABC()
         } else {
           // Using only DELAY_4 for delay based gain since other modes not
           // showing good gains
-          if (modes[curr_mode_idx] == Mode::DELAY_4) {
+          if (delay < best_delay_gain) {
```
maliberty, line 284: remove the comment as it is no longer true

src/rmp/src/Restructure.cpp
```diff
@@ -323,7 +337,7 @@ void Restructure::getEndPoints(sta::PinSet& ends,
       if (expanded.size() / 2 > max_depth) {
         ends.insert(end_point->pin());
         // Use only one end point to limit blob size for timing
-        break;
+        // break;
```
maliberty, line 340: This may produce a huge blob for while abc will take a long time.  It would be better to iteratively fix cones.

src/rmp/src/Restructure.cpp
```diff
@@ -219,15 +220,28 @@ void Restructure::runABC()
...
       Abc_Frame_t* abc_frame = Abc_FrameGetGlobalFrame();
       const std::string command = "source " + abc_script_file;
+
+      fflush(stdout);
+      int stdout_fd = dup(STDOUT_FILENO);
+      const std::string abc_log_name = logfile_ + std::to_string(curr_mode_idx);
+      std::ofstream{abc_log_name.c_str()};
+      int redir_fd = open(abc_log_name.c_str(), O_WRONLY);
+      dup2(redir_fd, STDOUT_FILENO);
+      close(redir_fd);
```
maliberty, line 230: You could make a RAII class to handle the redirect/reset.  It would ensure that the output get reset in case of any exception (probably abc doesn't throw but who knows).

src/rmp/src/Restructure.cpp
```diff
@@ -230,21 +236,20 @@ void Restructure::runABC()
         child_proc.push_back(c_pid);
       }
     }
+    open_sta_->setThreadCount(ord::OpenRoad::openRoad()->getThreadCount());
   }
 
   std::string best_blif;
   int best_inst_count = std::numeric_limits<int>::max();
-  float best_delay_gain = std::numeric_limits<float>::max();
+  float best_delay_gain = -1 * std::numeric_limits<float>::max();
```
rovinski, line 244:
```suggestion
  float best_delay_gain = std::numeric_limits<float>::lowest();
```

maliberty:
@osamahammad21 Is this ready for merging or should it be a draft?

maliberty:
Are you still working on this?  Its grown a lot of conflicts.

PR #2954, opened 2023-02-28
Fixed build error that related to GUI & DB Network

vic4key:
Build on Ubuntu 22.04 64-bit & Clang 14


vic4key:
Wrong commit format

PR #2955, opened 2023-02-28
Fixed build error that related to GUI & DB Network.

vic4key:
Build on Ubuntu 22.04 64-bit & Clang 14.

### DIFF
--- a/src/dbSta/include/db_sta/dbNetwork.hh
+++ b/src/dbSta/include/db_sta/dbNetwork.hh
@@ -253,7 +253,7 @@ class dbNetwork : public ConcreteNetwork
   void findConstantNets();
   virtual void visitConnectedPins(const Net* net,
                                   PinVisitor& visitor,
-                                  ConstNetSet& visited_nets) const;
+                                  NetSet& visited_nets) const;
   bool portMsbFirst(const char* port_name);
 
   dbDatabase* db_;
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -823,7 +823,7 @@ dbNetwork::termIterator(const Net* net) const
 void
 dbNetwork::visitConnectedPins(const Net* net,
                               PinVisitor& visitor,
-                              ConstNetSet& visited_nets) const
+                              NetSet& visited_nets) const
 {
   dbNet* db_net = staToDb(net);
   for (dbITerm* iterm : db_net->getITerms()) {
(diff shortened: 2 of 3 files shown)


maliberty:
You need to fix the DCO (see the details link).

@QuantamHD is this what you were running into?

src/dbSta/include/db_sta/dbNetwork.hh
```diff
@@ -253,7 +253,7 @@ class dbNetwork : public ConcreteNetwork
   void findConstantNets();
   virtual void visitConnectedPins(const Net* net,
                                   PinVisitor& visitor,
-                                  ConstNetSet& visited_nets) const;
+                                  NetSet& visited_nets) const;
```
maliberty, line 256: This no longer overrides the base class method and therefore seems wrong.  What is the point of this change?

vic4key:
Hi @maliberty, I already checked the DCO and found nothing to fix. My PR is satisfied the following commit's requirements, but I don't know why it's still unacceptable.

1. Add your Signed-off-by.
2. You are the only author of the commits in this branch.
3. You are absolutely certain nobody else is doing any work based upon this branch.
4. There are no empty commits in the branch (for example, a DCO Remediation Commit which was added using --allow-empty).

src/dbSta/include/db_sta/dbNetwork.hh
vic4key: The ConstNetSet class is no longer exist, and seems it's renamed to NetSet. So, without these changes, build will be failed.

maliberty:
The easiest way is to use 'git commit -s' to sign your commits.  However that is secondary to the fact that I don't think this PR is correct.

maliberty:
Are you using the git submodule currently reference from OR or are you trying to use the head or sta?  The former is what is normally used and the code is aligned to.

vijayank88:
@vic4key 
Don't change anything with `master` on your fork.
Use `git checkout -b dbsta_update` do your update and follow 
`git add .`
`git commit -s -m 'your message'`
`git push --set-upstream origin dbsta_update'

src/dbSta/include/db_sta/dbNetwork.hh
gadfort: It seems like the STA version is bumped since the fixes are similar to #2938

vic4key: @gadfort Yes. It was fixed in this PR and mine. When these changes are updated to the master branch, my issue will be solved.

maliberty:
I believe this problem has been resolved in the latest sta update.  Please reopen if you still have an issue.  Thanks for pointing out the issue.

PR #2963, opened 2023-03-01
ppl: ensure pin shape width/height is an even multiple of the dbu's since they are divided by 2 causing off grid violations

gadfort:
Fixes:
- off grid violations in pins since skywater has a mfg grid of 5nm and a DBU of 1000, this leads to rounding errors when computing the pin shape by /2

### DIFF
--- a/src/ppl/src/IOPlacer.cpp
+++ b/src/ppl/src/IOPlacer.cpp
@@ -1073,10 +1073,20 @@ void IOPlacer::updatePinArea(IOPin& pin)
       pin_width
           = mfg_grid * std::ceil(static_cast<float>(pin_width) / mfg_grid);
     }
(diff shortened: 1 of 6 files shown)


maliberty (approved):
I've never seen a PDK like this where a valid point on the manufacturing grid can't be represented in dbu.  It seems like a PDK bug...

gadfort:
> I've never seen a PDK like this where a valid point on the manufacturing grid can't be represented in dbu. It seems like a PDK bug...

I'm not sure (the skywatre130 PDK has lots of specialness), but the issue is the / 2 (since the minimum resolution is 5 DBUs), this causes a height of 485 DBUs be become 484 after the / 2 and * 2

PR #2964, opened 2023-03-01
pdn: adjust pad edge connection checking to allow for pins that extend beyond the instance boundary as in sky130

gadfort:


maliberty (approved):
It seems more like an openlane thing than a skywater thing.  I've always considered it a oddity

gadfort:
> 

Yes, it definitely seems wrong, but PDN is at the mercy of their LEFs unfortunately

PR #2967, opened 2023-03-02
remove par and par_py until they work on apple silicon.

zarkin404:
par and pay_py will cause `ERROR: Something is wrong with flag 'flagfile' in file 'Users/corentinl/work/main/build_make/_deps/absl-src/absl/flags/parse.cc'. One possibility: file 'Users/corentinl/work/main/build_make/_deps/absl-src/absl/flags/parse.cc' is being linked both statically and dynamically into this executable. e.g. some files listed as srcs to a test and also listed as srcs of some shared lib deps of the same test.` on apple silicon when launches executable, so temporarily remove until or-tools fix this problem.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -319,7 +319,6 @@ target_link_libraries(openroad
   mpl
   psm
   ant
-  par
   upf
   utl
   pdn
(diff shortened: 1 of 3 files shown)


maliberty (approved):
@vvbandeira have you seen this issue ?

vvbandeira:
@maliberty yes, this is the same error that lead us to disable mpl2 on macOS. Other users have reported the same problem, including Cherry and Eder.

PR #2969, opened 2023-03-03
Ifp: LEF58_PITCH

arthurjolo:
Adds support to the property LEF58_PITCH


src/ifp/src/InitFloorplan.cc
```diff
@@ -480,11 +480,24 @@ void InitFloorplan::makeTracks()
...
     if (layer->getType() == dbTechLayerType::ROUTING
         && layer->getRoutingLevel() != 0) {
-      makeTracks(layer,
-                 layer->getOffsetX(),
-                 layer->getPitchX(),
-                 layer->getOffsetY(),
-                 layer->getPitchY());
+      if (layer->getTechLayerPitchRules().empty()) {
+        std::cout << "\n\naqui esta entrado com a layer: " + layer->getName()
+                         + " \n\n";
```
rovinski, line 485: 👀

rovinski:
"Non Uniform" should be hyphenated to "Non-Uniform" in comments

src/ifp/src/InitFloorplan.cc
```diff
@@ -569,4 +579,80 @@ void InitFloorplan::makeTracks(odb::dbTechLayer* layer,
...
+                                         int x_pitch,
+                                         int y_offset,
+                                         int y_pitch,
+                                         int first_last_pitch)
+{
+  if (layer->getDirection() != dbTechLayerDir::HORIZONTAL) {
+    logger_->error(IFP, 40, "Non horizontal layer uses property LEF58_PITCH.");
+  }
+  auto rows = block_->getRows();
+  int cell_row_height = rows.begin()->getSite()->getHeight();
```
gadfort, line 593: You cannot assume the first row will not be of type PAD, you need to find the appropriate row

src/ifp/src/InitFloorplan.cc
```diff
@@ -480,11 +480,21 @@ void InitFloorplan::makeTracks()
...
+                   layer->getOffsetY(),
+                   layer->getPitchY());
+      } else {
+        auto pitch_rule = layer->getTechLayerPitchRules().begin();
+        makeTracksNonUniform(layer,
+                             layer->getOffsetX(),
+                             pitch_rule->getPitchX(),
+                             layer->getOffsetY(),
+                             pitch_rule->getPitchY(),
+                             pitch_rule->getFirstLastPitch());
```
gadfort, line 496: If more than 1 can be defined, why only generate one?

arthurjolo: According to the definition Matt sent me, this rule is supposed to be maximum 1 per layer. It allows to generate more than 1 because I used the codeGenerator, but never more than 1 is supposed to be created .

src/odb/src/db/dbTechLayer.cpp
```diff
@@ -733,6 +746,7 @@ dbIStream& operator>>(dbIStream& stream, _dbTechLayer& obj)
   stream >> *obj.width_table_rules_tbl_;
   stream >> *obj.min_cuts_rules_tbl_;
   stream >> *obj.area_rules_tbl_;
+  stream >> *obj.pitch_rules_tbl_;
```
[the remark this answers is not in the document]
arthurjolo: I don't understand, what version check?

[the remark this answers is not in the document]
gadfort: Here are the instructions: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/doc/OpenDB-AddFieldsInDbObjects.md

src/odb/src/lefin/lefTechLayerPitchRuleParser.cpp
```diff
@@ -0,0 +1,108 @@
...
+                                  &odb::dbTechLayerPitchRule::setPitchY)]
+         | lit("PITCH") >> double_[boost::bind(
+               &odb::lefTechLayerPitchRuleParser::setPitch,
+               this,
+               _1,
+               rule,
+               &odb::dbTechLayerPitchRule::setPitch)]);
+
+  qi::rule<std::string::iterator, space_type> FIRST_LAST_PTICH
+      = (PITCH >> (lit("FIRSTLASTPITCH") >> double_[boost::bind(
```
maliberty, line 64: FIRSTLASTPITCH is optional

src/odb/src/db/dbTechLayer.cpp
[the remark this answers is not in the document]
maliberty: If some has an old db without the new field we still want to be able to load it.  Look at how db_schema_wrongway_width is setup to see how we handled it for another addition.

src/odb/src/db/dbTechLayerPitchRule.cpp
```diff
@@ -0,0 +1,238 @@
...
+  // User Code Begin CopyConstructor
+  // User Code End CopyConstructor
+}
+
+dbIStream& operator>>(dbIStream& stream, _dbTechLayerPitchRule& obj)
+{
+  stream >> obj.first_last_pitch_;
+  stream >> obj.pitch_x_;
+  stream >> obj.pitch_y_;
+  stream >> obj.has_pitch_xy_;
```
maliberty, line 121: Why do we need to has_pitch_xy_?

maliberty (commented):
I haven't finished reviewing but some initial comments

src/odb/src/lefin/lefTechLayerPitchRuleParser.cpp
```diff
@@ -0,0 +1,108 @@
...
+void lefTechLayerPitchRuleParser::setPitchXY(
+    boost::fusion::vector<double, double>& params,
+    odb::dbTechLayerPitchRule* rule,
+    void (odb::dbTechLayerPitchRule::*func1)(int),
+    void (odb::dbTechLayerPitchRule::*func2)(int))
+{
+  (rule->*func1)(lefin_->dbdist(at_c<0>(params)));
+  (rule->*func2)(lefin_->dbdist(at_c<1>(params)));
+}
+
```
osamahammad21, line 107: I think you are copying the implementation of function binding but out of context here. The idea of having a class method with the rule's method as an argument is to avoid having multiple parser methods that does the same thing but only changes the called rule's method.
You should do one of two things:
- Have only one method "setInt" that takes the rule's method as an argument
- Do not pass the rule method as an argument and call it directly in the parser method.

src/odb/src/lefin/lefTechLayerPitchRuleParser.cpp
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright (c) 2022, The Regents of the University of California
```
osamahammad21, line 2: 2023

src/odb/src/db/dbTrackGrid.cpp
```diff
@@ -246,6 +246,38 @@ void dbTrackGrid::getGridPatternY(int i,
...
+  _dbTrackGrid* grid = (_dbTrackGrid*) this;
+  for (int i = 0; i < n_rows; i++) {
+    grid->_y_origin.push_back(origin_y);
+    grid->_y_count.push_back(line_count);
+    grid->_y_step.push_back(step);
+    origin_y
+        += (2 * first_last_offset) + (line_count * step) + first_last_offset;
+  }
+}
+
```
osamahammad21, line 280: I don't see any use for these methods in this PR. Also if there is a future use, I would prefer that this for loop to be implemented outside the dbTrackGrid class.

src/odb/src/db/dbTechLayer.cpp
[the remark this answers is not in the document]
osamahammad21: I am afraid that this part is handled by the db code generator and editing it would require editing the code generator itself to handle the version check. When we started the code generator, there was no db version to handle.

osamahammad21 (changes requested):

src/odb/src/db/dbTechLayer.cpp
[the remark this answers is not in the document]
gadfort: That's unfortunate, the right solution would be to fix the code generator so we can have backwards compatibility

src/ifp/src/InitFloorplan.cc
QuantamHD: I'm also unclear if only one can be defined the layer->getTechLayerPitchRules() should return a single instance.

src/odb/src/db/dbTrackGrid.cpp
arthurjolo: It can be done from the outside using the existing methods, and I agree it would be a beter solution, I will start changing it.

src/ifp/src/InitFloorplan.cc
arthurjolo: All the methods regarding the class dbTechLayerPitchRule were generated using the code generator, and the layer->getTechLayerPitchRules() made by the code generator returns a table. I didn't tried to change anything done by the code generator. @osamahammad21 is there a way to make it return just the class instance?

src/odb/src/lefin/lefTechLayerPitchRuleParser.cpp
arthurjolo: If the layer definition has "PROPERTY LEF58_PITCH" it is optional? because this parser is only called if there is a PROPERTY LEF58_PITCH on the layer definition

src/ifp/src/InitFloorplan.cc
osamahammad21: Unfortunately 1 to 1 relations are not supported yet in the odb code generator.

gadfort: It seems like maybe this would be better implemented outside of the code generator then? Otherwise we will have to go back and change this data type later which might not be so easy

src/ifp/src/InitFloorplan.cc
```diff
@@ -569,4 +579,80 @@ void InitFloorplan::makeTracks(odb::dbTechLayer* layer,
...
+    x_track_count--;
+  }
+
+  grid->addGridPatternX(origin_x, x_track_count, x_pitch);
+
+  auto y_track_count = int((cell_row_height - 2 * first_last_pitch) / y_pitch);
+  int origin_y = die_area.yMin() + first_last_pitch;
+
+  grid->addNonUniformGridY(
+      origin_y, y_track_count, y_pitch, first_last_pitch, rows.size());
```
gadfort, line 655: can't assume `rows.size()` only includes CORE rows

gadfort:
@maliberty how is this stored in the DEF file?

src/odb/src/lefin/lefTechLayerPitchRuleParser.cpp
maliberty: image
the square brackets indicate that the FIRSTLASTPITCH clause is optional.   Without the property this is equivalent to the older PITCH syntax but we should still handle it.

src/odb/src/db/dbTechLayer.cpp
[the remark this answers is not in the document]
maliberty: @osamahammad21 we could add a schema_check field to the json and a generate a corresponding check here.  It doesn't seem like a big change to the generator.  What do you think?

src/ifp/src/InitFloorplan.cc
maliberty: Looking at this I think it would be simpler to add a _first_last_pitch field to dbTechLayer (like pitch_x/y).  It would have a no-value (eg -1).  Simple hasFirstLastPitch() and getFirstLastPitch() methods on dbTechLayer would suffice.  It doesn't have to be a separate object.

maliberty:
Please add a unit test that mirrors asap7 m2's definition.

src/odb/src/db/dbTechLayer.h
```diff
@@ -85,6 +85,7 @@ struct dbTechLayerFlags
   uint has_protrusion_ : 1;
   uint has_alias_ : 1;
   uint has_xy_pitch_ : 1;
+  uint has_first_last_pitch_ : 1;
```
osamahammad21, line 88: This increases the number of bits in the bitfield struct to 33. in the stream functions, the flags are handled as uint32_t. I recommend removing the flag and depending on the value of first_last_patch_ variable. If it's greater than or equal to 0, then the layer has a first last pitch.
If you prefer to use the bool flag, then make sure to correctly read/write and the tech layer flags. Also move the has_first_last_pitch_ to the end of the struct so that when you read an older db, it doesn't mess up the variables. And for backward compatibility, you will need to handle older dbs with a 32-bit flag, and new dbs with a 64-bit flag.

src/odb/src/db/dbTechLayer.h
```diff
@@ -135,11 +136,11 @@ class _dbTechLayer : public _dbObject
   dbTable<_dbTechLayerMinCutRule>* min_cuts_rules_tbl_;
 
   dbTable<_dbTechLayerAreaRule>* area_rules_tbl_;
-
   // User Code Begin Fields
 
   uint _pitch_x;
   uint _pitch_y;
+  uint _first_last_pitch;
```
osamahammad21, line 143: please update the following functions to handle this variable:
- _dbTechLayer::out
- _dbTechLayer::differences
- _dbTechLayer::operator==
- **dbIStream& operator>>(dbIStream& stream, _dbTechLayer& obj)**
- **dbOStream& operator<<(dbOStream& stream, const _dbTechLayer& obj)**

osamahammad21 (changes requested):
Could you also add the necessary test under odb/test/cpp/TestLef58Properties.cpp?

arthurjolo:
I am trying to add the test, but for some reason it get en error when it parses the LEF

src/odb/src/db/dbTechLayer.cpp
```diff
@@ -1996,12 +2015,27 @@ void dbTechLayer::setPitchXY(int pitch_x, int pitch_y)
...
   return layer->flags_.has_xy_pitch_;
 }
 
+bool dbTechLayer::hasPitch()
+{
+  _dbTechLayer* layer = (_dbTechLayer*) this;
+  if (layer->_pitch_x) {
+    return true;
+  }
+  return false;
```
maliberty, line 2036:
```suggestion
  return layer->_pitch_x != 0;
```

src/odb/include/odb/db.h
```diff
@@ -7195,9 +7195,12 @@ class dbTechLayer : public dbObject
   int getPitch();
   int getPitchX();
   int getPitchY();
+  int getFirstLastPitch();
   void setPitch(int pitch);
   void setPitchXY(int pitch_x, int pitch_y);
+  void setFirstLastPitch(int first_last_pitch);
   bool hasXYPitch();
+  bool hasPitch();
```
maliberty, line 7203: Why is this need for this change?

arthurjolo: It is used to see if the pitch has alredy been set by the LEF58_PITCH, if it has then the PUTCH is ignored.

maliberty: The PITCH statement should precede the LEF58_PITCH so I don't think this is necessary.  

Its a bit confusing as an API as the PITCH will always be set when parsing is finished.

maliberty: I'm confused about how this happens.  The PITCH statement should come before the LEF58_PITCH.  Does the parser not return these in order?

arthurjolo: I belive when it first reads the LEF the parser reads it in order and store the information on the class lefiLayer, but it doesn't fineshes parsing the LEF58 properties. So on the class lefiLayer it first fineshes parsing the LEF58 properties and tranfering the information to the dbTechLayer, and after that it transfers the information already parsed from the lefiLayer to the dbTechLayer.

src/odb/src/db/dbTechLayer.h
```diff
@@ -157,6 +156,7 @@ class _dbTechLayer : public _dbObject
   int _min_step;
   int _min_step_max_length;
   int _min_step_max_edges;
+  uint _first_last_pitch;
```
osamahammad21, line 159: change to int. The null value is -1 in the constructor.

osamahammad21 (changes requested):
Everything else looks fine to me (in odb).

osamahammad21:
@maliberty This PR looks good for me in ODB. Could you please confirm the IFP changes are okay and approve?

osamahammad21:
@arthurjolo could you fix the build failures:
```
#8 377.0 Error: IFP 0040 used 2 times
#8 377.0 Error: IFP 0041 used 2 times
```

maliberty (approved):

maliberty:
I think this is fine.  It will be a separate step/PR to switch asap7 over.

maliberty (approved):

PR #2973, opened 2023-03-06
Net Assigned Tracks

osamahammad21:
This PR include changes to store the results of drt track assignment in opendb and show them in gui.


src/odb/src/db/dbNet.cpp
```diff
@@ -167,6 +170,7 @@ dbOStream& operator<<(dbOStream& stream, const _dbNet& net)
   stream << net._ccAdjustOrder;
   stream << net._groups;
   stream << net.guides_;
+  stream << net.tracks_;
```
maliberty, line 173: version compat

src/drt/src/io/io.cpp
```diff
@@ -3077,4 +3090,5 @@ void io::Writer::updateDb(odb::dbDatabase* db, bool pin_access)
     fillConnFigs(false);
     updateDbConn(block, db_tech);
   }
+  updateTrackAssignment(block);
```
maliberty, line 3093: This should be optional

maliberty:
@osamahammad21 please fix this up so we can get it merged

maliberty (approved):

PR #2974, opened 2023-03-06
Goto feature proposal

niv280:
I added goto feature.
It can be accessed through view menu or with 'Shift+G' shortcut.
The frame is look like that:

image

Once the user entered coordinates the Gui will zoomTo to the next rect:
(x - margin, y - margin, x + margin, y + margin)
When the default margin is 50 for now.

Do you think that the default margin is ok?
Do you want to add the option to enter the margin manually in different widget


QuantamHD:
Love the feature. A few thoughts

You could perhaps simplify the user interface a bit if the coordinates were dynamic. So instead of having an okay button as a user typed numbers in the location would change.

Having it in a pane form like photoshop or blender might also be helpful

image

src/gui/src/gotoDialog.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace gui {
+class GotoObjectDialog : public QDialog, public Ui::GotoObjDialog
+{
+  Q_OBJECT
+ public:
+  GotoObjectDialog(QWidget* parent = nullptr);
+ public slots:
+  void accept();
+  void reject();
```
maliberty, line 47: Please add the override keyword

src/gui/src/gotoDialog.h
```diff
@@ -0,0 +1,49 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <QDialog>
+
+#include "ui_gotoDlg.h"
+
+namespace gui {
+class GotoObjectDialog : public QDialog, public Ui::GotoObjDialog
```
maliberty, line 40: This doesn't go to an object but to a location so perhaps GotoLocationDialog

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,63 @@
...
+GotoObjectDialog::GotoObjectDialog(QWidget* parent) : QDialog(parent)
+{
+  setupUi(this);
+}
+
+void GotoObjectDialog::accept()
+{
+  int x_coord = xEdit->text().toDouble();
+  int y_coord = yEdit->text().toDouble();
+  int box_margin = 100;
```
maliberty, line 49: const

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,63 @@
...
+namespace gui {
+GotoObjectDialog::GotoObjectDialog(QWidget* parent) : QDialog(parent)
+{
+  setupUi(this);
+}
+
+void GotoObjectDialog::accept()
+{
+  int x_coord = xEdit->text().toDouble();
+  int y_coord = yEdit->text().toDouble();
```
maliberty, line 48: Is the user entering microns or dbus?  It should follow the display settings for the GUI (entry units = display units).

src/gui/src/gotoDialog.cpp
maliberty, line 49: depends on the units

src/gui/ui/gotoDlg.ui
```diff
@@ -0,0 +1,214 @@
...
+            <string>X</string>
+           </property>
+          </widget>
+         </item>
+        </layout>
+       </item>
+       <item>
+        <layout class="QVBoxLayout" name="verticalLayout_8">
+         <item alignment="Qt::AlignVCenter">
+          <widget class="QLineEdit" name="xEdit">
```
maliberty, line 104: A QDoubleSpinBox might work better

maliberty:
Visually I would prefer

X: <widget>
Y: <widget>

You should set the buddy property of the QLabel as well.

maliberty:
@QuantamHD would a modeless dialog with an apply button satisfy your needs?

maliberty:
Thanks for the contribution!

QuantamHD:
I do like reactive components more than models, but I think an apply with a panable window is a good alternative that doesn't require as much work.

maliberty:
> I do like reactive components more than models, but I think an apply with a panable window is a good alternative that doesn't require as much work.

I dont know what "reactive components more than models" means.  Panning is a different operation than goto.

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,63 @@
+//////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2019, The Regents of the University of California
```
gadfort, line 4: 2023

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,63 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include "gotoDialog.h"
+
+#include <string>
```
gadfort, line 35: where is string needed?

src/gui/src/gotoDialog.cpp
gadfort: it might be nice to be able to control this in the UI

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,63 @@
...
+                        y_coord - box_margin,
+                        x_coord + box_margin,
+                        y_coord + box_margin));
+  QDialog::accept();
+}
+
+void GotoObjectDialog::reject()
+{
+  QDialog::reject();
+}
```
gadfort, line 61: Do you need this if you are doing the exact same as the parent class?

src/gui/src/gotoDialog.h
```diff
@@ -0,0 +1,49 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2021, The Regents of the University of California
```
gadfort, line 4: 2023

src/gui/src/gotoDialog.h
gadfort: GotoLocationDialog makes sense to me

src/gui/src/mainWindow.cpp
```diff
@@ -513,6 +514,10 @@ void MainWindow::createActions()
 
   find_ = new QAction("Find", this);
   find_->setShortcut(QString("Ctrl+F"));
+
+  goto_position_ = new QAction("Goto position", this);
```
gadfort, line 518: should this be "Go to position"?

src/gui/src/mainWindow.cpp
```diff
@@ -1124,6 +1131,13 @@ void MainWindow::showFindDialog()
   find_dialog_->exec();
 }
 
+void MainWindow::showGotoDialog()
+{
+  if (getBlock() == nullptr)
+    return;
```
gadfort, line 1137: clang-format

src/gui/src/mainWindow.cpp
```diff
@@ -1124,6 +1131,13 @@ void MainWindow::showFindDialog()
...
+  goto_dialog_->exec();
```
gadfort, line 1138: when it opens, it should get the current center and margins to help guide the user (you would also be able to get the DBU vs. micron at this point)

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,63 @@
...
+void GotoObjectDialog::accept()
+{
+  int x_coord = xEdit->text().toDouble();
+  int y_coord = yEdit->text().toDouble();
+  int box_margin = 100;
+  auto gui = gui::Gui::get();
+  gui->zoomTo(odb::Rect(x_coord - box_margin,
+                        y_coord - box_margin,
+                        x_coord + box_margin,
+                        y_coord + box_margin));
```
gadfort, line 54: Default should be to enter in microns, this assumes that the user entered the coordinates in DBUs

src/gui/ui/gotoDlg.ui
```diff
@@ -0,0 +1,214 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gadfort, line 1: There is a lot of empty space in the widget, you might want to find a way to remove it

gadfort (changes requested):

src/gui/src/gotoDialog.cpp
niv280: In order to change the units according to the display units I thought of using the following code:
``` 
  if (block == nullptr) {
    const double dbu_per_microt = 1;
  } else {
    const double dbu_per_micron = block->getDbUnitsPerMicron();
  }
  const double x_coord = xEdit->text().toDouble() / dbu_per_micron;
  const double y_coord = yEdit->text().toDouble() / dbu_per_micron;
  const double box_margin = 100 / dbu_per_micron; 
 ```
The problem is that I didn't find way to get the block inside gotoDialog.
There is any way to do it without editing the gui.cpp file?

File: src/gui/src/mainWindow.cpp (hunk 10 above)
Comment (niv280) [reply src/gui/src/mainWindow.cpp:518]: In klayout is written the same way, I can separate it if you think it is better.

File: src/gui/src/gotoDialog.cpp (hunk 4 above)
Comment (gadfort) [reply src/gui/src/gotoDialog.cpp:48]: The gui.cpp has two helper functions:
Descriptor::Property::convert_dbu -> takes in dbus and returns a string in either DBUs or microns depending on the GUI setting
Descriptor::Property::convert_string -> takes a string (as microns or DBUs) and returns DBUs depending on GUI settings

```
bool convert_x_ok;
const int x_coord = Descriptor::Property::convert_string(xEdit->text().toStdString(), &convert_x_ok);
const int y_coord = Descriptor::Property::convert_string(yEdit->text().toStdString(), &convert_y_ok);
if convert_x_ok and convert_y_ok {
  okay to continue
} else { failure message }
```

when populating the current center:
```
xEdit->text(QString::fromStdString(Descriptor::Property::convert_dbu(x_center, false));
yEdit->text(QString::fromStdString(Descriptor::Property::convert_dbu(y_center, false));
```
currently the layout viewer doesn't have a way to get the center, but that would be easy to add as well.

Comment (niv280) [reply src/gui/src/gotoDialog.cpp:48]: Thanks!
I tried this, when the show DBU is off that worked but when he is on the convert_string returning 0 no matter what the input in the dialog.

Any idea why?

Comment (gadfort) [reply src/gui/src/gotoDialog.cpp:48]: 0 seems to indicate that the conversion failed, but I would need to take a look

File: src/gui/src/mainWindow.cpp (hunk 10 above)
Comment (maliberty) [reply src/gui/src/mainWindow.cpp:518]: "Go to" is better as it is an English phrase.  Goto is just a programming keyword.

File: src/gui/src/gotoDialog.cpp (hunk 4 above)
Comment (niv280) [reply src/gui/src/gotoDialog.cpp:48]: Found the issue, I change the ui from Line edit to spin box and it didn't converted to string.
BTW there is easy way to get the center coordinates?
I saw there is getVisibleCenter but it is in the LayoutViewer

**Comment by niv280** [comment]
Hi,
I improve a bit according to your comments, but it's not perfect yet.
There is few open items that I would love to get help with:
1. I removed the ok/cancel buttons and now after every change the gui will zoomTo automatically, do you want to leave it like that? or that it will be better if it will be panel in the main window. 
2. The margin of the box that will zoom into right now is not good, for now it is const value so it work either for DBU or micron. I can try to check the display units and then give 2 different value or that it will be controlled from the ui.
3. I tried to display to current center when the dialog opened, but I didn't succeed to use the function giveVisibleCenter even after I move it to public it required to create LayoutViewer object which break the build.

**Comment by niv280** [comment]
Hi,
I improve a bit according to your comments, but it's not perfect yet.
There is few open items that I would love to get help with:
1. I removed the ok/cancel buttons and now after every change the gui will zoomTo automatically, do you want to leave it like that? or that it will be better if it will be panel in the main window. 
2. The margin of the box that will zoom into right now is not good, for now it is const value so it work either for DBU or micron. I can try to check the display units and then give 2 different value or that it will be controlled from the ui.
3. I tried to display to current center when the dialog opened, but I didn't succeed to use the function giveVisibleCenter even after I move it to public it required to create LayoutViewer object which break the build.

**Comment by maliberty** [comment]
I would drop the margin and just use
```
  if (convert_x_ok && convert_y_ok){
    gui->centerAt({x_coord, y_coord});
  }
```

**Comment by maliberty** [comment]
You could add an extra field for window size rather than use a fixed margin.  In that case it should be prepopulated with the current value

**Comment by niv280** [comment]
Ok, I took @maliberty   advice and used centerAt instead of zoomTo.
The upside is that is more elegant but the downside is that when the gui is in see the whole design the goto seems like he didn't  focused.

Beside that I added the center coord when the dialog opened, and change the trigger from exec to show() which make is possible to move only the new dialog and switch between the windows ( I think the find dialog should also be like it).

**Comment by maliberty** [comment]
I agree that it is a bit non-intuitive when you are fully zoomed out.  Having a field for size would make that more obvious.

**Comment by niv280** [comment]
I added the box size as new field with default value for each display units:
image

**Comment by maliberty** [comment]
I think that will help a lot.  Please change 'S' to 'Size' in the dialog for clarity.

**Comment by maliberty** [comment]
You should buddy the labels with their fields (see https://doc.qt.io/qt-5/designer-buddy-mode.html)

File: src/gui/src/gotoDialog.h [hunk 15]
```diff
@@ -0,0 +1,48 @@
...
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <QDialog>
+
+#include "ui_gotoDlg.h"
```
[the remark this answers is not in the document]
maliberty: @QuantamHD do you know how to get clang-tidy to see this generated file?

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+void GotoLocationDialog::show_init(LayoutViewer* viewer_)
+{
+  xEdit->setText(QString::fromStdString(Descriptor::Property::convert_dbu(
+      viewer_->getVisibleCenter().x(), false)));
+  yEdit->setText(QString::fromStdString(Descriptor::Property::convert_dbu(
+      viewer_->getVisibleCenter().y(), false)));
+  sEdit->setText(
+      QString::fromStdString(Descriptor::Property::convert_dbu(300, false)));
```
maliberty, line 50: When the window appears the scale is not the current scale which is a bit confusing.  Could you populate this from the layout viewer?

maliberty:
When 'Show DBU' is toggled then the widget needs to switch its values to dbu.  You can see how the inspector does this based on the event callbacks.

niv280:
Hi @maliberty,
I'm trying to make the widget to switch the number according to showDBU by using the convert_string/convert_dbu functions but I encounter some problems.
In both of the function the condition for the converstion is that the showDBU should be checked, but in this case I can't use them to toggle between them:

```
std::string MainWindow::convertDBUToString(int value, bool add_units) const
{
  if (show_dbu_->isChecked()) {
    return std::to_string(value);
  } else {
    auto* block = getBlock();
    if (block == nullptr) {
      return std::to_string(value);
    } else {
      const double dbu_per_micron = block->getDbUnitsPerMicron();

      std::stringstream ss;
      const int precision = std::ceil(std::log10(dbu_per_micron));
      const double micron_value = value / dbu_per_micron;

      ss << std::fixed << std::setprecision(precision) << micron_value;

      if (add_units) {
        ss << " \u03BCm";  // micro meter
      }

      return ss.str();
    }
  }
}

int MainWindow::convertStringToDBU(const std::string& value, bool* ok) const
{
  QString new_value = QString::fromStdString(value).simplified();

  if (new_value.contains(" ")) {
    new_value = new_value.left(new_value.indexOf(" "));
  } else if (new_value.contains("u")) {
    new_value = new_value.left(new_value.indexOf("u"));
  } else if (new_value.contains("\u03BC")) {
    new_value = new_value.left(new_value.indexOf("\u03BC"));
  }

  if (show_dbu_->isChecked()) {
    //return new_value.toInt(ok);
    return 999;
  } else {
    auto* block = getBlock();
    if (block == nullptr) {
      return new_value.toInt(ok);
    } else {
      const int dbu_per_micron = block->getDbUnitsPerMicron();

      return new_value.toDouble(ok) * dbu_per_micron;
    }
  }
}
```

Is that make sense?
There is other way to do it beside copying the entire function to the widget code?

niv280:
Hi,
I did the following change in the widget:
1. Buddy the label with the lineEdit
2. Added intial box size based on the scale of the dbu_bounds.
   ( Was needed to add function to access the dbu_bounds from the LayoutViewer )
3. The conversation of the units is dynamically changing when the showDBU is toggle - 
    * For that I was needed to add new signal that emit from setUseDBU function in mainWindow.
    * I could not use the conversion function for that, so I did the calculation in specific function in my new code.

niv280:
Sure, I will update the max value.

The problem is when update the goto after every key press is that it looked to me a bit inconvenient, for example when entering the number 1000 it change after each key and looks kind of messy.

But I can change it back if you think it is better, or add apply key, what do you think it is better?

maliberty:
Perhaps a small timer delay - goto after a 1/2 sec pause (or enter or tab immediately)?

src/gui/src/layoutViewer.cpp
```diff
@@ -2788,7 +2786,7 @@ void LayoutViewer::updateBlockPainting(const QRect& area)
   // apply scaling
   block_painter.scale(pixels_per_dbu_, -pixels_per_dbu_);
 
-  const Rect dbu_bounds = screenToDBU(area);
+  dbu_bounds = screenToDBU(area);
```
gadfort, line 2789: Why this change?

src/gui/src/layoutViewer.h
```diff
@@ -169,6 +169,9 @@ class LayoutViewer : public QWidget
   virtual void mouseMoveEvent(QMouseEvent* event) override;
   virtual void mouseReleaseEvent(QMouseEvent* event) override;
 
+  odb::Rect dbu_bounds;
```
gadfort, line 172: This will not update on resizes. I think it would be better to ask the window for its visible size instead of storing this value. (which tends not work correctly with Qt)

src/gui/src/layoutViewer.h
```diff
@@ -169,6 +169,9 @@ class LayoutViewer : public QWidget
...
+  odb::Rect getDBUbounds() { return dbu_bounds; }
```
gadfort, line 173: This should be getDBUBound()

gadfort (changes requested):

src/gui/src/gotoDialog.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+  if (useDBU) {
+    xEdit->setText(QString::number(xEdit->text().toDouble() * dbu_per_micron));
+    yEdit->setText(QString::number(yEdit->text().toDouble() * dbu_per_micron));
+    sEdit->setText(QString::number(sEdit->text().toDouble() * dbu_per_micron));
+  } else {
+    xEdit->setText(QString::number(xEdit->text().toDouble() / dbu_per_micron));
+    yEdit->setText(QString::number(yEdit->text().toDouble() / dbu_per_micron));
+    sEdit->setText(QString::number(sEdit->text().toDouble() / dbu_per_micron));
+  }
```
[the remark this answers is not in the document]
niv280: the convert function converting only when the showDBU is set, but I need it to convert it to DBU if it set and convert to micron if it unset. I write about it above, I don't sure why the function act like it.

src/gui/src/layoutViewer.h
niv280: I was looking for this option but didn't find how to do it, there is easy way to get the visible size from the window?

src/gui/src/layoutViewer.cpp
niv280: Mistake, I changed it back

src/gui/src/layoutViewer.h
niv280: Changed

src/gui/src/layoutViewer.h
gadfort: you should change `getDBUBounds()` to return:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/4066c4222535a116eb01adabc4e73ffc44c0d5ab/src/gui/src/layoutViewer.cpp#L3152

And in the process, maybe `getVisibleBounds()` is a better name for the function.

niv280: Changed

niv280:
Changed the requested changed

maliberty (approved):
I think this can be refined further but would like to get this PR merged.

Please resolve the merge conflicts.

JohnPhilipBiggs:
Am I the only one who finds the "Goto Position" hard to use?

When I want to go to a specific position would expect to bring up the dialog box, delete the existing coordinates and enter some new ones.
However when I try an delete the existing coordinates using the back space key the UI wont let me delete all the digits - it insists on there being at least one digit in the box -  this means that I have to type a digit from my new coordinate as a place holder before I can delete the last digit from the old coordinate. 
A work around is to type CTRL-A to select all the old digits and then over type them with the new digits - but this is somewhat  counter intuitive.
Also, despite the dialog box popping up with default coordinates that have a decimal point - it wont let you enter new coordinates with a decimal point.

Can I suggest that this "Goto Position" be simplified - I dont think it needs to be interactive nor does it need a size parameter.  Simply taking X,Y coordinate and panning to that location at the current zoom level at the click of a button would be perfect.

maliberty:
I agree it is a bit awkward.  The goal is something like what klayout has.  It would be better to have an issue or discussion rather than use this old PR.

Issue #2891, opened 2023-02-13
Improve modeling of instance blockages in global routing


maliberty:
### Description

In GlobalRouter::findInstancesObstructions we merge all the instance obs into a single rectangle.  This is overly conservative and blocks areas where over-the-cell routing is possible.  This particularly happens with our LEF generator that exposes such area between power grid stripes.

### Suggested Solution

Store blocked track intervals in a style similar to what we do for blocked tracks elsewhere.

PR #2975, opened 2023-03-06
Grt: improve modeling of instance blockages

luis201420:
Solving Issue #2891


maliberty (commented):
Please add a unit test

maliberty:
Please run a secure CI with this change

luis201420:
> Please run a secure CI with this change

I ran the secure CI:
- In the public designs the **sky130hd/microwatt** design has an metric problem
`[ERROR] finish__timing__setup__ws fail test: -2.42731 >= -1.76`
- The **ng45/ariane136** and **ng45/bp** designs and the **private designs** are running now, but the previous run had no errors.

maliberty:
Please look at why uW degrades

luis201420:
> Please look at why uW degrades

The difference is in their critical path. This has two nets (soc0.\_331\_ and \_017339\_) that are routed differently.
The figure shows the routing of the net soc0.\_331\_ (In \_017339\_ you also get a detour around the macro)

diff_route_net

This deviation happens in the overflow iterations in GRT, because the resources of each layer change (the GRT resources are increased in some places ).

| Layer | Test branch | Master branch | Percentage |
|--------|--------|--------|--------|
| met1 | 1315034 | 1315034 | 0 % |
| met2 | **1318940** | 1318743 | **0.0149 %** |
| met3 | **910803** | 910801 | **0.0002195 %** |
| met4 | 401758 | 401758 | 0 % |
| met5 | 215642 | 215642 | 0 % |

The areas where resources are released are the following: (the yellow circles have a Ggrid that increases their resource and the green circles are all the ggrids that are on the border of the macro)

macros_uw

Increasing the grt resources in those zones causes the routing of the two nets to change, this may be due to other nets using the resources that were not used before or the congestion below is reduced faster (because there is more resources) and choose to route that path.
The following images show the congestion map in both versions:

diff_congestion

antonblanchard:
So we have more routing resources, but we've done a much worse job of global routing a critical net. I just noticed timing driven grt is disabled by default (`-critical_nets_percentage`), we should either update ORFS to use it or change the default to something more reasonable. gpl looks at the 10% most critical nets by default, so perhaps grt should too.

maliberty:
It suggests that this PR works correctly but just triggers the underlying concern about timing driven global routing.

@eder-matheus do you know why we never set critical_nets_percentage in ORFS?

eder-matheus:
> It suggests that this PR works correctly but just triggers the underlying concern about timing driven global routing.
> 
> @eder-matheus do you know why we never set critical_nets_percentage in ORFS?

Not really, but I can fix that. I'll start a secure-ci with this grt option and will update you when it's done. I'll start it with 10%, just as in gpl as @antonblanchard mentioned.

maliberty:
@eder-matheus any update?

eder-matheus:
> @eder-matheus any update?

I've run secure-ci using the 10% critical nets, and some public and private designs have ws failures and drv violations. I think it relates to the change of the topology of other nets that become critical after routing. I considered adding the `critical_nets_percentage` option after we have the slack calculated inside grt based on the intermediary net topologies generated in the grt flow.

eder-matheus:
@maliberty Luis and I performed a secure-ci run yesterday, and everything is passing now. The only two failures in the public CI are the same as we have in the master branch, so merging this is safe.

maliberty (approved):

PR #2978, opened 2023-03-06
IFP: Multi height cell support

mina1460:
This PR would serve as the foundation for supporting multi-height cells.

It also includes support for removing the need for the -site name flag for init_floorplan command. It would still work with and without it.


src/ifp/include/ifp/InitFloorplan.hh
```diff
@@ -35,8 +35,9 @@
 
 #pragma once
 
+#include <set>
 #include <string>
-
+#include <vector>
```
maliberty, line 40: I don't see a usage of vector in this header.  It is needed elsewhere please add it to the file(s) that use it.

src/ifp/src/InitFloorplan.cc
```diff
@@ -360,11 +389,53 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
...
+{
+  std::set<dbSite*> sites;
+
+  // loop over all instantiated cells in the block
+  for (dbInst* inst : block_->getInsts()) {
+    dbMaster* master = inst->getMaster();
+    // get the site of the master
+    auto site = master->getSite();
+    if (site == nullptr) {
+      continue;
```
maliberty, line 402: A warning would be good here as this would be unexpected.

src/ifp/src/InitFloorplan.cc
```diff
@@ -360,11 +389,53 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
...
+  });
+
+  // smallest site height
+  double min_site_height = site_vec[0]->getHeight();
+  double factor = 1.0;
+  for (auto site : site_vec) {
+    double site_height = site->getHeight();
+    factor = site_height / min_site_height;
+    // check if the site height is a multiple of the smallest site height
+    if (std::ceil(factor) != std::floor(factor)) {
```
maliberty, line 419: Use integers here not doubles.  Check with the modulo operator not ceil/floor.

src/ifp/src/InitFloorplan.cc
```diff
@@ -360,11 +389,53 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
...
+      logger_->error(
+          IFP,
+          228,  // TODO: add error code
```
maliberty, line 422: Why is this a TODO?

src/ifp/src/InitFloorplan.cc
```diff
@@ -360,11 +389,53 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
...
+          "site height {} is not a multiple of the smallest site height {}",
```
maliberty, line 423: Give the site names too

src/ifp/src/InitFloorplan.cc
```diff
@@ -360,11 +389,53 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
...
+          site_height,
+          min_site_height);
+      // erase from the set if the site height is not a multiple of the smallest
+      sites.erase(site);
```
maliberty, line 427: you'll never get here as error() throws an exception.  The erase wouldn't be safe while iterating anyways

src/ifp/src/InitFloorplan.cc
```diff
@@ -360,11 +389,53 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
...
+    }
+  }
+  return sites;
+}
+
 void InitFloorplan::makeRows(dbSite* site,
```
maliberty, line 433: Please add a unit test that inserts double height rows.

src/ifp/src/InitFloorplan.cc
```diff
@@ -360,11 +389,53 @@ void InitFloorplan::updateVoltageDomain(dbSite* site,
...
+    sites.insert(site);
+  }
+
+  std::vector<dbSite*> site_vec{sites.begin(), sites.end()};
+  sort(site_vec.begin(), site_vec.end(), [](dbSite* a, dbSite* b) {
+    return a->getHeight() < b->getHeight();
+  });
+
+  // smallest site height
+  double min_site_height = site_vec[0]->getHeight();
```
maliberty, line 413: site_vec[0] is invalid if sites.empty()

src/ifp/src/InitFloorplan.cc
```diff
@@ -166,11 +167,41 @@ void InitFloorplan::initFloorplan(const odb::Rect& die,
...
+    dbSite* site_found = findSite(site_name.c_str());
+    if (site_found != nullptr) {
+      sites.insert(site_found);
+    }
+  }
+  // if the sites set is empty, then there are no sites in the design
+  // and user didn't specify a site that we could find
+  if (sites.empty()) {
+    logger_->error(IFP, 33, "No sites found.");
+    return;
```
maliberty, line 188: rm, not reached

src/ifp/src/InitFloorplan.cc
```diff
@@ -166,11 +167,41 @@ void InitFloorplan::initFloorplan(const odb::Rect& die,
...
+  }
+  // get min site height using std::min_element
+  int min_site_height
```
maliberty, line 191: const

src/ifp/src/InitFloorplan.cc
```diff
@@ -166,11 +167,41 @@ void InitFloorplan::initFloorplan(const odb::Rect& die,
...
+      = (*std::min_element(sites.begin(),
+                           sites.end(),
+                           [](dbSite* a, dbSite* b) {
+                             return a->getHeight() < b->getHeight();
+                           }))
+            ->getHeight();
+
+  eval_upf(logger_, block_);
+  for (auto site : sites) {
+    if (core.xMin() >= 0 && core.yMin() >= 0) {
```
maliberty, line 201: The core size won't change with sites so this check should be hoisted.

src/ifp/src/InitFloorplan.cc
```diff
@@ -375,8 +446,9 @@ void InitFloorplan::makeRows(dbSite* site,
 
   int y = core_ly;
   for (int row = 0; row < rows_y; row++) {
-    dbOrientType orient = (row % 2 == 0) ? dbOrientType::R0   // N
-                                         : dbOrientType::MX;  // FS
+    dbOrientType orient = (factor == 2 or row % 2 == 0)
```
maliberty, line 449: any even factor not just 2 (eg quad height cells)

maliberty (changes requested):

src/ifp/src/InitFloorplan.cc
mina1460: I didn't know how I should pick an error code, I thought I should document the code somewhere as each code seems to be unique for a message. Should I just pick the first available number?

maliberty: Use the next available from ifp/messages.txt (ie 40)

src/ifp/src/InitFloorplan.cc
maliberty: This still needs addressing

QuantamHD:
@mina1460 Would you mind rebasing to master I introduced a clang-tidy action that leaves comments on your PR that I would like to see here https://github.com/The-OpenROAD-Project/OpenROAD/pull/2980

mina1460:
@QuantamHD Done

QuantamHD:
Thanks!

src/ifp/test/mock.tcl
```diff
@@ -0,0 +1,7 @@
+source "helpers.tcl"
+read_lef Nangate45/Nangate45.lef
+read_lef Nangate45/fake_macros.lef
+read_def mock.def
+initialize_floorplan -utilization 30 \
+  -aspect_ratio 0.5 \
+  -core_space 1
```
maliberty, line 7: The test needs to do something to validate the results.  Likely you want a write_def and a .defok file.  Please see other tests for examples.

mina1460: yes, absolutely. I just wanted to verify that the lef/def files and the mock macros are OK and this is indeed testing a double height cell before spending time on a defok file.

Should I proceed with it?

maliberty: yes

src/ifp/test/mock.def
```diff
@@ -5,7 +5,7 @@ DESIGN simple_design ;
 UNITS DISTANCE MICRONS 1000 ;
 DIEAREA ( 0 0 ) ( 2000 2000 ) ;
 ROW ROW_0 SingleHeightSite 0 0 FS DO 10 BY 1 STEP 1000 0 ;
-ROW ROW_1 DoubleHeightSite 1200 2200 N DO 12 BY 1 STEP 1100 0 ;
+ROW ROW_1 DoubleHeightSite 0 1000 N DO 10 BY 1 STEP 1000 0 ;
```
maliberty: Why are there rows in the input DEF?  The goal of ifp is to build the rows.  You could just read a .v for this test which would better mirror ORFS (ie the yosys output).

mina1460: makes sense. I will fix it

QuantamHD:
@maliberty @vvbandeira I think @mina1460 is getting a high number of notifications from the sync upstream, because of the waiting for approval status. We should probably add him to the list of approved workflow runners.

maliberty:
@QuantamHD  once his first PR is merged he will automatically be added.  I've never had to do it manually.

maliberty:
The CI failures need to be addressed before this can be merged

mina1460:
Hi @maliberty 

All CI/CD failures are now resolved. 

I modified the updateVoltageDomain and removed the site from it so it is no longer needed. As you said, this had the potential of cutting double-height rows into single-height rows, but I don't think this is a problem with this modification anymore.

maliberty (approved):

PR #2979, opened 2023-03-07
pdn: fix via count and check for via type before getting avgxy

gadfort:


src/pdn/src/via.cpp
```diff
@@ -2851,11 +2851,15 @@ void Via::writeToDb(odb::dbSWire* wire,
     const TechLayer tech_layer(lower_->getLayer());
     int x = 0;
     int y = 0;
+    int ripup_count = 0;
     for (auto* via : ripup_vias) {
       int via_x, via_y;
-      via->getViaXY(via_x, via_y);
-      x += via_x;
-      y += via_y;
+      if (via->getBlockVia() != nullptr || via->getTechVia() != nullptr) {
```
maliberty, line 2857: When would a via have neither of these conditions?

gadfort: The SBoxes also contain patch metals (in the case of a via stack), this caused an assert error when compiled with debug.

maliberty: So 'via' isn't necessarily a via?  That's a bit confusing.

gadfort: Yeah, I can change the variable name (it used to just be vias) :-)

maliberty:
FYI @vvbandeira is looking at the clang-tidy error.  This is the first PR since the action went in so some hiccup.

src/pdn/src/via.cpp
```diff
@@ -2831,44 +2831,48 @@ void Via::writeToDb(odb::dbSWire* wire,
       = check_shapes(lower_, shapes.bottom);
   std::set<odb::dbSBox*> ripup_vias_top = check_shapes(upper_, shapes.top);
 
-  std::set<odb::dbSBox*> ripup_vias;
-  ripup_vias.insert(ripup_vias_bottom.begin(), ripup_vias_bottom.end());
-  ripup_vias.insert(ripup_vias_top.begin(), ripup_vias_top.end());
+  std::set<odb::dbSBox*> ripup_shapes;
+  ripup_shapes.insert(ripup_vias_bottom.begin(), ripup_vias_bottom.end());
+  ripup_shapes.insert(ripup_vias_top.begin(), ripup_vias_top.end());
```
maliberty, line 2836: This makes me wonder if ripup_vias_bottom & ripup_vias_top should also be renamed.  I'll approve here and let you decide what makes sense.

maliberty (approved):

src/pdn/src/via.cpp
gadfort: good point

QuantamHD:
@gadfort You can fix the clang-tidy-review by rebasing to master

PR #2986, opened 2023-03-07
drt/rp: Remove unused GF12 minstep code

antonblanchard:
This code is never called and has via names and layer numbers hard coded in it. I presume it also predates prep_minStepViasCheck which should handle minstep violations in a more generic way.


QuantamHD:
@antonblanchard Can you rebase you PR to master. I made a change to the way the clang-tidy action is run.

maliberty (approved):

maliberty:
Since this is a pure delete I don't think we need to worry about tidy too much.

PR #2996, opened 2023-03-09
drt: fix track assignment for macro pins and bterms

eder-matheus:
This PR fixes how TA handles pins without preferred APs and updates the costing to avoid penalizing iroutes that touch obstructions to access a pin.


eder-matheus:
Secure-ci running now. In my tests with the array of macros in ASAP7, I see a large reduction in the number of weird detours. We still have a few of them, but they make sense since they're avoiding PDN routing.

src/drt/src/ta/FlexTA_init.cpp
```diff
@@ -310,6 +317,103 @@ void FlexTAWorker::initIroute_helper_generic_helper(frGuide* guide,
   }
 }
 
+void FlexTAWorker::initIroute_helper_generic_fix_endpoints(frGuide* guide,
```
maliberty, line 320: Please add some explanatory comments - its not clear what is being fixed

src/drt/src/ta/FlexTA_init.cpp
```diff
@@ -310,6 +317,103 @@ void FlexTAWorker::initIroute_helper_generic_helper(frGuide* guide,
...
+  }
+
+  frCoord newMinBegin = std::numeric_limits<int>::max();
+  frCoord newMaxEnd = std::numeric_limits<int>::min();
+  for (auto& term : result) {
+    switch (term->typeId()) {
+      case frcInstTerm: {
+        auto iterm = static_cast<frInstTerm*>(term);
+        if (iterm->getNet() != net
+            || !iterm->getInst()->getMaster()->getMasterType().isBlock()) {
```
maliberty, line 346: why?

eder-matheus: I've updated my code and removed this function. It was generating connectivity errors because the iroutes were not connected in some designs.

Also, I've noticed that just adding a check to avoid costing connections to the pins was enough to fix the track assignment and keep the straight connection when possible.

maliberty:
What is the result of secure CI?  Does it need to be run again with your changes?

eder-matheus:
> What is the result of secure CI? Does it need to be run again with your changes?

I've run it yesterday night. None of the designs fails, and I see a tiny percentage of wire length reduction in the designs with macros. The runtime wasn't impacted.

eder-matheus:
> So we can consider mock-big-array fixed? @oharboe FYI

Yes. We still have a few detours, but they are caused by PDN obstructions. I can add some figures later today, but the mock-big-array routing between macros is better than before.

Issue #2989, opened 2023-03-08
Build fails on Ubuntu arm64


hpretl:
### Describe the bug

The build of OpenROAD fails on Ubuntu 22.04 for `arm64`. The reason is the `-m64` flag in `src/par/CMakeLists.txt`.

### Expected Behavior

Building successfully.

### Environment

```shell
n/a
```

### To Reproduce

Try to build commit `0264023b6c2a8ae803b8d440478d657387277d93` on Ubuntu 22.04 LTS.

hpretl:
That's what I do now in our Docker build, but should be fixed in the source.

PR #3000, opened 2023-03-11
Fix for https://github.com/The-OpenROAD-Project/OpenROAD/issues/2989

hpretl:
Simple fix just removing the `-m64` compile flag.


maliberty:
Thanks for the fix.  Please fix DCO (see the details link for instructions)

maliberty:
You can avoid the problem by using git commit **-s** in the future to sign your commits.

PR #3012, opened 2023-03-13
odb: Avoid crash on unknown cut class in cut enclosure statement

antonblanchard:


antonblanchard:
I just noticed the fix from Matt, closing.

PR #3016, opened 2023-03-14
Update to latest OpenSTA 

openroadie:
Latest OpenSTA code + regression ok file update. PSM and RSZ regressions are passing


openroadie:
@maliberty @vvbandeira : All my commits are signed off as far as I can tell. So why is DCO failing here?

maliberty:
Your last two are unsigned.

maliberty (approved):

maliberty:
I waived the DCO but please remember it in the future.  I just use a git alias "ci = commit -s" to avoid thinking about it.

openroadie:
Thanks @maliberty. Much appreciated. Aliasing is done :).

PR #3018, opened 2023-03-15
drt/dr: Always apply via spacing to dst pin check

antonblanchard:
There is code in getEstCost() to add a penalty to vias placed too close to the destination pin, but it is prevented from running until iteration 30, and then only in non rip up iterations.

This helped a test case with fat vias, where it went from taking 3 iterations to 1 iteration to become drc clean.

### DIFF
--- a/src/drt/src/dr/FlexGridGraph_maze.cpp
+++ b/src/drt/src/dr/FlexGridGraph_maze.cpp
@@ -207,39 +207,35 @@ frCost FlexGridGraph::getEstCost(const FlexMazeIdx& src,
   // avoid propagating to location that will cause forbidden via spacing to
   // boundary pin
   if (dstMazeIdx1 == dstMazeIdx2 && gridZ == dstMazeIdx1.z()) {
-    if (drWorker_ && drWorker_->getDRIter() >= 30
-        && drWorker_->getRipupMode() == 0) {
-      auto layerNum = (gridZ + 1) * 2;
-      auto layer = getTech()->getLayer(layerNum);
-      if (layer->isUnidirectional()) {
-        bool isH = (layer->getDir() == dbTechLayerDir::HORIZONTAL);
-        if (isH) {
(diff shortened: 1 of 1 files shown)


maliberty:
@antonblanchard have you tested this change on the public test cases in ORFS?

antonblanchard:
> @antonblanchard have you tested this change on the public test cases in ORFS?

I can run some of the ORFS test cases, but running everything has proven problematic. If I run all tests serially it takes forever, if I run some in parallel I run out of memory (even with 64GB).

I know we've talked about it before, but having a way to run A/B ORFS comparisons using the OR infrastructure would be a great addition.

maliberty:
I started a secure CI

maliberty:
That infrastructure is in the works still.

maliberty:
So far I see:
microwatt
[ERROR] finish__timing__setup__ws fail test: -2.17081 >= -1.76

asap7/ethmac
[ERROR] finish__timing__drv__max_slew_limit fail test: -1.6743 >= -1.67

antonblanchard:
> That infrastructure is in the works still.

I might work on a script today that runs everything in parallel except detailed routing. That should reduce the time taken to run ORFS locally a bunch.

maliberty:
https://www.gnu.org/software/parallel/ is a handy tool for these sorts of tasks

maliberty:
Just those two above among the public designs.  Nothing on the private side (2 largest still running).

antonblanchard:
> So far I see: microwatt [ERROR] finish__timing__setup__ws fail test: -2.17081 >= -1.76
> 
> asap7/ethmac [ERROR] finish__timing__drv__max_slew_limit fail test: -1.6743 >= -1.67

Could this also have been caused by the STA update?

maliberty:
Possibly.  @vvbandeira is trying to get things back in order today as they gotten messy.

maliberty (approved):

maliberty:
Passes secure CI

Issue #3019, opened 2023-03-15
grt creates wrong direction routes


antonblanchard:
### Describe the bug

The attached test case shows M5 guides that go in the wrong direction. One example is the top left pin, which spans 3 gcells horizontally before switching to M6 (not displayed).

I'm guessing the pin extension code is buggy on corner cells.

![grt-wrong-dir](https://user-images.githubusercontent.com/2029971/225270594-cb2f42c8-b1eb-4fe8-b1a5-3394ce21aa01.png)

### Expected Behavior

Global guides should only go in the preferred direction on unidirectional layers.

### Environment

```shell
Git commit: edcd14ffec5d232d98e8b0c34315b1ce233b1af6
os: Fedora Linux 37 (Workstation Edition)
cmake version 3.26.0-rc5
```

### To Reproduce

Test case: [grt-bad-direction.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10978278/grt-bad-direction.tar.gz)

PR #3024, opened 2023-03-16
grt: update IO pin edge

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3019


src/grt/src/Pin.cpp
```diff
@@ -95,10 +95,17 @@ void Pin::determineEdge(const odb::Rect& bounds,
                         const odb::Point& pin_position,
                         const std::vector<odb::dbTechLayer*>& layers)
 {
+  odb::Point upper_right = pin_position;
+  // for IO pins, use the position that touches the die edge to detect the edge
+  // it is placed
+  if (is_port_) {
+    const int layer = layers_.back();
+    upper_right = boxes_per_layer_[layer].back().ur();
+  }
```
maliberty, line 104: Could something similar happen with a macro pin?

eder-matheus: Yes, I've included macro pins in this check. My only concern is about the pin shapes of a macro pin. Can we have a macro pin with multiple pin shapes in different layers? Do we have something like this in our designs?

maliberty: You can have a multi-layer pin in LEF (iterm or bterm).  I don't know if we have any test cases of such

eder-matheus: I've updated the code to iterate over all the pin shapes, and choose the pin edge in a voting system. I believe it will be more accurate than just getting the top layer's shape.

maliberty (approved):

PR #3025, opened 2023-03-17
pad: add warning message when creating new bterms

gadfort:


src/pad/src/ICeWall.cpp
```diff
@@ -157,9 +157,14 @@ void ICeWall::makeBTerm(odb::dbNet* net,
                         odb::dbTechLayer* layer,
                         const odb::Rect& shape) const
 {
-  odb::dbBTerm* bterm = odb::dbBTerm::create(net, net->getConstName());
+  odb::dbBTerm* bterm = net->get1stBTerm();
   if (bterm == nullptr) {
-    bterm = net->get1stBTerm();
+    logger_->warn(utl::PAD, 33, "Creating block terminal: {}", net->getName());
```
QuantamHD, line 162: I think as stated the message reads a bit too positive. What do you think about this rewording?

"Could not find a block terminal associated with net: {}, creating now."

gadfort: Sure, I can reword to as you are suggesting

maliberty (approved):

PR #3027, opened 2023-03-17
drt/gc: Don't iterate through targetObjs_ in initDesign_skipObj

antonblanchard:
Use find() instead of iterating through the set.

### DIFF
--- a/src/drt/src/gc/FlexGC_init.cpp
+++ b/src/drt/src/gc/FlexGC_init.cpp
@@ -139,32 +139,34 @@ bool FlexGCWorker::Impl::initDesign_skipObj(frBlockObject* obj)
   if (targetObjs_.empty()) {
     return false;
   }
+
   auto type = obj->typeId();
-  for (auto targetObj : targetObjs_) {
-    switch (targetObj->typeId()) {
-      case frcInst: {
-        if (type == frcInstTerm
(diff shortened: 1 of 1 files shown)


maliberty (approved):

src/drt/src/gc/FlexGC_init.cpp
```diff
@@ -139,32 +139,34 @@ bool FlexGCWorker::Impl::initDesign_skipObj(frBlockObject* obj)
...
-        logger_->error(
-            DRT, 40, "FlexGCWorker::initDesign_skipObj type not supported.");
+      return true;
+    }
+    case frcBTerm: {
+      auto term = static_cast<frTerm*>(obj);
+      if (targetObjs_.find(term) != targetObjs_.end()) {
+        return false;
+      }
+      return true;
```
osamahammad21, line 164: why do you need to check for type and static cast?

rovinski: `frBTerm` and `frInstTerm` used to both be part of `frTerm` and so this code was likely split when the split in the db model happened. `frInstTerm` still requires a cast to call `getInst()` but it looks like `frBTerm` does not. So yes, this cast should be safe to remove.

PR #3032, opened 2023-03-18
drt/ta: Rename wlen to nextIrouteDir

antonblanchard:
It's completely unclear what wlen is, so rename it and add some comments explaining what it does.


maliberty (approved):
Thanks for investigating

PR #3034, opened 2023-03-18
Update flow tutorial

akash-ambekar:
Added Git Clean Command

### DIFF
--- a/docs/tutorials/index.md
+++ b/docs/tutorials/index.md
@@ -1,5 +1,5 @@
 # Tutorials
-
+git clean -xdi
 ## OpenROAD Flow Scripts Tutorial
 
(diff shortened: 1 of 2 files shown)


docs/tutorials/index.md
```diff
@@ -1,5 +1,5 @@
 # Tutorials
-
+git clean -xdi
```
maliberty, line 2: This command is can result in loss of data if not careful.  What problem is being solved by adding this?

vijayank88: Depending on the OS it will clean `results` /`reports`/`logs` of previous runs. 
User must be careful, while selecting the option. So it is not recommended to include here.

PR #3041, opened 2023-03-19
drt: use dx()/dy()

antonblanchard:


QuantamHD:
Are dx and dy representing width and height?

maliberty (approved):

maliberty:
> Do dx and dy represent width and height?

Yes.  Is there a concern?

QuantamHD:
I think it would be clearer if they were renamed width and height.

rovinski:
"Width" and "height" are actually ambiguous because they can mean different things depending on the orientation of the layer. dx and dy are more specific.

QuantamHD:
dx and dy feel ambiguous to me because at a glance it's not always clear to me if it represents the distance between two unrelated objects, or a measurement against itself.

Would lengthX and lengthY be appropriate in this context? I feel like it helps differentiate between distances and measurements while not assuming an axis alignment.

rovinski:
`dx()` and `dy()` are always against the underlying `ll` and `ur` points of the `Rect`. They are member functions with no arguments, so I don't think it could be inferred there is a comparison to other objects.

There are already the `minDXDY()` and `maxDXDY()` functions, which are sort of the equivalent of "width" and "length".

QuantamHD:
Hmm, I still disagree, but this feels like the wrong forum to discuss this. Perhaps a discussion or issue instead?

maliberty:
I agree that the discussion is orthogonal to this change.  @QuantamHD please open an issue with your thoughts.

PR #3056, opened 2023-03-21
drt: Ensure vias conform to WIDTHTABLE constraints

antonblanchard:


src/drt/src/io/io_parser_helper.cpp
```diff
@@ -44,6 +44,88 @@ using namespace boost::polygon::operators;
...
+{
+  frVia via(viaDef);
+  const auto layer1Box = via.getLayer1BBox();
+  const auto layer2Box = via.getLayer2BBox();
+  const auto layer1Num = viaDef->getLayer1Num();
+  const auto layer2Num = viaDef->getLayer2Num();
+  const auto dbLayer1 = tech_->getLayer(layer1Num)->getDbLayer();
+  const auto dbLayer2 = tech_->getLayer(layer2Num)->getDbLayer();
+
+  // We ignore vias with default width pads because they wont cause width table
```
rovinski, line 98: Is "pads" the right term here? Do you mean "enclosure metals"?

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -44,6 +44,88 @@ using namespace boost::polygon::operators;
 
 using Rectangle = boost::polygon::rectangle_data<int>;
 
+bool io::Parser::validWidth(odb::dbTechLayer* layer,
+                            int width,
```
maliberty, line 48: const method and width is a const argument

isValidWidth makes more sense for a predicate

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -44,6 +44,88 @@ using namespace boost::polygon::operators;
...
+    // The width can be equal to or greater than the last entry
+    if (width >= last_width) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+bool io::Parser::viaWidthsValid(frViaDef* viaDef)
```
maliberty, line 88: isViaWidthsValid

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -44,6 +44,88 @@ using namespace boost::polygon::operators;
...
+      && (!isLayer2Fat
+          || (validWidth(
+                  dbLayer2, layer2Box.dy(), odb::dbTechLayerDir::HORIZONTAL)
+              && validWidth(
+                  dbLayer2, layer2Box.dx(), odb::dbTechLayerDir::VERTICAL)))) {
+    return true;
+  }
+
+  return false;
+}
```
maliberty, line 127: just return the expression, no need for

```
if (...) {
  return true;
}
return false;
```

maliberty (changes requested):

maliberty:
Replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/3059

PR #3057, opened 2023-03-21
drt: implement access for pins above max routing layers

eder-matheus:


src/drt/test/top_level_term.defok
```diff
@@ -0,0 +1,57 @@
...
+    - clk + NET clk + DIRECTION INPUT + USE SIGNAL
+      + PORT
+        + LAYER met5 ( -1250 -800 ) ( 1250 800 )
+        + PLACED ( 1250 155030 ) N ;
+END PINS
+NETS 1 ;
+    - clk ( PIN clk ) ( clkbuf_0_clk A ) ( clkbuf_2_3__f_clk A ) ( clkbuf_2_2__f_clk A ) ( clkbuf_2_1__f_clk A ) + USE SIGNAL
+      + ROUTED met3 ( 1250 155030 ) M3M4_PR
+      NEW met4 ( 1250 155030 ) M4M5_PR
+      NEW met3 ( 1440 155030 0 ) ( 6960 * )
```
maliberty, line 37: Why doesn't the m3 end at the same coordinate as the up vias?

src/drt/src/db/obj/frBTerm.h
```diff
@@ -116,10 +116,14 @@ class frBTerm : public frTerm
...
   }
 
+  void setIsAboveTopLayer(bool isAboveTopLayer) { isAboveTopLayer_ = isAboveTopLayer; }
+  const bool isAboveTopLayer() const { return isAboveTopLayer_; }
+
  protected:
   frBlock* block_;
   std::vector<std::unique_ptr<frBPin>> pins_;  // set later
   frNet* net_;
+  bool isAboveTopLayer_;
```
maliberty, line 126: ```bool isAboveTopLayer_ = false;``` avoids the need to repeat it in each ctor.

src/drt/src/io/io.cpp
```diff
@@ -904,14 +905,49 @@ void io::Parser::setBTerms(odb::dbBlock* block)
...
+                  }
+                }
+              }
+            }
+          }
+        }
+        frCoord xl = useViaBox ? viaBox.xMin() : box->xMin();
+        frCoord yl = useViaBox ? viaBox.yMin() : box->yMin();
+        frCoord xh = useViaBox ? viaBox.xMax() : box->xMax();
+        frCoord yh = useViaBox ? viaBox.yMax() : box->yMax();
```
maliberty, line 946: It seems simpler to just have a Rect to fill in rather than building it at the end:

```
Rect bbox = box->getBox();
...
if (...) {
  bbox = viaBox;
}
```

src/drt/src/TritonRoute.cpp
```diff
@@ -976,6 +953,65 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+{
+  odb::dbTech* tech = db_->getTech();
+  odb::dbSet<odb::dbTechVia> vias = tech->getVias();
+  std::map<int, odb::dbTechVia*> default_vias;
+  for (int i = top_layer_idx; i <= bterm_bottom_layer_idx; i++) {
+    for (odb::dbTechVia* via : vias) {
+      if (via->getBottomLayer()->getRoutingLevel() == i) {
+        default_vias[i] = via;
+        break;
+      }
```
maliberty, line 968: Grabbing any via on the right layer may not work well.  At a minimum you should take a default via but even then we may need to add more checks (eg https://github.com/The-OpenROAD-Project/OpenROAD/pull/3056)

src/drt/src/TritonRoute.cpp
```diff
@@ -976,6 +953,65 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+    }
+  }
+
+  // get bterm access points
+  std::vector<odb::dbAccessPoint*> access_points;
+  odb::Rect pin_rect;
+  for (odb::dbBPin* bpin : bterm->getBPins()) {
+    const std::vector<odb::dbAccessPoint*>& bpin_pas = bpin->getAccessPoints();
+    access_points.insert(
+        access_points.begin(), bpin_pas.begin(), bpin_pas.end());
```
maliberty, line 978: Is there a reason to front insert?  It will be slower on a vector

src/drt/src/TritonRoute.cpp
```diff
@@ -976,6 +953,65 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+
+  // set the via position as the first AP in the same layer of the bterm
+  odb::dbTechLayer* bterm_bottom_tech_layer
+      = tech->findRoutingLayer(bterm_bottom_layer_idx);
+  odb::Point via_position;
+  if (!access_points.empty()) {
+    for (const auto& ap : access_points) {
+      if (ap->getLayer()->getRoutingLevel()
+          == bterm_bottom_tech_layer->getRoutingLevel()) {
+        via_position = ap->getPoint();
```
maliberty, line 990: break?

src/drt/src/TritonRoute.cpp
```diff
@@ -976,6 +953,65 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+      }
+    }
+  } else {
+    via_position = odb::Point(pin_rect.xCenter(), pin_rect.yCenter());
```
maliberty, line 994: Does this code assume you did pin access before?

src/drt/src/TritonRoute.cpp
```diff
@@ -976,6 +953,65 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+  }
+
+  // insert the vias from the top routing layer to the bterm bottom layer
+  odb::dbNet* net = bterm->getNet();
+  odb::dbWire* wire = net->getWire();
+  if (wire == nullptr) {
```
maliberty, line 1000: If there is an existing route then this is dangerous as you are erasing it.  It seems like an error given the intended use model.

src/drt/src/TritonRoute.cpp
```diff
@@ -976,6 +953,65 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+    wire = odb::dbWire::create(net);
+  }
+  odb::dbWireEncoder wire_encoder;
+  wire_encoder.begin(wire);
+  for (int layer_idx = top_layer_idx; layer_idx < bterm_bottom_layer_idx;
+       layer_idx++) {
+    odb::dbTechLayer* tech_layer = tech->findRoutingLayer(layer_idx);
+    wire_encoder.newPath(tech_layer, odb::dbWireType::ROUTED);
```
maliberty, line 1008: Each via doesn't need to be in a separate path as they are stacked

src/drt/src/TritonRoute.tcl
```diff
@@ -456,4 +456,51 @@ proc check_drc { args } {
   drt::check_drc_cmd $output_file $x1 $y1 $x2 $y2
 }
 
+sta::define_cmd_args "process_ports_above_top_layer" {
+    [-top_routing_layer layer]
```
maliberty, line 460: This is mandatory so drop the []

src/drt/src/TritonRoute.tcl
```diff
@@ -456,4 +456,51 @@ proc check_drc { args } {
...
+    keys { -top_routing_layer } \
+    flags{}
+
+  if { [info exists keys(-top_routing_layer)] } {
+    set top_routing_layer $keys(-top_routing_layer)
+  } else {
+    set top_routing_layer ""
+  }
+
+  if { !($top_routing_layer eq "") } {
```
maliberty, line 474: Just reduce this to a check and an error if unset

src/drt/src/TritonRoute.tcl
```diff
@@ -456,4 +456,51 @@ proc check_drc { args } {
...
+    set tech [ord::get_db_tech]
+    set block [ord::get_db_block]
+
+    set top_tech_layer [$tech findLayer $top_routing_layer]
+
+    if { $top_tech_layer == "NULL" } {
+      utl::error GRT 5 "Layer $top_routing_layer not found."
```
maliberty, line 481: Give it a unique id

maliberty (changes requested):

src/drt/src/TritonRoute.cpp
eder-matheus: The `else` part intends to cover the case where we don't have PA information on the bterms.

src/drt/test/top_level_term.defok
eder-matheus: DRT chooses another position for the met3 via shape, while the vias are placed using the center of the pin shape:
image

src/drt/src/TritonRoute.cpp
eder-matheus: It's not clear how to get the default vias. To get it from DRT, it is necessary to run `initDesign` first, but my new command runs before it is initialized.

We also have this process on grt, but I think this method doesn't fit here:

```
for (odb::dbTechVia* via : vias) {
    odb::dbStringProperty* prop
        = odb::dbStringProperty::find(via, "OR_DEFAULT");

    if (prop == nullptr) {
      continue;
    } else {
      debugPrint(logger_,
                 utl::GRT,
                 "l2v_pitch",
                 1,
                 "Default via: {}.",
                 via->getConstName());
      default_vias[via->getBottomLayer()->getRoutingLevel()] = via;
    }
  }
```

eder-matheus: Actually, I think it can be simplified. Perhaps we can drop this new command and just run the stack vias process inside drt, without user input. It would be simpler for users and for the implementation, as it can be executed after defining the default vias.

eder-matheus:
@maliberty I think I'm done with the fixes in this PR, and it's ready to merge. I could simplify the code and stack the vias during DRT, before parsing the whole design. I think I've addressed all your comments as well.

src/drt/src/TritonRoute.i
```diff
@@ -238,4 +238,12 @@ void check_drc_cmd(const char* drc_file, int x1, int y1, int x2, int y2)
   auto* router = ord::OpenRoad::openRoad()->getTritonRoute();
   router->checkDRC(drc_file, x1, y1, x2, y2);
 }
+
+void stack_vias(odb::dbBTerm* bterm,
```
maliberty, line 242: Is this just for testing or is it obsolete?  I don't any tcl caller.

eder-matheus: Obsolete. I just removed it.

src/drt/src/TritonRoute.cpp
```diff
@@ -986,6 +964,110 @@ void TritonRoute::checkDRC(const char* filename, int x1, int y1, int x2, int y2)
...
+  }
+
+  // get bterm access points
+  std::vector<odb::dbAccessPoint*> access_points;
+  odb::Rect pin_rect;
+  for (odb::dbBPin* bpin : bterm->getBPins()) {
+    const std::vector<odb::dbAccessPoint*>& bpin_pas = bpin->getAccessPoints();
+    access_points.insert(access_points.end(), bpin_pas.begin(), bpin_pas.end());
+    pin_rect = bpin->getBBox();
+  }
```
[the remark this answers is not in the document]
osamahammad21: I might be wrong, so could you please modify the test case to have pin access?

osamahammad21 (changes requested):

src/drt/src/io/io.cpp
```diff
@@ -904,14 +906,49 @@ void io::Parser::setBTerms(odb::dbBlock* block)
...
+                      finalLayerNum = layerNum;
+                      termIn->setIsAboveTopLayer(true);
+                      break;
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
```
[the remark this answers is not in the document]
eder-matheus: Your comment made me notice that we can have a bug here. Specifically, when we have a net with multiple pins above the max routing layer, this code is not assigning the correct via box to the pins. I believe that this part **must be** dependent on the port itself to work properly. I'm updating it now.

src/drt/src/TritonRoute.cpp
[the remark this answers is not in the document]
eder-matheus: You're right; even when running pin_access before detailed_route, this function uses the pin center to place the vias. It's ok to use the center of the pin, since the use case of this new functionality is for the bumps in the upper layers, which usually have a large area. We can drop this part of the code and always use the pin center. @maliberty, do you agree with this approach?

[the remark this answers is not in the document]
maliberty: works for me

src/drt/src/io/io.cpp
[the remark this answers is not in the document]
osamahammad21: I don't think this part is handled. you are still duplicating the same pin fig for each port that intersects with the via box. you should add the pinfig of the viabox on the top routing layer only once.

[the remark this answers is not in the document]
eder-matheus: I think now it's handled. The code only iterates over all the bpins when the layer is below the max routing layer.

eder-matheus:
@maliberty @osamahammad21 Is there anything else I need to change in this PR?

maliberty:
@osamahammad21 please give a final review and approve if satisfied.  I have no further comments

osamahammad21 (approved):
@eder-matheus Could you please check the ci failure?

eder-matheus:
> @eder-matheus Could you please check the ci failure?

Not sure what caused the failure, but all the tests are passing. I think the pipeline for this run was killed before it completes.

eder-matheus:
@maliberty I think I need your approval too because you have requested changes.

maliberty (approved):

PR #3058, opened 2023-03-22
drt/ta: Fix and comment via to die edge adjustments

antonblanchard:
On unidrectional layers, assignIroute_availTracks() avoids tracks that are too close to the die edges to place a via. Add a comment and update the code to handle both up and down vias, since we might use either.


antonblanchard:
> A track could still be useful even if only one of up/down is available.

Yep, but since this is likely to just remove 1 track from the edge of the die I wasn't sure how much effort we should put into it. Should we just take the best case of up and down vias?

maliberty:
I'm wondering if it is worth changing at all.  It would have to be verified and the difference is likely to be small.  What motivated this PR?

antonblanchard:
> I'm wondering if it is worth changing at all. It would have to be verified and the difference is likely to be small. What motivated this PR?

Just uncommented code with an unclear purpose that took me an amount of time to understand. Can otherwise just add a comment and not change the code.

maliberty:
stale

PR #3063, opened 2023-03-22
Add RHEL requirements for building GUI

themperek:
Without this there will be no GUI.
Tested on RHEL8.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -216,8 +216,6 @@ _installRHELCleanUp() {
 
 _installRHELPackages() {
     yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
-    yum -y install http://ftp.nsc.ru/pub/centos/8-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-8-6.el8.noarch.rpm
-    yum -y install http://ftp.nsc.ru/pub/centos/8-stream/BaseOS/x86_64/os/Packages/centos-stream-repos-8-6.el8.noarch.rpm
 
     yum -y install \
         tzdata \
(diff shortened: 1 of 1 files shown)


vvbandeira:
This change actually breaks our test of RHEL using docker:
https://github.com/The-OpenROAD-Project/OpenROAD/actions/runs/4491586294/jobs/7900385085?pr=3063
Are you using some specific distribution, or maybe we are missing some other repo source?

themperek:
I may be wrong but if i run this on registered RHEL8 I am getting extra repositories namely (`rhel-8-for-x86_64-appstream-rpms`).
It seem to be related to this message:
```
This system is not registered with an entitlement server. You can use subscription-manager to register.
```
Those packages are available for example from here: https://centos.pkgs.org/8/centos-appstream-x86_64/

Feel free to close this or I can try to find a way to add those packages.

themperek:
Other way would be to check if the repository `rhel-8-for-x86_64-appstream-rpms` exit.

cbalint13:
> Other way would be to check if the repository `rhel-8-for-x86_64-appstream-rpms` exit.

Also, to enable epel-next extra there is a much simple way [see here](https://github.com/cbalint13/copr-packages/blob/main/docker-examples/Dockerfile.HDL%2BVLSI.rhel):
```
# centos extra repositories
RUN dnf -y config-manager --set-enabled powertools
RUN dnf -y install epel-release epel-next-release
```

maliberty:
Thanks @cbalint13 for the expert input

themperek:
> @themperek ,
> 
> As side note info regarding RHEL/Fedora:
> 
> * There are regular (weekly) builds on [COPR](https://copr.fedorainfracloud.org/coprs/rezso/VLSI/package/openroad)
> * Also some walk-through the full [VLSI suite](https://gist.github.com/cbalint13/1de29fbe77e8dedddd3861b5f49a1835)

That is quite nice! Thank you @cbalint13

cbalint13 (approved):

maliberty (approved):

PR #3064, opened 2023-03-22
pdn: fix spacing violations and routing through certain obstructions

gadfort:
Changes:
- Now the shape cutting uses the obstruction box instead of the shape box to ensure that spacing is honored for the new shape.
- Previously several instance types were getting filtered out from obstructions along with pins not connected, this was causing the PDN to place straps through them since it didn't know about the objects.

Test cases were updates were the new changes cased them to get modified.
@maliberty it might not be a bad idea to do a secure run on this.


maliberty:
I started a secure CI

maliberty:
The only thing I see so far (that isn't already failing on master) is:
black_parrot nangate45
[ERROR] finish__timing__drv__max_slew_limit fail test: -0.780891 >= -0.62

I'm not clear what changed but this is likely to be noise.  Still waiting on the two largest secure CI jobs.

PR #3079, opened 2023-03-25
Scan architect

fgaray:
Adding Scan Architect for DFT commands.

preview_dft:
- Performs scan replacement
- Perform scan architect.
- Shows a report of the scan chains that are going to be created.
- Performs a rollback to get to a state before scan replacement.

In Scan Architect
- We calculate the hash of each of the clock domains
- We calculate how many scan chains we need for the given max_length
- We put the biggest scan cells (in bits, for now we support only one bit cells) in the scan chains.
- We print the generated scan chains.

Output of preview_dft:

image

New commands:
- set_dft_config
- report_dft_config


fgaray:
Not sure what is the remaining action required.

maliberty:
Unit test failure, DCO

fgaray:
> Unit test failure, DCO

Thanks! I had to scroll to see it.

maliberty:
There are still two test failures.

fgaray:
So for future reference: unit tests needs to be inside record_pass_fail_tests and not record_tests, otherwise the test will fail in some platforms if the output of the test  is a colored string (like green for passed).

maliberty:
> So for future reference: unit tests needs to be inside record_pass_fail_tests and not record_tests, otherwise the test will fail in some platforms if the output of the test is a colored string (like green for passed).

Would you expand on this - I'm not sure what it means.

fgaray:
> Would you expand on this - I'm not sure what it means.

I think what was happening is that the code for the colors generated in the failing platforms was different that the one that I was getting on my machine that I copied to the .ok file. The record_tests tcl proc compares the logs files and was finding this difference making the test fail.

Looks like record_pass_fail_tests only check for pass/fail in the CPP unit test so it is more robust to changes in the .ok golden logs.

See the attached screenshot of what was being stored in the .ok file of the unit tests, note colored escaped sequence.

image

QuantamHD:
@maliberty is this ready to go?

maliberty:
> review

No review was requested.  I just commented on something that sounded interesting.  Is this ready for review?

QuantamHD:
Yes.

src/dft/src/ScanChain.hh
```diff
@@ -0,0 +1,93 @@
...
+
+#include "cells/ScanCell.hh"
+
+namespace dft {
+
+// A scan chain. This class contains all the cells that are going to be stitched
+// together into a scan chain. A scan chain can mix different clock domains
+// (Equivalent clocks) and different edges too. We are going to connect together
+// first the falling edge cells and then the rising edge to prevent
+// syncronization issues while shifting data in.
```
maliberty, line 45: typo: sync**h**ronization

src/dft/src/ScanChain.cpp
```diff
@@ -0,0 +1,124 @@
...
+  for (const auto& scan_cell : falling_edge_scan_cells_) {
+    if (current_clock_name != scan_cell->getClockDomain().getClockName()) {
+      // Change of clock domain, show the clock
+      logger->report("  {:s} ({:s}, falling)",
+                     scan_cell->getName(),
+                     scan_cell->getClockDomain().getClockName());
+    } else {
+      logger->report("  {:s}", scan_cell->getName());
+    }
+    current_clock_name = scan_cell->getClockDomain().getClockName();
```
maliberty, line 101: move this to the != block

src/dft/src/ScanChain.cpp
```diff
@@ -0,0 +1,124 @@
...
+  for (const auto& scan_cell : rising_edge_scan_cells_) {
+    if (current_clock_name != scan_cell->getClockDomain().getClockName()) {
+      // Change of clock domain, show the clock
+      logger->report("  {:s} ({:s}, rising)",
+                     scan_cell->getName(),
+                     scan_cell->getClockDomain().getClockName());
+    } else {
+      logger->report("  {:s}", scan_cell->getName());
+    }
+    current_clock_name = scan_cell->getClockDomain().getClockName();
```
maliberty, line 114: same

src/dft/src/Dft.cpp
```diff
@@ -65,6 +69,56 @@ void Dft::pre_dft()
...
+
+    // Scan Architect
+    std::unique_ptr<ScanCellsBucket> scan_cells_bucket
+        = std::make_unique<ScanCellsBucket>();
+    scan_cells_bucket->init(dft_config_.getScanArchitectConfig(),
+                            shared_scan_cells);
+
+    std::unique_ptr<ScanArchitect> scan_architect
+        = ScanArchitect::ConstructScanScanArchitect(
+            dft_config_.getScanArchitectConfig(), std::move(scan_cells_bucket));
```
maliberty, line 101: why not use an ordinary ctor here?

src/dft/src/Utils.hh
```diff
@@ -52,4 +54,38 @@ odb::dbInst* ReplaceCell(
...
+// Returns a vector of dbITerm for every clock that there is in the instance.
+// For black boxes or CTLs, we can have more than one clock
+std::vector<odb::dbITerm*> GetClockPin(odb::dbInst* inst);
+
+// Returns a sta::Clock of the given iterm
+std::optional<sta::Clock*> GetClock(sta::dbSta* sta, odb::dbITerm* iterm);
+
+// Runs the given function in a child process. We can use this function as a way
+// to rollback changes to the netlist
+void RunInForkForRollback(const std::function<void()>& fn);
```
maliberty, line 66: How will this work if you are in the GUI and the logger's output is being directed into the GUI's output pane?

src/dft/src/architect/ClockDomain.cpp
```diff
@@ -0,0 +1,68 @@
...
+  switch (config.getClockMixing()) {
+    // For NoMix, every clock domain is different
+    case ScanArchitectConfig::ClockMixing::NoMix:
+      return [](const ClockDomain& clock_domain) {
+        return std::hash<std::string>{}(clock_domain.getClockName())
+               ^ std::hash<ClockEdge>{}(clock_domain.getClockEdge());
+      };
+    default:
+      // Not implemented
+      abort();
```
maliberty, line 49: Use a logger error rather than abort()

src/dft/src/cells/ScanCell.hh
```diff
@@ -0,0 +1,67 @@
...
+  ScanCell(const std::string& name, ClockDomain clock_domain);
+  virtual ~ScanCell() = default;
+  // Not copyable or movable
+  ScanCell(const ScanCell&) = delete;
+  ScanCell& operator=(const ScanCell&) = delete;
+
+  virtual uint64_t getBits() const = 0;
+  virtual void connectSE() const = 0;
+  virtual void connectSI() const = 0;
+  virtual void connectSO() const = 0;
```
maliberty, line 57: Use full words, eg connectScanIn

src/dft/src/architect/ScanArchitect.cpp
```diff
@@ -0,0 +1,204 @@
...
+                      const std::shared_ptr<ScanCell>& rhs)
+{
+  // If they have the same number of bits, then we compare the names of the
+  // cells so they are ordered by name
+  if (lhs->getBits() == rhs->getBits()) {
+    return lhs->getName() < rhs->getName();
+  } else {
+    // Bigger elements last
+    return lhs->getBits() < rhs->getBits();
+  }
```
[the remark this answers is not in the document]
maliberty: This doesn't look resolved to me - the else is still there.

src/dft/src/architect/ScanArchitect.cpp
```diff
@@ -0,0 +1,204 @@
...
+}
+
+void SortScanCells(std::vector<std::shared_ptr<ScanCell>>& scan_cells)
+{
+  std::sort(scan_cells.begin(), scan_cells.end(), CompareScanCells);
+}
+
+}  // namespace
+
+ScanCellsBucket::ScanCellsBucket()
```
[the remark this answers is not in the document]
maliberty: not resolved

maliberty (changes requested):

maliberty:
I noticed you marked a number of clang-tidy warnings as resolved without actually resolving them.  If you disagree with the warning please say so and we can discuss.

src/dft/src/architect/ScanArchitect.cpp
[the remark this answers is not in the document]
fgaray: Should be solved now

src/dft/src/architect/ScanArchitect.cpp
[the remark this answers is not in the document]
fgaray: Should be solved now

src/dft/src/architect/ClockDomain.cpp
fgaray: Done

fgaray (commented):
Hey @maliberty , I solve some comments but saw the one where you ask about how this will work with the GUI and I wanted to generate a discussion there before continuing with the rest of the review.

Regarding the clang-tidy: I think I had an issue with the github UI and ended up running the tool in the command line. I think I now solved all of them.

Thanks for the comments!

src/dft/CMakeLists.txt
```diff
@@ -45,15 +45,28 @@ swig_lib(NAME dft
...
+    src/cells/ScanCell.cpp
+    src/cells/ScanCellFactory.cpp
 )
 
 target_include_directories(dft
   PUBLIC
     include
+  PRIVATE
+    src
+    src/architect
```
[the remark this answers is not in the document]
maliberty: In include/dft/Dft.hh you have ```#include "DftConfig.hh"``` which is in src.  This is not good as no outside client could use this header.  

You should remove src and use relative paths internally as needed.

src/dft/src/Utils.hh
maliberty: We don't have a generic roll back today.  rsz built a local one but it only works for the specific sorts of changes it makes (buffering/resizing).

As this won't work in this formulation I don't think we should include it.  You could make a blunt tool like delete all scan chains and start again.

fgaray: I think I can do something for scan replace where I keep track of the original cell and replace back the cell after preview. I will try to implement something like that and ping you once it is ready for review again. Thanks!

src/dft/CMakeLists.txt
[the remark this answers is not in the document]
fgaray: Hi @maliberty,

I created a simple rollback for Scan Replace in ScanReplace.cpp. The method name is rollbackScanReplace. This removes the dependency on forking.

Regarding the src/ path in the CMakeLists, I am not sure how to remove it. I removed the include DftConfig.hh but without that src/, I get several errors on other files for other headers inside src/ path (attached). I see other CMakeList.txt files are using something similar, for example ./dpl/CMakeLists.txt is also adding the src/ path in target_include_directories, maybe I am missing something?


image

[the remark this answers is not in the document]
maliberty: In dpl @QuantamHD has split it into dpl_lib and dpl.  Only the later has src as part of its includes.

In general we haven't used subdirs for sources the way you have.  There is no clear layering with cell and architect having mutual dependencies.

maliberty:
Is this to the point that it is ready for another review?

src/dft/CMakeLists.txt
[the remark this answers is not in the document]
fgaray: So I solved this with @QuantamHD's help: I defined several libs inside dft and include them where needed. The include of src is no longer needed.

fgaray:
Hi @maliberty ,

Yes!, this is ready for another review. Now doing the rollback of scan replace instead of forking and a refactor of the structure of the directories inside dft.

Thanks!

maliberty:
Please avoid using force push after an initial review.  Doing so breaks the 'see changes since last review' feature of GH.

fgaray:
Yeah, I forgot to sign off some commits and looks like the only way to correct that was by forcing push.

fgaray:
Hi @maliberty ,

I see what you mean with forcing push to the repo. I was trying to go and resolve some old comments but I wasn't able to see them since the commits do not exist anymore. However, I think the only missing comment to resolve is the one about "ctor" (see attached screenshot). Could you please refresh my memory what that was referring to?

image

Thanks!

maliberty:
> However, I think the only missing comment to resolve is the one about "ctor" (see attached screenshot). Could you please refresh my memory what that was referring to?

I believe that was about using ConstructScanScanArchitect which just calls a ctor.  I guess this might be an extension point in the future but it wasn't obvious what its purpose is.

maliberty:
Fwiw I have a git alias "ci = commit -s" to avoid having to remember to use -s.

fgaray:
Thanks @maliberty!,

I elaborated on the comment regarding supporting multiple scan architect algorithms in ScanArchitect.hh, that should make it more clear.

And thanks for that alias, I also saw that you can add a local git hook to auto sign your commits: https://gist.github.com/fgaray/c9c5a7e498a57237ff9c456824c1f145

That should prevent further issues from my side with force push.

Thanks!,
Felipe

maliberty (approved):
This is good progress.

PR #3087, opened 2023-03-26
drt/pa: Fix bounding box calculation in genPatterns_gc

antonblanchard:
genPatterns_gc miscalculates the bounding box used for gc, applying the lower x coordinate to all other coordinates. If we are unlucky this bounding box is huge and could contain over 100,000 objects. As a result gc wastes a lot of time initialising all the objects within the bounding box.

Fixing this significantly improves the start up time for the ispd19_test10 test case.


maliberty (approved):
Nice catch!  How did you find this?

maliberty:
@osamahammad21 FYI

antonblanchard:
> Nice catch! How did you find this?

Just trying to understand why it took so long for `ispd19_test10` to get through the pa code. The profiles showed lots of time in ` initDesign_skipObj`, even after some optimisation work. Eventually I realised there were sometimes 100,000+ objects in the gc region, which made no sense.

PR #3088, opened 2023-03-27
odb: support schema versions in code generator

osamahammad21:


maliberty (approved):

osamahammad21:
@maliberty is there anyway to skip the format check for the impl.cpp file? it's a template for jinja and cannot be formatted.

maliberty:
Add an exception for codeGenerator to https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/.github/workflows/github-actions-format-on-push.yml (there are already several)

PR #3089, opened 2023-03-27
drt/gc: Fix memory leak in checkMetalSpacing_short_obs()

antonblanchard:
We are leaking memory in checkMetalSpacing_short_obs() via gcRect rect3. There doesn't seem to be a reason for dynamically allocating this, so just move it to the stack.


antonblanchard:
I might be missing something here... does it need to live past function exit for some reason?

osamahammad21 (approved):

PR #3095, opened 2023-03-27
drt/pa: Better sizing of gc bounding box

antonblanchard:
During pin access, we currently hard wire the size of the gc bounding box. Depending on the technology, 3000 could be too big or too small.

Change this to be 5 wire pitches of the metal layer of the pin, which should be enough area to determine if we can safely access the pin.


osamahammad21 (commented):
Same could apply to genPatterns_gc.

maliberty:
The most systematic way would be to inspect all the constraints and take the maximum value that is required for any check.

antonblanchard:
> The most systematic way would be to inspect all the constraints and take the maximum value that is required for any check.

This calculation would also be useful for sizing the worker extended boxes which is currently hardwired to a different value:

```
int MTSAFEDIST = 2000;
```

maliberty (approved):
This can be refined more in the future but this looks good

PR #3098, opened 2023-03-27
DPO: disallow one site gap

mina1460:
The main idea behind this PR is:

1- Add a new command to the DPO dt script that supports detecting one site gaps violations. This command should run after the DPO is done optimizing. Earlier checks would be overwritten by the DPO anyway in the final randomization step. 
2- The returned violations can be easily shifted one site to the left, and if not possible, we try the right shift.

With one-site gaps          |  Without one-site gaps
:-------------------------:|:-------------------------:
Screenshot 2023-06-10 at 5 17 09 pm  | Screenshot 2023-06-10 at 5 19 09 pm
Detailed Improvement Results | Detailed Improvement Results
Original HPWL             7709.2 u | Original HPWL             7709.2 u
Final HPWL                7322.4 u | Final HPWL                7322.6 u
Delta HPWL                  -5.0 %  |  Delta HPWL                  -5.0 %

After the final fix with fixed blockage cells:

Screenshot 2023-06-14 at 2 05 09 am


src/dpo/src/detailed_manager.cxx
```diff
@@ -1384,6 +1384,84 @@ int DetailedMgr::checkEdgeSpacingInSegments()
...
+  violating_cells.resize(segments_.size() + 1);
+
+  std::vector<Node*> temp;
+  temp.reserve(network_->getNumNodes());
+  for (int s = 0; s < segments_.size(); s++) {
+    // To be safe, gather cells in each segment and re-sort them.
+    temp.clear();
+    for (Node* ndj : cellsInSeg_[s]) {
+      temp.push_back(ndj);
+    }
```
maliberty, line 1400: Just:
temp = cellsInSeg_[s];

src/dpo/src/detailed_manager.cxx
```diff
@@ -1384,6 +1384,84 @@ int DetailedMgr::checkEdgeSpacingInSegments()
...
+    std::sort(temp.begin(), temp.end(), compareNodesX());
```
maliberty, line 1401: compareNodesX is not a total order so it would be better to use stable_sort or to make it into one (ie handle tie breaking)

src/dpo/src/detailed_manager.cxx
```diff
@@ -1384,6 +1384,84 @@ int DetailedMgr::checkEdgeSpacingInSegments()
...
+    // it doesn't have to be the one directly before the current cell as
+    // there might be two cells on the same x but different y.
+    // if the difference between the last X and the current cell's X is equal to
+    // 1-site then we might have a violation. we still need to check the
+    // overlaps in the ys
+    // to do this efficiently we can simply loop and check for overlaps in the
+    // ys
+
+    auto isInRange = [](int value, int min, int max) -> bool {
+      return value >= min && value <= max;
```
maliberty, line 1413: purely stylistic but I prefer 

min <= value  && value <= max

as it reads more like the mathematical

min <= value <= max

src/dpo/src/detailed_manager.cxx
```diff
@@ -1384,6 +1384,84 @@ int DetailedMgr::checkEdgeSpacingInSegments()
...
+
+    std::vector<Node*> cellsAtLastX(1, temp[0]);
+
+    for (int node_idx = 1; node_idx < temp.size(); node_idx++) {
+      if (temp[node_idx]->getRight() != lastNode->getRight()) {
+        // we have a new X
+        // check if the difference between the last X and the current X is
+        // almost equal to 1-site (using tolerance)
+        if (abs(temp[node_idx]->getLeft() - lastNode->getRight() - one_site_gap)
+            < tolerance) {
```
maliberty, line 1435: Aren't these all integers?  I'm not clear why a tolerance is needed.

src/dpo/src/detailed_manager.cxx
```diff
@@ -85,7 +85,7 @@ DetailedMgr::DetailedMgr(Architecture* arch,
   targetUt_ = 1.0;
 
   // For generating a move list...
-  moveLimit_ = 10;
+  moveLimit_ = 100;
```
maliberty, line 88: What is the implication of this change?  How does it relate to 1-site gaps?

mina1460: assume there are 100 consequetive cells on the left and then a one-site gap followed by another 100 touching cells. We have to shift either of the 100 cells to the left or to the right.

If the move limit was 10, test cases like ibex failed because it needed to move more than 10 cells sometimes. If the move limit is small, it sees that the shift would result in an overlap so it will skip this move, which would result in a one-site gap violation that we didn't fix. 

Relaxing this to a 100 doesn't affect anything that doesn't need to move all of these cells. It only allows bigger test cases to fail less because the probability of having 100 consequetive cells is lower than having 10.

src/dpo/src/detailed_manager.cxx
maliberty: This one is still open

mina1460: resolved

src/dpo/src/detailed_manager.cxx
maliberty: We discussed having a larger threshold specifically for one-site

maliberty (approved):

PR #3100, opened 2023-03-28
ppl: fix segfault in error message

oharboe:


src/ppl/src/HungarianMatching.cpp
```diff
@@ -179,7 +179,7 @@ void HungarianMatching::assignMirroredPins(IOPin& io_pin,
                    "valid position for pin placement.",
                    mirrored_pos.getX(),
                    mirrored_pos.getY(),
-                   layer->getName());
+                   layer ? layer->getName() : "NA");
```
gadfort, line 182: How would we have gotten to this print statement with an invalid layer? Seems like there is an issue much earlier. Since a pin position has to be on a metal layer

maliberty: You could imagine a scenario where you have enough slots but they are not grouped together so that you can't place the group.

oharboe: I looked at the floorplan, at that point the group of pins are ca 30% or less of the edge, so I would say there is enough space. @eder-matheus and I have a debug call tomorrow to try to create a testcase.

maliberty (approved):
This is fine for what it is but we still need to find the real issue.

oharboe:
> This is fine for what it is but we still need to find the real issue.

Agreed.

PR #3104, opened 2023-03-28
drt: Add DEBUG_DRT_UNDERFLOW option

antonblanchard:
The -DDEBUG_DRT_UNDERFLOW=ON option checks for underflow in the drt cost calculations. It does bloat memory quite a lot, so isn't enabled by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,8 +69,6 @@ option(USE_SYSTEM_OPENSTA "Use system shared OpenSTA library" OFF)
 # Allow to use external shared abc libraries
 option(USE_SYSTEM_ABC "Use system shared ABC library" OFF)
 
-option(DEBUG_DRT_UNDERFLOW "Check for underflow in drt cost calculations" OFF)
-
 project(OpenROAD VERSION 1
   LANGUAGES CXX
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,8 @@ option(USE_SYSTEM_OPENSTA "Use system shared OpenSTA library" OFF)
 # Allow to use external shared abc libraries
 option(USE_SYSTEM_ABC "Use system shared ABC library" OFF)
 
+option(DEBUG_DRT_UNDERFLOW "Check for underflow in drt cost calculations" OFF)
+
 project(OpenROAD VERSION 1
   LANGUAGES CXX
 )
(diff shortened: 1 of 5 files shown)


src/drt/src/dr/FlexGridGraph.h
```diff
@@ -1029,12 +1039,21 @@ class FlexGridGraph
 
   frUInt4 addToByte(frUInt4 augend, frUInt4 summand)
   {
+    frUInt4 result = augend + summand;
+#ifndef DEBUG_DRT_UNDERFLOW
     constexpr frUInt4 limit = (1u << 8) - 1;
-    return std::min(augend + summand, limit);
+    result = std::min(result, limit);
+#endif
+    return result;
```
maliberty, line 1047: You could still have overflow for 16 bits

src/drt/src/dr/FlexGridGraph.h
```diff
@@ -892,8 +893,15 @@ class FlexGridGraph
...
  private:
   frTechObject* tech_;
+  Logger* logger_;
   FlexDRWorker* drWorker_;
   FlexDRGraphics* graphics_;  // owned by FlexDR
+			      //
+#ifdef DEBUG_DRT_UNDERFLOW
+#define COST_BITS 16
+#else
+#define COST_BITS 8
```
maliberty, line 903: Use constexpr int rather than #define.

maliberty (changes requested):

src/drt/src/dr/FlexGridGraph.h
maliberty: That's fine

src/CMakeLists.txt
```diff
@@ -122,6 +122,10 @@ else()
   set(ABC_LIBRARY "libabc")
 endif()
 
+if(DEBUG_DRT_UNDERFLOW)
+  add_definitions(-DDEBUG_DRT_UNDERFLOW=1)
+endif()
```
maliberty, line 127: Can this be moved in drt and use target_compile_definitions since this isn't affecting anything else?

CMakeLists.txt
```diff
@@ -69,6 +69,8 @@ option(USE_SYSTEM_OPENSTA "Use system shared OpenSTA library" OFF)
 # Allow to use external shared abc libraries
 option(USE_SYSTEM_ABC "Use system shared ABC library" OFF)
 
+option(DEBUG_DRT_UNDERFLOW "Check for underflow in drt cost calculations" OFF)
```
maliberty, line 72: Can this be moved to drt?

maliberty (approved):

maliberty:
I'm not sure where the conflict came from but please take a look.

PR #3105, opened 2023-03-29
setup: a safe, no choices to be made setup that safely installs dependencies

oharboe:
@yupferris The idea is to remove choices and sharp edges from install process.

Just run:

`sudo etc/setup.sh`

Then, to set up environment variables, run:

`. env.sh`

From ORFS, a local build is then made by running:

`./build_openroad.sh --local`

Then, after https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/928 merged, you can run the commands to set up env vars to yosys and OpenROAD binaries:

```
. env.sh
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ cmake-build-debug/
 
 coverage-output
 messages.txt
+dependencies/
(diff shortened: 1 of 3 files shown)


yupferris:
looks good and helpful to me at least

oharboe:
Moved everything into https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/928

PR #3107, opened 2023-03-29
Allow Disabling Tests

donn:
Adds a CMake boolean option `ENABLE_TESTS` (`ON` by default) that will disable tests if defined as `OFF`. Speeds up compile times significantly and allows compiling in environments where GoogleTest support is iffy, e.g. Nix.

### DIFF
--- a/src/rcx/src/CMakeLists.txt
+++ b/src/rcx/src/CMakeLists.txt
@@ -95,39 +95,40 @@ messages(
 ############################################################
 # Unit testing
 ############################################################
-enable_testing()
+if(ENABLE_TESTS)
+  enable_testing()
 
-add_executable(rcxUnitTest
-  ${PROJECT_SOURCE_DIR}/test/ext2dBoxTest.cpp
-)
-
-target_include_directories(rcxUnitTest
-  PRIVATE
-  ${PROJECT_SOURCE_DIR}/src
-  ${OPENROAD_HOME}/include
-)
-
-target_link_libraries(rcxUnitTest
-  rcx_lib
-)
-
-# Use the shared library if found.  We need to pass this info to
-# the code to select the corresponding include.  Using the shared
-# library speeds up compilation.
-if (Boost_unit_test_framework_FOUND)
-  message(STATUS "Boost unit_test_framework library found")
-  target_link_libraries(trTest
-    Boost::unit_test_framework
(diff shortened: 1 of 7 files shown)


src/rcx/src/CMakeLists.txt
```diff
@@ -95,39 +95,40 @@ messages(
...
+  target_link_libraries(rcxUnitTest
+    rcx_lib
+  )
 
+  # Use the shared library if found.  We need to pass this info to
+  # the code to select the corresponding include.  Using the shared
+  # library speeds up compilation.
+  if (Boost_unit_test_framework_FOUND)
+    message(STATUS "Boost unit_test_framework library found")
+    target_link_libraries(rcxUnitTest
```
maliberty, line 120: Thanks for fixing this

maliberty (approved):

QuantamHD:
@donn @maliberty Why not just run `make openroad`. This feels like unnecessary complexity when you could just specify to make/ninja which targets you are interested in.

donn:
@QuantamHD It still attempts to pull GTest mid-build from the internet. In any replicable build system, including that of OpenLane 2's Nix-based one, this is considered nondeterministic and will not work.

QuantamHD:
I'm familiar with nix. This only happens if there's no system gtest available though. Has no one managed to add gtest to the nix universe yet?

QuantamHD:
I point this out, because it's very likely that tests will be added without using your DISABLE_TEST flag. I submitted a PR today that does not https://github.com/The-OpenROAD-Project/OpenROAD/pull/3109

I'd like to find a less brittle solution if you're amenable.

QuantamHD:
> If it truly can use a system gtest, this would solve the problem, but yeah, I ran into this: https://github.com/NixOS/nixpkgs/blob/160d82edc1907a5b64d463726e0e3221c1601eac/pkgs/development/tools/parsing/antlr/4.nix#LL114C2-L114C2
> 
> I can look into if it'll work properly with system gtest tomorrow.

Cool Thanks!

maliberty:
Whatever you two decide is fine with me.  I keep this or revert it if a better solution is found.

donn:
@QuantamHD Very long tomorrow but, it should be fine to remove this now.

PR #3109, opened 2023-03-29
[mpl2, par, dbSta]: Implements library separation

QuantamHD:


src/dbSta/src/CMakeLists.txt
```diff
@@ -66,9 +86,6 @@ swig_lib(NAME          dbSta
 target_sources(dbSta
   PRIVATE
     dbSta.cc
```
maliberty, line 88: This file contains things that need to be in the library.  I think you'll have to split it up.

maliberty (approved):

src/dbSta/src/CMakeLists.txt
maliberty: I hope you address this in a future PR.  It doesn't matter in the current OR app

PR #3112, opened 2023-03-30
DO NOT MERGE - Fix crash collaboration

oharboe:
@eder-matheus Added missing header definition to make it compile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ cmake-build-debug/
 
 coverage-output
 messages.txt
+dependencies/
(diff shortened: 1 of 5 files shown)


oharboe:
@eder-matheus With these changes, I get:

```
Found 0 macro blocks.
Using 1u default distance from corners.
[ERROR PPL-0082] Mirrored position (42, 0) at layer NA is not a valid position for pin placement.
Error: io_placement.tcl, 13 PPL-0082
```

eder-matheus:
@oharboe Could you try this branch again: https://github.com/eder-matheus/OpenROAD/tree/ppl_groups?

I've made lots of changes that would conflict with this PR, so I'm keeping it in my fork for now.

PR #3114, opened 2023-03-30
deltaDebug: ignore irrelevant errors

oharboe:
fixes false negatives

### DIFF
--- a/etc/deltaDebug.py
+++ b/etc/deltaDebug.py
@@ -38,6 +38,7 @@ parser.add_argument('--error_string', type=str, help='The output that indicates 
 parser.add_argument('--step', type=str, help='Command used to perform step on the input odb file')
 parser.add_argument('--persistence', type=int, default=1, choices=[1,2,3,4,5,6], help= 'Indicates maximum input fragmentation; fragments = 2^persistence; value in [1,6]')
 parser.add_argument('--use_stdout', action='store_true', help='Enables reading the error string from standard output')
+parser.add_argument('--exit_early_on_error', action='store_true', help='Exit early on unrelated errors to speed things up, but risks exiting on false negatives.')
 parser.add_argument('--dump_def', action='store_true', help='Determines whether to dumb def at each step in addition to the odb')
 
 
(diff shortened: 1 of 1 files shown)


maliberty:
What is ERROR falsely matching?  I'm wondering if there is a way we an narrow the check rather than remove it.

oharboe:
> What is ERROR falsely matching? I'm wondering if there is a way we an narrow the check rather than remove it.

Two examples:

```
[deleted]
[*** LOG ERROR #0001 ***] [2023-03-29 14:05:24] [logger] ***argument not found***
[deleted]
[ERROR PPL-0061] Pins for set_io_pin_constraint command were not found.
```

- I don't know what the first error is about
- The second is a known harmless error, I've scaled down the design so that this pin is not present. I have a ```tcl catch``` statement around the `set_io_pin_constraint` statement to ignore that error.

QuantamHD:
The first error is a runtime issue with a log statement missing a template argument. See the following example.

`logger_->warn(UTL, 1000, "{} {} {}", "hi", "bye");`

oharboe:
@eder-matheus Probably not relevant here, but how can we find out which error doesn't have arguments specified?

eder-matheus:
> @eder-matheus Probably not relevant here, but how can we find out which error doesn't have arguments specified?

This looks trick. It is not necessarily an error message that doesn't have arguments. It can be any message. I will revise the PPL messages looking for this error.

maliberty:
A catch around an error was not anticipated.  However the common use case is that an unrelated error is failure.  I think it better to make a non-default option to continue on error instead or removing the check.

oharboe:
> The first error is a runtime issue with a log statement missing a template argument. See the following example.
> 
> `logger_->warn(UTL, 1000, "{} {} {}", "hi", "bye");`

@eder-matheus @QuantamHD  This hack removes the problem... https://github.com/The-OpenROAD-Project/OpenROAD/pull/3115

oharboe:
@maliberty So this is what we want I think, but I haven't tested it yet. If someone can vouch for it :-) it should be good to go in.

maliberty (approved):

maliberty:
I think I prefer the default to be on but it isn't that important to me.

oharboe:
> I think I prefer the default to be on but it isn't that important to me.

I prefer the default to be on, but the problem is that when it doesnt work, the user has to figure out why it isnt working and disable the option.... Default off gives the highest chance of a user succeeding, even if it takes longer.

PR #3115, opened 2023-03-30
escaping problems with Tcl logger utility functions

oharboe:
This removes bogus errors when curly or square brackets are interpreted as arguments in lower level logging formatting functions.

Needs a proper fix from someone who knows what is going on...


src/utl/src/LoggerCommon.cpp
```diff
@@ -52,28 +53,49 @@ void report(const char* msg)
...
   Logger* logger = getLogger();
-  logger->info(tool, id, msg);
+  logger->info(tool, id, escapeCurlyBraces(msg).c_str());
 }
 
 void warn(utl::ToolId tool, int id, const char* msg)
 {
   Logger* logger = getLogger();
-  logger->warn(tool, id, msg);
+  logger->warn(tool, id, escapeCurlyBraces(msg).c_str());
```
QuantamHD, line 86: This is not the preferred solution. 

Just change it to `logger->warn(tool, id, "{}", msg)`

QuantamHD (changes requested):

src/utl/src/LoggerCommon.cpp
maliberty: Same with info, error, and critical

maliberty (approved):

QuantamHD:
LGTM. @maliberty this is ready to go.

PR #3116, opened 2023-03-30
[gpl]: Improve performance of GPL by 13%

QuantamHD:
Removed binStor concept. Having a vector of Bin objects is faster due to cache locality considerations.

### DIFF
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -772,12 +772,12 @@ void BinGrid::initBins()
 
   log_->info(GPL, 29, "BinSize: {} {}", binSizeX_, binSizeY_);
 
-  size_t bin_dimensions = static_cast<size_t>(binCntX_ * binCntY_);
+  int bin_dimensions = binCntX_ * binCntY_;
   // initialize bins_ vector
   bins_.reserve(bin_dimensions);
   int x = lx_, y = ly_;
   int idxX = 0, idxY = 0;
-  for (size_t i = 0; i < bin_dimensions; i++) {
+  for (int i = 0; i < bin_dimensions; i++) {
     int sizeX = (x + binSizeX_ > ux_) ? ux_ - x : binSizeX_;
     int sizeY = (y + binSizeY_ > uy_) ? uy_ - y : binSizeY_;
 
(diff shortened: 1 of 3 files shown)


antonblanchard:
Nice speed up @QuantamHD!

src/gpl/src/nesterovBase.h
```diff
@@ -737,13 +707,13 @@ class BinGrid
   std::pair<int, int> getMinMaxIdxX(const Instance* inst) const;
   std::pair<int, int> getMinMaxIdxY(const Instance* inst) const;
 
-  const std::vector<Bin*>& bins() const;
+  std::vector<Bin>& bins();
+  const std::vector<Bin>& binsConst() const { return bins_; };
```
maliberty, line 711: Why is the non-const version now needed?  (Similarly where this API propagates upwards)

src/gpl/src/nesterovBase.cpp
```diff
@@ -773,19 +772,16 @@ void BinGrid::initBins()
 
   log_->info(GPL, 29, "BinSize: {} {}", binSizeX_, binSizeY_);
 
-  // initialize binStor_, bins_ vector
-  binStor_.resize(binCntX_ * binCntY_);
-  bins_.reserve(binCntX_ * binCntY_);
+  size_t bin_dimensions = static_cast<size_t>(binCntX_ * binCntY_);
```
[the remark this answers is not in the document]
maliberty: An int is fine - if we have 2^32 bins we have bigger problems.

maliberty (changes requested):
Surprisingly good gain!  Do you have a sense for how much was inlining vs data locality?

maliberty:
Btw, on what design(s) was the 13% measured?

QuantamHD:
medium06.tcl in gpl/test

src/gpl/src/nesterovBase.h
QuantamHD: bins is called in the non-const method `void NesterovBase::updateDensityForceBin()`

QuantamHD:
@antonblanchard @maliberty I should not that I think this test is sans `-timing_driven -congestion_driven` which tend to dominate runtime.

QuantamHD:
> Surprisingly

Mostly data locality. It was already inlining most of the lx, ly and friends.

src/gpl/src/nesterovBase.h
maliberty: I see the vector was constant but the pointers were not.  Now the vector itself must be mutable.

maliberty (approved):

PR #3130, opened 2023-04-04
ppl: place constrained groups with mirrored pins first

eder-matheus:


eder-matheus:
@oharboe Could you try this branch? It should fix the last error you're having.

src/ppl/test/add_constraint12.tcl
```diff
@@ -0,0 +1,17 @@
+# gcd_nangate45 IO placement
+source "helpers.tcl"
+read_lef Nangate45/Nangate45.lef
+read_def gcd.def
+
+set_io_pin_constraint -pin_names {resp_msg[3] resp_msg[2] resp_msg[14] req_val} -region top:0-18
```
maliberty, line 6: Does this cover the partially mirrored group case?

maliberty (approved):

src/ppl/test/add_constraint12.tcl
eder-matheus: I've added a new test covering it.

oharboe:
> @oharboe Could you try this branch? It should fix the last error you're having.

I see a different failure. I'll articulate a test-case.

PR #3134, opened 2023-04-05
grt: Correctly iterate Steiner Tree in gui

antonblanchard:
This fixes a crash when using global_route_debug -st


maliberty (approved):

maliberty:
/tmp/workspace/OpenROAD-Public_PR-3134-head/src/grt/src/fastroute/src/FastRoute.cpp:1392:15: error: unused variable ‘deg’ [-Werror=unused-variable]

     const int deg = stree_.deg;

PR #3137, opened 2023-04-06
Ignore warning on stderr

panicgh:
If the right locale is not available, the warning `sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)` is written to stderr. Since tcl's `exec` by default [treats output to stderr as failure](https://www.tcl-lang.org/man/tcl/TclCmd/exec.htm), the tcl script fails before any unit tests are executed.

### DIFF
--- a/src/dst/test/cpp_tests.tcl
+++ b/src/dst/test/cpp_tests.tcl
@@ -4,7 +4,7 @@ set test_dir [pwd]
 set openroad_dir [file dirname [file dirname [file dirname $test_dir]]]
 set tests_path [file join $openroad_dir "build" "src" "dst" "test" "cpp"]
 
-set tests_list [split [exec sh -c "find $tests_path -maxdepth 1 -name 'Test*'"] \n]
+set tests_list [split [exec -ignorestderr sh -c "find $tests_path -maxdepth 1 -name 'Test*'"] \n]
 
 foreach test $tests_list {
     set test_name [file tail $test]
--- a/src/odb/test/cpp_tests.tcl
+++ b/src/odb/test/cpp_tests.tcl
@@ -4,7 +4,7 @@ set test_dir [pwd]
 set openroad_dir [file dirname [file dirname [file dirname $test_dir]]]
 set tests_path [file join $openroad_dir "build" "src" "odb" "test" "cpp"]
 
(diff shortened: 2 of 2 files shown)


maliberty:
Thanks for the PR.  I'm concerned that this will have the effect of suppressing any real errors that might occur.  Is it possible for you to set your locale properly?

panicgh:
I might be wrong, but I think the root cause is that openroad expects the `en_US.UTF-8` locale to be available on the system, which may not always hold:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/0bacd5d5651bf429a5fe3721dc8518c5f0c7e4f2/src/Main.cc#L212-L217
The tcl test code is not executed standalone, but run from inside openroad itself, i.e. after it overrode the system's `LC_ALL` and `LANG`. A `sh -c "find ..."` would normally work just fine, but due to this specific setup, I expect the tests to fail on every system without `en_US.UTF-8`.

On an error (e.g. wrong command line argument,  file permission error, ...), I would expect `find` or `sh` to return a non-zero exit code, which would still be treated as error by the `exec` command. Therefore, I think this PR should silence only warning messages, i.e. the command succeeds but stderr was written to anyway. These may be rather few cases.

I'm a contributor to [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs/). Packages are built in isolation (no undeclared dependencies, e.g. locales, leak into the build environment) and the locales available by default are shown below. This regularly triggers upstream issues, e.g. hidden/implicit dependencies. It is probably possible to load additional locales just for the tests, but it is probably easier for us to continue applying this patch for our package even if it is not accepted here. This is fine, we just try to raise issues and maybe upstream fixes/workarounds.

```console
$ locale -a
C
C.utf8
POSIX
```

maliberty:
I would rather have a fallback for locale than suppress all errors.  I would be glad to try en_US.UTF-8 and failover to C.utf8, then C.   Instead of using setenv we could use setlocale and check the return type.  The problem we encountered before is that some locales use ',' as the decimal point and that breaks the LEF parser.

Do you want to try this change?

panicgh:
> Do you want to try this change?

Sounds good. If you ping me on the PR where you implement this, I'll rerun the tests.

PR #3138, opened 2023-04-06
Fix string formatting in tests

panicgh:
Hide the decimal point and digits after the decimal point when they are not needed.

This patch is necessary for building and packaging OpenROAD on [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs/blob/c59f6c75b60b31d2349b35ddb2777d230d02cabe/pkgs/applications/science/electronics/openroad/default.nix), which currently uses fmt 9.1.0. IIRC, fmt changed the default for printing floating point numbers to not show any decimal point and following digits, if they would be zero: https://fmt.dev/9.1.0/syntax.html

> The '#' option causes the “alternate form” to be used for the conversion. [...] For floating-point numbers the alternate form causes the result of the conversion to always contain a decimal-point character, even if no digits follow it. Normally, a decimal-point character appears in the result of these conversions only if a digit follows it.

Maybe this does not (yet) happen in your tests, because you use an older fmt. You may consider applying this patch when you update fmt, or format using the alternate form to make the decimal point and following digits explicit.

### DIFF
--- a/src/par/test/partition_gcd.ok
+++ b/src/par/test/partition_gcd.ok
@@ -9,7 +9,7 @@
 ========================================
 [INFO] Partitioning parameters**** 
 [PARAM] Number of partitions = 2
-[PARAM] UBfactor = 1.0
+[PARAM] UBfactor = 1
 [PARAM] Vertex dimensions = 1
 [PARAM] Hyperedge dimensions = 1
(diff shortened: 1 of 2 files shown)


maliberty:
Will revisit when we update fmt.

PR #3142, opened 2023-04-07
Separate the creation of dbSta from initialization.

cdleary:
Avoids dependency on OpenRoad object for creation of block-level dbSta instances.

Remove openroad object dependency from resizer.

Additional development context: we should be able to move dbSta.cc to dbSta_lib once we've abstracted the GUI dependency (and somehow addressed the OpenRoad observer). Those two things still couple it to the OpenRoad object after this change. This is an intermediate step that we felt was pretty self-contained and smaller / easier to review vs needing to "go for broke" of doing the full separation in this PR + GUI changes + OpenRoad observer changes.

### DIFF
--- a/src/dbSta/include/db_sta/MakeDbSta.hh
+++ b/src/dbSta/include/db_sta/MakeDbSta.hh
@@ -33,10 +33,6 @@
 
 #pragma once
 
-namespace odb {
-class dbDatabase;
-}
-
 namespace sta {
 class dbSta;
 }
@@ -47,6 +43,5 @@ class OpenRoad;
 
 sta::dbSta* makeDbSta();
 void deleteDbSta(sta::dbSta* sta);
-void initDbSta(OpenRoad* openroad);
 
 }  // namespace ord
--- a/src/dbSta/include/db_sta/MakeDbSta.hh
+++ b/src/dbSta/include/db_sta/MakeDbSta.hh
@@ -43,5 +43,6 @@ class OpenRoad;
 
 sta::dbSta* makeDbSta();
 void deleteDbSta(sta::dbSta* sta);
+void initDbSta(OpenRoad* openroad);
 
 }  // namespace ord
--- a/src/dbSta/src/CMakeLists.txt
+++ b/src/dbSta/src/CMakeLists.txt
@@ -85,6 +85,7 @@ swig_lib(NAME          dbSta
 
 target_sources(dbSta
   PRIVATE
+    MakeDbSta.cc
     dbSta.cc
     heatMap.cpp
 )
--- a/src/cts/src/TechChar.cpp
+++ b/src/cts/src/TechChar.cpp
@@ -755,7 +755,7 @@ void TechChar::createStaInstance()
 {
   // Creates a new OpenSTA instance that is used only for the characterization.
   // Creates the new instance based on the charcterization block.
-  openStaChar_ = sta::makeBlockSta(openroad_, charBlock_);
+  openStaChar_ = openSta_->makeBlockSta(charBlock_);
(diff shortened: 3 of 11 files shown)


maliberty:
ready for review?

QuantamHD:
Yes

src/dbSta/include/db_sta/MakeDbSta.hh
```diff
@@ -47,6 +43,5 @@ class OpenRoad;
 
 sta::dbSta* makeDbSta();
 void deleteDbSta(sta::dbSta* sta);
-void initDbSta(OpenRoad* openroad);
```
maliberty, line 50: The definition remains in MakeDbSta.cc so why is the declaration moved?

cdleary: I think there were a few factors:

- Nobody relies on the public declaration, so makes sense to remove
- We can't include the public declaration into dbSta (since we're trying to break that coupling)
- If there are multiple declarations of the same thing visible in the same scope, you get a warning from `-Wredundant-decls`

src/dbSta/src/CMakeLists.txt
```diff
@@ -85,6 +85,7 @@ swig_lib(NAME          dbSta
 
 target_sources(dbSta
   PRIVATE
+    MakeDbSta.cc
     dbSta.cc
```
maliberty, line 89: If dbSta.cc is in in dbSta and not dbSta_lib then its header dbSta.hh shouldn't be public.

src/dbSta/include/db_sta/MakeDbSta.hh
cdleary: (And to be a little more explicit about the context/reasoning, note that dbSta is just exposing its guts to the function as a friend, but the friend is not defined anywhere that it can actually see. You don't need to link to something to expose your guts via `friend`, you just need to acknowledge that it's going to exist in some namespace that will mess with your guts.)

src/dbSta/src/CMakeLists.txt
cdleary: I think we should be able to move `dbSta.cc` to `dbSta_lib` once we've abstracted the GUI dependency (and somehow addressed the OpenRoad observer). Those two things still couple it to the OpenRoad object after this change. This is an intermediate step that we felt was pretty self-contained and smaller / easier to review vs needing to "go for broke" of doing the full separation shown here plus gui plus openroad observer.

I think as a separate issue from whether `dbSta.cc` belongs in dbSta_lib (which would imply we could unit test it), `dbSta.hh` is used by a bunch of other components (which directly call methods on the dbSta object), so I think it would have to continue to be public for those purposes?

LMK if that makes sense or if I misunderstood.

maliberty: It's helpful to let me know that you view this as an intermediate step.  If you plan to continue on and complete the separation then I'm ok with this as just forward progress.

src/dbSta/include/db_sta/MakeDbSta.hh
maliberty: I'm afraid I don't follow your factors.  Why can't this follow the pattern of the other modules that have been converted?

src/dbSta/src/CMakeLists.txt
cdleary: I'll put the additional context in the PR next time of the succession of expected followup changes. Note, though, I was hoping this was a self contained change that made sense on its own even though it is towards the broader goal -- LMK if you don't feel like that's the case since that'd mean to me that some more discussion would be warranted.

maliberty (approved):

PR #3143, opened 2023-04-07
ppl: check for available contiguous slots when assigning groups to sections

eder-matheus:


src/ppl/src/IOPlacer.cpp
```diff
@@ -795,7 +795,8 @@ int IOPlacer::assignGroupToSection(const std::vector<int>& io_group,
       }
     }
     for (auto i : sortIndexes(dst)) {
-      if (sections[i].used_slots + group_size <= sections[i].num_slots) {
+      if (sections[i].used_slots + group_size
+          <= sections[i].getMaxContiguousSlots(slots_)) {
```
maliberty, line 799: Why add used_slots here?  Isn't it enough that there are enough contiguous slots for the group?

src/ppl/src/Slots.h
```diff
@@ -95,6 +106,22 @@ struct Section
   int used_slots;
   int num_slots;
   Edge edge;
+
+  int getMaxContiguousSlots(const std::vector<Slot>& slots)
+  {
+    int max_contioguous_slots = std::numeric_limits<int>::min();
```
maliberty, line 112: typo : max_contiguous_slots

src/ppl/src/Slots.h
```diff
@@ -95,6 +106,22 @@ struct Section
...
+    for (int i = begin_slot; i <= end_slot; i++) {
+      int contiguous_slots = 0;
+      while (i <= end_slot && !slots[i].blocked && !slots[i].used) {
```
maliberty, line 115: I see the blocked||used idiom in several places.  It would be nice to make a method on Slot for that.

src/ppl/src/Slots.h
```diff
@@ -95,6 +106,22 @@ struct Section
...
+        contiguous_slots++;
+        i++;
+      }
+
+      max_contioguous_slots = std::max(max_contioguous_slots, contiguous_slots);
```
maliberty, line 120: You could advance to the next free slot here.

src/ppl/src/Slots.h
```diff
@@ -95,6 +106,22 @@ struct Section
   int used_slots;
   int num_slots;
   Edge edge;
+
+  int getMaxContiguousSlots(const std::vector<Slot>& slots)
```
maliberty, line 110: This is a bit big for an inline function.

src/ppl/src/Slots.h
eder-matheus: I've decided to move to the next available slot at the beginning of the for loop to avoid missing a slot. Correct me if I'm wrong, but if we do it at the end of the loop, the `i` will be incremented in the for loop, and we will start the next iteration with `i+1`, losing one slot for the contiguous count.

src/ppl/src/Slots.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "Slots.h"
+
+namespace ppl {
+
+int Section::getMaxContiguousSlots(const std::vector<Slot>& slots)
+{
+  int max_contiguous_slots = std::numeric_limits<int>::min();
+  for (int i = begin_slot; i <= end_slot; i++) {
+    // advance to the next free slot
+    while (!slots[i].isAvailable()) {
```
maliberty, line 45: Need to add "i <= end_slot" to avoid going off the end.

maliberty (approved):

PR #3144, opened 2023-04-07
[odb]: Fixed memory leak in dbProperty

QuantamHD:
I decided to modernize the class with std::variant. The source of the leak was not freeing char* on replace. This new system relies on std::string which makes things simpler.

### DIFF
--- a/src/odb/src/db/dbProperty.cpp
+++ b/src/odb/src/db/dbProperty.cpp
@@ -49,8 +49,13 @@ namespace odb {
 template class dbTable<_dbProperty>;
 
 _dbProperty::_dbProperty(_dbDatabase*, const _dbProperty& n)
-    : _flags(n._flags), _name(n._name), _next(n._next), _owner(n._owner), _value(n._value)
-{}
+    : _flags(n._flags),
+      _name(n._name),
+      _next(n._next),
+      _owner(n._owner),
+      _value(n._value)
+{
+}
 
 _dbProperty::_dbProperty(_dbDatabase*)
 {
@@ -156,10 +161,10 @@ dbIStream& operator>>(dbIStream& stream, _dbProperty& prop)
   stream >> prop._name;
   stream >> prop._next;
   stream >> prop._owner;
-  
+
   uint boolean;
   int integer;
-  char* char_string; 
+  char* char_string;
   std::string string_property;
   double double_property;
 
--- a/src/odb/src/db/dbProperty.cpp
+++ b/src/odb/src/db/dbProperty.cpp
@@ -181,8 +181,11 @@ dbIStream& operator>>(dbIStream& stream, _dbProperty& prop)
 
     case DB_STRING_PROP:
       stream >> char_string;
-      prop._value = std::string(char_string);
-      free(char_string);
+      prop._value = "";
+      if (char_string != nullptr) {
+        prop._value = std::string(char_string);
+        free(char_string);
+      }
       break;
 
     case DB_DOUBLE_PROP:
@@ -501,7 +504,7 @@ bool dbBoolProperty::getValue()
 void dbBoolProperty::setValue(bool value)
 {
   _dbProperty* prop = (_dbProperty*) this;
-  prop->_value = value;
+  prop->_value = static_cast<uint>(value);
 }
 
 dbBoolProperty* dbBoolProperty::create(dbObject* object,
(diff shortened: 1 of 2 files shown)


src/odb/src/db/dbProperty.cpp
```diff
@@ -86,30 +88,8 @@ bool _dbProperty::operator==(const _dbProperty& rhs) const
...
-        return false;
-
-      break;
-
-    case DB_DOUBLE_PROP:
-      if (_value._double_val != rhs._value._double_val)
-        return false;
-
-      break;
+  if (_value == rhs._value) {
```
maliberty, line 91: Shouldn't this be != ?

In any case this can reduced to something like
return _value == rhs._value;

src/odb/src/db/dbProperty.h
```diff
@@ -73,13 +76,7 @@ class _dbProperty : public _dbObject
...
   uint _owner;
 
-  union
-  {
-    char* _str_val;
-    uint _bool_val;
-    int _int_val;
-    double _double_val;
-  } _value;
+  std::variant<std::string, uint, int, double> _value;
```
maliberty, line 79: While you are cleaning this up perhaps you could use bool rather than uint for a boolean property.

src/odb/src/db/dbProperty.cpp
```diff
@@ -182,21 +162,35 @@ dbIStream& operator>>(dbIStream& stream, _dbProperty& prop)
   stream >> prop._next;
   stream >> prop._owner;
 
+  uint boolean;
+  int integer;
+  char* char_string;
+  std::string string_property;
+  double double_property;
```
maliberty, line 169: Why not put these in the relevant case (ie the smallest scope)?

src/odb/src/db/dbProperty.cpp
```diff
@@ -182,21 +162,35 @@ dbIStream& operator>>(dbIStream& stream, _dbProperty& prop)
...
     case DB_INT_PROP:
-      stream >> prop._value._int_val;
+      stream >> integer;
+      prop._value = integer;
       break;
 
     case DB_STRING_PROP:
-      stream >> prop._value._str_val;
+      stream >> char_string;
+      prop._value = "";
```
maliberty, line 184: move to else block

maliberty (changes requested):

src/odb/src/db/dbProperty.cpp
QuantamHD: Not sure I agree. Seems like simpler control flow to just set the empty string up front since that's the default case.

i.e. no matter what we're setting the variant to a string.

maliberty: It is slightly inefficient to set and reset the string but I doubt it matters that much.  If you prefer it this way I don't care strongly.

antonblanchard:
I wonder if the original intent was to save memory. The union is 8 bytes and it looks like the std::variant version on gcc is 40 bytes (32 bytes for the string, + 8 bytes for the std::variant). Leaking memory is a worse issue of course :)

QuantamHD:
I definitely think that was in their minds when they wrote it, but variant probably wasn't around either leading them to pick the union option.

I think in this case the easier lifetime management is worth it for these objects.

src/odb/src/db/dbProperty.cpp
QuantamHD: Done.

src/odb/src/db/dbProperty.cpp
QuantamHD: I think I'd rather keep it as is. Feels like a deviation to just `return _value == rhs._value `

src/odb/src/db/dbProperty.cpp
```diff
@@ -135,7 +135,9 @@ dbOStream& operator<<(dbOStream& stream, const _dbProperty& prop)
 
   switch (prop._flags._type) {
     case DB_BOOL_PROP:
-      stream << std::get<uint>(prop._value);
+      // Older versions of the spec treated bools as uints
+      // retain backwards compatability
+      stream << static_cast<uint>(std::get<bool>(prop._value));
```
maliberty, line 140: You could make this schema version dependent for a bit of improved compactness.  Probably not worth the bother though.

maliberty (approved):

src/odb/src/db/dbProperty.cpp
QuantamHD: Good to know the option is there, but feels like that might be an improvement for a different PR.

PR #3154, opened 2023-04-11
fix: Latest OpenSTA

oharboe:
faster .lib extraction in generate_abstract.

In a case I tried locally, generate_abstract is 15x faster(from 15 mins to 1 minute). writing .lib is a tad faster than 15x in this case as there's other things like parsing .odb files that also take time.


oharboe:
> It appears you've set the submodule to something invalid or local: image

Ah, I guess I was using a fork that isn't set up to be used directly.

PR #3159, opened 2023-04-11
Split GUI dependencies from dbSta definition. 

cdleary:
This still does not split the OpenRoad observer object from the dbSta
class definition, so the file cannot be made private to the dbSta_lib
library at this point. A follow-on change intends to split the OpenRoad
observer dependency and make the dbSta.cc file not part of the
"application" target dbSta.

Waiting on https://github.com/The-OpenROAD-Project/OpenROAD/pull/3142 to re-diff


QuantamHD:
@maliberty I think this is ready to go

src/rsz/include/rsz/Resizer.hh
```diff
@@ -42,6 +42,7 @@
 #include "db_sta/dbSta.hh"
 #include "sta/UnorderedSet.hh"
 #include "sta/Path.hh"
+#include "gui/gui.h"
```
maliberty, line 45: Why is this needed in the public header?

cdleary: Switched to a forward decl of Gui class.

maliberty (approved):

PR #3160, opened 2023-04-11
Break Observer OpenRoad coupling

cdleary:
This makes the OpenRoadObserver not depend on the concrete
implementation of the OpenRoad object itself, so that the dbSta.cc file
can be moved to the non-application target dbSta_lib.

### DIFF
--- a/include/ord/OpenRoadObserver.hh
+++ b/include/ord/OpenRoadObserver.hh
@@ -1,0 +1,49 @@
+// Copyright 2019-2023 The Regents of the University of California, Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+#pragma once
+
+#include <functional>
+
+namespace odb {
+
+class dbBlock;
+class dbDatabase;
+class dbLib;
+class dbTech;
+
+}  // namespace odb
+
+namespace ord {
+
+class OpenRoad;
+
+// Observer interface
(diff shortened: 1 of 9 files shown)


cdleary:
@maliberty Ready to review

include/ord/OpenRoadObserver.hh
```diff
@@ -0,0 +1,49 @@
...
+
+class OpenRoad;
+
+// Observer interface
+class OpenRoadObserver
+{
+ public:
+  virtual ~OpenRoadObserver()
+  {
+    if (unregister_observer_ != nullptr) {
```
maliberty, line 30:
```suggestion
    if (unregister_observer_) {
```

src/upf/src/upf.cpp
```diff
@@ -41,6 +41,7 @@
 
 #include "db_sta/dbNetwork.hh"
 #include "db_sta/dbSta.hh"
+#include "ord/OpenRoad.hh"
```
maliberty, line 44: Why is this needed?  It seems to go against the grain of this change.

maliberty (changes requested):

src/upf/src/upf.cpp
cdleary: It was implicitly a necessary header that happened ot be dragged in by `dbSta.hh` before -- since we removed the inclusion from there we have to include what we use in this file now.

maliberty (approved):

maliberty:
The docker failures look unrelated

PR #3162, opened 2023-04-12
grt: Add -scenic_routes_report_file option

antonblanchard:
Add an option to create a report of the most scenic routes in global routing.

This is a first pass and needs work, but it allows me to quickly see where grt had congestion issues and had to route specific tracks all over the die. Some issues:

- Is "scenic route" a common enough name for this?
- Should we reuse the TritonRoute format, or use JSON?
- How should we format these reports? Right now each has their own top level entry, and unfortunately they get reverse sorted (shortest to longest). We also end up with a mess of DRC boxes making it impossible to see the design.
- What metrics do we want for flagging these routes? Below I flag any routes that are 1000 dbu larger than the Steiner tree, which is pretty arbitrary.

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -107,7 +107,8 @@ GlobalRouter::GlobalRouter()
       block_(nullptr),
       repair_antennas_(nullptr),
       heatmap_(nullptr),
-      congestion_file_name_(nullptr)
+      congestion_file_name_(nullptr),
+      scenic_routes_file_name_(nullptr)
 {
 }
 
@@ -286,6 +287,7 @@ void GlobalRouter::globalRoute(bool save_guides)
   updateDbCongestion();
 
   saveCongestion();
+  saveScenicRoutes();
 
   if (fastroute_->has2Doverflow()) {
     if (!allow_congestion_) {
@@ -1331,6 +1333,11 @@ void GlobalRouter::setOverflowIterations(int iterations)
 void GlobalRouter::setCongestionReportFile(const char* file_name)
 {
   congestion_file_name_ = file_name;
+}
+
+void GlobalRouter::setScenicRoutesReportFile(const char* file_name)
+{
+  scenic_routes_file_name_ = file_name;
 }
 
 void GlobalRouter::setGridOrigin(int x, int y)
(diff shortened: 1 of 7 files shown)


maliberty:
Scenic route is a common usage.

I would use JSON for new reports.  Eventually I think we should move entirely.

The bounding box for the net isn't especially useful as a visualization.  I don't have a better one in mind though.  Perhaps there is no marker for this violation and you have to look at the net to see anything.  It would be nice if there was a way to show the Steiner tree as well as the net.

You could have a reporting threshold flag.  The default shouldn't be in DBU as that varies by PDK.

src/grt/src/GlobalRouter.cpp
```diff
@@ -4057,4 +4064,52 @@ bool cmpById::operator()(odb::dbNet* net1, odb::dbNet* net2) const
   return net1->getId() < net2->getId();
 }
 
+void GlobalRouter::saveScenicRoutes()
+{
+  if (scenic_routes_file_name_ == nullptr) {
+    return;
+  }
+  std::ofstream out(scenic_routes_file_name_);
```
maliberty, line 4072: You should check if this succeeded

src/grt/src/GlobalRouter.cpp
```diff
@@ -4057,4 +4064,52 @@ bool cmpById::operator()(odb::dbNet* net1, odb::dbNet* net2) const
...
+
+  for (odb::dbNet* db_net : block_->getNets()) {
+    const auto sttLength = fastroute_->getSTTLength(db_net);
+    const auto grtLength = computeNetWirelength(db_net);
+    const int extraLength = dbuToMicrons(grtLength - sttLength);
+
+    lengths.emplace_back(extraLength, db_net);
+  }
+
+  std::sort(lengths.rbegin(), lengths.rend());
```
maliberty, line 4084: If you use begin/end then the order should reverse in the GUI.

src/grt/src/fastroute/src/RSMT.cpp
```diff
@@ -740,14 +740,16 @@ void FastRouteCore::gen_brk_RSMT(const bool congestionDriven,
         wl1 += sttrees_[i].edges[j].len;
     }
 
+    int thisWl = 0;
```
maliberty, line 743: net_wl is more descriptive

PR #3164, opened 2023-04-12
setup: single source of truth and user oriented documentation

oharboe:
DependencyInstaller.sh is an implementation detail of user setup procedures and Docker install scripts.

Also, remove some copy and paste from DepedencyInstaller.sh, will quickly become quickly out of date. Use single source of truth: the scripts themselves.

Major platforms are now supported and if some platform is not supported, the user will discover it and hopefully submit PRs to solve quirks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,38 +137,14 @@ transceivers, OpenPower-based Microwatt etc.
 
 ## Install dependencies
 
-For a limited number of configurations, the following script can be used
-to install dependencies.  The script `etc/DependencyInstaller.sh` supports
-Centos7, Ubuntu 20.04, Ubuntu 22.04, RHEL 8, Debian 10 and Debian 11. To correctly install the dependencies with the
-script you need to run as root or prepend `sudo` to the commands below.
+To install dependencies to develop or build and run OpenROAD locally,
+follow [this procedure](https://openroad-flow-scripts.readthedocs.io/en/latest/user/BuildLocally.html).
 
-``` shell
-./etc/DependencyInstaller.sh -help
-
-Usage: etc/DependencyInstaller.sh
-                                # Installs all of OpenROAD's dependencies no
-                                #     need to run -base or -common. Requires
-                                #     privileged access.
-                                #
-       etc/DependencyInstaller.sh -base
-                                # Installs OpenROAD's dependencies using
-                                #     package managers (-common must be
-                                #     executed in another command).
-       etc/DependencyInstaller.sh -common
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,38 +137,14 @@ transceivers, OpenPower-based Microwatt etc.
 
 ## Install dependencies
 
-For a limited number of configurations, the following script can be used
-to install dependencies.  The script `etc/DependencyInstaller.sh` supports
-Centos7, Ubuntu 20.04, Ubuntu 22.04, RHEL 8, Debian 10 and Debian 11. To correctly install the dependencies with the
-script you need to run as root or prepend `sudo` to the commands below.
```
maliberty, line 143: The list of supported platforms should be moved to user/BuildLocally

oharboe: Fixed: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/983

maliberty (approved):

README.md
```diff
@@ -137,38 +137,14 @@ transceivers, OpenPower-based Microwatt etc.
...
-                                #     with -common. This flag cannot be used
-                                #     with sudo or with root access.
-       etc/DependencyInstaller.sh -local
-                                # Installs common dependencies in
-                                #    "$HOME/.local". Only used with
-                                #    -common. This flag cannot be used with
-                                #    sudo or with root access.
-```
+To install dependencies to develop or build and run OpenROAD locally,
+follow [this procedure](https://openroad-flow-scripts.readthedocs.io/en/latest/user/BuildLocally.html).
```
vvbandeira, line 141: @oharboe, what do you think about adding a `setup.sh` to this repo instead of linking to ORFS?

With this PR, we are instructing users to:
1) clone the ORFS repo
2) install dependencies using ORFS `setup.sh`, which will install things that are not required by OR (e.g., `klayout`)
3) clone the OR repo (in the next section, `## Build`) even though OR is already cloned from ORFS.

Also, after redirecting the user to ORFS, is not clear when they should come back to this documentation.

oharboe: Yes, the flow needs work.

references rather than inlining is a known cost of single source of truth though.

oharboe: Although this does more than some users need in terms of installation, it has a simple and safe most common usecase.

The implementation scripts are available for those that need to drill down.

I wondering why normal users would care if a  hundred mb of extra disk space is used if it is fast and harmless.

PR #3177, opened 2023-04-14
fix: updated to latest OpenSTA

openroadie:


openroadie:
@maliberty Latest OpenSTA code. OpenRoad/test passed in my local build.

maliberty (approved):

Issue #3201, opened 2023-04-18
Support finer controls for row visibility in the GUI


maliberty:
### Description

We are going to support double height cells (#3099 ) which means we will have double height rows as well as single height rows overlapping each other.  The current GUI control 
![image](https://user-images.githubusercontent.com/761514/232908205-1611252a-e5f2-43b1-bd48-56efed2ffa9b.png)
 isn't fine grained enough.

For example:
![image](https://user-images.githubusercontent.com/761514/232908320-e725c751-5971-4d51-9595-ab0b3d573081.png)

### Suggested Solution

We should have the ability to turn on/off rows by SITE.  The rows item could become a twist-open like Layers.

### Additional Context

@gadfort do you have time for this?

gadfort:
@maliberty yeah, when I was working on the pad ring thing, that would have been helpful as well.
Any thoughts about color? keep them the same for all rows?

PR #3204, opened 2023-04-18
gui: add site based controls to rows

gadfort:
@maliberty try this out and see if this works as you want

Closes: #3201


maliberty (approved):
Thanks that was fast!

PR #3206, opened 2023-04-19
odb: error instead of crash on illegal malloc size

oharboe:


maliberty (approved):
This is fine but unlikely to be the real issue.  Its probably a memory corruption.

PR #3207, opened 2023-04-19
odb: add basic version check

oharboe:
Guard against some unnecessary confusion.

### DIFF
--- a/src/odb/include/odb/dbStream.h
+++ b/src/odb/include/odb/dbStream.h
@@ -49,6 +49,7 @@ class _dbDatabase;
 
 class dbOStream
 {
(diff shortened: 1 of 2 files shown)


gadfort:
@oharboe this already exists: https://github.com/The-OpenROAD-Project/OpenROAD/blob/9fdfef4f59ee8d27bea18af53a1bb76e3e170a1a/src/odb/src/db/dbDatabase.h#L69
I've found that the version updating is sometimes inconsistent which is likely what is causing the issue you are running into.
If you know the two commit numbers, it might be possible to track down the change that is causing a fault.

PR #3209, opened 2023-04-19
turn off tests in mpl2 when ENABLE_TESTS is off

gadfort:
Fixes:
- cmake will fail because GoogleTest is not included when ENABLE_TESTS is off, this mirrors the behavior of drt and openroad


QuantamHD:
That's fine with me.

gadfort:
@maliberty are you okay with this?

maliberty (approved):

PR #3210, opened 2023-04-19
dpl: filler masters use string match instead of regexp to avoid matching everything

gadfort:
This is a different solution to #2839 

This preserves the matching without requiring the -regexp flag.

### DIFF
--- a/src/dpl/src/Opendp.tcl
+++ b/src/dpl/src/Opendp.tcl
@@ -184,11 +184,11 @@ proc detailed_placement_debug { args } {
 }
 
 proc get_masters_arg { arg_name arg } {
-  set matched 0
   set masters {}
   # Expand master name regexps
   set db [ord::get_db]
   foreach name $arg {
+    set matched 0
     foreach lib [$db getLibs] {
       foreach master [$lib getMasters] {
         set master_name [$master getConstName]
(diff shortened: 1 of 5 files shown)


maliberty:
Probably this is fine for this usage but it punts on the fundamental question about how the APIs should work in general.  I hope in Python we can make this all more explicit.  It is a breaking change though maybe that is a good thing here.

@antonblanchard @QuantamHD thoughts?

QuantamHD:
I make use of the regex, but I'm fine with being more explicit in list form.

QuantamHD:
I think my only request would be to throw up a warning to inform the user of the change. Seems hard to execute, but maybe some kind of matching heuristic would let us inform users that they need to be more explicit going forward

QuantamHD:
Actually I have the perfect thing. We should error if a provided string matches no cells, and mention the regex change and provide a link to this issue

gadfort:
@QuantamHD it already generates warning if nothing matched (although looking at it I think it's in the wrong place).
I think the way this worked was incorrect to begin with (atleast if the idea was to be somewhat consistent with the other glob API calls). I can add a link to the PR.

donn:
I don't recall running into any issues with fill cells... But I have no objection to this in principle.

maliberty (approved):

PR #3222, opened 2023-04-21
Changing dbSigType and dbIoType from enum to enum class

fgaray:
Making dbSigType and dbIoType more type safe to prevent mixing up both classes.

I have been bitten by this in the past by mixing dbSigType and dbIoType.

If this is approved I will do the same for other types in odb/dbTypes.h

### DIFF
--- a/src/ant/src/AntennaChecker.cc
+++ b/src/ant/src/AntennaChecker.cc
@@ -383,14 +383,14 @@ std::pair<double, double> AntennaChecker::calculateWireArea(
   vector<std::pair<dbWireGraph::Edge*, dbIoType>> edge_vec;
   if (node->in_edge() != nullptr
       && nv.find(node->in_edge()->source()) == nv.end()) {
-    edge_vec.emplace_back(node->in_edge(), dbIoType::INPUT);
+    edge_vec.emplace_back(node->in_edge(), dbIoType::Value::INPUT);
   }
 
   dbWireGraph::Node::edge_iterator edge_it;
 
   for (edge_it = node->begin(); edge_it != node->end(); edge_it++) {
     if (nv.find((*edge_it)->source()) == nv.end()) {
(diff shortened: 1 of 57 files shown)


maliberty:
They are both already nested inside a class so I'm wondering how they could get mixed up.  Would you give an example of the problem.

fgaray:
Hi @maliberty , Yeah, they should be, but I have found that code like:


```
if (bterm->getIoType() == odb::dbIoType::OUTPUT) {
/// Something
}

if (bterm->getSigType() == odb::dbIoType::OUTPUT) {
///Something else
}
```

Compiles successfully in my system: gcc-10.2.1-6 inside the debian testing docker image.

maliberty:
Interesting clang will complain about it and g++ will complain about
```
if (bterm->getIoType().getValue == odb::dbIoType::OUTPUT) {
/// Something
}
```
Apparently if the 'operator Value()' is used then it doesn't complain which isn't nice.  I found
```
-  operator Value() const { return _value; }
+  bool operator==(const dbSigType& v) const { return _value == v._value; }
+  bool operator!=(const dbSigType& v) const { return _value != v._value; }
+  bool operator==(const Value v) const { return _value == v; }
+  bool operator!=(const Value v) const { return _value != v; }
```
works pretty well.  You do have to use getVaue() in a few places, like switch statements but its much less pervasive.  It avoids making the lengthy odb::dbIoType::INPUT into the even longer odb::dbIoType::Value::INPUT.

Thoughts?

fgaray:
Ah, I like that much better than my approach actually. It is less intrusive. How do you want to proceed?, do you want me to work on that and cancel this PR?

maliberty:
> Ah, I like that much better than my approach actually. It is less intrusive. How do you want to proceed?, do you want me to work on that and cancel this PR?

If you ware willing to work on it that would be great.

fgaray:
Sure, I will work on that tomorrow!, Thanks!

PR #3223, opened 2023-04-21
ifp: make -site take a list of sites

gadfort:
Changes:
- adds ability to add rows for cells that are not instantiated in the netlist.


src/ifp/README.md
```diff
@@ -9,13 +9,15 @@ initialize_floorplan
   [-site site_name]               LEF site name for ROWS
   -die_area "lx ly ux uy"         die area in microns
   [-core_area "lx ly ux uy"]      core area in microns
+  [-site site_names]              list of sites to make rows for
```
maliberty, line 12: -site now occurs twice - it isn't really part of the 'or' and was positioned first before..  If it takes a list it should be -sites

src/ifp/src/InitFloorplan.cc
```diff
@@ -160,32 +160,22 @@ static int divCeil(int dividend, int divisor)
...
                                   const odb::Rect& core,
-                                  const std::string& site_name)
+                                  const std::vector<odb::dbSite*>& user_sites)
 {
   Rect die_area(snapToMfgGrid(die.xMin()),
                 snapToMfgGrid(die.yMin()),
                 snapToMfgGrid(die.xMax()),
                 snapToMfgGrid(die.yMax()));
   block_->setDieArea(die_area);
-  auto sites = getSites();
```
maliberty, line 170: Why remove the auto-detection?  It forces the python API to require the site be named when there is only one.  I think -sites should be for additional sites over what the netlist requires rather than mandatory.

maliberty (changes requested):

src/ifp/src/InitFloorplan.cc
gadfort: The call is there, just changed it to `std::set<odb::dbSite*> sites = ...` (it's still on line 170). I don't think it changes the Python API, it can just pass in an empty list.

src/ifp/src/InitFloorplan.cc
```diff
@@ -160,32 +160,22 @@ static int divCeil(int dividend, int divisor)
...
-      logger_->warn(IFP, 41, "Site {} not found in Database.", site_name);
-    }
+  std::set<odb::dbSite*> sites = getSites();
+  sites.insert(user_sites.begin(), user_sites.end());
+
+  // Handle duplicated sites
+  std::map<std::string, odb::dbSite*> sites_by_name;
+  for (auto* site : sites) {
+    sites_by_name[site->getName()] = site;
   }
```
maliberty, line 177: You can't have two different sites with the same name - odb would error.  Is this necessary?

src/ifp/include/ifp/InitFloorplan.hh
```diff
@@ -76,11 +77,11 @@ class InitFloorplan
                      int core_space_top,
                      int core_space_left,
                      int core_space_right,
-                     const std::string& site_name);
+                     const std::vector<odb::dbSite*>& sites);
```
maliberty, line 80: The ```sites``` arg is really extra_sites now that we auto-detect the required sites from the netlist.  If you made it
```
const std::vector<odb::dbSite*>& sites = {}
```
then you wouldn't need
```
[floorplan.findSite("FreePDK45_38x28_10R_NP_162NW_34O")]
``` 
in all the python tests

src/ifp/src/InitFloorplan.cc
maliberty: Which testcase?  It looks like each LEF (dbLib) is allowed to define its own SITEs.

gadfort: init_floorplan9.tcl:

If you modify the script:
```
# init_floorplan called after a voltage domain is specified
source "helpers.tcl"
read_lef sky130hd/sky130hd.tlef
read_lef sky130hd/sky130_fd_sc_hd_merged.lef
read_liberty sky130hd/sky130_fd_sc_hd__tt_025C_1v80.lib
read_verilog reg2.v
link_design top

#create_voltage_domain TEMP_ANALOG -area {27 27 60 60}
#initialize_floorplan -die_area "0 0 150 150" \
#  -core_area "20 20 130 130" \
#  -site unithd
create_voltage_domain TEMP_ANALOG -area {33.58 32.64 64.86 62.56}
initialize_floorplan -die_area "0 0 155.48 146.88" \
  -core_area "18.4 16.32 137.08 130.56" 
  
set def_file [make_result_file init_floorplan9.def]
write_def $def_file
diff_files init_floorplan9.defok $def_file


foreach lib [[ord::get_db] getLibs] {
  foreach site [$lib getSites] {
    puts "[$lib getName] [$site getName] $site"
  }
}
```

It will output 2 unithd sites with a unique pointer to each (even if they are the same).
```
sky130hd unithd _d00fcd209d550000_p_odb__dbSite
sky130hd unithddbl _f00fcd209d550000_p_odb__dbSite
sky130_fd_sc_hd_merged unithd _c0b3c2209d550000_p_odb__dbSite
```

maliberty: In lefin::macro it copies the site, see https://github.com/The-OpenROAD-Project/OpenROAD/blob/97bb79043c39cac424604e50450f2df9484e4ded/src/odb/src/lefin/lefin.cpp#L1222

@osamahammad21 implemented [this code](https://github.com/The-OpenROAD-Project/OpenROAD/commit/d794f6dd75d39a2c75de2dd762fea927c058dfa8) so I'm not quite sure what the motivation was.  He is off for Eid so I'll approve and we can revisit when he's back.

maliberty (approved):

PR #3227, opened 2023-04-21
Fixes for default builds

andreili:
This patches are fixed a clean build on any system. 

1. As I see, official docker images contain a static linked executables. But, on default builds, all libraries are compiled as dynamicaly linked library. Default build - with "cmake -DCMAKE_INSTALL_PREFIX="/usr/" -DUSE_SYSTEM_BOOST=true -DCMAKE_CXX_FLAGS="-fPIC"" options only).
2. Fix optional build test on MPL2 library.
3. Fix Annealer class definition - "undefined vtable" error.


QuantamHD:
This should get reviewed on Monday

andreili:
About static/shared library.
Into all CMakeLists.txt don't exists section to install a shared library, if build it. That is, this type of build will be produced without installing all the dependencies and the program will not be able to run.
This section exists only on submodule for OpenSTA:

> install(TARGETS OpenSTA DESTINATION lib)

But, it's correct only for 32-bit systems - on 64 bit systems, we need to install binaries to "lib64" directory:

> set_property(LIB64 GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS)
> if ("${LIB64}" STREQUAL "TRUE")
>     set(LIBSUFFIX 64)
> else()
>     set(LIBSUFFIX "")
> endif()
> set(INSTALL_LIB_DIR lib${LIBSUFFIX} CACHE PATH "Installation directory for libraries")
> mark_as_advanced(INSTALL_LIB_DIR)

and then using INSTALL_LIB_DIR everywhere it is needed.

But as I noticed, this library is also built statically, so this can be omitted.

maliberty:
I don't know why you see dynamic libraries in a 'default' build.  I never see them.   From the cmake docs:
```
If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) is ON.
```
We don't set BUILD_SHARED_LIBS so you should be getting STATIC.

maliberty:
Would you expand on "Fix Annealer class definition - "undefined vtable" error.".  What file generates this error?  The nightly CI does no have this issue.  Perhaps it relates to using shared libs?

andreili:
> I don't know why you see dynamic libraries in a 'default' build. I never see them. From the cmake docs:
> 
> ```
> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) is ON.
> ```
> 
> We don't set BUILD_SHARED_LIBS so you should be getting STATIC.

On Gentoo system, ebuild override this settings to force SHARED:
`cat gentoo_common_config.cmake
set(CMAKE_GENTOO_BUILD ON CACHE BOOL "Indicate Gentoo package build")
set(LIB_SUFFIX 64 CACHE STRING "library path suffix" FORCE)
set(CMAKE_INSTALL_LIBDIR lib64 CACHE PATH "Output directory for libraries")
set(CMAKE_INSTALL_INFODIR "/usr/share/info" CACHE PATH "")
set(CMAKE_INSTALL_MANDIR "/usr/share/man" CACHE PATH "")
set(CMAKE_USER_MAKE_RULES_OVERRIDE "/var/tmp/portage/sci-electronics/openroad-9999/work/openroad-9999_build/gentoo_rules.cmake" CACHE FILEPATH "Gentoo override rules")
set(CMAKE_INSTALL_DOCDIR "/usr/share/doc/openroad-9999" CACHE PATH "")
set(BUILD_SHARED_LIBS ON CACHE BOOL "")
set(Python3_FIND_UNVERSIONED_NAMES FIRST CACHE STRING "")
set(CMAKE_INSTALL_ALWAYS 1)
set(CMAKE_ASM_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_ASM-ATT_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_C_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_Fortran_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_MODULE_LINKER_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_SHARED_LINKER_FLAGS_RELWITHDEBINFO "" CACHE STRING "")
set(CMAKE_STATIC_LINKER_FLAGS_RELWITHDEBINFO "" CACHE STRING "")`

maliberty:
This change effectively prevents us from ever using shared libraries which I don't think makes sense.  This sounds more like a gentoo issue than an OR one.  I am not inclined to accept this solution.  If you want to work on making shared libraries work I'm open to that.

PR #3238, opened 2023-04-24
Fix build issue with recent versions of spdlog

antonblanchard:


src/CMakeLists.txt
```diff
@@ -222,6 +222,7 @@ endif()
 
 find_package(spdlog REQUIRED)
 message(STATUS "spdlog: ${spdlog_VERSION}")
+add_definitions(-DFMT_DEPRECATED_OSTREAM=1)
```
maliberty, line 225: I think it would be more modern in utl/CMakeLists.txt to use

target_compile_definitions(utl PUBLIC FMT_DEPRECATED_OSTREAM=1)

All users of utl should then get the define.

maliberty:
Have you tested this with spdlog 9?

src/CMakeLists.txt
maliberty: https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#dont-use-macros-that-affect-all-targets-in-a-directory-tree-like-include_directories-add_definitions-or-link_libraries

maliberty: https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#declare-compile-definitions-with-target_compile_definitions

QuantamHD:
We should add a github action that builds with spdlog 9

maliberty:
> We should add a github action that builds with spdlog 9

If this works well enough we can just move forward to 9.

antonblanchard:
> Have you tested this with spdlog 9?

I build on Fedora 37 which has:

```
spdlog-1.10.0-3.fc37.x86_64
fmt-9.1.0-1.fc37.x86_64
```

src/CMakeLists.txt
antonblanchard: Thanks, I don't deal with cmake enough to know what's best.

antonblanchard:
The CI timed out when pulling the git tree, so force pushed to restart it.

maliberty (approved):

PR #3251, opened 2023-04-26
gui: Fix undefined behavior sanitizer error

antonblanchard:


maliberty (approved):
@gadfort any concerns?

src/gui/src/mainWindow.cpp
```diff
@@ -403,6 +403,9 @@ MainWindow::~MainWindow()
 {
   auto* gui = Gui::get();
   // unregister descriptors with GUI dependencies
```
gadfort, line 405: Keep this comment with the `unregisterDescriptor` below

gadfort (commented):
Looks good, just fix the comment so it's not floating away from the code it's associated with

PR #3284, opened 2023-05-02
setup: remove possibly harmful advice

oharboe:
if the user knows about the pros and cons
of using bashrc, then they don't need
the tip.

The documentation now encourages a harmless way to set up dependencies and environment that is not persisted across bash sessions/restart/logout, which means that a user can safely get back to "square one" by rebooting his computer.

Modifying the default cmake of the system can cause very subtle and hard to track down problems.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -640,8 +640,6 @@ EOF
 To install or run openroad, update your path with:
     export PATH="\$(brew --prefix bison)/bin:\$(brew --prefix flex)/bin:\$(brew --prefix tcl-tk)/bin:\${PATH}"
     export CMAKE_PREFIX_PATH=\$(brew --prefix or-tools)
-
-You may wish to add these lines to your .bashrc file.
 EOF
         ;;
     "openSUSE Leap" )
@@ -677,13 +675,3 @@ EOF
         _help
         ;;
 esac
-
-if [[ ! -z "${PREFIX}" ]]; then
-            cat <<EOF
-To use cmake, set cmake as an alias:
-    alias cmake='${PREFIX}/bin/cmake'
-    or  run
-    echo export PATH=${PREFIX}/bin:\${PATH} >> ~/.bash_profile
-    source ~/.bash_profile
-EOF
-fi


oharboe:
@maliberty There is a timeout or network outage error above. I didn't wait for them all to complete, I imagine it will be some time if setup is tested from scratch.

rovinski:
Many users _don't_ know about `.bashrc`. I would prefer to clarify that it makes the command persistent rather than remove it altogether.

oharboe:
> Many users _don't_ know about `.bashrc`. I would prefer to clarify that it makes the command persistent rather than remove it altogether.

I have an example of a user who followed the instructions blithely today. I dont think it is the role of OpenROAD to encourage or teach bashrc tricks.

maliberty (approved):
Users who don't know what a .bashrc is are better off using the ORFS build_openroad.sh.

rovinski:
@maliberty build_openroad.sh doesn't help at all with locating dependencies? To date, the dependency installer script and build script are two separate steps. If dependencies are installed in a non-default location (as they have to be for non-sudoers), then the build script will not find them.

maliberty:
@rovinski I believe https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/dev_env.sh takes care of the paths.  It is called from build_openroad.sh.

rovinski:
That only works if the dependencies are installed to `./dependencies`. It's not helpful if installed anywhere else. Is there a default or something that installs packages to `./dependencies`? Otherwise I don't know how the user would know to install there.

maliberty:
@rovinski https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/setup.sh

rovinski:
Requires sudo, so doesn't work for my case.

maliberty:
@oharboe would you comment on how you are setting things up?

oharboe:
> @oharboe would you comment on how you are setting things up?

We are using sudo ./setup.sh, then . env.sh and/or . dev_env.sh, no bashrc changes, no global changes(to /usr/local or the home folder equvivalent). This is as now according to the documentation. While this requires  sudo access, it also works fine for side by side project requirements of varying or-tools and cmake, etc. requirements. Also it works when cmake and or-tools requirements differ across versions of our project(we can check out and build an older version that requires a different cmake version).

PR #3286, opened 2023-05-02
DependencyInstaller.sh Fixes

rovinski:
Some small fixes and cleanups of the dependency installer script

* Check for and install pcre2 as a swig dependency (not all systems may have a suitable version)
* Eliminate some spurious warnings about "file not found" when grepping files that don't exist
* Formatting
* If installing to a non-default location, emit a sourceable script to set up the environment
* Convert relative paths to absolute when passing to installers

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -132,6 +132,15 @@ _installCommonDev() {
 
     cd "${lastDir}"
     rm -rf "${baseDir}"
+
+    if [[ ! -z ${PREFIX} ]]; then
+      # Emit an environment setup script
+      cat > ${PREFIX}/env.sh <<EOF
+depRoot="\$(dirname \$(readlink -f "\${BASH_SOURCE[0]}"))"
+PATH=\${depRoot}/bin:\${PATH}
+LD_LIBRARY_PATH=\${depRoot}/lib64:\${depRoot}/lib:\${LD_LIBRARY_PATH}
+EOF
+    fi
 }
 
 _installOrTools() {
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -56,6 +56,7 @@ _installCommonDev() {
         if [[ -z $(pcre2-config --version) ]]; then
           tarName="pcre2-${pcreVersion}.tar.gz"
           wget https://github.com/PCRE2Project/pcre2/releases/download/pcre2-${pcreVersion}/${tarName}
+          md5sum -c <(echo "${pcreChecksum} ${tarName}") || exit 1
           ./Tools/pcre-build.sh
         fi
         ./autogen.sh
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -30,7 +30,8 @@ _installCommonDev() {
 
     # CMake
     cmakePrefix=${PREFIX:-"/usr/local"}
-    if [[ -z $(${cmakePrefix}/bin/cmake --version | grep ${cmakeVersionBig}) ]]; then
+    cmakeBin=${cmakePrefix}/bin/cmake
+    if [[ ! -f ${cmakeBin} || -z $(${cmakeBin} --version | grep ${cmakeVersionBig}) ]]; then
         cd "${baseDir}"
         wget https://cmake.org/files/v${cmakeVersionBig}/cmake-${cmakeVersionSmall}-${osName}-x86_64.sh
         md5sum -c <(echo "${cmakeChecksum}  cmake-${cmakeVersionSmall}-${osName}-x86_64.sh") || exit 1
@@ -41,8 +42,9 @@ _installCommonDev() {
     fi
 
     # SWIG
-    swigPrefix=${PREFIX:-"/usr"}
-    if [[ -z $(${swigPrefix}/bin/swig -version | grep ${swigVersion}) ]]; then
+    swigPrefix=${PREFIX:-"/usr/local"}
+    swigBin=${swigPrefix}/bin/swig
+    if [[ ! -f ${swigBin} || -z $(${swigBin} -version | grep ${swigVersion}) ]]; then
         cd "${baseDir}"
         tarName="v${swigVersion}.tar.gz"
         wget https://github.com/swig/swig/archive/${tarName}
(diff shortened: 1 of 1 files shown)


etc/DependencyInstaller.sh
```diff
@@ -132,6 +132,15 @@ _installCommonDev() {
 
     cd "${lastDir}"
     rm -rf "${baseDir}"
+
+    if [[ ! -z ${PREFIX} ]]; then
```
oharboe, line 136: Does env.sh and setup_env.sh in ORFS need to be updated?

oharboe (changes requested):
Does env.sh and setup_env.sh in ORFS need to be updated?

etc/DependencyInstaller.sh
rovinski: "need to be" - no. This is just a convenience addition for installations making use of `-prefix`. It doesn't change any behavior with current scripts, although after merging they could make use if they wanted to.

rovinski:
@vvbandeira Any ideas about the failures? The all seem to be related to 
```
[ 43%] Linking CXX executable TestWorker
cd /OpenROAD/build/src/dst/test/cpp && /home/openroad-deps/bin/cmake -E cmake_link_script CMakeFiles/TestWorker.dir/link.txt --verbose=1
/usr/bin/clang++ -O3 -DNDEBUG CMakeFiles/TestWorker.dir/TestWorker.cc.o CMakeFiles/TestWorker.dir/stubs.cpp.o -o TestWorker  ../../dst.a /usr/lib64/libtcl8.6.so ../../../utl/utl.a ../../../utl/libutl_lib.a /home/openroad-deps/lib64/libspdlog.a /OpenROAD/src/sta/app/libOpenSTA.a /usr/lib64/libtcl8.6.so -pthread /usr/lib64/libtclreadline.so -ltcl /usr/lib64/libz.so /home/openroad-deps/lib/libboost_serialization.a /home/openroad-deps/lib/libboost_system.a /home/openroad-deps/lib/libboost_thread.a -lpthread 
/usr/bin/ld: cannot find -ltcl
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)

which I don't believe is related to this change.
```

etc/DependencyInstaller.sh
oharboe: So update env.sh and dev_env.sh to use a single source of truth instead of duplicating how to set up paths?

rovinski:
Confirmed - even without this PR those tests are failing in the same way. @vvbandeira

etc/DependencyInstaller.sh
rovinski: Sure

etc/DependencyInstaller.sh
```diff
@@ -28,25 +30,34 @@ _installCommonDev() {
...
+        md5sum -c <(echo "${swigChecksum} ${tarName}") || exit 1
         tar xfz ${tarName}
         cd swig-${tarName%%.tar*} || cd swig-${swigVersion}
+
+        # Check if pcre2 is installed
+        if [[ -z $(pcre2-config --version | grep ${pcreVersion}) ]]; then
+          tarName="pcre2-${pcreVersion}.tar.gz"
+          wget https://github.com/PCRE2Project/pcre2/releases/download/pcre2-${pcreVersion}/${tarName}
+          ./Tools/pcre-build.sh
+        fi
```
maliberty, line 60: Why is there a new dependency in this PR?

etc/DependencyInstaller.sh
maliberty: @rovinski you asked for a review but this item seems unresolved.  Is there more to do yet?

etc/DependencyInstaller.sh
rovinski: On some systems pcre2 isn't installed, pcre1 is. The SWIG install breaks if it can't find pcre2.

etc/DependencyInstaller.sh
rovinski: There would be nothing to resolve here, only in ORFS after this is merged.

etc/DependencyInstaller.sh
rovinski: I realized this was likely affecting all builds so I removed the version requirement. It should only install on systems which don't have pcre2 at all.

etc/DependencyInstaller.sh
```diff
@@ -28,25 +30,34 @@ _installCommonDev() {
...
         cd "${baseDir}"
         tarName="v${swigVersion}.tar.gz"
         wget https://github.com/swig/swig/archive/${tarName}
-        md5sum -c <(echo "${swigChecksum}  ${tarName}") || exit 1
+        md5sum -c <(echo "${swigChecksum} ${tarName}") || exit 1
         tar xfz ${tarName}
         cd swig-${tarName%%.tar*} || cd swig-${swigVersion}
+
+        # Check if pcre2 is installed
+        if [[ -z $(pcre2-config --version) ]]; then
```
maliberty, line 56: What happens in pcre2-config doesn't exist?

rovinski: It downloads the tarball and uses swig's built-in script to build it and statically link it during the swig build.

oharboe (commented):
Post-merge the ORFS env.sh and dev_env.sh should be changed to use the generated environment setup scripts in the installation to have a single source of truth.

oharboe (commented):
Post-merge the ORFS env.sh and dev_env.sh should be changed to use the generated environment setup scripts in the installation to have a single source of truth.

oharboe (commented):
Post-merge the ORFS env.sh and dev_env.sh should be changed to use the generated environment setup scripts in the installation to have a single source of truth.

oharboe:
@rovinski I wasn't able to change my request for changes to a review comment, so I had to dismiss my own review with "Post-merge the ORFS env.sh and dev_env.sh should be changed to use the generated environment setup scripts in the installation to have a single source of truth."

rovinski:
I raised the build errors as a separate issue in #3343

@maliberty anything else preventing this from merging?

rovinski:
Closes #3343

rovinski:
CentOS failures are unrelated. Network issues.

rovinski:
I should note that before this PR, OpenSuse, testinstaller-macos, and rhel were all failing. Now, at least rhel is passing. If centos is rerun, it should pass (I don't have permissions to rerun github actions).

vvbandeira:
I think it is OK to merge the PR as is; there are some changes coming that will make this more stable.

maliberty (approved):

PR #3295, opened 2023-05-05
Refactoring to enable a first CTS unit test.

cdleary:
This splits all the remaining targets necessary to run a "librarified" CTS unit test, that does not pull in the GUI and all openroad singleton-based dependencies.

It runs a simple unit test, which doesn't look like much, but it starts on a unit testing methodology where we can iterate on the algorithms (e.g. CTS) in isolation.


maliberty:
There is a lot more than CTS going on here.

src/grt/include/grt/GlobalRouter.h
```diff
@@ -142,13 +139,19 @@ class GlobalRouter
  public:
   GlobalRouter();
   ~GlobalRouter();
+
+#ifndef SWIG
```
maliberty, line 143: It is better to handle this in the .i if possible to keep swig concerns out of the c++

src/grt/include/grt/GlobalRouter.h
```diff
@@ -251,6 +253,20 @@ class GlobalRouter
       const Pin& pin,
       std::vector<std::pair<odb::Point, odb::Point>>& ap_positions);
 
+#ifndef SWIG
+  void setRenderer(std::unique_ptr<AbstractGrouteRenderer> groute_renderer);
+  AbstractGrouteRenderer* getRenderer();
+#endif
+
+  odb::dbDatabase* db()
```
maliberty, line 261: const

src/grt/src/heatMap.h
```diff
@@ -32,20 +32,25 @@
...
 namespace grt {
 
-class RoutingCongestionDataSource : public gui::HeatMapDataSource
+class RoutingCongestionDataSource : public gui::HeatMapDataSource,
+                                    public AbstractRoutingCongestionDataSource
 {
  public:
   RoutingCongestionDataSource(utl::Logger* logger, odb::dbDatabase* db);
-  ~RoutingCongestionDataSource() {}
+  ~RoutingCongestionDataSource() override = default;
```
maliberty, line 45: Why does this need overriding if we are using =default?

src/rsz/src/AbstractSteinerRenderer.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
maliberty, line 1: Missing copyright header

src/rsz/src/SteinerRenderer.cc
```diff
@@ -0,0 +1,31 @@
+#include "SteinerRenderer.h"
```
maliberty, line 1: Missing copyright header

src/rsz/src/SteinerRenderer.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
maliberty, line 1: Missing copyright header  (etc throughout)

src/rsz/src/SteinerRenderer.h
```diff
@@ -0,0 +1,21 @@
...
+#include "AbstractSteinerRenderer.h"
+#include "gui/gui.h"
+
+namespace rsz {
+
+class SteinerRenderer : public gui::Renderer, public AbstractSteinerRenderer
+{
+ public:
+  SteinerRenderer();
+  ~SteinerRenderer() override = default;
```
maliberty, line 12: unnecessary?

src/stt/src/LinesRenderer.cpp
```diff
@@ -0,0 +1,48 @@
...
+  if (gui::Gui::enabled()) {
+    if (LinesRenderer::lines_renderer == nullptr) {
+      LinesRenderer::lines_renderer = new LinesRenderer();
+      gui->registerRenderer(LinesRenderer::lines_renderer);
+    }
+    std::vector<std::pair<odb::Point, odb::Point>> lines;
+    for (int i = 0; i < tree.branchCount(); i++) {
+      const stt::Branch& branch = tree.branch[i];
+      int x1 = branch.x;
+      int y1 = branch.y;
```
maliberty, line 38: Part of my mental model is that things that aren't const *will* change value.  It makes code reading simpler.

maliberty (changes requested):

src/cts/src/Graphics.cpp
```diff
@@ -43,20 +43,28 @@
 
 namespace cts {
 
-Graphics::Graphics(HTreeBuilder* h_tree_builder, Clock* clock)
-    : clock_(clock), h_tree_builder_(h_tree_builder), sink_clustering_(nullptr)
+void Graphics::initializeWithClock(HTreeBuilder* h_tree_builder, Clock& clock)
 {
+  clock_ = &clock;
+  h_tree_builder_ = h_tree_builder;
+  sink_clustering_ = nullptr;
```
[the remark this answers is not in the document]
cdleary: The CTS implementation wants to initialize the object, and the object is now constructed outside of the CTS implementation, so it needs to be an initialization method (that inherits from the abstract API) instead of a constructor.

src/ant/src/CMakeLists.txt
```diff
@@ -37,19 +37,36 @@ swig_lib(NAME      ant
...
     MakeAntennaChecker.cc
-    AntennaChecker.cc
+)
+
+target_include_directories(ant_lib
+  PUBLIC
+    ../include
 )
 
 target_include_directories(ant
```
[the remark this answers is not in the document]
cdleary: We did this one here, generally we tend to try to do "include what you use" instead of assuming transitively exported dependencies. It's a little hard to keep it consistent across all the targets without some kind of "build cleaner" system. If you feel strongly there shouldn't be duplication in the wrapping targets I'll do another pass, but wanted to note this will probably be hard to keep consistent without some cleaner/checker system.

maliberty (approved):

PR #3299, opened 2023-05-07
gui: scale heatmap with gui scaling

gadfort:
This is an attempt to speed up handling of heatmaps in the gui.
1. change the underlying data type from an rtree to a 2d array.
2. only draw the portion of the heatmap that is currently visible
3. when zoomed out merges multiple bins together to avoid drawing boxes that are less an than a pixel.

### DIFF
--- a/src/gui/src/heatMap.cpp
+++ b/src/gui/src/heatMap.cpp
@@ -71,6 +71,8 @@ HeatMapDataSource::HeatMapDataSource(utl::Logger* logger,
       setup_(nullptr),
       color_generator_(SpectrumGenerator(100.0))
 {
+  clearMap();
+
   // ensure color map is initialized
   updateMapColors();
 }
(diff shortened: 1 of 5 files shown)


maliberty:
@oharboe would you see if this helps with your design.

oharboe:
> @oharboe would you see if this helps with your design.

Enormously.

gadfort:
@oharboe good to hear.

src/gui/src/heatMap.cpp
```diff
@@ -583,43 +622,92 @@ void HeatMapRenderer::drawObjects(Painter& painter)
...
+  // minimum box size is 2 pixels
+  const double min_dbu = 2.0 / painter.getPixelsPerDBU();
+
+  const double dbu_per_micron = datasource_.getBlock()->getDbUnitsPerMicron();
+  const int x_scale
+      = std::ceil(min_dbu / (datasource_.getGridXSize() * dbu_per_micron));
+  const int y_scale
+      = std::ceil(min_dbu / (datasource_.getGridYSize() * dbu_per_micron));
+
+  std::vector<HeatMapDataSource::MapColor> draw_pts;
```
maliberty, line 634: reserve at x_size * y_size ?

Can this be drawn immediately after down sampling rather than storing and later drawing?

src/gui/src/heatMap.cpp
```diff
@@ -583,43 +622,92 @@ void HeatMapRenderer::drawObjects(Painter& painter)
...
   const odb::Rect& bounds = painter.getBounds();
 
-  for (const auto& [bbox, map_pt] : datasource_.getMap()) {
-    if (!map_pt->has_value) {  // value not set so nothing to draw
+  // minimum box size is 2 pixels
+  const double min_dbu = 2.0 / painter.getPixelsPerDBU();
+
+  const double dbu_per_micron = datasource_.getBlock()->getDbUnitsPerMicron();
+  const int x_scale
+      = std::ceil(min_dbu / (datasource_.getGridXSize() * dbu_per_micron));
```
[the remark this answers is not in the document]
gadfort: microns, and it could, but when we save the display settings it's easier to do that in microns since when we open on a new design/technology the dbus might be different.

I think rewriting that would be a different PR as it's a bigger change.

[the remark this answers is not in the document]
maliberty: I think it would be good but am fine to separate it.

src/gui/src/heatMap.cpp
gadfort: yes, that would also be a good idea. I didn't think of it when doing this.

maliberty: I'm ok to go ahead with this as it already has benefit.  We can revisit it later if needed.  I think the reserve should be easy enough to add this to PR.

gadfort: Fixing the draw loop should be a pretty minor move of the code, I just didn't consider it. I'll have something in soon-ish

gadfort: @maliberty I made the change

maliberty (approved):

PR #3300, opened 2023-05-08
ppl: update fallback groups

eder-matheus:


src/ppl/test/large_groups3.tcl
```diff
@@ -0,0 +1,52 @@
...
+pin296 pin297 pin298 pin299}
+
+set_io_pin_constraint -region top:* -group -order -pin_names { pin300 pin301 pin302 pin303 pin304 pin305
+pin306 pin307 pin308 pin309 pin310 pin311 pin312 pin313 pin314 pin315
+pin316 pin317 pin318 pin319 pin320 pin321 pin322 pin323 pin324 pin325
+pin326 pin327 pin328 pin329 pin330 pin331 pin332 pin333 pin334 pin335
+pin336 pin337 pin338 pin339 pin340 pin341 pin342 pin343 pin344 pin345
+pin346 pin347 pin348 pin349 pin350 pin351 pin352 pin353 pin354 pin355
+pin356 pin357 pin358 pin359 pin360 pin361 pin362 pin363 pin364 pin365
+pin366 pin367 pin368 pin369 pin370 }
```
rovinski, line 44: Would this be more concise to generate in a for loop?

src/ppl/src/Slots.cpp
```diff
@@ -58,4 +58,10 @@ int Section::getMaxContiguousSlots(const std::vector<Slot>& slots)
   return max_contiguous_slots;
 }
 
+bool Interval::operator==(const Interval& interval) const
+{
+  return edge_ == interval.getEdge() && begin_ == interval.getBegin()
+         && end_ == interval.getEnd() && layer_ == interval.getLayer();
+}
+
```
maliberty, line 66: This looks like it is equivalent to what the compiler would do automatically.  What makes this required?

eder-matheus: `Interval` is a class.  When I use the `==`operator directly, the build fails with
```
IOPlacer.cpp:1421:29: error: invalid operands to binary expression ('ppl::Interval' and 'ppl::Interval')

    if (constraint.interval == interval) {
```

src/ppl/test/large_groups3.tcl
eder-matheus: Just made this update. Looks much better now. Thanks!

src/ppl/src/Slots.cpp
maliberty: Nvm, I forgot operator== isn't one that is default generated (I always think it should be).

maliberty (approved):

oharboe:
@eder-matheus Thanks! We're still testing, but so far looks good! We'll file a fresh github issue in case we see any issues.

PR #3304, opened 2023-05-09
DO NOT MERGE checking for negative total wire length

oharboe:


oharboe:
@maliberty Someone noticed hpwl being negative, so I thought I would check quickly in master on any of the tests there.

This assert should never trigger, right?

maliberty:
It should not fail.  Likewise in NesterovBase::getHpwl.  Note that we compile asserts out in optimized builds so they don't create a lot of coverage in the CI.

It seems unlikely that we are overflowing an int64_t.  Is there a testcase that shows the problem?

oharboe:
> It should not fail. Likewise in NesterovBase::getHpwl. Note that we compile asserts out in optimized builds so they don't create a lot of coverage in the CI.
> 
> It seems unlikely that we are overflowing an int64_t. Is there a testcase that shows the problem?

Yes, someone noticed it in the discussions, but how do I search discussions?

maliberty:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1001
(just use the search box at the top of the GH page)

oharboe:
> [The-OpenROAD-Project/OpenROAD-flow-scripts#1001](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1001) (just use the search box at the top of the GH page)

It doesnt show issues, pull requests or discussions NAICT

maliberty:
> > [The-OpenROAD-Project/OpenROAD-flow-scripts#1001](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1001) (just use the search box at the top of the GH page)
> 
> It doesnt show issues, pull requests or discussions NAICT

image

PR #3306, opened 2023-05-09
drt: tweak outOfDieVia() to do less pointerchasing

oharboe:
it is called a lot, so this helps


oharboe:
Easily resurrected, not shockingly better.

PR #3307, opened 2023-05-09
DPO: multi height cells support 

mina1460:
Main points:
- DPO treats the grid as an array of segments. A segment is a section of a row, and rows are made of single-heights. Our IFP treats rows differently as we have rows of all heights. Thus, while constructing the DPO::Architecture object, I only add rows that are single-height. 
Screenshot 2023-06-14 at 11 46 18 pm


maliberty:
ready for review?

mina1460:
@maliberty not yet, DPO places cells in unaligned locations. I am still trying to figure out why.

src/dpo/src/Optdp.cpp
```diff
@@ -729,7 +740,19 @@ void Optdp::createArchitecture()
...
-
+  for (auto skip : skip_list) {
+    std::string skip_string = "[";
+    for (int i = 0; i < skip.second.size(); i++) {
+      skip_string += skip.second[i] + ",]"[i == skip.second.size() - 1];
+    }
+    logger_->warn(DPO,
+                  108,
+                  "Skipping the rows with height {} as the minimum row height "
+                  "is {}. Here is the skip list: {}",
```
maliberty, line 751: A large design may have many rows.  I'm concerned this may produce a huge message string.  I think it might be better to say "skipping all rows with the following SITE types ..." rather than list all the rows themselves.  The skip_list could just be a set of sites.

src/dpo/src/Optdp.cpp
```diff
@@ -729,7 +740,21 @@ void Optdp::createArchitecture()
     unsigned orient = dbToDpoOrient(row->getOrient());
     archRow->setOrient(orient);
   }
-
+  for (const auto& skip : skip_list) {
+    std::string skip_string = "[";
+    int i = 0;
+    for (const auto& skipped_site : skip.second) {
+      skip_string += skipped_site + ",]"[i == skip.second.size() - 1];
```
maliberty, line 747: I've never seen this trick before - cute.

maliberty (approved):

PR #3318, opened 2023-05-10
Fix finding the clocks of the cells after scan replace and rollback

fgaray:
I found this bug while implementing scan stitching. The unit tests for this part will in the PR for scan stitching

### DIFF
--- a/src/dft/src/replace/ScanReplace.cpp
+++ b/src/dft/src/replace/ScanReplace.cpp
@@ -331,6 +331,7 @@ void ScanReplace::scanReplace()
 {
   odb::dbChip* chip = db_->getChip();
   scanReplace(chip->getBlock());
+  sta_->networkChanged();
 }
 
 // Recursive function that iterates over a block (and the blocks inside this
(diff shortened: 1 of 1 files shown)


maliberty:
I curious why this is required.  Sta should be listening for db callbacks and handling this automatically.  What problem did you observe?

fgaray:
Hi Matt,

The problem is that when I call Sta::clocks, I get an empty ClockSet after rollback, that is: I am unable to find the driving clock for the cell when I revert back from the scan cell to a non-scan one.

I agree that STA should be listening to changes, so maybe it is a bug with one of the observers?

I can provide a quick test case if you want to take a look, but adding this line solved the issue that I was seeing.

maliberty:
I wonder if ensureClkNetwork() would be better.  I'm afraid your approach might have performance implications.

fgaray:
I tried sta_->ensureClkNetwork() but doesn't seems to fix the issue. I wonder if there is another function that it is being called by networkChanged that is making the difference.

How big is the performance implication?, in total I should be calling networkChanged 3 times: 2 for preview_dft and 1 for insert_dft.

fgaray:
Ok, so I commented search_->clear() and now calling networkChanged is NOT fixing the issue.

So I guess there is some observer that is not updating search_ in STA?

maliberty:
I'm ok to merge this as is and we can sort it out later if it proves to be an issue.

maliberty (approved):

Issue #3290, opened 2023-05-03
Add select all buffer trees


maliberty:
### Description

From a selected instance if you right click and use Select/All nets you may only get up to the first set of buffers, eg:
![image](https://user-images.githubusercontent.com/761514/236072397-7bedd6a8-00f2-4b14-93de-fd3a5f342956.png)

### Suggested Solution

It would be helpful if there were a similar Select/All buffer nets.  Likewise for the other options as well as the Highlight/*.

niv280:
I will be happy to work on that

maliberty:
already fixed

PR #3332, opened 2023-05-13
Add select all buffer trees #3290

niv280:
Hi,
I added the option to select/highlight all buffer trees in the gui menu.

I didn't find very great case to test is, so I manually added to gcd test multiple buffer and connect them to random FF ( if you got better test case please let me know), the result looks like that:
image

Few notes:

- To get all the buffers I used the existing class of BufferTrees. In that class there is condition that only if the timing source of the inst is equal to TIMING it consider buffer, I'm not fully understood that condition so for now I added another option to work around it and create bufferTree even if the timing source don't defined.
-  When selected all buffer tree all the inst/nets/Bterms in of the buffer trees are selected, I can change it easily I just wasn't sure  if that was the original propose.
-  Right now it supporting only when select inst, I can add other object if you think it is needed too.

### DIFF
--- a/src/gui/src/bufferTreeDescriptor.cpp
+++ b/src/gui/src/bufferTreeDescriptor.cpp
@@ -46,9 +46,9 @@ namespace gui {
 
 sta::dbSta* BufferTree::sta_ = nullptr;
 
-BufferTree::BufferTree(odb::dbNet* net, bool timingSource)
+BufferTree::BufferTree(odb::dbNet* net)
 {
-  populate(net, timingSource);
+  populate(net);
 
   // Sort nets to ensure stable color order
   std::sort(
@@ -59,7 +59,7 @@ BufferTree::BufferTree(odb::dbNet* net, bool timingSource)
   name_ = nets_[0]->getName();
 }
 
-void BufferTree::populate(odb::dbNet* net, bool timingSource)
+void BufferTree::populate(odb::dbNet* net)
 {
   if (net == nullptr) {
     return;
@@ -77,7 +77,7 @@ void BufferTree::populate(odb::dbNet* net, bool timingSource)
 
   for (auto* iterm : net->getITerms()) {
     auto* inst = iterm->getInst();
-    if (!BufferTree::isAggregate(inst, timingSource)) {
+    if (!BufferTree::isAggregate(inst)) {
       iterm_terms_.insert(iterm);
       continue;
     }
@@ -86,20 +86,20 @@ void BufferTree::populate(odb::dbNet* net, bool timingSource)
 
     for (auto* next_iterm : inst->getITerms()) {
       if (!next_iterm->getSigType().isSupply()) {
-        populate(next_iterm->getNet(), timingSource);
+        populate(next_iterm->getNet());
       }
     }
   }
 }
 
-bool BufferTree::isAggregate(odb::dbNet* net, bool timingSource)
+bool BufferTree::isAggregate(odb::dbNet* net)
 {
   if (net == nullptr) {
     return false;
   }
 
   for (auto* iterm : net->getITerms()) {
-    if (BufferTree::isAggregate(iterm->getInst(), timingSource)) {
+    if (BufferTree::isAggregate(iterm->getInst())) {
       return true;
     }
   }
@@ -107,17 +107,13 @@ bool BufferTree::isAggregate(odb::dbNet* net, bool timingSource)
   return false;
 }
 
-bool BufferTree::isAggregate(odb::dbInst* inst, bool timingSource)
+bool BufferTree::isAggregate(odb::dbInst* inst)
 {
   if (inst == nullptr) {
     return false;
   }
 
   auto* network = sta_->getDbNetwork();
-  if (inst->getSourceType() != odb::dbSourceType::TIMING && timingSource) {
-    return false;
-  }
-
   auto* master = inst->getMaster();
   auto* cell = network->dbToSta(master);
   if (cell == nullptr) {
--- a/src/gui/src/bufferTreeDescriptor.cpp
+++ b/src/gui/src/bufferTreeDescriptor.cpp
@@ -46,9 +46,9 @@ namespace gui {
 
 sta::dbSta* BufferTree::sta_ = nullptr;
 
-BufferTree::BufferTree(odb::dbNet* net)
+BufferTree::BufferTree(odb::dbNet* net, bool timingSource)
 {
-  populate(net);
+  populate(net, timingSource);
 
   // Sort nets to ensure stable color order
   std::sort(
@@ -59,7 +59,7 @@ BufferTree::BufferTree(odb::dbNet* net)
   name_ = nets_[0]->getName();
 }
 
-void BufferTree::populate(odb::dbNet* net)
+void BufferTree::populate(odb::dbNet* net, bool timingSource)
 {
   if (net == nullptr) {
     return;
@@ -77,7 +77,7 @@ void BufferTree::populate(odb::dbNet* net)
 
   for (auto* iterm : net->getITerms()) {
     auto* inst = iterm->getInst();
-    if (!BufferTree::isAggregate(inst)) {
+    if (!BufferTree::isAggregate(inst, timingSource)) {
       iterm_terms_.insert(iterm);
       continue;
     }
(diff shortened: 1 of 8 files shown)


maliberty:
I tried to use this on the clk net of gcd:

image

The 'all buffer trees' is grayed out while the inspector shows that a buffer tree exists.  How should I use this?

niv280:
That because you selected the Net, that the output when I selected some clk buffer: 
image


Do you think it should be active when choosing the net as well?
I should be very easy to add that option.

maliberty:
> That because you selected the Net, that the output when I selected some clk buffer:
D'oh that makes sense.

src/gui/src/mainWindow.cpp
```diff
@@ -1244,6 +1244,45 @@ void MainWindow::selectHighlightConnectedNets(bool select_flag,
...
+{
+  SelectionSet connected_objects;
+  for (auto& sel_obj : selected_) {
+    if (sel_obj.isInst()) {
+      auto inst_obj = std::any_cast<odb::dbInst*>(sel_obj.getObject());
+      for (auto inst_term : inst_obj->getITerms()) {
+        auto inst_term_dir = inst_term->getIoType();
+        if (!inst_term->getSigType().isSupply()
+            && (inst_term_dir == odb::dbIoType::OUTPUT
+                || inst_term_dir == odb::dbIoType::INOUT)) {
```
maliberty, line 1258: Why not input iterms?

src/gui/src/layoutViewer.cpp
```diff
@@ -3263,6 +3263,17 @@ void LayoutViewer::selectHighlightConnectedNets(bool select_flag,
       select_flag, output, input, highlight_group);
 }
 
+void LayoutViewer::selectHighlightConnectedBufferTrees(bool select_flag)
+{
+  int highlight_group = 0;
+  if (!select_flag) {
+    HighlightGroupDialog dlg;
+    dlg.exec();
+    highlight_group = dlg.getSelectedHighlightGroup();
```
maliberty, line 3272: Why do you need to use a dialog for this?  @gadfort is this the best way?

maliberty:
Please fix the clang-tidy warnings

src/gui/src/layoutViewer.cpp
gadfort: In this case it might be easier to have a submenu that gives you the different highlight groups.

src/gui/src/layoutViewer.cpp
```diff
@@ -3426,6 +3443,8 @@ void LayoutViewer::addMenuAndActions()
       = highlight_menu->addAction(tr("Input Nets"));
   menu_actions_[HIGHLIGHT_ALL_NETS_ACT]
       = highlight_menu->addAction(tr("All Nets"));
+  menu_actions_[HIGHLIGHT_ALL_BUFFER_TREES_ACT]
+      = highlight_menu->addAction(tr("All buffer trees"));
```
gadfort, line 3447: @maliberty is there value in being able to select just the input trees vs. output trees?

src/gui/src/bufferTreeDescriptor.cpp
```diff
@@ -86,35 +86,35 @@ void BufferTree::populate(odb::dbNet* net)
...
+    if (BufferTree::isAggregate(iterm->getInst(), timingSource)) {
       return true;
     }
   }
 
   return false;
 }
 
-bool BufferTree::isAggregate(odb::dbInst* inst)
+bool BufferTree::isAggregate(odb::dbInst* inst, bool timingSource)
```
gadfort, line 110: why is `timingSource` needed?

src/gui/src/layoutViewer.cpp
niv280: As far I understand the buffer Tree object is that it find all the buffer in the fan out and facing of the specific net until it reach to none buffer inst.
So in order to choose only output or input I need to add support in buffer tree object, is that OK?

src/gui/src/layoutViewer.cpp
niv280: Did you mean only in buffer tree or in all the highlight options?

src/gui/src/bufferTreeDescriptor.cpp
niv280: This function consider buffer only if the TimingSource of the inst is TIMING, but I found that only inst that insert from ECO ( using RSZ ) are getting some kind of TimingSource value.
So the buffer tree don't found most of the buffers in that case, so I added some work around that issue, because Im not fully understood why there was from start the TimingSource check

maliberty: @gadfort I'm also a bit unclear on this check

gadfort: @maliberty I believe when we first added this, we decided that the tree should terminate on buffers and inverters inserted by synthesis (I'm not 100% if that really matters), so unless there is a compelling reason to do so, I would be happy to remove that limit

maliberty: We remove all the synthesis buffers so its only really inverters which wouldn't pass isBuffer() afaik.  I think we can remove it.  

That said I'm surprised it made a difference unless you were looking at a clock net.

niv280: I encounter that because I created test case with multiple place_cell command manually and then the buffer didn't added to the tree.
My thought was that it should be added to tree in case someone will insert buffer manually.

maliberty: That makes sense.  Go ahead and remove the check

niv280:
Ok so I commited to following change:
-  Fix clang warning
- Remove timingSource check ( and my work around ) 
-  Added sub menu for highlight with addition dialog, the code for this one  a bit ugly since I didn't find elegant way to represent all the colors names, if you got better idea please let me know.
image

maliberty:
With https://jenkins.openroad.tools/job/OpenROAD-flow-scripts-Nightly-Public/lastSuccessfulBuild/artifact/flow/final_report_swerv_wrapper_nangate45_base_2023-05-31_14-36.tar.gz I tried selecting the upper right instance and selecting all buffer nets:

image

If I do select all nets I see:
image

I expect the buffer nets to be a super set of the nets so I'm not sure what is going on.

maliberty:
Are you only showing the output nets?

maliberty:
On one proprietary design I tried I get:
```
/opt/rh/devtoolset-8/root/usr/include/c++/8/bits/stl_vector.h:932: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = odb::dbNet*; _Alloc = std::allocator<odb::dbNet*>; std::vector<_Tp, _Alloc>::reference = odb::dbNet*&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
Signal 6 received
Stack trace:
 0# 0x0000000000CF7EC7 in /workspaces/mliberty/w5/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 1# 0x00007F105DA95400 in /lib64/libc.so.6
 2# gsignal in /lib64/libc.so.6
 3# abort in /lib64/libc.so.6
 4# gui::BufferTree::BufferTree(odb::dbNet*) in /workspaces/mliberty/w5/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 5# gui::MainWindow::selectHighlightConnectedBufferTrees(bool, int) in /workspaces/mliberty/w5/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
```
any ideas?

niv280:
Any chance this is cell that connected to the io pin?
Because I think that the BufferTree crashed when it get nets that connected to io.

I'm showing the output and input because the BufferTree is going backward and forward for any given net.
Not sure regarding the: [final_report_swerv_wrapper_nangate45_base_2023-05-31_14-36.tar.gz](https://jenkins.openroad.tools/job/OpenROAD-flow-scripts-Nightly-Public/lastSuccessfulBuild/artifact/flow/final_report_swerv_wrapper_nangate45_base_2023-05-31_14-36.tar.gz) I will check and update

maliberty:
To fix the crash you need to check for a dangling iterm with no net:
```
diff --git a/src/gui/src/mainWindow.cpp b/src/gui/src/mainWindow.cpp
index 3500575a0..32fcede46 100644
--- a/src/gui/src/mainWindow.cpp
+++ b/src/gui/src/mainWindow.cpp
@@ -1257,6 +1257,9 @@ void MainWindow::selectHighlightConnectedBufferTrees(bool select_flag,
             && (inst_term_dir == odb::dbIoType::OUTPUT
                 || inst_term_dir == odb::dbIoType::INOUT)) {
           auto net_obj = inst_term->getNet();
+          if (!net_obj) {
+            continue;
+          }
```

niv280:
I will look into it, which design did you use in the pictures above? 
Just to be aligned

maliberty:
That design was the one I can't share

maliberty:
> * Didn't understand why you expect the all_nets to be smaller then all_buffer_tree, all_buffer_tree should included only nets that connected to buffers, no?
If a net has no buffers than the buffer tree is the net.

> * The sub menu of the highlight buffer tree is ok?
Fine with me

maliberty (approved):
Thanks, this will be quite useful.

PR #3340, opened 2023-05-16
layoutviewer: add more gui timers

oharboe:
The discrepancy is quite small locally, I suspect that the logging statements take some time, so they would have to be moved out to reduce the discrepancy between total and actual rendering time.

Hmm.... to reduce the discrepancy, we could include the logging statements in the rendering time. Or... we could sum up the logging time and print it out as a separate line so that the total should include logging time.


oharboe:
Ah, I can use, `git fetch https://github.com/Pinata-Consulting/OpenROAD.git gui-more-timers` to share w/Neria.

oharboe:
@maliberty Thoughts?

PR #3341, opened 2023-05-16
Opensta upgrade

openroadie:


maliberty:
Please run a secure CI for this

openroadie:
All regressions in OpenROAD/test pass with the code in this PR

> Please run a secure CI for this

Ok. So if I recall the process right:

- Make a branch in the private repo... make sure it is called "Secure" something 
- Commit the same code into that branch
- Wait for you secure-ci to finish.

Let me know if I missed anything.

maliberty:
Yes - in the private ORFS repo.

src/gpl/test/simple01-td.ok
```diff
@@ -46,54 +46,56 @@
...
-[NesterovSolve] Finished with Overflow: 0.099840
+[NesterovSolve] Iter: 360 overflow: 0.138573 HPWL: 5849121
+[NesterovSolve] Iter: 370 overflow: 0.119299 HPWL: 5847582
+[NesterovSolve] Iter: 380 overflow: 0.107724 HPWL: 5877406
+[NesterovSolve] Finished with Overflow: 0.099258
 worst slack 1.37
-No differences found.
+Differences found at line 63.
+    - _276_ NOR2_X2 + PLACED ( 9753 50834 ) N ;
+    - _276_ NOR2_X2 + PLACED ( 9965 51825 ) N ;
```
QuantamHD, line 101: Differences shouldn't be added to golden files.

maliberty:
I'm not sure why it only shows up in mac but:
```
2023-06-15T18:28:47.7178460Z /Users/openroad/github-action-runners/actions-runner-1/_work/OpenROAD/OpenROAD/src/gui/src/chartsWidget.cpp:212:54: error: no member named 'scaleAbreviation' in 'sta::Unit'; did you mean 'scaleAbbreviation'?
```

src/gpl/test/simple01-td-tune.ok
```diff
@@ -46,56 +46,57 @@
...
+[NesterovSolve] Iter: 340 overflow: 0.180172 HPWL: 5747118
+[NesterovSolve] Iter: 350 overflow: 0.156019 HPWL: 5741397
+[NesterovSolve] Iter: 360 overflow: 0.132409 HPWL: 5754790
+[NesterovSolve] Iter: 370 overflow: 0.112619 HPWL: 5768903
+[NesterovSolve] Finished with Overflow: 0.099853
 worst slack 1.37
-No differences found.
+Differences found at line 63.
+    - _276_ NOR2_X2 + PLACED ( 10430 52609 ) N ;
+    - _276_ NOR2_X2 + PLACED ( 10227 53342 ) N ;
```
maliberty, line 102: Needs fixing

maliberty (changes requested):

src/gpl/test/simple01-td-tune.ok
openroadie: Done.

maliberty:
Does this make sense in light of #3476 ?  Wouldn't that include this as well?

maliberty:
I don't see a secure CI result for this.  What branch name was used?

openroadie:
Yes. Don’t we have a policy of many small PRs. The previous PR is already
one month old. So let’s merge that and then worry about this incremental
change

openroadie:
I will do do a secure ci. Let’s close on the other one which now has
everything needed to merge. We don’t want to get caught with no upgrades
should secure ci show us problems with this pr

maliberty:
I don't see a secure ci.  If you have one that's fine but if you have to
start another it seems simpler to do it once.  If you prefer to do them
serially that's fine too.

openroadie:
The original pr was held up because of secure ci and issues with routing
etc or do you not remember that ? I will find out the branch name and post
here if you cannot find it

maliberty:
If I could find it I wouldn't have asked.

maliberty:
The result is likely too old to be meaningful.  I'll merge this and the next and we can deal with the fallout afterwards.  The drt fix had to be reverted anyways so this will fail again.

maliberty (approved):

Issue #2649, opened 2022-12-20
Add slack histogram to the GUI


maliberty:
### Description

Add a histogram of endpoint slack to the GUI

### Suggested Solution

https://doc.qt.io/qt-5/qtcharts-barchart-example.html looks like a good starting point

QuantamHD:
@maliberty Do you have an example of the format of such a chart?

QuantamHD:
I found this example online.

![image](https://user-images.githubusercontent.com/3784748/208764244-9dce0153-b98d-41d6-bb8a-35ece5376cb1.png)

maliberty:
That's what I'm thinking of.  Having a generic histogram capability would be useful for many things (fanout, congestion, etc).

Sov-trotter:
Hey All. First time contributor here!
Would love to work on this issue. :)

QuantamHD:
That's great!

The gui code is located here. https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/src/gui

maliberty:
Feel free to ask questions.  @gadfort and I can help

Sov-trotter:
Hey. I've been able to get a general layout done. 
![image](https://user-images.githubusercontent.com/43717431/210316737-d8fdf3fd-120e-41e8-bcc0-298f5417ae4e.png)

Looking at the TimingWidget code I am not able to figure out where to get the data from(number of paths and the slack). 
I assume that it's basic preprocessing before plotting, or do I need something like a `TimingModel`?

gadfort:
You will likely need to use: https://github.com/The-OpenROAD-Project/OpenROAD/blob/bb4801b7457d04f2ddf27c202803dfbe5c8f324d/src/gui/src/staGuiInterface.cpp#L724
At the moment it's configured to return more than one path per endpoint. This is something relatively easy (I assume) to modify and I'm happy to do that if you want.
Another thing to consider adding are controls for setup and hold slacks and corners (although those would be easy to get in once the widget is functional.

In the `TimingPath` class (which is what the `getTimingPaths` returns a list of) you should just be able to call `getSlack` for the slack on that path.
I hope this helps.

Sov-trotter:
> At the moment it's configured to return more than one path per endpoint. 

Thanks for adding this.

> Another thing to consider adding are controls for setup and hold slacks and corners (although those would be easy to get in once the widget is functional.

Sure. This is my first time with Qt, so learning as I go along. :)

[...]

PR #3344, opened 2023-05-17
Gui: Add charts widget to display endpoint slack histogram

AcKoucher:
Hello!
This is my first PR and I'm happy to be contributing to the project!

I tried to implement the widget to make as easy as possible to, in the future, add other plot categories (in the issue, it was said that the histogram widget could be used to display other things, e.g. congestion).

I added two API's to get the Endpoints and their slack that Matt kindly wrote to me. I put them in the place I thought to be convenient.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -56,7 +56,7 @@ if (Qt5_FOUND AND BUILD_GUI)
     src/gui_utils.cpp
     src/colorGenerator.cpp
     src/bufferTreeDescriptor.cpp
-    src/histogramWidget.cpp
+    src/chartsWidget.cpp
     resources/resource.qrc
   )
 
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -3,7 +3,7 @@ option(BUILD_GUI "Build the GUI" ON)
 
 # If Qt is not installed there will not be cmake
 # support for the package so this needs to be "quiet".
-find_package(Qt5 QUIET COMPONENTS Core Widgets)
+find_package(Qt5 QUIET COMPONENTS Core Widgets Gui Charts)
 
 include("openroad")
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
@@ -56,6 +56,7 @@ if (Qt5_FOUND AND BUILD_GUI)
     src/gui_utils.cpp
     src/colorGenerator.cpp
     src/bufferTreeDescriptor.cpp
+    src/slackWidget.cpp
     resources/resource.qrc
   )
 
@@ -64,7 +65,9 @@ if (Qt5_FOUND AND BUILD_GUI)
       odb
     PRIVATE
       Qt5::Core
+      Qt5::Gui
       Qt5::Widgets
+      Qt5::Charts
       utl
       Boost::boost
   )
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -1,12 +1,32 @@
 # Allow users to not build without the gui even if Qt is found
 option(BUILD_GUI "Build the GUI" ON)
 
-# If Qt is not installed there will not be cmake
-# support for the package so this needs to be "quiet".
-find_package(Qt5 QUIET COMPONENTS Core Widgets Gui Charts)
+# Allow users to build with the Charts widget
+option(CHARTS "Build GUI with Charts widget" OFF)
 
 include("openroad")
+
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+if(CHARTS)
+  message(STATUS "Charts widget enabled")
+
+  # If Qt is not installed there will not be cmake
+  # support for the package so this needs to be "quiet".
+  find_package(Qt5 QUIET COMPONENTS Core Widgets Gui Charts)
+
+  list(APPEND CHARTS_SRC
+    "src/chartsWidget.cpp"
+  )
+
+  list(APPEND CHARTS_LIBS
+    "Qt5::Charts"
+    "Qt5::Gui"
+  )
+
+else()
+  find_package(Qt5 QUIET COMPONENTS Core Widgets)
+endif()
 
 if (Qt5_FOUND AND BUILD_GUI)
   message(STATUS "GUI is enabled")
@@ -23,6 +43,10 @@ if (Qt5_FOUND AND BUILD_GUI)
            SCRIPTS   src/gui.tcl
            RUNTIME_HEADER tclSwig.h
   )
+
+  if(CHARTS)
+    target_compile_definitions(gui PRIVATE ENABLE_CHARTS)
+  endif()
 
   set_property(SOURCE ${GUI_WRAP} PROPERTY SKIP_AUTOMOC ON)
 
@@ -56,7 +80,7 @@ if (Qt5_FOUND AND BUILD_GUI)
     src/gui_utils.cpp
     src/colorGenerator.cpp
     src/bufferTreeDescriptor.cpp
-    src/chartsWidget.cpp
+    ${CHARTS_SRC}
     resources/resource.qrc
   )
 
@@ -65,9 +89,8 @@ if (Qt5_FOUND AND BUILD_GUI)
       odb
     PRIVATE
       Qt5::Core
-      Qt5::Gui
       Qt5::Widgets
-      Qt5::Charts
+      ${CHARTS_LIBS}
       utl
       Boost::boost
   )
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -56,7 +56,7 @@ if (Qt5_FOUND AND BUILD_GUI)
     src/gui_utils.cpp
     src/colorGenerator.cpp
     src/bufferTreeDescriptor.cpp
-    src/slackWidget.cpp
+    src/histogramWidget.cpp
     resources/resource.qrc
   )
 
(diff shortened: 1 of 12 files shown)


maliberty:
I see in the CI and on my centos7 machine that charts isn't found:
```
-- GUI is not enabled
```
What OS did you use for development and how did you install the new dependencies?

AcKoucher:
> I see in the CI and on my centos7 machine that charts isn't found:
> 
> ```
> -- GUI is not enabled
> ```
> 
> What OS did you use for development and how did you install the new dependencies?

I used Ubuntu 22.04 for development.
The new dependency is the package `libqt5charts5-dev`.

I installed it running via CL: `sudo apt install libqt5charts5-dev`

maliberty:
It needs to appear on the Windows menu like the other docking windows.

src/gui/src/histogramWidget.cpp
```diff
@@ -0,0 +1,242 @@
...
+    *pos_set << pos_buckets[i].size();
+    *neg_set << 0;
+    curr_slack = pos_buckets[i][0];
+    QString next_value;
+    time_values << open_bracket + curr_value.setNum(curr_slack) + comma + next_value.setNum(curr_slack+1) + close_parenthesis;
+    if(max_y < bucket_count)
+      max_y = bucket_count;
+  }
+
+  axis_x_->setTitleText("Slack [ns]");
```
maliberty, line 214: The time units depend on the .lib and should be taken from sta.  Some libraries may be in ps.

src/gui/src/histogramWidget.cpp
```diff
@@ -0,0 +1,242 @@
...
+  axis_x_->setVisible(true);
+
+  QStackedBarSeries* series = new QStackedBarSeries(this);
+
+  series->append(neg_set);
+  series->append(pos_set);
+  series->setBarWidth(1.0);
+  chart_->addSeries(series);
+
+  axis_y_->setTitleText("N. of Endpoints");
```
maliberty, line 226: N. -> Number

src/gui/src/histogramWidget.h
```diff
@@ -0,0 +1,82 @@
...
+  ~HistogramWidget() {} 
+
+  void setSTA(sta::dbSta* sta) { sta_ = sta; };
+
+  void setSlackMode();
+  void clearChart();
+  
+ signals:
+
+ public slots:
```
maliberty, line 66: No need for these if there are no signals or public slots

src/gui/src/histogramWidget.h
```diff
@@ -0,0 +1,82 @@
...
+
+namespace gui {
+
+class HistogramWidget : public QDockWidget
+{
+  Q_OBJECT
+
+ public:
+  HistogramWidget(QWidget* parent = nullptr);
+  ~HistogramWidget() {}
```
maliberty, line 57: Unneeded

src/gui/src/histogramWidget.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+const int SELECT = 0;
+const int SLACK_MODE = 1;
+
+HistogramWidget::HistogramWidget(QWidget* parent)
+    : QDockWidget("Histogram Plotter", parent),
+      sta_(nullptr),
+      label_(new QLabel(this)),
+      mode_menu_(new QComboBox(this)),
+      chart_(new QChart),
```
maliberty, line 59: Does chart_ have a parent?

src/gui/src/histogramWidget.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+  setObjectName("histogram_widget"); // for settings
+
+  QWidget* container = new QWidget(this);
+  QVBoxLayout* layout = new QVBoxLayout;
+  QFrame* controls_frame = new QFrame;
+  QHBoxLayout* controls_layout = new QHBoxLayout;
+
+  mode_menu_->addItem("Select Mode");
+  mode_menu_->addItem("Endpoint Slack");
```
maliberty, line 72: As there is only one choice currently, can we default it to Endpoint Slack rather than forcing the user to select it.

src/gui/src/histogramWidget.cpp
```diff
@@ -0,0 +1,242 @@
...
+  auto end_points = sta_gui.getEndPoints();
+  auto pin_iterator = end_points.begin();
+  std::vector<float> all_slack;
+  int unconstrained_count = 0;
+
+  for(int i = 0; i < end_points.size(); ++i) {
+    const int converter = 1e9;
+    float pin_slack = 0;
+    pin_slack = sta_gui.getPinSlack(*pin_iterator);
+    if(pin_slack != sta::INF)
```
maliberty, line 137: Please run clang-format

maliberty (changes requested):

maliberty:
Thanks this is a nice feature.  I do have some suggestions for enhancements but it can wait for subsequent PRs.

src/gui/src/histogramWidget.cpp
AcKoucher: The ownership of chart_ is passed to QChartView when the latter is initialized.
The order of ownership is `chart_`, `display_`, `layout`, `container` and finally the parent `QDockwidget`.

I just realized it would be more appropriate to initialize both QChart and QChartView like this:

```
display_(new QChartView(this)),
chart_(new QChart(display_))
```

AcKoucher: In QChart documentation, the constructor must be specified as:
`QChart::QChart(QGraphicsItem *parent = nullptr, Qt::WindowFlags wFlags = Qt::WindowFlags())`

There is no QGraphicsItem in the widget.

This is not a problem, because, when `display_` is initialized right after `chart_`, it takes its ownership.

To be more concise, I'll add a parent to `display_`, even though there is no leak, because `layout` takes its ownership in the constructor.

src/gui/src/histogramWidget.cpp
AcKoucher: Theoretically, using the QComboBox signal `activated()` instead of `currentIndexChanged()`  should produce the desired effect, but I tried that way and what happens in reality is that the dropdown menu, despite having only one option, needs to be clicked either way... Is too bad to leave like this? If yes, I can remove the dropdown menu.

maliberty: Hopefully we'll have some more choices soon.  We can leave it for now.

maliberty:
One general thought I have is that this should probably be a charts widget rather than a histogram widget.  As we add more choices they may not all be histograms.  If you want to save this change for another PR that's ok but I imagine it will be needed eventually.

AcKoucher:
> One general thought I have is that this should probably be a charts widget rather than a histogram widget. As we add more choices they may not all be histograms. If you want to save this change for another PR that's ok but I imagine it will be needed eventually.

No problem changing that now. I'll update the files' names. As for the title of the widget, what should the title be then? Right now it's Histogram Plotter.

maliberty:
> No problem changing that now. I'll update the widget. As for the title, what should it be then? Right now it's Histogram Plotter.

"Charts"

maliberty:
We will support OS with no qcharts (like centos7) by allowing the GUI to build but without this capability.  This is temporary until such OS support is dropped.

maliberty (changes requested):

src/gui/CMakeLists.txt
```diff
@@ -1,13 +1,33 @@
...
-# support for the package so this needs to be "quiet".
-find_package(Qt5 QUIET COMPONENTS Core Widgets Gui Charts)
+# Allow users to build with the Charts widget
+option(CHARTS "Build GUI with Charts widget" OFF)
 
 include("openroad")
+
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
+if(CHARTS)
```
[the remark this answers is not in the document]
AcKoucher: Should I check if QtCharts were found by using another `if(Qt5_FOUND)` inside `if(CHARTS)` ?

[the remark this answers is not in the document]
maliberty: You would use ```if (Charts_FOUND)``` inside

src/gui/CMakeLists.txt
```diff
@@ -1,13 +1,33 @@
 # Allow users to not build without the gui even if Qt is found
 option(BUILD_GUI "Build the GUI" ON)
 
-# If Qt is not installed there will not be cmake
-# support for the package so this needs to be "quiet".
-find_package(Qt5 QUIET COMPONENTS Core Widgets Gui Charts)
+# Allow users to build with the Charts widget
+option(CHARTS "Build GUI with Charts widget" OFF)
```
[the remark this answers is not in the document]
AcKoucher: Of course, but If I use `if (Charts_FOUND)` QtCharts is never found.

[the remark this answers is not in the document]
AcKoucher: It is possible to do:

`find_package(Qt5 QUIET COMPONENTS Core Widgets OPTIONAL_COMPONENTS Charts)`

so that `if(Qt5_FOUND)` will be true even if Qtcharts is not found. Then, check Qtchart using `if(TARGET Qt5::Charts)`.

It seems that Qt::Gui is not necessary.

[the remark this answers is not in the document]
maliberty: Sorry I had the syntax wrong but it looks like you've figured it out

AcKoucher:
Hey Matt, would it be possible to link this PR with #2649 ?

AcKoucher:
I just realized there is a problem in the case in which the sets are empty, for example, when there are no values in the positive or negative buckets. I'm fixing it. I'll push the changes as soon as I'm done.

maliberty:
> Hey Matt, would it be possible to link this PR with #2649 ?

The # reference acts as a link so I think you already did.

maliberty (approved):

PR #3356, opened 2023-05-18
Scan stitching

fgaray:
This is a first approach to scan stitching, we create by default new pins for the scan chain unless it is the scan out that we can reuse and only support one test mode, that is, there is a global scan enable for all the scan chains.

The result of scan stitch is the same as the one proposed when running preview_dft.


maliberty:
Is this ready for review?  I see it is still a draft.

fgaray:
Waiting for the unit tests to pass, I will send for review in a few minutes if they are green.

src/dft/src/cells/OneBitScanCell.hh
```diff
@@ -44,19 +46,25 @@ class OneBitScanCell : public ScanCell
...
+  void connectScanIn(const ScanDriver& driver) const override;
+  void connectScanOut(const ScanLoad& load) const override;
+  ScanDriver getScanOut() const override;
 
  private:
   odb::dbInst* inst_;
   sta::TestCell* test_cell_;
+  sta::dbNetwork* db_network_;
+
+  odb::dbITerm* findITerm(sta::LibertyPort* liberty_port) const;
```
maliberty, line 67: member functions before data members per https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/dft/src/utils/Either.hh
```diff
@@ -0,0 +1,127 @@
...
+{
+  Left,
+  Right
+};
+
+template <typename L>
+class Left
+{
+ public:
+  explicit Left(L left) : left_(left) {}
```
maliberty, line 48: Take arg by const& to avoid copying

src/dft/src/utils/Either.hh
```diff
@@ -0,0 +1,127 @@
...
+//      left();
+//    case EitherSide::Right:
+//      // do something with the double
+//      right();
+// }
+//
+// Since we use a union, you can't use C++ classes directly, only native types
+// and pointers
+template <typename L, typename R>
+class Either
```
maliberty, line 85: Why not use a std::variant?  It doesn't have these limitations and is more standard.

maliberty (commented):
I'd like to resolve the Either question before going further in the review as it is widely used.

src/dft/src/utils/Either.hh
fgaray: Thanks!, but I removed this file.

src/dft/src/utils/Either.hh
fgaray: Thanks!, much better now, I didn't know about std::variant

src/dft/src/cells/OneBitScanCell.hh
fgaray: Thanks!, I fixed other classes that were not correct.

src/dft/src/stitch/ScanStitch.cpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+  }
+
+  // Let's connect the last cell
+  ScanLoad scan_out_port
+      = FindOrCreateScanOut(block, last_scan_cell->getScanOut());
+  last_scan_cell->connectScanOut(scan_out_port);
+}
+
+void ScanStitch::FindOrCreateScanEnable(odb::dbBlock* block)
```
maliberty, line 104: Why doesn't this return anything?

src/dft/src/stitch/ScanStitch.cpp
```diff
@@ -0,0 +1,182 @@
...
+          = odb::dbBTerm::create(net, scan_enable_pin_name.c_str());
+      scan_enable_port->setSigType(odb::dbSigType::SCAN);
+      scan_enable_port->setIoType(odb::dbIoType::INPUT);
+      scan_enable_.emplace(scan_enable_port);
+      return;
+    }
+  }
+}
+
+ScanDriver ScanStitch::FindOrCreateScanIn(odb::dbBlock* block)
```
maliberty, line 125: FindOrCreateScanEnable, FindOrCreateScanIn, and FindOrCreateScanOut look to share a lot code that could be factored into a common method.

src/dft/src/cells/ScanCell.hh
```diff
@@ -47,23 +52,97 @@ class ClockDomain;
...
+    // Reuse any existing net if we are asked to preserve connections
+    if (preserve) {
+      odb::dbNet* driver_net = driver->getNet();
+      if (driver_net) {
+        debugPrint(logger_,
+                   utl::DFT,
+                   "scan_cell_connect",
+                   1,
+                   "    Preserving connections 1: Connecting driver net "
+                   "\"{:s}\" -> load pin \"{:s}\"",
```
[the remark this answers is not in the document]
fgaray: Done, thanks!

src/dft/src/stitch/ScanStitch.cpp
fgaray: It was setting scan_enable_ as a member of the ScanStitch object, but now I am returning a ScanDriver and passing it to the other methods.

src/dft/src/stitch/ScanStitch.cpp
fgaray: Done. I factored out the part of the creation into a private method in ScanStitch.

src/dft/src/stitch/ScanStitch.hh
```diff
@@ -48,18 +53,45 @@ class ScanStitch
...
+  ScanDriver FindOrCreateScanEnable(odb::dbBlock* block);
   ScanDriver FindOrCreateScanIn(odb::dbBlock* block);
   ScanLoad FindOrCreateScanOut(odb::dbBlock* block,
                                const ScanDriver& cell_scan_out);
 
+  // Typesafe function to create Ports for the scan chains.
+  template <typename Port>
+  Port CreateNewPort(odb::dbBlock* block, std::string_view name_pattern)
+  {
+    for (uint64_t port_number = 1;; ++port_number) {
```
maliberty, line 70: Are you expecting more the 2^32 ports?  From https://google.github.io/styleguide/cppguide.html#Integer_Types

```
We use int very often, for integers we know are not going to be too big, e.g., loop counters. Use plain old int for such things.```

fgaray: Good catch, thanks!

maliberty (approved):

PR #3360, opened 2023-05-19
etc/Env.sh: run from anywhere

oharboe:
@maliberty  Good to go I think.

Motivation, easier to file github issues correctly: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/1081

### DIFF
--- a/etc/Env.sh
+++ b/etc/Env.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 # Make sure we are on the correct folder before beginning
-cd "$(dirname $(readlink -f $0))"
+cd "$(dirname $(readlink -f $0))/../"
 
 mkdir -p build
 exec > >(tee -i build/openroad-env.log)
--- a/etc/Env.sh
+++ b/etc/Env.sh
@@ -2,8 +2,8 @@
 
 set -euo pipefail
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-cd $DIR/..
+# Make sure we are on the correct folder before beginning
+cd "$(dirname $(readlink -f $0))"
 
 mkdir -p build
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty Merge, fix or close?

etc/Env.sh
```diff
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $DIR/..
```
vvbandeira, line 6:
```suggestion
# Make sure we are on the correct folder before beginning
cd "$(dirname $(readlink -f $0))"
```

Since this script (`Env.sh`) is intended to be run and not sourced, I am partial to what we have in the `build_openroad.sh` from ORFS. Also, `$BASH_SOURCE` is bash-specific; macOS comes with `zsh` by default for example.

etc/Env.sh
```diff
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+# Make sure we are on the correct folder before beginning
+cd "$(dirname $(readlink -f $0))"
```
vvbandeira, line 6:
```suggestion
cd "$(dirname $(readlink -f $0))/../"
```

Sorry I jumped the gun on the copy-pasta

vvbandeira, line 6:
```suggestion
cd "$(dirname $(readlink -f $0))/../"
```

vvbandeira (changes requested):

vvbandeira (approved):

oharboe:
@maliberty Squash and rebase when merging to avoid git noise?

vvbandeira:
> @maliberty Squash and rebase when merging to avoid git noise?

Sounds good; I will do squash and merge after CI.

PR #3361, opened 2023-05-19
gui: Enable coalescing of mouse scroll events

rovinski:
Attempt to enable coalescing of mouse scroll events when the render time is very long. This should hopefully stop the event queue from backing up.

### DIFF
--- a/src/gui/src/gui.cpp
+++ b/src/gui/src/gui.cpp
@@ -1129,6 +1129,9 @@ int startGui(int& argc,
   font.setPointSize(12);
   QApplication::setFont(font);
 
+  // Enable coalescing of high-frequency events
+  QCoreApplication::setAttribute(Qt::AA_CompressHighFrequencyEvents);
+
   auto* open_road = ord::OpenRoad::openRoad();
 
   // create new MainWindow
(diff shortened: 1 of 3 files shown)


rovinski:
Doesn't seem to help.

src/gui/src/gui.cpp
```diff
@@ -1129,6 +1129,9 @@ int startGui(int& argc,
   font.setPointSize(12);
   QApplication::setFont(font);
 
+  // Enable coalescing of high-frequency events
+  QCoreApplication::setAttribute(Qt::AA_CompressHighFrequencyEvents);
```
oharboe, line 1133: doc says, `On the X11 windowing system, the default value is true`,  so I would expect no change...

https://doc.qt.io/qt-6/qt.html

rovinski: I'm using GWSL which I thought used Wayland, but maybe not. It was worth a shot because this is the only control I can find related to this issue.
I also tried `Qt::AA_CompressTabletEvents` because I thought maybe a laptop trackpad counts as that, but it didn't help either.

maliberty:
Do we have any empirical evidence that this helps?

rovinski:
No. Not sure what else to try.

PR #3363, opened 2023-05-21
gpl: improve error message upon too big die area

oharboe:
if the die area is too big, an out of error message is printed, rather than a std::vector() max_size() message.

out of memory is a hair better than the more internal error message from std::vector().

### DIFF
--- a/src/gpl/src/placerBase.cpp
+++ b/src/gpl/src/placerBase.cpp
@@ -1002,8 +1002,8 @@ void PlacerBase::initInstsForUnusableSites()
 {
   dbSet<dbRow> rows = db_->getChip()->getBlock()->getRows();
 
-  int siteCountX = (die_.coreUx() - die_.coreLx()) / siteSizeX_;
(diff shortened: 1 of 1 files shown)


src/gpl/src/placerBase.cpp
```diff
@@ -1002,8 +1002,8 @@ void PlacerBase::initInstsForUnusableSites()
 {
   dbSet<dbRow> rows = db_->getChip()->getBlock()->getRows();
 
-  int siteCountX = (die_.coreUx() - die_.coreLx()) / siteSizeX_;
-  int siteCountY = (die_.coreUy() - die_.coreLy()) / siteSizeY_;
+  int64 siteCountX = (die_.coreUx() - die_.coreLx()) / siteSizeX_;
+  int64 siteCountY = (die_.coreUy() - die_.coreLy()) / siteSizeY_;
```
maliberty, line 1006: int64_t is preferrable as it is defined in the c++ std and int64 is not and may not be present on all platforms.

maliberty (approved):

PR #3367, opened 2023-05-22
solving utf-8 bug when run GUI

luis201420:
Fixing bug discussed in #3314


maliberty:
@oharboe please verify this works for you as well

oharboe:
> @oharboe please verify this works for you as well

Aaah! 

Zero warnings! :-)


image

maliberty (approved):

Issue #3338, opened 2023-05-15
Fallback of io constraints is placed at oposite sides of macros at top level


oharboe:
### Description

When the io constraints fall back, because a full optimization is too time consuming, the leftmost position is used on the bottom of macro and rightmost positions are used on the top of the macro, which leads to a diagonal route:

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/d76755ca-a034-46db-8ae4-7724300fa7d4)

Another example of where placing the pins to the bottom left instead of bottom middle in the case of a fallback for placing a large group causes problems in placement. The pins at the bottom of the macro below are actually routed out of the macro. If the pins had been placed in the middle of the macro at the bottom of this macro, then this would have yielded a much better result in this case.

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/b364b311-ed5d-40c0-ad48-899815e6eed8)

### Suggested Solution

Place the pins in the middle of the assigned edge in the case of a fallback from full optimization.

Although it is possible to manually specify a range on the edge, this is incovenient during exploration where the dimensions of the design are in flux and we're iterating on finding a first cut reasonable floorplan.

eder-matheus:
Is this happening during random pin placement or during the regular pin placement?

eder-matheus:
@oharboe Could you try this: https://github.com/The-OpenROAD-Project/OpenROAD/pull/3374?

oharboe:
> @oharboe Could you try this: #3374?

gui_floorplan, pins are not centered. That's not a requirement, but would have been nice: faster turnaround times.

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/8cf8695f-50ed-409d-84c4-4d7ec37586e5)

eder-matheus:
Are they constrained to the edge? I have made a few tests with it and it was working on my end.

oharboe:
> Are they constrained to the edge? I have made a few tests with it and it was working on my end.

Yes. The bottom pin constraints are:

```
set_io_pin_constraint -group -order -region bottom:* -pin_names ...
```

oharboe:
Yes, this works!

gui_place gives the expected result, centered pins, at the bottom.

The only question that remains is whether you intended for pins to be centered already at the floorplanning stage...

[...]

PR #3374, opened 2023-05-23
ppl: place groups in fallback mode in the middle of the constraint region

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3338

Only constrained groups placed during fallback mode are placed in the middle of the constrained region.

### DIFF
--- a/src/ppl/test/large_groups3.ok
+++ b/src/ppl/test/large_groups3.ok
@@ -17,5 +17,5 @@ Found 0 macro blocks.
 [WARNING PPL-0042] Unsuccessfully assigned I/O groups.
 [WARNING PPL-0097] The max contiguous slots (35) is smaller than the group size (71).
 [WARNING PPL-0097] The max contiguous slots (35) is smaller than the group size (71).
-[ERROR PPL-0093] Pin group of size 71 does not fit in the constrained region 0-100.13 at TOP edge. First pin of the group is pin300.
+[ERROR PPL-0093] Pin group of size 71 does not fit in the constrained region 0.00-100.13 at TOP edge. First pin of the group is pin300.
 PPL-0093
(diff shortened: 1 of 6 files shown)


src/ppl/test/large_groups3.ok
```diff
@@ -12,13 +12,10 @@
...
-[INFO PPL-0006] Slots increase factor    0.01
-[INFO PPL-0008] Successfully assigned pins to sections.
-[INFO PPL-0012] I/O nets HPWL: 27153.80 um.
-No differences found.
+[WARNING PPL-0078] Not enough available positions (36) in section (47.215, 100.8)-(14.915, 100.8) at edge TOP to place the pin group of size 71.
+[WARNING PPL-0078] Not enough available positions (35) in section (85.215, 100.8)-(47.405, 100.8) at edge TOP to place the pin group of size 71.
+[WARNING PPL-0042] Unsuccessfully assigned I/O groups.
+[WARNING PPL-0097] The max contiguous slots (35) is smaller than the group size (71).
+[WARNING PPL-0097] The max contiguous slots (35) is smaller than the group size (71).
+[ERROR PPL-0093] Pin group of size 71 does not fit in the constrained region 0-100.13 at TOP edge. First pin of the group is pin300.
```
maliberty, line 20: This used to produce "Successfully assigned pins to sections" but now fails. Is this intended?

eder-matheus: Yes. In this case, placing the large group in the middle of the constrained region results in not enough room for the other group constrained to the same region.

In this case, I decided to generate an error instead of placing the group outside the constrained area.

maliberty (approved):

eder-matheus:
@maliberty This is ready to merge, but I can't do that myself. Could you merge it?

PR #3384, opened 2023-05-24
Multi power domain

ahmadelrouby:
Update GPL to handle multiple region placement. 
Inspired by the paper 'DREAMPlace 3.0: Multi-Electrostatics Based Robust VLSI Placement with Region Constraints', I divided each region into a separate placement instance with its own density targets. Wire length objective is common among all instances.


maliberty:
Conflicts and build errors need attention

maliberty:
@ZhiangWang033 please be aware that this change will land soon and may affect your GPU work

maliberty:
I've started a secure-CI

src/gpl/src/graphics.cpp
```diff
@@ -148,11 +157,13 @@ void Graphics::drawInitial(gui::Painter& painter)
 
 void Graphics::drawNesterov(gui::Painter& painter)
 {
+  // TODO: Support graphics for multiple Nesterov instances
```
maliberty, line 160: Wouldn't you just loop over nbVec_ and draw the bins of each?

src/gpl/src/initialPlace.cpp
```diff
@@ -150,39 +153,56 @@ void InitialPlace::doBicgstabPlace()
...
+        int domain_xMax = std::numeric_limits<int>::min();
+        int domain_yMax = std::numeric_limits<int>::min();
+        for (auto boundary : domain_region->getBoundaries()) {
+          domain_xMin = std::min(domain_xMin, boundary->xMin());
+          domain_yMin = std::min(domain_yMin, boundary->yMin());
+          domain_xMax = std::max(domain_xMax, boundary->xMax());
+          domain_yMax = std::max(domain_yMax, boundary->yMax());
+        }
+        inst->setCenterLocation(domain_xMax - (domain_xMax - domain_xMin) / 2,
+                                domain_yMax - (domain_yMax - domain_yMin) / 2);
```
maliberty, line 175: This will work for rectangular domains but might be outside of a domain with a more irregular shape (eg an L).  Will the placer recover if the starting point is outside?

maliberty (commented):
Generally looks good.  There are a few todo that we should revisit later but don't need to hold this PR for.  Waiting on CI results

PR #3386, opened 2023-05-25
dft: Adding support for clock mixing

fgaray:


src/dft/src/dft.tcl
```diff
@@ -47,7 +47,7 @@ proc insert_dft {} {
 
 proc set_dft_config { args } {
   sta::parse_key_args "set_dft_config" args \
-    keys {-max_length} \
+    keys {-max_length -clock_mixing} \
```
maliberty, line 50: The README should have command documentation.  Having that would help me to know what this option is for.

fgaray: I updated the README with the current status of the tool, thanks!

maliberty (approved):

Issue #2675, opened 2022-12-29
Resize the OpenROAD GUI Window vigourously and the GUI locks up when run from docker


oharboe:
### Describe the bug

Resize the OpenROAD GUI Window vigourously and the GUI locks up when run from docker.

Does not appear to happen when the OpenROAD GUI is run locally.

It is incredibly useful to be able to run OpenROAD GUI from a docker image when deploying OpenROAD.

### Expected Behavior

No lockup.

### Environment

```shell
Git commit: cf9e3b282573a32a7ae3678cd97074a5697276ae
kernel: Linux 5.15.0-56-generic
os: Ubuntu 22.04.1 LTS (Jammy Jellyfish)
cmake version 3.24.2
gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
Ubuntu clang version 14.0.0-1ubuntu1
```

### To Reproduce

1. create the bash script below somewhere, call it "or"
2. Run ```./or "make DESIGN_CONFIG=designs/asap7/gcd/config.mk floorplan && make DESIGN_CONFIG=designs/asap7/gcd/config.mk gui_floorplan"```
3. Resize vigourosly, at which point you get an error ```[WARNING GUI-0076] QXcbConnection: XCB error: 2 (BadValue), sequence: 1709, resource id: 1108, major code: 130 (Unknown), minor code: 3``` and the GUI freezes.

```shell
#!/bin/bash
set -ex
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

XSOCK=/tmp/.X11-unix
XAUTH=/tmp/.docker.xauth
xauth nlist :0 | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
ARGUMENTS=$@

if test -t 0; then
    DOCKER_INTERACTIVE=-ti
fi

docker run -u $(id -u ${USER}):$(id -g ${USER}) \
 -e "QT_X11_NO_MITSHM=1" \
 -e DISPLAY=$DISPLAY \
 -e QT_XKB_CONFIG_ROOT=/usr/share/X11/xkb \
 -v $XSOCK:$XSOCK -v $XAUTH:$XAUTH -e XAUTHORITY=$XAUTH \
 -v $DIR:$DIR \
 -v `pwd`:`pwd` \
 -v `pwd`/fakeram-results:/alpha-release/bsg_fakeram/results \
[...]

PR #3392, opened 2023-05-27
logging: fix lockup when resizing window in Docker

oharboe:
fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/2675

logging tried to render the log message synchronously, which
caused a lockup as the main event loop was processed during
the reporting of an Xcb error.

With the upcoming https://github.com/The-OpenROAD-Project/OpenROAD/pull/3370 fix to rendering where rendering is
cancelled and restarted when it takes to long, it's not
too bad to wait until the GUI is idle before updating the
log message in the GUI.

### DIFF
--- a/src/gui/src/scriptWidget.cpp
+++ b/src/gui/src/scriptWidget.cpp
@@ -380,7 +380,11 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
     // process widget event queue, if main thread will process new text,
     // otherwise there is nothing to process from this thread.
     if (QThread::currentThread() == widget_->thread()) {
-      QCoreApplication::sendPostedEvents(widget_);
+      if (!formatted_msg.contains("XcbConnection")) {
+        // there is a slim chance of a deadlock if we update the GUI
+        // from within a Qt error handler.
+        QCoreApplication::sendPostedEvents(widget_);
+      }
     }
   }
 
--- a/src/gui/src/scriptWidget.cpp
+++ b/src/gui/src/scriptWidget.cpp
@@ -376,6 +376,12 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
 
       widget_->addLogToOutput(formatted_msg, msg_color);
     }
+
+    // process widget event queue, if main thread will process new text,
+    // otherwise there is nothing to process from this thread.
+    if (QThread::currentThread() == widget_->thread()) {
+      QCoreApplication::sendPostedEvents(widget_);
+    }
   }
 
   void flush_() override {}
--- a/src/gui/src/scriptWidget.cpp
+++ b/src/gui/src/scriptWidget.cpp
@@ -376,12 +376,6 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
 
       widget_->addLogToOutput(formatted_msg, msg_color);
     }
-
-    // process widget event queue, if main thread will process new text,
-    // otherwise there is nothing to process from this thread.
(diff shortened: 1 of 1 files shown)


gadfort:
Once the threading pr @maliberty has is in I think we can get rid of this code, but right now it's needed to keep the logger responsive

oharboe:
> Once the threading pr @maliberty has is in I think we can get rid of this code, but right now it's needed to keep the logger responsive

Note that the other time than rendering that the gui event loop is blocked is during tcl scripts, loading, notably.

oharboe:
@maliberty OK to merge now that rendering is in a separate thread?

maliberty:
re " it's not too bad to wait until the GUI is idle before updating the log message in the GUI."

For long running commands we still want the output to appear as the command progresses.  This PR prevents that.  I don't see this as mergeable.

oharboe:
> re " it's not too bad to wait until the GUI is idle before updating the log message in the GUI."
> 
> For long running commands we still want the output to appear as the command progresses. This PR prevents that. I don't see this as mergeable.

I see  Any suggestions?

There really shouldnt be any long running blocking tasks on the GUI thread, but as a practical way forward could these long running tasks call the event handler?

maliberty:
That's what's happening now when the log a message in the code you suggest removing.

oharboe:
> That's what's happening now when the log a message in the code you suggest removing.

Yes, but this code is being called from Qt in an error handler under ill defined conditions where, obviously, it locks up.

Obviously it is not safe to call this method from anywhere under any conditions.

Logging should be safe to invoke anytime, anywhere.... Ideally.

src/gui/src/scriptWidget.cpp
```diff
@@ -376,12 +376,6 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
 
       widget_->addLogToOutput(formatted_msg, msg_color);
     }
-
-    // process widget event queue, if main thread will process new text,
-    // otherwise there is nothing to process from this thread.
-    if (QThread::currentThread() == widget_->thread()) {
-      QCoreApplication::sendPostedEvents(widget_);
```
oharboe, line 383: @maliberty How about if I checked for the XDG error message and skipped updates only for that message which is causing constipation?

oharboe:
@maliberty Tested. Works. What do you say, merge and forget all about it? :-)

src/gui/src/scriptWidget.cpp
maliberty: I'm ok with that.  Those messages are pretty useless anyways

gadfort:
@oharboe we have a place already for filtering messages (https://github.com/The-OpenROAD-Project/OpenROAD/blob/8d56294a54f42e0844ddf36cf7847f371bdb792a/src/gui/src/gui.cpp#L75) where this should go. However, I worry that the filtering is overly broad and doesn't have good context about the full message or set of messages that are causing trouble.

oharboe:
> @oharboe we have a place already for filtering messages (
> 
> https://github.com/The-OpenROAD-Project/OpenROAD/blob/8d56294a54f42e0844ddf36cf7847f371bdb792a/src/gui/src/gui.cpp#L75
> 
> ) where this should go. However, I worry that the filtering is overly broad and doesn't have good context about the full message or set of messages that are causing trouble.

I'm not filtering any messages, I'm just not updating the GUI immediately in the case where the error message contains this string. These messages are logged immediately to stderr.

The check is quite broad, but as the consequence is very slight, the GUI is updated a little bit later and there is no effect on stderr, I think this is good 'nuf to put lockup in Docker to bed.

maliberty (approved):

Issue #3385, opened 2023-05-24
Detailled Router complains about "Non orthogonal Routes" even though there aren't any


smunaut:
### Describe the bug

When trying to use OpenLane2 with a custom routing script that manually routes some signal before the Global+Detailled router passes, it fails with tons of 

```
[ERROR DRT-1010] Unsupported non-orthogonal wire begin=(1363820, 235100) end=(1513820, 1759000)
[ERROR DRT-1010] Unsupported non-orthogonal wire begin=(1363820, 235100) end=(1513820, 1759000)
[ERROR DRT-1010] Unsupported non-orthogonal wire begin=(1363820, 235100) end=(1513820, 1759000)
[ERROR DRT-1010] Unsupported non-orthogonal wire begin=(1363820, 235100) end=(1513820, 1759000)
[ERROR DRT-1010] Unsupported non-orthogonal wire begin=(1363820, 235100) end=(1513820, 1759000)
```

If in my custom routing script I leave the wires but remove the vias, then it works.

Test case data can be found on :
https://people.osmocom.org/tnt/stuff/tt/bug_data.zip

It contains the reproductable test case from OpenLane2 directly and the DEF output. ( Athough apparently it uses the ODB data directly rather than going through DEF between the different steps and the DEF is just "informational" ).

### Expected Behavior

Routing complete

### Environment

```shell
Git commit: 675438bce1190bd3ade9d24d674a161b175455ac
kernel: Linux 5.10.0-19-cloud-amd64
os: Debian GNU/Linux 11 (bullseye)
cmake version 3.18.4
-- The CXX compiler identification is GNU 10.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-8218-g675438bce
-- System name: Linux
-- Compiler: GNU 10.2.1
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- Found GTest: /usr/lib/x86_64-linux-gnu/libgtest.a  
-- The C compiler identification is GNU 10.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
[...]

PR #3411, opened 2023-05-31
drt: fix false non-orthogonal wires

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3385

### DIFF
--- a/src/drt/src/io/io.cpp
+++ b/src/drt/src/io/io.cpp
@@ -589,6 +589,7 @@ void io::Parser::setNets(odb::dbBlock* block)
     string shape = "";
     bool hasBeginPoint = false;
     bool hasEndPoint = false;
+    bool beginInVia = false;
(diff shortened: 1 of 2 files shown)


src/drt/src/io/io.cpp
```diff
@@ -589,12 +589,18 @@ void io::Parser::setNets(odb::dbBlock* block)
...
     frCoord beginY = -1;
     frCoord beginExt = -1;
+    frCoord nextX = -1;
+    frCoord nextY = -1;
     frCoord endX = -1;
     frCoord endY = -1;
     frCoord endExt = -1;
+    string prevLayer;
+    string lower_layer;
+    string top_layer;
```
maliberty, line 603: Why string/names and not dbTechLayer* ?

maliberty (approved):

PR #3415, opened 2023-06-01
Chain index

fgaray:
We need to call networkChanged otherwise we will get a crash when running report_worst_slack

image

### DIFF
--- a/src/dft/src/stitch/ScanStitch.cpp
+++ b/src/dft/src/stitch/ScanStitch.cpp
@@ -13,7 +13,7 @@ constexpr std::string_view kScanOutNamePattern = "scan_out_{}";
 
 namespace dft {
 
-ScanStitch::ScanStitch(odb::dbDatabase* db) : db_(db)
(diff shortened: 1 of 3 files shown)


src/dft/src/stitch/ScanStitch.cpp
```diff
@@ -35,6 +36,7 @@ void ScanStitch::Stitch(
   for (const std::unique_ptr<ScanChain>& scan_chain : scan_chains) {
     Stitch(block, *scan_chain, scan_enable);
   }
+  sta_->networkChanged();
```
maliberty, line 39: sta should be getting callbacks from odb dircectly.  This suggests some issue with callbacks that we should figure out rather than have to manually call this (which is also less efficient).

fgaray: I think this will take a while to debug, so for now I will cancel this PR. I think I found in dbInst where the calls to the callbacks are being made so I will try to add some logs there.

Thanks!,
Felipe

Issue #3422, opened 2023-06-02
Click on routing congestion and get some information on the cause


oharboe:
### Description

We sometimes have problems with routing congestion on top of a macro, from `make gui_cts` below. It is difficult to figure out where this routing congestion comes from. It is usually something simple to fix, like a pin being constrained to the wrong edge.

I have selected all nets less VSS and VDD below:

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/d5cd338c-df0b-49cb-9bc2-0ec1ef50202f)

In the image above, I *think* that it might be related to the clock that is the diagonal line seen in the picture, because I often see these sort of failures near the clock, but why would the routing congestion be outside of where the clock line is?

### Suggested Solution

In this case, it would be useful to click on the routing congestion to get more information on the matter(buffer tree? net?):

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/3ee25aa0-7d74-4cc9-9e24-f29c021c8f67)

oharboe:
@maliberty @eder-matheus @luis201420 FYI

eder-matheus:
@oharboe I've added the nets crossing the congested areas into the congestion report. Could you check it here: https://github.com/The-OpenROAD-Project/OpenROAD/pull/3424?

oharboe:
> @oharboe I've added the nets crossing the congested areas into the congestion report. Could you check it here: #3424?

I wont be able to test this before you merge it with master, I think, but I will definitely use this feature!

eder-matheus:
> > @oharboe I've added the nets crossing the congested areas into the congestion report. Could you check it here: #3424?
> 
> I wont be able to test this before you merge it with master, I think, but I will definitely use this feature!

Got it. Here's a screenshot of how it looks:
<img width="1022" alt="image" src="https://github.com/The-OpenROAD-Project/OpenROAD/assets/43450810/0f4c4a7b-d336-4c4f-8f70-4c2235ea8291">

eder-matheus:
> Sweet! And from this I can navigate to the buffertree to find the terminals, which I can then relate to Verilog?

Yes, I think you can. From the nets, you can go to all the terms connected to it, so I think you can get the information you want.

PR #3424, opened 2023-06-02
grt: update congestion report with net information

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3422


src/grt/src/GlobalRouter.cpp
```diff
@@ -221,17 +221,27 @@ void GlobalRouter::saveCongestion()
...
-  std::vector<std::pair<GSegment, TileCongestion>> congestionGridsV,
-      congestionGridsH;
+  std::vector<std::tuple<GSegment, TileCongestion, std::vector<std::string>>>
+      congestionGridsV, congestionGridsH;
   fastroute_->getCongestionGrid(congestionGridsV, congestionGridsH);
   for (auto& it : congestionGridsH) {
+    GSegment seg;
+    TileCongestion tile;
+    std::vector<std::string> srcs;
+    std::tie(seg, tile, srcs) = it;
```
maliberty, line 231: This could be simplified with a structured binding

const auto& [seg, tile, srcs] = it;

same below

src/grt/src/GlobalRouter.cpp
```diff
@@ -221,17 +221,27 @@ void GlobalRouter::saveCongestion()
...
-    const int usage = it.second.second;
-    out << "\tsrcs: \n";
+    const int capacity = tile.first;
+    const int usage = tile.second;
+    out << "\tsrcs: ";
+    if (!srcs.empty()) {
+      for (const auto& net : srcs) {
+        out << "net:" << net << " ";
+      }
+    }
```
maliberty, line 240: Why check if it is empty as the loop won't do anything in that case.  Does it make sense to print srcs: if there are none?  I'm not certain what the parser expects.

src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -150,9 +150,14 @@ class FastRouteCore
   int totalOverflow() const { return total_overflow_; }
   bool has2Doverflow() const { return has_2D_overflow_; }
   void updateDbCongestion();
+  std::vector<std::string> getNetsInCongestedEdge(int x, int y, bool vertical);
   void getCongestionGrid(
-      std::vector<std::pair<GSegment, TileCongestion>>& congestionGridV,
-      std::vector<std::pair<GSegment, TileCongestion>>& congestionGridH);
+      std::vector<
+          std::tuple<GSegment, TileCongestion, std::vector<std::string>>>&
+          congestionGridV,
```
maliberty, line 157: a struct rather than a tuple wold make this more compact and the fields more self-documenting

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -2130,34 +2130,99 @@ void FastRouteCore::mazeRouteMSMD(const int iter,
   v_cost_table_.clear();
 }
 
+std::vector<std::string> FastRouteCore::getNetsInCongestedEdge(int x,
```
maliberty, line 2133: Looping all the routing for each edge could be rather slow if there are many congested edges.  It might be better to store  all the edges in a hash by x/y->nets and the loop the routing once to update the nets.

src/grt/src/GlobalRouter.cpp
eder-matheus: I've moved the check above and will not print `"srcs: "` if it is empty.

eder-matheus:
@maliberty I've disabled the failing regression test. Since I'm using an unordered set, it will keep generating the ok file with a different order. But I decided to keep the test files in there, just to make it easier to compare possible new updates.

maliberty:
> @maliberty I've disabled the failing regression test. Since I'm using an unordered set, it will keep generating the ok file with a different order. But I decided to keep the test files in there, just to make it easier to compare possible new updates.

Can we sort the nets by name just before we write them out to make the results stable?

eder-matheus:
> > @maliberty I've disabled the failing regression test. Since I'm using an unordered set, it will keep generating the ok file with a different order. But I decided to keep the test files in there, just to make it easier to compare possible new updates.
> 
> Can we sort the nets by name just before we write them out to make the results stable?

It would work for the net names, but I'm also using the unordered_map to correlate the grid positions and the list of nets. It will also have instability and not be so straightforward to sort before write them out, I think.

src/grt/src/GlobalRouter.cpp
```diff
@@ -221,17 +221,25 @@ void GlobalRouter::saveCongestion()
...
+    const auto& [seg, tile, srcs] = it;
     out << "violation type: Horizontal congestion\n";
-    const int capacity = it.second.first;
-    const int usage = it.second.second;
-    out << "\tsrcs: \n";
+    const int capacity = tile.first;
+    const int usage = tile.second;
+    if (!srcs.empty()) {
+      out << "\tsrcs: ";
+      for (const auto& net : srcs) {
```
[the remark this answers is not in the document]
eder-matheus: I've decided to just use a `std::set` instead of `std::unordered_set`, since we're only adding values there once. I think it should not affect the runtime, and it prevents the necessity of copying the unordered set content to a new struct for sorting.

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -2130,34 +2130,83 @@ void FastRouteCore::mazeRouteMSMD(const int iter,
...
+        const std::vector<int16_t>& gridsX = treeedge->route.gridsX;
+        const std::vector<int16_t>& gridsY = treeedge->route.gridsY;
+        int lastX = tile_size_ * (gridsX[0] + 0.5) + x_corner_;
+        int lastY = tile_size_ * (gridsY[0] + 0.5) + y_corner_;
+
+        for (int i = 1; i <= routeLen; i++) {
+          const int xreal = tile_size_ * (gridsX[i] + 0.5) + x_corner_;
+          const int yreal = tile_size_ * (gridsY[i] + 0.5) + y_corner_;
+
+          nets_in_congested_edges[{lastX, lastY}].insert(
```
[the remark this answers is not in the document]
eder-matheus: Yeah, I think I misunderstood your previous suggestion. I've updated the code to add to the map only the edges with congestion, so the collection will not be very large.

maliberty:
I'll review once you sort out the unit test failures.

eder-matheus:
> I'll review once you sort out the unit test failures.

Now it should be working on all compilers.

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -2130,37 +2130,110 @@ void FastRouteCore::mazeRouteMSMD(const int iter,
...
-    std::vector<std::pair<GSegment, TileCongestion>>& congestionGridV,
-    std::vector<std::pair<GSegment, TileCongestion>>& congestionGridH)
+    std::vector<CongestionInformation>& congestionGridV,
+    std::vector<CongestionInformation>& congestionGridH)
 {
+  NetsPerCongestedArea nets_in_congested_edges;
+  std::map<std::pair<int, int>,
+                     TileCongestion,
+                     cmpPair>
+      edge_to_congestion;
```
maliberty, line 2190: Is there a need for two maps or can these be fused into one?

src/grt/src/GlobalRouter.cpp
```diff
@@ -221,17 +221,23 @@ void GlobalRouter::saveCongestion()
...
   fastroute_->getCongestionGrid(congestionGridsV, congestionGridsH);
   for (auto& it : congestionGridsH) {
+    const auto& [seg, tile, srcs] = it;
     out << "violation type: Horizontal congestion\n";
-    const int capacity = it.second.first;
-    const int usage = it.second.second;
-    out << "\tsrcs: \n";
+    const int capacity = tile.first;
+    const int usage = tile.second;
+    if (!srcs.empty()) {
```
maliberty, line 231: I don't think we need this check anymore as we can't have overflow with no sources

CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ project(OpenROAD VERSION 1
 
 set(OPENROAD_HOME ${PROJECT_SOURCE_DIR})
 
-# set(CMAKE_VERBOSE_MAKEFILE ON)
+set(CMAKE_VERBOSE_MAKEFILE ON)
```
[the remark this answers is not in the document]
eder-matheus: I've pushed the comment on this part by mistake. We have this on by default, and I remove it from my local file.
Not sure why we have it, but I can remove if you want.

maliberty (approved):

PR #3427, opened 2023-06-05
Add CMake variable for building with Address Sanitizer

### LINKED ISSUE #2738: CMake build flow support for ASAN
### Description

Address Sanitizer is a great way to discover memory leaks and memory corruption, and works even in optimized builds with small performance penalties. I'm not sure if there's a way today to configure the CMake based build to run with ASAN -- I don't see it mentioned in a search of the docs -- but it could be quite valuable, I think I see some instances where `Ath__array1D` objects are `new`'d and seem to fall out of scope, e.g. https://github.com/The-OpenROAD-Project/OpenROAD/blob/e41b1dcc4510b1a7dda30f4c7db7de4e60050579/src/rcx/src/extBenchDB.cpp#L407

### Suggested Solution

_No response_

### Additional Context

_No response_

Iyury1:
Created a new variable "ASAN" in the top-level CMakeLists.txt. This variable is OFF by default. To enable building with Address Sanitizer, set the ASAN variable to ON by passing the argument to CMake "-DASAN=ON".

When ASAN is ON, the src/CMakeLists.txt file adds compile and link options for using Address Sanitizer to the directory properties and also to the target properties. These compile and link options will be applied to all tools in the src directory and to openroad itself.

Fixes #2738

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -185,7 +185,16 @@ add_compile_options(
   $<$<CXX_COMPILER_ID:Clang>:-Wno-gnu-zero-variadic-macro-arguments>
   # Apple clang 14.0.0 deprecates sprintf, which generates 900 warnings.
   $<$<CXX_COMPILER_ID:AppleClang>:-Wno-deprecated-declarations>
+  $<$<BOOL:${ASAN}>:-fsanitize=address>
+  $<$<BOOL:${ASAN}>:-fno-omit-frame-pointer>
+  $<$<BOOL:${ASAN}>:-g>
 )
+
+if (ASAN)
+  add_link_options(
+    -fsanitize=address
+  )
+endif()
 
 ################################################################
 
@@ -273,7 +282,16 @@ add_executable(openroad
 target_compile_options(openroad
   PRIVATE
     -Wextra -pedantic -Wcast-qual
+  $<$<BOOL:${ASAN}>:-fsanitize=address>
+  $<$<BOOL:${ASAN}>:-fno-omit-frame-pointer>
+  $<$<BOOL:${ASAN}>:-g>
 )
+if (ASAN)
+  target_link_options(openroad
+    PRIVATE
+    -fsanitize=address
+  )
+endif()
 
 # Needed for boost stacktrace
 if(APPLE)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -191,9 +191,7 @@ add_compile_options(
 )
 
 if (ASAN)
-  add_link_options(
-    -fsanitize=address
-  )
+  add_link_options(-fsanitize=address)
 endif()
 
 ################################################################
@@ -287,10 +285,7 @@ target_compile_options(openroad
   $<$<BOOL:${ASAN}>:-g>
 )
 if (ASAN)
-  target_link_options(openroad
-    PRIVATE
-    -fsanitize=address
-  )
+  target_link_options(openroad PRIVATE -fsanitize=address)
 endif()
 
 # Needed for boost stacktrace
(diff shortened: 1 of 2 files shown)


QuantamHD:
I think something might be off with your history. It looks like something is a bit off with commit cbdc856d5062a3ecd98f7a03890459d8a7bcdc52. You might want to rebase and remove it.

Iyury1:
My fork was 331 commits behind OpenROAD:master, so I tried pulling the more recent commits on to the master branch of my fork, and then I used merge and squash to move those changes to my feature branch.

Should I be pulling the latest commits from OpenROAD into my fork before opening a pull request? I can try using rebase instead.

QuantamHD:
@maliberty Generally dislikes rebases, but in this case I think it's warranted.

I say drop https://github.com/The-OpenROAD-Project/OpenROAD/commit/cbdc856d5062a3ecd98f7a03890459d8a7bcdc52

and then `git fetch && git rebase origin/master` or whatever you've called the upstream repo in your setup.

Iyury1:
Does this look ok now?

QuantamHD:
Perfect!

src/CMakeLists.txt
```diff
@@ -185,8 +185,15 @@ add_compile_options(
   $<$<CXX_COMPILER_ID:Clang>:-Wno-gnu-zero-variadic-macro-arguments>
   # Apple clang 14.0.0 deprecates sprintf, which generates 900 warnings.
   $<$<CXX_COMPILER_ID:AppleClang>:-Wno-deprecated-declarations>
+  $<$<BOOL:${ASAN}>:-fsanitize=address>
+  $<$<BOOL:${ASAN}>:-fno-omit-frame-pointer>
+  $<$<BOOL:${ASAN}>:-g>
 )
 
+if (ASAN)
+  add_link_options(-fsanitize=address)
```
QuantamHD, line 194: Are both of the add_link_options required? What's the difference between this one and the one below on line 288?

Iyury1: I believe only add_compile_options and add_link_options are required. These commands modify the COMPILE_OPTIONS and LINK_OPTIONS properties for the directory, so the options will be added for all executables and libraries in the current directory and below _that are added after the command_. I was confused about this functionality and thought target_compile_options and target_link_options would be required as well.

I have removed the target_compile_options and target_link_options, and I am rebuilding to make sure it still works.

maliberty: I only dislike a rebase after an initial review as it breaks the show changes since last review. As this hasn't been reviewed yet it is fine to rebase before the initial review (at least by me)

Iyury1: I removed target_compile_options and target_link_options and rebuilt both Clang and GNU. Address sanitizer still produces the same output. Now only add_compile_options and add_link_options are used.

QuantamHD:
LGTM!

QuantamHD (approved):

QuantamHD (approved):

QuantamHD (approved):

QuantamHD (approved):

QuantamHD (approved):

QuantamHD (approved):

QuantamHD (approved):

QuantamHD (approved):

maliberty:
@vvbandeira please setup a pipeline with this enabled to run with unit tests and a small design or two.  Nightly should be enough.

rovinski:
Strongly approved 😂
image

QuantamHD:
@rovinski Lol. My computer lagged, and I clicked the button way too many times.

maliberty (approved):

PR #3438, opened 2023-06-07
dpl: Move Grid from pointers to vectors

QuantamHD:
I found a destructor related bug in one of Google's test cases. Rather than debug it I figured I'd just move it to vector, and not worry about it.

### DIFF
--- a/src/dpl/src/Grid.cpp
+++ b/src/dpl/src/Grid.cpp
@@ -191,6 +191,10 @@ void Opendp::initGrid()
   }
 }
 
+void Opendp::deleteGrid() {
+  grid_.clear();
+}
+
 Pixel* Opendp::gridPixel(int grid_idx, int grid_x, int grid_y) const
 {
   if (grid_idx < 0 || grid_idx >= grid_info_vector_.size()) {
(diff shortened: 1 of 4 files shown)


src/dpl/src/Grid.cpp
```diff
@@ -183,30 +183,14 @@ void Opendp::initGrid()
...
-          grid_[h_index][y][x].is_hopeless = true;
+          Pixel& pixel = grid_[h_index][y][x];
+          pixel.is_hopeless = true;
         }
       }
     }
   }
 }
 
-void Opendp::deleteGrid()
```
maliberty, line 193: This should do "grid_.clear();"

maliberty (changes requested):
I would have opted for boost::multi_array which is more efficient than either solution.

QuantamHD:
> I would have opted for boost::multi_array which is more efficient than either solution.

Seems like a good idea, but this one is about correctness, and not performance.

QuantamHD:
@maliberty Is this good to go?

maliberty (approved):

Issue #3039, opened 2023-03-19
Issues with timing driven global routing


antonblanchard:
### Describe the bug

Attached is a test case from https://github.com/The-OpenROAD-Project/OpenROAD/pull/2975. Some critical nets are getting badly routed (eg `soc0._331_`).

I had hoped `-critical_nets_percentage` would help, but it doesn't appear to. A few things:

- We don't do anything if the worst slack is positive:

```
  // Ensure the slack threshold is negative
  if (slack_th >= 0) {
    for (float slack : slacks) {
      if (slack >= 0)
        break;
      slack_th = slack;
    }
  }

  if (slack_th >= 0) {
    return;
  }
```

But it might be only slightly positive, and bad global routing it could make it much worse, driving the design into negative slack. I think we still want to prioritise the most critical nets.

- I worked around this (by removing the above code, and also setting a huge value for the default slack in `computeNetSlacks` so non critical nets get sorted last. Even with this, `-critical_nets_percentage` does a bad job of routing `soc0._331_`.

### Expected Behavior

Better routing of critical nets

### Environment

```shell
Git commit: f13f5828b53eb45ea040f46c2b6d9b41f2163f9b
os: Fedora Linux 37 (Workstation Edition)
cmake version 3.26.0-rc5
```

### To Reproduce

Test case is at https://ozlabs.org/~anton/junk/microwatt-grt-issue.tar.gz

antonblanchard:
While trying to fix this, I somehow managed to make it even worse. It makes me wonder if we should have a metric or a warning when global routes deviate a lot from their original Steiner tree.

![grt-long-route](https://user-images.githubusercontent.com/2029971/226206125-5e4b9089-808e-4e09-9c96-e1e0a4adeb3e.png)

maliberty:
We could have an ST wire length and a grt wirelength.  Some amount of deviation on a net is expected but this look pretty extreme.

antonblanchard:
> We could have an ST wire length and a grt wirelength. Some amount of deviation on a net is expected but this look pretty extreme.

Sounds good. `report_wire_length` gives us grt and drt wire length, so we could add an -st option

maliberty:
I think we need to update the net slacks during routing and not only take them at the start of routing.  @eder-matheus perhaps you could work with @arthurjolo on this?

PR #3452, opened 2023-06-12
Fastroute with make wire parasitics

arthurjolo:
Fixes [#3039](https://github.com/The-OpenROAD-Project/OpenROAD/issues/3039).

Currently the slack is estimated on the beginning of global routing, and never considers the changes made during the fastroute that can make a non-critical net to become one with a large detour from their original routing.
This changes update the slack estimates during the fastroute when the critical_nets_percentage is set, also for the nets in the critical percentage if they have a large detour from their original routing and weren't chosen for a rip up, decrease their search area and select them to be re-routed so they have another chance to get a routing closer to their original.


QuantamHD:
It feels like this option should be on by default, and have an option to turn it off.

arthurjolo:
> It feels like this option should be on by default, and have an option to turn it off.

Currently the slack is only considered if the critical_nets_percentage is set, so I decided to keep this property, but it culd be added a default percentage.

src/grt/include/grt/Grid.h
maliberty, line 1: I'm unclear why this and other private headers have been made public.  I don't see a change outside of grt requiring this.

arthurjolo: Some of them are needed to include the MakeWireParasitics.h in fastroute, but at the time I just made them all public and forgot to put the ones that are not needed back. I am going to look again and see witch ones can go back to private.

maliberty: The public include area is for clients outside of grt.  This is really about internal includes.  I would make an abstract interface in fastroute/include which MakeWireParasitics can implement in order to keep the layering and avoid having fastroute need to include from grt.

src/grt/src/fastroute/CMakeLists.txt
```diff
@@ -58,5 +58,6 @@ target_link_libraries(FastRoute4.1
     utl_lib
     stt_lib
     odb
+    dbSta_lib
```
maliberty, line 61: Why is this needed?

src/grt/src/MakeWireParasitics.cpp
maliberty, line 1: Is there any algorithmic change here or is this just moving code around?

src/grt/src/GlobalRouter.cpp
```diff
@@ -429,6 +429,8 @@ NetRouteMap GlobalRouter::findRouting(std::vector<Net*>& nets,
 {
   NetRouteMap routes;
   if (!nets.empty()) {
+    MakeWireParasitics builder(logger_, resizer_, sta_, db_->getTech(), this);
+    fastroute_->setMakeWireParasiticsBuilder(&builder);
     routes = fastroute_->run();
```
maliberty, line 434: It would be good to "fastroute_->setMakeWireParasiticsBuilder(nullptr)" after run() to ensure the builder isn't used after it goes out of scope.

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -682,6 +684,98 @@ NetRouteMap FastRouteCore::getRoutes()
   return routes;
 }
 
+NetRouteMap FastRouteCore::getPartialRoutes()
+{
+  NetRouteMap routes;
+  bool is_vertical = ((2 % 2) - layer_orientation_) != 0;
```
maliberty, line 690: 2 % 2 ??  Its a constant...

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -682,6 +684,98 @@ NetRouteMap FastRouteCore::getRoutes()
   return routes;
 }
 
+NetRouteMap FastRouteCore::getPartialRoutes()
```
maliberty, line 687: What is the definition of a partial route?

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -682,6 +684,98 @@ NetRouteMap FastRouteCore::getRoutes()
...
+        const std::vector<short>& gridsY = treeedge->route.gridsY;
+        const std::vector<short>& gridsL = treeedge->route.gridsL;
+        int lastX = tile_size_ * (gridsX[0] + 0.5) + x_corner_;
+        int lastY = tile_size_ * (gridsY[0] + 0.5) + y_corner_;
+        int lastL_h;
+        int lastL_v;
+        if (gridsL.empty()) {
+          if (is_vertical) {
+            lastL_h = 1;
+            lastL_v = 2;
```
maliberty, line 715: What are 1 & 2?  Are these directions?

src/grt/src/fastroute/src/RipUp.cpp
```diff
@@ -326,6 +326,71 @@ bool FastRouteCore::newRipupCheck(const TreeEdge* treeedge,
   }
 }
 
+bool FastRouteCore::newRipupCheck(const TreeEdge* treeedge,
```
maliberty, line 329: This looks to be a copy of the existing newRipupCheck.  Is it possible to avoid duplicating the code?

maliberty (changes requested):

src/grt/src/fastroute/CMakeLists.txt
arthurjolo: It is needed to include the definition of slack, but if needed I can change it to be just float instead.

src/grt/src/MakeWireParasitics.cpp
arthurjolo: Since your last rewiew just mooved some code, but in general there is a new function to estimate parasitics from information that fastroute has.

arthurjolo: But mostly it is the same code

src/grt/src/fastroute/src/FastRoute.cpp
arthurjolo: Yes, It can be one

src/grt/src/fastroute/src/FastRoute.cpp
arthurjolo: These are the layers definition for vertical edges and horizontal edges

src/grt/src/fastroute/src/RipUp.cpp
arthurjolo: Yes it can be one function, I created a new one because this fuction is used in other places than mazeRouteMSMD, and it should the added case should't be analized outside it, and at the time there was no way to difference them, but now there is, and I can change it.

src/grt/src/fastroute/src/FastRoute.cpp
arthurjolo: I think partial route was a bad naming, but it is the routes before layer assignment.

arthurjolo: I was trying to make it possible to use both before and after layer assignment, then it could be just one function with getRoutes(),  if this makes more sense I can do it.

maliberty:
> There is a public design failing and it is on investigation.

Is this resolved?  Has this been through secure CI?  Is there a default percentage?

arthurjolo:
> > There is a public design failing and it is on investigation.
> 
> Is this resolved? Has this been through secure CI? Is there a default percentage?

The public design failing was resolved, and it has been through a secure CI, I am waiting for @eder-matheus report on this run. As a default perecentege I didn't look that much into it, but for the test case attached to the issue 5% of critical nets seemd to be the best alternative, the secure run also used 5%.

eder-matheus:
> > > There is a public design failing and it is on investigation.
> > 
> > 
> > Is this resolved? Has this been through secure CI? Is there a default percentage?
> 
> The public design failing was resolved, and it has been through a secure CI, I am waiting for @eder-matheus report on this run. As a default perecentege I didn't look that much into it, but for the test case attached to the issue 5% of critical nets seemd to be the best alternative, the secure run also used 5%.

We actually have 2 separate runs in the CI. When using the default percentage of 0% (i.e., the updates from this PR are not executed), public and private CI are passing.

When using 5%, we still have some metric failures.

From the CI side, I'm fine with merging this PR and do the fix for the matric failures with 5% in a future PR.

maliberty:
> From the CI side, I'm fine with merging this PR and do the fix for the matric failures with 5% in a future PR.

agreed

src/grt/src/fastroute/src/FastRoute.cpp
maliberty: If it can convert both sorts of routes that is good.  If not we should rename it.

arthurjolo: Oh it seems it wasn’t pushed, but I made both methods into one. I will push it.

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -706,61 +708,53 @@ NetRouteMap FastRouteCore::getPartialRoutes()
...
-          int second_x = tile_size_ * (gridsX[1] + 0.5) + x_corner_;
-          lastL = (lastX == second_x) ? layer_v : layer_h;
+
+        if (layer_orientation_ != 0) {
+          layer_h = 1;
+          layer_v = 2;
         } else {
-          lastL = gridsL[0];
+          layer_h = 1;
+          layer_v = 2;
```
rovinski, line 722: If and else block are the same?

src/grt/src/fastroute/src/FastRoute.cpp
arthurjolo: lastL_h and lastL_v were renamed to layer_h and layer_v respectively.

src/grt/src/fastroute/src/FastRoute.cpp
arthurjolo: I tried to merge this method with the method getRoutes(), but this caused some errors in other parts of the flow. So this method was renamed to getPlanarRoutes(), for better understanding of it's purpose: generate the routes before layer assignment.

src/grt/src/MakeWireParasitics.cpp
```diff
@@ -307,6 +367,111 @@ void MakeWireParasitics::makeParasiticsToPin(
...
+    auto ap_position = ap_positions.front();
+    pt = ap_position.first;
+    grid_pt = ap_position.second;
+  }
+  if (pin.isConnectedToPadOrMacro() || pin.isPort()) {
+    grid_pt = grouter_->findFakePinPosition(pin, net);
+  }
+  // Use the route layer above the pin layer if there is a via
+  // to the pin.
+  int layer = 3;
```
maliberty, line 412: Why 3?

arthurjolo: I define the layers for planar routes as m2 or m3.

arthurjolo: with layer_h and layer_v beeing 1 (m2) or 2 (m3).

arthurjolo: And it had to be hardcoded because the it wasn't possible to use the pin access point before layer assignment.

maliberty (approved):

Issue #3404, opened 2023-05-30
Add inspector for dbBlock and dbTech


maliberty:
### Description

Add to dbDescriptor dbBlock and dbTech with their associated fields and links to related objects.

### Suggested Solution

See the other descriptors in src/gui/src/dbDescriptors.*

AcKoucher:
As #3552 and #3458  were merged and all the requested properties are available for inspection, can this issue be closed?

PR #3458, opened 2023-06-14
Gui: Add inspector for dbBlock and dbTech

AcKoucher:
This PR is related to #3404.

I left same-net rules without a link to Tech, because there was no API to get tech directly from it. If it's desirable to have this link, just let me know, as I can add it by getting the tech from the database instance db_.

### DIFF
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -374,6 +374,12 @@ Descriptor::Properties DbBlockDescriptor::getProperties(std::any object) const
   auto gui = Gui::get();
 
   Properties props;
+  SelectionSet modules;
+  for (auto module : block->getModules()) {
+    modules.insert(gui->makeSelected(module));
+  }
+  props.push_back({"Modules", modules});
+
   SelectionSet bterms;
   for (auto bterm : block->getBTerms()) {
     bterms.insert(gui->makeSelected(bterm));
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -249,17 +249,83 @@ std::string DbBlockDescriptor::getTypeName() const
 bool DbBlockDescriptor::getBBox(std::any object, odb::Rect& bbox) const
 {
   /*Why this returns a bool?
-    Because the rect dimensions are being passed by reference*/
+    Because the rect dimensions are being passed by reference.
+    Ok, but why not void?*/
   
   auto block = std::any_cast<odb::dbBlock*>(object);
   bbox = block->getBBox()->getBox();
   return true;
 }
 
-// void DbBlockDescriptor::highlight(std::any object, Painter& painter) const
-// {
+/*Now highlighting only the nets and instances*/
+void DbBlockDescriptor::highlight(std::any object, Painter& painter) const
+{
+  auto block = std::any_cast<odb::dbBlock*>(object);
 
-// }
+  auto inst_descriptor = Gui::get()->getDescriptor<odb::dbInst*>();
+  for(auto inst : block->getInsts()) {
+    inst_descriptor->highlight(inst, painter);
+  }
+
+  auto net_descriptor = Gui::get()->getDescriptor<odb::dbNet*>();
+  for(auto net : block->getNets()) {
+    net_descriptor->highlight(net, painter);
+  }
+}
+
+Descriptor::Properties DbBlockDescriptor::getProperties(std::any object) const
+{
+  auto block = std::any_cast<odb::dbBlock*>(object);
+
+  auto gui = Gui::get();
+
+  Properties props;
+  SelectionSet modules;
+  for (auto module : block->getModules()) {
+    modules.insert(gui->makeSelected(module));
+  }
+  props.push_back({"Modules", modules});
+
+  SelectionSet insts;
+  for (auto inst : block->getInsts()) {
+    insts.insert(gui->makeSelected(inst));
+  }
+  props.push_back({"Instances", insts});
+
+  SelectionSet nets;
+  for (auto net : block->getNets()) {
+    nets.insert(gui->makeSelected(net));
+  }
+  props.push_back({"Nets", nets});
+  //populateODBProperties(props, block);
+  return props;
+}
+
+Selected DbBlockDescriptor::makeSelected(std::any object) const
+{
+  if (auto block = std::any_cast<odb::dbBlock*>(&object)) {
+    return Selected(*block, this);
+  }
+  return Selected();
+}
+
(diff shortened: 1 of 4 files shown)


src/gui/src/dbDescriptors.cpp
```diff
@@ -230,6 +230,206 @@ static odb::dbTechLayer* getLayerSelection(odb::dbTech* tech,
...
+  for (auto inst : block->getInsts()) {
+    insts.insert(gui->makeSelected(inst));
+  }
+  props.push_back({"Insts", insts});
+
+  SelectionSet iterms;
+  for (auto iterm : block->getITerms()) {
+    iterms.insert(gui->makeSelected(iterm));
+  }
+  props.push_back({"ITerms", iterms});
```
maliberty, line 381: Let's drop this.  You can access these through the dbInst and this list is likely to be huge.

src/gui/src/dbDescriptors.cpp
```diff
@@ -230,6 +230,206 @@ static odb::dbTechLayer* getLayerSelection(odb::dbTech* tech,
...
+bool DbTechDescriptor::getBBox(std::any object, odb::Rect& bbox) const
+{
+  return false;
+}
+
+void DbTechDescriptor::highlight(std::any object, Painter& painter) const
+{
+}
+
+Descriptor::Properties DbTechDescriptor::getProperties(std::any object) const
```
maliberty, line 256: Please add getManufacturingGrid()  (check hasManufacturingGrid first), getNonDefaultRules(), and getMetalWidthViaMap()

maliberty (changes requested):

src/gui/src/dbDescriptors.cpp
AcKoucher: Should we keep the link to block inspection in the iterms?

maliberty: No - they should just link to the instance.  The normal hierarchy is block<->inst<->iterm.

maliberty:
@gadfort FYI in case you have comments

src/gui/src/dbDescriptors.cpp
```diff
@@ -230,6 +230,206 @@ static odb::dbTechLayer* getLayerSelection(odb::dbTech* tech,
...
+  for (auto tech_layer : tech->getLayers()) {
+    tech_layers.insert(gui->makeSelected(tech_layer));
+  }
+  props.push_back({"Tech Layers", tech_layers});
+
+  SelectionSet tech_vias;
+  for (auto tech_via : tech->getVias()) {
+    tech_vias.insert(gui->makeSelected(tech_via));
+  }
+  props.push_back({"Tech Vias", tech_vias});
```
gadfort, line 275: there are also getViaRules() and getViaGenerateRules() that should be included.

src/gui/src/dbDescriptors.cpp
```diff
@@ -230,6 +230,206 @@ static odb::dbTechLayer* getLayerSelection(odb::dbTech* tech,
...
+  for (auto region : block->getRegions()) {
+    regions.insert(gui->makeSelected(region));
+  }
+  props.push_back({"Regions", regions});
+
+  SelectionSet insts;
+  for (auto inst : block->getInsts()) {
+    insts.insert(gui->makeSelected(inst));
+  }
+  props.push_back({"Insts", insts});
```
gadfort, line 375: Maybe spell out `Instances`

src/gui/src/dbDescriptors.cpp
```diff
@@ -230,6 +230,206 @@ static odb::dbTechLayer* getLayerSelection(odb::dbTech* tech,
...
+    obstructions.insert(gui->makeSelected(obstruction));
+  }
+  props.push_back({"Obstructions", obstructions});
+
+  SelectionSet rows;
+  for (auto row : block->getRows()) {
+    rows.insert(gui->makeSelected(row));
+  }
+  props.push_back({"Rows", rows});
+
```
gadfort, line 400: What about `getTopModule()`, `getMOdules()`, and `getChildren()`, `getVias()`
core and die area?

gadfort (commented):
I think there are some additional low handing fruits to be added, but otherwise looks good.

src/gui/src/dbDescriptors.cpp
maliberty: I agree there are a bunch more but I don't want to overload this PR.  core and die area seem pretty easy.

src/gui/src/dbDescriptors.cpp
maliberty: Ideally all rules would have a getTech() method.  The db at this point is used with a single tech so I guess nobody bothered.  We may have to change that in the future but for now you can just the the tech from the db.

dbMetalWidthViaMap is usually a table in LEF (dummy example):
```
 LIBRARY LEF58_METALWIDTHVIAMAP STRING "
  METALWIDTHVIAMAP
   VIA   v1    1.000              1.000               VIA11
   VIA   v1    1.000              2.000               VIA12
   VIA   v1    1.000              3.000               VIA13
   VIA   v1    2.000              1.000               VIA21
   VIA   v1    2.000              2.000               VIA22
   VIA   v1    2.000              3.000               VIA23
```
If we could display it that way it would be nice.

src/gui/src/dbDescriptors.cpp
AcKoucher: Should I add the all or just core and die area?

src/gui/src/dbDescriptors.cpp
AcKoucher: There is not a descriptor for via-rules, what should it display in the tech inspection?

src/gui/src/dbDescriptors.cpp
AcKoucher: I'm sorry, i'm a bit confused. In your example there are two values per line, but, as far as I understood there are four values that can be returned from dbMetalWidthViaMap:

below_layer_width_low
below_layer_width_high
above_layer_width_low
above_layer_width_high

I was trying to print the values so I could understand better but, do the technologies in the test folder have a metal width via map?. I've run different ones and I got an empty map when calling `tech->getMetalWidthViaMap()`. Maybe I'm doing something wrong.

AcKoucher:
I opened #3477 to keep track of what I did not implement.

maliberty:
Is there still work pending here?

AcKoucher:
There are a few properties that could still be added in the future, I've listed them in #3477. I left them out as you said you didn't want to overload the PR. Apart from those, everything else commented in this PR is implemented.

maliberty (approved):

PR #3462, opened 2023-06-14
Logic cloning undo support2

openroadie:


openroadie:
@tspyrou This looks good at least on the one test. Let me check what I get from ORFS tests. Main changes:
- Changed the order of applying transforms.
- Added undo support
- No hill climbing (yet) for cloning. If cloning does not improve timing ... we undo it right away. 
Currently limited to high fanout nets (where cloning makes a lot more sense than buffering)

openroadie:
The code right now seems to work fine for high fanout nets (the limit for splitting a net was 20 ... using the same for gate cloning). I will run a few ORFS designs to see if we any of them exercises gate cloning.

PR #3472, opened 2023-06-15
Grt incremental updates

luis201420:
- **RSZ:** Checking resized instance positions after updating instance dimensions.

- **GRT:** Modifying the way to add new nets in case too high congestion is obtained in a repair_design iteration. All the nets that pass through the overflow zone are obtained, if there is no difference with the old dirty nets then the radius around the overflow zones is increased to obtain new nets that help to solve the congestion error.


src/rsz/src/Resizer.cc
```diff
@@ -960,6 +954,13 @@ Resizer::replaceCell(Instance *inst,
     sta_->replaceCell(inst, replacement_cell1);
     designAreaIncr(area(replacement_master));
 
+    // Legalize the position of the instance in case it leaves the die
+    if (parasitics_src_ == ParasiticsSrc::global_routing) {
+      opendp_->legalCellPos(db_network_->staToDb(inst));
+    } else if (parasitics_src_ == ParasiticsSrc::global_routing) {
+      logger_->error(RSZ, 91, "Opendp was not initialized before resized an instance");
+    }
```
gadfort, line 962: this doesn't look like it does anything since `parasitics_src_ == ParasiticsSrc::global_routing` is already checked

luis201420: I fixed the mistake, thanks

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -2254,23 +2259,64 @@ void FastRouteCore::setCongestionNets(int& posX, int& posY, int dir)
...
+        ys.push_back(i);
+        dirs.push_back(0);
+        n++;
+      }
+    }
+  }
+
+  int old_size = congestion_nets.size();
+
+  for (int ratio = 0; ratio < 5 && old_size == congestion_nets.size();
```
eder-matheus, line 2310: Could you explain what is the `ratio` variable, here and in `setCongestionNets`?

luis201420: The variable name is wrong, this variable serves to increase the radius around the congested zone in case new nets are not obtained.

eder-matheus: LGTM, but I'll let @maliberty give his review.

maliberty (approved):

PR #3474, opened 2023-06-16
Gate cloning off by default

openroadie:
@maliberty @tspyrou : Here is the PR to turn off gate cloning by default. It also changes the option in the TCL command to:

-enable_gate_cloning so we can evaulate the ORFS tests and enable as needed. The two regressions have also been updated.


maliberty:
ORFS is now using skip_gate_cloning so this will require a matching ORFS PR.

openroadie:
That ORFS  PR is already in draft.

maliberty (approved):

PR #3482, opened 2023-06-18
DPL: add full support for regions and groups 

mina1460:
From the LEF/DEF manual:

REGION regionName:
- Specifies a region in which the component must lie. regionName specifies a region already defined in the REGIONS section. 
- If the region is smaller than the bounding rectangle of the component itself, the DEF reader issues an error message and ignores the argument. 

- If the region does not contain a legal location for the component, the component remains unplaced after the placement step.

There are two regions types according to the guide:
* FENCE: All instances assigned to this type of region must be exclusively placed inside the region boundaries. No other instances are allowed inside this region.
* GUIDE: All instances assigned to this type of region should be placed inside this region; however, it is a preference, not a hard constraint. Other constraints, such as wire length and timing, can override this preference

Based on my understanding, DPL currently supports GUIDE or something very similar to it. I saw multiple passing tests for FENCE, but in this PR, I add a new unit test that uses the FENCE for regions and yet, DPL doesn't adhere to the constraint. 

After this PR merges, DPL should have full-support for these cases and the tests should pass. I follow the TDD approach.

Test case:
There are two regions, and 4 cells. I assign two cells to each region in the def file and assign the groups accordingly.
Screenshot 2023-06-18 at 5 54 05 pm

Running detailed_placement:

Running detailed_placement with disallow one site gaps:

With one-site gaps          |  Without one-site gaps
:-------------------------:|:-------------------------:
Screenshot 2023-06-18 at 5 54 17 pm  | Screenshot 2023-06-18 at 5 55 00 pm

Here is another test case with a the bottom row being a region on its own. then the upper row has the left part of it as a defined region, but not the right part. I then create a cell, and don't assign it to any region. We can also see that DPL correctly placed it in the right part that has no regions thanks to the fast lookups with the R-tree from Boost::Geometry.
Screenshot 2023-06-18 at 6 01 17 pm

If we were to keep the cell outside of all regions, yet assign both rows to regions completely:

We get this error:
```
>>> detailed_placement
[INFO DPL-0034] Detailed placement failed on the following 1 instances:
[INFO DPL-0035]  f2/_285_
[ERROR DPL-0036] Detailed placement failed.
[ERROR GUI-0070] DPL-0036
```

Added a multi-height test and it is working perfectly with @QuantamHD 's PR

Screenshot 2023-06-27 at 2 28 21 pm


src/dpl/test/regions2.ok
maliberty, line 1: why is this empty?

src/dpl/src/Place.cpp
```diff
@@ -938,6 +929,41 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
...
+        //            "detailed",
+        //            1,
+        //            "Point {} {} Cell {} has region x[{} {}] and y[{} {}]",
+        //            p.getX(),
+        //            p.getY(),
+        //            cell->name(),
+        //            cell->region_->xMin(),
+        //            cell->region_->xMax(),
+        //            cell->region_->yMin(),
+        //            cell->region_->yMax());
```
maliberty, line 947: rm

src/dpl/src/Place.cpp
```diff
@@ -946,6 +972,39 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
...
```
maliberty, line 990: rm

src/dpl/src/Place.cpp
```diff
@@ -946,6 +972,39 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
...
+        if (!cell->region_->intersects(p)) {
+          debugPrint(logger_,
+                     DPL,
+                     "detailed",
+                     2,
+                     "point and region doesn't interesect!");
```
maliberty, line 996: What is being checked?

src/dpl/src/Place.cpp
```diff
@@ -956,6 +1015,97 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
...
+                 "Overlap found between queryBox ({} , {}) - ({} , {}) and bbox"
+                 "({} , {}) - ({}, {})",
+                 queryBox.min_corner().x(),
+                 queryBox.min_corner().y(),
+                 queryBox.max_corner().x(),
+                 queryBox.max_corner().y(),
+                 box.min_corner().x(),
+                 box.min_corner().y(),
+                 box.max_corner().x(),
+                 box.max_corner().y());
```
maliberty, line 1074: It would be nice to have stream operators so you can just write
 "Overlap found between queryBox {} and bbox {}", queryBox, box

src/dpl/src/Place.cpp
mina1460: instead of doing the rtree query, I check if the initial point intersects with the region or not. If not, I simply skip the rest of the logic. If it does, we check if the whole cell will fit inside the region.

src/dpl/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@
 
 include("openroad")
 find_package(TCL)
+find_package(Boost REQUIRED COMPONENTS serialization)
```
maliberty, line 41: Why do you need serialization ?

src/dpl/src/dbToOpendp.cpp
```diff
@@ -261,7 +262,16 @@ void Opendp::makeGroups()
...
           // offset region to core origin
           box.moveDelta(-core_.xMin(), -core_.yMin());
-
+          if (height == *(unique_heights.begin())) {
+            bgBox bbox(
+                bgPoint(box.xMin(), box.yMin()),
+                bgPoint(
+                    box.xMax() - 1,
+                    box.yMax()
+                        - 1));  /// the -1 is to prevent imagianring overlaps
```
maliberty, line 271: imagianring  ?

An alternative is to check that the overlap area is non-zero.  Does shrinking the rect imply an issue if an instances touches the region edge?

src/dpl/test/region_constraint.tcl
maliberty, line 1: What is the purpose of this file?  It doesn't seem to be part of any tests.

src/dpl/src/Place.cpp
```diff
@@ -61,6 +61,13 @@ using std::vector;
 
 using utl::DPL;
 
+std::string printBgBox(const boost::geometry::model::box<bgPoint>& queryBox)
+{
+  return "(" + std::to_string(queryBox.min_corner().x()) + ", "
```
maliberty, line 66: fmt::format would simplify this a bit

src/dpl/src/dbToOpendp.cpp
mina1460: I noticed this issue when I was debugging the failing fence01.tcl test
The regions are defined in the file as follows:
```
REGIONS 2 ;
- er0 ( 0 0 ) ( 67780 30800 ) + TYPE FENCE ;
- er1 ( 0 30800 ) ( 67780 61600 ) + TYPE FENCE ;
```

The point 0, 30800 and all the points on that horizontal line seems to be inside both regions, which I think is meaningless. Thus, I assumed the first region would start at 0 and end at (30800 - 1)  and the new one starts at 30800. The same applies to the vertical boundaries in other test cases. 

Without this, if the instance touched 30800, the r-tree found 2 overlaps, one with each region. I think my code doesn't change anything but the r-tree regions. The cell regions are untouched, and it would still be from 0 to 30800.

mina1460: tl; dr;
This is not affecting the cell regions themselves, but rather the r-tree regions that is used to count the overlaps. I assume 0 to 30800 is exclusive of the right-boundary -> [0, 30800) and [30800, 61600)

This is because the chip size in the y-dimensions is 61600, yet from 0 to 61600 is 61601. So my assumption makes sense. Otherwise, the regions defined in all the DEF files need modification to avoid overlaps.

mina1460: imagianring is a typo. I meant imaginary.

maliberty: If you are insetting only the top/right but not the bottom/left wouldn't you have a similar issue on those edges?

mina1460: I am a bit confused. Can you elaborate what is the problem you are referring to? 
I am inserting two points that define the bottom-left and the top-right corners. They are used to define the rectangle between them that is inserted in the r-tree.

maliberty: You subtract one from the max points but you don't add one to the min points.

mina1460: yes, but for example if region 1 is from 0 to 5 and the other region is from 5 to 10. Then region 1 includes the 0, but not 5. similarly, region 2 would include 5 and not the 10.

I don't think I need to add 1 to the min as subtracting 1 from the max assures no overlaps.

src/dpl/src/Place.cpp
```diff
@@ -938,6 +869,15 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
 
   if (x > bin_x) {
     for (int i = bin_search_width_ - 1; i >= 0; i--) {
+      Point p(bin_x + i, bin_y);
+      p.setX(p.getX() * site_width_);
+      p.setY(p.getY() * row_height);
+      if (cell->region_) {
+        if (!cell->region_->intersects(p)) {
```
maliberty, line 876: no need to nest - just

 if (cell->region_ && !cell->region_->intersects(p)) {

src/dpl/src/Place.cpp
```diff
@@ -938,6 +869,15 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
 
   if (x > bin_x) {
     for (int i = bin_search_width_ - 1; i >= 0; i--) {
+      Point p(bin_x + i, bin_y);
+      p.setX(p.getX() * site_width_);
+      p.setY(p.getY() * row_height);
```
maliberty, line 874: collapse to a single statement

src/dpl/src/Place.cpp
```diff
@@ -946,6 +886,14 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
...
   } else {
     for (int i = 0; i < bin_search_width_; i++) {
+      Point p(bin_x + i, bin_y);
+      p.setX(p.getX() * site_width_);
+      p.setY(p.getY() * row_height);
+      if (cell->region_) {
+        if (!cell->region_->intersects(p)) {
+          continue;
+        }
+      }
```
maliberty, line 896: same as above

src/dpl/src/Place.cpp
```diff
@@ -956,6 +904,61 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
...
+  if (cell->region_) {
+    if (result.size() == 1) {
+      if (boost::geometry::covered_by(queryBox, result[0])) {
+        // the queryBox must be fully contained in the region or else there
+        // might be a part of the cell outside of any region
+        return true;
+      }
+      // if we are here, then the queryBox is not fully contained in the
+      // region.
+      return false;
```
maliberty, line 947: just

return boost::geometry::covered_by(queryBox, result[0]);

src/dpl/src/Place.cpp
```diff
@@ -956,6 +904,61 @@ PixelPt Opendp::binSearch(int x, const Cell* cell, int bin_x, int bin_y) const
...
+    // both are invalid. The overlap size should be 1
+    return false;
+  }
+  if (!result.empty()) {  // cell doesn't have a region, so it
+                          // should be placed outside of all
+                          // regions. So ther overalp must be zero
+
+    return false;
+  }
+  return true;
```
maliberty, line 959: return result.empty();

maliberty (changes requested):
Just a few stylistic points but otherwise it looks good.

maliberty (approved):

PR #3486, opened 2023-06-19
fix: mismatch between .hh and .cc in argument name

openroadie:
missed changing this file when flipping the gate cloning argument


maliberty (approved):

openroadie:
@maliberty if you can please press the merge button ...

PR #3493, opened 2023-06-21
odb: Modifies abstract LEF writer to work on ostreams 

QuantamHD:
Moves file output mechanism to ostreams instead of FILE*. I want an ostream so I can run unit tests without having to rely on the filesystem. Also it's just nice to modernize this interface

### DIFF
--- a/src/OpenRoad.cc
+++ b/src/OpenRoad.cc
@@ -390,10 +390,16 @@ void OpenRoad::writeLef(const char* filename)
       if (cnt > 0) {
         name += "_" + std::to_string(cnt);
         std::ofstream os(name);
+        if (!os.is_open()) {
+          logger_->error(utl::ODB, 2001, "Cannot open LEF file {}\n", name);
+        }
         odb::lefout lef_writer(logger_, os);
         lef_writer.writeLib(lib);
       } else {
         std::ofstream os(name);
+        if (!os.is_open()) {
+          logger_->error(utl::ODB, 2002, "Cannot open LEF file {}\n", name);
+        }
         odb::lefout lef_writer(logger_, os);
         lef_writer.writeTechAndLib(lib);
       }
@@ -401,6 +407,9 @@ void OpenRoad::writeLef(const char* filename)
     }
   } else if (db_->getTech()) {
     std::ofstream os(filename);
+    if (!os.is_open()) {
+      logger_->error(utl::ODB, 2003, "Cannot open LEF file {}\n", filename);
+    }
     odb::lefout lef_writer(logger_, os);
     lef_writer.writeTech(db_->getTech());
   }
--- a/src/OpenRoad.cc
+++ b/src/OpenRoad.cc
@@ -378,7 +378,6 @@ void OpenRoad::writeLef(const char* filename)
 {
   auto libs = db_->getLibs();
   int num_libs = libs.size();
-  odb::lefout lef_writer(logger_);
   if (num_libs > 0) {
     if (num_libs > 1) {
       logger_->info(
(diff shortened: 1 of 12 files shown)


src/odb/src/lefout/lefout.cpp
```diff
@@ -897,57 +908,59 @@ void lefout::writeVia(dbTechVia* via)
...
-            lefdist(P.getXTopEnclosure()),
-            lefdist(P.getYTopEnclosure()));
+    fmt::print(
+        _out, " + LAYERS {} {} {} ", bot.c_str(), cut.c_str(), top.c_str());
+    fmt::print(_out,
+               " + CUTSPACING {:.11g} {:.11g} ",
+               lefdist(P.getXCutSpacing()),
+               lefdist(P.getYCutSpacing()));
+    fmt::print(_out,
+               " + ENCLOSURE {:.11g} {:.11g} {:.11g} {:.11g} ",
```
maliberty, line 930: While not an issue with the translation, it seems there should be new lines in these statements.

src/odb/src/swig/common/swig_common.cpp
```diff
@@ -127,22 +127,28 @@ int write_def(odb::dbBlock* block,
 
 int write_lef(odb::dbLib* lib, const char* path)
 {
+  std::ofstream os(path);
```
maliberty, line 130: Check if successful

src/odb/src/swig/common/swig_common.cpp
```diff
@@ -127,22 +127,28 @@ int write_def(odb::dbBlock* block,
...
-  odb::lefout writer(logger);
-  return writer.writeTechAndLib(lib, path);
+  odb::lefout writer(logger, os);
+  writer.writeTechAndLib(lib);
+  return true;
 }
 
 int write_tech_lef(odb::dbTech* tech, const char* path)
 {
+  std::ofstream os(path);
```
maliberty, line 139: Check if successful

src/odb/src/swig/common/swig_common.cpp
```diff
@@ -127,22 +127,28 @@ int write_def(odb::dbBlock* block,
...
   utl::Logger* logger = new utl::Logger(NULL);
-  odb::lefout writer(logger);
-  return writer.writeTech(tech, path);
+  odb::lefout writer(logger, os);
+  writer.writeTech(tech);
+  return true;
 }
 int write_macro_lef(odb::dbLib* lib, const char* path)
 {
+  std::ofstream os(path);
```
maliberty, line 147: Check if successful

maliberty (changes requested):

maliberty (approved):

src/OpenRoad.cc
```diff
@@ -390,17 +390,26 @@ void OpenRoad::writeLef(const char* filename)
       if (cnt > 0) {
         name += "_" + std::to_string(cnt);
         std::ofstream os(name);
+        if (!os.is_open()) {
```
oharboe, line 393: Exceptions are enabled when reading. The advantage is that *all* IO operations are checked, including each individual read/write operation, so if one runs out of e.g. disk space, an exception is thrown.

https://github.com/The-OpenROAD-Project/OpenROAD/blob/7586d03811ca43cc2615077afb0f4821c81f82e2/src/OpenRoad.cc#L426

The code below has is manually adding the exceptions.

oharboe (changes requested):
enable exceptions in ofstream to get error checking everywhere, as is done for read:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/7586d03811ca43cc2615077afb0f4821c81f82e2/src/OpenRoad.cc#L426

src/OpenRoad.cc
QuantamHD: @maliberty Do we correctly catch these exceptions in TCL land?

maliberty: https://github.com/The-OpenROAD-Project/OpenROAD/blob/7586d03811ca43cc2615077afb0f4821c81f82e2/src/Exception.i#L28

should catch them but you can always verify

maliberty (approved):

PR #3497, opened 2023-06-21
ppl: place fallback groups and pins after placing the remaining pins

eder-matheus:


maliberty (approved):

maliberty:
I'll squash this to get rid of the reverts when merging.

oharboe:
@eder-matheus @maliberty I tried it on my design where I had this problem and it now works :-)

PR #3505, opened 2023-06-22
Fix for openlane #1823

openroadie:
Don't touch pins connected to don't touch instances when fixing fanout.

### DIFF
--- a/src/rsz/src/RepairDesign.cc
+++ b/src/rsz/src/RepairDesign.cc
@@ -1236,8 +1236,9 @@ RepairDesign::makeRepeater(const char *reason,
       Instance *inst = network_->instance(pin);
 
       // do not disconnect/reconnect don't touch instances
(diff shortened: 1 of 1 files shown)


src/rsz/src/RepairDesign.cc
```diff
@@ -1234,6 +1234,11 @@ RepairDesign::makeRepeater(const char *reason,
     for (const Pin *pin : load_pins) {
       Port *port = network_->port(pin);
       Instance *inst = network_->instance(pin);
+
+      // do not disconnect/reconnect don't touch instances
+      if (resizer_->dontTouch(inst)) {
+        continue;
+}
```
QuantamHD, line 1241: Looks like the indenting on this is incorrect.

openroadie: I guess github's web interface does not do correct indentations by default :). Will fix. Thx

PR #3507, opened 2023-06-22
Use our own fork of lemon

### LINKED ISSUE #2791: Should we fork LEMON
### Description

The [LEMON](https://lemon.cs.elte.hu/trac/lemon) library is a required by OpenROAD, but the maintainers are not very responsive to bug fixes.

The last official release of LEMON was in 2014, but it has commits as recent as 2020. We've since run into some problem with the library not supporting C++20, and were required to introduce patches.

LEMON is pretty widely distributed by package managers so by requiring a custom version of this library we're potentially introducing a conflict. If we do so I think it would be best to hard fork LEMON and make sure to rename its install directory. 

### Suggested Solution

_No response_

### Additional Context

_No response_

vvbandeira:
This change is to avoid recent intermittent network issues.

Closes #2791

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -18,7 +18,6 @@ _installCommonDev() {
     boostChecksum="077f074743ea7b0cb49c6ed43953ae95"
     eigenVersion=3.4
     lemonVersion=1.3.1
-    lemonChecksum="e89f887559113b68657eca67cf3329b5"
     spdlogVersion=1.8.1
 
     # temp dir to download and compile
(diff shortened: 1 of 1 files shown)


maliberty (approved):

rovinski:
@vvbandeira The homebrew repo we host is going to have to be updated as well:
https://github.com/The-OpenROAD-Project/homebrew-lemon-graph

vvbandeira:
@rovinski I merged this since the changes needed for macOS are not affected by this PR. I have pushed the update `.rb` file to the repo, is there anything else I need to do to propagate these changes?

rovinski:
I believe homebrew simply looks at the latest version on the master/main branch, so if the updated version is there, it's all set.

PR #3508, opened 2023-06-22
odb: Maps Pin SigType(SCAN, RESET, TIEOFF) to SIGNAL

QuantamHD:
Fixes abstract LEF writer to emit the correct sigtype for pins.


src/odb/src/lefout/lefout.cpp
```diff
@@ -398,7 +398,19 @@ void lefout::writeBlockTerms(dbBlock* db_block)
...
+      case odb::dbSigType::TIEOFF:
+      case odb::dbSigType::RESET:
+        // LEF Pins can only be pin : [USE { SIGNAL | ANALOG | POWER | GROUND |
+        // CLOCK } ;] whereas nets can be net: [+ USE {ANALOG | CLOCK | GROUND |
+        // POWER | RESET | SCAN | SIGNAL | TIEOFF}] BTerms get their sigType
+        // from the net. So we map other sigtypes to SIGNAL in exported LEFs
+        fmt::print(_out, "    USE SIGNAL ;\n");
+        break;
+      default:
+        fmt::print(_out, "    USE {} ;\n", b_term->getSigType().getString());
```
maliberty, line 412: It is preferrable to list the other cases rather than use default.  That way the compiler will complain if a new enum is added.

A break is missing.

maliberty (approved):

PR #3062, opened 2023-03-22
fix: Add missing gTest dependency

openroadie:

PR #3514, opened 2023-06-23
Remove libgtest-dev from DependencyInstaller

vvbandeira:
This is installed from ./third-party/gtest/CMakeLists.txt

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -188,7 +188,6 @@ _installUbuntuPackages() {
         libpcre3-dev \
         python3-dev \
         libreadline-dev \
-        libgtest-dev \
         tcl-dev \
         tcllib \
         wget \
--- a/third-party/gtest/CMakeLists.txt
+++ b/third-party/gtest/CMakeLists.txt
@@ -1,4 +1,3 @@
-find_package(GTest)
 if(NOT GTest_FOUND)
   include(FetchContent)
   FetchContent_Declare(
--- a/third-party/gtest/CMakeLists.txt
+++ b/third-party/gtest/CMakeLists.txt
@@ -1,10 +1,8 @@
-if(NOT GTest_FOUND)
-  include(FetchContent)
-  FetchContent_Declare(
-    googletest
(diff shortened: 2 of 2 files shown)


etc/DependencyInstaller.sh
```diff
@@ -188,7 +188,6 @@ _installUbuntuPackages() {
         libpcre3-dev \
         python3-dev \
         libreadline-dev \
-        libgtest-dev \
```
oharboe, line 191: Was there a library that should be uninstalled too?

vvbandeira: Not sure, I looked at when the dependency was introduced and it was only this line. See 9964e664f on #3062.

oharboe: See https://github.com/The-OpenROAD-Project/OpenROAD/discussions/3513#discussioncomment-6265721

vvbandeira: This is the only reference for `gtest` that I found. @QuantamHD can you confirm that the `gtest` you suggested uninstalling was just `libgtest-dev`?

QuantamHD: Yes. I think the other thing we might consider removing is the option of using the system Gtest in the cmake file. It seems like Gtest in general prefers to be included in your build rather that being a system dep.

vvbandeira: Sounds good. I will remove and also test this in a fresh Ubuntu 20.04 VM to make sure this does not break Harsh's or Peter's systems.

third-party/gtest/CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
-find_package(GTest)
 if(NOT GTest_FOUND)
```
gadfort, line 1: Why keep the if()? Seems like it should be removed if we are not checking for the GTest package anymore.

vvbandeira: Addressed on [9324a29](https://github.com/The-OpenROAD-Project/OpenROAD/pull/3514/commits/9324a299b638ca02fa8ff443663732fe2fa1f219).

maliberty (approved):

PR #3516, opened 2023-06-23
cts: Rename graphics class to disambiguate header files

QuantamHD:
Some header files are named similarly which presents a problem in Google's version of the build. I could fix it other ways, but this would be helpful for us.


maliberty (approved):
That's an odd restriction.

QuantamHD:
It's just because we glob all the files together, and when you import Graphics.h it isn't sure which one it needs. This could be fixed with more module like compiles, but I haven't gotten around to reorganizing it yet.

PR #3538, opened 2023-06-26
dpl: Fixes double height placement already present error

QuantamHD:
Checks that the row heights of double height cells match the row height of the current cell. If they don't, it means that this pixel is partially filled by a single-height or shorter cell, which is allowed. However, if they do match, it means that we are trying to overwrite a double-height cell placement, which is an error.

### DIFF
--- a/src/dpl/src/Grid.cpp
+++ b/src/dpl/src/Grid.cpp
@@ -638,14 +638,27 @@ void Opendp::paintPixel(Cell* cell, int grid_x, int grid_y)
       for (int y = layer_y; y < layer_y_end; y++) {
         Pixel* pixel = gridPixel(layer.second.grid_index, x, y);
         if (pixel->cell) {
-          logger_->error(
-              DPL,
-              41,
-              "Cannot paint grid because another layer is already occupied.");
-        } else {
-          pixel->cell = cell;
-          pixel->util = 1.0;
+          // Checks that the row heights of the found cell match the row height
+          // of this layer. If they don't, it means that this pixel is partially
+          // filled by a single-height or shorter cell, which is allowed.
+          // However, if they do match, it means that we are trying to overwrite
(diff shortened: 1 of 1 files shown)


QuantamHD:
@mina1460 FYI.

src/dpl/src/Grid.cpp
```diff
@@ -638,14 +638,27 @@ void Opendp::paintPixel(Cell* cell, int grid_x, int grid_y)
...
-        } else {
-          pixel->cell = cell;
-          pixel->util = 1.0;
+          // Checks that the row heights of the found cell match the row height
+          // of this layer. If they don't, it means that this pixel is partially
+          // filled by a single-height or shorter cell, which is allowed.
+          // However, if they do match, it means that we are trying to overwrite
+          // a double-height cell placement, which is an error.
+
+          GridInfo grid_info_candidate = getGridInfo(pixel->cell);
```
mina1460, line 647: Your changes make perfect sense. Thank you

Just one comment, getGridInfo makes two tree traversals and loses any information about cell height (the grid index would work, but you would have to map it back to the height). Instead, if you use getRowInfo, you make only one traversal and would have a pair where the first is the height of the cell and the second is the gridInfo you get from getGridInfo.

The cell height might be useful to check if the cell is getting placed on top of another cell of bigger height. the exact equality of grid index doesn't cover this.

mina1460 (approved):
I tested it and it is fixed an issue I was facing with the region placement PR.

src/dpl/src/Grid.cpp
maliberty: @QuantamHD this sounds easy to address

maliberty: I'll merge this to unblock Mina's PR.  Please address this in a follow on.

maliberty (approved):

PR #3550, opened 2023-06-29
sync readme OR images

vijayank88:
Update OpenROAD flow stages images as per latest features.


vvbandeira (approved):

gadfort:
@maliberty @vijayank88 how come the background in these images is not black but is gray instead?  That seems to be a bug in the GUI?

maliberty:
I recall Seyi had a similar screen but I've never seen it.  @vijayank88 do you see the gray background when running interactively?

gadfort:
@maliberty that is why it caught my eye

vijayank88:
I don't recall which commit has changed, but I see gray background past one month... But if I save image through command line arguments getting black for layout area.

maliberty:
@vijayank88 what platform are you running on?

vijayank88:
GCP CentOS 7

PR #3552, opened 2023-06-29
gui: Add remaining properties to Tech and Block inspection

AcKoucher:
This PR is to be linked with the Issue #3404.

With these changes I'm adding the remaining properties in the Technology and Block inspection which were requested by @gadfort and listed in #3477.

Some important considerations:
In the dbVia, I created a property list Shapes, because a dbVia may have more than 3 geometries as the cut layer may have more than one. As the layers can be inspected through this new property, I thought there was no need to create a "Layers" property of its own.

The way the DbTechViaLayerRuleDescriptor is currently implemented, there will be inevitably via-layer rules with the same name. I couldn't find a solution to that, because there is neither a getName() method nor a way to access the ViaRule through the dbTechViaLayerRule object (The latter would make possible to create a name such as: via_rule_name + layer_name).

### DIFF
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -2015,7 +2015,7 @@ Descriptor::Properties DbViaDescriptor::getProperties(std::any object) const
 
   Properties props({{"Block", gui->makeSelected(via->getBlock())}});
 
-  if (via->getPattern() != "") {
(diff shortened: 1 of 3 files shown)


maliberty:
LGTM - @gadfort any comments?

gadfort (approved):
LGTM as well. Thanks

PR #3553, opened 2023-06-29
gui: refactor with new function drawTextInBBox()

Iyury1:
Additional refactoring after #3528

I added a function that removes most code duplication from drawInstanceNames() and drawITermLabels(). It takes parameters for a string, color, font, rect, but also a painter pointer. Perhaps this is passing around the QPainter pointer too many times?


maliberty (approved):

maliberty:
Pointers are cheap to pass so no worries about that.  Thanks for the improvement.

maliberty:
FYI - it would be good to update your master branch in your fork.  The message id error has been fixed there.

PR #3554, opened 2023-06-29
dft: misc refactors

fgaray:
Several refactors:

* changing shared_ptrs to unique_ptr in dft code. The owner should be the scan chains
* Changing the responsibility of sorting the scan cells inside the scan chains to architect instead of stitch. Stitch should only create the connectivity and not decide the order of the scan cells.
* Fix a small bug where we didn't show  the change of clock if we were changing clock edges.

### DIFF
--- a/src/dft/src/architect/ScanChain.hh
+++ b/src/dft/src/architect/ScanChain.hh
@@ -50,7 +50,13 @@ namespace dft {
 //  - Find the scan enable of the chain
 class ScanChain
 {
+ private:
+  using ScanCells = std::vector<std::unique_ptr<ScanCell>>;
+
  public:
+  // Function to sort the scan cells. (falling, rising, output vector)
+  using SortFn = const std::function<void(ScanCells&, ScanCells&, ScanCells&)>&;
+
   explicit ScanChain(const std::string& name);
   // Not copyable or movable
   ScanChain(const ScanChain&) = delete;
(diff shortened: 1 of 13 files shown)


src/dft/src/clock_domain/ClockDomain.cpp
```diff
@@ -68,4 +68,26 @@ ClockEdge ClockDomain::getClockEdge() const
...
+    case ClockEdge::Rising:
+      return "rising";
+      break;
+    case ClockEdge::Falling:
+      return "falling";
+      break;
+    default:
+      // TODO replace with std::unreachable() once we reach c++23
+      return "";
+      break;
```
maliberty, line 83: Its best not to use a default case with an enum that way the compiler will complain if any case is unused

maliberty (changes requested):

src/dft/src/clock_domain/ClockDomain.cpp
fgaray: Right. I changed this to use the tip from here: [abseil.io/tips/147](https://abseil.io/tips/147)

We need to return something anyway, otherwise the compiler with complain with "control reaches end of non-void function "

src/dft/src/architect/ScanChain.hh
```diff
@@ -50,15 +50,21 @@ namespace dft {
 //  - Find the scan enable of the chain
 class ScanChain
 {
+ private:
+  using ScanCells = std::vector<std::unique_ptr<ScanCell>>;
+
  public:
+  // Function to sort the scan cells. (falling, rising, output vector)
+  using SortFn = const std::function<void(ScanCells&, ScanCells&, ScanCells&)>&;
+
```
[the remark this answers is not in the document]
fgaray: Yeah, you are right. I think the type is not so bad so I will just remove the alias. Thanks!

maliberty (approved):

PR #3555, opened 2023-06-29
ppl: add initial version of pin placement with simulated annealing

eder-matheus:
This first version only assigns regular pins. Therefore, features like pin groups, mirrored pins and pin constraints are not supported yet.

Use the `-annealing` flag for testing. Not adding this flag to README yet as it is not ready to all users.


src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -0,0 +1,206 @@
...
+  float temperature = init_temperature_;
+
+  for (int iter = 0; iter < max_iterations_; iter++) {
+    for (int perturb = 0; perturb < perturb_per_iter_; perturb++) {
+      int prev_slot = -1;
+      int new_slot = -1;
+      int pin1 = -1;
+      int pin2 = -1;
+      perturbAssignment(prev_slot, new_slot, pin1, pin2);
+      int cost = getAssignmentCost();
```
maliberty, line 80: Recomputing the assignment cost visits every net.  When you perturb you only touch one or two nets.   It would be faster to compute a cost delta for the move.

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -0,0 +1,206 @@
...
+      int delta_cost = cost - pre_cost;
```
maliberty, line 81: const

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -0,0 +1,206 @@
...
+  pin_assignment_[pin] = new_slot;
+}
+
+void SimulatedAnnealing::restorePreviousAssignment(const int prev_slot,
+                                                   const int new_slot,
+                                                   const int pin1,
+                                                   const int pin2)
+{
+  if (prev_slot != -1 && new_slot != -1) {
+    // restore single pin to previous slot
```
maliberty, line 193: This is requires too much knowledge of how the perturbation works.  Once you extend to groups you'll need a move the knows how to undo itself.  Just noting - it can wait until your next step.

maliberty (changes requested):

src/ppl/src/SimulatedAnnealing.cpp
eder-matheus: My idea is to have a list of pins instead of just two. I'll start this part tomorrow and will keep you updated on my progress.

maliberty (approved):

PR #3561, opened 2023-07-01
pad: allow manual setting of bump routing to pad iterms

gadfort:
Changes:
- added `-terminal` to `assign_io_bump` to control which iterm the rdl router will attempt to route to + associated testcases.


src/pad/src/pad.tcl
```diff
@@ -95,17 +95,35 @@ proc remove_io_bump {args} {
...
   sta::check_argc_eq1 "assign_io_bump" $args
 
+  set terminal NULL
+  if { [info exists keys(-terminal)] } {
+    set terms [split $keys(-terminal) {/}]
+    set iterm [lindex $terms end]
+    set inst [join [lrange $terms 0 end-1] {/}]
+
+    set inst [pad::find_instance $inst]
+    set terminal [$inst findITerm $iterm]
```
maliberty, line 115: why not dbBlock::findITerm ?

gadfort: I didn't realize that was there. That is a much better solution. Will make that change.

maliberty (approved):

PR #3564, opened 2023-07-03
ppl: small fix on annealing

eder-matheus:


maliberty (approved):

oharboe:
@eder-matheus @maliberty How many pins does this deal with before fallback?

PR #3565, opened 2023-07-04
Recover power

openroadie:
repair_timing -recover_power 

- Not sure whether a separate new command should be added.
- For now "-recover_power" will block setup/hold fixing (mutually exclusive).


openroadie:
@QuantamHD @tspyrou @maliberty  See the ORFS results here with an additional call to repair timing (for power recovery):

I believe we can merge this (CI/CD is clean). Currently the default is to fix the top 20% (in terms of slack available) paths. This seems a safe enough limit based on ORFS results for now. I will add an option to control this and also look at why certain tests show a wide variation in timing beyond a certain number of changes.

Other work in this area involves running this after global routing. I tried to do this but the incremental global router crashed. Merging this PR would also enable the global router team can take a look at what is going on.

maliberty:
The CI runs with the head of master.  If this PR is merged then the test cases in your test PR will fail.  I'm ok to update the metrics, but a PR should be prepared that does so and can be merged shortly after this one.

openroadie:
@maliberty I am sorry if I was not clear on my intent. The ORFS branch I ran runs "repair_timing -recover_power" right after repair_timing and shows that the few failures are reasonable. I was not intending to make this call a default one in the ORFS scripts. So merging this PR is not going to impact anything unless we want to make power recovery a default option post CTS. I personally believe that we should not make this the default behavior and work on the following issues post merge:

- Incremental route crashes if I try to use GR parasitics for powe recovery.
- I don't get good results if I optimize too many paths for power (this is semi expected if we do large perturbations). We need to see if we can do a better job with power recovery once global route parasitics are used (it could also be an issue in power recovery code... I don't know yet).

maliberty (approved):
Sorry I missed that this is off by default.

Issue #3484, opened 2023-06-19
Detailed routing increases dramatically in running times with area, surprisingly


oharboe:
Create `settings.mk` in flow and use updated mock-array-big https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1097:

```
export MOCK_ARRAY_PITCH_SCALE=20
export MOCK_ARRAY_DATAWIDTH=64
export DESIGN_CONFIG=designs/asap7/mock-array-big/config.mk
```

Regenerate Verilog for mock-array-big to get a 64 bit datapath:

```
make verilog
```

Now run:

```
rm -rf results/ objects/ logs/ && make
```

Output:

```
[deleted]
(/usr/bin/time -f 'Elapsed time: %E[h:]min:sec. CPU time: user %U sys %S (%P). Peak memory: %MKB.' /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -exit -no_init  ./scripts/detail_route.tcl -metrics ./logs/asap7/mock-array-big_Element/base/5_2_TritonRoute.json) 2>&1 | tee ./logs/asap7/mock-array-big_Element/base/5_2_TritonRoute.log
[deleted]
[INFO DRT-0199]   Number of violations = 76.
Viol/Layer          M3     M4     M5
EOL                  0      3      5
Metal Spacing        1     22     22
Recheck              0      1      0
Short                0     13      0
eolKeepOut           0      9      0
[INFO DRT-0267] cpu time = 00:00:15, elapsed time = 00:00:07, memory = 5626.13 (MB), peak = 6145.72 (MB)
[deleted]
```

- after 10 iterations the Element still as 76 DRC violations

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/23af58e5-bf67-488b-b4ad-f7577bdd5dd3)

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/faec11e3-4021-480b-a6b3-f71d500708bf)

Detailed routing converges to 0 violations on the 3rd iteration, which is quick. However, detailed routing takes 3h30m on my machine, which I believe is mainly a function of the *area* not the complexity of routing.

```
[INFO DRT-0195] Start 3rd optimization iteration.
    Completing 10% with 186 violations.
    elapsed time = 00:00:02, memory = 25160.54 (MB).
    Completing 20% with 157 violations.
[...]

PR #3568, opened 2023-07-04
Reduce data transferred in distributed detailed routing

osamahammad21:
In this PR, I am reducing the data sent to servers and received by leaders to the minimal needed form. I studied the data needed by both ends. The server only needs the initial data of worker including the routebox, extBox, costs and markers. The leader only needs the nets and the bestMarkers. The nets don't need to include the initial segments, it only needs the final segments to be committed to the design.
The main part that has beed reduced here is the gridGraph. The server doesn't need the gridGraph since it initializes it. and the leader doesn't need the gridGraph back. This minimization showed noticeable speedup and memory recduction in distributed detailed routing.

### DIFF
--- a/src/drt/test/gcd_nangate45_distributed.tcl
+++ b/src/drt/test/gcd_nangate45_distributed.tcl
@@ -5,7 +5,7 @@ set server1 [exec $OR server1.tcl > results/server1.log &]
 set server2 [exec $OR server2.tcl > results/server2.log &]
 set balancer [exec $OR balancer.tcl > results/balancer.log &]
 set base [exec $OR -exit gcd_nangate45.tcl > results/base.log &]
-
+exec sleep 3
 read_lef Nangate45/Nangate45_tech.lef
 read_lef Nangate45/Nangate45_stdcell.lef
 read_def gcd_nangate45_preroute.def
--- a/src/drt/src/db/drObj/drNet.cpp
+++ b/src/drt/src/db/drObj/drNet.cpp
@@ -49,9 +49,6 @@ template <class Archive>
 void drNet::serialize(Archive& ar, const unsigned int version)
 {
   (ar) & boost::serialization::base_object<drBlockObject>(*this);
-  // (ar) & pins_;
-  // (ar) & extConnFigs_;
-  // (ar) & routeConnFigs_;
   (ar) & bestRouteConnFigs_;
   (ar) & modified_;
   (ar) & numMarkers_;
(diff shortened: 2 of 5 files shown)


src/drt/test/gcd_nangate45_distributed.tcl
```diff
@@ -5,7 +5,7 @@ set server1 [exec $OR server1.tcl > results/server1.log &]
 set server2 [exec $OR server2.tcl > results/server2.log &]
 set balancer [exec $OR balancer.tcl > results/balancer.log &]
 set base [exec $OR -exit gcd_nangate45.tcl > results/base.log &]
-
+exec sleep 3
```
maliberty, line 8: Is this intended?

oharboe:
@osamahammad21 What is the observable effect of this change on detailed routing performance?

Ref: https://github.com/The-OpenROAD-Project/OpenROAD/issues/3484

src/drt/test/gcd_nangate45_distributed.tcl
osamahammad21: yes. to slow down the leader and let the servers be spawned before sending them any work.

osamahammad21:
@oharboe Previously when I tried to run the in-question test case in distributed mode, it ran out of memory. That's because in distributed detailed routing, unlike normal detailed routing, the number of workers to be processed at one batch is not restricted. That led to having 130K+ workers running at the same time on the cloud. The issue appears when the servers return the processed workers back to the leader. Without these changes the memory blows up.

At first as a workaround, I limited the number of workers per batch in the distributed mode. But that led the design to take even more time than the original runtime on one machine (2h+ in the first iteration). That's due to serializing and transferring gigabytes of unneeded data over the cloud.

After this change, even without applying the changes from [PR 3567](https://github.com/The-OpenROAD-Project/OpenROAD/pull/3567), it ended in 7 minutes which is **24x** faster than the original runtime on one machine.

Single Machine (96 CPU):
`[INFO DRT-0267] cpu time = 87:08:00, elapsed time = 02:48:43, memory = 10662.26 (MB), peak = 28213.59 (MB)
`Distributed (10 servers of 96 CPU machines):
`[INFO DRT-0267] cpu time = 00:07:06, elapsed time = 00:07:09, memory = 10402.31 (MB), peak = 15573.68 (MB)
`

oharboe:
@osamahammad21 So this improves distributed mode only, or should it have an effect on local detailed routing too?

osamahammad21:
@oharboe This should only affect distributed mode only.

maliberty (approved):

src/drt/test/gcd_nangate45_distributed.tcl
maliberty: We generally need a better worker startup process.  This can hold for more productization of distributed drt.

PR #3575, opened 2023-07-04
gui: pin marker texts must fit in the layout viewer

### LINKED ISSUE #3333: It is not possible to scroll to the leftmost letter of long port names
### Description

There's text to the left that I can't see for ports with long names:

image

Same for top, right and bottom.

### Suggested Solution

Allow scrolling far enough to read the entire portname.

### Additional Context

_No response_

AcKoucher:
Resolves #3333 for now.

The problem that @oharboe described happens in a specific situation in which the amount of zoom is not big enough so that the port name fits in the distance between the boundary of the die are and the limit of the widget, and not small enough so that the port names are not displayed in the layout viewer. 

Apparently it is not possible to follow @oharboe 's suggestion to make the scroll area bigger, because doing that will just allow leaving the margins of the widget, in other words, the text is being painted outside of the widget.

As @maliberty said he is working on the zoom mechanism and these things seem related, I chose to, for now, paint the pin marker text only if it fits between the die area and the widget’s limit.


AcKoucher:
The problem that was happening in the build was due to the usage of the QFontMetrics method `horizontalAdvance()`.

A had to use the obsolete method `width()` instead for the OR test build in CentOS to succeed.

Reference: [Obsolete Members for QFontMetrics](https://doc.qt.io/qt-5/qfontmetrics-obsolete.html#width)

maliberty:
So if a label doesn't fit then it just won't be drawn at all?

Btw I am done with my gui work for now

AcKoucher:
Yes, that's the idea, as this situation only happens in a specific zoom range. Do you have a suggestion of another approach?Should it behave differently?

maliberty:
I think the idea is to see the labels in full rather than hide them

AcKoucher:
I'm closing this and opening another PR, because I had to update the branch and I messed up with my changes.

PR #3576, opened 2023-07-04
ppl: initial support for pin groups in simulated annealing

eder-matheus:


maliberty (approved):
I'll review in more detail later when I'm back. I'll approve for now

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -163,66 +209,151 @@ int SimulatedAnnealing::getPinCost(int pin_idx)
...
   }
 }
 
-int SimulatedAnnealing::swapPins(int& pin1, int& pin2)
+int SimulatedAnnealing::swapPins(std::vector<int>& pins)
 {
-  pin1 = (int) (std::floor(distribution_(generator_) * num_pins_));
-  pin2 = (int) (std::floor(distribution_(generator_) * num_pins_));
+  int pin1 = (int) (std::floor(distribution_(generator_) * num_pins_));
+  int pin2 = (int) (std::floor(distribution_(generator_) * num_pins_));
```
maliberty, line 250: Perhaps use uniform_int_distribution ? (many places)

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -163,66 +209,151 @@ int SimulatedAnnealing::getPinCost(int pin_idx)
...
     pin2 = (int) (std::floor(distribution_(generator_) * num_pins_));
   }
 
+  while (netlist_->getIoPin(pin1).isInGroup()) {
+    pin1 = (int) (std::floor(distribution_(generator_) * num_pins_));
+  }
+
+  while (netlist_->getIoPin(pin2).isInGroup()) {
+    pin2 = (int) (std::floor(distribution_(generator_) * num_pins_));
+  }
```
maliberty, line 261: Are you guaranteed to have to non-group pins?  

This could be slow if there are mostly pins in groups.  It might be better to separate them and then choose.  (Similarly in other places)

Is it not possible to swap a group with a block of whatever signals are in a set of slots?  If not it seems it will be hard to move groups as they are easily blocked by single bit pins.

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -85,12 +90,16 @@ void SimulatedAnnealing::run()
       if (delta_cost <= 0 || accept_prob > rand_float) {
         // accept new solution, update cost and slots
         pre_cost = cost;
-        if (prev_slot != -1 && new_slot != -1) {
-          slots_[prev_slot].used = false;
-          slots_[new_slot].used = true;
+        if (!prev_slots.empty() && !new_slots.empty()) {
```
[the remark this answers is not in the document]
eder-matheus: Yes, because all prev_slots are marked as `used` during the iterations, and the new_slots are only free slots.

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -127,14 +136,50 @@ void SimulatedAnnealing::randomAssignment()
...
+      slot_idx++;
+      slot = slot_indices[slot_idx];
+    }
+    pin_assignment_[i] = slot;
+    slots_[slot].used = true;
+    slot_idx++;
   }
 }
 
+int SimulatedAnnealing::randomAssignmentForGroups(
```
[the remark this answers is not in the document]
eder-matheus: It is. The `slot_indices` vector is shuffled before being passed to the `randomAssignmentForGroupts` function. I've made a small change to avoid overlaps between groups. This and the remaining fixes will be added in my following PR.

src/ppl/src/SimulatedAnnealing.cpp
eder-matheus: I agree with both suggestions. The second suggestion is more complex to implement because I need an efficient way to know what pin is placed in a specific slot. I can try to implement it in a future PR.

src/ppl/src/SimulatedAnnealing.cpp
eder-matheus: I've tried using uniform_int_distribution, but it's generating different results across platforms. My unit tests fail in the ci while they are passing in my Mac.

PR #3581, opened 2023-07-05
ppl: add checks to ensure properly placement of pin groups

eder-matheus:


eder-matheus:
@maliberty My latest commit closes the more significant part of the pin grouping approach inside the SA. This is ready for review.

After getting this merged, I'll go to the io constraints.

src/ppl/test/annealing2.ok
```diff
@@ -10,5 +10,5 @@
 Found 0 macro blocks.
 Using 1u default distance from corners.
 Using 2 tracks default min distance between IO pins.
-[INFO PPL-0012] I/O nets HPWL: 1962.48 um.
+[INFO PPL-0012] I/O nets HPWL: 1962.64 um.
```
maliberty, line 13: Are you expecting a change in this test which doesn't use groups?

maliberty:
I think we are going to need more way to move/swap groups.  This set of choices seems to make a hard separation between lone and group pins but we may need to be able to operate across that barrier.

src/ppl/test/annealing2.ok
eder-matheus: This test has two pin groups.

eder-matheus:
> I think we are going to need more way to move/swap groups. This set of choices seems to make a hard separation between lone and group pins but we may need to be able to operate across that barrier.

As mentioned in the other PR, I agree with that. After having the PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/3603 and this PR merged, I'll work on allowing the swap between pin groups and lone pins.

maliberty (approved):

PR #3594, opened 2023-07-08
pad/pdn: improve bump handling

gadfort:
Adds:
- dont_route to `assign_io_bump` command to allow for assignment of bumps without attempting to route the bump
- `-bump` flag to pdn macro grid definition to skip adding grids to bumps that have already been routed to.


maliberty:
What is the use case? Pre routing?

src/pad/src/RDLRouter.cpp
```diff
@@ -1152,6 +1152,8 @@ std::vector<RDLRouter::TargetPair> RDLRouter::generateRoutingPairs(
   odb::dbTechLayer* bump_pin_layer = get_other_layer(bump_accessvia_);
   odb::dbTechLayer* pad_pin_layer = get_other_layer(pad_accessvia_);
 
+  const double dbus = block_->getDbUnitsPerMicron();
```
maliberty, line 1155: This looks to be defined far from it's use

gadfort:
> What is the use case? Pre routing?

The bumps in the center of chip cannot be routed with the RDL router. This still allows them to be correctly assigned (ie. a bpin is created and the bump pins are connected to the correct net), but the RDL router will not attempt to route them. PDNgen can then add a grid around them to connect to the bump properly.

src/pad/src/RDLRouter.cpp
gadfort: I can move it down closer.

maliberty (approved):

PR #3601, opened 2023-07-10
pick the centroid of targets and driver pin to insert new cloned cell

openroadie:
Use the centroid of all receivers and driver pin to pick a location for the cloned gate.

### DIFF
--- a/src/rsz/src/RepairSetup.cc
+++ b/src/rsz/src/RepairSetup.cc
@@ -406,8 +406,8 @@ RepairSetup::repairSetup(PathRef &path,
       // Gate cloning
       if (enable_gate_cloning && fanout > split_load_min_fanout_ &&
           !tristate_drvr &&
-          !resizer_->dontTouch(net)) {
-          cloneDriver(drvr_path, drvr_index, path_slack, &expanded);
+          !resizer_->dontTouch(net) &&
+          cloneDriver(drvr_path, drvr_index, path_slack, &expanded)) {
           changed = true;
           break;
       }
@@ -659,7 +659,7 @@ Point RepairSetup::computeCloneGateLocation(const Pin *drvr_pin,
   return (Point(centroid_x/count, centroid_y/count));
 }
 
-void
+bool
 RepairSetup::cloneDriver(PathRef* drvr_path, int drvr_index,
                          Slack drvr_slack, PathExpanded *expanded)
 {
@@ -701,7 +701,7 @@ RepairSetup::cloneDriver(PathRef* drvr_path, int drvr_index,
   Instance *drvr_inst = db_network_->instance(drvr_pin);
 
   if (!resizer_->isSingleOutputCombinational(drvr_inst)) {
-    return;
+    return false;
   }
 
   Net *net = network_->net(drvr_pin);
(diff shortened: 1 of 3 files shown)


src/rsz/test/repair_fanout7.ok
```diff
@@ -11,8 +11,8 @@ load DFF_X1 CK D 300000 metal3 1000
 [INFO ODB-0132]     Created 2 special nets and 306 connections.
 [INFO ODB-0133]     Created 4 nets and 307 connections.
 worst slack -23.63
-[INFO RSZ-0040] Inserted 34 buffers.
-[INFO RSZ-0041] Resized 139 instances.
-[INFO RSZ-0049] Cloned 6 instances.
+[INFO RSZ-0040] Inserted 81 buffers.
+[INFO RSZ-0041] Resized 326 instances.
```
maliberty, line 15: Do you understand the jump in these metrics?

src/rsz/src/RepairSetup.cc
```diff
@@ -641,6 +638,27 @@ RepairSetup::upsizeCell(LibertyPort *in_port,
...
+Point RepairSetup::computeCloneGateLocation(const Pin *drvr_pin,
+                              const vector<pair<Vertex*, Slack>> &fanout_slacks)
+{
+  int count(1); // driver_pin counts as one
+  int centroid_x(0), centroid_y(0); // (0, 0)
+
+  centroid_x += db_network_->location(drvr_pin).getX();
+  centroid_y += db_network_->location(drvr_pin).getY();
+
+  int split_index = fanout_slacks.size() / 2;
```
maliberty, line 650: Is this fanout splitting arbitrary?

src/rsz/test/repair_fanout7.ok
openroadie: Different optimization path ... because the cloned gates are now placed differently. And as you can see... the timing has improved (for whatever that is worth since the design is not yet detailed placed).

src/rsz/src/RepairSetup.cc
maliberty: I don't see a response but this is marked resolved.

openroadie: github somehow lost my comment. The fanout list is sorted.

openroadie:
@vvbandeira Can you please take a look at the CI/CD failures. They have nothing to do with my code. I can also make a tiny change to retrigger the pipeline.

src/rsz/src/RepairSetup.cc
maliberty: It is sorted by slack not spatial proximity afaik.  It seems odd to take the centroid of such.  I'll approve but I suspect there is further work to do here.

maliberty (approved):

maliberty:
Please address the build errors

maliberty:
```
/tmp/workspace/OpenROAD-Public_PR-3601-head/src/rsz/src/RepairSetup.cc:707:8: error: unused variable ‘net’ [-Werror=unused-variable]

   Net *net = network_->net(drvr_pin);
```
seems related to this change.

openroadie:
@maliberty Just noticed that as well. Thx.

PR #3603, opened 2023-07-10
ppl: fixes for annealing

eder-matheus:
Updates and fixes following the comments on PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/3576.

### DIFF
--- a/src/ppl/src/SimulatedAnnealing.cpp
+++ b/src/ppl/src/SimulatedAnnealing.cpp
@@ -319,7 +319,8 @@ int SimulatedAnnealing::movePinToFreeSlot(std::vector<int>& prev_slots,
 
   bool free_slot = false;
   int new_slot;
-  distribution = boost::random::uniform_int_distribution<int>(0, num_slots_ - 1);
+  distribution
+      = boost::random::uniform_int_distribution<int>(0, num_slots_ - 1);
   while (!free_slot) {
     new_slot = distribution(generator_);
     free_slot = slots_[new_slot].isAvailable() && new_slot != prev_slot;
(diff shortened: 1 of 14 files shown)


src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -185,9 +187,11 @@ int SimulatedAnnealing::randomAssignmentForGroups(
       pin_assignment_[pin_idx] = group_slot;
       slots_[group_slot].used = true;
       group_slot++;
-      slot_idx++;
       placed_pins.insert(pin_idx);
     }
+    while (slots_[slot_idx].used) {
+      slot_idx++;
+    }
```
maliberty, line 194: This guarantees that you start on an empty slot but I'm not clear it prevents overlap (you might start the next group too close to the next group to fit).

maliberty (approved):

maliberty:
conflicts to resolve

eder-matheus:
> conflicts to resolve

Working on it. Also, I still see the different results in Mac vs Linux when using the `uniform_int_distribution`. Using the `uniform_real_distribution` is awkward in some cases, but at least we have consistent results. Should I keep the `uniform_real_distribution`, or accept the diff results when using `uniform_int_distribution`?

maliberty:
> > conflicts to resolve
> 
> Working on it. Also, I still see the different results in Mac vs Linux when using the `uniform_int_distribution`. Using the `uniform_real_distribution` is awkward in some cases, but at least we have consistent results. Should I keep the `uniform_real_distribution`, or accept the diff results when using `uniform_int_distribution`?

Please look into why you get differences.  There should be no platform dependence with uniform_int_distribution afaik.  It just relies on the underlying generator for the randomness.

maliberty (approved):

Issue #3588, opened 2023-07-07
OpenROAD slowdown when timing driven placement/CTS/routing is used. 


openroadie:
### Describe the bug

>>>>>>>> I've noticed a significant slowdown in OpenROAD over the last two weeks.
>>>>>>>> I'm trying to narrow down what might be to blame.
>>>>>>>> But for reference two weeks ago our ZeroSOC example took:
>>>>>>>> v2.0-8780 - a7f84d55d849dd56b4b06b8e3ef36fb68ee500f5
>>>>>>>> Place: 02:46.110
>>>>>>>> CTS: 02:59.520
>>>>>>>> Route: 22:46.849
>>>>>>>>
>>>>>>>> Now:
>>>>>>>> v2.0-8997 - ffa127cb22da12c0fd1fe787e2119651cad06806
>>>>>>>> Place: 43:50.960 (a 21x slowdown)
>>>>>>>> CTS: 23:57.950 (a 11x slowdown)
>>>>>>>> Route: 17:53.140 (5 mins faster)

### Expected Behavior

The runtime should remain more or less the same as before. 

### Environment

```shell
N/A for now.
```

### To Reproduce

N/A for now

### Relevant log output

```shell
N/A for now
```

### Screenshots

N/A for now

cc: @maliberty @gadfort @louiic @tspyrou @vvbandeira

openroadie:
![sc1](https://github.com/The-OpenROAD-Project/OpenROAD/assets/114895267/f4b975e1-04d6-4545-851f-42d1531e5dd5)

Here is a runtime comparison between the following SHAs:

a7f84d55d849dd56b4b06b8e3ef36fb68ee500f5
ffa127cb22da12c0fd1fe787e2119651cad06806

As you can see, the differences in runtime for various functions are comparable barring some noise.

QuantamHD:
Can you get a linux perf profile for both runs, and then run them in [perf diff ](https://manpages.ubuntu.com/manpages/jammy/man1/perf-diff.1.html)?

The fact that both CTS and GPL have slowed down makes me think it's an STA related issue.

openroadie:
@QuantamHD Please see screenshot above.

openroadie:
@gadfort Try as I might, I just do not see any slowdown. I have tried the ZeroAsic design as well as ORFS asap7/aes and asap7/riscv32. I double checked that we are running timing driven placement for my runs and I am comparing the latest master against  3ec606c02

[...]

PR #3606, opened 2023-07-11
rsz: dont estimate parasitics on special nets

gadfort:
Closes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3588

Avoid special nets in wire estimation.
@openroadie FYI

### DIFF
--- a/src/rsz/src/EstimateWireParasitics.cc
+++ b/src/rsz/src/EstimateWireParasitics.cc
@@ -356,7 +356,8 @@ Resizer::estimateWireParasitic(const Pin *drvr_pin,
 {
   if (!network_->isPower(net)
       && !network_->isGround(net)
-      && !sta_->isIdealClock(drvr_pin)) {
+      && !sta_->isIdealClock(drvr_pin)
+      && !db_network_->staToDb(net)->isSpecial()) {
     if (isPadNet(net))
       // When an input port drives a pad instance with huge input
       // cap the elmore delay is gigantic. Annotate with zero


maliberty (approved):

maliberty:
Do you see special nets that are not power or ground?

gadfort:
@maliberty the ones that form the io padring rings are special and the rdl routing.

maliberty:
> @maliberty the ones that form the io padring rings are special and the rdl routing.

Do we need running estimates for those signals?

gadfort:
@maliberty I would say yes for some of them. The rdl routed ones for sure, but then they don't get a good estimate anyways since they are not routed and the estimation assumes a wrong metal layer for them. The ones that are in the io ring don't have any parasitics to extract because they are formed from abutment connections, so there is no routing there either.
I think the parasitic estimation has some issues that need to be addressed to handle these cases correctly and handle the routing as well, however, that is a huge task. So if you are okay with merging this, it at least gets the runtimes back to normal for us.

rovinski:
I think it's an interesting question because the vast majority of the time, the timing between the pad cell and pad/bump isn't meaningful. The package and board parasitics will dominate the timing. So unless you're trying to balance skew in very high-speed off-chip interconnects, that timing shouldn't be important during _optimization_. Reporting the timing after routing is still useful.

It's interesting that the parasitics estimation is causing a large slowdown. Is it because of the estimation, or because the high latency of the net makes resizer work really hard on the I/O nets?

In general, this problem also occurs in commercial tools. Pad ring nets and pad cell to pad nets often have to be specified with `set_disable_timing` or extra runtime gets consumed. This could be an opportunity for OpenROAD to be better, however, by automatically determining this scenario and adjusting accordingly.

PR #3610, opened 2023-07-12
gui: scale pin markers' text based on available space

### LINKED ISSUE #3333: It is not possible to scroll to the leftmost letter of long port names

AcKoucher:
Resolves #3333.


src/gui/src/layoutViewer.h
```diff
@@ -384,6 +384,10 @@ class LayoutViewer : public QWidget
   // dbu)
   qreal pixels_per_dbu_;
 
+  // holds a value in dbu added to the block bounds to ensure pin markers'
+  // text don't leave the scroll area
+  const int marker_text_offset_ = 15000;
```
gadfort, line 389: not a fan of magic numbers. how does this work if the dbuPerMicron is different?

gadfort (changes requested):
It might be better to cache the size in DBUs for the largest/longest pin name and update it on zooming so it's not a hardcoded constant.

gadfort (changes requested):
It might be better to cache the size in DBUs for the largest/longest pin name and update it on zooming so it's not a hardcoded constant.

AcKoucher:
As the name is just a string, is there a way to convert the width of the string directly to dbu? I just know how to get the size in pixels with Qt using `QFontMetrics::horizontalAdvance(QString qstr)`.

gadfort:
If you have the number of pixels, you can can convert that to DBUs with `getPixelsPerDBU()`
```
QFontMetrics::horizontalAdvance(QString qstr) / getPixelsPerDBU()
```

AcKoucher:
I am trying to follow the instruction to cache the size of the longest text in DBU, but I am having some difficulties understanding how to do that.
Do the pin texts have a fixed size in DBU like the other objects from db? It looks like the texts have a fixed size in pixels (that I can get using the Qt method of my previous comment), so:
If the texts don't have a fixed size in DBU and the only way to know it is by converting as you showed, it seems complicated, because `getBounds()` is used to compute the new `pixels_per_dbu_` on which the text size in dbu depends.

Iyury1:
I think the pin texts are being drawn in RenderThread::drawPinMarkers(). The text comes from odb::dbBterm::getName() on line 1441 which returns a std::string, so I suppose it would need to be converted to a QString first and then you could use the conversion @gadfort mentioned.

AcKoucher:
Maybe I'm having difficulties expressing myself:
The problem is: to add the correct offset to the bounds, we need to know the size of the text in DBU, but to know the size of the text in DBU we need the value in pixels_per_dbu_, which is calculated using the bounds. One thing depends on the other.

@gadfort I'm sorry if this is getting too confusing.

@Iyury1 I appreciate the comment, but I think I know how to make the conversion:

```
void LayoutViewer::generateLongestPinName()
{
  QFont marker_font = options_->pinMarkersFont();
  QFontMetrics font_metrics(marker_font);

  std::vector<int> text_sizes;

  for (auto term : block_->getBTerms()) {
    QString pin_text = QString::fromStdString(term->getName());
    text_sizes.push_back(font_metrics.width(pin_text));
  }

  auto longest_text = std::max_element(text_sizes.begin(),
                                       text_sizes.end());

  pin_longest_text_ = *longest_text/pixels_per_dbu_;
}
```

gadfort:
@AcKoucher It looks like we allocate 2% of the boundary to draw the markers: https://github.com/The-OpenROAD-Project/OpenROAD/blob/dcfd631a4a67c225e5a13372a33b27ae0d024c25/src/gui/src/renderThread.cpp#L1291

Maybe the solution isn't to add an additional offset, but to correct this code to ensure the text will get scaled to ensure it fits properly in the space allotted?

AcKoucher:
I agree with you, I think this is solution. I'll take a look at it. Thanks!

src/gui/src/renderThread.cpp
```diff
@@ -1306,15 +1306,37 @@ void RenderThread::drawPinMarkers(Painter& painter,
...
   const int text_margin = 2.0 / viewer_->pixels_per_dbu_;
 
+  const int viewer_width = viewer_->geometry().width();
+  const int viewer_height = viewer_->geometry().height();
+
+  const int text_gap
+      = (std::min(viewer_width, viewer_height)
+         - std::max(die_width, die_height) * viewer_->pixels_per_dbu_)
+        / 2;
+  std::vector<int> text_lenghts;
```
maliberty, line 1319: typo: text_lenghts -> text_lengths

maliberty (changes requested):
@gadfort any further concerns?

src/gui/src/renderThread.cpp
```diff
@@ -1306,15 +1306,76 @@ void RenderThread::drawPinMarkers(Painter& painter,
...
+  const QFontMetrics font_metrics(marker_font);
+  std::vector<int> text_widths;
+
+  for (auto pin : block->getBTerms()) {
+    QString pin_text = QString::fromStdString(pin->getName());
+    text_widths.push_back(font_metrics.width(pin_text));
+  }
+
+  auto longest_text_size
+      = std::max_element(text_widths.begin(), text_widths.end());  // pixels
```
maliberty, line 1345: Building a vector to just take the max is quite wasteful.  Just compute the max as you go and skip all the memory allocations.  Same below.

maliberty:
The gui feels noticeable slower when I use this branch.  I'm not sure if it the vector noted above or the sizing loop or both.

AcKoucher:
I removed the unneeded allocations and, instead of iterating inside the size loop, I kept the largest text in a string to compute the new size based on the font. It improved a lot, but I think I can still feel a little slower than master, in one or two specific ranges of zoom particularly.

maliberty:
Something is broken - if I zoom in:
image

maliberty (approved):

PR #3621, opened 2023-07-14
psm: use layer names and add metrics recording

gadfort:
Changes:
- uses layer names in logger messages for PSM
- add metrics for IR drop

Open to suggestions to the naming of the metrics


src/psm/src/pdnsim.cpp
```diff
@@ -196,19 +196,28 @@ void PDNSim::analyze_power_grid()
...
+  const double worst_drop = std::abs(irsolve_h->getSupplyVoltageSrc()
+                                     - irsolve_h->getWorstCaseVoltage());
+  logger_->report("Average IR drop  : {:3.2e} V", avg_drop);
+  logger_->report("Worstcase IR drop: {:3.2e} V", worst_drop);
   logger_->report("######################################");
+
+  logger_->metric("powergrid__worst__voltage",
+                  irsolve_h->getWorstCaseVoltage());
+  logger_->metric("powergrid__average__drop", avg_drop);
+  logger_->metric("powergrid__worst__drop", worst_drop);
```
maliberty, line 210: Based on https://github.com/ieee-ceda-datc/datc-rdf-Metrics4ML#metrics-naming-convention I think the modifier is supposed to be last.  There is no powergrid category so maybe:

design_powergrid__voltage__worst
design_powergrid__drop__average
design_powergrid__drop__max

@ravi-varadarajan any comments?

maliberty (approved):

src/psm/src/pdnsim.cpp
gadfort: in looking at the spec, it looks like it needs to have the net and corner appended at the end so I added them.

maliberty (approved):

PR #3622, opened 2023-07-15
rsz: print progress

gadfort:
Adds:
- adds number of violating endpoints to setup repair (to mirror the behavior of the repair hold).
- adds a `-verbose` to `repair_timing` to print the status of the repair.

Motivation:
repair_timing can run for a very long time on some designs and it would be helpful to be able to see what progress if any is actually being made.

Current format:
```
Iteration | Resized | Buffers | Cloned Gates | Pin Swaps |   WNS   |   TNS   | Endpoint
```
printed every 10 repairs (could increase of decrease that). It doesn't seem to impact the runtime (atleast from what I could tell).
If there are suggestions for alternative headings or phrasing I think that would be helpful, since I just grabbed the first set that came to mind.

@maliberty @openroadie I noticed that the pin swap goes negative (this seems counter intuitive to me, I assume that is not correct).

### DIFF
--- a/src/rsz/test/repair_setup4_verbose.ok
+++ b/src/rsz/test/repair_setup4_verbose.ok
@@ -1,0 +1,23 @@
+[INFO ODB-0222] Reading LEF file: Nangate45/Nangate45.lef
+[INFO ODB-0223]     Created 22 technology layers
+[INFO ODB-0224]     Created 27 technology vias
+[INFO ODB-0225]     Created 135 library cells
+[INFO ODB-0226] Finished LEF file:  Nangate45/Nangate45.lef
+[INFO ODB-0128] Design: reg1
+[INFO ODB-0130]     Created 1 pins.
+[INFO ODB-0131]     Created 17 components and 92 component-terminals.
+[INFO ODB-0132]     Created 2 special nets and 34 connections.
+[INFO ODB-0133]     Created 7 nets and 30 connections.
+worst slack -1.95
+[INFO RSZ-0094] Found 6 endpoints with setup violations.
+Iteration | Resized | Buffers | Cloned Gates | Pin Swaps |   WNS   |   TNS   | Endpoint
+---------------------------------------------------------------------------------------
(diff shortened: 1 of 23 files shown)


maliberty:
My guess on the negative swaps is that undos are not properly counted

maliberty:
10 repairs sounds too low.  ca53 has > 3k hold buffers so that is a lot of messages.  Maybe 100?

src/rsz/test/repair_setup4_verbose.ok
```diff
@@ -0,0 +1,23 @@
...
+[INFO ODB-0132]     Created 2 special nets and 34 connections.
+[INFO ODB-0133]     Created 7 nets and 30 connections.
+worst slack -1.95
+[INFO RSZ-0094] Found 6 endpoints with setup violations.
+Iteration | Resized | Buffers | Cloned Gates | Pin Swaps |   WNS   |   TNS   | Endpoint
+---------------------------------------------------------------------------------------
+        0 |       0 |       0 |            0 |         0 |  -1.954 |  -6.294 | r2/D
+       10 |       8 |       3 |            0 |         0 |  -0.865 |  -3.085 | r2/D
+       19 |      17 |       3 |            0 |        -2 |  -0.654 |  -2.908 | r2/D
+    final |      17 |       3 |            0 |        -2 |  -0.654 |  -2.908 | r2/D
```
maliberty, line 18: Please sort out the negative values before we merge.

openroadie: I don't think we should combine what is possibly a counting error in pin swapping with a reporting change (which exposed the error). I would suggest that we merge this. I can take a look at the -ve pin swap number (will look right now).

openroadie: https://github.com/The-OpenROAD-Project/OpenROAD/pull/3630

maliberty (approved):

PR #3625, opened 2023-07-15
gui dbITerm selection

### LINKED ISSUE #3409: Allow selection of dbITerm in the GUI
### Description

If you click on an iterm the instance is selected today.  It would be good if you could directly select the iterm.

### Suggested Solution

The existing instance selection rtree can be kept but then a secondary search inside the master to see if an mterm is selected should be made.  If one is found then the corresponding iterm can be found.

### Additional Context

_No response_

Iyury1:
Fixes #3409 

I am having some trouble getting clang-tidy to run locally, but I want to finish with this issue first.

BTW I asked some questions on the issue page if someone could take a look at those! thanks

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -887,10 +887,28 @@ void LayoutViewer::selectAt(odb::Rect region, std::vector<Selected>& selections)
                                    region.yMax(),
                                    instanceSizeLimit());
 
-  for (auto& [box, inst] : insts) {
-    if (options_->isInstanceVisible(inst)
-        && options_->isInstanceSelectable(inst)) {
-      selections.push_back(gui_->makeSelected(inst));
+  if (options_->areITermsVisible() && options_->areITermsSelectable()) {
+    for (auto& [box, inst] : insts) {
+      if (options_->isInstanceVisible(inst)
+          && options_->isInstanceSelectable(inst)) {
+        selections.push_back(gui_->makeSelected(inst));
+      }
+      for (auto iterm : inst->getITerms()) {
+        Rect iterm_bbox = iterm->getBBox();
+        if ((region.xMin() > iterm_bbox.xMax()
+             || region.xMax() < iterm_bbox.xMin())
+            || (region.yMin() > iterm_bbox.yMax()
+                || region.yMax() < iterm_bbox.yMin())) {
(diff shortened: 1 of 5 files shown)


maliberty:
I'm not sure why iterm selection is tied to iterm labels.  You should be able to select the iterms even if the labels aren't visible.

maliberty:
Selection seems odd - when I click on an instance it seems all the iterms are overlapping.  Repeated clicking cycles through all of them, even those not under the cursor

Iyury1:
Now the ITerms are selectable depending on their Instances.

My logic for checking overlap of rectangles was wrong. It should work fine now.

I tested this change with nangate45 gcd. For the VSS and VDD ITerms, the placement of the labels doesn't seem to align completely with where you can select the ITerm. I think the text may have been translated too far.

src/gui/src/layoutViewer.cpp
```diff
@@ -887,19 +887,19 @@ void LayoutViewer::selectAt(odb::Rect region, std::vector<Selected>& selections)
...
-            || region.yMax() < iterm_bbox.yMin()) {
-          selections.push_back(gui_->makeSelected(iterm));
+        for (auto iterm : inst->getITerms()) {
+          Rect iterm_bbox = iterm->getBBox();
+          if (region.xMin() <= iterm_bbox.xMax()
+              && region.xMax() >= iterm_bbox.xMin()
+              && region.yMin() <= iterm_bbox.yMax()
+              && region.yMax() >= iterm_bbox.yMin()) {
+            selections.push_back(gui_->makeSelected(iterm));
+          }
```
rovinski, line 901: Use [`Rect::contains()`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/74b75f764fb96332c08f7af9c073b43cba617c4d/src/odb/include/odb/geom.h#L266-L267)

Iyury1: I replaced it with Rect::instersects() since it matches the logic I was using. Intersect will allow for a points on the edge of an iterm bbox to count as a selection, as well as rubber bands partially covering an iterm bbox.

rovinski: It depends on which Rect's perspective but either works.

maliberty (approved):
Thanks!

PR #3627, opened 2023-07-15
psm: dont require liberty models or set voltages when running check_power_grid

gadfort:
Changes:
- Avoids the check for voltages when doing connectivity checks as this is not required.

Adds:
- test for functionality of new behavior.

PR #3628, opened 2023-07-15
psm: cleanup tests to use common platform files where possible

gadfort:
While working on #3627 I noticed the tests are still using copies of the platform files.

Changes:
- no functional changes.
- Remove platform files in favor of /test files, asap7.lib is the only one remaining as this is not the FF corner and results in very very different answer.

.ok files were updated and only contain changes to based on the file paths.


maliberty (approved):
Thanks for the cleanup

PR #3629, opened 2023-07-15
sta: ensure heatmap and renderer is installed

gadfort:
Changes:
- make sure the power density heatmap is available when invoked via `gui::show`

### DIFF
--- a/src/dbSta/src/MakeDbSta.cc
+++ b/src/dbSta/src/MakeDbSta.cc
@@ -49,11 +49,9 @@ void initDbSta(OpenRoad* openroad)
   sta->initVars(openroad->tclInterp(), openroad->getDb(), logger);
   sta::Sta::setSta(sta);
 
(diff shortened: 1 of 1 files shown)


maliberty (approved):

QuantamHD:
@gadfort This breaks my version of openroad when you don't build the GUI? I need to make sure that the register renderer is gated. How can we make sure both of our use cases work?

gadfort:
@QuantamHD I suspect the issue is that stub.cpp isn't handling things correctly.

gadfort:
What breakage are you seeing?

QuantamHD:
Yeah I'm seeing a segfault, because GUI::GUI::get() returns nullptr

gadfort:
Do you have a stacktrace for that?

gadfort:
I think you will need to be more specific about the segfault, I'm not able to reproduce it

PR #3630, opened 2023-07-16
Negative pin swap count

openroadie:
Fix an issue with pin swap counting.


openroadie:
@vvbandeira @maliberty CI/CD fails with this:

+ ./test/regression aes_sky130hd
aes_sky130hd (tcl) *FAIL* DRT::max_capacitance_slack  -1% <  -1%

How do we handle these situations ?

maliberty:
> @vvbandeira @maliberty CI/CD fails with this:
> 
> * ./test/regression aes_sky130hd
>   aes_sky130hd (tcl) _FAIL_ DRT::max_capacitance_slack  -1% <  -1%
> 
> How do we handle these situations ?  

I assume it is failing but the printing is rounded and not showing something like -1.000001% < -1.000000%.  The format string is "%3.0f%%".  I would update the metrics as usual.  (@vvbandeira )

openroadie:
@vvbandeira : I tried updating the metrics file in OpenROAD/tests/ from OpenRoad/tests/results but it did not really help. Please let me know how to fix this and I will be happy to do so.

vvbandeira:
@openroadie I believe that to update the metrics in OR you have to run:

```
cd test
./save_flow_metrics aes_sky130hd
./save_flow_metrics_limits aes_sky130hd
```

Did you run both or only the first one?

openroadie:
@maliberty ready to merge this PR unless there are other concerns.

maliberty (approved):

PR #3633, opened 2023-07-16
rsz: fix layerRC for vias

gadfort:
Issue:
- `set_layer_rc` is using the routing level to store the layer RCs, this causes the vias to be stored and overwrite eachother.

Changes:
- use `getNumber` instead of `getRoutingLevel` to store and recover the layer RCs so vias are recorded correctly.

### DIFF
--- a/src/rsz/src/EstimateWireParasitics.cc
+++ b/src/rsz/src/EstimateWireParasitics.cc
@@ -81,8 +81,8 @@ Resizer::setLayerRC(dbTechLayer *layer,
     }
   }
 
-  layer_res_[layer->getRoutingLevel()][corner->index()] = res;
-  layer_cap_[layer->getRoutingLevel()][corner->index()] = cap;
+  layer_res_[layer->getNumber()][corner->index()] = res;
+  layer_cap_[layer->getNumber()][corner->index()] = cap;
 }
 
 void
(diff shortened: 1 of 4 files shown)


src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -92,11 +92,11 @@ Resizer::layerRC(dbTechLayer *layer,
                  double &res,
                  double &cap) const
 {
-  layerRC(layer->getRoutingLevel(), corner, res, cap);
+  layerRC(layer->getNumber(), corner, res, cap);
 }
 
 void
-Resizer::layerRC(int routing_level,
+Resizer::layerRC(int layer_level,
```
maliberty, line 99: Have you verified all callers (particular in grt)?

gadfort: yeah, they all call with the pointer to the techlayer (which resolves it into the correct number), I couldn't find any instances that call based on the routing_level.

maliberty: In that case let's get rid of this one and merge it into the other overload so nobody can make a mistake

gadfort: @maliberty good suggestion, turns out there was one user of it in BufferNet (so when I compiled with it protected it would still have been available). I made BufferNet use the layerRC(dbTechLayer*).

maliberty (approved):

PR #3636, opened 2023-07-17
deltaDebug.py: improve heuristic to speed up case of using many cuts

oharboe:
@maliberty Thoughts? This speeds up the case where the level of detail for the cuts is high as it isn't necessary to restart from scratch to find many errors at the same level of detail.


oharboe:
I need more time on this. Ignore for now.

PR #3638, opened 2023-07-17
psm: analysis across corners

gadfort:
Changes:
- adds a `-corner` flag to `analyze_power_grid` and `write_pg_spice` to allow PSM to get power numbers for a specific corner and parasitic values from resizer.
- If `-corner` is not used, there are no changes in behavior.

### DIFF
--- a/src/psm/src/ir_solver.cpp
+++ b/src/psm/src/ir_solver.cpp
@@ -128,7 +128,7 @@ IRSolver::IRSolver(odb::dbDatabase* db,
   net_voltage_map_ = net_voltage_map;
 
   if (corner == nullptr) {
-    logger_->error(utl::PSM, 84, "A valid corner is required.");
+    corner = sta_->cmdCorner();
   }
   corner_ = corner;
 }
--- a/src/psm/src/ir_solver.cpp
+++ b/src/psm/src/ir_solver.cpp
@@ -52,6 +52,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "gmat.h"
 #include "node.h"
 #include "odb/db.h"
+#include "rsz/Resizer.hh"
 
 namespace psm {
 using odb::dbBlock;
@@ -93,6 +94,7 @@ using Eigen::VectorXd;
 
 IRSolver::IRSolver(odb::dbDatabase* db,
                    sta::dbSta* sta,
+                   rsz::Resizer* resizer,
                    utl::Logger* logger,
                    const std::string& vsrc_loc,
                    const std::string& power_net,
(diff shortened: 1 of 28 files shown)


src/psm/src/ir_solver.cpp
```diff
@@ -1504,4 +1513,30 @@ bool IRSolver::buildConnection()
   result_ = res;
   return result_;
 }
+
+double IRSolver::getResistance(odb::dbTechLayer* layer) const
+{
+  double res;
+
+  if (layer->getRoutingLevel() == 0) {
```
maliberty, line 1521: Why the special case for this?

src/psm/test/corners.ok
```diff
@@ -0,0 +1,34 @@
...
+[INFO PSM-0015] Reading location of VDD and VSS sources from Vsrc_gcd_vdd.loc.
+[INFO PSM-0076] Setting metal node density to be standard cell height times 5.
+[WARNING PSM-0030] VSRC location at (50.000um, 50.000um) and size 20.000um, is not located on an existing power stripe node. Moving to closest node at (68.070um, 53.115um).
+[INFO PSM-0031] Number of PDN nodes on net VDD = 604.
+[INFO PSM-0064] Number of voltage sources = 1.
+[INFO PSM-0040] All PDN stripes on net VDD are connected.
+########## IR report #################
+Worstcase voltage: 6.12e-01 V
+Average IR drop  : 4.42e-01 V
+Worstcase IR drop: 4.88e-01 V
```
maliberty, line 21: I think it would be good if the output indicated the corner it comes from.

maliberty (changes requested):

src/psm/src/ir_solver.cpp
gadfort: For vias, resizer records that layer as the resistance of a via, while the routing layers get recorded as ohm/meter (and PSM needs the vias resistance, so no change ot that, but the routing layers it needs the sheet resistance of the metal).

src/psm/test/corners.ok
gadfort: I can add that, anything else that would be helpful while I'm in there?

maliberty: Nothing comes to mind

src/psm/src/ir_solver.cpp
maliberty: That makes sense - a little comment might be useful for others.

maliberty:
Won't this require some .ok to be updated?

gadfort:
> Won't this require some .ok to be updated?

Yes, they were updated in: https://github.com/The-OpenROAD-Project/OpenROAD/pull/3638/commits/5c048e7c8d5fc2ab022f6d9a07e64f97441eb202
I just had to correct the formatting in the last push

maliberty (approved):

PR #3645, opened 2023-07-17
psm: add ability to read in bterms for sources

gadfort:
Changes:
- adds ability to read in bterms to place sources, if they are available, otherwise the default checkerboard pattern is used (as before).
- reworks the debug gui to see IR drop values

Also includes changes to remove C4 bump naming scheme and replace with sources.

@maliberty I defaulted to a pitch of 10 tracks from above (turns out that is a lot more dense than the nodes in the solver). We can play with the spacing if you want from above.


src/psm/src/ir_solver.cpp
```diff
@@ -452,14 +456,74 @@ void IRSolver::readC4Data(bool require_voltage)
...
+                              bump_size_ * unit_micron,
+                              supply_voltage_src,
+                              true});
         }
       }
     }
   }
 }
 
+bool IRSolver::createBTerms(dbNet* net, double voltage)
```
maliberty, line 470: Confusingly named as it doesn't create bterms.  Perhaps createSourcesFromBTerms?

maliberty (changes requested):
Does it make sense to use node_density_ as the default rather than 10 pitches?

gadfort:
@maliberty from what I can tell the node_density only applies to the bottom layer (the node is incredibly hard to follow for node creation), using it on a high metal layer would result it a huge number of source nodes that eventually would just be assigned to the few nodes already present. At 10 x pitch there were a lot more source nodes than actual nodes. I would suggest keeping it 10, and maybe look at cleaning up the node creation code so it's much clearer to figure out what is going on.

maliberty (approved):

gadfort:
@maliberty I don't know why the docs failed to build, I don't see a place to retrigger it to see if its an actual issue

maliberty:
> @maliberty I don't know why the docs failed to build, I don't see a place to retrigger it to see if its an actual issue

@vvbandeira any ideas?  I'm inclined to ignore it but I also don't know how to retrigger it.

PR #3646, opened 2023-07-18
drt: remove reading from parameter file

rovinski:


src/drt/src/TritonRoute.tcl
```diff
@@ -79,131 +78,124 @@ proc detailed_route { args } {
   # development.  It is not listed in the help string intentionally.
   set single_step_dr  [expr [info exists flags(-single_step_dr)]]
   set save_guide_updates  [expr [info exists flags(-save_guide_updates)]]
-  if { [info exists keys(-param)] } {
-    if { [array size keys] > 1 } {
-      utl::error DRT 251 "-param cannot be used with other arguments"
```
maliberty, line 84: I would call this removing not deprecating but I'm ok with it in any case.

maliberty (approved):

PR #3647, opened 2023-07-18
ppl: update annealing to support pin constraints

eder-matheus:


src/ppl/src/IOPlacer.cpp
```diff
@@ -1507,11 +1507,12 @@ void IOPlacer::addVerLayer(odb::dbTechLayer* layer)
...
 }
 
-void IOPlacer::getPinsFromDirectionConstraint(Constraint& constraint)
+void IOPlacer::getPinsFromDirectionConstraint(Constraint& constraint,
+                                              int constraint_idx)
 {
   if (constraint.direction != Direction::invalid
       && constraint.pin_list.empty()) {
-    for (const IOPin& io_pin : netlist_io_pins_->getIOPins()) {
+    for (IOPin& io_pin : netlist_io_pins_->getIOPins()) {
```
maliberty, line 1515: why drop the const?

src/ppl/src/IOPlacer.cpp
```diff
@@ -1507,11 +1507,12 @@ void IOPlacer::addVerLayer(odb::dbTechLayer* layer)
   ver_layers_.insert(layer->getRoutingLevel());
 }
 
-void IOPlacer::getPinsFromDirectionConstraint(Constraint& constraint)
+void IOPlacer::getPinsFromDirectionConstraint(Constraint& constraint,
+                                              int constraint_idx)
```
maliberty, line 1511: How is the new argument used?

maliberty (commented):
Is this handling mirroring or only grouping?

src/ppl/src/IOPlacer.cpp
eder-matheus: Fixed this and the above/ I've started with these two changes, but they're not necessary anymore.

eder-matheus:
> Is this handling mirroring or only grouping?

For this PR, only region constraints for lone pins and groups. Mirroring is my next (and probably last) step to have a complete version of PPL with Simulated Annealing.

maliberty (approved):

PR #3651, opened 2023-07-18
Recover power

openroadie:
Coverity issues.

ORFS testing here:

a few tests are expected to fail since we are testing with 

repair_timing -recover_power

But this will not be a default change in ORFS yet.


src/rsz/src/RecoverPower.cc
```diff
@@ -380,12 +380,10 @@ RecoverPower::downsizeCell(LibertyPort *in_port,
...
-           return drive1 < drive2
-             || ((drive1 == drive2
-                  && intrinsic1 > intrinsic2)
-                 || (intrinsic1 == intrinsic2
-                 || (intrinsic1 == intrinsic2
-                     && port1->capacitance() > port2->capacitance())));
+
+           return ( drive1 < drive2 || 
+		    (drive1 == drive2 && intrinsic1 > intrinsic2) ||
+		    (intrinsic1 == intrinsic2 && port1->capacitance() > port2->capacitance()));
```
maliberty, line 386: Is this right?  If drive1 > drive2 and  intrinsic1 == intrinsic2 then you will use port cap not drive strength.

I tend to use std::tie in these cases (see operator< example in https://en.cppreference.com/w/cpp/utility/tuple/tie)

openroadie: I tested with 10+ designs and we don't seem to need the capacitance business. ORFS is also running here with this code change + call to recover_power added to cts.tcl 

maliberty (approved):

PR #3660, opened 2023-07-19
psm: add sources from pads

gadfort:
Changes:
- add sources from pads to the power grid to enable wirebond IR drop


maliberty (approved):

maliberty:
I looked at zerosoc and I still see what look like the signs of bumps despite the pads:
image

gadfort:
image
@maliberty this is what I get when I run the rest (no appearance of the bumps)

maliberty:
It would be nice if the heatmap defaulted to the lowest populated layer.  Here it defaults to li1 which has no data and confusingly seems like a no-op.

maliberty:
An aside - this def is odd in that the tie cells are not placed that close to the pads:
image

gadfort:
@maliberty the smooth ness would be nice to find a good solution for.  I think we could default to something different (maybe), just depends on how and when the default it chosen.
As for tiecells, we aren't doing anything special here, so this is how OR chose to place them (and I agree that does look a little less ideal). Maybe they are connected to multiple things and that is causing them to move around? Might be worth investigating since this will cause a lot of extra routing.

maliberty:
At least that one tie I looked at had only a single connection.  Do you have a test case for gpl for this design?

PR #3667, opened 2023-07-20
pad: add place_io_terminals command to place bpins on iterms in the padring

gadfort:
Adds:
- ability to place bpins on iterms in the event that the pads contain integrated bondpads


src/pad/src/ICeWall.cpp
```diff
@@ -921,6 +921,65 @@ void ICeWall::placeBondPads(odb::dbMaster* bond,
...
+  for (auto* iterm : iterms) {
+    if (iterm == nullptr) {
+      continue;
+    }
+    auto* net = iterm->getNet();
+    if (net == nullptr) {
+      continue;
+    }
+    auto* inst = iterm->getInst();
+    if (!inst->isFixed()) {
```
maliberty, line 943: Should we check it is an IO?

rovinski:
Might be good to add an info message e.g. "placed X terminals" just to know that the command did something. It seems like if all iTerms get filtered out then nothing will appear to happen.

src/pad/src/ICeWall.cpp
gadfort: Good, point, we should.

maliberty (approved):

PR #3668, opened 2023-07-20
Gate cloning default

openroadie:
Gate cloning on by default. This needs to be merged together with the corresponding ORFS PR below or CI/CD will break

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1262


vijayank88:
regression fails
```
 ./test/regression aes_sky130hs
aes_sky130hs (tcl) *FAIL* DRT::max_capacitance_slack  -0% <   0%
```

```
./test/regression aes_sky130hd
aes_sky130hd (tcl) *FAIL* GRT::ANT::errors 371 > 365 ; DRT::max_capacitance_slack  -1% <  -1%
```

openroadie:
@vvbandeira ^^. I am going to skip gate cloning for these tests for now by making the requisite change in flow.tcl

vvbandeira:
@openroadie
If the default for ORFS will be gate cloning enabled, keeping this option enabled here as well might help prevent CI failure downstream on ORFS in the future. Wouldn't it be better to update the metrics for these two test cases and keep gate cloning enabled on these tests? 
I can do the update, or you can download the results from CI from here:
https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-3668-merge/1/artifact/test/results.tgz
https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-Public/detail/PR-3668-merge/1/artifacts

openroadie:
@vvbandeira The failures of not of a nature where I am entirely comfortably changing the metrics file. It is also surprising that the ORFS tests for the same designs pass without any issues (with gate cloning enabled). So I would leave this as is. We do have tests for gate cloning in rsz (it is enabled for all tests except maybe 1-2... but it is triggered only for 1 of them). I can help debug why ORFS results are different from OR results for these designs ... if you are sure that the input data (libs/netlists etc) is exactly the same.

openroadie:
secure CI/CD is here:

https://github.com/The-OpenROAD-Project-private/OpenROAD-flow-scripts/pull/670

@maliberty

maliberty:
> I have a DRC issue post detailed route on nangate45/bp_multi. Should we update the metrics or should we look at this issue in more detail? @maliberty

You can update the metrics but also open an issue for @osamahammad21 to look into it later.

openroadie:
@maliberty @osamahammad21 

https://github.com/The-OpenROAD-Project/OpenROAD/issues/3726

PR #3675, opened 2023-07-20
rsz: return if nothing to repair

gadfort:
Changes:
- avoids printing the verbose messages when there is nothing to do.

Fixes:
- segfault when clock network has been corrupted by CTS, causing OpenSTA to return nullptr vertexes during printing.


src/rsz/src/RepairSetup.cc
```diff
@@ -979,7 +982,7 @@ RepairSetup::printProgress(int iteration, bool force, bool end) const
 {
   const bool start = iteration == 0;
 
-  if (start) {
+  if (start && !end) {
```
maliberty, line 985: when would start && end both happen?

PR #3677, opened 2023-07-20
drt: keep searching pin access until two valid vias have been found

gadfort:
@maliberty I was thinking that PA could do something more like this.
In theory, if the first 2 vias are acceptable, then there is no additional runtime. If the first 2 are not, it will just continue until the entire list has been exhausted.

This is just an idea and we should definitely test it out before commit anything like this (this however does seem to find the vias in my testcase)


maliberty:
@osamahammad21 please review and test as needed

osamahammad21:
I ran a secure CI just to check if it messes with any of our test cases before reviewing the change and found 105 drvs on gf12/bp_single and double runtime in ispd19_test10.
I am looking into the change now to understand more what's going on.

gadfort:
@osamahammad21 the run time can probably be improved by adjusting the search exit criteria. The DRV was unfortunate seems to imply that the some of the accepted AP might not be so good

maliberty:
@osamahammad21 is the extra runtime in pa or dr?

osamahammad21:
The extra runtime in DR. All the signs suggest that the chosen vias in pa are not good for dr and make the router struggle, either with runtime or DRVs

gadfort:
Maybe the exit criteria is alright, but PA is not rejecting ones it should?

QuantamHD:
@maliberty I think this is also affecting me on my nodes could you have another look to get this over the line?

maliberty:
@osamahammad21 is working on it. It breaks other designs and so needs more work before it can be considered a complete solution and merged.  

@osamahammad21 please provide an update

QuantamHD:
Thanks for the update 👍

osamahammad21:
This change produces many DRVs on gf12 designs.  Most of the violations look as following:
Screenshot from 2023-07-31 22-53-04
They are MINSTEP and CORNER SPACING violations. The violations cannot be avoided because we don't do wrongway routing on gf12.  PA doesn't consider how we reach the via so it doesn't know that this via would always cause this violation because the M2 shapes would always form that T shape that causes these violations.
As a workaround, I changed the criteria to reject vias with enclosures that are not along the preferred routing direction. This brought down the number of violations from 106 violations to 1 violation. However, the runtime is still higher which means that the router is still struggling with the chosen vias. Moreover, the 1 violation is not directly connected to the chosen vias.

gadfort:
@osamahammad21 thanks for the update. It seems to me that there are several things that could be improved:
1. reject the APs when unidirectional routing is required on a given layer and the enclosure does not align (that seems completely reasonable).
2. the original APs would still be in this set, so why does the router not attempt those when it's struggling with the routing? Does the router only try a few of the APs or only one?

When you say the runtime was higher, do you have a measurement of the runtime impact?

maliberty:
> @gadfort Could you try out this branch https://github.com/The-OpenROAD-Project-private/OpenROAD/tree/TR-drt-pa-try-harder ? It has the suggested edits as discussed in the meeting.

I don't think he has access to the private repo so you need to make a PR and sync to public

gadfort:
Closing: #4095

PR #3679, opened 2023-07-20
drt: set VIA_ACCESS_LAYERNUM when adding the first routing level 

gadfort:
Issue:
if multiple vias are defined before the first routing level, the `VIA_ACCESS_LAYERNUM` will be pointing to a via instead of the first routing level.

Changes:
- set VIA_ACCESS_LAYERNUM when adding the first routing level to ensure it matches the layer stack
- just loop over layers to init gcells to avoid assumptions about the start of the layer stack
- minor changes to error message for guideIN to make it clear what the via access is.

### DIFF
--- a/src/drt/src/db/tech/frTechObject.h
+++ b/src/drt/src/db/tech/frTechObject.h
@@ -38,7 +38,9 @@
 #include "db/tech/frLayer.h"
 #include "db/tech/frViaRuleGenerate.h"
 #include "frBaseTypes.h"
+#include "global.h"
 #include "utl/Logger.h"
+
 namespace odb {
 class dbTechLayer;
 }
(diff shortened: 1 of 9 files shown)


maliberty:
drt has strong assumptions about the layer stack.  I think it is best to hide the extra via layer.  I expect more problems will arise otherwise.

gadfort:
@maliberty alright, it's skipping successive cut layers now

gadfort:
@maliberty anything else needed on this one?

maliberty:
Is there any need to mess with  VIA_ACCESS_LAYERNUM or buildGCellPatterns now that you skip successive layers?

gadfort:
I've removed it, but it looks like this PR and my branch are no longer connected... Might need to debug a bit

maliberty:
> I've removed it, but it looks like this PR and my branch are no longer connected... Might need to debug a bit

Maybe you forgot to push your change?

gadfort:
@maliberty thats what I thought too: https://github.com/gadfort/OpenROAD/tree/drt-skip-double-cuts
maybe I'll just close this one and reopen it again

gadfort:
@maliberty a force push fixed it, sorry about that

maliberty (approved):
clang-tidy nit

PR #3682, opened 2023-07-21
Proposal: percentage progress indication

joennlae:
This PR introduces percentage indications for `repairDesign`, `repairHold`, and `repairSetup`. I know that there is a debug logging option. 

We found this helpful for larger designs that can spend multiple hours in these stages. We wanted to propose this, but we also understand if there is an effort to keep terminal output to a minimum.

Thank you again for this fantastic piece of software 💯

### DIFF
--- a/src/rsz/src/RepairDesign.cc
+++ b/src/rsz/src/RepairDesign.cc
@@ -159,7 +159,13 @@ RepairDesign::repairDesign(double max_wire_length, // zero for none (meters)
 
   resizer_->incrementalParasiticsBegin();
   int max_length = resizer_->metersToDbu(max_wire_length);
+  int previous_progress = 0;
   for (int i = resizer_->level_drvr_vertices_.size() - 1; i >= 0; i--) {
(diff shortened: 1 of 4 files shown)


gadfort:
@joennlae seems like for repairSetup and repairHold the `-verbose` option might work for you. I agree that repairDesign should also have a `-verbose` option. It might be nice to have a complementary output for that with buffers inserted, violations remaining, etc. I'm not sure the percent really mean much in these contexts.

maliberty:
I agree with @gadfort that the report he created for -verbose is more descriptive.  What is your thought on moving in a similar direction for repair design & hold?

gadfort:
@maliberty it would be relatively easy to add something like those to both repairDesign and the new power command as well.

maliberty:
@gadfort please do 👍

gadfort:
@maliberty here you go: https://github.com/The-OpenROAD-Project/OpenROAD/pull/3688
I decided to skip recoverPower for now because it doesn't look like there are any unit tests for that functionality for me to borrow to test on

maliberty:
Closed in favor of https://github.com/The-OpenROAD-Project/OpenROAD/pull/3688

PR #3685, opened 2023-07-21
Discussion: allow different placement margin per instance also in global placement

joennlae:
There are better solutions than this here... But it is the fastest fix we saw atm.

We had the issue that we could not place our design when using `set_placement_density -instances` in detailed placement.

We used the way shown in this PR to get the information from the detailed placement command to the global placer, as the global placer supports per-instance padding.

Examples of our hotspots:

```tcl
set_placement_padding -global -left 2 -right 2
set_placement_padding -instances [get_cells *i_bootrom*] -right 5 -left 5
set_placement_padding -instances [get_cells *gen_asic_regfile_i_ariane_regfile*] -right 7 -left 7
set_placement_padding -instances [get_cells *gen_asic_fp_regfile_i_ariane_fp_regfile*] -right 7 -left 7
set_placement_padding -instances [get_cells *i_scoreboard*] -right 5 -left 5
set_placement_padding -instances [get_cells *i_multiplier*] -right 5 -left 5
```

As mentioned above, this is not a nice solution as it mixes global placer and detailed placer. But we believe that having the per instance placement information in the global placer should help with the overall backend flow.

We are happy to discuss any other ideas or also if we have missed something :-).

### DIFF
--- a/src/gpl/include/gpl/Replace.h
+++ b/src/gpl/include/gpl/Replace.h
@@ -44,6 +44,10 @@ namespace sta {
 class dbSta;
 }
 
+namespace dpl {
+class Opendp;
+};
+
 namespace grt {
 class GlobalRouter;
 }
@@ -76,6 +80,7 @@ class Replace
   ~Replace();
 
   void init(odb::dbDatabase* odb,
+            dpl::Opendp* dp,
             rsz::Resizer* resizer,
             grt::GlobalRouter* router,
             utl::Logger* logger);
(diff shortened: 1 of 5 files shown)


maliberty:
I'm curious about the background of this solution.  What problem were you seeing that required such large per instance paddings?  Is it something you can share?

joennlae:
@maliberty The instances need a lower placement density, and this is the solution we could come up with. There may be a better solution that we could not find. 

We are very close to the open-source release. (Some cleanup to be NDA conform still ongoing). I will post here when it's finished.

maliberty:
> We are very close to the open-source release. (Some cleanup to be NDA conform still ongoing). I will post here when it's finished.

any update?

maliberty:
This pull request has been automatically marked as stale because it has not had recent activity. It will be closed in 21 days if no further activity occurs. Remove the Stale label or comment to keep it open.

PR #3688, opened 2023-07-21
rsz: add verbose to repair design

gadfort:
Alternative to https://github.com/The-OpenROAD-Project/OpenROAD/pull/3682

Adds:
- `-verbose` flag to `repair_design`

Prints the number of iterations, nets fixed, reized count, buffers inserted, and remaining vertices to check


gadfort:
@joennlae FYI

src/rsz/src/RepairDesign.cc
```diff
@@ -1385,4 +1404,35 @@ RepairDesign::metersToDbu(double dist) const
   return dist * dbu_ * 1e+6;
 }
 
+void
+RepairDesign::printProgress(int iteration, bool force, bool end, int repaired_net_count) const
+{
+  const bool start = iteration == 0;
+
+  if (start && !end) {
+    logger_->report("Iteration | Resized | Buffers | Nets repaired | Left to check");
```
rovinski, line 1413: Maybe "nets remaining" instead of "left to check"?

src/rsz/test/repair_design3_verbose.ok
```diff
@@ -0,0 +1,29 @@
...
+      300 |     298 |       0 |             0 | 702
+      400 |     398 |       0 |             0 | 602
+      500 |     498 |       0 |             0 | 502
+      600 |     598 |       0 |             0 | 402
+      700 |     698 |       0 |             0 | 302
+      800 |     798 |       0 |             0 | 202
+      900 |     898 |       0 |             0 | 102
+     1000 |     998 |       0 |             0 | 2
+    final |    1000 |       0 |             0 | 0
+-------------------------------------------------------------
```
rovinski, line 28: Right justify last column?

src/rsz/src/RepairDesign.cc
gadfort: I thought about doing that, but it's not actually nets, it's `Vertex`'s in OpenSTA ,so there are usually more than the number of nets. If you have a better suggestion for the phrasing I'm open since it's not great as it is.

src/rsz/test/repair_design3_verbose.ok
gadfort: sure, good point, I copied this from the other -verbose flags where this is the worstcase endpoint

src/rsz/src/RepairDesign.cc
rovinski: Maybe just "remaining" or "checks remaining"? It's fine either way I just like the formal-sounding term.

gadfort: I'll make it remaining

src/rsz/src/RepairDesign.cc
```diff
@@ -158,8 +161,21 @@ RepairDesign::repairDesign(double max_wire_length, // zero for none (meters)
...
+    print_interval_ = min_print_interval_;
+  }
+  if (verbose) {
+    printProgress(print_iteration, false, false, repaired_net_count);
+  }
   int max_length = resizer_->metersToDbu(max_wire_length);
   for (int i = resizer_->level_drvr_vertices_.size() - 1; i >= 0; i--) {
+    print_iteration++;
+    if (verbose) {
+      printProgress(print_iteration, false, false, repaired_net_count);
```
maliberty, line 177: How much output does this generate on a large design?

maliberty:
https://github.com/The-OpenROAD-Project/OpenROAD/pull/3682 also metions repair hold - do you plan to do that as well?

maliberty (approved):

src/rsz/src/RepairDesign.cc
gadfort: I implemented some back off for larger designs to only print every 1000, but if it turns out to be too verbose we can look into different strategies for printing

gadfort:
@maliberty repair hold already has it

PR #3689, opened 2023-07-21
cts: Move length units to DBU

QuantamHD:
Fixes divide by zero in some PDKs


src/cts/src/CtsOptions.h
```diff
@@ -148,9 +149,16 @@ class CtsOptions
   {
     clusteringCapacity_ = capacity;
   }
-  double getBufferDistance() const { return bufDistance_; }
+
+  // BufferDistance is set in microns, but returned in DBU
+  uint32_t getBufferDistance() const { return bufDistance_ * getDbUnits(); }
```
maliberty, line 154: why not int?  https://google.github.io/styleguide/cppguide.html#Integer_Types
```In particular, do not use unsigned types to say a number will never be negative.```

In general I would switch this over to dbu internally and put the user micron conversion near UI

Comment (QuantamHD) [reply src/cts/src/CtsOptions.h:154]: I did think about that, but these units in particular have default values in microns that makes it hard to initialize without knowing what the DBU conversion is.

I think it would be cleaner to have DBU internally, but it seems difficult to accommodate the default behavior.

Comment (maliberty) [reply src/cts/src/CtsOptions.h:154]: Perhaps the default could be a sentinel value (eg -1, std::optional) that is converted to dbu at the command start.

**Comment by maliberty** [comment]
I started a secure CI for safety since there are small changes.

**Comment by QuantamHD** [comment]
Done.

File: src/cts/src/CtsOptions.h [hunk 2]
```diff
@@ -148,12 +150,40 @@ class CtsOptions
...
+      return *bufDistance_;
+    }
+
+    if (dbUnits_ == -1) {
+      logger_->error(
+          utl::CTS, 542, "Must provide a dbUnit conversion though setDbUnits.");
+    }
+
+    return 100 /*um*/ * dbUnits_;
+  }
```
maliberty, line 167: Out-of-line.  Also why int32_t and not just int?  Similarly  getVertexBufferDistance

maliberty (changes requested):

maliberty:
@vvbandeira will manage the metrics updates

PR #3693, opened 2023-07-22
ppl: support for mirrored pins inside annealing

eder-matheus:
For now, mirrored pins and groups are only moved inside the movePinToFreeSlot function.

Swapping mirrored pins has a level of complexity similar to swapping pin groups, and they will be added to the swap in a future PR.

### DIFF
--- a/src/ppl/src/SimulatedAnnealing.cpp
+++ b/src/ppl/src/SimulatedAnnealing.cpp
@@ -42,11 +42,13 @@ namespace ppl {
 
 SimulatedAnnealing::SimulatedAnnealing(
     Netlist* netlist,
+    Core* core,
     std::vector<Slot>& slots,
     const std::vector<Constraint>& constraints,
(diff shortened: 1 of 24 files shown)


src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -177,19 +183,48 @@ void SimulatedAnnealing::randomAssignment()
                                                                 last_slot);
 
       int slot = distribution(generator_);
-      while (slots_[slot].used) {
+      int mirrored_slot;
+      bool free = !slots_[slot].used;
+      if (io_pin.isMirrored()) {
+        free = isFreeForMirrored(slot, mirrored_slot);
```
maliberty, line 189: Should this be free = free && isFreeForMirrored(slot, mirrored_slot); ?  Likewise below

maliberty (approved):

Issue #3583, opened 2023-07-05
Save DRC reports intermittently for global routing for hard runs


oharboe:
### Description

If global routing is having problmes converging, it can run for many hours.

However, even after a few minutes, a DRC report could contain valuable information about where the routing problems are.

### Suggested Solution

Save out DRC reports "occasionally". This should be in some idempotent interval, i.e. wall clock time or type of machine that global routing is running on should not matter.

Furthermore, this should be the default behavior. Spelunking in .tcl scripts shouldn't be required to get information on why global routing is struggling.

maliberty:
In general it isn't safe to issue a message from a signal handler (see https://stackoverflow.com/questions/16891019/how-to-avoid-using-printf-in-a-signal-handler).  It is also non-standard that ctrl-c make run another 10+ minutes.  I know of no other programs that do so.

I think it better to just write the report on each iteration.  We could have some upper limit on the number of violations to prevent it from generating huge reports on the early iterations (say 100k).

QuantamHD:
I think that's acceptable @maliberty, but dd is an example where you can interrupt the program for a status report. https://askubuntu.com/a/215521

maliberty:
dd gives an immediate result - I've seen that in other programs.  This would be quite different where the response occurs after an unbounded amount of time.

oharboe:
> dd gives an immediate result - I've seen that in other programs. This would be quite different where the response occurs after an unbounded amount of time.
> 
> It also not a status report - it still eventually exits the program.

The approach may require some care and not be entirely typical, but it is not unprecedented. I'm fine with any solution to this reporting problem where I don't have to configure the build or go spelunking in .tcl.

I haven't checked, but I asked chatgpt...


MySQL Server: MySQL Server is a highly popular relational database management system. When you send a SIGINT signal to MySQL, it doesn't immediately terminate. It attempts to finish active transactions, roll back uncommitted transactions, and clean up the system, in order to maintain data consistency before shutting down.

Apache HTTP Server: As a popular web server application, Apache has to handle many active connections simultaneously. When a SIGINT signal is sent, it tries to gracefully stop by allowing active connections to complete while not accepting new connections.

TensorFlow: This machine learning library used for high performance numerical computation can run processes that take a significant amount of time, such as training neural networks. When interrupted, it might need to save the state of the model before exiting.

[...]

PR #3712, opened 2023-07-25
Grt: Add -drc_report_iter_step option

luis201420:
Report the DRC on each iteration of overflow. By default only the final report is generated. (Discussed in #3583)

### DIFF
--- a/src/grt/src/GlobalRouter.tcl
+++ b/src/grt/src/GlobalRouter.tcl
@@ -193,7 +193,7 @@ proc set_global_routing_random { args } {
 sta::define_cmd_args "global_route" {[-guide_file out_file] \
                                   [-congestion_iterations iterations] \
                                   [-congestion_report_file file_name] \
-                                  [-drc_report_iter_step steps] \
+                                  [-congestion_report_iter_step steps] \
                                   [-grid_origin origin] \
                                   [-overflow_iterations iterations] \
                                   [-critical_nets_percentage percent] \
@@ -207,7 +207,7 @@ sta::define_cmd_args "global_route" {[-guide_file out_file] \
 proc global_route { args } {
   sta::parse_key_args "global_route" args \
     keys {-guide_file -congestion_iterations -congestion_report_file \
-          -overflow_iterations -grid_origin -critical_nets_percentage -drc_report_iter_step
+          -overflow_iterations -grid_origin -critical_nets_percentage -congestion_report_iter_step
          } \
     flags {-allow_congestion -allow_overflow -verbose -start_incremental -end_incremental}
 
@@ -248,11 +248,11 @@ proc global_route { args } {
     grt::set_congestion_report_file $file_name
   }
 
-  if { [info exists keys(-drc_report_iter_step) ] } {
-    set steps $keys(-drc_report_iter_step)
-    grt::set_drc_report_iter_step $steps
+  if { [info exists keys(-congestion_report_iter_step) ] } {
+    set steps $keys(-congestion_report_iter_step)
+    grt::set_congestion_report_iter_step $steps
   } else {
-    grt::set_drc_report_iter_step 0
+    grt::set_congestion_report_iter_step 0
   }
 
   if { [info exists keys(-overflow_iterations)] } {
--- a/src/grt/src/GlobalRouter.tcl
+++ b/src/grt/src/GlobalRouter.tcl
@@ -193,7 +193,7 @@ proc set_global_routing_random { args } {
 sta::define_cmd_args "global_route" {[-guide_file out_file] \
                                   [-congestion_iterations iterations] \
                                   [-congestion_report_file file_name] \
-                                  [-drc_report_iter_step] \
+                                  [-drc_report_iter_step steps] \
                                   [-grid_origin origin] \
                                   [-overflow_iterations iterations] \
                                   [-critical_nets_percentage percent] \
@@ -249,8 +249,8 @@ proc global_route { args } {
   }
 
   if { [info exists keys(-drc_report_iter_step) ] } {
-    set drc_report_iter_step $keys(-drc_report_iter_step)
-    grt::set_drc_report_iter_step $drc_report_iter_step
+    set steps $keys(-drc_report_iter_step)
+    grt::set_drc_report_iter_step $steps
   } else {
     grt::set_drc_report_iter_step 0
   }
--- a/src/grt/src/GlobalRouter.tcl
+++ b/src/grt/src/GlobalRouter.tcl
@@ -193,6 +193,7 @@ proc set_global_routing_random { args } {
 sta::define_cmd_args "global_route" {[-guide_file out_file] \
                                   [-congestion_iterations iterations] \
                                   [-congestion_report_file file_name] \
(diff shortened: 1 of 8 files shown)


src/grt/src/GlobalRouter.tcl
```diff
@@ -193,6 +193,7 @@ proc set_global_routing_random { args } {
 sta::define_cmd_args "global_route" {[-guide_file out_file] \
                                   [-congestion_iterations iterations] \
                                   [-congestion_report_file file_name] \
+                                  [-drc_report_iter_step] \
```
maliberty, line 196: this takes an argument eg [-drc_report_iter_step step]

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2302,6 +2304,107 @@ int FastRouteCore::edgeShiftNew(Tree& t, int net)
...
+    final_x = route.init_x;
+  }
+
+  if (route.init_y < route.final_y) {
+    init_y = route.init_y;
+    final_y = route.final_y;
+  } else {
+    init_y = route.final_y;
+    final_y = route.init_y;
+  }
```
maliberty, line 2326: Simplify to:
const auto [init_x, final_x] = std::minmax(route.init_x, route.final_x);
const auto [init_y, final_y] = std::minmax(route.init_y, route.final_y);

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2302,6 +2304,107 @@ int FastRouteCore::edgeShiftNew(Tree& t, int net)
...
+
+  int llX = init_x - (tile_size_ / 2);
+  int llY = init_y - (tile_size_ / 2);
+
+  int urX = final_x + (tile_size_ / 2);
+  int urY = final_y + (tile_size_ / 2);
+
+  if ((x_grid_max_ - urX) / tile_size_ < 1) {
+    urX = x_grid_max_;
+  }
```
maliberty, line 2336: Is this just
```
urX = std::min(urX, x_grid_max_);
```
or is it something more complex?  Its a bit hard to follow. 

Why is nothing similar needed with llX?

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2302,6 +2304,107 @@ int FastRouteCore::edgeShiftNew(Tree& t, int net)
...
+  if ((y_grid_max_ - urY) / tile_size_ < 1) {
+    urY = y_grid_max_;
+  }
+
+  odb::Point lower_left = odb::Point(llX, llY);
+  odb::Point upper_right = odb::Point(urX, urY);
```
maliberty, line 2342: const

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -1167,6 +1168,11 @@ NetRouteMap FastRouteCore::run()
     }
 
     last_total_overflow = total_overflow_;
+
+    // generate DRC report each interval
+    if (drc_report_iter_step_ && i % drc_report_iter_step_ == 0) {
+      saveCongestion();
```
maliberty, line 1174: Is this overwriting the same file each time?
(@oharboe is that what you want?)

src/grt/src/GlobalRouter.tcl
luis201420: The argument was added

src/grt/src/fastroute/src/utility.cpp
luis201420: The change was applied

src/grt/src/fastroute/src/utility.cpp
luis201420: I used the same GlobalRouter::globalRoutingToBox operation, 
I think it's when the last tile is less width than the `tile_size`

src/grt/src/fastroute/src/utility.cpp
luis201420: The change was applied

src/grt/src/fastroute/src/FastRoute.cpp
luis201420: I modified to generate a file for each interval (eg. `congestion-step.rpt` where `step` is the iteration number)

src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -99,7 +99,7 @@ class FastRouteCore
   ~FastRouteCore();
 
   void clear();
-  void saveCongestion();
+  void saveCongestion(const int& iter = -1);
```
maliberty, line 102: Just use int - pass by value is better for primitive types.

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2351,57 +2335,72 @@ double FastRouteCore::dbuToMicrons(int64_t dbu)
   return (double) dbu / (block->getDbUnitsPerMicron());
 }
 
-void FastRouteCore::saveCongestion()
+void FastRouteCore::saveCongestion(const int& iter)
```
maliberty, line 2338: const int

maliberty (changes requested):

maliberty:
@eder-matheus any comments?

src/grt/src/fastroute/include/FastRoute.h
oharboe: use std::option instead of overloading int? self documenting...

src/grt/src/fastroute/src/utility.cpp
eder-matheus: Yes, this is checking if the last tile is smaller than tile_size, and creating a larger guide for the last row/column of the route grid.

eder-matheus:
@luis201420 What is the default behavior? In the https://github.com/The-OpenROAD-Project/OpenROAD/issues/3583 issue, it is required that the reports should be generated by default. If I understand correctly, this code is not generating the reports by default.

@oharboe What should be the default interval between iterations to generate the congestion report?

oharboe:
@eder-matheus The OpenROAD tool should not change behavior, but I have discussed with @maliberty the ORFS policy on when and how to generate interim .odb files, so I suggest you discuss with him. I am not the only user who has concerns on this and Matt is a bit more on top of other inputs.

maliberty:
I think it  should be off by default at the OR level and we can discuss the default in ORFS

src/grt/src/GlobalRouter.tcl
```diff
@@ -193,6 +193,7 @@ proc set_global_routing_random { args } {
 sta::define_cmd_args "global_route" {[-guide_file out_file] \
                                   [-congestion_iterations iterations] \
                                   [-congestion_report_file file_name] \
+                                  [-drc_report_iter_step steps] \
```
eder-matheus, line 196: I'm not sure if this is the best name for this option. Maybe "report_partial_congestion iteration_interval", or something like that.

Also, please add the description of this new option to the README.md file.

eder-matheus (approved):
LGTM.

maliberty (approved):

PR #3717, opened 2023-07-25
[Docs] rmp documentation

luarss:
Fix missing documentation and formatting issues

### DIFF
--- a/src/rmp/README.md
+++ b/src/rmp/README.md
@@ -10,56 +10,71 @@ The ABC output is read back by a `blif` reader which is integrated to OpenDB.
 back of constants requires insertion of tie cells which should be provided
 by the user as per the interface described below.
 
-
 ## Commands
+
+```{note}
+- Parameters in square brackets `[-param param]` are optional.
+- Parameters without square brackets `-param2 param2` are required.
+```
 
 Restructuring can be done in two modes: area or delay.
 
 ### Area Mode
 
-```
-restructure -liberty_file <liberty_file>
-            -target "area"
-            -tielo_pin  <tielo_pin_name>
-            -tiehi_pin  <tiehi_pin_name>
+```tcl
+restructure 
+    -liberty_file liberty_file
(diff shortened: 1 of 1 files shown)


src/rmp/README.md
```diff
@@ -10,56 +11,71 @@ The ABC output is read back by a `blif` reader which is integrated to OpenDB.
...
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-liberty_file` | Liberty file with description of cells used in design. This is passed to ABC. |
+| `-target` | Either `area` or `delay`. In area mode, the focus is area reduction and timing may degrade. In delay mode, delay is likely reduced but area may increase. |
+| `-slack_threshold` | Specifies a (setup) timing slack value below which timing paths need to be analyzed for restructuring. |
+| `-depth_threshold` | Specifies the path depth above which a timing path would be considered for restructuring. |
+| `tielo_pin` | Specifies the tie cell pin which can drive constant zero. Format is `lib/cell/pin`. |
+| `tiehi_pin` | Specifies the tie cell pin which can drive constant one. Format is `lib/cell/pin`. |
```
vvbandeira, line 54:
```suggestion
| `-liberty_file` | Liberty file with description of cells used in design. This is passed to ABC. |
| `-target` | Either `area` or `delay`. In area mode, the focus is area reduction, and timing may degrade. In delay mode, delay is likely reduced, but the area may increase. |
| `-slack_threshold` | Specifies a (setup) timing slack value below which timing paths need to be analyzed for restructuring. |
| `-depth_threshold` | Specifies the path depth above which a timing path would be considered for restructuring. |
| `-tielo_pin` | Tie cell pin that can drive constant zero. The format is `lib/cell/pin`. |
| `-tiehi_pin` | Tie cell pin that can drive constant one. The format is `lib/cell/pin`. |
```

vvbandeira (changes requested):

PR #3721, opened 2023-07-25
ppl: debug visualization for simulated annealing

### LINKED ISSUE #3720: Develop debug visualization for PPL
### Description

The new PPL solver uses Simulated Annealing. A visualization of the state of the iterations of the Simulated Annealing can be helpful to understand how the algorithm is converging to a solution, and if it is converging to a good result.

The visualization of the state can be the "fly lines" between the IO pins and the remaining pins connected to them.

### Suggested Solution

You can add a debug option to the place_pins command where, when it is used, the solution found on an iteration of the annealing run is drawn on the GUI.

For the debug settings, add an option to define the interval of iterations that the solution will be drawn (e.g., draw the solution at every 100 iterations).
Also, add a flag to define if the GUI will pause at each iteration drawn, or if it will keep updating the visualization at each second.

### Additional Context

_No response_

AcKoucher:
Resolves #3720

I created a new Renderer to be used by any of the solver (as OR never uses more than one solver in the flow, it looks like there is no problem).

Currently there are two modes:
Default: One key argument for number of iterations between paintings.
No Pause: Additional flag for continuous printing.

In the future it is possible to think about updating odb data for correctly displaying the pin markers as the iterations go along.


maliberty:
Something is wrong with this PR - there are too many modified files.

AcKoucher:
I'm sorry, I messed up while updating my branch with the changes from master. I'll fix it and reopen.

maliberty:
Please remember to sign your commits with 'git commit -s' to avoid the DCO error.  I can waive it this time.

src/ppl/include/ppl/IOPlacerRenderer.h
```diff
@@ -0,0 +1,64 @@
...
+{
+ public:
+  IOPlacerRenderer();
+  void setPinAssignment(const std::vector<IOPin>& assignment) override;
+  void setSinks(const std::vector<std::vector<InstancePin>>& sinks) override;
+  void setIsNoPauseMode(const bool& isNoPauseMode) override;
+
+  void redrawAndPause() override;
+
+  void drawObjects(gui::Painter& /* painter */) override;
```
maliberty, line 55: No need to comment the argument name here.

src/ppl/src/IOPlacer.cpp
```diff
@@ -62,7 +63,10 @@ IOPlacer::IOPlacer()
   top_grid_ = std::make_unique<TopLayerGrid>();
 }
 
-IOPlacer::~IOPlacer() = default;
+IOPlacer::~IOPlacer()
+{
+  delete annealing_;
```
maliberty, line 68: Prefer a unique_ptr to manual memory management

src/ppl/include/ppl/IOPlacer.h
```diff
@@ -164,6 +165,19 @@ class IOPlacer
...
 
+  void setRenderer(std::unique_ptr<AbstractIOPlacerRenderer> ioplacer_renderer);
+  AbstractIOPlacerRenderer* getRenderer();
+
+  // annealing debug functions
+  void setAnnealingDebugOn();
+  bool isAnnealingDebugOn() const;
+
+  void setAnnealingDebugPaintInterval(const int iters_between_paintings);
+  int getAnnealingDebugPaintInterval() const;
```
eder-matheus, line 176: Is this used somewhere? If not, you can remove it

AcKoucher:
I missed signing while reverting the incorrect merge from master. I didn't know that it was necessary to use git revert -s. I'll be more careful. Should I consider clang-tidy suggestions?

src/ppl/src/SimulatedAnnealing.h
```diff
@@ -51,6 +51,16 @@ class Logger;
 }  // namespace utl
 
 namespace ppl {
+
+struct DebugSettings
+{
+  std::unique_ptr<AbstractIOPlacerRenderer> renderer_;
+  int painting_interval_ = 0;
+  bool isNoPauseMode_ = false;
```
eder-matheus, line 59: Use the same name pattern as the other structs. They use the underline as a separator and don't have the underline at the end.
`painting_interval_` --> `painting_interval`
`isNoPauseMode_` --> `is_no_pause_mode`

src/ppl/src/IOPlacer.tcl
```diff
@@ -336,6 +336,25 @@ proc set_simulated_annealing { args } {
...
+}
+
+proc simulated_annealing_debug { args } {
+  sta::parse_key_args "simulated_annealing_debug" args \
+  keys {-iters_between_paintings} \
+  flags {-no_pause_mode}
+
+  set no_pause_mode [info exists flags(-no_pause_mode)]
+
+  if [info exists keys(-iters_between_paintings)] {
```
maliberty, line 351: This looks to be mandatory.  Is that intentional?  If so, an error should be given if not provided.  If not there should be a default value.

eder-matheus:
> I missed signing while reverting the incorrect merge from master. I didn't know that it was necessary to use git revert -s. I'll be more careful. Should I consider clang-tidy suggestions?

I didn't read all of them, but usually, clang-tidy suggestions are very helpful. I suggest you follow them.

src/ppl/src/IOPlacerRenderer.cpp
```diff
@@ -0,0 +1,90 @@
...
+void IOPlacerRenderer::setPinAssignment(const std::vector<IOPin>& assignment)
+{
+  pin_assignment_ = assignment;
+}
+
+void IOPlacerRenderer::drawObjects(gui::Painter& painter)
+{
+  painter.setPen(gui::Painter::yellow);
+  painter.setBrush(gui::Painter::yellow);
+  painter.setPenWidth(200);
```
maliberty, line 65: Use a cosmetic pen so that it has the same unit width at all zoom scales.

src/ppl/include/ppl/IOPlacerRenderer.h
maliberty, line 1: This should be in src/ as it isn't a public interface.

src/ppl/src/IOPlacer.cpp
```diff
@@ -1931,11 +1976,18 @@ void IOPlacer::runAnnealing(bool random)
   initMirroredPins(true);
   initConstraints(true);
 
-  ppl::SimulatedAnnealing annealing(
+  annealing_ = new ppl::SimulatedAnnealing(
       netlist_io_pins_.get(), core_.get(), slots_, constraints_, logger_, db_);
```
maliberty, line 1980: Why does this need to be a member rather than local?

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -121,6 +123,29 @@ void SimulatedAnnealing::run(float init_temperature,
       }
 
       temperature *= alpha_;
+
+      if (debug_->isOn()) {
+        bool isNoPauseMode = false;
+
+        if (debug_->isNoPauseMode_) {
+          isNoPauseMode = true;
+        }
```
maliberty, line 132: Why is this variable needed - it seems to just be an awkward copy.

src/ppl/include/ppl/IOPlacer.h
```diff
@@ -291,12 +305,21 @@ class IOPlacer
...
+
   float init_temperature_ = 0;
   int max_iterations_ = 0;
   int perturb_per_iter_ = 0;
   float alpha_ = 0;
 
+  // simulated annealing debugger variables
+  bool annealing_debug_mode_ = false;
+  bool no_pause_mode_ = false;
+  int iters_between_paintings_ = 0;
```
maliberty, line 321: These are concerns of the renderer and should be stored there.

src/ppl/include/ppl/IOPlacer.h
```diff
@@ -291,12 +305,21 @@ class IOPlacer
...
   // simulated annealing variables
+  ppl::SimulatedAnnealing* annealing_;
+
   float init_temperature_ = 0;
   int max_iterations_ = 0;
   int perturb_per_iter_ = 0;
   float alpha_ = 0;
 
+  // simulated annealing debugger variables
+  bool annealing_debug_mode_ = false;
```
maliberty, line 319: Does this replicate the null/non-null status of annealing_ ?

src/ppl/src/SimulatedAnnealing.cpp
```diff
@@ -121,6 +123,29 @@ void SimulatedAnnealing::run(float init_temperature,
...
+
+        if (iter == 0 || (iter + 1) % debug_->painting_interval_ == 0) {
```
maliberty, line 134: Too much logic here for what isn't the annealer's concern.  Just call the renderer and let it decide whether to draw or not.

maliberty (changes requested):

maliberty:
Please address the clang-tidy as well.

src/ppl/src/SimulatedAnnealing.h
AcKoucher: I followed the pattern used in FastRoute:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/2313c4895460a815bf1bce666ded4a52e34074f2/src/grt/src/fastroute/include/FastRoute.h#L78C1-L89C2

Should I change it either way?

eder-matheus: This usage in FastRoute needs to be corrected. I will update it when I have time. Follow the pattern we have in IOPlacer for structs:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/ppl/src/Slots.h#L70-L84

src/ppl/src/IOPlacer.tcl
AcKoucher: It's intentional. I see the errors have numbers, (e. g. PPL 17). Which number must I use?

eder-matheus: Look at the file `ppl/messages.txt`. You must choose a number that is not present in this file yet.

src/ppl/src/IOPlacer.cpp
AcKoucher: When the following method is called, SimulatedAnnealing object takes ownership of the renderer.
https://github.com/The-OpenROAD-Project/OpenROAD/blob/814b7ea72d6b7fdd8f1dea2ac4ba149c97ec85b5/src/ppl/src/SimulatedAnnealing.cpp#L167-L171
If the SimulatedAnnealing object was to remain local, then the renderer would run out of scope when `runAnnealing()` ends and for the renderer to work it needs to exist as long as the IOPlacer exists.

maliberty: Why - if there is nothing left to render then it can be deregistered and destroyed.

AcKoucher: I had misunderstood the way the renderer works. Indeed, it will be possible to see the annealing iterations if the object is local.
It only makes sense to make the object a member if it's desired to keep the last state in the layoutviewer after the annealing run is over (this way it will be possible to see the fly wires of the solution and the correct placement of the markers). Should I keep it local?

eder-matheus: IMO, we don't need to keep the final state after completing the annealing. We can just select the pins in their final positions to visualize the fly wires.

maliberty: Does the renderer store enough state to draw even after the annealer is destroyed?  If so then I would keep it as it but not pass ownership to the annealer.

maliberty: The only advantage I can see to keeping afterwards it is if it grows to show more data than the fly lines (eg constraints)

AcKoucher: The renderer stores the state in a vector of IOPins and a vector of vectors of InstancePins.

How about keep it local for now and later, if desired, change to show more data?

maliberty: fine

src/ppl/include/ppl/IOPlacer.h
AcKoucher: As annealing is going to remain local, annealing_debug_mode_ should be set to false after the annealing run?

maliberty (approved):

PR #3741, opened 2023-07-28
grt: fix overlaps with unnecessary vias

eder-matheus:


maliberty:
Does this resolve black_parrot nangate45?

eder-matheus:
> Does this resolve black_parrot nangate45?

Yes, it does. I'm still investigating how to solve the other overlap between vias, but the crash doesn't happen anymore with this fix.

maliberty (approved):

PR #3749, opened 2023-07-31
Fix anchor link

vvbandeira:


vvbandeira:
@luarss I started this PR to test. The second image looks good, but the first is a bit too big.

luarss:
Another try is to set `useMaxWidth` to false [link](https://github.com/mermaid-js/mermaid/issues/1800)

PR #3756, opened 2023-08-01
Warning when pins are not placed on routing grid

osamahammad21:
Instead of checking for min-width pins only, I check on all pins and see if they intersect with any routing track on their layer. If and only if none of the pins of a term intersect with any routing track, I fire a warning that this term has no pins on the routing grid.

### DIFF
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -507,8 +507,6 @@ void TritonRoute::init(Tcl_Interp* tcl_interp,
 
 bool TritonRoute::initGuide()
 {
-  if (DBPROCESSNODE == "GF14_13M_3Mx_2Cx_4Kx_2Hx_2Gx_LB")
-    USENONPREFTRACKS = false;
   io::Parser parser(db_, getDesign(), logger_);
   bool guideOk = parser.readGuide();
   parser.postProcessGuide();
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -819,6 +819,9 @@ void TritonRoute::sendDesignUpdates(const std::string& globals_path)
 
 int TritonRoute::main()
 {
+  if (DBPROCESSNODE == "GF14_13M_3Mx_2Cx_4Kx_2Hx_2Gx_LB") {
+    USENONPREFTRACKS = false;
+  }
   asio::thread_pool pa_pool(1);
   if (!distributed_)
     pa_pool.join();
--- a/src/drt/src/io/io_parser_helper.cpp
+++ b/src/drt/src/io/io_parser_helper.cpp
@@ -649,8 +649,9 @@ void io::Parser::checkPins()
                     |= vertTracks.find(box.yCenter()) != vertTracks.end();
               }
             }
-            if (foundTracks && box.minDXDY() > layer->getMinWidth())
+            if (foundTracks && box.minDXDY() > layer->getMinWidth()) {
               foundCenterTracks = true;
+            }
           } else if (uFig->typeId() == frcPolygon) {
             hasPolys = true;
             auto polygon = static_cast<frPolygon*>(uFig.get());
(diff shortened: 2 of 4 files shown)


maliberty:
If you have a pin where the track is off center it may cause issues for drt - particularly if an on track wire wouldn't be a legal connection.  Does this cover such a case?

osamahammad21:
> If you have a pin where the track is off center it may cause issues for drt - particularly if an on track wire wouldn't be a legal connection. Does this cover such a case?

No It doesn't. This only checks if a track passes through the pin. Should we consider this?

maliberty:
> > If you have a pin where the track is off center it may cause issues for drt - particularly if an on track wire wouldn't be a legal connection. Does this cover such a case?
> 
> No It doesn't. This only checks if a track passes through the pin. Should we consider this?

Yes as it seems to be difficult for drt when the pins are at all off track, even if the track passes through them.   Does that match your experience?

maliberty:
I'm just looking for an early warning - pa will still do its usual analysis.   Even if pa can find an ap it is often the case that dr will struggle.

Rect-only checking is ok.  Polygon's are rare.

osamahammad21:
@maliberty ready for review

src/drt/src/TritonRoute.cpp
```diff
@@ -819,6 +819,9 @@ void TritonRoute::sendDesignUpdates(const std::string& globals_path)
 
 int TritonRoute::main()
 {
+  if (DBPROCESSNODE == "GF14_13M_3Mx_2Cx_4Kx_2Hx_2Gx_LB") {
+    USENONPREFTRACKS = false;
+  }
```
maliberty, line 824: This is in TritonRoute::initGuide already.  Are you needing to move it earlier?  If so the other can be removed.

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -551,8 +552,186 @@ void io::Parser::convertLef58MinCutConstraints()
   }
 }
 
+inline void getTrackLocs(bool isHorzTracks,
```
maliberty, line 555: Is this a copy of FlexDRWorker::getTrackLocs ?  Please have one copy 
No need for inline in a .cpp

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -551,8 +552,186 @@ void io::Parser::convertLef58MinCutConstraints()
...
+                  foundTracks |= !horzTracks.empty();
+                } else {
+                  foundTracks |= !vertTracks.empty();
+                }
+              }
+            }
+          }
+        }
+      }
+      if (iTerm->hasNet() && !iTerm->getNet()->isSpecial()) {
```
maliberty, line 715: Why not skip such iTerm at the start ?

src/drt/src/io/io_parser_helper.cpp
osamahammad21: not a copy

maliberty: Perhaps not a literal copy but it seems to do the same thing.  Is there some difference I'm missing?

osamahammad21: When I looked at FlexDRWorker::getTrackLocs, I didn't understand what it was trying to do. First there is this part:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/6a66a6b07a3231b9400ed35d0f4e71eace5f5f4f/src/drt/src/dr/FlexDR_init.cpp#L1394-L1401
trackLocs is a set.  At the first line, you already add the trackLoc to the set. The if condition after that is completely useless. And this is the only part where the bool isHorzTracks is used, so it is completely useless as well. 
I didn't focus on this when I started writing the check, so I just changed the implementation, but when you mentioned it, I came up with the mentioned findings. However, I am not sure which part is incorrect, inserting the trackLoc regardless of its direction or validating on isHorzTracks bool.
I am gonna start a GH issue to look at this when I have time.

maliberty: Ok I'll approve here and you can figure this out when address your PR.  My goal is one method that does this correctly and is shared.

maliberty (approved):

Issue #3752, opened 2023-08-01
ZASSERT error in check_antenna


openroadie:
### Describe the bug

Running the debug build, OR asserts in antenna checking in this code:

void dbWireEncoder::newPathVirtualWire(int jct_id,
                                       dbTechLayer* layer,
                                       dbWireType type)
{
  **ZASSERT((jct_id >= 0) && (jct_id < _idx));**

### Expected Behavior

No assertion error. The latest master should show this issue. 

### Environment

```shell
[WARNING] Your current OpenROAD version is outdated.
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Linux 5.8.7-050807-lowlatency
os: Ubuntu 20.04.6 LTS (Focal Fossa)
cmake version 3.24.2
CMake Warning at CMakeLists.txt:98 (message):
  OpenROAD git describe failed, using sha1 instead


-- The CXX compiler identification is GNU 9.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: ab7625fe5843c6579d28068b2a7e4dfb38f9dcb5
-- System name: Linux
-- Compiler: GNU 9.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 9.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.8 (found version "3.8.10") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
[...]

PR #3757, opened 2023-08-01
grt: check jct_id before creating new path or virtual path

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3752

### DIFF
--- a/src/grt/src/RepairAntennas.cpp
+++ b/src/grt/src/RepairAntennas.cpp
@@ -249,12 +249,7 @@ void RepairAntennas::addWireTerms(Net* net,
       }
 
       if (conn_layer >= grouter_->getMinRoutingLayer()) {
-        if (jct_id != -1) {
-          wire_encoder.newPathVirtualWire(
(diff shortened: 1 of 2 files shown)


src/grt/src/RepairAntennas.cpp
```diff
@@ -272,16 +272,24 @@ void RepairAntennas::addWireTerms(Net* net,
             = layer1->getDirection() == odb::dbTechLayerDir::VERTICAL ? layer1
                                                                       : layer2;
         // create horizontal wire to connect to the pin
-        wire_encoder.newPathVirtualWire(
-            jct_id, h_layer, odb::dbWireType::ROUTED);
+        if (jct_id != -1) {
```
openroadie, line 275: Not that important but it might be nice to factor out these 4 lines and the next 4 lines into a small function if that makes sense. The code is virtually identical except for the h_layer/v_layer  difference.

eder-matheus: It makes sense. I've pushed this refactor.

maliberty (approved):

PR #3764, opened 2023-08-02
Use relative paths for submodules so they use the same protocol as the top one.

titan73:
Hi,
My proxy only supports git over ssh so I have to clone the repo with ssh but cloning submodules fails are they are explicitely using https. This MR makes paths relative so the same protocol is used.
Chris


gadfort:
This makes forks have to also fork those projects too. Would it be easier for the user to use:
```
git config --global --add url."git@github.com".insteadOf "https://github.com/"
```

vvbandeira:
@gadfort I think that would be true if the change was to `../OpenSTA`, but using two levels down `../../`+ explicitly setting `The-OpenROAD-Project` make it resolve correctly. I just run:
```
$ git clone --recursive  https://github.com/titan73/OpenROAD OpenROAD-titan
$ cat OpenROAD-titan/.git/config
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true
[submodule]
        active = .
[remote "origin"]
        url = https://github.com/titan73/OpenROAD
        fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
        remote = origin
        merge = refs/heads/master
        rebase = true
[submodule "module/OpenSTA"]
        url = https://github.com/The-OpenROAD-Project/OpenSTA.git
[submodule "src/abc"]
        url = https://github.com/The-OpenROAD-Project/abc.git
```

vvbandeira:
> @vvbandeira any concerns? ORFS is similar.

ORFS could also use relative paths except for yosys and lsoracle, but not OpenRAOD due to the private org and how our CI is currently configured.

vvbandeira:
> [submodule "tools/OpenROAD"]
> path = tools/OpenROAD
> url = ../OpenROAD.git

"One level up", yes.

To be more clear, changing `../OpenROAD` to `../../The-OpenROAD-Project/OpenROAD` in ORFS would break our CI.

titan73:
> > [submodule "tools/OpenROAD"]
> > path = tools/OpenROAD
> > url = ../OpenROAD.git
> 
> "One level up", yes.
> 
> To be more clear, changing `../OpenROAD` to `../../The-OpenROAD-Project/OpenROAD` in ORFS would break our CI.

Ah ok. :)

titan73:
Hi. Is there any news about this PR? Should I do something more?

vvbandeira (approved):

PR #3778, opened 2023-08-03
tap: add check if unable to complete boundaries

### LINKED ISSUE #3760: Overlapping endcap & corner cells
### Describe the bug

In gf180/aes tapcell is now placing overlapping cells:
image

### Expected Behavior

No overlaps

### Environment

```shell
ORFS 392edae54a52586959855228f9fbfee50ebb56e4
OR d27c60005b12791b2a60547689bfe9e7c72b9afb
```

### To Reproduce

DESIGN_CONFIG=./designs/gf180/aes/config.mk make floorplan

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

gadfort:
Adds:
- Check for endcaps on horizontal rows to ensure they can be inserted without gaps.
Closes #3760 (the errors setup is incorrect and therefore you will have CI issues unless you change the tapcell command)


maliberty:
on gf180/aes with the current tapcell.tcl I now get:
```[ERROR TAP-0020] Unable to fill Bottom boundary in ROW_0 from 1473.9200um to 1474.4800um```
which doesn't seem right

gadfort:
@maliberty the cells are are 2 sites wide, but the rows are an odd number of sites, so not sure how you want the tapcells to be inserted?

gadfort:
The previous version just leaves a gap:
image

maliberty:
I think we can go ahead with this change but it seems we have a new constraint to be aware of.  Do you think we should try to handle this in ifp or tap?

maliberty (approved):

gadfort:
I think it might be something that cut_rows could handle by telling it there are no 1x or odd width end caps so it makes the rows even

maliberty:
> I think it might be something that cut_rows could handle by telling it there are no 1x or odd width end caps so it makes the rows even

Do you think that is better than telling ifp and having it not even generate such rows?

gadfort:
Both should probably be made aware

maliberty:
I'm guessing this only comes up in gf180 because it isn't intending for us to use this tapcell on the top & bottom.  In the other PDKs I've looked at they have a 1X tapcell available.

gadfort:
Yeah there usually is an odd width one 1x or 3x to avoid this.

Issue #3591, opened 2023-07-08
Crash in pin constraints


oharboe:
### Describe the bug

```
Using 2 tracks default min distance between IO pins.
[INFO PPL-0007] Random pin placement.
[WARNING PPL-0096] Pin group of size 8 does not fit constraint region. Adding to fallback mode.

[WARNING PPL-0096] Pin group of size 8 does not fit constraint region. Adding to fallback mode.
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = int; _Alloc = std::allocator<int>; std::vector<_Tp, _Alloc>::reference = int&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
 0# 0x00005580A3390F2B in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 1# 0x00007F22CF642520 in /lib/x86_64-linux-gnu/libc.so.6
 2# pthread_kill in /lib/x86_64-linux-gnu/libc.so.6
 3# raise in /lib/x86_64-linux-gnu/libc.so.6
 4# abort in /lib/x86_64-linux-gnu/libc.so.6
 5# 0x00005580A338F198 in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 6# ppl::IOPlacer::randomPlacement(std::vector<int, std::allocator<int> >, std::vector<int, std::allocator<int> >, bool, bool) in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 7# ppl::IOPlacer::randomPlacement() in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 8# ppl::IOPlacer::run(bool) in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 9# 0x00005580A34E4D47 in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
10# TclNRRunCallbacks in /lib/x86_64-linux-gnu/libtcl8.6.so
11# 0x00007F22D3B48028 in /lib/x86_64-linux-gnu/libtcl8.6.so
[...]

PR #3784, opened 2023-08-03
ppl: fix crash in random pin placement

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3591


src/ppl/src/IOPlacer.cpp
```diff
@@ -664,6 +675,11 @@ double IOPlacer::dbuToMicrons(int64_t dbu)
   return (double) dbu / (getBlock()->getDbUnitsPerMicron());
 }
 
+int64_t IOPlacer::micronsToDbu(double microns)
```
maliberty, line 678: int?

maliberty (approved):

PR #3786, opened 2023-08-04
messages: print out next free message id on duplicate

oharboe:


maliberty (approved):

maliberty:
Thanks, this will be useful

PR #3790, opened 2023-08-04
gui: pin selecting and drawing 

gadfort:
Adds:
- pin selection control for instances

Changes:
- only allow selection of pins when visible and when layers are visible for that pin
- draw the pin label in the first mpin box that it will fit it otherwise pins are left floating off the metal shape.

### DIFF
--- a/src/gui/src/renderThread.cpp
+++ b/src/gui/src/renderThread.cpp
@@ -557,7 +557,6 @@ void RenderThread::drawInstanceNames(QPainter* painter,
     return;
   }
 
-  const QTransform initial_xfm = painter->transform();
   const QColor text_color = viewer_->options_->instanceNameColor();
   const QFont initial_font = painter->font();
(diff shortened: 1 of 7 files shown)


src/gui/src/renderThread.cpp
```diff
@@ -603,31 +601,51 @@ void RenderThread::drawITermLabels(QPainter* painter,
...
-      painter->setTransform(initial_xfm);
+      bool drawn = false;
+      for (auto* mpin : inst_iterm->getMTerm()->getMPins()) {
+        for (auto* geom : mpin->getGeometry()) {
+          const auto layer = geom->getTechLayer();
+          if (layer == nullptr) {
+            continue;
+          }
+          if (viewer_->options_->isVisible(layer)
+              && viewer_->options_->isSelectable(layer)) {
```
maliberty, line 615: Why is selectability controlling the drawing of the label?

maliberty (approved):

PR #3792, opened 2023-08-04
Recover power

openroadie:
Clean up the interface for "repair_timing -recover_power". Add a percent argument to the option which works the same as TNS_END_PERCENT. The corresponding ORFS PR is here:


QuantamHD:
@openroadie General comment, I would like to see this step integrated into the standard repair_timing flow. We should just recover power where there's slack to do so rather than relying on a flow level setting.

I did try the recover_power flow on one of my designs with negative slack, and it went from -400ps to -550ps WNS. Is that expected, in my mind we would only resize instances if there was no effect on timing or if we had positive slack.

openroadie:
@QuantamHD You are absolutely right. In theory we should be able do just recover slack where it is available and I will work to figure out how to achieve that goal in a more automated fashion.

The reason this is a flow level setting right now is twofold:

- Too much perturbation tends to degrade TNS/WNS (even though the perturbation itself was deemed to be neutral to timing violations). We don't quite understand the root cause for. This lines up with what you are reporting as your testing results.  
- I think it would be a better idea to recover slack with global routing based parasitics (and this is why there is a flow setting with a value which is similar to the TNS value we have for repair timing).

maliberty:
https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1233 ran the public tests and required some updates.  Is there something similar for the private/secure CI?

src/rsz/src/Resizer.tcl
```diff
@@ -389,7 +389,7 @@ proc repair_tie_fanout { args } {
 
 # -max_passes is for developer debugging so intentionally not documented
 # in define_cmd_args
-sta::define_cmd_args "repair_timing" {[-setup] [-hold] [-recover_power]\
+sta::define_cmd_args "repair_timing" {[-setup] [-hold] [-recover_power percent_of_paths_with_slack]\
```
maliberty, line 392: Please document the new option in the README.md

src/rsz/src/RecoverPower.cc
```diff
@@ -140,7 +140,12 @@ RecoverPower::recoverPower()
              int(ends_with_slack.size() / double(endpoints->size()) * 100));
 
   int end_index = 0;
-  int max_end_count = ends_with_slack.size()/5; // 20%
+  int max_end_count = ends_with_slack.size()*recover_power_percent;
+
+  // As long as we are herem fix at least one path
```
maliberty, line 145: type: herem

maliberty:
Something is odd with the formatting:
image

perhaps a tab?

openroadie:
> Something is odd with the formatting: image
> 
> perhaps a tab?

Fixed. It was a tab.

PR #3799, opened 2023-08-06
[Docs] Addtool patch

luarss:
@vvbandeira @maliberty 

This addresses the issue in https://github.com/The-OpenROAD-Project/OpenROAD/discussions/3735#discussioncomment-6569842

How it works: Instead of manually maintaining the `AddTool.patch` file I separated into static file `AddToolFiles.patch` and dynamic file `AddTool.patch`, the latter will be programmatically generated by running `python getPatch.py`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -160,3 +160,7 @@ def setup(app):
     # for populating OR Messages page.
     command = "python getMessages.py"
     _ = os.popen(command).read()
+
+    # for populating AddTool.patch file
+    command = "python getPatch.py"
+    _ = os.popen(command).read()
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -160,7 +160,3 @@ def setup(app):
     # for populating OR Messages page.
     command = "python getMessages.py"
     _ = os.popen(command).read()
-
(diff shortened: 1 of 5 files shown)


vvbandeira (commented):
The idea of making this process more programmatic and future-proof is great, but this approach has an unnecessary level, in my opinion. If we generate the patch file on the fly, doesn't it make more sense to directly apply it or modify the files and let the user review and make the final commit? Example: running `./add_tool.py --name NewShiningTool --abbrev nst ...` would create the folder in `src/nst` add to CMake, create a README, etc.

Note: This is a rare occurrence, so it should be a stretch goal for the GSoC project with very low priority.

luarss:
But in the meantime, would we need to update the patch file?

vvbandeira:
> But in the meantime, would we need to update the patch file?

Yes, please update it once and also include the correct command to apply the patch.

luarss:
@vvbandeira Removed the references to python script.

luarss:
@vvbandeira Ready for review

vvbandeira (approved):

PR #3802, opened 2023-08-06
Creating rows for hybrid-cells in IFP

mina1460:
This draft PR is 90% completed.
It needs a couple of test cases, and some manipulations in the dbSite to allow for seeing ROWPATTERNS and using them.


src/ifp/include/ifp/InitFloorplan.hh
```diff
@@ -111,6 +112,13 @@ class InitFloorplan
                int core_uy,
                int factor,
                int row_index);
+  int makeHybridRows(const std::vector<dbSite*>& hybrid_sites,
```
QuantamHD, line 115: I think we should name this more generically, seems like we should tie it more to the concept of "patterns" rather than hybrid rows. 

"makeRowPattern"
"generateContiguousRowPattern"

src/ifp/src/InitFloorplan.cc
```diff
@@ -482,6 +490,114 @@ int InitFloorplan::makeRows(dbSite* site,
...
+    int y = core_l.y(), pattern_iterator = 0;
+
+    while (y < core_height) {
+      dbOrientType orient = (pattern_iterator % 2 == 0)
+                                ? dbOrientType::R0   // N
+                                : dbOrientType::MX;  // FS
+      string row_name = stdstrPrint("ROW_%d", row_index + row);
+      dbSite* site_it;
+      if (pattern.size() == 1) {
+        site_it = pattern[pattern_iterator % pattern.size()];
```
maliberty, line 559: why mod if size==1?

maliberty:
Please remember to sign your commits.  I've set DCO to pass for this PR.

src/odb/include/odb/db.h
```diff
@@ -5330,6 +5330,27 @@ class dbSite : public dbObject
   ///
   bool getSymmetryR90();
 
+  ///
+  /// set the row pattern of this site
+  ///
+  void setRowPattern(
+      std::vector<std::pair<std::string, std::string>>& row_pattern);
```
maliberty, line 5337: This should mirror getSitePattern and take a "const std::vector<std::pair<dbSite*, dbOrientType>>&" argument.  The conversion from strings should happen in lefin.

src/odb/src/db/dbSite.cpp
```diff
@@ -230,6 +232,62 @@ void dbSite::setClass(dbSiteClass type)
...
+         {"S", dbOrientType::R180},
+         {"E", dbOrientType::R90},
+         {"FN", dbOrientType::MYR90},
+         {"FW", dbOrientType::MY},
+         {"FS", dbOrientType::MXR90},
+         {"FE", dbOrientType::MX}};
+
+  for (auto& row : row_pattern) {
+    dbOrientType orient = dbOrientType::R0;
+    auto _site = (_dbSite*) getLib()->findSite(row.first.c_str());
```
maliberty, line 253: You should check the site is found and error if it is not found.

src/odb/src/db/dbSite.cpp
```diff
@@ -230,6 +232,62 @@ void dbSite::setClass(dbSiteClass type)
...
+    _site->_flags._is_hybrid = 1;
+    auto it = orientationMap.find(row.second);
+    if (it != orientationMap.end()) {
```
maliberty, line 256: If not found it should be an error

src/odb/src/db/dbSite.cpp
```diff
@@ -230,6 +232,62 @@ void dbSite::setClass(dbSiteClass type)
...
+  }
+}
+
+bool dbSite::hasRowPattern()
+{
+  _dbSite* site = (_dbSite*) this;
+  return !site->_row_patterns.empty();
+}
+
+bool dbSite::isHybrid()
```
maliberty, line 269: What is the difference between isHybrid & hasRowPattern?  Do we need both (same about _row_pattern & _is_hybird)?

mina1460: Yeah there is a big difference.
in the case of having a hybridCell AB that is made of A&B. All 3 are hybrid cells, but only cell AB would have a row pattern set of <A,orientation> and <B,orientation>

src/odb/src/db/dbSite.h
```diff
@@ -64,6 +65,7 @@ class _dbSite : public _dbObject
   uint _height;
   uint _width;
   dbId<_dbSite> _next_entry;
+  std::vector<std::pair<std::string, dbOrientType::Value>> _row_patterns;
```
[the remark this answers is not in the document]
mina1460: how can I do that?

src/odb/src/db/dbSite.cpp
mina1460: there is a possibility that it is not constructed yet.
It depends on the order of parsing and finding rows and row patterns, so it might not be an issue here. It will be flagged as an error in other places in the code, but here it is not an issue (yet).

src/odb/src/db/dbSite.cpp
mina1460: The same comment below applies here. 
please let me know if you think otherwise, but this was intentional from me

src/odb/src/db/dbSite.h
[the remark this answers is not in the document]
maliberty: Look at how the other fields are handled in dbSite (operator<<, operator>>, operator==, differences, and out).

For backward compatability see db_schema_block_component_mask_shift as an example.

mina1460:
@maliberty 
I believe most, if not all, of the comments are now resolved. 
There is however one remaining thing here. Since you suggested that the lefin should handle the creation of the rowpattern vector, I had to make it a member of dbSite and not _dbSite (the inner cast). 

This means this info is no longer available during serialization. I also couldn't see how to handle the schema pump to maintain backward compatibility.

maliberty:
> Since you suggested that the lefin should handle the creation of the rowpattern vector, I had to make it a member of dbSite and not _dbSite (the inner cast).

I don't understand this comment.  Why is _dbSite needed in lefin?  lefin should call the public methods

maliberty:
replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/4167

PR #3804, opened 2023-08-06
psm: code cleanup

gadfort:
No functional changes, just cleaning up the code in IRSolver to use fewer strings and instead pass in odb objects and make function names consistent with other tools.
Updates a few tests to account for long standing "diff" in ok files.


src/psm/src/pdnsim.tcl
```diff
@@ -87,49 +81,50 @@ proc analyze_power_grid { args } {
...
 }
 
 sta::define_cmd_args "check_power_grid" {
+  [-error_file error_file]
   [-net power_net]}
 
 proc check_power_grid { args } {
   sta::parse_key_args "check_power_grid" args \
-    keys {-net} flags {}
+    keys {-net -error_file} flags {}
```
maliberty, line 113: README update

maliberty:
Please look at clang-tidy, especially bugprone-unchecked-optional-access

maliberty:
Thanks for the big cleanup!

FYI @ahmadelrouby will start looking at the performance issues with matrix setup

gadfort:
@maliberty thats great to hear. I have one more thing after this to address the handling of macro pins, but that should not be a big thing.

maliberty (approved):

PR #3807, opened 2023-08-07
ppl: fix infinite loop in simulated annealing

oharboe:
@eder-matheus I'm shooting from the hip here, but there does seem to be an infinite loop...


eder-matheus:
> @eder-matheus I'm shooting from the hip here, but there does seem to be an infinite loop...

Yes, we have an infinite loop, and I'm looking into it now. I've tested your suggestion with asap7/mock-array and still see the infinite loop. I'll let you know when I have a fix for it. I believe this is the only issue blocking us from turning the annealing solver as the default in ORFS.

eder-matheus (approved):

maliberty (approved):

PR #3808, opened 2023-08-07
Revert "grt: generalize layer name regex in layer range parsing"

vvbandeira:
This reverts commit 7f449467ade1eb06af581e0ab683cc07319a5a60, which breaks support for gf12.


maliberty:
Replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/3809

PR #3814, opened 2023-08-08
gui: render instance text centered, easier to read

oharboe:
Before:

image

image

After:

image

image

### DIFF
--- a/src/gui/src/renderThread.cpp
+++ b/src/gui/src/renderThread.cpp
@@ -573,7 +573,9 @@ void RenderThread::drawInstanceNames(QPainter* painter,
 
     Rect instance_box = inst->getBBox()->getBox();
     QString name = inst->getName().c_str();
-    drawTextInBBox(text_color, text_font, instance_box, name, painter);
+    auto master = inst->getMaster();
+    auto center = master->isBlock() || master->isPad();
+    drawTextInBBox(text_color, text_font, instance_box, name, painter, center);
   }
   painter->setFont(initial_font);
 }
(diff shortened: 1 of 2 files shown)


gadfort:
@oharboe can you provide before and after images of this change?

gadfort:
How does this look on standard cells?

gadfort:
@oharboe I think for the macros it helps, but the standard cells I don't know if that is an improvement, it's actually harder to determine the boundaries of the cells (since it used to be where the text starts).

oharboe:
> @oharboe I think for the macros it helps, but the standard cells I don't know if that is an improvement, it's actually harder to determine the boundaries of the cells (since it used to be where the text starts).

Is there a way, from within that fn to determine when this should be done and not done?

oharboe:
> You should be able to check the type of the cell (`inst->getMaster()->isMacro()`) or something like that. I suspect pads and cover cells also benefit from something like this.

Fixed.

maliberty:
In the after picture, one of the srams has its label horizontally while the other three are vertical.  What explains that?

src/gui/src/renderThread.cpp
```diff
@@ -697,13 +700,23 @@ bool RenderThread::drawTextInBBox(const QColor& text_color,
   painter->scale(scale_adjust, -scale_adjust);
   if (do_rotate) {
     text_bounding_box = font_metrics.boundingRect(name);
+    auto xOffset = (bbox_in_px.height() - text_bounding_box.width()) / 2;
+    auto yOffset = (bbox_in_px.width() - text_bounding_box.height()) / 2;
```
maliberty, line 704: Move inside 'if (center)' since they are only needed at that scope.  Also const.  Same in the else block.

oharboe:
@maliberty horizontal vs. vertical text direction is chosen based on the number of pixels wide the text is. the macros have different labels, so if you zoom just right, you can illustrate horizontal and vertical text in the same screenshot

PR #3815, opened 2023-08-08
gui: make same_clocktree_image a regular tcl command

gadfort:
Adds:
- `save_clocktree_image` with options to save with width/height and specify timing corner

This allows for better control of the resulting image since before it was just using the widget sizes which could result in poor quality images.

### DIFF
--- a/src/gui/src/clockWidget.h
+++ b/src/gui/src/clockWidget.h
@@ -47,6 +47,7 @@
 #include <QSpinBox>
 #include <QTabWidget>
 #include <QToolTip>
+#include <optional>
 #include <variant>
 
 #include "gui/gui.h"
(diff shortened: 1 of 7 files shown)


src/gui/src/clockWidget.h
```diff
@@ -466,7 +466,11 @@ class ClockWidget : public QDockWidget, sta::dbNetworkObserver
   void setLogger(utl::Logger* logger);
   void setSTA(sta::dbSta* sta);
 
-  void saveImage(const std::string& clock_name, const std::string& path);
+  void saveImage(const std::string& clock_name,
+                 const std::string& path,
+                 const std::string& corner,
+                 int width_px,
+                 int height_px);
```
maliberty, line 473: std::optional?

gadfort: sure, I assume we don't want this at the gui.h too because the swig wrapper wouldn't be right for python? If I recall correctly what SWIG doesn't really know what to do with optional

maliberty: Yeah swig doesn't like optional

maliberty (approved):

maliberty:
Do you think we should make any change to ORFS's save_images?

gadfort:
> Do you think we should make any change to ORFS's save_images?

it shouldn't be required, the signature for `gui::save_clocktree_image` will still work as it is used now.
But it can't hurt to cleanup the code to avoid using the old command

PR #3816, opened 2023-08-08
ppl: give more flexibility when moving pin groups

eder-matheus:


eder-matheus:
@oharboe Could you try it with the design you have with the infinite loop issue? Btw, you're facing it during random or non-random IO placement?

oharboe:
> @oharboe Could you try it with the design you have with the infinite loop issue? Btw, you're facing it during random or non-random IO placement?

Looks stuck:

```
openroad -no_init /home/oyvind/OpenROAD-flow-scripts//flow//scripts/io_placement.tcl
[deleted]
Not all elements are present, skipping io constraint: resetOut
Not all elements are present, skipping io constraint: ceEnableOut
[INFO PPL-0048] Restrict pins [clock] to region 0.0u-120.0u at the top edge.
Found 0 macro blocks.
[INFO PPL-0001] Number of slots          6194
[INFO PPL-0002] Number of I/O            686
[INFO PPL-0003] Number of I/O w/sink     686
[INFO PPL-0004] Number of I/O w/o sink   686
^C
```

eder-matheus:
> > @oharboe Could you try it with the design you have with the infinite loop issue? Btw, you're facing it during random or non-random IO placement?
> 
> Looks stuck:
> 
> ```
> openroad -no_init /home/oyvind/OpenROAD-flow-scripts//flow//scripts/io_placement.tcl
> [deleted]
> Not all elements are present, skipping io constraint: resetOut
> Not all elements are present, skipping io constraint: ceEnableOut
> [INFO PPL-0048] Restrict pins [clock] to region 0.0u-120.0u at the top edge.
> Found 0 macro blocks.
> Using 2 tracks default min distance between IO pins.
> [INFO PPL-0001] Number of slots          6194
> [INFO PPL-0002] Number of I/O            686
> [INFO PPL-0003] Number of I/O w/sink     686
> [INFO PPL-0004] Number of I/O w/o sink   686
> ^C
> ```

For how much time did you let it run? The asap7/mock-array design took about 3 minutes to complete.

oharboe:
@eder-matheus I made this change to get it unstuck yesterday.... Is this a real bug? https://github.com/The-OpenROAD-Project/OpenROAD/pull/3807

eder-matheus:
> @eder-matheus I made this change to get it unstuck yesterday.... Is this a real bug? #3807

This should only affect random mode.

oharboe:
@eder-matheus 3 minutes is a long time... I imagine it could be substantially longer for other designs. Would it be possible to store the result in source control? I can imagine that one could make do with the same pin-layout for a very long time, so no need to re-run it all time...

oharboe:
> > @eder-matheus I made this change to get it unstuck yesterday.... Is this a real bug? #3807
> 
> This should only affect random mode.

I see, but should I restore the PR? Is it a real fix?

eder-matheus:
> 

Yes, I think you can restore the PR. It changes a bit the logic of the function, but it looks correct.

maliberty:
The test case you sent didn't fail for me with the old placer.  I haven't tried the new one.

oharboe:
> The test case you sent didn't fail for me with the old placer. I haven't tried the new one.

I tried with e2e062580a3994cdea22a618a260e393c83c27a3, infinite loop...

eder-matheus:
> @eder-matheus 3 minutes is a long time... I imagine it could be substantially longer for other designs. Would it be possible to store the result in source control? I can imagine that one could make do with the same pin-layout for a very long time, so no need to re-run it all time...

@oharboe Yes, I know. 3 minutes is far from what I expect with this new solver, and reducing this runtime is my goal for this week.

I don't understand what you mean about storing the result in source control. Could you clarify how it would work?

maliberty (approved):

oharboe:
> I don't understand what you mean about storing the result in source control. Could you clarify how it would work?

Let's say there was a `-keep foo.tcl` option. If foo.tcl exists, then no placement is run, otherwise the placement is run and a foo.tcl is written out. foo.tcl would contain pin commands for pin placement that was found. The user could then have foo.tcl under source control. To generate a new one, delete foo.tcl, run the io placement and commit.

maliberty:
I think it is better to just address the runtime first and only deal with that if we can't make it acceptably fast.

oharboe:
> I think it is better to just address the runtime first and only deal with that if we can't make it acceptably fast.

Agreed.

PR #3818, opened 2023-08-08
gui: add context menu in Timing Report for "writing" report_checks in tcl input

### LINKED ISSUE #3739: Right click on timing path and inject OpenSTA command to report that path
### Description

Today I have to manually type up the report_checks command for a timing path if I want to drill down for more information or create some variant.

### Suggested Solution

It would be nice to be able to right click on a path and have an OpenSTA command injected into the console that the user can then refine.

### Additional Context

_No response_

AcKoucher:
Resolves #3739


maliberty:
The merge conflict is because I've removed all usage of the old style SIGNAL/SLOT statements.  Please make a similar change in this PR and resolve the conflict.  (see https://doc.qt.io/qt-5/signalsandslots-syntaxes.html)

src/gui/src/staGui.cpp
```diff
@@ -1289,4 +1289,21 @@ const std::vector<std::set<const sta::Pin*>> TimingControlsDialog::getThruPins()
...
+/////////
+TimingPathsTableView::TimingPathsTableView(QWidget* parent) : QTableView(parent)
+{
+}
+
+void TimingPathsTableView::mousePressEvent(QMouseEvent* event)
+{
+  if (event->button() == Qt::RightButton) {
+    QModelIndex index = indexAt(event->pos());
+    if (index.isValid() && selectionModel()->isSelected(index)) {
```
maliberty, line 1302: Its non-obvious that you have to select the row before you can right click it.

maliberty:
I feel that the widgets are getting too directly connected.  I would have the timing widget emit a setCommand signal and the script widget have a setCommand slot.  The mainWindow can connect these up.  It will leave the possibility for other widgets to use that slot in the future and reduce coupling.

src/gui/src/staGui.cpp
maliberty: @oharboe user opinion?

maliberty:
Strictly speaking the command doesn't ensure you get exactly the same path.  You can have multiple paths between the same start/end points.  

To be thorough you would have to specify -through on every intermediate point but I suspect that would be unwieldy.  I think use -(rise|fall)_from and -(rise|fall)_to, as matches the path, would be a help.

src/gui/src/staGui.cpp
oharboe: I would go with what EDA users are accustomed to...



EDA Tools Right-Clicking Behavior

GPT-4

User
is it normally necessary to select before right clicking to get a menu in EDA programs.

list example of.what well known EDA tools do

ChatGPT
In Electronic Design Automation (EDA) programs, the right-click context menu behavior may vary depending on the specific tool and what you are trying to do. Here's a general overview of how right-clicking might work in some well-known EDA tools:

Cadence Virtuoso:

Typically, you can right-click on a specific object without selecting it first to bring up a context menu related to that object.
Mentor Graphics PADS:

Usually, you might need to select a particular element to interact with it via the right-click context menu.
Altium Designer:

Right-clicking without selection might still present a context menu, and clicking on a specific object will usually provide more context-sensitive options.
EAGLE by Autodesk:

You can usually right-click without selecting to get a general menu, but right-clicking on a specific object can give you more detailed options.
KiCad:

In this tool, you can often right-click without selecting an item first to access various context-sensitive options.
Synopsys VCS:

Right-clicking might give access to general options, but selecting an item first may provide more specific choices depending on the context.
ANSYS Electronics Desktop:

Right-clicking without selection might provide a generic menu, but more specific options might be available when an object is selected first.
So, while it's not always necessary to select an object before right-clicking in EDA programs, it might provide more specific or relevant options depending on what you are attempting to do.

oharboe:
> Strictly speaking the command doesn't ensure you get exactly the same path. You can have multiple paths between the same start/end points.
> 
> To be thorough you would have to specify -through on every intermediate point but I suspect that would be unwieldy. I think use -(rise|fall)_from and -(rise|fall)_to, as matches the path, would be a help.

I think those are different alternatives: something like this that I can modify and another option, exactly what is shown...

src/gui/src/staGui.cpp
AcKoucher: In fact, I just implemented that way, because I thought it looked a bit odd to have a selected row, but being able to write the command from a different row which is not selected. Still, it's not clear to me, should I keep mandatory to select the row?

maliberty: I think this would work better as a right click menu with choices (eg exact path, start-end path).  When you right click make the row under the cursor selected at the same time.

AcKoucher: How about, for this PR, the menu have 2 options with the following names?

`Closest Match` (As apparently there's no way to get the exact path.)
`From Start to End`

maliberty: Sure.

maliberty:
For the exact choice you should include -through_(rise|fall) for the intermediate points.

AcKoucher:
Done. I also removed the methods I added in staGuiInterface's Timing Path class. As it was necessary to iterate the selected path node list, there was no need to use them.

maliberty (approved):

PR #3821, opened 2023-08-08
gui: add new attribute filtering for selection

### LINKED ISSUE #3768: Add command to highlight only the IO nets
### Description

When debugging results from the pin placer, it is helpful to see the fly wires of the IO nets after the pin placement. I have to manually select all the IO pins to see these fly wires.

Since it would be used mainly during developer debugging, a command to highlight only the IO nets should be enough to simplify this process.

### Suggested Solution

This command can live inside PPL using the existing IOPlacerRendered. I suggest having something like `highlight_io_nets <net_names>`. This way, we can draw the fly wires for a subset of nets and not only for all nets. The `*` symbol can be used to draw the fly wires for all nets.

Some examples:
`highlight_io_nets {net1}` - Draw the fly wires of net1
`highlight_io_nets {net1 net2 net3}` - Draw the fly wires of the nets net1, net2 and net3
`highlight_io_nets {net*}` - Draw the fly wires of all nets with the prefix `net`.
`highlight_io_nets {*}` - Draw the fly wires of all IO nets.

### Additional Context

_No response_

AcKoucher:
Create a GUI selection filter based on the properties of the selected object.

The filter covers also the cases in which the property holds a SelectionSet or a PropertyList.

I created a special case to resolve #3768.


oharboe:
@AcKoucher Would it be an idea to inject the command that does the selection into the Tcl console so the user can tinker with more advanced options? That way we wouldn't have to clutter the GUI for more obtuse options and there is something "market" the more advanced feature. The user would start with a simple selection, do up-arrow in the console and then add some more obscure options for which there are no GUI buttons. Just a tought...

AcKoucher:
@oharboe I think that injecting the selection command in the Tcl console can make sense - especially if we're trying to keep the same line of #3739 - if triggered by right-clicking the selectable cell in the inspector, although the layout viewer's selection ability is already very powerful by itself and the selection commands are shorter when compared to sta long commands. @maliberty what do you think?

oharboe:
One thing I like about injecting GUI operations as Tcl commands into the command line is that it will be easier to write github issues and provide instructions :-) "Paste X into the command console". Though... it is tougher to go from command line to illustrate where in the GUI you can do the same....

src/gui/src/gui.tcl
```diff
@@ -170,6 +171,24 @@ proc select { args } {
...
+  set value ""
+  if { [info exists keys(-filter)] } {
+    if { $name != "" } {
+      utl::error GUI 59 "-filter not available when using -name."
+    } else {
+      set filter $keys(-filter)
+      set filter [split $filter "="]
+      set length [llength $filter]
+      if { $length != 2 } {
+        utl::error GUI 56 "Invalid syntax for -filter. Use -filter attribute=value."
```
maliberty, line 184: It think we could eventually allow a list of key=value filters that would be and'ed together.  Let's revisit that once we get the single case working.

src/gui/src/gui.tcl
```diff
@@ -170,6 +171,24 @@ proc select { args } {
     set name $keys(-name)
   }
   
+  set attribute ""
+  set value ""
+  if { [info exists keys(-filter)] } {
+    if { $name != "" } {
+      utl::error GUI 59 "-filter not available when using -name."
```
maliberty, line 178: Why not both?

src/gui/src/gui.cpp
```diff
@@ -476,6 +484,28 @@ int Gui::select(const std::string& type,
   logger_->error(utl::GUI, 35, "Unable to find descriptor for: {}", type);
 }
 
+void Gui::netAttributeFilter(const std::string& attribute,
```
maliberty, line 487: I was envisioning this to be more generic.  The descriptor has getProperties that lets you traverse them generically.  If the property name matches the 'attribute' then compare the property value to 'value'.  I think you could cover all types/properties in a carefully constructed method rather than having to enumerate every case.

src/gui/src/gui.tcl
AcKoucher: I was asking myself if it would make sense to apply this filter if we know the name of what we're trying to select. On second thought, I guess it would if we want to know if a specific object is, in this case, an IO net. Should I allow both?

maliberty: I don't see a reason not to.

AcKoucher:
Some considerations after the last changes:
I think it would be nice to review the properties' syntax (I'd be happy to do that). There are some properties which are written like: Tech Layer, but in other objects it's written as Tech layer (with lower case 'l').
It would be cool to have an option in the TCL Input context menu for inserting special characters (Ω, μ, etc.). It could be use to enter something like:
`select -type -filter {Resistance=75.32 mΩ/sq}`

src/gui/src/gui.cpp
```diff
@@ -450,32 +452,96 @@ int Gui::select(const std::string& type,
...
+                                    const std::string& attribute,
+                                    const std::any& value,
+                                    bool& is_valid_attribute)
+{
+  for (Descriptor::Property property : properties) {
+    if (attribute == property.name) {
+      is_valid_attribute = true;
+      if (auto props_selected_set
+          = std::any_cast<SelectionSet>(&property.value)) {
+        if (Descriptor::Property::toString(value) == "CONNECTED"
```
maliberty, line 516: This should be documented

src/gui/src/gui.tcl
```diff
@@ -182,31 +182,45 @@ proc save_clocktree_image { args } {
...
+  set value ""
+  if { [info exists keys(-filter)] } {
+    set filter $keys(-filter)
+    set filter [split $filter "="]
+    if {[llength $filter] != 2} {
+      utl::error GUI 56 "Invalid syntax for -filter. Use -filter attribute=value."
+    } else {
+      set attribute [lindex $filter 0]
+      set value [lindex $filter 1]
+    }
```
maliberty, line 221: error will stop execution so there is no need for an else
```
    if {[llength $filter] != 2} {
      utl::error GUI 56 "Invalid syntax for -filter. Use -filter attribute=value."
    } 
    set attribute [lindex $filter 0]
    set value [lindex $filter 1]
```

maliberty:
I really like this - its very powerful.

maliberty (approved):

PR #3832, opened 2023-08-09
gui: fix text rendering, recalculate dimensions after elide

### LINKED ISSUE #3829: Render labels on top of the macro instead of to the left when zoomed out
### Description

open floorplan.odb in the GUI, zoom in/out.

zoomed out, the labels are to the left of the macros in some cases, but not all:

image

[floorplan_microwatt_sky130hd_base_2023-08-09_13-52.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/12302250/floorplan_microwatt_sky130hd_base_2023-08-09_13-52.tar.gz)

### Suggested Solution

_No response_

### Additional Context

_No response_

oharboe:
We don't need great accuracy in choosing horizontal or vertical direction for text rendering, but once we have chosen direction, we want to center the text more precisely, so recalculate bounding box of text for elided text.

fixes #3829

### DIFF
--- a/src/gui/src/renderThread.cpp
+++ b/src/gui/src/renderThread.cpp
@@ -690,6 +690,7 @@ bool RenderThread::drawTextInBBox(const QColor& text_color,
     name = font_metrics.elidedText(
         name, Qt::ElideLeft, size_limit * bbox_in_px.width());
   }
+  text_bounding_box = font_metrics.boundingRect(name);
 
   painter->setPen(QPen(text_color, 0));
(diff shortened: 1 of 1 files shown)


maliberty (approved):

gadfort:
@maliberty I'm still getting vertical text that is not on an instance

gadfort:
I am still debugging to find a fix, but there is something else wrong with the solution.

gadfort:
@maliberty nevermind, I think I pulled the wrong version

PR #3833, opened 2023-08-09
gui: Fixes stub initialization

QuantamHD:
See discussion https://github.com/The-OpenROAD-Project/OpenROAD/pull/3629#issuecomment-1668739671

The root of the issue is that gui::Gui::get() returns nullptr, and some tools try to run tools against it yielding segfaults.

### DIFF
--- a/src/gui/src/stub.cpp
+++ b/src/gui/src/stub.cpp
@@ -67,6 +67,9 @@ Gui::Gui()
 
 Gui* gui::Gui::get()
 {
+  if (singleton_ == nullptr) {
+    singleton_ = new Gui();
+  }
   return singleton_;
 }
 


maliberty:
I disagree with this change.  The clients should be checking for Gui::enabled() and not calling get().  Returning a Gui object is dangerous and will likely lead to more crashes later.  Please file a bug or PR to fix clients as needed.

QuantamHD:
I disagree with your disagreement. You approved @gadfort's change https://github.com/The-OpenROAD-Project/OpenROAD/pull/3629 which previously checked enabled.

I think the issue is that it's a one time init, and if you call gui::show it won't be correctly init'd

gadfort:
@QuantamHD where are we calling the ::get()? I agree with @maliberty we should not be returning an object, but instead fix the underlying issue

gadfort:
@QuantamHD no, that takes a way functionality from the GUI.
Either the solution is:
1. wrap the renderer set with a check (a partial revert).
2. make sure the path renderer does not unconditionally call gui::get()

maliberty:
@gadfort 1 makes sense to me.  Would you make a PR for it?

@QuantamHD would that resolve your concern?

gadfort:
@maliberty #3836

QuantamHD:
Works for me. Thanks @gadfort

PR #3835, opened 2023-08-09
ppl: use the number of lone pins and number of groups to define the default number of perturbations

eder-matheus:
Previously, the number of perturbations per iteration was 80% of the total number of pins. However, with the presence of pin groups, using the total number of pins makes little sense since a pin group is moved as a single object. Also, moving pin groups is more expensive than moving lone pins, so designs with many groups were having trouble with the original approach.

With this update, the number of pin groups is considered when calculating the number of perturbations per iteration, avoiding running too many perturbations when the number of groups is large.


maliberty (approved):
looks fine but please format

PR #3841, opened 2023-08-10
Or 3783

### LINKED ISSUE #3783: Virtual Resizer running out of memory
### Describe the bug

From https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/1296

This test case sees memory blow up.  It happens during the call to the virtual resizer from the global placer.  It seems to be stuck buffering a net forever.

### Expected Behavior

Reasonable memory and runtime.

### Environment

```shell
Head of master
```

### To Reproduce

https://drive.google.com/file/d/1B--Yyo3EOvMHq2ySyXwNk24psujtZEoZ/view?usp=sharing

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

openroadie:
Fixes #3783 
Also fixes https://github.com/The-OpenROAD-Project/OpenLane/issues/1930


src/rsz/src/RepairDesign.cc
```diff
@@ -635,9 +636,31 @@ RepairDesign::repairNet(BufferedNetPtr bnet,
...
+    best_case_slew_load_ = ref_cap;
+    best_case_slew_ = slew;
+  }
+
+  if (max_load_slew < best_case_slew_) {
+    logger_->error(RSZ, 90, "Max transition time from SDC is {}. Best achievable transition time is {} with {} load",
+                   max_load_slew, best_case_slew_, best_case_slew_load_);
+  }
+}
+
```
QuantamHD, line 661: Can we move this to a seperate file that's conceptually targeted at prechecks? I think it would be good to build an infrastructure around prechecks and preconditions.

src/rsz/src/RepairDesign.cc
```diff
@@ -746,16 +776,20 @@ RepairDesign::repairNetWire(BufferedNetPtr bnet,
...
+          split_length = metersToDbu(l);
         split_wire = true;
         resize = false;
       }
       else {
         split_length = 0;
-        split_wire = true;
-        resize = false;
+        split_wire = false;
+        resize = true;
```
QuantamHD, line 789: Why the change here?

src/rsz/src/RepairDesign.hh
```diff
@@ -250,6 +251,12 @@ protected:
   static constexpr float elmore_skew_factor_ = 1.39;
   static constexpr int min_print_interval_ = 10;
   static constexpr int max_print_interval_ = 100;
+
+  // best slew numbers to ensure the max_slew in SDC is reasonable
+  float best_case_slew_;
+  float best_case_slew_load_;
+  bool best_case_slew_computed_;
+
```
QuantamHD, line 259: Harkening back to my comment above I think it would be good to encapsulate these into their own class.

src/rsz/test/repair_slew13.ok
```diff
@@ -21,11 +21,10 @@ b1/A                                    0.03    0.07   -0.04 (VIOLATED)
...
 [INFO RSZ-0058] Using max wire length 2393um.
 [INFO RSZ-0034] Found 1 slew violations.
-[INFO RSZ-0038] Inserted 1 buffers in 1 nets.
 [INFO RSZ-0039] Resized 1 instances.
 max slew
 
 Pin                                    Limit    Slew   Slack
 ------------------------------------------------------------
-b1/A                                    0.03    0.04   -0.01 (VIOLATED)
+b1/A                                    0.03    0.07   -0.04 (VIOLATED)
```
QuantamHD, line 29: Is this change expected?

QuantamHD (changes requested):

src/rsz/src/RepairDesign.cc
openroadie: This looks to be a bug if you look at the code in it's wider context.

src/rsz/src/RepairDesign.cc
openroadie: @QuantamHD sure thing. Since the transition is sensitive to load, it is a bit tricky to implement this as an upfront computation esp. as we move towards using global routing information for parasitic computations.

src/rsz/src/RepairDesign.hh
openroadie: I will create a new file. As I mentioned earlier, since these checks are dependent on loads, doing them too early might not be the best idea. On the other hand, the infinite loop of inserting a buffer at the same spot  had two root causes. One of them is the slew limit and the other one had to do with the bug in logic where we are splitting a net of length L into a net of length L and a subnet of length 0.

src/rsz/test/repair_slew13.ok
openroadie: Yes... because the previous code logic was buggy.

src/rsz/src/RepairDesign.cc
QuantamHD: I feel like we should pull this change into its own PR since it seems unrelated to the infinite loop bug. Unless I'm mistaken.

src/rsz/src/PreChecks.cc
```diff
@@ -0,0 +1,84 @@
...
+    float slew = resizer_->bufferSlew(resizer_->buffer_lowest_drive_, ref_cap, resizer_->tgt_slew_dcalc_ap_);
+    if (equiv_cells != nullptr) {
+      for (LibertyCell *buffer: *equiv_cells) {
+        slew = std::min(slew, resizer_->bufferSlew(buffer, ref_cap, resizer_->tgt_slew_dcalc_ap_));
+      }
+    }
+    best_case_slew_computed_ = true;
+    best_case_slew_load_ = ref_cap;
+    best_case_slew_ = slew;
+  }
```
QuantamHD, line 77: It seems like we can avoid best_case_slew_computed_, depending on how this code interacts with resizer.

Can we move this code to the init function, and just calculate it once? That would obviate the need for the best_case_slew_computed_ variable since we would just do it once during init.

If it's not possible to move it to init then I think we should wrap best_case_slew_load_ and best_case_slew_ in struct, and mark it as std::optional, and check for its presence. It would make it clearer to the future readers that this is really just a cache. I think it could potentially also make sense to move this into a private method.

```c++
private: 
struct LibrarySlewProperties {
  float best_case_slew_load_;
  float best_case_slew_load_;
};

LibrarySlewProperties getLibrarySlewProperties() {
  if (slew_properties_) {
    return *slew_properties_;
  }

  // Calculate and set slew_properties_
}

std::optional<LibrarySlewProperties> slew_properties_;
```

openroadie: @QuantamHD The problem here is that the best case slew depends on the load. I do not quite comprehend how we can compute it just once when load might need to depend on cell placement. It is a cache.... but a cache that is only valid as long as we don't get a lower best case load (when the cache becomes invalid). Too much abstraction does not necessarily lead to better code or more maintainable code ... but let me think about that some more.

src/rsz/src/RepairDesign.cc
openroadie: Moved to PreChecks.cc/PreChecks.hh

src/rsz/src/PreChecks.cc
QuantamHD: Sounds good, and generally agree on not adding unnecessary levels of abstraction.

I would generally like to see more structure, and separation of concerns in the codebase, but you're in the best position to judge the trade-off between simplicity and abstraction.

openroadie: @QuantamHD I looked into your suggestions and for now I think I will leave it as it is. We could theoretically do this check at a different level but that is most likely a larger project which takes in the user settings (max transition in this case), other user/design requirements ... and then looks at the library and design data to figure out if they are even reasonable. For now, the changes in this PR are as follows:

- We do check that transition for a specific gate at nominal load is actually going to meet the SDC max transition requirements so the buffer insertion code does not keep trying to fix that error without knowing that it is impossible to fix.
-  There was a genuine bug in the logic where we specify a net of length "X" to be split into lengths "X and 0" (normal expectation is to split it into two non-zero parts. 

The reason both the changes are in the same PR is because both these issues need to be addressed to fix the reported bug. I hope this makes sense.

src/rsz/src/PreChecks.cc
```diff
@@ -0,0 +1,84 @@
...
+
+PreChecks::PreChecks(Resizer* resizer) :
+        logger_(nullptr),
+        sta_(nullptr),
+        resizer_(resizer),
+        best_case_slew_computed_(false) {
+}
+
+void
+PreChecks::init()
```
maliberty, line 57: Why isn't this part of the ctor?  The only usage I see is
```
  pre_checks_ = new PreChecks(resizer_);
  pre_checks_->init();
```

openroadie: Mostly to match the style of the rest of the code. I will make the change.

maliberty (approved):

PR #3843, opened 2023-08-10
dpl: Fixes out of bounds in grid paint

QuantamHD:
In some cases you can have a double height row
with only a single layer row below it instead of two. In that case if you were painting a double height
cell then you would go out of bounds.

This commit adds a check to ensure, that does not happen.

### DIFF
--- a/src/dpl/src/Grid.cpp
+++ b/src/dpl/src/Grid.cpp
@@ -656,7 +656,7 @@ void Opendp::paintPixel(Cell* cell, int grid_x, int grid_y)
       ++layer_y_end;
     }
 
-    if (layer_y_end >= layer.second.row_count) {
+    if (layer_y_end > layer.second.row_count) {
       logger_->error(DPL,
                      49,
                      "Cannot paint grid index: {} because it is out of bounds. "
--- a/src/dpl/src/Grid.cpp
+++ b/src/dpl/src/Grid.cpp
@@ -657,13 +657,13 @@ void Opendp::paintPixel(Cell* cell, int grid_x, int grid_y)
     }
 
     if (layer_y_end >= layer.second.row_count) {
-      // In some cases you can have a double height row
-      // with only a single layer row below it instead of two.
-      // In that case if you were painting a double height
-      // cell then you would go out of bounds.
-      //
-      // This check ensures that does not happen.
-      layer_y_end = layer.second.row_count;
+      logger_->error(DPL,
+                     49,
+                     "Cannot paint grid index: {} because it is out of bounds. "
+                     "Calculated (row end {}) > (rows {}). This is a bug",
+                     layer.first,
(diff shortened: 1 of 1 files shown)


maliberty:
I don't understand how this could happen.  I could see a single height row with no associated double height but not visa versa.  Would you give an example?

QuantamHD:
Can't give you any databases, but this did segfault on my PDK.

maliberty:
I suspecting it is a bug in row generation and we need to fix it there.  Can you look at the relevant rows and see what's happening?

QuantamHD:
I'm having a hard time replicating the bug, but I might be remembering it backwards. Either way this precheck shouldn't ever be incorrect.

maliberty:
I'm concerned that you will get a result that isn't valid if it is as you describe.  This seems more like an internal error.

QuantamHD:
We can go with an error condition if it's violated if you would prefer.

maliberty:
> We can go with an error condition if it's violated if you would prefer.

I would.

QuantamHD:
Done.

maliberty (approved):

maliberty:
Please address the unit tests

QuantamHD:
It looks like these failures are related to the double height cell work. Can you have @mina1460 take a look

mina1460:
@QuantamHD looking into it now

src/dpl/src/Grid.cpp
```diff
@@ -655,6 +656,16 @@ void Opendp::paintPixel(Cell* cell, int grid_x, int grid_y)
       ++layer_y_end;
     }
 
+    if (layer_y_end >= layer.second.row_count) {
```
mina1460, line 659: The equal sign is what is causing the error.
The error message says:
`[ERROR DPL-0049] Cannot paint grid index: 2800 because it is out of bounds. Calculated (row end 6) > (rows 6). This is a bug`

Which is not really a bug since layer_y_end is an exclusive bound. We never touch it as you can see in the loop below.

mina1460 (changes requested):

src/dpl/src/Grid.cpp
mina1460: Removing it results in all tests passing.

maliberty: @QuantamHD the ball is in your court

maliberty: @QuantamHD ??

QuantamHD: Done.

QuantamHD:
Should be good to go.

maliberty (approved):
This isn't a fix, just an error message.  Is that intended?

QuantamHD:
Yeah. I left this in waiting for a repro to show up in our tests

PR #3853, opened 2023-08-12
Supporting ROWPATTERNs in  ODB

mina1460:


oharboe:
@mina1460 Isn't this a schema change? Doesn't the .odb version need to be bumped?

src/odb/src/db/dbSite.cpp
```diff
@@ -230,6 +232,50 @@ void dbSite::setClass(dbSiteClass type)
   site->_flags._class = type.getValue();
 }
 
+void dbSite::setRowPattern(
+    std::vector<std::pair<std::string, std::string>>& row_pattern)
```
QuantamHD, line 236: I think it would be better to replace std::pair with a class named something like RowPatternElement where the definition is something like.
```c++
class RowPatternElement {
public:
  RowPatternElement(dbSite*, dbOrientType)
}
```

I think it would be a good idea to force the user to provide a dbSite* and dbOrientType upfront. That way ODB doesn't need to do as much error handling. As it is you can set an invalid name, and it will just get silently removed during get.

gadfort: I agree, that is much easier to understand

QuantamHD:
> @mina1460 Isn't this a schema change? Doesn't the .odb version need to be bumped?

It looks like this PR doesn't add the necessary code to actually persist this information in ODB. You should take a look at how dbBlock is implemented. Specifically, the ones implemented with dbVector https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/src/db/dbBlock.h#L187

@maliberty might have more specific advice for you on how you should implement this, but from my limited understanding you're going to need to implement the follow

- [ ] dbRowPatternElement, where it will hold a dbId<dbSite> and a dbOrientType as memebers. 
- [ ] add dbVector<dbRowPatternElement> to dbSite

src/odb/src/db/dbSite.cpp
maliberty: The conversion to RowPatternElement  should happen in lefin along with validation

src/odb/src/db/dbSite.cpp
```diff
@@ -230,6 +232,50 @@ void dbSite::setClass(dbSiteClass type)
...
+  site->_row_patterns.reserve(row_pattern.size());
+  std::unordered_map<std::string, dbOrientType::Value> orientationMap
+      = {{"N", dbOrientType::R0},
+         {"W", dbOrientType::R270},
+         {"S", dbOrientType::R180},
+         {"E", dbOrientType::R90},
+         {"FN", dbOrientType::MYR90},
+         {"FW", dbOrientType::MY},
+         {"FS", dbOrientType::MXR90},
+         {"FE", dbOrientType::MX}};
```
maliberty, line 248: Handle in lefin

src/odb/src/db/dbSite.h
```diff
@@ -64,6 +64,7 @@ class _dbSite : public _dbObject
   uint _height;
   uint _width;
   dbId<_dbSite> _next_entry;
+  std::vector<std::pair<std::string, dbOrientType::Value>> _row_patterns;
```
maliberty, line 67: Please see dbDatabase.h for how schema updates are handled for backwards compatability.

PR #3865, opened 2023-08-16
Remove redundant call to compute incremental parasitics + bug fix in incremental routing

openroadie:


openroadie:
Adding @eder-matheus and @luis201420 to make sure my fix is legitimate. Before, the number of incrementally routed nets was monotonically increasing (which does not seem right). With these changes it was limited to a reasonable number of nets (usually less than 5).

openroadie:
Results are identical before/after... but more interestingly.... the runtime also did not change much. @luis201420 @eder-matheus Any idea why that would be the case? Is it because the routing code is smart enough not to reroute nets that we were previously sending it (but had no change)?

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -219,6 +219,8 @@ Resizer::incrementalParasiticsBegin()
     break;
   case ParasiticsSrc::global_routing:
     incr_groute_ = new IncrementalGRoute(global_router_, block_);
+    // Don't print verbose messages for incremental routing
+    global_router_->setVerbose(false);
```
eder-matheus, line 223: Great catch! I believe many messages generated in incremental runs can make the runtime worse.

src/rsz/src/RecoverPower.cc
```diff
@@ -140,17 +140,18 @@ RecoverPower::recoverPower(float recover_power_percent)
...
 
   // As long as we are here fix at least one path
   if (max_end_count == 0) {
     max_end_count = 1;
-}
+  }
 
   resizer_->incrementalParasiticsBegin();
+  resizer_->updateParasitics();
+  sta_->findRequireds();
```
eder-matheus, line 152: I'm not an expert on resizer or sta, but is this the key to having fewer nets being incrementally routed?
It's more my curiosity than something that should be fixed :)

eder-matheus:
> Results are identical before/after... but more interestingly.... the runtime also did not change much. @luis201420 @eder-matheus Any idea why that would be the case? Is it because the routing code is smart enough not to reroute nets that we were previously sending it (but had no change)?

AFAIK, the router will not reroute nets that did not have changes regarding the pin positions in the global routing grid. So small changes on the nets do not trigger the reroute of them.

What design are you using to compare the runtime? Also, did you see a reduction in the number of iterations when running repair_design/repair_timing?

openroadie:
> > Results are identical before/after... but more interestingly.... the runtime also did not change much. @luis201420 @eder-matheus Any idea why that would be the case? Is it because the routing code is smart enough not to reroute nets that we were previously sending it (but had no change)?
> 
> AFAIK, the router will not reroute nets that did not have changes regarding the pin positions in the global routing grid. So small changes on the nets do not trigger the reroute of them.
> 
> What design are you using to compare the runtime? Also, did you see a reduction in the number of iterations when running repair_design/repair_timing?

@eder-matheus : No reduction in runtime. I was using asap7/ethmac_lvt where the goal of "repair_timing -recover_power" is to downsize cells (or replace low Vt cells with high Vt cells). I agree that the router does not reroute nets which have no changes but it does not make much sense to carry around an ever growing list of nets for incremental route (even though only the last 5 nets or so are being actually rerouted).

src/rsz/src/RecoverPower.cc
openroadie: No. This is just to initialize the code once and then call it again when we make changes. The original code was calling updateParasitics/findRequireds twice per iteration.

src/grt/src/GlobalRouter.cpp
```diff
@@ -545,6 +545,7 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
       dirty_nets.push_back(db_net_map_[db_net]);
     }
   }
+  dirty_nets_.clear();
```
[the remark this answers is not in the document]
openroadie: @eder-matheus I do not think this original code is correct. I suspect that if we hit the return statement then the dirty_nets_.clear() call will not be made. The only reason I  made this PR was because I saw that the dirty_nets_ vector size somehow kept growing linearly with every call. 

```
    std::vector<Net*> dirty_nets;
    updateDirtyNets(dirty_nets);

    if (dirty_nets.empty()) {  
      return;  
    }
```

I will delete the original call since you also like the new place where the vector is cleared :+1:

maliberty (approved):

src/grt/src/GlobalRouter.cpp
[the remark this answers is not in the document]
eder-matheus: You're right. I didn't notice this `return` could be called even when the `dirty_nets_` vector is not empty. Thanks for the clarification!

PR #3869, opened 2023-08-16
psm: check grid fix for floating marcos

gadfort:
Changes:
- adds separate function to find instances which are not connected to the grid.

Adds:
- test cases for disconnected marcos and padrings


src/pad/src/ICeWall.cpp
```diff
@@ -1123,7 +1123,7 @@ std::set<odb::dbNet*> ICeWall::connectByAbutment(
 
 std::vector<std::pair<odb::dbITerm*, odb::dbITerm*>> ICeWall::getTouchingIterms(
     odb::dbInst* inst0,
-    odb::dbInst* inst1) const
+    odb::dbInst* inst1)
```
maliberty, line 1126: Why did this stop being const?

maliberty (changes requested):

src/pad/src/ICeWall.cpp
gadfort: I needed it in psm, so I made it static instead

maliberty (approved):

PR #3876, opened 2023-08-18
gui: fix placement and power density heatmaps

### LINKED ISSUE #3863: placement heatmap does not extend to the top and the right, like it does to the left and bottom
### Describe the bug

Using https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1356#top, run `make DESIGN_CONFIG=designs/asap7/mock-alu/config.mk` to produce 6_final.odb, attached here [mock-alu.zip](https://github.com/The-OpenROAD-Project/OpenROAD/files/12345871/mock-alu.zip)

Then launch `openroad -gui` and load results/asap7/mock-alu/base/6_final.odb and enable the placement heatmap

Notice that the heatmap extends to the end of the die on the left and bottom, but not at the top and right.

image

### Expected Behavior

The heatmap should extend to the top and the right, like it does to the left and bottom.

### Environment

```shell
OpenROAD v2.0-9875-g7430f83d9
```

### To Reproduce

See above

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
Fixes #3863

When creating the mapview of the heatmap, the last column and row were not being taken into account. The situations where the wrong behavior was visible were only those in which the cells of the last column and row were smaller than the grid.

Power density was having the same problem.

### DIFF
--- a/src/gui/src/heatMap.cpp
+++ b/src/gui/src/heatMap.cpp
@@ -349,7 +349,7 @@ HeatMapDataSource::MapView HeatMapDataSource::getMapView(
   const int x_high = std::min(
       static_cast<int>(
           std::ceil(bounds.xMax() / (getGridXSize() * dbu_per_micron))),
-      static_cast<int>(map_.shape()[0]) - 1);
+      static_cast<int>(map_.shape()[0]));
   const int y_low = std::max(
       static_cast<int>(
           std::floor(bounds.yMin() / (getGridYSize() * dbu_per_micron))),
@@ -357,7 +357,7 @@ HeatMapDataSource::MapView HeatMapDataSource::getMapView(
   const int y_high = std::min(
       static_cast<int>(
           std::ceil(bounds.yMax() / (getGridYSize() * dbu_per_micron))),
-      static_cast<int>(map_.shape()[1]) - 1);
+      static_cast<int>(map_.shape()[1]));
 
   return map_[boost::indices[Map::index_range(x_low, x_high)]
                             [Map::index_range(y_low, y_high)]];


AcKoucher:
Still trying to figure out the routing (#3862). Apparently, there's something else going on.

src/gui/src/heatMap.cpp
```diff
@@ -349,15 +349,15 @@ HeatMapDataSource::MapView HeatMapDataSource::getMapView(
   const int x_high = std::min(
       static_cast<int>(
           std::ceil(bounds.xMax() / (getGridXSize() * dbu_per_micron))),
-      static_cast<int>(map_.shape()[0]) - 1);
+      static_cast<int>(map_.shape()[0]));
```
gadfort, line 352: wouldn't this be out of bounds? ie. is the array is 10 wide the indexes go from 0 -> 9?

AcKoucher: Now, I got a little confused. I guess it makes sense, but:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/2f2fd3dbe7d377126468b6f79bd7407da81afc6c/src/gui/src/heatMap.cpp#L631-L635

Here, the drawing iterations are not getting to the last raw/column.

In your example (considering only x coordinate): array is 10 wide,
x_size must be 10 so that the last element 9 is covered, but, as it currently is, x_size=9.

I am missing something?

gadfort: is `::index_range` inclusive or exclusive of the endpoint?

gadfort (approved):
I believe this is correct.

maliberty (approved):

src/gui/src/heatMap.cpp
AcKoucher: Although this is already merged, just FYI: https://www.boost.org/doc/libs/1_82_0/libs/multi_array/doc/reference.html#generators

As described in Associated Types for Views section, it looks like the interval is open in the upper bound.

`::index_range(lower_bound, upper_bound)`
lower_bounds is inclusive, upper_bound is exclusive

PR #3877, opened 2023-08-18
grt: fix updateDbCongestion with the correct values for the last col/row

eder-matheus:


src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -791,9 +791,14 @@ void FastRouteCore::updateDbCongestion()
     const unsigned short capH = h_capacity_3D_[k];
     const unsigned short capV = v_capacity_3D_[k];
     for (int y = 0; y < y_grid_; y++) {
-      for (int x = 0; x < x_grid_ - 1; x++) {
-        const unsigned short blockageH = capH - h_edges_3D_[k][y][x].cap;
-        const unsigned short blockageV = capV - v_edges_3D_[k][y][x].cap;
+      for (int x = 0; x < x_grid_; x++) {
+        // use std::abs because in the last col/row, the edge capacity can be
+        // greater than the default cap of the layer, since these col/row are
+        // larger
```
maliberty, line 797: I understand the last entry can be larger but how does that relate to abs?

maliberty:
Does this affect results in any way or is it just for GUI display?

src/grt/src/fastroute/src/FastRoute.cpp
eder-matheus: I made confusion here, my bad. My last commit should properly cover the calculation of the congestion for every case, including the last entries.

eder-matheus:
> Does this affect results in any way or is it just for GUI display?

I should affect only the GUI.

maliberty (approved):

PR #3879, opened 2023-08-19
[Docs] Mermaid anchor fix test 2

luarss:
Continuing here because the RTD docs stopped uploading changes to the commits.


luarss:
@vvbandeira Looks like this now. 

image

PR #3887, opened 2023-08-21
Add support for LEF57_MINSTEP in ROUTING & LEF57_SPACING in CUT

refaay:


src/odb/src/lefin/lefin.cpp
```diff
@@ -694,7 +694,7 @@ void lefin::layer(lefiLayer* layer)
       } else
         supported = false;
     } else if (type.getValue() == dbTechLayerType::CUT) {
-      if (!strcmp(layer->propName(iii), "LEF58_SPACING")) {
+      if (!strcmp(layer->propName(iii), "LEF58_SPACING") | !strcmp(layer->propName(iii), "LEF57_SPACING")) {
```
QuantamHD, line 697: Even though bitwise operator `|` is technically correct in this case it's much preferable to use the `||` logical operator when dealing with truth values. Please change to the logical operator `||`

QuantamHD:
You should also include a unit test for this functionality

https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/test/cpp/TestAbstractLef.cc

src/odb/src/lefin/lefin.cpp
refaay: Done

maliberty (approved):

maliberty:
Please clang-format the code to fix the error

maliberty:
Please add the test as requested

osamahammad21 (approved):

PR #3888, opened 2023-08-21
Grt: iterating only the ggrids that are modified

luis201420:
This update avoids iterating ggrids that were not modified in certain fastroute functions.

With the modification, the runtime for the secure-ci designs is reduced and the private designs `bp_single` and `ca53` that previously did not finish executing now manage to finish without errors.


src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -590,6 +591,9 @@ class FastRouteCore
       vertical_blocked_intervals_;
   std::unordered_map<Tile, interval_set<int>, boost::hash<Tile>>
       horizontal_blocked_intervals_;
+
+  std::set<std::pair<int, int>> h_used_ggrid_;
+  std::set<std::pair<int, int>> v_used_ggrid_;
```
eder-matheus, line 596: Perhaps using `std::unordered_set` instead of `std::set`? If I understand correctly, your code doesn't care about the ordering, and an `unordered_set` will have better runtime to insert the elements.

You can try it with a large public design and check if the runtime is better and if the result keeps the same.

eder-matheus (changes requested):

eder-matheus:
@luis201420 It looks reasonable, but you should run secure-ci for these changes to ensure it is not breaking anything.

Please, add a secure branch with these changes in OR without any changes in the ORFS scripts (i.e., without running incremental groute).

src/grt/src/fastroute/include/FastRoute.h
luis201420: For the public designs (ng45/bp and ng45/ariane136) I didn't find much difference in the runtime when I use `std::set` and `std::unordered_set` (I think it's due to the small number of ggrids that are modified in each iteration), even so I am going to add that modification since in private designs there may be a greater impact.

eder-matheus (approved):
The changes look good, but let's wait for the last CI run to make sure it does not crash anything.

maliberty:
> The changes look good, but let's wait for the last CI run to make sure it does not crash anything.

Any update on this?

maliberty (approved):

PR #3889, opened 2023-08-22
gui: consider last row/column when populating routing map

### LINKED ISSUE #3862: routing heatmap is missing data on the right and on the top
### Describe the bug

Using https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1356#top, run `make DESIGN_CONFIG=designs/asap7/mock-alu/config.mk` to produce 6_final.odb, attached here [mock-alu.zip](https://github.com/The-OpenROAD-Project/OpenROAD/files/12345871/mock-alu.zip)

Then launch `openroad -gui` and load results/asap7/mock-alu/base/6_final.odb and enable the routing heatmap

Notice that the heatmap extends to the end of the die on the left and bottom, but not at the top and right.

image

### Expected Behavior

The heatmap should extend to the top and the right, like it does to the left and bottom.

### Environment

```shell
OpenROAD v2.0-9875-g7430f83d9
```

### To Reproduce

See above

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
Fixes #3862 

The routing map last column/row were being skipped when populating the map.

### DIFF
--- a/src/grt/src/heatMap.cpp
+++ b/src/grt/src/heatMap.cpp
@@ -203,12 +203,14 @@ bool RoutingCongestionDataSource::populateMap()
     for (uint y_idx = 0; y_idx < gcell_congestion_data.numCols(); ++y_idx) {
       const auto& cong_data = gcell_congestion_data(x_idx, y_idx);
 
-      if (x_idx + 1 >= x_grid_sz || y_idx + 1 >= y_grid_sz) {
-        continue;
-      }
+      const int next_x = (x_idx + 1) == x_grid_sz
+                             ? getBlock()->getDieArea().xMax()
+                             : x_grid[x_idx + 1];
+      const int next_y = (y_idx + 1) == y_grid_sz
+                             ? getBlock()->getDieArea().yMax()
+                             : y_grid[y_idx + 1];
 
-      const odb::Rect gcell_rect(
-          x_grid[x_idx], y_grid[y_idx], x_grid[x_idx + 1], y_grid[y_idx + 1]);
+      const odb::Rect gcell_rect(x_grid[x_idx], y_grid[y_idx], next_x, next_y);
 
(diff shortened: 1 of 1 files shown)


src/grt/src/heatMap.cpp
```diff
@@ -203,12 +203,14 @@ bool RoutingCongestionDataSource::populateMap()
     for (uint y_idx = 0; y_idx < gcell_congestion_data.numCols(); ++y_idx) {
       const auto& cong_data = gcell_congestion_data(x_idx, y_idx);
 
-      if (x_idx + 1 >= x_grid_sz || y_idx + 1 >= y_grid_sz) {
```
maliberty, line 206: The code used to check greater than or equal but now only checks equal. Is that safe?

AcKoucher: x_grid_sz and y_grid_sz come from the GCellGrid.
https://github.com/The-OpenROAD-Project/OpenROAD/blob/b989394af4cf28e07b9bea3617ac1ffdd68f3da2/src/grt/src/heatMap.cpp#L186
https://github.com/The-OpenROAD-Project/OpenROAD/blob/b989394af4cf28e07b9bea3617ac1ffdd68f3da2/src/grt/src/heatMap.cpp#L196-L200
The loops are iterating the grid with limits being set from `gcell_congestion_data.numRows()` `gcell_congestion_data.numCols()`, which also comes from the GCellGrid.
https://github.com/The-OpenROAD-Project/OpenROAD/blob/b989394af4cf28e07b9bea3617ac1ffdd68f3da2/src/grt/src/heatMap.cpp#L191
I think in the original code == would be enough, because these loops can't reach out of the map scope, unless for, any reason, the CongestionMap can be larger than the GCellGrid.

Either way, I can change for safety if you wish :)

PR #3893, opened 2023-08-22
support map_power_switch in upf

ahmadelrouby:
Added support for the "map_power_switch" command in UPF


src/odb/include/odb/db.h
```diff
@@ -7583,8 +7583,13 @@ class dbPowerSwitch : public dbObject
   void setOutSupplyPort(const std::string& out_port);
   void addControlPort(const std::string& control_port);
   void addOnState(const std::string& on_state);
+  void addLibCell(const std::string& lib_cell);
```
gadfort, line 7586: this should take in a dbMaster

src/odb/include/odb/db.h
```diff
@@ -7583,8 +7583,13 @@ class dbPowerSwitch : public dbObject
...
+  void addPortMap(const std::string& model_port,
+                  const std::string& switch_port);
```
gadfort, line 7588: this should take in the dbMTerm for the terminal that maps to a master

src/odb/include/odb/db.h
```diff
@@ -7583,8 +7583,13 @@ class dbPowerSwitch : public dbObject
...
   std::vector<std::string> getControlPorts();
   std::vector<std::string> getOnStates();
+  std::vector<std::string> getLibCells();
+  std::map<std::string, std::string> getPortMap();
```
gadfort, line 7592: return correct types, probably which from map to a struct of some sort @maliberty  might have a good idea of the form it needs to take

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -118,6 +118,8 @@ dbIStream& operator>>(dbIStream& stream, _dbPowerSwitch& obj)
   stream >> obj._control_port;
   stream >> obj._on_state;
   stream >> obj._control_net;
+  stream >> obj._lib_cells;
+  stream >> obj._port_map;
```
gadfort, line 122: db version checking is needed here

src/upf/include/upf/upf.h
```diff
@@ -104,4 +104,15 @@ bool eval_upf(sta::dbNetwork* network,
               utl::Logger* logger,
               odb::dbBlock* block);
 
+bool update_power_switch_cell(utl::Logger* logger,
+                              odb::dbBlock* block,
+                              const char* name,
+                              const char* cell);
```
gadfort, line 110: dbMaster

src/upf/include/upf/upf.h
```diff
@@ -104,4 +104,15 @@ bool eval_upf(sta::dbNetwork* network,
...
+
+bool update_power_switch_port_map(utl::Logger* logger,
+                                  odb::dbBlock* block,
+                                  const char* name,
+                                  const char* model_port,
+                                  const char* switch_port);
```
gadfort, line 116: dbMTerm

src/upf/src/upf.i
```diff
@@ -106,4 +106,18 @@
                         x2,
                         y2);
   }
+
+  void set_power_switch_cell(char* name, char* cell)
+  {
+    odb::dbDatabase* db = getOpenRoad()->getDb();
+    upf::update_power_switch_cell(
+        getOpenRoad()->getLogger(), db->getChip()->getBlock(), name, cell);
```
gadfort, line 114: a user most definitely would be to call this function before a Block or Chip has been created causing a segfault, same below. Probably needs to have some safeguards in the tcl

maliberty:
@gadfort does this meet you needs?

src/upf/src/upf.tcl
```diff
@@ -301,5 +301,52 @@ proc set_domain_area { args } {
...
+    }
+
+    foreach {switch} $switch_name_list {
+
+        foreach {cell} $lib_cells {
+            upf::set_power_switch_cell $switch $cell
+        }
+
+        foreach {port} $port_map {
+            upf::set_power_switch_port $switch [lindex $port 0] [lindex $port 1]
```
gadfort, line 349: need to check that the port is 2 elements wide and not 1 or 3+

gadfort (changes requested):

gadfort:
@maliberty I think this stores the information I need.

ahmadelrouby:
The original idea for these objects were to store the raw text in case it was needed in another way later on. What I could do is provide methods to return the information in the format you'd prefer (dbMaster, dbMTerm, etc). As for the setter functions, these are made for reading in the upf text commands so it has to be a string(which again could be processed when you query the db). If there's supposed to be another source for mapping the ports other than upf, I'll add other methods that can actually use dbMaster, dbMTerm, etc.

src/upf/src/upf.tcl
```diff
@@ -301,5 +301,52 @@ proc set_domain_area { args } {
...
+#
+#
+# Argument list: 
+#  - switch_name_list []: A list of switches (as defined by create_power_switch) to map.
+#  - lib_cells [] : A list of library cells.
+#  - port_map {{mapped_model_port switch_port_or_supply_net_ref} *} : mapped_model_ port is a port on the model being mapped.
+#                                                                     switch_ port_or_supply_net_ref indicates a supply or logic port on a
+#                                                                     switch: an input supply port, output supply port, control port, or
+#                                                                     acknowledge port; or it references a supply net from a supply set
+#                                                                     associated with the switch.
```
vvbandeira, line 315: @luarss 
FYI, not sure if we will merge the docs PR on upf before or after this, but this will need to be included.

gadfort:
@ahmadelrouby The dbMaster and dbMTerm information would already have been loaded before calling upf, so it seems better to store those pointers instead of a string.

maliberty:
Since this data is just LEF dependent we can store the db objects directly.  The UPF related to the design is kept unevaluated so it can be reapplied as the design is modified.

src/upf/src/upf.tcl
vvbandeira: @ahmadelrouby we have merged the UPF docs PR, could you update the README to include the new command?

vvbandeira (changes requested):

src/odb/include/odb/db.h
ahmadelrouby: This one was added for the upf commands, I've added another one that takes dbMaster

src/odb/include/odb/db.h
ahmadelrouby: same as above

src/odb/include/odb/db.h
ahmadelrouby: I've updated the types, libCell returns now just a dbMaster, and portMap maps string -> dbMTerm

src/upf/include/upf/upf.h
ahmadelrouby: This one is read by the tcl interface so it has to stay char*

src/upf/include/upf/upf.h
ahmadelrouby: same

src/upf/src/upf.tcl
ahmadelrouby: I'm not sure I understand what you mean. Could you please elaborate?

src/upf/src/upf.i
ahmadelrouby: I've added safeguards in the tcl. should be okay now

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: done

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -118,6 +121,9 @@ dbIStream& operator>>(dbIStream& stream, _dbPowerSwitch& obj)
   stream >> obj._control_port;
   stream >> obj._on_state;
   stream >> obj._control_net;
+  stream >> obj._lib_cell;
+  stream >> obj._lib;
+  stream >> obj._port_map;
```
maliberty, line 126: You need to provide backward compatibility here (load is conditional db_schema_upf_power_switch_mapping)

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -240,16 +237,142 @@ void dbPowerSwitch::addOnState(const std::string& on_state)
   obj->_on_state.push_back(on_state);
 }
 
+void dbPowerSwitch::addLibCell(const std::string& lib_cell)
```
maliberty, line 240: Can a power switch have more than one cell?

src/odb/include/odb/db.h
```diff
@@ -7579,12 +7575,31 @@ class dbPowerSwitch : public dbObject
   // User Code Begin dbPowerSwitch
   static dbPowerSwitch* create(dbBlock* block, const char* name);
   static void destroy(dbPowerSwitch* ps);
-  void setInSupplyPort(const std::string& in_port);
-  void setOutSupplyPort(const std::string& out_port);
+  void addInSupplyPort(const std::string& in_port);
+  void addOutSupplyPort(const std::string& out_port);
```
maliberty, line 7579: should these take dbMTerm?

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: According to my understanding of the spec not really. Are you referring to the word "add" instead of "set"?

src/odb/include/odb/db.h
ahmadelrouby: this is the logical port defined the upf. It's a string that is then later mapped to a dbMTerm

src/odb/src/db/dbPowerSwitch.cpp
maliberty: yes

src/odb/include/odb/db.h
ahmadelrouby: updated

src/upf/include/upf/upf.h
ahmadelrouby: done

src/upf/src/upf.tcl
ahmadelrouby: I added it to the UPF readme file

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: done

src/odb/include/odb/db.h
ahmadelrouby: so I double checked, and these ports are the modeled one specified while creating the power switch. These are later then mapped to dbMTerms by specifying the name of these ports and there equivalent on the dbMaster

src/upf/src/upf.tcl
vvbandeira: Should we remove the comments or replace them with a pointer to the README file? I am worried that in the future if one changes the code it will forget to change both places where the arguments are defined (code and docs).

src/odb/include/odb/db.h
```diff
@@ -7579,12 +7575,30 @@ class dbPowerSwitch : public dbObject
...
   std::vector<std::string> getControlPorts();
+  std::vector<std::string> getInputSupplyPorts();
+  std::vector<std::string> getOutputSupplyPorts();
   std::vector<std::string> getOnStates();
+
+  // Returns all lib cells that were defined in the upf for this power switch
+  // In case of multiple cells, the first one is the cell associated with the
+  // dbMTerms defined
+  //  in the port map
+  dbMaster* getLibCell();
```
maliberty, line 7596: The comment talks about returning multiple cells but the api just a single cell.  What is intended here?

src/upf/src/upf.tcl
maliberty: You are suing linex to get two values from $port.  You should ensure that it is a two element list (not too few or too many entries).

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -132,6 +142,13 @@ dbOStream& operator<<(dbOStream& stream, const _dbPowerSwitch& obj)
   stream << obj._on_state;
   stream << obj._control_net;
   stream << obj._power_domain;
+  // User Code Begin <<
+  if (obj.getDatabase()->isSchema(db_schema_upf_power_switch_mapping)) {
```
maliberty, line 146: There's no need to check on output as this will always be true.  We don't support writing in old formats, only reading them.

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -240,16 +245,115 @@ void dbPowerSwitch::addOnState(const std::string& on_state)
...
+    obj->getImpl()->getLogger()->error(
+        utl::ODB,
+        32001,
+        "Cannot map port {} to {} because no lib cell is added",
+        model_port,
+        switch_port);
+    return;
+  }
+
+  dbMTerm* mterm = master->findMTerm(switch_port.c_str());
```
maliberty, line 273: what if mterm is null - that should be an error.

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -240,16 +245,115 @@ void dbPowerSwitch::addOnState(const std::string& on_state)
...
+  obj->_port_map[model_port] = mterm->getImpl()->getOID();
+}
+
+void dbPowerSwitch::addPortMap(const std::string& model_port, dbMTerm* mterm)
+{
+  _dbPowerSwitch* obj = (_dbPowerSwitch*) this;
+
+  dbMaster* master = (dbMaster*) dbMaster::getMaster(
+      dbLib::getLib((dbDatabase*) obj->getImpl()->getDatabase(), obj->_lib),
+      obj->_lib_cell);
```
maliberty, line 283: Call getLibCell

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -240,16 +245,115 @@ void dbPowerSwitch::addOnState(const std::string& on_state)
...
+  obj->_lib = master->getLib()->getImpl()->getOID();
+}
+
+void dbPowerSwitch::addPortMap(const std::string& model_port,
+                               const std::string& switch_port)
+{
+  _dbPowerSwitch* obj = (_dbPowerSwitch*) this;
+  dbMaster* master = (dbMaster*) dbMaster::getMaster(
+      dbLib::getLib((dbDatabase*) obj->getImpl()->getDatabase(), obj->_lib),
+      obj->_lib_cell);
```
maliberty, line 261: call getLibCell

src/odb/test/data/design.odb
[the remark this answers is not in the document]
ahmadelrouby: Last time you mentioned that I need to run update_db which updates this file when we do versioning changes.

src/odb/include/odb/db.h
ahmadelrouby: I apologize, the comment was outdated, and I've fixed it now

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: got it, fixed

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: done

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: done

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: fixed

src/upf/src/upf.tcl
ahmadelrouby: I get that, but all these methods are mirrors to the spec of UPF, but I'll make sure I updated in both places from now on.

src/odb/src/db/dbPowerSwitch.cpp
```diff
@@ -256,31 +254,35 @@ void dbPowerSwitch::addPortMap(const std::string& model_port,
...
+                  "Cannot map port {} to {} because no lib cell is added",
+                  model_port,
+                  switch_port);
     return;
   }
 
   dbMTerm* mterm = master->findMTerm(switch_port.c_str());
+  if (mterm == nullptr) {
+    logger->error(utl::ODB,
+                  33002,
```
rovinski, line 272: Logger codes should be at most 4 digits.

src/odb/test/data/design.odb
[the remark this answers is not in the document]
maliberty: This is a test for backwards compatibility so you should be able to pass this test with the original file.  That demonstrates backward compatible load.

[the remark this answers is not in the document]
ahmadelrouby: I updated it. thanks

src/odb/src/db/dbPowerSwitch.cpp
ahmadelrouby: Is this a hard requirement, or just a suggestion?

vvbandeira: It is a strong suggestion unless you have a reason why you need five digits. We hope to keep uniformity across the tool; all other tools use four digits, and our code automatically pads them with `0` if needed, i.e., `1` --> `0001`. Also, `9999` should be plenty of messages 😉 .

ahmadelrouby: alright, I've changed as many as I could for this PR. If I come across any others that I've done I'll change them in another PR.

src/upf/src/upf.tcl
ahmadelrouby: got it, updated.

src/odb/src/db/dbPowerSwitch.cpp
maliberty: I believe you did this backwards.  The operator>> needs the check not the operator<<

ahmadelrouby: updated

maliberty:
@gadfort would you give this a last look as you'll be the main consumer of this change for PDN.

gadfort:
@maliberty looks okay to me

maliberty (approved):

maliberty:
@gadfort you should be enabled for power switches in pdn.  Let me know if you need anything further to proceed.

PR #3896, opened 2023-08-23
Improved undo

openroadie:


openroadie:
The new code is not yet enabled since I need to add better statistics (it will also enable us to correctly count number of moves made vs. number of moves that were rejected/undone) but I would like this PR merged so it does not become an unmanageable mess. The next steps are:

- Add the missing corner cases/more thorough testing
- Once ORFS and maybe even secure ORFS successfully passes, we can move to the new undo method
- Delete the existing undo code if there are no issues

src/rsz/test/repair_fanout7.ok
```diff
@@ -12,9 +12,9 @@ load DFF_X1 CK D 300000 metal3 1000
...
-[INFO RSZ-0045] Inserted 78 buffers, 1 to split loads.
-[INFO RSZ-0041] Resized 527 instances.
-[INFO RSZ-0049] Cloned 7 instances.
+[INFO RSZ-0045] Inserted 41 buffers, 1 to split loads.
+[INFO RSZ-0041] Resized 312 instances.
+[INFO RSZ-0049] Cloned 4 instances.
 [WARNING RSZ-0062] Unable to repair all setup violations.
 Repair timing output passed/skipped equivalence test
-worst slack -3.75
+worst slack -2.84
```
maliberty, line 20: If "The new code is not yet enabled" why are there changes to the results?

src/rsz/include/rsz/Resizer.hh
```diff
@@ -514,13 +514,16 @@ protected:
...
                                 SteinerPt pt,
                                 const ParasiticAnalysisPt *parasitics_ap);
-
-  bool replaceCell(Instance *inst,
-                   LibertyCell *cell,
-                   bool journal);
-
   void findResizeSlacks1();
+  
+public:
```
maliberty, line 519: Consolidate the public & protected sections

src/rsz/src/Resizer.cc
```diff
@@ -2701,38 +2718,98 @@ Resizer::journalSwapPins(Instance *inst, LibertyPort *port1,
...
+    if (network_->direction(pin)->isInput()) {
+      Instance *inst = network_->instance(pin);
+      auto term_port = network_->port(pin);
+      sta_->connectPin(inst, term_port, original_out_net);
+    }
+  }
+  // Final cleanup
+  // sta_->deleteNet(clone_out_net);
+  sta_->deleteInstance(cloned_inst);
+  sta_->graphDelayCalc()->delaysInvalid();
```
maliberty, line 2795: Shouldn't sta do this automatically?

src/rsz/src/Journal.cc
```diff
@@ -0,0 +1,240 @@
...
+UndoBuffer::UndoBuffer(Instance* inst)
+{
+  // Original cell used for resize and the instance
+  buffer_inst_ = inst;
+}
+
+int UndoBuffer::UndoOperation(Resizer *resizer)
+{
+  // Remove the inserted buffer
+  resizer->removeBuffer(const_cast<Instance*>(buffer_inst_));
```
maliberty, line 73: Why is a const_cast needed as the field isn't const?

src/rsz/src/Journal.cc
```diff
@@ -0,0 +1,240 @@
...
+    resizer_ = resizer;
+    logger_ = logger;
+}
+
+void Journal::begin()
+{
+  debugPrint(logger_, RSZ, "journal", 1, "journal begin");
+  while (!journal_stack_.empty()) {
+    journal_stack_.pop();
+  }
```
maliberty, line 148: Why not "journal_stack_ = {};" like Jounal::end() ?

src/rsz/src/Journal.cc
```diff
@@ -0,0 +1,240 @@
...
+  // TODO We need to do something with cloned instances here
+  journal_stack_ = {};
+}
+
+void Journal::swapPins(Instance* inst, LibertyPort* port1,
+                              LibertyPort* port2)
+{
+  debugPrint(logger_, RSZ, "journal", 1, "journal swap pins {} ({}->{})",
+             network()->pathName(inst), port1->name(), port2->name());
+  std::unique_ptr<Undo> element(new UndoPinSwap(inst, port1, port2));
```
maliberty, line 163: Use std::make_unique

src/rsz/src/Resizer.cc
openroadie: I am assuming you mean the delaysInvalid call. Fixed.

src/rsz/test/repair_fanout7.ok
openroadie: General cleanup and restructuring of the code so we can actually switch over at some point.

openroadie:
@maliberty This is finally ready. The next step would be to do some testing of the new undo mode and then migrate to it (and delete all the existing undo code). The big difference between new/old Undo is that the new code actually does use a stack (instead of the old ad hoc approach where certain undo operations can simply not be supported).

src/rsz/test/repair_fanout7.tcl
```diff
@@ -34,5 +34,4 @@ estimate_parasitics -placement
 report_worst_slack -max
 write_verilog_for_eqy repair_fanout7 before "None"
 repair_timing -setup -repair_tns 100 -skip_pin_swap
-run_equivalence_test repair_fanout7 ./Nangate45/work_around_yosys/ "None"
```
maliberty, line 37: Why remove this?

src/rsz/test/repair_fanout7.ok
```diff
@@ -12,9 +12,8 @@ load DFF_X1 CK D 300000 metal3 1000
...
-[INFO RSZ-0045] Inserted 78 buffers, 1 to split loads.
-[INFO RSZ-0041] Resized 527 instances.
-[INFO RSZ-0049] Cloned 7 instances.
+[INFO RSZ-0045] Inserted 40 buffers, 1 to split loads.
+[INFO RSZ-0041] Resized 364 instances.
+[INFO RSZ-0049] Cloned 4 instances.
 [WARNING RSZ-0062] Unable to repair all setup violations.
-Repair timing output passed/skipped equivalence test
-worst slack -3.75
+worst slack -2.84
```
maliberty, line 19: I thought this is off by default - why is there a diff?

src/rsz/src/Journal.hh
```diff
@@ -0,0 +1,173 @@
...
+enum class BufferedNetType;
+using BufferedNetPtr = std::shared_ptr<BufferedNet> ;
+using BufferedNetSeq = vector<BufferedNetPtr>;
+
+class Undo
+{
+ public:
+  // base class for undo elements
+  Undo() = default;
+  virtual int UndoOperation(Resizer *resizer) = 0;
```
maliberty, line 71: This needs a virtual dtor

maliberty, line 71: What does the return value indicate?

src/rsz/src/Journal.cc
```diff
@@ -0,0 +1,238 @@
...
+    logger_->info(RSZ, 97, "Cloned {} instances.", cloned_gate_count_);
+  }
+  }
+
+sta::Network * Journal::network()
+{
+  return resizer_->network_;
+}
+
+dbSta *Journal::sta()
```
maliberty, line 232: What is this used for?

maliberty (changes requested):

src/rsz/test/repair_fanout7.tcl
openroadie: Because it fails on CENTOS7 and I do not know whether it is related to an eqy issue or some other problem.

src/rsz/test/repair_fanout7.ok
openroadie: What is "off my default"

src/rsz/src/Journal.cc
openroadie: Removed for now. Will pit it back when needed.

src/rsz/src/Journal.hh
openroadie: Not much for the moment but we might need success/failure once we try to make this the default.

src/rsz/test/repair_fanout7.ok
openroadie: The diff comes from some cleanup that happened during the implementation of the infrastructure to redo the "undo" operation. The results are better.

maliberty: typo : off by default

maliberty: Does "The results are better." imply you ran a secure CI?

openroadie: No. All it implies is that these particular results are better (-3.75 vs. -2.84). If you think there is a need to run the secure CI for this change then I can try to remember how to do that. LMK.

maliberty:
I went ahead and started a secure CI

openroadie:
Thank you @maliberty . Much appreciated.

maliberty:
aes/ng45:  Error: cts.tcl, 111 ERROR: A problem occurred during equivalence check.
aes_lvt/asap7: [ERROR] finish__timing__wns_percent_delay fail test: -12.204497 >= -10.0

The first needs investigating but the second could be updated.

openroadie:
> aes/ng45: Error: cts.tcl, 111 ERROR: A problem occurred during equivalence check. aes_lvt/asap7: [ERROR] finish__timing__wns_percent_delay fail test: -12.204497 >= -10.0
> 
> The first needs investigating but the second could be updated.

Thank you @maliberty . I agree with you and luckily can run aes/n45 locally. I will investigate and report back. Sorry for the delay in replying (been kinda sick this week).

maliberty:
stale - replaced by odb undo

PR #3898, opened 2023-08-23
ppl: add new movements for pin groups to reduce annealing runtime

eder-matheus:


maliberty:
annealing_constraint6 (tcl) *ERROR* SIGABRT
Please cleanup clang-tidy

src/ppl/src/Netlist.h
```diff
@@ -184,7 +184,7 @@ class Netlist
                     bool order,
                     int group_idx);
   void addIOGroup(const std::vector<int>& pin_group, bool order);
-  const std::vector<PinGroupByIndex>& getIOGroups() { return io_groups_; }
+  std::vector<PinGroupByIndex>& getIOGroups() { return io_groups_; }
```
maliberty, line 187: Why is this returning a non-const reference?  It breaks the encapsulation.

eder-matheus: I overlooked the issue with the encapsulation, but I know how to fix it. My idea was to return a non-const to sort the pin indices according to the edge on which the group is placed. I will create a function in the Netlist class that will perform this sorting to fix the issue.

eder-matheus:
> annealing_constraint6 (tcl) _ERROR_ SIGABRT Please cleanup clang-tidy

I didn't face this when testing on my Mac, but I will check it out in our servers.

maliberty (approved):

PR #3911, opened 2023-08-27
Fix build with fmt 10.1

tobim:
continuation of #2696.

This fixes build failures when fmt-10.1 is used.


maliberty:
Thanks for looking at this.  I would rather see a solution based on https://github.com/fmtlib/fmt/issues/3088 than introduce a bunch of fmt::formatter specializations.

PR #3913, opened 2023-08-27
gui: dont draw if image is null

gadfort:


maliberty (approved):

maliberty:
Does this resolve the problem you were seeing?

gadfort:
@maliberty yes, I still need to make a change to one of our scripts, but I've confirmed that this + that change solves my issue.

maliberty:
I find this odd as it just suppresses drawing entirely.  What is the full issue?

gadfort:
The issue is that when drawing on a machine that hasn't opened OpenROAD in the GUI (CI runners), the shape of the layout window has not been computed yet, so the shape is 0x600 (0 by something), which is invalid and therefore unable to draw and unable to save an image. By skipping the drawing it avoids the warnings about the painter not being valid.
I need to provide an `area` to the save image command to ensure a shape is provided avoiding that the tool attempts to use the 0x600 shape.

Once the settings file is saved (second time you open OpenROAD), a shape is available for all windows and this issue is gone.

maliberty:
It's odd as the public Jenkins spins up workers frequently.  @vvbandeira is there an openroad settings in the base image?

vvbandeira:
@maliberty
I think we only save images for ORFS, and there we have `flow/Makefile: export QT_QPA_PLATFORM ?= offscreen`.
Other than that, nothing special.

PR #3916, opened 2023-08-28
Add ParallelOverlap to LEF58_Spacing property

refaay:


src/odb/src/db/dbTechLayerCutSpacingRule.h
```diff
@@ -73,6 +73,7 @@ struct dbTechLayerCutSpacingRuleFlags
   bool same_mask_ : 1;
   bool except_same_pgnet_ : 1;
   bool side_parallel_overlap_ : 1;
+  bool parallel_overlap_ : 1;
```
rovinski, line 76: need to decrement `spare_bits_` by 1

src/drt/src/io/io.cpp
```diff
@@ -1573,13 +1573,20 @@ void io::Parser::setCutLayerProperties(odb::dbTechLayer* layer,
...
+      case odb::dbTechLayerCutSpacingRule::CutSpacingType::PARALLELOVERLAP: {
+        auto con = make_unique<frLef58CutSpacingConstraint>();
+        con->setCutSpacing(rule->getCutSpacing());
+        con->setCenterToCenter(rule->isCenterToCenter());
+        con->setSameNet(rule->isSameNet());
+        con->setSameMetal(rule->isSameMetal());
+        con->setSameVia(rule->isSameVia());
+        con->setSameCut(rule->isSameCut());
+        con->setSameMask(rule->isSameMask());
+        con->setParallelOverlap(rule->isParallelOverlap());
```
osamahammad21, line 1585: just set to true

src/drt/test/fixture.cpp
```diff
@@ -544,6 +544,26 @@ void Fixture::makeKeepOutZoneRule(frLayerNum layer_num,
   design->getTech()->addUConstraint(std::move(con));
 }
 
+frLef58CutSpacingConstraint*
+Fixture::makeLef58CutSpacingConstraint_parallelOverlap(frLayerNum layer_num,
+                                                       frCoord spacing,
+                                                       bool parallelOverlap)
```
osamahammad21, line 550: why do you need bool parallelOverlap? I am assuming from the name of this function that it should only create PARALLELOVERLAP rules

src/drt/test/gcTest.cpp
```diff
@@ -1128,6 +1128,32 @@ BOOST_AUTO_TEST_CASE(metal_width_via_map)
              Rect(100, 0, 200, 100));
 }
 
+BOOST_AUTO_TEST_CASE(cut_spc_parallel_overlap)
```
osamahammad21, line 1131: Very good. Could you now make it a BOOST_DATA_TEST_CASE with two tests; one should have a violation marker and the other shouldn't?

src/odb/src/db/dbTechLayerCutSpacingRule.cpp
```diff
@@ -1191,6 +1191,20 @@ bool dbTechLayerCutSpacingRule::isSideParallelOverlap() const
...
+  obj->flags_.parallel_overlap_ = parallel_overlap;
+}
+
+bool dbTechLayerCutSpacingRule::isParallelOverlap() const
+{
+  _dbTechLayerCutSpacingRule* obj = (_dbTechLayerCutSpacingRule*) this;
+
+  return obj->flags_.parallel_overlap_;
+}
+
```
osamahammad21, line 1207: not needed. A paralleloverlap rule should have their type as odb::dbTechLayerCutSpacingRule::CutSpacingType::PARALLELOVERLAP. you can get that from getType.

osamahammad21 (changes requested):

src/drt/test/fixture.cpp
osamahammad21: Also you don't need here to create an odb rule. Some rules require a pointer to the odb rule as they use that pointer directly without having to recreate all the rule data. However, this rule doesn't do that.

src/drt/src/io/io.cpp
refaay: Do I only set ParallelOverlap to true? or all of the mentioned four rules?
Because I think at this point we are only sure that parallelOverlap is true.

osamahammad21: just paralleloverlap

maliberty:
What is the status of this PR?

refaay:
> What is the status of this PR?

I finished all the edits suggested.
@osamahammad21 can you review again and approve?

src/odb/src/db/dbTechLayerCutSpacingRule.h
osamahammad21, line 76: This should be removed now

osamahammad21 (changes requested):

src/drt/src/db/tech/frLayer.h
```diff
@@ -545,9 +545,8 @@ class frLayer
   {
     if (samenet) {
       return (!lef58CutSpacingSamenetConstraints.empty());
-    } else {
-      return (!lef58CutSpacingConstraints.empty());
     }
+    return (!lef58CutSpacingConstraints.empty());
```
QuantamHD, line 549: No need to put this in parens.

src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -1158,6 +1163,24 @@ void FlexDRWorker::modAdjCutSpacingCost_fixedObj(const frDesign* design,
...
+            if (con->isParallelOverlap()) {
+              if (prl > 0 && currDistSquare < reqDistSquare) {
+                hasViol = true;
+              }
+            }
+            if (hasViol) {
+              break;
+            }
+          }
+        }
```
osamahammad21, line 1183: This neglects many options of the rule which could lead to false positive costs. For example if the rule specifies a restriction on the upper/lower cut layer not the same layer. I would prefer if we remove the maze cost updates from this PR and have another PR that handles all such cases.

osamahammad21 (changes requested):

osamahammad21 (approved):

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -2299,6 +2299,101 @@ void FlexGCWorker::Impl::checkCutSpacing_spc_diff_layer(
...
+  frSquaredDistance reqSpcValSquare
+      = checkLef58CutSpacing_spc_getReqSpcVal(rect1, rect2, con);
+  reqSpcValSquare *= reqSpcValSquare;
+
+  gtl::point_data<frCoord> center1, center2;
+  gtl::center(center1, *rect1);
+  gtl::center(center2, *rect2);
+  frSquaredDistance distSquare = 0;
+  if (con->isCenterToCenter()) {
+    distSquare = gtl::distance_squared(center1, center2);
```
maliberty, line 2357: move the center code inside the isCenterToCenter block

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -2299,6 +2299,101 @@ void FlexGCWorker::Impl::checkCutSpacing_spc_diff_layer(
...
+  } else {
+    distSquare = gtl::square_euclidean_distance(*rect1, *rect2);
+  }
+  if (distSquare >= reqSpcValSquare) {
+    return;
+  }
+  // no violation if fixed shapes
+  if (rect1->isFixed() && rect2->isFixed()) {
+    return;
+  }
```
maliberty, line 2367: move this check to the top as there is no reason to do all the other work and then reject it here.

maliberty (changes requested):

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -2299,6 +2299,101 @@ void FlexGCWorker::Impl::checkCutSpacing_spc_diff_layer(
...
+  auto net2 = rect2->getNet();
+
+  auto prlX = gtl::delta(markerRect, gtl::HORIZONTAL);
+  auto prlY = gtl::delta(markerRect, gtl::VERTICAL);
+  frCoord prl = std::max(prlX, prlY);
+
+  // skip if no parallel overlap
+  if (prl <= 0) {
+    return;
+    // skip if parallel overlap but shares the same above/below metal
```
rovinski, line 2318: I'm mildly confused which piece of code this comment belongs to.

refaay: I moved the comment to the correct location

maliberty (approved):

Issue #3909, opened 2023-08-27
Update io placement to have an option to output the placement to a .tcl file


oharboe:
### Description

I'm using routing by abutment for two different macros that need to have the same pin-layout.

If I use `PLACE_PINS_ARGS=-annealing`, I get different pin layout for these two macros.

### Suggested Solution

Add option to io placement to write out the pin-placement to a .tcl file. This .tcl file can then be fed into the second macro and pin layout will be identical.

Furthermore, it would be great to have the pins in this file grouped by edges (generate a comment to indicate edge). This would allow using a combination where only one edge is copy and pasted from an io placement annealing run.

Another use-case for this .tcl file is to write out best found solutions for long running io placement efforts. The user can then hand-tweak the best result.

There should be some sort of output for long running io placement efforts.

Ideally the io placement would print out a recommendation to improve convergence speed, like "reduce number of pins on left edge by 10%", or something.

eder-matheus:
@oharboe If it is guaranteed that the two different macros have the same track positions, we can use the `place_pin` command for each IO pin of the design. So the output .tcl file would be a list of calls for the `place_pin` command, like this:

```
place_pin -pin_name pin0 -layer layer -location "x0 y0"
place_pin -pin_name pin1 -layer layer -location "x1 y1"
...
place_pin -pin_name pinN -layer layer -location "xN yN"
```

We could easily group the pins by edge in this format. The only problem is that this file can be huge with one line per pin, but I think this is not a blocker.

About the output for long runtime, do you think an `-timeout` option could help? The annealing run would end after reaching the timeout, and the current result would be the final assignment, followed by a warning message and the suggestion of a new configuration for the constraints.

oharboe:
> @oharboe If it is guaranteed that the two different macros have the same track positions, we can use the `place_pin` command for each IO pin of the design. So the output .tcl file would be a list of calls for the `place_pin` command, like this:
> 
> ```
> place_pin -pin_name pin0 -layer layer -location "x0 y0"
> place_pin -pin_name pin1 -layer layer -location "x1 y1"
> ...
> place_pin -pin_name pinN -layer layer -location "xN yN"
> ```
> 
> We could easily group the pins by edge in this format.

:+1: 

This would allow me to cut and paste some pins for my use-case and also to use the annealing.

> The only problem is that this file can be huge with one line per pin, but I think this is not a blocker.

[...]

PR #3923, opened 2023-08-29
ppl: add option to write pin placement to tcl file

eder-matheus:
Partially fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3909


eder-matheus:
@oharboe FYI

src/ppl/test/top_layer6.ok
```diff
@@ -7,7 +7,7 @@
 [INFO ODB-0130]     Created 54 pins.
 [INFO ODB-0131]     Created 88 components and 422 component-terminals.
 [INFO ODB-0133]     Created 54 nets and 88 connections.
-[INFO PPL-0070] Pin clk placed at (4um, 4um).
+[INFO PPL-0070] Pin clk placed at (4.8um, 4.8um).
```
maliberty, line 10: Is this intentional?   Similarly with other existing tests with diffs.

src/ppl/src/IOPlacer.cpp
```diff
@@ -600,8 +604,8 @@ bool IOPlacer::checkBlocked(Edge edge, int pos, int layer)
     if (blocked_interval.getLayer() == -1
         || blocked_interval.getLayer() == layer) {
       if (blocked_interval.getEdge() == edge
-          && pos >= blocked_interval.getBegin()
-          && pos <= blocked_interval.getEnd()) {
+          && pos > blocked_interval.getBegin()
+          && pos < blocked_interval.getEnd()) {
```
maliberty, line 608: how does this relate to writing the tcl?

src/ppl/test/top_layer6.ok
eder-matheus: Yes. I fixed how this was being printed to use the `dbuToMicrons` function. The previous was incorrect.

src/ppl/src/IOPlacer.cpp
eder-matheus: Not directly to writing the tcl, but when reading it. The pin placement when loading the tcl file differed from the tool run result. It was happening because of this check. When placing an individual pin, the tool creates a blocked region around it, and this check changes the position of the pins placed near the previous one. With this update, it allows touching the blocked area.

maliberty:
clang-format needed

maliberty (approved):

src/ppl/src/IOPlacer.tcl
```diff
@@ -359,13 +359,14 @@ sta::define_cmd_args "place_pin" {[-pin_name pin_name]\
...
-                                  [-force_to_die_boundary]
+                                  [-force_to_die_boundary]\
+                                  [-placed_status]
 }
 
 proc place_pin { args } {
   sta::parse_key_args "place_pin" args \
   keys {-pin_name -layer -location -pin_size}\
-  flags {-force_to_die_boundary}
+  flags {-force_to_die_boundary -placed_status}
```
vvbandeira, line 369: @eder-matheus
Should this flag be documented? Can you make a PR or provide the info to @luarss ?

eder-matheus: It will not be necessary. I'm about to remove this flag since it isn't working as I expected and we will not need it anymore.

PR #3925, opened 2023-08-29
gui: allow layout view to be dragged while holding middle mouse button

### LINKED ISSUE #3840: Add ability to drag the view with the mouse
It would make moving around the design a lot easier.

The arrow keys is a bit clunky compared to dragging the view with the mouse.

_Originally posted by @oharboe in https://github.com/The-OpenROAD-Project/OpenROAD/discussions/3812_

AcKoucher:
Resolves #3840

Layout can be dragged by holding middle mouse button like in KLayout.

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -133,6 +133,7 @@ LayoutViewer::LayoutViewer(
       min_depth_(0),
       max_depth_(99),
       rubber_band_showing_(false),
+      is_view_dragging_(false),
       gui_(gui),
       usingDBU_(usingDBU),
       showRulerAsEuclidian_(showRulerAsEuclidian),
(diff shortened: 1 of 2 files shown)


maliberty:
I observed some strange scrolling if I first create a ruler and then immediately middle mouse drag.  I think those two features are interacting in an odd way.

oharboe:
I tried it on a Dell laptop. Use two fingers on the touchpad and drag the view around :+1:

AcKoucher:
I pushed a change to prevent dragging when the rubber band is showing (used KLayout as reference)

@maliberty I couldn't notice any strange scrolling when creating a ruler and then immediately middle mouse dragging. Could you please describe the weird scrolling?

oharboe:
@AcKoucher Works great! :+1:

maliberty:
I demo'ed the ruler problem to @AcKoucher yesterday so this will hold until it can be resolved.

maliberty (approved):

maliberty:
I realized the problem I was seeing could happen even without the ruler.  My middle mouse wheel can also tilt left/right and I was getting a combination of hold and tilt causing the left scrolling.

PR #3930, opened 2023-08-31
pad: add error checking for master cell

gadfort:
Adds:
- check for master cell mismatch on placePad
- unit test for master cell mismatch and unit test for no master in command


maliberty:
CI errors..

src/pad/test/place_pad_no_master.ok
maliberty, line 1: shouldn't there be an error here?

maliberty (approved):

PR #3934, opened 2023-09-01
dpl: Allows CORE type cells to be used as fillers more seamlessly

QuantamHD:
This PR repurposes a user_flag in dbInst to store whether or not
that inst was placed by the filler engine. If it was then downstream
tools that operate on fillers will recognize these cells as fillers
regardless of their LEF type.

There's probably other places we should pipe this, but for now it's
integrated into check_placement and remove_fillers.

### DIFF
--- a/src/dpl/src/CheckPlacement.cpp
+++ b/src/dpl/src/CheckPlacement.cpp
@@ -457,6 +457,12 @@ bool Opendp::isOverlapPadded(const Cell* cell1, const Cell* cell2) const
 bool Opendp::isCrWtBlClass(const Cell* cell) const
 {
   dbMasterType type = cell->db_inst_->getMaster()->getType();
+  dbInst* inst = cell->db_inst_;
+
+  if (inst->isPlacedAsFiller()) {
+    return false;
+  }
+
   // Use switch so if new types are added we get a compiler warning.
   switch (type) {
     case dbMasterType::CORE:
--- a/src/dpl/src/FillerPlacement.cpp
+++ b/src/dpl/src/FillerPlacement.cpp
@@ -207,7 +207,7 @@ void Opendp::removeFillers()
 bool Opendp::isFiller(odb::dbInst* db_inst)
 {
   dbMaster* db_master = db_inst->getMaster();
-  
+
   if (db_inst->isPlacedAsFiller()) {
     return true;
   }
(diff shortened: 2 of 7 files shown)


gadfort:
Wouldn't it be better to be able specify the user type of the masters instead? If you read in a def from a different tool it's a lot easier to specify the masters that are fillers than to try and replicate this information in the insts

maliberty:
I agree with @gadfort .  The only confusion I'm aware of is if you use a decap cell as a filler.  Perhaps we should just have a (sub)type for that.

QuantamHD:
I'm not too concerned about tool inter-op, I think this is a benefit of an all OpenROAD flow. I'm trying to achieve two things in this PR that I think the other two solutions proposed don't quite address.

1. Yes, it's definitely possible to specify the master type, but it's mildly annoying this PR is trying to make that more seamless.
2. The DECAP is one common occurrence, but if the user decides to use some other fill type that's not strictly a filler like a NAND gate that solution won't work

By keeping track of which cells were placed by the filler we know exactly which ones we need to remove in `remove_fillers`, and the user only needs to specify their fill cell choice list instead of relying on the upstream PDK following standards. Overall I think it's a good balance between usability and internal complexity.

maliberty:
Filling with a nandgate is nonsensical (leakage for no benefit).

We don't use remove_fillers in the flow so I'm not sure why it even needed.  What problem are you solving here?

gadfort:
@maliberty if we want to be able to do post global routing timing repairs (or diode insertion) you'll need to use remove_fillers to clear the space.

@QuantamHD using logic gates would be highly unusual, are you trying to insert FIB cells? Those really should not be handled by fillers.

maliberty:
@gadfort when we have post-grt repair we'll just delay filler insertion until after.  There is no need to fill before you have finished optimizing.

QuantamHD:
My fillers are just marked CORE, and I've seen that in enough technology nodes that I thought we should be tolerant to it

maliberty:
An alternate suggestion:

In filler_placement, when you see a CORE cell, with no non-power pins, just change its type to CORE SPACER.

QuantamHD:
@maliberty I'm personally fine with that solution, and what I initially considered, but it felt like it could have more side effects than intended. Which is why I went with the keep track of the filler cells approach. If you're fine with just marking those cells as SPACER I'm happy to make the change.

maliberty:
I prefer it.  @gadfort any opinion?

maliberty:
Since @gadfort had no comment please go ahead with my suggestion

maliberty:
stale - reopen if still relevant and update

Issue #3936, opened 2023-09-02
tapcell.tcl runs 'forever', [INFO ODB-0303] The initial 119565 rows (-1535139212 sites) were cut with


oharboe:
### Describe the bug

Use `make tapcell_issue` [archive](https://drive.google.com/file/d/1PjqyemlycfRku4lQeTZuIaW5oi7kYzQa/view?usp=sharing)

```
./run-me-megaboom-asap7-base.sh
OpenROAD v2.0-10090-g067d1f4be 
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
[INFO-FLOW] Tap and End Cap cell insertion
[INFO-FLOW]   TAP Cell          : TAPCELL_ASAP7_75t_R
[INFO-FLOW]   ENDCAP Cell       : TAPCELL_ASAP7_75t_R
[INFO-FLOW]   Halo Around Macro : 2 2
[INFO-FLOW]   TAP Cell Distance : 25
[INFO ODB-0303] The initial 119565 rows (-1535139212 sites) were cut with 18 shapes for a total of 153977 rows (-1894716766 sites).
[INFO TAP-0004] Inserted 239130 endcaps.
```

I set CORE_UTILIZATION=0.5 as a first guess and I got this result, which was surprising... The die area ended up very large... 1024mm^2 or so...

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/9ed86262-bb12-4af3-a91a-36ca869b049f)

Lower right corner:

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/51553905-5ec9-4d2c-b3b2-5e85448f2a7d)

### Expected Behavior

complete in a reasonable amount of time

### Environment

```shell
OpenROAD v2.0-10090-g067d1f4be
```

### To Reproduce

See above.

vijayank88:
@oharboe 
The core area seems very high.
```
>>> ord::get_core_area
2.052 2.16 32284.764 32284.71
```
Is that created by automatically or manual size passed?

I have seen such cases, if users give big die size more than 10k micron, tapcell insertion takes too long.

oharboe:
> `CORE_UTILIZATION=0.5` this will increase die size. Normally `30-60` reasonable for automatic die size calculation.

[...]

PR #3941, opened 2023-09-04
odb: add error message for overflow during cut_rows and fix overflow in message

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3936


src/odb/src/zutil/util.cpp
```diff
@@ -165,6 +165,16 @@ int makeSiteLoc(int x, double site_width, bool at_left_from_macro, int offset)
   return site_x1 * site_width + offset;
 }
 
+template <typename T>
+bool hasOverflow(T a, T b)
+{
+  if ((b > 0 && a > std::numeric_limits<T>::max() - b)
+      || (b < 0 && a < std::numeric_limits<T>::min() - b)) {
```
maliberty, line 172: You want lowest not min (for floating point min is the smallest positive value).

maliberty:
This feels too targeted as a fix. I'm sure there are many places that will fail on such an unreasonably large design. Also the message doesn't offer much insight into the problem.

I suggest in initialize_floorplan we error on any block with more than 2^32 sites.

oharboe:
> This feels too targeted as a fix. I'm sure there are many places that will fail on such an unreasonably large design. Also the message doesn't offer much insight into the problem.
> 
> I suggest in initialize_floorplan we error on any block with more than 2^32 sites.

2^31 sites? Otherwise it's easy to run into signed integer overflow?

How big a design would this be equivalent to, roughly?

maliberty:
2^31 is reasonable.   In asap7:
```
SITE asap7sc7p5t
 SIZE 0.054 BY 0.270 ;
```

So 2^31 * .054 * .27 = 31310311 um^2 [about 5.6mm x 5.6mm]

oharboe:
> 2^31 is reasonable. In asap7:
> 
> ```
> SITE asap7sc7p5t
>  SIZE 0.054 BY 0.270 ;
> ```
> 
> So 2^31 * .054 * .27 = 31310311 um^2 [about 5.6mm x 5.6mm]

This is on the small side when doing exploration. I would rather not have such a hard limit. While OpenROAD can't be used at these scales, things can be learned by exploring designs at these scales.

Also, exploration involves automation, so getting *some* result at larger sizes can help create a gradient when doing automatic searches.

maliberty:
> Also, exploration involves automation, so getting some result at larger sizes can help create a gradient when doing automatic searches.

Then this overflow will not work either.   We would need to change site count to a int64_t instead.  I expect it will be a pipe cleaning exercise if you try to push such a large design through the rest of the flow.

src/odb/src/zutil/util.cpp
```diff
@@ -177,10 +187,16 @@ void cutRows(dbBlock* block,
   }
   auto rows = block->getRows();
   const int initial_rows_count = rows.size();
-  const int initial_sites_count
-      = std::accumulate(rows.begin(), rows.end(), 0, [](int sum, dbRow* row) {
-          return sum + row->getSiteCount();
-        });
+  const std::int64_t initial_sites_count = std::accumulate(
```
eder-matheus, line 190: @maliberty I've updated the data type to int64_t, but I still see the overflow in the message INFO-303. Debugging it, I've noticed that the `sum` variable overflows when it reaches the max limit for an `int`. I don't get why it happens, since I'm using int64_t for this variable too.

maliberty: Use int64_t(0), [](int64_t sum, ....

eder-matheus:
@maliberty This is ready to merge, I believe.

src/odb/src/zutil/util.cpp
```diff
@@ -221,10 +240,19 @@ void cutRows(dbBlock* block,
...
-  const int final_sites_count
-      = std::accumulate(rows.begin(), rows.end(), 0, [](int sum, dbRow* row) {
-          return sum + row->getSiteCount();
-        });
+  const std::int64_t final_sites_count = std::accumulate(
+      rows.begin(),
+      rows.end(),
+      (std::int64_t) 0,
+      [&](std::int64_t sum, dbRow* row) {
+        if (hasOverflow(sum, (std::int64_t) row->getSiteCount())) {
```
maliberty, line 248: We should not need the overflow check when using an int64_t

eder-matheus: I've removed the check, but kept the `hasOverflow` function in case we need it in the future.

maliberty (approved):

Issue #3867, opened 2023-08-16
No nets under "Sources:" in global routing congestion progress reports


oharboe:
To reproduce:

1. Comment out GLOBAL_ROUTE_ARGS, see below:
2. Run `make DESIGN_CONFIG=designs/sky130hd/microwatt/config.mk results/sky130hd/microwatt/base/5_1_grt.odb`
3. *While* global route is running, start a new shell, run `ls reports/sky130hd/microwatt/base/congestion*.rpt`. When you see .rpt files appearing, run `make DESIGN_CONFIG=designs/sky130hd/microwatt/config.mk gui_cts` and load for instance `reports/sky130hd/microwatt/base/congestion-10.rpt`
4. Look at DRC errors, there's no information about which nets are congested.

```
$ git diff
diff --git a/flow/designs/sky130hd/microwatt/config.mk b/flow/designs/sky130hd/microwatt/config.mk
index 4b4eb79d..90745a5d 100644
--- a/flow/designs/sky130hd/microwatt/config.mk
+++ b/flow/designs/sky130hd/microwatt/config.mk
@@ -31,4 +31,4 @@ export CTS_CLUSTER_SIZE = 30
 
 export export SETUP_SLACK_MARGIN = 0.2
 
-export GLOBAL_ROUTE_ARGS=-congestion_iterations 100 -verbose
+#export GLOBAL_ROUTE_ARGS=-congestion_iterations 100 -verbose
```

_Originally posted by @oharboe in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/1338#discussioncomment-6688226_

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/48772425-7b3c-4da6-b22a-ed922a76342d)

What I expected was to see nets in the progress reports, like I get in the final report:

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/68c110c7-cf3e-443f-b109-0af706890a3e)

rovinski:
Resolved?

PR #3942, opened 2023-09-04
grt: iterating all the nets when generating the progress report

luis201420:
Iterating all the nets when checking the nets that cross the congested ggrids
Finxing git issue #3867


eder-matheus (changes requested):
Please, add a unit test for it. Perhaps you can use the congestion7 unit test as a base, and generate at least two intermediary congestion files.

eder-matheus (approved):

maliberty (approved):

PR #3348, opened 2023-05-17
openroad: add more GUI timers

oharboe:
With this change, all rendering time is pretty much accounted for, previously there could be significant discrepancies between what was logged and the total rendering time.

<!DOCTYPE html>

<html>
<head>
	
	
	
</head>

<body>

[DEBUG | GUI-draw] | manufacturing | grid |   | 2.62222E-04 | sec
-- | -- | -- | -- | -- | -- | --
[DEBUG | GUI-draw] | inst | search |   | 4.86770E-01 | sec
[DEBUG | GUI-draw] | inst | outline | render | 1.13920E+00 | sec
[DEBUG | GUI-draw] | blockages |   |   | 9.15498E-04 | sec
[DEBUG | GUI-draw] | layer | V0 | render | 9.15700E-06 | sec
[DEBUG | GUI-draw] | layer | M1 | render | 2.43998E-01 | sec
[DEBUG | GUI-draw] | layer | V1 | render | 6.62100E-06 | sec
[DEBUG | GUI-draw] | layer | M2 | render | 3.04803E-01 | sec
[DEBUG | GUI-draw] | layer | V2 | render | 7.07200E-06 | sec
[DEBUG | GUI-draw] | layer | M3 | render | 2.07964E-01 | sec
[DEBUG | GUI-draw] | layer | V3 | render | 6.77800E-06 | sec
[DEBUG | GUI-draw] | layer | M4 | render | 1.61031E-01 | sec
[DEBUG | GUI-draw] | layer | V4 | render | 6.84600E-06 | sec
[DEBUG | GUI-draw] | layer | M5 | render | 1.41925E-01 | sec
[DEBUG | GUI-draw] | layer | V5 | render | 4.94600E-06 | sec
[DEBUG | GUI-draw] | layer | M6 | render | 1.36262E-01 | sec
[DEBUG | GUI-draw] | layer | V6 | render | 5.25500E-06 | sec
[DEBUG | GUI-draw] | layer | M7 | render | 1.28346E-01 | sec
[DEBUG | GUI-draw] | layer | V7 | render | 4.66900E-06 | sec
[DEBUG | GUI-draw] | layer | M8 | render | 1.28116E-01 | sec
[DEBUG | GUI-draw] | layer | V8 | render | 5.02000E-06 | sec
[DEBUG | GUI-draw] | layer | M9 | render | 1.28121E-01 | sec
[DEBUG | GUI-draw] | layer | V9 | render | 4.80500E-06 | sec
[...]

PR #3944, opened 2023-09-05
[DO NOT MERGE] gui: multithreaded renderer

Iyury1:
This PR shows the approach I am taking to implement multithreading inside renderThread.cpp. An approach was discussed in  #3359 to have individual layers drawn as separate images, and then collate these images after the threads have finished.

However as shown by the GUI timers in #3348, there are other functions that would benefit from multithreading besides drawLayers() that cannot necessarily be divided by layer, such as drawInstanceOutlines().

I borrowed from @oharboe in #3824 to divide the layout by subrectangle, but then used these subrectangles to divide the inst_range into sub instance ranges.

I created two additional functions drawConcurrent() and drawBlockConcurrent(). The idea is a thread pool is created inside drawConcurrent() for collating images. The thread pool can then be reused by each function inside drawBlockConcurrent(). I also created a second thread pool just for the purpose of returning sub insts vectors from Search::searchInsts().

I still have several changes to make, and I need to determine if there is any benefit performance from this. This is more or less an update on my progress.

Also, can anyone tell me what is the purpose of the GuiPainter class?

### DIFF
--- a/src/gui/src/renderThread.cpp
+++ b/src/gui/src/renderThread.cpp
@@ -117,17 +117,21 @@ void RenderThread::run()
     highlighted.swap(highlighted_);
     rulers.swap(rulers_);
     mutex_.unlock();
+
+
     QImage image(draw_bounds.width(),
-                 draw_bounds.height(),
-                 QImage::Format_ARGB32_Premultiplied);
+                      draw_bounds.height(),
+                      QImage::Format_ARGB32_Premultiplied);
+    image.fill(Qt::transparent);
+
     // drawing can be interrupted by setting restart_
-    draw(image,
-         draw_bounds,
-         selected,
-         highlighted,
-         rulers,
-         1.0,
-         Qt::transparent);
+    drawConcurrent(image,
+                   draw_bounds,
+                   selected,
+                   highlighted,
+                   rulers,
+                   1.0,
+                   Qt::transparent);
     if (!restart_) {
       emit done(image, draw_bounds);
     }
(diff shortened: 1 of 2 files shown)


oharboe:
@Iyury1 Any numbers on rendering speeds before and after?

Iyury1:
> 

I will check later this evening

maliberty:
GuiPainter is the implementation of the Painter API.  The general idea is that clients outside of the gui/ directory should have no Qt dependency.  They draw on the window using the abstract Painter.

src/gui/src/renderThread.cpp
```diff
@@ -979,6 +1045,176 @@ void RenderThread::drawLayer(QPainter* painter,
...
+              }
+              if (viewer_->options_->isInstanceVisible(inst)) {
+                subInsts.push_back(inst);
+              }
+            }
+            return subInsts;
+          }));
+    }
+    for (auto& future : inst_futures) {
+      future.waitForFinished();
```
maliberty, line 1123: Is there a need to synchronize here or could the subInsts flow directly into the drawInstanceOutlines from the next loop using then()?

src/sta
maliberty, line 1: This seems unrelated

maliberty:
Thanks very much for working on this.  I'm very interested to see how the performance turns out.

Iyury1:
It appears that the inst search time was halved and the time taken drawInstanceOutlines was reduced by 400 us in my environment.

However, the time for all other functions that are still running single threaded was increased. So the overall render time increased. I will proceed with adding multithreading to the other functions and keep testing.

oharboe:
> It appears that the inst search time was halved and the time taken drawInstanceOutlines was reduced by 400 us in my environment.
> 
> However, the time for all other functions that are still running single threaded was increased. So the overall render time increased. I will proceed with adding multithreading to the other functions and keep testing.

@maliberty Can instance search be made multithreaded?

Iyury1:
Actually I am not sure yet if there is any benefit. It seems to be inconsistent depending on the GUI action. So far I just opened the nangate45 GCD, enabled debug logging, clicked once outside of the layout, zoomed in one level, and pressed the up arrow once.

Here is the log output. Honestly its not looking great so far.

| inst outline render (parallel) |  inst outline render (normal) | inst search (parallel) | inst search (normal) |
|--------------------------------|-------------------------------|------------------------|----------------------|
|  0.0135371 sec                 | 0.01126 sec                   | 0.000615308 sec        | 0.000628172 sec      |
| 1.7742e-05 sec                 | 1.458e-05 sec                 | 2.0134e-05 sec         | 1.6157e-05 sec       |
| 0.0171375 sec                  | 0.0182989 sec                 | 0.000590983 sec        | 0.000383854 sec      |
| 0.0177183 sec                  | 0.0174756 sec                 | 0.000459324 sec        | 0.000993236 sec      |

oharboe:
@Iyury1 nangate45 GCD is much, much to small to test rendering speed on. You need something that takes seconds to render... There's no point in speeding up nangate45 gcd, it is instantanenous. :-)

maliberty:
> Can instance search be made multithreaded?
@oharboe he is doing that in this PR by dividing the area search up into multiple queries.  The underlying boost library doesn't have any internal MT.

oharboe:
> > Can instance search be made multithreaded?
> > @oharboe he is doing that in this PR by dividing the area search up into multiple queries.  The underlying boost library doesn't have any internal MT.

I found in my experiment that searching a small rectangle took as long as a big rectangle, so something else is needed to speed up search than splitting searches into rectangles, one peer thread...

src/gui/src/renderThread.cpp
```diff
@@ -979,6 +1042,224 @@ void RenderThread::drawLayer(QPainter* painter,
...
+  std::vector<dbInst*> insts;
+  std::vector<std::vector<dbInst*>> sub_insts;
+  insts.reserve(10000);
+  {
+    QVector<QFuture<std::vector<dbInst*>>> inst_futures;
+    for (const auto& subRect : subRects) {
+      const Rect dbu_subrect = viewer_->screenToDBU(subRect);
+      inst_futures.push_back(
+          QtConcurrent::run([=]() {
+            auto sub_inst_range = viewer_->search_.searchInsts(block,
```
oharboe, line 1095: I have not studied this PR, I just want to say that I found the searching for triangles to be what dominated performance.

Any ideas on how this part can be parallelized?

I think @maliberty said something to the effect that making rtree run with multithreading is "not easy"...

PR #3961, opened 2023-09-07
Add checkers for Block Term off-grid pin

refaay:


rovinski:
Seems like mostly duplicated code. I would recommend factoring it out into a function.

maliberty:
This still has CI errors.  Do you have an ETA to complete?

refaay:
> This still has CI errors. Do you have an ETA to complete?

This should be completed by the next weekly meeting to have time for any reviews and further edits.

The change I pushed now fixed the warnings issue locally.

For the error "grt/test/results/pin_access2-tcl.log: "[ERROR DRT-0416] Term clk contains offgrid pin shape. Pin shape ( 100010 299647 ) ( 100150 300130 ) is not a multiple of the manufacturing grid 5.",
we need to fix pin_access2.tcl in the CI by making the pin shape divisible by 5.

maliberty:
You can fix pin_access2 by correcting the DEF

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -592,148 +592,222 @@ inline void getTrackLocs(bool isHorzTracks,
...
+              grid);
+        } else {
+          logger_->error(DRT,
+                         420,
+                         "BTerm {} contains offgrid pin shape. Polygon point "
+                         "{} is not a multiple of the manufacturing grid {}.",
+                         term_name,
+                         pt,
+                         grid);
+        }
```
osamahammad21, line 689: replace ITerm and BTerm with just "Term" like in error 416

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -592,148 +592,222 @@ inline void getTrackLocs(bool isHorzTracks,
   }
 }
 
+void io::Parser::checkFig(frPinFig* uFig,
+                          const frString& term_name,
+                          bool is_iTerm,
+                          const frString& inst_name,
```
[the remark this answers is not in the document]
refaay: rror "logger_->error(
DRT,
417,
"Term {} of {} contains offgrid pin shape. Polygon point "
"{} is not a multiple of the manufacturing grid {}.",
term_name,
inst_name,
pt,
grid);" prints both term and inst names for ITerm
I can make it print term name only for ITerm like BTerm, what do you think?

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -592,138 +592,191 @@ inline void getTrackLocs(bool isHorzTracks,
...
+                         pt,
+                         grid);
+        } else {
+          logger_->error(DRT,
+                         420,
+                         "Term {} contains offgrid pin shape. Polygon point "
+                         "{} is not a multiple of the manufacturing grid {}.",
+                         term_name,
+                         pt,
+                         grid);
```
osamahammad21, line 681: Combine into one error

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -592,138 +592,191 @@ inline void getTrackLocs(bool isHorzTracks,
   }
 }
 
+void io::Parser::checkFig(frPinFig* uFig,
+                          const frString& term_name,
+                          bool is_iTerm,
+                          const dbTransform& xform,
```
osamahammad21, line 598: is_iTerm is not needed. Just apply the transformation directly. if it's a bterm, just pass a default dbTransform which should not change the shape.

osamahammad21 (changes requested):

PR #3969, opened 2023-09-09
odb: def parsing and writing of MASK property on VIAs

gadfort:


maliberty:
def_parser (tcl) *ERROR*

src/odb/test/data/design58.def
maliberty, line 1: is there a test for defout?

src/odb/src/db/dbWireCodec.cpp
```diff
@@ -642,6 +642,32 @@ void dbWireEncoder::clearColor()
...
+  for (const auto color : {bottom_color, cut_color, top_color}) {
+    if (color > 3) {
+      utl::Logger* logger = _wire->getImpl()->getLogger();
+      logger->error(
+          utl::ODB, 1103, "Mask color: {}, but must be between 0 and 3", color);
+    }
+  }
+
+  // encode as XX BB CC TT
+  const uint8_t mask_color = bottom_color << 4 | cut_color << 2 | top_color;
```
QuantamHD, line 660: You should just make a struct with bit fields, and then cast it to uint8 in the WOP. It'll save you from having to do the bit twiddling manually

gadfort: I would be all for that, but I haven't tried that before and a quick stab at it didn't work, so I'll probably need some help as to how to do that.

src/odb/test/data/design58.def
gadfort: added a test for `write_def58`

src/odb/src/db/dbWireCodec.cpp
maliberty: @QuantamHD I'm inclined to merge and you can PR your idea as an improvement.

maliberty (approved):

PR #3971, opened 2023-09-10
[Docs] Fix prebuilt instructions and comments

luarss:
Addresses the following comments @vvbandeira @dralabeing 

1. The OpenROAD installation does not currently link to pre-built binaries. This is only shown for ORFS. This should be added here as well: https://openroad.readthedocs.io/en/latest/user/Build.html
This is also missing the OS information as in ORFS. 
2. The link from OpenROAD to ORFS lands on this page : https://openroad.readthedocs.io/en/latest/user/ORFS.html.  I think this can be removed since it is redirected to another page  : https://openroad-flow-scripts.readthedocs.io/en/latest/.
3. There is a redundant/recursive link reference the readme from the readme landing here: 81c8ee17-ef2e-45bf-a0cd-351f8787aca2.png
Remove the line "Documentation is also...."

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,6 @@ flowchart LR
 
 ```
 
-Documentation is also available [here](https://openroad.readthedocs.io/en/latest/main/README.html).
 
 ## OpenROAD Mission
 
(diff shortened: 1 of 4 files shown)


dralabeing (approved):
Looks good.

PR #3973, opened 2023-09-12
mpl2: add final pass to improve macro orientation

AcKoucher:
Currently, in the simulated annealing for macro placement, we model the macros as a bundled pin located in their center. This causes the SA to not properly see the orientation when we calculate the wire length.

Due to the fact that the part of the code which handles the creation of the bundled nets is too abstracted, it was better to, before actually digging into conceiving another solution, creating a pass in the final stage of HierRTLMP to adjust macro orientation. This PR is to create this new pass.

I chose to deal directly with odb objects, so the function that handles the orientation adjustment is called after we update odb with the correct macro location.

The orientation adjustment is being made by computing the wire length after mpl2 finishes based on: other macros' pins, std cells (we now create a temporary location for them based on the location of the physical clusters)  and anything else which isPlaced().

In order to snap the macros after the orientation adjustment, I refactored the function that updates odb in mpl2 and used it to avoid code duplication.


stefanottili:
> I'm only considering the connections between macros.
> We can also consider the connections between macros and IO pins if desired.

In which scenario do macro to macro or macro to io connections matter ?
Unless it has built in scan muxing, every ram/rom will be surrounded by boundary scan and/or access logic, no ?

maliberty:
@stefanottili no, you can look at asap7/mock-array which connects by abutment.  Not every macro is a ram/rom.  Ditto for FPGAs.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -5787,6 +5789,83 @@ void HierRTLMP::FDPlacement(std::vector<Rect>& blocks,
...
+        }
+
+        if (net_iterm->getInst()->isBlock()) {
+          const float x1 = dbuToMicron(iterm->getBBox().xCenter(), dbu_);
+          const float y1 = dbuToMicron(iterm->getBBox().yCenter(), dbu_);
+          const float x2 = dbuToMicron(net_iterm->getBBox().xCenter(), dbu_);
+          const float y2 = dbuToMicron(net_iterm->getBBox().yCenter(), dbu_);
+
+          wirelength += (std::abs(x2 - x1) + std::abs(y2 - y1));
+        }
```
maliberty, line 5817: This will tend to overestimate the wire length.  I think the hpwl would be better estimator.

maliberty:
I think the approach is general good.  It could be sped up but I'm not sure its worth the extra complexity.

src/mpl2/src/object.cpp
```diff
@@ -1033,7 +1033,7 @@ void HardMacro::updateDb(float pitch_x, float pitch_y, odb::dbBlock* block)
 
   inst_->setOrigin(snap_origin_x, snap_origin_y);
   inst_->setOrient(orientation_);
-  inst_->setPlacementStatus(odb::dbPlacementStatus::LOCKED);
+  inst_->setPlacementStatus(odb::dbPlacementStatus::PLACED);
```
maliberty, line 1036: Macros should be locked as we don't want global placement to move them.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -5914,6 +5918,133 @@ void HierRTLMP::FDPlacement(std::vector<Rect>& blocks,
...
+
+    odb::dbNet* net = iterm->getNet();
+    if (net != nullptr) {
+      for (odb::dbITerm* net_iterm : net->getITerms()) {
+        if (net_iterm == iterm) {
+          continue;
+        }
+
+        if (net_iterm->getInst()->getPlacementStatus()
+            == odb::dbPlacementStatus::PLACED) {
```
maliberty, line 5989: Use net_iterm->getInst()->getPlacementStatus().isPlaced() which covers placed & locked

src/mpl2/src/object.cpp
AcKoucher: This is where mpl2 is updating db with the orientation that we are correcting later with the new pass. If they're locked here, we won't be able to adjust the orientation. 
https://github.com/AcKoucher/OpenROAD/blob/82f3cabf7abce001fd8ae8672802a517f2b2fb3f/src/mpl2/src/hier_rtlmp.cpp#L638-L644

The new pass now is responsible for locking the macros:

https://github.com/AcKoucher/OpenROAD/blob/82f3cabf7abce001fd8ae8672802a517f2b2fb3f/src/mpl2/src/hier_rtlmp.cpp#L6033-L6035

src/mpl2/src/hier_rtlmp.cpp
AcKoucher: Is it really necessary to check placed and locked at this point? We will lock the macros only after adjusting their orientation with the new pass which runs after we set the temporary location for std cells.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -5914,6 +5918,139 @@ void HierRTLMP::FDPlacement(std::vector<Rect>& blocks,
...
+  for (odb::dbInst* inst : block_->getInsts()) {
+    if (!inst->isBlock()) {
+      continue;
+    }
+
+    const float original_wirelength = calculateRealMacroWirelength(inst);
+    odb::Point macro_location = inst->getLocation();
+
+    flipRealMacro(inst, is_vertical_flip);
+    inst->setLocation(macro_location.getX(), macro_location.getY());
```
eder-matheus, line 6020: Perhaps a small comment explaining why setting the location after flipping would be good for clarity.

src/mpl2/src/object.cpp
```diff
@@ -955,66 +977,36 @@ void HardMacro::updateDb(float pitch_x, float pitch_y, odb::dbBlock* block)
...
-                  = dbuToMicron(static_cast<float>(layer->getPitchX()), dbu_);
-              offset_x
-                  = dbuToMicron(static_cast<float>(layer->getOffsetX()), dbu_);
-            }
-            pin_width_x = dbuToMicron(static_cast<float>(box->getDX()), dbu_);
+            computeCoordinateSpacingParameters(
+                block, layer, box, offset_y, pitch_y, pin_width_y, false);
+          }
+
+          if (layer->getDirection() == odb::dbTechLayerDir::VERTICAL) {
```
eder-matheus, line 986: You can merge the second `if` statement with the previous one into an `else`, so you don't need to evaluate the second check.

src/mpl2/src/object.cpp
```diff
@@ -1031,9 +1023,92 @@ void HardMacro::updateDb(float pitch_x, float pitch_y, odb::dbBlock* block)
...
-  inst_->setOrigin(snap_origin_x, snap_origin_y);
-  inst_->setOrient(orientation_);
-  inst_->setPlacementStatus(odb::dbPlacementStatus::LOCKED);
+  const odb::Point snap_origin(snap_origin_x, snap_origin_y);
+
+  return snap_origin;
+}
+
+// Compute for each pin: layer offset, pitch and pin width
+void HardMacro::computeCoordinateSpacingParameters(
```
eder-matheus, line 1032: Why the `Coordinate` word in the function names? Perhaps `computeDirectionSpacingParameters` is more precise.

eder-matheus (changes requested):
Overall, it looks great. The large refactor in the `object.cpp` file is a good improvement in the code readability.

@maliberty CI results from last week show a few metrics failures, listed below:

- `asap7/aes-block: [ERROR] cts__design__instance__count__hold_buffer fail test: 1979.0 <= 1966.0`
- `nangate45/swerv_wrapper: [ERROR] finish__timing__drv__hold_violation_count fail test: 196.0 <= 139.0`
- `nangate45/tinyRocket: [ERROR] finish__timing__setup__ws fail test: -0.00283884 >= 0.0`

Secure-CI has only one failure in the `finish__timing__setup__ws` metric. We could update these metrics and investigate the reason for the worst numbers later.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -5914,6 +5918,139 @@ void HierRTLMP::FDPlacement(std::vector<Rect>& blocks,
...
+    odb::dbNet* net = iterm->getNet();
+    if (net != nullptr) {
+      for (odb::dbITerm* net_iterm : net->getITerms()) {
+        if (net_iterm == iterm) {
+          continue;
+        }
+
+        if (net_iterm->getInst()->getPlacementStatus().isPlaced()) {
+          const float x1 = dbuToMicron(iterm->getBBox().xCenter(), dbu_);
+          const float y1 = dbuToMicron(iterm->getBBox().yCenter(), dbu_);
```
maliberty, line 5987: These two don't depend on the net_iterm and can be hoisted out of the inner for loop for efficiency.

src/mpl2/src/object.cpp
```diff
@@ -1031,9 +1023,92 @@ void HardMacro::updateDb(float pitch_x, float pitch_y, odb::dbBlock* block)
...
+{
+  float pitch = 0.0;
+
+  if (is_vertical_direction) {
+    pitch = layer->getPitchX();
+  } else {
+    pitch = layer->getPitchY();
+  }
+
+  return static_cast<float>(pitch);
```
maliberty, line 1072: why is the static_cast necessary when pitch is already a float?

src/mpl2/src/object.cpp
```diff
@@ -1031,9 +1023,92 @@ void HardMacro::updateDb(float pitch_x, float pitch_y, odb::dbBlock* block)
...
+{
+  float offset = 0.0;
+
+  if (is_vertical_direction) {
+    offset = layer->getOffsetX();
+  } else {
+    offset = layer->getOffsetY();
+  }
+
+  return static_cast<float>(offset);
```
maliberty, line 1086: why is the static_cast necessary when offset is already a float?

src/mpl2/src/object.cpp
```diff
@@ -1031,9 +1023,92 @@ void HardMacro::updateDb(float pitch_x, float pitch_y, odb::dbBlock* block)
...
+{
+  float pin_width = 0.0;
+
+  if (is_vertical_direction) {
+    pin_width = box->getDX();
+  } else {
+    pin_width = box->getDY();
+  }
+
+  return static_cast<float>(pin_width);
```
maliberty, line 1100: why is the static_cast necessary when pin_width is already a float?

src/mpl2/src/object.cpp
AcKoucher: The intent was to make clear that the the function works for one coordinate.
I agree with you. Direction is more precise.

src/mpl2/src/object.cpp
maliberty: There should be no pins on a layer with no direction.

maliberty (approved):

PR #3981, opened 2023-09-13
grt: print all pins out of die area before erroring

rovinski:
It seems like the code meant to do this, but got warn/error switched around. Otherwise, `GRT-0028` is actually unreachable.


rovinski:
@eder-matheus it seems like duplicate messages are printed if multiple shapes of the same pin are outside. Should we break out of the loop after the first warning?

eder-matheus (approved):

eder-matheus:
> @eder-matheus it seems like duplicate messages are printed if multiple shapes of the same pin are outside. Should we break out of the loop after the first warning?

Perhaps we can count the number of blockage shapes out of the die for each pin, and print this number instead of one print for shape.

maliberty (approved):

PR #3983, opened 2023-09-13
Fix fanout for eqy

openroadie:
Make changes to repair_fanout7 to not leave outputs of flops as dangling (since dangling outputs do not get checked for equivalence). Verified that the gate cloning bug would be caught by this new netlist by undoing the code fix.


openroadie:
Added skip_pin_swap to the test because the current equivalence recipe does not seem to handle it well and it is not important to use that option for this test.

maliberty (approved):

maliberty:
What happens with pin swapping on?  I'm surprised that doesn't work as it should be one of the easiest things to verify.

PR #3984, opened 2023-09-13
Grt default cn percentage

arthurjolo:
Related to  [#3452](https://github.com/The-OpenROAD-Project/OpenROAD/pull/3452).

Change the default percentage of critical nets from 0% to 10%.
Also it changes how the critical nets are treated during the overflow iterations on 2D maze route, in order to improve the tns and ws metrics. This is done by changing the critical nets priority when maze router orders the nets. When ordering, nets with overflow get the highest priority and the critical nets are ranked right after them, the non critical nets without overflow get the lowest priority.


src/grt/src/fastroute/src/utility.cpp
```diff
@@ -1122,6 +1122,19 @@ void FastRouteCore::StNetOrder()
   std::stable_sort(
       tree_order_cong_.begin(), tree_order_cong_.end(), compareTEL);
 
+  // Set the 70% (or less) of non critical nets that doesn't have overflow
```
eder-matheus, line 1125: Could you explain in more detail the idea of this update? What set of nets should have the highest priority?

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -721,12 +721,12 @@ NetRouteMap FastRouteCore::getPlanarRoutes()
         // defines the layer used for horizontal edges are still 2D
         int layer_v = 0;
 
-        if (layer_orientation_ != 0) {
-          layer_h = 1;
-          layer_v = 2;
+        if (((nets_[netID]->getMinLayer() % 2) - layer_orientation_) != 0) {
```
eder-matheus, line 724: Can you remind me what is this function used for? Why layers 1 and 2 were the default before?

src/grt/src/GlobalRouter.cpp
```diff
@@ -99,7 +99,7 @@ GlobalRouter::GlobalRouter()
       verbose_(false),
       min_layer_for_clock_(-1),
       max_layer_for_clock_(-2),
-      critical_nets_percentage_(0),
+      critical_nets_percentage_(10),
```
eder-matheus, line 102: @maliberty Should we set the 10% of critical nets as default in the tool code or enable it only in ORFS?

eder-matheus (changes requested):
Jenkins has a lot of failures in the unit tests. Could you check it? It seems to be a segfault issue in different tool unit tests.

src/grt/src/fastroute/src/FastRoute.cpp
arthurjolo: This is checking the orientation of the wires on a certain layer. In this function we are converting the nets routing from fastroute to segments so the MakeWireParasitics can understand them.
It was defaulted as 1 and 2 before because they were the first routing layers for the designs, but I changed it because it was an assumption, and some nets might have the minimal layer changed.

eder-matheus: Thanks! Your change looks good and makes sense with your new modeling.

src/grt/src/fastroute/src/utility.cpp
arthurjolo: The  idea is to first ripup at least 30% of the nets(all non critical), with the highest priority to non critical nets with overflow, than the critical nets, and with the lowest priority the non critical nets without overflow. 
The non critical nets with overflow are ripped up first to decrease the chances of critical nets also getting ripped up, and if the do the should have more space to find routings that doesn't has a big deroute.

src/grt/src/GlobalRouter.cpp
maliberty: I think we should change the default as other flows are unlikely to update and this looks to be a good change.

eder-matheus:
@arthurjolo It still has some failures in the ci. You will probably need to update some ok files. Other than that, I'm fine with the changes in this PR.

arthurjolo:
> @arthurjolo It still has some failures in the ci. You will probably need to update some ok files. Other than that, I'm fine with the changes in this PR.

I still haven't updated the OK files. There is another crash that wasn't resolved, once it is resolved I will update the OK files.

src/grt/src/GlobalRouter.cpp
```diff
@@ -785,8 +785,16 @@ void GlobalRouter::initNets(std::vector<Net*>& nets)
...
-    computeNetSlacks();
+    if (sta_->getDbNetwork()->defaultLibertyLibrary() != nullptr) {
+      fastroute_->setUpdateSlack(critical_nets_percentage_);
+      computeNetSlacks();
+    } else {
+      critical_nets_percentage_ = 0;
+      logger_->warn(
+          GRT,
+          300,
+          "Timing is not available, setting critical nets percentage to 0\n");
```
eder-matheus, line 796: You can remove the extra \n. Logger already adds a new line at the end of each message.

src/grt/src/GlobalRouter.cpp
```diff
@@ -785,8 +785,16 @@ void GlobalRouter::initNets(std::vector<Net*>& nets)
   }
 
   if (critical_nets_percentage_ != 0) {
-    fastroute_->setUpdateSlack(critical_nets_percentage_);
-    computeNetSlacks();
+    if (sta_->getDbNetwork()->defaultLibertyLibrary() != nullptr) {
+      fastroute_->setUpdateSlack(critical_nets_percentage_);
+      computeNetSlacks();
+    } else {
```
eder-matheus, line 791: I think you could also add a flag to the `initNets` function to detect if it's being called during `readGuides`/`loadGuidesFromDB` or during the normal grt flow. Currently, the warning message can be generated when using the `read_guides` command, which I think is not correct.

eder-matheus:
@arthurjolo You need to clang-format, and I believe a few unit tests will still fail. Other than that, I'm fine with the PR and will approve after the final changes.

arthurjolo:
> @arthurjolo You need to clang-format, and I believe a few unit tests will still fail. Other than that, I'm fine with the PR and will approve after the final changes.

Yes, I am updating the OK files now, and clang formating

eder-matheus:
@arthurjolo Some rsz failures:
`repair_hold10 (tcl) *FAIL*`
`repair_setup6 (tcl) *FAIL*`

They probably need an update in their ok files, too.

eder-matheus (approved):

eder-matheus:
@vvbandeira This can be merged after PR https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1508 be ready to merge too.

vvbandeira (approved):

PR #3986, opened 2023-09-13
gui: display message to inform that design is loading before first render

### LINKED ISSUE #3937: OpenROAD GUI doesn't show any progress indication when drawing the first time
### Description

2_5_floorplan_tapcell.odb in my case draws quickly:

image

2_6_floorplan_pdn.odb when ready and responding to zooming commands is just black...  No matter how far  I zoom in...

image

Eventually, I get something. After the image is drawn the first time, it responds in seconds in my case(ca. 5mm^2 megaboom design).

image

### Suggested Solution

There should be some sort of progress indication of rendering the first time, because the first time the renderinger takes a lot longer than subsequently. Perhaps go from busy & black screen, to showing "Rendering..."  when rendering the first time?

### Additional Context

_No response_

AcKoucher:
Resolves #3937

Just drawing a string in the center of the layout viewer.

### DIFF
--- a/src/gui/src/renderThread.cpp
+++ b/src/gui/src/renderThread.cpp
@@ -145,15 +145,22 @@ void RenderThread::run()
   }
 }
 
-void RenderThread::drawRenderIndication(gui::Painter& painter,
+void RenderThread::drawRenderIndication(Painter& painter,
                                         const odb::Rect& bounds)
 {
-  painter.setPen(gui::Painter::white, true);
+  QPainter* qpainter = static_cast<GuiPainter&>(painter).getPainter();
+  const QFont initial_font = qpainter->font();
+  QFont indication_render_font = qpainter->font();
+  indication_render_font.setPointSize(16);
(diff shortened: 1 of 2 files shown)


AcKoucher:
This is not working properly, I'll close the PR to investigate.

AcKoucher:
This is ready for review.

rovinski:
> This is not working properly, I'll close the PR to investigate.

Is `is_first_render_done_` a variable shared between threads? If it's not working as expected it could be because it's not declared as `std::atomic<bool>`.

AcKoucher:
@rovinski Apparently it's working fine now. 
Previously I didn't realize that I could use the same thread run to draw the message, erase it and then draw the design. Before the last commit, we would have to wait for the subsequent run to draw the design, but, in some cases this run does not happen, e. g. when the .odb has little data and loads fast.

maliberty (approved):

PR #3989, opened 2023-09-14
mpl: Fixes missing implementation of Annealer

QuantamHD:
The Annealer appears to be unimplemented, and
was causing errors when linking to openroad as a library.


QuantamHD:
I think no on the clang-tidy suggestions. These files seem to have a lot of problems with them already.

Issue #3964, opened 2023-09-07
Detailled routing runs forever when I/O pins on top layer


lustefan97:
### Describe the bug

I try to implement a Face-to-face 3D flow. I use the Ibex core design as a base to develop it, I place all its IO pins on the top layer (Pad in ASAP7) but the detailled routing doesn't seem to converge towards an optimized solution with 0 violations. It always "oscillates" around a certain number of violations but never seems to tend to zero.

### Expected Behavior

Detailled routing finishing and optimization converging towards 0 violations.

### Environment

```shell
Git commit: 972137093aa4850b04b51185956b51b8a28b2ccf
kernel: Linux 5.15.90.1-microsoft-standard-WSL2
os: Ubuntu 22.04.2 LTS (Jammy Jellyfish)
```

### To Reproduce

[Test3D.zip](https://github.com/The-OpenROAD-Project/OpenROAD/files/12552899/Test3D.zip)

In this zip file should be all the files needed to reproduce the issue. 
Create a directory "Test3D" in _/OpenROAD/test/_, enter it, open the GUI and enter _source "ibex3D_asap7_bottom.tcl"_ in the GUI TCL command box.

eder-matheus:
@lustefan97 Thanks for the test case. Debugging it, I've noticed that some IO pins placed on the PAD layer are very close to each other in adjacent tracks:
<img width="556" alt="image" src="https://github.com/The-OpenROAD-Project/OpenROAD/assets/43450810/bd154fec-c780-49ee-9836-81ddab21e984">

I'm trying a different config for the define_pin_shape_pattern command. Can you try a larger x/y pitch?
I'm running this config now:
```
set top_layer Pad
set top_pin_size {0.04 0.04}
set x_pitch 0.16
set y_pitch 0.16
```

With one track of difference between each pin, I expect the detailed router will have more room to route those pins.

eder-matheus:
@lustefan97 I found a good setup for your test case, where the detailed route finishes in 4 iterations with zero DRCs.

I suggest you using this configuration for global and detailed routing:

```
# Global Routing
set global_routing_layers M2-M9
set global_routing_clock_layers M4-M6
set global_routing_layer_adjustments {{M1 0.5} {M2 0.5} {M3 0.5} {M4 0.5} {M5 0.5} {M6 0.5} {M7 0.5} {M8 0.5} {M9 0.5}}

# Detailled Routing
set min_routing_layer M2
[...]

PR #3994, opened 2023-09-14
drt: update stackVias for pins above the max routing layer

eder-matheus:
Include some fixes for issues detected in the test case provided in https://github.com/The-OpenROAD-Project/OpenROAD/issues/3964.


src/drt/src/TritonRoute.cpp
```diff
@@ -1040,6 +1040,14 @@ void TritonRoute::stackVias(odb::dbBTerm* bterm,
                             int bterm_bottom_layer_idx,
                             bool has_routing)
 {
+  // if pin access runs before the detailed route step, the nets with pins above
+  // the max routing layer will have the stacked vias inserted. added this check
+  // to avoid trying to stack new vias on the same nets.
+  odb::dbNet* net = bterm->getNet();
+  if (nets_with_stacked_vias_.count(net) == 1) {
+    return;
+  }
```
maliberty, line 1049: What if PA was run in a separate process (like grt) and we are just opening the db?

eder-matheus: I've updated the code to look at the odb objects instead of relying on non-persistent information.

maliberty:
drt
top_level_term (tcl) *FAIL*
top_level_term (py) *FAIL*
top_level_term2 (tcl) *FAIL*
top_level_term2 (py) *FAIL*

maliberty (approved):

Issue #1747, opened 2022-04-03
[Feature request] Routing only specific nets


armleo:
Can OpenRoad implement a TCL command for routing only specific nets? Looks like FastRoute will route the nets it has been asked to route. So this should be a fairly simple feature to implement.

The reason for this is that I wanted to extend OpenLane to route some PDN/Mixed signals. However, it seems that the only reasonable way to do this is by asking FastRoute to route these signals before Place. Any suggestions?

maliberty:
PDN signals are routed by pdngen not fastroute.  Would you describe further what you are trying to do.  Note that you would also have to enhance the detailed router as well.

armleo:
I am trying to route mixed signals from PADs to GPIOs located inside a Caravan template without any manual work. For this I made a custom NONDEFAULTRULE in LEF and I can observe that these rules are followed by both global and detailed router.

However, since the timings for these nets are disabled in SDC the observed routes are not the shortest paths, since there signals between PADs and GPIO module which are routed with shortest path, as they have timings enabled.

So I am currently trying to implement GPIO routing in the floorplanning stage, before placement (as the placement of GPIO is done manually) and then fixing them, preventing routers from modifying these nets. This would also allow user to see the nets and analyze the optimality before fully completing the flow.

I could just remove all nets from the DEF, do the routing, then add the routed nets back to DEF, but I thought if the router allowed to select the nets to route or prioritize this would solve this issue.

vijayank88:
@maliberty 
Any update on this request?

maliberty:
Another option would be a user defined high priority set of nets.

PR #4010, opened 2023-09-18
grt: add command to route specific nets

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/1747


src/grt/src/GlobalRouter.tcl
```diff
@@ -340,6 +340,26 @@ proc repair_antennas { args } {
...
+                 flags {}
+  sta::check_argc_eq1 "set_nets_to_route" $args
+  set net_names [lindex $args 0]
+  set block [ord::get_db_block]
+  if { $block == "NULL" } {
+    utl::error GRT 252 "Missing dbBlock."
+  }
+
+  foreach net [get_nets $net_names] {
+    if { $net != "NULL" } {
```
maliberty, line 357: warn if not found

src/grt/src/GlobalRouter.cpp
```diff
@@ -2743,35 +2743,47 @@ static bool nameLess(const Net* a, const Net* b)
...
+    std::sort(non_clk_nets.begin(), non_clk_nets.end(), nameLess);
+
+    nets = clk_nets;
+    nets.insert(nets.end(), non_clk_nets.begin(), non_clk_nets.end());
+  } else {
+    for (odb::dbNet* db_net : nets_to_route_) {
+      Net* net = addNet(db_net);
+      if (net) {
+        nets.push_back(net);
+      }
```
maliberty, line 2783: shouldn't we still sort these by clk/non-clk?

maliberty (changes requested):

src/grt/src/GlobalRouter.tcl
eder-matheus: The `get_nets` function already has a warning for nets not found.

src/grt/src/GlobalRouter.cpp
eder-matheus: Fixed it, and the code is better now, with fewer changes than before.

maliberty (approved):

PR #4013, opened 2023-09-18
gui: fixes to segfaults and allow for better debugging

gadfort:
Adds:
- tcl commands to access heatmap settings
- avoids using corner in dbSta for power density as this can segfault
- ensure nodes are saved after failed IRDrop analysis for better debugging.

### DIFF
--- a/src/psm/src/pdnsim.cpp
+++ b/src/psm/src/pdnsim.cpp
@@ -149,9 +149,7 @@ void PDNSim::saveIRDrop(psm::IRSolver* ir_solver)
         = std::abs(ir_solver->getSupplyVoltageSrc() - voltage);
   }
   ir_drop_ = ir_drop;
-  min_resolution_ = ir_solver->getMinimumResolution();
 
-  heatmap_->update();
   if (debug_gui_) {
     debug_gui_->setSources(ir_solver->getSources());
   }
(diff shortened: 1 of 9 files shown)


maliberty:
> avoids using corner in dbSta for power density as this can segfault
What is the issue?  Is this something that needs fixing?

src/psm/src/pdnsim.cpp
```diff
@@ -135,10 +163,10 @@ void PDNSim::analyzePowerGrid(const std::string& voltage_file,
   auto irsolve_h = getIRSolver(enable_em);
 
   if (!irsolve_h->build(error_file)) {
+    saveIRDrop(irsolve_h.get());
     logger_->error(
         utl::PSM, 78, "IR drop setup failed.  Analysis can't proceed.");
```
maliberty, line 168: Why do we save the results on failure?

gadfort:
@maliberty if the gui is opened and the `corner_` gets set, but then a new set of corners are defined, the `corner_` is destroyed by OpenSTA, but this reference doesn't know about it causing it to segfault when attempting to run.

src/psm/src/pdnsim.cpp
gadfort: the GUI debugger relies on this to be saved. At present, we can only debug PSM when it is successful. This ensures the information about the nodes is copied over so we can debug it when it "claims" the PDN is not connected

maliberty (approved):

PR #4016, opened 2023-09-19
etc: add yosys, eqy and sby as deps

vvbandeira:
(draft for now, will likely add a switch to turn on/off as it is not required for now)


gadfort:
@vvbandeira is it possible to make the yosys, etc optional? These are only needed for some developers.

PR #4020, opened 2023-09-20
pad: ensure pad is placed inside the row

gadfort:
Ensures the pads are in the row they are assigned to.


src/pad/src/ICeWall.cpp
```diff
@@ -1239,7 +1286,8 @@ std::vector<odb::dbInst*> ICeWall::getPadInstsInRow(odb::dbRow* row) const
 
     const odb::Rect instbbox = inst->getBBox()->getBox();
 
-    if (row_bbox.intersects(instbbox)) {
+    if (row_bbox.intersects(instbbox)
+        && row_bbox.intersect(instbbox).area() > 0) {
```
maliberty, line 1290: I think this is equivalent to 
````
row_bbox.overlaps(instbbox)
```

gadfort: you are correct, I made the change

PR #4021, opened 2023-09-20
gui: add gui::supported command and test for it

gadfort:
As mentioned in: https://github.com/The-OpenROAD-Project/OpenROAD/pull/3997


src/gui/src/stub.cpp
```diff
@@ -195,12 +195,19 @@ void initGui(OpenRoad* openroad)
 {
   auto interp = openroad->tclInterp();
   // Tcl requires this to be a writable string
-  std::string cmd(
+  std::string cmd_save_imame(
```
maliberty, line 198: typor imame -> image

src/gui/test/supported.ok
```diff
@@ -0,0 +1 @@
+Found / Pass
```
maliberty, line 1: Will this test fail in the non-gui build?

src/gui/src/stub.cpp
gadfort: fixed

PR #4023, opened 2023-09-20
dpl: Handles odd numbers of single height rows in double height context

QuantamHD:
As mentioned in the comment.

If there's an uneven number of single row height cells, say 21. The above layer mapping coordinates on double height rows will round up, because they don't know if there's 11 or 10 rows of double height cells. This just caps it to the right amount.


maliberty (approved):

maliberty:
@mina1460 any concerns?

PR #4027, opened 2023-09-21
flip-flop clustering

therealshreyas:
flip-flop clustering based on ICCAD16 paper (https://vlsicad.ucsd.edu/Publications/Conferences/344/c344.pdf)


src/gpl/src/mbff.h
```diff
@@ -0,0 +1,77 @@
+#include <vector>
```
QuantamHD, line 1: Try running clang format on these files. The indenting is using tabs, and not spaces.

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,77 @@
...
+using namespace std;
```
QuantamHD, line 2: Remove `using namespace std;`

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+using namespace lemon;
+using namespace operations_research;
+using namespace sat;
+
+
+using Graph = ListDigraph;
+using Node = ListDigraph::Node;
+using Edge = ListDigraph::Arc;
+
+auto seed = chrono::high_resolution_clock::now().time_since_epoch().count();
```
QuantamHD, line 30: This will make the results of this process non-deterministic. Please replace with a user-configurable fixed seed.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+#include <ortools/sat/cp_model.h>
+#include <ortools/sat/cp_model.pb.h>
+#include <ortools/sat/cp_model_solver.h>
+#include <lemon/list_graph.h>
+#include <lemon/network_simplex.h>
+#include <lemon/maps.h>
+using namespace std;
+using namespace lemon;
+using namespace operations_research;
+using namespace sat;
```
QuantamHD, line 23: Remove `using namespace *;`. It's much better to leave the namespace alone. If you must shorten the namespace prefer to use `using ortools = operations_research` or something like that.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+using Node = ListDigraph::Node;
+using Edge = ListDigraph::Arc;
+
+auto seed = chrono::high_resolution_clock::now().time_since_epoch().count();
+std::mt19937 mt(seed);
+
+/* hard coded values for height, width, power consumption, etc. */
+float WIDTH = 2.448, HEIGHT = 1.2;
+float RATIOS[6] = {1, 0.875, 0.854, 0.854, 0.844, 0.844}, POWER[6] = {1, 0.875, 0.854, 0.854, 0.844, 0.844};
+int BITCNT[6] = {1, 4, 8, 16, 32, 64};
```
QuantamHD, line 36: Magic numbers are hard to read, and should be used sparingly. If they are required try adding a significant comment on exactly what they are used for, and why they were chosen.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+namespace gpl {
+	// begin or-tools usage
```
QuantamHD, line 38: Unnecessary comment.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+#include <lemon/maps.h>
+using namespace std;
+using namespace lemon;
+using namespace operations_research;
+using namespace sat;
+
+
+using Graph = ListDigraph;
+using Node = ListDigraph::Node;
+using Edge = ListDigraph::Arc;
```
QuantamHD, line 28: I think I would prefer to use the Lemon standard names. @maliberty ?

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+			new_flop = prob_dist[i];
+			if (cum_sum * 100.0 >= rand_num * tot_dist) break;
+		}
+
+		return new_flop;
+	}
+
+	vector<Tray> MBFF::GetStartTrays(vector<Flop> &flops, int num_trays, float AR) {
+
+		int num_flops = (int)flops.size();
```
QuantamHD, line 431: Prefer to use size_t instead of int when dealing with vector sizes. Also if you need to down convert use `static_cast<int>()` instead of c-style casts like `(int)`

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+
+		NUM_THREADS = threads;
+	}
+
+	// dector
+	MBFF::~MBFF() {
+		FLOPS.clear();
+		PATHS.clear();
+		NUM_THREADS = 0;
+	}
```
QuantamHD, line 1008: Not necessary. The default destructor will handle this.

Please replace with MBFF::~MBFF() = default;

src/gpl/src/mbff.cpp
QuantamHD, line 36: I think we can replace BITCNT with a function instead. Something like

getBinSize(int bin_index) {1 << bin_index;}

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1024 @@
...
+		        trays[i] = new_tray;
+			 }
+
+			return tot_d;
+		}
+
+
+		/* 
+			NOTE: CP-SAT constraints only work with INTEGERS
+			so, all coefficients (shift_x and shift_y) are multiplied by 100
```
QuantamHD, line 113: Can we use DBU in the internal solver. Those units are integer by default.

QuantamHD (changes requested):

src/gpl/README.md
```diff
@@ -47,6 +47,10 @@ global_placement
     [-timing_driven_nets_percentage]
     [-pad_left pad_left]
     [-pad_right pad_right]
+    [-alpha alpha]
+    [-beta beta]
+    [-max_split_size max_split_size]
+    [-num_threads num_threads]
```
rovinski, line 53: Thread count is set at the application level. You should remove this flag and instead call `OpenRoad::getThreadCount()`.

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,72 @@
+#include <vector>
```
maliberty, line 1: Needs the standard BSD header (see any other file).  You can assign the copyright to "Precision Innovations Inc."

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,72 @@
...
+};
+
+namespace gpl {
+
+class MBFF {
+
+  public:
+    MBFF(int num_flops, int num_paths, std::vector<float> x,
+         std::vector<float> y, std::vector<std::pair<int, int> > paths,
+         int threads);
```
maliberty, line 34: Take vectors by const-reference for efficiency

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,72 @@
...
+    ~MBFF();
+
+    void Run(int mx_sz, float ALPHA, float BETA);
+
+  private:
+    std::vector<Flop> FLOPS;
+    std::vector<Path> PATHS;
+    int NUM_THREADS;
```
maliberty, line 42: methods before data members per https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/gpl/src/replace.i
```diff
@@ -43,6 +43,16 @@ replace_nesterov_place_cmd()
   replace->doNesterovPlace();
 }
 
+
+void
+replace_run_mbff_cmd(int max_sz, float ALPHA, float BETA)
```
maliberty, line 48: lowercase alpha & beta (variable names not macros)

src/gpl/src/replace.tcl
```diff
@@ -64,6 +64,9 @@ sta::define_cmd_args "global_placement" {\
     [-timing_driven_nets_percentage timing_driven_nets_percentage]\
     [-pad_left pad_left]\
     [-pad_right pad_right]\
+    [-alpha alpha]\
+    [-beta beta]\
+    [-max_split_size max_split_size]\
```
maliberty, line 69: Why are these added to global_placement ?

src/gpl/src/replace.tcl
```diff
@@ -82,6 +85,7 @@ proc global_placement { args } {
       -timing_driven_net_reweight_overflow \
       -timing_driven_net_weight_max \
       -timing_driven_nets_percentage \
+      -alpha -beta -max_split_size \
```
maliberty, line 88: same

src/gpl/src/replace.tcl
```diff
@@ -384,4 +388,18 @@ proc get_global_placement_uniform_density { args } {
   }
   return $uniform_density
 }
+
+proc get_mbff_results { args } {
```
maliberty, line 392: Why is the command get_mbff_results but it invokes replace_run_mbff_cmd.  The user command should be something like cluster_flops and should be in the global namespace not the gpl namespace.

src/gpl/README.md
maliberty, line 53: This seems related to the confusion in the .tcl as to what the user command is.

src/gpl/src/replace.cpp
```diff
@@ -296,6 +297,28 @@ void Replace::doInitialPlace()
   ip_->doBicgstabPlace();
 }
 
+void Replace::runMBFF(int max_sz, float alpha, float beta, int threads) {
+
+  int num_flops = 0, num_paths = 0;
+  vector<float> X, Y;
```
maliberty, line 303: lowercase x, y

src/gpl/src/replace.cpp
```diff
@@ -296,6 +297,28 @@ void Replace::doInitialPlace()
...
+  vector<float> X, Y; 
+  vector<pair<int, int> > paths;
+
+
+  auto block = db_->getChip()->getBlock();
+  for (auto inst : block->getInsts()) {
+    auto mstr = inst->getMaster();
+    if (mstr->isSequential()) {
+      int x, y; inst->getOrigin(x, y);
+      X.push_back(x), Y.push_back(y);
```
maliberty, line 312: split into separate statements

src/gpl/src/replace.cpp
```diff
@@ -296,6 +297,28 @@ void Replace::doInitialPlace()
...
+      num_flops++;
+    }
+  }
+
+  MBFF pntset = MBFF(num_flops, num_paths, X, Y, paths, threads);
```
maliberty, line 317: MBFF pntset(num_flops, num_paths, X, Y, paths, threads);

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+/* 
+height and width of a flip-flop
+ratios are used for:
+    (1) finding all slot locations in a tray
+    (2) finding power consumption per slot 
+*/
+
+float WIDTH = 2.448, HEIGHT = 1.2;
+float RATIOS[6] = { 1, 0.875, 0.854, 0.854, 0.844, 0.844 };
```
maliberty, line 29: constexpr
move inside the gpl namespace

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+namespace gpl {
+
+int MBFF::GetBitCnt(int bit_idx) { return (1 << bit_idx); }
+
+float MBFF::RunLP(std::vector<Flop> flops, std::vector<Tray> &trays,
+                  std::vector<std::pair<int, int> > clusters, int sz) {
+
+    int num_flops = static_cast<int>(flops.size()), num_trays = static_cast<int>(trays.size());
```
maliberty, line 38: split into separate statements.  This occurs in many places

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+float WIDTH = 2.448, HEIGHT = 1.2;
+float RATIOS[6] = { 1, 0.875, 0.854, 0.854, 0.844, 0.844 };
+
+namespace gpl {
+
+int MBFF::GetBitCnt(int bit_idx) { return (1 << bit_idx); }
+
+float MBFF::RunLP(std::vector<Flop> flops, std::vector<Tray> &trays,
+                  std::vector<std::pair<int, int> > clusters, int sz) {
```
maliberty, line 36: flops & clusters should be passed by const-ref.  Many other similar cases elsewhere.
sz is unused

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    objective->SetMinimization();
+
+    solver->Solve();
+
+    float tot_d = 0;
+    for (int i = 0; i < num_trays; i++) {
+        Tray new_tray;
+        new_tray.pt.x = tray_x[i]->solution_value();
+        new_tray.pt.y = tray_y[i]->solution_value();
+        tot_d += MBFF::GetDist(trays[i].pt, new_tray.pt);
```
maliberty, line 103: MBFF:: is unneeded

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    for (int i = 0; i < 6; i++) {
+        for (int j = 0; j < static_cast<int>(all_trays[i].size()); j++)
+            trays.push_back(all_trays[i][j]);
+    }
+
+    int num_flops = static_cast<int>(flops.size()), num_paths = static_cast<int>(paths.size()),
+        num_trays = static_cast<int>(trays.size());
+
+    operations_research::sat::CpModelBuilder cp_model;
+    const int inf = 2000000000;
```
maliberty, line 129: Why this magic number rather than numeric_limits<int>::max() ?

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    float maxD = 0;
+    for (int i = 0; i < num_flops; i++) {
+        for (int j = 0; j < num_trays; j++) {
+
+            std::vector<Point> slots = trays[j].slots;
+
+            for (int k = 0; k < static_cast<int>(slots.size()); k++)
+                if (trays[j].cand[k] == i) {
+
+                    float shift_x = slots[k].x - flops[i].pt.x,
```
maliberty, line 161: const

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    // add constraints for Z
+    std::vector<operations_research::sat::IntVar> z_x(num_paths),
+        z_y(num_paths);
+    for (int i = 0; i < num_paths; i++) {
+        z_x[i] = cp_model.NewIntVar(operations_research::Domain(0, inf));
+        z_y[i] = cp_model.NewIntVar(operations_research::Domain(0, inf));
+    }
+
+    for (int i = 0; i < num_paths; i++) {
+        int flop_a_idx = paths[i].a, flop_b_idx = paths[i].b;
```
maliberty, line 200: const

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+    int num_flops = static_cast<int>(flops.size()), num_paths = static_cast<int>(paths.size()),
+        num_trays = static_cast<int>(trays.size());
+
+    operations_research::sat::CpModelBuilder cp_model;
+    const int inf = 2000000000;
+
+    std::vector<operations_research::sat::BoolVar> B[num_flops][num_trays];
+    std::vector<operations_research::sat::IntVar> d_x[num_flops][num_trays],
+        d_y[num_flops][num_trays];
```
maliberty, line 133: Please pick variable names that are more meaningful

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+    return tot_d;
+}
+
+/*
+        NOTE: CP-SAT constraints only work with INTEGERS
+        so, all coefficients (shift_x and shift_y) are multiplied by 100
+*/
+
+void MBFF::RunILP(std::vector<Flop> flops, std::vector<Path> paths,
```
maliberty, line 115: Give a high level description of what this method is doing.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+*/
+
+float WIDTH = 2.448, HEIGHT = 1.2;
+float RATIOS[6] = { 1, 0.875, 0.854, 0.854, 0.844, 0.844 };
+
+namespace gpl {
+
+int MBFF::GetBitCnt(int bit_idx) { return (1 << bit_idx); }
+
+float MBFF::RunLP(std::vector<Flop> flops, std::vector<Tray> &trays,
```
maliberty, line 35: Give a high level description

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+                    d_x[i][j][k] =
+                        cp_model.NewIntVar(operations_research::Domain(0, inf));
+                    d_y[i][j][k] =
+                        cp_model.NewIntVar(operations_research::Domain(0, inf));
+                    B[i][j][k] = cp_model.NewBoolVar();
+                }
+        }
+    }
+
+    // add constraints for D
```
maliberty, line 151: 'D' is not descriptive.  Likewise below

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+    const operations_research::sat::CpSolverResponse response =
+        operations_research::sat::Solve(cp_model.Build());
+    if (response.status() ==
+        operations_research::sat::CpSolverStatus::INFEASIBLE) {
+        std::cout << "No solution found.\n";
+        return;
+    }
+
+    std::cout << "Total = " << (response.objective_value()) << "\n";
```
maliberty, line 333: Use the logger rather than cout (throughout).  Consider whether this is for developer use (debugPrint) or the user (info).  See https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/docs/contrib/Logger.md

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+    std::cout << "\n";
+
+} // end ILP
+
+float MBFF::GetDist(Point a, Point b) {
+    return (abs(a.x - b.x) + abs(a.y - b.y));
+}
+
+int MBFF::GetRows(int K) {
```
maliberty, line 404: K is not descriptive so its unclear what is being mapped to rows

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+void MBFF::Run(int mx_sz, float ALPHA, float BETA) {
+    /*
+            (1) Run k-means++ based pointset decomposition
+            (2) Run Silhouette metric on each pointset from (1)
+            (3) Run capacitated k-means clustering on each pointset from (1)
+            (4) Merge pointsets from (1)
+            (5) Run ILP
+    */
+    srand(time(NULL));
```
maliberty, line 862: Use a fixed seed for reproducible results.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    omp_set_num_threads(NUM_THREADS);
+    int num_flops = static_cast<int>(FLOPS.size());
+    std::vector<Flop> flops = FLOPS;
+    std::vector<Path> paths = PATHS;
+
+    // Run k-means++ based pointset decomposition
+    std::vector<std::vector<Flop> > pointsets = KMeansDecomp(flops, mx_sz);
+
+    // all_trays[i] = all trays of size GetBitCnt(i)
+    std::vector<std::vector<Tray> > all_trays(6);
```
maliberty, line 872: Explain the use of '6' here and below.  Perhaps a named constant?

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    return ret;
+}
+
+void MBFF::Run(int mx_sz, float ALPHA, float BETA) {
+    /*
+            (1) Run k-means++ based pointset decomposition
+            (2) Run Silhouette metric on each pointset from (1)
+            (3) Run capacitated k-means clustering on each pointset from (1)
+            (4) Merge pointsets from (1)
+            (5) Run ILP
```
maliberty, line 860: each of these steps should be its own method

maliberty (changes requested):

src/gpl/README.md
```diff
@@ -68,7 +72,11 @@ global_placement
...
 - `-timing_driven_net_reweight_overflow`: set overflow threshold for timing-driven net reweighting. Allowed values are `tcl list of [0-100, int]`.
 - `-timing_driven_net_weight_max`: Set the multiplier for the most timing critical nets. Default value is 1.9.
-- `-timing_driven_nets_percentage`: Set the percentage of nets that are reweighted in timing-driven mode. Default value is 10. Allowed values are `[0-100, float]`
+- `-timing_driven_nets_percentage`: Set the percentage of nets that are reweighted in timing-driven mode. Default value is 10. Allowed values are `[0-100, float]`  
+- `alpha`: Set the weighting factor for tray cost in flip-flop clustering (recommended to be `[20.0, float]`).    
+- `beta`: Set the weighting factor for timing-critical paths in flip-flop clusering (recommended to be `[1.0. float]`).   
+- `max_split_size`: The maximum size of a single pointset after running the pointset decomposition algorithm for runtime improvement in flip-flop clustering (to not run pointset decomposition, set as `1000000`).
```
eder-matheus, line 78: An invalid number to not run pointset decomposition would be better, like -1. It's not clear why 1000000 would lead to not running this decomposition.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    for (int i = 0; i < num_trays; i++) {
+        tray_x[i] = solver->MakeNumVar(0, inf, "");
+        tray_y[i] = solver->MakeNumVar(0, inf, "");
+    }
+
+    std::vector<const operations_research::MPVariable *> d_x(num_flops),
+        d_y(num_flops);
+
+    for (int i = 0; i < num_flops; i++) {
+
```
eder-matheus, line 55: You can remove the extra new line.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+        d_x[i] = solver->MakeNumVar(0, inf, "");
+        d_y[i] = solver->MakeNumVar(0, inf, "");
+
+        int tray_idx = clusters[i].first, slot_idx = clusters[i].second;
+
+        Point flop = flops[i].pt;
+        Point tray = trays[tray_idx].pt;
+        Point slot = trays[tray_idx].slots[slot_idx];
```
eder-matheus, line 63: Perhaps `const Point&`? I think you are not changing these variables below, so you don't need to copy it.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+Tray MBFF::GetOneBit(Point pt) {
+    Tray tray;
+    tray.pt.x = pt.x - WIDTH / 2.0, tray.pt.y = pt.y - HEIGHT / 2.0;
+    tray.slots.push_back(pt);
+    return tray;
+}
+
+std::vector<std::pair<int, int> >
+MBFF::MinCostFlow(std::vector<Flop> flops, std::vector<Tray> &trays, int sz) {
+
```
eder-matheus, line 506: Remove the extra new lines.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    return clusters;
+}
+
+float MBFF::GetSilh(std::vector<Flop> flops, std::vector<Tray> trays,
+                    std::vector<std::pair<int, int> > clusters) {
+    int num_flops = static_cast<int>(flops.size()), num_trays = static_cast<int>(trays.size());
+
+    float tot = 0;
+    for (int i = 0; i < num_flops; i++) {
+        float min_num = 2000000000;
```
eder-matheus, line 599: `std::numeric_limits<int>::max()`

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+    float prev = -1;
+    while (1) {
+
+        for (int i = 0; i < K; i++)
+            clusters[i].clear();
+
+        // remap flops to clusters
+        for (int i = 0; i < num_flops; i++) {
+
+            float min_cost = 2000000000;
```
eder-matheus, line 674: std::numeric_limits<int>::max()

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+            for (int k = 0; k + 1 < static_cast<int>(tmp[j].size()); k++) {
+                cur_cost += MBFF::GetDist(tmp[j][k].pt, tmp[j].back().pt);
+            }
+        }
+
+        tmp_clusters[i] = tmp;
+        tmp_costs[i] = cur_cost;
+    }
+
+    float best_cost = 2000000000;
```
eder-matheus, line 763: std::numeric_limits<int>::max()

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1058 @@
...
+
+    std::set<int> chosen;
+    // choose initial center
+    int seed = rand() % num_flops;
+    chosen.insert(seed);
+
+    std::vector<Flop> centers;
+    centers.push_back(flops[seed]);
+
+    float d[num_flops];
```
eder-matheus, line 631: I prefer using C++ containers such as `std::vector` or `std::array` instead of a "raw" array.

QuantamHD: +1

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,111 @@
...
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+///////////////////////////////////////////////////////////////////////////////
+
+#include <vector>
+
+#pragma once
+
+struct Point {
```
QuantamHD, line 38: Points should be represented as odb::dbPoint.

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,111 @@
...
+    float x = 0, y = 0;
+};
+
+struct Tray {
```
QuantamHD, line 42: Unless for a particular reason structs should be defined in a namespace. Otherwise they are added to the global namespace which can cause conflicts in other code.

https://en.cppreference.com/w/cpp/language/definition

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,111 @@
...
+    MBFF(const int num_flops, const int num_paths, const std::vector<float>& x,
+         const std::vector<float>& y, const std::vector<std::pair<int, int> >& paths,
+         const int threads);
+    ~MBFF();
+
+    void Run(const int mx_sz, const float alpha, const float beta);
+
+  private:
+
+    float GetDist(const Point& a, const Point& b);
```
QuantamHD, line 74: If you use odb::dbPoint this method would already be defined.

However, assuming we needed a point class it would make more sense to define Point as a class with a distance method. 

The object oriented programming model attempts to merge code with data. So in general it's best to keep methods as close to their data container as possible

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,98 @@
...
+#include "odb/db.h"
+
+#pragma once
+
+namespace gpl {
+
+struct Tray {
+    odb::Point pt;
+    std::vector<odb::Point> slots;
+    int cand[70];
```
QuantamHD, line 44: Prefer std::array<int, 70>. 

Also is there an explanation for the 70 number? Perhaps a constexpr variable with a name or comment would be helpful for readability

therealshreyas: Hi Ethan, thank you very much for the comment.
There are a maximum of 64 slots in a tray -- I just created the "70" as a hard limit. 

Perhaps it would be better to create a vector, and then resize it given the size of a tray?

QuantamHD: Generally, if there's a maximum value we would want the array size to represent exactly that number.

Otherwise a future reader might be confused why there are 70 elements in a tray, but the algorithm can only place 64. Could you summarize where the limit comes from? Is it an algorithm defined constraint, or just an arbitrary constant? Either way it would be good to introduce a variable name called MAX_SLOTS

The other consideration to make is that std::array<int, MAX_SLOTS>  and int[MAX_SLOTS] consume MAX_SLOTS worth of memory regardless if they're used or not. 

Sometimes that's good if we expect that slots will always be filled or close to filled. If we expect that most slots will be empty then a std::vector can be a better choice.

therealshreyas: We are working with a maximum of 64-bit trays, so there are a maximum of 64-bit slots. 
However, there are many trays created that only contain <= 2^5 slots (so, there would be a bunch of wasted memory if I just went with the array of size 64). 

In the most recent commit, I replaced the cand array with a vector to support resizing.

src/gpl/src/mbff.cpp
therealshreyas: To my understand, shift_x and shift_y will always be floats because they involve slot locations (which will be floats because of the ratio multiplier).

src/gpl/src/mbff.h
therealshreyas: Creating the cand vector and not adding a default value, even if the size is initialized, seems to increase runtime by ~1 minute (tested for #flops <= 5K, split-size = 500).

I initialized the cand vector to have a size of 1, and made the default value -1.
The vector then gets resized when we create a Tray that doesn't need as many slots. 

I'm not sure why this happens... 

Any suggestions?

QuantamHD: The resize penalty can be high. In this case it might be worth it to move to static arrays

therealshreyas: Hi, the issue doesn't persist if I just create a vector of size 1 with the value of -1. 
Should I just keep it as is (and then resize when trays are created), or should I create a static array of size 64 for every tray? 

The runtime differences are negligible with these methods.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1219 @@
...
+            }
+        }
+
+        Remap(pointsets[t], trays);
+
+        // append trays
+        for (int i = 0; i < NUM_SIZES; i++) {
+            for (int j = 0; j < static_cast<int>(trays[i].size()); j++) {
+                all_trays[i].push_back(trays[i][j]);
+            }
```
rovinski, line 1183: Use `std::vector.insert()`
https://stackoverflow.com/questions/201718/concatenating-two-stdvectors

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,119 @@
...
+    std::vector<std::vector<Flop> > KMeans(const std::vector<Flop> &flops,
+                                           int K);
+    std::vector<std::vector<Flop> > KMeansDecomp(const std::vector<Flop> &flops,
+                                                 int MAX_SZ);
+
+    std::vector<std::pair<int, int> >
+    RunCapacitatedKMeans(const std::vector<Flop> &flops,
+                         std::vector<Tray> &trays, int sz, int iter);
+
+    std::vector<std::vector<Tray> > RunSilh(const std::vector<Flop> &pointset);
```
rovinski, line 104: Most, if not all, of these should be returning references to avoid high-overhead copying.

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,119 @@
...
+    float RunLP(const std::vector<Flop> &flops, std::vector<Tray> &trays,
+                const std::vector<std::pair<int, int> > &clusters);
+    void RunILP(const std::vector<Flop> &flops, const std::vector<Path> &paths,
+                const std::vector<std::vector<Tray> > &all_trays, float alpha,
+                float beta);
+
+    utl::Logger *log_;
+    std::vector<Flop> all_flops;
+    std::vector<Path> all_paths;
+    int num_threads;
```
rovinski, line 117: Class member variable names should end with a trailing `_`.

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,119 @@
...
+    int idx = 0;
+    float prob = 0;
+
+    bool operator<(Flop &a) { return prob < a.prob; }
+};
+
+struct Path {
+    int a = 0;
+    int b = 0;
+};
```
rovinski, line 62: Is there a difference between `Path` and `odb::Point`?

`a` and `b` are not clear names.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1219 @@
...
+}
+
+// ctor
+MBFF::MBFF(int num_flops, int num_paths, const std::vector<float> &x,
+           const std::vector<float> &y,
+           const std::vector<std::pair<int, int> > &paths, int threads,
+           utl::Logger *logger) {
+    all_flops.resize(num_flops);
+    for (int i = 0; i < num_flops; i++) {
+        odb::Point new_pt(x[i] + 70.0, y[i] + 70.0);
```
rovinski, line 1197: Where does `70.0` come from? Magic constants are frowned upon.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1219 @@
...
+
+        Flop new_flop;
+        new_flop.pt = new_pt, new_flop.idx = i;
+        all_flops[i] = new_flop;
+    }
```
rovinski, line 1202: Reserve instead of resize to avoid default initialization.
```suggestion
    all_flops.reserve(num_flops);
    for (int i = 0; i < num_flops; i++) {
        odb::Point new_pt(x[i] + 70.0, y[i] + 70.0);
        all_flops.emplace_back(new_pt, i, 0.0);
    }
```

src/gpl/src/replace.cpp
```diff
@@ -296,6 +297,29 @@ void Replace::doInitialPlace()
...
+  vector<pair<int, int> > paths;
+
+  auto block = db_->getChip()->getBlock();
+  for (auto inst : block->getInsts()) {
+    auto mstr = inst->getMaster();
+    if (mstr->isSequential()) {
+      int x_i, y_i; 
+      inst->getOrigin(x_i, y_i);
+      x.push_back(x_i);
+      y.push_back(y_i);
```
rovinski, line 313: Does it make more sense to have a single vector of locations rather than 2 vectors of x and y?

src/gpl/src/mbff.h
rovinski: Is the number of flops known at allocation time? If so you should leave the vector empty and then use `std::vector.reserve()` before pushing elements.

src/gpl/src/mbff.h
therealshreyas: a and b represent the start and end points of timing-critical path, respectively

therealshreyas: will remain so that it's clear

eder-matheus: Why not `startpoint` and `endpoint`?

src/gpl/src/replace.cpp
therealshreyas: I don't think so... at that point it would be better to have vector<odb::Point>

rovinski: IMO the only reason to have a separate x and y vector is if x and y were accessed separately. To me it appears that x and y are always accessed together, so it should be a vector of `odb::Point`. It's more concise and probably gets better locality.

src/gpl/src/mbff.h
therealshreyas: would it instead be better to have all of these functions return void, and have all the functions modify a vector passed in by reference?

eder-matheus: I don't think you need vectors passed in by reference. Returning a reference would be more readable and solve the copy issue.

src/gpl/README.md
```diff
@@ -47,10 +47,18 @@ global_placement
     [-timing_driven_nets_percentage]
     [-pad_left pad_left]
     [-pad_right pad_right]
+    [-num_threads num_threads]
```
eder-matheus, line 50: You can remove this from README since it is no longer in the Tcl file.

src/gpl/src/mbff.h
eder-matheus: @therealshreyas I'm sorry, I thought you were returning class members in this function, not a vector created inside the function. The approach you suggested is better for this and potentially other cases in your code.

src/gpl/src/mbff.cpp
```diff
@@ -0,0 +1,1219 @@
...
+
+        for (auto x : tray_sz) {
+            log_->error(utl::GPL, 512, "Tray size = {}: {}", x.first, x.second);
+        }
+
+    }
+
+} // end ILP
+
+std::vector<odb::Point> MBFF::GetSlots(const odb::Point &tray, int rows,
```
eder-matheus, line 503: Similar to the other comment. As already mentioned, you could use a reference here to avoid copying the whole vector.

eder-matheus (changes requested):

src/gpl/src/replace.cpp
```diff
@@ -296,6 +297,29 @@ void Replace::doInitialPlace()
...
+void Replace::runMBFF(int max_sz, float alpha, float beta, int threads) {
+
+  int num_flops = 0, num_paths = 0;
+  vector<odb::Point> points; 
+  vector<pair<int, int> > paths;
+
+  auto block = db_->getChip()->getBlock();
+  for (auto inst : block->getInsts()) {
+    auto mstr = inst->getMaster();
+    if (mstr->isSequential()) {
```
rovinski, line 309: Pass by const reference
```suggestion
  for (const auto& inst : block->getInsts()) {
    if (inst->getMaster()->isSequential()) {
```

src/gpl/src/replace.cpp
```diff
@@ -296,6 +297,29 @@ void Replace::doInitialPlace()
   ip_->doBicgstabPlace();
 }
 
+void Replace::runMBFF(int max_sz, float alpha, float beta, int threads) {
+
+  int num_flops = 0, num_paths = 0;
```
rovinski, line 302: One per line
```suggestion
  int num_flops = 0;
  int num_paths = 0;
```

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,118 @@
...
+struct Path {
+    int start_point;
+    int end_point;
+};
+
+class MBFF {
+
+  public:
+    MBFF(int num_flops, int num_paths, const std::vector<odb::Point> &points,
+         const std::vector<std::pair<int, int> > &paths, int threads, utl::Logger *logger);
```
rovinski, line 69: why not `const std::vector<Path>` instead of `const std::vector<std::pair<int, int> >`?

src/gpl/README.md
```diff
@@ -68,9 +75,12 @@ global_placement
 - `-initial_place_max_fanout`: set net escape condition in initial place when 'fanout >= initial_place_max_fanout'. Default value is 200. Allowed values are `[1-MAX_INT, int]`.
 - `-timing_driven_net_reweight_overflow`: set overflow threshold for timing-driven net reweighting. Allowed values are `tcl list of [0-100, int]`.
 - `-timing_driven_net_weight_max`: Set the multiplier for the most timing critical nets. Default value is 1.9.
-- `-timing_driven_nets_percentage`: Set the percentage of nets that are reweighted in timing-driven mode. Default value is 10. Allowed values are `[0-100, float]`
+- `-timing_driven_nets_percentage`: Set the percentage of nets that are reweighted in timing-driven mode. Default value is 10. Allowed values are `[0-100, float]`   
 - `-verbose_level`: set verbose level for RePlAce. Default value is 1. Allowed values are `[0-5, int]`.
 - `-force_cpu`: Force to use the CPU solver even if the GPU is available.
+- `alpha`: Set the weighting factor for tray cost in flip-flop clustering (recommended to be `[20.0, float]`).    
+- `beta`: Set the weighting factor for timing-critical paths in flip-flop clusering (recommended to be `[1.0. float]`).
```
maliberty, line 82: alpha and beta are fine for the paper but for the user are not that meaningful.  At the UI level please switch
```
alpha -> flop_weight
beta -> timing_weight
```
Please also give their default values

(also typo: clusering)

therealshreyas: alpha should be tray weight, right?

Issue #3573, opened 2023-07-04
GUI designs with origin not at (0 0)


osamahammad21:
### Describe the bug

GUI is not showing the die if the die area does not start at ( 0 0 ). This is the gui of ispd18_sample.tcl
![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/56893454/096a3c66-3ad7-424c-b456-8b22747768c4)

### Expected Behavior

The routed design to be shown in the gui

### Environment

```shell
Git commit: 914d6d23c442175889a2cca9b8e6fc37fe66e244
kernel: Linux 5.19.0-46-generic
os: Ubuntu 22.04.2 LTS (Jammy Jellyfish)
cmake version 3.24.2
-- The CXX compiler identification is GNU 11.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-9039-g914d6d23c
-- System name: Linux
-- Compiler: GNU 11.3.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.10 (found version "3.10.6") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
[...]

PR #4031, opened 2023-09-22
gui: fix fit bad behavior when DIEAREA is outside the first quadrant

### LINKED ISSUE #3998: gui fit and pan don't work well with DIEAREA outside the first quadrant
### Describe the bug

f(it) doesn't fit def DIEAREA ( -10000 -10000 ) ( 31000 10000 ) ;

panning also doesn't take the lower left portion of the design into account.

### Expected Behavior

fit should also work for DIEAREA that are not ( 0 0 ) ( width height )

### Environment

```shell
v2.0-10263-ga70c5128c (Sept 14 2023) 
M1 MacBook, compiled with mpl2 enabled.
```

### To Reproduce

[test_origin.tgz](https://github.com/The-OpenROAD-Project/OpenROAD/files/12613506/test_origin.tgz)

openroad -gui origin.or

### Relevant log output

_No response_

### Screenshots

image

### Additional Context

_No response_

AcKoucher:
Resolves #3998 

Two things are being done here:

1. I am reverting what I have previously changed in #3585. With this change we got back to the problem reported in #3573, which is now fixed (now properly) by this PR.

2. When computing the necessary shift for centralizing the design in the layout viewer, we were not considering the distance that can exist between bound's xMin / yMin and the dbu origin. This was causing the design to the drawn, but not properly displayed in the layout viewer as the center was not being placed correctly. This is distance is now being considered.

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -218,9 +218,7 @@ Rect LayoutViewer::getBounds() const
 
   Rect die = block_->getDieArea();
 
-  Rect visible(0, 0, die.xMax(), die.yMax());
-
-  bbox.merge(visible);
+  bbox.merge(die);
 
   return bbox;
 }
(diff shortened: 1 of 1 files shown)


gadfort (approved):
looks okay to me

vvbandeira (approved):

PR #4043, opened 2023-09-25
gui: zoom triggered by keyboard now focus on cursor

### LINKED ISSUE #4033: zoom in/out to the position of the mouse in the view, not the view center
### Description

The current OpenRoad behavior is to zoom in/out of the center of the view.

It's much more useful when the point to zoom in or out to is determined by the position of the mouse cursor in the view.
This is the default behavior of klayout.

### Suggested Solution

"Zoom in window" is already using the mouse cursor to determine the new view coordinates by drawing the new view bounding box.
So using the mouse position for keyboard or mouse wheel zoom in/out instead of the view center should be fairly straight forward.

### Additional Context

_No response_

AcKoucher:
Resolves #4033

### DIFF
--- a/src/gui/src/layoutTabs.cpp
+++ b/src/gui/src/layoutTabs.cpp
@@ -125,14 +125,28 @@ void LayoutTabs::setLogger(utl::Logger* logger)
 void LayoutTabs::zoomIn()
 {
   if (current_viewer_) {
-    current_viewer_->zoomIn();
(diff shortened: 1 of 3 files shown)


maliberty:
@gadfort do you have time to review this?

gadfort:
@maliberty yes, I'll take a look tomorrow

gadfort (approved):
LGTM

vvbandeira (approved):

PR #4048, opened 2023-09-26
Fixed the issue of skipping mpl2 directly after initializing the cell position

ZhuLinsen:
- Fixed the issue of skipping mpl2 directly after initializing the cell position
    - For example, in OpenROAD-flow-scripts, mixed sized placement is executed first, and then mpl2 is executed, so mpl2 will be skipped directly in present.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -384,7 +384,7 @@ void HierRTLMP::hierRTLMacroPlacer()
   // Check if placement is feasible in the core area when considering
   // the macro halos
   float macro_with_halo_area = 0;
-  int unplaced_macros = 0;
+  int unfixed_macros = 0;
   for (auto inst : block_->getInsts()) {
     auto master = inst->getMaster();
     if (master->isBlock()) {
@@ -393,7 +393,7 @@ void HierRTLMP::hierRTLMacroPlacer()
       const auto height
           = dbuToMicron(master->getHeight(), dbu_) + 2 * halo_width_;
       macro_with_halo_area += width * height;
-      unplaced_macros += !inst->getPlacementStatus().isFixed();
+      unfixed_macros += !inst->getPlacementStatus().isFixed();
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -393,7 +393,7 @@ void HierRTLMP::hierRTLMacroPlacer()
       const auto height
           = dbuToMicron(master->getHeight(), dbu_) + 2 * halo_width_;
       macro_with_halo_area += width * height;
-      unplaced_macros += !inst->getPlacementStatus().isPlaced();
+      unfixed_macros += !inst->getPlacementStatus().isFixed();
```
eder-matheus, line 396: @ZhuLinsen Could you describe how we can reproduce the issue with mpl2 being skipped?

I don't understand why changing the check from `isPlaced` to `isFixed` solves the issue. AFAIK, mixed-sized placement doesn't move the macros, but I may be wrong.

ZhuLinsen: - Of course，I have tested it in **OpenROAD-flow-scripts** with command `make floorplan  DESIGN_CONFIG=./designs/nangate45/black_parrot/config.mk`, and then I got the following message: 
 `[INFO MPL-0017] There are no unplaced macros so placement is skipped.`
- Actually, after mixed-sized placement, the macros' status change to `PLACED ` which leads to skip mpl2. In my opinion, any case involving macros can be reproduced with flow `ifp->ppl->gpl->mpl2`

maliberty: There should be no mixed size placement of you are using mpl2. How are you enabling it?

ZhuLinsen: - Oh, I found that this has been fixed in the latest flow/Makefile by skipping mixed-sized placement when using mpl2.
- But anyway, I highly recommend using 'FIXED' to check. This approach offers several advantages. Firstly, it provides better compatibility. For example, if I need to switch between mpl and mpl2, or if I want to execute mpl2 after mixed-sized placement, it is possible. Secondly, `FIXED `aligns better with the `dbPlacementStatus` setting in OpenROAD. **Mpl2 should be able to handle macros in the 'PLACED' state.**

eder-matheus: Sounds reasonable to me. I'll approve it, but will check if mpl2 is already able to handle macros in the `PLACED` state.

eder-matheus (approved):

vvbandeira (approved):

PR #4051, opened 2023-09-26
Add missing <cstdint> header for uint16_t, uint32_t etc.

hzeller:


hzeller:
gcc 13 started complaining if these are not there.

eder-matheus:
@hzeller Did you have compile issues due to the lack of these headers? The changes look reasonable, but I'm wondering why we didn't had build failures before it.

eder-matheus (approved):

vvbandeira (approved):

PR #4052, opened 2023-09-26
ci/etc: make eqy an optional dependency

vvbandeira:


gadfort:
Why make the default install it vs making the default to skip it?

vvbandeira:
@gadfort, the default is to install only if creating a docker image. For local installs the default is to skip.

gadfort:
Ahh, okay. Thanks for clarifying

PR #4055, opened 2023-09-27
etc: avoid issues with multiple users/runs

vvbandeira:


etc/DependencyInstaller.sh
```diff
@@ -584,6 +582,8 @@ PREFIX=""
 option="all"
 isLocal="false"
 equivalenceDeps="no"
+# temp dir to download and compile
+baseDir=/tmp/installers-$(date +%s)
```
rovinski, line 586: Recommend using `mktemp` as guaranteed not to alias.

PR #4056, opened 2023-09-28
modified inside blockage criteria and added debug codes

precisionmoon:
This is a minor change off the public branch.  Inside blockage now includes boundary points.

### DIFF
--- a/src/cts/src/HTreeBuilder.cpp
+++ b/src/cts/src/HTreeBuilder.cpp
@@ -123,6 +123,14 @@ void HTreeBuilder::preSinkClustering(
                                   options_->getSinkBuffer(),
                                   legalCenter.getX() * wireSegmentUnit_,
                                   legalCenter.getY() * wireSegmentUnit_);
+      if (logger_->debugCheck(CTS, "legalizer", 2) && (center != legalCenter)) {
+        // clang-format off
+	debugPrint(logger_, CTS, "legalizer", 2,
+		   "preSinkClustering legalizeOneBuffer {}: {} => {}",
+		   baseName + std::to_string(clusterCount),
+		   center, legalCenter);
+        // clang-format on
+      }
 
       if (!secondLevel) {
         addFirstLevelSinkDriver(&rootBuffer);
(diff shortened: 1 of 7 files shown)


src/cts/src/HTreeBuilder.cpp
```diff
@@ -123,6 +123,14 @@ void HTreeBuilder::preSinkClustering(
                                   options_->getSinkBuffer(),
                                   legalCenter.getX() * wireSegmentUnit_,
                                   legalCenter.getY() * wireSegmentUnit_);
+      if (logger_->debugCheck(CTS, "legalizer", 2) && (center != legalCenter)) {
+        // clang-format off
+	debugPrint(logger_, CTS, "legalizer", 2,
+		   "preSinkClustering legalizeOneBuffer {}: {} => {}",
+		   baseName + std::to_string(clusterCount),
+		   center, legalCenter);
+        // clang-format on
```
eder-matheus, line 132: Why the `clang-format` comments? Here and in the other debugPrints

src/cts/src/HTreeBuilder.cpp
```diff
@@ -123,6 +123,14 @@ void HTreeBuilder::preSinkClustering(
                                   options_->getSinkBuffer(),
                                   legalCenter.getX() * wireSegmentUnit_,
                                   legalCenter.getY() * wireSegmentUnit_);
+      if (logger_->debugCheck(CTS, "legalizer", 2) && (center != legalCenter)) {
```
eder-matheus, line 126: You don't need the `logger_->debugCheck(CTS, "legalizer", 2)` in the if statement. The `debugPrint` function already checks this for you. Here and in the other debugPrints.

eder-matheus (changes requested):

src/cts/src/HTreeBuilder.cpp
precisionmoon: clang wants to expand debugPrint line into like 10+ lines.  I want to make debug statements compact so this doesn't impact  readability of the rest of the codes.

src/cts/src/HTreeBuilder.cpp
precisionmoon: I want to issue debugPrint only if legalizer moved things (center != legalCenter).  Is there an alternate way without debugCheck?

eder-matheus: You can keep the `center != legalCenter`, but remove the `logger_->debugCheck(CTS, "legalizer", 2)` from the if statement.

precisionmoon: Good idea.  Done.

eder-matheus (approved):
@vvbandeira FYI

vvbandeira (approved):

PR #4057, opened 2023-09-28
fixed Coverity error related to uninitialized db_

precisionmoon:


src/cts/src/TreeBuilder.h
```diff
@@ -59,7 +59,7 @@ class TreeBuilder
               Clock& clk,
               TreeBuilder* parent,
               utl::Logger* logger,
-              odb::dbDatabase* db)
+              odb::dbDatabase* db = nullptr)
```
eder-matheus, line 62: Are there places where this constructor is called without initializing db?

eder-matheus: If not, I don't see why having the `db = nullptr` as the default value.

precisionmoon: Yes, HTreeBuilder is a derived class of TreeBuilder.   TreeBuilder has a member db_ which was initialized properly while HTreeBuilder also had a member db_ which was not.  I removed the redundant member db_ from HTreeBuilder.  This was a good catch from Coverity.  This also explains why I had to reset db for HTreeBuilder in PR 3988.

eder-matheus (approved):

vvbandeira (approved):

PR #4058, opened 2023-09-29
odb: Fixes UBSAN violation in odb classes

QuantamHD:
The initialization of these flags is undefined behavior, and breaks Google binaries which are often build with ubsan by default.

As @rovinski pointed out below this affected more than just two classes. I updated the generator, and added a new CI action that fails if the generator output doesn't match what is at the tip of the submitted branch.

I also stabilized the generator output which was non-hermetic due to the reliance on the dict order which isn't stable across systems.


rovinski:
I think these files were produced from the code generator. The problem probably needs to be identified and fixed there so it doesn't produce bad code.

QuantamHD:
@rovinski Good callout, this is definitely some dangerous code

https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/src/codeGenerator/templates/impl.cpp#L179-L183

QuantamHD:
@rovinski Fixed, and I added a new CI action that makes sure that the generate command matches the current output in the branch.

osamahammad21 (approved):

vvbandeira:
@QuantamHD, good to merge?

vvbandeira (approved):

rovinski:
@QuantamHD that code segment is also used in `dbIStream` and `dbOStream`. Should those be fixed too?

https://github.com/The-OpenROAD-Project/OpenROAD/blob/2dd3af4e2d8ea0272ccdd91f4a40f15ef2345412/src/odb/src/codeGenerator/templates/impl.cpp#L213-L217

https://github.com/The-OpenROAD-Project/OpenROAD/blob/2dd3af4e2d8ea0272ccdd91f4a40f15ef2345412/src/odb/src/codeGenerator/templates/impl.cpp#L237-L241

QuantamHD:
@rovinski I think that code is technically legal. Since there's no assignment happening I think there aren't aliasing problems. I'm not an expert though.

QuantamHD:
@rovinski @maliberty It looks like the whole odb serialization scheme is somewhat broken. The above does not respect endianness. I strongly suspect that if we generated an db file on ARM it would not  work on x86 and vice versa. Is this intentional?

QuantamHD:
Upon looking at this further there's a huge number of problems of what we're doing as a whole. The first problem is obviously the aliasing issues, but the larger problem is that bitfields have no defined order in any compiler, and there's zero way to enforce it.

We have to manually decompose each struct into a integer/long by shifting. We also need to take care to conserve endianness. This is a major rewrite. As it is now the only guarantee we could make is that a file written from a binary will only be compatible with *that binary.*

https://stackoverflow.com/questions/61593309/serializing-deserializing-a-bitfield-structure-in-c

maliberty:
@QuantamHD I'm curious about the original problem.  What was undefined about the previous code?

QuantamHD:
In general casting incompatible pointers creates aliasing problems.

maliberty:
I'm fine with the change but I'm not clear there was an actual failure here which is what I was trying to understand.

QuantamHD:
This was triggered by the undefined behavior sanitizer. Google test targets are built with ubsan by default, and a few Google teams link to OpenROAD rather than using it as an app.

When running their tests they ran into a failure triggered by the aforementioned strict aliasing violations

PR #4066, opened 2023-10-02
flip-flop clustering

therealshreyas:
flip-flop clustering based on https://vlsicad.ucsd.edu/Publications/Conferences/344/c344.pdf

### DIFF
--- a/src/gpl/include/gpl/Replace.h
+++ b/src/gpl/include/gpl/Replace.h
@@ -70,7 +70,6 @@ class InitialPlace;
 class NesterovPlace;
 class Debug;
 
-class MBFF;
 
 class Replace
 {
--- a/src/gpl/include/gpl/Replace.h
+++ b/src/gpl/include/gpl/Replace.h
@@ -39,6 +39,7 @@
 namespace odb {
 class dbDatabase;
 class dbInst;
+
 }  // namespace odb
 namespace sta {
 class dbSta;
@@ -69,6 +70,8 @@ class InitialPlace;
 class NesterovPlace;
 class Debug;
 
+class MBFF;
+
 class Replace
 {
  public:
@@ -83,6 +86,7 @@ class Replace
 
   void doIncrementalPlace();
   void doInitialPlace();
+  void runMBFF(int max_sz, float alpha, float beta, int threads);
 
   int doNesterovPlace(int start_iter = 0);
 
@@ -203,7 +207,6 @@ class Replace
   // temp variable; OpenDB should have these values.
   int padLeft_;
   int padRight_;
-
   bool gui_debug_;
   int gui_debug_pause_iterations_;
   int gui_debug_update_iterations_;
--- a/src/gpl/include/gpl/Replace.h
+++ b/src/gpl/include/gpl/Replace.h
@@ -70,7 +70,6 @@ class InitialPlace;
 class NesterovPlace;
 class Debug;
 
-
 class Replace
 {
  public:
--- a/src/gpl/src/mbff.h
+++ b/src/gpl/src/mbff.h
@@ -32,8 +32,9 @@
 ///////////////////////////////////////////////////////////////////////////////
 #pragma once
 
-#include <vector>
 #include <set>
+#include <vector>
+
 #include "odb/db.h"
 #include "utl/Logger.h"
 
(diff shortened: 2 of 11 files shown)


src/gpl/src/routeBase.cpp
```diff
@@ -377,9 +377,6 @@ void RouteBase::getGlobalRouterResult()
   grouter_->setAllowCongestion(true);
   grouter_->setOverflowIterations(1);
 
-  // this option must be off
-  grouter_->setCriticalNetsPercentage(0);
-
```
maliberty, line 382: How does this relate to this PR?

src/gpl/include/gpl/Replace.h
```diff
@@ -69,6 +70,8 @@ class InitialPlace;
 class NesterovPlace;
 class Debug;
 
+class MBFF;
```
maliberty, line 73: Not needed in this header as it is unused

src/gpl/src/replace.tcl
```diff
@@ -321,6 +321,25 @@ proc global_placement { args } {
...
+proc cluster_flops { args } {
+  sta::parse_key_args "cluster_flops" args \
+    keys { -tray_weight -timing_weight -max_split_size }
+
+  if { [info exists keys(-tray_weight)] } {
+    set tray_weight $keys(-tray_weight)
+    set timing_weight $keys(-timing_weight)
+    set max_split_size $keys(-max_split_size)
+    gpl::replace_run_mbff_cmd $max_split_size $tray_weight $timing_weight
+  }
```
maliberty, line 339: It appears you are treating all the flags as mandatory - otherwise you get a silent no-op.  Is this intentional?

src/gpl/src/replace.cpp
```diff
@@ -296,6 +297,29 @@ void Replace::doInitialPlace()
   ip_->doBicgstabPlace();
 }
 
+void Replace::runMBFF(int max_sz, float alpha, float beta, int threads) {
+
+  int num_flops = 0;
+  int num_paths = 0;
```
maliberty, line 303: num_paths is always 0 - is that right?

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,123 @@
...
+    std::vector<odb::Point> slots;
+    std::vector<int> cand;
+};
+
+struct Flop {
+    odb::Point pt;
+    int idx;
+    float prob;
+
+    bool operator<(const Flop &a) const { return prob < a.prob; }
```
maliberty, line 57: This is only a partial order - is that correct?

maliberty:
I think you lost the or-tools fix we discussed (moving it up to src/)

src/gpl/src/replace.tcl
therealshreyas: Yes -- all these flags must exists or else flip-flop clustering shouldn't run. The user must define alpha, beta and split-size.

maliberty: Are there no reasonable defaults?

therealshreyas: fixed in latest commit (default: tray_weight = 20.0, timing_weight = 1.0, max_split_size = -1)

maliberty:
I clang-formatted your code for you.

src/gpl/src/replace.cpp
maliberty: to be addressed in a subsequent PR

src/gpl/src/mbff.h
maliberty: Does the thumbs up mean you are expecting it to be a partial order?

therealshreyas: Sorry for the confusion -- it is *not* partial order.

src/gpl/src/mbff.h
```diff
@@ -0,0 +1,148 @@
...
+  odb::Point pt;
+  int idx;
+  float prob;
+
+  bool operator<(const Flop& a) const
+  {
+    if (prob == a.prob) {
+        return idx < a.idx;
+    }
+    return prob < a.prob;
```
rovinski, line 65: More conventional.
Requires `#include <tuple>`
```suggestion
    return std::tie(prob, idx) < std::tie(a.prob, a.idx);
```

maliberty (approved):
This is a sufficient baseline to merge.  More work to follow in unit tests and production testing.

PR #4067, opened 2023-10-02
[rsz, pdn]: Fixes Strict Weak Ordering Violations

QuantamHD:
Having non-compliant comparators can lead to crashes, infinite loops, or undefined behavior.

This commit makes the comparators follow strict weak ordering.

See:
https://github.com/bashrc-real/Codearchive/blob/master/cpp/Strict_weak_ordering_and_stl.md

### DIFF
--- a/src/rsz/src/Rebuffer.cc
+++ b/src/rsz/src/Rebuffer.cc
@@ -239,14 +239,27 @@ RepairSetup::rebufferBottomUp(BufferedNetPtr bnet,
     // larger required and smaller capacitance.
     // This is fanout^2.
     // Presort options to hit better options sooner.
-    sort(Z.begin(), Z.end(),
-         [=](BufferedNetPtr option1,
-             BufferedNetPtr option2)
-         { Slack slack1 = slackPenalized(option1);
+    sort(Z.begin(),
+         Z.end(),
+         [=](BufferedNetPtr option1, BufferedNetPtr option2) {
+           Slack slack1 = slackPenalized(option1);
            Slack slack2 = slackPenalized(option2);
-           return fuzzyGreater(slack1, slack2)
-             || (fuzzyEqual(slack1, slack2)
-                 && fuzzyLess(option1->cap(), option2->cap()));
+
+           if (slack1 > slack2) {
+             return true;
+           }
+           if (slack2 > slack1) {
+             return false;
+           }
+
+           if (option1->cap() < option2->cap()) {
+             return true;
+           }
+           if (option2->cap() < option1->cap()) {
+             return false;
+           }
+
+           return false;
          });
     int si = 0;
     for (size_t pi = 0; pi < Z.size(); pi++) {
--- a/src/pdn/src/via.cpp
+++ b/src/pdn/src/via.cpp
@@ -1015,8 +1015,8 @@ bool ViaGenerator::isPreferredOver(const ViaGenerator* other) const
              getName(),
              other->getName());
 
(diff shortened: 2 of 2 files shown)


src/rsz/src/Rebuffer.cc
```diff
@@ -239,14 +239,27 @@ RepairSetup::rebufferBottomUp(BufferedNetPtr bnet,
     // larger required and smaller capacitance.
     // This is fanout^2.
     // Presort options to hit better options sooner.
-    sort(Z.begin(), Z.end(),
-         [=](BufferedNetPtr option1,
-             BufferedNetPtr option2)
-         { Slack slack1 = slackPenalized(option1);
+    sort(Z.begin(),
+         Z.end(),
+         [=](BufferedNetPtr option1, BufferedNetPtr option2) {
```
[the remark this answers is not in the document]
QuantamHD: These are shared pointers.

src/rsz/src/Rebuffer.cc
```diff
@@ -239,14 +239,27 @@
...
```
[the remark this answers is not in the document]
QuantamHD: These are shared pointers.

src/rsz/src/Rebuffer.cc
```diff
@@ -239,14 +239,27 @@ RepairSetup::rebufferBottomUp(BufferedNetPtr bnet,
...
+           Slack slack1 = slackPenalized(option1);
            Slack slack2 = slackPenalized(option2);
-           return fuzzyGreater(slack1, slack2)
-             || (fuzzyEqual(slack1, slack2)
-                 && fuzzyLess(option1->cap(), option2->cap()));
+
+           if (slack1 > slack2) {
+             return true;
+           }
```
maliberty, line 250: This is not semantically equivalent - you dropped all the fuzzy-ness from the comparisons.

QuantamHD: You cannot have fuzzy-ness in comparators.

One of the properties is transitivity of equivalence.

if equiv(a,b) and equiv(b,c) then equiv (a,c) is true. The following example demonstrates a the problem with the fuzzy equiv comparators which is fundamentally what STA does.

```
struct MyComparator {
  bool operator()(double a, double b) const {
    if (std::abs(a - b) <= 1.0) {
      return false;  // Close-enough values are considered equal.
    }
    return a < b;
  }
};

TEST(MyComparator, HasAProblem) {
  std::set<double, MyComparator> s1 = gtl::Container(1, 2, 3);
  std::set<double, MyComparator> s2 = gtl::Container(2, 3, 1);
  EXPECT_EQ(s1.size(), s2.size());
  // The above fails: s1.size() == 2, but s2.size() == 1.
}
```

QuantamHD: In the example above comp(1,2) is equal and comp(2,3) is equal, but comp(1,3) is not equal.

maliberty: Ok but this will need a full CI testing as it changes the semantics (for the better but even so).

QuantamHD: Sounds good. This one is high priority if you can make sure to run it soon.

maliberty: running

maliberty (approved):
ci passed

maliberty:
@gadfort any comments on the pdn change?

gadfort:
@maliberty seems reasonable to me

PR #4071, opened 2023-10-03
ppl: fix check for available positions in the top layer

eder-matheus:
Address issues reported in https://github.com/The-OpenROAD-Project/OpenROAD/discussions/4061.


maliberty (approved):
It looks like we need a unit test for the top layer pins case as some test should have new results.

PR #4072, opened 2023-10-03
gui: fix save image black output when using small resolution

### LINKED ISSUE #4000: `save_image` outputs empty black image if size is too big
### Describe the bug

It seems like `save_image` will fail to create a proper image when the output file width / height gets too big (by setting `-resolution` too small). Instead you seem to get an empty black image. 

It doesn't look like OpenROAD is failing or segfaulting. You get a black image that is of roughly the size you would expected.

The same database file produces valid images at lower image sizes.

### Expected Behavior

You get an image.

### OpenROAD Environment

```shell
I believe commit hash 65053db0233ffb6e3235b3f28774afbcd063d569
```

### OpenLane Environment

```shell
N/A - Using `bazel_hdl_rules`
```

### To Reproduce

Failing pixel image output size is 3816 x 7200, while a working image size is 1100 x 2075.

`cmd script` is

```tcl
read_db "add_sky130_pnr_straight__export_def_global_placement.db"
save_image -resolution 0.01 "add_sky130_pnr_straight__export_def_global_placement.png"
puts "Wrote add_sky130_pnr_straight__export_def_global_placement.png"
```

### Relevant log output

```shell
INFO: Running command line: blaze-bin/third_party/open_road/openroad -exit /tmp/tmpc_t7kbcl.tcl
OpenROAD 1.0.0-Google bazel_rules_hdl
Wrote /google/obj/workspace/a980c46c0b22c44d96ddb3e865bf9b7262f4b004a10a578f713293a558ad5b35/7178466f-ed9b-41c1-9fc2-2a96c3a140a9/blaze-out/k8-fastbuild/bin/experimental/users/tansell/adders/add_sky130_pnr_straight__place_pins.png
```

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
Resolves #4000 

What was happening was: If the user put a resolution small enough to make `Utils::adjustMaxImageSize` resize the image, we were not adapting the resolution again based on this adjusted size.

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -1926,7 +1926,7 @@ void LayoutViewer::saveImage(const QString& filepath,
                                       screen_region.width(),
                                       screen_region.height());
 
-  const QRect bounding_rect = save_region.boundingRect();
+  QRect bounding_rect = save_region.boundingRect();
 
   // We don't use Utils::renderImage as we need to have the
   // rendering be synchronous.  We directly call the draw()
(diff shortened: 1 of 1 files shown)


maliberty (approved):

rovinski:
Where does the 7200 px limit come from?

maliberty:
@rovinski https://github.com/The-OpenROAD-Project/OpenROAD/blob/03f219802d0ee569eb86846529260e15a020c26c/src/gui/src/gui_utils.cpp#L86

"~1.5GB in memory max @ 7200"

I recall we had somebody make an error and we tried to dump a many GB file so we set a 'reasonable' cap.

rovinski:
Gotcha. Should probably be:
```
    logger_->warn(utl::GUI,
                  94,
                  "Can't save image with the specified size (max width/height "
                  "is {} pixels). Saved image dimensions = {} x {}.",
                  max_size,
                  img_size.width(),
                  img_size.height());
```
@AcKoucher

AcKoucher:
@rovinski I'm working on adding support for target image dimensions.  I'll include this correction in the PR.

AcKoucher:
@rovinski I was thinking here... Would it really be necessary to make this change?

I'm asking, because as the variable that holds the 7200 limit is created inside `Utils::adjustImageSize` and `Utils` is a class of static methods, I don't see a good alternative to where we could cache the limit.

AcKoucher:
@rovinski Nevermind... we can just use std::max dimension of the new size if the image was adjusted...

rovinski:
I was just thinking we shouldn't have 2 sources of truth in case the limit gets changed in one place but not the other. Whatever works best to do that.

Issue #4069, opened 2023-10-03
Add support for saving IO placement


oharboe:
### Description

Add support in GUI to manually move and to save the IO placement.

Initially, a constraint file might be written, but in later stages, manually moving pins can be quite useful.

### Suggested Solution

1. Add possibly to move pin, snapping to grids such that detailed placement and other stages do not run into trouble.
2. add `write_io_placement foo.tcl` command.

rovinski:
@eder-matheus can PPL be adjusted to make `write_pin_placement` a separate command?

eder-matheus:
> @eder-matheus can PPL be adjusted to make `write_pin_placement` a separate command?

Done

maliberty:
Turn OR into an interactive layout editing system is an endless task that we have avoided.  This and #4068 are just the start once we open this door.  Is it possible to define improvements that would obviate the need to this manually?

oharboe:
@maliberty I just spoke with @louiic: he spoke of constraints to place macros. I think that coming up with some additional constraints is more idiomatic for OpenROAD's approach than to start down the path of GUI editing features...

That said, I think a write_io_placement and write_macro_placement Tcl commands is a good idea. For macros and io placement, this can be used to do some tinkering in text editors and also to speed up the flow by skipping macro and io placement...

maliberty:
Eder already did write_pin_placement in #4073 so I'll close this.  write_macro_placement wouldn't be hard and I'll revise #4068 to be just that.

mpl2 has some support for constraints that we could build on.

PR #4073, opened 2023-10-03
ppl: add write_pin_placement command

eder-matheus:
Partially cover issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/4069


src/ppl/src/Netlist.cpp
```diff
@@ -184,6 +185,7 @@ void Netlist::clear()
   io_pins_.clear();
   io_groups_.clear();
   _db_pin_idx_map.clear();
+  net_pointer_.push_back(0);
```
maliberty, line 188: Why is this required?

eder-matheus: This is required because I've moved the call for the Netlist::clear() to the initialization of the netlist data. The `net_pointer_` vector is cleared above, but it needs to start with 0.

Perhaps renaming this function to Netlist::reset() will make things clearer since the idea is to restore the Netlist state to the beginning of the PPL flow execution in cases where PPL is called multiple times in a single run.

maliberty (approved):

src/ppl/README.md
```diff
@@ -255,7 +255,19 @@ microns to 60.5 microns, and the left edge from its beginning to 50 microns.
...
+
+```tcl
+write_pin_placement -file_name file_name
+```
+
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-file_name` | The name of the file with the pin placement. |
```
rovinski, line 270: Seems like this should just be a positional argument?

src/ppl/src/IOPlacer.cpp
```diff
@@ -684,14 +683,14 @@ int IOPlacer::micronsToDbu(double microns)
...
 {
-  std::string file_name = parms_->getPinPlacementFile();
-  if (file_name.empty()) {
+  std::string filename = file_name;
+  if (filename.empty()) {
     return;
   }
 
-  std::ofstream out(file_name);
+  std::ofstream out(filename);
```
rovinski, line 693: Does an error get thrown if the file can't be opened?

maliberty: It isn't related to this PR but it is a good point.  We should check 'if (!out)'  ...

src/ppl/README.md
maliberty: Agreed

PR #4078, opened 2023-10-04
ord: Some more timing APIs in python

VidyaChhabria:
Some more sample timing APIs for Python


src/Design.cc
```diff
@@ -208,6 +208,35 @@ bool Design::isSequential(odb::dbMaster* master)
   return lib_cell->hasSequentials();
 }
 
+bool Design::isInClock(odb::dbInst* inst) {
+  for (auto* iterm : inst->getITerms()) {
+    auto* net = iterm->getNet();
+    if (net != NULL && net->getSigType() == odb::dbSigType::CLOCK){
```
maliberty, line 214: Note that this will only work post-CTS.  If you want it to work earlier you'll have to query sta directly.

src/Design.cc
```diff
@@ -208,6 +208,35 @@ bool Design::isSequential(odb::dbMaster* master)
...
+      return true;
+    }
+  }
+  return false;
+}
+
+std::uint64_t Design::getNetRoute(odb::dbNet* net) {
```
maliberty, line 221: getNetRoute is rather vague, perhaps getNetRoutedLength?

Btw, what would you do with the power supply wire length?

maliberty (changes requested):
Please address the clang-tidy items as well.

maliberty:
Note that DCO & clang-format are also failing.

maliberty:
@VidyaChhabria this is pending on your to resolve the issues

src/Design.cc
```diff
@@ -190,24 +190,104 @@ float Design::getNetCap(odb::dbNet* net, sta::Corner* corner, MinMax minmax)
...
+float Design::staticPower(odb::dbInst* inst, sta::Corner* corner)
+{
+  sta::dbSta* sta = getSta();
+  sta::dbNetwork* network = sta->getDbNetwork();
+
+  sta::Instance* sta_inst = network->dbToSta(inst);
+  if (!sta_inst) {
+    return 0.0;
+  }
+  // sta::LibertyCell* lib_cell = network->libertyCell(sta_inst);
```
maliberty, line 241: rm

src/Design.cc
```diff
@@ -190,24 +190,104 @@ float Design::getNetCap(odb::dbNet* net, sta::Corner* corner, MinMax minmax)
...
+float Design::dynamicPower(odb::dbInst* inst, sta::Corner* corner)
+{
+  sta::dbSta* sta = getSta();
+  sta::dbNetwork* network = sta->getDbNetwork();
+
+  sta::Instance* sta_inst = network->dbToSta(inst);
+  if (!sta_inst) {
+    return 0.0;
+  }
+  // sta::LibertyCell* lib_cell = network->libertyCell(sta_inst);
```
maliberty, line 255: rm

src/Design.cc
```diff
@@ -190,24 +190,104 @@ float Design::getNetCap(odb::dbNet* net, sta::Corner* corner, MinMax minmax)
...
+  sta::PowerResult power = sta->power(sta_inst, corner);
+  return (power.internal() + power.switching());
+}
+
+bool Design::isInClock(odb::dbInst* inst)
```
maliberty, line 260: Do you intend for FF to return true based on the clock pin?

maliberty (changes requested):

maliberty (approved):

src/Design.cc
VidyaChhabria: We will include FF as well. So it will return true. As long as is_seq is true we should be good CircuitOps can differentiate them.

PR #4086, opened 2023-10-04
grt: add check for adjacent layers direction

eder-matheus:
Following discussion in https://github.com/The-OpenROAD-Project/OpenROAD/discussions/4080

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -180,6 +180,7 @@ std::vector<Net*> GlobalRouter::initFastRoute(int min_routing_layer,
   }
 
   initRoutingLayers();
+  checkAdjacentLayersDirection(min_routing_layer, max_routing_layer);
   reportLayerSettings(min_routing_layer, max_routing_layer);
   initRoutingTracks(max_routing_layer);
   initCoreGrid(max_routing_layer);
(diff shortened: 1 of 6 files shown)


src/grt/src/GlobalRouter.cpp
```diff
@@ -498,6 +499,28 @@ void GlobalRouter::initRoutingLayers()
   fastroute_->setLayerOrientation(vertical);
 }
 
+void GlobalRouter::checkAdjacentLayersDirection(int min_routing_layer,
+                                                int max_routing_layer)
+{
+  odb::dbTech* tech = db_->getTech();
+  for (int l = min_routing_layer; l < max_routing_layer; l++) {
+    odb::dbTechLayer* layer_a = tech->findRoutingLayer(l);
+    odb::dbTechLayer* layer_b = tech->findRoutingLayer(l + 1);
```
maliberty, line 508: should the loop end at max_routing_layer - 1 ?

src/grt/src/GlobalRouter.cpp
```diff
@@ -498,6 +499,28 @@ void GlobalRouter::initRoutingLayers()
...
+    if (layer_a->getDirection() == layer_b->getDirection()) {
+      std::string dir = layer_a->getDirection() == odb::dbTechLayerDir::VERTICAL
+                            ? "VERTICAL"
+                            : "HORIZONTAL";
```
maliberty, line 512: you can use layer_a->getDirection().getString()

src/grt/src/GlobalRouter.cpp
eder-matheus: No. The range is [min_routing_layer, max_routing_layer], so when `l` is `max_routing_layer - 1`, it will get the correct layers without going out of the range.

maliberty (approved):

PR #4089, opened 2023-10-04
ppl: small fixes

eder-matheus:
Address comments from other PRs.


src/ppl/src/IOPlacer.cpp
```diff
@@ -693,6 +693,10 @@ void IOPlacer::writePinPlacement(const char* file_name)
 
   std::ofstream out(filename);
 
+  if (!out.is_open()) {
```
maliberty, line 696: it is simpler and more robust to just use "if (!out)" as that also checks for a good state.

maliberty (approved):

PR #4097, opened 2023-10-05
gui: add support for entering -width in save_image

### LINKED ISSUE #3999: It would be nice for `save_image` to support a target image width (or height) specified in pixels
### Description

Currently `save_image` has a `-resolution` argument which is microns per pixel. This means if you want an image a certain number of pixels wide, you have to calculate the value by first extracting the size of your design in microns and then working out a resolution value which gives you the right number of pixels.

It also took me a while to figure out it was microns per pixel and not pixels per micron and thus for 0.13 micron process I'm going to want a value very small number like `0.1` for a high resolution image (and **not** a big number like `10,000`).

### Suggested Solution

It would be nice if you could just specify the image output width in pixels.

`save_image -width 10000px` or `save_image -height 10000px`

### Additional Context

_No response_

AcKoucher:
Resolves #3999

@rovinski FYI, I'm making the correction you pointed out by creating a cache of the max size.

### DIFF
--- a/src/gui/README.md
+++ b/src/gui/README.md
@@ -62,6 +62,7 @@ This command can be both be used when the GUI is active and not active.
 ```
 save_image [-resolution microns_per_pixel]
            [-area {x0 y0 x1 y1}]
+           [-width width]
            [-display_option {option value}]
            filename
 ```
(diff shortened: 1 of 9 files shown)


maliberty:
In the C++20 build
layoutViewer.cpp:(.text+0xf7f3): undefined reference to `gui::Utils::MAX_IMAGE_SIZE'

rovinski:
Update docs with new flag?

maliberty (approved):

PR #4100, opened 2023-10-05
grt: fix coverity issues

eder-matheus:


src/grt/src/Grid.cpp
```diff
@@ -180,19 +180,25 @@ int Grid::computeTileReduce(const odb::Rect& obs,
   int reduce = -1;
   if (direction == odb::dbTechLayerDir::VERTICAL) {
     if (obs.xMin() >= tile.xMin() && obs.xMax() <= tile.xMax()) {
-      reduce = ceil(std::abs(obs.xMax() - obs.xMin()) / track_space);
+      reduce = ceil(static_cast<double>(std::abs(obs.xMax() - obs.xMin()))
+                    / track_space);
```
maliberty, line 184: Kinda long and repeated.  I think it would be simpler to do something like delay the common part to the end

int obs_size;
if ....
  if ...
      obs_size = std::abs(obs.xMax() - obs.xMin());
...
  return ceil(obs_size / static_cast<double>(track_space));

maliberty: You could take track_space as a double argument since you only use it in that form

maliberty (approved):

PR #4116, opened 2023-10-10
odb: support for PROPERTY LEF58_FORBIDDENSPACING

eder-matheus:


eder-matheus:
@maliberty Jenkins was failing, but nothing was wrong with the code. I think this PR is ready for review.

src/odb/test/data/design.odb
maliberty, line 1: This should not be updated - its purpose is to prove that we can still load old .odb files.

src/odb/include/odb/db.h
```diff
@@ -8895,6 +8899,38 @@ class dbTechLayerEolKeepOutRule : public dbObject
   // User Code End dbTechLayerEolKeepOutRule
 };
 
+class dbTechLayerForbiddenSpacingRule : public dbObject
```
maliberty, line 8902: There is no way to tell if the optional fields have been given in the LEF or not.  Usually we have has<field>() methods (see dbTechLayerSpacingRule for an example)

src/odb/test/data/design.odb
eder-matheus: I've updated it because it was failing. I'll try reverting it and check if the error persists.

src/odb/include/odb/db.h
eder-matheus: I've added in the following way:
image

I'm comparing against the default value of those fields. Is this the correct way to do it?

src/odb/src/db/dbTechLayer.cpp
```diff
@@ -740,6 +752,7 @@ dbIStream& operator>>(dbIStream& stream, _dbTechLayer& obj)
   stream >> *obj.width_table_rules_tbl_;
   stream >> *obj.min_cuts_rules_tbl_;
   stream >> *obj.area_rules_tbl_;
+  stream >> *obj.forbidden_spacing_rules_tbl_;
```
maliberty, line 755: You need to make this conditional of the schema - see db_schema_keepout_zone below

eder-matheus: Thanks! That fixed the read_db unit test.

maliberty:
Failing "Check That ODB Files Are Generated "

eder-matheus:
> Failing "Check That ODB Files Are Generated "

Fixed. I didn't notice the missing "schema" field for the new rule.

maliberty (approved):

PR #4120, opened 2023-10-11
mpl2: add command for writing macro placement

### LINKED ISSUE #4068: Add saving of macro placement
### Description

Add support for moving and rotating macros from the GUI, snapping to the grid as well as saving of macro placement.

### Suggested Solution

1. grab and move macro, by default snapping to the grid so that detailed router and other stages do not run into trouble.
2. add a `write_placement foo.tcl` command.

### Additional Context

_No response_

AcKoucher:
Resolves #4068 

In this PR I'm also implementing the command `place_macro` to place macros individually.
The output file of write_macro_placement will be written in the format of `place_macro` commands
as we currently do in PPL.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -643,7 +643,7 @@ void HierRTLMP::hierRTLMacroPlacer()
 
   correctAllMacrosOrientation();
 
-  writeMacroPlacement(macro_placement_file_.c_str());
+  writeMacroPlacement(macro_placement_file_);
 
   // Clear the memory to avoid memory leakage
   // release all the pointers
@@ -6066,23 +6066,21 @@ void HierRTLMP::correctAllMacrosOrientation()
   }
 }
 
-void HierRTLMP::setMacroPlacementFile(const char* file_name)
+void HierRTLMP::setMacroPlacementFile(const std::string& file_name)
 {
   macro_placement_file_ = file_name;
 }
 
-void HierRTLMP::writeMacroPlacement(const char* file_name)
+void HierRTLMP::writeMacroPlacement(const std::string& file_name)
 {
-  std::string filename = file_name;
-
-  if (filename.empty()) {
+  if (file_name.empty()) {
     return;
   }
 
-  std::ofstream out(filename);
+  std::ofstream out(file_name);
 
   if (!out) {
-    logger_->error(MPL, 11, "Cannot open file {}.", filename);
+    logger_->error(MPL, 11, "Cannot open file {}.", file_name);
   }
 
   // Use only insts that were placed by mpl2
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -6085,8 +6085,8 @@ void HierRTLMP::writeMacroPlacement(const std::string& file_name)
 
   // Use only insts that were placed by mpl2
   for (auto& [inst, hard_macro] : hard_macro_map_) {
-    float x = dbuToMicron(inst->getLocation().x(), dbu_);
-    float y = dbuToMicron(inst->getLocation().y(), dbu_);
+    const float x = dbuToMicron(inst->getLocation().x(), dbu_);
+    const float y = dbuToMicron(inst->getLocation().y(), dbu_);
 
     out << "place_macro -macro_name " << inst->getName() << " -location {" << x
         << " " << y << "} -orientation " << inst->getOrient().getString()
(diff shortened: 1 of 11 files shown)


AcKoucher:
There was a check failing in the CI which I thought it could be a problem from the CI itself, but it was actually a conflict between mpl2's new messages and the old mpl's messages IDs.

Ready for review.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -6064,6 +6066,36 @@ void HierRTLMP::correctAllMacrosOrientation()
...
 }
 
+void HierRTLMP::setMacroPlacementFile(const char* file_name)
+{
+  macro_placement_file_ = file_name;
+}
+
+void HierRTLMP::writeMacroPlacement(const char* file_name)
+{
+  std::string filename = file_name;
```
maliberty, line 6076: Why not

void HierRTLMP::writeMacroPlacement(const std::string& file_name)

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -6064,6 +6066,36 @@ void HierRTLMP::correctAllMacrosOrientation()
...
+  std::ofstream out(filename);
+
+  if (!out) {
+    logger_->error(MPL, 11, "Cannot open file {}.", filename);
+  }
+
+  // Use only insts that were placed by mpl2
+  for (auto& [inst, hard_macro] : hard_macro_map_) {
+    float x = dbuToMicron(inst->getLocation().x(), dbu_);
+    float y = dbuToMicron(inst->getLocation().y(), dbu_);
```
maliberty, line 6091: const

src/mpl2/src/mpl.tcl
```diff
@@ -252,8 +258,70 @@ proc rtl_macro_placer { args } {
     return true
 }
 
+sta::define_cmd_args "place_macro" {[-macro_name macro_name] \
+                                    [-location location] \
+                                    [-orientation orientation] \
```
maliberty, line 263: mandatory options should not be in [] which denotes optional

src/mpl2/src/rtl_mp.cpp
```diff
@@ -121,6 +123,95 @@ bool MacroPlacer2::place(const int max_num_macro,
...
+    orientation = odb::dbOrientType::R0;
+  } else if (orientation_string == "MY") {
+    orientation = odb::dbOrientType::MY;
+  } else if (orientation_string == "MX") {
+    orientation = odb::dbOrientType::MX;
+  } else if (orientation_string == "R180") {
+    orientation = odb::dbOrientType::R180;
+  } else {
+    logger_->error(MPL, 33, "Invalid orientation {}.", orientation_string);
+  }
```
maliberty, line 141: dbOrientType::dbOrientType(const char* orient);  already does this for you.

src/mpl2/src/rtl_mp.cpp
```diff
@@ -121,6 +123,95 @@ bool MacroPlacer2::place(const int max_num_macro,
...
+
+  return orientation;
+}
+
+void MacroPlacer2::placeMacro(odb::dbInst* inst,
+                              float x_origin,
+                              float y_origin,
+                              const std::string& orientation_string)
```
maliberty, line 149: Its better to take a dbOrientType here rather than a string

maliberty (changes requested):

src/mpl2/src/rtl_mp.cpp
AcKoucher: This function returns only the orientations we allow in mpl2. (we don't allow rotation so, `R90`, `R270` and `MYR90`, `MXR90` are not allowed).

If I used the `dbOrientType::dbOrientType(const char* orient)` I would have to check for the illegal ones either way.

maliberty: While we wouldn't generate a placement with a 90 degree rotation there is no reason a user couldn't chose to that if they knew it was ok.  I think it is at most a warning.

In general I'd prefer to avoid duplication and check the dbOrientType.  You could add a a method there for isRightAngleRotation() that could be general purpose.

src/mpl2/src/hier_rtlmp.cpp
AcKoucher: I thought I was having some sort of compilation problem using std::string, but, apparently, it was related to the C++ swig methods.

I'll change it.

I'll change in `setMacroPlacementFile(const char* file_name)` as well.

maliberty: You can use a string in swig as well but you need to have
```
%include <std_string.i>
```
in the .i

maliberty (approved):

PR #4122, opened 2023-10-11
gui: use one control for drawing bterms' marker, geometry and text in the layout

### LINKED ISSUE #4050: OR gui: add a switch to be able to toggle PIN geometry
### Description

Currently a def PIN and NET are both displayed depending on the Net switch in the gui.

Trying to debug https://github.com/KLayout/klayout/issues/1490 I would have liked to be able to switch the pins off and see where the net is.

### Suggested Solution

add pin switch

### Additional Context

_No response_

AcKoucher:
#Resolves #4050

Also renaming "Misc->Instances->Pin labels" to "Misc->Instances->Pin Name".

### DIFF
--- a/src/gui/src/displayControls.cpp
+++ b/src/gui/src/displayControls.cpp
@@ -457,7 +457,7 @@ DisplayControls::DisplayControls(QWidget* parent)
   makeLeafItem(
       instance_shapes_.pins, "Pins", instance_shape, Qt::Checked, true);
   makeLeafItem(instance_shapes_.iterm_labels,
-               "Pin labels",
+               "Pin Names",
                instance_shape,
                Qt::Unchecked,
                false,
--- a/src/gui/src/displayControls.cpp
+++ b/src/gui/src/displayControls.cpp
@@ -423,9 +423,9 @@ DisplayControls::DisplayControls(QWidget* parent)
   makeParentItem(site_group_, "Rows", root, Qt::Unchecked, true);
 
   // Rows
-  makeParentItem(pin_markers_, "Pin Markers", root, Qt::Checked);
+  makeParentItem(io_pins_, "Pins", root, Qt::Checked);
   pin_markers_font_ = QApplication::font();  // use default font
-  setNameItemDoubleClickAction(pin_markers_, [this]() {
+  setNameItemDoubleClickAction(io_pins_, [this]() {
     pin_markers_font_ = QFontDialog::getFont(
         nullptr, pin_markers_font_, this, "Pin marker font");
   });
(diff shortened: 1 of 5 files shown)


maliberty (approved):

maliberty:
@AcKoucher it is better to use "Fixes #4050" rather than #Resolves #4050" that way GH will automatically close the issue when the PR is merged.

stefanottili:
a) one can't distinguish pin from net geometry
b) viewing hierarchical def, the distinction between "Pin" and "Misc->Instances->Pin" is getting blurry, no ?

openroad -gui mock-array.or

cat mock-array.or:
read_lef platforms/asap7/lef/asap7_tech_1x_201209.lef
read_lef platforms/asap7/lef/asap7sc7p5t_28_R_1x_220121a.lef
read_lef results/asap7/mock-array_Element/base/Element.lef
read_def results/asap7/mock-array/base/6_final.def
read_def -child results/asap7/mock-array_Element/base/6_final.def

maliberty:
Turn off the net and you'll see the pin.  Often you want to see the net as a whole and totally separating the pin would be confusing (eg focus nets, selection).

stefanottili:
a) Nets on, Pins off: you still see the PIN geometry, THAT's confusing.

b) Look at the pins of mock-array_Element.
Switch of NET and neither Pins nor Misc->Instances->Pins will draw the PIN geometry as if there is no pin geometry at all.
Misc->Instances->Pins will toggle the pin name, so if it's off it doesn't matter that Misc->Instances->Pins Name is on, it won't show the pin name.

maliberty:
I dislike that the pins are not in the layer color as that loses information.  I am not in favor of the images you show.

gds doesn't really have the concept of a net so I'm not sure klayout is a good guide.  Does it support net selection?

stefanottili:
klayout reads lef/def and distingishes between PIN and NET.

In the case of sky130 the lefdef2gds map maps, PIN geometry to both net and pin layer and the pin name to a third.
So yes, it becomes trickier in gds, but even there one can distinguish the pin geometry.

I'm not saying that I want different colors for PIN and NET geometry.
This was just done to show that in this case the NET doesn't overlap the PIN.

I would argue that the PINS switch should be toggling PIN geometry and the NETS switch NET geometry and not NET + PIN.

And looking at hierarchical def PIN's: even when both Pins and Misc->Instances->Pins are on, you won't see any PIN geometry when Nets is off.

maliberty:
How in klayout do you select a net?  Whenever I click a shape it just selects the individual shape.

maliberty:
There are a number of net types with their own visibility controls.  If someone is only showing clock nets should we show all pins or only clock pins?  I would find it less helpful to show all.  In this case net visibility controls pin visibility.  How do you think this should be handled?

PR #4130, opened 2023-10-13
mpl2: corrections in place_macro

AcKoucher:
I'm making three corrections here:
1. I used wrongly the function to snap the macro. The name I chose for the function was misleading its use, so I changed it. I'm using the function correctly now.

2. I realized then that this same function would not work if a macro have pins in only one direction (e. g. `src/mpl2/test/mp_test1.tcl`), so now we calculate the coordinate of the snap origin only when we actually have pins in that direction.

3. Finally, now we also don't try to snap the macro placed by `place_macro` if it has a rotation which is not allowed in HierRTLMP.

Also:
We're now using not only iterms, but bterms in the wirelength computation for adjusting orientation.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -6058,7 +6058,7 @@ void HierRTLMP::correctAllMacrosOrientation()
 
     const odb::dbOrientType inst_orientation = inst->getOrient();
 
-    const odb::Point snap_origin = hard_macro->alignOriginWithGrids(
+    const odb::Point snap_origin = hard_macro->computeSnapOrigin(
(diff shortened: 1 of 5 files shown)


maliberty:
I think this should go through a secure CI.  

One other thing I noticed is that the wirelength computation uses iterms but not bterms.  You might fix that before starting the CI.

AcKoucher:
@maliberty  Secure-CI passed.

maliberty:
gf12/ca53 is happening on master and Osama is looking at it.

gf12/bp_single on master has 2 drvs currently not 10.  I'll take a quick look at the new violations.

maliberty (approved):
The code changes look fine

PR #4135, opened 2023-10-16
Change ripupMode to an enum

refaay:


src/drt/src/TritonRoute.cpp
```diff
@@ -607,14 +620,29 @@ void TritonRoute::stepDR(int size,
...
+    case 0:
+      tempMode = fr::RipUpMode::ripupDRC;
+      break;
+    case 1:
+      tempMode = fr::RipUpMode::ripupAll;
+      break;
+    case 2:
+      tempMode = fr::RipUpMode::ripupAroundDRC;
+      break;
+  }
```
rovinski, line 637: This code is used twice; you can wrap it in a function.

src/drt/src/frBaseTypes.h
```diff
@@ -287,6 +287,14 @@ enum class frAccessPointEnum
   NearbyGrid = 4  // nearby grid or 1/2 grid
 };
 
+enum class RipUpMode
+{
+  UNKNOWN = -1,
+  ripupDRC = 0,
+  ripupAll = 1,
+  ripupAroundDRC = 2
```
rovinski, line 295: If the value is already part of the `RipUpMode` enum, I think it's a bit redundant to include `ripup` in the value name. e.g.
```
  UNKNOWN = -1,
  DRC = 0,
  ALL = 1,
  AROUND_DRC = 2
```

I think the style guidelines also prefer enums to be all capital letters.
Then the value would simply be e.g. `RipUpMode::DRC`

src/drt/src/TritonRoute.cpp
```diff
@@ -190,7 +190,20 @@ void TritonRoute::setDebugWorkerParams(int mazeEndIter,
   debug_->markerCost = markerCost;
   debug_->fixedShapeCost = fixedShapeCost;
   debug_->markerDecay = markerDecay;
-  debug_->ripupMode = ripupMode;
+  switch (ripupMode) {
+    case -1:
+      debug_->ripupMode = fr::RipUpMode::UNKNOWN;
```
rovinski, line 195: Is it necessary to use the `fr::` namespace here and everywhere else? I think it doesn't need it if it's already in the fr namespace. If it does, I might recommend using an alias for simplicity.

src/drt/src/frBaseTypes.h
```diff
@@ -287,6 +287,14 @@ enum class frAccessPointEnum
   NearbyGrid = 4  // nearby grid or 1/2 grid
 };
 
+enum class RipUpMode
+{
+  UNKNOWN = -1,
```
osamahammad21, line 292: why do we need this UNKNOWN value?

src/drt/src/dr/FlexDR.cpp
```diff
@@ -956,75 +957,75 @@ void FlexDR::end(bool done)
...
+    {7, -5, 32, 32 * shapeCost,  8 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}, // 55
+    {7, -6, 32, 32 * shapeCost,  8 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}, // 56
+    {3, -1,  8,      shapeCost,      MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::ALL, false}, // 57
+    {7,  0, 64,      shapeCost,      MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::NEARDRC, false}, // 58
+    {7, -1, 64, 64 * shapeCost, 16 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}, // 59
+    {7, -2, 64, 64 * shapeCost, 16 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}, // 60
+    {7, -3, 64, 64 * shapeCost, 16 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}, // 61
+    {7, -4, 64, 64 * shapeCost, 16 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}, // 62
+    {7, -5, 64, 64 * shapeCost, 16 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}, // 63
+    {7, -6, 64, 64 * shapeCost, 16 * MARKERCOST, 100 * shapeCost, 0.999, RipUpMode::DRC, false}  // 64
```
osamahammad21, line 1028: Could you add the needed indentation for the rip up mode so that they are all aligned? makes it easier to read.

osamahammad21 (changes requested):

src/drt/src/gr/FlexGR.cpp
```diff
@@ -1165,10 +1180,10 @@ bool FlexGR::initGR_patternRoute_route_iter(
...
       switch (mode) {
-        case 0:
+        case RipUpMode::DRC:
           patternRoute_LShape(startNode, endNode);
           break;
-        case 1:
+        case RipUpMode::ALL:
           break;
         default:;
       }
```
osamahammad21, line 1189: I don't think This should be the same RipUpModeENUM. No DRC in GR

osamahammad21 (approved):

maliberty (approved):

PR #4136, opened 2023-10-16
rsz: ensure recover_power will pick a smaller cell

eder-matheus:


src/rsz/src/RecoverPower.cc
```diff
@@ -353,8 +353,8 @@ RecoverPower::meetsSizeCriteria(LibertyCell *cell, LibertyCell *equiv,
     if (!match_size) {
       return true;
     }
-    dbMaster* lef_cell1 = db_network_->staToDb(cell);
-    dbMaster* lef_cell2 = db_network_->staToDb(equiv);
+    dbMaster* lef_cell1 = db_network_->staToDb(equiv);
+    dbMaster* lef_cell2 = db_network_->staToDb(cell);
     if (lef_cell1->getWidth() <= lef_cell2->getWidth()) {
```
maliberty, line 358: We should also check the height to avoid swapping a single height for a double height.

maliberty (approved):

PR #4139, opened 2023-10-17
DPO bug fixes for MPD

ahmadelrouby:
fixed two issues in DPO that appeared while testing multi power domain

- One issue that couldn't find any segments for instances belonging to a fenced domain
- The other issue was related to min cost flow calculation, double to int casting was overflowing resulting in infinite execution.


rovinski:
The sky130 files should link to the ones already in `/test`. Please force push to remove.

maliberty:
The code looks ok.  This seems to be one test that is spread/copied to multiple modules gpl/dpl/dpo.  I think it would be better to turn it into a single test for the application.  Unit tests for one tool should not involve running other tools (eg dpl should not require running global_placement).

ahmadelrouby:
@maliberty I must have missed those ones. I've removed them as well now.

maliberty (approved):

PR #4152, opened 2023-10-18
rsz: call incremental groute for repair_timing -recover_power

eder-matheus:
FIxes error with the detailed router when using recover_power.

### DIFF
--- a/src/rsz/src/EstimateWireParasitics.cc
+++ b/src/rsz/src/EstimateWireParasitics.cc
@@ -296,9 +296,8 @@ Resizer::ensureWireParasitic(const Pin *drvr_pin,
       parasitics_invalid_.erase(net);
       break;
     case ParasiticsSrc::global_routing: {
-      grt::IncrementalGRoute incr_groute(global_router_, block_);
       global_router_->addDirtyNet(db_network_->staToDb(net));
-      incr_groute.updateRoutes();
+      incr_groute_->updateRoutes();
       global_router_->estimateRC(db_network_->staToDb(net));
       parasitics_invalid_.erase(net);
       break;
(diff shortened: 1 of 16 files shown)


eder-matheus:
@maliberty I will apply clang-format after the code review, since it changes a lot of other places unrelated to this PR.

eder-matheus:
@maliberty Now it is ready for review. GF55/aes is passing with this branch, and a new secure-ci run was started.

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -296,9 +296,8 @@ Resizer::ensureWireParasitic(const Pin *drvr_pin,
       parasitics_invalid_.erase(net);
       break;
     case ParasiticsSrc::global_routing: {
-      grt::IncrementalGRoute incr_groute(global_router_, block_);
       global_router_->addDirtyNet(db_network_->staToDb(net));
-      incr_groute.updateRoutes();
+      incr_groute_->updateRoutes();
```
maliberty, line 300: Is the call to addDirtyNet still needed?

eder-matheus: The unit test that checks it didn't fail when I removed it, so I believe it is ok to remove. Added a new secure-ci to ensure it will no break anything.

eder-matheus:
@maliberty Public CI is green, and gf55/aes is passing locally. For some reason, the secure-ci pipeline is not starting, and I'm not sure how long it will take to start.

maliberty:
Secure CI is just a bit backed up but I see one pipeline has started.  There is only one private design with recover-power enabled so public CI seems good enough.

maliberty (approved):

PR #4156, opened 2023-10-19
dpl: avoid moving legal cell

eder-matheus:


eder-matheus:
@luis201420 I thing you're using this function for incremental groute. Does this change make sense with your use case?

maliberty:
My recollection was there was a case where we upsized a cell on the edge of the core area and then the cell then extended outside the core area.  grt couldn't route to those pins and we had an error.  We legalized the cell to avoid such.

maliberty (approved):

Issue #455, opened 2020-07-23
Router is assuming all the pins to be in the first layer


mukul19147:
Even if I supply pins in the different layer the router is assuming them in the first layer and routing because of which many higher-layer pins remain unconnected. How to fix it?

vijayank88:
@eder-matheus 
Is required any further response?

eder-matheus:
@vijayank88 @mukul19147 I would only add an observation regarding the `set_routing_layers` command, where you can set specific layer ranges for signal and clock nets. But note that even with the layer range defined, the global router needs to create guides to the pin layers for pin access. So you may see routing out of the layer range only for pin access.

maliberty:
@arthurjolo your current work should resolve this

PR #4157, opened 2023-10-20
Grt pin access layer

arthurjolo:
Solves [#455
](https://github.com/The-OpenROAD-Project/OpenROAD/issues/455)
Currently during fastroute it is never considered the pins access layers, the guides that connect the routing to the pin are only added at the end of the global route. This was causing some additional vias adding and also it wasn't allowing fastroute to fully understand net, preventing it from finding possible better routings considering the pins correct access layer. These changes make fastroute fully aware of the pins access layers and makes the necessary connection still on fastroute, removing the need of a post-processing to add these guides.


src/grt/src/fastroute/src/maze3D.cpp
```diff
@@ -142,12 +142,14 @@ void FastRouteCore::setupHeap3D(int netID,
   dest_heap_3D.clear();
 
   if (num_terminals == 2) {  // 2-pin net
-    d1_3D[0][y1][x1] = 0;
-    directions_3D[0][y1][x1] = Direction::Origin;
-    src_heap_3D.push_back(&d1_3D[0][y1][x1]);
-    d2_3D[0][y2][x2] = 0;
-    directions_3D[0][y2][x2] = Direction::Origin;
-    dest_heap_3D.push_back(&d2_3D[0][y2][x2]);
+    const int nt1 = treenodes[n1].stackAlias;
```
eder-matheus, line 145: Could you comment on why the stackAlias is used here? Also, a more meaningful name than `nt1` would make it clearer.

eder-matheus (changes requested):
@maliberty FYI, both public and private CIs are passing.

src/grt/src/fastroute/src/maze3D.cpp
arthurjolo: The stackAlias is used the get the correct layer in case there is more than one node in the same GCell, so with this it is to get the correct layer and ensure that both pins will be covered. As for the name I just followed the pattern used in maze3D but I can change it.

src/grt/src/GlobalRouter.cpp
```diff
@@ -1941,8 +1861,6 @@ void GlobalRouter::addRemainingGuides(NetRouteMap& routes,
       if (route.empty()) {
         addGuidesForLocalNets(
             db_net, route, min_routing_layer, max_routing_layer);
-      } else {
-        addGuidesForPinAccess(db_net, route);
```
eder-matheus, line 1945: It's great to see this function going away!

src/grt/src/fastroute/src/maze3D.cpp
eder-matheus: The names in maze3D could be better. I think you can change these var names here, and at some point, we should review the remaining of this file to improve its readability.

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -197,6 +197,24 @@ void FastRouteCore::fillVIA()
         int n1a = treeedge->n1a;
         int n2a = treeedge->n2a;
 
+        if (n1a < num_terminals) {
+          if (treenodes[n1a].hID == BIG_INT && edgeID == treenodes[n1a].lID) {
+            const int n1a_access_layer = nets_[netID]->getPinL()[n1a];
+
+            // Concection edge to pin n1 if edge is not on the same layer
```
eder-matheus, line 204: "Concection" typo. Maybe "Connect edge to pin..."

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -242,24 +260,48 @@ void FastRouteCore::fillVIA()
...
-            treeedge->route.gridsX[k] = tmpX[k];
-            treeedge->route.gridsY[k] = tmpY[k];
-            treeedge->route.gridsL[k] = tmpL[k];
+        }
+
+        if (n2a < num_terminals && treenodes[n2a].hID == BIG_INT
+            && edgeID == treenodes[n2a].lID) {
+          const int n2a_access_layer = nets_[netID]->getPinL()[n2a];
+
+          // Concection edge to pin n2 if edge is not on the same layer
```
eder-matheus, line 276: Ditto.

eder-matheus (changes requested):

src/grt/src/GlobalRouter.cpp
```diff
@@ -1756,7 +1756,8 @@ void GlobalRouter::saveGuides()
             odb::dbTechLayer* layer = routing_layers_[layer_idx];
             odb::dbGuide::create(db_net, layer, box);
             if (isCoveringPin(net, segment)) {
-              odb::dbTechLayer* layer = routing_layers_[segment.final_layer];
+              int layer_idx = std::max(segment.init_layer, segment.final_layer);
+              odb::dbTechLayer* layer = routing_layers_[layer_idx];
```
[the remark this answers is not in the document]
arthurjolo: It is not selectin de max layer, it is selcting the hiest layer of the segment(that it is a via)

src/grt/src/GlobalRouter.cpp
```diff
@@ -1784,9 +1785,11 @@ void GlobalRouter::saveGuides()
...
 {
   for (const auto& pin : net->getPins()) {
-    if (pin.getConnectionLayer() == segment.final_layer
-        && pin.getOnGridPosition()
-               == odb::Point(segment.final_x, segment.final_y)
+    int seg_top_layer = std::max(segment.final_layer, segment.init_layer);
+    int seg_x = segment.final_x;
+    int seg_y = segment.final_y;
+    if (pin.getConnectionLayer() == seg_top_layer
+        && pin.getOnGridPosition() == odb::Point(seg_x, seg_y)
```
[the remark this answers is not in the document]
arthurjolo: The lowest layer is always added to the guides, what needs to be checked is if the highest also needs to be added(if it conects a pin).

src/grt/src/GlobalRouter.cpp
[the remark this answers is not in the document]
arthurjolo: In this method the lowest layer is always added, so it is needed to check if the other layer is connected to a pin(meaning that no other segment reaches that point and it is needed to add that layer).

eder-matheus (approved):
@arthurjolo The PR looks good. You just need to fix the typos and the var names. Other than that, this is ready to merge.

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -703,7 +703,10 @@ NetRouteMap FastRouteCore::getRoutes()
         int y2 = tile_size_ * (nodes[treeedge->n2].y + 0.5) + y_corner_;
         int l2 = nodes[treeedge->n2].botL;
         GSegment segment(x1, y1, l1 + 1, x2, y2, l2 + 1);
-        if (net_segs.find(segment) == net_segs.end()) {
+        // It is possible to have nodes that are not in adjacent layers so
+        // this check only adds the segmente if the node ate in adjacent layer
```
eder-matheus, line 707: Typo in "segment" and missing plural.

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -703,7 +703,10 @@ NetRouteMap FastRouteCore::getRoutes()
         int y2 = tile_size_ * (nodes[treeedge->n2].y + 0.5) + y_corner_;
         int l2 = nodes[treeedge->n2].botL;
         GSegment segment(x1, y1, l1 + 1, x2, y2, l2 + 1);
-        if (net_segs.find(segment) == net_segs.end()) {
+        // It is possible to have nodes that are not in adjacent layers so
```
eder-matheus, line 706: Elaborate on why it can have nodes that are not in adjacent layers.

eder-matheus (changes requested):

src/grt/src/fastroute/src/FastRoute.cpp
arthurjolo: If an edge with length 0 is between a pin and a steiner node, since this edge is never analyze the attributes botL will have the default value that is the max routing layer. In these cases the steiner node considers the pin layer on it's stack alias and the other edges that use the stiener node will consider the correct pin layer.

eder-matheus: Could you add this explanation (perhaps more simplified) to the code? I want to make sure we'll have this well-documented in the code for future reference.

eder-matheus (approved):

maliberty:
@arthurjolo note that "Solves #455" has no effect on that issue.  It is better to use a keyword from https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #4174, opened 2023-10-25
fix gitignore and report file generation

eder-matheus:
Minor fixes to avoid creating unwanted report files. Currently, a bunch of files is created when running the regression tests, leaving the branch "dirt" with git status.

### DIFF
--- a/src/cts/.gitignore
+++ b/src/cts/.gitignore
@@ -1,3 +1,2 @@
 test/src
 test/results
-test/cts.clk.buffer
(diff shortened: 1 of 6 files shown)


src/cts/.gitignore
```diff
@@ -1,2 +1,3 @@
 test/src
 test/results
+test/cts.clk.buffer
```
maliberty, line 3: What is this file?

maliberty: I think this comes from 
```
set_debug_level CTS legalizer 3
```
in check_buffers_blockage.tcl.  I think you could lower that to 2 and avoid this file being written.  It isn't used in the test.

maliberty (approved):

PR #4175, opened 2023-10-25
Grt: slack calculation fix

luis201420:
Avoiding calculating slack to nets that are already previously routed


eder-matheus (approved):
Waiting for secure-ci results.

PR #4181, opened 2023-10-26
odb: add support for exceptpgnets on dbObstructions

gadfort:
Adds support for DEF blockages with EXCEPTPGNETS flag

### DIFF
--- a/src/odb/src/db/dbObstruction.cpp
+++ b/src/odb/src/db/dbObstruction.cpp
@@ -61,6 +61,7 @@ _dbObstruction::_dbObstruction(_dbDatabase*)
 {
   _flags._slot_obs = 0;
   _flags._fill_obs = 0;
+  _flags._except_pg_nets = 0;
   _flags._pushed_down = 0;
   _flags._has_min_spacing = 0;
   _flags._has_effective_width = 0;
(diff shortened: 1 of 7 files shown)


src/odb/src/db/dbObstruction.cpp
```diff
@@ -93,6 +94,12 @@ dbIStream& operator>>(dbIStream& stream, _dbObstruction& obs)
   stream >> obs._min_spacing;
   stream >> obs._effective_width;
 
+  _dbDatabase* db = obs.getImpl()->getDatabase();
+  if (!db->isSchema(db_schema_except_pg_nets_obstruction)) {
+    // assume false for older databases
+    obs._flags._except_pg_nets = false;
+  }
```
maliberty, line 101: don't you need an else to actually read the field?

gadfort: the flags size didn't change, so I don't think thats needed (since it's still a 32 bit word it's reading (a few lines before this), but it's not clear if the unused bits are set of not, so this just makes sure for older versions it's set to false. I could be wrong, but I don't see how to implement this with the bit_fields thing.

maliberty:
I'm curious if you are building some flow around this that OR might benefit from.

gadfort:
@maliberty eventually it would be nice to update PDN to support this, but this is mainly to ensure the DEFs we have are parse-able.

maliberty (approved):

PR #4183, opened 2023-10-27
mpl2: display cluster type when printing physical tree

AcKoucher:
I also changed the names of the methods related to bundledIO clusters to improve readability.

### DIFF
--- a/src/mpl2/src/object.cpp
+++ b/src/mpl2/src/object.cpp
@@ -258,6 +258,36 @@ void Cluster::clearHardMacros()
   hard_macros_.clear();
 }
 
+std::string Cluster::getClusterTypeString() const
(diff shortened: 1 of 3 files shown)


src/mpl2/src/object.cpp
```diff
@@ -258,6 +258,41 @@ void Cluster::clearHardMacros()
   hard_macros_.clear();
 }
 
+std::string Cluster::getClusterTypeString() const
+{
+  std::string cluster_type = "";
```
maliberty, line 263: no need for ```= ""``` as the default ctor does that.

maliberty (approved):

maliberty (approved):

PR #4184, opened 2023-10-27
include netlist updating in mbff clustering code

therealshreyas:
…in clusterable chunks; rewrite netlist


src/gpl/src/mbff.cpp
```diff
@@ -43,48 +41,206 @@
...
+    }
+    if (port->direction()->isOutput()) {
+      q_pins.push_back(port->name());
+    }
   }
-  if (slot_cnt == 8) {
-    return 2;
+
+  std::map<std::string, std::string> ret;
+  for (int i = 0; i < static_cast<int>(d_pins.size()); i++) {
```
QuantamHD, line 139: Probably cleaner to just make i size_t

src/gpl/src/mbff.cpp
```diff
@@ -957,116 +1160,374 @@ void MBFF::KMeansDecomp(const std::vector<Flop>& flops,
...
+    pin_mappings_[i].resize(num_sizes_);
+
+    slot_to_tray_x_[i].resize(num_sizes_);
+    slot_to_tray_y_[i].resize(num_sizes_);
   }
 
+  for (auto lib : db_->getLibs()) {
+    for (auto master : lib->getMasters()) {
+      odb::dbInst* tmp_tray = odb::dbInst::create(block_, master, "test_tray");
+      if (!IsValidTray(tmp_tray)) {
```
maliberty, line 1391: Is you pass every type of cell to IsValidTray you can expect IsDPin not to work on most (eg a nand2 will think it has 2 D pins).

src/gpl/src/mbff.cpp
```diff
@@ -957,116 +1160,374 @@ void MBFF::KMeansDecomp(const std::vector<Flop>& flops,
...
+        odb::dbInst::destroy(tmp_tray);
+        continue;
+      }
+
+      int num_slots = GetNumD(tmp_tray);
+      int idx = GetBitIdx(num_slots);
+      int setting = (GetNumQ(tmp_tray) - num_slots == 0 ? 0 : 1);
```
maliberty, line 1398: What is 'setting' ?  Its not a descriptive name

src/gpl/src/mbff.cpp
```diff
@@ -43,58 +43,277 @@
...
+    }
+    if (port->direction()->isOutput()) {
+      if (q_pins.size()) {
+        if (sta::FuncExpr::equiv(q_pins.back()->function(), port->function())) {
+          q_pins.push_back(port);
+        } else {
+          qn_pins.push_back(port);
+        }
+      } else {
+        q_pins.push_back(port);
```
maliberty, line 175: Why assuming the first pin is Q and not QN?

therealshreyas: It ends up not mattering since we don't actually care about whether it is Q or QN.
As long as we group pins by different functions, the logic works.

src/gpl/src/mbff.cpp
therealshreyas: setting = 0 if #D = #Q pins
setting = 1 if #Q - #D pins > 0 (2 Q pins per 1 D pin case)

src/gpl/src/mbff.cpp
therealshreyas: Does the following pseudocode seem fine? 

IsValidTray(inst):
 slots = 0
 for (seq : sequentials):
   slots += (seq->PortIterator->hasNext())
 return slots > 1
 

Since we scan sequentials (and are essentially checking that it's not an FF), I believe this should work

src/gpl/src/mbff.cpp
maliberty: How about has_dual_outputs ?  Setting doesn't mean much.

src/gpl/src/mbff.cpp
maliberty: you can't group a d/clk/q with a d/clk/qn

therealshreyas: This won't happen because my program checks that q and qn have the same function.
In your given given, q and qn have different functions

src/gpl/src/mbff.cpp
maliberty: ok

src/gpl/src/mbff.cpp
therealshreyas: Maybe it makes more sense to think of the vectors q and qn has a collection of ports that have the same function.
If a "Q" port gets placed into the vector qn, then the "QN" port has to be placed into q since we check for the function equivalence.

When we cluster flops and modify the netlist, if I want to disconnect the FF's "QN" port and connect the tray's "QN" port, I check the vector q.
Since the port's in q have the same function as our "QN" port, we make a valid connection.

src/gpl/src/mbff.cpp
```diff
@@ -43,58 +43,385 @@
...
+{
+  sta::Cell* cell = network_->dbToSta(inst->getMaster());
+  if (cell == nullptr) {
+    return false;
+  }
+  sta::LibertyCell* lib_cell = network_->libertyCell(cell);
+  if (lib_cell == nullptr) {
+    return false;
+  }
+  return lib_cell->isInverter();
```
maliberty, line 62: This checks if the cell is an inverter which no FF is.  It doesn't not tell you if the edge is inverting.

src/gpl/src/mbff.cpp
```diff
@@ -43,58 +43,385 @@
...
+  sta::Cell* cell = network_->dbToSta(inst->getMaster());
+  if (cell == nullptr) {
+    return false;
+  }
+  sta::LibertyCell* lib_cell = network_->libertyCell(cell);
+  if (lib_cell == nullptr) {
+    return false;
+  }
+  for (auto seq : lib_cell->sequentials()) {
+    if (seq->clear()) {
```
maliberty, line 76: should be seq->preset()

src/gpl/src/mbff.cpp
```diff
@@ -43,58 +43,385 @@
...
+  sta::Cell* cell = network_->dbToSta(inst->getMaster());
+  if (cell == nullptr) {
+    return false;
+  }
+  sta::LibertyCell* lib_cell = network_->libertyCell(cell);
+  if (lib_cell == nullptr) {
+    return false;
+  }
+  for (auto seq : lib_cell->sequentials()) {
+    if (seq->preset()) {
```
maliberty, line 94: should be seq->clear()

src/gpl/src/mbff.cpp
```diff
@@ -43,58 +43,385 @@
...
+bool MBFF::IsSupplyPin(odb::dbITerm* iterm)
+{
+  return iterm->getSigType().isSupply();
+}
+
+bool MBFF::IsDPin(odb::dbITerm* iterm)
+{
+  const bool exclude = (IsClockPin(iterm) || IsSupplyPin(iterm));
+  const bool yes = (iterm->getIoType() == odb::dbIoType::INPUT);
+  return (yes & !exclude);
```
maliberty, line 140: This wil return true for set/reset pins too

maliberty:
My feeling is that there will be subsequent work on this PR but it is ready to be merged and more testing will follow.

PR #4187, opened 2023-10-30
Turn gcNet into drNet

refaay:


refaay:
Closing this PR because of importing other users' commits here.
PR 4260 created instead.

PR #4188, opened 2023-10-30
dpl: Avoiding moving cells that are outside of macros and inside the die

luis201420:
Fixing bug when avoiding moving legal cells.

### DIFF
--- a/src/dpl/src/Place.cpp
+++ b/src/dpl/src/Place.cpp
@@ -1280,10 +1280,11 @@ void Opendp::legalCellPos(dbInst* db_inst)
 {
   Cell cell;
   convertDbToCell(db_inst, cell);
-  Point legal_pt = pointOffMacro(cell);      // return real position
-  Point new_pos = legalPt(&cell, legal_pt);  // return real position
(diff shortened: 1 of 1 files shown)


src/dpl/src/Place.cpp
```diff
@@ -1280,10 +1280,15 @@ void Opendp::legalCellPos(dbInst* db_inst)
...
   convertDbToCell(db_inst, cell);
-  Point legal_pt = pointOffMacro(cell);      // return real position
-  Point new_pos = legalPt(&cell, legal_pt);  // return real position
-
-  if (legal_pt == new_pos) {
+  Point init_pos = initialLocation(
+      &cell, false);  // returns the initial position of the cell
+  Point legal_pt = pointOffMacro(
+      cell);  // returns the modified position if the cell is in a macro
+  Point new_pos = legalPt(
```
eder-matheus, line 1287: How do you ensure that `legalPt` will not move the cell inside a macro again?

luis201420: The `legalPt` function will not ensure that, that case would happen when there is no minimum distance to place a cell between a macro and a side of the die.

eder-matheus (approved):

Issue #4148, opened 2023-10-18
global route crash


oharboe:
### Describe the bug

This is from MegaBoom:

```
./orfs make DESIGN_NAME=RegisterFileSynthesizable_1 DESIGN_CONFIG=block.mk VERILOG_FILES="rtl/RegisterFileSynthesizable_1.sv\ rtl/regfile_128x64.sv" CORE_UTILIZATION=20 PLACE_DENSITY=0.40
```

```
[deleted]
[INFO GRT-0103] Extra Run for hard benchmark.
/usr/include/c++/12/bits/stl_vector.h:1123: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = int; _Alloc = std::allocator<int>; reference = int&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Signal 6 received
Stack trace:
 0# 0x00005563040E3323 in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 1# 0x00007F3F3E03C460 in /lib/x86_64-linux-gnu/libc.so.6
 2# pthread_kill in /lib/x86_64-linux-gnu/libc.so.6
 3# gsignal in /lib/x86_64-linux-gnu/libc.so.6
 4# abort in /lib/x86_64-linux-gnu/libc.so.6
 5# 0x00007F3F3E4E21CF in /lib/x86_64-linux-gnu/libstdc++.so.6
 6# grt::FastRouteCore::copyGrids(grt::TreeNode const*, int, int, grt::TreeEdge const*, int, std::vector<int, std::allocator<int> >&, std::vector<int, std::allocator<int> >&) in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 7# grt::FastRouteCore::updateRouteType1(int, grt::TreeNode const*, int, int, int, int, int, grt::TreeEdge*, int, int) in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 8# grt::FastRouteCore::mazeRouteMSMD(int, int, float, int, int, bool, int, float, int, int, int, float&) in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
 9# grt::FastRouteCore::run() in /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad
[...]

PR #4200, opened 2023-11-02
[DO NOT MERGE] deltaDebug: interleave elimination of nets and instances

oharboe:
this will hopefully avoid pahtologically slow cases, such as when trying to whittle down a testcase for global routing where speeding up the testing will come from the nets, which previously would be whittled down only after all instances were whittled down.

### DIFF
--- a/etc/deltaDebug.py
+++ b/etc/deltaDebug.py
@@ -111,17 +111,19 @@ class deltaDebugger:
             while self.n <= (2 ** self.persistence):
                 error_in_range = None
                 for j in range(self.n):
-                    current_err = self.perform_step(cut_index = j)
-                    if(current_err == "NOCUT"):
-                        break
-                    elif(current_err != None):
-                        # Found the target error with the cut DB
-                        #
-                        # This is a suitable level of detail to look for more errors,
-                        # complete this level of detail.
-                        err = current_err
-                        error_in_range = current_err
-                        self.prepare_new_step()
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty Thoughts?

I'm trying it out on #4148, which didn't complete, even after 2 weeks.

maliberty:
It looks like you toggle between nets & insts now.  That would help though I guess for grt test cases it might just be better to have an option to say do nets first and then insts.

oharboe:
> It looks like you toggle between nets & insts now. That would help though I guess for grt test cases it might just be better to have an option to say do nets first and then insts.

Then again, every option makes deltaDebug harder to learn, reducing potential beneficiaries and value of the feature.

maliberty:
Making alternating the default makes sense but I would see value in having control.

oharboe:
> Making alternating the default makes sense but I would see value in having control.

Makes sense. Though there is something to be said for fewer options and just tinkering with the script as needed.

I am doing a run now to see if there is any benefit in my pathological case to alternating.

maliberty:
Tinkering with the script takes even more expertise than using a switch.

oharboe:
> Tinkering with the script takes even more expertise than using a switch.

True. If this test works out, i.e. converges faster, then I will reread the above and propose a policy/command line switches.

oharboe:
The case I have is just very tough to whittle down. Incredibly long turnaround times, a failure only after 2 hours, not much to remove.

PR #4203, opened 2023-11-02
mpl2: draw macro blockages in graphics

AcKoucher:
Draw macro blockages during SA for SoftMacros.


maliberty (approved):
It looks fine but please clean up the clang-tidy items.

PR #4222, opened 2023-11-07
grt: create gcell grid based on input guides

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/4192


src/drt/test/ispd18_sample.defok
```diff
@@ -27,8 +27,8 @@ TRACKS X 83800 DO 52 STEP 400 LAYER Metal2 ;
 TRACKS Y 72010 DO 51 STEP 380 LAYER Metal2 ;
 TRACKS X 83800 DO 52 STEP 400 LAYER Metal1 ;
 TRACKS Y 72010 DO 51 STEP 380 LAYER Metal1 ;
-GCELLGRID X 83600 DO 18 STEP 5700 ;
-GCELLGRID Y 71820 DO 16 STEP 5700 ;
+GCELLGRID X 83600 DO 3 STEP 6000 ;
+GCELLGRID Y 71820 DO 3 STEP 5700 ;
```
maliberty, line 31: 18->3 is a surprisingly large change - is this correct?

maliberty:
Have you run the ispd benchmarks?

src/drt/test/ispd18_sample.defok
eder-matheus: Yes, it is correct because previously, GRT was not considering the origin of the design to calculate the number of GCells. Since this design has a non-zero origin, the resulting value is much larger than it should be.

eder-matheus:
> Have you run the ispd benchmarks?

I've started a run for the ispd benchmarks now.

maliberty:
I recall there is something similar to the voting code in drt when it reads guides as well.  Might it be shareable?

eder-matheus:
> I recall there is something similar to the voting code in drt when it reads guides as well. Might it be shareable?

We have a similar voting code in drt, but each tool's data structures are very different. Adapting drt code to use grt structs would be a considerable overhead.

maliberty (approved):

maliberty:
@eder-matheus you can merge once ispd passes

PR #4243, opened 2023-11-09
gpl: check value of clock_on in flop clustering

therealshreyas:


src/gpl/src/mbff.cpp
```diff
@@ -332,6 +359,9 @@ MBFF::DataToOutputsMap MBFF::GetPinMapping(odb::dbInst* tray)
   // all output pins are Q pins
   while (port_itr.hasNext()) {
     sta::LibertyPort* port = port_itr.next();
+    if (port->isBus() || port->isBundle()) {
+      log_->info(utl::GPL, 11000, "Found bus/bundle port");
```
rovinski, line 363: Messages should be max of 4 digits

maliberty (approved):

PR #4244, opened 2023-11-10
odb: Fixes more ubsan violations in odb

QuantamHD:
std::memcpy instead of unsafe casts.

### DIFF
--- a/src/odb/src/codeGenerator/gen.py
+++ b/src/odb/src/codeGenerator/gen.py
@@ -265,6 +265,11 @@ for klass in schema["classes"]:
             field["setterArgumentType"] = field["getterReturnType"] = field["type"]
 
     klass["fields"] = [field for field in klass["fields"] if "bits" not in field]
+
+    klass["hasBitFields"] = False
+    if flag_num_bits > 0:
+        klass["hasBitFields"] = True
+
     total_num_bits = flag_num_bits
     if flag_num_bits > 0 and flag_num_bits % 32 != 0:
         spare_bits_field = {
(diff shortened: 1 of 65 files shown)


QuantamHD:
@maliberty looks like there's something wrong with the ci runner

maliberty:
> @maliberty looks like there's something wrong with the ci runner

"Timeout has been exceeded"

maliberty:
What makes the casts unsafe?

maliberty:
I restarted the failed step to see if it times out again

QuantamHD:
> What makes the casts unsafe?


```
  uint32_t* flags__bit_field = (uint32_t*) &obj.flags_;
  stream >> *flags__bit_field;
```

Is a violation of strict aliasing. https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8

QuantamHD:
I guess union type punning is also UB in C++, but not C. So I changed to std::memcpy which I guess is the only legal way to do this in C++

QuantamHD:
Added some static asserts to ensure things are the right size

QuantamHD:
I think the clang tidy is confused.

maliberty (approved):

maliberty:
Note that this PR is only addressing the problem in code the use the code generator.  Older code is still 'wrong' eg 
```
dbOStream& operator<<(dbOStream& stream, const _dbBTerm& bterm)
{
  uint* bit_field = (uint*) &bterm._flags;
  stream << *bit_field;
````

PR #4245, opened 2023-11-10
rtlmp: change 'total area' with Core area in debug output

oharboe:
It is not clear from the printed information what 'total area' means.

Why should total area not include halos of macros?

The individual area information is now printed and the user can add those together to get the numbers they need.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -409,6 +409,7 @@ void HierRTLMP::hierRTLMacroPlacer()
       "\tNumber of macros: {}\n"
       "\tArea of macros: {:.2f}\n"
       "\tArea of macros with halos: {:.2f}\n"
+      "\tArea of std cell instances + Area of macros: {:.2f}\n"
       "\tCore area: {:.2f}\n"
       "\tDesign Utilization: {:.2f}\n"
       "\tCore Utilization: {:.2f}\n"
@@ -418,6 +419,7 @@ void HierRTLMP::hierRTLMacroPlacer()
       metrics_->getNumMacro(),
       metrics_->getMacroArea(),
       macro_with_halo_area,
+      metrics_->getStdCellArea() + metrics_->getMacroArea(),
       core_area,
       util,
       core_util,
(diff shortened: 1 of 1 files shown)


src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -404,21 +404,21 @@ void HierRTLMP::hierRTLMacroPlacer()
...
+      "\tDesign Utilization: {:.2f}\n"
       "\tCore Utilization: {:.2f}\n"
       "\tManufacturing Grid: {}\n",
       metrics_->getNumStdCell(),
       metrics_->getStdCellArea(),
       metrics_->getNumMacro(),
       metrics_->getMacroArea(),
       macro_with_halo_area,
-      metrics_->getStdCellArea() + metrics_->getMacroArea(),
+      core_area,
```
maliberty, line 421: Core area is a completely different thing.  Would total instance area be clearer?

oharboe: perhaps spell out the addition?

maliberty: What wording would make sense to you

oharboe: `Area of std cell instances + Area of macros:`

oharboe: Though, given the error message below, I would want to list core_area as well, it is relevant:

```C++
if (macro_with_halo_area + metrics_->getStdCellArea() > core_area) {
error
```

maliberty: That should be a separate entry then.

maliberty: Would 'Area of instances' suffice?

oharboe: In this case, I have to guess that a macro is considered an instance. 

Why not:

`Area of std cell instances + Area of macros:`

?

Much less than 80 chars...

maliberty: How could a macro not be an instance?  I don't care that strongly.

oharboe: I don't know how it could not.

You asked me what label would make the report easier to understand. I think `Area of std cell instances + Area of macros` is self documenting, whereas `Area of instance` leaves me with questions and I wouldn't have guessed that it is an addition of two other pieces of information listed.

I'm thinking that this report is going to be read on those(such as me) that are being trained in the art of backend tools, in which case a bit of spelling out is helpful.

maliberty (approved):

Issue #3803, opened 2023-08-06
segfault in `check_antenna`


gadfort:
### Describe the bug

When running `check_antenna` OpenROAD segfaults.

```
 0# 0x000055C036B030AB in openroad
 1# 0x00007F98FDBB2520 in /lib/x86_64-linux-gnu/libc.so.6
 2# odb::dbWireEncoder::addITerm(odb::dbITerm*) in openroad
 3# odb::tmg_conn::addToWire(int, int, int, bool, bool) in openroad
 4# odb::tmg_conn::treeReorder(bool) in openroad
 5# odb::tmg_conn::analyzeNet(odb::dbNet*) in openroad
 6# grt::RepairAntennas::makeNetWire(odb::dbNet*, std::vector<grt::GSegment, std::allocator<grt::GSegment> >&, std::map<int, odb::dbTechVia*, std::less<int>, std::allocator<std::pair<int const, odb::dbTechVia*> > >&) in openroad
 7# grt::RepairAntennas::makeNetWires(std::map<odb::dbNet*, std::vector<grt::GSegment, std::allocator<grt::GSegment> >, grt::cmpById, std::allocator<std::pair<odb::dbNet* const, std::vector<grt::GSegment, std::allocator<grt::GSegment> > > > >&, int) in openroad
 8# ant::AntennaChecker::checkAntennas(odb::dbNet*, bool) in openroad
 9# ant::check_antennas(char const*, bool) in openroad
10# 0x000055C0376BABD4 in openroad
```

### Expected Behavior

No segfault

### Environment

```shell
kernel: Linux 5.15.0-1040-azure
os: Ubuntu 22.04.2 LTS (Jammy Jellyfish)
cmake version 3.26.4
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-9702-g92f8e07b0
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]

PR #4248, opened 2023-11-10
grt: fix wire creation during antenna check

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3803


eder-matheus:
@maliberty I still have a crash that only happens sometimes in ant/src/AntennaChecker.cc, which I believe is related to how we create the connections to the pins.

Previously, we were using the virtual paths, leading to another crash. Using a regular path is the best choice, but I'm likely doing something wrong when creating these pin connections. I'll investigate more on what should be done to fix it, but any insights would be appreciated.

eder-matheus:
@maliberty I think this can be reviewed now. It will not fully solve the related issue, but at least if fixes a couple of bugs that I found in the code. I'll start a secure-ci for it as well, and will investigate the remaining issues later.

maliberty (approved):
you can merge once the CI is passing

maliberty:
@eder-matheus looks ok to me

eder-matheus:
@maliberty FYI, CI is passing now. I have the artifacts for the previous failure in gf180/jpeg that I will continue investigating, but this PR is safe to merge.

PR #4251, opened 2023-11-11
odb: Migrate DB methods to generic iostreams

QuantamHD:
In preparation for adding compression to the odb format.

### DIFF
--- a/src/OpenRoad.cc
+++ b/src/OpenRoad.cc
@@ -484,11 +484,14 @@ void OpenRoad::readDb(const char* filename)
 
 void OpenRoad::writeDb(const char* filename)
 {
-  FILE* stream = fopen(filename, "w");
-  if (stream) {
-    db_->write(stream);
-    fclose(stream);
-  }
+  std::ofstream stream;
+  stream.exceptions(std::ifstream::failbit | std::ifstream::badbit
+                     | std::ios::eofbit);
+  stream.open(filename, std::ios::binary);
+  
+  db_->write(stream);
+  stream.close();
+  
 }
 
 void OpenRoad::diffDbs(const char* filename1,
--- a/src/OpenRoad.cc
+++ b/src/OpenRoad.cc
@@ -35,6 +35,7 @@
 
 #include "ord/OpenRoad.hh"
 
+#include <fstream>
 #include <iostream>
 #include <thread>
 #ifdef ENABLE_PYTHON3
@@ -490,7 +491,6 @@ void OpenRoad::writeDb(const char* filename)
   stream.open(filename, std::ios::binary);
 
   db_->write(stream);
-  stream.close();
 }
 
 void OpenRoad::diffDbs(const char* filename1,
--- a/src/OpenRoad.cc
+++ b/src/OpenRoad.cc
@@ -486,12 +486,11 @@ void OpenRoad::writeDb(const char* filename)
 {
   std::ofstream stream;
   stream.exceptions(std::ifstream::failbit | std::ifstream::badbit
-                     | std::ios::eofbit);
+                    | std::ios::eofbit);
   stream.open(filename, std::ios::binary);
-  
+
   db_->write(stream);
   stream.close();
-  
 }
 
 void OpenRoad::diffDbs(const char* filename1,
--- a/src/odb/include/odb/dbStream.h
+++ b/src/odb/include/odb/dbStream.h
@@ -56,6 +56,15 @@ class dbOStream
   double _lef_area_factor;
   double _lef_dist_factor;
 
+  // By default values are written as their string ("255" vs 0xFF)
+  // representations when using the << stream method. In dbOstream we are
+  // primarly writing the byte representation which the below accomplishes.
+  template <typename T>
+  void writeValueAsBytes(T type)
+  {
+    _f.write(reinterpret_cast<char*>(&type), sizeof(T));
+  }
+
  public:
   dbOStream(_dbDatabase* db, std::ostream& f);
 
@@ -69,67 +78,67 @@ class dbOStream
 
   dbOStream& operator<<(char c)
   {
-    _f << c;
+    _f.put(c);
     return *this;
   }
 
   dbOStream& operator<<(unsigned char c)
   {
-    _f << c;
+    _f.put(c);
     return *this;
   }
 
   dbOStream& operator<<(int16_t c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(uint16_t c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(int c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(uint64_t c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(unsigned int c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(int8_t c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(float c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(double c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
   dbOStream& operator<<(long double c)
   {
-    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
+    writeValueAsBytes(c);
     return *this;
   }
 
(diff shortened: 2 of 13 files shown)


src/odb/include/odb/dbStream.h
```diff
@@ -74,100 +70,85 @@ class dbOStream
...
 
   dbOStream& operator<<(int16_t c)
   {
-    int n = fwrite(&c, sizeof(c), 1, _f);
-    if (n != 1) {
-      write_error();
-    }
+    char dst[sizeof(c)];
+    std::memcpy(dst, &c, sizeof(c));
+    _f.write(dst, sizeof(c));
```
maliberty, line 87: why not just ```_f << c``` ?

src/odb/src/db/dbBlock.cpp
```diff
@@ -3351,24 +3351,14 @@ void dbBlock::writeDb(char* filename, int allNode)
       dbname = fmt::format("{}.remote.{}.db", filename, getpid());
   } else
     dbname = fmt::format("{}.db", filename);
-  FILE* file = fopen(dbname.c_str(), "wb");
-  if (!file) {
+  std::ofstream file(dbname, std::ios::binary);
+  if (!file.is_open()) {
```
maliberty, line 3355: more general error check
```if (!file)```

File: src/odb/src/db/dbBlock.cpp [hunk 3]
```diff
@@ -3351,24 +3351,14 @@ void dbBlock::writeDb(char* filename, int allNode)
...
-    getImpl()->getLogger()->warn(
-        utl::ODB, 20, "Memory allocation failed for io buffer");
-    fclose(file);
-    return;
-  }
-  setvbuf(file, buffer, _IOFBF, io_bufsize);
   getDataBase()->write(file);
-  free((void*) buffer);
-  fclose(file);
+  file.close();
```
maliberty, line 3361: not needed

src/odb/src/db/dbDatabase.cpp
```diff
@@ -701,8 +701,8 @@ void dbDatabase::writeEco(dbBlock* block_, const char* filename)
 {
   _dbBlock* block = (_dbBlock*) block_;
 
-  FILE* file = fopen(filename, "w");
-  if (!file) {
+  std::ofstream file(filename, std::ios::binary);
+  if (!file.is_open()) {
```
maliberty, line 705: same

src/odb/src/db/dbDatabase.cpp
```diff
@@ -714,7 +714,7 @@ void dbDatabase::writeEco(dbBlock* block_, const char* filename)
     stream << *block->_journal_pending;
   }
 
-  fclose(file);
+  file.close();
```
maliberty, line 717: same

src/odb/src/swig/common/swig_common.cpp
```diff
@@ -182,15 +182,15 @@ odb::dbDatabase* read_db(odb::dbDatabase* db, const char* db_path)
 
 int write_db(odb::dbDatabase* db, const char* db_path)
 {
-  FILE* fp = fopen(db_path, "wb");
-  if (!fp) {
+  std::ofstream fp(db_path, std::ios::binary);
+  if (!fp.is_open()) {
```
maliberty, line 186: same

src/odb/src/swig/common/swig_common.cpp
```diff
@@ -182,15 +182,15 @@ odb::dbDatabase* read_db(odb::dbDatabase* db, const char* db_path)
...
     int errnum = errno;
     fprintf(stderr, "Error opening file: %s\n", strerror(errnum));
     fprintf(stderr, "Errno: %d\n", errno);
     return errno;
   }
   db->write(fp);
-  fclose(fp);
+  fp.close();
```
maliberty, line 193: same

maliberty (changes requested):

src/odb/src/db/dbBlock.cpp
```diff
@@ -3351,24 +3351,14 @@ void dbBlock::writeDb(char* filename, int allNode)
       dbname = fmt::format("{}.remote.{}.db", filename, getpid());
   } else
     dbname = fmt::format("{}.db", filename);
-  FILE* file = fopen(dbname.c_str(), "wb");
-  if (!file) {
+  std::ofstream file(dbname, std::ios::binary);
```
oharboe, line 3354: add missing exception handling to this stream, as the rest of the code assumes it.

src/odb/include/odb/dbStream.h
```diff
@@ -178,21 +159,17 @@ class dbOStream
...
 
   dbOStream& operator<<(dbObjectType c)
   {
-    int n = fwrite(&c, sizeof(c), 1, _f);
-    if (n != 1) {
-      write_error();
-    }
+    char dst[sizeof(c)];
+    std::memcpy(dst, &c, sizeof(c));
+    _f.write(dst, sizeof(c));
```
oharboe, line 172: This is using exceptions, which is good, but exceptions for this file handle needs to be enabled below.

I'm a bit mystified by the std::memcpy(). Perhaps break out a function with generics(to take any type) and document in that function why std::memcpy() is used?

src/odb/test/cpp/TestAccessPoint.cpp
```diff
@@ -28,10 +28,9 @@ BOOST_AUTO_TEST_CASE(test_default)
   ap->setHighType(dbAccessType::HalfGrid);
   ap->setAccess(true, dbDirection::DOWN);
   iterm->setAccessPoint(pin, ap);
-  FILE* write;
   std::string path
       = std::string(std::getenv("BASE_DIR")) + "/results/TestAccessPointDbRW";
-  write = fopen(path.c_str(), "w");
+  std::ofstream write(path, std::ios::binary);
```
oharboe, line 33: add exceptions for all of these std::ofstream() constructors

oharboe (changes requested):
nits

src/odb/include/odb/dbStream.h
```diff
@@ -74,100 +69,67 @@ class dbOStream
...
     return *this;
   }
 
   dbOStream& operator<<(int16_t c)
   {
-    int n = fwrite(&c, sizeof(c), 1, _f);
-    if (n != 1) {
-      write_error();
-    }
+    _f.write(reinterpret_cast<char*>(&c), sizeof(c));
```
maliberty, line 84: why doesn't _f << c work here?

QuantamHD: I think if the file is open in text mode it will potentially write out string(c) instead of bytes(c). Could be mistaken on the behavior of 16 bit values, but I think 32 and 64 bit values have that behavior.

maliberty: Why not use binary mode?  This isn't really a text file.

QuantamHD: The dbOstream constructor takes a generic stream std::ostream which we don't know anything about. If we're passed a ofstream in text mode we wouldn't know it. My C++ isn't that great though so my understanding of the situation is if that happens and we _f << uint16_t it will literally write out the string representation of that number whereas if we use the write above no matter the stream we're passed it will always write out the byte representation of the number. 

To your understand of the stdlib is that correct or am I wrong.

src/odb/src/db/dbDatabase.cpp
```diff
@@ -701,8 +702,8 @@ void dbDatabase::writeEco(dbBlock* block_, const char* filename)
 {
   _dbBlock* block = (_dbBlock*) block_;
 
-  FILE* file = fopen(filename, "w");
-  if (!file) {
+  std::ofstream file(filename, std::ios::binary);
```
oharboe, line 705: use exceptions, otherwise subsequent io operations won't be checked(such as running out of disk space).

src/odb/test/cpp/TestLef58Properties.cpp
```diff
@@ -54,10 +54,9 @@ BOOST_AUTO_TEST_CASE(test_default)
 
   lefParser.createTechAndLib("tech", libname, path.c_str());
 
-  FILE* write;
   path = std::string(std::getenv("BASE_DIR"))
          + "/results/TestLef58PropertiesDbRW";
-  write = fopen(path.c_str(), "w");
+  std::ofstream write(path, std::ios::binary);
```
oharboe, line 59: add exceptions, reduces questions upon search the code for lack of io exceptions...

oharboe (changes requested):
nits

src/odb/include/odb/dbStream.h
maliberty: On further research it appears your code is correct.

oharboe: Can we break out a function with generics here where we can document what is going on. Seems like it is worth doing so, there is something new and subtle to learn here for future readers of the code...

src/odb/include/odb/dbStream.h
```diff
@@ -74,100 +78,67 @@ class dbOStream
 
   dbOStream& operator<<(char c)
   {
-    int n = fwrite(&c, sizeof(c), 1, _f);
-    if (n != 1) {
-      write_error();
-    }
+    _f.put(c);
```
maliberty, line 81: Use writeValueAsBytes to match operator<<

src/odb/include/odb/dbStream.h
```diff
@@ -74,100 +78,67 @@ class dbOStream
...
     return *this;
   }
 
   dbOStream& operator<<(unsigned char c)
   {
-    int n = fwrite(&c, sizeof(c), 1, _f);
-    if (n != 1) {
-      write_error();
-    }
+    _f.put(c);
```
maliberty, line 87: same

maliberty (changes requested):

maliberty (approved):

src/OpenRoad.cc
```diff
@@ -484,11 +485,12 @@ void OpenRoad::readDb(const char* filename)
...
-  if (stream) {
-    db_->write(stream);
-    fclose(stream);
-  }
+  std::ofstream stream;
+  stream.exceptions(std::ifstream::failbit | std::ifstream::badbit
+                    | std::ios::eofbit);
+  stream.open(filename, std::ios::binary);
+
+  db_->write(stream);
```
oharboe, line 493: @maliberty @QuantamHD If you are adding compression, then you can easily get an exception within a desreuctor, which is nasty.
call close.

PR #4254, opened 2023-11-12
Fix DRC standalone command

osamahammad21:
- Fix io reading from wire encoder
- Fix initialization of gcWorker nets from design directly (no drWorker)
- add drc_test


src/drt/src/io/io.cpp
```diff
@@ -750,8 +750,6 @@ void io::Parser::setNets(odb::dbBlock* block)
               nextX = beginX;
               nextY = beginY;
             }
-            prevLayer = decoder.getLayer();
-            layerName = prevLayer->getName();
```
osamahammad21, line 754: @eder-matheus this part was added in your [PR](https://github.com/The-OpenROAD-Project/OpenROAD/pull/3411) previously which messed up the layer assignment of the read segments from wire encoder. I reran the test case that your PR should solve and it ran through without any problem.
Could you please confirm that this does not conflict with your initial fix?

maliberty:
unit test failure

osamahammad21:
> unit test failure

fixed

src/drt/src/io/io.cpp
eder-matheus: From what I reviewed, it doesn't conflict with my changes.

maliberty (approved):

PR #4260, opened 2023-11-13
Turn frNet into drNet inside gcNet

refaay:


osamahammad21:
@refaay  In general, you need to account that there are 2 types of frNets; signal and special.
The signal nets are the nets that are supposed to be routed and handled by the detailed router. They're gonna be initialized by the drWorker as drNets and passed through gc.
The special nets are supply nets (POWER/GROUND). They are routed before DRT by PDN. They are not supposed to be touched by DRT. We should only consider them while we are routing other nets and avoid violating them. Thus, they **do not** get initialized at the drWorker as drNets. So the owner of their gcNets should be the frNet.

maliberty:
@osamahammad21 do you want to keep this and complete it or close?

osamahammad21:
@maliberty I am converting it to draft till I find time to work on it.

maliberty:
I'm closing old PRs to minimize Jenkins load (pr-merge still runs).  Reopen when you are ready to proceed.

Issue #4253, opened 2023-11-12
don't write congestion map outside of the routing layers


maliberty:
### Description

I noticed in asap7 that although the min routing layer is M2 we still write out useless congestion map data for M1 into odb.  All the entries are 0.

### Suggested Solution

Skip writing outside of min/max layers

maliberty:
@eder-matheus why was this re-opened?

eder-matheus:
I had to revert my change due to some failures in the CI. I haven't had the chance to get back to it yet.

PR #4262, opened 2023-11-14
grt: skip layers out of min/max layer range

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/4253

### DIFF
--- a/src/grt/src/fastroute/src/FastRoute.cpp
+++ b/src/grt/src/fastroute/src/FastRoute.cpp
@@ -806,7 +806,8 @@ NetRouteMap FastRouteCore::getPlanarRoutes()
   return routes;
 }
 
-void FastRouteCore::updateDbCongestion()
+void FastRouteCore::updateDbCongestion(int min_routing_layer,
+                                       int max_routing_layer)
 {
   auto block = db_->getChip()->getBlock();
   auto db_gcell = block->getGCellGrid();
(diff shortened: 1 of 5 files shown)


src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -818,8 +819,8 @@ void FastRouteCore::updateDbCongestion()
   db_gcell->addGridPatternX(x_corner_, x_grid_, tile_size_);
   db_gcell->addGridPatternY(y_corner_, y_grid_, tile_size_);
   auto db_tech = db_->getTech();
-  for (int k = 0; k < num_layers_; k++) {
-    auto layer = db_tech->findRoutingLayer(k + 1);
+  for (int k = min_routing_layer - 1; k < max_routing_layer - 1; k++) {
```
maliberty, line 822: why min_routing_layer -1 ?

eder-matheus: In the GlobalRoute class, the min layer starts in 1. Inside FastRouteCore, the min layer is zero. So, we need the `-1` here to correctly access the vectors with k.

maliberty (approved):

maliberty:
Please look at the CI failures

src/grt/src/fastroute/src/FastRoute.cpp
eder-matheus: It should be fine with this. The changes in #4263 don't change the "global" layer range. It just makes the specific net layer ranges between FastRouteCore and GlobalRoute classes consistent.

maliberty: We should put it on the list to resolve this difference in indexing as it seems to be a bug source.

eder-matheus: I can create a GH issue to keep track of it.

eder-matheus: @maliberty FYI: https://github.com/The-OpenROAD-Project/OpenROAD/issues/4266

Issue #4196, opened 2023-11-01
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_1_bits_rob_idx[3].


oharboe:
### Describe the bug

unzip https://drive.google.com/file/d/1t_yi8pM2GSOqkkBR2IxDmVzehg5vH1_4/view?usp=sharing

NOTE! The example file is 7gByte, even after deltaDebug.py, but it runs in ca. 2 minutes.

run:

```
./run-me-BoomCore-asap7-base.sh
```

```
[deleted]
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_1_bits_rob_idx[3].
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_0_bits_rob_idx[3].
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_2_bits_rob_idx[3].
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_3_bits_rob_idx[3].
[WARNING GRT-0350] Missing route to pin rob/io_enq_uops_3_rob_idx[3].
[WARNING GRT-0350] Missing route to pin rob/io_enq_uops_0_rob_idx[3].
[WARNING GRT-0350] Missing route to pin rob/io_enq_uops_1_rob_idx[3].
[WARNING GRT-0350] Missing route to pin rob/io_enq_uops_2_rob_idx[3].
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_0_bits_rob_idx[2].
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_1_bits_rob_idx[2].
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_2_bits_rob_idx[2].
[WARNING GRT-0350] Missing route to pin mem_issue_unit/io_dis_uops_3_bits_rob_idx[2].
[deleted]
```

### Expected Behavior

@maliberty Said that GRT-0350 should never be observed.

### Environment

```shell
OpenROAD v2.0-10770-g6719c92e6
```

### To Reproduce

See above

eder-matheus:
@oharboe Could you try this branch: https://github.com/The-OpenROAD-Project/OpenROAD/pull/4263?

oharboe:
> @oharboe Could you try this branch: #4263?

[...]

PR #4263, opened 2023-11-14
Grt layer prediction on planar routing

arthurjolo:
Fixes [#4196](https://github.com/The-OpenROAD-Project/OpenROAD/issues/4196) 

When fastroute generates planar routes it uses the net's minimum layer information as a prediction to where the routes will be assigned, on MakeWireParasitics when estimating partial parasitics to pins, it checks if there is a routing to the pins on the net's minimum layer information from GlobalRouter (since layers haven't been assigned). However, the information of a net's minimum layer is different on fastroute and GlobalRouter, causing MakeWireParasitics to not find any routing to some pins, and it is fixed by making this information consistent between  fastroute and GlobalRouter.
Also this PR corrects the guides order for VIAs when saving them.


eder-matheus (approved):

src/grt/src/GlobalRouter.cpp
```diff
@@ -866,12 +866,11 @@ bool GlobalRouter::makeFastrouteNet(Net* net)
     // See https://github.com/The-OpenROAD-Project/OpenROAD/pull/2893 and
     // https://github.com/The-OpenROAD-Project/OpenROAD/discussions/2870
     // for a detailed discussion
-    int min_pin_layer = std::numeric_limits<int>::max();
+
     for (RoutePt& pin_pos : pins_on_grid) {
       fr_net->addPin(pin_pos.x(), pin_pos.y(), pin_pos.layer() - 1);
-      min_pin_layer = std::min(min_pin_layer, pin_pos.layer());
     }
-    fr_net->setMinLayer(std::max(min_pin_layer - 1, min_layer - 1));
```
eder-matheus, line 874: When the FrNet min layer is defined now?

arthurjolo: When it is created, on line 856. Previously the information was calculated again and then it was updated. Now the all the computing for the net's min layer is done on _getNetLayerRange_, so ther is no need to update it

PR #4264, opened 2023-11-14
Via not found above max routing layer bugfix

refaay:
…max layer


src/drt/src/io/io_parser_helper.cpp
```diff
@@ -45,6 +45,29 @@ using namespace boost::polygon::operators;
...
+        for (const auto& pin : iTerm->getTerm()->getPins()) {
+          for (const auto& fig : pin->getFigs()) {
+            if (((frShape*) (fig.get()))->getLayerNum() > TOP_ROUTING_LAYER) {
+              return true;
+            }
+          }
+        }
+      }
+    }
+  }
```
osamahammad21, line 67: Add a check to see if the pin is connected to a net that is to be routed by drt. If it's not then just skip it.

osamahammad21 (changes requested):

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -45,6 +45,33 @@ using namespace boost::polygon::operators;
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+  return false;
+}
+
```
osamahammad21, line 74: I think there is a better approach for this. We should find the top_pin_layer and check below against the max(TOP_ROUTING_LAYER, top_pin_layer). We don't need to get default vias for layer met10 when the top routing layer is met5 and the top_pin_layer is met7.

osamahammad21 (changes requested):

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -45,6 +45,38 @@ using namespace boost::polygon::operators;
...
+{
+  int topPinLayer = 0;
+  if (design_->getTopBlock()) {
+    for (const auto& bTerm : design_->getTopBlock()->getTerms()) {
+      if (bTerm->getNet() && !bTerm->getNet()->isSpecial()) {
+        for (const auto& pin : bTerm->getPins()) {
+          for (const auto& fig : pin->getFigs()) {
+            if (((frShape*) (fig.get()))->getLayerNum() > topPinLayer) {
+              topPinLayer = ((frShape*) (fig.get()))->getLayerNum();
+            }
```
osamahammad21, line 58: replace with 
`topPinLayer = std::max(topPinLayer, ((frShape*) (fig.get()))->getLayerNum())`

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -45,6 +45,38 @@ using namespace boost::polygon::operators;
...
+      }
+    }
+    for (const auto& inst : design_->getTopBlock()->getInsts()) {
+      for (const auto& iTerm : inst->getInstTerms()) {
+        if (iTerm->getNet() && !iTerm->getNet()->isSpecial()) {
+          for (const auto& pin : iTerm->getTerm()->getPins()) {
+            for (const auto& fig : pin->getFigs()) {
+              if (((frShape*) (fig.get()))->getLayerNum() > topPinLayer) {
+                topPinLayer = ((frShape*) (fig.get()))->getLayerNum();
+              }
```
osamahammad21, line 70: ditto

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -59,6 +91,9 @@ void io::Parser::initDefaultVias()
     auto viaDef = tech_->getVia(userDefinedVia);
     tech_->getLayer(viaDef->getCutLayerNum())->setDefaultViaDef(viaDef);
   }
+  // Check whether there are pins above top routing layer
+  int topPinLayer = getTopPinLayer();
```
osamahammad21, line 95: frLayerNum

osamahammad21 (changes requested):

src/drt/src/io/io.h
```diff
@@ -69,6 +69,7 @@ class Parser
   bool readGuide();
   void postProcess();
   void postProcessGuide();
+  int getTopPinLayer();
```
osamahammad21, line 72: frLayerNum

src/drt/src/io/io_parser_helper.cpp
```diff
@@ -45,6 +45,38 @@ using namespace boost::polygon::operators;
 using Rectangle = boost::polygon::rectangle_data<int>;
 namespace gtl = boost::polygon;
 
+int io::Parser::getTopPinLayer()
+{
+  int topPinLayer = 0;
```
osamahammad21, line 50: frLayerNum

osamahammad21 (changes requested):

osamahammad21 (approved):

PR #4272, opened 2023-11-15
mpl2: draw shapes based on the simulated annealing outline

AcKoucher:
Trying to make it easier to understand what is going on during both Cluster and Macro placement in the Hierarchical Macro Placement stage of HierRTLMP flow.

I also added a method to clear the drawning made by Graphics so we can see the result after running mpl2 flow while OR is still running.


maliberty (approved):

maliberty:
Please resolve the conflict with your previous PR

AcKoucher:
Done

PR #4279, opened 2023-11-17
RUDY Implementation in OpenROAD

ApeachM:
## Heatmap

| bench    | RUDY                                                         | GRT                                                          |
| -------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
| simple01 | image-20231115195718257-1700107163919-2 | image-20231115195733519 |
| medium01 | image-20231115205433996 | image-20231115205421309 |

## Time comparison with GRT (with `perf` profiling)

| BENCH    | GRT time                                                     | RUDY time                                                    |
| -------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
| medium01 | image-20231115212548216 | image-20231115212537824 |

ABOUT 8.85 TIMES FASTER

### Time reducing by R-tree usage

Reduce the runtime of `RUDY::calculate()` function as much as 90% by r-Tree

image-20231115211832072


maliberty:
What's the difference with #4278 ?

ApeachM:
> What's the difference with #4278 ?

The difference is [here](https://github.com/ApeachM/OpenROAD/blob/abab48a2df3a335bbbc124b567215b468299a518/src/gui/src/stub.cpp#L61-L66). Due to the lack of this part, the build was failed in $4278.

maliberty (changes requested):
partial review

maliberty:
The tiles form a grid so I'm not sure why an rtree is needed.  Is this for some future use case?

ApeachM:
> The tiles form a grid so I'm not sure why an rtree is needed. Is this for some future use case?

Constructing RTree is just for reducing the complexity of calculation in `calculateRUDY()`

src/odb/include/odb/util.h
```diff
@@ -486,6 +487,48 @@ unsigned int prime)
...
+    void setRect(int lx, int ly, int ux, int uy);
+    void setCoordinate(int x, int y);
+    void addRUDY(double_t rudy);
+    double_t getRUDY() { return rudy_; }
+
+   private:
+    odb::Rect rect_;
+    std::pair<int, int> coordinate_;
+    double_t rudy_ = 0;  // This value is real RUDY value * 1e2
+                         // for reducing floating point calculation
```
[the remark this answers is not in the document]
ApeachM: Oh, I forgot to remove the comments.. I will remove it and push again.

src/odb/src/zutil/util.cpp
```diff
@@ -43,6 +45,129 @@ namespace odb {
...
+    std::vector<GridValue> queryResults;
+    rtree.query(bgi::intersects(box(point(netBox.xMin(), netBox.yMin()),
+                                    point(netBox.xMax(), netBox.yMax()))),
+                std::back_inserter(queryResults));
+
+    for (const auto& value : queryResults) {
+      auto grid = value.second;
+      auto gridBox = grid->getRect();
+      if (netBox.intersects(gridBox)) {
+        auto s_k = netBox.intersect(gridBox).area();
```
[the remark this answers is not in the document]
ApeachM: I will refactor it! Thank you.

src/odb/src/zutil/util.cpp
```diff
@@ -43,6 +45,129 @@ namespace odb {
...
+
+  for (auto net : block_->getNets()) {
+    auto netBox = net->getTermBBox();
+    auto netArea = netBox.area();
+    if (netArea == 0) {
+      continue;
+    }
+    auto hpwl = static_cast<int64_t>(netBox.dx() + netBox.dy());
+    auto wireArea = hpwl * wireWidth;
+    auto netCongestion = wireArea * 1e4 / netArea;
```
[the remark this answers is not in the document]
ApeachM: For keeping 4 significant figures.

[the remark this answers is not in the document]
maliberty: Why do you want to limit the precision of the calculation?

maliberty:
> > What's the difference with #4278 ?
> 
> The difference is [here](https://github.com/ApeachM/OpenROAD/blob/abab48a2df3a335bbbc124b567215b468299a518/src/gui/src/stub.cpp#L61-L66). Due to the lack of this part, the build was failed in $4278.

The PR is to merge the branch not a specific commit.  When you push to the branch the current PR is updated.  You don't need to close/open a new one.

maliberty:
> > The tiles form a grid so I'm not sure why an rtree is needed. Is this for some future use case?
> 
> Constructing RTree is just for reducing the complexity of calculation in `calculateRUDY()`

How does it reduce the complexity?  It seems you can calculate the tiles covered by a rectangle with some simple math rather than needing an rtree.

ApeachM:
> > > What's the difference with #4278 ?
> > 
> > 
> > The difference is [here](https://github.com/ApeachM/OpenROAD/blob/abab48a2df3a335bbbc124b567215b468299a518/src/gui/src/stub.cpp#L61-L66). Due to the lack of this part, the build was failed in $4278.
> 
> The PR is to merge the branch not a specific commit. When you push to the branch the current PR is updated. You don't need to close/open a new one.

I didn't know that.. Thank you for letting me know this..! 
Is there any way to remove the closed PR?

ApeachM:
> > > The tiles form a grid so I'm not sure why an rtree is needed. Is this for some future use case?
> > 
> > 
> > Constructing RTree is just for reducing the complexity of calculation in `calculateRUDY()`
> 
> How does it reduce the complexity? It seems you can calculate the tiles covered by a rectangle with some simple math rather than needing an rtree.

At the first time, I implement the calculation like [this](https://github.com/The-OpenROAD-Project/OpenROAD/blob/1bbfb68a17def40b91c2e7aa9393597bb9e45529/src/gpl/src/routeBase.cpp#L982-L997), but it takes time as much as grt.

So I used the rtree method and it reduces the time as much as 10 [times](image).

Will be there any better method for this?

maliberty:
It looks like previously you iterated over all tiles which would be inefficient.  However in a rectangular array you can compute the X & Y intervals of overlap directly and just iterate over their cross product.

ApeachM:
> It looks like previously you iterated over all tiles which would be inefficient. However in a rectangular array you can compute the X & Y intervals of overlap directly and just iterate over their cross product.

Thank you very much. It is much faster as much as 3 times.
I made a commit for the change and pushed it.
image

src/odb/include/odb/util.h
```diff
@@ -486,6 +487,61 @@ unsigned int prime)
...
+  {
+   public:
+    odb::Rect getRect() { return rect_; }
+    void setRect(int lx, int ly, int ux, int uy);
+    void addRUDY(double_t rudy);
+    double_t getRUDY() { return rudy_; }
+
+   private:
+    odb::Rect rect_;
+    double_t rudy_ = 0;
```
maliberty, line 503: A float should be sufficient accuracy for an estimation and saves memory

src/odb/include/odb/util.h
```diff
@@ -486,6 +487,61 @@ unsigned int prime)
...
+  void setGridConfig(odb::Rect block, int tileCntX, int tileCntY);
+
+  /**
+   * Set the wire length for calculate RUDY.
+   * If the layer which name is metal1 and it has getWidth value, then this
+   * function will not applied, but it will apply that information.
+   * */
+  void setWireWidth(int wireWidth) { wireWidth_ = wireWidth; }
+
+  Tile& getTile(int x, int y) { return grid_.at(x).at(y); }
```
maliberty, line 527: This should be private and a const version should be public to preserve encapsulation.

src/gui/src/heatMapRUDY.h
```diff
@@ -0,0 +1,64 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2021, The Regents of the University of California
```
maliberty, line 4: 2023

src/gui/src/heatMapRUDY.cpp
```diff
@@ -0,0 +1,125 @@
+//////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2019-2023, The Regents of the University of California, Google
```
maliberty, line 4: 2023

rovinski: Additionally, don't include Google unless code was derived from Google.

rovinski:
Can there be a Docs update which describes this in more detail? RUDY is not a standard acronym so I had to look up what it is.

src/gui/src/heatMapRUDY.cpp
```diff
@@ -1,7 +1,7 @@
 //////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (c) 2019-2023, The Regents of the University of California, Google
+// Copyright (c) 2023, The Regents of the University of California
```
rovinski, line 4: rm LLC on the next line

src/gui/src/heatMapRUDY.cpp
```diff
@@ -0,0 +1,125 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include "heatMapRUDY.h"
+
+#include "odb/db.h"
+
+namespace gui {
+
+RUDYDataSource::RUDYDataSource(utl::Logger* logger, odb::dbDatabase* db)
+    : gui::HeatMapDataSource(logger, "RUDY", "RUDY", "RUDY")
```
maliberty, line 40: I am concerned that many users will not know about RUDY or what it stands for.  I think it is fine as an internal name for EDA developers in the code.

We already have 'Routing Congestion' heatmap so it needs to be distinguished from that.  Perhaps 'Estimated Congestion'?

@gadfort @rovinski @oharboe any user input?

oharboe: I think I perefer "Esimated Congestion", though that leaves the question as to why it is estimated and not actual. Also, what kind of congestion? Placement? Routing?

oharboe: Perhaps "Fast routing congestion estimate"?

maliberty: It is estimated from placement without actually running a router.  It can be used earlier in the flow.  My goal is to replace the calls to grt from gpl with it.

I'm afraid long names will get truncated in the GUI as the docking window is usually pretty narrow.

rovinski: Maybe "Estimated Congestion (RUDY)"?
Does the GUI support tooltips? Maybe more info could be added to a tooltip.

oharboe: Why would the user want to use this? Perhaps a name that is relevant to the user?

"Congestion Preview"?

+ manual for details...

QuantamHD: Here are my suggestions. Go one of two ways

1. Merge RUDY with the existing routing congestion option. Run Rudy first and display the result, and in the background calculate the grt solution then replace the heat map when complete.

2. Merge with existing routing congestion option, and in the double click options panel add a new drop down called algorithm. And include RUDY (fast) and GRT (Accurate). Choose a default, and then let a user choose which one they want.

rovinski: A user might want to use it if running GRT would take a long time. I imagine this would also work before DPL is run.

oharboe: > Here are my suggestions. Go one of two ways
> 
> 1. Merge RUDY with the existing routing congestion option. Run Rudy first and display the result, and in the background calculate the grt solution then replace the heat map when complete.
> 2. Merge with existing routing congestion option, and in the double click options panel add a new drop down called algorithm. And include RUDY (fast) and GRT (Accurate). Choose a default, and then let a user choose which one they want.

I like 1 best. Fast preview is an implementation detail of usability...

gadfort: I think "Estimated congestion" would be a good choice. It is clear to the user what it is and avoids confusion with the routing congestion data.

gadfort:
@ApeachM @maliberty it looks to be like the images posted scaled incorrectly. The RUDY data source is from 0% to 1%, and it should be 0% to 100%

maliberty:
> @ApeachM @maliberty it looks to be like the images posted scaled incorrectly. The RUDY data source is from 0% to 1%, and it should be 0% to 100%

This has been fixed, the images are just old.

maliberty (approved):

PR #4285, opened 2023-11-19
Fix `rsz` metric types

donn:
* Fix types of `timing__drv__floating__nets` and `timing__drv__floating__pins`, which were mistakenly reported as strings.


maliberty (approved):
Thanks

maliberty:
@vvbandeira FYI in case any infra cares

PR #4286, opened 2023-11-19
mpl2: fix SASoftMacro result update

AcKoucher:
After merging #4272, for some designs, the bundled io clusters were still being displayed in a weird location in debug mode after we ran SA for specifically placing children clusters of the root.

I realized then that when incorporating the SASoftMacro results for the root cluster, we were cheating the `fixed_` flag that we set to true when setting a cluster as an IO Cluster. This was being done by changing the SoftMacro pointer address directly and, this way, bypassing the flag. With this, when trying to move these clusters to their real location in the die, they would be stuck in a wrong location, however they should not have been displaced in the first place.

I'm running Secure-CI on this.

Also doing some tiny refactoring/cleaning.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -4074,12 +4074,8 @@ void HierRTLMP::multiLevelMacroPlacement(Cluster* parent)
   }
   file.close();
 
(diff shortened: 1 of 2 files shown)


AcKoucher:
Secure-CI passed.

eder-matheus (approved):

PR #4294, opened 2023-11-21
mpl2: consider placement blockages & some refactoring

AcKoucher:
Three things are being done here:

1. I'm creating a function to compute the blockages overlap with the parent's outline during cluster placement. Small steps into getting rid of too much duplicated (in this case triplicate..) code.
2. As I was adding the new function declaration I thought it could be a good idea to organize a little bit the private functions in hier_rtlmp.h, so I separated the declarations according HierRTLMP flow stages.
3. I'm ensuring we actually use the placement blockages in the design by changing `SACoreSoftMacro::setBlockages` to `SACoreSoftMacro::addBlockages` and properly populating the placement blockage list.

Obs: I also had to update Graphics to ensure we draw the placement blockages. I chose to use a different color.

I'm opening this as a draft, because I'm not sure if these changes are too messy. 
If it looks good I'll run CI.


src/mpl2/src/Mpl2Observer.h
```diff
@@ -65,6 +65,10 @@ class Mpl2Observer
   virtual void setMacroBlockages(const std::vector<mpl2::Rect>& macro_blockages)
   {
   }
+  virtual void setPlacementBlockages(
+      const std::vector<mpl2::Rect>& placement_blockages)
+  {
+  }
```
maliberty, line 71: Where is this overridden?  I don't see it in graphics

maliberty:
A quick look suggests this is reasonable.  Go ahead with a CI

AcKoucher:
Secure-CI passed.

src/mpl2/src/SACoreSoftMacro.cpp
```diff
@@ -1062,4 +1062,12 @@ void SACoreSoftMacro::calSegmentLoc(float seg_start,
   }
 }
 
+// The blockages here are only those that overlap with the annealing outline.
+void SACoreSoftMacro::addBlockages(const std::vector<Rect>& blockages)
+{
+  for (const auto& blockage : blockages) {
+    blockages_.push_back(blockage);
+  }
```
eder-matheus, line 1070: Perhaps you could use std::vector::insert? Having a one-line function might be clearer:
`blockages_.insert(blockages_.end(), blockages .begin(), blockages.end());`

eder-matheus (approved):
LGTM, but I'll let @maliberty check before merging it.

PR #4295, opened 2023-11-21
odb: generating a default destructor for classes without destructible fields

fgaray:
To fix clang-tidy's warning:

image

### DIFF
--- a/src/odb/src/codeGenerator/gen.py
+++ b/src/odb/src/codeGenerator/gen.py
@@ -264,6 +264,12 @@ for klass in schema["classes"]:
         else:
             field["setterArgumentType"] = field["getterReturnType"] = field["type"]
 
+        if field["name"] == '_name' and 'no-destruct' not in field["flags"]:
+          klass["has_destructible_fields"] = True
+
+        if "table" in field:
+          klass["has_destructible_fields"] = True
+
     klass["fields"] = [field for field in klass["fields"] if "bits" not in field]
 
     klass["hasBitFields"] = False
--- a/src/odb/src/codeGenerator/gen.py
+++ b/src/odb/src/codeGenerator/gen.py
@@ -266,7 +266,7 @@ for klass in schema["classes"]:
 
         # For fields that we need to free/destroy in the destructor
         if field["name"] == '_name' and 'no-destruct' not in field["flags"] or "table" in field:
-          klass["has_destructible_fields"] = True
+          klass["needs_non_default_destructor"] = True
 
     klass["fields"] = [field for field in klass["fields"] if "bits" not in field]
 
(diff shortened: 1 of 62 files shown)


src/odb/src/codeGenerator/helper.py
```diff
@@ -168,3 +168,8 @@ def get_ref_type(type_name):
         return None
 
     return type_name[6:-1] + "*"
+
+def has_destructible_field(klass):
+  for field in klass["fields"]:
+    return True
+  return False
```
maliberty, line 175: What is this for?

src/odb/src/db/dbAccessPoint.h
```diff
@@ -58,7 +58,8 @@ class _dbAccessPoint : public _dbObject
  public:
   _dbAccessPoint(_dbDatabase*, const _dbAccessPoint& r);
   _dbAccessPoint(_dbDatabase*);
-  ~_dbAccessPoint();
+
+  ~_dbAccessPoint() = default;
```
oharboe, line 62: Don't know if this applies here: this destructor is not virtual, so if there are any subclasses of this class, there could be a memory leak since the destructor of the subclass won't be run.

maliberty: There is no change in that regard. There classes are not used with inheritance

src/odb/src/codeGenerator/helper.py
fgaray: Was originally for checking the fields but I ended implementing it directly in gen.py. Removed!

src/odb/src/codeGenerator/gen.py
```diff
@@ -264,6 +264,10 @@ def get_json_files(directory):
         else:
             field["setterArgumentType"] = field["getterReturnType"] = field["type"]
 
+        # For fields that we need to free/destroy in the destructor
+        if field["name"] == '_name' and 'no-destruct' not in field["flags"] or "table" in field:
+          klass["has_destructible_fields"] = True
```
maliberty, line 269: has_destructible_fields isn't a clear name to me.  All the fields are destructible, what varies is whether custom destructor code is needed.   Also a user might need to write dtor code that isn't captured by the above conditions in the "//User Code" section.

Perhaps has_non_default_destructor ?

fgaray: I think that makes sense, but a better name  may be "needs_non_default_destructor" because klass do not have a destructor yet, wdyt?

maliberty: Works for me

fgaray: Changed!

maliberty (approved):

PR #4303, opened 2023-11-23
WRONGDIRECTION SPACING routing rule support

refaay:


maliberty:
> Hi @vvbandeira , do you know why I am getting "Check That ODB Files Are Generated / format (pull_request)" failing? It tells me "Your ODB generated files are out of date". I updated the db schema versions. Is there something else that I should do? Thanks

You've changed the odb schema files but not re-run the code generator so things are out of sync.  Re-run the code generator.

refaay:
> > Hi @vvbandeira , do you know why I am getting "Check That ODB Files Are Generated / format (pull_request)" failing? It tells me "Your ODB generated files are out of date". I updated the db schema versions. Is there something else that I should do? Thanks
> 
> You've changed the odb schema files but not re-run the code generator so things are out of sync. Re-run the code generator.

I re-ran the codeGenerator and confirmed this in the commit before this reply, but it still fails odb files gen check.

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -905,6 +911,192 @@ void FlexGCWorker::Impl::checkMetalSpacing()
...
+    }
+    if (con->hasPrlLength()) {
+      auto prlLength = con->getPrlLength();
+      if (prlLength < 0) {
+        logger_->warn(DRT,
+                      334,
+                      "Unsupported negative value for prlLength in branch PRL "
+                      "in checkMetalSpacing_wrongDir.");
+      }
+    }
```
osamahammad21, line 965: Could you move these checks to IO so that we don't get a warning each time we do a gc check?

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -905,6 +911,192 @@ void FlexGCWorker::Impl::checkMetalSpacing()
...
+    for (auto& edges : pin->getPolygonEdges()) {
+      for (auto& edge : edges) {
+        // Check wrongDir edge
+        if (edge->isVertical() != layer->isVertical()) {
+          // Check noneol flag
+          if (con->hasNoneolWidth()) {
+            // Get edge length and compare
+            auto edgeLength = gtl::length(*edge);
+            auto noneolLength = con->getNoneolWidth();
+            if (edgeLength < noneolLength) {
```
osamahammad21, line 977: I feel it should be <=. Do you have any justification for using < only?

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -905,6 +911,192 @@ void FlexGCWorker::Impl::checkMetalSpacing()
...
+            if (ptr->isVertical() != layer->isVertical()) {
+              if (edge.get() == ptr) {
+                continue;
+              }
+
+              auto net1 = edge->getNet();
+              auto net2 = ptr->getNet();
+              if (net1 == net2) {
+                continue;
+              }
```
osamahammad21, line 1001: I see nothing in the rule that excludes same net checks

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -905,6 +911,192 @@ void FlexGCWorker::Impl::checkMetalSpacing()
...
+              const frCoord edge1_min = std::min(edge1_low, edge1_high);
+              const frCoord edge1_max = std::max(edge1_low, edge1_high);
+
+              const frCoord edge2_low = ptr->low().get(orient);
+              const frCoord edge2_high = ptr->high().get(orient);
+              const frCoord edge2_min = std::min(edge2_low, edge2_high);
+              const frCoord edge2_max = std::max(edge2_low, edge2_high);
+              const frCoord prl = std::min(edge1_max, edge2_max)
+                                  - std::max(edge1_min, edge2_min);
+              // Check prl run length > 0
```
osamahammad21, line 1025: I feel this part is a replica of the prl check in checkMetalEndOfLine_eol_hasEol_check, Could you please reduce it to a callable method by both functions. In general we want to minimize code.

src/drt/src/db/tech/frConstraint.h
```diff
@@ -1145,6 +1148,62 @@ class frLef58SpacingEndOfLineConstraint : public frConstraint
...
+  }
+
+ protected:
+  frCoord wrongDirSpace{0};
+  bool noneolValid{false};
+  frCoord noneolWidth{0};
+  bool lengthValid{false};
+  frCoord length{0};
+  bool prlLengthValid{false};
+  frCoord prlLength{0};
```
osamahammad21, line 1204: I see no reason for having both prlLengthValid and prlLength. if no PRL specified, then prlLength would be 0. Same for the odb object

osamahammad21 (changes requested):
In general, I prefer if we avoid data replication. I don't see a need for replicating all the variables in both the odb object and the drt constraint object. It would be better if the drt constraint object held a pointer of odb object and you use the later's APIs directly. Check the frLef58KeepOutZoneConstraint for reference

src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -905,6 +911,182 @@ void FlexGCWorker::Impl::checkMetalSpacing()
...
+              }
+
+              // Get edges prl
+              const gtl::orientation_2d orient = getOrientation(edge.get());
+              const frCoord prl = getPrl(edge.get(), ptr, orient);
+              // Check prl run length > 0 (avoid error with negative
+              // prlLength)
+              if (prl <= 0) {
+                continue;
+              }
```
osamahammad21, line 1018: not needed

src/drt/src/io/io.cpp
```diff
@@ -1403,6 +1403,23 @@ void io::Parser::setRoutingLayerProperties(odb::dbTechLayer* layer,
...
+    if (rule->isLengthValid()) {
+      logger_->warn(
+          DRT, 333, "Unsupported branch LENGTH in WrongDirSpacingRule.");
+    }
+    if (rule->getPrlLength() < 0) {
+      logger_->warn(DRT,
+                    334,
+                    "Unsupported negative value for prlLength in branch PRL "
+                    "in WrongDirSpacingRule.");
+    }
```
osamahammad21, line 1419: you should skip those rules. add continue; after the warn

src/drt/src/gc/FlexGC_impl.h
```diff
@@ -218,6 +218,14 @@ class FlexGCWorker::Impl
   void initRegionQuery();
 
   void checkMetalSpacing();
+  void checkMetalSpacing_wrongDir_getQueryBox(gcSegment* edge,
+                                              frCoord spcVal,
+                                              box_t& queryBox);
+  gtl::orientation_2d getOrientation(gcSegment* edge) const;
```
osamahammad21, line 224: Good idea to have this as a method, But I think it should be a public method for gcSegment

osamahammad21 (changes requested):

src/drt/src/gc/FlexGC_impl.h
refaay: I moved getOrientation() inside gcSegment as noted.
Did you mean the same for checkMetalSpacing_wrongDir_getQueryBox() too?

src/drt/src/db/gcObj/gcShape.h
```diff
@@ -209,6 +209,12 @@ class gcSegment : public gtl::segment_data<frCoord>, public gcShape
         return frDirEnum::UNKNOWN;
     }
   }
+  gtl::orientation_2d getOrientation() const
+  {
+    const frDirEnum dir = this->getDir();
```
osamahammad21, line 214: just getDir(), why this->getDir() ?

osamahammad21 (changes requested):

osamahammad21 (approved):

src/odb/test/data/design.odb
maliberty, line 1: Why is this modified?  It is to test backward compatibility so it shouldn't be changed.

maliberty:
@osamahammad21 are you satisfied with the drt changes?

src/odb/test/data/design.odb
refaay: I reverted it

maliberty (approved):

Issue #4301, opened 2023-11-23
Feature request - remove confusing warning in io_placement_random.tcl - [WARNING PPL-0015] Macro core/FpPipeline/fregfile/regfile_ext is not placed.


oharboe:
### Describe the bug

unzip https://drive.google.com/file/d/17-H3xSv_emeozTsXrW7eKzUb2y-uAjYS/view?usp=sharing

```
$ ./run-me-BoomTile-asap7-base.sh 
OpenROAD v2.0-11072-g2d83b7137 
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
[WARNING PPL-0015] Macro core/FpPipeline/fregfile/regfile_ext is not placed.
[WARNING PPL-0015] Macro core/iregfile/regfile_ext is not placed.
[WARNING PPL-0015] Macro dcache/data/array_0_0_0_ext is not placed.
[WARNING PPL-0015] Macro dcache/data/array_0_0_1_ext is not placed.
[WARNING PPL-0015] Macro dcache/data/array_0_1_0_ext is not placed.
[WARNING PPL-0015] Macro dcache/data/array_0_1_1_ext is not placed.
[deleted]
```

### Expected Behavior

No warning

### Environment

```shell
OpenROAD v2.0-11072-g2d83b7137
```

### To Reproduce

See above

### Additional Context

This is from the same test project as for which #4299 was filed, but possibly a separate problem.

maliberty:
Are you not running macro placement before pin placement?

oharboe:
> Are you not running macro placement before pin placement?

I'm using ORFS and doing whatever ORFS is doing...

The command I ran before running `make io_placement_random_issue` was:

```
./orfs make do-floorplan
```

It runs macro placement *after* io_placement_random.tcl:

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/c72ffd7dcd317c337df2535314289f8d0a593467/flow/Makefile#L609

oharboe:
@maliberty Perhaps worth fixing #4299 and then retesting?

oharboe:
@maliberty Tried with the #4302 fix, still problems.

unzip [io_placement_random.zip](https://github.com/The-OpenROAD-Project/OpenROAD/files/13451004/io_placement_random.zip)

```
$ ./run-me-L1MetadataArray-asap7-base.sh 
[...]

PR #4304, opened 2023-11-23
ppl: show warn message on macro not placed only for non-random runs

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/4301

### DIFF
--- a/src/ppl/test/macro_not_placed.tcl
+++ b/src/ppl/test/macro_not_placed.tcl
@@ -1,0 +1,11 @@
+# region blocked by macro in die boundary
+source "helpers.tcl"
+read_lef sky130hd/sky130hd.tlef
+read_lef sky130hd/sky130_fd_sc_hd_merged.lef
+read_lef blocked_region.lef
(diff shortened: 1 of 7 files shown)


src/ppl/test/macro_not_placed.tcl
```diff
@@ -0,0 +1,17 @@
...
+
+place_pin -pin_name clk -layer met4 -location {40 30} -force_to_die_boundary
+place_pin -pin_name resp_rdy -layer met5 -location {50 15} -force_to_die_boundary
+place_pins -hor_layer met3 -ver_layer met2
+
+set def_file [make_result_file macro_not_placed.def]
+
+write_def $def_file
+
+diff_file macro_not_placed.defok $def_file
```
maliberty, line 17: Since you are only testing the presence or absence of the message you can skip having/checking defok.  You can updated and force push to remove the files from the history

maliberty (approved):

PR #4322, opened 2023-11-28
grt: Converting TreeNode and TreeEdge pointers to vectors

luis201420:


src/grt/src/fastroute/include/DataType.h
```diff
@@ -169,7 +169,8 @@ struct TreeNode
   short heights[max_connections];
   int eID[max_connections];
 
-  short x, y;   // position in the grid graph
+  short x, y;  // position in the grid graph
+  int nbrCount = 0;
```
eder-matheus, line 173: Use snake_case for member names: `nbr_count`

src/grt/src/fastroute/src/maze3D.cpp
```diff
@@ -895,8 +895,8 @@ void FastRouteCore::mazeRouteMSMDOrder3D(int expand,
 
     int enlarge = expand;
     const int num_terminals = sttrees_[netID].num_terminals;
-    const auto& treeedges = sttrees_[netID].edges;
-    const auto& treenodes = sttrees_[netID].nodes;
+    auto& treeedges = sttrees_[netID].edges;
+    auto& treenodes = sttrees_[netID].nodes;
```
eder-matheus, line 899: Why they're losing const? Here and below.

eder-matheus (changes requested):

src/grt/src/fastroute/src/maze3D.cpp
luis201420: Because those variables are then modified in the function

eder-matheus: I didn't notice some function calls that use these variables were marked as non-const.

eder-matheus (approved):
Waiting for secure-ci before merging.

PR #4326, opened 2023-11-29
adding the get_pin_arr API

LukeChristianMartin:
Adding pin arrival Python API for CircuitOps


maliberty:
You need to sign your commits (git commit -s) in order to pass DCO.  Please see the 'Details' link for info on how to correct this commit.

QuantamHD:
Please address the clang-tidy comments and run clang-format on the code.

VidyaChhabria:
@LukeChristianMartin you can clean up as requested above. Clang and the signing off commits.  Feel free to utilize the same PR to add in the transitions as well.

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+sta::Network* Design::cmdLinkedNetwork()
+{
+  sta::Network *network = sta::Sta::sta()->cmdNetwork();;
+  if (network->isLinked()) {
+    return network;
+  }
+
+  getLogger()->error(utl::ORD, 104, "STA network is not linked.");
+  return nullptr;
+
```
maliberty, line 110: error throws an exception so return is unnecessary

include/ord/Design.h
```diff
@@ -175,6 +186,19 @@ class Design
...
+  bool isInPower(odb::dbITerm* iterm);
+  bool isInGround(odb::dbITerm* iterm);
+  sta::ClockSeq findClocksMatching(const char *pattern, bool regexp, bool nocase);
+  sta::Clock* defaultArrivalClock();
+  std::string getITermName (odb::dbITerm* ITerm);
+  float getPinArrival(odb::dbITerm* db_pin, const std::string& rf);
+  std::vector<float> arrivalsClk(const sta::RiseFall *rf, sta::Clock *clk, const sta::RiseFall *clk_rf, sta::Vertex *vertex);
+  float getPinArrivalTime(sta::Clock *clk, const sta::RiseFall *clk_rf, sta::Vertex *vertex, const std::string& arrrive_or_hold);
+  sta::Graph* cmdGraph();
+  sta::Network* cmdLinkedNetwork();
```
maliberty, line 198: What does 'cmd' denote?

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+sta::Vertex** Design::vertices(const sta::Pin *pin)
+{
+  sta::Vertex *vertex, *vertex_bidirect_drvr;
+  static sta::Vertex *vertices[2];
+
+  cmdGraph()->pinVertices(pin, vertex, vertex_bidirect_drvr);
+  vertices[0] = vertex;
+  vertices[1] = vertex_bidirect_drvr;
+  return vertices;
+}
```
maliberty, line 128: Don't use a static.  Just return std::array<sta::Vertex*, 2>

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+
+std::string Design::getITermName (odb::dbITerm* ITerm)
+{
+    auto MTerm_name = ITerm->getMTerm()->getName();
+    auto inst_name  = ITerm->getInst()->getName();
+    return inst_name + "/" + MTerm_name;
+}
+
+bool Design::isTimeInf(float time) {
+  return (time > 1e+10 || time < -1e+10);
```
maliberty, line 157: Use INF from sta

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+}
+
+bool Design::isTimeInf(float time) {
+  return (time > 1e+10 || time < -1e+10); 
+}
+
+float Design::getPinArrivalTime(sta::Clock *clk,
+                               const sta::RiseFall *clk_rf,
+                               sta::Vertex *vertex,
+                               const std::string& arrrive_or_hold)
```
maliberty, line 163: Why is this a string and not sta::RiseFall ?

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+{
+  const sta::RiseFall *rf = (arrrive_or_hold == "arrive")? sta::RiseFall::rise(): sta::RiseFall::fall();
+  std::vector<float> times = arrivalsClk(rf, clk, clk_rf, vertex);
+  float delay = -1e+10;
```
maliberty, line 167: INF

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+  for (float delay_ : times){
```
maliberty, line 168: Don't use names with _ for local variables - they are for class members.

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+  sta::PatternMatch matcher(pattern, regexp, nocase, sta->tclInterp());
+  return sta::Sta::sta()->sdc()->findClocksMatching(&matcher);
+}
+
+sta::Clock* Design::defaultArrivalClock()
+{
+  return sta::Sta::sta()->sdc()->defaultArrivalClock();
+}
+
+float Design::getPinArrival(odb::dbITerm* db_pin, const std::string& rf) {
```
maliberty, line 191: Don't use a string, use the sta type

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+    std::vector<float> pin_arr;
+    int num_vertex_elements = 2;
```
maliberty, line 193: const

src/Design.cc
```diff
@@ -98,6 +98,120 @@ void Design::readDef(const std::string& file_name,
...
+    sta::Pin* sta_pin = sta->getDbNetwork()->dbToSta(db_pin);
+
+    sta::Vertex** vertex_arrray = vertices(sta_pin);
+    float delay = -1;
+    for (int i = 0; i < num_vertex_elements; i++) {
+      sta::Vertex* vertex = vertex_arrray[i];
+      if (vertex != nullptr) {
+        std::string arrival_or_hold = (rf == "rise")? "arrive":"hold";
+        delay = std::max(delay, getPinArrivalTime(nullptr, sta::RiseFall::rise(), vertex, arrival_or_hold));
+        delay = std::max(delay, getPinArrivalTime(defaultArrivalClock(), sta::RiseFall::rise(), vertex, arrival_or_hold));
```
maliberty, line 205: why does the first arg vary?

maliberty (changes requested):

src/Design.cc
LukeChristianMartin: I believe this is done this way because because we pass in a string object from the python interface!

src/Design.cc
LukeChristianMartin: We've tried putting the filter on time with -sta::INF and sta::INF.

We need a tighter constraint!

src/Design.cc
LukeChristianMartin: This is facing the python interface so the user would place a string for this input variable

src/Design.cc
LukeChristianMartin: @VidyaChhabria Can you help me answer this!

src/Design.cc
maliberty: Why is a string necessary?  The dp_pin isn't a string...

src/Design.cc
maliberty: If you really want to use a string then you should provide a typemap in the .i but keep the c++ interface strongly typed.

maliberty:
Please avoid force pushing after a review has started as it makes it impossible to see the changes from the last review.

rovinski:
@maliberty I think it was necessary in this case to fix DCO

maliberty:
> @maliberty I think it was necessary in this case to fix DCO

ok that makes sense.  Hopefully it won't be an issue in the future

src/Design.cc
LukeChristianMartin: It wasn't able to filter out the Infinities that are generated by the STA application

VidyaChhabria: @maliberty internally sta uses these numbers to filter out infinities. Its not checking for an exact infinity. We have replicated the proc times_are_inf from sta search.tcl.

maliberty:
There are some unanswered questions

src/Design.cc
```diff
@@ -266,6 +380,18 @@ bool Design::isInClock(odb::dbInst* inst)
...
+  auto* net = iterm->getNet();
+  return (net != nullptr && net->getSigType() == odb::dbSigType::POWER);
+}
+
+bool Design::isInGround(odb::dbITerm* iterm)
+{
+  auto* net = iterm->getNet();
+  return (net != nullptr && net->getSigType() == odb::dbSigType::GROUND);
+}
+
```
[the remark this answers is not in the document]
VidyaChhabria: Yes I think an isSupply would suffice.  And yes getMterm()->getSigType would be better. @LukeChristianMartin  could you make this change?

PR #4337, opened 2023-12-04
pdn: add missing expr to returns

nmoroze:
Several of the procs in PdnGen.tcl return an expression that's missing an `expr` wrapper. I believe this is a bug, since a `return` won't evaluate an expression (but rather, the last word will be returned directly):

```tcl
proc foo {} {
  return 0 && 1
}

puts [foo]
```

prints 1 rather than 0, as expected.

I caught this with a tool I'm working on, [`tclint`](https://github.com/nmoroze/tclint): 

```
$ tclint . | grep return
```

### DIFF
--- a/src/pad/src/PdnGen.tcl
+++ b/src/pad/src/PdnGen.tcl
@@ -799,7 +799,7 @@ proc is_inst_in_voltage_domain {inst domain_name} {
   set power_net [dict get $voltage_domain primary_power]
   set ground_net [dict get $voltage_domain primary_ground]
 
-  return [match_inst_connection $inst $power_net] && [match_inst_connection $inst $ground_net]
+  return [expr {[match_inst_connection $inst $power_net] && [match_inst_connection $inst $ground_net]}]
 }
 
 proc get_block_inst_masters {} {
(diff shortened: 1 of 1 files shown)


maliberty (approved):
@gadfort FYI

gadfort:
@maliberty I think we should incorporate @nmoroze tool into the CI for OR for the tcl files.

maliberty:
@vvbandeira re: tclint in the CI

nmoroze:
I'd definitely love to get tclint into the CI and am willing to help. I opened this issue with just a quick fix since it's a functional issue (and is the only one I've found so far during my early experiments on this codebase). 

Overall, a run on the codebase (with the following settings to quiet things up a tad) returns quite a few violations:

```console
$ cat tclint.toml # settings
[style]
indent = 2
line-length = 100
$ tclint . | wc -l
23965
```

These are style related, and my tool doesn't do any sort of auto-formatting yet, so it might be a big effort to clean this up (and I imagine there'll have to be some improvements on the tool side, to make sure the style enforcement is correct & configurable to your taste). 

It would probably be best to adopt this incrementally, by enabling linting only for certain groups of files and/or certain checks at a time. 

I can open a new issue for further discussion if there's interest.

maliberty:
If you are willing to help that would be great.  Please open an issue to track it.

nmoroze:
Issue opened here: #4347

PR #4339, opened 2023-12-04
mpl2: fix bus planning flag, add first unit test, included mpl2 tests in OR regression and cleanup

AcKoucher:
What is being done here:
1) Turning bus planning into an actual flag - renaming related functions/variables;
2) Changing the way we check for the existence of bundledIO cluster;
3) Cleaning up the mpl2 test directory;
4) Adding test case for macro only case to test for when we have neither bundledIO cluster nor std cells;
5) Including this added test case in regression in OR
6) Suppressing thread count message;
7) Fixing fake_macro.lib to get rid of ODB warnings


src/mpl2/README.md
```diff
@@ -65,7 +65,7 @@ rtl_macro_placer
 | `-target_dead_space` | Specifies the target dead space percentage, which influences the utilization of `StandardCellCluster`. The allowed values are floats, and the default value is `0.05`. |
 | `-min_ar` | Specifies the minimum aspect ratio $a$, or the ratio of its width to height of a `StandardCellCluster` from $[a, \frac{1}{a}]$. The allowed values are floats, and the default value is `0.33`. |
 | `-snap_layer` | Snap macro origins to this routing layer track. The default value is 4, and the allowed values are integers `[1, MAX_LAYER]`). |
-| `-bus_planning_flag` | Flag to enable bus planning. The recommendation is to turn on bus planning for SKY130 and off for NanGate45/ASAP7. The default value is disabled. |
+| `-bus_planning` | Flag to enable bus planning. The recommendation is to turn on bus planning for SKY130 and off for NanGate45/ASAP7. The default value is disabled. |
```
maliberty, line 68: This appears to be based on https://github.com/The-OpenROAD-Project/OpenROAD/blob/e4ca4c0b0bbf5fe69cb4e6083f4793e9d2a2ffca/src/mpl2/src/hier_rtlmp.h#L272

I think the description about limited routing layers is more helpful as people may use this with other PDKs.  GF180 sounds like another candidate for bus planning.

I know this is incidental to your change but let's fix it while we are looking at it.

maliberty (changes requested):

maliberty:
Have you tried running any design with only macros to make sure it works?

src/mpl2/README.md
AcKoucher: Should I leave these technologies used as examples or just use the description about limited routing layers?

AcKoucher:
> Have you tried running any design with only macros to make sure it works?

I tried running `src/mpl2/test/mp_test1.tcl` which has macros and IO pins, but no std cells - it worked fine.  (I had to use smaller halos, as, apparently, the way it is doesn't work - I can also fix that).

Would it be necessary to come up with a test with only macros at all, i. e., no IO pins?

maliberty:
> Would it be necessary to come up with a test with only macros at all, i. e., no IO pins?
No

AcKoucher:
Secure-CI passed.

maliberty (approved):

AcKoucher:
The regression diff is showing that the failure is regarding the difference in the number of threads used, but it also seems like the result from the pipeline differs from the result I get in my end.

Is there a way to not account for the `[INFO ORD-0030] Using 20 thread(s).`? I also thought with the changes in #4288 it wouldn't have any problems regarding different results.

maliberty:
The choices I see
1) run the test single threaded if fast enough
2) suppress the message in test/helpers.tcl & py

If you are getting a different result with a different number of threads that is concerning.  We expect the same result regardless.  I fixed a bug like that a little while back but perhaps there are others.

AcKoucher:
I'll test the singled threaded version, if it's to slow, I'll suppress the message.

I'll also investigate the different result, I think it could be because there's was a similar bug when we check results for HardMacro SA.

AcKoucher:
@maliberty Suppressing the message that informs the number of threads results in failing of the Restructure test rmp/test/gcd_restructure.tcl (and .py). 

This case shows always the ORD message with number of threads = 3 (it's manually set in the test). May I update the .ok for this one?

src/mpl2/test/macro_only.tcl
```diff
@@ -0,0 +1,20 @@
...
+read_lef "./Nangate45/Nangate45_tech.lef"
+read_lef "./Nangate45/fake_macros.lef"
+read_liberty "./Nangate45/fake_macros.lib"
+
+read_verilog "./testcases/macro_only.v"
+link_design "macro_only"
+
+read_def "./testcases/macro_only.def" -floorplan_initialize
+
+set_thread_count 0
```
maliberty, line 14: Are you confident that changing the number of threads will not change the result?

maliberty (approved):

src/mpl2/test/macro_only.tcl
AcKoucher: I tested all different possibilities at my end (1 ~ 8), and as CI runs with a larger number of threads  (20 if I'm not mistaken), it should be fine.

PR #4350, opened 2023-12-06
Grt delete fr net

arthurjolo:
Replaces the temporary solution of marking FrNets as deleted and not deleting them. Now FrNets are properly deleted, however it still isn't updating fastroute's nets_ vector, it sets the entry of the deleted net as _nullptr_, and fastrotue then always skips _nullptr_ entries. Since there aren't many deleted nets even for the biggest designs the memory waste isn't big, but it is something to look in the future.


src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -1020,8 +1022,8 @@ NetRouteMap FastRouteCore::run()
   // debug mode Rectilinear Steiner Tree before overflow iterations
   if (debug_->isOn() && debug_->rectilinearSTree_) {
     for (int netID = 0; netID < netCount(); netID++) {
-      if (nets_[netID]->getDbNet() == debug_->net_
-          && !nets_[netID]->isRouted()) {
+      if (nets_[netID]->getDbNet() == debug_->net_ && !nets_[netID]->isRouted()
+          && nets_[netID] != nullptr) {
```
eder-matheus, line 1026: Move the nullptr check to the beginning of the if statement.

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -1272,8 +1274,8 @@ NetRouteMap FastRouteCore::run()
   // Debug mode Tree 2D after overflow iterations
   if (debug_->isOn() && debug_->tree2D_) {
     for (int netID = 0; netID < netCount(); netID++) {
-      if (nets_[netID]->getDbNet() == debug_->net_
-          && !nets_[netID]->isRouted()) {
+      if (nets_[netID]->getDbNet() == debug_->net_ && !nets_[netID]->isRouted()
+          && nets_[netID] != nullptr) {
```
eder-matheus, line 1278: Move the nullptr check to the beginning of the if statement.

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -1330,8 +1332,8 @@ NetRouteMap FastRouteCore::run()
   // Debug mode Tree 3D after layer assignament
   if (debug_->isOn() && debug_->tree3D_) {
     for (int netID = 0; netID < netCount(); netID++) {
-      if (nets_[netID]->getDbNet() == debug_->net_
-          && !nets_[netID]->isRouted()) {
+      if (nets_[netID]->getDbNet() == debug_->net_ && !nets_[netID]->isRouted()
+          && nets_[netID] != nullptr) {
```
eder-matheus, line 1336: Move the nullptr check to the beginning of the if statement.

eder-matheus (changes requested):

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -1765,7 +1765,7 @@ void FastRouteCore::copyRS(void)
 
   if (!sttrees_bk_.empty()) {
     for (netID = 0; netID < netCount(); netID++) {
-      if (nets_[netID]->isRouted() || nets_[netID]->isDeleted()) {
+      if (nets_[netID] == nullptr || nets_[netID]->isRouted()) {
```
maliberty, line 1768: This is repeated many times so you might make a ```FastRouteCore::isRouted(int netID)``` method to encapsulate this pattern.

arthurjolo: I created a method `FastRouteCore::skipNet(int netID)`. Since this check isn't only checking if the net is routed I belive skipNet was a more appropriate name, if you hink other wise I can change it to `FastRouteCore::isRouted(int netID)`

PR #4351, opened 2023-12-06
Add format_as for fmt v10 formatting

CircuitCoder:
This commit adds `format_as` for fmt v10.

Because of ADL, every namespace that contains serializable types needs to include this function. Right now, this is done by adding `using utl::format_as` in the same header that defines these types. We may also write all namespaces in `utl/Logger.h`. Which do you think is better?


src/utl/include/utl/Logger.h
```diff
@@ -327,3 +327,40 @@ class Logger
...
+} // anonymous namespace
+
+template <class T, class = std::enable_if_t<decltype(test_ostream::test<T>(0))::value>>
+auto format_as(T t) { return fmt::streamed(t); }
+
+} // namespace rtl
+
+#include <boost/asio/ip/address.hpp>
+namespace boost::asio::ip {
+inline static auto format_as(const boost::asio::ip::address f) { return fmt::streamed(f); }
```
maliberty, line 356: Why does this need to be here as most of the program doesn't need to depend on asio?

src/utl/include/utl/Logger.h
```diff
@@ -327,3 +327,40 @@ class Logger
...
+  template <class T>
+  static auto test(int) -> decltype(std::declval<std::ostream>() << std::declval<T>(), std::true_type());
+
+  template <class>
+  static auto test(...) -> std::false_type;
+};
+
+} // anonymous namespace
+
+template <class T, class = std::enable_if_t<decltype(test_ostream::test<T>(0))::value>>
```
maliberty, line 349: What is the '0' for?

maliberty (commented):
This looks good to me.  Is this sufficient to compile with fmt v10?

maliberty:
I prefer it this way to "We may also write all namespaces in utl/Logger.h" which would pull too much into utl.

maliberty:
It looks like this doesn't work with the older fmt:
```
#8 9.471 /OpenROAD/src/utl/include/utl/Logger.h: In function 'auto utl::format_as(T)':
#8 9.471 /OpenROAD/src/utl/include/utl/Logger.h:350:35: error: 'streamed' is not a member of 'fmt'
#8 9.471   350 | auto format_as(T t) { return fmt::streamed(t); }
```

CircuitCoder:
> I prefer it this way to "We may also write all namespaces in utl/Logger.h" which would pull too much into utl.

OK. Then should the asio part be moved to the involved source file, or stay at `util/Logger.h`?

> It looks like this doesn't work with the older fmt:

I apparently made a typo regarding version check (10000 / 100000). That was corrected in the commit I just pushed.

src/utl/include/utl/Logger.h
CircuitCoder: It can also be moved to the cpp file which uses asio. There are two file that depends on this:
- dst/src/LoadBalancer.cc
- dst/src/LoadBalancer.cc

I will move this piece of code into those two files.

CircuitCoder: Updated in 928fe1a375ab0630fe97490ba4b1e9bd06ef7aa2

src/utl/include/utl/Logger.h
```diff
@@ -327,3 +327,40 @@ class Logger
...
+auto format_as(T t) { return fmt::streamed(t); }
+
+} // namespace rtl
```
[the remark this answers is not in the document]
CircuitCoder: Updated in 928fe1a375ab0630fe97490ba4b1e9bd06ef7aa2

src/utl/include/utl/Logger.h
```diff
@@ -327,3 +327,40 @@ class Logger
 #undef GENERATE_STRING
 
 }  // namespace utl
+
+// Define stream_as for fmt > v10
+#if !SWIG && FMT_VERSION >= 100000
+
+namespace utl {
+
+namespace {
```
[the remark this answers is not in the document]
CircuitCoder: This unnamed namespace has been removed in 0d0991c12

CircuitCoder:
Fixed clang-format errors reported by CI in 0d0991c12

src/dst/src/BalancerConnection.cc
```diff
@@ -236,3 +236,14 @@ void BalancerConnection::handle_read(boost::system::error_code const& err,
     sock_.close();
   }
 }
+
+#if !SWIG && FMT_VERSION >= 100000
+namespace boost::asio::ip {
+
+inline static auto format_as(const boost::asio::ip::address f)
```
maliberty, line 243: The use of inline in a cpp is unnecessary.  Same in src/dst/src/LoadBalancer.cc

maliberty:
Thanks for the contribution!

PR #4206, opened 2023-11-02
mpl2: fix macro blockage penalty

AcKoucher:
Fix #4780 

This PR is to:
1) Change the way we compute the penalty for overlapping between clusters with macros and macro blockages during SA for SoftMacros. We now just compute the overlapping area between the `SoftMacro` and the `macro_blockage`.
2) Adapt the calculation of cluster placement penalties (SoftSA) to the fixed sequence pair.
3) Ensure we actually use the new sequence pair in when running enhanced macro placement function.

_For documentation: Not only were we using an overly complex expression to compute the penalty, but there was also a bug: even when there was no overlap, the computed `macro_blockage_penalty` between a `SoftMacro` and a `macro_blockage` was different from 0, because, even if the ratio of one of the coordinates was 0 the other one was not. This was due to way std::max was being used in the ratio of each coordinate._

_To keep the original computing, we would have to ensure that, if there was no overlap between a `SoftMacro` and a`macro_blockage` in any of the coordinates, the penalty would be 0._

PR #4354, opened 2023-12-06
mpl2: improve annealing and debugger

AcKoucher:
When running CI for #4206, gf12/ariane failed in hierarchical macro placement for a specific macro cluster. Some debugging showed that some good results generated when SA temperature is high are being wasted. This PR is to make sure we keep the  lowest cost valid result.

Also, we now know in which annealing step and perturb we are when debugging.

Some small rearrangement and cleaning as well.


AcKoucher:
Running Secure-CI.

maliberty (approved):
The code looks fine but I'll hold the merge until the CI results are in

AcKoucher:
CI is showing some failures, but I'll way to dig until #4357 is merged. I think it can have some meaningful impact.

AcKoucher:
Running Secure-CI with updates.

maliberty:
@AcKoucher what the status of this?

AcKoucher:
@maliberty I'm addressing the question in #4206, as it will be easier to not split the answer.

PR #4357, opened 2023-12-07
mpl2: fix weights variations in macro placement

AcKoucher:
_CI for #4339 showed that results were still varying depending on the number of threads.
Testing the changes in this PR with the test case showed consistency across results._

Weights for macro placement don't depend on the number of threads anymore.

Running Secure-CI..


AcKoucher:
Secure-CI passed.

AcKoucher:
@maliberty FYI, I as was testing this change, I realized that a bunch of situations where we were seeing some weird orientations could be happening because of this bug.

PR #4360, opened 2023-12-08
mpl2: ensure we don't get stuck in cluster placement

### LINKED ISSUE #4335: rtlmp should complete in a reasonable amount of time or fail quickly if it is hopeless
### Description

Never completes...

To reproduce(I suppose it should be possible to make a small text only based unit-test that reproduces the problem, but I don't know how to):

unzip https://drive.google.com/file/d/1e2ebRJ80NInmfF3xhgY96wOHQsz_7Czc/view?usp=sharing

```
./run-me-BoomTile-asap7-base.sh 
OpenROAD v2.0-11160-g042cdcfa4 
[INFO ORD-0030] Using 16 thread(s).
HierRTLMP Flow enabled...
Call Macro Placer  -halo_width 20 -report_directory bazel-bin/build//objects/asap7/BoomTile/base/rtlmp
Floorplan Outline: (0.0, 0.0) (4740.732, 4740.732),  Core Outline: (2.052, 2.16) (4738.716, 4738.5)
Traversed logical hierarchy
	Number of std cell instances: 2130872
	Area of std cell instances: 273473.66
	Number of macros: 7
	Area of macros: 4194455.50
	Area of macros with halos: 4534412.00
	Area of std cell instances + Area of macros: 4467929.00
	Core area: 22434450.00
	Design Utilization: 0.20
	Core Utilization: 0.01
	Manufacturing Grid: 1

[INFO MPL-0024] [Multilevel Autoclustering] Creating clustered netlist.
[INFO MPL-0039] [Coarse Shaping] Determining shape functions for clusters.
[INFO MPL-0028] [Hierarchical Macro Placement] Placing clusters and macros.
[no further output]
```

### Suggested Solution

Actionable error message if it is hopeless(what is wrong and what needs to be fixed) or complete in a practical amount of time. At most hours, not days.

Also, it would be great to have a write_macro_placement command so as to store the macro placement. Once the floorplan has been identified, there may not be a reason and also reasons not to recompute it. Also writing out the macro placement allows for manual tinkering afterewards.

### Additional Context

_No response_

AcKoucher:
Fix #4335 

We were getting stuck in when synchronizing the threads results.

Running Secure-CI


src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -3781,11 +3782,12 @@ void HierRTLMP::multiLevelMacroPlacement(Cluster* parent)
         }
         ++begin_check;
       }
-      // add early stop mechanism
-      if (best_sa) {
+      const int remaining_checks = total_runs - begin_check;
+
+      if (best_sa || remaining_checks == 0) {
```
maliberty, line 3787: if remaining_checks==0 wouldn't remaining_runs ==0 too?

AcKoucher: I think I see your point, but the problem is the last iteration of checks.

I had a hard time writing a proper explanation, so I'll go with a practical example:

In a situation with
`Total_runs = 12`
`Threads = 8`
`Check Interval = 10`

After we end executing the first batch of 8 runs, we won't enter the check loop, because we haven't finished 10 runs yet.
..
So we execute the last 4 runs and by the time we get to the loop, `remaining_runs == 0`, so we'll check the first 10 runs and break the loop without checking the last 2. That's why we need `remaining_checks`.

maliberty: The goal is to check every 10 runs.  If we find a solution then we should ignore the remaining two runs.  Otherwise we would get a different result if we used 10 threads.

AcKoucher: Oh.. I'm sorry, indeed. I got confused with the checking. I'm fixing it.

AcKoucher:
Secure-CI passed.

maliberty (approved):

PR #4361, opened 2023-12-09
Adds functionality to the Python API to check if a pin is a start point

LukeChristianMartin:


include/ord/Design.h
```diff
@@ -175,6 +189,32 @@ class Design
   bool isSequential(odb::dbMaster* master);
   bool isBuffer(odb::dbMaster* master);
   bool isInverter(odb::dbMaster* master);
+  bool isInPower(odb::dbITerm* iterm);
+  bool isInGround(odb::dbITerm* iterm);
+  sta::ClockSeq findClocksMatching(const char *pattern, bool regexp, bool nocase);
+  sta::Clock* defaultArrivalClock();
+  std::string getITermName (odb::dbITerm* ITerm);
```
maliberty, line 196: Design is starting to become the timing API which I think isn't a good idea.  That will be two large APIs lumped into one. Would you split out a timing API and move the timing relevant methods?

src/Design.cc
```diff
@@ -98,6 +98,249 @@ void Design::readDef(const std::string& file_name,
                child);
 }
 
+
+
+odb::dbITerm* Design::staToDBPin(const sta::Pin *pin) {
```
maliberty, line 103: This isn't a good API as iterm may be null - a Pin is either an iterm or a bterm not always an iterm.

src/Design.cc
```diff
@@ -98,6 +98,249 @@ void Design::readDef(const std::string& file_name,
...
+    // This will need to be edited
+    sta::PinSet endPoints = findEndpoints();
+
+    for (const auto& endpoint : endPoints) {
+
+        odb::dbITerm* dbITermPin = staToDBPin(endpoint);
+
+        if (dbITermPin == nullptr) {
+            continue;
+        }
```
maliberty, line 140: This ignores all endpoints the are bterms?

src/Design.cc
```diff
@@ -98,6 +98,249 @@ void Design::readDef(const std::string& file_name,
...
+  sta::PinSet pins(network);
+  sta::VertexPinCollector visitor(pins);
+  sta::Sta::sta()->visitEndpoints(&visitor);
+  return pins;
+}
+
+std::vector<std::string> Design::extractEndpointNames() {
+    std::vector<std::string> endpointNames;
+
+    // This will need to be edited
```
maliberty, line 131: ??

src/Design.cc
```diff
@@ -98,6 +98,249 @@ void Design::readDef(const std::string& file_name,
...
+
+sta::PinSet Design::findEndpoints() {
+  sta::Network *network = sta::Sta::sta()->cmdNetwork();;
+  sta::PinSet pins(network);
+  sta::VertexPinCollector visitor(pins);
+  sta::Sta::sta()->visitEndpoints(&visitor);
+  return pins;
+}
+
+std::vector<std::string> Design::extractEndpointNames() {
```
maliberty, line 128: Why does this need to be a method?  It appears to be written in terms of public methods

src/Design.cc
```diff
@@ -98,6 +98,249 @@ void Design::readDef(const std::string& file_name,
...
+        }
+
+        endpointNames.push_back(getITermName(dbITermPin));
+    }
+
+    return endpointNames;
+}
+
+bool Design::isPinEndpoint(const std::vector<std::string>& endpointNames, odb::dbITerm* db_pin) {
+    std::string pinName = getITermName(db_pin);
```
maliberty, line 149: Please avoid string based APIs

src/Design.cc
```diff
@@ -98,6 +98,249 @@ void Design::readDef(const std::string& file_name,
...
+
+
+float Design::getPinSlew(odb::dbITerm* db_pin) {
+    const int num_vertex_elements = 2;
+
+    sta::dbSta* sta = getSta();
+
+    sta::Pin* sta_pin = sta->getDbNetwork()->dbToSta(db_pin);
+
+    std::array<sta::Vertex*, 2> vertex_array = vertices(sta_pin);
```
maliberty, line 216: replace 2 with num_vertex_elements

src/Design.cc
```diff
@@ -98,6 +98,249 @@ void Design::readDef(const std::string& file_name,
...
+
+    float pinSlew = -sta::INF;
+    for (int i = 0; i < num_vertex_elements; i++) {
+      sta::Vertex* vertex = vertex_array[i];
+      if(vertex != nullptr){
+        float pinSlewTemp = slew_corner(vertex);
```
maliberty, line 222: const

maliberty (changes requested):

PR #4363, opened 2023-12-10
Fixes the distance between two coordinates bug in DPL

mina1460:


src/dpl/src/Opendp.cpp
```diff
@@ -478,9 +478,28 @@ int Opendp::gridPaddedWidth(const Cell* cell) const
...
+    int remaining_rows = y_coordinate % grid_sites.size();
+    int height = patterns_below * total_height
+                 + remaining_rows
+                       * std::accumulate(
+                           grid_sites.begin(),
+                           grid_sites.begin() + remaining_rows,
+                           0,
+                           [](int sum, const dbSite::OrientedSite& entry) {
+                             return sum + entry.site->getHeight();
+                           });
```
rovinski, line 499: I would break up the expression for better readability.

maliberty:
please run a secure CI

maliberty (approved):

PR #4370, opened 2023-12-12
Reduce the number of copies and destructions of std::shared_ptr.

rmlarsen:
This is done by making read-only copies and iterators const-ref. The time spent incrementing and decrementing the refcounts drops by about 6x when running the OpenROAD Gate Resize tool on a circuit we care about, as shown by the profile snippets below.

Botttom-up profile view before:
image

Bottom-up profile view after:
image

 [Notice that the "before" profile corresponds to a slightly older snapshot where `slackPenalized` has not been updated to take a const-ref argument, so this change probably reduces the cost by at most 5x.]


maliberty:
Please follow the 'Details' link by DCO for steps to fix the missing DCO.

maliberty (approved):

rmlarsen:
DCO fixed. Thanks for the pointer @maliberty

PR #4379, opened 2023-12-13
Speed up initLUT in flute.cpp by ~14x

rmlarsen:
The speedup is accomplished by adding a function for parsing decimal integers instead of calling `sscanf(...,"%d%n",...)`.

Flame graph before:
image

Flame graph after:
image


maliberty:
The screenshot links are inaccessible

maliberty:
There are multiple test failures that need addressing.

rmlarsen:
Screenshot links fixed. Will look at test failures later today.

rmlarsen:
The comment above the helper function is also wrong now... Ooops!

maliberty:
Is it really necessary to write a low level parser?  Would atoi suffice?

rmlarsen:
@maliberty atoi wouldn't work since it doesn't return the number of characters processed, but `std::stoi` would. Let me give it a try.

rmlarsen:
@maliberty sadly, `stoi` is _much_ slower, even slower than `sscanf` on my system on compiler (Skylake + Clang)

rmlarsen:
@maliberty the tests passed now. I just had a silly cut-and-paste error.

maliberty:
how about  std::from_chars ?

maliberty:
"This is intended to allow the fastest possible implementation that is useful in common high-throughput contexts such as text-based interchange" from https://en.cppreference.com/w/cpp/utility/from_chars

rmlarsen:
@maliberty that is not the right API either as it assumes you've already scanned to find the end point.

rmlarsen:
@maliberty FWIW, at Google we use the abseill library, which has a lot of handy optimized functions for various types of string processing. Unfortunately, the standard library does not seem to provide anything near the same performance.

QuantamHD:
@maliberty I think this is ready to go.

maliberty (approved):

maliberty:
Its a bit sad that there is no standard solution to this ordinary need

rovinski:
@maliberty wasn't there some effort a few years ago to stitch the Flute LUTs directly into the binary so  it didn't have to be loaded? Further, it seems more like the main problem is not caching the files after the first initialization?

maliberty:
@rovinski the file is stored as a string in OR but it is still parsed to 'load'.  It should only be parsed once - do you see something to suggest otherwise?

rovinski:
So according to the flame graph, this one-time init is about 30% of the whole GRT time? Why is it even stored as a string that needs to be parsed?

maliberty:
I suppose you mean gpl time.  @rmlarsen what design is this from?  I'm guessing it is quite small.

rmlarsen:
@maliberty @rovinski Yes, this is a relatively small circuit, but it is processed many times, so this startup time matters. Ideally this should only be done once, but it would require refactoring the tool chain.

rovinski:
I see, it's probably not worth the effort beyond this PR then.

PR #4386, opened 2023-12-14
Replace Diamond search with a BFS

mina1460:
This PR is a step forward to enable us to do two things:
1- have a readable version of the existing cryptic diamond search that is potentially parallelizable.
2- Allow us to fix this [bug](https://drive.google.com/file/d/1jy9h0fuS4szo6ZI-EMvxIVwy4EIjII3u/view) by introducing a priority queue based on the Manhattan distance instead of the order of traversal. 

Notes for code reviewers:
- the hashset of the visited points can be replaced by a normal tree-based set, but hashmaps with simple hash functions can outperform the log search for large trees, which is the case in this scenario.


QuantamHD:
Looks nice! Does the change to BFS affect the PPA?

src/dpl/src/Place.cpp
```diff
@@ -796,42 +809,37 @@ PixelPt Opendp::diamondSearch(const Cell* cell,
     return avail_pt;
   }
 
-  for (int i = 1; i < std::max(scaled_max_displacement_y_, max_displacement_x_);
-       i++) {
+  std::queue<std::pair<int, int>> positionsQueue;
```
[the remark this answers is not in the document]
mina1460: yeah, good "point" xD
I will update the code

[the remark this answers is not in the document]
mina1460: what does PPA mean @QuantamHD ?

src/dpl/src/Place.cpp
```diff
@@ -841,14 +836,29 @@ PixelPt Opendp::diamondSearch(const Cell* cell,
...
+        }
+
+        // Mark the current position as visited
+        visitedPositions.insert({x_offset, y_offset});
+
+        // Enqueue neighboring positions for exploration
+        positionsQueue.push({x_offset - 1, y_offset});
+        positionsQueue.push({x_offset + 1, y_offset});
+        positionsQueue.push({x_offset, y_offset - 1});
+        positionsQueue.push({x_offset, y_offset + 1});
```
maliberty, line 854: rather than queue and later skipping you could keep a collection of enqued points and skip adding redundant ones here.  Checking should be cheaper once using a heap.

maliberty:
is "which is the case in this scenario" based on measurement?

maliberty:
@QuantamHD this PR right now doesn't affect PPA (power, performance, area) but once we switch to a priority queue it will - hopefully for the better.  That change will be made in this PR, this review was just for an initial look.

ApeachM:
Hi, Mina. I have some question for making sure my understanding.

1. Is it correct that the reason for changing the existing diamond search to BFS is to make the code more readable?
2. Are `x_offset` and `y_offset` grid coordinates rather than Manhattan distances? If then, I don't think this could solve the bug that you mentioned. Would my thougt be correct?

mina1460:
Hi @ApeachM 
Sorry, I think I was not clear in the PR description. 
1- Yes you are 100% correct. This PR helps make diamond search easier to maintain and understand.
2- You are also correct. As it stands now, using the queue alone, doesn't solve the bug you discovered. But, it makes it much easier to fix when we move the queue to a priority queue (min heap) where the ordering of the elements in the heap is based on their manhattan distance. This will solve the bug in the gif in the PR description

ApeachM:
Thank you for your answer.

Maybe we need to expand the search space as much as the aspect ratio, and collect the points and sort them by the distances in the future work. 

This is not by the BFS, but diamond search, but you may refer to the code.
This code searches following GIF, when aspect ratio is 5 which is extreme case.

Thank you for your work!

src/dpl/include/dpl/Opendp.h
```diff
@@ -617,7 +617,7 @@ class Opendp
   std::unique_ptr<DplObserver> debug_observer_;
 
   // Magic numbers
-  static constexpr int bin_search_width_ = 10;
+  static constexpr int bin_search_width_ = 2;
```
maliberty, line 620: Why 2 ?

src/dpl/src/Place.cpp
```diff
@@ -796,58 +793,97 @@ PixelPt Opendp::diamondSearch(const Cell* cell,
...
+                        cur_dist);
+
+      if (cur_pt.pixel && cur_dist < best_dist) {
+        best_pt = cur_pt;
+        best_dist = cur_dist;
+        best_for = 0;
+      } else if (best_dist < cur_dist) {
+        ++best_for;
       }
+      if (best_for >= 7) {
```
maliberty, line 853: What is 7?

src/dpl/include/dpl/Opendp.h
mina1460: this was the bug that made the PQ approach not work as expected. The way it was used is that after a bin is identified from the old diamond search, there was a for loop that checked 10 bins to the left and right of it, and if any of them were available, it returned them.

there could have been an empty bin to the top at a shorter Manhattan distance that would not be discovered because the 9th bin on the row below it was available.

2 or 1 should work, but I experimented with all numbers from 1 to 10, and 2 was slightly better than 1 always (and a tiny bit faster).

src/dpl/src/Place.cpp
mina1460: 7 is a magic number that I found from experimenting. When you find an available point, you could return immediately or you can wait until you are sure that it was indeed the best after you do some more search.

7 means that your point remained the best you could find for 7 additional searches. any number would work, but numbers below 7 result in worse metrics, and above 7 doesn't improve.

of course it depends on the test case, but remember that everytime you find a better point, you still need to make sure it will remain the "best for" X additional searches, so you reset the counter.

Larger numbers slow down runtime with no additional gains

rovinski: If it's a magic number then it should probably be changed to a self-documenting variable and included with the others as `static constexpr`

maliberty: How would you get a better result later - aren't we expanding outwards?

mariam606:
@maliberty 
Hello, I spoke with Mina and I am interested in taking this up. I agreed with him and he will guide me through it. Can I do so?
Thanks in advance!

maliberty:
That's sounds great - I haven't had time to investigate the failures in this PR.

maliberty:
replaced by #6693

PR #4388, opened 2023-12-15
odb codegen: adds support for C++ enum classes

fgaray:


src/odb/src/codeGenerator/templates/db.h
```diff
@@ -22,7 +22,11 @@ class {{klass.name}} : public dbObject
   {% endfor %}
   {% for _enum in klass.enums %}
     {% if _enum.public %}
-    enum {{ _enum.name }}{% if "type" in _enum %} :{{ _enum.type }}{% endif %}
+      {% if _enum.class %}
+        enum class {{ _enum.name }}{% if "type" in _enum %} :{{ _enum.type }}{% endif %}
```
maliberty, line 26: Would
```
enum {{ class if _enum.class }} {{ _enum.name ...
```
work?  It might reduce the duplication.

fgaray: I changed this to one line with a different syntax, wdyt? Thanks!

maliberty (approved):

src/odb/include/odb/db.h
```diff
@@ -7634,7 +7634,7 @@ class dbScanChain : public dbObject
 class dbScanInst : public dbObject
 {
  public:
-  enum SCAN_INST_TYPE
+  enum class SCAN_INST_TYPE
```
maliberty, line 7637: Fwiw I find dbScanInst::SCAN_INST_TYPE::.. rather redundant.

fgaray: Totally agree, I will address this in a future PR

Issue #4347, opened 2023-12-05
Add `tclint` to OpenROAD CI


nmoroze:
### Description

This issue is a follow-up from discussion on #4337, and tracks adding [`tclint`](https://github.com/nmoroze/tclint.git) to the OpenROAD CI. 

`tclint` is a lint tool for Tcl that checks both style and functional errors. I've done some early experiments on this codebase, and with a baseline config there are ~23,000 lint violations across all the Tcl scripts. Nearly all of these are style-related. 

I think the best strategy would be to add linting incrementally to certain parts of the codebase. This will also give the maintainers a chance to review a reasonably-sized initial PR to make sure the style enforced by `tclint` is to their liking.

Do any maintainers have a preference for a chunk of the Tcl codebase that might be a good candidate for the first linting pass? If given a pointer somewhere I'll open a PR that adds linting to CI for that part of the code and does the cleanup required to get it to pass.

maliberty:
How about src/rsz/src/Resizer.tcl as a sample

rovinski:
Should create a mirrored issue for ORFS.

On that note, it might be good to create separate issues for a Python linter as well.

@nmoroze are you planning on enhancing it reformat files similar to `clang-format -i`? That would be super useful.

nmoroze:
> Should create a mirrored issue for ORFS.

Done: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1686

> are you planning on enhancing it reformat files similar to clang-format -i? That would be super useful.

yeah, this was on my medium/long-term roadmap but I can consider prioritizing it, since it definitely seems useful for getting lint going on a large existing codebase like this one. At the very least, I can do some early experiments to see how hard it will be. I think `tclint` has most of the key pieces already, and "just" needs an emitter for the syntax tree to spit out well-formatted Tcl scripts. I tried to be mindful of this use case when designing the parse tree data structure to make sure it captures enough information, the devil's in the details and I have a feeling there will be some tricky bits :) 

I opened an issue to track this: https://github.com/nmoroze/tclint/issues/8

nmoroze:
I've submitted the first PR (#4397) towards this effort, cleaning up `Resizer.tcl` as suggested. 

One facilitating change to `tclint` was adding an option to not require an extra level of indentation within `namespace eval` blocks - this seems to be a common pattern in the OpenROAD code, and it seems to halve the overall number of lint violations found in the codebase.

[...]

PR #4397, opened 2023-12-16
Add `tclint` to CI and clean `Resizer.tcl`

nmoroze:
Towards #4347, this PR adds `tclint` to the CI (with a hardcoded file-list) and cleans up the lint warnings in `Resizer.tcl`, which was the first file chosen as a test.

### DIFF
--- a/.github/workflows/github-actions-lint-tcl.yml
+++ b/.github/workflows/github-actions-lint-tcl.yml
@@ -1,0 +1,21 @@
+name: Lint Tcl code
+
+on:
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install tclint
+        run: |
+          python -m pip install tclint==0.1.6
+
+      # TODO: hardcoded list of files while we incrementally lint codebase
+      # See issue #4347 for tracking
+      - name: Lint
+        run: |
+          tclint src/rsz/src/Resizer.tcl
--- a/src/rsz/src/Resizer.tcl
+++ b/src/rsz/src/Resizer.tcl
@@ -156,12 +156,6 @@ proc set_wire_rc { args } {
       set wire_res [rsz::layer_resistance $layer $corner]
       set wire_cap [rsz::layer_capacitance $layer $corner]
     }
-
-    # Unfortunately this does not work very well with technologies like sky130
-    # that use inappropriate kohm/pf units.
-    if { 0 } {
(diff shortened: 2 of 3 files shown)


src/rsz/src/Resizer.tcl
```diff
@@ -156,12 +156,6 @@ proc set_wire_rc { args } {
       set wire_res [rsz::layer_resistance $layer $corner]
       set wire_cap [rsz::layer_capacitance $layer $corner]
     }
-
-    # Unfortunately this does not work very well with technologies like sky130
-    # that use inappropriate kohm/pf units.
-    if { 0 } {
-      utl::info RSZ 61 "$signal_clk wire resistance [sta::format_resistance [expr $wire_res * 1e-6] 6] [sta::unit_scale_abbreviation resistance][sta::unit_suffix resistance]/um capacitance [sta::format_capacitance [expr $wire_cap * 1e-6] 6] [sta::unit_scale_abbreviation capacitance][sta::unit_suffix capacitance]/um."
-    }
```
nmoroze, line 164: `tclint` flagged this line for a length violation, I made the call to delete it wholesale myself (since it's dead code). If you'd prefer to keep it, I can change the commit to just fix the line length violation instead.

src/rsz/src/Resizer.tcl
```diff
@@ -612,7 +606,8 @@ proc check_corner_wire_caps {} {
   set have_rc 1
   foreach corner [sta::corners] {
     if { [rsz::wire_signal_capacitance $corner] == 0.0 } {
-      utl::warn RSZ 14 "wire capacitance for corner [$corner name] is zero. Use the set_wire_rc command to set wire resistance and capacitance."
+      utl::warn RSZ 14 "wire capacitance for corner [$corner name] is zero.\
+        Use the set_wire_rc command to set wire resistance and capacitance."
```
maliberty, line 610: Won't this introduce extra whitespace into the message?

.github/workflows/github-actions-lint-tcl.yml
maliberty, line 1: @vvbandeira any concerns on this new action?

src/rsz/src/Resizer.tcl
nmoroze: The `\` character will actually replace the newline and all leading whitespace on the next line with just a single space, so it should be the same as before. As an example you can run in `tclsh`, this should just print `Hello World`: 

```tcl
puts "Hello\
  World"
```

Happy to tweak this if you have a different preference on how to do it, though!

maliberty (approved):

.github/workflows/github-actions-lint-tcl.yml
vvbandeira: No, looks simple enough.

PR #4399, opened 2023-12-18
meta: Moves GPL tests to new CTest Runner

QuantamHD:
You can now run the GPL tests in parallel with the ctest command.

`ctest -j 40`

Integration tests can be excluded with.

`ctest -j 40 -LE IntegrationTest`

### DIFF
--- a/src/gpl/test/CMakeLists.txt
+++ b/src/gpl/test/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_program (BASH_PROGRAM bash)
+find_program (BASH_PROGRAM bash REQUIRED)
 
 function(gpl_test test_name)
   add_test (
@@ -38,8 +38,6 @@ clust01
 clust02
 )
 
-if (BASH_PROGRAM)
-  foreach(TEST_NAME IN LISTS TEST_NAMES)
-    gpl_test(${TEST_NAME})
-  endforeach()
-endif (BASH_PROGRAM)
+foreach(TEST_NAME IN LISTS TEST_NAMES)
+  gpl_test(${TEST_NAME})
+endforeach()
--- a/src/gpl/test/CMakeLists.txt
+++ b/src/gpl/test/CMakeLists.txt
@@ -1,0 +1,45 @@
+find_program (BASH_PROGRAM bash)
+
+function(gpl_test test_name)
+  add_test (
+    NAME gpl.${test_name} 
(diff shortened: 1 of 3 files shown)


src/gpl/test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+error01
+diverge01
+density01
+convergence01
+nograd01
+clust01
+clust02
+)
+
+if (BASH_PROGRAM)
```
maliberty, line 41: If this 'if' fails then will no tests will be run?  That should be an error and not a silent skip.  With REQUIRED above I don't think this if is needed.

src/gpl/test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+find_program (BASH_PROGRAM bash)
```
maliberty, line 1: REQUIRED ?

src/gpl/test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+function(gpl_test test_name)
+  add_test (
+    NAME gpl.${test_name} 
+    COMMAND ${BASH_PROGRAM} ${CMAKE_CURRENT_SOURCE_DIR}/regression ${test_name}
+  )
+  set_tests_properties(gpl.${test_name}  PROPERTIES LABELS "IntegrationTest")
+endfunction()
+
+set(TEST_NAMES
```
maliberty, line 11: If this PR is merged, will these test runs twice or not at all in the current GH action?

QuantamHD: Not at all, tests labeled IntegrationTest are not run in the github action

maliberty: So no tests are run by this change by default?

QuantamHD: Correct, no change to the CI. The only change is that if you run `ctest` manually these integration tests will now run by default.

They can be excluded with the `-LE` flag as mentioned above, but the default `ninja build_and_test` rule excludes them by default which means they won't run in the CI.

maliberty: Is the goal to eventually convert over? Is the diffs file overwritten when these run in parallel?

QuantamHD: The goal is eventually to convert over. The diffs are overwritten in parallel mode. We probably want to switch over to writing them to stdout instead of a file.

maliberty (approved):

PR #4410, opened 2023-12-19
mpl2: refactoring of the top level function & removing more messages

### LINKED ISSUE #4394: Convert more mpl2 messages to debug output
### Description

Debug output... all of this is information already contained in previous output.

```
[INFO MPL-0024] [Multilevel Autoclustering] Creating clustered netlist.
[INFO MPL-0039] [Coarse Shaping] Determining shape functions for clusters.
[INFO MPL-0028] [Hierarchical Macro Placement] Placing clusters and macros.
[INFO MPL-0037] Updated location of 95 macros
Delete buffers for RTLMP flow...
[INFO RSZ-0026] Removed 0 buffers.
```

### Suggested Solution

Convert to debug output...

### Additional Context

_No response_

AcKoucher:
Resolve #4394

My intention with the names of each stage method (e. g., runCoarseShaping) is to make it easier if we want split the HierRTLMP class into different classes for each engine at some point.

With this new top level interface it became easy to transform the root into a HardMacroCluster and get rid of that weird check for the macro only case, so I also did that.


AcKoucher:
Running Secure-CI.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -329,28 +331,82 @@ void HierRTLMP::setDefaultThresholds()
...
+// 3) Fine Shaping -> Top - Down:
+//      Refine the possible shapes of each cluster based on the fixed
+//      outline and location of its parent cluster.
+//      *This is executed within hierarchical placement method.
+// 4) Hierarchical Macro Placement -> Top - Down
+//      a) Placement of Cluster (one level at a time);
+//      b) Placement of Macros (one macro cluster at a time).
+// 5) Orientation Improvement
+//      Attempts macro flipping to improve WR
+void HierRTLMP::run()
```
maliberty, line 366: Much nicer!

maliberty:
Waiting to merge based on CI results

AcKoucher:
Secure-CI showed some increase in the DRCs reported by Calibre which seems to be on master as well.

PR #4413, opened 2023-12-19
Grt: Fixing overlapping edges in net

luis201420:
Fixing overlapping edges generated after overflow iterations


src/grt/src/fastroute/src/maze.cpp
```diff
@@ -769,170 +769,150 @@ void FastRouteCore::setupHeap(const int netID,
...
-          // add the neighbor of cur node into queue
-          queue[queuetail] = nbr;
-          queuetail++;
-        }  // loop i (3 neigbors for cur node)
-      }    // while queue is not empty
-    }      // else n1 is not a Pin node
+        // add the neighbor of cur node into queue
+        queue[queuetail] = nbr;
+        queuetail++;
+      }  // loop i (3 neigbors for cur node)
```
eder-matheus, line 843: Typo: neigbors --> neighbor

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -1327,6 +1307,116 @@ float getCost(const int i,
   return cost;
 }
 
+int FastRouteCore::splitEdge(std::vector<TreeEdge>& treeedges,
+                             std::vector<TreeNode>& treenodes,
+                             int n1,
+                             int n2,
+                             int edge_n1n2)
```
eder-matheus, line 1314: Add const int for these parameters.

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -1327,6 +1307,116 @@ float getCost(const int i,
...
+  new_node.x = n2x;
+  new_node.y = n2y;
+  new_node.stackAlias = treenodes[n2].stackAlias;
+
+  // create new edge
+  const int new_edge_id = treeedges.size();
+  TreeEdge new_edge;
+
+  // find one neighbor node different to n1
+  int B1;
```
eder-matheus, line 1331: What B1 is representing? Can it have a clearer name?

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -1327,6 +1307,116 @@ float getCost(const int i,
...
+  if (treenodes[n2].nbr[0] == n1) {
+    B1 = treenodes[n2].nbr[1];
+    edge_n2B1 = treenodes[n2].edge[1];
+  } else if (treenodes[n2].nbr[1] == n1) {
+    B1 = treenodes[n2].nbr[0];
+    edge_n2B1 = treenodes[n2].edge[0];
+  } else {
+    B1 = treenodes[n2].nbr[0];
+    edge_n2B1 = treenodes[n2].edge[0];
+  }
```
eder-matheus, line 1342: Why the `else if` and the `else` do the same thing? Can they be merged into a single `else`, or should they do different things?

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -1327,6 +1307,116 @@ float getCost(const int i,
...
+  new_edge.route.gridsY.push_back(n2y);
+
+  // config new node
+  new_node.nbr_count = 3;
+  new_node.nbr[0] = B1;
+  new_node.nbr[1] = n2;
+  new_node.nbr[2] = n1;
+  new_node.edge[0] = edge_n2B1;
+  new_node.edge[1] = new_edge_id;
+  new_node.edge[2] = edge_n1n2;
```
eder-matheus, line 1412: Will the new node you're creating always have three neighbors and three edges?

eder-matheus (changes requested):

src/grt/src/fastroute/src/maze.cpp
luis201420: Yes, it will always have 3 neighboring nodes because I will place it in the middle of node n2 and one of its neighbors. Then it will have as neighbors the node n,2, the node which connects that edge n1 and the neighbor node to n2.
Node n2 will always have a neighbor apart from n1 (neighbors > 1), since if it did not have one then its position does not change and it would not be used in this function.

eder-matheus (approved):
The code looks good, but I'm investigating some issues in the CI. We have a non-determinism problem unrelated to this PR, which is currently blocking it from passing. I'll hold the merge until I fix the problem.

PR #4419, opened 2023-12-20
Min area patching bugfix

refaay:


src/drt/src/gc/FlexGC_main.cpp
```diff
@@ -2020,10 +2019,9 @@ void FlexGCWorker::Impl::checkMetalShape_addPatch(gcPin* pin, int min_area)
     patchBx.set_ylo(-chosenEdg->length() / 2);
     patchBx.set_yhi(chosenEdg->length() / 2);
     offset.setY((chosenEdg->low().y() + chosenEdg->high().y()) / 2);
-    if (chosenEdg->getOuterDir() == frDirEnum::E) {
+    if (chosenEdg->getOuterDir() == frDirEnum::E
+        || chosenEdg->getOuterDir() == frDirEnum::W) {
       offset.setX(chosenEdg->low().x());
-    } else if (chosenEdg->getOuterDir() == frDirEnum::W) {
-      offset.setX(chosenEdg->low().x() - length);
     } else
```
osamahammad21, line 2025: How does this work? you didn't do any modification to the patchBx to account for the new implementation that the offset is to the right of the box. I think this will lead to getBBox to fail because it moves the offset_box right to the offset point.

osamahammad21 (changes requested):

PR #4428, opened 2023-12-21
fix copy const std::pair<int, int> in for loop

fgaray:


fgaray:
Seems to be some infra issue 
image

Also, seeing the same in #4429 

Is there any way to trigger again the tests?

QuantamHD:
Just push a new commit with maybe a comment change, or reword the commit and force push

fgaray:
I tried again but seems to be happening again. Looks like it is affecting other PRs too: #4413, #4396, #4427

maliberty:
@vvbandeira is looking into it now

maliberty:
The pair is destructured so I don't see a copy happening.

fgaray:
Thanks @vvbandeira !

@maliberty I think you may be right, I tried looking at the output of some toy example (https://godbolt.org/z/eob4jPv3f) and  I can't see a difference in the generated assembly for this case, even with optimizations disabled.

However, I see a [warning](https://github.com/The-OpenROAD-Project/OpenROAD/assets/1619948/71334615-caba-41ba-a20b-03ade4328ccf) in gcc 13.2 (and in my local copy of OpenROAD) for this for loop. Maybe is it a false positive? 

Anyway, it doesn't seem too bad to add the reference here, even if it is just to prevent the warning.

maliberty:
weird but not harmful

maliberty (approved):

PR #4429, opened 2023-12-21
odb: adding support to serialize and de-serialize std::unordered_map

fgaray:
This is based on the already existing case for std::map

### DIFF
--- a/src/odb/include/odb/dbStream.h
+++ b/src/odb/include/odb/dbStream.h
@@ -38,6 +38,7 @@
 #include <istream>
 #include <ostream>
 #include <string>
+#include <unordered_map>
 
 #include "ZException.h"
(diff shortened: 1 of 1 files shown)


maliberty:
We already have dbHashTable that supports serialization.  Is there a reason to add unordered_map?

fgaray:
So, I was looking at dbHashTable.h and dbHashTable.hpp and it seems to be a custom implementation of a hash table, I would say that it is better to just use the standard implementations and reduce the surface for possible bugs, **unless they have some feature that would be useful when dealing with db objects.**

We already have std::map in the serialization code, so there is at least some use case for the standard containers.

Would you recommend me to use dbHashTable instead?

maliberty (approved):

PR #4444, opened 2023-12-25
mpl2: add array support

AcKoucher:
This draft is to show a first approach.

The idea is to create a classification based on direct connection between single-macro macro clusters && to prioritize array macro cluster grouping over connection signature grouping.

Important observation: For this initial draft I'm considering only direct connection (Not based on hops);

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -2253,7 +2253,7 @@ void HierRTLMP::breakMixedLeafCluster(Cluster* root_cluster)
     // Split by replacement if macro dominated
     if (mixed_leaf->getNumStdCell() * macro_dominated_cluster_threshold_
         < mixed_leaf->getNumMacro()) {
-      parent = mixed_leaf->getParent();  // replacement
+      parent = mixed_leaf->getParent();
     }
 
     mapMacroInCluster2HardMacro(mixed_leaf);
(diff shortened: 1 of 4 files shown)


src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -2231,12 +2231,12 @@ void HierRTLMP::breakMixedLeafCluster(Cluster* root_cluster)
...
-  std::vector<Cluster*> leaf_clusters;
+  std::vector<Cluster*> mixed_leaves;
   for (auto& child : root_cluster->getChildren()) {
     setInstProperty(child);
     if (child->getNumMacro() > 0) {
       if (child->getChildren().empty()) {
-        leaf_clusters.push_back(child);
+        mixed_leaves.push_back(child);
       } else {
         breakMixedLeafCluster(child);
```
maliberty, line 2241: The terminology is getting confusing.  You have 'mixed_leaves' which are leaf cluster but the else has breakMixedLeafCluster which are not leaves?

AcKoucher: As far as I understand, the function breakMixedLeafCluster(Cluster* root_cluster) traverses the tree searching for mixed clusters. - the argument should be (Cluster* parent). I'll fix this. - If a parent has mixed leaves, it pushes them into a vector and break them into stdcell clusters and hard macro clusters.

I was thinking of something like
findAndBreakMixedLeavesIn(Cluster* parent) 
or
separateStdCellAndMacroClusters(Cluster* parent)

What do you think?

maliberty: If we take the positive side of the 'if' would we have a cluster that is not been separated?  Shouldn't all clusters get separated?

AcKoucher: Entering the if scope means we found a mixed leaf in the parent's children, so we push it to the mixed_leaves vector.

The else means that there is a mixed child that is not leaf, so we need to run breakMixedLeafCluster on it.

After we're done searching the parent's children, we enter the loop that actually splits the mixed_leaves into std cells clusters and macro clusters one at the time.

_As far as I understand, we won't have a non-leaf mixed cluster that will not be split._

Mixed non-leaf clusters will keep being mixed.

AcKoucher: Any comments on this? If possible, I'd like to resolve it before pushing my changes for the tight packing shaping.

maliberty: I think the root problem is that breakMixedLeafCluster mixes a recursive tree traversal and the actual breaking operation.  It makes the name confusing as it doesn't necessarily operate on a leaf cluster.   I think your rename is ok but really we split this up (it doesn't have to be in this PR).

maliberty:
It seems like this PR is a mix of a refactoring and the new array grouping code.  It would have been simpler to review separately.

AcKoucher:
I wouldn't mind closing this and opening firstly one PR for refactor and then another to add the array support. Should I keep this PR?

maliberty:
Do you plan to do more changes in this PR?  If so I would split it

AcKoucher:
Yes, more changes are needed in this PR to add the array support, I'll close this and go on with the refactoring first.

PR #4474, opened 2024-01-02
Fix rudy

gudeh:
Fix grid resolution. Previously, setGridConfig() was never called. Using same resolution as GCell grid.

Making RUDY aware of macro blockages.


maliberty:
DCO requires that you use git commit -s to sign your commits.   I use a git alias for this so I don't have to remember each time.

maliberty (changes requested):

maliberty:
Use 'clang-format -i <file>' to fix the formatting errors.

gudeh:
> DCO requires that you use git commit -s to sign your commits. I use a git alias for this so I don't have to remember each time.

I usually use the "sync fork" button on github webpage, it is really convenient. Although it seems the only way to sync the fork with a signed commit is to update the fork manually on my end.

maliberty:
A bunch of the GH automation doesn't play well with DCO unfortunately.  Suggested changes are similar.

PR #4481, opened 2024-01-03
odb: Rudy fix

gudeh:
Fix grid resolution. Previously, setGridConfig() was never called. Using same resolution as GCell grid.

Making RUDY aware of macro blockages.


maliberty:
Replaces #4474

src/odb/src/zutil/util.cpp
```diff
@@ -47,8 +49,35 @@ using std::vector;
...
+  // from RUDY paper: "The wire width p is defined by the
+  //  average wire-to-wire pitch ~p and number of routing layers l: p = ~p/l.
   wireWidth_ = block_->getTech()->findRoutingLayer(1)->getWidth();
-  makeGrid();
+
+  int track_spacing = 0;
+  odb::dbTechLayer* tech_layer = block_->getTech()->findRoutingLayer(3);
+  odb::dbTrackGrid* track_grid = block_->findTrackGrid(tech_layer);
+
+  // This code is the same on GRT, make a method for it? For example:
```
eder-matheus, line 60: You can move this implementation to the odb::dbTechLayer class and just call it here.

src/odb/src/zutil/util.cpp
```diff
@@ -47,8 +49,35 @@ using std::vector;
 RUDYCalculator::RUDYCalculator(dbBlock* block) : block_(block)
 {
   gridBlock_ = block_->getDieArea();
+  // from RUDY paper: "The wire width p is defined by the
+  //  average wire-to-wire pitch ~p and number of routing layers l: p = ~p/l.
```
eder-matheus, line 53: I didn't understand this comment with the line below. The current implementation isn't correct, then?

src/odb/src/zutil/util.cpp
```diff
@@ -113,20 +145,78 @@ void RUDYCalculator::calculateRUDY()
...
-        const auto gridBox = grid.getRect();
-        if (netBox.overlaps(gridBox)) {
-          const auto intersectArea = netBox.intersect(gridBox).area();
-          const auto gridArea = gridBox.area();
-          const auto gridNetBoxRatio = static_cast<float>(intersectArea)
-                                       / static_cast<float>(gridArea);
-          const auto rudy = netCongestion * gridNetBoxRatio * 100;
-          grid.addRUDY(rudy);
+        Tile& tile = getEditableTile(x, y);
+        const auto tileBox = tile.getRect();
```
eder-matheus, line 149: Just a comment for future reference: variable names should use the snake case pattern, not the camel case. I know this code uses the camel case pattern everywhere, but we should avoid adding more "wrong" var names. In the future, you could refactor the var names of this code to use the correct pattern.

eder-matheus (changes requested):

src/odb/src/zutil/util.cpp
gudeh: > I didn't understand this comment with the line below. The current implementation isn't correct, then?

The current implementation is considering the width of the first layer. Although, the paper mentions this averaged value. I tried implementing it the way the paper mentions and the congestion values got too high. 

My idea is to test the current behavior with my modifications on multiple designs so we could further discuss what to do.

eder-matheus: Got it. Perhaps update this comment with a TODO message, something like:
`TODO: Match the wire width with the paper definition...`, completing with the definition you have there.

eder-matheus (approved):
Looks good to me, but I'll let @maliberty give his final review and merge.

src/odb/src/db/dbTechLayer.cpp
```diff
@@ -1401,6 +1401,21 @@ int dbTechLayer::getSpacing()
   return layer->_spacing;
 }
 
+int dbTechLayer::getTrackSpacing(odb::dbTrackGrid* track_grid)
```
maliberty, line 1404: This appears to assume a single track pattern.  In a layer like m2 in asap7 we have more than one track pattern and the spacing may be non-uniform.  I don't think this matters for grt because it usually uses a high layer but if we are making a general purpose API then this is somewhat misleading.

maliberty: Perhaps getAverageTrackSpacing might be more useful and add handling for the >1 track pattern case

eder-matheus: @gudeh Global router has an implementation for this. You can look at the `averageTrackPattern` function.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2637,37 +2617,8 @@ void GlobalRouter::initRoutingTracks(int max_routing_layer)
...
-                       "Vertical tracks for layer {} not found.",
-                       tech_layer->getName());
-        return;  // error throws
-      }
-    } else {
-      logger_->error(
-          GRT, 148, "Layer {} has invalid direction.", tech_layer->getName());
-      return;  // error throws
-    }
+    tech_layer->getAverageTrackPattern(
```
gudeh, line 2620: This method "getAverageTrackPattern()" is the same as the removed piece of code in GlobalRouter.cpp in "initRoutingTracks()".

src/odb/include/odb/db.h
```diff
@@ -7816,6 +7816,22 @@ class dbTechLayer : public dbObject
...
+  /// Get the spacing between tracks for this layer.
+  /// If the layer has a multi pattern spacing returns the average.
+  ///
+  int getAverageTrackSpacing(odb::dbTrackGrid* track_grid);
+  
+  ///
+  /// Populate parameters with average track information
+  /// for tracks with multiple patterns.
+  ///
+  void getAverageTrackPattern(odb::dbTrackGrid* grid,
```
gudeh, line 7829: This is equivalent the to removed method GlobalRouter::averageTrackPattern()

src/odb/src/db/dbTechLayer.cpp
```diff
@@ -1407,6 +1407,60 @@ void dbTechLayer::setSpacing(int spacing)
...
+      track_grid->getGridPatternY(0, track_init, num_tracks, track_step);
+    } else if (track_grid->getNumGridPatternsY() > 1) {
+      getAverageTrackPattern(
+          track_grid, false, track_init, num_tracks, track_step);
+    } else {
+      getImpl()->getLogger()->error(utl::ODB,
+                                    414,
+                                    "Horizontal tracks for layer {} not found.",
+                                    getName());
+      return;  // error throws
```
maliberty, line 1426: no need for this (same below)

src/odb/include/odb/db.h
```diff
@@ -7816,6 +7816,25 @@ class dbTechLayer : public dbObject
...
+  void getAverageTrackSpacing(odb::dbTrackGrid* track_grid,
+                              int& track_step,
+                              int& track_init,
+                              int& num_tracks);
+
+  ///
+  /// Populate parameters with average track information
+  /// for tracks with multiple patterns.
+  ///
+  void getAverageTrackPattern(odb::dbTrackGrid* grid,
```
maliberty, line 7832: When would a user call this version?  It seems like a private internal method.

maliberty (changes requested):

src/odb/include/odb/db.h
gudeh: Indeed, it makes sense to be private. The private methods are the ones starting with an underscore?

maliberty: Move the method to 'class _dbTechLayer' and remove it from db.h.  Be sure to put it in the "// User Code Begin Methods" block in dbTechLayer.h

maliberty (approved):

maliberty:
You've got a odb code gen diff to fix:
```
  diff --git a/src/odb/src/db/dbTechLayer.h b/src/odb/src/db/dbTechLayer.h
  index 59341cb3f..364bf63d2 100644
  --- a/src/odb/src/db/dbTechLayer.h
  +++ b/src/odb/src/db/dbTechLayer.h
  @@ -44,7 +44,6 @@ namespace odb {
   class dbIStream;
   class dbOStream;
   class dbDiff;
  -class dbTrackGrid;
```

maliberty:
@eder-matheus feel free to merge once the tests pass

eder-matheus:
@gudeh One of the checks is failing. You can fix it by following the suggestion in the check pipeline:
```
Your ODB generated files are out of date
run ./generate in the src/odb/src/codeGenerator folder
```

PR #4482, opened 2024-01-03
fix coverity issues

eder-matheus:


src/ppl/src/IOPlacer.cpp
```diff
@@ -2245,10 +2248,12 @@ void IOPlacer::placePin(odb::dbBTerm* bterm,
     const int min_area = layer->getArea() * database_unit * database_unit;
     if (layer->getDirection() == odb::dbTechLayerDir::VERTICAL) {
       width = layer->getMinWidth();
-      height = int(std::max((double) width, ceil(min_area / width)));
+      height = int(std::max((double) width,
+                            ceil(static_cast<double>(min_area) / width)));
```
QuantamHD, line 2252: You should just factor out min_area into its own variable outside the if, and I would not mix c style casts with C++ casts for width.

eder-matheus: Done, thanks for the suggestion!

maliberty (approved):

PR #4483, opened 2024-01-03
Fix misc gcc 13.2 compiler warnings

maliberty:


src/odb/src/db/dbTech.cpp
```diff
@@ -572,7 +572,7 @@ void _dbTech::_setLefVersion(double inver)
              minor_version,
              opt_minor_version);
   else
-    snprintf(_version_buf, 10, "%d.%d", major_version, minor_version);
+    snprintf(_version_buf, 12, "%d.%d", major_version, minor_version);
```
QuantamHD, line 575: I don't think we should leave this. It's clearly pretty dangerous. Should just replace _version_bug with std::string

QuantamHD (changes requested):

src/odb/src/db/dbTech.cpp
maliberty: I've fixed this

maliberty: In practice this was never going to be an issue as LEF version are simple, like 5.8, and not some huge number.

QuantamHD: LGTM. Thanks for doing that. I know in practice it would never be an issue, but the c style string stuff just feels like more complexity than its worth when we have much more powerful tools. I think it made a nice improvement to the code overall.

QuantamHD (approved):

vvbandeira (approved):

PR #4488, opened 2024-01-04
Sta gui 50 paths is enough

oharboe:
1000 paths can take a LONG time to update, who needs more than 50 initially?

Alternatively, the GUI could return immediately and populate in the background


maliberty:
You can already change this in the settings:
image

maliberty:
@gadfort you picked the default - any opinion?

gadfort:
@maliberty I actually just copied the default from the initial implementation, I think 50 is fine.

maliberty (approved):

PR #4491, opened 2024-01-04
Adding support for pin slack and capacitance to Python APIs

VidyaChhabria:


include/ord/Timing.h
```diff
@@ -104,6 +107,7 @@ class Timing
   bool isEndpoint_(sta::Pin* sta_pin);
   float getPinSlew_(sta::Pin* sta_pin, MinMax minmax);
   float getPinArrival_(sta::Pin* sta_pin, RiseFall rf, MinMax minmax);
+  float getPinSlack_(sta::Pin* sta_pin, RiseFall rf, MinMax minmax);
```
maliberty, line 110: Why is there an underscore in the name?

include/ord/Design.h
```diff
@@ -172,6 +172,7 @@ class Design
     Max
   };
   float getNetCap(odb::dbNet* net, sta::Corner* corner, MinMax minmax);
+  float getPortCap(odb::dbITerm* pin, sta::Corner* corner, MinMax minmax);
```
maliberty, line 175: I would rather see APIs that require sta objects moved to Timing.h so that Design is more about the netlist (obd).

maliberty (changes requested):

include/ord/Timing.h
VidyaChhabria: This was just a convention that I had followed, indicating that the function is a private version of the user-facing public API. These private functions can access internal objects like STA pins. If there is a better coding style, I can change it; please let me know what would be appropriate.

maliberty: I would just drop the underscore as the overload resolution is pretty clear.  If you prefer a different name then I would replace underscore with Priv or Impl.

maliberty (approved):

PR #4493, opened 2024-01-04
odb: refactoring out serialization code for std::variant

fgaray:
- Makes the code in dbScanPin more simple
- std::variant now is available to be use for serialization by others

### DIFF
--- a/src/odb/include/odb/dbStream.h
+++ b/src/odb/include/odb/dbStream.h
@@ -39,6 +39,7 @@
 #include <ostream>
 #include <string>
 #include <unordered_map>
+#include <variant>
 
 #include "ZException.h"
 #include "dbObject.h"
@@ -49,6 +50,8 @@
 namespace odb {
 
 class _dbDatabase;
+
+inline constexpr size_t kTemplateRecursionLimit = 16;
 
 class dbOStream
 {
@@ -173,6 +176,9 @@ class dbOStream
   template <size_t I = 0, typename... Ts>
   constexpr dbOStream& operator<<(const std::tuple<Ts...>& tup)
   {
+    static_assert(I <= kTemplateRecursionLimit,
+                  "OpenROAD disallows of std::tuple larger than 16 "
+                  "elements. You should look into alternate solutions");
     if constexpr (I == sizeof...(Ts)) {
       return *this;
     } else {
@@ -220,6 +226,23 @@ class dbOStream
     *this << tmp;
     free((void*) tmp);
     return *this;
+  }
+
+  template <uint32_t I = 0, typename... Ts>
+  dbOStream& operator<<(const std::variant<Ts...>& v)
+  {
+    static_assert(I <= kTemplateRecursionLimit,
+                  "OpenROAD disallows of std::variants larger than 16 "
+                  "elements. You should look into alternate solutions");
+    if constexpr (I == sizeof...(Ts)) {
+      return *this;
+    } else {
+      if (I == v.index()) {
+        *this << (uint32_t) v.index();
+        *this << std::get<I>(v);
+      }
+      return ((*this).operator<< <I + 1>(v));
+    }
   }
 
   double lefarea(int value) { return ((double) value * _lef_area_factor); }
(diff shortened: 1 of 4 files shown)


src/odb/include/odb/dbStream.h
```diff
@@ -398,9 +413,31 @@ class dbIStream
...
+  template <uint32_t I = 0, typename... Ts>
+  dbIStream& variantHelper(uint32_t index, std::variant<Ts...>& v)
+  {
+    if constexpr (I == sizeof...(Ts)) {
+      return *this;
+    } else {
+      if (I == index) {
+        *this >> std::get<I>(v);
+      }
+      return ((*this).operator>><I + 1>(v));
```
QuantamHD, line 438: What's the template recursion limit?

fgaray: Line 432, when I is equal to the number of elements in the variant, then we just return dbIStream, else we do a recursion to the next element.

QuantamHD: There's a C++ compiler limit to the recursion depth though, at least I thought there was. I also wouldn't be entirely surprised if its lower inside Google than default configurations of LLVM.

QuantamHD: looks like it's around 500 https://stackoverflow.com/questions/12277732/template-metaprogramming-recursion-up-limits

QuantamHD: Looking inside Google it appears this limit is 40, which is a lot lower. Can we include a compile error in the template at 39 to let the user they need to do something else.

fgaray: Yeah, seems to be controlled by -ftemplate-depth. I looked and I could not find one source of true, but values seems to vary from 256 to 1024.

I don't think there is a use case for variants bigger than a few elements, do you want me to put a limit, let's say 16 so future usage of big variants fail to compile? I think it is better than introducing possible sources of incompatibility for future compilations.

maliberty: Anybody using a variant with 40+ members will have to use a real data structure instead....

QuantamHD: Yeah 16 feels like the right number

src/odb/include/odb/dbStream.h
```diff
@@ -222,6 +226,21 @@ class dbOStream
     return *this;
   }
 
+  template <uint32_t I = 0, typename... Ts>
+  dbOStream& operator<<(const std::variant<Ts...>& v)
+  {
+    static_assert(I < kTemplateRecursionLimit);
```
QuantamHD, line 232: Let's add a message to the static assert. "OpenROAD disallows of std::variants larger than 16 elements. You should look into alternate solutions."

Also we probably want <=.

src/odb/include/odb/dbStream.h
```diff
@@ -381,6 +400,7 @@ class dbIStream
   template <size_t I = 0, typename... Ts>
   constexpr dbIStream& operator>>(std::tuple<Ts...>& tup)
   {
+    static_assert(I <= kTemplateRecursionLimit, "OpenROAD disallows of std::variants larger than 16 elements. You should look into alternate solutions");
```
QuantamHD, line 403: Probably should be std::tuple in this message.

QuantamHD (approved):

QuantamHD (approved):

src/odb/include/odb/dbStream.h
fgaray: Thanks!, good catch!

fgaray:
I will squash the commits and for push to clean up the PR and re-trigger the CI

fgaray:
Hi @maliberty , this should be ready for a final review, thanks!

maliberty (changes requested):
You have changed the schema without any backward compatibility changes.

fgaray:
> You have changed the schema without any backward compatibility changes.

I think the change is equivalent to the previous code:

- I moved the serialization code from dbScanPin to dbStream. The templated code in dbStream should generate the same code that was used before in dbScanPin.
- The std::variant pin_ variable added in dbScanPin.json schema file was added since now I no longer need to do the serialization manually, but the code should be the same. Previously I was adding it manually to the private _dbScanPin class.
- I don't think anyone was using std::tuple bigger than 16 elements so that should also be a nop change.

Maybe there is something that am I missing?

maliberty (approved):

Issue #3020, opened 2023-03-15
grt creates guides that go through blockages


antonblanchard:
### Describe the bug

The attached test case shows guides that go through a blockage. M7 is completely blocked, but some guides go right through it from M5 through to M8:

![grt-through-blockage](https://user-images.githubusercontent.com/2029971/225280426-d1b05b81-42d7-4dde-92ba-23a50e31b143.png)

### Expected Behavior

Guides should not cross through obstructions

### Environment

```shell
Git commit: edcd14ffec5d232d98e8b0c34315b1ce233b1af6
os: Fedora Linux 37 (Workstation Edition)
cmake version 3.26.0-rc5
```

### To Reproduce

Test case:  [grt-through-blockage.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/10978590/grt-through-blockage.tar.gz)

eder-matheus:
This is a good case to start considering via resources. The current grt core code doesn't check the resources when adding vias. Also, the 2D resources will consider the M8 tracks, even if they are inaccessible from lower layers.

The first step could be to detect inaccessible layers and "block" them from the actual resources. The via-resource model is a more complex work that can be implemented next.

maliberty:
@antonblanchard would you give some more context as to what you are seeking to have happen by blocking m7?

antonblanchard:
> @antonblanchard would you give some more context as to what you are seeking to have happen by blocking m7?

Trying to build simple test cases to exercise specific parts of drt. In this particular case I could obstruct all layers other than the two I'm interested in.

maliberty:
What does it mean to block all but two non-adjacent layers?

antonblanchard:
> What does it mean to block all but two non-adjacent layers?

To force specific vias to be used in test cases. I'm not sure how much of an issue this is outside of test cases, and I can see how grt would struggle since it starts with a 2D view of the capacity.

PR #4494, opened 2024-01-05
grt: handle obstructions that completely block the die area

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/3020

When a routing obstruction completely blocks the die area, block the remaining layers above it since those layers are unreachable from the lower layers. Also error out when a pin is placed inside the blocked layers range.

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -3140,15 +3140,17 @@ int GlobalRouter::findObstructions(odb::Rect& die_area)
     }
   }
 
-  for (odb::dbBTerm* bterm : block_->getBTerms()) {
-    for (odb::dbBPin* bterm_pin : bterm->getBPins()) {
-      for (odb::dbBox* bpin_box : bterm_pin->getBoxes()) {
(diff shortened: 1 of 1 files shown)


maliberty:
I have mixed feeling about this approach.  It is narrowly tailored to a specific and unusual situation that would be more generally handled by via modelling.

It also assumes that you should always block layers above the blockage.  You could have two M1 pins that are connected in one net, two M6 connected in another net and an M4 blockage between them.  Both nets are still routable but this PR would prevent that.  Extending it to that even more obscure case doesn't seem worth the effort.

eder-matheus:
> I have mixed feeling about this approach. It is narrowly tailored to a specific and unusual situation that would be more generally handled by via modelling.
> 
> It also assumes that you should always block layers above the blockage. You could have two M1 pins that are connected in one net, two M6 connected in another net and an M4 blockage between them. Both nets are still routable but this PR would prevent that. Extending it to that even more obscure case doesn't seem worth the effort.

I see your point and I agree with it. I was thinking on a quick solution for the issue, but I believe it's better to spend some time on the via modeling instead. I'll close this.

Issue #2560, opened 2022-12-01
Investigate the monotonic router in fastroute


maliberty:
The FastRoute 2.0 and later papers all mention monotonic routing.   FastRouteCore::routeMonotonicAll is never called anywhere in the code so it seems to be unused in practice.  Someone should investigate:

1) Does it work as the papers describe?
2) Does it help if inserted into FastRouteCore::run before maze routing?

QuantamHD:
Paper link https://ieeexplore.ieee.org/document/4196040

PR #4503, opened 2024-01-09
rename routLV to reoutMonotonic

arthurjolo:
Fixes [#2560](https://github.com/The-OpenROAD-Project/OpenROAD/issues/2560) 

As the issue mentioned above mentions, the fastroute papers talk about a monotonic routing step. In our fastrout we had a method called FastRouteCore::::routeMonotonicAll but it as never used.

After investigating more about the monotonic routing I found that the method FastRouteCore::routeLVAll that is used during fastroute is doing a monotonic routing step, but it was labeled as a patter routing step. This PR renames the method  FastRouteCore::routeLVAll to FastRouteCore::::routeMonotonicAll, and removes the unused implementation of the monotonic routing.


src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -405,22 +405,20 @@ class FastRouteCore
...
   void routeSegH(Segment* seg);
   void routeSegLFirstTime(Segment* seg);
   void spiralRoute(int netID, int edgeID);
-  void routeLVEnew(int netID,
-                   int edgeID,
-                   multi_array<float, 2>& d1,
-                   multi_array<float, 2>& d2,
-                   int threshold,
-                   int enlarge);
+  void routeMonotonicNew(int netID,
```
eder-matheus, line 416: Since we don't have another function with the same name, you can rename it to just "routeMonotonic".

eder-matheus (changes requested):

eder-matheus (approved):

PR #4506, opened 2024-01-09
Optimize `pdr::get_nearest_neighbors`

kbieganski:
It's a hot function in global placement. Probably the biggest win here is capturing `pts` by reference instead of by copy, but also there's fewer allocations.

| `gpl` run | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| `ariane133` (before) | 486.971 ± 0.833 | 485.887 | 487.765 |
| `ariane133` (after) | 447.314 ± 6.410 | 441.101 | 454.566 |
| `black_parrot` (before) | 562.767 ± 2.765 | 553.641 | 565.986 |
| `black_parrot` (after) | 537.563 ± 5.191 | 524.795 | 543.223 |

Global placement speedup is about 5-8%.

### DIFF
--- a/src/stt/src/pdr/src/pd.cpp
+++ b/src/stt/src/pdr/src/pd.cpp
@@ -63,6 +63,9 @@ using utl::Logger;
 using Neighbors = std::vector<int>;
 static vector<Neighbors> get_nearest_neighbors(const vector<Point>& pts)
 {
+  thread_local static vector<int> data;
+  data.clear();
+
   const int pt_count = pts.size();
 
   vector<Neighbors> neighbors(pt_count);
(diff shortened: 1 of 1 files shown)


src/stt/src/pdr/src/pd.cpp
```diff
@@ -63,6 +63,9 @@ using utl::Logger;
 using Neighbors = std::vector<int>;
 static vector<Neighbors> get_nearest_neighbors(const vector<Point>& pts)
 {
+  thread_local static vector<int> data;
+  data.clear();
```
maliberty, line 67: There is no need to clear as you are going to resize and overwrite.

src/stt/src/pdr/src/pd.cpp
```diff
@@ -72,17 +75,23 @@ static vector<Neighbors> get_nearest_neighbors(const vector<Point>& pts)
...
-  vector<int> ur(pt_count, std::numeric_limits<int>::max());
-  vector<int> ul(pt_count, std::numeric_limits<int>::min());
-  vector<int> lr(pt_count, std::numeric_limits<int>::max());
-  vector<int> ll(pt_count, std::numeric_limits<int>::min());
+  data.reserve(pt_count * 5);
+  data.resize(pt_count, std::numeric_limits<int>::max());
+  data.resize(pt_count * 2, std::numeric_limits<int>::min());
+  data.resize(pt_count * 3, std::numeric_limits<int>::max());
+  data.resize(pt_count * 4, std::numeric_limits<int>::min());
+  data.resize(pt_count * 5);
```
maliberty, line 83: Why resize repeatedly - why not resize once (no need to reserve) and then use std::fill.

maliberty (changes requested):

src/stt/src/pdr/src/pd.cpp
kbieganski: `resize` does zero initialization. It's not much of a difference (maybe it even gets optimized away?) so I can change it if you insist.

maliberty: That's a fair point.  You could reorder the blocks so that you only resize once for the max() values and once for the min() values.

kbieganski: Good idea, done.

kbieganski:
Right! Should be OK now.

maliberty (approved):

PR #4507, opened 2024-01-09
Reduce number of pin location lookups in Steiner tree construction

kbieganski:
There are some unnecessary pin location lookups during Steiner tree construction in global placement. This change fetches these locations once and reuses them.

| `gpl` run | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| `ariane133` (before) | 486.971 ± 0.833 | 485.887 | 487.765 |
| `ariane133` (after) | 447.517 ± 1.311 | 446.120 | 449.581 |
| `black_parrot` (before) | 562.767 ± 2.765 | 553.641 | 565.986 |
| `black_parrot` (after) | 525.773 ± 1.841 | 522.481 | 530.099 |

Global placement speedup is about 7-8%.

### DIFF
--- a/src/rsz/src/SteinerTree.cc
+++ b/src/rsz/src/SteinerTree.cc
@@ -58,8 +58,9 @@ using sta::NetConnectedPinIterator;
 static void
 connectedPins(const Net *net,
               Network *network,
+              dbNetwork* db_network,
               // Return value.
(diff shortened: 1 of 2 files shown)


maliberty (approved):
Thanks

PR #4514, opened 2024-01-10
grt: evalue coeffADJ only if needed

RRozak:
It moves variables and a call to `coeffADJ` to the scopes where their values are needed. It isn't a big optimization. As I checked by running black_parrot under callgrind, program spends 0.57% of time of grt phase in that call.

### DIFF
--- a/src/grt/src/fastroute/src/RSMT.cpp
+++ b/src/grt/src/fastroute/src/RSMT.cpp
@@ -662,17 +662,6 @@ void FastRouteCore::gen_brk_RSMT(const bool congestionDriven,
 
     FrNet* net = nets_[i];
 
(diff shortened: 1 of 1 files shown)


eder-matheus (commented):
@RRozak The code looks fine, but two regression tests are failing. Apparently, the grt results are different for these tests. Could you look at what is causing the differences?

RRozak:
Thanks for the review. Tests passed locally when I rebased it to the current master.

eder-matheus:
> Thanks for the review. Tests passed locally when I rebased it to the current master.

Thanks! I see it is passing now. I'll merge the PR when the remaining checks end.

eder-matheus (approved):

PR #4516, opened 2024-01-10
mpl2: refactor mixed leaves splitting

AcKoucher:
Refactoring to prepare for macro array support.

Originally, we had one function to both search and split the mixed leaves in the multilevel autoclustering stage.

This PR is to split this function into two: one to fetch the mixed leaves and other to actually split them into std cells and macro clusters.

If the approach is ok, I'll run Secure-CI


maliberty:
go ahead with CI

AcKoucher:
Secure-CI passed.

maliberty (approved):

PR #4518, opened 2024-01-10
Fix non-determinism in global placement caused by different FFT output values

mithro:
The global placement depends on the output of the FFT library. Without `-ffp-contract=off` the current FFT library produces different results dependent on if the compiler is targeting an x86 architecture which has the "FMA instruction".

The FMA instruction omits the rounding done in the multiplication instruction, making the calculations produce different results for some input values

See https://kristerw.github.io/2021/11/09/fp-contract/

This pull request;
 * Adds a test for the FFT library which checks the floating point output is correct.
 * Turns `-ffp-contract=off` on for the whole of OpenROAD to make sure the floating point values are deterministic.


maliberty (approved):

QuantamHD (approved):
Incredible Debugging!

PR #4521, opened 2024-01-11
rsz: clang-format

eder-matheus:


eder-matheus:
@maliberty I'm not sure if there is a reason why all these files were not clang-formatted before, but I noticed a bunch of this while investigating resizer and decided to fix that before any other code update.

maliberty:
We should wait for https://github.com/The-OpenROAD-Project/OpenROAD/pull/3896 to finish

QuantamHD:
I wonder if this is an actual bug

image

maliberty:
@QuantamHD I don't think its a bug as swapping is a commutative operation so it doesn't really matter.  You could make tidy happy by swapping the call

maliberty:
This is a good idea but the PR is so old it would be good to redo it on the current head.

PR #4523, opened 2024-01-11
Rudy supply net fix

gudeh:
odb: Enhance RUDY with supply net handling and guard clauses
- Add guard clauses in RUDY for zero die area and missing grid object.
- Relocate getAverageTrackSpacing() from dbTechLayer to dbTrackGrid.
- Introduce private methods for enhanced code organization and readability in RUDY.


src/odb/src/zutil/util.cpp
```diff
@@ -49,15 +49,19 @@ using std::vector;
 RUDYCalculator::RUDYCalculator(dbBlock* block) : block_(block)
 {
   gridBlock_ = block_->getDieArea();
+  if (gridBlock_.area() == 0) {
```
eder-matheus, line 52: Shouldn't this be an error? I can't think of a case where the die area would be zero.

src/odb/src/zutil/util.cpp
```diff
@@ -49,15 +49,19 @@ using std::vector;
...
+    return;
+  }
   // TODO: Match the wire width with the paper definition
   wireWidth_ = block_->getTech()->findRoutingLayer(1)->getWidth();
 
   odb::dbTechLayer* tech_layer = block_->getTech()->findRoutingLayer(3);
   odb::dbTrackGrid* track_grid = block_->findTrackGrid(tech_layer);
+  if (track_grid == nullptr) {
```
eder-matheus, line 60: Similar comment here. I believe that if a layer doesn't have a track grid pattern defined at the stage RUDY is called, we should throw an error.

src/odb/src/zutil/util.cpp
```diff
@@ -76,71 +80,47 @@ void RUDYCalculator::setGridConfig(odb::Rect block, int tileCntX, int tileCntY)
...
+      const auto netBox = net->getTermBBox();
+      processTilesIntersectionSignalNet(netBox, tile_width, tile_height);
+    } else {
+      for (odb::dbSWire* swire : net->getSWires()) {
+        for (odb::dbSBox* s : swire->getWires()) {
+          if (s->isVia()) {
+            continue;
+          } else {
+            odb::Rect wire_rect = s->getBox();
+            processTilesIntersection(wire_rect, tile_width, tile_height, 2);
```
eder-matheus, line 122: Is there a reason for setting the `nets_per_tile` as 2 in this function call?

src/odb/src/zutil/util.cpp
```diff
@@ -170,34 +150,84 @@ void RUDYCalculator::processMacroObstruction(odb::dbMaster* macro,
...
-          const auto wireArea = hpwl * wireWidth_;
-          const auto obstr_congestion = wireArea / tileBox.area();
-          const auto intersectArea = macroObstrRect.intersect(tileBox).area();
-          const auto tileArea = tileBox.area();
-          const auto tileObstrBoxRatio = static_cast<float>(intersectArea)
-                                         / static_cast<float>(tileArea);
-          const auto rudy = obstr_congestion * tileObstrBoxRatio * 100 * 2;
-          tile.addRUDY(rudy);
-        }
+void RUDYCalculator::processTilesIntersectionSignalNet(const odb::Rect netBox,
```
eder-matheus, line 196: I would create a specific function for the supply nets instead of a function for the signal nets since the signal nets are handled according to the paper, and you added a specific way to support supply nets.

eder-matheus (changes requested):
Besides my comments, please address the clang-tidy comments as well.

src/odb/src/zutil/util.cpp
gudeh: The function for signal nets would not be strictly necessary, I just created it because I found it easier to read the code. You think I remove it either way?

eder-matheus: It is good to have the two separate functions; it helps readability a lot. I would rename the functions to make it clear that the "special case" is the supply nets and not the signal nets.

But this is just a minor comment, so feel free to choose what you think is best.

src/odb/src/zutil/util.cpp
maliberty: I ran into trouble with opening the GUI in the floorplan step as the rudy object is created before the flooplan has actually been run.

src/odb/src/db/dbTrackGrid.cpp
```diff
@@ -271,4 +312,24 @@ void dbTrackGrid::destroy(dbTrackGrid* grid_)
...
+// User Code Begin PrivateMethods
+void _dbTrackGrid::getAverageTrackPattern(bool is_x,
+                                          int& track_init,
+                                          int& num_tracks,
+                                          int& track_step)
+{
+  std::vector<int> coordinates;
+  dbTrackGrid* track_grid = (dbTrackGrid*) this;
+  if (is_x) {
+    track_grid->getGridX(coordinates);
```
gudeh, line 324: I am not sure if this is the proper way to call a public method in the private context. I am accessing the "this" object without the underline. Is this allowed?

src/odb/src/zutil/util.cpp
gudeh: This was a mistake. The number 2 seemed to work better with the macros, and was accidentally added here.

gudeh: Furthermore, I find this solution of adding "nets per tile" (hypothetically assuming their presence) is somewhat inelegant, but it provided a good congestion result on most designs, considering the macros.

src/odb/src/db/dbTrackGrid.cpp
maliberty: That's fine.

PR #4526, opened 2024-01-12
grt: Remove unnecessary ifs

RRozak:
It removes unnecessary if statements from nested loops.


eder-matheus (approved):
The clang-tidy comments are not directly related to your changes, so I can take care of them later today.

PR #4535, opened 2024-01-15
gui: add charts event when qtcharts is not installed

### LINKED ISSUE #4433: Add Charts even if qt charts is not installed, but with a text saying qt charts not installed
### Description

If there are features that are invisible, because some dependency was not installed at compile time, how can the user know about them?

### Suggested Solution

Add Charts even if qt charts is not installed, but with a text saying qt charts not installed. This will tell users what features they are missing out on and think to look up the documentation to build OpenROAD correctly

### Additional Context

_No response_

AcKoucher:
Resolves #4433 

With these changes, Charts will always be built within the GUI, but if QtCharts is not installed, a message will be displayed "QtCharts is not installed".

Screenshot from 2024-01-15 14-58-16

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -29,10 +29,6 @@ if (Qt5_FOUND AND BUILD_GUI)
 
     target_compile_definitions(gui PRIVATE ENABLE_CHARTS)
 
-    list(APPEND CHARTS_SRC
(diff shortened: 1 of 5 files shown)


eder-matheus (approved):
Looks good to me. @oharboe could check if this covers what you've asked?

oharboe (approved):

PR #4544, opened 2024-01-16
mpl2: add support for array of interconnected macros

### LINKED ISSUE #4085: Improve RTLMP so that it can lay out mock-array in an array
### Description

unzip [rtlmp.zip](https://github.com/The-OpenROAD-Project/OpenROAD/files/12808850/rtlmp.zip)

```
OpenROAD v2.0-10574-gd04de2f32 
[deleted]
```

Placement result, not an array:

image

### Suggested Solution

Should result in an array.

### Additional Context

_No response_

AcKoucher:
Now that we have:
- #4614
- #4815
- #4206
- #4705

It's possible to resolve #4085 with the changes here and by setting the boundary penalty to 0.

Changes:
1. Adding another classification for single macro clusters based on interconnection during mixed leaves splitting.
2. Using this classification to merge macro clusters
3. Adding tight packing tiling computation for arrays of interconnected macros.
4. Add possibility to manually initialize sequence pair
5. Macro arrays have their sequence pair initialized to their respectful shape and SA only applies exchange and flip moves.


maliberty:
are the macros within the array in their expected positions?

AcKoucher:
Running without any further changes leads to mpl2 not finding a valid solution (i.e., all macros inside the outline). My suspicions are that this array is a bit large, so there are many sequence pair representation options and the annealer would need more steps to actually find the optimal solution.
image


Increasing the number of steps considerably to see if it would converge (number of steps from 2000 to 5000):
image

maliberty:
It may make sense to pick a shape and then just use macro swap moves rather than the full set of transforms (many of which will destroy the shape).

maliberty:
Secure CI?

maliberty (approved):

oharboe:
@AcKoucher Could you make a followup ORFS PR to update mock-array to demonstrate this feature?

AcKoucher:
@oharboe yes. I'll do it.

AcKoucher:
@maliberty  Secure-CI passed

PR #4546, opened 2024-01-17
Build with LTO in release mode by default

kbieganski:
Enables link time optimizations by default in release mode, yielding 3-12% better performance for the entire flow.

| `nangate45/ibex` run | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No LTO | 256.417 ± 1.089 | 252.747 | 258.500 | 1.12 ± 0.01 |
| With LTO | 228.625 ± 1.979 | 223.222 | 230.904 | 1.00 |

| `nangate45/black_parrot` run | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No LTO | 2961.542 ± 14.714 | 2937.151 | 2987.307 | 1.06 ± 0.01 |
| With LTO | 2795.283 ± 26.873 | 2725.632 | 2822.192 | 1.00 |

| `nangate45/ariane133` run | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No LTO | 5807.242 ± 17.377 | 5791.852 | 5832.0591 | 1.03 ± 0.02 |
| With LTO | 5651.473 ± 71.611 | 5550.869 | 5739.271 | 1.00 |


QuantamHD:
I've tried enabling this before, but the GUI fails to run under LTO. Have you tried running the GUI?

QuantamHD:
https://github.com/InBetweenNames/gentooLTO/issues/444

QuantamHD:
It also might be possible to compile the GUI without LTO and the other parts with, but I never looked into how that might work.

QuantamHD:
If LTO doesn't work out AutoFDO might be a more stable option https://gcc.gnu.org/wiki/AutoFDO/Tutorial

maliberty:
Please also give a summary of the impact on link time of using whatever option you settle on.

kbieganski:
| | Full build | Linking only |
|:-|-:|-:|
| No LTO | 8 min 23 s | 5 s |
| With LTO | 9 min 23 s | 1 min 30 s |

On GCC 13.2.1, i7-9700 (8 cores)

The GUI actually works for me with LTO (Arch Linux, Qt 5.15.12). I'll try with Clang. Also, seems like the CI is failing on Clang.

rovinski:
If it's not possible to build only the GUI without LTO, then maybe CMake can just error if both LTO and the GUI are enabled.

kbieganski:
If we're not going to enable it by default, there's no point in this PR. Closing it.

For future reference:
* If you want to build with LTO, use `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` (or [`-DLINK_TIME_OPTIMIZATION=ON`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/589dee1c8b0d678bd2989dbb1150df86a0b5271f/CMakeLists.txt#L57)) when configuring
* The GUI worked for me with LTO (Qt 5.15.12), so this issue may have been resolved. (Also, for CI I recommend `-DBUILD_GUI=no`)
* Use a new-ish compiler and avoid the BFD linker (the default GNU linker). For me, BFD segfaulted with Clang. Use `gold`, `lld`, or `mold`. They're much faster anyway: LTO link time with Clang+any of these is about twice as fast as GCC+BFD.

maliberty:
We could make it the default but there needs to be an option to turn it off for developers in an edit/compile/run loop.

kbieganski:
For sure, you don't want LTO while developing. I'll get back to this then.

CMakeLists.txt
```diff
@@ -107,6 +107,11 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE RELEASE)
 endif()
 
+# Build with link time optimizations in release mode by default. Brings noticeable speedup.
+if(CMAKE_BUILD_TYPE STREQUAL "RELEASE")
+  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+endif()
```
gadfort, line 113: shouldn't this be gated by the BUILD_GUI as well?

kbieganski: Yes, I hadn't gotten around to it then. It is gated now.

maliberty: If we don't use it with BUILD_GUI then it will hardly ever be used as almost everyone builds with the GUI.   We need to determine if it works or not with the GUI enabled.

maliberty: Its unclear to me if that means you just don't get LTO on Qt without building from source or if you can't link such a library at all when using LTO.  I suspect it is the former as people don't rebuild libc from scratch to use lto.

maliberty:
@kbieganski are you still planning to get back to this?

kbieganski:
I'm busy with other stuff at the moment, so if anyone is willing to take over, go ahead.

maliberty:
handled elsewhere

PR #4547, opened 2024-01-17
odb: add support for MASK on bpins

gadfort:
Adds:
- support for layer mask in dbBox
- adds DEF in and out support for LAYER x MASK x for PINS
- test case for def

### DIFF
--- a/src/odb/src/db/dbBox.cpp
+++ b/src/odb/src/db/dbBox.cpp
@@ -80,6 +80,9 @@ bool _dbBox::operator==(const _dbBox& rhs) const
   if (_flags._layer_id != rhs._flags._layer_id)
     return false;
 
+  if (_flags._layer_mask != rhs._flags._layer_mask)
+    return false;
+
   if (_flags._via_id != rhs._flags._via_id)
     return false;
   if (_flags._octilinear != rhs._flags._octilinear)
@@ -136,6 +139,8 @@ int _dbBox::equal(const _dbBox& rhs) const
     }
   }
   if (_flags._octilinear != rhs._flags._octilinear)
+    return false;
+  if (_flags._layer_mask != rhs._flags._layer_mask)
     return false;
   if (design_rule_width_ != rhs.design_rule_width_)
     return false;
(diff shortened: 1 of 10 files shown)


src/odb/src/db/dbBox.cpp
```diff
@@ -741,12 +750,22 @@ dbTechLayer* dbBox::getTechLayer()
   return (dbTechLayer*) box->getTechLayer();
 }
 
+uint dbBox::getLayerMask()
+{
+  _dbBox* box = (_dbBox*) this;
+  if (getTechLayer() == nullptr) {
```
maliberty, line 756: Could you have a non-zero mask but no layer?  I'm wondering if this check is needed here or this should be enforced in dbBox::create

maliberty:
Please resolve clang-tidy

src/odb/src/db/dbBox.cpp
gadfort: looking at the DEF I don't think you can. I can remove that check since the `_layer_mask` gets initialized to 0 anyways. I'll look at the ::create and add checks there.

maliberty (approved):
@eder-matheus please merge if the test pass

QuantamHD:
@gadfort I think the change looks good, but the one thing I'll say is the limit on Via IDs seems pretty low now at 8192. I could honestly see us hitting that limit at some point. It might require us to rethink how boxes get their via_ids, and maybe factoring it out into its own table.

gadfort:
> @gadfort I think the change looks good, but the one thing I'll say is the limit on Via IDs seems pretty low now at 8192. I could honestly see us hitting that limit at some point. It might require us to rethink how boxes get their via_ids, and maybe factoring it out into its own table.

@QuantamHD that would require that either there are more than 8192 tech vias or block vias. If it becomes an issue we could also increase the size of the flags field from 4bytes to something slightly larger.

maliberty:
Some of the bits could be saved - is_block & is_tech could have been one bit as there are only two choices.  Marked isn't important today.

QuantamHD:
In the most extreme case we could create a table of dbBox<Id> and Via<Ids> and then anytime you needed to know which via it was you pay an indirection cost. It's just a question of if the memory saved is worth the cost of indirection.

QuantamHD:
The trend seems to be more vias and more metal layers as process technology advances so this limit is something we should be aware of.

maliberty:
We should at least have a check against overflow

PR #4551, opened 2024-01-18
tap: insert missing cells that partially overlaps with endcaps

eder-matheus:
When inserting tapcells, we assume that the row is filled when it overlaps with other placed cells, usually endcaps at the end of the row or rows that are actually filled.

This PR updates this process to make two checks: if the tapcell that would be inserted partially overlaps with another cell (i.e., the x_min of the tapcell is lesser than the x_min of the fixed cell), it will check if placing the tapcell avoiding the overlap with the former fixed cell will not cause any other overlaps. If this check passes, the tool will place the tapcell with this "left shift" from the fixed cell.

This way, we ensure that the minimum distance required between the tapcells in a row will always be respected, but we also avoid placing the tapcell in a completely filled row.


maliberty (approved):

maliberty:
If you shift left a tapcell, is the next tapcell aware of the offset?  Feel free to merge if yes.

PR #4555, opened 2024-01-19
deltaDebug: add --start parameter

oharboe:
When runs are hours, then starting at a given level can be important.

### DIFF
--- a/etc/deltaDebug.py
+++ b/etc/deltaDebug.py
@@ -31,12 +31,22 @@ from math import ceil
 import errno
 import enum
 
+persistence_range = [1, 2, 3, 4, 5, 6]
+persistence_start_range = persistence_range[1:]
 
 parser = argparse.ArgumentParser('Arguments for delta debugging')
 parser.add_argument('--base_db_path', type=str, help='Path to the db file to perform the step on')
 parser.add_argument('--error_string', type=str, help='The output that indicates target error has occured')
 parser.add_argument('--step', type=str, help='Command used to perform step on the input odb file')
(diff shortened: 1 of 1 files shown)


etc/deltaDebug.py
```diff
@@ -31,12 +31,16 @@
...
 
+persistence_range = [1, 2, 3, 4, 5, 6]
 
 parser = argparse.ArgumentParser('Arguments for delta debugging')
 parser.add_argument('--base_db_path', type=str, help='Path to the db file to perform the step on')
 parser.add_argument('--error_string', type=str, help='The output that indicates target error has occured')
 parser.add_argument('--step', type=str, help='Command used to perform step on the input odb file')
-parser.add_argument('--persistence', type=int, default=1, choices=[1, 2, 3, 4, 5, 6], help='Indicates maximum input fragmentation; fragments = 2^persistence; value in [1,6]')
+parser.add_argument('--start', type=int, default=1,
+                    choices=persistence_range[1:],
```
maliberty, line 41: Is it ok that the default is not in choices?

etc/deltaDebug.py
```diff
@@ -110,7 +116,6 @@ def debug(self):
 
         while (True):
             err = None
-            self.n = 2  # Initial Number of cuts
```
maliberty, line 113: You are changing the default from 2 to 1 - why?

oharboe: good catch. I have this set to draft until I can test it....

maliberty (approved):

PR #4556, opened 2024-01-19
odb: add MASK support for special nets

gadfort:
Adds:
- DEFin and DEFout support for MASK for both VIA and LAYER in Special nets + added to CI testing for read_def58 and write_def58

Fixes:
- Ensure dbWireShapeType is initialized in the `const char*` constructor
- Fixes reading of RECTs in DEFin for special nets to preserve their TYPE and SHAPETYPE (+ CI check for this)


src/odb/src/defin/definSNet.cpp
```diff
@@ -516,6 +537,9 @@ void definSNet::pathViaArray(const char* via_name,
                                     numY,
                                     stepX,
                                     stepY,
+                                    0,
+                                    0,
+                                    0,
```
QuantamHD, line 542: Any time you have constants like these it's good to append the param name as a block comment.

```c++
/*x=*/ 0,
/*y=*/ 0,
/*z=*/ 0,

```

src/odb/src/defin/definSNet.cpp
```diff
@@ -491,6 +509,9 @@ void definSNet::pathViaArray(const char* via_name,
...
```
QuantamHD, line 514: Param names as comments

QuantamHD (changes requested):

src/odb/src/db/dbBox.cpp
```diff
@@ -781,19 +796,15 @@ dbBox* dbBox::create(dbBPin* bpin_,
         layer_->getName(),
         layer_id);
   }
-  if (mask >= 4) {
-    bpin->getLogger()->error(utl::ODB, 434, "Mask must be between 0 and 3.");
-  }
-  if (layer_ == nullptr && mask != 0) {
-    bpin->getLogger()->error(
-        utl::ODB, 435, "Mask must be 0 when no layer is provided.");
-  }
```
maliberty, line 790: should this be in _dbBox::checkMask ?

QuantamHD (approved):

src/odb/src/db/dbSBox.h
```diff
@@ -46,14 +46,25 @@ class dbIStream;
...
 
+struct _dbSBoxFlags
+{
+  dbWireShapeType::Value _wire_type : 6;
+  uint _direction : 2;  // 0 = undefiend, 1 = horizontal, 2 = vertical, 3 =
+                        // octilinear
+  uint _via_bottom_mask : 2;
+  uint _via_cut_mask : 2;
+  uint _via_top_mask : 2;
+  uint _spare_bits : 18;
```
maliberty, line 65: I don't think you actually need to rev the schema and add this backward compatibility as you are just converting spare bits to used.  The spare bits are 0 already in old dbs.

src/odb/src/db/dbBox.cpp
gadfort: got lost in the shuffle, restored now.

maliberty (approved):

PR #4557, opened 2024-01-19
mpl2 & odb: move write_macro_placement command from mpl2 to odb

### LINKED ISSUE #4511: write_macro_placement writes out empty file
### Describe the bug

```
$ make DESIGN_CONFIG=designs/asap7/mock-array/config.mk floorplan
[deleted]
$ make DESIGN_CONFIG=designs/asap7/mock-array/config.mk open_floorplan
[deleted]
OpenROAD v2.0-11723-gde78d0e07 
[deleted]
openroad> write_macro_placement test.tcl
openroad> 
$ cat test.tcl
$ ls -l test.tcl
$ wc -l test.tcl 
0 test.tcl
```

### Expected Behavior

Write out macro placement regardless of how the macros were placed: manually, classic mpl and mpl2.

### Environment

```shell
OpenROAD v2.0-11723-gde78d0e07
```

### To Reproduce

See above.

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
Fix #4511 

The idea here is that the mpl2 flow's flag `-write_macro_placement` is kept and continues to use the function `HierRTLMP::writeMacroPlacement(const std::string& file_name)`. The flag will only generate the placement of the macros that were actually placed by mpl2, differently from the command write_macro_placement, which will generate the placement of all the macros in the design.

Instead of writing the file inside C++, we return a string and write the file in .tcl.

Adding a unit test for this new utility based on a simple DEF with only fake macros.


src/odb/src/db/dbBlock.cpp
```diff
@@ -3453,6 +3453,35 @@ void dbBlock::writeGuides(const char* filename) const
...
+  float y = 0.0f;
+
+  for (odb::dbInst* inst : this->getInsts()) {
+    if (inst->isBlock()) {
+      x = (inst->getLocation().x()) / dbu;
+      y = (inst->getLocation().y()) / dbu;
+
+      out << "place_macro -macro_name " << inst->getName() << " -location {"
+          << x << " " << y << "} -orientation " << inst->getOrient().getString()
+          << '\n';
```
rovinski, line 3480: Probably more concise with a fmt string?

src/odb/src/db/dbBlock.cpp
```diff
@@ -3453,6 +3453,35 @@ void dbBlock::writeGuides(const char* filename) const
   guide_file.close();
 }
 
+void dbBlock::writeMacroPlacement(const std::string& file_name)
```
QuantamHD, line 3456: Prefer taking a ostream instead of a string. That way you can unit test this without file IO.

QuantamHD (changes requested):

src/odb/src/db/dbBlock.cpp
QuantamHD: I'm going to take that back actually. Since this is getting automatically mapped into SWIG. I think it's potentially better to return a string instead of a write command, and then implement the file writing in TCL.

QuantamHD (changes requested):

QuantamHD:
There's also something kind of odd about writing macro placement commands in ODB which is generic to TCL which is just one wrapper around odb.

AcKoucher:
@QuantamHD I see your point, but then what would be the purpose of writing the macro placement? It seems like the main idea is to be able to just paste the generated commands and alter something manually if needed. 

And, as write_macro_placement does not rely on mpl2 internals, it seems reasonable for it to live in odb.

QuantamHD:
I think this is future work, but we probably want to add a enum parameter for the particular format we want the string in. However, I think keeping the method as a string return is preferable and doing the file IO in TCL.

Testing these methods in C++ is much harder when you have to deal with file io, and my general preference is for C++ unit tests over integration tests.

oharboe (changes requested):
Can we have a single write_macro_placement and add a comment in the generated .tcl about origin of placement?

QuantamHD:
@oharboe I think that should be handled as a separate enhancement. That would require keeping track of that metadata in odb, and I think that's beyond the scope of this PR.

oharboe:
> @oharboe I think that should be handled as a separate enhancement. That would require keeping track of that metadata in odb, and I think that's beyond the scope of this PR.

That works for me. One step at the time...

I was just thinking that the rtlmp write macro arg is superfluous at this point.... This is information one can find out from the logs.

If automation is required, then perhaps such information is better collected out of band of the .odb file?

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -6371,6 +6371,11 @@ void HierRTLMP::setMacroPlacementFile(const std::string& file_name)
 
 void HierRTLMP::writeMacroPlacement(const std::string& file_name)
 {
+  logger_->warn(MPL,
+                39,
+                "The flag -write_macro_placement is deprecated. Use the .tcl "
+                "command -write_macro_placement instead.");
```
oharboe, line 6377: spelling "command -write_macro_placement instead" => "command write_macro_placement instead"

src/odb/src/db/odb.tcl
```diff
@@ -533,5 +533,9 @@ proc write_macro_placement { args } {
     utl::error ODB 439 "No design loaded. Cannot write macro placement."
   }
   set block [$chip getBlock]
-  $block writeMacroPlacement $file_name
+  set macro_placement [$block generateMacroPlacementString]
```
oharboe, line 536: much nicer breaking out the generateMacroPlacementString as an api function.

oharboe (changes requested):
spelling nit

src/odb/src/db/dbBlock.cpp
```diff
@@ -3453,6 +3453,31 @@ void dbBlock::writeGuides(const char* filename) const
   guide_file.close();
 }
 
+std::string dbBlock::generateMacroPlacementString()
```
maliberty, line 3456: This method is written entirely in terms of the public API and therefore doesn't need to be a member function.  It should be moved to a utility function.

maliberty:
This could use a unit test

maliberty (approved):

Issue #4548, opened 2024-01-18
Global place: Command terminated by signal 11 


oharboe:
### Describe the bug

This is from the MegaBoom design. Very slightly changed initial conditions and global place is now crashing.

1. untar https://drive.google.com/file/d/1ZHu7RxqYa_1FMVhqgps9Uxr3LPzZJA0l/view?usp=sharing
2. `./run-me-ChipTop-asap7-flat.sh`

```
OpenROAD v2.0-11865-g6471c0995 
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
global_placement -density 0.25 -pad_left 2 -pad_right 2 -routability_driven -timing_driven
[INFO GPL-0002] DBU: 1000
[INFO GPL-0003] SiteSize: 54 270
[INFO GPL-0004] CoreAreaLxLy: 2052 2160
[INFO GPL-0005] CoreAreaUxUy: 1497960 1497960

[deleted]
[NesterovSolve] Iter: 400 overflow: 0.400787 HPWL: 24931330112
[NesterovSolve] Iter: 410 overflow: 0.353168 HPWL: 24745674434
[NesterovSolve] Iter: 420 overflow: 0.305717 HPWL: 24289745133
[INFO GPL-0100] worst slack -3.76e-07
[INFO GPL-0103] Weighted 174499 nets.
[NesterovSolve] Iter: 430 overflow: 0.275646 HPWL: 23622015587
[NesterovSolve] Iter: 440 overflow: 0.244673 HPWL: 22883279513
[NesterovSolve] Iter: 450 overflow: 0.218247 HPWL: 22215706593
[NesterovSolve] Iter: 460 overflow: 0.198184 HPWL: 21474367018
[INFO GPL-0100] worst slack -3.76e-07
[INFO GPL-0103] Weighted 174498 nets.
[INFO GPL-0075] Routability numCall: 1 inflationIterCnt: 1 bloatIterCnt: 0
./run-me-ChipTop-asap7-flat.sh: line 7: 1596838 Segmentation fault      (core dumped) openroad -no_init ${SCRIPTS_DIR}/global_place.tcl
```

### Expected Behavior

No crash

### Environment

```shell
OpenROAD v2.0-11865-g6471c0995
```

### To Reproduce

See above

### Additional Context

Will try to whittle down...

```
[...]

PR #4558, opened 2024-01-21
deltaDebug: --multiplier increases number of cuts, --timeout can be specified explicitly

oharboe:
This is useful when there is an .odb file that needs very fine level cuts. These are lessons learned when trying to whittle down #4548.

Also added a basic test script. Takes a few minutes to run, worth hooking up to CI . Only needs to run if deltaDebug.py is changed.

To run smoke-test:

```
cd flow
../tools/OpenROAD/etc/test_delta_debug.sh
```

### DIFF
--- a/etc/deltaDebug.py
+++ b/etc/deltaDebug.py
@@ -32,15 +32,17 @@ import errno
 import enum
 
 persistence_range = [1, 2, 3, 4, 5, 6]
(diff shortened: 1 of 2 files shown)


vvbandeira (approved):
Added to CI ToDo's.

PR #4559, opened 2024-01-21
ord: provide access functions to version information to avoid recompiling modules on commit changes

gadfort:
No functional changes.

Changes:
- Moves OpenROAD version functionality to the OpenROAD object and removing the dependency on Version.hh everywhere except for the openroad object.

### DIFF
--- a/include/ord/OpenRoad.hh
+++ b/include/ord/OpenRoad.hh
@@ -247,8 +247,8 @@ class OpenRoad
   void addObserver(OpenRoadObserver* observer);
   void removeObserver(OpenRoadObserver* observer);
 
-  const char* getVersion() const;
-  const char* getGitDescribe() const;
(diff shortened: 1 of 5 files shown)


include/ord/OpenRoad.hh
```diff
@@ -247,6 +247,9 @@ class OpenRoad
   void addObserver(OpenRoadObserver* observer);
   void removeObserver(OpenRoadObserver* observer);
 
+  const char* getVersion() const;
+  const char* getGitDescribe() const;
```
maliberty, line 251: These could be static methods as they don't depend on any memberes afaik.

gadfort: makes sense, I'll switch them to static

maliberty (approved):

PR #4565, opened 2024-01-22
gui: make IO pins honor signal and layer settings

gadfort:
Changes:
- gui bterms only visible when Net is visible
- gui bterms only visible when layer is visible

I don't recall if there is a particular reason these pins do not honor the same constraints but without control it makes it hard to inspect pins on different layers and of different signal types

### DIFF
--- a/src/gui/src/renderThread.cpp
+++ b/src/gui/src/renderThread.cpp
@@ -1483,6 +1483,9 @@ void RenderThread::drawIOPins(Painter& painter,
     if (restart_) {
       break;
     }
+    if (!viewer_->isNetVisible(term->getNet())) {
+      continue;
+    }
     for (odb::dbBPin* pin : term->getBPins()) {
       odb::dbPlacementStatus status = pin->getPlacementStatus();
       if (!status.isPlaced()) {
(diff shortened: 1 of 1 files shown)


AcKoucher:
@gadfort I didn't realize that  we should honor the layer/net constraints when I implemented the alternative of displaying the pins' geometries by using the display control.

gadfort:
@AcKoucher I don't recall if there was a reason why they weren't originally set up that way. I can definitely see the use case where you want to be able to see them without the routing, so maybe we need additional controls if that is important @maliberty

maliberty:
See the comment https://github.com/The-OpenROAD-Project/OpenROAD/issues/4050#issuecomment-1758558032

The user wants to distinguish the pin shape from the other shapes in the net.

maliberty:
Perhaps we should have a shape type control section.  I sometimes want to distinguish segs vs via but we have no control for that either.

gadfort:
I like the idea of shape type control.
We have:
- Segments / Routing
- Vias
- Pins
- Special Routing

Anything else?

maliberty:
> Anything else?

Move fills into there

Issue #4436, opened 2023-12-22
When CORE_UTILIZATION is used, use a minimum area such that io pins fit


oharboe:
### Description

Building on https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1722, it would be useful to use CORE_UTILIZATION, but that the area is increased so that the pins, respecting io constraints, can fit.

### Suggested Solution

CORE_UTILIZATION sets a minimum size, but the die area is increased such that io pins fit.

### Additional Context

...

oharboe:
@maliberty Perhaps the more idiomatic solution for openroad and ORFS is to print actionable error: increase CORE_UTILIZATION from xx to yy?

Settings specified should be used as is and the user should adjust if necessary based on information in errors, this reduces coupling between stages in the flow amongst other benefits.

maliberty:
It might be better to adjust CORE_MARGIN rather than spread things out more just to get more perimeter.

maliberty:
Though strictly speaking ppl doesn't know if you used CORE_UTILIZATION/CORE_MARGIN or DIE_AREA in ifp so we might need to suggest both.

oharboe:
> Though strictly speaking ppl doesn't know if you used CORE_UTILIZATION/CORE_MARGIN or DIE_AREA in ifp so we might need to suggest both.

Makes sense. An error message that it didnt fit would be an improvement even if much more specific advise in the error isnt approperiate.

eder-matheus:
@oharboe Maybe I'm missing something, but it´s unclear what the request is here. If I remember correctly, PPL already has error messages that suggest how much the edges should be increased.

Would the idea here be to suggest alternatives for CORE_UTILIZATION, CORE_MARGIN, and DIE_AREA when PPL can't fit the pins in the boundaries?

oharboe:
> @oharboe Maybe I'm missing something, but it´s unclear what the request is here. If I remember correctly, PPL already has error messages that suggest how much the edges should be increased.

If it shows a) how much it is today b) how much it would need to be as well as how many pins that fit and how many that were requested to fit, then the user can carry the water the rest of the way. There are going to be differences in the action to take depending on build system and how it is configured(ORFS, SiliconCompiler, etc.)

eder-matheus:
[...]

PR #4566, opened 2024-01-22
ppl: add change suggestions in the error messages

eder-matheus:
Partial fix for https://github.com/The-OpenROAD-Project/OpenROAD/issues/4436

### DIFF
--- a/src/ppl/src/IOPlacer.cpp
+++ b/src/ppl/src/IOPlacer.cpp
@@ -682,9 +682,8 @@ int IOPlacer::micronsToDbu(double microns)
 
 double IOPlacer::areaDbuToMicrons(int64_t dbu)
 {
-  return static_cast<double>(dbu)
-         / ((getBlock()->getDbUnitsPerMicron()
-             * (getBlock()->getDbUnitsPerMicron())));
+  const int units_per_micron = getBlock()->getDbUnitsPerMicron();
+  return static_cast<double>(dbu) / (units_per_micron * units_per_micron);
 }
 
 void IOPlacer::writePinPlacement(const char* file_name)
@@ -931,8 +930,8 @@ void IOPlacer::defineSlots()
       min_dist = std::max(layer_min_dist, min_dist);
     }
 
-    int64_t die_margin = getBlock()->getDieArea().margin();
-    int64_t new_margin
+    const int64_t die_margin = getBlock()->getDieArea().margin();
+    const int64_t new_margin
         = computeIncrease(min_dist, regular_pin_count, die_margin) + die_margin;
 
     logger_->error(
@@ -1794,8 +1793,8 @@ void IOPlacer::initConstraints(bool annealing)
     getPinsFromDirectionConstraint(constraint);
     constraint.sections = createSectionsPerConstraint(constraint);
     int num_slots = 0;
-    int region_begin = constraint.interval.getBegin();
-    int region_end = constraint.interval.getEnd();
+    const int region_begin = constraint.interval.getBegin();
+    const int region_end = constraint.interval.getEnd();
     const std::string region_edge
         = getEdgeString(constraint.interval.getEdge());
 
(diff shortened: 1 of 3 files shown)


src/ppl/src/IOPlacer.cpp
```diff
@@ -907,12 +921,26 @@ void IOPlacer::defineSlots()
...
+    }
+
+    int64_t die_margin = getBlock()->getDieArea().margin();
+    int64_t increase = computeIncrease(min_dist, regular_pin_count, die_margin);
+
     logger_->error(PPL,
                    24,
                    "Number of IO pins ({}) exceeds maximum number of available "
-                   "positions ({}).",
+                   "positions ({}). Increase the die width or height in {}um.",
```
oharboe, line 940: be specific here, if possible. Is it width or height?

The user now has to guess

oharboe: increase from ... to .... will be less ambiguous

src/ppl/src/IOPlacer.cpp
```diff
@@ -1775,9 +1808,13 @@ void IOPlacer::initConstraints(bool annealing)
         logger_->warn(PPL,
                       110,
                       "Constraint has {} pins, but only {} available slots.\n"
-                      "Increase the region in at least {}um.",
+                      "Increase the region {}-{} on the {} edge in at least "
+                      "{}um.",
```
oharboe, line 1812: to at least

oharboe (changes requested):
nits

src/ppl/src/IOPlacer.cpp
eder-matheus: I also think of something like "Increase the die perimeter from {} to {}". Does it make sense, in your opinion?

eder-matheus: It is more generic, but it gives more room for the user to choose where they want to increase. It can be only width, only height, or divide the increase value between both.

src/ppl/src/IOPlacer.cpp
eder-matheus: "to at least" would imply that the message would inform the new total length of the region, right? The way it is currently written informs how much the user needs to increase the region.

oharboe: from ... to ...

src/ppl/src/IOPlacer.cpp
oharboe: i see, so not constrained to an edge here.

eder-matheus: Yes. Mainly because in this error, there is no room for the pins, independent of where they're constrained to. So the suggested increase value is to ensure that exists enough positions to place the pins in the boundaries.

src/ppl/src/IOPlacer.cpp
```diff
@@ -1763,22 +1794,36 @@ void IOPlacer::initConstraints(bool annealing)
     getPinsFromDirectionConstraint(constraint);
     constraint.sections = createSectionsPerConstraint(constraint);
     int num_slots = 0;
+    int region_begin = constraint.interval.getBegin();
+    int region_end = constraint.interval.getEnd();
```
maliberty, line 1798: const

src/ppl/src/IOPlacer.cpp
```diff
@@ -907,12 +921,29 @@ void IOPlacer::defineSlots()
...
+      min_dist = std::max(layer_min_dist, min_dist);
+    }
+    for (int layer_idx : hor_layers_) {
+      const int layer_min_dist = core_->getMinDstPinsY()[layer_idx];
+      min_dist = std::max(layer_min_dist, min_dist);
+    }
+
+    int64_t die_margin = getBlock()->getDieArea().margin();
+    int64_t new_margin
+        = computeIncrease(min_dist, regular_pin_count, die_margin) + die_margin;
```
maliberty, line 936: const

src/ppl/src/IOPlacer.cpp
```diff
@@ -680,6 +680,13 @@ int IOPlacer::micronsToDbu(double microns)
   return (int64_t) (microns * getBlock()->getDbUnitsPerMicron());
 }
 
+double IOPlacer::areaDbuToMicrons(int64_t dbu)
+{
+  return static_cast<double>(dbu)
+         / ((getBlock()->getDbUnitsPerMicron()
+             * (getBlock()->getDbUnitsPerMicron())));
```
maliberty, line 687: call once & use a const var

maliberty (approved):
Minor points - please update and you can merge

PR #4569, opened 2024-01-22
odb: derating relocation from GRT to dbTechLayer

gudeh:
- Implemented the insertion of derating (layer adjustment) into RUDY, using an average value from the routing layers.
      - Note: RUDY does not have access to layer information. Instead, derating values for each layer are set within GRT.
- Moved derating from GRT to dbTechLayer through the use of a code generator.


src/odb/src/zutil/util.cpp
```diff
@@ -130,6 +130,32 @@ void RUDYCalculator::calculateRUDY()
...
   }
+
+  float rudy_adjustment = 0.0;
+  odb::dbTech* tech = block_->getTech();
+  for (int l = 1; l <= tech->getRoutingLayerCount(); l++) {
+    odb::dbTechLayer* tech_layer = tech->findRoutingLayer(l);
+    if (tech_layer == nullptr) {
+      continue;
+    }
+    rudy_adjustment += tech_layer->getLayerAdjustment();
```
maliberty, line 141: Averaging the adjustments isn't precise.  The number of tracks varies by layer so the effect of a given derate is larger on layers with more tracks.

gudeh: Alright. What do you think of using the greatest value then?

src/grt/src/GlobalRouter.cpp
```diff
@@ -1132,8 +1130,8 @@ void GlobalRouter::computeUserGlobalAdjustments(int min_routing_layer,
     return;
 
   for (int l = min_routing_layer; l <= max_routing_layer; l++) {
-    if (adjustments_[l] == 0) {
-      adjustments_[l] = adjustment_;
+    if (db_->getTech()->findRoutingLayer(l)->getLayerAdjustment() == 0.0) {
+      db_->getTech()->findRoutingLayer(l)->setLayerAdjustment(adjustment_);
```
QuantamHD, line 1134: @maliberty How much null checking do we need to do on such lines. Is it guaranteed that all these pointers will exist i.e. because we checked it elsewhere?

src/odb/src/zutil/util.cpp
maliberty: A pitch weighted average would be closer (though still subject to some round-off).

src/grt/src/GlobalRouter.cpp
maliberty: We should have verified the min/max routing layers when they were set.  I don't think we need to check here.

src/odb/src/zutil/util.cpp
gudeh: The idea is that layers with higher pitch values should weight more on the derating average for RUDY, correct?

maliberty: lower pitch implies more tracks so you want to more weight on lower pitch

src/grt/src/GlobalRouter.cpp
gudeh: Actually, we get multiple test fails in GRT if this condition is removed. If it is kept, all tests from GRT pass.

src/odb/src/zutil/util.cpp
gudeh: You can check the behavior for [three methods on this link](https://drive.google.com/drive/folders/1T-ECrPp1tulA_uU4xfXOWSH5nAK45KcJ?usp=sharing):  1. Using the maximum derating value across all layers, 2. Calculating a simple average of the derating values across layers, and 3. the current implementation on this PR, with a weighted average based on the pitch value of each layer.

gudeh: You can see [on this link](https://docs.google.com/spreadsheets/d/1FpQRPGZ4ImhqMTdEhdUEsyUx0nXm6fZtQTUh8IQDWOg/edit#gid=0) that the correlation metrics (RUDY versus GRT) are quite similar. Their differences are more evident on the heatmap images.

maliberty:
mock array should be quite readily predicted.  I'm wondering why rudy w/weight-avg is showing >100 congestion.

gudeh:
There are two things that influence the > 100% values: the derating itself, and the fact that RUDY is considering the presence of 2 nets per tile per blockage on the macros. Here you can see the case with 0 derating for reference. Notice RUDY values are actually lower when derating is 0.

RUDY: image

GRT: image

There are other examples where RUDY overestimates congestion due to derating, in the current format. I thought of a condition, prior to derating adjustment, to not add too much if the value is already high, but seems strange to do that, any ideas?

maliberty:
When you consider the macro for track usage do you consider derating?  If the macro blocks all the tracks, it should be blocking all the derated tracks not the underated tracks

gudeh:
> When you consider the macro for track usage do you consider derating? If the macro blocks all the tracks, it should be blocking all the derated tracks not the underated tracks

If you look at `RUDYCalculator::calculateRUDY()` you will notice I first calculate RUDY for the nets (different methods for supply versus signal nets), afterwards I loop over the instances to process the macros.  Then only afterwards I adjust every tile based on the derating at dbTechLayer. 

So the derating and the macro blockages are processed separately first the blockages than the derating on all tiles (considering RUDY values are stored on the tiles).

gudeh:
I tried using a sigmoid function to reduce the derating addition on values that are already high. Some examples look better:
-IBEX
ibex_core-nangate45-rudy_ibex_core-nangate45-grt

-Mock-array
MockArray-asap7-rudy_MockArray-asap7-grt

I am executing more examples right now.

maliberty:
@eder-matheus please advise on how derating and macros are handled in grt

eder-matheus:
> @eder-matheus please advise on how derating and macros are handled in grt

Regarding user-defined derating, GRT iterates through every tile of the gcell grid and applies the reduction percentage over the current resource value. With that, we avoid over-reducing the tile resources because it doesn't reduce anymore when it reaches zero.

For the macro obstructions, it iterates over every macro obstruction, gets the tiles overlapping with the obstruction, and sets the resources to zero when it is completely blocked. GRT will estimate how many tracks are blocked to define the reduction value when it is partially blocked.

maliberty:
@eder-matheus do you think we can share the code/logic with rudy so there is one way to handle macros?

eder-matheus:
> @eder-matheus do you think we can share the code/logic with rudy so there is one way to handle macros?

I think @gudeh can explain better, but I understand that rudy doesn't see resources. It only considers the congestion in the tiles. Also, GRT looks at layers, while rudy doesn't have this capability.

I suggested to @gudeh to check if the tile already has 100% congestion and avoid adding the user-defined reductions to these tiles. This is what we do in GRT.

gudeh:
For future reference, we have decided during a sync meeting to implement a generic function in ODB that can be called by GRT, RUDY, or any other process. This approach will provide us with a unified method for calculating congestion for the power grid, macro obstructions, and derating.

The difference among RUDY and GRT should be only for signal nets.

maliberty:
I think this is obsolete in light of #4927 - close?

gudeh:
> I think this is obsolete in light of #4927 - close?

I believe you are correct. With the new code structure I see no reason to relocate derating to odb.

PR #4572, opened 2024-01-23
tap: fix missing cells that overlaps with endcaps 

eder-matheus:
Similar to https://github.com/The-OpenROAD-Project/OpenROAD/pull/4551, but covering the cases where the tapcell overlaps with an endcap on the left side.


eder-matheus:
@maliberty This fixes the last violations caused by missing tapcells.

maliberty (approved):

PR #4576, opened 2024-01-23
grt: remove arrays with static size

eder-matheus:
As discussed in https://github.com/The-OpenROAD-Project/OpenROAD/issues/4548, these static limits can potentially cause the issue. Even if it doesn't solve the issue, getting rid of the static limits is desirable in any case.


maliberty:
There could be runtime performance impact to replacing these fixed arrays with vectors.  Have you checked?  It might be good to reserve space in these to avoid repeated reallocation if needed.

eder-matheus:
@maliberty I've added the reserves to the new vectors. Tomorrow morning, I'll check how the runtime was affected.

rovinski:
Are changes to the guides expected just from changing the underlying data structure?

src/grt/src/fastroute/src/maze3D.cpp
```diff
@@ -395,6 +395,11 @@ int FastRouteCore::copyGrids3D(std::vector<TreeNode>& treenodes,
   const int n1x = treenodes[n1].x;
   const int n1y = treenodes[n1].y;
   const int n1l = treenodes[n1].botL;
+  const int routelen = treeedges[edge_n1n2].route.routelen;
+
+  gridsX_n1n2.reserve(routelen + 1);
```
oharboe, line 400: This is of course a performance optimization only and it doesn't have to be particularly precise... .resize() sets the size of the vector .reserve() only speeds up the reallocation by making that much space available for push_back & resize() later...

eder-matheus:
> Are changes to the guides expected just from changing the underlying data structure?

@rovinski Not really, I'm looking into them. My first guess is that it relates to the congestion of them, which is much larger than the regular unit tests.

eder-matheus:
> @rovinski Not really, I'm looking into them. My first guess is that it relates to the congestion of them, which is much larger than the regular unit tests.

@rovinski Never mind, I've caused confusion when updating these files. For some reason, we have different results for them when running on MacOS, and I accidentally updated the ok files with the run from my Mac. After merging it with the latest master branch, I was able to fix the ok files. My update didn't caused any difference in the end.

eder-matheus:
@maliberty I've completed my tests regarding runtime, and it doesn't have a great impact, even with the larger designs. I will start a secure-ci just to ensure nothing will break, but this PR should be ready to merge.

maliberty (approved):
I'll approve and you can merge once the CI passes

oharboe:
@maliberty @eder-matheus It still crashes (ca. 8 hours in in the debugger...).

So here we are allocating on the order of:

4  * (448826 + 224413 * 2) = 3.5mByte on the stack.

The maximum on Ubuntu 23.10 (by default) in kBytes is 8mBytes. It surprised me that it is this low, but it makes sense if you want to have a ton of threads in an application.

```
$ ulimit -s
8192
```

Let us see if we are in the neighbourhood of running out of stack:

Reading rsp at top and bottom of stack trace:

```
register read rsp
     rsp = 0x00007ffcdd0d2ec0
register read rsp
     rsp = 0x00007ffcdc8d6630
```

The difference is 0x7FC890.

Aha! Out of stack...

Solution: stop allocating non-trivial arrays on the stack... Here, switch to std::vector...



image

oharboe:
@eder-matheus I think this PR could go in as one concern and the next problem can be fixed in a separate PR?

eder-matheus:
> @eder-matheus I think this PR could go in as one concern and the next problem can be fixed in a separate PR?

Yes, that's my plan. I will update the array you mentioned this afternoon and create a PR for it.

oharboe:
> > @eder-matheus I think this PR could go in as one concern and the next problem can be fixed in a separate PR?
> 
> Yes, that's my plan. I will update the array you mentioned this afternoon and create a PR for it.

Do you know of other places large arrays are allocated on the stack...? Is there a way to search for them?

eder-matheus:
> Do you know of other places large arrays are allocated on the stack...? Is there a way to search for them?

I'm looking at the code for more large arrays. I don't know an easy way to do this search.

maliberty:
Use -Wvla to get:

foo.cc:3:10: warning: variable length array 'x' is used [-Wvla]

At least it will be easier to see what needs inspecting

maliberty:
Note that once we can use rudy to replace grt in gpl this sort of thing won't be a big issue.  This only happens because we haven't been through buffering yet and have some massive net.

oharboe:
> Note that once we can use rudy to replace grt in gpl this sort of thing won't be a big issue. This only happens because we haven't been through buffering yet and have some massive net.

👍 Good to do the code cleanup though.

PR #4581, opened 2024-01-24
etc: add qtcharts to Dependency installer

gadfort:
Adds:
- qtcharts to Ubuntu and RHEL/CentOS installs


maliberty:
There is no qtcharts for centos7.  Will this change work on that?

gadfort:
@maliberty I don't know about CentOS7, I know it's there for RHEL8, so I could limit it to just CentOS8 and RHEL8

maliberty:
> @maliberty I don't know about CentOS7, I know it's there for RHEL8, so I could limit it to just CentOS8 and RHEL8

Please do.

PR #4588, opened 2024-01-25
add report on cells count

arthurjolo:
Fixes [#1695](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1695)

This PR adds a command (report_cell_count) to report the cells count before CTS, after CTS and after Filler cells have been added.


src/odb/src/db/odb.tcl
```diff
@@ -465,6 +465,102 @@ proc report_voltage_domains {} {
...
+  set time_repr_buffer "0"
+  
+  utl::report "\nReporting Cells:"
+
+  foreach inst $insts {
+    set master [$inst getMaster]
+    set master_type [$master getType]
+    set source_type [$inst getSourceType]
+    set cell_name [$master getName]
+    if { bool([$master isBlock]) == bool(true) } {
```
maliberty, line 497: can you write
```
    if { [$master isBlock] } {
```
?

arthurjolo: I don't think tcl supports this, but I can try.

arthurjolo: Yes, It works, I will change for the other cases as well.

QuantamHD:
Should we be writing this in TCL? It seems like we should add something to the C++ api that returns a map<cell_type, int> or something like that instead.

maliberty:
I agree this shouldn't be in TCL as it won't be available in python.  You could separate the stats summary, as Ethan suggests, from the actual reporting. 

I don't think the "Cells before/after" messages are needed.

arthurjolo:
A small change now that the code is in C++, the command to report the cells count is now: `sta::report_cell_count`

src/dbSta/src/dbSta.cc
```diff
@@ -270,6 +271,129 @@ std::set<dbNet*> dbSta::findClkNets(const Clock* clk)
   return clk_nets;
 }
 
+std::map<dbSta::InstType, int> dbSta::getInstancesType()
```
QuantamHD, line 274: I suggest change the name of this function countInstancesByType. It's more descriptive of it's actual functionality.

QuantamHD:
> A small change now that the code is in C++, the command to report the cells count is now: `sta::report_cell_count`

Nice! I like the change a lot.

src/dbSta/src/dbSta.cc
arthurjolo: I agree, this is a better name.

maliberty:
This looks very close to gui's DbInstDescriptor::getInstanceType & DbInstDescriptor::getInstanceTypeText.  I'd like to avoid having two pieces of code that do nearly, but not quite, the same thing.

Could we move the guts of those functions to your code and call them from GUI?  The count function would iterate and sum by calling this new method per inst.

arthurjolo:
> This looks very close to gui's DbInstDescriptor::getInstanceType & DbInstDescriptor::getInstanceTypeText. I'd like to avoid having two pieces of code that do nearly, but not quite, the same thing.
> 
> Could we move the guts of those functions to your code and call them from GUI? The count function would iterate and sum by calling this new method per inst.

@maliberty I belive this last commit  addresses this. The DbInstDescriptor::getInstanceTypeText I maintained as it was in DbInstDescriptor but using the new dbSta::getInstanceType().

maliberty:
Please show a sample output

arthurjolo:
> Please show a sample output

An output example would be:

Reporting Cells count:
  Macro: 6
  Fill: 523662
  Tapcell: 58169
  Antenna: 412
  Buffer/inverter: 242
  Clock buffer/inverter: 81
  Buffer/inverter from timing repair: 288
  Sequential: 525
  Combinational: 1928

precisionmoon:
> > Please show a sample output
> 
> An output example would be:
> 
> ```
> Reporting Cells count:
>   Macro: 6
>   Fill: 523662
>   Tapcell: 58169
>   Antenna: 412
>   Buffer/inverter: 242
>   Clock buffer/inverter: 81
>   Buffer/inverter from timing repair: 288
>   Sequential: 525
>   Combinational: 1928
> ```

Does "Combinational" include buffers, clock buffers, buffers from timing repair?
The report can be made clearer with something like the following:

Cell usage report:
  Macros: 6
  Fill cells: 523662
  Tap cells: 58169
  Antenna cells: 412
  Data buffers/inverters: 242
  Clock buffers/inverters: 81
  Timing repair data buffers/inverters: 288
  Sequential cells: 525
  Complex combinational cells: 1928

Alternate Tcl command name like report_cell_usage may be good also.

arthurjolo:
> Complex

Combinational do not include buffers/inverters. I changed the report.
That wxample now is:
 
```
Cell usage report:
  Macros: 6
  Fill cells: 523662
  Tap cells: 58169
  Antenna cells: 412
  Buffer/inverters: 242
  Clock buffer/inverters: 81
  Timming Repair Buffer/inverters: 288
  Sequential cells: 525
  Complex combinational cells: 1928
```

precisionmoon:
> > Complex
> 
> Combinational do not include buffers/inverters. I changed the report. That wxample now is:
> 
> ```
> Cell usage report:
>   Macros: 6
>   Fill cells: 523662
>   Tap cells: 58169
>   Antenna cells: 412
>   Buffer/inverters: 242
>   Clock buffer/inverters: 81
>   Timming Repair Buffer/inverters: 288
>   Sequential cells: 525
>   Complex combinational cells: 1928
> ```

Just a small typo:  "Timming" should read "Timing"

maliberty:
Please add a unit test and command documentation.

arthurjolo:
@maliberty I am having some difficulties on finding where to add the documentation for the command. The dbSta directory doesn’t have a README.md file, is there another directory I can add the documentation?

precisionmoon (approved):

maliberty (approved):

PR #4591, opened 2024-01-26
grt: use vectors instead of large arrays

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/4548


eder-matheus:
@maliberty secure-ci is passing.

maliberty (approved):

PR #4604, opened 2024-01-30
Optimize rebuffering

kiryk:
This PR optimizes `RepairSetup::rebufferBottomUp` with three types of changes:
1. Vectors are preallocated wherever their size can be predicted
2. Slack values, used in sorting, are precalculated.
3. Option pruning is modified from pessimistically quadratic to linear complexity.

I gathered averaged outcomes from 10 runs on chosen designs (`cts` target):
 Version \ Design | Ariane | BlackParrot | TinyRocket | Ibex
-|-|-|-|-
Unoptimized | 63m40s | 3m39s | 0m27s | 1m24s
Optimized | 12m32s | 2m57s | 0m21s | 0m28s

### DIFF
--- a/src/rsz/src/Rebuffer.cc
+++ b/src/rsz/src/Rebuffer.cc
@@ -216,7 +216,14 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
     const BufferedNetSeq& Z1 = rebufferBottomUp(bnet->ref(), level + 1);
     const BufferedNetSeq& Z2 = rebufferBottomUp(bnet->ref2(), level + 1);
     BufferedNetSeq Z;
-    Z.resize(Z1.size()*Z2.size());
+
+    size_t size = Z1.size()*Z2.size();
+    // This assumption is used only in the final loop
+    // but we can quit as early as we know there's nothing to do here.
+    if (size == 0)
+      return Z;
+    Z.resize(size);
+
     // Combine the options from both branches.=
     for (size_t i = 0, base = 0; i < Z1.size(); i++, base += Z2.size()) {
       const BufferedNetPtr& p = Z1[i];
@@ -235,8 +242,8 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
     }
     // Prune the options if there exists another option with
     // larger required and smaller capacitance.
-    // This is fanout^2.
-    // Presort options to hit better options sooner.
+    // This is fanout*log(fanout) if options are
+    // presorted to hit better options sooner.
     sort(Z.begin(),
          Z.end(),
          [this](const BufferedNetPtr& option1, const BufferedNetPtr& option2) {
@@ -259,26 +266,24 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
 
            return false;
          });
-    int si = 0;
-    for (size_t pi = 0; pi < Z.size(); pi++) {
+    float Lsmall = Z[0]->cap();
+    size_t si = 1;
+    // Remove options by shifting down with index si.
+    // Because the options are sorted we don't have to look
+    // beyond the first option. We also know that slack
+    // is nonincreasing, so we can remove everything that has
+    // higher capacitance than the lowest found so far.
+    for (size_t pi = si; pi < size; pi++) {
       const BufferedNetPtr& p = Z[pi];
       float Lp = p->cap();
-      // Remove options by shifting down with index si.
-      si = pi + 1;
-      // Because the options are sorted we don't have to look
-      // beyond the first option.
-      for (size_t qi = pi + 1; qi < Z.size(); qi++) {
-        const BufferedNetPtr& q = Z[qi];
-        float Lq = q->cap();
-        // We know Tq <= Tp from the sort so we don't need to check req.
-        // If q is the same or worse than p, remove solution q.
-        if (fuzzyLess(Lq, Lp)) {
-          // Copy survivor down.
-          Z[si++] = q;
-        }
+      // If Tp is the same or worse than Tsmall, remove solution p.
+      if (fuzzyLess(Lp, Lsmall)) {
+        // Otherwise copy the survivor down.
+        Z[si++] = p;
+        Lsmall = Lp;
       }
-      Z.resize(si);
     }
+    Z.resize(si);
     return Z;
   }
   case BufferedNetType::load: {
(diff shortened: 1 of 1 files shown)


maliberty:
I see these failures
```
array_ins_delay (tcl) *ERROR* segmentation violation
insertion_delay (tcl) *ERROR* segmentation violation
```
they were broken in master a while ago and then fixed.  I suspect you need to merge master into your branch.

QuantamHD:
Wow! Amazing work

src/rsz/src/Rebuffer.cc
```diff
@@ -305,8 +306,10 @@ RepairSetup::addWireAndBuffer(const BufferedNetSeq& Z,
                               int level)
 {
   BufferedNetSeq Z1;
+  Z1.resize(Z.size());
   Point wire_end = bnet_wire->location();
-  for (const BufferedNetPtr& p : Z) {
+  for (int i = 0; i < Z.size(); i++) {
+    const BufferedNetPtr& p = Z[i];
```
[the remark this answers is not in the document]
kiryk: Nice, right.

src/rsz/src/Rebuffer.cc
```diff
@@ -227,55 +240,44 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
...
+        slacks[junc.get()] = slackPenalized(junc);
+        Z[base + j] = std::move(junc);
       }
     }
     // Prune the options if there exists another option with
     // larger required and smaller capacitance.
-    // This is fanout^2.
-    // Presort options to hit better options sooner.
+    // This is fanout*log(fanout) if options are
+    // presorted to hit better options sooner.
```
maliberty, line 250: I think both comments are wrong.  If the fanouts are evenly split in two between Z1 & Z2 then Z is size (fanout/2)^2.  The sort is big-O N*logN of N=fanout^2 in that case.

src/rsz/src/Rebuffer.cc
```diff
@@ -227,55 +240,44 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
...
-        const BufferedNetPtr& q = Z[qi];
-        float Lq = q->cap();
-        // We know Tq <= Tp from the sort so we don't need to check req.
-        // If q is the same or worse than p, remove solution q.
-        if (fuzzyLess(Lq, Lp)) {
-          // Copy survivor down.
-          Z[si++] = q;
-        }
+      // If Tp is the same or worse than Tsmall, remove solution p.
+      if (fuzzyLess(Lp, Lsmall)) {
```
maliberty, line 274: The comment doesn't match the code Tp vs Lp and Tsmall vs Lsmall.

src/rsz/src/Rebuffer.cc
```diff
@@ -227,55 +240,44 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
...
+        // Otherwise copy the survivor down.
+        Z[si++] = p;
+        Lsmall = Lp;
```
maliberty, line 277: The previous code did not update Lp so it likely kept many more options.  This is a change in functionality not a pure performance tuning and will require validation if is intended - please confirm.   It sounds reasonable.  I expect this is where most of the performance gain comes from.

src/sta
maliberty, line 1: Is this related to this change?

maliberty (changes requested):

maliberty:
Started a secure CI

src/rsz/src/Rebuffer.cc
kiryk: The complexity is a function of two variables here, but I decided to keep the original intent. Hm, I think we can just avoid the issue saying that the problem is N*logN, without any explicit reference to fanout.

src/rsz/src/Rebuffer.cc
kiryk: Yes, it is intended. I changed it because an improvement of complexity seemed reasonable, but in the designs I tested most of improvement came not from this, but from caching the slack values (and so avoiding redundant floating point calculations).

src/sta
maliberty: I see a bunch of failures that look related to this change in the CI.  I think you should back it out of this PR.

maliberty: I have reverted it in my copy of this and restarted the CI

src/rsz/src/Rebuffer.cc
```diff
@@ -216,9 +216,22 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
...
-    for (const BufferedNetPtr& p : Z1) {
-      for (const BufferedNetPtr& q : Z2) {
+
+    size_t size = Z1.size()*Z2.size();
+    // This assumption is used only in the final loop
+    // but we can quit as early as we know there's nothing to do here.
+    if (size == 0) {
+      return Z;
+    }
+    Z.resize(size);
```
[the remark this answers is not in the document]
kiryk: Done.

src/rsz/src/Rebuffer.cc
```diff
@@ -227,55 +240,44 @@ RepairSetup::rebufferBottomUp(const BufferedNetPtr& bnet,
...
+    // This is fanout*log(fanout) if options are
+    // presorted to hit better options sooner.
     sort(Z.begin(),
          Z.end(),
-         [this](const BufferedNetPtr& option1, const BufferedNetPtr& option2) {
-           Slack slack1 = slackPenalized(option1);
-           Slack slack2 = slackPenalized(option2);
+         [&slacks](const BufferedNetPtr& option1, const BufferedNetPtr& option2) {
+           Slack slack1 = slacks[option1.get()];
+           Slack slack2 = slacks[option2.get()];
```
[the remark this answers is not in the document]
kiryk: Done.

maliberty:
There was some issue in the secure CI with the SQL server going down while this was running but as far as I can see everything passed.

src/rsz/src/Rebuffer.cc
kiryk: The previous code would iterate over the vector and for each element it'd remove its successors with a higher or similar capacitance. If some element is not removed, it must have a lower capacitance, so on the next iteration the inner loop will iterate over the survivors again and challenge them even more. It should be enough to iterate once, while keeping track of the lowest cap so far.

I think If this argument is correct, then both versions should produce the same output.

QuantamHD:
This ready to go @maliberty?

PR #4605, opened 2024-01-30
pdn: ensure tech vias can be snapped to routing tracks

gadfort:
Fixes:
- tech vias not getting snapped to routing tracks
- test case that was supposed to test this

Does not fix:
- Generate rule vias still not snapping to routing tracks


maliberty:
> test case that was supposed to test this

Does it not?

maliberty (approved):

gadfort:
> > test case that was supposed to test this
> 
> Does it not?

The vias aligned to the routing tracks, so the test appeared to be functional, but didn't actually test anything. Now it does.

PR #4613, opened 2024-02-01
Added aarch64 architecture support for Ubuntu 22.04

liav-vorak:
The following has been done:

- Enhanced CMake installation to dynamically handle x86_64 and aarch64 architectures, adjusting the download URL accordingly for improved system flexibility.
- Modified the Or-Tools installation to check and build from source for aarch64 architecture, as prebuilt binaries are not available.
- Modified  part of libQt5Core.so strip in _installUbuntuPackages function, added 'arch' variable to dynamically define machine architecture, which makes the script more adaptive to different architectures.

The modifications has been developed and tested successfully on an AWS Graviton (aarch64) instance using the following specifications: Instance type - t4g.xlarge (4 vCPU, 16 GiB Memory), Storage - 128 GiB.

This pull request is associated with another [PR](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1787) in the OpenROAD-flow-scripts repository. Specifically, adjustments were made in the DependancyInstaller.sh script to accommodate dependencies for the aarch64 architecture. .


etc/DependencyInstaller.sh
```diff
@@ -213,15 +218,30 @@ _installOrTools() {
     if [[ ! -z "${PREFIX}" ]]; then mkdir -p "${PREFIX}"; fi
     cd "${baseDir}"
 
-    orToolsFile=or-tools_${arch}_${os}-${version}_cpp_v${orToolsVersionSmall}.tar.gz
-    wget https://github.com/google/or-tools/releases/download/v${orToolsVersionBig}/${orToolsFile}
     orToolsPath=${PREFIX:-"/opt/or-tools"}
-    if command -v brew &> /dev/null; then
-        orToolsPath="$(brew --prefix or-tools)"
+    if [ "$(uname -m)" == "aarch64" ]; then
+        LIST=($(sudo find / -type f -name "libortools.so*" 2>/dev/null))
```
vvbandeira, line 223: I would like to avoid using sudo in this script. One of the reasons to subdivide the install process for each OS and the common packages (i.e., those installed from source) is to support some users that do not have sudo access. If a function has a command that needs sudo it should be executed with the other `-dev` functions.

vvbandeira (changes requested):

vvbandeira (approved):

PR #4614, opened 2024-02-01
mpl2: fix sequence pair

AcKoucher:
_This fix is to allow the approach that will be used in #4544_

The changes in this PR are to make sure that the sequence pair includes only the clusters / macros that we're actually trying to place during Cluster Placement and Macro Placement.

Changes:
1.  Populate the macros' vector (Soft for cluster placement, Hard for macros placement) in such a manner that its beginning will contain the macros that should be part of the sequence pair;
2. Make all the annealing moves deal with the macros of the sequence pair instead of all the macros used for calculating the penalties;
3. Some renaming to facilitate reading and review 

This draft is to show the approach


src/mpl2/src/SimulatedAnnealingCore.cpp
```diff
@@ -355,65 +366,87 @@ void SimulatedAnnealingCore<T>::packFloorplan()
...
+
+  width_ = accumulated_length[pos_seq_.size() - 1];
 
   // calulate Y position
-  std::vector<int> pos_seq(pos_seq_.size());
-  for (int i = 0; i < macros_.size(); i++) {
-    pos_seq[i] = pos_seq_[macros_.size() - 1 - i];
+  std::vector<int> inverted_pos_seq(pos_seq_.size());
+  for (int i = 0; i < inverted_pos_seq.size(); i++) {
+    inverted_pos_seq[i] = pos_seq_[inverted_pos_seq.size() - 1 - i];
```
maliberty, line 411: reversed_pos_seq feels a bit more descriptive

src/mpl2/src/SimulatedAnnealingCore.cpp
```diff
@@ -510,11 +542,6 @@ void SimulatedAnnealingCore<T>::fastSA()
   if (graphics_) {
     graphics_->startSA();
   }
-  // perturb();  // Perturb from beginning
-  std::iota(pos_seq_.begin(), pos_seq_.end(), 0);
-  std::iota(neg_seq_.begin(), neg_seq_.end(), 0);
-  std::iota(pre_pos_seq_.begin(), pre_pos_seq_.end(), 0);
-  std::iota(pre_neg_seq_.begin(), pre_neg_seq_.end(), 0);
```
maliberty, line 517: What has replaced this?

maliberty:
Go ahead with a CI

src/mpl2/src/SimulatedAnnealingCore.cpp
AcKoucher: this std::iota usage looks unneeded due to the way we initialize the sequences:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/bdda927aed0ece8efa42706bd847024ce9ec5f37/src/mpl2/src/SimulatedAnnealingCore.cpp#L114-L123

maliberty (approved):

Issue #4618, opened 2024-02-03
RUDY needs 25GB for ispd24 ariane133_51 (120k std cells)


stefanottili:
### Describe the bug

openroad needs 25GB to just read/display the smallest of the ispd24 test cases, ariane133_51.def with 120k std cells.

The placement density map looks nice, but 'I think I need a bigger boat' to be able to have a look at the other test cases.

I just recently ran the ispd19_test6 detailed routing with 178K std cells without issues on my 16GB M1, so why does OR all the sudden need 25GB to display 120k placed cells ?

cluster-001.def with 9.8M std cells uses 10GB after read def, but then it keeps allocating memory

### Expected Behavior

a) use less then 25GB to display 120k placed std cells with placement density map
b) give some feedback on how much memory the tool is allocating

### Environment

```shell
OpenROAD v2.0-12068-ge92b2439a as of Feb 3rd 2024

MacOS M1 16GB
```

### To Reproduce

openroad -gui ispd24.or
```
read_lef ../chips/ispd24/lef/Nangate.lef
read_def ../chips/ispd24/def/ariane133_68.def.gz
#read_def ../chips/ispd24/def/ariane133_51.def.gz
#read_def ../chips/ispd24/def/mempool_tile.def.gz
#read_def ../chips/ispd24/def/nvdla.def.gz
#read_def ../chips/ispd24/def/bsg_chip.def.gz
#read_def ../chips/ispd24/def/mempool_group.def.gz
#read_def ../chips/ispd24/def/cluster-001.def.gz
```
https://liangrj2014.github.io/ISPD24_contest/

### Relevant log output

```shell
[INFO ODB-0222] Reading LEF file: ../chips/ispd24/lef/Nangate.lef
[INFO ODB-0223]     Created 22 technology layers
[INFO ODB-0224]     Created 27 technology vias
[INFO ODB-0225]     Created 148 library cells
[INFO ODB-0226] Finished LEF file:  ../chips/ispd24/lef/Nangate.lef
[INFO ODB-0127] Reading DEF file: ../chips/ispd24/def/ariane133_68.def.gz
[INFO ODB-0128] Design: ariane
[INFO ODB-0094] 		Created 100000 Insts
[INFO ODB-0097] 		Created 100000 Nets
[INFO ODB-0130]     Created 435 pins.
[INFO ODB-0131]     Created 120202 components and 732255 component-terminals.
[...]

PR #4621, opened 2024-02-05
gui: allow for non-uniform grids in heatmaps and apply gcells to heatmaps

gadfort:
Changes:
- allow for heatmaps to have non-uniform grids
- add reset to RUDY calculation, otherwise each call just sums on the previous
- add db callbacks to RUDY heatmap to update when instances move
- add mutex to heatmap `ensureMap` to prevent RenderThread from calling it more than once at a time

Closes: #4618

### DIFF
--- a/src/gui/src/heatMap.cpp
+++ b/src/gui/src/heatMap.cpp
@@ -32,6 +32,7 @@
 
 #include "gui/heatMap.h"
 
+#include <algorithm>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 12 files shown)


src/gui/src/heatMapRUDY.cpp
```diff
@@ -57,14 +57,17 @@ double RUDYDataSource::getGridXSize() const
     return default_grid_;
   }
 
-  std::vector<int> grid;
-  gCellGrid->getGridX(grid);
+  int max_gridx = 0;
+  for (int i = 0; i < gCellGrid->getNumGridPatternsX(); i++) {
+    int origin, count, step;
+    gCellGrid->getGridPatternX(i, origin, count, step);
+    max_gridx = std::max(max_gridx, step);
```
maliberty, line 64: In OR's grt that right most gcell may be wider than the others if the design's width is not a gcell multiple (same with top and height).  I think therefore using the mode would be better than that max.

gadfort: I can fix that.
Should I also fix the code in the GRT heatmap to match as well?

maliberty:
What is the memory usage for https://github.com/The-OpenROAD-Project/OpenROAD/issues/4618 with this change?

src/gui/src/heatMap.cpp
```diff
@@ -340,24 +340,36 @@ void HeatMapDataSource::setSettings(const Renderer::Settings& settings)
...
 {
-  const double dbu_per_micron = getBlock()->getDbUnitsPerMicron();
+  const auto x_low_find = std::find_if(
+      map_x_grid_.begin(), map_x_grid_.end(), [&bounds](const int x) {
+        return x >= bounds.xMin();
+      });
+  const auto x_high_find = std::find_if(
+      map_x_grid_.begin(), map_x_grid_.end(), [&bounds](const int x) {
+        return x > bounds.xMax();
+      });
```
maliberty, line 350: Does this linear search hurt runtime?  I'm wondering if binary search is needed.

For high you can start from where you found low rather than the start.

maliberty:
Connecting and disconnecting a net's iterm/bterm can also change the results (the net's bbox may change).

src/gui/src/heatMap.cpp
gadfort: I didn't notice much, but that is not to say on a larger design it would be apparent.
I'm open to suggestions if you have any.

gadfort:
> What is the memory usage for #4618 with this change?

I only saw a 0.1% increase while generating the map, so that would be around ~16MB, I think, if there is a better method to measure I'd be happy to try and get a better number.
Runtime wise it only takes a few seconds to compute.

src/gui/src/heatMap.cpp
maliberty: std::lower_bound

maliberty:
16MB is great

src/gui/src/heatMap.cpp
gadfort: I could try it out and see

maliberty (approved):

PR #4627, opened 2024-02-06
Grt merge segments

oharboe:


oharboe:
@eder-matheus This algorithm should be a bit better, it avoids O(n^2). I don't have a test-case w.r.t. performance. This was just tinkering based upon observation of one suspend/resume. Perhaps valuable since, as you mentioned, the code was written the way it was written w.r.t. performance concerns?

PR #4628, opened 2024-02-06
grt: use std::vector::erase() in mergeSegments

oharboe:


oharboe:
Superceeded by #4628

maliberty:
> Superceeded by #4628

self-reference??

oharboe:
Before:

3_3_place_gp                    7576

After:

PR #4638, opened 2024-02-08
gui: draw ellipsis animation while rendering

### LINKED ISSUE #4450: Add a descreet indication that background rendering is running
### Description

There is no way of knowing if rendering is still running. This can be confusing when e.g. trying out various heat-maps where rendering can take quite some time.

### Suggested Solution

Add a small visual indication that rendering is running, like a slowly animating ".", "..", "..." at the top left corner of the image.

### Additional Context

_No response_

AcKoucher:
Fix #4450 

Also making a small change to prevent interactions between the cursor and the layoutViewer while the design loading message is drawn.

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -1093,7 +1093,7 @@ odb::Point LayoutViewer::findNextSnapPoint(const odb::Point& end_pt,
 
 void LayoutViewer::mousePressEvent(QMouseEvent* event)
 {
-  if (!hasDesign()) {
+  if (!hasDesign() || !viewer_thread_.isFirstRenderDone()) {
     return;
   }
 
@@ -1124,7 +1124,7 @@ void LayoutViewer::mousePressEvent(QMouseEvent* event)
 
 void LayoutViewer::mouseMoveEvent(QMouseEvent* event)
 {
-  if (!hasDesign()) {
+  if (!hasDesign() || !viewer_thread_.isFirstRenderDone()) {
     return;
   }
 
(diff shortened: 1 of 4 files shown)


AcKoucher:
@oharboe Could you check if this covers what you want?

oharboe:
> @oharboe Could you check if this covers what you want?

I found the visual clue :-)

I'm a little bit concerned that others won't.... Perhaps if you added animation then it would catch the eye.


image

maliberty:
Would making it more obvious be acceptable or do you think animation is essential?  It is simpler as is.

oharboe:
> A more obvious symbol like image
> 
> it could also be placed more centrally
Consider myself a GUI expert, but I dont think a feature that only I notice and like is what we want.

I think we have to experiment and ask others?

A graphics icon like this is better, but these days progress indications animate, so that may be the expectation....

QuantamHD:
+1 @oharboe animating is key.

stefanottili:
Isn't the common thing to just use the "busy" cursor ?
Unless of course you can provide some feedback on how long things will take, then an additional progress bar is always welcome.

AcKoucher:
@stefanottili I guess one interesting thing about the animation is that it makes clear that the loading process is in the application widget that handles the drawing of the design. Wouldn't the busy cursor feel somewhat generic?

oharboe:
Ideally, the GUI would be fully operational while rendering is taking place, so only the rendering would be "busy". So the mouse pointer should not be "busy".

oharboe:
> @oharboe Could you test it again?

Very nice.

I think the graphics could be adjusted, where to render it, how big and possibly with some icons, can be adjusted later. I don't have any strong opinions on that.

I do think the functionality w.r.t. animation is very nice: a non-intrusive indication that rendering is happening.

maliberty (approved):

PR #4642, opened 2024-02-09
gpl: don't run overflow iterations during routability

eder-matheus:


eder-matheus:
@maliberty Secure-ci is passing.

maliberty (approved):

PR #4648, opened 2024-02-09
Fix pin swap of nested not operator

### LINKED ISSUE #4647: Pins swapped incorrectly on and2b cell
### Describe the bug

During the Resizer Timing step of the flow for my Tiny Tapeout design, the inputs of a `sky130_fd_sc_hd__and2b_1` cell were swapped.

The function of the cell is !A & B, so swapping the inputs changes behaviour of the design.

### Expected Behavior

The resizer should not swap pins for cells where the behaviour of the design would be changed by the swap.

### OpenROAD Environment

```shell
Using the docker image: efabless/openlane-tools:openroad_app-75f2f325b7a42e56a92404f33af8e96530d9b202-centos-7-amd64
```

### OpenLane Environment

```shell
open_pdks cd1748bb197f9b7af62a54507de6624e30363943
Kernel: Linux v5.15.133.1-microsoft-standard-WSL2
Distribution: ubuntu 22.04
Python: v3.10.12 (OK)
Container Engine: docker v20.10.25 (OK)
OpenLane Git Version: 9dbd8b5ea2bd891bed4dcc97df5c7439083f0368
python-venv: INSTALLED
---
PDK Version Verification Status: OK
---
Git Log (Last 3 Commits)

9dbd8b5 2024-01-11T15:26:56+02:00 Fix `-synth_explore` crash (#2085) - Kareem Farid -  (grafted, HEAD)
---
Git Remotes

origin  https://github.com/The-OpenROAD-Project/OpenLane.git (fetch)
origin  https://github.com/The-OpenROAD-Project/OpenLane.git (push)
```

### To Reproduce

Clone the `openroad-bug` branch of https://github.com/MichaelBell/tt06-tinyQV recursively
Follow https://docs.google.com/document/d/1aUUZ1jthRpg4QURIIyzlOaPWlmQzr-jBn3wZipVUPt4/edit#heading=h.wwc5ldl01nl5 to set up the OpenLane flow for TinyTapeout, using the above project instead of the factory test project.
Run:

    tt/tt_tool.py --create-user-config
    tt/tt_tool.py --harden

Inspect `runs/wokwi/results/placement/tt_um_MichaelBell_tinyQV.nl.v` and `runs/wokwi/results/routing/tt_um_MichaelBell_tinyQV.nl.v` and note that the wires for the input for instance `_2354_` have been swapped:

```
 sky130_fd_sc_hd__and2b_1 _2354_ (.A_N(_0535_),
    .B(_0536_),
    .X(_0537_));
```
becomes
```
 sky130_fd_sc_hd__and2b_1 _2354_ (.A_N(_0536_),
    .B(_0535_),
    .X(_0537_));
```

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

MichaelBell:
`RepairSetup::getEquivPortList2` determines which ports on a cell are equivalent, and therefore can be swapped to improve timing.  However, it incorrectly treats a single port inside a not operator as equivalent to a port that is not inside a not operator.  This means the output of the cell is negated if the pins are swapped.

If all ports are inside the not operator it should be allowed, but if the not operator only applies to some ports it must not be.  Therefore `op_not` should only be permitted if the status is `op_zero` (no operations performed yet).

This fixes #4647.

I have tested this fix on my repro case for #4647 and it resolves the problem.

### DIFF
--- a/src/rsz/src/RepairSetup.cc
+++ b/src/rsz/src/RepairSetup.cc
@@ -890,7 +890,7 @@ RepairSetup::getEquivPortList2(sta::FuncExpr *expr,
     using Operator = sta::FuncExpr::Operator ;
     const Operator curr_op = expr->op();
 
-    if (curr_op == Operator::op_not) {
+    if (status == Operator::op_zero && curr_op == Operator::op_not) {
         getEquivPortList2(expr->left(), ports, status);
     }
     else if (status == Operator::op_zero &&


QuantamHD:
All commits need to be signed off using `-s` during `git commit`. You can add a sign off to your existing commits with the following commands.

image

MichaelBell:
Apologies, I had missed that you were using signed-off-by in that way.  Done.

maliberty (approved):

maliberty:
This fix is too restrictive and will block swapping on a nand2 with function : "(!A) | (!B)".  I'll think about a more precise fix.

QuantamHD:
This needs to be redesigned completely in my view @maliberty. We should evaluate the truth tables under swaps, and ensure that they are equal. Messing around with this symbolic stuff has proven to be error prone.

MichaelBell:
Hmm, yes it's a shame that nand2 is described that way.
I agree that building truth tables by evaluating the function would be a better approach.  Ideally would make something that works for cells with > 2 input ports, too.

maliberty:
Replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/4651

I agree we should rethink this but the above resolves this issue.

mattvenn:
@maliberty @QuantamHD do you have any idea how long this issue has been present? how many tapeouts could it have affected?

Issue #4630, opened 2024-02-07
display "busy" cursor while RUDY is calculating the map


stefanottili:
### Describe the bug

For ispd24 cluster.def (9.8M std cell) it takes RUDY ~30 seconds to compute the map.

### Expected Behavior

Please change the cursor from "pointer" to "busy" while computing RUDY to give the user feedback that OR is doing something.

### Environment

```shell
Feb 6 OR M1 macOS Sonoma
```

### To Reproduce

openroad -gui 
read_lef lef/Nangate.lef
read_def def/cluster.def.gz

show RUDY map

maliberty:
@gudeh

gudeh:
References: 
https://liangrj2014.github.io/ISPD24_contest/
#4618

gudeh:
I noticed that other heatmap calculations also do not indicate when they are busy, and these heatmaps take some time with larger designs as well. Therefore, I have made modifications to show a busy indicator for any heatmap calculation, not just for RUDY.

gudeh:
Covered by #4638.

PR #4657, opened 2024-02-12
Display busy cursor during heatmap calculations.

gudeh:
Asked by user on issue #4630.

### DIFF
--- a/src/gui/src/heatMap.cpp
+++ b/src/gui/src/heatMap.cpp
@@ -532,6 +532,7 @@ void HeatMapDataSource::ensureMap()
 
   if (build_map || !isPopulated()) {
     debugPrint(logger_, utl::GUI, "HeatMap", 1, "Populating map");
+    
     QApplication::setOverrideCursor(Qt::WaitCursor);    
     populated_ = populateMap();
     QApplication::restoreOverrideCursor();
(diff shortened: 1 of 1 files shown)


oharboe:
@gudeh @maliberty already adressed by the animating elipsis, no?

gudeh:
Hi @oharboe! I am not aware about this change you mentioned. Although, the requested change is specifically for when we click the button to show the heatmap, since it takes some time on larger designs. Until now, we haven't had any indicator to show that it's loading.

oharboe:
@gudeh Try #4638, it might need a bit of bling, bu

maliberty:
I think this is different from the rendering busy signal.  The heat map generation runs in the main thread

oharboe:
> I think this is different from the rendering busy signal. The heat map generation runs in the main thread

Ideally, the calculation could be moved to the rendering thread...?

maliberty:
> Ideally, the calculation could be moved to the rendering thread...?

Its not a rendering operation.  Perhaps another thread could be used but that would add a lot more complexity.

oharboe:
> > Ideally, the calculation could be moved to the rendering thread...?
> 
> Its not a rendering operation. Perhaps another thread could be used but that would add a lot more complexity.

Why have multiple non UI threads? 

Currently rendering is locked when the GUI thread is locked.... 

So locking rendering thread is an improvement over locking.main thread.

maliberty:
Rendering is not the only concern.  Any modifications to odb while there are readers is unsafe.

PR #4669, opened 2024-02-15
cmake: move add_subdirectory mpl2 and par

vvbandeira:
The add_subdirectory statements for mpl2 and par are now conditional to the OS to avoid build and link errors on macOS.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -270,10 +270,8 @@ add_subdirectory(cts)
 add_subdirectory(grt)
 add_subdirectory(tap)
 add_subdirectory(mpl)
-add_subdirectory(mpl2)
 add_subdirectory(rcx)
 add_subdirectory(psm)
-add_subdirectory(par)
 add_subdirectory(ant)
 add_subdirectory(gui)
 add_subdirectory(drt)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -353,6 +353,8 @@ if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   target_link_libraries(openroad mpl2 par)
   target_compile_definitions(openroad PRIVATE ENABLE_MPL2)
   target_compile_definitions(openroad PRIVATE ENABLE_PAR)
+  add_subdirectory(mpl2)
+  add_subdirectory(par)
```
maliberty, line 357: Do the ENABLE_* no longer work?

vvbandeira: Cherry mentioned that just adding the subdir was causing problems and suggested this patch.

maliberty: ok then we might as well drop the options

stefanottili: mpl2 and par compile and link just fine on macOS M1.

The resulting openroad executable will throw a runtime error. 
Putting the ortools.dylib at the end of the link command fixes that.

vvbandeira: the `ENABLE_PAR` and `ENABLE_MPL2` are required for correct includes in `Main.cc` and `OpenRoad.cc`

maliberty (approved):

PR #4673, opened 2024-02-15
rsz: Makes pin swap equivalence algorithm more robust

QuantamHD:
Replaces the existing symbolic manipulation code with truth table evaluation.

Also changes the code to keep track of pin equivalence instead of "cell pin group" equivalence. Not all pins are equal to other pins as sky130_fd_sc_hd__o311ai_1 shows. The code would allow a more broad set of swapping to occur.

No functional change should be observed since the pin swap is limited to two term cells which would have worked under the old scheme.


QuantamHD:
Do I need to rebase against master @maliberty?

image

maliberty:
rebasing is a good idea

maliberty:
secure CI started

src/rsz/test/pin_swap1.ok
```diff
@@ -10,431 +10,1314 @@
...
-sky130_fd_sc_hd__xor2_2 ->
-sky130_fd_sc_hd__xor2_4 ->
-sky130_fd_sc_hd__xor3_1 ->
-sky130_fd_sc_hd__xor3_2 ->
-sky130_fd_sc_hd__xor3_4 ->
+sky130_fd_sc_hd__a2111o_1/A1 -> A2
+sky130_fd_sc_hd__a2111o_1/A2 -> A1
+sky130_fd_sc_hd__a2111o_1/B1 -> C1 D1
+sky130_fd_sc_hd__a2111o_1/C1 -> B1 D1
+sky130_fd_sc_hd__a2111o_1/D1 -> B1 C1
```
maliberty, line 17: could this be reduced to

sky130_fd_sc_hd__a2111o_1 -> [A1, A2] [B1, C1, D1] 

?

src/rsz/test/pin_swap1.ok
```diff
@@ -10,431 +10,1314 @@
...
+sky130_fd_sc_hd__einvp_1/A ->
+sky130_fd_sc_hd__einvp_1/TE ->
+sky130_fd_sc_hd__einvp_2/A ->
+sky130_fd_sc_hd__einvp_2/TE ->
+sky130_fd_sc_hd__einvp_4/A ->
+sky130_fd_sc_hd__einvp_4/TE ->
+sky130_fd_sc_hd__einvp_8/A ->
+sky130_fd_sc_hd__einvp_8/TE ->
+sky130_fd_sc_hd__fa_1/A ->
+sky130_fd_sc_hd__fa_1/B ->
```
maliberty, line 553: Extra credit for a multi output gate :trophy:

src/rsz/src/RepairSetup.hh
```diff
@@ -168,7 +178,10 @@ private:
   const MinMax *min_;
   const MinMax *max_;
 
-  sta::UnorderedMap<LibertyCell *, sta::LibertyPortSet> equiv_pin_map_;
+  sta::UnorderedMap<std::pair<LibertyCell*, LibertyPort*>,
```
maliberty, line 181: Why do you need the pair?  A LibertyPort can only belong to one LibertyCell so it seems redundant to mention the cell and would remove the need for Boost.

src/rsz/src/RepairSetup.cc
```diff
@@ -965,23 +1007,34 @@ RepairSetup::equivCellPins(const LibertyCell *cell, sta::LibertyPortSet &ports)
...
-            if (expr != nullptr) {
-                getEquivPortList(expr, ports);
-            }
+          LibertyPort* candidate_port = port_iter2.next();
+          if (!candidate_port->direction()->isInput()) {
+            continue;
+          }
+
+          if (expr != nullptr && input_port != candidate_port
+              && isPortEqiv(expr, cell, input_port, candidate_port)) {
```
maliberty, line 1035: Won't this lead to duplicate checks (eg A can be swapped with B implies B can be swapped with A)?

src/rsz/src/RepairSetup.cc
```diff
@@ -885,76 +886,117 @@ RepairSetup::fanout(Vertex *vertex)
...
     }
+    var_index++;
+  }
+
+  if (port_stimulus.find(port_a) == port_stimulus.end()) {
+    return false;
+  }
+  if (port_stimulus.find(port_b) == port_stimulus.end()) {
+    return false;
+  }
```
maliberty, line 973: when would this happen?

src/rsz/src/RepairSetup.hh
QuantamHD: That does seem correct, but it does seem a bit less readable to my eye. The reader would need to know that I guess. I can make the change, but I'll probably add a comment so it's clear.

src/rsz/src/RepairSetup.cc
QuantamHD: Yeah, but that's an optimization, and this runs over the entire library in less than a second. No point in complicating it unless it's too slow.

src/rsz/src/RepairSetup.cc
QuantamHD: If someone used this function with a port that didn't belong to the cell.

src/rsz/test/pin_swap1.ok
QuantamHD: Yeah, but we would need to implement the merging, and really doesn't represent what the underlying code is actually calculating. I think the above is simpler and actually a better test.

src/rsz/src/RepairSetup.cc
maliberty: In that case I would just check port_a->libertyCell() == cell (likewise port_b) at the top of the method.

src/rsz/src/RepairSetup.hh
maliberty: This is less readable to my eye because I wonder why the cell is there - could it not be the cell associated with this port? I guess it depends on experience.   I'd prefer the change with a comment

src/rsz/test/pin_swap1.ok
QuantamHD: Done, but don't go asking for any Christmas presents this year.

image

maliberty:
Looks like two flow tests need a metric update.

QuantamHD:
Metrics degraded slightly, but I created to PR to allow more than just 2 input nets. Once enabled there I see these metric regressions disappear.

src/rsz/src/RepairSetup.cc
```diff
@@ -524,12 +524,13 @@ bool RepairSetup::swapPins(PathRef *drvr_path,
...
+        // and that should apply to all instances of that cell with this input_port.
+        if (equiv_pin_map_.find(input_port) == equiv_pin_map_.end()) {
+          equivCellPins(cell, input_port, ports);
+          equiv_pin_map_.insert(input_port, ports);
         }
-        ports = equiv_pin_map_[cell];
+        ports = equiv_pin_map_[input_port];
         if (ports.size() > 1) {
             resizer_->findSwapPinCandidate(input_port, drvr_port, load_cap,
                                            dcalc_ap, &swap_port);
```
maliberty, line 536: `ports.size() > 1` needs to become `!ports.empty()` with you change from keying on the input port rather than the cell.  Once you fix this please see if you actually need the metric updates that are in this PR.

maliberty:
In asap7/aes-mbff I see

````
[INFO RSZ-0094] Found 162 endpoints with setup violations.
Error: cts.tcl, 107 _Map_base::at
````

Please take a look.

PR #4676, opened 2024-02-16
odb: adding a helper function to get the dbObjects of a dbVector with a generic OutputIterator

fgaray:
Use case:
  The original getObjects function fills a vector with the private class
  (underscore). In some cases, we want to return the public class.

  Without this implementation, we would need to create a temporal vector
  to hold the private class and then cast this into a new vector for the
  public class.

### DIFF
--- a/src/odb/src/db/dbTable.h
+++ b/src/odb/src/db/dbTable.h
@@ -171,6 +171,17 @@ class dbTable : public dbObjectTable, public dbIterator
   dbObject* getObject(uint id, ...) override;
   void getObjects(std::vector<T*>& objects);
 
+  template <class OutputIterator>
+  void getObjects(OutputIterator back_inserter)
+  {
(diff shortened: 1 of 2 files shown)


maliberty:
We should not be returning containers but iterators.  I'm unclear how you plan to use this change.

fgaray:
So, I have this class:

image


That then I can use to get dbScanInst* (for example) instead of _dbScanInst*:

image

Because I don't want to return the private class (_dbScanInst) from dbScanChain.

I can change this if there is another way, thanks!

maliberty:
We shouldn't be returning a std::vector; dbSet is used by most APIs (eg dbBlock::getInsts()) and it provides iterators that work with dbTable without copying.

fgaray:
Sounds good!, I will use dbSet then, thanks!

PR #4682, opened 2024-02-19
mpl2: consider macro dominance when computing io blockages' depth

AcKoucher:
This PR is to ensure we don't generate huge io blockages and end up deteriorating QoR when using the changes in #4206.
The quadratic power is to make the factor more aggressive as the problems below were not solved using the factor without it.

Secure-CI for #4206 showed, among some other metrics failures, problems for the following designs:

gf121/ariane: [ERROR GPL-0307] RePlAce divergence detected. Re-run with a smaller max_phi_cof value.
tsmc65lp/bp_be: [ERROR DPL-0036] Detailed placement failed.

Both of these designs had huge io blockages for pin access. Running the changes here combined with #4206 lead both designs to finish without metric failures.

I'm starting a Secure-CI only for this.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -3150,8 +3150,8 @@ void HierRTLMP::setIOClustersBlockages()
   const float macro_dominance_factor
       = macro_with_halo_area_
         / (root_cluster_->getWidth() * root_cluster_->getHeight());
-  const float depth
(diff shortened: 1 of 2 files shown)


eder-matheus:
@maliberty @AcKoucher CI results are mostly passing. Only one design has a metric failure:
`nangate45/bp_be: [ERROR] detailedroute__route__wirelength fail test: 2636966.0 <= 2630349.0`

maliberty:
If that's the only issue than I'm ok to update and merge.

maliberty (approved):

Issue #4108, opened 2023-10-07
Global routing fails to route, even if there is a lot of space


oharboe:
### Description

unzip https://drive.google.com/file/d/1BFUHnN-5qJI7WUqug3ehs6fxSkNoZCqJ/view?usp=sharing

Run:

```
$ ./run-me-mock-aluregisterfile-asap7-base.sh 
OpenROAD v2.0-10613-g75796f6ec 
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
[INFO GRT-0020] Min routing layer: M2
[INFO GRT-0021] Max routing layer: M7
[INFO GRT-0022] Global adjustment: 0%
[INFO GRT-0023] Grid origin: (0, 0)
[INFO GRT-0043] No OR_DEFAULT vias defined.
[INFO GRT-0088] Layer M1      Track-Pitch = 0.0360  line-2-Via Pitch: 0.0360
[INFO GRT-0088] Layer M2      Track-Pitch = 0.0390  line-2-Via Pitch: 0.0360
[INFO GRT-0088] Layer M3      Track-Pitch = 0.0360  line-2-Via Pitch: 0.0360
[INFO GRT-0088] Layer M4      Track-Pitch = 0.0480  line-2-Via Pitch: 0.0480
[INFO GRT-0088] Layer M5      Track-Pitch = 0.0480  line-2-Via Pitch: 0.0480
[INFO GRT-0088] Layer M6      Track-Pitch = 0.0640  line-2-Via Pitch: 0.0640
[INFO GRT-0088] Layer M7      Track-Pitch = 0.0640  line-2-Via Pitch: 0.0640
[INFO GRT-0019] Found 892 clock nets.
[INFO GRT-0001] Minimum degree: 2
[INFO GRT-0002] Maximum degree: 31
[INFO GRT-0003] Macros: 4
[INFO GRT-0004] Blockages: 2496

[INFO GRT-0053] Routing resources analysis:
          Routing      Original      Derated      Resource
Layer     Direction    Resources     Resources    Reduction (%)
---------------------------------------------------------------
M1         Vertical            0             0          0.00%
M2         Horizontal   15151320       4140683          72.67%
M3         Vertical     17502480       5557969          68.24%
M4         Horizontal   12835800       4107283          68.00%
M5         Vertical     12835800       3683262          71.30%
M6         Horizontal    9335520       2665179          71.45%
[...]

PR #4685, opened 2024-02-19
Grt fix pinoffset adjustment

arthurjolo:
Fixes [#4108](https://github.com/The-OpenROAD-Project/OpenROAD/issues/4108)

GlobalRoute's pin offset adjustment was not taking into account that 1 net could have more then one pin that needed a fake position in the same GCell. With this it was creating a duplicated guide to reach those pins and then it was decreasing the capacity of the GCells contained in those guides more then it should.
Also the fake pin creation was failing to create the fake pins when a net had more then one fake pin in the same GCell.

This PR fixes that by checking if there is repeated guides to connect fake pins and only adding the adjustment once, and it also allows the creation of fake pins on the same GCell.


src/grt/src/GlobalRouter.cpp
```diff
@@ -1071,8 +1071,11 @@ void GlobalRouter::computeTrackAdjustments(int min_routing_layer,
 
 void GlobalRouter::computePinOffsetAdjustments()
 {
-  for (auto const& map_obj : pad_pins_connections_) {
-    for (const auto& segment : map_obj.second) {
+  for (auto& net_route_fake_pins : pad_pins_connections_) {
```
eder-matheus, line 1074: You can use `for (auto& [db_net, segments] : pad_pins_connections_` to avoid using the `.first` and `.second` below.

eder-matheus (approved):
Will wait for my comment and for secure-ci before merging it. But this is a nice bug fix!

PR #4689, opened 2024-02-20
grt: fix crash caused by overlaps and ordering of wires creation during repair_antennas

eder-matheus:
Fixes crash in the CI. A few metrics failures in the regression tests, but no failures in the public-ci.
Private-ci is running.


eder-matheus:
@maliberty Secure-ci is passing too.

maliberty (approved):

PR #4694, opened 2024-02-20
odb: adding support for persistent DFT structures

fgaray:
This is a basic set of structs to support scan chains and scan cells in odb.

Also, bumping the db_schema_minor to 80


src/odb/test/cpp/scan/TestScanChain.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+constexpr std::string_view kPartition1 = "CHAIN_1_FALLING_1";
+constexpr std::string_view kPartition2 = "CHAIN_1_RISING_1";
+
+template <class>
+inline constexpr bool always_false_v = false;
+
+std::string_view GetName(const std::variant<dbBTerm*, dbITerm*>& pin)
+{
+  return std::visit(
```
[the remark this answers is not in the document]
fgaray: Seems like a false warning.

src/odb/src/db/dbBlock.cpp
```diff
@@ -1361,6 +1379,12 @@ bool _dbBlock::operator==(const _dbBlock& rhs) const
   if (*_cc_seg_tbl != *rhs._cc_seg_tbl)
     return false;
 
+  if (_dft != rhs._dft)
+    return false;
```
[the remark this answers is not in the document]
fgaray: Generated code

fgaray:
Hi @maliberty , this is a big PR but it is mostly code generated by the code generator script.

Thanks!

maliberty:
I see many resolved comments but not the usual clang-tidy thumbs-up.  Is there something not yet resolved?

src/odb/test/data/design.odb
maliberty, line 1: This file should not be modified as it is for backwards compatibility testing.

src/odb/src/db/dbBlock.cpp
```diff
@@ -2211,6 +2239,20 @@ dbExtControl* dbBlock::getExtControl()
   return (block->_extControl);
 }
 
+dbDft* dbBlock::createDft()
+{
+  _dbBlock* block = (_dbBlock*) this;
+  _dbDft* dft = block->_dft_tbl->create();
+  block->_dft = dft->getId();
+  return (dbDft*) dft;
+}
```
maliberty, line 2248: Can you have more than one dft per block?  It doesn't seem that way but this code would allow repeated calls.

src/odb/test/data/design.odb
fgaray: I will rollback this file to the previous state.

src/odb/src/db/dbBlock.cpp
fgaray: Just to be clear, the issue that you are seeing here is that we may create more than one dft per block (agree that's an issue), not that I am using a table (_dft_tbl) to hold the pointer right?

fgaray:
> I see many resolved comments but not the usual clang-tidy thumbs-up. Is there something not yet resolved?

Yeah, most of them are in code that I am keeping the already existing style: 
image

Or something that I already fixed (I was missing a header): 
image

Other seems to be false positive (I don't think this is a temporal object): 
image


As well as this one too (std::always_false_v is used in a constexpr code so it is not in the final binary): 
image

Does that makes sense?

src/odb/src/db/dbBlock.cpp
maliberty: If there can only be one then there is no need for a table.  The table optimizes allocation of many objects.  Its not too harmful (we allocate a page of objects but only use one) but I'm sure it is helpful

fgaray: So, the tricky thing is that I found no way of setting the owner to my dbDft class without using the dbTable as a "wrapper". I looked into other objects inside the block and I don't think they use a pattern similar to that. However, I changed the code and now we will only have one DFT object per block and the user of odb will not be able to create more.

fgaray:
Hi @maliberty, I think this is ready for another look, thanks!

maliberty:
#4617 and this PR are about to contend on the db schema.  Would you be ok if that one goes first and you take the merge?

fgaray:
> #4617 and this PR are about to contend on the db schema. Would you be ok if that one goes first and you take the merge?

Sure, no problem! Thanks!

src/odb/src/db/dbScanChain.cpp
```diff
@@ -80,79 +85,222 @@ void _dbScanChain::differences(dbDiff& diff,
...
+}
+
+void dbScanChain::setName(std::string_view name)
+{
+  _dbScanChain* scan_chain = (_dbScanChain*) this;
+  scan_chain->name_ = name;
+}
+
+dbBTerm* _dbScanChain::getPin(_dbScanChain* scan_chain,
+                              dbId<dbBTerm> _dbScanChain::*field)
```
maliberty, line 210: Why not `const dbId<dbBTerm>& bterm_id`?  It would remove the need for the cast below.

src/odb/src/db/dbScanInst.cpp
```diff
@@ -116,5 +130,134 @@ dbOStream& operator<<(dbOStream& stream, const _dbScanInst& obj)
...
+  scan_inst->clock_edge_ = static_cast<uint>(clock_edge);
+}
+
+dbScanInst::ClockEdge dbScanInst::getClockEdge() const
+{
+  _dbScanInst* scan_inst = (_dbScanInst*) this;
+  return static_cast<ClockEdge>(scan_inst->clock_edge_);
+}
+
+void dbScanInst::setBits(uint bits)
```
maliberty, line 158: What does 'bits' hold?  It is rather vague to me.

maliberty (changes requested):

src/odb/include/odb/db.h
```diff
@@ -7067,6 +7073,22 @@ class dbAccessPoint : public dbObject
...
+ public:
+  void setScanInserted(bool scan_inserted);
+
+  bool isScanInserted() const;
+
+  dbSet<dbScanChain> getScanChains() const;
+
+  // User Code Begin dbDft
+  dbId<dbScanPin> CreateScanPin(dbBTerm* bterm);
+  dbId<dbScanPin> CreateScanPin(dbITerm* iterm);
```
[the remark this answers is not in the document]
fgaray: Thanks!, agree, I changed the naming style and also moved the create methods as static for this and other dft classes.

No reason to do this on the original code so let's better use the usual pattern.

src/odb/include/odb/db.h
```diff
@@ -7617,22 +7639,91 @@ class dbPowerSwitch : public dbObject
...
+
+  void setScanEnable(dbBTerm* scan_enable);
+  void setScanEnable(dbITerm* scan_enable);
+  std::variant<dbBTerm*, dbITerm*> getScanEnable() const;
+
+  void setAccessPins(const AccessPins& access_pins);
+
+  AccessPins getAccessPins() const;
+
+  std::vector<dbInst*> getInsts() const;
```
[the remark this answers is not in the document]
fgaray: The logic is to be able to express the DEF file section for scan chains for FLOATING and ORDERED elements.

image

FLOATING is easy, it is just one element in each scan inst. For ORDERED, we need to keep the order of the scan elements so we abstract the access pins and we treat that sequence of elements as one "virtual" cell.

So, we could have something like this:

[FF1, FF2, [FF3, FF4], FF5]

scanInst 1: FF1 -> FLOATING
scanInst 2: FF2 -> FLOATING
scanInst 3: [FF3, FF4] -> ORDERED
scanInst 4: FF5 -> FLOATING

One example of ORDERED elements are shift-registers that you are able to use it for shifting patterns but they should not be reorder (because you will break the functional part of your design).

src/odb/include/odb/db.h
```diff
@@ -7617,22 +7639,91 @@ class dbPowerSwitch : public dbObject
...
+
+  void setStop(dbBTerm* bterm);
+
+  void setStop(dbITerm* iterm);
+
+  std::variant<dbBTerm*, dbITerm*> getStart() const;
+
+  std::variant<dbBTerm*, dbITerm*> getStop() const;
+
+  std::string_view getName() const;
```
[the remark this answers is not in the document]
fgaray: So, I did a quick test and you are right, this do not play well with swig, instead of printing the string, it prints the objects address + class name.

I change back every getter that was returning std::string_view to std::string, thanks for pointing this out!

src/odb/test/cpp/scan/TestScanChain.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    for (const auto& name : instances_names) {
+      instances_.push_back(block_->findInst(name.c_str()));
+    }
+  }
+
+  dbDatabase* writeReadDb()
+  {
+    writeDb();
+    return readDb();
```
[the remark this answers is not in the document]
fgaray: Yes, so that we get the same objects back after writing. Does it make sense?

src/odb/src/db/dbScanInst.cpp
fgaray: The number of bits of the scan inst.

For simple flops it is just 1, for shift registers it could be the number of elements, for multibit cells with internal scan it would be the number of bits in the cell, etc.

src/odb/src/db/dbScanChain.cpp
fgaray: Makes sense, thanks!

I will be back in a few days as now I am realizing that I need to also support iterms as scan_in and scan_out for the scan chains for future scan compression implementation. I will tag you later once this is ready for another review, thanks!

src/odb/src/db/dbScanInst.cpp
maliberty: The seems rather overloaded.  At a minimum a descriptive comment is needed in db.h.

src/odb/include/odb/db.h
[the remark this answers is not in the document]
maliberty: dbScanInst is an odd name to express this as this seems to be an ordered collection of instances.  It feels like there is a missing intermediate class, something like

dbScanChain -> dbScanInstSequence -> dbScanInst

src/odb/test/cpp/scan/TestScanChain.cpp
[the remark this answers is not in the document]
maliberty: Ok a small comment would clarify

src/odb/include/odb/db.h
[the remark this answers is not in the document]
fgaray: That makes sense, what do you think if I reuse the name ScanCell here:

dbScanChain -> dbScanCell -> dbScanInst

Similar to the ScanCell in src/dft/src/cells/ScanCell.hh where we hold dbInsts ?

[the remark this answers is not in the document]
maliberty: Instance and Cell sound to me like placeable objects rather than abstract collections.  Do you have thoughts on how we might avoid terminology conflict?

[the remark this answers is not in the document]
maliberty: My concern with (2) is can you get it through swig into tcl/python.  If you want to see if that's possible then I'm fine with (2).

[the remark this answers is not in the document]
fgaray: That's fair. I always forget about tcl. I will go with (1) creating dbScanList to hold a sequence of dbScanInst then.

Thanks!

maliberty:
ready for review or still working?

fgaray:
>ready for review or still working?

Yep, ready for review  :). The git history is a mess in this PR so let me know before merging so I can clean it up a little bit. Thanks!

src/odb/include/odb/db.h
```diff
@@ -7642,22 +7659,110 @@ class dbPowerSwitch : public dbObject
...
+  void setScanClock(std::string_view scan_clock);
+  const std::string& getScanClock() const;
+
+  void setClockEdge(ClockEdge clock_edge);
+  ClockEdge getClockEdge() const;
+
+  // The number of bits that are in this scan inst from the scan in to the scan
+  // out. For simple flops this is just 1.
+  void setBits(uint bits);
+  uint getBits() const;
```
maliberty, line 7718: Just use int  (https://google.github.io/styleguide/cppguide.html#Integer_Types)

src/odb/include/odb/db.h
```diff
@@ -7642,22 +7659,110 @@ class dbPowerSwitch : public dbObject
...
+  void setScanIn(dbBTerm* scan_in);
+  void setScanIn(dbITerm* scan_in);
+  std::variant<dbBTerm*, dbITerm*> getScanIn() const;
+
+  void setScanOut(dbBTerm* scan_out);
+  void setScanOut(dbITerm* scan_out);
+  std::variant<dbBTerm*, dbITerm*> getScanOut() const;
+
+  void setScanEnable(dbBTerm* scan_enable);
+  void setScanEnable(dbITerm* scan_enable);
```
maliberty, line 7680: What does it mean for an iterm to be the scan_enable for a whole chain?  Does it imply some sort of chained connection?

src/odb/include/odb/db.h
```diff
@@ -7642,22 +7659,110 @@ class dbPowerSwitch : public dbObject
...
+
+  void setClockEdge(ClockEdge clock_edge);
+  ClockEdge getClockEdge() const;
+
+  // The number of bits that are in this scan inst from the scan in to the scan
+  // out. For simple flops this is just 1.
+  void setBits(uint bits);
+  uint getBits() const;
+
+  void setScanEnable(dbBTerm* scan_enable);
```
maliberty, line 7720: What does it mean for the enable of a scan inst to be a bterm?  Similarly with scan in/out

src/odb/include/odb/db.h
```diff
@@ -7642,22 +7659,110 @@ class dbPowerSwitch : public dbObject
...
+  // User Code End dbScanInst
+};
+
+class dbScanList : public dbObject
+{
+ public:
+  dbSet<dbScanInst> getScanInsts() const;
+
+  // User Code Begin dbScanList
+  dbScanInst* add(dbInst* inst);
```
maliberty, line 7739: Adds to where in the list?  A more descriptive name would be better

src/odb/include/odb/db.h
```diff
@@ -7642,22 +7659,110 @@ class dbPowerSwitch : public dbObject
 class dbScanChain : public dbObject
 {
  public:
+  dbSet<dbScanPartition> getScanPartitions() const;
+
+  dbSet<dbScanList> getScanLists() const;
```
maliberty, line 7664: How do lists and and partitions interact?  Can a partition span multiple lists?  Include part of a list?

maliberty:
It would be good to have descriptive comments for the new classes in db.h

fgaray:
>It would be good to have descriptive comments for the new classes in db.h

maliberty:
Can you just add the comments to db.h before the class?

fgaray:
No, it disappears. I see other comments for some classes there (ex: dbTechSameNetRule), but I am not sure how they are preserving those comments.

Meanwhile I am adding comment for the hand made methods.

maliberty:
dbTechSameNetRule is not generated

maliberty:
How about adding
```
diff --git a/src/odb/src/codeGenerator/templates/db.h b/src/odb/src/codeGenerator/templates/db.h
index 04ac32e09..6159423ef 100644
--- a/src/odb/src/codeGenerator/templates/db.h
+++ b/src/odb/src/codeGenerator/templates/db.h
@@ -5,7 +5,11 @@ class {{klass.name}};
 //Generator Code End ClassDeclarations
 //Generator Code Begin ClassDefinition
 {% for klass in schema.classes|sort(attribute='name') %}
-
+{% if klass.description %}
+  {% for line in klass.description %}
+    // {{ line }}
+  {% endfor %}
+{% endif %}
 class {{klass.name}} : public dbObject
```

fgaray:
Hi @maliberty , That makes sense! Do you want me to include that in this PR or would you send one with your code? Thanks!

maliberty:
You can just include it here.

src/odb/include/odb/db.h
fgaray: > Can a partition span multiple lists?

Yes but not partial lists. Let's say that I have a mux -> SFF1 -> SFF2. I need to express that like [mux, SFF1], [SFF2]. If they are of the same clock and edge then they will be in the same partition.

I changed this, It think this model handles better what we want to express here:

scan chain -> scan_partition -> scan_list -> scan_inst

Wdyt?

src/odb/include/odb/db.h
fgaray: Added a comment to say that we are appending at the end

src/odb/include/odb/db.h
fgaray: I added  a comment about this.

The idea is that a scan in/scan out/scan enable may be top level ports (bterms) but also pins in dbInsts. Pins in cells (iterms) are used if the scan chain is part of a compressor/decompressor and the scan enable may be controlled by a TAP Controller in the case of boundary scan.

src/odb/include/odb/db.h
fgaray: Is the scan enable pin that puts the scan chain into shifting mode. We need this pin for simulations and well, to let the ATE (Automatic Test Equipment) what pin needs to be set to 1 for shifting.

maliberty: When would that pin be an iterm and not a bterm?  iterms are internal to the block.

src/odb/include/odb/db.h
maliberty: Perhaps a quick picture showing what you have in mind would help me to understand.

src/odb/include/odb/db.h
fgaray: Maybe I have a misconception of what an iterm is. Is this correct?

If I have a cell (let's say a mux) and I want to take the output pin of that mux, would that be an iterm?

If I have a top-level input port (let's say, in verilog, "module (my_port)", would that be a bterm?

Thanks!

src/odb/include/odb/db.h
fgaray: So, in this image, SI, SE, SO, and CLK would be bterms, because they are input ports at the top level of the module.

Screenshot 2024-04-05 at 10-50-35 TestPrinciples pdf

Basically:

```
module (SI, SE, CLK, SO);
   input SI, SE, CLK;
   output SO;
endmodule
````

An example of when SE would be an iterm, is for pipelined scan enabled:

Screenshot 2024-04-05 at 10-52-49 Architecture for Pipelining of Scan Enable

Where the scan enabled signal comes from the output of the or gate.

For a case where CLK is coming from a gate output, we have clock gating:

Screenshot 2024-04-05 at 10-51-57 TestPrinciples pdf

GCK would be an iterm right?

And finally, for the case of compression, we will have SI and SO also as iterms (only showing the decompressor -> SI case here)

Screenshot 2024-04-05 at 10-55-13 TestPrinciples pdf

Again, this is from my understanding of the difference between bterms and iterms, I may be wrong here.

Thanks!

src/odb/include/odb/db.h
maliberty: That's correct, in iterm blocks to an instance

src/odb/include/odb/db.h
maliberty: I think I get it.  The key point is that the iterm isn't actually an iterm of any element in the dbScainChain but instead of some other control logic.

maliberty (approved):

maliberty:
DFT is a new area for me so I've been slow to get it.  I think this is clear enough for me to move forward.  I'm sure it will become clearer as you build out the rest of the solution.

PR #4696, opened 2024-02-21
Opendp: Fix missing include

ubfx:
We have been seeing this error in some build environments:
```
src/dpl/include/dpl/Opendp.h:230:17: error: 'accumulate' is not a member of 'std'
```

This is fixed by this patch.


maliberty (approved):

rovinski:
@maliberty requires workflow approval

maliberty:
Thanks for the fix!

PR #4698, opened 2024-02-22
pdn: ensure adjustment to min area is rounded up

gadfort:
@eder-matheus please check if this resolves the min-area issue you saw. I was able to trigger an issue with the new test, but please check on the test you showed me

Fixes: min area getting rounded down causing a min area violation


rovinski:
If the whole lef file needed for the test case?

gadfort:
@rovinski I'll see, it would be nice to not need it (I was just being a little lazy and copying it forward)

maliberty:
@eder-matheus you should do a squash commit if it is removed to eliminate it from the history (or @gadfort could rebase/squash/force push)

gadfort:
@maliberty I just forced pushed to remove it and replace it with a call to setArea on metal6 (much nicer solution). Thanks @rovinski

PR #4701, opened 2024-02-23
ci: remove docker centos7 builds

vvbandeira:


stefanottili:
Just out of curiosity, could you please elaborate on the reason ?

maliberty:
We will be dropping centos7 as it is heading out of support and is ancient. Security concerns require manual patching at this point

rovinski:
Are we going to be adding CentOS 9 Stream support? It's important to have at least 1 CentOS OS in the CI

maliberty:
It's not clear the uptake of stream.  We are thinking opensuse. We will likely add build on rhel and slse for pre built binaries

maliberty:
Given that stream is not upstream of rhel anymore it isn't clear it performs that function.

rovinski:
Not sure what you mean? CentOS stream is just upstream of RHEL instead of downstream like CentOS 7/8. https://www.redhat.com/en/topics/linux/what-is-centos-stream
They both work off the same codebase and from what I read they are kept pretty well in-sync.

maliberty:
It isn't clear how the stream/rocky/alma split will play out in practice.

rovinski:
I guess the main question is will there be some platform that takes over the role of CentOS 7? I don't think OpenSUSE is viable for this purpose, because it needs to be something which lends to compatibility to RHEL, whether upstream or downstream. I think the furor over CentOS Stream vs. RHEL is overblown. And especially for CI purposes, there's probably negligible differences between all of them.

vvbandeira:
CI is moving to use docker images. The current plan is to run the full suite of tests on Ubuntu 22 (or the latest LTS version available) and minimal tests (build + one small design) for other OSes we choose to support. As Matt mentioned, we will test on RHEL and SUSE for the prebuilt binaries. We can consider adding RHEL to the minimal tests in CI as RH releases [UBI](https://developers.redhat.com/blog/2020/03/24/red-hat-universal-base-images-for-docker-users#), which should be enough for our CI needs.

rovinski:
@vvbandeira ok thanks. I just wanted to make sure support/CI for the RHEL ecosystem was disappearing completely. This plan sounds fine to me.

maliberty:
We will drop centos7 once the docker CI transition is complete.  The plan is stream8/rhel8 should be supported but not 7 as it is EOL.

PR #4702, opened 2024-02-25
etc: fix path in test delta script

habibayassin:


etc/test_delta_debug.sh
```diff
@@ -17,4 +19,4 @@ mkdir test_delta_debug
 cd test_delta_debug
 tar --strip-components=1 -xzf ../$latest_file
 sed -i 's/openroad -no_init/openroad -exit -no_init/g' run-me-$testname-asap7-base.sh
-openroad -exit -python ~/OpenROAD-flow-scripts/tools/OpenROAD/etc/deltaDebug.py --persistence 3 --use_stdout --error_string "Iter: 100 " --base_db_path results/asap7/$testname/base/3_2_place_iop.odb --step ./run-me-$testname-asap7-base.sh --multiplier 2
+openroad -exit -python ../tools/OpenROAD/etc/deltaDebug.py --persistence 3 --use_stdout --error_string "Iter: 100 " --base_db_path results/asap7/$testname/base/3_2_place_iop.odb --step ./run-me-$testname-asap7-base.sh --multiplier 2
```
vvbandeira, line 22: Shouldn't it be two levels above? i.e., `../../tools/OpenROAD`?
In line 8 we change to `<ORFS>/flow`.
In line 16-19 er create `test_delta_debug` inside `flow` so we would be at `<ORFS>/flow/test_delta_debug` to get to `<ORFS>/tools/OpenROAD` we would need `../../` right?

habibayassin: yes makes sense

vvbandeira (approved):

PR #4704, opened 2024-02-26
mpl2: fix boundary penalty and bug in macro placement

AcKoucher:
The way that the boundary penalty is currently implemented causes SA to push the macros based on the outline of the level in which it's currently working instead of pushing them to the core boundaries. This leads many macros to be placed near the center in many designs, deteriorating QoR.

These changes are to ensure that macros are pushed in the direction of the design's boundaries.

I'm also making another fix here. It looks like there was a bug that was probably introduced by copy and pasting code during the original development:
- Removing the varying factor applied to outline during macro placement && ensuring we actually use run_id == 0 in the first SA thread build.


AcKoucher:
@maliberty CI Results:
[ERROR] finish__timing__drv__hold_violation_count fail test: 182.0 <= 141.0

Update metrics?

maliberty:
> Update metrics?

If that's the only issue then just update.

maliberty (approved):
@eder-matheus you can merge once your update the metrics.  I have no code comments.

eder-matheus:
@maliberty PR with the metrics: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1823

PR #4705, opened 2024-02-26
mpl2: attempt floorplan centralization after SA steps in cluster placement

AcKoucher:
The sequence pair representation for the floorplan uses the lower left origin of the current outline as reference for the location of the clusters. There are many situations in which the floorplan not only fits the outline, but its height/width are considerably smaller than the outline's width/height.


AcKoucher:
This change depends on #4704.

AcKoucher:
Now that both the penalty for overlapping between clusters with macros / macro blockages (#4206) and the penalty for pushing macros to the boundaries of the core (#4815) are fixed. It's possible to benefit from the changes in this PR.

Running Secure-CI.

maliberty (approved):
Looks fine - pending CI results

AcKoucher:
@maliberty 

Secure-CI showed a degradation in the cts hold buffer count for  for asap7/aes-block.

```
MASTER [INFO] cts__design__instance__count__hold_buffer pass test: 966.0 <= 990.0
MY-BRANCH [ERROR] cts__design__instance__count__hold_buffer fail test: 1104.0 <= 990.0
```

After some digging I found out that these changes were messing up the behavior of the function that fills the dead space by expanding mixed clusters (`void SACoreSoftMacro::fillDeadSpace()`).

 Apparently this happens, because, if we revert the attempted centralization by arithmetic's operations, a tiny floating-point divergence might appear between the end of one SoftMacro and the beginning of the next one when we construct the grid that will be used to fill the space (we use a `std::set<float>`..). So I'm pushing changes to cache the SoftMacros' locations if reverting is needed to avoid this problem.
  
 Ideally, we'd have everything being computed in dbu...
 
Obs: Taking a closer look at how this mechanism works, turns out maybe this will be a more targeted approach for mock-array than a general improvement for our designs.

 I'll talk about it more in the R&D synch to not overload this conversation.
 
Starting a new Secure-CI run.

maliberty:
ci results?

AcKoucher:
CI passed

PR #4713, opened 2024-02-28
mpl2: add debug flag to draw bundled nets during cluster and macro placement

AcKoucher:


maliberty (approved):
Do you have a screen shot of what this looks like?  Should we try to indicate the weight of the bundled net in some way?

AcKoucher:
For three different levels in nangate45/bp_multi:







This last image reveals what seems to be a bug, because the center pin is not exactly on the center of the hard macros.

AcKoucher:
I think knowing the weight of each net can be useful. Perhaps just draw the value in the middle? Include in this PR?

maliberty:
> I think knowing the weight of each net can be useful. Perhaps just draw the value in the middle? Include in this PR?

You could use line width, color, or text label.  I'll merge this one and leave it up to you if you want to address it now or another time.

PR #4728, opened 2024-03-02
Grt: Improving the incremental GRT runtime

luis201420:
Improving the incremental GRT runtime.
- Avoiding checking and iterating all the nets in each Fastrouter step.
- Avoiding calling the checkPinPlacement function in each incremental GRT iteration.


src/grt/src/fastroute/src/utility.cpp
```diff
@@ -1292,7 +1256,7 @@ void FastRouteCore::StNetOrder()
 
   // Set the 70% (or less) of non critical nets that doesn't have overflow
   // with the lowest priority
-  for (int ord_elID = 0; ord_elID < netCount(); ord_elID++) {
+  for (int ord_elID = 0; ord_elID < dirty_net_ids_.size(); ord_elID++) {
```
eder-matheus, line 1259: Why use the dirty_net_ids_ size here instead of iterating over each vector item, like everywhere else?

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2022,7 +1955,7 @@ void FastRouteCore::copyBR(void)
         v_edges_[i][j].usage = 0;
       }
     }
-    for (netID = 0; netID < netCount(); netID++) {
+    for (int netID = 0; netID < netCount(); netID++) {
```
eder-matheus, line 1958: Similar question here. Why use `netCount()` if you have the vector with the net IDs?

src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -622,6 +621,7 @@ class FastRouteCore
 
   std::set<std::pair<int, int>> h_used_ggrid_;
   std::set<std::pair<int, int>> v_used_ggrid_;
+  std::vector<int> dirty_net_ids_;
```
eder-matheus, line 624: You can move this vector to the line after the `std::vector<FrNet*> nets_;` definition. Also, I think the name could be only `net_ids_`. FastRouteCore doesn't know about the "dirty nets" concept. It only makes sense in the GlobalRouter class.

eder-matheus (changes requested):

eder-matheus (approved):

PR #4730, opened 2024-03-02
Add in compile flags in splash

### LINKED ISSUE #4550: Add command for displaying all of the options OR was compiled with.
### Description

It would be nice to have a way to know whether or not we're running with an exclusive feature such as Charts, which is not supported in CentOS 7.

### Suggested Solution

```
OpenROAD v2.0-11454-g34687e2a9
Features included (+) or not (-):
-charts -eqy +GPU +GUI +Python  ...
```

### Additional Context

_No response_

luarss:
Fixes  #4550 

Output
image


rovinski:
Thanks, @luarss. Please add one more for QtCharts, because it's possible to compile the GUI without charts. Also, please list options alphabetically.

luarss:
New version
image

rovinski:
GUI after GPU

CMakeLists.txt
```diff
@@ -153,3 +154,16 @@ if(ENABLE_TESTS)
 endif()
 
 add_subdirectory(src)
+
+# After all compiled, look for the respective flags
+if(NOT OPENROAD_GPU_COMPILED)
```
maliberty, line 159: Why is this condition needed?

include/ord/OpenRoad.hh
```diff
@@ -250,6 +250,11 @@ class OpenRoad
   static const char* getVersion();
   static const char* getGitDescribe();
 
+  static const char* getGPUCompileOption();
+  static const char* getPythonCompileOption();
+  static const char* getGUICompileOption();
+  static const char* getChartsCompileOption();
```
maliberty, line 256: It would be nicer if these returned a bool

src/Main.cc
```diff
@@ -495,6 +495,14 @@ static void showSplash()
   logger->report("OpenROAD {} {}",
                  ord::OpenRoad::getVersion(),
                  ord::OpenRoad::getGitDescribe());
+  logger->report("Features included (+) or not (-):\n"
+                "{} {} {} {}",
+                strcmp(ord::OpenRoad::getChartsCompileOption(), "ON") == 0 ? "+Charts" 
+                                                                           : "-Charts",
```
maliberty, line 501: move the common portion to the format string eg

"{}Charts", ord::OpenRoad::getChartsCompileOption() ? '+' : '-'

oharboe:
Can we modify ORFS so it lists splash screen once?

CMakeLists.txt
luarss: I wrote it to be consistent with the same syntax as the above commands. 
image

luarss:
Updated. 
image

gadfort:
Would it make sense to make this a single line since it's not that many items? Also should par/mpl2 be on the list since they are conditionally compiled for MacOS?

maliberty:
> Can we modify ORFS so it lists splash screen once?

Each step is a separate run so I think it is best to record it.  You could in principle restart the flow at any step with a new build giving a mixed result.

oharboe:
> > Can we modify ORFS so it lists splash screen once?
> 
> Each step is a separate run so I think it is best to record it. You could in principle restart the flow at any step with a new build giving a mixed result.

While it would be nice to reduce logging noise, idempotency of the steps is  a requirement....

Minimally the version is needed in every log. Dumping copyright only to console once would be Ok.

luarss:
Just to clarify, I see `versions.txt` recording the OR, yosys and klayout versions already. Do we want something like 
`@echo openroad -splash` in ORFS Makefile?

luarss:
image

maliberty:
I don't feel the copyright is excessive.  If we really want to save a line you could be the features after the commit id.

AcKoucher:
> Addresses https://github.com/The-OpenROAD-Project/OpenROAD/issues/4550

@luarss FYI It's better to use one of the special keywords such as "fix" or "resolve" to link the PR with the issue so that when merged the issue will be closed: [Linking a PR to an Issue in GH](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

maliberty:
fails on mac

rovinski:
Doesn't this solution feel like overkill? Why does it need a whole CMake generated config file? Can't you just simply use `add_definition()` or similar and then reference that directly?

luarss:
@maliberty 
The error is because we disabled compile definitions for macos in `src/CMakeLists`.

image. Fixing this with an else init.

@rovinski That is a good suggestion. Is there any preference for using `add_compile_definitions` vs `add_definitions`?

rovinski:
> @rovinski That is a good suggestion. Is there any preference for using `add_compile_definitions` vs `add_definitions`?

Just whichever one works and simplifies the code.

src/CMakeLists.txt
```diff
@@ -357,6 +357,8 @@ if (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   add_subdirectory(par)
 else()
   message(STATUS "Removing MPL2 and PAR to avoid run time fatal error.")
+  target_compile_definitions(openroad PRIVATE ENABLE_MPL2=0)
+  target_compile_definitions(openroad PRIVATE ENABLE_PAR=0)
```
rovinski, line 361: I don't think this helps. The code checks whether it's defined, not what the value is.

src/Main.cc
```diff
@@ -491,10 +491,29 @@ static void showUsage(const char* prog, const char* init_filename)
...
+#ifdef ENABLE_MPL2
+  bool enable_mpl2 = true;
+#else
+  bool enable_mpl2 = false;
+#endif
+#ifdef ENABLE_PAR
+  bool enable_par = true;
+#else
+  bool enable_par = false;
+#endif
```
luarss, line 503: @rovinski Not sure if this is the best way to do it but we can repeat the same for the other headers too were we to avoid creating new cmake generated files

rovinski: I think you might be able to simplify this with a macro. Also use `static constexpr`
Something like:
```
#define IS_COMPILED(DEF,VAR) \
  #ifdef DEF \
    static constexpr VAR = true; \
  #else \
    static constexpr VAR = false; \
  #endif

IS_COMPILED(ENABLE_PAR,enable_par)
IS_COMPILED(ENABLE_MPL2,enable_mpl2)
...
```

maliberty:
> Doesn't this solution feel like overkill? Why does it need a whole CMake generated config file? Can't you just simply use add_definition() or similar and then reference that directly?

I agree with this.  The file generation isn't needed.

luarss:
@maliberty Updated, now uses less overhead

PR #4733, opened 2024-03-04
Add APIs for LEF macro ANTENNADIFFAREA and ANTENNAGATEAREA 

kareefardi:
* Add dbTechAntennaPinModel.hasGateArea
* Add dbMTerm.hasDiffArea

---

The incentive behind this PR is to access the antenna information using python APIs. Currently, I was unable to call `getDiffArea` and `getGateArea` using the python APIs as they require passing a specific object which I was unable to create.

### DIFF
--- a/src/odb/src/db/dbTechLayerAntennaRule.cpp
+++ b/src/odb/src/db/dbTechLayerAntennaRule.cpp
@@ -985,6 +985,16 @@ void _dbTechAntennaPinModel::getAntennaValues(
   }
 }
 
+bool dbTechAntennaPinModel::hasGateArea()
+{
+  vector<std::pair<double, dbTechLayer*>> gate_areas;
+  getGateArea(gate_areas);
+  if (gate_areas.size() > 0)
+    return true;
+  else
+    return false;
+}
+
 void dbTechAntennaPinModel::getGateArea(
     std::vector<std::pair<double, dbTechLayer*>>& data)
 {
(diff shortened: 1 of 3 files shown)


src/odb/src/db/dbTechLayerAntennaRule.cpp
```diff
@@ -985,6 +985,16 @@ void _dbTechAntennaPinModel::getAntennaValues(
...
 
+bool dbTechAntennaPinModel::hasGateArea()
+{
+  vector<std::pair<double, dbTechLayer*>> gate_areas;
+  getGateArea(gate_areas);
+  if (!gate_areas.empty()) {
+    return true;
+  }
+  return false;
+}
```
rovinski, line 996:
```suggestion
  return !gate_areas.empty();
```

maliberty:
It would be better to just fix the python API.  This still doesn't get you access to the data.

kareefardi:
@maliberty I tried to do something like 
```
bool dbTechAntennaPinModel::hasGateArea()
{
  vector<std::pair<double, dbTechLayer*>> gate_areas;
  gate_areas = getGateArea();
  return !gate_areas.empty();
}
```
but I would get warnings:
```
swig/python detected a memory leak of type 'std::vector< std::pair< double,odb::dbTechLayer * >,std::allocator< std::pair< double,odb::dbTechLayer * > > > *', no destructor found.
```
Furthermore, the returned object is still inaccessible `<Swig Object of type 'std::vector< std::pair< double,odb::dbTechLayer * >,std::allocator< std::pair< double,odb::dbTechLayer * > > > *' at 0x7f815092aa30>`.

In addition, I am not really after the data itself. I am only checking for it's existence. Having said that, I wouldn't mind fixing the APIs that returns the data for Python but it is not something I am good with so it would be nice if you can give me some pointers on doing that or I can leave it up to the OpenROAD team to handle it and close this PR.

maliberty:
You would need to add a swig typemap to handle the collection. in odb/src/swig you can see how we have handled other such.  In general messing with swig is headache but great when it finally works.

PR #4735, opened 2024-03-04
Report total usage of cells on report_cell_usage

arthurjolo:
Fixes [#1695](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1695)

Adds a report of the total usage of cells to the command report_cell_usage

### DIFF
--- a/src/dbSta/test/report_cell_usage.ok
+++ b/src/dbSta/test/report_cell_usage.ok
@@ -11,5 +11,5 @@ Cell usage report:
   Fill cells: 168
   Buffer/inverters: 232
   Sequential cells: 35
-  Complex combinational cells: 241
+  Multi-Input combinational cells: 241
   Total: 676


stefanottili:
Complex combinational cells. (vs sequential).

Just out of curiosity, what is complex supposed to indicate ? 
Are there 'non complex' combinational cells ?
If not, why not remove "Complex" ?

maliberty:
Perhaps "Multi-input" instead of "Complex"?

stefanottili:
I would argue that ‘combinatorial’ implies that there are multiple inputs to be combined. Just remove ‘complex’.

maliberty:
> I would argue that ‘combinatorial’ implies that there are multiple inputs to be combined. Just remove ‘complex’.

You don't consider an inverter to be combinational?

stefanottili:
I don’t recall an inverter to be categorized as a combinatorial cell.
Especially not if there is a category buffer/inverter just two lines above.

stefanottili:
but according to Toshiba, if you divide cmos into sequential and combinatorial, buffer/inverter are combinatorial.

https://toshiba.semicon-storage.com/us/semiconductor/knowledge/e-learning/cmos-logic-basics/chap3/chap3-1.html

maliberty:
https://en.wikipedia.org/wiki/Combinational_logic splits the same way and it is the usage I'm used to.

maliberty:
@dalyles any comment on the naming discussion above?

dalyles:
> @dalyles any comment on the naming discussion above?

I would argue that combinational logic also includes inverters and buffers.

https://www.allaboutcircuits.com/textbook/digital/chpt-9/combinational-logic-functions/

According to this source, the term combinational came from mathematics and implies that
the order of the inputs coming into the circuits do not matter because the output result
is the same regardless of the input order. 

Combinational is not about the number of the inputs but the order of the inputs. 
Both have no impact of the end result anyway. 

Also, an unordered set of inputs can also have just one element.

maliberty:
@dalyles do you have a preference between "Complex" and "Multi-input" ?

dalyles:
@maliberty I prefer "Multi-Input." 

Unlike "complex" it's unambiguous and clearly describes the type of cell, and it will be much clearer for other users who come across the cell count report.

Thanks!

dalyles:
Should I wait until the pr-merge is complete to update ORFS?

dalyles:
One detail that we requested was to have the total cell count
outputted in the output file <code>6_report.log</code> along with the rest of
the cell count information.

image

we want a total count here so we don't have to tally the numbers up ourselves.
Thanks :)

dalyles:
Can you please provide a screenshot of a sample run, like gcd?
I reinstalled the repo and gave it a run, but i still do not see a 
cell count with the total tallied at the bottom.

arthurjolo:
I just checked and the OpenROAD tool on ORFS is still outdated, but an example output is:

```
Cell usage report:
  Fill cells: 168
  Buffer/inverters: 232
  Sequential cells: 35
  Complex combinational cells: 241
  Total: 676
```

dalyles:
Ah I see. Please let me know once the tool gets updated so I can sync on my end and retry the run.

arthurjolo:
Yes I will take a look into that and will get back to you.

maliberty:
You can use the --latest flag to build.sh to use the head of OpenROAD in ORFS.  That will let you get this change to try out.

dalyles:
Success! It works!!

PR #4737, opened 2024-03-05
gui: use iterator insert for rtrees

gadfort:


gadfort:
@maliberty I don't see a particular drop in loading time, so I'm wondering if you see anything that could be further improved.

src/gui/src/search.cpp
```diff
@@ -257,10 +257,22 @@ void Search::updateShapes(odb::dbBlock* block)
...
+  LayerMap<std::vector<SNetValue<odb::dbNet*>>> snet_shapes;
+  LayerMap<std::vector<SNetSBoxValue<odb::dbNet*>>> snet_net_via_shapes;
   for (odb::dbNet* net : block->getNets()) {
-    addNet(net);
-    addSNet(net);
+    addNet(net, net_shapes);
+    addSNet(net, snet_shapes, snet_net_via_shapes);
   }
+  for (const auto& [layer, layer_shapes] : snet_shapes) {
+    data.snet_shapes_[layer].insert(layer_shapes.begin(), layer_shapes.end());
```
maliberty, line 268: Rather than insert I think you need to call the ctor.  In drt it does:

```
  for (auto i = 0; i < numLayers; i++) {
    shapes_.at(i) = boost::move(RTree<frBlockObject*>(allShapes.at(i)));
```

gadfort: thanks.

gadfort:
@maliberty this dropped the initial render time from ~19 seconds to 1.1 seconds for one of our designs

src/gui/src/search.cpp
```diff
@@ -257,10 +257,23 @@ void Search::updateShapes(odb::dbBlock* block)
...
 
+  LayerMap<std::vector<RouteBoxValue<odb::dbNet*>>> net_shapes;
+  LayerMap<std::vector<SNetValue<odb::dbNet*>>> snet_shapes;
+  LayerMap<std::vector<SNetSBoxValue<odb::dbNet*>>> snet_net_via_shapes;
   for (odb::dbNet* net : block->getNets()) {
-    addNet(net);
-    addSNet(net);
+    addNet(net, net_shapes);
+    addSNet(net, snet_shapes, snet_net_via_shapes);
   }
```
maliberty, line 266: The memory peak can be lowered by moving splitting this loop to only handle snet here and separately handle nets after ` snet_net_via_shapes.clear();`.  The extra loop over all nets should be minor in runtime.

maliberty (approved):

PR #4738, opened 2024-03-05
grt: fix gridsize calculation if diesize is smaller than tilesize

TheMightyDuckOfDoom:
GRT: Make sure that the grid size is at least 1 by 1.

With small designs the tile size could be larger than the width or height of the die area, for example in a single row macro.

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -2477,8 +2477,8 @@ void GlobalRouter::initGrid(int max_layer)
 
   int tile_size = grid_->getPitchesInTile() * track_spacing;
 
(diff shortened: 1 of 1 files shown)


eder-matheus:
@TheMightyDuckOfDoom Thanks for your contribution! Could you add a unit test for this case?

TheMightyDuckOfDoom:
I added a regression test.

Without the fix it fails as expected, while with the fix it should work.

TheMightyDuckOfDoom:
I missed to add a .ok file for the regression test, but have added it now.

rovinski:
> I added a regression test.
> 
> Without the fix it fails as expected, while with the fix it should work.

Extra credit for a nice, concise test case 👏

PR #4739, opened 2024-03-05
rsz: update repair_design command with skip_resizing flag

eder-matheus:


src/rsz/src/RepairDesign.cc
```diff
@@ -354,7 +356,8 @@ RepairDesign::repairNet(Net *net,
                         int &slew_violations,
                         int &cap_violations,
                         int &fanout_violations,
-                        int &length_violations)
+                        int &length_violations,
+                        bool skip_driver_resizing)
```
maliberty, line 360: I find the name confusing.  resize_drvr already controls driver resizing - this seems more about skipping resizeToTargetSlew specifically.  Perhaps skip_target_slew_resizing?

eder-matheus: Thanks, it is a better name. Should the TCL option be renamed as well? It may make the purpose of the flag more explicit, but at the same time, it might be too specific for users to understand.

PR #4741, opened 2024-03-06
Update minimum SWIG version in `src/CMakeLists.txt`

donn:
As per https://github.com/The-OpenROAD-Project/OpenROAD/issues/4401#issuecomment-1980207992, "swig3 is ancient and we don't support it. Please use swig 4".

This updates the build files to reflect that statement.


maliberty (approved):
Thanks

PR #4745, opened 2024-03-06
grt: check max routing layer before fetching blockage data

eder-matheus:
Fix two public CI crashes.


eder-matheus:
@vvbandeira FYI.

maliberty (approved):

PR #4746, opened 2024-03-07
rsz: update repairNet to avoid multiple resizing

eder-matheus:


eder-matheus:
@maliberty This update fixes the issue @AcKoucher is facing in his mpl2 update. I've started a secure-ci for this and created the PR to begin the code revision.

src/rsz/src/RepairDesign.cc
```diff
@@ -391,92 +387,47 @@ RepairDesign::repairNet(Net *net,
...
+                                      cap_violations,
+                                      length_violations);
+        
+        if (need_repair) {
           if (resize_drvr) {
             resize_count_ += resizer_->resizeToTargetSlew(drvr_pin);
           }
+          wire_length = bnet->maxLoadWireLength();
+          need_repair = needRepair(drvr_pin,
+                                   corner,
```
maliberty, line 408: If we don't do resizeToTargetSlew there is no need to redo needRepair so this call can be moved inside the if.

src/rsz/src/RepairDesign.cc
```diff
@@ -486,6 +437,132 @@ RepairDesign::repairNet(Net *net,
...
+                         const int wire_length,
+                         const bool check_cap,
+                         const bool check_slew,
+                         float& max_cap,
+                         int &slew_violations,
+                         int &cap_violations,
+                         int &length_violations)
+{
+  bool repair_cap = false;
+  bool repair_wire = false;
```
maliberty, line 545: repair_wire  can be defined below as there is no if condition for it.

maliberty (changes requested):
The reduction in changes looks good.

src/rsz/test/repair_design3_verbose.ok
```diff
@@ -14,16 +14,15 @@
...
+      200 |       0 |       0 |             0 |       802
+      300 |       0 |       0 |             0 |       702
+      400 |       0 |       0 |             0 |       602
+      500 |       0 |       0 |             0 |       502
+      600 |       0 |       0 |             0 |       402
+      700 |       0 |       0 |             0 |       302
+      800 |       0 |       0 |             0 |       202
+      900 |       0 |       0 |             0 |       102
+     1000 |       0 |       0 |             0 |         2
+    final |       0 |       0 |             0 |         0
```
maliberty, line 27: Is it right that there are zero changes here?

eder-matheus: If I understand it correctly, this design has no violations to repair, so the resizing is never executed. That's why we have zeros now.

maliberty (approved):
The code looks fine; just needs good CI results.

eder-matheus:
@maliberty In the public CI, we have only two failures
* asap7/aes-block: [ERROR] finish__timing__drv__setup_violation_count fail test: 417.0 <= 376.0
* nangate45/aes: [ERROR] finish__timing__setup__ws fail test: -0.273632 >= -0.17

The private pipelines haven't started yet.

PR #4747, opened 2024-03-07
fix directory

habibayassin:


etc/test_delta_debug.sh
```diff
@@ -5,7 +5,8 @@
 # variable names such as $(false), unused variables, etc.
 set -ue -o pipefail
 
-cd "$(dirname $(readlink -f $0))/../flow"
+cd "$(dirname $(readlink -f $0))"
+cd ../../../flow
```
vvbandeira, line 9: Is this script exclusive to ORFS?

vvbandeira:
@oharboe
I might have missed the discussion, but why is this script in the OR and not ORFS repo?

maliberty:
It can be used to reduce any OR test case, not just ORFS.

etc/test_delta_debug.sh
habibayassin: it seems like it is mainly targeting ORFS

oharboe: To clarify: this integration test is testing the generic deltaDebug.py, which is independent of ORFS, using test-data from ORFS.

oharboe: Perhaps the integration test of deltaDebug.py could be moved to ORFS? Lots of OpenROAD integration testing happens in ORFS.

vvbandeira (approved):

PR #4754, opened 2024-03-10
dbsta: swap pin ordering to make iterms handling faster

gadfort:
Since there are far more Iterms in a design, this avoids calling `.size()` when handling iterms speeding up `::id()`.

### DIFF
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -582,10 +582,10 @@ ObjectId dbNetwork::id(const Pin* pin) const
   dbBTerm* bterm;
   staToDb(pin, iterm, bterm);
 
-  if (iterm) {
-    return iterm->getId() + iterm->getBlock()->getBTerms().size();
+  if (iterm != nullptr) {
+    return iterm->getId();
   }
-  return bterm->getId();
+  return bterm->getId() + bterm->getBlock()->getITerms().size();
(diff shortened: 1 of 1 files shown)


maliberty (approved):

gadfort:
@maliberty I think this answers the question whether timing/timing repair is sensitive to the pin ordering

maliberty:
Have you looked at the failures to see if they are just text differences or actual changes in results?

maliberty:
I won't be surprised if there is an order dependence

maliberty:
We don't generally create bterms so I think this why the old way works.

gadfort:
@maliberty PDNgen can makes bterms

maliberty:
How about 
```
  if (iterm != nullptr) {
    return iterm->getId() << 1;
  }
  return (bterm->getId() << 1) + 1;
```
Does that help with performance?  Shifting should be pretty fast.

maliberty:
> @maliberty PDNgen can makes bterms

We don't use timing in pdngen.  I'm not saying it is good, just why we get away with it.

gadfort:
@maliberty I added your suggestion and it drops the call time to `::id()` to ~200ms from vtune, so between the sta changes and this most of the runtime for report_power has been brought back to normal.

Thanks

maliberty (approved):

PR #4766, opened 2024-03-11
meta: Moving tool tests to ctest

QuantamHD:
Now we can run all the regression tests in parallel.

```
$ ctest -j 128
...
721/740 Test #501: grt.congestion5 ...............................................   Passed   64.21 sec
722/740 Test #503: grt.congestion7 ...............................................   Passed   70.06 sec
723/740 Test #488: cts.array_no_blockages ........................................   Passed   71.07 sec
724/740 Test #342: dpo.aes .......................................................   Passed   61.13 sec
725/740 Test #177: rsz.repair_fanout6 ............................................   Passed   73.91 sec
726/740 Test #240: rsz.repair_design3_verbose ....................................   Passed   81.89 sec

100% tests passed

Label Time Summary:
IntegrationTest    = 4664.65 sec*proc (714 tests)

Total Test time (real) = 354.23 sec
```

Flow tests haven't been moved, because I'm not sure how they should be handled. I figure someone else can figure that out.


maliberty:
@vvbandeira please take a look and comment on how we can integrate this with the CI.

gadfort:
@maliberty @QuantamHD this will require maintaining two lists of CI tests. @QuantamHD would it be possible to source the regression_test.tcl file to build the test list in CMake? I fear that otherwise the list will grow stale or add an un-needed overhead to adding testing

QuantamHD:
I did think about that, but I think long term we want to transition to CTest fully, and drop the lists in TCL.

I can probably add a mode that defines the test list from an env var.

maliberty:
> I did think about that, but I think long term we want to transition to CTest fully, and drop the lists in TCL.

I agree about dropping TCL.  My hope is the window of overlap is short.  

Can we wildcard test/*tcl (*py) rather than list them all explictly?  We would need a way to manually exclude some but I'd rather default to running.  I just learned today there are some gpl tests that we don't run and I'm not sure if there is any good reason why not.

QuantamHD:
I think both tcl and py tests get run automatically if it's built with python support

vvbandeira:
> I did think about that, but I think long term we want to transition to CTest fully, and drop the lists in TCL.
> 
> I can probably add a mode that defines the test list from an env var.

What if we use CTest in CI for the converted tools and remove their `tcl` list files? This way we can do the move incrementally and avoid keeping two lists of tests.

> I just learned today there are some gpl tests that we don't run and I'm not sure if there is any good reason why not.

Not sure if these are the ones you are talking about, but there are some `gpl` tests that we run on a separate pipeline and not for every PR because they take too long.

QuantamHD:
> > I did think about that, but I think long term we want to transition to CTest fully, and drop the lists in TCL.
> > I can probably add a mode that defines the test list from an env var.
> 
> What if we use CTest in CI for the converted tools and remove their `tcl` list files? This way we can do the move incrementally and avoid keeping two lists of tests.
> 

Sounds great to me! I think we would just need a slight change to the root level regression_var.tcl to read the test name in from an env var.

```tcl
# Record tests in the /test directory.
# Compare results/<test>.log to <test>.ok for pass/fail.
proc record_tests { tests } {
  record_tests1 $tests "compare_logfile"
}

# Record tests in the /test directory.
# Last line of results/<test>.log should be pass/fail.
proc record_pass_fail_tests { tests } {
  record_tests1 $tests "pass_fail"
}

proc record_flow_tests { tests } {
  record_tests1 $tests "check_metrics"
  define_test_group "flow" $tests
}

proc record_tests1 { tests cmp_logfile } {
  global test_dir
  foreach test $tests {
    # Prune commented tests from the list.
    if { [string index $test 0] != "#" } {
      record_test $test $test_dir $cmp_logfile
    }
  }
}
```

Without doing so no tests will appear as available to run.

gadfort:
I would like to which to CMake as well, seems easier, I think if a test is not in the regression_tests.tcl the ./regression fails, might just need a simple fix to by pass that

maliberty:
> Not sure if these are the ones you are talking about, but there are some gpl tests that we run on a separate pipeline and not for every PR because they take too long.

No there are some that are in neither (eg routability)

QuantamHD:
I pushed a change that will allow CTest to not have to rely on the definitions in regression_vars.tcl file. It still depends on the file calling record_tests which can be removed once we fully transition over to ctest.

QuantamHD:
@maliberty seems like a weird failure in the CI

maliberty:
#4759 just updated the same metric - perhaps you need a merge?  I have another PR coming for a parallel conflict in rsz.

QuantamHD:
I'll wait to merge

QuantamHD:
rebased on master

gadfort:
@QuantamHD can you check that it will still compile with ENABLE_TESTS=OFF?

maliberty:
@vvbandeira please switch this CI to use this runner.  Once we have done so we can start to remove the old TCL files.  I think there is some more work but it can be done in subsequent PRs.

PR #4767, opened 2024-03-11
gui: resize viewer based on its actual size rather than the padded rect dimensions

### LINKED ISSUE #4033: zoom in/out to the position of the mouse in the view, not the view center

AcKoucher:
Fix #4033

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -285,6 +285,10 @@ void LayoutViewer::setPixelsPerDBU(qreal pixels_per_dbu)
   qreal target_pixels_per_dbu
       = std::min(pixels_per_dbu, maximum_pixels_per_dbu_);
 
+  if (target_pixels_per_dbu == maximum_pixels_per_dbu_) {
+    return;
+  }
+
   const QSize new_size(ceil(current_viewer.dx() * target_pixels_per_dbu),
                        ceil(current_viewer.dy() * target_pixels_per_dbu));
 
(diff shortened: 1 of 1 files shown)


maliberty:
@stefanottili would you try this out and see if it satisfies your request?

stefanottili:
Nope, doesn't fix the issue.

Zoom into an area close to the chip boundary.
Use shift-Z to zoom-out to where your mouse is pointing.
Keep on zooming-out.

Eventually you hit an artificial OpenROAD "you're not allowed to zoom out any further" limit and the view shifts from under your mouse and then you can't zoom out any further with only half the chip in view.

The fix would be to completely remove the limit or at least allow zoom-out until the chip only covers a fifth of the width/height of the view.

If in doubt about OR gui behavior, compare it to klayout.
Its gui controls, mouse focus behavior and draw speed have seen decades of refinement and optimization.

AcKoucher:
@stefanottili Did you test with the changes in this PR's branch?

stefanottili:
yes, of course.

Use the "draw rectangle to zoom into" gesture to zoom into any corner of a block, so that the corner is in the middle of the view. Then place the mouse on the corner in the middle of the view and hit "shift-z" to zoom out. This keeps the corner in the middle of the view only up to the point where the size of the block matches the size of the view (fit) and the view shifts.
it wouldn't if the artificial "can't zoom out further then this" limit wouldn't kick in. See klayout.

maliberty (approved):

PR #4790, opened 2024-03-14
pdn: switch initial building of rtrees to vectors and use bulk loading

gadfort:
Changes:
- initial loading of shape and obstruction rtrees to bulk loading

I'm sure there are more places that can benefit from this in PDN.
For ZeroSOC this takes off ~50% of the build time for the power grid.

### DIFF
--- a/src/pdn/src/renderer.cpp
+++ b/src/pdn/src/renderer.cpp
@@ -74,32 +74,26 @@ void PDNRenderer::update()
   vias_.clear();
   repair_.clear();
 
+  ShapeVectorMap shapes;
+  ShapeVectorMap obs;
+  std::vector<ViaValue> vias;
   for (const auto& domain : pdn_->getDomains()) {
     for (auto* net : domain->getBlock()->getNets()) {
-      ShapeTreeMap net_shapes;
-      Shape::populateMapFromDb(net, net_shapes);
-      for (const auto& [layer, net_obs_layer] : net_shapes) {
-        auto& obs_layer = grid_obstructions_[layer];
-        for (const auto& [box, shape] : net_obs_layer) {
-          obs_layer.insert({shape->getObstructionBox(), shape});
-        }
-      }
+      Shape::populateMapFromDb(net, obs);
     }
 
     for (const auto& grid : domain->getGrids()) {
-      grid->getGridLevelObstructions(grid_obstructions_);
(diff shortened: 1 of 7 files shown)


src/pdn/src/sroute.cpp
```diff
@@ -666,15 +666,11 @@ void SRoute::createSrouteWires(
...
-      auto& obs_layer = obstructions[layer];
-      for (const auto& [box, shape] : net_obs_layer) {
-        obs_layer.insert({shape->getObstructionBox(), shape});
-      }
-    }
+    ShapeVectorMap obstructions_vec;
+    Shape::populateMapFromDb(net, obstructions_vec);
+    const ShapeTreeMap obstructions
+        = Shape::convertVectorToTree(obstructions_vec);
+    obstructions_vec.clear();
```
maliberty, line 673: Add shrink_to_fit() otherwise the capacity is unchanged  (or swap with an empty vector).

src/pdn/src/shape.cpp
```diff
@@ -617,6 +617,23 @@ Shape* Shape::extendTo(
   return new_shape.release();
 }
 
+ShapeTreeMap Shape::convertVectorToTree(const ShapeVectorMap& vec)
```
maliberty, line 620: Would it be better to clear/shrink the vec here rather than in each caller?

src/pdn/src/renderer.cpp
```diff
@@ -125,6 +119,11 @@ void PDNRenderer::update()
     }
   }
 
+  shapes_ = Shape::convertVectorToTree(shapes);
+  grid_obstructions_ = Shape::convertVectorToTree(obs);
+  shapes_ = Shape::convertVectorToTree(shapes);
+  vias_ = Shape::convertVectorToTree(vias);
```
maliberty, line 125: Need to clear/shrink as you go to minimize the memory peak (here or in convertVectorToTree)

maliberty (changes requested):

src/pdn/src/shape.cpp
gadfort: done

maliberty:
I just fixed it to my satisfaction rather than go around the loop again.

maliberty (approved):

PR #4797, opened 2024-03-15
gpl: add simple tests for routability mode in GPL

gudeh:
small modification on simple01-rd to include the target RC parameter. And included simple02-rd to test considering a lower target RC, which activates routability.

### DIFF
--- a/src/gpl/test/regression_tests.tcl
+++ b/src/gpl/test/regression_tests.tcl
@@ -7,7 +7,7 @@ record_tests {
   simple01-ref
   simple01-skip-io
   simple01-rd
-  simple02-rd  
+  simple02-rd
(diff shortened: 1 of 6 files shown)


src/gpl/test/regression_tests.tcl
```diff
@@ -7,6 +7,7 @@ record_tests {
   simple01-ref
   simple01-skip-io
   simple01-rd
+  simple02-rd
```
QuantamHD, line 10: Could you add it to CMakeLists.txt as well. We're moving away from this runner to CTest

QuantamHD (commented):
We're moving away from the

maliberty (approved):

Issue #4006, opened 2023-09-18
Improve the buffering of simple wires


maliberty:
### Describe the bug

The new [buffer_varying_lengths](https://github.com/The-OpenROAD-Project/OpenROAD/pull/4005#top) test for rsz shows cases where the buffering needs improvement.  Two-pin wires are increasing length are generated and buffered:

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/761514/04e3ff73-a8b3-4273-a393-d662035056ad)

The buffering is far from optimal with some wires showing one buffer right next to the driver.

### Expected Behavior

Optimal buffering for all lengths.

### Environment

```shell
Head of master
```

### To Reproduce

Run the test

maliberty:
For example moving instance wire15 to the middle (x=450um) reduces the slack from -0.14 to -0.12

PR #4801, opened 2024-03-15
Rsz update longe wire buffer insertion

arthurjolo:
Fixes [#4006](https://github.com/The-OpenROAD-Project/OpenROAD/issues/4006)

This PR changes the position of buffers inserted to fix long wires.

The new position is set to be the the middle of the wire, previously it was set to be on the max wire length allowed causing some buffers to be inserted near the drivers.


maliberty (approved):

src/rsz/src/RepairDesign.cc
```diff
@@ -851,7 +851,8 @@ void RepairDesign::repairNetWire(
                  level,
                  units_->distanceUnit()->asString(dbuToMeters(wire_length), 1),
                  units_->distanceUnit()->asString(dbuToMeters(max_length_), 1));
-      split_length = min(split_length, max(max_length_ - wire_length_ref, 0));
+      split_length = min(max_length_, length / 2);
```
povik, line 854: This reintroduces what was fixed in 3b97c566

maliberty: @arthurjolo please comment

arthurjolo: I believe this was overwritten during a merge. But I think it easy to fix: `split_length = min(length / 2, max(max_length_ - wire_length_ref, 0));`
This should account for both changes, @povik do you agree?

povik: Yes, that looks good to me! (I was lazy to type that out myself...)

maliberty: @arthurjolo please make a PR for your fix.

PR #4802, opened 2024-03-16
etc: use detachedDB in deltaDebug

habibayassin:


etc/deltaDebug.py
```diff
@@ -219,11 +220,6 @@ def perform_step(self, cut_index=-1):
             odb.write_def(self.base_db.getChip().getBlock(),
                           self.base_def_file)
 
-        # Destroy the DB in memory to avoid being out-of-memory when
-        # the step code is running
-        if (self.base_db is not None):
-            self.base_db.destroy(self.base_db)
-
```
maliberty, line 226: Why is this removed?

habibayassin: Because the database instance is being used later in the code, that’s why there was a segmentation fault

maliberty: Once the step is over it shouldn't be needed.  Where is it used later?  Why wasn't this an issue before?

habibayassin: get_cuts function gets called after perform_step, get_cuts calls get_elms which calls get_insts and get_nets and both of these functions access the database instance.
I do not know why it was not an issue before, however, from the CI, the deltaDebug test has never passed.

maliberty: @oharboe it looks like a bug you introduced in https://github.com/The-OpenROAD-Project/OpenROAD/commit/c18ba0d6172d9741acb4e9a4adabbd3503d423a2 where you added the `while j < self.get_cuts():` loop

maliberty: I don't think removing the call to destroy is the correct fix.  Probably it is better to gather whatever stats we need before the db is destroyed.

oharboe: Let me know if there is anything I can do to help. I dont think the python interface should ever segfault, even if there is a python error, right?

A helpful error message as the nature of the problem rather, from which it should be a small matter to fix the python.

The python code now has unit and integration tests, but did @vvbandeira hook them up yet?

etc/deltaDebug.py
```diff
@@ -203,7 +204,7 @@ def debug(self):
     # and calls the step function, then returns the stderr of the step.
     def perform_step(self, cut_index=-1):
         # read base db in memory
-        self.base_db = odb.dbDatabase.create()
+        self.base_db = Design.createDetachedDb()
```
oharboe, line 207: What is the significance of this change?

etc/deltaDebug.py
maliberty: Habiba is doing that now and hit this issue.  Accessing objects from a destroyed db is equivalent to access deallocated memory.

etc/deltaDebug.py
maliberty: Logger handling

etc/deltaDebug.py
```diff
@@ -187,6 +193,11 @@ def debug(self):
                 if err is None or self.get_cuts() == 0:
                     break
 
+        # Destroy the DB in memory to avoid being out-of-memory when
+        # the step code is running
+        if (self.base_db is not None):
+            self.base_db.destroy(self.base_db)
```
maliberty, line 199: Rather than moving the destroy call out of perform_step I think it would be better to capture the relevant info inside that step (num insts & nets) and return it to the caller.  Moving it here still increases the high memory point.

oharboe (changes requested):
I contibut

maliberty:
The point of this is that the test can't be run as they currently fail.  As I pointed out earlier you introduced a change that broke it.

maliberty:
The other option is to revert your change...

oharboe:
> The other option is to revert your change...

Is a SEGFAULT expected when using the Python interface? I would have thought not.

Would you like me to have a look?

etc/deltaDebug.py
```diff
@@ -18,6 +18,7 @@
 # where the script manipulates base.odb between steps to reduce its size.
 ################################
 
+from openroad import Design, Tech
```
oharboe, line 21: "Tech" is unused.

oharboe (changes requested):
add this to test_deltaDebug.py so that the unittest keeps working:

```
$ git diff
diff --git a/etc/test_deltaDebug.py b/etc/test_deltaDebug.py
index c1cc42b4e..af7c2f15b 100644
--- a/etc/test_deltaDebug.py
+++ b/etc/test_deltaDebug.py
@@ -5,6 +5,8 @@ from unittest.mock import patch, MagicMock
 
 odb = MagicMock()
 sys.modules['odb'] = odb
+openroad = MagicMock()
+sys.modules['openroad'] = openroad
 import deltaDebug
 
 default_args = argparse.Namespace(persistence=4,
```

oharboe:
@maliberty How do I reproduce the problem here?

I do find that this PR is needed to fix the logging problem and that #4847 is needed, but I don't get an error.

I tried:

```
$ cd OpenROAD
$ etc/test_delta_debug.sh
```

Output:

```
[deleted]
Step 141, Nets level debugging, Insts 13, Nets 3, cut elements 1, timeout 2 minutes
[#C#]
Step 142, Nets level debugging, Insts 13, Nets 3, cut elements 1, timeout 2 minutes
[##C]
___________________________________
Resultant file is results/asap7/uart/base/deltaDebug_base_result_3_2_place_iop.odb
Delta Debugging Done!
oyvind@gale:~/OpenROAD-flow-scripts/tools/OpenROAD$ echo $?
0
```

Unittests work fine too:

```
oyvind@gale:~/OpenROAD-flow-scripts/tools/OpenROAD/etc$ python -m unittest discover -s . -p "test_*.py"
[deleted]
Step 16, Nets level debugging, Insts 1, Nets 1, cut elements 1, timeout 2 minutes
[C]
___________________________________
Resultant file is /dummy/path/deltaDebug_base_result_base.db
Delta Debugging Done!
.
----------------------------------------------------------------------
Ran 6 tests in 0.229s

OK
```

maliberty:
> Is a SEGFAULT expected when using the Python interface? I would have thought not.

The database is destroyed but pointer is kept to it in python and subsequent APIs calls are made.  The fact that it is python doesn't remove the need to not use dead pointers.  It would be better to null out the pointer in python after destroy so that you get a python error instead.

oharboe:
> > Is a SEGFAULT expected when using the Python interface? I would have thought not.
> 
> The database is destroyed but pointer is kept to it in python and subsequent APIs calls are made. The fact that it is python doesn't remove the need to not use dead pointers. It would be better to null out the pointer in python after destroy so that you get a python error instead:
> 
> self.base_db = None

Makes sense. So both OpenROAD should be fixed(no SEGFAULT) and the Python code using the OpenROAD API.

I dont have access to  failing test case currently.

oharboe:
OpenROAD should implement a file open/close idiom for the database and any references to the database should throw an exception(not SEGFAULT) when referenced for a closed database.


# Implementing 'with' Concept in Python using SWIG

This guide explains how to use SWIG (Simplified Wrapper and Interface Generator) to implement the 'with' concept in Python, particularly for managing database connections.

## Step 1: Define a Context Manager in C++

First, implement a context manager in C++ for your database. This should manage the opening and closing of the database.

```cpp
class DatabaseManager {
public:
    DatabaseManager(/* ... */) { /* Open the database */ }
    ~DatabaseManager() { /* Close the database */ }

    DatabaseManager* __enter__() { return this; }
    void __exit__(/* ... */) { /* Close the database */ }

    // Other database methods
};
```

### SWIG Interface File Example

```cpp
%module databasemodule

%{
#include "DatabaseManager.h"
%}

%include "DatabaseManager.h"
```

### C++ Exception Handling Example

```cpp
void DatabaseManager::someMethod() {
    if (/* database is closed */) {
        throw std::runtime_error("Database is closed");
    }
    // Normal operation
}
```


### SWIG Exception Handling Example

```cpp
%exception {
    try {
        $action
    } catch (std::runtime_error &e) {
        PyErr_SetString(PyExc_RuntimeError, e.what());
        return NULL;
    }
}
```

### Python Usage Example

```python
import databasemodule

with databasemodule.DatabaseManager() as db:
    # Use the database
    pass  # Your code here

# The database is automatically closed here
```

habibayassin:
> > > Is a SEGFAULT expected when using the Python interface? I would have thought not.
> > 
> > 
> > The database is destroyed but pointer is kept to it in python and subsequent APIs calls are made. The fact that it is python doesn't remove the need to not use dead pointers. It would be better to null out the pointer in python after destroy so that you get a python error instead:
> > self.base_db = None
> 
> Makes sense. So both OpenROAD should be fixed(no SEGFAULT) and the Python code using the OpenROAD API.
> 
> I dont have access to failing test case currently.

to recreate the failing test case (no logger), navigate to OpenROAD-flow-scripts repo and execute
`./tools/OpenROAD/etc/test_delta_debug.sh`

to recreate the failing test case (SEGFAULT), change the instantiation of the database in  OpenROAD-flow-scripts path tools/OpenROAD/etc/deltaDebug.py line 206 to
`self.base_db = Design.createDetachedDb()`
and add this import at the top
`from openroad import Design` 
then in OpenROAD-flow-scripts main directory execute
`./tools/OpenROAD/etc/test_delta_debug.sh`

oharboe:
@habibayassin FYI https://github.com/The-OpenROAD-Project/OpenROAD/pull/4853

maliberty:
The above idiom doesn't prevent errors.  You can get a pointer to a db object and still try to access it outside of the scope.

PR #4806, opened 2024-03-17
gui: ensure setLoadingState is called from layout viewer thread

gadfort:
This needs to be validated for correct behavior


maliberty:
@oharboe this is attempting to address your heatmap timer error.  Please give it a try.

maliberty (approved):

PR #4810, opened 2024-03-18
floorplan: output one line per lef to make output less noisy and easier to read

oharboe:
still logs the file in which an error occurred, but only if there was an error


oharboe:
> I'm not sure these should be demoted, they don't add up to much additional text, but provide useful feedback about the technology getting loaded. you can always use `suppress_message` if you want to get rid of this in your runs

The reports should be a match by default for most anyone.

I have never used or read this information.

I think it belongs in a report, if it cant be trivially seen in OpenROAD GUI, not the log.

maliberty:
I'm ok but would like agreement from the community that these messages aren't useful (@gadfort @QuantamHD @rovinski @donn or anyone else).

oharboe:
Context: OpenROAD is substantially less "chatty" than it used to be. A lot of debug output from early implementations of e.g. CTS and mpl2 was converted to debugPrints.

This PR is in that same vein: if I never read or used output, then for my part it should not be logged by default, but of course my use case is only one use case.

gadfort:
mpl2 was extremely noisy, this represents 3 lines (btw, the same can be said about the defin and macro lefin).
When setting up new technologies it's somewhat helpful and I don't think any user should be asked to look at the debugging messages if they run into issues (those should be limited to developers).

oharboe:
> mpl2 was extremely noisy, this represents 3 lines (btw, the same can be said about the defin and macro lefin). When setting up new technologies it's somewhat helpful and I don't think any user should be asked to look at the debugging messages if they run into issues (those should be limited to developers).

Is this information available nowhere else?

oharboe:
> mpl2 was extremely noisy, this represents 3 lines (btw, the same can be said about the defin and macro lefin).

This PR doesn't address other cases of log spam, one step at the time...

maliberty:
> Is this information available nowhere else?

The final db doesn't know from what files it was constructed.  We could try to capture that though it would get complicated for verilog which can have multiple files, defines, etc.

oharboe:
> sn't know from what files it was constructed. We could try to capture that though it would get complicated for verilog which can have multiple files, defines, etc.

Is there a way to log this information to a report instead of spamming the output log?

gadfort:
@oharboe how about adding a flag `-verbose` to read_lef and read_def which will allow the users to decide if they want to see this or not? This would be consistent with how other commands work.

donn:
We're ambivalent. On the one hand, you're right, OpenROAD is a lot more quiet than it used to be and these are too chatty.

On the other hand, it doesn't happen that frequently within a flow and OpenLane isn't exactly the epitome of being quiet. So whatever you guys agree on we're okay with.

maliberty:
We could move the reporting of  layers, via, and library cells counts to a user command as those things are saved in the db.  At that point this PR would be reasonable for compressing the reading/finished messages.  @gadfort opinion?

maliberty:
Unless you really need to know from which LEF they came.

gadfort:
@maliberty that would also be okay (knowing the file is helpful, and the command would need to come with (along side) this PR so we don't loose any debugging capability), but the lefin (for macros) can be very helpful to know that a cell has a lot of shapes in it (this is a common reason PDN slows down or uses a lot of memory) and if the readin-log is lost it will be just that much harder to fix.
If everyone else wants to get rid of them, then I would say don't bother making `debugPrint` statements since it's not going to be documented anyways we'll just forget them anyways.
If all the warnings reference the file they came from then there is no lost information for those.

maliberty:
If the file scope if useful then -verbose is a better solution.

oharboe:
> If the file scope if useful then -verbose is a better solution.

Is `-verbvose` an existing feature? If so, how does it work?

oharboe:
> It would be a new command option

To the openroad binary?

maliberty:
To read_lef

oharboe:
> To read_lef

Hmmm....Why not a global cmd line option for openroad to add verbosity? no need to hack scripts then. lots of tools have this.

maliberty:
> Hmmm....Why not a global cmd line option for openroad to add verbosity? no need to hack scripts then. lots of tools have this.

That would be a very large project to evaluate every message for verbosity.  All your prior changes to debugPrint would need to be revisited as well.

rovinski:
* I generally don't find that these lines add clutter. Sometimes it's a nice reminder if you read in the wrong library by spotting the wrong number of cells, etc.
* I do find that it is information that I like to see as a designer
* I also wouldn't care greatly if it went away
* Unless I'm mistaken, I think commercial tools usually print similar information
* I agree with @gadfort that if this becomes a debugPrint, it will be completely forgotten about.
* I would be fine with moving this info to a -verbose flag. Seems like a good compromise.

oharboe:
I don't like options, they make applications more complicated and harder to test. I have another idea I will try: rework the output to include what there is today, but put it all on one line to make it more compact and to output context only in the case of an exception being thrown.

oharboe:
New output:

- no options
- one line per lef file. This doubles as progress output
- full path for file is only printed in the event that an exception was thrown
- single line with file name only (no path) to .lef file + all the other information that was printed

```
[INFO ODB-0227] LEF file: NangateOpenCellLibrary.tech.lef, created 22 layers, 27 vias, 0 library cells
[INFO ODB-0227] LEF file: NangateOpenCellLibrary.macro.mod.lef, created 0 layers, 0 vias, 135 library cells
```

gadfort:
@oharboe I think you should keep the same behavior for the one line so if no vias/layers/cells are created, don't add it to the list.

oharboe:
> @oharboe I think you should keep the same behavior for the one line so if no vias/layers/cells are created, don't add it to the list.

Fixed.

src/odb/src/lefin/lefin.cpp
```diff
@@ -2221,17 +2230,40 @@ bool lefin::readLef(const char* lef_file)
...
+
+  if (_layer_cnt > 0) {
+    std::ostringstream ss;
+    ss << _layer_cnt << " layers";
+    parts.push_back(ss.str());
+  }
+
+  if (_via_cnt > 0) {
+    std::ostringstream ss;
+    ss << _via_cnt << " vias";
```
maliberty, line 2245: Just FYI fmt::format is a nice replacement for this pattern.

maliberty (approved):

src/odb/src/lefin/lefin.cpp
oharboe: Is CentOS 7 phased out?

maliberty: Yes but how does that relate?

oharboe: I dumbed down this PR so it would work on centos

maliberty: We have been using it on centos7, eg

```
dbname = fmt::format("{}.main.{}.db", filename, getpid());
````

PR #4811, opened 2024-03-18
Assume port direction is input if io_type is unspecified in LEF.

mikesinouye:
in an offline chat with @maliberty, he suggests that even with an unknown type we should default to input.


maliberty:
I am not in favor of this change as I think this is fixing a symptom rather than the root cause.  The four legal values of dbIoType are handled in the current code.  Somehow you are getting here with an invalid type.  That is a critical problem and should be treated as such.  How such an invalid value got into odb should be figured out and resolved at the source.

Issue #4821, opened 2024-03-20
Fix problem or remove unactionable warning 


oharboe:
### Describe the bug

Fix the cause of the following errors warnings or remove them if they are bogus or inactionable:

- [ ] `Error: duplicated via definition for via1_2_70092_18_1_1947_36_36`
- [ ] `[WARNING DRT-0240] CUT layer V3 does not have square single-cut via, cut layer width may be set incorrectly. `

To reproduce:

The file is big, but running the test until you get the warnings only takes a minute:

untar https://drive.google.com/file/d/1nHLO6_kBF4hhUBYYEDDPbirIYIGbMu7b/view?usp=sharing

```
$ ./run-me-BoomTile-asap7-base.sh
OpenROAD v2.0-12661-g64b9d673c
[deleted]
[INFO DRT-0150] Reading design.
Error: duplicated via definition for via1_2_70092_18_1_1947_36_36
Error: duplicated via definition for via5_6_120_288_1_2_58_322
[deleted]
[WARNING DRT-0240] CUT layer V3 does not have square single-cut via, cut layer width may be set incorrectly.
[WARNING DRT-0240] CUT layer V5 does not have square single-cut via, cut layer width may be set incorrectly.
[deleted]
```

### Expected Behavior

Only actionable real warnings

### Environment

```shell
OpenROAD v2.0-12661-g64b9d673c
```

### To Reproduce

See above

osamahammad21:
>  Error: duplicated via definition for via1_2_70092_18_1_1947_36_36

This via appears in both the tech lef and the def. I don't see it in the original asap7 tech lef but it appears when I write lef from 5_2_fillcell.odb. An earlier step in the flow did something wrong by adding it to the tech. I am thinking of making this an actual error. However it seems more fit to catch it in odb not drt so that it could be handled as early as possible.

> [WARNING DRT-0240] CUT layer V3 does not have square single-cut via, cut layer width may be set incorrectly.

This should be handled in the specified commit in this issue. I am removing this warning as well as the dependency on the cut layer width.

maliberty:
@oharboe can you find which step produces that via in the tech lef and make a test case for it.

osamahammad21:
@maliberty @oharboe It happens at PDN in hierarchical designs. PDN creates its via definitions for the sub-modules (block vias). Then when the block is hardened into a macro, the pdn via moves to the tech of the top module. PDN then creates the same via definition for the top module.
[...]

PR #4822, opened 2024-03-20
detailed route: reduce frequency of progress update

oharboe:
computers are faster than they used to be, some of these updates spammed the log several times a second.

10x rarer still gives progress updates several times per minute plenty fast for detailed route.

This is helpful for larger designs, such as the one in this github issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/4821

### DIFF
--- a/src/drt/src/frRegionQuery.cpp
+++ b/src/drt/src/frRegionQuery.cpp
@@ -731,12 +731,12 @@ void frRegionQuery::Impl::init()
     }
     cnt++;
     if (VERBOSE > 0) {
-      if (cnt < 100000) {
-        if (cnt % 10000 == 0) {
+      if (cnt < 1000000) {
+        if (cnt % 100000 == 0) {
           logger_->info(DRT, 18, "  Complete {} insts.", cnt);
         }
       } else {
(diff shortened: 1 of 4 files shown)


src/drt/src/frRegionQuery.cpp
```diff
@@ -731,12 +731,12 @@ void frRegionQuery::Impl::init()
     }
     cnt++;
     if (VERBOSE > 0) {
-      if (cnt < 100000) {
-        if (cnt % 10000 == 0) {
+      if (cnt < 1000000) {
```
gadfort, line 734: could this be a `static constexpr int some_name_ = 100000;` in a header (same for all the rest)? might be cleaner to read and easier to update later if we need to.

maliberty: Since we know how much work we are going to do it is easy to convert this to 10% , 20%, etc which is more scale invariant.

maliberty (approved):

src/drt/src/frRegionQuery.cpp
oharboe: Good future improvement.

PR #4839, opened 2024-03-21
mpl2: ensure io blockages are created only when there are io pins in the boundary

### LINKED ISSUE #4836: Mpl2 generates macro blockage for pin access when there are no io pins
### Describe the bug

When there are io pins in between the limits of the core and the boundaries of the die a macro blockage for pin access is being created for a boundary that has no io pins at all.

### Expected Behavior

Mpl2 should only create macro blockages for io pins in the boundaries that actually have io pins.

### Environment

```shell
HEAD of Master
```

### To Reproduce

Run mpl2 flow in a design that has IO pins in the region between the die and the core.

### Relevant log output

_No response_

### Screenshots

asap7/swerv_wrapper: upper left corner (this design has no IO pins in the left boundary)

image

### Additional Context

_No response_

AcKoucher:
Fix #4836 

We were not taking into account the case where we might have IO pins in the empty region in between the core and the die.

This should fix CI. (Failing in asap7/swerv_wrapper)

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -3044,52 +3044,52 @@ void HierRTLMP::setIOClustersBlockages()
 {
   debugPrint(logger_,
              MPL,
-             "hierarchical_macro_placement",
+             "coarse_shaping",
              1,
              "Creating the pin blockage for root cluster");
-  floorplan_lx_ = root_cluster_->getX();
-  floorplan_ly_ = root_cluster_->getY();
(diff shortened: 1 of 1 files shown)


maliberty (approved):

maliberty:
Pending on secure CI

maliberty:
Don't worry about clust02

AcKoucher:
@maliberty 
Secure-CI showed the Calibre failures for gf12: bp_single, ca53_cpu, coyote and intel16: gcd.

These are on master  as well so it's good to merge.

PR #4842, opened 2024-03-23
correct realvalued heatmap handling

gadfort:
Issue:
* when converting to a `%`, the previous code was applying the scaling factor to the the percent value, which was be incorrect.

Changes:
* avoid applying scaling factor to percent and delay setting min and max with the scaling factor until after processing map
* provide access to logger (no functional change, but helpful when attempting to debug)
* use `lowest` vs. `min` for finding max values.

### DIFF
--- a/src/gui/src/heatMap.cpp
+++ b/src/gui/src/heatMap.cpp
@@ -850,14 +850,15 @@ void RealValueHeatMapDataSource::correctMapScale(HeatMapDataSource::Map& map)
 {
   determineMinMax(map);
   determineUnits();
-  min_ = roundData(min_);
-  max_ = roundData(max_);
 
   for (const auto& map_col : map) {
     for (const auto& map_pt : map_col) {
       map_pt->value = convertValueToPercent(map_pt->value);
     }
   }
+
+  min_ = roundData(min_ * scale_);
+  max_ = roundData(max_ * scale_);
 
   // reset since all data has been scaled by the appropriate amount
   scale_ = 1.0;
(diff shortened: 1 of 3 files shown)


src/gui/src/heatMap.cpp
```diff
@@ -850,31 +850,31 @@ void RealValueHeatMapDataSource::correctMapScale(HeatMapDataSource::Map& map)
...
   scale_ = 1.0;
 }
 
 double RealValueHeatMapDataSource::roundData(double value) const
 {
   const double precision = 1000.0;
-  double new_value = value * scale_;
-  return std::round(new_value * precision) / precision;
+  return std::round(value * precision) / precision;
 }
```
maliberty, line 871: Does this change the precision now that you are not multiplying by scale_?

src/gui/src/heatMap.cpp
```diff
@@ -850,31 +850,31 @@ void RealValueHeatMapDataSource::correctMapScale(HeatMapDataSource::Map& map)
...
 
 void RealValueHeatMapDataSource::determineMinMax(
     const HeatMapDataSource::Map& map)
 {
   min_ = std::numeric_limits<double>::max();
-  max_ = std::numeric_limits<double>::min();
+  max_ = std::numeric_limits<double>::lowest();
```
maliberty, line 877: This is such an unfortunate choice in the c++ standard - I've seen this bug before.

src/gui/src/heatMap.cpp
gadfort: I moved the min and max values to be scaled after all the values are processed. So the scaling should not be needed since all the values have the same scale. (Doing the scaling across two functions was also somewhat confusing when debugging).

maliberty: Sure but does it affect the final rounded value?  If scale_ != 1.0 then this will result in a different return value.

gadfort: it should not, there were only two places where the `scale_` was used (for the min and max rounding) and I just make sure it's multiplied in the argument in the function call, effectively should be the same answer.

maliberty: My question is about:
std::round(value * precision * 1000.0) != std::round(value * precision) * 1000.0

gadfort: The way it was rewritten (unless I'm missing something), it should be as follows since the value is scaled by `scale_` when calling `roundData` instead of inside of `roundData`:
```
std::round(value * precision * 1000.0) == std::round((value * 1000.0) * precision)
```

maliberty (approved):

PR #4844, opened 2024-03-23
pdn: remove non-fixed bpins during power grid generation

gadfort:
Changes:
- removes any bpins that ppl might have added which are not marked as LOCKED+

Issue:
- Since pdngen doesn't attempt to connect to them they will be left floating causing connectivity errors.


maliberty (approved):

maliberty:
I'm assuming this only applies to power/ground nets.

gadfort:
> I'm assuming this only applies to power/ground nets.

Yes, only the nets that PDNgen is handling

PR #4846, opened 2024-03-24
ubi9 dep

habibayassin:


etc/DependencyInstaller.sh
```diff
@@ -342,10 +339,86 @@ _installRHELPackages() {
...
-        https://forensics.cert.org/centos/cert/7/x86_64/flex-2.6.1-9.el7.x86_64.rpm
+        https://rpmfind.net/linux/centos-stream/9-stream/AppStream/x86_64/os/Packages/tcl-devel-8.6.10-7.el9.x86_64.rpm \
+        https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/bison-3.7.4-5.el9.x86_64.rpm \
+        https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/flex-2.6.4-9.el9.x86_64.rpm
 }
 
+# _installRHELPackages() {
+#     yum -y update
+#     if [[ $1 == 9* ]]; then
+#         if [[ $(yum repolist | egrep -c "rhel-8-for-x86_64-appstream-rpms") -eq 0 ]]; then
```
vvbandeira, line 350: Remove

vvbandeira (changes requested):

etc/DependencyInstaller.sh
vvbandeira: Also, there's a merge conflict now.

vvbandeira (changes requested):
We need a version check as we do for Ubuntu to ensure we do not try to install the packages on an unsupported version.

etc/DependencyInstaller.sh
```diff
@@ -715,14 +713,20 @@ EOF
...
+                _checkIsLocal
+                _installRHELPackages
+                _installRHELCleanUp
+            fi
+            if [[ "${option}" == "common" || "${option}" == "all" ]]; then
+                _installCommonDev
+                _installOrTools "centos" "8" "amd64"
+            fi
+        else
+            echo "ERROR: RHEL 9 is only supported" >&2
```
rovinski, line 728:
```suggestion
            echo "[ERROR] Only version 9 of RHEL supported" >&2
```

Issue #2590, opened 2022-12-08
add fit to cts viewer


maliberty:
### Description

There should be a way to do a fit to window.

### Suggested Solution

Hotkey: f 
Button

PR #4849, opened 2024-03-24
Add fit to cts viewer

jlowpc:
Attempt to add fit to cts viewer Issue #2590.  Attached a simple ibex test case (with 2 clock tabs) to test the hotkey 'f'.
[test_cts_viewer_fit.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/14736293/test_cts_viewer_fit.tar.gz)


src/gui/src/mainWindow.cpp
```diff
@@ -588,6 +588,7 @@ void MainWindow::createActions()
   connect(exit_, &QAction::triggered, this, &MainWindow::exit);
   connect(this, &MainWindow::exit, viewers_, &LayoutTabs::exit);
   connect(fit_, &QAction::triggered, viewers_, &LayoutTabs::fit);
+  connect(fit_, &QAction::triggered, clock_viewer_, &ClockWidget::fit);
```
maliberty, line 591: Thanks for looking into this.

Fit in the main window and fit in the clock viewer should be separate.  You might well want to fit one without fitting the other.   You should create an independent action for the ClockWidget.

jlowpc: When you say independent action for the ClockWidget, do you want a different hot key for the fit for clock viewer?  If so, what key should I use?  If not, please let me know how do we decide when user press 'f', whether it goes to the layout window or the clock tree viewer?  Thanks.

AcKoucher:
How about also having a fit button in the clock widget?

Instead of having two hotkeys we couldn't we fit based on the cursor position? (Only fit the clock tree if cursor is inside this widget)

maliberty:
I was thinking the fit key would be window dependent as well.  There is already a fit on the right click menu so I'm not sure another button is needed.

jlowpc:
@maliberty please confirm that you want the 'f' key to ONLY work on the CTS window when the mouse cursor is within the CTS window, and you want the 'f' key to ONLY work on the layout when the mouse cursor is over the layout window.  The menu and button for fit will work ONLY for the layout regardless of where the mouse cursor is.

maliberty:
My thought was it would depend on the mouse location or focus but perhaps this is getting too tricky both for implementation and somewhat unintuitive as a UI.  A fit button next to the update button might be a simpler solution.

jlowpc:
Please review and comment on my attempt #2 to add a fit button instead of using 'f' hotkey for CTS viewer.  Please use attached smaller and improved test case for testing [source run.tcl].  
[test_cts_fit2.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/14877506/test_cts_fit2.tar.gz)

maliberty:
Thanks this look good.  Please run clang-format  on the code and fix the clang-tidy warning above.

maliberty (approved):

maliberty:
Thanks!

Issue #4402, opened 2023-12-18
PSM power grid check lists false-positive disconnected nodes inside macros


donn:
### Describe the bug

Trying a smoke-test multi-macro design with core rings: `check_power_grid -net VPWR` reports a multitude of disconnected PDN nets. Key is, the nets are not in fact disconnected there are vias at the claimed location: see this example

![Screenshot 2023-12-18 at 17 15 43](https://github.com/The-OpenROAD-Project/OpenROAD/assets/12652988/763bacc1-8f29-45b5-aeb6-5ef6023e098d)

When disabling the check, the design does indeed pass LVS.

### Expected Behavior

A design passing LVS should not fail the PSM power grid check.

### Environment

```shell
Git commit: 75f2f325b7a42e56a92404f33af8e96530d9b202
kernel: Darwin 23.1.0
os: macOS 14.1.1
cmake version 3.27.7
CMake Deprecation Warning at third-party/abc/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Warning:
  Manually-specified variables were not used by the project:

    VERBOSE


-- The CXX compiler identification is Clang 16.0.6
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /nix/store/zh206nri10wwp68g7qf9ccya1irynx5g-clang-wrapper-16.0.6/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-11533-g75f2f325b
-- System name: Darwin
-- Compiler: Clang 16.0.6
-- Build type: RELEASE
-- Install prefix: /var/empty/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is Clang 16.0.6
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /nix/store/zh206nri10wwp68g7qf9ccya1irynx5g-clang-wrapper-16.0.6/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
[...]

Issue #3599, opened 2023-07-09
Reporting ultimately takes longer than detailed routing for mock-array


oharboe:
### Description

Using https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1212

Create:

```
export DESIGN_CONFIG?=designs/asap7/mock-array/config.mk
export MOCK_ARRAY_DATAWIDTH?=8
export MOCK_ARRAY_TABLE?=8 8 4 4 5 5
export MOCK_ARRAY_SCALE?=640
```

Run `make verilog` and then `make`.

These are the running times reported for a 12 thread machine:

Still waiting for report after ca. 1 hour...

```
Log                       Elapsed seconds
1_1_yosys                          1
2_1_floorplan                      1
2_2_floorplan_io                   1
2_3_tdms_place                     1
2_5_tapcell                        1
2_6_pdn                          222
3_1_place_gp_skip_io              11
3_2_place_iop                      3
3_3_place_gp                      42
3_4_resizer                       16
3_5_opendp                        23
4_1_cts                           27
4_2_cts_fillcell                  25
5_1_fastroute                     42
5_2_TritonRoute                 1065
6_1_merge                         47
6_report                        5665
```

Tail of log:

```
[WARNING PSM-0030] VSRC location at (131.200um, 971.200um) and size 10.000um, is not located on an existing power stripe node. Moving to closest node at (129.882um, 965.404um).
[WARNING PSM-0030] VSRC location at (971.200um, 971.200um) and size 10.000um, is not located on an existing power stripe node. Moving to closest node at (967.818um, 965.404um).
[INFO PSM-0031] Number of PDN nodes on net VSS = 13251525.
[no further output after 1 hour... then some output before reporting completes relatively quickly.]
```

### Suggested Solution

Find and fix some low-hanging fruit in the scaling of reporting stage.

maliberty:
There are a number of different steps. Can you narrow it  down?

oharboe:
> There are a number of different steps. Can you narrow it down?

The log pause is not indicative enough? Anyone working on this will have to run this locally anyway at which point they can drill down...

oharboe:
Some quick debugger suspend/resume profiling.

This is the step that takes a long time.

```
[...]

PR #4850, opened 2024-03-25
psm: rework solver

gadfort:
Changes:
- reworks the IR solver to speed it up on larger designs
- adds caching of networks to speed up analysis across corners
- add alternative voltage source types (bumps, straps, and full for now)
- heatmap can select different nets/corners/layers

Breaking changes:
- The previous implementation would "snap" instances to the power grid, this does not and therefore `check_power_grid` will return an error if the places instances are not detailed placed.

Metrics:
- ariane136 (previous implementation takes about ~360s, this PR takes ~50s.
- on small designs the speedup will not be noticeable or possible a slight increase, but for larger designs where the previous one didn't complete within an hour, the current one can finish in about ~300s (2x2mm^2 with 13 metal layers).

Closes: #4402
Closes: #3599


src/utl/src/timer.cpp
```diff
@@ -55,4 +53,25 @@ std::ostream& operator<<(std::ostream& os, const Timer& t)
...
+      msg_(msg),
+      tool_(tool),
+      group_(group),
+      level_(level)
+{
+}
+
+DebugScopedTimer::~DebugScopedTimer()
+{
+  debugPrint(logger_, tool_, group_.c_str(), level_, msg_, *this);
```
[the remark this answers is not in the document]
gadfort: cannot convert ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’} to ‘const char*’

oharboe:
Crashes. To reproduce:

untar [final-report-crash.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/14753537/final-report-crash.tar.gz)

```
OpenROAD v2.0-12730-gdf94f5a0f 
[WARNING STA-0450] virtual clock clock_vir can not be propagated.
[INFO] Deleted 0 routing obstructions
[INFO RCX-0431] Defined process_corner X with ext_model_index 0
[INFO RCX-0029] Defined extraction corner X
[INFO RCX-0008] extracting parasitics of MockArray ...
[INFO RCX-0435] Reading extraction model file /home/oyvind/OpenROAD-flow-scripts/flow/foo/platforms/asap7/rcx_patterns.rules ...
[INFO RCX-0436] RC segment generation MockArray (max_merge_res 50.0) ...
[INFO RCX-0040] Final 44262 rc segments
[INFO RCX-0439] Coupling Cap extraction MockArray ...
[INFO RCX-0440] Coupling threshhold is 0.1000 fF, coupling capacitance less than 0.1000 fF will be grounded.
[INFO RCX-0043] 105163 wires to be extracted
[INFO RCX-0442] 10% completion -- 10576 wires have been extracted
[INFO RCX-0442] 16% completion -- 17117 wires have been extracted
[INFO RCX-0442] 23% completion -- 24387 wires have been extracted
[INFO RCX-0442] 32% completion -- 34649 wires have been extracted
[INFO RCX-0442] 39% completion -- 41440 wires have been extracted
[INFO RCX-0442] 46% completion -- 48508 wires have been extracted
[INFO RCX-0442] 56% completion -- 58917 wires have been extracted
[INFO RCX-0442] 71% completion -- 75498 wires have been extracted
[INFO RCX-0442] 76% completion -- 80875 wires have been extracted
[INFO RCX-0442] 82% completion -- 87030 wires have been extracted
[INFO RCX-0442] 88% completion -- 92837 wires have been extracted
[INFO RCX-0442] 92% completion -- 97540 wires have been extracted
[INFO RCX-0442] 100% completion -- 105163 wires have been extracted
[INFO RCX-0045] Extract 23324 nets, 59969 rsegs, 59969 caps, 49108 ccs
[INFO RCX-0015] Finished extracting MockArray.
[INFO RCX-0016] Writing SPEF ...
[INFO RCX-0443] 23324 nets finished
[INFO RCX-0017] Finished writing SPEF ...
Signal 11 received
Stack trace:
 0# 0x000055E7DD1C2483 in openroad
 1# 0x00007BFABCE42990 in /lib/x86_64-linux-gnu/libc.so.6
 2# odb::dbTechLayer::getRoutingLevel() in openroad
 3# psm::IRNetwork::generatePolygonsFromITerms(std::vector<psm::TerminalNode*, std::allocator<psm::TerminalNode*> >&) in openroad
 4# psm::IRNetwork::generateRoutingLayerShapesAndNodes() in openroad
 5# psm::IRNetwork::construct() in openroad
 6# psm::IRNetwork::IRNetwork(odb::dbNet*, utl::Logger*, bool) in openroad
 7# psm::IRSolver::IRSolver(odb::dbNet*, bool, sta::dbSta*, rsz::Resizer*, utl::Logger*, std::map<odb::dbNet*, std::map<sta::Corner*, float, std::less<sta::Corner*>, std::allocator<std::pair<sta::Corner* const, float> > >, std::less<odb::dbNet*>, std::allocator<std::pair<odb::dbNet* const, std::map<sta::Corner*, float, std::less<sta::Corner*>, std::allocator<std::pair<sta::Corner* const, float> > > > > > const&, psm::PDNSim::GeneratedSourceSettings const&) in openroad
 8# psm::PDNSim::getIRSolver(odb::dbNet*, bool) in openroad
 9# psm::PDNSim::checkConnectivity(odb::dbNet*, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) in openroad
10# psm::PDNSim::analyzePowerGrid(odb::dbNet*, sta::Corner*, psm::GeneratedSourceType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) in openroad
11# analyze_power_grid_cmd(odb::dbNet*, sta::Corner*, psm::GeneratedSourceType, char const*, bool, char const*, char const*, char const*) in openroad
12# 0x000055E7DDE75067 in openroad
13# TclNRRunCallbacks in /lib/x86_64-linux-gnu/libtcl8.6.so
14# 0x00007BFAC17A1C43 in /lib/x86_64-linux-gnu/libtcl8.6.so
15# Tcl_EvalEx in /lib/x86_64-linux-gnu/libtcl8.6.so
16# Tcl_Eval in /lib/x86_64-linux-gnu/libtcl8.6.so
17# sta::sourceTclFile(char const*, bool, bool, Tcl_Interp*) in openroad
18# ord::tclAppInit(Tcl_Interp*) in openroad
19# Tcl_MainEx in /lib/x86_64-linux-gnu/libtcl8.6.so
20# main in openroad
21# 0x00007BFABCE28150 in /lib/x86_64-linux-gnu/libc.so.6
22# __libc_start_main in /lib/x86_64-linux-gnu/libc.so.6
23# _start in openroad
./run-me-mock-array-asap7-base.sh: line 7: 15548 Segmentation fault      (core dumped) openroad -no_init ${SCRIPTS_DIR}/final_report.tcl
```

oharboe:
@gadfort Added PR to add the failing test-case to CI https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1886

src/psm/src/ir_network.cpp
```diff
@@ -0,0 +1,1299 @@
...
+        }
+
+        auto* layer = geom->getTechLayer();
+
+        has_routing_term = true;
+
+        odb::Rect pin_shape = geom->getBox();
+        transform.apply(pin_shape);
+
+        shapes_by_layer[layer] += rectToPolygon(pin_shape);
```
maliberty, line 281: I don't understand why you are adding the geom shapes here when it seems you already did that above (265-267)

src/psm/src/connection.cpp
```diff
@@ -0,0 +1,262 @@
...
+    std::swap(node0_, node1_);
+
+    return;
+  }
+
+  if (node1_ == nullptr) {
+    std::swap(node0_, node1_);
+
+    return;
+  }
```
maliberty, line 57: Is this a copy-paste error?  Why do you want to swap in both cases?

src/psm/src/connection.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+void Connection::print(utl::Logger* logger) const
+{
+  logger->report(describe());
+}
+
+int Connection::getDBUs() const
+{
+  const Node* node = node0_ ? node0_ : node1_;
+  if (node) {
```
maliberty, line 101: Can you have two null nodes?

src/psm/src/connection.cpp
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  if (node1_ != nullptr) {
+    const auto& pt = node1_->getPoint();
+    node1_x = pt.getX();
+    node1_y = pt.getY();
+    node1_l = node1_->getLayer()->getNumber();
+  }
+
+  return {node0_l, node0_x, node0_y, node1_l, node1_x, node1_y};
```
maliberty, line 141: Would (a,b) != (b,a) be correct?

src/psm/src/node.h
```diff
@@ -1,108 +1,167 @@
...
-CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
-OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-*/
+class Node
+{
+ public:
+  enum class NodeType
+  {
+    None,
```
maliberty, line 62: Is `None` needed?  When would it occur?

src/psm/src/node.h
```diff
@@ -1,108 +1,167 @@
...
-using odb::dbInst;
-using odb::Point;
-using NodeIdx = int;
+  std::string getName() const;
+  std::string getTypeName() const;
 
-struct NodeEnclosure
+ protected:
+  virtual NodeType getType() const { return NodeType::None; }
+  double getDBUs() const;
```
maliberty, line 95: Do we expect each Node to have its own DBUs?  This seems rather low for this API.

src/psm/src/ir_network.cpp
```diff
@@ -0,0 +1,1299 @@
...
+
+        // create iterm nodes
+        auto center = std::make_unique<TerminalNode>(pin_shape, layer);
+        terminals.push_back(center.get());
+
+        connections_.push_back(
+            std::make_unique<TermConnection>(base_node.get(), center.get()));
```
maliberty, line 288: If the base_node is at the same location as the center then can you skip making the center node and the connection?  For simple pins this would often occur.

maliberty: Same with generatePolygonsFromBTerms

src/psm/src/ir_network.cpp
```diff
@@ -0,0 +1,1299 @@
...
+  using Rectangle = boost::polygon::rectangle_data<int>;
+
+  auto get_layer_orientation
+      = [](odb::dbTechLayer* layer) -> boost::polygon::orientation_2d {
+    switch (layer->getDirection()) {
+      case odb::dbTechLayerDir::NONE:
+      case odb::dbTechLayerDir::HORIZONTAL:
+        return boost::polygon::orientation_2d(
+            boost::polygon::orientation_2d_enum::VERTICAL);
+        break;
```
maliberty, line 362: `return boost::polygon::orientation_2d_enum::VERTICAL;` suffices - same below

src/psm/src/ir_network.cpp
```diff
@@ -0,0 +1,1299 @@
...
+  shape_poly_set += polygon;
+
+  std::vector<Rectangle> rect_shapes;
+  shape_poly_set.get_rectangles(rect_shapes, get_layer_orientation(layer));
+
+  std::set<odb::Rect> search_rect_shapes;
+  for (const auto& poly_rect : rect_shapes) {
+    search_rect_shapes.emplace(
+        xl(poly_rect), yl(poly_rect), xh(poly_rect), yh(poly_rect));
+  }
```
maliberty, line 381: With `#include "odb/geom_boost.h"` this can be reduced to 

````
  std::vector<odb::Rect> search_rect_shapes;
  shape_poly_set.get_rectangles(search_rect_shapes, get_layer_orientation(layer));
````

[auto for the iterator type below avoids needing to update them]

File: src/psm/src/ir_network.cpp [hunk 11]
```diff
@@ -0,0 +1,1273 @@
...
+  min_node_pitch_.clear();
+
+  const double dbus = getBlock()->getDbUnitsPerMicron();
+  const int min_pitch = dbus * min_node_pitch_um_;
+
+  for (auto* layer : getTech()->getLayers()) {
+    if (layer->getRoutingLevel() != 0) {
+      min_node_pitch_[layer] = std::min(
+          min_pitch,
+          min_node_pitch_multiplier_ * std::max(1, layer->getPitch()));
```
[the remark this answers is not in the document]
gadfort: I couldn't remember if pitch is guaranteed to present and sent to something > 0. If < 1 is not possible, then I can drop this.

src/psm/src/ir_network.h
```diff
@@ -0,0 +1,213 @@
...
+
+  std::vector<std::unique_ptr<Connection>> connections_;
+
+  std::vector<std::unique_ptr<ITermNode>> iterm_nodes_;
+  std::vector<std::unique_ptr<BPinNode>> bpin_nodes_;
+
+  std::map<odb::dbTechLayer*, int> min_node_pitch_;
+
+  static constexpr int min_node_pitch_multiplier_ = 10;
+  static constexpr double min_node_pitch_um_ = 10.0;
```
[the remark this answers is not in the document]
gadfort: "magic numbers", the pitch multiplier is "kind of like" the row multiplier in the previous implementation (defaulted to 5), 10 tracks is much less than 5 row heights, so this seemed reasonable, but I'm not wed to these numbers.
10um is just to ensure the top layer has enough nodes to intersect with the voltage sources (10 happen to the smallest in the test cases), but it's possible that is not needed any longer since the voltages sources were updated to "snap" to the nearest node.

src/psm/src/ir_network.cpp
gadfort: oversight, fixed.

src/psm/src/ir_network.cpp
gadfort: thats awesome (this will appear in a lot of places)

src/psm/src/node.h
gadfort: It's just a regular (non-special) node, I could give it a better name

src/psm/src/connection.cpp
gadfort: should have had an else

src/psm/src/connection.cpp
gadfort: at one point that was possible, that shouldn't be possible, but this is really only supposed to be triggering when debugging

src/psm/src/connection.cpp
gadfort: the `ensureNodeOrder` function should force the order of the nodes, so I don't have to account for `b,a` case.

src/psm/src/node.h
gadfort: it's just a helper function to the layer to get the DBUs when printing debugging information. This is not meant for the public access.

src/psm/src/ir_network.cpp
gadfort: The `base_node` is special (only one is created per ITerm and BPin), to avoid a bunch of logic to prevent merging the iterm and bpin nodes away, they are kept separate from the rest of the nodes, but this requires a few extra connections to track them.

src/psm/README.md
```diff
@@ -5,16 +5,11 @@ an open-source static IR analyzer.
...
--   Report worst IR drop.
--   Report worst current density over all nodes and wire segments in the
-    power distribution network, given a placed and PDN-synthesized design.
--   Check for floating PDN stripes on the power and ground nets.
--   Spice netlist writer for power distribution network wire segments.
-
-| | |
-| - | - |
-| ![Image 1](doc/current_map.jpg) | ![Image 2](doc/IR_map.jpg) |
-<p style="text-align: center;">(Left): Current Map, (Right): IR drop map</p>
```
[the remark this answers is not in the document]
gadfort: that was aspirational, I wanted to recreate the images using the openroad heatmap, but didn't get around to it. I can restore them, or create a new image (except the current map is not supported in openroad right now), if you have a suggestion for the design, I could do that.

src/psm/README.md
```diff
@@ -27,74 +22,74 @@ Features:
...
 | `-enable_em` | Report current per power grid segment. |
-| `-outfile` | Write per-instance voltage into the file. |
 | `-em_outfile` | Write the per-segment current values into a file. This option is only available if used in combination with `-enable_em`. |
-| `-voltage` | Sets the voltage on a specific net. If this option is not set, the Liberty file's voltage value is obtained from operating conditions. |
-| `-node_density` | Node density (in microns) on the standard cell rails. It cannot be used together with `-node_density_factor`. |
-| `-node_density_factor` | Factor which is multiplied by standard cell height to determine the node density on the std cell rails. It cannot be used together with `-node_density`. The default value is `5`, and the allowed values are integers `[0, MAX_INT]`. |
-| `-corner` | Corner to use for analysis. | 
+| `-voltage_file` | Write per-instance voltage into the file. |
+| `-source_type` | Indicate the type of voltage source grid to model. FULL uses all the nodes on the top layer as voltage sources, BUMPS will model a bump grid array, and STRAPS will model power straps on the layer above the top layer. |
```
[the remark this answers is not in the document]
gadfort: I have a feeling pictures will make this easier to follow.

src/psm/README.md
```diff
@@ -27,74 +22,74 @@ Features:
...
-    [-dx bump_pitch_x]
-    [-dy bump_pitch_y]
-    [-node_density val_node_density]
-    [-node_density_factor val_node_density_factor]
+    -net net_name
     [-corner corner]
+    [-error_file error_file]
+    [-voltage_file voltage_file]
+    [-enable_em]
+    [-em_file em_file]
```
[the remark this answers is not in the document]
gadfort: fixed

src/psm/src/ir_network.cpp
gadfort: When I tried I get:
```
/usr/local/include/boost/polygon/polygon_90_set_data.hpp: In instantiation of ‘void boost::polygon::polygon_90_set_data<T>::get_rectangles(output_container&, boost::polygon::orientation_2d) const [with output_container = std::set<odb::Rect>; T = int]’:
..../OpenROAD/src/psm/src/ir_network.cpp:371:32:   required from here
/usr/local/include/boost/polygon/polygon_90_set_data.hpp:193:38: error: no matching function for call to ‘transform(const odb::Rect&, boost::polygon::axis_transformation)’
  193 |           ::boost::polygon::transform(*itr, axis_transformation(axis_transformation::SWAP_XY));`
```

src/psm/src/connection.cpp
maliberty: I'm still not clear what invariant you are establishing.  Are you trying to ensure one of these is non-null?  The code doesn't do that.

src/psm/src/ir_network.cpp
maliberty: I just made the change and pushed it to your branch.

src/psm/src/connection.cpp
```diff
@@ -0,0 +1,260 @@
...
+{
+  if (node0_ == nullptr) {
+    std::swap(node0_, node1_);
+
+    return;
+  } else if (node1_ == nullptr) {
+    std::swap(node0_, node1_);
+
+    return;
+  }
```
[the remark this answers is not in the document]
maliberty: The else wasn't needed ....

src/psm/test/sky130hd_data/zerosoc_pads.def
maliberty, line 1: Why did this need to change?

src/psm/test/gcd_no_vsrc.tcl
```diff
@@ -4,5 +4,6 @@ read_lef Nangate45/Nangate45.lef
 read_def Nangate45_data/gcd.def
 read_liberty Nangate45/Nangate45_typ.lib
 read_sdc Nangate45_data/gcd.sdc
-set_pdnsim_net_voltage -net VDD -voltage 1.1
+set_pdnsim_net_voltage -net VDD -voltage 1.5
```
maliberty, line 7: Why the change?

src/psm/test/corners.ok
```diff
@@ -4,29 +4,25 @@
...
-Corner: max
-Worstcase voltage: 6.25e-01 V
-Average IR drop  : 4.30e-01 V
-Worstcase IR drop: 4.75e-01 V
+Corner           : max
+Supply voltage   : 9.50e-01 V
+Worstcase voltage: 9.30e-01 V
+Average voltage  : 9.41e-01 V
+Average IR drop  : 8.54e-03 V
+Worstcase IR drop: 2.02e-02 V
```
maliberty, line 26: That's a fairly large change in results - what do you attribute it to?

src/psm/test/aes_test_multiple_bterms.ok
```diff
@@ -4,25 +4,23 @@
...
-Corner: default
-Worstcase voltage: 6.75e-03 V
-Average IR drop  : 1.14e-04 V
-Worstcase IR drop: 6.75e-03 V
+Corner           : default
+Supply voltage   : 0.00e+00 V
+Worstcase voltage: 4.25e-03 V
+Average voltage  : 3.49e-04 V
+Average IR drop  : 3.49e-04 V
+Worstcase IR drop: 4.25e-03 V
```
maliberty, line 24: Why are the voltage values equal to the IR drop values?  It would be helpful to print the net name in the report.

src/psm/src/connection.cpp
gadfort: I think some of this is left over, you are correct, the code doesn't do what it should. I'll take a look to see how safe it is to remove.

src/psm/test/sky130hd_data/zerosoc_pads.def
gadfort: turns out all the pads in the file were connected to ground (which would be wrong), this change just removes it so I was able to test on both VDD and VSS.

src/psm/test/gcd_no_vsrc.tcl
gadfort: because the default value is 1.1V, so setting it to 1.1 doesn't actually prove that this is working. I just picked a different number so I could see the value change from the default.

src/psm/test/aes_test_multiple_bterms.ok
gadfort: I think adding the net name makes sense, I'll add it.

src/psm/src/shape.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+  const IRNetwork::NodeTree tree = getNodeTree(getNodes(layer_nodes));
+
+  while (shape_.overlaps(start)) {
+    const IRNetwork::Box box(
+        IRNetwork::Point(start.x() - radius, start.y() - radius),
+        IRNetwork::Point(start.x() + radius, start.y() + radius));
+
+    if (tree.qbegin(boost::geometry::index::intersects(box)) == tree.qend()) {
+      new_nodes.push_back(std::make_unique<Node>(start, layer_));
```
maliberty, line 137: This has the possibility to have nodes more than max_distance apart.  For example image a wire from [0,100] with a max_distance of 10.  If you have existing nodes at 6 (in [5,15]) and 24 (in [15,25]) you won't add a node but they are 18 apart.

src/psm/test/corners.ok
gadfort: There is one bug in there, which I can fix. There is another source of differences which relate to how the old and new implementations handle parallel vias (the old one just made it 1 via, while mine will merge the vias and reduce the resistance of that connection).

src/psm/src/shape.cpp
gadfort: removed this logic

src/psm/test/corners.ok
gadfort: If I triple the resistance of the vias (the emulate only using a single via instead instead of 3 parallel vias) I get:
```
########## IR report #################
Net              : VDD
Corner           : min
Supply voltage   : 1.10e+00 V
Worstcase voltage: 5.37e-01 V
Average voltage  : 5.76e-01 V
Average IR drop  : 5.24e-01 V
Worstcase IR drop: 5.63e-01 V
Percentage drop  : 51.19 %
######################################
```

Compared with the solution on master:
```
########## IR report #################
Corner: min
Worstcase voltage: 5.76e-01 V
Average IR drop  : 4.75e-01 V
Worstcase IR drop: 5.24e-01 V
######################################
```

This is much closer to the original solution.

src/psm/README.md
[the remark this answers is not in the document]
gadfort: added back a picture of picorv32 in sky130hd

maliberty:
Started a secure CI (not expecting any metrics failures, more just a broader test)

src/psm/src/ir_solver.h
```diff
@@ -48,207 +57,190 @@ class Resizer;
...
+  bool hasSolution(sta::Corner* corner) const;
+  Voltage getNetVoltage(sta::Corner* corner) const;
+  std::optional<Voltage> getVoltage(sta::Corner* corner, Node* node) const;
+
+  std::pair<bool, Voltage> getSDCVoltage(sta::Corner* corner,
+                                         odb::dbNet* net) const;
+  std::pair<bool, Voltage> getPVTVoltage(sta::Corner* corner) const;
+  std::pair<bool, Voltage> getUserVoltage(sta::Corner* corner,
+                                          odb::dbNet* net) const;
+  std::pair<bool, Voltage> getSolutionVoltage(sta::Corner* corner) const;
```
maliberty, line 140: I believe the bool here is for success.  Perhaps std::optional might be clearer.

maliberty:
I'm seeing a lot of failures.  First is aes/sky130hd
```
[INFO PSM-0073] Using bump pattern with x-pitch 140.0000um, y-pitch 140.0000um, and size 70.0000um with an reduction factor of 3x.

[ERROR PSM-0010] LU factorization of the G Matrix failed. SparseLU solver message: THE MATRIX IS STRUCTURALLY SINGULAR ... ZERO COLUMN AT 26640.
```

maliberty:
All designs for sky130hd/hs

gadfort:
@maliberty for the hd designs it looks like I wasn't filtering out the nwell and pwell pins, that should be fixed when I push (tested on the aes design).
for the hs designs, I think it's the wrong tap cell that is getting used (assuming it's the ORFS setup). Currently it's using `sky130_fd_sc_hs__tap_1` but it should be `sky130_fd_sc_hs__tapvpwrvgnd_1`. On the aes design it passes if I change the tap cell (without that it reports a bunch of opens which is correct).

src/psm/src/ir_solver.h
gadfort: good, point, I forget the std::optional is available

maliberty:
I think you are right about the tapcell so I'll change it.

maliberty:
tapcell fix https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1919

maliberty:
I started a new CI with the latest code + tapcell change.

maliberty:
I think we should do further verification but I don't think we need to hold the PR for it.  The CI was fine.

maliberty (approved):

PR #4853, opened 2024-03-26
deltaDebug: do not communicate with destroyed database

oharboe:


maliberty:
@habibayassin @vvbandeira FYI

PR #4856, opened 2024-03-26
gpl: Update README.md with debug mode instructions

gudeh:
Added detailed description of parameters in debug mode with an usage example.


src/gpl/README.md
```diff
@@ -114,19 +114,40 @@ global_placement
 | `-timing_driven_net_weight_max` | Set the multiplier for the most timing-critical nets. The default value is `1.9`, and the allowed values are floats. |
 | `-timing_driven_nets_percentage` | Set the reweighted percentage of nets in timing-driven mode. The default value is 10. Allowed values are floats `[0, 100]`. |
 
+### Debug Mode
+
+The `gpl::global_placement_debug` command initiates a debug mode, enabling real-time visualization of the algorithm's progress on the layout. Use the command prior to executing the `global_placement` command, for example in the `global_place.tcl` script.
```
maliberty, line 119: If we are going to make this public then we should promote it out of the gpl:: namespace.

gudeh: Alright, I added it because I could not run the command without the "gpl::"

maliberty:
I mean you'll have to modify the .i file to move the command to the global namespace.  Generally we discourage users from using commands inside a namespace

maliberty (approved):

PR #4862, opened 2024-03-27
mpl2: fix bug introduced in io blockages computation fix

AcKoucher:
A bug was introduced in #4839. I realized now when refactoring.

The bug was avoiding the generation of an io blockage for the left boundary.


maliberty (approved):

maliberty:
I see a few to investigate already in https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-All-Tests-Private/detail/secure-mpl2-io-bug/1/pipeline/

AcKoucher:
@maliberty It looks like not creating the blockage for pin access in the left boundary was generating results that were not so compromised by the boundary penalty bug.

I think that merging that one before this fix might be a better approach. I'll start a secure-ci for the boundary fix now that @eder-matheus has merged the fix for the congestion problem @gudeh and I were seeing in sky130hd/microwatt.

maliberty:
ok, the other alternative is to combine them into a single change

PR #4866, opened 2024-03-28
Fix pin acccess for 130BCD IO library

gatecat:
The middle of the pin needs to be used even though it intersects two routing tracks, as either track is too close to the blockages around the pin.

As a result, increase the threshold for checking the middle of the pin as an option from two to three routing tracks.

Let me know if you prefer a more specific solution.


maliberty:
It seems reasonable.  @osamahammad21 please run a full CI to verify

maliberty (approved):

PR #4874, opened 2024-03-30
[Docs] Manpage hotfixes for CI 

luarss:
- Hotfix for CI: try to avoid using `&&` in workflow + fix bug in preprocessing script that caused `./md/man3` not being created.
- Update new tcl syntax to be compliant with docs test.

This is a TODO: our current parser cannot handle this new format `ord::parse_list_args` so we are ignoring problematic functions for now (4 functions affected)

image


.github/workflows/github-actions-docs-test.yml
```diff
@@ -7,28 +7,34 @@ jobs:
...
-    - name: Set up Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: '3.10'  
+        apt-get update 
+        apt-get install -y pandoc
+
+    - name: Build OpenROAD
+      run: |
+        ./etc/Build.sh -compiler='clang-16' -cmake='-DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_STANDARD=20'
```
maliberty, line 26: Do you actually need to compile OR in order to generate the docs?  I'm wondering if there is docs target.

luarss: We just need the `messages.txt` generated as part of the compilation. I can simulate it with a script i suppose

.github/workflows/github-actions-docs-test.yml
```diff
@@ -7,28 +7,33 @@ jobs:
...
+        cd docs
+        make preprocess -j${nproc}
     - name: Run Tcl syntax parser
       run: |
         python3 docs/src/test/man_tcl_params.py
     - name: Run readme parser
       run : |
-        cd docs && make clean
+        cd docs
+        make clean
```
maliberty, line 38: Above you do `make preprocess` and here `make clean` - is this intentional?

luarss: Yes, the second one is a README parsing check. So it doesn't require manpages.

vvbandeira:
@luarss, please merge master into your branch so CI can pass.

vvbandeira:
Tests preventing merge:
```
gui_man_tcl_check (py) *FAIL*
upf_man_tcl_check (py) *FAIL*
ppl_man_tcl_check (py) *FAIL*
stt_man_tcl_check (py) *FAIL*
```

luarss:
@maliberty @vvbandeira, may I remove the dynamic tests for now and add them back in a separate PR? It needs some work:

- One common point of failure is the printing of specific number of readme / proc / help commands parsed. This should not be the case for dynamic tests. 
- Rather they should be a simple check whether commands parsed are equal or not.

vvbandeira (commented):
LGTM

maliberty (approved):

rovinski:
does a doc check even need a .ok file? Seem like it should just be able to return non-zero on failure.

PR #4894, opened 2024-04-03
gui: display list of clocksand their period/frequency in charts

### LINKED ISSUE #4820: Add clock period information to histogram
### Description

I found myself not knowing the clock period when I was explaining the histogram.

image

### Suggested Solution

Some sort of indication of clock period, perhaps, "Slack [ps] clock period 8500"?

### Additional Context

_No response_

AcKoucher:
Fix #4820

### DIFF
--- a/src/gui/src/chartsWidget.cpp
+++ b/src/gui/src/chartsWidget.cpp
@@ -44,6 +44,8 @@
 #include <cmath>
 #include <vector>
 
+#include "sta/Clock.hh"
+#include "sta/Fuzzy.hh"
 #include "sta/MinMax.hh"
 #include "sta/Units.hh"
 #include "staGuiInterface.h"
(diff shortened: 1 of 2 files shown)


AcKoucher:
@oharboe 

Some examples with these changes:

<p float="left">
  
   
</p>

Does this covers what you asked?

maliberty:
ready for review?

oharboe:
@AcKoucher Nice! I would remove the GHz duplicate information. Too busy. Also, generally in the GUI, scripts and reports only time(not frequency) is listed, so this would be more consistent.

In a future improvement, it would be nice to be able to filter what goes into the histogram: by clocks, endpoints, etc.

maliberty:
I actually asked him to add GHz thinking it is useful :disappointed:

oharboe:
> I actually asked him to add GHz thinking it is useful 😞

:-) I was torn about what I thought about it, but I landed on consistency and minimalism.

Then I wondered why here? Why not time and frequency *everywhere*?

So I landed on time only everywhere and consistency with scripts/reports, etc.

Let's say there are rounding errors. Which is authoritative?

Time period is the underlying data, no?

maliberty:
Time period is authoritative but even there we are rounding for display to 4 digits (you don't want 3.999999999998).

I thought it useful here because you are likely to think about the frequency implied but I'm not attached.  I don't think most tools even display the clock period never mind the frequency.

maliberty:
An earlier incarnation:  
image

oharboe:
Looking at that earlier version, I expect rounding to be consistent in the diagram: same as x-axis labels and mouse tip.

I think that there is enough width to list all the clocks on a single line, breaking into multiple lines only when there isn't space. Normally, I think, there would be a single clock.

oharboe:
> @oharboe Regarding the rounding part, are you referring to the decimal part here? image

Yes, though looking at it, it makes sense to have different precision on labels for bins and in clock period.

maliberty:
Let's clone on whether to have the frequency or not.  My impression is @oharboe would prefer it removed.  I don't have a strong feeling about it.  @oharboe do you want it removed?

oharboe:
> Let's clone on whether to have the frequency or not. My impression is @oharboe would prefer it removed. I don't have a strong feeling about it. @oharboe do you want it removed?

That would be my recommendation. No strong feelings and I dont consider myself a UX expert.

maliberty:
Ok let's remove it.  Sorry for the wasted work

oharboe:
@AcKoucher Final screenshot?

maliberty (approved):

PR #4895, opened 2024-04-03
odb: avoid skipping bbox computation for inst when block is not valid

### LINKED ISSUE #4032: zooming into macro with large offset, they disappear from view
### Describe the bug

This is a follow up of #3998.

When zooming in, the contents of macros with large offset disappears from view.

Same behavior whether top references macros for a, b and c or whether it's hierarchical def.

It seems that the "is in view" calculation doesn't properly take offset into account.

### Expected Behavior

Display should always show the macro contents.

### Environment

```shell
v2.0-10398-gecc03c290 Sept 23
M1 compiled with mpl2/par patch
```

### To Reproduce

make or 

and then zoom into macro b or c and you eventually get a blank screen.

[test_origin_hier.tgz](https://github.com/The-OpenROAD-Project/OpenROAD/files/12707685/test_origin_hier.tgz)

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
Fix #4032

The callback cpp test was considering that, for a created inst whose block has not a valid bbox, calling setOrigin would only be effective in the first call. However, with the fix here, as long as the block has not a valid bbox, every setOrigin call should trigger callback events.

### DIFF
--- a/src/odb/test/cpp/TestCallBacks.cpp
+++ b/src/odb/test/cpp/TestCallBacks.cpp
@@ -61,10 +61,11 @@ BOOST_AUTO_TEST_CASE(test_inst_and_iterm)
   BOOST_TEST(cb->events.size() == 2);
   BOOST_TEST(cb->events[0] == "PreMove inst i1");
   BOOST_TEST(cb->events[1] == "PostMove inst i1");
+  cb->clearEvents();
   i1->findITerm("a")->connect(n1);
-  BOOST_TEST(cb->events.size() == 4);
-  BOOST_TEST(cb->events[2] == "PreConnect iterm to net n1");
-  BOOST_TEST(cb->events[3] == "PostConnect iterm to net n1");
+  BOOST_TEST(cb->events.size() == 2);
+  BOOST_TEST(cb->events[0] == "PreConnect iterm to net n1");
+  BOOST_TEST(cb->events[1] == "PostConnect iterm to net n1");
   cb->clearEvents();
   i1->findITerm("a")->connect(n1);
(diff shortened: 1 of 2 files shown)


maliberty (approved):

maliberty:
Please look at the odb test failure

src/odb/test/cpp/TestCallBacks.cpp
```diff
@@ -58,11 +58,13 @@ BOOST_AUTO_TEST_CASE(test_inst_and_iterm)
...
   cb->clearEvents();
   i1->setOrigin(100, 100);
-  BOOST_TEST(cb->events.size() == 0);
-  i1->findITerm("a")->connect(n1);
   BOOST_TEST(cb->events.size() == 2);
-  BOOST_TEST(cb->events[0] == "PreConnect iterm to net n1");
-  BOOST_TEST(cb->events[1] == "PostConnect iterm to net n1");
+  BOOST_TEST(cb->events[0] == "PreMove inst i1");
+  BOOST_TEST(cb->events[1] == "PostMove inst i1");
+  i1->findITerm("a")->connect(n1);
```
maliberty, line 64: Add a `cb->clearEvents();` before connect to separate events between setOrigin and connect.

AcKoucher:
@maliberty I ran Secure-CI just to be sure that there wouldn't be an problem. It only shows the asap7 aes-block failure which we see on master.

maliberty (approved):

maliberty:
mac is having IT issues and can be ignored.

PR #4896, opened 2024-04-03
Build libspdlog.a with -fPIC by default

dlmiles:
-fPIC is required so the static lib*.a we are building can also be used inside a DSO on Linux.

See also https://github.com/gabime/spdlog/issues/1405

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -178,7 +178,7 @@ _installCommonDev() {
         cd "${baseDir}"
         git clone --depth=1 -b "v${spdlogVersion}" https://github.com/gabime/spdlog.git
         cd spdlog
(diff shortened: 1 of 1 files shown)


maliberty:
Thanks for the fix but I am confused about the issue.  Aren't we building a static .a and not a DSO?  OR is statically linked.

Can you describe the scenario in which this problem manifests?

dlmiles:
> Can you describe the scenario in which this problem manifests?

Language bindings to other languages require the component (OR) to be loaded as a DSO.

An additional SWIG configuration is added to OR tree, built alongside existing SWIG and the output is a DLL and linkage files that provide access to APIs.  At the moment ODB API access is the main goal.

When auditing it was found spdlog the only component preventing this from happening as-is, because it emitted a particular non-PIC symbol by default.  As this seemed a simple fix putting this commit in is to understand if there are any negative impacts on doing this.

Another strategy maybe to replace the spdlog component completely with an C++ ABI compatible component that builds and links with the DLL to provides equivalent function (i.e. not using spdlog at all), with the other language taking control of logging.

Another strategy is to build spdlog twice, once like now and once with -fPIC.

PR #4898, opened 2024-04-03
odb: dbTypes API uniformity

dlmiles:
REQUEST FOR FEEDBACK  (hence draft)

In working on language bindings for other languages these things came up

There are a number of matters in the one commit (please advise if you want commit broken up) :

Typo corrections
Renaming variables that were copy-and-paste but whose name is now a poor choice
Completely incorrect type naming use in command (due to copy-and-paste error)

More major concerns:

Discrepancies between the comment and the API
Mismatches over the default value
Missing methods that exist in header
Missing else clauses when nothing else matches (one appears fixed by 08026ed496)
Use of std::optional to allow parsing to return no Value, so we don't have to rewrite all string parsers for bindings to correctly detect input error.  Maybe you don't like the header introduced here, but I think it is allowed by your C++ language minimum now.
Unification of constructor choices
mismatch with noarg constructor: there all warrant an inspection to determine what is the correct default value


rovinski:
Iff is usually shorthand for "if and only if". It might not be a typo.

maliberty:
You have a good eye for detail!  My only concern is whether std::optional plays well with swig into tcl & python.  Have you tested the news APIs from those languages?

dlmiles:
> You have a good eye for detail! My only concern is whether std::optional plays well with swig into tcl & python. Have you tested the news APIs from those languages?

I don't believe I changed the behaviour of any external symbol concerning the use of std::optional.

I did add new ABI symbols, maybe the language bindings don't see or pickup at this time (will look and confirm).

The language bindings may already have their own string to enum parser (a duplication of code).  Part of the reason to move to std::optional is to consider removing that duplication.  Since I was about to write yet another version for another language when there was an obvious use case to make it available via `fromString(...)` API.

Python also supports `Optional[]` directly if that was a concern.  TCL maybe `nil` is more appropriate, should the language bindings expose the API more directly.

As this is draft I can check over those things to report back exactly how TCL/Python handled the extra APIs.


The main concern is to highlight areas that are inconsistent concerning the concept of a default value for each type and if there is any code reliant on those inconsistencies.  Maybe an audit, coverage can be done later


With the commit ctidy-style-bot what is necessary ?  If I implement the changes just on the lines it requests the file will look strange.  If that is necessary then maybe a separate commit that is just whitespace/format changes applied to the whole file first.  Then I commit this stuff on top.

maliberty:
My concern is whether swig is able to establish a useable API in practice.   You can ignore the clang-tidy issues in odb as it hasn't been cleaned up yet.

dlmiles:
> My concern is whether swig is able to establish a useable API in practice. You can ignore the clang-tidy issues in odb as it hasn't been cleaned up yet.

That consideration has not been a concern to me, I am not intending at this time to expose the `std::optional` automatically via any SWIG mechanism.  I am expecting SWIG to ignore the `fromString()` and emit no API in the foreign language for this method.  I shall get an answer to your queries and respond more fully on them in this PR (also explore exactly what the options available are, as I want to know too).

There is already an equivalent API in TCL/Python for translating a `string` into an `enum` type.  So it is not necessary nor intended for `fromString()` to be exposed via SWIG at this time.  This existing API code is in C++ as well, so is well capable of handling `std::optional` when seen.

At this time duplicate mechanisms exist to perform this translation, one for each language, this results in increased code size and maybe increased string table sizes (although maybe the linker is really good here).

The intent of the change, is to provide a sharable C++ code version of the important parts of this string to enum conversion.  Instead of me writing yet another version of the same making 3 copies of similar code in the code base, one for TCL, one for Python one for the next language.

At a later point in time I may offer an additional patch to optimize and rework the existing TCL and Python string to enum conversions using the shared API, this should result in an overall code size reduction.   I would however prefer the foreign languages (where possible) to have a appropriate unit test coverage (just like the new language will :)

Here is an example if such duplication (concerning use of strcasecmp() in a switch like if):

https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/src/swig/tcl/dbenums.i#L34  (TCL dbOrientType)

https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/src/swig/python/dbenums.i#L26  (Python dbOrientType)

If we then imagine another language is added, do we copy-and-paste ?

Note these are C++ snippets that are a shim layer to translate original C++ API into the foreign language C++ and are well capable of handling `std::optional` the resulting new method might look like:

```
// where "langstring_t" is the new language specific type for a string
// where "env" is some kind of language handle to environment
static langstring_t SWIG_Custom_ConvertEnum_dbOrientType_to_langstring(LangEnv *env, const odb::dbOrientType::Value &value) {
  auto opt = dbOrientType.fromString(value)  // using the new <std::optional> that is the subject of this PR comment
  if(opt.has_value()) {
    return env->makeStringInForeignLanguage(opt.value())
  } else {
    SWIG_ThrowException(jenv, SWIG_LanguageExceptionType, "Exception Message")
    return (langstring_t) 0
  }
}
```

I believe all language binding snippets can be transformed into only what is necessary, all the string conversion, comparisons, what is the DEFAULT is all pushed back towards the origin (the C++ API) and managed there.
Reducing errors, reducing duplication, making maintenance easier (a new type added as C++ is automatically propagated to the SWIG API, without the need for fixes)

At the moment I see the usual SWIG maintenance nightmare, where the language bindings become out-of-date, out-of-sync, so I will take efforts to remove the problem altogether when I can and eventually reduce overall code size.  See my recent commit #4897 as an example of the problem I believe my approach here should eliminate.


Sorry for the longer replies but it maybe necessary to fully explain and set the scene as a new contributor to this project of the nature of the work being done.  Language bindings are a slow burn matter, that do require an attention to detail to vet and inspect every item in due course.

maliberty:
> db.h is swig'ed so you are.

Sorry you are touching dbTypes.h which isn't directly swigged so this should be ok.  You do need to run clang-format on your change.  Is there any other feedback you are seeking?

dlmiles:
I have checked out, cleanup and CI tested (the original commit) on a few designs the original patch just to check it seems ok.

Separated the commit into:
 * low risk (should be comment only changes)
 * medium risk (additional code added, there should be no breaking ABI changes and just new symbols added)
 * fix for missing initializer (removed the other one as 08026ed496 already in tree has the same purpose)

This should pass clang-format checking as well now.

So this can be picked up if there is a hurry.

With the std::optional there will be additional cases to follow ?

Would you accept a future commit to fix the parsed string discrepancies (mark by comment at this time), so the DEFAULT value is set, helps make the external foreign language API more consistent ?

---

$ clang-format --dump-config | grep ColumnLimit
ColumnLimit:     80

Is this a sane default in 2024 the era of 4K monitors and high DPI displays ?   132 Columns was the next size up and still feels sane maximum today, I think such terminals have existed for over 40 years.  Some polices might use 150% of 80 which ends up at 120 as a happy medium between the factions of opinion.

ColumnLimit: 80 makes this seem preferred, if that is how the robot wants it
```
+std::optional<dbTechLayerType::Value> dbTechLayerType::fromString(
+    const char* value)
```

dlmiles:
> Iff is usually shorthand for "if and only if". It might not be a typo.

On this... and I though the "IFF" was a VBA VisualBasic language keyword and was attempting to have a similar meaning here.

When I read out load the comment and try to understand the comment, how does using IFF improve clarity of meaning ?  Did replacing it with just a simple "if" change the understanding somehow ?  Maybe someone can explain.  Who would understand use of "if" incorrectly ?  Then thinking about choice of language and words internationally what readers more likely to understand.  This is why it was changed.

dlmiles:
https://github.com/The-OpenROAD-Project/OpenROAD/compare/master..7095e16b8cdb1c98cd87a51e87296982b5d2b33b#diff-61f7e663d03c2ab600e5251319960ed6eaba3dc8021e555bb743a57a63ca92e9R1374  "TECH_VIA" or "TECH VIA" this has the best choice been made to remove underscore here?   The change is to make the string<>enum<>string and enum<>string<>enum reciprocal.

rovinski:
> how does using IFF improve clarity of meaning ?

https://en.wikipedia.org/wiki/If_and_only_if

If means implication, but iff means bi-implication. For example:
```
  ///
  ///  True Iff value corresponds to POWER or GROUND
  ///
  bool isSupply() const;
```
Semantically speaking, this means that `isSupply()` is true if value is POWER or value is GROUND. Otherwise it will be false.

```
  ///
  ///  True If value corresponds to POWER or GROUND
  ///
  bool isSupply() const;
```
Semantically speaking, this means that `isSupply()` is true if value is POWER or value is GROUND. There is no implication on what `isSupply()` returns if the value is neither POWER or GROUND.

Does this clarify the meaning beyond "if"? No, not really. I was just mentioning that it probably wasn't a typo.

maliberty:
I think the TECH VIA change is fine.  The 'iff' is mathematically correct though not especially important.

I think you just need to resolve the conflict and this can move forward.

maliberty:
> My concern is whether swig is able to establish a useable API in practice. You can ignore the clang-tidy issues in odb as it hasn't been cleaned up yet.

odb has been clang-tidy'ed so I went ahead and fixed the issues in this PR.

maliberty (approved):

PR #4899, opened 2024-04-03
Gpl sycl

UDXS:


maliberty:
I guess this was a mistake but just FYI we do have some work going on with gpl & cuda.  I've been thinking about https://github.com/kokkos/kokkos as a cross-platform way to handle such code.  Perhaps it would be of interest to you.

UDXS:
Indeed, this was a mistake. I was intending to eventually subsume the CUDA support for gpl with a SYCL implementation (as  a more cross-platform accelerator API) but I will look into Kokkos as well

PR #4910, opened 2024-04-05
[Docs] Add manpages to Readthedocs

luarss:
@vvbandeira @dralabeing

### DIFF
--- a/src/utl/README.md
+++ b/src/utl/README.md
@@ -33,11 +33,11 @@ man
 
 ## Example scripts
 
-You may run various commands
+You may run various commands or message IDs for manpages.
 ```
-man openroad; level 1
-man clock_tree_synthesis; level 2
-man CTS-0005; level 3
+man openroad
+man clock_tree_synthesis
+man CTS-0005
 ```
 
 ## Regression tests
--- a/src/utl/README.md
+++ b/src/utl/README.md
@@ -12,7 +12,7 @@ The utility module contains the `man` command.
 ## Man installation
 
 The `man` command can be installed optionally as part of the OpenROAD
-binary. For more instructions, please refer to [here](../../docs/README.md).
+binary. For more instructions, please refer to [here](manREADME.md).
 
 ### Man
 
--- a/src/utl/README.md
+++ b/src/utl/README.md
@@ -33,6 +33,13 @@ man
 
 ## Example scripts
 
+You may run various commands
+```
+man openroad; level 1
+man clock_tree_synthesis; level 2
+man CTS-0005; level 3
+```
+
 ## Regression tests
 
 There are a set of regression tests in `./test`. For more information, refer to this [section](../../README.md#regression-tests).
--- a/src/utl/README.md
+++ b/src/utl/README.md
@@ -31,20 +31,20 @@ This can be used for a range of commands in different levels as follows:
 man
     name
     [-manpath manpath]
-    [-no_query]
+    [-no_pager]
 ```
 
 #### Options
 
 | Switch Name | Description | 
 | ----- | ----- |
-| `-name` | Name of the command/message to query. |
-| `-manpath` | Include optional path to manpage pages (e.g. ~/OpenROAD/docs/cat). |
-| `-no_query` | This flag determines whether you wish to see all of the man output at once. Default value is `False`, which shows a buffered output. |
+| `name` | Name of the command/message to query. |
+| `-manpath` | Include optional path to man pages (e.g. ~/OpenROAD/docs/cat). |
+| `-no_pager` | This flag determines whether you wish to see all of the man output at once. Default value is `False`, which shows a buffered output. |
 
 ## Example scripts
 
-You may run various commands or message IDs for manpages.
+You may run various commands or message IDs for man pages.
 ```
 man openroad
 man clock_tree_synthesis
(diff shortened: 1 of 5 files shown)


src/utl/README.md
```diff
@@ -33,6 +33,13 @@ man
 
 ## Example scripts
 
+You may run various commands
+```
+man openroad; level 1
+man clock_tree_synthesis; level 2
+man CTS-0005; level 3
```
maliberty, line 40: The 'level #' are confusing - are they actually commands?

luarss: Updated

src/utl/README.md
```diff
@@ -33,11 +33,11 @@ man
 
 ## Example scripts
 
-You may run various commands
+You may run various commands or message IDs for manpages.
```
dralabeing, line 36: These three examples are very cryptic. Do you want to add a note on how to get help on man itself? Also, include a few examples showing levels as switches to show what these levels mean and also how much is displayed.

Will you be adding the installation section as well ? (pandoc, build) 

Are you planning to add any docs for man pages integration for other flows? Is this supported?

dralabeing (changes requested):

src/utl/README.md
luarss:
```
These three examples are very cryptic. Do you want to add a note on how to get help on man itself? Also, include a few examples showing levels as switches to show what these levels mean and also how much is displayed.
```
This is the page currently for manpages. I am not sure what you refer to by "levels as switches", the commands can be run as-is. 
https://openroad--4910.org.readthedocs.build/en/4910/main/src/utl/README.html

```
Will you be adding the installation section as well ? (pandoc, build)
```
Good point, we do have a `docs` README that is not being used. Does this look okay?
https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/docs

```
Are you planning to add any docs for man pages integration for other flows? Is this supported?
```
It is supported as long as the flows upgrade to a recent version of OR, as it is optionally compiled with the OR binary. Do you happen to have the list of flows that depends on OR?

dralabeing: @luarss here are some points that aren't clear (to me):
1. there is some confusion in the naming. e.g in the man page syntax here :
   man
    name
    [-manpath manpath]
    [-no_query]
the name seems as required. however it referred to as optional:
https://openroad--4910.org.readthedocs.build/en/4910/main/src/utl/README.html#options
also is manpath, no_query a flag or a switch?

2. also "level" is an implementation detail, correct? Should users care about it? It is a useful thing to know that this is a feature that has been implemented but I am not clear about specifying levels as a command. Anyway, this has been updated but I am wondering if it has other explicit uses.
There was also some suggestion about the use of "_no_query" (name) and support for tcl blocks (level 2). Has this been done ? Should users care about this?

4. The introduction of man under utilities is very brief. https://openroad--4910.org.readthedocs.build/en/4910/main/src/utl/README.html#utilities
Can you add a brief description its use since this is the first landing page for this command? 
Something like :" The man command in OpenROAD is similar in functionality to Unix (and Unix-like operating systems such as Linux) . It is used to display the manual pages for various applications, tools and error messages.  These manual pages provide detailed information about how to use a particular command or function, along with its syntax and options. 

This is an optional package. You will need to explicitly build and install this in OpenROAD and OpenROAD-flow-scripts to be able to use this command. 
Refer to <> for details on installation" 

Currently, these instructions for build and usage are disconnected and navigation to access this info is not clear. Are users likely to look for the docs README? It seems to me not since it seems to have generic information about doc installation.

5. The link to regressions under https://openroad--4910.org.readthedocs.build/en/4910/main/src/utl/README.html#regression-tests (./test)  does not work.
6. Change the text in the side bar and all other doc references from Man pages to Man pages.
7.  I am not sure if the docs README here :https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/docs is appropriate for man README. Check with @vvbandeira on this. I think there should be intuitive navigation from building, installing and using man pages. 

8. Developer section. This would refer to instructions for new tool additions and incorporation into man. What do they need to know to compile and add man?  I was also referring to support of man for other flows built on top of OpenROAD like ORFS, OpenLane. Is there intended support for this? This is a general use question.

If it is helpful to close we can have a quick call.

thanks

luarss: 1. `name` is required, the other two are optional flags/variables as shown here: 
image

2. `There was also some suggestion about the use of "_no_query" (name) and support for tcl blocks (level 2). Has this been done ? Should users care about this?` What was the suggestion? 

5. It does work. It's purpose is to link back to the main README regarding the regression tests. It just takes some time to load.
image

6. You wrote the same thing? I am guessing you want `Manpages`?
7. Yes, I can add a link to this document via the utl README.
8. Developer section is already in with ReadmeFormat and TclFormat, Manpage Test Framework. Perhaps I can also add a link to these. 
image

dralabeing: 1. I was referring to the syntax  "-name" as explained as opposed to "name" . - implies it is optional.
2.  Please check the recording of your presentation for details for feedback on this. for e.g we talked about not using no_query as the arg name. it is similar to -no--pager option in git so should be similarly named.  

Also, can error messages be made more actionable by allowing developers to annotate extra text to the pre-existing messages in messages.txt.  is there a plan for this implementation?
3. I was referring to the ./test link. The link to regression tests works fine. It would be good to enable navigation nagivation to this directory--(this is the boiler plate for all README files0. please ignore if it is too much of big deal.

4. Change Manpages to Man pages --it is not one word

luarss: 2. Updated the `no-pager` in code and in documentation. 
3. The objective is to say `./test` is the folder containing the useful regression test scripts. It is not meant as a link, to be consistent with the other module READMEs.

luarss:
Changes made:
- Add install instructions for Manpage in utl/README and Build.md
image

image

dralabeing (approved):

dralabeing:
Looks good!

dralabeing (approved):

maliberty (approved):

PR #4918, opened 2024-04-07
gui: use boost geom for shapes

gadfort:
Changes:
- converts shapes.h and shapes.cpp to use the boost geom interface.

Notes:
- dbBox* could be further simplified to Rect, but it appears that accessing the dbBox requires several calculations to find the pointer, so I left that in (still reduces the total memory needed).

### DIFF
--- a/src/gui/src/search.cpp
+++ b/src/gui/src/search.cpp
@@ -289,10 +289,8 @@ void Search::updateShapes(odb::dbBlock* block)
         if (!box) {
           continue;
         }
-        const Box bbox(Point(box->xMin(), box->yMin()),
-                       Point(box->xMax(), box->yMax()));
         odb::dbTechLayer* layer = box->getTechLayer();
-        net_shapes[layer].emplace_back(bbox, false, term->getNet());
+        net_shapes[layer].emplace_back(box->getBox(), false, term->getNet());
       }
     }
   }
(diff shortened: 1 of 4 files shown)


src/gui/src/search.cpp
```diff
@@ -363,16 +354,12 @@ void Search::updateBlockages(odb::dbBlock* block)
...
 
-  std::vector<BoxValue<odb::dbBlockage*>> blockages;
+  std::vector<DBoxValue<odb::dbBlockage*>> blockages;
   for (odb::dbBlockage* blockage : block->getBlockages()) {
-    odb::dbBox* bbox = blockage->getBBox();
-    const Point ll(bbox->xMin(), bbox->yMin());
-    const Point ur(bbox->xMax(), bbox->yMax());
-    const Box box(ll, ur);
-    blockages.emplace_back(box, blockage);
+    blockages.emplace_back(blockage->getBBox(), blockage);
```
maliberty, line 359: Why do we need to store both?  Couldn't we just have the index getter work directly on the blockage?  Other types may be similar.

gadfort: I think we can, I got confused by some slowdowns in the rendering (I just sent you an email about it), which I thought might have been related to this, but now I don't. I'll try switching the dbBox stuff over and if there isn't an measurable rendering time impact, I'll push it up.

maliberty (approved):

PR #4922, opened 2024-04-07
pdn: switch pdn to use geom boost structures

gadfort:
No functional changes expected, just cleaning up the rtree to use the odb boost geom.

### DIFF
--- a/src/pdn/src/via.h
+++ b/src/pdn/src/via.h
@@ -74,12 +74,8 @@ class Shape;
 class Via;
 class ViaGenerator;
 
-using Point = bg::model::d2::point_xy<int, bg::cs::cartesian>;
-using Box = bg::model::box<Point>;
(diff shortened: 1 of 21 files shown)


src/pdn/src/via.h
```diff
@@ -729,6 +723,13 @@ class TechViaGenerator : public ViaGenerator
 class Via
 {
  public:
+  struct ViaIndexableGetter
+  {
+    using result_type = odb::Rect;
+    odb::Rect operator()(const ViaPtr& via) const { return via->getArea(); }
```
maliberty, line 729: What is the 'area' of the via?  It seems like a bbox.

maliberty (approved):

src/pdn/src/via.h
gadfort: Yeah, it's the bounding box

Issue #4785, opened 2024-03-13
GRT: Move the resources reduction calculation from GRT to ODB


eder-matheus:
Move the resource reduction calculation for fixed objects (instance obstructions, macros, PG nets) from GRT to ODB, in order to share this calculation between GRT and RUDY.

Issue #4917, opened 2024-04-06
Failing to run globalroute for MegaBoom on Nangate45.


sakundu:
### Description

For one of our projects, we used OpenROAD and ran P&R for MegaBoom on Nangate45. We found that even after reducing the floorplan utilization to 20%, OpenROAD fails to run global routing. We have updated the fastroute.tcl file to allow more routing resources than the default. It would be good if you could suggest some solutions or enhance your global route engine to fix the problem.

We have uploaded the CTS, [DEF, Verilog, and SDC](https://drive.google.com/file/d/1NW_r3H2cDUbDIgkyVFP-Cf6W3WjD65YN/view?usp=sharing) files that were the input to the global route stage. We use the default Nangate 45 lef and lib available  in the ORFS GitHub.

Please let us know if you have any suggestions to fix the problem, and if you need any more information to debug it further.

Thanks,
Sayak

stefanottili:
Using the macro.mod.lef to get the TAP cell to load def and then display the RUDY routeablilty heat map  ..

OpenROAD-flow-scripts/flow/platforms/nangate45/lef/NangateOpenCellLibrary.tech.lef, created 22 layers, 27 vias
OpenROAD-flow-scripts/flow/platforms/nangate45/lef/NangateOpenCellLibrary.macro.mod.lef, created 135 library cells

This doesn't look like a GRT problem at all, this looks like GIGO, a bad netlist being unplaceable.
Simply querying the instances in the dark middle of the chip shows "regfile" as instance names.

Just synthesizing the abstract description of the "regfile" of a VLIW will not get you a placeable netlist.

Maybe physical synthesis tools got better over the last couple of years, but a regular structure like a regfile
will at the very least need careful attention to the structure of the muxes and very likely some data path placement.
You'll have to isolate the muxes in separate RTL hierarchy and script the placement of the sequential elements.

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/25827408/8e44d3ab-386b-492c-bcd6-6faef5cacf93)

gadfort:
I've seen this behavior a few times before when I synthesized a register file. I noticed that during global placement the register file would get to densities much greater than requested (90+%, even when the density requirement is set to 40%). The commercial tools struggle with this too, but the placement density they achieve is usually around 70%.

I would suggest taking a look at the global placement step maybe try turning off routabilty driven placement since that appears to be when the density climbs up.

oharboe:
[...]

PR #4927, opened 2024-04-08
grt: move rudy inside grt and share resource reduction calculation

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/4785


maliberty:
Please cleanup the failures and I'll review after.

src/grt/include/grt/GRoute.h
```diff
@@ -116,6 +117,9 @@ struct CongestionInformation
   std::set<odb::dbNet*, cmpById> sources;
 };
 
+using CapacityReductionData
+    = std::vector<std::vector<std::pair<uint8_t, uint8_t>>>;
```
maliberty, line 121: Give a comment as its not obvious what the indices or the pair are.  Better still would be to use a struct with meaningful names rather than a pair.

src/grt/src/Rudy.cpp
```diff
@@ -0,0 +1,190 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2024, The Regents of the University of California
```
maliberty, line 4: Its better to start using "Precision Innovations Inc." for new code

src/grt/src/Rudy.h
```diff
@@ -0,0 +1,106 @@
+///////////////////////////////////////////////////////////////////////////////
+// BSD 3-Clause License
+//
+// Copyright (c) 2024, The Regents of the University of California
```
maliberty, line 4: same

maliberty (approved):

maliberty:
@gudeh FYI

stefanottili:
[WARNING GUI-0066] Heat map "Estimated Congestion (RUDY)" has not been populated with data.

Before this commit, I could just read_def a placed design and have a look at the RUDY congestion
What do I have to do now ?

gudeh:
Hi @stefanottili, I am sorry for the inconvenience, I will have a look at what is happening.

stefanottili:
Is there a command to "compute RUDY" ?

I didn't quite like that fact that when having looked at RUDY data in a previous session, the gui keept that display state and then computed RUDY when loading a new def in a new session on. This behavior was unexpected, even though RUDY is lightning fast, it takes some time on larger designs. I'd prefer it to be only computed "on demand".

stefanottili:
Ah, I see RUDY is now computed when running global route ... 

On the mega boom issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/4917 global route and Rudy maps look quite different ... 
Is there room for calibrating RUDY or is that the price for a fast estimation ?

global route
image

rudy
image

eder-matheus:
> Ah, I see RUDY is now computed when running global route ...

This behavior is not expected. Rudy should be able to run without GRT run. I'm looking at this bug now.
Regarding the differences in megaboom, we did this PR to calibrate RUDY with GRT data about fixed objects, but the way both handle the nets is different. So I would expect a difference between both.
@gudeh Do you think the difference between the figures is reasonable, or they should be closer?

gudeh:
I believe this difference might be too much. The RUDY code is still under work and should be further modified. This divergence on the heatmaps is mainly due to  the fact that RUDY is unaware of layers, while GRT is.

gudeh:
@stefanottili this prints are from stage 6 right? The final stage. If it is, can you also send the pair from stage 3, please?

stefanottili:
All I did was to read_lef read_def of the placed mega boom def, run global_route and look at the maps.

PR #4939, opened 2024-04-10
psm: use double instead of float

gadfort:
Changes:
- switch Voltages to doubles to avoid some casting during solution generation
- use long-double when generating conductance matrix to avoid rounding error

Fixes:
- numerical instability in CI (hopefully).


src/psm/src/ir_solver.cpp
```diff
@@ -807,7 +807,7 @@ void IRSolver::buildCondMatrixAndVoltages(
       J[node_idx] = find_node->second;
     }
 
-    Connection::Conductance node_cond = 0.0;
+    long double node_cond = 0.0;
```
QuantamHD, line 810: Maybe we should add another using declartation called `Connection::ConductanceHighPrecision`

maliberty: long double may have performance implications (using x87 and not SSE).  Is this strictly necessary?

gadfort: I think I found a seconds stop that could avoid this... so I converted to draft until I can verify.

QuantamHD:
LGTM. Sort looks good

QuantamHD (approved):

src/psm/src/ir_solver.h
```diff
@@ -187,7 +187,8 @@ class IRSolver
   bool wasNodeVisited(const std::unique_ptr<Node>& node) const;
   bool wasNodeVisited(const Node* node) const;
 
-  std::map<Node*, std::set<Connection*>> getNodeConnectionMap(
+  std::map<Node*, std::set<Connection*, Connection::Compare>>
```
maliberty, line 190: `std::set<Connection*, Connection::Compare>` get repeated alot, perhaps `using ConnectionSet = std::set<Connection*, Connection::Compare>` ?

src/psm/src/ir_solver.h
```diff
@@ -62,7 +62,7 @@ class IRNetwork;
 class IRSolver
 {
  public:
-  using Voltage = float;
+  using Voltage = double;
   using Current = double;
   using Power = float;
```
maliberty, line 67: Any reason to stick with float for Power?

PR #4943, opened 2024-04-12
rsz: clang-format

eder-matheus:


maliberty:
Please remove the rsz exclusion from github/workflows/github-actions-format-on-push.yml

PR #4944, opened 2024-04-12
Odb: user derating relocation

gudeh:
Bringing back this modification, since it showed necessary. RUDY requires some way to load the user derating without overwriting the GRT information, now that it is shared. For that we opted to put the user layer adjustment in ODB, in dbTechLayer. The code generator was used to do so.


src/grt/src/GlobalRouter.cpp
```diff
@@ -1202,8 +1201,8 @@ void GlobalRouter::computeUserGlobalAdjustments(int min_routing_layer,
     return;
 
   for (int l = min_routing_layer; l <= max_routing_layer; l++) {
-    if (adjustments_[l] == 0) {
-      adjustments_[l] = adjustment_;
+    if (db_->getTech()->findRoutingLayer(l)->getLayerAdjustment() == 0.0) {
```
eder-matheus, line 1204: Refactor this long function calls:
```
odb::dbTechLayer* tech_layer = getTech()->findRoutingLayer(l);
tech_layer->getLayerAdjustment()
```

And use the tech_layer in the lines below.

src/grt/src/GlobalRouter.cpp
```diff
@@ -1218,7 +1217,8 @@ void GlobalRouter::computeUserLayerAdjustments(int max_routing_layer)
   const std::vector<int>& ver_capacities = grid_->getVerticalEdgesCapacities();
 
   for (int layer = 1; layer <= max_routing_layer; layer++) {
-    float adjustment = adjustments_[layer];
+    float adjustment
```
eder-matheus, line 1220: Same here. Even if you only use it once, I think it's good to separate it for readability.

eder-matheus (changes requested):

rovinski:
Would be a good idea to rebase this PR to avoid the spurious changes.

gudeh:
Closed due to too many unnecessary commits and failed rebase.

PR #4952, opened 2024-04-15
odb: relocate user derating value from GRT to dbTechLayer

gudeh:
RUDY requires some way to load the user derating without overwriting the GRT information, now that it is shared. For that we opted to put the user layer adjustment in ODB, in dbTechLayer. The code generator was used to do so.


eder-matheus (approved):
Let's wait for the secure-ci run before merging it.

PR #4954, opened 2024-04-15
grt: avoid breaking the connectivity between wire and pin

eder-matheus:


maliberty (approved):

maliberty:
Needs a format.  Secure CI might be good just to ensure nothing breaks in order wires

eder-matheus:
@maliberty Yeah, I already started a secure-ci and will update the formatting too.

PR #4976, opened 2024-04-19
mpl2: make hardSA final costs comparable across multiple runs

AcKoucher:
The weight variations (decrease WL weight & increase Outline weight) we use to ease valid results generation in macro placement (Hard SA) causes runs to have final costs which are actually not comparable.

The goal of these changes is to ensure we reset the weights to a common configuration so we can compare results across runs.

My intentions are to substitute the floats parameters for the weights' struct in a subsequent refactoring PR so we can eliminate those parameters when instantiating SACore.


AcKoucher:
@maliberty Is this what you had in mind for solving the weights' variations issues?

maliberty (approved):

PR #4978, opened 2024-04-20
etc: use brew to install or-tools in macOS

eder-matheus:
Reopening this as it was accidentally merged before without applying the changes.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -456,7 +456,8 @@ _installOpenSusePackages() {
 _installHomebrewPackage() {
     package=$1
     commit=$2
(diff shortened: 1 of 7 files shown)


etc/DependencyInstaller.sh
```diff
@@ -490,7 +491,7 @@ Then, rerun this script.
 EOF
       exit 1
     fi
-    brew install bison boost cmake eigen flex groff libomp pandoc pyqt5 python swig tcl-tk zlib
+    brew install bison boost cmake eigen flex groff libomp or-tools pandoc pyqt5 python tcl-tk zlib
```
maliberty, line 494: Why remove swig?

maliberty: nvm I see it is below

maliberty (approved):

maliberty:
I will be so glad to be done with this saga

PR #4986, opened 2024-04-23
Cts enable dummy load by default

arthurjolo:
This PR makes the CTS option to insert dummy load buffers at the end of CTS to be on by default.


maliberty:
I see a lot of failure that need addressing.  A secure CI will be required once those are cleaned up

src/cts/README.md
```diff
@@ -89,7 +89,7 @@ clock_tree_synthesis
 | `-obstruction_aware` | Enables obstruction-aware buffering such that clock buffers are not placed on top of blockages or hard macros. This option may reduce legalizer displacement, leading to better latency, skew or timing QoR.  The default value is `False`, and the allowed values are bool. |
 | `-apply_ndr` | Applies 2X spacing non-default rule to all clock nets except leaf-level nets. The default value is `False`. |
 | `-no_insertion_delay` | Don't consider insertion delays in macro timing models in balancing latencies between macro cells and registers. This option prevents construction of separate clock trees for macro cells and registers.  The default value is `False`. |
-| `-use_dummy_load` | Applies dummy buffer or inverter cells at clock tree leaves to balance loads.  The default values is `False`. |
+| `-dont_use_dummy_load` | Don't apply dummy buffer or inverter cells at clock tree leaves to balance loads.  The default values is `True`. |
```
precisionmoon, line 92: Shouldn't the default be 'False', meaning use dummy loads?

precisionmoon (approved):

maliberty:
There are a bunch of conflicts to resolve

arthurjolo:
All the conflicts were resolved and now there is just one regression design that gets an error during DRT.

maliberty:
Are you referring to `aes_sky130hd (tcl) *ERROR* ` ?

arthurjolo:
> Are you referring to `aes_sky130hd (tcl) *ERROR* ` ?

Yes, I have been talking with Osama on how to solve this.

arthurjolo:
Yes it has been throw a CI, there are a lot of benefits. Cho has a table with the results.
However there are 2 designs where the CI is failing and need to be rebased.

src/cts/test/array_ins_delay.ok
```diff
@@ -103,26 +103,32 @@
 [INFO CTS-0015]     Created 366 clock nets.
 [INFO CTS-0016]     Fanout distribution for the current clock = 2:1, 4:3, 5:9, 6:5, 7:4, 8:3, 9:4, 10:230..
 [INFO CTS-0017]     Max level of the clock tree: 5.
+[INFO CTS-0121] Buffer 'clkload3' has unconnected output pin.
+[INFO CTS-0121] Buffer 'clkload2' has unconnected output pin.
+[INFO CTS-0121] Inverter 'clkload1' has unconnected output pin.
+[INFO CTS-0121] Inverter 'clkload0' has unconnected output pin.
```
maliberty, line 109: We should avoid warnings if these are intentionally dangling

precisionmoon (commented):
Just a minor suggestion.  Why don't you add unordered_set to TritonCTS class?   You don't have to pass the extra argument around then.

arthurjolo:
@precisionmoon Since we only use it at the end and just once I decided to make it an argument, but it can be added to the TritonCTS class if you believe to the best option.

maliberty (approved):

Issue #1933, opened 2022-06-04
After run the flow successfully, the DRC is not clean


Elon-J:
When I try to increase the output load ability of my block, after passing the OpenRoad flow successfully, the DRC is not clean. What should I adjust?
Attach my detail_route_issue_file:

[detail_route_NON_CLK_GEN_sky130hd_base_2022-06-04_07-30.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/8837249/detail_route_NON_CLK_GEN_sky130hd_base_2022-06-04_07-30.tar.gz)

Attach my block repository:
https://github.com/Elon-J/OpenROAD-flow-scripts/tree/master/flow/designs/sky130hd/NON_CLK_GEN
Attach my report file, magic DRC report and layout DRC report:

[6_report.log](https://github.com/The-OpenROAD-Project/OpenROAD/files/8837223/6_report.log)

[sky130_drc.txt](https://github.com/The-OpenROAD-Project/OpenROAD/files/8837220/sky130_drc.txt)

[6_final_drc.txt](https://github.com/The-OpenROAD-Project/OpenROAD/files/8837222/6_final_drc.txt)

Elon-J:
<img width="437" alt="image" src="https://user-images.githubusercontent.com/88715865/171989190-1551505f-5146-4763-8252-ce3b018dbfe9.png">

Elon-J:
I also have try many verisons including adjust different core areas.

vijayank88:
@Elon-J 
Please execute following commands from `flow` directory and attach zip file.
```
make detail_route_issue
```

Elon-J:
OK

Elon-J:
I have updated my first comment.

vijayank88:
@Elon-J 
Detail routing completed with 0 DRC errors.
```
[INFO DRT-0195] Start 57th optimization iteration.
    Completing 10% with 0 violations.
    elapsed time = 00:00:00, memory = 149.02 (MB).
    Completing 20% with 0 violations.
    elapsed time = 00:00:00, memory = 149.02 (MB).
    Completing 30% with 0 violations.
    elapsed time = 00:00:00, memory = 149.03 (MB).
    Completing 40% with 0 violations.
    elapsed time = 00:00:00, memory = 149.03 (MB).
[INFO DRT-0199]   Number of violations = 0.
```
Its something error coming from KLayout.

Please try to load and check those overlapping error.

[...]

PR #4991, opened 2024-04-24
mpl2: post-annealing step to push macros to boundaries

AcKoucher:
[#1930](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1930) showed that we actually don't allow different x & y coordinates for macro placement halos in mpl2, so [#1933](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1933) was opened.

When I was testing the distinct x & y halos for mock-array's test case used in #4085, two problems came up:
- The macros' locations inside the array were wrong (should be fixed by #4976)
- The array was getting split in two as mpl2 was trying to push the macros to the boundaries by a post-annealing function.

This PR is to change the alignHardMacroGlobal function by a new step in mpl2 that happens before we improve orientation. Basically, we push macros that are close enough to the boundaries if:
1. They don't overlap with blockages generated for Pin Access
2. They don't overlap with other macros
3. Pushing won't destroy the shape of an array of macros.


AcKoucher:
It looks like there's something going on with clang-tidy.

maliberty:
secure ci?

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -6638,4 +6241,235 @@ void HierRTLMP::setDebugShowBundledNets(bool show_bundled_nets)
...
+  std::vector<HardMacro*> hard_macros = macro_cluster->getHardMacros();
+
+  int min_dx_to_core = std::numeric_limits<int>::max();
+  int min_dy_to_core = std::numeric_limits<int>::max();
+
+  for (HardMacro* hard_macro : hard_macros) {
+    hard_macros_.push_back(hard_macro);
+
+    min_dx_to_core = std::min(min_dy_to_core, hard_macro->getWidthDBU());
+    min_dy_to_core = std::min(min_dy_to_core, hard_macro->getHeightDBU());
```
maliberty, line 6329: The variable name seems to suggest a distance but the value is w/h.  Please explain your thinking.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -6638,4 +6241,235 @@ void HierRTLMP::setDebugShowBundledNets(bool show_bundled_nets)
...
+        macro_cluster, vertical_move_allowed, horizontal_move_allowed);
+
+    pushMacroClusterToCoreBoundaries(macro_cluster, boundaries_distance);
+  }
+}
+
+std::map<Boundary, int> Pusher::getDistanceToCloseBoundaries(
+    Cluster* macro_cluster,
+    bool vertical_move_allowed,
+    bool horizontal_move_allowed)
```
maliberty, line 6316: const

src/mpl2/src/hier_rtlmp.cpp
AcKoucher: The names should just be w/h as the value, those names weren't a good choice. I also realized that we don't need to iterate the HardMacros within the macro cluster and make all this std::min stuff as we only group macros of the same size (I'll comment this on the code).

AcKoucher:
I started a secure-ci

maliberty:
@osamahammad21 this comes with one new drc in ca53 but it seems to be just a case of perturbing things happens to trigger a routing issue.  I don't see a problem in the macro placement so I'll ask you to look at it but don't want to hold this PR for it.

PR #4999, opened 2024-04-26
rudy: avoid heatmap crash 

gudeh:
avoid crash when clicking on rudy heatmap and there are unplaced cells.

### DIFF
--- a/src/grt/src/heatMapRudy.cpp
+++ b/src/grt/src/heatMapRudy.cpp
@@ -63,7 +63,13 @@ bool RUDYDataSource::populateMap()
   if (!getBlock()) {
     return false;
   }
-
+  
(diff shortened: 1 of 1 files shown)


maliberty (commented):
Will the user get any notification as to why the heatmap isn't shown?

gudeh:
> Will the user get any notification as to why the heatmap isn't shown?

Yes. It generates the usual "not populated with data".

maliberty:
> It generates the usual "not populated with data". Should it be more specific?

I think so as it won't be obvious why it isn't populated (which normally happens on demand rather than through an explicit step).

gudeh:
We get the following when trying to activate rudy's heatmap after a `make gui_floorplan`
Screenshot from 2024-04-26 11-34-18

maliberty (approved):

PR #5001, opened 2024-04-26
staGui: End column is now visible

oharboe:
there's no great way in C++ to get the maximum value/count for an enum, but at least it's no longer hardcoded if new columns are added.

The "End" column was hidden since there was a gaffe when adding End and not updating the hardcoded count of the enum everywhere:

image

### DIFF
--- a/src/gui/src/staGui.cpp
+++ b/src/gui/src/staGui.cpp
@@ -110,7 +110,7 @@ int TimingPathsModel::rowCount(const QModelIndex& parent) const
 
 int TimingPathsModel::columnCount(const QModelIndex& parent) const
 {
-  return 6;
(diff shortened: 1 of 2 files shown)


oharboe:
@maliberty Possible future refinement, use [QMetaEnum](https://doc.qt.io/qt-6/qmetaenum.html), since there is a dependency on Qt anyway?

QMetaEnum might help reduce brittle unqiue solutions within each module to deal with enums and make the code easier to read as there is a lot of implicit documentation in using QMetaEnum to e.g. translate from enum to string, get count of enums and such.

maliberty (approved):

maliberty:
Q_ENUM could be used (though only in gui).

PR #5006, opened 2024-04-26
DRT skip adding access points for db pins with no net

arthurjolo:
Fixes [#5005](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5005)

This PR avoids creating access points for db pins that aren't connected to any net. This is to prevent errors from happening to pins that are not connected to any net.


arthurjolo:
This PR has been throw a secure-CI and everything is passing

maliberty (approved):

PR #5017, opened 2024-04-29
mpl2: decrease connection signature weight threshold

AcKoucher:
When running Secure-CI for #4991 asap7/aes-block showed some metrics degradation. After some digging it looked like the problem wasn't specifically related to that PR itself, but rather there was something weird going on with the clustering.

Actually, asap7/aes-block shows a very narrow case where the minimum weight threshold for MPL2 to consider two clusters connected was impacting on how the clusters were formed.

These changes are to ensure that, for smaller designs, we won't end up splitting macros unconveniently.


AcKoucher:
Running Secure-ci

PR #5022, opened 2024-04-30
ord: add build type to splash if other than release

gadfort:
No function changes expected to the "Release" builds.

Adds:
- indicator to splash to indicate if the build is something other than Release.

Reasoning:
- in the past I've been trying to determine runtime issues and it's frequently because the build type is Debug and I forgot to switch it back to Release.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -491,11 +491,13 @@ static void showSplash()
                  ord::OpenRoad::getGitDescribe());
   logger->report(
       "Features included (+) or not (-): "
-      "{}Charts {}GPU {}GUI {}Python",
+      "{}Charts {}GPU {}GUI {}Python{}",
       ord::OpenRoad::getChartsCompileOption() ? "+" : "-",
       ord::OpenRoad::getGPUCompileOption() ? "+" : "-",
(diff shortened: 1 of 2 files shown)


maliberty (approved):

src/Main.cc
```diff
@@ -491,11 +491,13 @@ static void showSplash()
...
   logger->report(
       "Features included (+) or not (-): "
-      "{}Charts {}GPU {}GUI {}Python",
+      "{}Charts {}GPU {}GUI {}Python{}",
       ord::OpenRoad::getChartsCompileOption() ? "+" : "-",
       ord::OpenRoad::getGPUCompileOption() ? "+" : "-",
       ord::OpenRoad::getGUICompileOption() ? "+" : "-",
-      ord::OpenRoad::getPythonCompileOption() ? "+" : "-");
+      ord::OpenRoad::getPythonCompileOption() ? "+" : "-",
+      strcmp(BUILD_TYPE, "Release") == 0 ? ""
```
oharboe, line 499: following the "+" / "-" convention above, absent features are not represented by a blank string, because then it is hard to tell what is missing.

Similarly, using a blank string for release, makes it hard to tell that there should have been a check here.

PR #5028, opened 2024-05-02
rmp: Adds support to create an abc standard cell library from opensta

QuantamHD:
Creates a factory class to map an sta library into an abc one. This library can be used to unmap a netlist in ABC into an AIG for optimization, and then back into a mapped netlist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,6 +140,7 @@ configure_file(
 
 ################################################################
 
+link_libraries(stdc++fs)
 # Ask CMake to output a compile_commands.json file for use with things like clang-tidy
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
--- a/src/dbSta/src/MakeDbSta.cc
+++ b/src/dbSta/src/MakeDbSta.cc
@@ -29,11 +29,6 @@ namespace ord {
 
 using sta::dbSta;
 
-dbSta* makeDbSta()
-{
-  return new dbSta;
-}
-
 void deleteDbSta(sta::dbSta* sta)
 {
   delete sta;
--- a/src/rmp/src/CMakeLists.txt
+++ b/src/rmp/src/CMakeLists.txt
@@ -80,12 +80,14 @@ add_library(rmp_abc_library
 target_include_directories(rmp_abc_library
   PUBLIC
     ../include
+  PRIVATE
     .
 )
 
 target_link_libraries(rmp_abc_library
   PUBLIC
     OpenSTA
+    dbSta_lib
     libabc
     utl_lib
 )
--- a/src/rmp/src/CMakeLists.txt
+++ b/src/rmp/src/CMakeLists.txt
@@ -73,6 +73,23 @@ target_link_libraries(rmp
     ${ABC_LIBRARY}
  )
 
+add_library(rmp_abc_library 
+  abc_library_factory.cpp
+)
+
+target_include_directories(rmp_abc_library
+  PUBLIC
+    ../include
+    .
+)
+
+target_link_libraries(rmp_abc_library
+  PUBLIC
+    OpenSTA
+    libabc
+    utl_lib
+)
+
 if (Python3_FOUND AND BUILD_PYTHON)
   swig_lib(NAME          rmp_py
            NAMESPACE     rmp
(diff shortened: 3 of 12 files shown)


src/rmp/src/abc_library_factory.cpp
```diff
@@ -0,0 +1,468 @@
...
+      continue;
+    }
+
+    if (HasNonInputOutputPorts(cell)) {
+      continue;
+    }
+
+    if (HasOnlyPowerAndGroundPorts(cell)) {
+      continue;
+    }
```
maliberty, line 379: Would it be simpler to just have IsAbcCompatible and traverse the pins once?

QuantamHD: I feel like that will end up being less clear overall since you need to traverse all the pins for some of the checks so you'll end up with a while loop with a bunch of counters that you then need to check. I feel like these checks could be moved into IsAbcCompatible, but I'd rather keep them seperate.

src/rmp/src/abc_library_factory.cpp
```diff
@@ -0,0 +1,468 @@
...
+#include "sta/PortDirection.hh"
+#include "sta/Sta.hh"
+#include "sta/TableModel.hh"
+#include "sta/TimingArc.hh"
+#include "sta/Units.hh"
+#include "utl/deleter.h"
+
+namespace rmp {
+
+static bool IsCombinational(sta::LibertyCell* cell)
```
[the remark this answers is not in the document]
QuantamHD: No that function deals with insts not masters which is what we have here. The function looks at net source type which is not meaningful in this context. If you create a version that worked on masters I could use. I transitioned everything over to dbSta so it wouldn't be hard to change.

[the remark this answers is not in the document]
maliberty: It makes finer distinction that I guess you care about that do require an instance.  If you stick with this scheme you should consider isLevelShifter and isIsolationCell.

src/rmp/src/abc_library_factory.cpp
maliberty: If you have an output, how can you have only power and ground ports?

QuantamHD: They are considered inout ports

src/rmp/src/abc_library_factory.cpp
```diff
@@ -0,0 +1,468 @@
...
+// https://developers.google.com/open-source/licenses/bsd
+
+#include "abc_library_factory.h"
+
+#include <cmath>
+#include <unordered_set>
+#include <vector>
+
+// Poor include definitions in ABC
+// clang-format off
```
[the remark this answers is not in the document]
QuantamHD: Yes.

QuantamHD:
Needs https://github.com/berkeley-abc/abc/pull/294 to be merged here and then in the openroad copy

src/rmp/src/abc_library_factory.cpp
maliberty: inout are not outputs unless you mean to use isAnyOutput() not isOutput()

maliberty: Also they are not considered inout either but have their own categories.

QuantamHD:
@maliberty Can you merge this

src/rmp/src/CMakeLists.txt
```diff
@@ -73,6 +73,23 @@ target_link_libraries(rmp
...
  )
 
+add_library(rmp_abc_library 
+  abc_library_factory.cpp
+)
+
+target_include_directories(rmp_abc_library
+  PUBLIC
+    ../include
+    .
```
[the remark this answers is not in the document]
QuantamHD: Done.

src/rmp/src/abc_library_factory.cpp
QuantamHD: You're right. I removed it.

src/dbSta/src/MakeDbSta.cc
```diff
@@ -29,11 +29,6 @@ namespace ord {
 
 using sta::dbSta;
 
-dbSta* makeDbSta()
```
maliberty, line 32: Why is this moved?

QuantamHD: In order to construct this from tests the linker needs to see this function in the lib version of dbsta instead of the binary version. 

The new operator itself isn't callable directly because of the abstract gui class pointers.

CMakeLists.txt
```diff
@@ -140,6 +140,7 @@ configure_file(
 
 ################################################################
 
+link_libraries(stdc++fs)
```
maliberty, line 143: I tried this previously but had lots of trouble on mac.  Hopefully this will work ok a few years later.

maliberty (approved):

CMakeLists.txt
QuantamHD: C++ makes me sad

PR #5031, opened 2024-05-02
Banner cleanup

oharboe:
No conditional code and always print build type.

Before:

```
Features included (+) or not (-): +Charts +GPU +GUI +Python : RELEASE
```

After:

```
Features included (+) or not (-): +Charts +GPU +GUI +Python RELEASE
```

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -491,13 +491,12 @@ static void showSplash()
                  ord::OpenRoad::getGitDescribe());
   logger->report(
       "Features included (+) or not (-): "
-      "{}Charts {}GPU {}GUI {}Python{}",
+      "{}Charts {}GPU {}GUI {}Python {}",
       ord::OpenRoad::getChartsCompileOption() ? "+" : "-",
       ord::OpenRoad::getGPUCompileOption() ? "+" : "-",
       ord::OpenRoad::getGUICompileOption() ? "+" : "-",
       ord::OpenRoad::getPythonCompileOption() ? "+" : "-",
-      strcmp(BUILD_TYPE, "Release") == 0 ? ""
-                                         : fmt::format(" : {}", BUILD_TYPE));
+      BUILD_TYPE);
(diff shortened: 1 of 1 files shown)


src/Main.cc
```diff
@@ -491,13 +491,12 @@ static void showSplash()
...
   logger->report(
       "Features included (+) or not (-): "
-      "{}Charts {}GPU {}GUI {}Python{}",
+      "{}Charts {}GPU {}GUI {}Python {}",
       ord::OpenRoad::getChartsCompileOption() ? "+" : "-",
       ord::OpenRoad::getGPUCompileOption() ? "+" : "-",
       ord::OpenRoad::getGUICompileOption() ? "+" : "-",
       ord::OpenRoad::getPythonCompileOption() ? "+" : "-",
-      strcmp(BUILD_TYPE, "Release") == 0 ? ""
-                                         : fmt::format(" : {}", BUILD_TYPE));
```
oharboe, line 500: Printing out the buid type unconditionally is congruent with always printing out the feature above, even if it is not compiled in.

I wonder what the strcmp() was doing. In my case BUILD_TYPE == "RELEASE", which doesn't match the strcmp() above.

maliberty:
I am worried that the word "release" has different meanings.  Someone might think this implies it is a formal release rather than a compilation mode.  I would like to keep the suppression of that particular string.  You could use strcasecmp if that helps.

oharboe:
> I am worried that the word "release" has different meanings. Someone might think this implies it is a formal release rather than a compilation mode. I would like to keep the suppression of that particular string. You could use strcasecmp if that helps.

Hmm... I think this solves one problem and creates another by adding complexity. Master is currently broken and the behavior is actually what you didn't want...

I'm going to close this PR, because I think it is probably the best way to go about this, all told, but if something else is wanted, that's more a question of policy than coding...

PR #5036, opened 2024-05-02
par & mpl2: get inst area from .lef instead of .lib

AcKoucher:
When debugging using sky130hd/uW, SoftSA overlap penalty was resulting in -nan. The reason was that the APIs from `Cluster` were returning 0.0 area for the macro area of clusters with macros. The problem doesn't look like it's mpl2 itself, but rather the fact that `LibertyCell::area()` is returning 0.0 for the macros in this design during mpl2 flow.


AcKoucher:
running secure-ci

maliberty (approved):

maliberty:
merge pending ci results

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -704,33 +708,39 @@ Metrics* HierRTLMP::computeMetrics(odb::dbModule* module)
...
 {
-  unsigned int num_std_cell = 0;
-  unsigned int num_macro = 0;
-  float std_cell_area = 0.0;
-  float macro_area = 0.0;
-  num_std_cell += cluster->getLeafStdCells().size();
-  num_macro += cluster->getLeafMacros().size();
-  for (odb::dbInst* inst : cluster->getLeafStdCells()) {
-    const sta::LibertyCell* liberty_cell = network_->libertyCell(inst);
-    std_cell_area += liberty_cell->area();
```
oharboe, line 727: @maliberty @AcKoucher When I fix a bug, I try to remember to ask myself if this bug exists elsewhere too: there are 25 other places `->area()` is used in OpenROAD, should any of the other be reconsidered?

maliberty: I considered it.  Note that not all ->area() are calls in sta as there are other methods with the same name that do not rely on sta.

The problem shows up in uW as the rams were generated a long time ago when we didn't write the area into Liberty abstracts.  That was fixed a long time ago but we don't have a good way to regenerate these.  It is isn't a problem on std cell libraries where most of these calls are relevant.

There are two uses in par that are worth looking at @AcKoucher

rovinski: If the only issue is the area property on those .libs, can't they just be added manually?

oharboe: How about a warning when .lib is broken, delete the broken fakrams and replace them with mock srams that are more easily maintained?

AcKoucher:
CI showed a large increasing in drt WL for sky130hd/uW:
`[ERROR] detailedroute__route__wirelength fail test: 9237052.0 <= 8799572.0`

I'll fix the PAR .lib area calls and test it again locally, as it can have some meaningful impact in mpl2 clustering.

maliberty:
They are not fake rams, they are real dfframs that were used in a tapeout.  I think it is generally better to rely on the LEF in any case as Liberty doesn't even require the area field to be present.

maliberty:
Yes the par change would need another secure CI

maliberty:
@AcKoucher how does the uW macro placement change with this PR?

AcKoucher:
Running CI again now that RUDY was merged.

PR #5055, opened 2024-05-06
Adding Timing::makeEquivCells and Timing::equivCells Python APIs

bingyuew:
adding void Timing::makeEquivCells() and std::vector<odb::dbMaster*> Timing::equivCells(odb::dbMaster* master) Python APIs in the src/Timing.cc file

### DIFF
--- a/src/Timing.cc
+++ b/src/Timing.cc
@@ -406,7 +406,6 @@ void Timing::makeEquivCells()
   sta->makeEquivCells(&libs, nullptr);
 }
 
-//std::vector<odb::dbMaster*> Timing::equivCells(odb::dbMaster* master)
 std::vector<std::string> Timing::equivCells(odb::dbMaster* master)
 {
   sta::dbSta* sta = getSta();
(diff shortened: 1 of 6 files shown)


maliberty:
You have missed the DCO.  Please see the directions on the Details like to fix that.

maliberty:
You need to run clang-format as well

bingyuew:
Fix the clang-format issue and add sign-offs to each commit.

src/Timing.cc
```diff
@@ -385,4 +385,44 @@ float Timing::dynamicPower(odb::dbInst* inst, sta::Corner* corner)
   return (power.internal() + power.switching());
 }
 
+void Timing::makeEquivCells()
+{
+  sta::Network* network = cmdLinkedNetwork();
+  sta::LibertyLibrarySeq libs;
+  sta::LibertyLibraryIterator* lib_iter = network->libertyLibraryIterator();
```
QuantamHD, line 392: using std::unique_ptr on the iterators makes the deletes much safer. That way you don't have to manually manage the delete.

`std::unique_ptr<sta::LibertyLibraryIterator> lib_iter = std::unique_ptr(network->libertyLibraryIterator()); `

QuantamHD (changes requested):

src/Timing.cc
```diff
@@ -385,4 +385,44 @@ float Timing::dynamicPower(odb::dbInst* inst, sta::Corner* corner)
   return (power.internal() + power.switching());
 }
 
+void Timing::makeEquivCells()
```
maliberty, line 388: Rather than copying this code from Resizer I think it would be better to just make it a public method there and call it here.

src/Timing.cc
bingyuew: fix that!

src/Timing.cc
bingyuew: I think makeEquivCells and equivCells should be placed together, either in Timing.cc or in Resizer.cc, However, since equivCells is from sta, so I put both of them in Timing.cc.

PR #5064, opened 2024-05-08
misc: use micron and dbu converters from odb

eder-matheus:


src/grt/src/GlobalRouter.cpp
```diff
@@ -3907,7 +3907,7 @@ void GlobalRouter::reportLayerWireLengths()
         odb::dbTechLayer* layer = routing_layers_[i];
         logger_->report("{:5s} {:8d}um {:3d}%",
                         layer->getName(),
-                        static_cast<int64_t>(dbuToMicrons(length)),
+                        static_cast<int64_t>(block_->dbuToMicrons(length)),
```
rovinski, line 3910:
```suggestion
                        block_->dbuToMicrons(length),
```

rovinski (commented):
It looks to me like the antenna checker needs a wholesale conversion to DBU, because I don't see why it's using doubles right now.

maliberty:
> It looks to me like the antenna checker needs a wholesale conversion to DBU, because I don't see why it's using doubles right now.

@luis201420 is in the process of rewriting it so hopefully this will be better.

maliberty (approved):
I agree with @rovinski 's comment but otherwise fine.

PR #5068, opened 2024-05-08
Gain-based (ABC-style) buffering

povik:
This branch implements a buffering algorithm similar to ABC's `buffer` command.

### DIFF
--- a/src/rsz/src/RepairDesign.cc
+++ b/src/rsz/src/RepairDesign.cc
@@ -159,13 +159,15 @@ void RepairDesign::repairDesign(
   resizer_->resized_multi_output_insts_.clear();
 
   if (gain_buffering) {
-    // If gain-based buffering is enabled, we need to override slews in order to get
-    // good required time estimates.
+    // If gain-based buffering is enabled, we need to override slews in order to
+    // get good required time estimates.
     for (int i = resizer_->level_drvr_vertices_.size() - 1; i >= 0; i--) {
       Vertex* drvr = resizer_->level_drvr_vertices_[i];
       for (auto rf : {RiseFall::rise(), RiseFall::fall()}) {
-        sta_->setAnnotatedSlew(drvr, resizer_->tgt_slew_corner_,
-                               sta::MinMaxAll::all(), rf->asRiseFallBoth(),
+        sta_->setAnnotatedSlew(drvr,
+                               resizer_->tgt_slew_corner_,
+                               sta::MinMaxAll::all(),
+                               rf->asRiseFallBoth(),
                                resizer_->tgt_slews_[rf->index()]);
       }
     }
@@ -232,7 +234,8 @@ void RepairDesign::repairDesign(
     if (gain_buffering) {
       for (auto mm : sta::MinMaxAll::all()->range()) {
         for (auto rf : sta::RiseFallBoth::riseFall()->range()) {
-          const DcalcAnalysisPt *dcalc_ap = resizer_->tgt_slew_corner_->findDcalcAnalysisPt(mm);
+          const DcalcAnalysisPt* dcalc_ap
+              = resizer_->tgt_slew_corner_->findDcalcAnalysisPt(mm);
           drvr->setSlewAnnotated(false, rf, dcalc_ap->index());
         }
       }
@@ -418,7 +421,7 @@ bool RepairDesign::getCin(const Pin* drvr_pin, float& cin)
   return false;
 }
 
-static float bufferCin(const LibertyCell *cell)
+static float bufferCin(const LibertyCell* cell)
 {
   LibertyPort *a, *y;
   cell->bufferPorts(a, y);
@@ -429,23 +432,25 @@ void RepairDesign::findBufferSizes()
 {
   buffer_sizes_.clear();
   buffer_sizes_ = resizer_->buffer_cells_;
-  std::sort(buffer_sizes_.begin(), buffer_sizes_.end(),
-            [=](LibertyCell *a, LibertyCell *b) {
-    return bufferCin(a) < bufferCin(b);
-  });
+  std::sort(buffer_sizes_.begin(),
+            buffer_sizes_.end(),
+            [=](LibertyCell* a, LibertyCell* b) {
+              return bufferCin(a) < bufferCin(b);
+            });
 }
 
 void RepairDesign::performGainBuffering(Net* net,
-                                        const Pin *drvr_pin,
+                                        const Pin* drvr_pin,
                                         int max_fanout)
 {
-  struct EnqueuedPin {
-    Pin *pin;
+  struct EnqueuedPin
+  {
+    Pin* pin;
     PathRef required_path;
     Delay required_delay;
     int level;
 
-    Required required(const StaState *sta) const
+    Required required(const StaState* sta) const
     {
       if (required_path.isNull()) {
         return INF;
@@ -453,12 +458,12 @@ void RepairDesign::performGainBuffering(Net* net,
       return required_path.required(sta) - required_delay;
     }
 
-    std::pair<Required, int> sort_label(const StaState *sta) const
+    std::pair<Required, int> sort_label(const StaState* sta) const
     {
       return std::make_tuple(required(sta), -level);
     }
 
-    float capacitance(const Network *network)
+    float capacitance(const Network* network)
     {
       return network->libertyPort(pin)->capacitance();
     }
@@ -466,21 +471,23 @@ void RepairDesign::performGainBuffering(Net* net,
 
   class PinRequiredHigher
   {
-  private:
-    const Network *network_;
+   private:
+    const Network* network_;
 
-  public:
-    PinRequiredHigher(const Network *network) : network_(network) {};
+   public:
+    PinRequiredHigher(const Network* network) : network_(network){};
 
-    bool operator()(const EnqueuedPin &a, const EnqueuedPin &b) const {
+    bool operator()(const EnqueuedPin& a, const EnqueuedPin& b) const
+    {
       auto la = a.sort_label(network_), lb = b.sort_label(network_);
-      if (la > lb)
+      if (la > lb) {
         return true;
-      else if (la < lb)
+      }
+      if (la < lb) {
         return false;
-      else
-        return sta::stringLess(network_->pathName(a.pin),
-                               network_->pathName(b.pin));
+      }
+      return sta::stringLess(network_->pathName(a.pin),
+                             network_->pathName(b.pin));
     }
   };
 
@@ -490,28 +497,24 @@ void RepairDesign::performGainBuffering(Net* net,
   NetConnectedPinIterator* pin_iter = network_->connectedPinIterator(net);
   while (pin_iter->hasNext()) {
     const Pin* pin = pin_iter->next();
-    if (pin != drvr_pin && !network_->isTopLevelPort(pin) &&
-        network_->direction(pin) == sta::PortDirection::input() && network_->libertyPort(pin)) {
+    if (pin != drvr_pin && !network_->isTopLevelPort(pin)
+        && network_->direction(pin) == sta::PortDirection::input()
+        && network_->libertyPort(pin)) {
       Vertex* vertex = graph_->pinLoadVertex(pin);
       PathRef req_path = sta_->vertexWorstSlackPath(vertex, sta::MinMax::max());
-      sinks.push_back({
-        const_cast<Pin*>(pin),
-        req_path,
-        0.0,
-        0
-      });
+      sinks.push_back({const_cast<Pin*>(pin), req_path, 0.0, 0});
     }
   }
   delete pin_iter;
 
   // Keep track of the vertices at the boundary of the tree so we know where
   // to ask for delays to be recomputed
-  std::vector<Vertex *> tree_boundary;
+  std::vector<Vertex*> tree_boundary;
 
   float cin;
   if (getCin(drvr_pin, cin)) {
     float load = 0.0;
-    for (auto &sink : sinks) {
+    for (auto& sink : sinks) {
       load += sink.capacitance(network_);
     }
     std::sort(sinks.begin(), sinks.end(), PinRequiredHigher(network_));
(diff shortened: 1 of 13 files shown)


src/rsz/src/Resizer.cc
```diff
@@ -264,7 +266,70 @@ void Resizer::removeBuffers()
...
+
+  int remove_count = 0;
+  for (dbInst *db_inst : block_->getInsts()) {
+    LibertyCell *lib_cell = db_network_->libertyCell(db_inst);
+    Instance *cell = db_network_->dbToSta(db_inst);
+    if (!db_inst->isDoNotTouch()
+        && !db_inst->isFixed()
+        && lib_cell) {
+      // TODO: don't use tgt_slew_dcalc_ap_
+      LibertyCell *equiv = findSmallestEquiv(tgt_slew_dcalc_ap_, lib_cell);
```
maliberty, line 321: You are finding the weakest drive strength rather than the small area.  Likely they are the same cell but its a bit unclear from the naming?

src/rsz/src/Resizer.cc
```diff
@@ -264,7 +266,70 @@ void Resizer::removeBuffers()
...
+          ArcDelay intrinsic1 = port1->intrinsicDelay(this);
+          ArcDelay intrinsic2 = port2->intrinsicDelay(this);
+          return (std::tie(drive1, intrinsic2) < std::tie(drive2, intrinsic1));
+       });
+
+  return equiv_cells->back();
+}
+
+void
+Resizer::downsizeAllCells()
```
maliberty, line 305: What is the use model for this?

src/rsz/src/SynthesizeBuffers.hh
```diff
@@ -0,0 +1,33 @@
...
+using sta::Pin;
+using sta::dbSta;
+using sta::dbNetwork;
+
+class SynthesizeBuffers : StaState
+{
+public:
+  explicit SynthesizeBuffers(Resizer *resizer);
+  ~SynthesizeBuffers();
+  bool getCin(Pin *drvr, float &cin);
```
maliberty, line 23: private?

src/rsz/src/Resizer.cc
povik: Buffering interacts with sizing so it's a potential debugging aid. Also we can test if for some designs, e.g. with multiple clock domains, it might be better to start with a clean slate with regards to sizing since ABC wasn't aware of clock domains.

I wrote this command first because it sounded like a simple thing I could finish end-to-end and familiarize myself with the STA/OR interfaces on it.

src/rsz/src/SynthesizeBuffers.cc
```diff
@@ -0,0 +1,369 @@
...
+      }
+
+      Pin *new_input_pin = network_->findPin(inst, size_in);
+      Pin *new_output_pin = network_->findPin(inst, size_in);
+
+      graph_delay_calc_->findDelays(graph_->pinDrvrVertex(new_output_pin));
+      arrival_visitor->visit(graph_->pinLoadVertex(new_input_pin));
+      arrival_visitor->visit(graph_->pinLoadVertex(new_output_pin));
+      req_visitor->visit(graph_->pinLoadVertex(new_output_pin));
+      req_visitor->visit(graph_->pinLoadVertex(new_input_pin));
```
povik, line 246: STA calls here

src/rsz/src/SynthesizeBuffers.cc
```diff
@@ -0,0 +1,369 @@
...
+  {
+    VertexPathIterator path_iter(graph->pinLoadVertex(pin), nullptr,
+                                 nullptr, MinMax::max(), sta);
+    Required required = MinMax::min()->initValue();
+    while (path_iter.hasNext()) {
+      Required path_required = path_iter.next()->required(sta);
+      if (delayGreater(path_required, required, MinMax::min(), sta)) {
+        required = path_required;
+      }
+    }
```
povik, line 110: Here's where we get the required time from STA structures

src/rsz/src/SynthesizeBuffers.cc
```diff
@@ -0,0 +1,369 @@
...
+    if (work_queue[i]) {
+      Pin* drvr_pin = vertex->pin();
+      Net* net = network_->isTopLevelPort(drvr_pin)
+                    ? network_->net(network_->term(drvr_pin))
+                    : network_->net(drvr_pin);
+      synthesizeBuffers(net, drvr_pin, max_fanout, gain);
+    }
+
+    // as we go and repair the nets also backpropagate the required times
+    req_visitor->visit(vertex);
```
povik, line 360: STA call here

src/rsz/src/Resizer.cc
```diff
@@ -263,7 +255,69 @@ void Resizer::removeBuffers()
   logger_->info(RSZ, 26, "Removed {} buffers.", remove_count);
 }
 
-bool Resizer::bufferBetweenPorts(Instance* buffer)
+LibertyCell* Resizer::findSmallestEquiv(const DcalcAnalysisPt* dcalc_ap,
```
maliberty, line 258: Strictly speaking this is the weakest not the smallest though those are generally correlated.  If you truly want the smallest you can take the area from LEF (odb) and it will work on multi-output gates as well.  Liberty may also have area but it is optional

src/rsz/src/Resizer.i
```diff
@@ -745,6 +753,14 @@ set_worst_slack_nets_percent(float percent)
   resizer->setWorstSlackNetsPercent(percent);
 }
 
+void
+synthesize_buffers_cmd(int max_fanout, float gain, float slew)
```
maliberty, line 757: How would a user choose gain and slew?

povik: That's TBD. Once we confirm this algorithm works I would like for us to experiment a bit with other algorithms which may need less user input

maliberty:
@povik what is the status on this work?

povik:
@maliberty Experiments with running a re-implementation of naive ABC-style buffering at the time of `repair_design` gave bad results. Because the buffer tree structure is generated with no regard for placement, there's a lot of extra wiring which worsens both the signal delay and routing congestion (I would say this was to be expected). The crucial difference to the earlier experiments with keeping ABC buffers is that here we buffer post placement, and it looks like the extra wiring is an issue no matter how we place the newly inserted buffers. I looked at some options and have more to say tomorrow.

The branch has the ABC-style buffering as the `synthesize_buffers` command. There's no work to integrate it into `repair_design`.

maliberty:
> It will take `max_fanout` `gain` `slew` for its parameters.

Why will max_fanout need to be a parameter?  You should take that from sta as the rest of rsz does.  What is the slew parameter for?

povik:
> Why will max_fanout need to be a parameter? You should take that from sta as the rest of rsz does.

On second thought I agree. The algorithm works without a max_fanout limit too.

> What is the slew parameter for?

That's the slew override on unrepaired nets. There's `Resizer::targetSlew()`, should we make it the default for that parameter?

maliberty:
> That's the slew override on unrepaired nets. There's Resizer::targetSlew(), should we make it the default for that parameter?

That seems reasonable.  I'm wondering if this can be reduced to -gain on the existing repair_design command.

povik:
Is `repair_design` on unplaced designs supported? In any case I would make it `-buffering_gain` since gain can be used for resizing but that's not what we do.

maliberty:
> Is `repair_design` on unplaced designs supported? 

Only in the sense that it would be a no-op as it will skip any net with unplaced instances.  It would be now with your option.  

Do you feel it is different enough to warrant a separate command?

povik:
> Only in the sense that it would be a no-op as it will skip any net with unplaced instances. It would be now with your option.
> 
> Do you feel it is different enough to warrant a separate command?

Given the above, yes. To the user it's weird that repair_design would buffer but not resize on an unplaced design. If we cannot make it do resizing, then I would argue for keeping the gain buffering separate.

povik:
I am looking at the code
```
    // Resize the driver to normalize slews before repairing limit violations.
    if (parasitics_src_ == ParasiticsSrc::placement && resize_drvr) {
      resize_count_ += resizer_->resizeToTargetSlew(drvr_pin);
    }
```
maybe making it do resizing when run pre placement would only be a question of removing that part of the conditional?

maliberty:
No you need a Steiner tree to drive the code and there is none without placement.

povik:
I don’t think that’s true of the resizing bit (looking at the insides of resizeToTargetSlew)

src/rsz/src/RepairDesign.cc
```diff
@@ -153,9 +158,25 @@ void RepairDesign::repairDesign(
   resize_count_ = 0;
   resizer_->resized_multi_output_insts_.clear();
 
+  if (gain_buffering) {
+    // If gain-based buffering is enabled, we need to override slews in order to get
+    // good required time estimates.
+    for (int i = resizer_->level_drvr_vertices_.size() - 1; i >= 0; i--) {
+      Vertex* drvr = resizer_->level_drvr_vertices_[i];
+      for (auto rf : {RiseFall::rise(), RiseFall::fall()}) {
+        sta_->setAnnotatedSlew(drvr, resizer_->tgt_slew_corner_,
```
maliberty, line 167: Any true user annotated slews will be lost.

src/rsz/src/RepairDesign.cc
```diff
@@ -383,6 +624,21 @@ void RepairDesign::repairNet(Net* net,
                sdc_network_->pathName(drvr_pin));
     const Corner* corner = sta_->cmdCorner();
     bool repaired_net = false;
+
+    if (buffer_gain_ != 0.0) {
+      float fanout, max_fanout, fanout_slack;
+      // TODO: is it correct to unconditionally use max_fanout? what's fanout_slack?
+      sta_->checkFanout(drvr_pin, max_, fanout, max_fanout, fanout_slack);
+
+      int resized = resizer_->resizeToTargetSlew(drvr_pin);
```
maliberty, line 633: Is this needed?  abc won't do this.

src/rsz/src/RepairDesign.cc
```diff
@@ -383,6 +624,21 @@ void RepairDesign::repairNet(Net* net,
...
+      performGainBuffering(net, drvr_pin, max_fanout);
+      // Resize again post buffering as the load changed
+      resized += resizer_->resizeToTargetSlew(drvr_pin);
```
maliberty, line 636: same

src/rsz/src/RepairDesign.cc
povik: It’s not needed, it should be better if we do it than if we don’t (we can save on buffers if we can resize instead)

maliberty: The gain based buffering should be doing the sizing rather than relying on the target slew resizing (it tends to oversize and we are working to reduce/eliminate it).

src/rsz/src/RepairDesign.cc
povik: That’s an awkward one. I was hoping people don’t use annotations in this context. So should we annotate if there isn’t an annotation already, later revert only those annotations we added?

src/rsz/src/RepairDesign.cc
maliberty: I thought they were a single algorithm.  In that case I guess you need to leave it until we have a better solution for sizing.  How is it in the current yosys/abc flow in terms of sizing and buffering?  Are they both run? In which order?

povik: > How is it in the current yosys/abc flow in terms of sizing and buffering? Are they both run? In which order?

Buffering then sizing

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/scripts/abc_speed.script

maliberty: So this would be sizing then buffering?

povik: Sizing, buffering, then sizing again, but on each driver at a time. So we don't do a separate pass of buffering and a separate pass of sizing where the buffering sees unsized loads or the other way around

povik: > Sizing, buffering, then sizing again

Now that I think of it there may be an interaction between buffering and sizing which explains why the combined process isn't idempotent, which I considered an outstanding issue with the code

maliberty:
At least one unit test is needed in this PR.

povik:
OK, I assume a Tcl one is good enough

src/rsz/src/Resizer.tcl
```diff
@@ -437,19 +437,24 @@ sta::define_cmd_args "repair_design" {[-max_wire_length max_wire_length] \
 
 proc repair_design { args } {
   sta::parse_key_args "repair_design" args \
-    keys {-max_wire_length -max_utilization -slew_margin -cap_margin} \
+    keys {-max_wire_length -max_utilization -slew_margin -cap_margin -buffer_gain} \
```
maliberty, line 440: -buffer_gain is missing from the define_cmd_args above and the README.

maliberty:
> OK, I assume a Tcl one is good enough

Yes

povik:
I have rebased, added a test, and handled don't-care sinks (they are ignored for buffering but warned about)

povik:
The CI failure looks like something unrelated

```
OpenROAD
error1 (tcl) pass
get_core_die_areas (tcl) pass
timing_api (py) pass
timing_api_2 (py) *FAIL*
timing_api_3 (py) pass
upf_test (tcl) pass
upf_aes (tcl) pass
two_designs (py) pass
commands_without_load (tcl) pass
------------------------------------------------------
Failed 1/9
See /tmp/workspace/OpenROAD-Public_PR-5068-head/test/results/diffs for differences
See /tmp/workspace/OpenROAD-Public_PR-5068-head/test/results for log files
```

maliberty:
The unit test is unrelated and occurs in multiple PRs but not for me locally.

povik:
> handled don't-care sinks (they are ignored for buffering but warned about)

Err, don't-touch sinks

src/rsz/src/RepairDesign.cc
```diff
@@ -359,6 +406,243 @@ void RepairDesign::repairNet(Net* net,
...
+    float load = 0.0;
+    for (auto& sink : sinks) {
+      load += sink.capacitance(network_);
+    }
+    std::sort(sinks.begin(), sinks.end(), PinRequiredHigher(network_));
+
+    // Iterate until we satisfy both the gain condition and max_fanout
+    // on drvr_pin
+    while (sinks.size() > max_fanout || load > cin * buffer_gain_) {
+      float max_load = bufferCin(buffer_sizes_.back()) * buffer_gain_;
```
maliberty, line 546: Can this be hoisted out of the loop / made const?

povik: Yes, I have pulled it out and renamed to `max_buf_load` (since it's the maximum load on a buffer, not on the driver pin at the root of the buffer tree)

PR #5072, opened 2024-05-08
gui: add logic depth, logic delay and implement column visibility in timing report

### LINKED ISSUE #4973: Add "Logic delay" column to Timing Report
### Description

It is useful to see the logic delay of a path in the Timing Report when doing triage in architectural exploration.

### Suggested Solution

Add "Logic delay" column to Timing Report

### Additional Context

_No response_

AcKoucher:
Resolve #4973 

Add to Timing Report:
- Column Visibility control
- Logic Delay
- Logic Depth 

*Also fix crash when right clicking timing report if no odb is loaded.

### DIFF
--- a/src/gui/src/staGui.cpp
+++ b/src/gui/src/staGui.cpp
@@ -125,6 +125,7 @@ QVariant TimingPathsModel::data(const QModelIndex& index, int role) const
         return Qt::AlignLeft;
       case Required:
       case Arrival:
+      case LogicDelay:
       case LogicDepth:
       case Slack:
       case Skew:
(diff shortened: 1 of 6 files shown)


AcKoucher:
@oharboe @maliberty  Here's how the visibility control looks like:

image

Would you check to see if this changes attend your needs?

oharboe:
Perfect!

oharboe:
Sweet!

image

maliberty:
I think it would be good to store the set of visible columns in the settings so it is preserved from run to run.  If someone doesn't want to see the new columns they shouldn't have to hide them each time.

maliberty:
The logic depth is off by a factor of two as we discussed.

oharboe:
@maliberty @AcKoucher Could you add a tooltip that defines exactly what logic depth and logic delay means?

I think it is more important that "logic depth" and "logic delay" is clearly defined than exactly what the definition is. For instance: does logic delay/depth exclude wires/buffers?

I would be more interested in separating out the "wire/buffer delay/depth" in separate columns; separate the concerns of routing/placement as much as possible from logic.

See also #5083 w.r.t. tooltip.

AcKoucher:
@maliberty @oharboe It's possible, but I'd like to not further overload this PR. How about a subsequent PR with the tool-tips?

For the depth: we exclude buffers and back-to-back inverters.

maliberty (approved):

PR #5081, opened 2024-05-09
drt: handle PDN repair for block vias

gadfort:
Changes:
- repair via will now "smash" a block via array into individual vias and then remove all violating vias. This avoids the issue where a single via in a block via array causes drt to remove the entire via array disconnecting the power grid.

### DIFF
--- a/src/odb/src/db/dbSBox.cpp
+++ b/src/odb/src/db/dbSBox.cpp
@@ -43,6 +43,7 @@
 #include "dbTechVia.h"
 #include "dbVia.h"
 #include "odb/db.h"
+#include "odb/dbShape.h"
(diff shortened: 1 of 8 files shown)


src/odb/src/db/dbSBox.cpp
```diff
@@ -459,4 +460,68 @@ void dbSBox::destroy(dbSBox* box_)
   block->_sbox_tbl->destroy(box);
 }
 
+std::vector<dbSBox*> dbSBox::smashVia()
```
maliberty, line 463: It would be good to have a test for this API

maliberty (approved):

PR #5089, opened 2024-05-11
Fix issue with linking dependency of -lstdc++fs when using flatpak

sebinho:
I had to fork OpenROAD as there is an issue during linking of `stdc++fs`. It seems to be an old practice to put this line and is an issue when building with Flatpak. 

I made a similar request for Surelog that has been accepted (maybe it can be done for OpenROAD as well):
https://github.com/chipsalliance/Surelog/pull/3966

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,7 +140,8 @@ configure_file(
 
 ################################################################
 
-if (NOT APPLE)
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "9.1")
+  MESSAGE(STATUS "Older version of GCC detected. Linking against stdc++fs")
   link_libraries(stdc++fs)
 endif()
 


sebinho:
Build seems to fail for very old distros, I see that CentOS7 failed. Probably need to add a flag to detect CentOS.

sebinho:
I will add a check to detect CentOS and update pull request

sebinho:
I added an attempt to detect Centos in `CMakelists.txt`. I cannot test it as I don't have Centos (I am using ArchLinux).
Should be visible in the Jenkins build though.

sebinho:
Looking at the logs in Jenkins, it seems to properly run and detect Centos.

QuantamHD:
I'm pretty sure the limitation here is on older versions of gcc < 9 so we should just base the check on that

maliberty:
There is no issue on centos8 so I think the compiler version is more relevant.

sebinho:
ok. So I changed the check from distribution to GCC version. I used version < 9.0 for linking against stdc++fs.
Is this good enough?

CMakeLists.txt
```diff
@@ -140,7 +152,8 @@ configure_file(
 
 ################################################################
 
-if (NOT APPLE)
+if (NOT APPLE AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "9.0")
```
maliberty, line 155: You also need to check CMAKE_CXX_COMPILER_ID STREQUAL "GNU"

maliberty:
Please fix the missing DCO

CMakeLists.txt
maliberty: Actually it should be 9.1 per the notes in https://en.cppreference.com/w/cpp/filesystem

sebinho: ok, I made the requested changes. Currently running on Jenkins. Should be good to go.
Thx for your replies

CMakeLists.txt
```diff
@@ -131,6 +131,18 @@ message(STATUS "C++ Standard: ${CMAKE_CXX_STANDARD}")
...
+  execute_process (
+          COMMAND bash -c "awk -F= '/^ID=/{print $2}' /etc/os-release |tr -d '\n' | tr -d '\"'"
+          OUTPUT_VARIABLE LinuxOS
+  )
+ELSE()
+  set(LinuxOS "n/a")
+ENDIF()
+
+MESSAGE(STATUS "Linux Distro: ${LinuxOS}")
+
```
gadfort, line 145: This isn't doing anything for the linking issue, might be better to remove?

sebinho: done, just removed it. I thought it could be interesting to get this info, but indeed no added value so I removed it

CMakeLists.txt
```diff
@@ -140,7 +140,8 @@ configure_file(
 
 ################################################################
 
-if (NOT APPLE)
+if (NOT APPLE AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "9.1")
```
maliberty, line 143: I don't think "NOT APPPLE" is needed.

sebinho: I cleaned up my commits (to only have a single commit) and removed the APPLE check.

maliberty:
DCO still needs to be addressed.

sebinho:
> DCO still needs to be addressed.

done

maliberty (approved):

maliberty:
Thanks for contributing.

sebinho:
> Thanks for contributing.

It was a pleasure!

PR #5108, opened 2024-05-15
Fix building `rmp_abc_library` when `USE_SYSTEM_ABC` is set

donn:
(Small change)

### DIFF
--- a/src/rmp/src/CMakeLists.txt
+++ b/src/rmp/src/CMakeLists.txt
@@ -90,6 +90,7 @@ target_link_libraries(rmp_abc_library
     dbSta_lib
     libabc
     utl_lib
+    ${ABC_LIBRARY}
 )
 
 if (Python3_FOUND AND BUILD_PYTHON)


maliberty:
I'm a bit confused by this.  USE_SYSTEM_ABC is not mentioned anywhere in the PR beyond the title.  What issue is being solved?

donn:
When `USE_SYSTEM_ABC` is set to true, `ABC_LIBRARY` is set to a non-null value. Things that require abc link against `ABC_LIBRARY` so ABC symbols won't be missing for those who set `USE_SYSTEM_ABC` to true.

`rmp_abc_library` was just missing `ABC_LIBRARY`.

maliberty:
rmp already links against it https://github.com/The-OpenROAD-Project/OpenROAD/blob/f0cc0ee1a62d7a6c35b43bd46a20980969eeb9a2/src/rmp/src/CMakeLists.txt#L73 - why does rmp_abc_library also need it (or should it just be moved)?

donn:
Because otherwise `rmp_abc_library` which is I believe compiled as a standalone unit will complain about being unable to find things in the `abc::` namespace.

Issue #3658, opened 2023-07-18
make write_lef and others write out files atomically


oharboe:
### Description

A half-written .lef file can be lying around after a crash which complicates e.g. `deltaDebug.py`.

Suggestion:

1. modify all `write_*` commands to write to a temp file
2. when the write completes, rename temp file to final name

This way there are no partial files lying around

annapetrosyan26:
Hello @maliberty @oharboe , I'd like to work on this issue. Could you please assign it to me ?

oharboe:
@annapetrosyan26 You can simply create a pull request without being assigned this task, I don't have the role to assign github issues in this project.

I haven't seen that @maliberty assigns github issues to new contributors, but perhaps that's something that would be approperiate?

annapetrosyan26:
Thank you @oharboe for the clarification. Then I'll work on the issue and once the code is ready I will create a pull request.

oharboe:
@maliberty Fixed, no?

maliberty:
Fixed in https://github.com/The-OpenROAD-Project/OpenROAD/pull/5109

PR #5109, opened 2024-05-15
make write_* commands to write out files atomically 

annapetrosyan26:
@maliberty  @oharboe  this PR addresses the #3658 issue. Could you please take a look?
The implementation is done according to the suggestion in the issue #3658 description. The changes have been locally tested.


src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +33,19 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
   }
 }
 
+std::string createTmpFileName(const char* filename)
```
maliberty, line 36: std::tmpnam or std::tmpfile should be used.

src/OpenRoad.cc
```diff
@@ -378,17 +384,21 @@ void OpenRoad::writeAbstractLef(const char* filename,
...
+
   std::ofstream os;
   os.exceptions(std::ofstream::badbit | std::ofstream::failbit);
-  os.open(filename);
+  os.open(tmp_filename);
   odb::lefout writer(logger_, os);
   writer.setBloatFactor(bloat_factor);
   writer.setBloatOccupiedLayers(bloat_occupied_layers);
   writer.writeAbstractLef(block);
+  fs::rename(tmp_filename, filename);
```
maliberty, line 396: os.close should happen before the rename to ensure all output has been written.

maliberty (changes requested):

src/OpenRoad.cc
```diff
@@ -355,15 +359,17 @@ static odb::defout::Version stringToDefVersion(const string& version)
...
   if (chip) {
     odb::dbBlock* block = chip->getBlock();
     if (block) {
       odb::defout def_writer(logger_);
       def_writer.setVersion(stringToDefVersion(version));
-      def_writer.writeBlock(block, filename);
+      def_writer.writeBlock(block, tmp_filename.c_str());
     }
   }
+  fs::rename(tmp_filename, filename);
```
oharboe, line 372: does this overwrite if the file exists?

oharboe: if the rename fails, the tmp file is left behind. intentional?

oharboe (changes requested):
some things to check...

QuantamHD:
This feels like a dangerous change to me. Not all users have a tmp folder, and using a time based file name is also a bit dangerous if you have multiple instances running on the same machine. Usually you would use a UUID for this purpose.

I would strongly advocate for making this an option.

oharboe:
> This feels like a dangerous change to me. Not all users have a tmp folder, and using a time based file name is also a bit dangerous if you have multiple instances running on the same machine. Usually you would use a UUID for this purpose.
> 
> I would strongly advocate for making this an option.

Also, isn't rename broken if tmp and the destination folder are on different devices? I think the filename has to be alongside the destination file.

1. delete destination file, if it exists
2. write a foo.odb.tmp
3. rename foo.odb.tmp to foo.odb

Better?

QuantamHD:
I like that change a lot better.

> > This feels like a dangerous change to me. Not all users have a tmp folder, and using a time based file name is also a bit dangerous if you have multiple instances running on the same machine. Usually you would use a UUID for this purpose.
> > I would strongly advocate for making this an option.
> 
> Also, isn't rename broken if tmp and the destination folder are on different devices? I think the filename has to be alongside the destination file.
> 
> 1. delete destination file, if it exists
> 2. write a foo.odb.tmp
> 3. rename foo.odb.tmp to foo.odb
> 
> Better?

I like that flow much better

src/OpenRoad.cc
gadfort, line 396: shouldn't this check for failure and report an error accordingly?

oharboe: Doesnt it throw an exception? It is 2024 😵‍💫

maliberty: It does https://en.cppreference.com/w/cpp/filesystem/rename

src/utl/src/ScopedTemporaryFile.cpp
annapetrosyan26: I have used original file name plus ISO 8601 style date time for the tmp file name, which I thought is preferable over just unique file name. Anyway this approach will not be used anymore as @oharboe suggested alternative solution which is accepted by other contributors as well:
> 
> 1. delete destination file, if it exists
> 2. write a foo.odb.tmp
> 3. rename foo.odb.tmp to foo.odb

src/OpenRoad.cc
annapetrosyan26: 1. The  `std::filesystem::rename` will overwrite existing file (`filename` is first deleted, then, without allowing other processes to observe `filename` as deleted, the pathname `filename` is linked to the file and `tmp_filename` is unlinked from the file.) 
2. In the initial implementation the tmp file was left behind intentionally as the tmp file was written in the `/tmp` directory, which is cleaned by the OS. With your suggested solution we will remove it.

src/OpenRoad.cc
annapetrosyan26: Thank you, I'll add the exception handling.

annapetrosyan26:
> I like that change a lot better.
> 
> > > This feels like a dangerous change to me. Not all users have a tmp folder, and using a time based file name is also a bit dangerous if you have multiple instances running on the same machine. Usually you would use a UUID for this purpose.
> > > I would strongly advocate for making this an option.
> > 
> > 
> > Also, isn't rename broken if tmp and the destination folder are on different devices? I think the filename has to be alongside the destination file.
> > 
> > 1. delete destination file, if it exists
> > 2. write a foo.odb.tmp
> > 3. rename foo.odb.tmp to foo.odb
> > 
> > Better?
> 
> I like that flow much better

I'll change the code according suggested flow. @QuantamHD do you still prefer to make this optional?

src/OpenRoad.cc
oharboe: No need to add exception handling, it is already there.

To be a bit more specific, no exception handling:

1. check if the destination file exists, if it exists, then delete it. If deleting fails, then this should throw an exception.
2. open the .tmp file in overwrite mode alongside the final .odb file. If this fails, throw exception. Nothing to be done and the exception is handled higher up.
3. rename .tmp to final destination file. If this fails, then an exception is thrown and it is handled higher up. Nothing for the code to do here.

oharboe:
> > I like that change a lot better.
> > > > This feels like a dangerous change to me. Not all users have a tmp folder, and using a time based file name is also a bit dangerous if you have multiple instances running on the same machine. Usually you would use a UUID for this purpose.
> > > > I would strongly advocate for making this an option.
> > > 
> > > 
> > > Also, isn't rename broken if tmp and the destination folder are on different devices? I think the filename has to be alongside the destination file.
> > > 
> > > 1. delete destination file, if it exists
> > > 2. write a foo.odb.tmp
> > > 3. rename foo.odb.tmp to foo.odb
> > > 
> > > Better?
> > 
> > 
> > I like that flow much better
> 
> I'll change the code according suggested flow. @QuantamHD do you still prefer to make this optional?

Answering for @QuantamHD due to time-zones, any mistakes answering on his behalf are mine: no options are required with the apove approach.

oharboe:
also, I think this code should be a utility fn using [RAII](https://en.cppreference.com/w/cpp/language/raii) so that we can create a stream that is written atomically everywhere.

annapetrosyan26:
@oharboe I'd like to summarize what I'm going to do, that you can confirm if that will work for you.

1. create an utility `WriteFile` function which takes as argument filename and a lambda function, the `WriteFile` will do the following:
    a. check for the filename and filename.tmp files existence, if they exist, then delete them. If deleting fails, then throw an exception.
    b. call the lambda function with the filename.tmp as an argument.
    c. rename filename.tmp to filename.
2. encapsulate the code into lambda functions in all places where currently writing is done into the filename and call `utl::WriteFile` with the filename and the lambdas.

oharboe:
> @oharboe I'd like to summarize what I'm going to do, that you can confirm if that will work for you.
> 
> 1. create an utility `WriteFile` function which takes as argument filename and a lambda function, the `WriteFile` will do the following:
>    a. check for the filename and filename.tmp files existence, if they exist, then delete them. If deleting fails, then throw an exception.
>    b. call the lambda function with the filename.tmp as an argument.
>    c. rename filename.tmp to filename.
> 2. encapsulate the code into lambda functions in all places where currently writing is done into the filename and call `utl::WriteFile` with the filename and the lambdas.

I recommend using the RAII pattern here, not lambda functions.

maliberty:
You need to flush/fsync the file before the rename (see https://www.slideshare.net/nan1nan1/eat-my-data)

maliberty:
Please also note you are missing DCO and need to run clang-format before the merge.

src/OpenRoad.cc
```diff
@@ -469,12 +462,10 @@ void OpenRoad::readDb(const char* filename)
 
 void OpenRoad::writeDb(const char* filename)
 {
-  std::ofstream stream;
-  stream.exceptions(std::ifstream::failbit | std::ifstream::badbit
-                    | std::ios::eofbit);
-  stream.open(filename, std::ios::binary);
+  utl::StreamHandler stream_handler(
+      filename, std::ofstream::badbit | std::ios::eofbit, std::ios::binary);
```
oharboe, line 466: remove "badbit" from arguments and eofbit, just add them inside the StreamHandler implementation as they are required for correct operation of StreamHandler.

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,61 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+std::ofstream& StreamHandler::getStream()
+{
+  return os_;
+}
+
+FileHandler::FileHandler(const char* filename, const char* mode)
+    : filename_(filename), mode_(mode)
+{
+  if (fs::exists(filename_)) {
+    fs::remove(filename_);
```
oharboe, line 73: Remove the mode argument.

It implies that "r" and "rb" and "w" and "wb" are all supported, but they are not. Bad unexpected things will happen...

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,61 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+  }
+  std::string tmp_filename = filename_ + ".tmp";
+  file = fopen(tmp_filename.c_str(), mode);
+  if (!file) {
+    throw std::runtime_error("Failed to open temporary file");
```
oharboe, line 78: read errno and elaborate the exception message a bit with filename that could not be opened.

Search code for errno to find example.

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,61 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+    : filename_(filename), flag_(flag), mode_(mode)
+{
+  if (fs::exists(filename_)) {
+    fs::remove(filename_);
+  }
+  std::string tmp_filename = filename_ + ".tmp";
+  if (fs::exists(tmp_filename)) {
+    fs::remove(tmp_filename);
+  }
+  os_.exceptions(flag | std::ofstream::failbit);
```
oharboe, line 51: add eof bit too

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,61 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
   }
 }
 
+StreamHandler::StreamHandler(const char* filename,
+                             std::ios_base::iostate flag,
+                             std::ios_base::openmode mode)
+    : filename_(filename), flag_(flag), mode_(mode)
+{
+  if (fs::exists(filename_)) {
```
oharboe, line 44: Move the deletion until the very last rename() in the destructor so that the original file is not deleted upon a failed write.

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,61 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+    fs::remove(filename_);
+  }
+  std::string tmp_filename = filename_ + ".tmp";
+  if (fs::exists(tmp_filename)) {
+    fs::remove(tmp_filename);
+  }
+  os_.exceptions(flag | std::ofstream::failbit);
+  os_.open(tmp_filename, mode);
```
oharboe, line 52: The exception thrown here does not include a filename, that would be nice to have in the exception. Articulate a an exception with filename when file can not be created.

oharboe (changes requested):
nits + write unit-tests

src/utl/src/ScopedTemporaryFile.cpp
annapetrosyan26: I'll change mode to binary, which will be boolean type and if its set then file will be opened in `wb` mode otherwise in `w`

annapetrosyan26:
> nits + write unit-tests

@oharboe I propose to add unit test in the `CFileUtilsTest.cpp` file, could you please clarify if this is the right approach?
Also could you please provide some info regarding nits?

oharboe:
> > nits + write unit-tests
> 
> @oharboe I propose to add unit test in the `CFileUtilsTest.cpp` file, 

Should be fine.

> could you please clarify if this is the right approach? Also could you please provide some info regarding nits?

See my inline comments in this review.

src/utl/src/ScopedTemporaryFile.cpp
maliberty: What does the exception contain?  Its odd as the filename is a arg.

src/utl/src/ScopedTemporaryFile.cpp
maliberty: Why would you add eof to an ostream?

src/utl/src/ScopedTemporaryFile.cpp
oharboe: I was going by memory. Could very well be wrong, which would be good!

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,70 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+  }
+  os_.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+  try {
+    if (binary) {
+      os_.open(tmp_filename, std::ios::binary | std::ios_base::out);
+    } else {
+      os_.open(tmp_filename, std::ios_base::out);
+    }
+  } catch (std::ios_base::failure& e) {
+    std::string error = e.what();
```
oharboe, line 54: I don't want to loose the stack trace by replacing the exception.

Please try with https://en.cppreference.com/w/cpp/error/throw_with_nested and confirm that it works:

```
  } catch (const std::ios_base::failure& e) {
    std::string error = e.what();
    std::throw_with_nested(std::ios_base::failure(error + " (failed to open '" + tmp_filename + "')"));
  }
}

// Helper function to print nested exceptions
void printNestedException(const std::exception& e, int level = 0) {
  std::cerr << std::string(level, ' ') << "Exception: " << e.what() << std::endl;
  try {
    std::rethrow_if_nested(e);
  } catch (const std::exception& nested) {
    printNestedException(nested, level + 1);
  } catch (...) {
    // No further nested exception
  }
}

int main() {
  try {
    StreamHandler sh("example.txt", true);
  } catch (const std::exception& e) {
    printNestedException(e);
  }
  return 0;
}
```

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,70 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+    : filename_(filename)
+{
+  std::string tmp_filename = filename_ + ".tmp";
+  if (fs::exists(tmp_filename)) {
+    fs::remove(tmp_filename);
+  }
+  os_.exceptions(std::ofstream::failbit | std::ofstream::badbit);
+  try {
+    if (binary) {
+      os_.open(tmp_filename, std::ios::binary | std::ios_base::out);
```
oharboe, line 49: Nit: I think the following reads a bit better than an if-statement.

os_.open(tmp_filename, (binary ? std::ios::binary : 0) | std::ios_base::out);

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,70 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+}
+
+StreamHandler::~StreamHandler()
+{
+  if (os_.is_open()) {
+    os_.close();
+  }
+  std::string tmp_filename = filename_ + ".tmp";
+  fs::rename(tmp_filename,
+             filename_);  // If filename_ exists it will be overwritten
```
oharboe, line 67: move comment to line before statement instead of using end of line comments. Better with git-changes and I've seen that style of commenting more in OpenROAD code.

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,70 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
...
+{
+  return os_;
+}
+
+FileHandler::FileHandler(const char* filename, bool binary)
+    : filename_(filename)
+{
+  std::string tmp_filename = filename_ + ".tmp";
+  if (binary) {
+    file_ = fopen(tmp_filename.c_str(), "wb");
```
oharboe, line 80: binary?"wb":"w"

src/utl/test/CFileUtilsTest.cpp
```diff
@@ -107,4 +108,111 @@ BOOST_AUTO_TEST_CASE(read_all_of_file_exactly_1025B)
...
+  const char* filename = "test_exception_handling.txt";
+
+  // Ensure the temporary file is handled correctly if an exception occurs
+  try {
+    StreamHandler sh(filename);
+    throw std::runtime_error("Simulated exception");
+  } catch (...) {
+    std::string tmp_filename = std::string(filename) + ".tmp";
+    BOOST_TEST(!std::filesystem::exists(
+        tmp_filename));  // Ensure temporary file is cleaned up
```
oharboe, line 159: comments on line before statement instead of end of line comments

oharboe (changes requested):
Some more changes. Looking good!

There is a reason that this github issue was left alone for a long time, seemingly simple, but error handling & unit-tests need to be added carefully for all cases.

Also, this error handling is an important user-facing interface for several use-cases: normal operation as well as debugging.

src/utl/src/ScopedTemporaryFile.cpp
annapetrosyan26: I don't think that we loose stack trace in this case as both nested exception and the original one have the same stack trace. Anyway do you prefer `std::throw_with_nested`  ?

oharboe: > I don't think that we loose stack trace in this case as both nested exception and the original one have the same stack trace. Anyway do you prefer `std::throw_with_nested` ?

You are right. To be a bit nit-picky: we do loose a bit of stack information, but nothing interesting: inside libstdc++ when opening a file, not application stack trace.

However, it would be nice to have an example of std::throw_with_nested, so it is a "bonus point" for this PR, but not a concern for this PR as such.

Up to you.

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -45,15 +45,13 @@ StreamHandler::StreamHandler(const char* filename, bool binary)
...
-    }
+    os_.open(tmp_filename,
+             (binary ? (std::ios::binary | std::ios_base::out)
+                     : std::ios_base::out));
   } catch (std::ios_base::failure& e) {
     std::string error = e.what();
-    throw std::ios_base::failure(error + " (failed to open '" + tmp_filename
-                                 + "')");
+    std::throw_with_nested(std::ios_base::failure(error + " (failed to open '"
+                                                  + tmp_filename + "')"));
```
oharboe, line 54: did you try it? What does the stack trace now look like?

oharboe (changes requested):
it would be nice to see what the stack trace looks like.

src/utl/src/ScopedTemporaryFile.cpp
annapetrosyan26: - Please see stack trace for both cases below:

throw:
```
#0  0x00007ffff78969fc in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff7842476 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff78287f3 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ffff7ca2b9e in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7cae20c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7cae277 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7cae4d8 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00005555555568c9 in StreamHandler::StreamHandler(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) ()
#8  0x0000555555556be6 in main ()
```
std::throw_with_nested:
```
#0  0x00007ffff78969fc in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff7842476 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff78287f3 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ffff7ca2b9e in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff7cae20c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7cae277 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7cae4d8 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00005555555574d0 in void std::__throw_with_nested_impl<std::ios_base::failure[abi:cxx11]>(std::ios_base::failure[abi:cxx11]&&, std::integral_constant<bool, true>) ()
#8  0x000055555555723c in void std::throw_with_nested<std::ios_base::failure[abi:cxx11]>(std::ios_base::failure[abi:cxx11]&&) ()
#9  0x0000555555556930 in StreamHandler::StreamHandler(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) ()
#10 0x0000555555556c4a in main ()
```

- Please see the error messages for both cases below:

throw:
```
Exception: basic_ios::clear: iostream error (failed to open '/invalid/path/example.txt.tmp'): iostream error
```
std::throw_with_nested:
```
Exception: basic_ios::clear: iostream error (failed to open '/invalid/path/example.txt.tmp'): iostream error
 Exception: basic_ios::clear: iostream error
```

oharboe: 👍 The exception handler in OpenROAD needs to be updated, but I think that is out of scope for this PR.

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -33,4 +36,64 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
   }
 }
 
+StreamHandler::StreamHandler(const char* filename, bool binary)
+    : filename_(filename)
+{
+  std::string tmp_filename = filename_ + ".tmp";
+  if (fs::exists(tmp_filename)) {
+    fs::remove(tmp_filename);
```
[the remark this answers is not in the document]
oharboe: hmmm... now I remember my concern. We really dont want the destructor to throw an exception, but just log errors. otherwise the system can leak memory when the rest of the destructor(member variables) does not run. I must admit that I dont know these C++ details on top of my head, but it is a mentioned concern with RAII.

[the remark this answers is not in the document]
oharboe: Also, when we finally do arrive at the exact implementation, we need some succinct comments why the code is exactly this way or ideally tests to enforce it...

oharboe (approved):
I think this is a good improvement over what is in master and that we can take this as a starting point in master for any future work.

oharboe:
> Does this PR ensure that the tmp file has a unique name which isn't already in use? If a user already has a manually generated file named `filename.tmp` it should not be overwritten.

I disagree. Keep it simple. OpenROAD will "own" .odb.tmp file as tmp file, dont put anything there 

OpenROAD and ORFS overwrites lots of files, because by convention those files are used: reports, logs, objects, results, etc.

rovinski:
I 100% disagree. I often add random extensions to files like `.tmp` `.bak` `.old` etc. in order to make backup files that don't get overwritten. If I do `write_db file.odb` I would have no expectation or knowledge that `file.odb.tmp` would be overwritten in this process. `std::tmpnam()` should be used - it was made for this purpose.

oharboe:
The same is done in ORFS to write logs atomically, BTW.

rovinski:
> The same is done in ORFS to write logs atomically, BTW.

I disagree with that too. I would ask why a log would need to be written atomically, but that's unrelated to this issue. Temporary files should use [`mktemp`](https://www.gnu.org/software/autogen/mktemp.html).

gadfort:
Assuming this change is the direction we want to go (not sure why this wouldn't apply to all files and not a few select ones, except that this is a fix to a limitation in ORFS in which case this feels overkill), I would think the proper naming scheme would be to use a hidden file so `file.odb`, would first go to `.file.odb.tmp` and then renamed to `file.odb`. This would avoid @rovinski concern (I think) while preserving the need to keep the file writing to a specific directory (which @QuantamHD suggested and I agree with).

rovinski:
@gadfort the issue stems from not checking that a file exists before overwriting, especially when that file is not the user-provided filename. Again, here's an example of unexpected behavior from the user side:
```
% ls
file.odb
# User thinks they're making a backup
% cp file.odb file.odb.tmp
% ls
file.odb
% write_db file.odb
# Backup file is now gone
% ls
file.odb
```
There is no checking that the temp file already exists, so it could be overwritten without prompting the user. Using the std/coreutil functions will guarantee that a unique filename is generated. This is what e.g. Google Chrome does when downloading files.

Good behavior:
```
% ls
file.odb
# User thinks they're making a backup
% cp file.odb file.odb.tmp
% ls
file.odb
% write_db file.odb
# Process is interrupted
^C
% ls
file.odb file.odb.tmp file.odb.tmp.D0zMeJoW9g
# temp file with guaranteed unique string is written
```

maliberty:
I already made the same suggestion above: `std::tmpnam or std::tmpfile should be used.`

oharboe:
> I already made the same suggestion above: `std::tmpnam or std::tmpfile should be used.`

I thought @gadfort had a good suggestion, though I think the code is fine as-is. Although I think the tmpfile solves a problem that doesn't exist and creates a problem in ORFS for log files(I want to tail -f a partially written log file), if tmpfile makes this change paletable to the group, then that works for me. Unlike log files that are tee'ed, there are no such files written by openroad.

The reason we need a tmp file name and write atomically is to avoid strange problems with dependencies with a partially written file. ORFS has dependencies on .log and .odb files. The dependency on a log file might be reworked in ORFS, I forget the details.

rovinski:
> I want to tail -f a partially written log file

I think this is orthogonal because this PR doesn't concern log files. You wouldn't `tail -f` an odb, def, etc.

> I already made the same suggestion above: `std::tmpnam or std::tmpfile should be used.`

I thought I remember seeing that comment but somehow I missed it.

oharboe:
Some more color on the motivation for atomic writes: it avoids partial files upon error/crashes, which improves the user experience

Also, I like fixed  tmp names, because random files do not accrue. Not a biggie, but I thought I would mention it. /tmp is full of orphaned temporary files sometimes....

maliberty:
POSIX has tempnam (not tmpnam) which allows a directory specification if that is helpful.

maliberty (approved):
I don't want to drag this out over too many nits.  We can address the name of the temp file in a follow on PR.

maliberty:
Thanks for contributing!

maliberty:
The nit picking can migrate to #5181

PR #5110, opened 2024-05-15
odb: provide dbuToMicrons for various datatypes to avoid casting errors

gadfort:
Fixes overflow:
```
original HPWL         -964503.2 u
legalized HPWL       -2147483.6 u
```
->
```
original HPWL         3330464.1 u
legalized HPWL        3341133.7 u
```

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -1185,10 +1185,10 @@ class dbBlock : public dbObject
   ///
   /// Convert a length from database units (DBUs) to microns.
   ///
-  double dbuToMicrons(const int dbu);
-  double dbuToMicrons(const unsigned int dbu);
-  double dbuToMicrons(const int64_t dbu);
-  double dbuToMicrons(const double dbu);
+  double dbuToMicrons(int dbu);
+  double dbuToMicrons(unsigned int dbu);
+  double dbuToMicrons(int64_t dbu);
+  double dbuToMicrons(double dbu);
 
   ///
   /// Convert an area from database units squared (DBU^2) to square microns.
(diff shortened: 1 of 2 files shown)


src/odb/include/odb/db.h
```diff
@@ -1185,7 +1185,10 @@ class dbBlock : public dbObject
   ///
   /// Convert a length from database units (DBUs) to microns.
   ///
-  double dbuToMicrons(int dbu);
+  double dbuToMicrons(const int dbu);
+  double dbuToMicrons(const unsigned int dbu);
+  double dbuToMicrons(const int64_t dbu);
```
maliberty, line 1190: I think it would suffice to have int64_t as the shorter integral types can always be promoted to this one.

gadfort: I get if I don't include all four (if I don't have the double I we can get away with just a single int64_t)
```
src/ant/src/AntennaChecker.cc:2203:46: error: call of overloaded ‘dbuToMicrons(int)’ is ambiguous
 2203 |           wire_length += block_->dbuToMicrons((abs(x2 - x1) + abs(y2 - y1)));

note: candidate: ‘double odb::dbBlock::dbuToMicrons(double)’
 1189 |   double dbuToMicrons(double dbu);
note: candidate: ‘double odb::dbBlock::dbuToMicrons(int64_t)’
 1188 |   double dbuToMicrons(int64_t dbu);
```

maliberty: Looks like I'm just wrong so nvm

maliberty (approved):

PR #5134, opened 2024-05-18
Adding Timing::equivCells and Timing::makeEquivCells python APIs

bingyuew:
Move makeEquivCells from protected to public in rsz::Resizer and call it in Timing::makeEquivCells. Create Timing::equivCells.

### DIFF
--- a/include/ord/Timing.h
+++ b/include/ord/Timing.h
@@ -110,7 +110,7 @@ class Timing
   sta::Corner* findCorner(const char* name);
 
   void makeEquivCells();
-  std::vector<std::string> equivCells(odb::dbMaster* master);
+  std::vector<odb::dbMaster*> equivCells(odb::dbMaster* master);
 
  private:
   sta::dbSta* getSta();
--- a/include/ord/Timing.h
+++ b/include/ord/Timing.h
@@ -63,6 +63,10 @@ class Vertex;
 class Pin;
 }  // namespace sta
 
+namespace rsz {
+  class Resizer;
+}
+
 namespace ord {
 
 class Design;
--- a/include/ord/Timing.h
+++ b/include/ord/Timing.h
@@ -63,10 +63,6 @@ class Vertex;
 class Pin;
 }  // namespace sta
 
-namespace rsz {
-  class Resizer;
-}
-
 namespace ord {
 
 class Design;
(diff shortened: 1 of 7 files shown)


include/ord/Timing.h
```diff
@@ -63,6 +63,10 @@ class Vertex;
 class Pin;
 }  // namespace sta
 
+namespace rsz {
+  class Resizer;
+}
+
```
maliberty, line 69: Why is this needed?  I see no mention of Resizer in this header.

src/Timing.cc
```diff
@@ -385,4 +386,32 @@ float Timing::dynamicPower(odb::dbInst* inst, sta::Corner* corner)
...
+  sta::dbSta* sta = getSta();
+  sta::dbNetwork* network = sta->getDbNetwork();
+  sta::Cell* cell = network->dbToSta(master);
+  std::vector<odb::dbMaster*> masterSeq;
+  if (cell) {
+    sta::LibertyCell* libcell = network->libertyCell(cell);
+    sta::LibertyCellSeq* equiv_cells = sta->equivCells(libcell);
+    if (equiv_cells) {
+      for (sta::LibertyCell* equiv_cell : *equiv_cells) {
+        odb::dbMaster* master_
```
maliberty, line 407: master_ looks like a member variable.  Perhaps equiv_master?

src/Timing.cc
```diff
@@ -385,4 +386,32 @@ float Timing::dynamicPower(odb::dbInst* inst, sta::Corner* corner)
...
+            = reinterpret_cast<odb::dbMaster*>(equiv_cell->extCell());
```
maliberty, line 408: Use dbNetwork::staToDb for conversion from LibertyCell to dbMaster.

maliberty (changes requested):

include/ord/Timing.h
bingyuew: Fixed that!

src/Timing.cc
bingyuew: Fixed that!

src/Timing.cc
bingyuew: Fixed that!

maliberty (approved):

maliberty:
Needs a clang-format

PR #5139, opened 2024-05-20
gui: add filter to Charts based on Start/End path type

### LINKED ISSUE #5000: Add filtering in Timing Report and End point slack histogram for register to register paths only
### Description

Timing Reports and Endpoint slack histogram does not distinguish between timing close for register to register paths and timing paths that are there simply to give an optimization target. For more detail on this see https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1969.

To reproduce:

```
make DESIGN_CONFIG=designs/asap7/mock-array/Element/config.mk
```

Then launch GUI:

```
make DESIGN_CONFIG=designs/asap7/mock-array/Element/config.mk gui_final
```

image

### Suggested Solution

Add a filtering option to Timing Report and Endpoint Slack histogram to only consider register to register paths.

I think it would be nice to be able to articulate a report_checks command type filtering in Timing Report, rather than have lots of buttons.

Endpoint slack histogram could be representing what is in the Timing Report (except that the endpoint slack histogram should not have a limit on number of paths).

I would like to see an endpoint slack histogram for this query:

```
report_checks -from [all_registers] -to [all_registers] -path_delay max
```

### Additional Context

_No response_

AcKoucher:
Resolve #5000

### DIFF
--- a/src/gui/src/chartsWidget.cpp
+++ b/src/gui/src/chartsWidget.cpp
@@ -65,8 +65,7 @@ ChartsWidget::ChartsWidget(QWidget* parent)
       axis_x_(new QValueAxis(this)),
       axis_y_(new QValueAxis(this)),
       buckets_(std::make_unique<Buckets>()),
-      previous_mode_(SELECT),
-      prev_filter_index_(0),
+      prev_filter_index_(2),  // start with no filter
 #endif
       label_(new QLabel(this))
 {
@@ -117,24 +116,16 @@ ChartsWidget::ChartsWidget(QWidget* parent)
 #ifdef ENABLE_CHARTS
 void ChartsWidget::changeMode()
 {
-  if (previous_mode_ == SELECT) {
-    filters_menu_->hide();
-    clearChart();
+  if (mode_menu_->currentIndex() == SELECT) {
+    return;
   }
 
-  Mode current_mode = static_cast<Mode>(mode_menu_->currentIndex());
+  clearChart();
 
-  switch (current_mode) {
-    case SELECT: {
-      return;
-    }
-    case SLACK_HISTOGRAM: {
-      filters_menu_->show();
-      setSlackHistogram();
-    }
+  if (mode_menu_->currentIndex() == SLACK_HISTOGRAM) {
+    filters_menu_->show();
+    setSlackHistogram();
   }
-
-  previous_mode_ = current_mode;
 }
 
 void ChartsWidget::setModeMenu()
@@ -232,14 +223,7 @@ void ChartsWidget::setSlackHistogram()
   connect(neg_set, &QBarSet::hovered, this, &ChartsWidget::showToolTip);
   connect(pos_set, &QBarSet::hovered, this, &ChartsWidget::showToolTip);
 
-  for (int i = 0; i < buckets_->negative.size(); ++i) {
-    *neg_set << buckets_->negative[i].size();
-    *pos_set << 0;
-  }
-  for (int i = 0; i < buckets_->positive.size(); ++i) {
-    *neg_set << 0;
-    *pos_set << buckets_->positive[i].size();
-  }
+  populateBarSets(*neg_set, *pos_set);
 
   QStackedBarSeries* series = new QStackedBarSeries(this);
   series->append(neg_set);
@@ -298,8 +282,8 @@ SlackHistogramData ChartsWidget::fetchSlackHistogramData() const
       continue;
     }
 
+    // Default filter is register to register so we don't need to check it.
     if (start_node->isPinITerm()) {
-      // Default is register to register.
       if (end_node->isPinBTerm()) {
         histogram_end_point.path_type = RegisterToIO;
       }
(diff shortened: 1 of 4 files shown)


AcKoucher:
@oharboe @maliberty Would you happen to know a design in which we'd have all the path start/end type for a more consistent test?

oharboe:
> @oharboe @maliberty Would you happen to know a design in which we'd have all the path start/end type for a more consistent test?

Try designs/asap7/mock-array/Element

oharboe:
@maliberty I tried on mock-array/Element:

- Why are IO to REG listed with clock and reg to io listed with <no clock>?
- IO - IO have no clock. This isn't entirely surprising, it is a max delay path only. Can this path be shown in the endpoint slack histogram? If so, under what clock, if any? Add "<no clock>" to endpoint histogram?
- register to register paths exist with "clock".
- "No filter" and "Register to register" show the same. The other filters show now end-paths.


image


No endpoints shown in histogram:

image

maliberty:
When it first loads I see:
image

which appears to be equivalent to
image

Would you set the initial state to "No Filter"

maliberty:
@AcKoucher is looking at the lack of IO related paths

maliberty:
The set of clocks is not filter dependent and just shows what are defined in the design.

oharboe:
If I click on one of the buckets, I get a filtered list in Timing Report. I think the same dropdown filter would be useful in the Timing Report.

I am sure this raises some questions to sort through...

oharboe:
If I choose "Register to register" (or any non-filtered), I get a crash:

untar [gui-crash.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/15383934/gui-crash.tar.gz)


```
$ ./run-me-mock-array_Element-asap7-base.sh 
OpenROAD v2.0-13792-g7457e3a7e 
read_spef ./results/asap7/mock-array_Element/base/6_final.spef
Signal 11 received
Stack trace:
 0# 0x00006335527B8693 in openroad
 1# 0x0000729FC5C42990 in /lib/x86_64-linux-gnu/libc.so.6
 2# std::_Rb_tree_increment(std::_Rb_tree_node_base const*) in /lib/x86_64-linux-gnu/libstdc++.so.6
 3# gui::ChartsWidget::fetchConstrainedPins(std::set<sta::Pin const*, std::less<sta::Pin const*>, std::allocator<sta::Pin const*> >&, bool) in openroad
 4# gui::ChartsWidget::changeStartEndFilter() in openroad
 5# 0x0000729FC6F06312 in /lib/x86_64-linux-gnu/libQt5Core.so.5
 6# QComboBox::currentIndexChanged(int) in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
 7# 0x0000729FC7C732E6 in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
 8# 0x0000729FC7C76C5B in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
 9# 0x0000729FC7C76ECB in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
10# 0x0000729FC6F0602D in /lib/x86_64-linux-gnu/libQt5Core.so.5
11# QComboBoxPrivateContainer::itemSelected(QModelIndex const&) in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
12# QComboBoxPrivateContainer::eventFilter(QObject*, QEvent*) in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
13# QCoreApplicationPrivate::sendThroughObjectEventFilters(QObject*, QEvent*) in /lib/x86_64-linux-gnu/libQt5Core.so.5
14# QApplicationPrivate::notify_helper(QObject*, QEvent*) in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
15# QApplication::notify(QObject*, QEvent*) in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
16# QCoreApplication::notifyInternal2(QObject*, QEvent*) in /lib/x86_64-linux-gnu/libQt5Core.so.5
17# QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
18# 0x0000729FC7BC886A in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
19# 0x0000729FC7BCB12F in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
20# QApplicationPrivate::notify_helper(QObject*, QEvent*) in /lib/x86_64-linux-gnu/libQt5Widgets.so.5
21# QCoreApplication::notifyInternal2(QObject*, QEvent*) in /lib/x86_64-linux-gnu/libQt5Core.so.5
22# QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) in /lib/x86_64-linux-gnu/libQt5Gui.so.5
23# QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) in /lib/x86_64-linux-gnu/libQt5Gui.so.5
24# 0x0000729FC1EF9F7E in /lib/x86_64-linux-gnu/libQt5XcbQpa.so.5
25# 0x0000729FC5F13B2C in /lib/x86_64-linux-gnu/libglib-2.0.so.0
26# 0x0000729FC5F6F46F in /lib/x86_64-linux-gnu/libglib-2.0.so.0
27# g_main_context_iteration in /lib/x86_64-linux-gnu/libglib-2.0.so.0
28# QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) in /lib/x86_64-linux-gnu/libQt5Core.so.5
29# QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) in /lib/x86_64-linux-gnu/libQt5Core.so.5
30# QCoreApplication::exec() in /lib/x86_64-linux-gnu/libQt5Core.so.5
31# gui::startGui(int&, char**, Tcl_Interp*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) in openroad
32# ord::tclAppInit(Tcl_Interp*) in openroad
33# Tcl_MainEx in /lib/x86_64-linux-gnu/libtcl8.6.so
34# main in openroad
35# 0x0000729FC5C28150 in /lib/x86_64-linux-gnu/libc.so.6
36# __libc_start_main in /lib/x86_64-linux-gnu/libc.so.6
37# _start in openroad
./run-me-mock-array_Element-asap7-base.sh: line 7: 11889 Segmentation fault      (core dumped) openroad -no_init -gui ${SCRIPTS_DIR}/gui.tcl
```

AcKoucher:
@oharboe The changes is this draft are neither ready for testing nor review.

AcKoucher:
@oharboe Results for the the element test case of your previous comment:

All endpoints:
<image src="https://github.com/The-OpenROAD-Project/OpenROAD/assets/104802710/0d76e65f-b038-46cd-8007-322a6b8c9206" width="500">

Filters:
<p float="left">
  
   
</p>

<p float="left">
  
   
</p>

oharboe:
@AcKoucher Perfect! Ready for testing?

AcKoucher:
@oharboe Yes. Go ahead :)

oharboe:
@AcKoucher Should the same dropdown filter be available from the Timing Report menu?

Of course I can click on one column in the filtered histogram, which gives me the top paths in for that filter in that slack range, so I have what I need.

I'm unsure if that complication to the GUI is worth the potential benefit.

AcKoucher:
@oharboe Well, something to have in mind is that we're dealing with an endpoint slack histogram. The report that you get from clicking in a certain range gives you the paths of those endpoints, but that doesn't mean you'll get only the paths within that filter.

An endpoint shown in the reg2reg filter might as well appear in the io2reg filter. That's why the "no filter" option - which presents all the endpoints - is not necessarily a superposition of the filtered ones.

oharboe:
@AcKoucher My domain knowledge isn't up to the task of saying what exactly would be useful here. The limited case that I am interested in and think I understand is that every macro must have timing closure for all reg2reg paths, because those paths are no longer visible where the macro is used as the .lib file doesn't contain information about those paths anymore.

oharboe:
@AcKoucher But from a usability point of view, it is surprising that clicking on an empty column can show paths in the Timing Report. And purely from a usability point of view, surprises should be minimized...

oharboe:
Hmm.... choosing a filter is *much* slower than showing the no-filter version of the histogram. Is that expected.

Still running minutes later....

A few suspend in debugger shows no debug info(so release build) and the code is gnawing on the same bone:

image

image

AcKoucher:
@oharboe That is expected, because for and endpoints we just get the worst slack to show in the histogram. For the filters we have to query the paths from sta based on the start/end combination which can be slow if there are a lot of endpoints.

As for the empty column I agree with you and I think it shouldn't be that way either, I'll wait for @maliberty 's input.

oharboe:
@AcKoucher Thanks for the clarification. So how much slower could it be? Looks stuck here. I'm thinking 1 minute for the first one and still running after 20 minutes or so.

AcKoucher:
@oharboe I need to make some more testing to confirm. Perhaps there's more to it.

oharboe:
The filter did complete in the end, perhaps an hour compared to 1 minute with no filter?

maliberty:
Let's prove the viability of this before we worry about adding it to the timing report.

oharboe:
Viability in the sense that it needs to be fast enough or something else?

oharboe:
"No filter", less than a second. reg2reg, 1 hour?

AcKoucher:
@oharboe Do you have any idea of how much time does the Timing Report update on the GUI takes in this test case you're using?

oharboe:
I tried on `make DESIGN_CONFIG=designs/sky130hd/microwatt/config.mk cts gui_cts` and the filter is acceptably fast. Less than a minute.

So I think that the performance problem is for large designs, such as ours and possibly megaboom. Those designs are known to have serious performance problems that must be carefully worked around already, such as disabling timing reports in builds, skipping timing repair in cts, disable power analysis/fixing, etc.

oharboe:
@maliberty @AcKoucher  I think this is ready to merge, it is fast on the larges sky130hd/microwatt design.

That it is slow on larger designs is a separate concern from this PR.

maliberty (approved):

Issue #4982, opened 2024-04-22
recover_power is very slow


oharboe:
### Description

Tar made with `make DESIGN_CONFIG=designs/asap7/jpeg_lvt/config.mk  global_route_issue`

untar https://drive.google.com/file/d/1Aax8fTJl9KyA0rl7YNv8mwu5yG4kaBeo/view?usp=sharing

Run:

```
./run-me-jpeg_lvt-asap7-base.sh
```

```
$ ./run-me-jpeg_lvt-asap7-base.sh 
OpenROAD v2.0-13348-gd423155d6
[deleted]
[quickly, a few minutes, gets to here and then it is stuck for ca. 1-2 hours]
Downsizing/switching to higher Vt  for non critical gates for power recovery
Percent of paths optimized 100
tns 0.00
wns 0.00
Group                  Internal  Switching    Leakage      Total
                          Power      Power      Power      Power (Watts)
----------------------------------------------------------------
Sequential             1.34e-02   2.66e-03   6.57e-06   1.61e-02  20.9%
Combinational          2.25e-02   3.23e-02   1.04e-05   5.48e-02  71.2%
Clock                  3.64e-03   2.43e-03   1.39e-07   6.07e-03   7.9%
Macro                  0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
Pad                    0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
----------------------------------------------------------------
Total                  3.96e-02   3.74e-02   1.71e-05   7.70e-02 100.0%
                          51.4%      48.6%       0.0%
```

Compared to the rest of the flow, global route is surprisingly slow:

```
Log                            Elapsed seconds Peak Memory/KB
1_1_yosys                                  243         971412
2_1_floorplan                               27         510688
2_2_floorplan_io                             4         363812
2_3_floorplan_tdms                           4         362360
2_4_floorplan_macro                          4         367364
2_5_floorplan_tapcell                        4         325804
2_6_floorplan_pdn                            6         381020
3_1_place_gp_skip_io                        38         461672
3_2_place_iop                                4         372996
3_3_place_gp                               613        1048224
3_4_place_resized                           94         602152
3_5_place_dp                                78         657092
4_1_cts                                    152         747132
5_1_grt                                   6137        1298604
5_2_fillcell                                 5         494576
5_3_route                                  543       11470100
6_1_merge                                   15         815860
6_report                                   198        2751824
[...]

PR #5150, opened 2024-05-22
Skip known bad vertices in power recovery

kbieganski:
In power recovery, vertices that have already been tried and rejected can be tried again if they happen to be on another path we look at. Usually, the result is the same – we reject the vertex. Not only is this work unnecessary, it also prevents us from trying another (possibly better) vertex later in the path. Some stats on vertices that are tried in #4982:

| Vertex | Total tries | Better | Worse |
| - | - | - | - |
| `max_length492/Y` | 1325 | 1 | 1324 |
| `load_slew487/Y` | 130 | 1 | 129 |
| `load_slew465/Y` | 309 | 1 | 308 |
| `load_slew447/Y` | 231 | 1 | 230 |
| `load_slew409/Y` | 49 | 0 | 49 |
| `load_slew400/Y` | 294 | 2 | 292 |
| `load_slew375/Y` | 262 | 1 | 261 |
| `load_slew57/Y` | 240 | 0 | 240 |
| `load_slew407/Y` | 124 | 0 | 124 |
| `fdct_zigzag.dct_mod.dct_block_0.dct_unit_0.ddin\[7\]$_DFFE_PN0P_/QN` | 141 | 0 | 141 |
| `max_length63/Y` | 63 | 1 | 62 |
| `_085272_/Y` | 481 | 0 | 481 |
| `load_slew402/Y` | 56 | 1 | 55 |
| `_083048_/Y` | 2 | 1 | 1 |
| `wire301/Y` | 26 | 1 | 25 |
| `load_slew62/Y` | 4 | 2 | 2 |
| `load_slew446/Y` | 12 | 2 | 10 |
| `_078958_/Y` | 3 | 0 | 3 |
| `load_slew399/Y` | 2 | 1 | 1 |
| `load_slew486/Y` | 14 | 1 | 13 |
| `load_slew462/Y` | 27 | 2 | 25 |
| `load_slew461/Y` | 8 | 1 | 7 |
| `_118922_/Y` | 2 | 0 | 2 |
| `_122581_/CON` | 3 | 0 | 3 |
| `load_slew177/Y` | 3 | 0 | 3 |

In the worst case, one vertex is visited 1325 times, 1324 of which result in worse worst slack. That's also about a quarter of all attempts.

This patch makes it so we don't revisit the vertices that were already rejected.

In the example in #4982, this speeds up GRT from ~1.5h to around 6 minutes, so we're looking at ~15x the performance on that particular design.

Also, the actual power recovery seems overall more effective (not sure how to weigh these values against each other).
power-diff

### DIFF
--- a/src/rsz/src/RecoverPower.cc
+++ b/src/rsz/src/RecoverPower.cc
@@ -66,7 +66,8 @@ using sta::Edge;
 using sta::PathExpanded;
 using sta::VertexOutEdgeIterator;
 
-RecoverPower::RecoverPower(Resizer* resizer) : resizer_(resizer)
+RecoverPower::RecoverPower(Resizer* resizer)
+    : resizer_(resizer), bad_vertices_(resizer->graph_)
 {
 }
 
@@ -142,7 +143,7 @@ void RecoverPower::recoverPower(const float recover_power_percent)
     //=====================================================================
     resizer_->journalBegin();
     PathRef end_path = sta_->vertexWorstSlackPath(end, max_);
-    const bool changed = recoverPower(end_path, end_slack_before);
+    Vertex* const changed = recoverPower(end_path, end_slack_before);
     if (changed) {
       resizer_->updateParasitics(true);
       sta_->findRequireds();
@@ -179,6 +180,8 @@ void RecoverPower::recoverPower(const float recover_power_percent)
                    worst_slack_before,
                    worst_slack_after);
       } else {
+        // Save the vertex to avoid trying it again.
+        bad_vertices_.insert(changed);
         // Undo the change here.
         ++failed_move_threshold;
         if (failed_move_threshold > failed_move_threshold_limit_) {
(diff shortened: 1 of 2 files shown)


kbieganski:
Results for other cores (all on `nangate45`):
### Ibex
Ibex
### BlackParrot
BlackParrot
### Ariane
Ariane

maliberty:
bad_vertices_ needs to be cleared at the end of recoverPower so it doesn't persist between calls.

kbieganski:
Sure

maliberty:
I tested the current recover power designs and they look fine.  It just need the clear mentioned above to be ready to merge.

kbieganski:
Added the clear.

src/rsz/src/RecoverPower.cc
```diff
@@ -223,10 +226,11 @@ void RecoverPower::recoverPower(const float recover_power_percent)
   if (resizer_->overMaxArea()) {
     logger_->error(RSZ, 125, "max utilization reached.");
   }
+  bad_vertices_.clear();
```
maliberty, line 229: On any exception this code won't be execute (eg "max utilization reached").

kbieganski: Ok, didn't know that throws. I'll move it.

kbieganski: Pushed the clear before the logs.

maliberty (approved):

PR #5152, opened 2024-05-22
Implementing callbacks for ScanDef parser and writer

fgaray:
Tested: read/write test in scandef_sky130.tcl


maliberty:
ready for review?

fgaray:
Yes, this is ready, thanks!

src/odb/src/defout/defout_impl.cpp
```diff
@@ -913,6 +928,69 @@ void defout_impl::writeGroups(dbBlock* block)
...
+          dbScanInst::AccessPins access_pins = scan_inst->getAccessPins();
+          const std::string scan_in_name = std::visit(
+              [](auto&& pin) { return getPinName(pin); }, access_pins.scan_in);
+          const std::string scan_out_name = std::visit(
+              [](auto&& pin) { return getPinName(pin); }, access_pins.scan_out);
+          fprintf(_out,
+                  "  %s ( IN %s ) ( OUT %s )\n",
+                  scan_inst->getInst()->getName().c_str(),
+                  scan_in_name.c_str(),
+                  scan_out_name.c_str());
```
maliberty, line 982: Is this correct if you have a bterm for scan_in/out ?

src/odb/src/defin/definReader.cpp
```diff
@@ -72,6 +72,107 @@
   }
 namespace odb {
 
+namespace {
+
+// Helper function to get the correct number of bits of a cell for scandef.
+int calculateBitsForCellInScandef(int bits, dbInst* inst)
```
maliberty, line 78: bits can be const

src/odb/src/defin/definReader.cpp
```diff
@@ -72,6 +72,107 @@
...
+}
+
+void populateScanInst(definReader* reader,
+                      dbBlock* block,
+                      defiScanchain* scan_chain,
+                      dbScanList* db_scan_list,
+                      const char* inst_name,
+                      const char* in_pin_name,
+                      const char* out_pin_name,
+                      int bits)
```
maliberty, line 140: bits can be const

src/odb/src/defin/definReader.cpp
```diff
@@ -72,6 +72,107 @@
...
+      = findScanITerm(reader, inst, out_pin_name, scan_chain->commonOutPin());
+  if (!scan_out) {
+    reader->_logger->error(utl::ODB,
+                           257,
+                           "SCANDEF OUT pin {} does not exist in cell {}",
+                           in_pin_name,
+                           inst_name);
+  }
+
+  scan_inst->setAccessPins({.scan_in = scan_in, .scan_out = scan_out});
```
maliberty, line 170: Nice - I tend to forget you can do this now.

src/odb/src/defin/definReader.cpp
```diff
@@ -1898,14 +2089,17 @@ bool definReader::createBlock(const char* file)
...
-
     defrSetStartPinsCbk(pinsStartCallback);
     defrSetStylesStartCbk(stylesCallback);
     defrSetTechnologyCbk(technologyCallback);
   }
 
+  if (_mode == defin::INCREMENTAL || _mode == defin::DEFAULT) {
+    defrSetScanchainsStartCbk(scanchainsStartCallback);
+    defrSetScanchainCbk(scanchainsCallback);
+    defrSetSlotStartCbk(slotsCallback);
```
maliberty, line 2100: Why the change for slots?

src/odb/src/defin/definReader.cpp
fgaray: :)

src/odb/src/defout/defout_impl.cpp
fgaray: Hey Matt,

I am having issues coming up with a test case for this.

I have in verilog something like:

```
module top();
   sub sub1(scan_in_1, scan_out_1);
   sub sub2(scan_in_2, scan_out_2);
endmodule
```


But looks like odb flattens the modules and the instances take names like "sub1/cell". The only way that I found that may work is getting the dbModInst at the top, then get the dbModule master and then get a dbModBTerm:

```
for (dbModInst* inst: block->getModInsts()) {
    if (inst->getName() == std::string_view(inst_name)) {
        dbModule* master = inst->getMaster();
        for (dbModBTerm* bterm:  master->getModBTerms()) {
            std::cout << "GARAY " << bterm->getName() << std::endl; // This gives me scan_in_1 and scan_out_1
        }
    }
}
```

Q1) What is the relationship between dbModBTerm and a dbBTerm? I don't think I can get a BTerm from a ModBTerm.

Q2) Do I need to go back to my dbScanPin and change the BTerm for a dbModBTerm?


Thanks!

src/odb/src/defin/definReader.cpp
maliberty: I don't see a change, did you push it?

src/odb/src/defout/defout_impl.cpp
maliberty: dbModBTerm is for terms of a dbModule.   Such a term may just represent an intermediate level of the hierarchy with no corresponding dbBTerm.

Are scan pins ever intermediate hierarchy levels?

src/odb/src/defin/definReader.cpp
fgaray: I pushed them now.

src/odb/src/defout/defout_impl.cpp
fgaray: So, I am not sure about that concept in odb so let me try to explain what I want here:

I want to be able to represent a CORE where we don't want to expose the internals in the scandef:

image

So, this CORE will have two scan ins and two scan outs. For this core, we have this scandef:

image

Note that test_si1,  test_so1, test_si2 and test_so2 are pins (are they ITerms or BTerms?).

So now my question would be: how can OpenROAD represent this? Is this going to be a hier? if this is a hier, how can I access those pins?

Thanks for your help Matt!

maliberty: DEF can only refer to instances that appear in the DEF afaik.  That would imply that CORE is a module with a LEF and these are iterms.  Does that seem right?

fgaray: Thanks!, I created a regresion for the cores.

Responding at your first question, I think I got the Bterm/ITerm relationship wrong:

Looking at this: 
image

I think that I will not have a BTerm ever for the cells/blocks and I will always have an ITerm. In that case, the scaninsts do not need to hold a BTerm. Wdyt?

I still think I need that std::variant for BTerm ITerm for scan chains.

fgaray:
Hi @maliberty , sorry for the delay, this is ready for another look. Thanks!

src/dft/test/scandef_core_sky130.def.ok
maliberty, line 1: Should be named .defok not .def.ok

src/odb/src/defout/defout_impl.cpp
maliberty: Can a scan chain start/end on a primary (ie top-level) bterm?  If not then you will always have iterms.

src/odb/src/defin/definReader.cpp
```diff
@@ -72,6 +72,107 @@
...
+                       const char* pin_name,
+                       const char* common_pin)
+{
+  if (!pin_name) {
+    if (!common_pin) {
+      reader->_logger->error(utl::ODB,
+                             259,
+                             "SCANDEF is missing either component pin or a "
+                             "COMMONSCANPINS for instance {}",
+                             inst->getName());
```
maliberty, line 107: Errors should be handled though definReader::error (same throughout).  Note that it will not stop execution and you should be prepared to continue (generates more than the first error, also -continue_on_errors).

maliberty (changes requested):

src/odb/src/defout/defout_impl.cpp
fgaray: Yes!, they can. I think then it makes sense to have the std::variant. Thanks!

maliberty: Ok but bterms don't belong to an instance so is this format correct?

src/odb/src/defin/definReader.cpp
fgaray: Thanks for the suggestion!, I think I got your comment about continuing showing more than the first error, I tried to do that when possible. Thanks!, this is ready for another look.

fgaray:
I am looking at the failures now

fgaray:
Solved! I did not correctly updated sta/ when doing a merge from master.

maliberty (approved):

PR #5156, opened 2024-05-24
Python timing API for max liberty pin capacitance limit

VidyaChhabria:
Adding API for max cap limit in Python for the gate sizing contest.


maliberty:
Please fix the clang-tidy and a unit test

maliberty (approved):

PR #5160, opened 2024-05-25
mpl2: group snap functionalities in a class and avoid snapping twice

AcKoucher:
Changes:
- Prepare mpl2 for removing the dbu <--> micron internal methods of conversion for the new dbBlock ones
- Make `place_macro` not relying on mpl2 internals anymore
- Avoid snapping when updating odb the first time

The minimal changes in macro_only test results are due to removing the unneeded snap.

If the approach is ok I'll run CI


maliberty:
go ahead w/CI

AcKoucher:
Opened [#2036](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2036) for the metrics. So this should be good to merge.

maliberty (approved):

PR #5180, opened 2024-05-29
Fix type errors in odb I/BTerm iteration logic.

mikesinouye:
Some minor tweaks to return types and included headers were necessary to build openroad using our build setup.

### DIFF
--- a/src/odb/src/db/dbModuleBTermItr.cpp
+++ b/src/odb/src/db/dbModuleBTermItr.cpp
@@ -34,6 +34,9 @@
 #include "dbModuleBTermItr.h"
 
 #include "dbModBTerm.h"
+#include "dbModITerm.h"
+#include "dbModule.h"
+#include "dbObject.h"
 #include "dbTable.h"
 
 namespace odb {
@@ -80,7 +83,7 @@ uint dbModuleBTermItr::begin(dbObject* parent)
 {
   // User Code Begin begin
   _dbModule* _module = (_dbModule*) parent;
-  _dbModBTerm* _modbterm = _module->_modbterms;
+  return _module->_modbterms;
   // User Code End begin
 }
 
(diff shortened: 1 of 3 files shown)


maliberty (approved):

src/odb/src/db/dbModuleBTermItr.cpp
```diff
@@ -34,6 +34,9 @@
 #include "dbModuleBTermItr.h"
 
 #include "dbModBTerm.h"
+#include "dbModITerm.h"
+#include "dbModule.h"
+#include "dbObject.h"
```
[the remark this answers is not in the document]
maliberty: should be odb/dbObject.h

maliberty:
Please fix the DCO

src/odb/src/db/dbModuleBTermItr.cpp
```diff
@@ -80,7 +82,7 @@ uint dbModuleBTermItr::begin(dbObject* parent)
 {
   // User Code Begin begin
   _dbModule* _module = (_dbModule*) parent;
-  _dbModBTerm* _modbterm = _module->_modbterms;
+  return _module->_modbterms;
```
rovinski, line 85: Actually seems like a bug fix? I'm surprised the compiler didn't complain about not returning a value?

maliberty: I see this file isn't in the CMakeLists.txt so it isn't even compiled.  How did you happen upon this error?

maliberty: @andyfox-rushc did you forget to the commit the json for this file?

maliberty: @andyfox-rushc I see dbModInstITermItr, dbModInstModITermItr, dbModuleBTermItr, dbModuleModInstITermItr all have this issue.

mikesinouye: We are using our own bazel scripts instead of cmake to build. In our setup we were building everything in this directory, but this seemed better to fix upstream than to exclude.

andyfox-rushc: These files need to be removed as they are redundant. The came from a prior mis-naming in the schema. All the new dbModule iterators are prefixed by dbModuleMod*. So this is my error. New pull request to be issued removing these unused files.

maliberty:
#5182 should take precedence on the erroneous files

maliberty:
Please merge master and resolve the conflicts.

mikesinouye:
Great, thanks for getting this fixed quickly.

PR #5183, opened 2024-05-30
Added a new function to remove unused dbMasters.

### LINKED ISSUE #2910: deltaDebug.py leaves files with a few insts and nets that is megabytes
### Description

One great thing about deltaDebug.py is that it can reduce confidential examples to the point where they can be shared as bug reports.

However, when the resulting .odb file with a few insts and nets is  megabytes, that doesn't install confidence that no confidential information is divulged.

### Suggested Solution

The .odb files should be small and then converted to a textual representation that can be inspected for confidential information.

When there are a few insts and nets I would have expected a text file in the kilobytes, not a binary in megabytes...

### Additional Context

_No response_

NairiBaveyann:
This PR closes #2910. C++ and Python new functions have been added:

- `int dbDatabase::removeUnusedMasters()` - removes unused masters
- `def remove_unused_masters(self)` - to prepare database and call above C++ function

### DIFF
--- a/etc/deltaDebug.py
+++ b/etc/deltaDebug.py
@@ -191,6 +191,9 @@ class deltaDebugger:
                 if err is None or cuts == 0:
                     break
 
+        # Delete unused master-cells from design
+        self.remove_unused_masters()
+
         # Change deltaDebug resultant base_db file name to a representative name
         if os.path.exists(self.temp_base_db_file):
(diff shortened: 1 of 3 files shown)


maliberty (approved):

maliberty:
Please resolve the clang-tidy

PR #5188, opened 2024-05-31
gui: Module view enables detailed

oharboe:
Module view doesn't do anything when detailed view isn't enabled, so enable detailed view when Module view is enabled.

This helps "market" this, in my opinion, undersold feature and improves usability a tad; enable it and it works without having to know that detailed view is needed.

This feature, much to my pleasant surprise, works even without hierarchical synthesis.

Megaboom:

image

### DIFF
--- a/src/gui/src/displayControls.cpp
+++ b/src/gui/src/displayControls.cpp
@@ -1651,7 +1651,7 @@ bool DisplayControls::areSelectedVisible()
 
 bool DisplayControls::isDetailedVisibility()
 {
-  return isModelRowVisible(&misc_.detailed);
+  return isModelRowVisible(&misc_.detailed) || isModelRowVisible(&misc_.module);
 }
 
 bool DisplayControls::arePrefTracksVisible()


src/gui/src/displayControls.cpp
```diff
@@ -1651,7 +1651,7 @@ bool DisplayControls::areSelectedVisible()
 
 bool DisplayControls::isDetailedVisibility()
 {
-  return isModelRowVisible(&misc_.detailed);
+  return isModelRowVisible(&misc_.detailed) || isModelRowVisible(&misc_.module);
```
gadfort, line 1654: I think the correct place to adjust this behavior would be in the renderthread.cpp. This causes the detailed view to be drawn even if the user has un-selected it.

oharboe: Could you rephrase/elaborate, I'm missing something that makes me not understand...

gadfort: without this do the memories get colored?
Looking at the code for drawing the module view: https://github.com/The-OpenROAD-Project/OpenROAD/blob/38a3b37e1a0bde4ef919c6a8bc4717516c3ad871/src/gui/src/renderThread.cpp#L1399C20-L1399C34
the instance list is controlled by: https://github.com/The-OpenROAD-Project/OpenROAD/blob/38a3b37e1a0bde4ef919c6a8bc4717516c3ad871/src/gui/src/renderThread.cpp#L1062
I think it would be cleaner to adjust the instance limit here, instead of changing the definition of detailed view.

oharboe: Hmm.... I'm out of my depth here. If we agree that this would be a nice feature, then it needs a 5 minute implementation by someone who knows what they are doing.

Of course nothing ever takes 5 minutes... but I'm thinking this is a small well defined task...

gadfort: I think it would be nice (I use the module view from time to time). I can take a stab at this sometime today

oharboe:
Nice idea, needs a better implementation.

PR #5194, opened 2024-06-03
gui: slack histogram filter fix

AcKoucher:
While working on improving performance I realized that there was a bug making us show the wrong slack data when using the filter feature. These changes are to ensure we actually show the correct data and reset the axis and limits to improve the presentation.

Changes:
- Use the API from the TimingPath rather than the sta API which gives always the worst slack
- Ensure that for tiny intervals of min/max slack, we use a reasonable slack interval for the x axis
- Clicking on empty bins won't trigger timing report (based on [Oyvind's Comment](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5139#issuecomment-2125400329))

### DIFF
--- a/src/gui/src/chartsWidget.cpp
+++ b/src/gui/src/chartsWidget.cpp
@@ -65,14 +65,12 @@ ChartsWidget::ChartsWidget(QWidget* parent)
       axis_y_(new QValueAxis(this)),
       buckets_(std::make_unique<Buckets>()),
       prev_filter_index_(0),  // start with no filter
-      max_slack_(0.0f),
-      min_slack_(std::numeric_limits<float>::max()),
       resetting_menu_(false),
       default_number_of_buckets_(15),
-      largest_slack_count_(0),
-      precision_count_(0),
+      max_slack_(0.0f),
+      min_slack_(std::numeric_limits<float>::max()),
       bucket_interval_(0.0f),
-      neg_count_offset_(0),
+      precision_count_(0),
 #endif
       label_(new QLabel(this))
 {
@@ -175,7 +173,9 @@ void ChartsWidget::showToolTip(bool is_hovering, int bar_index)
 
     QString scaled_suffix = sta_->units()->timeUnit()->scaledSuffix();
 
-    const float lower = (bar_index - neg_count_offset_) * bucket_interval_;
+    const int neg_count_offset = static_cast<int>(buckets_->negative.size());
+
+    const float lower = (bar_index - neg_count_offset) * bucket_interval_;
     const float upper = lower + bucket_interval_;
 
     QString time_info
@@ -197,8 +197,6 @@ void ChartsWidget::clearChart()
   // reset limits
   max_slack_ = 0;
   min_slack_ = std::numeric_limits<float>::max();
-
-  largest_slack_count_ = 0;
 
   chart_->setTitle("");
   chart_->removeAllSeries();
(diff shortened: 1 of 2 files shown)


AcKoucher:
@oharboe FYI

src/gui/src/chartsWidget.cpp
```diff
@@ -583,40 +637,58 @@ void ChartsWidget::setSTA(sta::dbSta* sta)
...
+  for (const std::unique_ptr<TimingPath>& path : paths) {
+    const float slack = time_unit->staToUser(path->getSlack());
+
+    if (slack < min_slack_) {
+      min_slack_ = slack;
+    }
+
+    if (slack > max_slack_) {
+      max_slack_ = slack;
+    }
```
eder-matheus, line 688: You can use std::min and std::max for these:
`min_slack_ = std::min(slack, min_slack_);`
`max_slack_ = std::max(slack, max_slack_);`

src/gui/src/chartsWidget.cpp
```diff
@@ -514,18 +550,36 @@ QString ChartsWidget::createXAxisTitle(const std::set<sta::Clock*>& clocks)
...
 {
-  int y_interval = computeYInterval();
+  int largest_slack_count = 0;
+
+  for (const std::vector<const sta::Pin*>& bucket : buckets_->negative) {
+    const int bucket_slack_count = static_cast<int>(bucket.size());
+
+    if (bucket_slack_count > largest_slack_count) {
+      largest_slack_count = bucket_slack_count;
+    }
```
eder-matheus, line 565: Use `std::max`.

src/gui/src/chartsWidget.cpp
```diff
@@ -514,18 +550,36 @@ QString ChartsWidget::createXAxisTitle(const std::set<sta::Clock*>& clocks)
...
+  }
+
+  for (const std::vector<const sta::Pin*>& bucket : buckets_->positive) {
+    const int bucket_slack_count = static_cast<int>(bucket.size());
+
+    if (bucket_slack_count > largest_slack_count) {
+      largest_slack_count = bucket_slack_count;
+    }
```
eder-matheus, line 573: Use `std::max`.

eder-matheus (changes requested):

oharboe:
First impressions:

- Register-register still takes much longer than no filter. Last time 2 hours. Still running 15 minutes later compared to 60 seconds or so with no filter.
- A progress indicator w/cancel would be good... Though I understand it can be trick to know how much of the work has been done.

oharboe:
I suppose it is the insert() that is the O(n^2), it is checking if the state is already in the set.

Here is a crazy idea based upon overconfidence after looking at the problem for 30 seconds: create a secure(i.e. guaranteed not to be duplicate, so 256-512 bits or something, I'm unsure exactly what is required) hash of all the IDs and then check if the hash is in the set when inserting. This requires calcluating the has of all the things being insert into this set once instead of comparing O(n^2).

https://github.com/The-OpenROAD-Project/OpenSTA/blob/e01d3f163f483f233db00410b6515a767a6ca03b/sdc/Sdc.cc#L5223

Looks like the ids are already sorted, from the looks of this comparison algorithm, so creating a guaranteed unique hash should not be so bad:

https://github.com/The-OpenROAD-Project/OpenSTA/blob/e01d3f163f483f233db00410b6515a767a6ca03b/network/Network.cc#L2140C1-L2149C6

AcKoucher:
@oharboe This PR is to fix the result of the current approach. The performance improvement will be made in a subsequent PR in a different way.

eder-matheus (approved):

PR #5198, opened 2024-06-04
mpl2: change internal dbu <-> microns conversion for the new methods from odb

AcKoucher:


AcKoucher:
I got Failure emails from All-Tests-Private and public_tests_small. However, it showed no failing designs, so I'm assuming the failure is from the Docker build. Based on that this should be good to merge.

eder-matheus (approved):

eder-matheus:
@vvbandeira This is ready to merge.

Issue #5197, opened 2024-06-04
Global Placement Crash


kareefardi:
### Describe the bug

Crash when calling `global_placement`. Crash is avoided by passing `-skip_nesterov_place`.

### Expected Behavior

Graceful exit or no crash

### Environment

```shell
[WARNING] Your current OpenROAD version is outdated.
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Linux 6.8.10-zabbly+
os: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
cmake version 3.25.1
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-13997-gad54bbe88
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.10 (found version "3.10.12") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds - Success
[...]

PR #5200, opened 2024-06-05
gpl: avoid crash by dividing to zero

gudeh:
This fix avoids a crash reported at issue #5197.


src/gpl/src/nesterovBase.cpp
```diff
@@ -602,12 +602,21 @@ void BinGrid::initBins()
...
       = static_cast<int64_t>(ux_ - lx_) * static_cast<int64_t>(uy_ - ly_);
 
-  int64_t averagePlaceInstArea
-      = pb_->placeInstsArea() / pb_->placeInsts().size();
+  int64_t averagePlaceInstArea = 0;
+  if (!pb_->placeInsts().empty()) {
+    averagePlaceInstArea = pb_->placeInstsArea() / pb_->placeInsts().size();
+  }
+
+  int64_t idealBinArea = 0;
```
QuantamHD, line 610: Is this a situation we want to recover from or send an error to the user? Separately, should placeInsts ever be empty at this point?

maliberty: If there are no instances to place then we should skip all of gpl. Does this design really have no instances to place?

gudeh: I do see instances placed at the beginning of GPL with debug mode. I will investigate further. Either way, I think the checking I have added so far does not hurt.

gudeh:
As mentioned in the issue #5197, the reason we had a PlacerBase object without any placed instances is because the code creates a PlacerBase object for each group and another one for instances not related to any group (the first one in `pbVec_` on replace.cpp). Consequently, along the algorithm we would end up with division by zero and NANs.

To solve this issue I erased the first position of `pbVec_` when there are no placed instances related to it.

I ran a secure-CI and it passes (there are some failings matching master).

What do you guys think?

src/gpl/src/nesterovBase.cpp
```diff
@@ -602,12 +602,23 @@ void BinGrid::initBins()
   int64_t totalBinArea
       = static_cast<int64_t>(ux_ - lx_) * static_cast<int64_t>(uy_ - ly_);
 
-  int64_t averagePlaceInstArea
-      = pb_->placeInstsArea() / pb_->placeInsts().size();
+  int64_t averagePlaceInstArea = 0;
+  if (!pb_->placeInsts().empty()) {
```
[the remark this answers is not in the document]
gudeh: I am not sure. I understand this protection to avoid a crash is only beneficial, right?

maliberty (approved):

src/gpl/src/replace.cpp
```diff
@@ -230,6 +230,10 @@ void Replace::doInitialPlace()
       }
     }
 
+    if (pbVec_.front()->placeInsts().empty()) {
+      pbVec_.erase(pbVec_.begin());
+    }
```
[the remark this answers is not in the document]
gudeh: I tried defining a third region in the TCL and the PDN provided this error:
image

PR #5202, opened 2024-06-05
global placement: make routability info message more actionable

### LINKED ISSUE #5201: Make `[INFO GPL-0075] Routability numCall: 5 inflationIterCnt: 0 bloatIterCnt: 1` more actionable
### Description

When global place is taking a long time, it is hard for the user to know what it is that is needed to speed it up based upon the info printed.

During architectural exploration, it is oftentimes more important to get an approximate result quickly than to get the best result. The user has to state his intention through modifying ORFS options to e.g. skip routability calls, but connecting the dots from the info below to the ORFS could be easier.

```
[INFO GPL-0075] Routability numCall: 5 inflationIterCnt: 0 bloatIterCnt: 1
```

### Suggested Solution

Change wording of this info to be more actionable. I know there is something that can be adjusted in ORFS to speed things up here, like disable placement w.r.t. routability?

### Additional Context

_No response_

oharboe:
fixes #5201

### DIFF
--- a/src/gpl/src/routeBase.cpp
+++ b/src/gpl/src/routeBase.cpp
@@ -887,7 +887,8 @@ void RouteBase::increaseCounter()
 
   log_->info(GPL,
              75,
-             "Routability numCall: {} inflationIterCnt: {} bloatIterCnt: {}",
+             "To speed up, remove '-routability_driven' option, numCall: {} "
+             "inflationIterCnt: {} bloatIterCnt: {}",
              numCall_,
              inflationIterCnt_,
              bloatIterCnt_);


oharboe:
@maliberty I'll fix the failures in testing if you think the change in info message makes sense.

gadfort:
@maliberty it seems like this will encourage users to remove it all together. I would think that making the message clearer for what is actually happening instead of just saying interaction x

oharboe:
@gadfort Idea for wording?

gadfort:
@oharboe no, because I have no clue what the numbers mean in that message. As an aside this message only works for the tcl interface and not python.
I don't think it makes sense ask users to remove a flag they selected (the tcl script that invoked this command could print a warning about runtime), I'd rather see the message become more informative.

oharboe:
@gadfort The user did not select this option, it is on by default in ORFS.

gadfort:
@oharboe then it seems like this message should be emitted in ORFS and not in OpenROAD. Since the OpenROAD default is off.

oharboe:
> @oharboe then it seems like this message should be emitted in ORFS and not in OpenROAD. Since the OpenROAD default is off.

This is a message that is enabled when the option is enabled and the option is enabled in ORFS.

Wouldnt another framework than ORFS that enables it by default also want the same info message?

gadfort:
@oharboe if you think this is critical to the useability of OpenROAD then it's fine (each framework uses a different control knob to turn this option on or off, so this is only useful to users using this interactively) (begs the question why not add it for timing driven too and any other option that has runtime implications). I do think that the disconnect from Python vs tcl is an issue.

gudeh:
Hey guys! The message @oharboe edited is showing how many times routability iterations have started. I think the suggestion of turning routability off could go somewhere else. 

I recently added some comments to README to help with that. 

>When the routability-driven option is enabled, each of its iterations will execute FastRoute to provide an estimation of routing congestion, which is typically costly in terms of runtime. Congested tiles will have the area of their logic cells inflated to reduce >routing congestion. The iterations will attempt to achieve the target RC (routing congestion) by comparing it to the final RC. If the >algorithm takes too long during routability-driven execution, consider raising the target RC value to alleviate the constraints.

A not so drastic approach would be to raise the target routing congestion, instead of turning routability off, as I suggested in the README.

Furthermore, I have an open PR #5124 to substitute fastroute (slow) for RUDY (fast) during routability, and this drastically reduces the routability runtime. Unfortunately we are getting some issues with private designs, I am working on it right now.

rovinski:
Agree that any message about turning off to speed up should be in the flow, not in the tool.
Also agree that the routability runtime penalty should greatly reduce after switching to RUDY.
Realistically the better approach is to comment in the docs if a flag increases or decreases runtime.

oharboe:
@rovinski When I see messages from OpenROAD, I wonder who it is trying to communicate what to and what the t

If the only target audience are developers of the code, then it is the

oharboe:
As to whether this information belongs in ORFS or OpenROAD: let's say we figure out and agree some information about tips on options belong in ORFS, how would the error message be put into ORFS?

gudeh:
I believe the challenge to define where to put such message lies in the fact that it is not necessarily an error that is happening, but a long runtime issue. DRT faces a similar issue with its iterations (they are costly in runtime), and it shows a time measurement after each iteration.

On the other hand, the idea is that when RUDY is incorporated, this problem should not happen, so maybe we should not take any action ATM?

rovinski:
I agree that the information in this message is not helpful and should explain what it's printing a bit better.
The "disable this option to improve runtime" is unrelated to the message that's printed and it really raises the question of why wouldn't this be on all commands or flags have this message? The solution is that it should be in the docs.
A solution that some other flows use is a very coarse-grained set of knob for flow-level runtime settings. Usually there's either "max speed" or "max quality" and sometimes even a "balanced" setting, but they often prove difficult to maintain or just simply too coarse to be useful for intermediate design stages.

Honestly I already have an idea for doing ML-based parameter tuning to optimize quality vs. runtime for the flow. A user would set the effort as a float between 0.0 (max speed) and 1.0 (max quality) and the ML model would translate the smooth continuum of effort into a discrete set of flags and parameters to use in the flow. But, I need a student to work on it.

oharboe:
Good discussion. For now await the RUDY upgrade. Perhaps this particular problem is moot at that point?

PR #5207, opened 2024-06-05
Cts prevent removing sinks

arthurjolo:
During the HTree building step of cts we can create more Clock subNets then needed, these extra Clock subNets were deleted at the end of CTS, but this can cause crashes (like it was mentioned in [PR#4938](https://github.com/The-OpenROAD-Project/OpenROAD/pull/4938)).

To prevent having to delete these nets at the end of CTS, this PR identifies and removes theses nets during the HTree building step.


src/cts/src/Clock.h
```diff
@@ -39,6 +39,7 @@
 #include <cstdint>
 #include <deque>
 #include <functional>
+#include <iostream>
```
eder-matheus, line 42: Do you need this include? I think you're not using nothing from iostream

src/cts/src/Clock.h
```diff
@@ -218,6 +223,44 @@ class Clock
     return subNets_.back();
   }
 
+  void removeSubNet(ClockSubNet sub_net)
+  {
+    std::set<ClockInst*> sinksToRemove;
+    sinksToRemove.insert(sub_net.getDriver());
+    for (auto it = subNets_.begin(); it != subNets_.end();) {
```
eder-matheus, line 230: What is the cost of iterating over every subnet? Wouldn't the runtime of it be too large?

src/cts/src/Clock.h
```diff
@@ -218,6 +223,44 @@ class Clock
...
+      } else {
+        ++it;
+      }
+    }
+    removeClockBuffer(sub_net.getDriver());
+  }
+
+  void removeClockBuffer(ClockInst* clock_buffer)
+  {
+    for (auto it = clockBuffers_.begin(); it != clockBuffers_.end();) {
```
eder-matheus, line 254: Similar question here. Isn't there a way to avoid these loops? Perhaps a hash table, a map...

eder-matheus (changes requested):

src/cts/src/Clock.h
arthurjolo: The run time is O(n) for the number of subNets, I don't think there is a better runtime for removing this an element from a std::deque. Also this code shouldn't be run very often, it is to fix some corner cases. But if the runtime becomes a problem we can maintain the subNet deletion at the end of CTS that would not affect the runtime.

arthurjolo:
This PR generated a crash in one testcases, and the approach to solve the problem had to change. So I will be closing this PR and open a new one with the correct fix.

Issue #5204, opened 2024-06-05
DPL-0026 legalPt called on fixed cell


kareefardi:
### Describe the bug

Not sure if this is a bug or not. I have a macro that I want to place inside a region. I don't have much experience with regions so I might be doing something wrong. If I remove the region and the group in the input def the error is not thrown

### Expected Behavior

Some guidance maybe?

### Environment

```shell
Git commit: 4fe54e666f3bf7120b9c22edd2087e537fe19a29
kernel: Linux 6.5.0-35-generic
os: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
cmake version 3.25.1
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-14016-g4fe54e666
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.10 (found version "3.10.12") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds - Success
[...]

PR #5209, opened 2024-06-06
dpl: don't touch fixed cells inside groups

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5204


maliberty (approved):
Looks good, do you see any other places that need the same?

eder-matheus:
> Looks good, do you see any other places that need the same?

This was the only place with the missing check in DPL. I inspected the code for possible causes of the issue, and the other places already checked if the Cell is a fixed inst.

Issue #5205, opened 2024-06-05
Global Placement Crash


kareefardi:
### Describe the bug

Crash when using `global_placement`

### Expected Behavior

Graceful exit

### Environment

```shell
Git commit: 4fe54e666f3bf7120b9c22edd2087e537fe19a29
kernel: Linux 6.9.3-zabbly+
os: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
cmake version 3.25.1
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-14016-g4fe54e666
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.10 (found version "3.10.12") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull - Success
[...]

PR #5210, opened 2024-06-06
dpl: check if db_inst_ exists before accessing it

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5205

### DIFF
--- a/src/dpl/src/Place.cpp
+++ b/src/dpl/src/Place.cpp
@@ -1364,7 +1364,7 @@ DbuPt Opendp::legalPt(const Cell* cell, const bool padded) const
     // If that didn't do the job fall back on the old move to nearest
     // edge strategy.  This doesn't consider site availability at the
     // end used so it is secondary.
-    if (block && isBlock(block)) {
+    if (block && block->db_inst_ && isBlock(block)) {
       const Rect block_bbox(block->x_.v,
                             block->y_.v,
                             block->x_.v + block->width_.v,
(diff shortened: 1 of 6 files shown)


eder-matheus:
@maliberty I found that the db_inst being nullptr was the cause of the issue, but I couldn't see why it is nullptr at this point. The position of the Cell that is returned by the pixel doesn't have any cell on it.

maliberty (approved):
Let's take this as a quick fix and dig in more afterwards

maliberty:
Which issue is this related to?

eder-matheus:
> Which issue is this related to?

This issue: https://github.com/The-OpenROAD-Project/OpenROAD/issues/5205
The only difference I first noticed is using REGION and GROUP, but we don't have this specific crash in other test cases from the same user.

maliberty:
The cell is the dummy_cell used by groupInitPixels and groupInitPixels2.  It is used to block the area outside the region during placement of the group.  I think it is better to move the check into isBlock itself.  

isBlock should be a method of Cell not Opendp.

src/dpl/src/Place.cpp
```diff
@@ -1364,7 +1364,7 @@ DbuPt Opendp::legalPt(const Cell* cell, const bool padded) const
     // If that didn't do the job fall back on the old move to nearest
     // edge strategy.  This doesn't consider site availability at the
     // end used so it is secondary.
-    if (block && isBlock(block)) {
+    if (block && block->db_inst_ && isBlock(block)) {
```
maliberty, line 1367: Move the check to isBlock

maliberty (changes requested):

src/dpl/src/Place.cpp
eder-matheus: Done, both moving the check to isBlock and move isBlock to Cell struct.

maliberty (approved):

PR #4636, opened 2024-02-07
grt: fastroute: maze.cpp: Parallelized getOverflow3D with OpenMP

glatosinski:
This PR:

* introduces usage of OpenMP to `grt::FastRouteCore::getOverflow3D` method to run overflow computation in parallel
* removes unused `max_H_overflow` and `max_V_overflow` computation

The comparison of speed (checked on 16 threads machine):

| grt        | master                      | PR                          |
|------------|-----------------------------|-----------------------------|
| ariane     | 9:20 (min: 9:11, max: 9:36) | 8:51 (min: 8:44, max: 8:47) |
| tinyRocket | 0:25 (min: 0:25, max: 0:25) | 0:22 (min: 0:22, max: 0:22) |
| ibex       | 0:43 (min: 0:42, max: 0:43) | 0:42 (min: 0:42, max: 0:42) |

PR #5215, opened 2024-06-09
grt: parallel getOverflow3D

eder-matheus:
Completing PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/4636


src/grt/src/fastroute/src/maze.cpp
```diff
@@ -2447,32 +2449,25 @@ int FastRouteCore::getOverflow2D(int* maxOverflow)
...
-    }
+  int kstep = y_grid_ * x_grid_;
+  int istep = x_grid_;
+
+#pragma omp parallel for schedule(static) reduction(+ : H_overflow) \
+    reduction(+ : V_overflow) reduction(+ : total_usage)
+  for (int iter = 0; iter < num_layers_ * y_grid_ * x_grid_; iter++) {
+    int k = iter / kstep;
+    int i = (iter % kstep) / istep;
+    int j = (iter % kstep) % istep;
```
maliberty, line 2464: It think it would be simpler to keep the original loop structure and apply `collapse(2)` to the prama.  The original loop was over h_used_ggrid_ rather than all points.

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -2447,32 +2449,25 @@ int FastRouteCore::getOverflow2D(int* maxOverflow)
...
+    int overflow = h_edges_3D_[k][i][j].usage - h_edges_3D_[k][i][j].cap;
+    H_overflow += overflow > 0 ? overflow : 0;
```
maliberty, line 2466: I find
```
      const auto& h_edge = h_edges_3D_[k][i][j];
      H_overflow += std::max(h_edge.usage - h_edge.cap, 0);
```
clearer.  Likewise V

maliberty (changes requested):

PR #5222, opened 2024-06-10
cts avoid removing sinks

arthurjolo:
During the HTree building step of cts we can create more Clock subNets then needed, these extra Clock subNets were deleted at the end of CTS, but this can cause crashes (like it was mentioned in https://github.com/The-OpenROAD-Project/OpenROAD/pull/4938).

To prevent having to delete these nets at the end of CTS, this PR avoids creating subnets that aren't connected to anything.

### DIFF
--- a/src/cts/src/HTreeBuilder.cpp
+++ b/src/cts/src/HTreeBuilder.cpp
@@ -1776,10 +1776,10 @@ void HTreeBuilder::createClockSubNets()
   bool isFirstPoint = true;
   topLevelTopology.forEachBranchingPoint([&](unsigned idx,
                                              Point<double> branchPoint) {
-    
-    // If the branch point is a leaf and has no sinks that will be connected to it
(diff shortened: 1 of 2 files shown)


src/cts/src/HTreeBuilder.cpp
```diff
@@ -1776,6 +1776,13 @@ void HTreeBuilder::createClockSubNets()
   bool isFirstPoint = true;
   topLevelTopology.forEachBranchingPoint([&](unsigned idx,
                                              Point<double> branchPoint) {
+    // If the branch point is a leaf and has no sinks that will be connected to
+    // it
+    // don't create a clock sub net for it
```
maliberty, line 1781: Fix the awkward line break (same below)

maliberty (changes requested):
Looks good otherwise.

maliberty (approved):

PR #5227, opened 2024-06-11
ant: parallelize check_antennas

eder-matheus:


eder-matheus:
@maliberty Secure-ci is running.

src/ant/src/DSU.hh
```diff
@@ -0,0 +1,54 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <vector>
+
+namespace ant {
+
+class DSU
```
maliberty, line 38: What is DSU?

maliberty, line 38: Could boost's disjoint_sets replace this?

eder-matheus: I'll try!

src/grt/src/RepairAntennas.cpp
```diff
@@ -94,7 +97,10 @@ bool RepairAntennas::checkAntennaViolations(
 
   makeNetWires(routing, nets_to_repair, max_routing_layer);
   arc_->initAntennaRules();
-  for (odb::dbNet* db_net : nets_to_repair) {
+  omp_set_num_threads(num_threads);
+#pragma omp parallel for schedule(dynamic)
+  for (auto it = nets_to_repair.begin(); it != nets_to_repair.end(); ++it) {
+    odb::dbNet* db_net = *it;
     checkNetViolations(db_net, diode_mterm, ratio_margin);
   }
```
eder-matheus, line 105: @maliberty I added parallelization here, but I'm facing some strange non-determinism in the results. Testing with sky130hd/chameleon, the number of antenna violations reported varies slightly across different runs.

I've made additional changes in the antenna_checker to prevent non-determinism, but now I don't see any other possible sources for it. The only thing I can think of is the access to the antenna_violations_ map that is made in parallel. However, we access it using the dbNet pointer as key, and we should not have duplicated nets in this iteration. Do you have some insights on why this is happening?

maliberty: Is this map pre-populated (all the keys are present)?  If not you are doing an insertion which is not safe.

eder-matheus: Oh, ok. This is the problem then. I will initialize the map with the keys before going to the parallel part.

src/ant/src/AntennaChecker.cc
```diff
@@ -1125,24 +1103,34 @@ int AntennaChecker::checkAntennas(odb::dbNet* net, bool verbose)
...
-                 net_violation_count,
-                 pin_violation_count);
+        nets_.push_back(net);
       }
     }
+    omp_set_num_threads(num_threads);
+#pragma omp parallel for schedule(dynamic)
+    for (int i = 0; i < nets_.size(); i++) {
+      odb::dbNet* net = nets_[i];
+      ViolationList antenna_violations;
```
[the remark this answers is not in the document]
eder-matheus: Can't do this due to omp parallelization.

src/ant/src/DSU.hh
eder-matheus: Done, much better now!

eder-matheus:
@maliberty I checked the speedup on repair_antennas with sky130hd with chameleon, and the results are good. Using 16 threads, it went from 13 sec to 7sec, almost 50% of reduction (only repair_antennas). I will investigate more in the future to check what else can be parallelized in repair_antennas and GRT.

src/grt/src/RepairAntennas.cpp
```diff
@@ -76,13 +78,15 @@ bool RepairAntennas::checkAntennaViolations(
...
     int max_routing_layer,
     odb::dbMTerm* diode_mterm,
-    float ratio_margin)
+    float ratio_margin,
+    const int num_threads)
 {
   // safe copy net wires before orderWires calls
   // TODO: remove this copy when antenna checker update is done
   std::map<odb::dbNet*, std::pair<std::vector<int>, std::vector<unsigned char>>>
       copy_wires;
```
maliberty, line 87: Should this be removed now?

src/ant/include/ant/AntennaChecker.hh
```diff
@@ -122,6 +118,13 @@ struct Violation
   int diode_count_per_gate;
 };
 
+using LayerInfoMap = std::map<odb::dbTechLayer*, NodeInfo>;
+using GraphNodeVector = std::vector<GraphNode*>;
+using GraphNodeVectorMap
+    = std::unordered_map<odb::dbTechLayer*, GraphNodeVector>;
+using GateInfoMap = std::map<std::string, LayerInfoMap>;
+using ViolationList = std::vector<Violation>;
```
maliberty, line 126: GraphNodeVector is a vector and so is ViolationList.  I find that confusing.  We don't have a standard naming scheme here (and I don't see one in the Google style guide).  

My personal preference is to using a plural for simple collections, eg GraphNodes and "To" for mapping, eg LayerToNodeInfo.  You can combine it for LayerToGraphNodes (replacing GraphNodeVectorMap).  That way you don't have to rework the code if you change a vector to a deque or a map to an unordered_map, etc.

maliberty (approved):
Please fix the remaining clang-tidy

eder-matheus:
> Please fix the remaining clang-tidy

This can't be resolved since I need to use the "ugly" for-loop to have the parallelism working in all compilers.

PR #5234, opened 2024-06-12
Enforce iteration order determinism for ppl::PinSet.

mikesinouye:
We are observing non-determinism in the output of the `place_pins` step across multiple runs of OpenROAD with fixed initial conditions. The root cause seems to be iteration over a set of `odb::dbBTerm*`.

With this change, we observe deterministic pin placement. Using `dbBTerm.getId()` for the comparison was considered but we suggest using `getObjNames()` to avoid being affected by upstream allocation order.


maliberty (approved):

eder-matheus:
@mikesinouye Thanks for the contribution! Could you update the PPL unit tests as well? They're failing in our CI, which prevents the PR from merging.

maliberty:
getObjName is wrong, I suppose you meant getName.  However getId is relied on elsewhere so I see no reason to be different here.

PR #5235, opened 2024-06-13
rsz: consider vias for drvr_route_pt when net is local

eder-matheus:


eder-matheus:
@maliberty CI is passing.

maliberty (approved):

PR #5242, opened 2024-06-13
ant: return CAR and CSR violations to repair_antennas

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1981

This is an initial implementation, as mentioned in a comment in the code. Using this naive approach, we can reduce the number of nets with violations from 81 to 55 in ihp-sg13g2/riscv32i.

### DIFF
--- a/src/ant/src/AntennaChecker.cc
+++ b/src/ant/src/AntennaChecker.cc
@@ -740,122 +740,70 @@ bool AntennaChecker::checkCSR(odb::dbTechLayer* tech_layer,
   return violation;
 }
 
-int AntennaChecker::checkGates(odb::dbNet* db_net,
-                               bool verbose,
-                               bool report_if_no_violation,
-                               std::ofstream& report_file,
-                               odb::dbMTerm* diode_mterm,
-                               float ratio_margin,
-                               GateToLayerToNodeInfo& gate_info,
-                               Violations& antenna_violations)
+bool AntennaChecker::checkRatioViolations(odb::dbTechLayer* layer,
+                                          const NodeInfo& node_info,
+                                          bool verbose,
+                                          bool report,
+                                          std::ofstream& report_file)
 {
-  ratio_margin_ = ratio_margin;
-  int pin_violation_count = 0;
+  bool node_has_violation = false;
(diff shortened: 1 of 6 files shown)


src/ant/src/AntennaChecker.cc
```diff
@@ -740,6 +740,92 @@ bool AntennaChecker::checkCSR(odb::dbTechLayer* tech_layer,
...
+  node_has_violation = par_violation.first || car_violation;
+
+  if (layer->getRoutingLevel() != 0) {
+    auto psr_violation
+        = checkPSR(layer, node_info, verbose, report, report_file);
+    bool csr_violation
+        = checkCSR(layer, node_info, verbose, report, report_file);
+    node_has_violation
+        = node_has_violation || psr_violation.first || csr_violation;
+  }
```
maliberty, line 761: Does check* have a side-effect or should we skip this if node_has_violation was already set by PAR/CAR?

src/ant/src/AntennaChecker.cc
```diff
@@ -740,6 +740,92 @@ bool AntennaChecker::checkCSR(odb::dbTechLayer* tech_layer,
...
+bool AntennaChecker::checkRatioViolations(odb::dbTechLayer* layer,
+                                          const NodeInfo& node_info,
+                                          bool verbose,
+                                          bool report,
+                                          std::ofstream& report_file)
+{
+  bool node_has_violation = false;
+  auto par_violation = checkPAR(layer, node_info, verbose, report, report_file);
+  bool car_violation = checkCAR(layer, node_info, verbose, report, report_file);
+  node_has_violation = par_violation.first || car_violation;
```
maliberty, line 752: There's no need to define node_has_violation and then set it afterwards, just combine 749 & 752

maliberty (approved):
Minor cleanup before merging

src/ant/src/AntennaChecker.cc
eder-matheus: We should not skip because a net can have both PAR/CAR and PSR/CSR. Skipping it will give us the wrong number of violations and make the report incorrect.

PR #5243, opened 2024-06-14
[Docs] Docs CI fix

luarss:
Fixed a missing flag in readme @maliberty


maliberty (approved):

maliberty:
@gudeh it would be good to add a description in a separate PR

gudeh:
> @gudeh it would be good to add a description in a separate PR

For sure. Sorry I forgot the readme.

I will also fix another detail. Currently routability is being set on through TCL. I understand we want the actual default to be routability on, and only use the TCL command to turn it off when desired.

PR #5246, opened 2024-06-14
Initial placement-aware scan stitching

gatecat:
This isn't quite complete as I haven't updated tests yet, but I'd appreciate feedback on the direction here before I do that.

This PR makes two changes to the DFT code:
 
 - `insert_dft` is now only performing scan cell replacement, a new command `connect_dft` is used to perform scan architect and stitch. This is because replacement needs to be done before placement (as it affects cell areas) but stitching needs to be done after placement (where it can use placement information to minimise scan wirelength)
 - A basic wirelength minimising sort is used on the scan chain before stitching. This is a simple greedy nearest neighbour approach using Boost rtrees - I also briefly explored using the or-tools TSP solver a bit but all the modes I tried were considerably too slow (and the only way to improve this seemed to be by setting a timeout, which is obviously a no-go due to nondeterminism).

After this PR, I plan to work on the scan chain connectivity, so it's easier to create scan ports ahead of time or connect the scan chain to existing signals (like top level IO pad cells, if not using a macro based flow). Any feedback on what might be wanted here is appreciated, too.

### DIFF
--- a/src/dft/src/architect/ScanArchitectHeuristic.cpp
+++ b/src/dft/src/architect/ScanArchitectHeuristic.cpp
@@ -36,14 +36,16 @@
 #include <sstream>
 
 #include "ClockDomain.hh"
+#include "Opt.hh"
 #include "ScanArchitect.hh"
 
 namespace dft {
 
 ScanArchitectHeuristic::ScanArchitectHeuristic(
     const ScanArchitectConfig& config,
-    std::unique_ptr<ScanCellsBucket> scan_cells_bucket)
-    : ScanArchitect(config, std::move(scan_cells_bucket))
+    std::unique_ptr<ScanCellsBucket> scan_cells_bucket,
+    utl::Logger* logger)
+    : ScanArchitect(config, std::move(scan_cells_bucket)), logger_(logger)
 {
 }
 
@@ -61,12 +63,14 @@ void ScanArchitectHeuristic::architect()
         current_chain->add(std::move(scan_cell));
       }
 
-      // TODO: Perform an actual sorting to reduce the wire length
       current_chain->sortScanCells(
-          [](std::vector<std::unique_ptr<ScanCell>>& falling,
-             std::vector<std::unique_ptr<ScanCell>>& rising,
-             std::vector<std::unique_ptr<ScanCell>>& sorted) {
+          [this](std::vector<std::unique_ptr<ScanCell>>& falling,
+                 std::vector<std::unique_ptr<ScanCell>>& rising,
+                 std::vector<std::unique_ptr<ScanCell>>& sorted) {
             sorted.reserve(falling.size() + rising.size());
+            // Sort to reduce wire length
+            OptimizeScanWirelength(falling, logger_);
+            OptimizeScanWirelength(rising, logger_);
             // Falling edge first
             std::move(
                 falling.begin(), falling.end(), std::back_inserter(sorted));
--- a/src/dft/test/cpp/TestScanArchitectHeuristic.cpp
+++ b/src/dft/test/cpp/TestScanArchitectHeuristic.cpp
@@ -35,8 +35,8 @@ TEST(TestScanArchitectHeuristic, ArchitectWithOneClockDomainNoMix)
   scan_cells_bucket->init(config, scan_cells);
 
   std::unique_ptr<ScanArchitect> scan_architect
-      = ScanArchitect::ConstructScanScanArchitect(config,
-                                                  std::move(scan_cells_bucket));
+      = ScanArchitect::ConstructScanScanArchitect(
+          config, std::move(scan_cells_bucket), logger);
   scan_architect->init();
   scan_architect->architect();
   std::vector<std::unique_ptr<ScanChain>> scan_chains
@@ -99,8 +99,8 @@ TEST(TestScanArchitectHeuristic, ArchitectWithTwoClockDomainNoMix)
   scan_cells_bucket->init(config, scan_cells);
 
   std::unique_ptr<ScanArchitect> scan_architect
-      = ScanArchitect::ConstructScanScanArchitect(config,
-                                                  std::move(scan_cells_bucket));
+      = ScanArchitect::ConstructScanScanArchitect(
+          config, std::move(scan_cells_bucket), logger);
   scan_architect->init();
   scan_architect->architect();
   std::vector<std::unique_ptr<ScanChain>> scan_chains
@@ -156,8 +156,8 @@ TEST(TestScanArchitectHeuristic, ArchitectWithTwoEdgesNoMix)
   scan_cells_bucket->init(config, scan_cells);
 
   std::unique_ptr<ScanArchitect> scan_architect
-      = ScanArchitect::ConstructScanScanArchitect(config,
-                                                  std::move(scan_cells_bucket));
+      = ScanArchitect::ConstructScanScanArchitect(
+          config, std::move(scan_cells_bucket), logger);
   scan_architect->init();
   scan_architect->architect();
   std::vector<std::unique_ptr<ScanChain>> scan_chains
--- a/src/dft/README.md
+++ b/src/dft/README.md
@@ -45,11 +45,21 @@ Prints the current DFT configuration to be used by `preview_dft` and
 report_dft_config
 ```
 
+### Scan replace
+
+Replaces flipflops with equivalent scan flipflops. This will generally be called before
+placement, as it changes the area of cells.
+
+```tcl
+scan_replace
+```
+
 ### Preview DFT
 
 Prints a preview of the scan chains that will be stitched by `insert_dft`. Use
-this command to iterate and try different DFT configurations. This command do
-not perform any modification to the design.
+this command to iterate and try different DFT configurations. This command does
+not perform any modification to the design, and should be run after `scan_replace`
+and global placement.
 
 ```tcl
 preview_dft
@@ -64,19 +74,13 @@ preview_dft
 
 ### Insert DFT
 
-Implements the scan chains into the design by performing the following actions:
-
-1. Scan Replace.
-2. Scan Architect.
-3. Scan Stitch.
-
-The end result will be a design with scan flops connected to form the scan
-chains.
-
+Architect scan chains and connect them up in a way that minimises wirelength. As a result, this
+should be run after placement, and after `scan_replace`.
 
 ```tcl
 insert_dft
 ```
+
 
 ## Example scripts
 
@@ -86,6 +90,8 @@ scan flops in the scan chains.
 ```
 set_dft_config -max_length 10 -clock_mixing clock_mix
 report_dft_config
+scan_replace
+# Run global placement...
 preview_dft -verbose
 insert_dft
 ```
--- a/src/dft/include/dft/Dft.hh
+++ b/src/dft/include/dft/Dft.hh
@@ -84,7 +84,7 @@ class Dft
   //
   // If verbose is true, then we show all the cells that are inside the scan
   // chains
-  void preview_dft(bool verbose);
+  void previewDft(bool verbose);
 
   // Inserts the scan chains into the design. For now this just replace the
   // cells in the design with scan equivalent. This functions mutates the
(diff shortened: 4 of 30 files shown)


maliberty:
@fgaray would you please comment

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -38,15 +38,84 @@
...
 ScanArchitectHeuristic::ScanArchitectHeuristic(
     const ScanArchitectConfig& config,
-    std::unique_ptr<ScanCellsBucket> scan_cells_bucket)
-    : ScanArchitect(config, std::move(scan_cells_bucket))
+    std::unique_ptr<ScanCellsBucket> scan_cells_bucket, utl::Logger* logger)
+    : ScanArchitect(config, std::move(scan_cells_bucket)), logger_(logger)
 {
 }
 
+void ScanArchitectHeuristic::order_reduce_wirelength(std::vector<std::unique_ptr<ScanCell>>& cells)
```
QuantamHD, line 57: method should follow Google style order_reduce_wirelength -> OrderReduceWirelength.

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -38,15 +38,84 @@
...
+{
+
+  // Nothing to order
+  if (cells.empty())
+    return;
+  // No point running this if the cells aren't placed yet
+  if (std::all_of(cells.begin(), cells.end(), [](const std::unique_ptr<ScanCell> &cell) {
+        return !cell->isPlaced();
+      }))
+    return;
```
QuantamHD, line 67: I do generally like std::all_of, but in this case I think 

```c++
for (const auto& cell: cells) {
  if (!cell->isPlaced()) {
     return;
  }
}
```

would be a little bit clearer.

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -38,15 +38,84 @@
...
+  // Define the starting node as the upperleftmost
+  size_t start_index = 0;
+  int64_t lowest_dist = cells.at(0)->getOrigin().x() + cells.at(0)->getOrigin().y();
```
QuantamHD, line 70: The coordinates of odb allow for negative locations, I don't think code accounts for that with -inf, inf being the upper leftmost point. In practice the coordinates are int_32t so inf = max_int32_t

src/dft/src/architect/ScanArchitectHeuristic.cpp
QuantamHD: To represent the functionality better I would suggest OptimizeScanWirelength, but that's more of a nit.

fgaray:
Thanks for working on this! Some high level comments:

>insert_dft is now only performing scan cell replacement, a new command connect_dft is used to perform scan architect and stitch. This is because replacement needs to be done before placement (as it affects cell areas) but stitching needs to be done after placement (where it can use placement information to minimise scan wirelength)

I think this is a good idea but I would propose a change to be more "standard". I would create a standalone command called "scan_replace" to perform the scan replacement and leave architect and stitching to insert_dft.

Similar to other tools, scan_replace is usually performed in a command like "compile -scan" and insert_dft performs architect and stitching.

>A basic wirelength minimising sort is used on the scan chain before stitching. This is a simple greedy nearest neighbour approach using Boost rtrees - I also briefly explored using the or-tools TSP solver a bit but all the modes I tried were considerably too slow (and the only way to improve this seemed to be by setting a timeout, which is obviously a no-go due to nondeterminism).

That's my experience too. Besides what you tried, I would see if we can get some kind of generic heuristic inside OpenROAD, a genetic algorithm would be a good thing to try. Also, I would not try to optimize that much the result of insert_dft as we can always do a second optimization stage later in the flow once we have more info about where the cells are really going to be placed.

I will review the code itself now.

fgaray:
I guess if you go with my suggestion the code will change. Let me know once this is ready for another look or if my suggestions do not make sense. Thanks!

gatecat:
@fgaray Thanks for your feedback, I've implemented your recommendations and I think it's ready for another review.

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -32,21 +32,97 @@
...
+  while (true) {
+    result.emplace_back(std::move(cells.at(cursor.second)));
+    if (result.size() >= cells.size()) {
+      break;
+    }
+    // Search for next nearest
+    auto it = rtree.qbegin(bgi::nearest(cursor.first, kMaxCellsToSearch));
+    auto next = cursor;
+    while (it != rtree.qend()) {
+      next = *it++;
```
fgaray, line 107: This is creating an implicit temporal copy of the iterator. Please try to avoid postfix ++ as much as possible.

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -61,12 +137,14 @@ void ScanArchitectHeuristic::architect()
         current_chain->add(std::move(scan_cell));
       }
 
-      // TODO: Perform an actual sorting to reduce the wire length
       current_chain->sortScanCells(
-          [](std::vector<std::unique_ptr<ScanCell>>& falling,
-             std::vector<std::unique_ptr<ScanCell>>& rising,
-             std::vector<std::unique_ptr<ScanCell>>& sorted) {
+          [this](std::vector<std::unique_ptr<ScanCell>>& falling,
```
fgaray, line 141: I think if you convert OptimizeScanWireLength to a function you will not need to pass this anymore

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -32,21 +32,97 @@
...
     const ScanArchitectConfig& config,
-    std::unique_ptr<ScanCellsBucket> scan_cells_bucket)
-    : ScanArchitect(config, std::move(scan_cells_bucket))
+    std::unique_ptr<ScanCellsBucket> scan_cells_bucket,
+    utl::Logger* logger)
+    : ScanArchitect(config, std::move(scan_cells_bucket)), logger_(logger)
 {
 }
 
+void ScanArchitectHeuristic::OptimizeScanWirelength(
```
fgaray, line 57:
```suggestion
void ScanArchitectHeuristic::OptimizeScanWireLength(
```

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -32,21 +32,97 @@
...
+  }
+  // Search nearest neighbours
+  std::vector<std::unique_ptr<ScanCell>> result;
+  // Max number of cells we assume could be placed at the same point (hopefully,
+  // after DP, this is never more than one)
+  const int kMaxCellsToSearch = 10;
+
+  while (true) {
+    result.emplace_back(std::move(cells.at(cursor.second)));
+    if (result.size() >= cells.size()) {
```
fgaray, line 100: I think you can use this condition inside the while so we don't need to do a while (true) anymore.

Also, maybe you want to get the original size of the cells vector before so the ending condition is when the new vector has the same size.

src/dft/src/cells/OneBitScanCell.hh
```diff
@@ -59,6 +59,9 @@ class OneBitScanCell : public ScanCell
   void connectScanOut(const ScanLoad& load) const override;
   ScanDriver getScanOut() const override;
 
+  odb::Point getOrigin() const override;
```
fgaray, line 62: odb::Point is a class, I think it is a good idea to return a const reference to it:

```suggestion
  const odb::Point& getOrigin() const override;
```

src/dft/test/cpp/TestScanArchitectHeuristic.cpp
```diff
@@ -35,8 +35,8 @@ TEST(TestScanArchitectHeuristic, ArchitectWithOneClockDomainNoMix)
   scan_cells_bucket->init(config, scan_cells);
 
   std::unique_ptr<ScanArchitect> scan_architect
-      = ScanArchitect::ConstructScanScanArchitect(config,
-                                                  std::move(scan_cells_bucket));
+      = ScanArchitect::ConstructScanScanArchitect(
+          config, std::move(scan_cells_bucket), nullptr);
```
fgaray, line 39: You have the logger available in line 15, no need to pass nullptr here.

fgaray (commented):
Thanks for the changes!, please take a look at the new comments.

Also, did you have a chance to try this with a bigger design?

In the test/ top level folder there is a design called ./jpeg_sky130hs.tcl that you can use to benchmark your algorithm, I am curious to see how it performs there.

You could do a HEAD vs your changes test and report back the runtime.

Thanks!

gatecat:
I'll get back to your other comments in a bit, but just as an update - I've been looking at the benchmarking and the good news is that with these changes, there does appear to be a noticeable drop in routed wirelength - from 2260771um (HEAD) to 1881677um (scan). The reordering doesn't seem to be a significant part of runtime, as far as I can see.

However, weirdly, which is what I'm currently trying to investigate (it could be that I messed something up with my checks), is that the with the reordering, antenna diode insertion takes a ridiculously long amount of time, increasing the overall runtime by more than double - even though with the reordered chains (and therefore less wirelength), there are fewer diodes to insert.

gatecat:
Ah, found why, I was running DFT insertion in the reordering case after all placement for the reordering, which meant that the newly inserted scan enable signal wasn't getting buffered by repair. Moving it to before repair means the runtime is essentially the same as before, maybe 30 seconds faster (10 minutes +/- noise in each case). The wirelength comparison above doesn't change significantly as a result either.

src/dft/src/cells/OneBitScanCell.hh
gatecat: The idiom in OpenROAD elsewhere (e.g. the `dbInst::getPoint` method it wraps) is to just pass it by value, being a 64 bit object (2 integers) anyway there's no real benefit to using a reference.

src/dft/src/architect/ScanArchitectHeuristic.cpp
```diff
@@ -32,21 +32,97 @@
...
+  // No point running this if the cells aren't placed yet
+  for (const auto& cell : cells) {
+    if (!cell->isPlaced()) {
+      return;
+    }
+  }
+  // Define the starting node as the upperleftmost
+  size_t start_index = 0;
+  int64_t lowest_dist = std::numeric_limits<int64_t>::max();
+  for (size_t i = 0; i < cells.size(); i++) {
```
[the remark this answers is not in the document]
gatecat: I've added a bit of a comment to the code - we need to find an (arbitrary) starting point, and this finds the cell with the lowest manhattan distance from (0, 0) for that.

src/dft/src/architect/ScanArchitectHeuristic.cpp
gatecat: wirelength is normally a single word in EDA contexts, so I think this okay

maliberty:
@fgaray @gatecat any more concerns before final review/merge?

fgaray:
The code LG Thanks for working on it!, I will leave this bit of argument to you @maliberty : 
image

@gatecat says that in other parts of the code we return a copy (and it is only two ints) but I think we should return Point as a reference. I will leave the final decision to you (@maliberty ) if you think we should return a reference here or use the semantics of the rest of the code.

One last bit, could it be possible to add a test case with physical information?, the current test cases are not really testing your code as they are not doing global placement, is it possible to include such case? (or modify one or all the test cases that we have in dft/test).

That's all from my side!, thanks!

maliberty:
godbolt says it doesn't matter whether you return by reference or by value so the current code is fine (with #5279)

maliberty (changes requested):

maliberty:
I agree about a physical test cast.  I don't think it is necessary to run global placement but I do think something that shows the scan ordering being driven by WL would be good.

gatecat:
Thanks for the review comments. I've addressed them, and added a test that manually places 10 cells and checks they are sorted according to the placement.

maliberty:
In the future please avoid using a force push after review as it prevents use of the GH feature to see changes since the last review.

maliberty (approved):

maliberty:
Thanks @gatecat !

PR #5251, opened 2024-06-17
Added sta registered commands tracing and tests

dav77227:
The solution is implemented using the Tcl_CreateObjTrace function. It performs tracing for all execution modes: prompt, batch and GUI modes. Only OpenROAD registered commands are being logged. 
During the testing I have encountered that there are some commands which are logged twice, and I have not yet found the exact reason of this behavior, so I have applied a workaround solution to this issue (maintained the last executed command). @maliberty @QuantamHD @gadfort @stefanottili because of this workaround, this PR is currently a draft, so please let me know if you have a clue what might be the reason of double logging?

Please see the sample log files attached with and without the workaround
[2_2_floorplan_io.log without the workaround](https://github.com/user-attachments/files/15832890/2_2_floorplan_io.log)
[2_2_floorplan_io.log with the workaround](https://github.com/user-attachments/files/15865791/2_2_floorplan_io.log)


maliberty:
> create_clock -name core_clock -period 0.46 _c08bf46215630000_p_Port

It isn't ideal to include the swig'ed pointer instead of the name as it won't be reproducible in another run.

maliberty:
When running ng45/gcd I see in the sdc:
```
current_design gcd

set clk_name  core_clock
set clk_port_name clk
set clk_period 0.46 
set clk_io_pct 0.2

set clk_port [get_ports $clk_port_name]
```
while in the trace
```
OpenROAD> current_design gcd 
OpenROAD> get_name _e095730700000000_p_Cell 
OpenROAD> get_ports clk 
```
The call to `get_name` is internal to `current_design` and should not be logged despite being a public command (it wasn't directly issued by the user).

dav77227:
> When running ng45/gcd I see in the sdc:
> 
> ```
> current_design gcd
> 
> set clk_name  core_clock
> set clk_port_name clk
> set clk_period 0.46 
> set clk_io_pct 0.2
> 
> set clk_port [get_ports $clk_port_name]
> ```
> 
> while in the trace
> 
> ```
> OpenROAD> current_design gcd 
> OpenROAD> get_name _e095730700000000_p_Cell 
> OpenROAD> get_ports clk 
> ```
> 
> The call to `get_name` is internal to `current_design` and should not be logged despite being a public command (it wasn't directly issued by the user).

The tcl_CreateObjTrace function that I used for tracing has a level argument, which makes it possible to trace commands based on their nesting level. A value of 0 means that commands at any level are traced. A value of 1 means only top-level commands are traced, 2 means top-level commands or those invoked as immediate consequences of executing top-level commands, and so on.

As an experiment I have printed all commands and all levels and discovered that if we run design through flow scripts (batch mode), then the command `current_design gcd` in the `ng45/gcd 1_synth.sdc` file, is at level 20, and `get_name _e095730700000000_p_Cell` is at level 23. By filtering logging of traced commands where the level is 20 or less, we can log only user issued commands. However, in case of interactive mode level 20 is too much, so I think we need to do extensive testing to find correct levels for every mode. 

@maliberty Please let me know what you think.

dav77227:
> > create_clock -name core_clock -period 0.46 _c08bf46215630000_p_Port
> 
> It isn't ideal to include the swig'ed pointer instead of the name as it won't be reproducible in another run.

We evaluate all arguments of the TCL command and log the evaluated arguments. For this particular command it's not that much informative, but in general it's always more informative than printing just some TCL variable names.

maliberty:
> We evaluate all arguments of the TCL command and log the evaluated arguments. For this particular command it's not that much informative, but in general it's always more informative than printing just some TCL variable names.

Ideally this would be the most descriptive of what the user did:
```
set clk_port_name clk
set clk_port [get_ports $clk_port_name]
```
the next best
```
set clk_port [get_ports clk]
```
the fully evaluated version doesn't tell you anything about what port was used.

maliberty:
Can you keep track if you are already inside of a public command and not log any further public commands until the outer one is done?

dav77227:
I made changes to the code and improved TCL command tracing to avoid logging nested commands, also no more any workaround is needed. With these changes, we log only public commands and do not log any further public commands until the outer one is done. Please see attached log file.
[2_2_floorplan_io.log](https://github.com/user-attachments/files/15970200/2_2_floorplan_io.log)

As for the evaluation of the commands arguments: after changing the code, as I already mentioned, we do not log commands called inside public commands, so we no longer see the internal `get_name` (`get_name` is called from inside `read_sdc -> current_design`) command in this example, and accordingly, the argument passed to the `get_name` command is not logged. But if the `get_name` command is called in a not-nested place, then its arguments will be evaluated as before. One way that I can think of to fix this issue is to keep a list of public commands, for which we will not evaluate arguments (I think logging `set` commands will create a mess in log files), @maliberty please let me know what you think about this?

Regarding test failures: once we agree on a solution, I'll re-golden all tests.

src/Main.cc
```diff
@@ -45,6 +45,11 @@
 #include <cstdlib>
 #include <iostream>
 #include <string>
+#include <unordered_set>
+
+#define TCL_TRACE_LEAVE_EXEC 2
+#define TCL_TRACE_ENTER_DURING_EXEC 4
```
maliberty, line 51: Instead of using a macro to store a constant, use a const variable.
(https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros)

src/utl/test/test_logged_commands_batch.py
maliberty, line 1: This style of test seems odd as the regression runner will normally invoke `openroad -python test_logged_commands_batch.py` and diff the output against a .ok file.  Is there a reason to deviate?

dav77227: I have changed the test. Now it is like the other tests.

Please let me know if you are ok with this solution, and I'll re-golden the tests.

src/utl/test/test_command_loging.ok
maliberty, line 1: test filenames: 2 'g' in logging

maliberty:
With this update I see you solved the problem of hiding internal calls to public commands.  Otoh I see
```
set num_instances [llength [get_cells -hier *]]
puts "number instances in verilog is $num_instances"
```
turned into
```
OpenROAD> get_cells -hier *
number instances in verilog is 508
```
The user did call get_cells but the bigger context is lost.  Do you think it would work to log at the outermost level and hide anything nested inside an OR command?

dav77227:
> With this update I see you solved the problem of hiding internal calls to public commands. Otoh I see
> 
> ```
> set num_instances [llength [get_cells -hier *]]
> puts "number instances in verilog is $num_instances"
> ```
> 
> turned into
> 
> ```
> OpenROAD> get_cells -hier *
> number instances in verilog is 508
> ```
> 
> The user did call get_cells but the bigger context is lost. Do you think it would work to log at the outermost level and hide anything nested inside an OR command?

I think we can't log at the outermost level and hide anything nested inside an OR command. The problem is that the outermost level is not fixed, and we are unable to dynamically retrieve it.

dav77227:
@maliberty did you have a chance to take a look at this?

maliberty:
Last week was short due to the US holiday.  Have you looked at TCL_ALLOW_INLINE_COMPILATION:

> Tracing has a significant effect on runtime performance because it causes the bytecode compiler to refrain from generating in-line code for Tcl commands such as if and while in order that they may be traced. If traces for the built-in commands are not required, the flags parameter may be set to the constant value TCL_ALLOW_INLINE_COMPILATION.

maliberty:
Please merge master to this branch to resolve recent CI issues.

dav77227:
I have done some experiments using the TCL_ALLOW_INLINE COMPILATION flag in conjunction with TCL_TRACE_ENTER_DURING_EXEC and TCL_TRACE_LEAVE_EXEC flags. Using the bitwise OR operator, we can achieve desired results.

gadfort:
A command like this is still resolving to the swig objects
```
set_input_delay -clock clk  1.0  [all_inputs]
OpenROAD> set_input_delay -clock clk 1.0 _f01edae1105c0000_p_Port _b020dae1105c0000_p_Port _f021dae1105c0000_p_Port _c023dae1105c0000_p_Port _4025dae1105c0000_p_Port _5027dae1105c0000_p_Port _d028dae1105c0000_p_Port _502adae1105c0000_p_Port _d02bdae1105c0000_p_Port _502ddae1105c0000_p_Port _d02edae1105c0000_p_Port _5030dae1105c0000_p_Port _d031dae1105c0000_p_Port _5033dae1105c0000_p_Port _d034dae1105c0000_p_Port _5036dae1105c0000_p_Port ...
```

Also, I think you will need to handle lists in the printing.
For example:
```
create_clock -period 1 {clk1 clk2 clk3}
OpenROAD> create_clock -period 1 clk1 clk2 clk3
```
are not equivalent.

dav77227:
Hi @gadfort. I changed the code to get the expected result as you said. At this point I need to update the tests results to make them pass. Can you take a look, if the changes are no longer required, maybe I can update them?

PR #5255, opened 2024-06-17
mpl2: avoid groupping macros with empty conn signature

AcKoucher:
This is an idea to make it possible for mpl2 to find a valid tiling for the root in cases such as sky130hd/chameleon. For this design, multilevel autoclustering engine generates a cluster with three very big macros that have no connection signature at all. This cluster will be further fine shaped to what we see in the image below .

The existence of this cluster ends up preventing SA from fitting all clusters in the outline later on.


AcKoucher:
Running CI

AcKoucher:
CI passed.

PR #5256, opened 2024-06-17
mpl2: avoid checking nullptr LibertyCell

AcKoucher:
Resolve [#1850](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1850)

This solves the crash, but mpl2 stage only completes with the changes in #5255.

My suspicions are that these checks were there, because mpl2 used to get the instances area in micron from their LibertyCell. This was fixed in #5036.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -1568,14 +1568,14 @@ void HierRTLMP::createDataFlow()
       continue;
     }
 
-    // mark sequential instances
-    odb::dbIntProperty::create(inst, "vertex_id", stop_flag_vec.size());
-    std_cell_vertex[stop_flag_vec.size()] = inst;
-
     const sta::LibertyCell* liberty_cell = network_->libertyCell(inst);
     if (!liberty_cell) {
       continue;
     }
+
+    // Mark registers
+    odb::dbIntProperty::create(inst, "vertex_id", stop_flag_vec.size());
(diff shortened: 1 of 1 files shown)


AcKoucher:
Running CI

maliberty:
How did the checks cause a crash?

AcKoucher:
Sky130hd/chameleon has macros, and so mpl2 is not skipped, because `unfixed_macros != 0`

https://github.com/The-OpenROAD-Project/OpenROAD/blob/73ad7caa801b7f6d27100d38df2668031856bfef/src/mpl2/src/hier_rtlmp.cpp#L446-L456

However, on master, we only populate the HardMacro objects if the `LibertyCell*` is not `nullptr`:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/73ad7caa801b7f6d27100d38df2668031856bfef/src/mpl2/src/hier_rtlmp.cpp#L674-L693

So mpl2 flow would go on with no macros to place which shouldn't happen.
On [#1850](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1850) log:

```
Traversed logical hierarchy
	Number of std cell instances: 2971
	Area of std cell instances: 25033.99
	Number of macros: 0
	Area of macros: 0.00
	Area of macros with halos: 6491855.00
	Area of std cell instances + Area of macros: 25033.99
	Core area: 10009950.00
	Design Utilization: 0.00
	Core Utilization: 0.00
	Manufacturing Grid: 5

maliberty (approved):

PR #5260, opened 2024-06-18
Added long running commands runtime to log files.

annapetrosyan26:
Added logging of the runtime and peak memory for the following commands:

- `global_place`
- `rtl_macro_placer` 
- `repair_design`
- `repair_timing`
- `global_route`
- `detailed_route`

Updated the regression tests diff checking to ignore runtime lines.


src/utl/src/timer.cpp
```diff
@@ -74,4 +76,23 @@ DebugScopedTimer::~DebugScopedTimer()
...
+size_t ScopedStatistics::getPeakMemoryUsage()
+{
+  struct rusage rusage;
+  getrusage(RUSAGE_SELF, &rusage);
+  return (size_t) (rusage.ru_maxrss * 1024L);
+}
+
+ScopedStatistics::~ScopedStatistics()
+{
+  logger_->report("Runtime {} seconds, memory used {} KB.",
```
oharboe, line 93: This output leaves me with questions...

Scope? What took this long and used this much memory. What was the memory usage at start of scope compared to the peak usage.

src/utl/src/timer.cpp
```diff
@@ -74,4 +76,23 @@ DebugScopedTimer::~DebugScopedTimer()
...
 
+ScopedStatistics::ScopedStatistics(utl::Logger* logger)
+    : Timer(), logger_(logger)
+{
+}
+
+size_t ScopedStatistics::getPeakMemoryUsage()
+{
+  struct rusage rusage;
+  getrusage(RUSAGE_SELF, &rusage);
```
oharboe, line 87: error handling for getrusage missing. There may not be access rights to get this information or otherwise it is not available?

oharboe: Actually, what I want to know is maximum virtual memory, not maximum resident memory set. The only reason why ORFS is settling for maximum resident memory set is because `ps` doesn't measure maximum virtual memory.

A discussion on cores and max virtual memory  vs. maximum resident memory set.

test/regression.tcl
```diff
@@ -230,14 +230,30 @@ proc run_test_lang { test lang } {
             set tmp_file [file join $result_dir $test.tmp]
             exec tr -d "\r" < $log_file > $tmp_file
             file rename -force $tmp_file $log_file
-            if [catch [concat exec diff $diff_options $ok_file $log_file \
-                         >> $diff_file]] {
-              puts " *FAIL*$error_msg"
-              append_failure $test
-              incr errors(fail)
+
```
oharboe, line 233: this proc is already quite long. Break out a new proc and give the proc a good name so that it is clear what is being tested. Testing for regressions in maximum resident set is bound to have very significant variations...

oharboe (changes requested):
Wrote up some feedback on what I can think of.

Also, there was a request to make this non-default option so that logs/reports don't become any more non-idempotent than they already are.

If OpenROAD wants to be an "examplary citizen" for bazel type build systems, then it should be a way to configure OpenROAD logs and reports should to idempotent... This of course means forgoing some types of logging and reporting like running times and memory consumption but this is understood and accepted in a bazel context.

src/utl/test/test_statistics.py
```diff
@@ -0,0 +1,42 @@
+import os
```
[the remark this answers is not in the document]
oharboe: What I mean with "CI" is continuous integration, I wanted to know if this test is hooked up to the tests run on pull requests in github.

Regarding os vs pathlib, I think `pathlib` is more recent and a bit more elegant(but a little bit more complex than `os`) in the Python world but that `os` is more familiar(I know it better). I have no strong opinion on either for such simple scripts as these. It depends on what my colleagues prefer.

src/utl/src/timer.cpp
oharboe: This doesn't describe the scope, what was run during this time.

stefanottili:
Would it be possible to put the name of the command in the same line ?
The whole idea was to be able to grep for a keyword to easily get runtime info out of the verbose log file.

maliberty:
> Also, there was a request to make this non-default option so that logs/reports don't become any more non-idempotent than they already are.

I disagreed then and still do.  The goal is to have an informative log file by default.  If you want to give up information to get idempotency that should be the non-default option.

src/utl/src/timer.cpp
maliberty: DRT already outputs in this format
```
[INFO DRT-0267] cpu time = 00:00:04, elapsed time = 00:00:01, memory = 2197.27 (MB), peak = 4031.35 (MB)
```
Is there a reason to have another?

src/utl/src/timer.cpp
maliberty: If you can't measure it you can't report it so I'm not sure what is requested here.

I would argue that max virtual memory is not very interesting in 64bit computing.  You'll never run out of virtual address space and thrashing is a resident set size issue.

oharboe: max virtual address space can be measured in C++ easily enough, it just the `time` command that doesn't support it, which makes it awkward for ORFS to report it.

I disagree that virtual memory size isn't interesting. resident set size will always be smaller than physical memory and and doesn't give any indication that you have too large a working set and that the system could be thrashing. That said, even if max virtual memory is larger than physical memory, doesn't mean that there will be thrashing.

src/utl/src/timer.cpp
oharboe: That prefix is nice and gives some context, but there should be a human description of the scope of what is being measured. In some cases it would be unmabigious from that prefix, but e.g. detailed routing measures for each iteration, for instance. So some further description of context could be useful.

oharboe:
> > Also, there was a request to make this non-default option so that logs/reports don't become any more non-idempotent than they already are.
> 
> I disagreed then and still do. The goal is to have an informative log file by default. If you want to give up information to get idempotency that should be the non-default option.

I wanted to mention it, because it came up last time.

I do prefer this sort of information in logs by default and I don't know bazel well enough to understand why or how non-idempotent log artifacts cause a problem or if they are just an eyesore of sorts. I haven't observed any problems as such. Non-idempotent logs don't surprise me and isn't a problem for me. Looking forward to an explanation of the issue here for such build systems. bazel is one example, there are other... Certainly with build times of 24 hours, artifacts are required, ORFS alone doesn't cut it.

src/utl/src/timer.cpp
maliberty: How is it easily measured?

rss ~= physical memory is a warning sign.  vsz > physical memory doesn't say much.

oharboe: I don't have a mac, but when I compile and run this on Ubuntu, I get the output below:

```C++
#include <iostream>
#include <unistd.h>

#ifdef __linux__
#include <fstream>
#elif __APPLE__
#include <mach/mach.h>
#endif

void getMemoryUsage(size_t &rss, size_t &vsz) {
#ifdef __linux__
    std::ifstream statm("/proc/self/statm");
    if (!statm.is_open()) {
        std::cerr << "Error opening /proc/self/statm" << std::endl;
        return;
    }

    size_t pageSize = sysconf(_SC_PAGESIZE);
    size_t size, resident, share, text, lib, data, dt;

    statm >> size >> resident >> share >> text >> lib >> data >> dt;

    rss = resident * pageSize; // Convert from pages to bytes
    vsz = size * pageSize;     // Convert from pages to bytes

    statm.close();
#elif __APPLE__
    mach_task_basic_info info;
    mach_msg_type_number_t infoCount = MACH_TASK_BASIC_INFO_COUNT;

    if (task_info(mach_task_self(), MACH_TASK_BASIC_INFO, (task_info_t)&info, &infoCount) != KERN_SUCCESS) {
        std::cerr << "Error getting task info" << std::endl;
        return;
    }

    rss = info.resident_size;
    vsz = info.virtual_size;
#endif
}

int main() {
    size_t rss = 0, vsz = 0;
    getMemoryUsage(rss, vsz);
    std::cout << "RSS: " << rss << " bytes" << std::endl;
    std::cout << "VSZ: " << vsz << " bytes" << std::endl;
    return 0;
}
```

Output:

```
$ ./a.out
RSS: 3440640 bytes
VSZ: 6377472 bytes
```

annapetrosyan26:
> Would it be possible to put the name of the command in the same line ? The whole idea was to be able to grep for a keyword to easily get runtime info out of the verbose log file.

We can put the name of the command in the same line.

src/utl/test/test_statistics.py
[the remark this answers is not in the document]
annapetrosyan26: This test is included in the regression, so it's in CI.

src/utl/src/timer.cpp
annapetrosyan26: We can also add the command name before reporting its runtime to be clear about the scope.

src/utl/src/timer.cpp
annapetrosyan26: Would it be more convenient to have both RSS and VSZ information?

oharboe: For my part, I think we need both. Over time, we may develop better understanding of how they interact with OpenROAD and decide we only need rsz or rss and why only one is sufficient, but for now I'm definitely missing rsz when analyzing how much more resources I need to give a build server...

oharboe:
Could you show some example output in this PR after the next update?

annapetrosyan26:
> Could you show some example output in this PR after the next update?

The log file should contain an information like the example below:
`repair_design: runtime = 0.000545613 (seconds), rsz start = 2197.27 (MB), vsz start = 4031.35 (MB), rsz peak = 2197.27 (MB), vsz peak = 4031.35 (MB).`
Will this work?

oharboe:
> > Could you show some example output in this PR after the next update?
> 
> The log file should contain an information like the example below: `repair_design: runtime = 0.000545613 (seconds), rsz start = 2197.27 (MB), vsz start = 4031.35 (MB), rsz peak = 2197.27 (MB), vsz peak = 4031.35 (MB).` Will this work?

- Round off to integer seconds & megabytes.
- Now I have to mentally do the math of peak - start, which is the information I care about. So the log line could do this math for me. net usage rsz = 1234/MB, vsz 12552/MB, peak rsz = 13534/MB, vsz 1552/MB
- If this log line line can be squeezed into 80 chars while being legible, that's a boon...

src/utl/include/utl/timer.h
```diff
@@ -78,4 +80,23 @@ class DebugScopedTimer : public Timer
...
+ public:
+  ScopedStatistics(utl::Logger* logger, std::string msg);
+  ~ScopedStatistics() override;
+
+ private:
+  size_t getMemoryUsage(const char* tag);
+  size_t getStartRSZ();
+  size_t getPeakRSZ();
+  size_t getStartVSZ();
+  size_t getPeakVSZ();
```
oharboe, line 94: I asked chatgpt to add comments and it got it all correct, so I don't see a point in adding those comments back in here.... That is just going to poison the language learning models by feeding it it's own output...

```C++
// ScopedStatistics class is used to collect memory usage statistics during its lifetime
class ScopedStatistics : public Timer
{
public:
  // Constructor that initializes the logger and message, and starts collecting statistics
  ScopedStatistics(utl::Logger* logger, std::string msg);

  // Destructor that finalizes the statistics collection and logs the results
  ~ScopedStatistics() override;

private:
  // Gets the current memory usage based on the provided tag
  size_t getMemoryUsage(const char* tag);

  // Gets the Resident Set Size (RSZ) at the start
  size_t getStartRSZ();

  // Gets the peak Resident Set Size (RSZ) during the scope
  size_t getPeakRSZ();

  // Gets the Virtual Memory Size (VSZ) at the start
  size_t getStartVSZ();

  // Gets the peak Virtual Memory Size (VSZ) during the scope
  size_t getPeakVSZ();

  // Message to be logged
  std::string msg_;

  // Initial Resident Set Size (RSZ)
  size_t start_rsz_;

  // Initial Virtual Memory Size (VSZ)
  size_t start_vsz_;

  // Logger to log the statistics
  utl::Logger* logger_;
};
```

oharboe (approved):
:+1: Looks good to me. I don't know anything about that tcl testing code/CI integration. Leaving that for others to review...

src/utl/src/timer.cpp
```diff
@@ -74,4 +78,71 @@ DebugScopedTimer::~DebugScopedTimer()
...
+}
+
+size_t ScopedStatistics::getPeakRSZ()
+{
+  return getMemoryUsage("VmHWM:");
+}
+
+ScopedStatistics::~ScopedStatistics()
+{
+  logger_->report(msg_ + ": runtime {} seconds, usage: rsz = {} MB, vsz = {} MB, peak: rsz = {} MB, vsz = {} MB", static_cast<int>(Timer::elapsed()), (getPeakRSZ() - start_rsz_),
```
maliberty, line 144: runtime is unclear as it could be elapsed or cpu time.  I think using elapsed is clearer.

Personally I think peak is sufficient (@oharboe any opinion)?

maliberty (changes requested):

src/utl/src/timer.cpp
oharboe: If we don't know the starting memory consumption, we can miss false positives: the problem is not with the algorithm that is reported as too high usage, but something in the setup.

maliberty: This doesn't tell you the starting memory - only the end - start

oharboe: correct. pick a way to present two numbers....

start memory is trivially calculated from peak and usagr, true for any cpmbination of these two numbers. most time usage and peak is what is of interested

maliberty: usage and peak do not give start.  The peak may well be in the middle not at the end.

oharboe: Now I am confused as to what this code is measuring. Seems like that should be clear before merging(if not to me, then everyone else....)

What I really care about is to know how much ram that I should have for a stage. I dont care what my machine has, I just.need to know what the correct amount of ram for the job is so I can make it so.

maliberty: Which is what peak should say and why I don't think usage matters much.

oharboe: Agreed. Keep it simple. If there is something wrong(large initial memory allocation outside of scope), then it can be debugged otherwise. I think normal operation, determining required RAM, is more important and other information clouds that...

src/utl/src/timer.cpp
maliberty: I'm fine to add a prefix

src/utl/src/timer.cpp
maliberty: so let's go with the DRT format above (with prefix)

src/utl/src/timer.cpp
annapetrosyan26: The command name before reporting its runtime is already added. But by adding prefix like this `[INFO DRT-0267]` we should also create a unique message ID for each tool as `Logger::info` requires it, is this your suggestion?

src/utl/src/timer.cpp
annapetrosyan26: From the conversation below it's not clear what the final message should be.  Please confirm if it should look like the DRT message you provided.
 `[INFO DRT-0267] cpu time = 00:00:04, elapsed time = 00:00:01, memory = 2197.27 (MB), peak = 4031.35 (MB)`
Otherwise please provide the exact way you would like the message to be.

maliberty: I was thinking the prefix would be about what is being timed, eg

`repair_design: cpu time = 00:00:04, elapsed time = 00:00:01, memory = 2197.27 (MB), peak = 4031.35 (MB)`

Though the prefix would be up to the client code.

stefanottili:
It would be nice to have one common prefix that one can grep for.

src/utl/src/timer.cpp
```diff
@@ -74,4 +78,71 @@ DebugScopedTimer::~DebugScopedTimer()
...
+      logger_(logger)
+{
+}
+
+size_t ScopedStatistics::getMemoryUsage(const char* tag)
+{
+  FILE* file = fopen("/proc/self/status", "r");
+  if (file == nullptr) {
+    perror("Failed to open /proc/self/status");
+    exit(EXIT_FAILURE);
```
maliberty, line 95: This is too harsh a form of error handling.  Issuing a logger warning is sufficient.

src/utl/src/timer.cpp
```diff
@@ -74,4 +78,71 @@ DebugScopedTimer::~DebugScopedTimer()
...
+      msg_(std::move(msg)),
+      start_rsz_(getStartRSZ()),
+      start_vsz_(getStartVSZ()),
+      logger_(logger)
+{
+}
+
+size_t ScopedStatistics::getMemoryUsage(const char* tag)
+{
+  FILE* file = fopen("/proc/self/status", "r");
```
maliberty, line 92: procfs doesn't exist on OSX unfortunately.  src/drt/src/db/infra/frTime_helper.cpp attempts to handle OSX

src/utl/src/timer.cpp
```diff
@@ -74,4 +90,73 @@ DebugScopedTimer::~DebugScopedTimer()
...
+{
+#if defined(__APPLE) && defined(__MACH__)
+  struct mach_task_basic_info info;
+  mach_msg_type_number_t info_count = MACH_TASK_BASIC_INFO_COUNT;
+  if (task_info(mach_task_self(), MACH_TASK_BASIC_INFO,
+                (task_info_t) &info, &info_count) {
+    return (size_t) 0L;
+  }
+  return (size_t) info.resident_size / 1024;
+#elif defined(__linux) || defined(__linux) || defined(linux) \
```
rovinski, line 112: The first two conditions look identical to me?

annapetrosyan26:
@maliberty  the appropriate changes have been made. Could you please review them?

stefanottili:
Who would have thought that logging runtimes could be controversial … especially considering how verbose drt is.

Issue #5263, opened 2024-06-19
PPL: update constraints to save the constraint details of each pin


eder-matheus:
Make the constraint information of each pin available for other tools in the flow (like GPL and MPL2).
The goal is to give flexibility for these tools to place the instances near the region of the pin, not near an exact location

PR #5264, opened 2024-06-19
ppl: add constraint information to odb::dbBTerm

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5263

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -1886,6 +1886,16 @@ class dbBTerm : public dbObject
 
   uint32_t staVertexId();
   void staSetVertexId(uint32_t id);
+
+  ///
+  /// Set the region where the BTerm is constrained
+  ///
+  void setConstraintRegion(const std::pair<Point, Point>& constraint_region);
+
+  ///
+  /// Get the region where the BTerm is constrained
+  ///
+  std::pair<Point, Point> getConstraintRegion();
 };
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -1895,7 +1895,7 @@ class dbBTerm : public dbObject
   ///
   /// Get the region where the BTerm is constrained
   ///
-  std::pair<Point, Point> getConstraintRegion();
+  std::optional<std::pair<Point, Point>> getConstraintRegion();
 };
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/src/odb/src/db/dbBTerm.cpp
+++ b/src/odb/src/db/dbBTerm.cpp
@@ -76,7 +76,6 @@ _dbBTerm::_dbBTerm(_dbDatabase*)
   _ext_id = 0;
   _name = nullptr;
   _sta_vertex_id = 0;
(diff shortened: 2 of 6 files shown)


src/odb/src/db/dbBTerm.h
```diff
@@ -92,6 +92,7 @@ class _dbBTerm : public _dbObject
   dbId<_dbBTerm> _ground_pin;
   dbId<_dbBTerm> _supply_pin;
   std::uint32_t _sta_vertex_id;  // not saved
+  std::pair<Point, Point> _constraint_region;
```
maliberty, line 95: This is a schema change and requires backward compatibility code along with serialization code.

src/ppl/src/IOPlacer.cpp
```diff
@@ -2802,6 +2802,40 @@ void IOPlacer::initNetlist()
   }
 }
 
+void IOPlacer::commitConstraintsToDB()
+{
+  for (Constraint& constraint : constraints_) {
```
maliberty, line 2807: does this handle mirroring?

src/odb/src/db/dbBTerm.h
maliberty: how does this work with bumps?

eder-matheus: Do you mean the pins that PPL places at the top layer? I'm adding the lower left and the upper right points to indicate the box the pins can be placed.

src/ppl/src/IOPlacer.cpp
eder-matheus: It wasn't, but my last commits are covering mirrored pins. I've tested with PPL unit tests and the values are correct.

eder-matheus:
@maliberty This is ready to the final review.

src/odb/include/odb/db.h
```diff
@@ -1886,6 +1886,16 @@ class dbBTerm : public dbObject
...
+
+  ///
+  /// Set the region where the BTerm is constrained
+  ///
+  void setConstraintRegion(const std::pair<Point, Point>& constraint_region);
+
+  ///
+  /// Get the region where the BTerm is constrained
+  ///
+  std::pair<Point, Point> getConstraintRegion();
```
maliberty, line 1898: If there is no constraint then you get back (-1,-1),(-1,-1) as a special value which is non-obvious.  It would be cleaner to return a std::optional<std::pair<Point, Point>>

src/odb/src/db/dbBTerm.cpp
```diff
@@ -921,4 +927,17 @@ void dbBTerm::staSetVertexId(uint32_t id)
...
+    const std::pair<Point, Point>& constraint_region)
+{
+  _dbBTerm* bterm = (_dbBTerm*) this;
+  bterm->_constraint_region = constraint_region;
+}
+
+std::optional<std::pair<Point, Point>> dbBTerm::getConstraintRegion()
+{
+  _dbBTerm* bterm = (_dbBTerm*) this;
+  return bterm->_constraint_region;
```
maliberty, line 940: If it is undefined you need to return an empty optional

eder-matheus: Now it should be complete.

maliberty (approved):

PR #5265, opened 2024-06-20
gui: add tooltips to Timing Report

### LINKED ISSUE #5083: Add tooltips to Skew column
### Description

Add a tooltip to explain in a sentence or two exactly what the Skew column is displaying.

image

See also https://github.com/The-OpenROAD-Project/OpenROAD/issues/5002 that has been following some bugfixes in Skew column.

### Suggested Solution

Add a terse exact definition of the Skew column in tooltips.

Using this definition, it should be possible to corroborate that information in the Skew column with report_clock_skew or other sources in mock-array and mock-array/Element from within the OpenROAD GUI.

### Additional Context

_No response_

AcKoucher:
Resolve #5083 

Adding tool tips to logic delay/depth and skew.

### DIFF
--- a/src/gui/src/staGui.cpp
+++ b/src/gui/src/staGui.cpp
@@ -165,9 +165,35 @@ QVariant TimingPathsModel::headerData(int section,
                                       Qt::Orientation orientation,
                                       int role) const
 {
-  if (role == Qt::DisplayRole && orientation == Qt::Horizontal) {
-    return getColumnNames().at(static_cast<Column>(section));
+  Column column = static_cast<Column>(section);
+
+  if (role == Qt::ToolTipRole) {
+    switch (column) {
+      case Clock:
+      case Required:
+      case Arrival:
+      case Slack:
+      case Fanout:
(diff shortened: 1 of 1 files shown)


AcKoucher:
@oharboe It looks like this:

oharboe:
what about back to back inverters for logic depth/delay. Should not the same filtering criteria apply to both? If so, use same language.

AcKoucher:
Logic Delay: Delay from instances apart from buffers and back-to-back inverters.
Logic Depth: Number of instances apart from buffers and back-to-back inverters.

Reasonable?

AcKoucher:
fixed

rovinski:
I would do:
Logic Delay: Path delay from instances (excluding buffers and consecutive inverter pairs)
Logic Depth: Path instances (excluding buffers and consecutive inverter pairs)

I don't think there need to be periods in the tooltip.

AcKoucher:
Done

maliberty (approved):

PR #5269, opened 2024-06-21
Add getting involved section to the readme

robtaylor:
Noticed that the comms channels for OpenRoad weren't immediately obvious to a newcomer, so added this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,8 +108,16 @@ includes GlobalFoundries shuttles, design contests and IC design
 workshops. The OpenROAD flow has been successfully used to date
 in over 600 silicon-ready tapeouts for technologies up to 12nm.
 
-There is a [Matrix](https://matrix.org/) channel for discussion at 
+## Getting Involved
+
+There is a users mailing list, which you can join [here](https://groups.google.com/g/openroad-users).
+
+There is also a [Matrix](https://matrix.org/) channel for discussion at 
 [#openroad:matrix.org](https://matrix.to/#/%23openroad:matrix.org).
+
+For guidance on contributing to OpenRoad you can see the 
+[Getting Involved section of the documentation](https://openroad.readthedocs.io/en/latest/contrib/GettingInvolved.html). 
+We encourage contribution and collaboration!
 
 ## Getting Started with OpenROAD-flow-scripts
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,9 @@ includes GlobalFoundries shuttles, design contests and IC design
 workshops. The OpenROAD flow has been successfully used to date
 in over 600 silicon-ready tapeouts for technologies up to 12nm.
 
+There is a [Matrix](https://matrix.org/) channel for discussion at 
+[#openroad:matrix.org](https://matrix.to/#/%23openroad:matrix.org).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,6 +108,17 @@ includes GlobalFoundries shuttles, design contests and IC design
 workshops. The OpenROAD flow has been successfully used to date
 in over 600 silicon-ready tapeouts for technologies up to 12nm.
 
+## Getting Involved
+
+There is a users mailing list, which you can join [here](https://groups.google.com/g/openroad-users).
+
+There is also a [Matrix](https://matrix.org/) channel for discussion at 
+[#openroad:matrix.org](https://matrix.to/#/%23openroad:matrix.org).
```
maliberty, line 116: I'm also unfamiliar with this service.  It appears to require an account to even see the content so I don't know how relevant it is.  I don't believe anyone from the OR team interacts here.  Do you see a lot of useful content here?

README.md
```diff
@@ -108,6 +108,17 @@ includes GlobalFoundries shuttles, design contests and IC design
 workshops. The OpenROAD flow has been successfully used to date
 in over 600 silicon-ready tapeouts for technologies up to 12nm.
 
+## Getting Involved
+
+There is a users mailing list, which you can join [here](https://groups.google.com/g/openroad-users).
```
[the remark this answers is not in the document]
oharboe: Never heard of it :-)

README.md
oharboe: Being a social media ludite, haven't heard of it...

robtaylor:
Matrix is common open messaging service used by many open source projects.
It has been gradually replacing or being bridged with irc for a few years
now.

It’s an open standard, with a number of open source clients.

OpenRoad users was the only mailing list I could find to add here. Is there
another? If not I suggest the project grows one.

maliberty:
Most conversations happen in GH or slack.  If there is organic demand for a mailing list I'm open but I haven't see any so far.

maliberty:
I don't see either of these as forums to promote at this point.

robtaylor:
I think any forum would be useful to promote.

What do you consider the correct forum to be?

oharboe:
I dont know what "correct forum" means, but currently there is a little bit of activity in github discussions under ORFS and OpenROAD... It has good participation from maintainers, hard to compete with... Strangely there is no such thing as github chat&chat rooms....

robtaylor:
Yeah, that makes sense. I’ll fix the patch to cover that instead.

I really think a matrix or irc room would be very helpful for contributors
and users - these can be the same room and for most OSS projects that’s the
common practice.

Obviously there’s the open-silicon-dev.org slack room as well, but I think
we’d all agree a proprietary chat platform is inappropriate.

(Again just for your info, matrix is a an open federated protocol, which
gives slack-like capabilities. There are bridges to many other chat-like
protocols.)

robtaylor:
I’ll just add that we could bridge to the slack room if that room is seen
as useful/important. (
https://matrix.org/ecosystem/bridges/slack/)

maliberty:
Most chit-chat is on open-source-silicon.slack.com

PR #5281, opened 2024-06-25
DRT: fix incremental updates in DRT

osamahammad21:
DRT used to clear special nets routing if it was used more than once i.e. by calling detailed_route command after pin_access command or calling detailed_route command multiple times. This PR fixes this issue.


maliberty:
ready for review?

maliberty:
secure CI needed?

maliberty (approved):

osamahammad21:
@maliberty I am having troubles with the metric updates resulting from this PR. Now, it's failing sky130hd_ibex with this which doesn't make any sense:
`ibex_sky130hd (tcl) *FAIL* DRT::max_capacitance_slack -11% < -11%
`

maliberty:
> `ibex_sky130hd (tcl) *FAIL* DRT::max_capacitance_slack -11% < -11% `

My guess is that rounding is making this look odd (@vvbandeira ) and you should just update it.  In the metric_limits

`"DRT::max_capacitance_slack" : "-11.363490057317899"`

osamahammad21:
@maliberty The weird thing is that I updated it with the new value that I got from running the design on my local machine which is ` "DRT::max_capacitance_slack" : "-10.870980671686256"`. But I am getting a different value on jenkins!

vvbandeira:
> > `ibex_sky130hd (tcl) *FAIL* DRT::max_capacitance_slack -11% < -11% `
> 
> My guess is that rounding is making this look odd (@vvbandeira ) and you should just update it. In the metric_limits
> 
> `"DRT::max_capacitance_slack" : "-11.363490057317899"`

We could change the formatting here: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/test/flow_metrics.tcl#L133

PR #5282, opened 2024-06-25
mpl2: split macros between partitions when using PAR

### LINKED ISSUE #4539: rtlmp fails without SYNTH_HIERARCHICAL=1
### Describe the bug

I want to have RTLMP and flatten the MegaBoom design. MegaBoom relies heavily on register retiming and cloning per its documentation.

1. untar https://drive.google.com/file/d/1aGCnX_eXryTUB5RYXcMJ59_xdw9lVIeK/view?usp=sharing
2. Run below, ca. 26 minutes on my machine.

```
$ NUM_CORES=16 ./run-me-ChipTop-asap7-flat.sh 
OpenROAD v2.0-11843-gc74f6a2f0 
[INFO ORD-0030] Using 16 thread(s).
HierRTLMP Flow enabled...
rtl_macro_placer -halo_width 20 -report_directory bazel-bin/objects/asap7/ChipTop/flat/rtlmp
Floorplan Outline: (0.0, 0.0) (1500.0, 1500.0),  Core Outline: (2.052, 2.16) (1497.96, 1497.96)
Traversed logical hierarchy
	Number of std cell instances: 940301
	Area of std cell instances: 120349.56
	Number of macros: 76
	Area of macros: 181911.06
	Area of macros with halos: 565030.88
	Area of std cell instances + Area of macros: 302260.62
	Core area: 2237579.00
	Design Utilization: 0.14
	Core Utilization: 0.06
	Manufacturing Grid: 1

[ERROR MPL-0040] Failed on cluster (root)_glue_logic_0_0_0_0_0
Error: macro_place_util.tcl, 138 MPL-0040
openroad> 
```

### Expected Behavior

No failure

### Environment

```shell
OpenROAD v2.0-11843-gc74f6a2f0
```

### To Reproduce

See above

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
This is an idea to resolve #4539.

MPL2 uses TritonPart to partition what it calls big flat clusters which are made of a single logical module. The methodology consists of breaking the cluster in two until we get the number of std cells to be within the leaf threshold. However, nothing is done regarding the macros of this flat cluster, i. e., what we currently do is split the std cells keeping the macros in the same cluster until the end of the clustering stage. That makes all the macros be kept in the same cluster leading to something like we see in the image below:

The logic here is to add the macros to a certain partition based on the number of std cell iterms connected to it. With these changes, we get something like:

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -2167,28 +2167,8 @@ void HierRTLMP::breakLargeFlatCluster(Cluster* parent)
     }
   }
 
-  std::map<odb::dbInst*, int> parent_macro_to_iterm_count
-      = getMacroToStdCellPinsCountMap(parent);
-  std::map<odb::dbInst*, int> partition_macro_to_iterm_count
-      = getMacroToStdCellPinsCountMap(cluster_part_1);
-
-  std::vector<odb::dbInst*> kept_macros, macros_to_move;
-  for (const auto& macro : parent->getLeafMacros()) {
-    if (parent_macro_to_iterm_count.at(macro)
-        > partition_macro_to_iterm_count.at(macro)) {
-      kept_macros.push_back(macro);
-    } else {
-      macros_to_move.push_back(macro);
-    }
-  }
-
-  parent->clearLeafMacros();
-
-  for (const auto& macro : kept_macros) {
-    parent->addLeafMacro(macro);
-  }
-  for (const auto& macro : macros_to_move) {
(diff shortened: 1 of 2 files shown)


AcKoucher:
@oharboe Would you know if I can find this megaBoom test case so I can run the whole flow on it?

oharboe:
> @oharboe Is it possible to have access to this megaBoom test case so I can run the whole flow on it?

https://github.com/The-OpenROAD-Project/megaboom

maliberty:
It looks a lot better

PR #5296, opened 2024-06-27
mpl2: add simple test for setMacro()

amuritna:
Added a simple test for setMacro that instantiates a database with 3 dbInst objects (default/no configuration), and then runs setMacro(inst) for each instance.

Tagging my GSoC mentors: @maliberty @vvbandeira


src/mpl2/test/cpp/TestSnapper.cpp
```diff
@@ -0,0 +1,72 @@
...
+  odb::dbDatabase::beginEco(block_);
+  odb::dbInst* inst1 = odb::dbInst::create(block_, master_, "cells_1");
+  odb::dbInst* inst2 = odb::dbInst::create(block_, master_, "cells_2");
+  odb::dbInst* inst3 = odb::dbInst::create(block_, master_, "cells_3");
+  odb::dbDatabase::endEco(block_);
+
+  Snapper snapper;
+  snapper.setMacro(inst1);
+  snapper.setMacro(inst2);
+  snapper.setMacro(inst3);
```
maliberty, line 69: Do you plan to add to this test or just verify that setMacro doesn't crash?  It doesn't invoke snapMacro.

maliberty (approved):

src/mpl2/test/cpp/TestSnapper.cpp
amuritna: I thought I'd start with something easier, so this one only verifies that setMacro doesn't crash. The one that I showed you on Monday is a different one I'm still working on right now.

PR #5303, opened 2024-06-28
dbSta: Honor dont_touch attribute in Verilog

gatecat:
This enables a cell to be marked as "don't touch" (for example, a buffer you want kept), directly inside the HDL using the attribute (as per the included test) without needing any custom Tcl scripting.


maliberty (approved):
Nice.  Is this an attribute that yosys passes through synthesis?

gatecat:
Yosys passes all attributes through synthesis.

Yosys itself honours the `keep` attribute to stop things being optimised away, but I decided to pick `dont_touch` to match OpenROAD terminology and because the semantics are a bit different (`dont_touch` is stronger in OpenROAD than `keep` is in Yosys).

In other commercial FPGA flows I've also seen the `dont_touch` attribute used.

PR #5313, opened 2024-07-01
dft: Fix the place_sort_sky130 test

gatecat:
I didn't sport that `dft/test/helpers.tcl` was a symlink, so forgot to add my change to it adding the `place_inst` function for the test, so the test fails on master. As this probably isn't a very useful placement function outside of this specific test, I have decided to add it to the test this time round rather than modifying the global `helpers.tcl`.

Does something else also need to be fixed so that issues in Tcl tests get caught in CI?

### DIFF
--- a/src/dft/test/place_sort_sky130.tcl
+++ b/src/dft/test/place_sort_sky130.tcl
@@ -13,6 +13,13 @@ set_dft_config -max_length 10
 
 scan_replace
 
+proc place_inst { inst x y } {
+  set db_inst [[ord::get_db_block] findInst $inst]
+  $db_inst setLocation $x $y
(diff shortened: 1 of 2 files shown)


maliberty:
You need to add place_sort_sky130 to src/dft/test/regression_tests.tcl in order for the test to be run.

maliberty:
You can test by running the `regression` script in dft/test.

gatecat:
Thanks, I've updated this PR - I added the test to `tests/CMakeLists.txt` and was checking with `ctest`, I hadn't spotted the tcl file too.

maliberty (approved):

PR #5316, opened 2024-07-01
Dpl: Inserting decap cells with the insert_decap command

### LINKED ISSUE #4970: Create a decap insertion routine
OpenROAD doesn't currently have any automated method for inserting decap cells. It could provide a simple routine after static IR drop analysis to insert a user-specified amount of decap (in Liberty capacitance units) around instances with the highest static IR drop because there is a high correlation to the worst dynamic IR drop instances.

I can imagine a simple, greedy approach where decap cells are added within X um of the worst instances until the user-provided threshold is met. Note that decap most useful when it is connected to the power rail of the problematic instance (so same row or +/- 1 row depending on the row orientation).

> @maliberty usually decaps should be added after IR drop analysis and inserted next to instances with the worst IR drop. A loose spec off the top of my head would be:

* list of decap cells to use
* maximum amount of decap to insert (in liberty cap units), or
* voltage margin to fix up to
  * e.g. if worst IR drop is 0.15 V, and the user specifies 0.10 V, the tool would try to insert decap around all failing instances to ensure the worst IR drop is no more than 0.10 V

_Originally posted by @rovinski in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1945#issuecomment-2062955341_

luis201420:
Creating the `insert_decap` command to insert decap cells in the areas with the highest IR DROP.
To use the command you must first execute `analyze_power_grid` for the net VDD, to obtain the areas with the highest IR DROP.
Fixes #4970


src/dpl/src/Opendp.i
```diff
@@ -152,6 +152,20 @@ filler_placement_cmd(dpl::dbMasterSeq *filler_masters,
   opendp->fillerPlacement(filler_masters, prefix);
 }
 
+void
+set_decap_master(odb::dbMaster *master, float cap)
```
eder-matheus, line 156: add_decap_master

src/dpl/src/Opendp.tcl
```diff
@@ -136,6 +136,46 @@ proc filler_placement { args } {
   dpl::filler_placement_cmd $filler_masters $prefix
 }
 
+sta::define_cmd_args "insert_decap" { [-cap target] [-cells cell_info] }
```
eder-matheus, line 139: Perhaps change `-cap` to `-target_cap`?

eder-matheus (changes requested):
Could you also add a small unit test?

src/dpl/src/DecapPlacement.cpp
```diff
@@ -0,0 +1,335 @@
...
+            irdrop_points.end(),
+            [](std::pair<odb::Point, double>& point1,
+               std::pair<odb::Point, double>& point2) {
+              return point1.second > point2.second;
+            });
+
+  // If fillers are placed
+  if (have_fillers_) {
+    logger_->error(
+        DPL, 54, "Filler cells found when running insert decap cells");
```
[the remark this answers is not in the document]
maliberty: More actionable: Run remove_fillers before inserting decap cells

maliberty:
@luis201420 do you have an ETA to resolve the feedback?

luis201420:
Hi @maliberty,
I was investigating the compilation error. I can now reproduce the error, I think the difference in GNU version was the problem.
I tried first running the master branch without my modifications to verify that everything compiles without problems, then I added the `psm` to the `dpl` CMakeList file and compiled again. The latter resulted in the same problem that is had in Jenkins.
The cause is probably that, when placing `psm` in `dpl`, it needs `psm` to be compiled first, but also in `psm` it has dependencies like `rsz`, and `rsz` has dependencies like `grt` and `grt` has `dpl` as a dependency. Which will generate a loop in compilation.
Perhaps a possible solution is to place the command `insert_decap` in `psm`, to avoid adding `psm` in `dpl`. The idea would be to execute the command in `psm` and then using a `dpl` function to add the decap cells passing the IR DROPS as a parameter. (add `dpl` in `psm` instead of `psm` in `dpl`).
What do you think about it?

src/psm/include/psm/pdnsim.h
```diff
@@ -38,8 +38,11 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <optional>
 #include <string>
 
+#include "dpl/Opendp.h"
 #include "odb/dbBlockCallBackObj.h"
 
+using odb::dbMaster;
```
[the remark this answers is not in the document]
eder-matheus: @luis201420 You should move this inside the psm namespace.

src/psm/test/insert_decap1.tcl
```diff
@@ -0,0 +1,19 @@
...
+read_liberty sky130hd/sky130_fd_sc_hd__ss_n40C_1v40.lib
+read_lef "sky130hd/sky130hd.tlef"
+read_lef "sky130hd/sky130hd_std_cell.lef"
+read_def "insert_decap_gcd.def"
+
+source sky130hd/sky130hd.rc
+
+analyze_power_grid -net {VDD}
+insert_decap -target_cap 1000.5 -cells {"sky130_fd_sc_hd__decap_3" 0.93 "sky130_fd_sc_hd__decap_4" 0.124 "sky130_fd_sc_hd__decap_6" 0.186 "sky130_fd_sc_hd__decap_8" 0.248 "sky130_fd_sc_hd__decap_12" 0.362}
+#insert_decap -target_cap 1000.5 -cells {"sky130_fd_sc_hd__decap_3" 0.93 "sky130_fd_sc_hd__decap_4" 1.24 "sky130_fd_sc_hd__decap_6" 1.86 "sky130_fd_sc_hd__decap_8" 2.48 "sky130_fd_sc_hd__decap_12" 3.62}
```
eder-matheus, line 13: Remove the commented code.

eder-matheus (changes requested):

src/psm/src/pdnsim.tcl
```diff
@@ -117,6 +117,62 @@ proc analyze_power_grid { args } {
...
+      utl::warn "PSM" 280 "$cell_name did not match any masters."
+    }
+  }
+  # Get net name
+  set net_name ""
+  if { [info exists keys(-net)] } {
+    set net_name $keys(-net)
+  }
+
+  puts $net_name
```
maliberty, line 171: rm ?

maliberty:
@eder-matheus I have no further comments - I'll leave it to you to comment or merge.

eder-matheus (approved):

PR #5317, opened 2024-07-01
coverity fixes

eder-matheus:


src/mpl2/src/SimulatedAnnealingCore.cpp
```diff
@@ -518,7 +518,10 @@ void SimulatedAnnealingCore<T>::exchangeMacros()
       neg_index2 = i;
     }
   }
-  std::swap(neg_seq_[neg_index1], neg_seq_[neg_index2]);
+
+  if (neg_index1 >= 0 && neg_index2 >= 0) {
```
maliberty, line 522: If this ever fails it would be a bug.

maliberty (approved):

src/mpl2/src/SimulatedAnnealingCore.cpp
eder-matheus: Should I add an error message if it fails?

maliberty: That would be nice.  You can then move the swap out/after the if block.

Issue #5291, opened 2024-06-26
Integrate repair_design and global_placement


maliberty:
### Description

Running repair_design after global_placement can lead to a placement density hotspots that deteriorate results.

### Suggested Solution

Early in global placement the positions are subject to too much change to commit to the actual buffering/sizing of repair_design.  As the placement overflow gets low enough and the placement is stabilizing it makes sense to switch to actually modifying the design so that the density can be seen.

This will require significant tuning to determine the right point at which to switch over and also to ensure that multiple calls to repair operate well.

@gadfort @QuantamHD @rovinski for any opinions.  I want to align on the goal before we get too far into implementation.

maliberty:
@rovinski in at least some designs I've seen that the repairs are concentrated in local areas.  I also considered doing some global adjustment for the amount of area predicted during virtual resizer but I'm concerned that it won't resolve local issues.

gudeh:
Hi everyone. This is what happens on GPL timing-driven placement: during GPL, we call timing-driven based on a list of 6 fixed values of overflow, defined in src/gpl/src/replace.tcl: `set overflow_list [list 79 64 49 29 21 15]`.

When a timing-driven iteration happens, the resizer does all its work and afterwards removes all the work done (we call it a virtual resizing), such as resizing, buffer insertion, and pin swapping (possibly more modifications, I am not sure yet). What is kept in GPL are only the slack weights calculated by the resizer's work and placement changes based on the slack weights only (due to timing-driven adjustments).

I tried a first experiment by keeping the work done by the resizer only on the last overflow value (15), meaning the last timing-driven iteration is not a virtual one, and we get a nice density correlation between steps 3_3_place_gp and 3_4_place_resized.
<img src="https://github.com/The-OpenROAD-Project/OpenROAD/assets/37420320/ff616120-caf4-40a9-a59d-e1436c3259f2" width="300"> <img src="https://github.com/The-OpenROAD-Project/OpenROAD/assets/37420320/74ee771a-e666-4b1d-a543-de2946cd0c8c" width="300">

We currently have a mismatch between the same steps in master:
<img src="https://github.com/The-OpenROAD-Project/OpenROAD/assets/37420320/9718491b-243e-4671-b516-985bf190709a" width="300"> <img src="https://github.com/The-OpenROAD-Project/OpenROAD/assets/37420320/c231e384-42cc-4345-8d1c-35bb177cb9d3" width="300">

[...]

Issue #5203, opened 2024-06-05
Make info message more actionable `[INFO GPL-0100] worst slack -1.54e-09`


oharboe:
### Description

What is OpenROAD trying to tell me here?

Is OpenROAD trying to tell me that removing `-timing_driven` could speed things up?

For architectural exploration, I initially want fast turnaround time with less accurate results.

```
[INFO GPL-0100] worst slack -1.54e-09
[INFO GPL-0103] Weighted 41603 nets.
```

### Suggested Solution

Make the error message more actionable.

gudeh:
Hi @oharboe, this message is not an error message actually. This is printing the worst slack after calling repair design with resizer. Maybe we should transform this into a debug message.

oharboe:
@gudeh  Understood. However, it may be useful for the user to know that they can relax the optimization target to improve running times. If message can be rearticulated to communicate this, it would be helpful.

maliberty:
It seems that having some document or global options on fast analysis is better than sprinkling hints throughout the log

oharboe:
@rovinski Had an idea to collect data on running times vs quality of results and creating a tool that recommended options?

gudeh:
I am closing this issue. I am currently working on the timing driven mode in global placer, which is what generates the mentioned logs. We should have further modifications on timing driven in the near future.

PR #5318, opened 2024-07-01
gpl: modify log and readme for timing driven mode clarity

gudeh:
Include more clear messages during timing driven mode so the user can better understand what is happening. 
Include possible actions for user to improve runtime due to slow timing driven mode.
Include new tcl command in README for using grt(previous) routing congestion estimator during routability.
Fix default inflation parameters in README.

### DIFF
--- a/src/gpl/README.md
+++ b/src/gpl/README.md
@@ -40,7 +40,7 @@ Timing-driven iterations are triggered based on a list of overflow threshold
 values. Each time the placer execution reaches these overflow values, the 
 resizer is executed. This process can be costly in terms of runtime. The 
 overflow values for recalculating weights can be modified with 
-`-timing_driven_net_reweight_overflow`, you may use less overflow threhsold 
+`-timing_driven_net_reweight_overflow`, you may use less overflow threshold 
 values to decrease runtime, for example.
 
 When the routability-driven option is enabled, each of its iterations will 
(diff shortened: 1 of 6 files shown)


gudeh:
These changes should address issue #5203.

gudeh:
I included the message "Timing-driven: executing resizer for reweighting nets." because the resizer may take a long time, this way the user can know why the tool is loading.

maliberty:
@oharboe please comment as it is your issue

src/gpl/README.md
```diff
@@ -34,17 +34,24 @@ weight nets with low slack. It adjusts the worst slacks (modified with
...
-attempt to achieve the target RC (routing congestion) by comparing it to the 
-final RC. If the algorithm takes too long during routability-driven execution, 
-consider raising the target RC value to alleviate the constraints. The final 
-RC value is calculated based on the weight coefficients. The algorithm will 
-stop if the RC is not decreasing for three consecutive iterations.
+execute RUDY to provide an estimation of routing congestion. Congested tiles 
+will have the area of their logic cells inflated to reduce routing congestion. 
+The iterations will attempt to achieve the target RC (routing congestion) 
+by comparing it to the final RC at each iteration. If the algorithm takes too 
+long during routability-driven execution, consider raising the target RC value
```
oharboe, line 51: as above. Reduce guesswork; be explicit about which option to use to "the target RC value".

oharboe (changes requested):
Just a nit, see comments.

More general comments:

I think the main usability improvement here is that RUDY is used and that is adequately documented.

The other routing options require a strong mental model of what is going on. They are not "user-servicable parts" to my mind.

A user-servicable knob would be something like a value between 0 and 1 where 0 is fastest and 1 is best quality of results...

I'm not suggesting that such a feature should be implemented, but are interested in hearing thoughts.

I think the explanation of the "non user-servicable parts" of OpenROAD here are good.

gudeh:
I realized we do not have the number of buffers inserted (and other modifications by resizer) in the gpl log. Which makes sense because we do a virtual resizing. I will include this info in a new PR, since it is more related to issue #5291.

Do you think this modifications are good to go? @maliberty @oharboe

oharboe (approved):
This level of detail on global router is too deep for my current understanding, so I can't comment much.

maliberty (approved):

maliberty:
> I realized we do not have the number of buffers inserted (and other modifications by resizer) in the gpl log.

This may be better as a debugPrint as I'm not sure users will care.

PR #5321, opened 2024-07-02
ant: insert diodes near all gates with CAR violations, except near other diodes

eder-matheus:


src/ant/src/AntennaChecker.cc
```diff
@@ -478,24 +478,24 @@ void AntennaChecker::calculatePAR(GateToLayerToNodeInfo& gate_info)
...
     while (iter_layer) {
-      if (gate_it.second.find(iter_layer) != gate_it.second.end()) {
+      if (layer_to_node_info.find(iter_layer) != layer_to_node_info.end()) {
         if (iter_layer->getRoutingLevel() == 0) {
-          sumVia += gate_it.second[iter_layer];
-          gate_it.second[iter_layer].CAR += sumVia.PAR;
-          gate_it.second[iter_layer].CSR += sumVia.PSR;
-          gate_it.second[iter_layer].diff_CAR += sumVia.diff_PAR;
-          gate_it.second[iter_layer].diff_CSR += sumVia.diff_PSR;
+          sumVia += layer_to_node_info[iter_layer];
```
QuantamHD, line 488: You should just store `layer_to_node_info[iter_layer]` in a variable between line 486 and 487 and then you can use it both the if and else. Plus probably save yourself some look ups.

maliberty (approved):

PR #5322, opened 2024-07-02
mpl2: add debug flag for drawing hierarchical macro placement final result

AcKoucher:


AcKoucher:
@maliberty This should help the private design debugging.

maliberty (approved):
looks good except for the clang-tidy

PR #5326, opened 2024-07-02
ppl: use Rect instead of pair to represent the constraint region

eder-matheus:


src/odb/src/db/dbBTerm.cpp
```diff
@@ -927,19 +927,17 @@ void dbBTerm::staSetVertexId(uint32_t id)
...
 }
 
-std::optional<std::pair<Point, Point>> dbBTerm::getConstraintRegion()
+std::optional<Rect> dbBTerm::getConstraintRegion()
 {
   _dbBTerm* bterm = (_dbBTerm*) this;
   const auto& constraint_region = bterm->_constraint_region;
-  if (constraint_region.first == Point(0, 0)
-      && constraint_region.second == Point(0, 0)) {
+  if (constraint_region == Rect(0, 0, 0, 0)) {
```
maliberty, line 940: While we are improving this, it would be better to use an inverted rect rather than the origin as the no value - see mergeInit & isInverted

maliberty (approved):

Issue #4040, opened 2023-09-25
Make OpenROAD-flow-script's `report_metrics.tcl` part of OpenROAD core


mithro:
### Description

OpenROAD-flow-scripts has a pretty nice [`report_metrics.tcl` script](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/scripts/report_metrics.tcl) which makes nice looking report about important metrics in an design. As many people are likely to want to see these metrics, it would be nice if this was wrapped in a function and provided to anyone using OpenROAD.

### Suggested Solution

Move the function into the OpenROAD core.

vvbandeira:
@mithro 
Not a solution/directly related, but just to make sure you are aware, `openroad` can output some metrics in a JSON format with `openroad -metrics <PATH>.json`.

PR #5328, opened 2024-07-03
Commiting solution and test for issue4040

dav77227:
This PR addresses [OpenROAD_issue_4040](https://github.com/The-OpenROAD-Project/OpenROAD/issues/4040))

Made `report_metrics` procedure part of the OpenROAD. Once this is merged we can remove the `report_metrics` from the flow_scripts.


src/utl/test/test_report_metrics.tcl
```diff
@@ -0,0 +1,4 @@
+read_lef ./../../../test/Nangate45/Nangate45.lef
+read_verilog ./../../../test/gcd_nangate45.v
+link_design gcd
+report_metrics 2 final
```
maliberty, line 4: You should pass `-metrics_report_dir results` to put the result file in the expected location.  You should also have an .rptok file and use diff_files to compare the report.

src/utl/test/test_report_metrics.tcl
```diff
@@ -0,0 +1,4 @@
+read_lef ./../../../test/Nangate45/Nangate45.lef
```
maliberty, line 1: You should do read_liberty and read_sdc as well to get timing values.

src/Metrics.tcl
```diff
@@ -254,5 +254,253 @@ proc report_design_area_metrics {args} {
...
+
+define_cmd_args "report_metrics" {[-stage][-when][-include_erc][-include_clock_skew][-metrics_report_dir]}
+proc report_metrics { stage when {include_erc true} {include_clock_skew true} {metrics_report_dir ""}} {
+  if {[info exists ::env(SKIP_REPORT_METRICS)] && $::env(SKIP_REPORT_METRICS) == 1} {
+    return
+  }
+  puts "Report metrics stage $stage, $when..."
+  if {$metrics_report_dir eq ""} {
+    set metrics_report_dir [file join  $::env(HOME) "OpenROAD" "results"]
+  }
```
maliberty, line 273: Creating directories in the user's home is unexpected. I think '.' is a more reasonable default.

maliberty (changes requested):

src/Metrics.tcl
dav77227: If a user has only read access in the '.' directory, it will not work, but the user usually has write access to his home directory.

maliberty: If they don't they can use -metrics_report_dir.  I would not be happy to use a tool that behaved like this.

dav77227: Well. I updated the code, based on your suggestion.

src/Metrics.tcl
```diff
@@ -254,5 +254,250 @@ proc report_design_area_metrics {args} {
...
+}
+
+define_cmd_args "report_metrics" {[-stage][-when][-include_erc][-include_clock_skew][-metrics_report_dir]}
+proc report_metrics { stage when {include_erc true} {include_clock_skew true} {metrics_report_dir "."}} {
+  if {[info exists ::env(SKIP_REPORT_METRICS)] && $::env(SKIP_REPORT_METRICS) == 1} {
+    return
+  }
+  puts "Report metrics stage $stage, $when..."
+  if {![file isdirectory $metrics_report_dir/results]} {
+    file mkdir ${metrics_report_dir}/results
```
maliberty, line 272: /results is now unnecessary

src/Metrics.tcl
```diff
@@ -254,5 +254,250 @@ proc report_design_area_metrics {args} {
...
 
+proc report_puts { out } {
+  upvar 1 when when
+  upvar 1 filename filename
+  set fileId [open $filename a]
+  puts $fileId $out
+  close $fileId
+}
+
+define_cmd_args "report_metrics" {[-stage][-when][-include_erc][-include_clock_skew][-metrics_report_dir]}
```
maliberty, line 265: [-stage][-when] are both mandatory args not switches

[-metrics_report_dir] should be [-metrics_report_dir dir]

maliberty (changes requested):
Thanks for the change.  Two small points left.

src/Metrics.tcl
```diff
@@ -254,5 +254,250 @@ proc report_design_area_metrics {args} {
...
+  upvar 1 when when
+  upvar 1 filename filename
+  set fileId [open $filename a]
+  puts $fileId $out
+  close $fileId
+}
+
+define_cmd_args "report_metrics" {[-stage][-when][-include_erc][-include_clock_skew][-metrics_report_dir dir]}
+proc report_metrics { stage when {include_erc true} {include_clock_skew true} {metrics_report_dir "."}} {
+  if {[info exists ::env(SKIP_REPORT_METRICS)] && $::env(SKIP_REPORT_METRICS) == 1} {
```
gadfort, line 267: Should this be here? I think this should be removed.

src/Metrics.tcl
```diff
@@ -254,5 +254,250 @@ proc report_design_area_metrics {args} {
...
+    report_puts "\n=========================================================================="
+    report_puts "$when slack div critical path delay"
+    report_puts "--------------------------------------------------------------------------"
+    report_puts "[format "%4f" [expr $path_slack / $path_delay * 100]]"
+  }
+
+  report_puts "\n=========================================================================="
+  report_puts "$when report_power"
+  report_puts "--------------------------------------------------------------------------"
+  if {[info exists ::env(CORNERS)]} {
```
gadfort, line 482: This should use the corners defined in sta

src/Metrics.tcl
dav77227: Am I correct in understanding that checking these variables here makes no sense, and we need to remove these pieces of code from this part?

src/Metrics.tcl
dav77227: So we need to check sta::env(CORNERS) ?

src/Metrics.tcl
maliberty: Correct, the command as a whole should be skipped at the ORFS level based on these vars.

src/Metrics.tcl
gadfort: you need to use `sta::corners` which returns the list of defined corners.

dav77227: Thanks for noticing. I changed the code.

maliberty:
Please note that you need to sign your commits with `git commit -s`.  To fix this PR see the steps in the details link by DCO.

maliberty:
`test_report_metrics (tcl) *FAIL*` needs fixing

dav77227:
> `test_report_metrics (tcl) *FAIL*` needs fixing

The branch has been updated to reflect the latest changes, and the test results have been reviewed and updated accordingly. The tests were successful.

src/Metrics.tcl
```diff
@@ -254,5 +254,250 @@ proc report_design_area_metrics {args} {
...
+    report_puts "\n=========================================================================="
+    report_puts "$when slack div critical path delay"
+    report_puts "--------------------------------------------------------------------------"
+    report_puts "[format "%4f" [expr $path_slack / $path_delay * 100]]"
+  }
+
+  report_puts "\n=========================================================================="
+  report_puts "$when report_power"
+  report_puts "--------------------------------------------------------------------------"
+  if {[info exists sta::CORNERS]} {
```
gadfort, line 482: You don't need to check if this exists. This is masking that you should be using `sta::corners` not `sta::CORNERS`

dav77227: Do I understand correctly that I just need to check sta::corners ? `if {[sta::corners]}`

gadfort: I don't think you need the `if` since that was left over from the ORFS which doesn't always use corners. Sta always defines at least one corner so you can just iterate over those corners.

src/utl/test/test_report_metrics.tcl
```diff
@@ -0,0 +1,6 @@
+read_liberty ./../../../test/Nangate45/Nangate45_fast.lib
```
gadfort, line 1: I think your CI test should include corners

src/Metrics.tcl
dav77227: Thank you for the suggestion! You’re right. Since Sta always defines at least one corner, I’ve removed the condition and now iterate directly over the corners. The updated code has been pushed to the repository.

src/utl/test/test_report_metrics.tcl
dav77227: Do I understand correctly that you mean that we need to call the report_metrics command with the include_erc and include_clock_skew flags?

gadfort: I mean that you need to define a couple of corners and load liberty files in for those
```
define_corners fast slow
read_liberty -corner fast  ./../../../test/Nangate45/Nangate45_fast.lib
read_liberty -corner slow ./../../../test/Nangate45/Nangate45_slow.lib
```
that should allow you to test the `sta::corners` code you have.

And yes, if the function has flags we should be testing that behavior as well.

dav77227: Hi @gadfort. I changed the test on your advice, and now I have two corners in the test:
```
define_corners fast slow
read_liberty -corner fast  ./../../../test/Nangate45/Nangate45_fast.lib
read_liberty -corner slow ./../../../test/Nangate45/Nangate45_slow.lib
read_lef ./../../../test/Nangate45/Nangate45.lef
read_verilog ./../../../test/gcd_nangate45.v
link_design gcd
read_sdc ./../../../test/timing_api_2.sdc
report_metrics 2 final false false
```
But I can't generate the ok file - I get error in the output:
```
[INFO ODB-0227] LEF file: ./../../../test/Nangate45/Nangate45.lef, created 22 layers, 27 vias, 135 library cells
Report metrics stage 2, final...
Metrics results directory: /home/dav77227/OpenROAD/results
[ERROR STA-0102] _f0aa93ee7b550000_p_Corner is not the name of process corner.
Error: test_report_metrics.tcl, 8 STA-0102
ERROR 
```
I suppose that the case has something incorrect. Could you please help to fix the test?

src/Metrics.tcl
```diff
@@ -254,5 +254,245 @@ proc report_design_area_metrics {args} {
   utl::metric_float "design__instance__utilization__stdcell" $stdcell_util
 }
 
+proc report_puts { out } {
```
gadfort, line 257: this probably shouldn't been in the public namespace, it should go into `ord::`

src/Metrics.tcl
```diff
@@ -254,5 +254,245 @@ proc report_design_area_metrics {args} {
...
+  upvar 1 when when
+  upvar 1 filename filename
+  set fileId [open $filename a]
+  puts $fileId $out
+  close $fileId
+}
+
+define_cmd_args "report_metrics" {[-stage][-when][-include_erc][-include_clock_skew][-metrics_report_dir dir]}
+proc report_metrics { stage when {include_erc true} {include_clock_skew true} {metrics_report_dir "."}} {
```
gadfort, line 266: this should be using: https://github.com/The-OpenROAD-Project/OpenROAD/blob/42ba30d730851a73bd2996c5673a05fd79520f87/docs/contrib/TclFormat.md?plain=1#L17

src/Metrics.tcl
```diff
@@ -254,5 +254,245 @@ proc report_design_area_metrics {args} {
...
+
+  report_puts "\n=========================================================================="
+  report_puts "$when report_power"
+  report_puts "--------------------------------------------------------------------------"
+  
+  foreach corner $sta::CORNERS {
+    report_puts "Corner: $corner"
+    report_power -corner $corner >> $filename
+    report_power_metric -corner $corner >> $filename
+  }
```
gadfort, line 487:
```

  foreach corner $sta::corners {
    report_puts "Corner: [$corner name]"
    report_power -corner [$corner name] >> $filename
    report_power_metric -corner [$corner name] >> $filename
  }
```

src/utl/test/test_report_metrics.tcl
```diff
@@ -0,0 +1,6 @@
...
+read_lef ./../../../test/Nangate45/Nangate45.lef
+read_verilog ./../../../test/gcd_nangate45.v
+link_design gcd
+read_sdc ./../../../test/timing_api_2.sdc
+report_metrics 2 final
```
gadfort, line 6: this really should be `report_metrics -stage 2 -when final`? @maliberty that feels somewhat awkward to have a numbering scheme in OpenROAD.
`-stage` and `-when` is really just a name for the report, so maybe that should be renamed/simplified?

src/utl/test/test_report_metrics.tcl
gadfort: see my comment above

gadfort (changes requested):

PR #5338, opened 2024-07-04
ci: update tclint to v0.3.2

nmoroze:
This PR updates `tclint` to v0.3.2. Since CI was a bit behind the latest version, there are a few new categories of violations that require fixes. I've gone ahead and fixed all the newline escape spacing violations in this PR (there weren't that many), but I ignored [`spaces-in-braces`](https://github.com/nmoroze/tclint/blob/main/docs/violations.md#spaces-in-braces) and [`unbraced-expr`](https://github.com/nmoroze/tclint/blob/main/docs/violations.md#unbraced-expr), since there were too many instances of these violations to hit in one go. 

This latest version also includes a bug fix for the issue in `mpl.tcl`. With that fix + updating the indentation pattern to fix the other violations, the comment annotations to disable checks can be removed. 

It might be worth looking into setting up Dependabot to update Python dependencies, which could help stay up to date with `tclint` going forward.

### DIFF
--- a/tclint.toml
+++ b/tclint.toml
@@ -35,8 +35,14 @@
     "src/utl/test/"
 ]
 
+ignore = [
+    "spaces-in-braces",
+    "unbraced-expr",
+]
+
 [style]
 indent = 2
 line-length = 100
 allow-aligned-sets = true
 indent-namespace-eval = false
+spaces-in-braces = true
--- a/.github/workflows/github-actions-lint-tcl.yml
+++ b/.github/workflows/github-actions-lint-tcl.yml
@@ -12,7 +12,7 @@ jobs:
 
       - name: Install tclint
         run: |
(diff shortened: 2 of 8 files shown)


tclint.toml
```diff
@@ -35,8 +35,14 @@
...
+    "spaces-in-braces",
+    "unbraced-expr",
+]
+
 [style]
 indent = 2
 line-length = 100
 allow-aligned-sets = true
 indent-namespace-eval = false
+spaces-in-braces = true
```
nmoroze, line 48: Even though this check is being ignored for now, this setting reduces the overall number of violations and will make it easier to transition to include this check.

maliberty (approved):
Thanks for the quick fix and the cleanup!

maliberty:
Do you have experience setting up Dependabot ?

PR #5345, opened 2024-07-04
test: set thread count before grt to use parallel repair_antennas

eder-matheus:


test/flow.tcl
```diff
@@ -191,6 +191,8 @@ write_verilog $verilog_file
 ################################################################
 # Global routing
 
+set_thread_count [exec getconf _NPROCESSORS_ONLN]
```
maliberty, line 194: Is there a reason not to put this at the start of the flow?  It won't make much difference today but it could in the future.

AcKoucher: The problems with mpl2 generating different results for different number of threads were resolved a while ago. It should be fine.

maliberty (approved):

eder-matheus:
@maliberty I see this failure in the CI:
```
./test/regression jpeg_sky130hd
jpeg_sky130hd (tcl) *FAIL* RSZ::hold_buffer_count   66 >   50
```
But I can't reproduce it on my GCP machine. We have another source of possible non-determinism since this metric is generated before grt. I can't download the artifacts of this design from the CI. Is there a way to access it? It would be nice to see where the discrepancies begin.

maliberty:
Does https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-5345-head/3/artifact/test/results.tgz contain it?

eder-matheus:
> Does https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-5345-head/3/artifact/test/results.tgz contain it?

It has only for sky130hs/gcd.

eder-matheus:
Closing it due to instability in the results. Will create a new PR from the latest master branch to check if the instability persists.

Issue #5325, opened 2024-07-02
Repair Antennas non-deterministic/faulty results


osamahammad21:
### Describe the bug

In https://github.com/The-OpenROAD-Project/OpenROAD/pull/5281 The CI is giving different result for sky130hd_ibex than my local machine run. The differences start at repair_antennas part, while the groute guides are identical. What's more weird is that at the end of repair_antennas, the number of antenna violations reported is 0 while after check antennas which is called right after it, the antenna violations are 1 (This is on the CI's result). 

### Expected Behavior

Same number of antenna violations reported by repair_antennas and check_antennas
Same result on my machine and the CI's

### Environment

```shell
kernel: Linux 6.5.0-41-generic
os: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
cmake version 3.24.2
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-14378-gc1c65bb8c
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.10 (found version "3.10.12") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
[...]

PR #5346, opened 2024-07-04
ant: consider violations in cut layers for repair_antennas

eder-matheus:
Partially fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5325

The ant API was not returning violations related to CUT layers, leading to the mismatch between the repair_antennas report and the check_antennas report. This simple update fixes it, and the mismatch reported in the issue is not happening anymore (all violations are fixed).


maliberty:
What does it mean to have a violation on a cut layer?

maliberty (approved):

maliberty:
I'm ok with this as a solution to the current problem but separately I want to clarify what the checker is checking.

eder-matheus:
> What does it mean to have a violation on a cut layer?

The violation that repair_antennas wasn't seeing was a PAR violation in the cut layer via4. The calculated PAR on that layer was higher than the allowed. We were not checking the cut layers in the ant code, so repair_antennas never knew such a violation existed. Now, we are returning PAR violations on both routing and cut layers.

eder-matheus:
> I'm ok with this as a solution to the current problem but separately I want to clarify what the checker is checking.

Does my last comment clarify this or do you have other questions about the ant check?

maliberty:
I didn't think we have par/car values on cut layers.  Is there a PDK that has those?

eder-matheus:
> I didn't think we have par/car values on cut layers.  Is there a PDK that has those?

The design that had the violation was sky130hd/ibex. I didn't check the LEF, but I believe I also saw PAR values in cut layers for ihp platform.

eder-matheus:
Here's the LEF definition for via4 in sky130hd:
image

maliberty:
You would think the area on the cut layer would be rather small.  I guess such violation occur based on this PR but I'm surprised.

PR #5348, opened 2024-07-05
mpl2: Add C++ test for snapMacro 

amuritna:
Add test that will input a macro instance with various "invalid" origins (not yet aligned with the track grid) and then check whether the new origin, after running setMacro->snapMacro, is as expected.

Tagging my GSoC mentors: @vvbandeira @maliberty 
and @AcKoucher


AcKoucher:
@maliberty I made a feel general comments.

@amuritna I feel that there's too much setup code in the test. This code should be shared among tests. Methods to instantiate odb objects such as the block, layer, pins, etc. shouldn't be repeated if possible.

amuritna:
> @amuritna I feel that there's too much setup code in the test. This code should be shared among tests.

I started creating helper functions when working on the Pusher tests (#5362) and was thinking of updating the Snapper tests after that PR got merged, but now that I've thought about it there's no reason I can't do that now. I'll do it in a bit.

AcKoucher (changes requested):

src/mpl2/test/cpp/TestSnapper.cpp
```diff
@@ -57,16 +57,155 @@ TEST_F(Mpl2SnapperTest, CanSetMacroForEmptyInstances)
...
   snapper.setMacro(inst3);
-}
+}  // CanSetMacroForEmptyInstances
+
+TEST_F(Mpl2SnapperTest, CanSnapMacros)
+{
+  // This test checks the snapMacro functionality of the
+  // Snapper class defined in hier_rtlmp.cpp.
+  //
+  // snapMacro works by computing a valid origin for a
```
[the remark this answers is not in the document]
amuritna: You're right, the wording isn't right. I've edited it to the following; does this work?
```
snapMacro works by adjusting the position of the
macro so that the signal pins are aligned with the
track-grid.

Therefore, this test inputs instances that have
not been aligned correctly as testcases, and then
checks whether snapMacro has aligned said instances
correctly.
...

At the moment, this test only receives instance unaligned with
the track-grid as inputs, because right now, Snapper will not
necessarily compute the most optimal origin for instances that
are  already aligned (origins should've been left unchanged).
```

src/mpl2/test/cpp/TestSnapper.cpp
```diff
@@ -31,42 +32,254 @@ class Mpl2SnapperTest : public ::testing::Test
...
+
+// At the moment, these tests only receive instances unaligned with
+// the track-grid as inputs, because right now, Snapper will not
+// necessarily compute the most optimal origin for instances that
+// are already aligned (origins should've been left unchanged).
+//
+// However, this might change in the future if Snapper is optimized
+// further.
+
+TEST_F(Mpl2SnapperTest, SnapMacrosHorizontal)
```
AcKoucher, line 83: Here the name  "HorizontalSnap"  seems better.

src/mpl2/test/cpp/TestSnapper.cpp
```diff
@@ -31,42 +32,254 @@ class Mpl2SnapperTest : public ::testing::Test
...
+  EXPECT_TRUE(inst->getOrigin().y() == 535);
+
+  inst->setOrigin(519, 579);
+  snapper.snapMacro();
+  EXPECT_TRUE(inst->getOrigin().x() == 515);
+  EXPECT_TRUE(inst->getOrigin().y() == 535);
+
+}  // SnapMacrosHorizontal
+
+TEST_F(Mpl2SnapperTest, SnapMacrosVertical)
```
AcKoucher, line 141: "VerticalSnap"

src/mpl2/test/cpp/TestSnapper.cpp
```diff
@@ -31,42 +32,254 @@ class Mpl2SnapperTest : public ::testing::Test
...
+  EXPECT_TRUE(inst->getOrigin().y() == 515);
+
+  inst->setOrigin(579, 519);
+  snapper.snapMacro();
+  EXPECT_TRUE(inst->getOrigin().x() == 535);
+  EXPECT_TRUE(inst->getOrigin().y() == 515);
+
+}  // SnapMacrosVertical
+
+TEST_F(Mpl2SnapperTest, SnapMacrosDouble)
```
AcKoucher, line 198: The idea of this test should be having one macro with one pin in L1 and other in L2 not two macros with both pins in the same layer.

src/mpl2/test/cpp/TestSnapper.cpp
```diff
@@ -31,42 +32,254 @@ class Mpl2SnapperTest : public ::testing::Test
...
+
+  Snapper snapper;
+  snapper.setMacro(inst);
+
+  // Considering the grid pattern configuration (0 20 40 ... 980)
+  // manufacturing grid size (5), and direction preference (horizontal):
+  // Valid alignments for x would include 15 35 55 75 95 ...
+  // with 40 to 59 snapping to 15, 60 to 79 mapping to 35, etc.
+  // Valid alignments for y would include 10 15 20 25 30 35 40 ...
+  // with 14 to 11 snapping to 10, 19 to 16 snapping to 15, etc.
```
AcKoucher, line 174: We should be able to see clearly the pins dimensions from the scope of the test in order to avoid having to read a long comment such as this one in order to figure it out.

AcKoucher (changes requested):

src/mpl2/test/cpp/TestSnapper.cpp
amuritna: I've re-organized the helper functions so that the MPin instantiations are now also "visible", does this work?

Issue #4857, opened 2024-03-26
PDN short circuiting Metal1 in endcap/tbtie cells


gatecat:
### Description

We are using OpenROAD (with ORFS) to tape out on a proprietary PDK. The cell library contains endcap, corner and tbtie cells that contain a metal1 ring (that needs to be connected to VSS in our case).

Those cells are placed fine using the `tapcell` command, but the metal1 PDN stripes generated by `add_pdn_stripe -followpins` don't take this into account and short onto the metal1.

Is there a way of avoiding skipping the first/last metal1 stripe and trimming the length of the stripes, such that they don't short the metal1 inside this tap cell ring, other than ad hoc patching the PDN generation?

Incidentally, what would be the nicest way to connect a pin in every cell in this ring to VSS, and generate the necessary routing for that?

maliberty:
A few pictures would help as I don't understand your description.  Normally tapcells/endcaps have power pins in the same location as any other standard cell and follow pins isn't an issue.

gatecat:
Please excuse my "not to scale" sketch to avoid violating any NDAs, hope this still helps.

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/78621419/99e51c40-7cc1-4291-a640-e748e0bfb78e)

As well as the standard cell power strips, there is also a "guard ring" that connects a continuous ring of metal1 (pins called DTI in the LEF) around the edge of the core area - this needs to be connected to P-Substate bias e.g. Vss - but at the moment, it's also shorted to the extended Vdd pins.

maliberty:
That's quite odd, if I look at the gf180 endcap I see:

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/761514/14cedac2-8616-4f64-9262-d0081f52910a)

The rails are not going to be shorted due to the gap.  Are you saying your endcap cells has a metal strap between the rails?

maliberty:
In the gds for m1:
![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/761514/def4209f-a248-4feb-8989-772805970fd4)

gatecat:
Nope - the rails inside the cell end well before that Metal1 ring (which is outside the placement boundary of the cell) - it's OpenROAD that's extending the rails such that they short the outer ring.

maliberty:
How did you place cells outside the core area?  tapcell will not normally do that.

maliberty:
[...]

PR #5349, opened 2024-07-05
[WIP] pdn: Add option to cut followpins around cell obstructions

gatecat:
This is not yet finished, but I'd appreciate feedback on what I've got so far. It's a solution to https://github.com/The-OpenROAD-Project/OpenROAD/issues/4857 which affects the gf130 PDK, where we have a ring of cells containing a metal1 ring around the core (DTI, connected to VSS), that the PDN will short to causing a VSS-VDD short without a workaround.

Some specific things I'd appreciate feedback on:

 - the name of the option
  - currently I have a hack to prevent very short stubs being generated filtering out segments whose length is shorter than `20 * width`. Another solution to this could be to take the largest segment in the middle only (i.e. instead of cutting the shape, shortening it so it excludes obstructions at the left and right edges), but that's less generic in another case where there's an obstruction somewhere in the middle of the core
  - to prevent the shapes being "repaired" back to full length and the short being reintroduced when creating vias, I have to call `setLocked` on it. This seems a bit extreme, but the only other thing I can think of is adding the obstructions to the general obstruction set (but then having more overhead and having to deal with power/ground same-net filtering carefully, as well as the above issue of short stubs).

I still need to add a test once this is a bit more final, this will involve creating something that replicates the problem but with an open PDK - not sure if any specific PDK is preferred for tests like this.

Thanks for any feedback or help!


src/pdn/src/straps.cpp
```diff
@@ -383,6 +385,59 @@ void FollowPins::makeShapes(const Shape::ShapeTreeMap& other_shapes)
...
+      }
+
+      const odb::dbTransform transform = inst->getTransform();
+
+      for (odb::dbITerm *iterm : inst->getITerms()) {
+
+        bool is_power = iterm->getNet() == power;
+        bool is_ground = iterm->getNet() == ground;
+
+        // Non-power pins might obstruct
```
maliberty, line 403: This will generate a huge amount of obstructions.  Have you looked at the runtime/memory impact?

gatecat: As this is being run straight after tapcell, there aren't that many placed cells other than those - I wasn't noticing any significant impact on our medium (~100k cells) design.

maliberty: That's true in ORFS but nothing prevents someone from loading a design with pre-placed cells or many macros.  Would it be reasonable to limit the check to endcaps?

src/pdn/src/straps.cpp
```diff
@@ -383,6 +385,59 @@ void FollowPins::makeShapes(const Shape::ShapeTreeMap& other_shapes)
   const int x_start = boundary.xMin();
   const int x_end = boundary.xMax();
   odb::dbTechLayer* layer = getLayer();
+
+  std::vector<std::shared_ptr<Shape>> power_obs, ground_obs;
+  if (getExtendMode() == CLIP_CELLS) {
+    for (odb::dbInst *inst : grid->getBlock()->getInsts()) {
+      if (!inst->isPlaced() || !boundary.contains(inst->getBBox()->getBox())) {
+        continue;
+      }
```
gadfort, line 394: this needs to be limited to fixed cells only.

src/pdn/src/straps.cpp
```diff
@@ -383,6 +385,59 @@ void FollowPins::makeShapes(const Shape::ShapeTreeMap& other_shapes)
   const int x_start = boundary.xMin();
   const int x_end = boundary.xMax();
   odb::dbTechLayer* layer = getLayer();
+
+  std::vector<std::shared_ptr<Shape>> power_obs, ground_obs;
```
gadfort, line 389: why keep them seperate?

src/pdn/src/straps.cpp
```diff
@@ -383,6 +385,59 @@ void FollowPins::makeShapes(const Shape::ShapeTreeMap& other_shapes)
...
+          auto shape = std::make_shared<Shape>(obstr->getTechLayer(), rect, Shape::OBS);
+          shape->generateObstruction();
+          power_obs.emplace_back(shape);
+          ground_obs.emplace_back(shape);
+        }
+      }
+    }
+  }
+
+  Shape::ObstructionTree power_obs_tree(power_obs), ground_obs_tree(ground_obs);
```
gadfort, line 439: minor formatting, I would prefer these on separate lines

src/pdn/src/straps.cpp
```diff
@@ -383,6 +385,59 @@ void FollowPins::makeShapes(const Shape::ShapeTreeMap& other_shapes)
   const int x_start = boundary.xMin();
   const int x_end = boundary.xMax();
   odb::dbTechLayer* layer = getLayer();
+
+  std::vector<std::shared_ptr<Shape>> power_obs, ground_obs;
+  if (getExtendMode() == CLIP_CELLS) {
+    for (odb::dbInst *inst : grid->getBlock()->getInsts()) {
+      if (!inst->isPlaced() || !boundary.contains(inst->getBBox()->getBox())) {
```
gadfort, line 392: move this into a private function

src/pdn/src/straps.cpp
```diff
@@ -404,12 +459,40 @@ void FollowPins::makeShapes(const Shape::ShapeTreeMap& other_shapes)
...
+            cut_shape->setLocked();
+            addShape(cut_shape);
+          }
+        }
+      } else {
+        addShape(power_strap);
+      }
+    } else {
+      addShape(power_strap);
+    }
```
gadfort, line 476: this is pretty ugly, it might be cleaner to have the `FollowPins` class override the `addShape` to make this a little less messy.

src/pdn/src/straps.cpp
```diff
@@ -404,12 +459,40 @@ void FollowPins::makeShapes(const Shape::ShapeTreeMap& other_shapes)
...
     auto* ground_strap = new FollowPinShape(
         layer, ground, odb::Rect(x0, ground_y_bot, x1, ground_y_bot + width));
     ground_strap->addRow(row);
-    addShape(ground_strap);
+    if (getExtendMode() == CLIP_CELLS) {
+      std::vector<Shape*> replacements;
+      if (ground_strap->cut(ground_obs_tree, grid, replacements)) {
+        for (Shape *cut_shape : replacements) {
+          if (cut_shape->getRect().dx() > 20 * width) {
+            cut_shape->setLocked();
```
gadfort, line 486: I think we need to dig into the issue that is forcing you to add this.
It might be helpful to have a testcase that demonstrate the issue so we can debug it.

src/pdn/src/pdn.tcl
```diff
@@ -289,7 +289,7 @@ sta::define_cmd_args "add_pdn_stripe" {[-grid grid_name] \
 proc add_pdn_stripe {args} {
   sta::parse_key_args "add_pdn_stripe" args \
     keys {-grid -layer -width -pitch -spacing -offset -starts_with -number_of_straps -nets} \
-    flags {-followpins -extend_to_core_ring -extend_to_boundary -snap_to_grid}
+    flags {-followpins -extend_to_core_ring -extend_to_boundary -clip_around_cells -snap_to_grid}
```
gadfort, line 292: needs to be added to the `sta::` list above and documented in the README.

src/pdn/src/pdn.tcl
```diff
@@ -366,6 +370,8 @@ proc add_pdn_stripe {args} {
     set extend "Rings"
   } elseif {[info exists flags(-extend_to_boundary)]} {
     set extend "Boundary"
+  } elseif {[info exists flags(-clip_around_cells)]} {
+    set extend "Clip_Cells"
```
gadfort, line 374: no `_`.

gadfort (changes requested):
@maliberty makes a good point about the runtime impact on large designs, although all this code only gets triggered if this option is on and I have yet to encounter a PDK where this was required out in the wild.
A CI test should be included.

src/pdn/src/straps.cpp
gatecat: Hopefully this simplified illustration helps from a public github point of view - the detailed testcase is covered by NDA but could also be discussed further by email if necessary

image

gadfort: since you will need to make a test for this PR, you can just use that test as the example of the bad behavior.

gatecat:
Thanks for your comments - I hope I've addressed them, and also added a test with a replica of the situation we're facing reproduced in nangate45.

maliberty:
Please merge master into your branch to resolve the CI issue.

src/pdn/src/straps.cpp
```diff
@@ -347,6 +348,98 @@ FollowPins::FollowPins(Grid* grid, odb::dbTechLayer* layer, int width)
...
+
+  const int width = getWidth();
+  auto* grid = getGrid();
+
+  if (shape->cut(clip_obstructions, grid, replacements)) {
+    for (Shape* cut_shape : replacements) {
+      if (cut_shape->getRect().dx() > 20 * width) {
+        cut_shape->setLocked();
+        addShape(cut_shape);
+      } else {
```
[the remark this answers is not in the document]
gatecat: Yeah - because of the two-rings nature, which mimicks our PDK, we get this little unconnected stub.

image

maliberty:
Is this still WIP/draft?

gatecat:
Yes - for this to be mergeable I still need to replace the `> 20 * width` hack for eliminating small stub segments. My current plan, short of alternative suggestions, is to only consider the case where the obstructions are at the edges, and only include the largest segment in the middle.

maliberty:
Are there rows in the space where the stub goes?  If you don't want to power them then you should remove them when generating the floorplan.

gatecat:
It's a short stub - about 1.5um long at either end of each row. If pruning the length of the rows so that only go as far as the power pins of the endcaps rather than all the way to edge is possible, then maybe this PR isn't necessary at all?

maliberty:
Is it valid to only connect to the corner of the endcap?  You could potentially connect to the stub from above.  What does the power pin shape look like for the endcap?

gatecat:
Not sure if I've been explaining this great, but the power stripes of the endcaps are only a small section on the inner edge (what's selected in this screenshot) - it's perfectly fine to connect to them normally
image

so long as the PDN strip is kept short enough that it doesn't short with the parts that belong with the ring.

maliberty:
I'm trying to understand if this case is actually invalid:
image

Is that stub actually covering a power pin in the endcap?  If so it should be connected.  If not then there is no need for the follow pin stripe to cover that area.

gatecat:
The stub is totally floating, not connected to any pins and potentially not reached by any vias, and shouldn't be there indeed - the ongoing discussion is a better heuristic to avoid generating these stubs other than the current hack in this pull request of pruning anything shorter than 20*width.

I think a reasonable approach that will work here is to prune any segments produced as a result of cutting that are exclusively over a fixed cell, and isn't connecting to any pins of that cell.

gatecat:
Seems like a good solution to me

gatecat:
ping @gadfort  ^ ?

gadfort:
@gatecat sorry this slipped by. @maliberty I think the cutting is alright, but maybe we need to add a check to ensure the followpin is actually overlapping a pin otherwise remove it (this might cause some trouble with power grids where we have M1 and M2 as followpins, but only M1 is a pin layer). Another possibility would be to only enable this check when the cutting option is enabled (avoiding the M2 problem)?

gatecat:
See https://github.com/The-OpenROAD-Project/OpenROAD/pull/5457

PR #5350, opened 2024-07-05
mpl2: create class for clustering engine

AcKoucher:
With this, we'll be able to implement unit tests only for clustering.

There was a discrepancy between thresholds and cached constants in MPL2 as they were defined/set in both Cpp and Tcl. With the changes here, everything is being set in Tcl - my intentions were to keep the setters that come from Tcl to allow future auto-tuning.


AcKoucher:
@maliberty There was a problem with the macro_only regression test that was solved by the last change I pushed.

Running Secure-CI.

src/mpl2/src/clusterEngine.cpp
```diff
@@ -0,0 +1,2160 @@
...
+  design_metrics_ = design_metrics;
+}
+
+void ClusteringEngine::setTree(PhysicalHierarchy* tree)
+{
+  tree_ = tree;
+}
+
+// Fetch the design's logical data
+void ClusteringEngine::fetchDesignMetrics()
```
maliberty, line 101: This is an odd name as nothing is fetch and it is confusing with computeMetrics.  What is the conceptual operation here?

src/mpl2/src/clusterEngine.cpp
```diff
@@ -0,0 +1,2160 @@
...
+    std_cell_area += metrics->getStdCellArea();
+    num_macro += metrics->getNumMacro();
+    macro_area += metrics->getMacroArea();
+  }
+
+  Metrics* metrics
+      = new Metrics(num_std_cell, num_macro, std_cell_area, macro_area);
+  tree_->maps.module_to_metrics[module] = metrics;
+
+  return metrics;
```
maliberty, line 214: It is better to use a unique_ptr to transfer ownership.

src/mpl2/src/clusterEngine.cpp
```diff
@@ -0,0 +1,2160 @@
...
+    float inst_area = computeMicronArea(inst);
+
+    if (master->isBlock()) {  // a macro
+      num_macro += 1;
+      macro_area += inst_area;
+
+      // add hard macro to corresponding map
+      HardMacro* macro
+          = new HardMacro(inst, tree_->halo_width, tree_->halo_width);
+      tree_->maps.inst_to_hard[inst] = macro;
```
maliberty, line 191: Is inst_to_hard the owner of this memory?  If so it should use a unique_ptr.  If not, who is the owner?

src/mpl2/src/clusterEngine.cpp
```diff
@@ -0,0 +1,2160 @@
...
+      design_metrics_->getStdCellArea() + design_metrics_->getMacroArea(),
+      core_area,
+      util,
+      core_util,
+      block_->getTech()->getManufacturingGrid());
+}
+
+void ClusteringEngine::initTree()
+{
+  tree_->root = new Cluster(id_, std::string("root"), logger_);
```
maliberty, line 250: Same about ownership and unique_ptr

src/mpl2/src/clusterEngine.cpp
```diff
@@ -0,0 +1,2160 @@
...
+        odb::dbInst* inst = iterm->getInst();
+        tree_->maps.bterm_to_inst[bterm] = inst;
+      }
+    }
+  }
+}
+
+// Dataflow is used to improve quality of macro placement.
+// Here we model each std cell instance, IO pin and macro pin as vertices.
+void ClusteringEngine::createDataFlow()
```
maliberty, line 524: This is rather long and should be broken down

src/mpl2/src/clusterEngine.cpp
```diff
@@ -0,0 +1,2160 @@
...
+
+  if (cluster->getClusterType() == HardMacroCluster) {
+    cluster->setMetrics(
+        Metrics(0, metrics.getNumMacro(), 0.0, metrics.getMacroArea()));
+  } else if (cluster->getClusterType() == StdCellCluster) {
+    cluster->setMetrics(
+        Metrics(metrics.getNumStdCell(), 0, metrics.getStdCellArea(), 0.0));
+  } else {
+    cluster->setMetrics(metrics);
+  }
```
maliberty, line 1086: The cluster knows its type so the caller should only need to do `cluster->setMetrics(metrics)` and the setMetrics method should handle the rest.

maliberty (changes requested):
diffs do a poor job for code movement so I'll leave correctness to the CI.  I've made some general code comments.

src/mpl2/src/clusterEngine.cpp
AcKoucher: The conceptual operation would be to populate the Metrics object stored in HierRTLMP with the design's data.

maliberty: computeDesignMetrics perhaps?  Though computeMetrics might need a more specific name to distinguish what it does from this.

AcKoucher: Within this logic, computeMetrics should be computeModuleMetrics. I usually like to use computeSomething for getters that return an object, that's why I thought of fetch. But I think computeDesignMetrics and computeModuleMetrics seems good. What do you think?

maliberty: fetch sounds like a getter to me though I always use get naming for that purpose.  I can see expecting compute to return a result.  

I think the problem is that this method is a mixture of things: calling computeMetrics, reporting, assigning macro_with_halo_area , and error checking.   Making it more single purpose would probably ease naming.

AcKoucher: After thinking about this, I thought it would be better to remove this fetchDesignMetrics as we'll actually populate the Metrics when doing
```
design_metrics_ = computeModuleMetrics(block_->getTopModule());
```
 So we'll have an init() function to check if macro placement is both needed and feasible after it. We'd do this reporting for the user at the end of this function. With that said this should be resolved.

src/mpl2/src/clusterEngine.h
```diff
@@ -0,0 +1,260 @@
...
+  const int register_dist = 5;
+  const float factor = 2.0;
+  const float weight = 1;
+
+  std::vector<std::pair<odb::dbITerm*, std::vector<std::set<odb::dbInst*>>>>
+      macro_pin_to_regs;
+  std::vector<std::pair<odb::dbBTerm*, std::vector<std::set<odb::dbInst*>>>>
+      io_to_regs;
+  std::vector<std::pair<odb::dbITerm*, std::vector<std::set<odb::dbInst*>>>>
+      macro_pin_to_macros;
```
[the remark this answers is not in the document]
AcKoucher: I think I don't quite understand the question. It's exhaustive in the sense that we traverse all signal pins so we take into account all Insts (macros and stdcells) and IOs.

However, we only consider the connections between
IO to Registers
Macro Terms to Registers
Macro Terms to Macros

I.e., we won't be adding connections between clusters based on reg_to_macro_pin, reg_to_io, etc.

AcKoucher:
I resolved the requests related to memory management, because, to not further overload this PR by mixing refactoring with other changes, I'll be opening a subsequent PR for switching from manual to automatic memory management.

src/mpl2/src/clusterEngine.cpp
AcKoucher: These checks are not even needed. The module Metrics is populated according to its instances so if a module has e.g., no std cells its Metrics will also not have std cells and it's not needed to create a new Metrics for HardMacroCluster as the Metrics created earlier in the function will have the correct data.

AcKoucher:
Running Secure-CI

maliberty (approved):

PR #5353, opened 2024-07-08
Add format_as for compilation using fmt v10

kamilrakoczy:
This MR fixes compilation of OpenROAD with fmt v10. It uses the same approach as in: https://github.com/The-OpenROAD-Project/OpenROAD/pull/4351

### DIFF
--- a/src/dpl/src/CheckPlacement.cpp
+++ b/src/dpl/src/CheckPlacement.cpp
@@ -50,6 +50,8 @@ using std::vector;
 
 using utl::DPL;
 
+using utl::format_as;
+
 void Opendp::checkPlacement(const bool verbose,
                             const bool disallow_one_site_gaps,
                             const string& report_file_name)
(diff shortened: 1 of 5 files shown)


maliberty (approved):

rovinski:
If this really works then it should be possible to remove the ugly homebrew hack: https://github.com/The-OpenROAD-Project/OpenROAD/blob/943ac03e173b8e5b250dce913c303cb627b63617/etc/DependencyInstaller.sh#L516-L519

I don't have a mac to test it on anymore though.

maliberty:
@eder-matheus @vvbandeira can you test on a mac per @rovinski above?

eder-matheus:
> @eder-matheus @vvbandeira can you test on a mac per @rovinski above?

I've tested it now, and it builds with no errors on my Mac. I will create a PR removing it.

PR #5357, opened 2024-07-08
test: set thread count for the whole flow

eder-matheus:


eder-matheus:
@maliberty I had to set the thread count after the GPL run because it uses omp, but setting more than one thread leads to different results and some metric failures.

maliberty (approved):

maliberty:
Ok but we should track down and fix that as parallelism shouldn't change the result.

eder-matheus:
> Ok but we should track down and fix that as parallelism shouldn't change the result.

Sure, I'll create an issue in GH to keep track of it.

PR #5358, opened 2024-07-08
ci: eqy always uses yosys version

vvbandeira:


etc/DependencyInstaller.sh
```diff
@@ -10,7 +10,7 @@ _versionCompare() {
 
 _equivalenceDeps() {
     yosysVersion=yosys-0.33
-    eqyVersion=8327ac7
+    eqyVersion=702af89
```
maliberty, line 13: From https://github.com/YosysHQ/eqy/issues/67
```
Please make sure to always update to the EQY version tagged yosys-0.xy whenever you update to yosys 0.xy.
```
It sounds like we should use the same tag from yosysVersion

maliberty (changes requested):

maliberty (approved):

PR #5367, opened 2024-07-11
init: HOME can be unset

oharboe:
getenv(HOME) can actually return null. Other places in OpenROAD, such as OpenSTA, this is checked.

Typical use-cases are highly controlled build environments, like bazel hermetic builds.


oharboe:
@hovind FYI

maliberty (approved):

PR #5370, opened 2024-07-11
gpl: ignore clock gating cells in mbff clustering

maliberty:


maliberty:
@mikesinouye @QuantamHD FYI

mikesinouye:
Great, thanks for the investigation and quick fix!

vvbandeira (approved):

PR #5372, opened 2024-07-12
etc: remove pinned fmt and spdlog versions for mac build

eder-matheus:


etc/DependencyInstaller.sh
```diff
@@ -513,10 +513,6 @@ EOF
     # Lemon is not in the homebrew-core repo
     brew install The-OpenROAD-Project/lemon-graph/lemon-graph
 
-    # Install fmt 8.1.1 because fmt 9 causes compile errors
-    _installHomebrewPackage "fmt" "8643c850826702923f02d289e0f93a3b4433741b" ""
-    # Install spdlog 1.9.2
-    _installHomebrewPackage "spdlog" "0974b8721f2f349ed4a47a403323237e46f95ca0" ""
```
rovinski, line 519: Doesn't these have to be added as a normal package on line 508?

maliberty (approved):

PR #5397, opened 2024-07-18
gui: add save DB action

gadfort:
Adds save option in GUI toolbar menu to write out the database (mirrors open option).

### DIFF
--- a/src/gui/src/mainWindow.cpp
+++ b/src/gui/src/mainWindow.cpp
@@ -432,6 +432,8 @@ void MainWindow::setBlock(odb::dbBlock* block)
     const std::string title
         = fmt::format("{} - {}", window_title_, block->getName());
     setWindowTitle(QString::fromStdString(title));
+
+    save_->setEnabled(true);
   }
   for (auto* heat_map : Gui::get()->getHeatMaps()) {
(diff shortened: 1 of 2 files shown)


src/gui/src/mainWindow.cpp
```diff
@@ -1614,6 +1620,22 @@ void MainWindow::openDesign()
...
+      this, "Save Design", QString(), filefilter);
+
+  if (file.isEmpty()) {
+    return;
+  }
+
+  try {
+    ord::OpenRoad::openRoad()->writeDb(file.toStdString().c_str());
+  } catch (const std::exception&) {
+  }
```
maliberty, line 1636: On an exception we should give some indication - perhaps an error dialog with the message.

gadfort: if an error occurs it will be printed via the logger, does it make sense to also display a dialog box? we don't do that anywhere else in the gui code.
The reason for the try/catch is to avoid a fault when openroad throws an exception during an error.

maliberty: Ok so long as it appears somewhere

maliberty (approved):

PR #5398, opened 2024-07-18
pdn: separate repair channels area and repairable area to avoid failing on channels with partial overlap

gadfort:
Fixes:
- repair channels would fail if there is partial overlap in narrow channels, this separates the channel area and the actually repairable area so both can be tracked separately


maliberty:
Can you create a unit test for this?

gadfort:
@maliberty unit test created

maliberty (approved):

PR #5401, opened 2024-07-18
gui: fix exit logic on hide

gadfort:
Corrects the behavior of the GUI when hide is selected.
Fixes one bad format error in PDN debug print.

### DIFF
--- a/src/pdn/src/straps.cpp
+++ b/src/pdn/src/straps.cpp
@@ -1661,7 +1661,7 @@ void RepairChannelStraps::determineParameters(
           utl::PDN,
           "Channel",
           2,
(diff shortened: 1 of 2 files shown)


src/pdn/src/straps.cpp
```diff
@@ -1661,7 +1661,7 @@ void RepairChannelStraps::determineParameters(
           utl::PDN,
           "Channel",
           2,
-          "Failed on channel width check, group {:.4f} and channel {.4f}.",
+          "Failed on channel width check, group {:.4f} and channel {:.4f}.",
```
maliberty, line 1664: sneaky

maliberty (approved):

maliberty:
Thanks!

PR #5403, opened 2024-07-20
psm: avoid warnings for pins without any geometries

gadfort:
Changes:
- no functional changes, just avoids generating a node when a pin has no geometries, like well contacts.
- move some files around in the insert decap tests to be consistent with other tests


maliberty (approved):

gadfort:
@maliberty the failure seems to be unrelated to my PR (it's coming from GRT.congestion1 (py)).

PR #5409, opened 2024-07-21
gui: add enablement for painting all layers

gadfort:
Changes:
- moves implant layers into a sub category of layers (in case of a lot of implant layers this allows the routing and cut layers to be a little easier to find.)
- adds masterslice and other groups to enable non-implant layers too, there are very few shapes in this so should not add much to the render time.
- masterslice and other are not enabled by default

Before):
Screenshot 2024-07-21 111958

New default view:
Screenshot 2024-07-21 111912

Expanded:
Screenshot 2024-07-21 112226


maliberty (approved):

gadfort:
@maliberty that failure looks to be unrelated to this PR.

PR #5410, opened 2024-07-21
gui: add layers to ITerm descriptor

gadfort:
Adds:
- list of layers an iterm can be found on. For most it will only be one, but on memory macros and IO pads it can be be several

### DIFF
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -797,9 +797,9 @@ Descriptor::Properties DbMasterDescriptor::getProperties(std::any object) const
   if (site != nullptr) {
     props.push_back({"Site", gui->makeSelected(site)});
   }
-  std::vector<std::any> mterms;
+  SelectionSet mterms;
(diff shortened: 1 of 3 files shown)


maliberty:
This is more a property of the mterm than than the iterm.  Does it make sense to put it there?

gadfort:
@maliberty we could, but there is no mterm descriptor at the moment. I could add it in this PR and move the layers there.

gadfort:
@maliberty I created an MTerm descriptor and moved that information there. This opens up the possibility to expose some of the antenna model information in the future.

maliberty (approved):

maliberty:
Now that  "IO type" is on the mterm I don't think we need it on iterm

gadfort:
@maliberty both IO type and Signal type come from the MTerm, so both could be removed if we wanted it.

maliberty:
I think we should do so - do you want to make the change or should I?

gadfort:
@maliberty I can do it, should only take a few minutes

Issue #5415, opened 2024-07-22
Make `info script` return path to script executing


oharboe:
### Description

```
$ cat blah.tcl
puts "Script running: [info script]"
$ tclsh blah.tcl 
Script running: blah.tcl
$ openroad -exit blah.tcl
OpenROAD v2.0-14658-gd48987e36 
[deleted]
Script running: 
```

Use case:

Implement the following `source` idiom to remove dependency on pwd, just like `#include "foo.h"` in C++ that includes the foo.h in the same folder as the compilation unit.

```
source [file dirname [info script]]/util.tcl
```

### Suggested Solution

Reduce surprises, make `info script` return same as for `tclsh`

oharboe:
Looks like the problem is that `Tcl_Eval` instead of `Tcl_EvalFile` is used.

maliberty:
where?

oharboe:
> where?

https://github.com/The-OpenROAD-Project/OpenSTA/blob/1c7f022cd0a02ce71d047aa3dbb64e924b6efbd5/app/StaMain.cc#L95

maliberty:
So not an openroad issue but an sta one

oharboe:
> So not an openroad issue but an sta one

I wouldn't say so. It is the choise of OpenROAD to use "source ..." instead of Tcl_EvalFile(). This choice is encoded into the OpenROAD GUI code:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/0e7dcb1f9100f6667a7d76a39f7d2f0a43749b17/src/Main.cc#L395

https://github.com/The-OpenROAD-Project/OpenROAD/blob/0e7dcb1f9100f6667a7d76a39f7d2f0a43749b17/src/Main.cc#L441

https://github.com/The-OpenROAD-Project/OpenROAD/blob/0e7dcb1f9100f6667a7d76a39f7d2f0a43749b17/src/Main.cc#L401

https://github.com/The-OpenROAD-Project/OpenROAD/blob/0e7dcb1f9100f6667a7d76a39f7d2f0a43749b17/src/Main.cc#L433

oharboe:
If #5427 and #5428 are merged, then the remaining failing case for `info script` is when the GUI is enabled, which I think is a somewhat marginal use-case(`info script` is most critical in batch processing) that can be fixed eventually by someone who has a need for it.

```
$ cat blah.tcl
puts "Script running: [info script]"
[...]

PR #5427, opened 2024-07-23
tcl: fixes 'info string' for normal use case

oharboe:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5415 for normal, most important, use case where a file is executed from a batch file without a GUI.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -333,6 +333,18 @@ static int tclReadlineInit(Tcl_Interp* interp)
 }
 #endif
 
+int tclEvalFile(const char* filename, Tcl_Interp* interp)
+{
+  int code = Tcl_EvalFile(interp, filename);
+  if (code != TCL_OK) {
+    const char* result = Tcl_GetStringResult(interp);
+    if (result[0] != '\0') {
+      printf("%s\n", result);
+    }
+  }
+  return code;
+}
+
 // Tcl init executed inside Tcl_Main.
 static int tclAppInit(int& argc,
                       char* argv[],
@@ -430,7 +442,7 @@ static int tclAppInit(int& argc,
         char* cmd_file = argv[1];
         if (cmd_file) {
           if (!gui_enabled) {
-            int result = sourceTclFile(cmd_file, false, false, interp);
+            int result = tclEvalFile(cmd_file, interp);
             if (exit_after_cmd_file) {
               int exit_code = (result == TCL_OK) ? EXIT_SUCCESS : EXIT_FAILURE;
               exit(exit_code);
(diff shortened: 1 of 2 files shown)


src/rsz/test/hi_fanout.tcl
```diff
@@ -185,7 +185,7 @@ proc write_clone_test_def1 { filename clone_gate fanout
     puts $stream "UNITS DISTANCE MICRONS $dbu ;"
     # Write out the die area 
     write_diearea $stream $fanout $load_spacing
-    puts "$filename $clone_gate $fanout $drvr_inst $drvr_cell $drvr_clk $drvr_out "
+    puts "./[string trimleft [string map [list [pwd]/ ""] [file normalize  $filename]] "/"] $clone_gate $fanout $drvr_inst $drvr_cell $drvr_clk $drvr_out "
```
maliberty, line 188: These two puts seem like debug code as everything else writes to $stream.  I would just remove them.

src/Main.cc
```diff
@@ -333,6 +333,18 @@ static int tclReadlineInit(Tcl_Interp* interp)
 }
 #endif
 
+int tclEvalFile(const char* filename, Tcl_Interp* interp)
+{
+  int code = Tcl_EvalFile(interp, filename);
+  if (code != TCL_OK) {
+    const char* result = Tcl_GetStringResult(interp);
+    if (result[0] != '\0') {
+      printf("%s\n", result);
```
maliberty, line 342: Won't tcl print the result itself?

oharboe: I was trying not to change this aspect of the existing behavior. I will check

src/rsz/test/hi_fanout.tcl
oharboe: Separate concern, separate PR... https://github.com/The-OpenROAD-Project/OpenROAD/pull/5435

src/Main.cc
oharboe: Ah. There more to unravel here...

There's substantial amount of useful error context that isn't printed when using sta::sourceTclFile compared to  tclsh, like a stack trace.

If I disable the monkeypatched "source" implementation in sta and modify tclEvalFile() to:

```C++
int tclEvalFile(const char* filename, Tcl_Interp* interp)
{
  int code = Tcl_EvalFile(interp, filename);
  if (code != TCL_OK) {
    const char* errorMessage = Tcl_GetStringResult(interp);
    const char* errorStack = Tcl_GetVar(interp, "errorInfo", TCL_GLOBAL_ONLY);

    printf("Error: %s\n", errorMessage);
    if (errorStack != NULL) {
        printf("Stack Trace:\n%s\n", errorStack);
    }
  }
  return code;
}
```

I get:

```
$ tclsh blah.tcl 
blah.tcl
invalid command name "asdf"
    while executing
"asdf"
    (file "blah.tcl" line 4)
$ openroad -exit blah.tcl 
OpenROAD v2.0-14709-g55cdc9f21 
blah.tcl
Error: 
Stack Trace:
invalid command name "asdf"
    while executing
"::unknown {*}$args"
    (procedure "sta_unknown" line 28)
    invoked from within
"asdf"
    (file "blah.tcl" line 4)
$ cat blah.tcl 
puts [info script]


asdf
```

oharboe:

PR #5428, opened 2024-07-23
tcl: read in config file without verbose and echo

oharboe:
Use Tcl_EvalFile() to run config files.

This also reduces surprises with e.g. 'info script' not working:

https://github.com/The-OpenROAD-Project/OpenROAD/issues/5415

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -333,6 +333,18 @@ static int tclReadlineInit(Tcl_Interp* interp)
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


gadfort:
I actually like that the scripts echo the commands so it's easier to see what is going on. I don't think this is the right change.

oharboe:
> I actually like that the scripts echo the commands so it's easier to see what is going on. I don't think this is the right change.

Any thoughts on how to fix `info script` and minimize surprises?

oharboe:

PR #5429, opened 2024-07-23
tap: Add -ignore_row_orientation

gatecat:
This is required for correct tapcell orientation regardless of how the rows are laid out on gf130; because as the top/bottom edge cell contains no power pins and instead forms part of a ring round the design, the same cell is used with different mirroring at the top and bottom edges, and the row orientation doesn't matter.

This option therefore ignores the row orientation; always placing the nwin cells with orientation R0 at the bottom edge and the nwout cells with orientation MX at the top edge.


gadfort:
Doesn't this cause check_placement to generate an error?

gatecat:
I haven't seen any failures there, from a quick look at the `check_placement` code, it checks that cells are exactly placed in a row but not that the orientation matches the row orientation.

gadfort:
Could this be accomplished with the place_endcap command instead? There you can specify the top and bottom instances directly

gatecat:
As far as I can tell, place_endcap uses the same code paths as tapcell, so the PR would be needed in any case, just the flag added to a different command. I could add it to place_endcap too, it wasn't really clear which command is preferably used.

maliberty:
> I haven't seen any failures there, from a quick look at the `check_placement` code, it checks that cells are exactly placed in a row but not that the orientation matches the row orientation.

This sounds like a bug/oversight more than a feature you should rely on.

gadfort:
@gatecat check_placement should honor the symmetry of the cells, so it seems problematic to exclude certain cells. If the orientation doesn't matter why not just insert the cell to match the row?

maliberty:
If these cells can truly have any orientation then the LEF should give that information in the SYMMETRY statement.  Do you see that in your PDK?  Any enhancement to check_placement would need to honor SYMMETRY.

I do find this all a bit strange.  We have no other PDK with this behavior including gf180 & gf12.  Are you certain this is intended?

gadfort:
@gatecat with the `place_endcaps` command you can control which cells go in which row so that should provide you the access needed to specify a given cell type for the top and bottom rows.  The only thing missing would be the orientation issue, which I think would be a problem later in the flow if we allowed incorrect orientations to be added to the design.

gatecat:
Closing this PR after discussion with @maliberty - tomorrow I will work on an alternative solution of limiting the floorplanning to an even (or odd in different cases), which also solves the problem without worrying about incorrect row orientations.

PR #5430, opened 2024-07-23
Gui tcl eval file fix

oharboe:


src/Main.cc
```diff
@@ -412,7 +423,7 @@ static int tclAppInit(int& argc,
       init_path += init_filename;
       if (is_regular_file(init_path.c_str())) {
         if (!gui_enabled) {
-          sourceTclFile(init_path.c_str(), true, true, interp);
+          tclEvalFile(init_path.c_str(), interp);
```
gadfort, line 426: sourceTclFile provides some error checking etc and feedback. It feels odd to me that if I run source I get some kind of feedback but if I run openroad with a tcl script I dont

oharboe: you get the same errors and feedback as you would get if you ruh "tslsh foo.tcl"

src/gui/src/gui.cpp
```diff
@@ -1338,10 +1338,10 @@ int startGui(int& argc,
       &app, &QApplication::aboutToQuit, main_window, &MainWindow::saveSettings);
 
   // execute commands to restore state of gui
-  std::string restore_commands;
-  for (const auto& cmd : gui->getRestoreStateCommands()) {
-    restore_commands += cmd + "\n";
-  }
+  std::vector<std::string> restore_commands;
```
gadfort, line 1341: I don't see any reason why the restore commands need any access to script info and therefore there should not be any changes in this. There is no CI checks for this, so this needs to be tested manually.

oharboe: You're right that there are no tests at all for `openroad -gui`, seems like that should be remedied.

We already peer into the restore commands(to look for exit), so storing them as separate commands leaves it to the backend(during userinterface) to inspect the commands and give them(source & exit) any special treatment.

src/Main.cc
gadfort: But not those that match `source`. It feels like the correct solution would be to wait for opensta to implement this so it's consistent throughout openroad

oharboe: I think OpenSTA is a red herring here. This is about how OpenROAD chooses to use Tcl, it has nothing to do with OpenSTA.

Here is the OpenSTA function, what would you have it do differently?

https://github.com/The-OpenROAD-Project/OpenSTA/blob/1c7f022cd0a02ce71d047aa3dbb64e924b6efbd5/app/StaMain.cc#L84C1-L100C1

gadfort: Actually this is the function https://github.com/parallaxsw/OpenSTA/blob/6110a6b96563676228bd51f95a21ae0a2da2aa92/tcl/Sdc.tcl#L68

src/gui/src/gui.cpp
oharboe: Aha! That was surprising. Any idea why "source" has been overriden instead of keeping an OpenSTA specific name?

src/Main.cc
oharboe: Aha! That was surprising. Any idea why "source" has been overriden instead of keeping an OpenSTA specific name?

gadfort: I suspect to provide better error messaging than what is available in tcl. Because of this, i think the right solution is to have the opensta issue resolved, which should solve this globally for us

oharboe: This is the tcl behavior of "source":

```
$ cat blah.tcl 
puts "Script running: [info script]"
$ tclsh
% source blah.tcl
Script running: blah.tcl
$ tclsh 
% source ~/OpenROAD-flow-scripts/blah.tcl
Script running: ~/OpenROAD-flow-scripts/blah.tcl
$ tclsh 
% source ~/OpenROAD-flow-scripts/blah.tcl
Script running: ~/OpenROAD-flow-scripts/blah.tcl
% source OpenROAD-flow-scripts/blah.tcl
Script running: OpenROAD-flow-scripts/blah.tcl
```

From inside `~/OpenROAD-flow-scripts`:

```
$ tclsh
% source ../OpenROAD-flow-scripts/blah.tcl
Script running: ../OpenROAD-flow-scripts/blah.tcl
```

oharboe: > I suspect to provide better error messaging than what is available in tcl. Because of this, i think the right solution is to have the opensta issue resolved, which should solve this globally for us

That's still possible without replacing the existing `source` function. I wonder why the default tcl behavior was overriden instead of having a separate sta_source command.

oharboe:
It does not look like error handling in the OpenSTA source implementation is always better than tclsh. tclsh has stack traces. IMO, simpler errors are easier to read with the OpenSTA syntax, but harder errors are harder to track down.

```
$ openroad -exit foo.tcl 
OpenROAD v2.0-14708-g0e7dcb1f9 
Script running: 
Error: blah.tcl, 3 invalid command name "asdf"
$ tclsh foo.tcl 
Script running: blah.tcl
invalid command name "asdf"
    while executing
"asdf"
    (file "blah.tcl" line 3)
    invoked from within
"source blah.tcl"
    (file "foo.tcl" line 1)
$ cat foo.tcl
source blah.tcl

$ cat blah.tcl 
puts "Script running: [info script]"

asdf
```

gadfort:
@oharboe I was wrong I think it is the echo and verbose flags that get added plus the std redirection that is added by opensta

oharboe:
@gadfort Thanks for all the information :-)

I think at this point, it is useful to just gather some more info in these pull requests so leaving open for review in that regard.

Not sure where this is all going to lead after the information has come all in though...

oharboe:

PR #5432, opened 2024-07-23
grt: remove fake pin creation

eder-matheus:
Secure-ci is running. A PR in ORFS will be created to update the necessary metrics.


src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -409,6 +409,10 @@ void FastRouteCore::addAdjustment(int x1,
     if (!isReduce) {
       const int increase = reducedCap - cap;
       h_edges_[y1][x1].cap += increase;
+      int new_red = static_cast<int>(h_edges_[y1][x1].red) - increase;
```
maliberty, line 412: If reducedCap was a uint16_t you could avoid this casting.

src/grt/src/fastroute/src/FastRoute.cpp
```diff
@@ -409,6 +409,10 @@ void FastRouteCore::addAdjustment(int x1,
...
+      int new_red_3D = static_cast<int>(h_edges_3D_[k][y1][x1].red) - increase;
+      h_edges_[y1][x1].red = new_red >= 0 ? new_red : 0;
```
maliberty, line 414: fwiw I would write this as `h_edges_[y1][x1].red = std::max(new_red, 0);`

maliberty (approved):
I'll approve as the comments are stylistic.

PR #5456, opened 2024-07-26
ifp: Allow constraining row count to be odd or even

gatecat:
This is an alternative solution to #5429 - for gf130 we need an even row count.

I haven't considered the hybrid row case, as that's not needed for us and it seems like supporting it there would introduce a various edge cases that would need to be carefully considered.


src/ifp/README.md
```diff
@@ -72,6 +73,7 @@ initialize_floorplan
 | `-site` | Site name. |
 | `-die_area` | Die area coordinates in microns (lower left x/y and upper right x/y coordinates). |
 | `-core_area` | Core area coordinates in microns (lower left x/y and upper right x/y coordinates). |
+| `-row_parity` | Snap to either an odd (`ODD`) or even (`EVEN`) number of rows. |
```
maliberty, line 76: Indicate the default (presumably NONE).

src/ifp/src/InitFloorplan.i
```diff
@@ -97,11 +110,12 @@ init_floorplan_core(int die_lx,
 		    int core_ux,
 		    int core_uy,
 		    odb::dbSite* site,
-		    const std::vector<odb::dbSite*>& additional_sites)
+		    const std::vector<odb::dbSite*>& additional_sites,
+        ifp::RowParity row_parity)
```
maliberty, line 114: whitespace alignment is off

src/ifp/src/InitFloorplan.tcl
```diff
@@ -39,12 +39,13 @@ sta::define_cmd_args "initialize_floorplan" {[-utilization util]\
 					       [-die_area {lx ly ux uy}]\
 					       [-core_area {lx ly ux uy}]\
 					       [-additional_sites site_names]\
-					       [-site site_name]}
+					       [-site site_name]\
+                 [-row_parity NONE|ODD|EVEN]}
```
maliberty, line 43: alignment

src/ifp/src/InitFloorplan.cc
```diff
@@ -416,7 +426,22 @@ void InitFloorplan::makeUniformRows(odb::dbSite* base_site,
...
+
+    switch (row_parity) {
+      case RowParity::NONE:
+        break;
+      case RowParity::EVEN:
+        rows_y = (rows_y / 2) * 2;
+        break;
+      case RowParity::ODD:
+        if (rows_y > 0) {
+          rows_y = ((rows_y - 1) / 2) * 2 + 1;
```
maliberty, line 439: Fwiw this works but I find `rows_y = rows_y % 2 ? rows_y - 1 : rows_y` more obvious

maliberty (changes requested):

src/ifp/src/InitFloorplan.i
gatecat: Fixed - these files are using a mix of tabs and spaces which I guess should eventually be fixed, but I won't do that in this PR to keep the diff consistent.

src/ifp/src/InitFloorplan.cc
gatecat: Thanks, I agree, I've changed it.

maliberty (approved):

PR #5457, opened 2024-07-26
pdn: Add -stop_at_endcaps to limit stripes to endcap power pins

gatecat:
This follows the alternative approach described in https://github.com/The-OpenROAD-Project/OpenROAD/pull/5349#issuecomment-2223108708, with a new option to stop at the extent of the power pins of endcaps.

This prevents the power stripes extending further into the contents of the endcaps, which on gf130 causes a short on a ring contained within them.


gadfort:
@gatecat is there any reason this shouldn't be the default behavior? ie. no need for the new flag and just stop at the ends of the pins?

gatecat:
I think that could also be okay, it did seem like something that could subtly change behaviour in ways that cause problems, although I think in all other known supported PDKs endcaps have power pins extending to the end, so there'd be no change there.

However, if we had it by default, we perhaps want a better solution to repair stages down the line re-extending the straps and spoiling the fix than just [setting the shapes as locked](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5457/files#diff-917e05e86be3efcbdbc3448422a303209735c80c0b1fca4efad0fac508ed7256R581-R584).

src/pdn/src/straps.cpp
```diff
@@ -413,6 +414,70 @@ FollowPins::FollowPins(Grid* grid, odb::dbTechLayer* layer, int width)
...
+          extent = std::max(extent, rect.xMax());
+        } else {
+          extent = std::min(extent, rect.xMin());
+        }
+      }
+    }
+  }
+
+  return extent;
+}
```
gadfort, line 479: it seems like what you really want is for each "row" to have a new BBox that is bounded by the pin on that layer (or by the row itself if no pins are available)
You may need to check that these pins actually overlap with the follow pin you are aiming for to avoid issues with m2 followpins on powerswitches.

src/pdn/src/straps.cpp
```diff
@@ -413,6 +414,70 @@ FollowPins::FollowPins(Grid* grid, odb::dbTechLayer* layer, int width)
...
+        }
+        odb::Rect rect = box->getBox();
+        transform.apply(rect);
+        if (right) {
+          extent = std::max(extent, rect.xMax());
+        } else {
+          extent = std::min(extent, rect.xMin());
+        }
+      }
+    }
```
gadfort, line 475: what would happen if the where pins on the other side of the vertical pin? this seems to ignore that case all together

gadfort (changes requested):
I think cut rows will be an issue with the current implementation.
Part of me feels like this is right approach for all followpins, so I'm still on the fence about the new flag (since this a default would eliminate the need).
As a new default here are the pitfalls I can see:
1. follow pins on M2 which don't actually follow anything, but the m1 strap, that could be solved by just using the row width if no m2 pins are found and intersect the strap of interest.
2. there are probably some edge cases where the slight difference in strap extent would be an issue, but shouldn't be a huge problem I think.
3. I would expect all the CI tests to have some slight changes on M1, but really nothing else.

src/pdn/src/straps.cpp
gatecat: Do you have a good example of the m2 followpins case that I could use as a reference?

src/pdn/src/straps.cpp
```diff
@@ -413,6 +414,70 @@ FollowPins::FollowPins(Grid* grid, odb::dbTechLayer* layer, int width)
...
+    if (found_row == y_to_row.end()) {
+      continue;
+    }
+    odb::dbRow* row = found_row->second;
+    if (box->xMin() == row->getBBox().xMin()) {
+      result[row].left = inst;
+    }
+    if (box->xMax() == row->getBBox().xMax()) {
+      result[row].right = inst;
+    }
```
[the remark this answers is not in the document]
gatecat: To check I've understood the cut row issue correctly and can implement fixes, also for the above comment, when a row is cut by a macro, there will be two rows either side of it on the same y-location but with different x bounding boxes?

Importantly, what I'd like to confirm is there's never a case where you have a single dbRow with a "hole" inside of it?

src/pdn/src/straps.cpp
gatecat: I'm not sure I quite understand the problem here?

src/pdn/src/straps.cpp
gatecat: For now, I've added a check to make sure the y-position overlaps.

src/pdn/src/straps.cpp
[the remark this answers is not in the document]
gatecat: I've fixed cut rows by indexing on lower-left and lower-right points instead of just on y-coordinates. This seems to work on a local test.

 I haven't committed the test yet, because although it works well enough to check that the PDN stops at the right point, the endcap/corner insertion around macros isn't really playing well with the ring design of endcap cells that's causing problems here, so it would be quite a strange looking test floorplan with short circuits elsewhere.

src/pdn/src/straps.cpp
```diff
@@ -413,6 +414,70 @@ FollowPins::FollowPins(Grid* grid, odb::dbTechLayer* layer, int width)
...
+  std::unordered_map<int, odb::dbRow*> y_to_row;
+  for (auto row : rows) {
+    y_to_row[row->getBBox().yMin()] = row;
+  }
+
+  for (auto inst : getGrid()->getBlock()->getInsts()) {
+    if (!inst->isFixed()
+        || (!inst->isEndCap()
+            && inst->getMaster()->getType()
+                   != odb::dbMasterType::CORE_WELLTAP)) {
```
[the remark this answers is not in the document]
gatecat: Fixed

maliberty:
This seems to have stalled.  What remains to resolve here?

gatecat:
This dropped off my radar, as we did have an albeit somewhat ugly but working workaround in odbpy. I think the remaining issues where whether or not this should be default (and if it was, finding a better solution to the issues in the repair stage undoing the work here than the current one of marking the stripes as locked); and a few other of the review comments.

maliberty:
reopen when ready to proceed

PR #5461, opened 2024-07-26
rmp: Adds new logic cut library to remapper

QuantamHD:
This new logic remapper is step 2 of 3 to begin to do restructuring in the loop of global placement. This new logic library is extensible and can implement diferent stratigies.

This version of the LogicExtractor takes a list of endpoints and will walk backwards until it reaches a cell that ABC cannot handle.


src/rmp/src/logic_extractor.h
```diff
@@ -0,0 +1,63 @@
...
+#include "abc_library_factory.h"
+#include "db_sta/dbSta.hh"
+#include "logic_cut.h"
+#include "sta/Graph.hh"
+#include "sta/NetworkClass.hh"
+#include "sta/SearchPred.hh"
+
+namespace rmp {
+
+class SearchPredNonReg2AbcSupport : public sta::SearchPredNonReg2
```
maliberty, line 21: Please add a class comment as I can't guess what this is from the name.

QuantamHD: Done.

maliberty (approved):

PR #5462, opened 2024-07-26
Changing last sprintf in definReader  to use fmt:format

fgaray:


maliberty (approved):

src/odb/src/defin/definReader.cpp
```diff
@@ -1562,13 +1562,11 @@ int definReader::unitsCallback(defrCallbackType_e, double d, defiUserData data)
...
   if (d > reader->_tech->getDbUnitsPerMicron()) {
-    char buf[256];
-    sprintf(buf,
-            "The DEF UNITS DISTANCE MICRONS convert factor (%d) is "
-            "greater than the database units per micron (%d) value.",
-            (int) d,
-            reader->_tech->getDbUnitsPerMicron());
-    UNSUPPORTED(buf);
+    UNSUPPORTED(
+        fmt::format("The DEF UNITS DISTANCE MICRONS convert factor ({}) is "
```
maliberty, line 1566: Does this work?  I thought it would take a char* and fmt::format returns a string.  Am I missing something?

fgaray: Yeah, I changed in the prev PR #5152 definReader::error from const char* to std::string_view, so now it supports either const char* or std::string.

PR #5467, opened 2024-07-29
delta_debug: reduce review noise by getting some simple formatting changes in

oharboe:
Broke out some changes from https://github.com/The-OpenROAD-Project/OpenROAD/pull/5223 top make review easier.

@maliberty We need to pick some Python code formatting standard and add that to the repository, otherwise "format source code" is going to give different results depending on locally configured Python IDE...

### DIFF
--- a/etc/.flake8
+++ b/etc/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+extend-ignore =
+    # Formatting, yapf handles
+    E501, E251, E128, E131, E125, E129
--- a/etc/.style.yapf
+++ b/etc/.style.yapf
@@ -1,0 +1,2 @@
+[style]
+based_on_style = pep8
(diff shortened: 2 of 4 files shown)


oharboe:
@maliberty Picked some formatting rules as a starting point. Fairly forgiving, can be tightened. Only one "flake8 ." error now(a bit tricky to fix because it has to do with mocking and imports, so beyond this PR).

oharboe:
@habibayassin FYI

maliberty:
I would pick black over yapf.  flake8 is fine.

oharboe:
> I would pick black over yapf. flake8 is fine.

Sounds good to me, I havent used it.  

Could you institute it?

oharboe:
@maliberty Instituting black makes this PR moot

PR #5468, opened 2024-07-29
Tcl readline init

oharboe:


oharboe:
@hovind FYI

maliberty (approved):

maliberty:
Was this a bazel problem?

oharboe:
> Was this a bazel problem?

It was certainly an nit with bogus warnings with Bazel as our use case, yes.

PR #5470, opened 2024-07-29
ant: refactor logs and write report

eder-matheus:
* To create the report for the nets, ant called checkRatioViolations twice. This was potentially causing issues regarding stability, and it wasn't efficient.
* Also, the report was not consistent when multithreading was used.
* The counters for the total number of violation needed reduction when using multithreading.

I created a map to store the report per net, sorting it by the net IDs. The report is written at the end of the ant flow, creating consistent reports independently of threading.

The downside is that the nets with violations will have all the ratio information in their report.


src/ant/src/AntennaChecker.cc
```diff
@@ -539,12 +550,7 @@ bool AntennaChecker::checkPAR(odb::dbTechLayer* tech_layer,
           info.PAR,
           PAR_ratio,
           violation ? "(VIOLATED)" : "");
-      if (verbose) {
-        logger_->report("{}", par_report);
-      }
-      if (report_file.is_open() && (violation || verbose)) {
-        report_file << par_report << "\n";
-      }
+      net_to_report_[db_net].report += par_report + "\n";
```
maliberty, line 553: As it required for MT that db_net be present it would be better to use `net_to_report_.at(db_net)` to enforce it.  Likewise elsewhere.

maliberty (approved):

PR #5473, opened 2024-07-30
python: ran black .

oharboe:
If someone has pull requests in flight, they can run "black ." and create a commit, then merge with master, to minimize conflicts.


QuantamHD:
@oharboe You should add a github action to ensure that the files remain formatted in this PR as well.

https://black.readthedocs.io/en/stable/integrations/github_actions.html

maliberty (approved):

oharboe:
@maliberty The CI change is simple(trivial, right), but I prefer to leave it to the OpenROAD project: when, exactly how, etc.

oharboe:
@maliberty Added CI integration per suggestion from @QuantamHD, so your "automerge" was disabled...

PR #5479, opened 2024-07-30
nix: fix and update shell.nix to 24.05

widlarizer:
Tested to create a functional openroad executable, though on nixOS it doesn't work with `-gui` as-is. I assume it didn't before either, though

### DIFF
--- a/etc/shell.nix
+++ b/etc/shell.nix
@@ -5,8 +5,7 @@
 #   3. `cd build`
 #   4. `cmake $CMAKE_FLAGS ..`
 {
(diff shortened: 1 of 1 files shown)


donn (approved):

maliberty (approved):

widlarizer:
That command works fine. I'm on a rather up to date nixOS running sway as WM. What I get with openroad built with this shell.nix is `This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.`

Issue #4859, opened 2024-03-26
Routing congestion heat map with values over 500 at extreme top and right with ispd18_test7


Felipe-Savi:
### Describe the bug

After loading ispd18_test7 from ispd 2018 benchmark, the heat map shows red lines at the top and right of the circuit, with values over 500:

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/100985480/0fa6879d-cfdd-4639-9c90-7c69c393ea10)

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/100985480/8f49cad3-05c9-466c-b3ab-839eb6073edc)

### Expected Behavior

The congestion at the borders should be 0

### Environment

```shell
Git commit: 2a8ed3debbac47b99675f9b09acaaf6c6cc8408a
kernel: Linux 6.5.0-26-generic
os: Ubuntu 22.04.3 LTS (Jammy Jellyfish)
cmake version 3.24.2
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-12729-g2a8ed3deb
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python: /usr/bin/python3.10 (found version "3.10.12") found components: Interpreter 
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
[...]

PR #5492, opened 2024-08-01
Grt fix phantom congestion on border (Issue 4859)

bnmfw:
Fixes [#4859](https://github.com/The-OpenROAD-Project/OpenROAD/issues/4859). Following there is an explanation of problems found and their solutions.

### Bad Blockage Calculation
This was the first problem to be found and what was causing overflow in congestion (congestion above 500). This was due to borders having more capacity than the rest of the circuit, but not having this fact taken into consideration when calculating blockage. For this fix **Fastrouter.h** was altered to include getters for the capacity at last column and row of a circuit. These are used in **GlobalRouter::updateDbCongestionFromGuides()** in which the capacity considered in a border is the correct one.
### Bad Border Capacity Initialization
In layers of vertical routing the upper border of cells did not have correct initialization, the same is true for the right border in horizontal routing. To fix this **FastRouteCore::initEdges** had two for loops altered to not ignore the last column and row anymore depending on the layer. 
These rows were ignored for a reason, while v_edges_3D_ has values for this border, v_edges does not (the same is true for h_edges_) therefore an if was included to not alter v_edges_ in this border cases. For the same reason **FastRouteCore::addAdjustment** was altered in a similar manner.
### Bad Corner Adjusment
Finally **GlobalRouter::computeGridAdjustments** was altered to addjust the capacity of the top right corner of cells.


maliberty:
Please resolve the clang-tidy feedback.

maliberty:
Nice description!

eder-matheus (approved):

PR #5493, opened 2024-08-01
mpl2: fix SoftMacro leaks and beginning of automated memory management 

AcKoucher:
Change memory management from manual to automated for the following objects:
- Metrics (Owned by a map within the tree)
- HardMacros (Owned by a map within the tree - this can be further improved so that the clustering engine doesn't deal with objects from the other mpl2 stages).
- SoftMacros (Owned by their cluster)

SoftMacros were neither being correctly allocated nor destroyed.

Also removing some useless variables.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -260,9 +260,7 @@ void HierRTLMP::run()
   correctAllMacrosOrientation();
 
   commitMacroPlacementToDb();
-
   writeMacroPlacement(macro_placement_file_);
-  clear();
 }
 
 void HierRTLMP::init()
@@ -3379,8 +3377,6 @@ void HierRTLMP::placeMacros(Cluster* cluster)
              cluster->getName());
 
   std::vector<HardMacro*> hard_macros = cluster->getHardMacros();
-  num_hard_macros_cluster_ += hard_macros.size();
-
   std::vector<HardMacro> sa_macros;
   std::vector<Cluster*> macro_clusters;  // needed to calculate connections
   std::map<int, int> cluster_to_macro;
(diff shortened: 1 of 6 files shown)


src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -1232,11 +1234,12 @@ void HierRTLMP::runHierarchicalMacroPlacement(Cluster* parent)
     }
     // for other clusters
     soft_macro_id_map[cluster->getName()] = macros.size();
-    SoftMacro* soft_macro = new SoftMacro(cluster);
+    std::unique_ptr<SoftMacro> soft_macro
+        = std::make_unique<SoftMacro>(cluster);
```
maliberty, line 1238: this is a good place for auto to avoid repeating the same info twice
```
auto soft_macro = std::make_unique<SoftMacro>(cluster);
```
Similar elsewhere

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -4068,16 +4072,6 @@ void HierRTLMP::writeMacroPlacement(const std::string& file_name)
 
 void HierRTLMP::clear()
 {
-  for (auto& [module, metrics] : tree_.maps.module_to_metrics) {
-    delete metrics;
-  }
-  tree_.maps.module_to_metrics.clear();
```
maliberty, line 4074: Do these no longer need to be cleared?

maliberty (changes requested):

src/mpl2/src/hier_rtlmp.cpp
AcKoucher: They do. I'm fixing it.

AcKoucher: When HierRTLMP::run() finishes the tree will be deleted and so these maps which should destroy the unique_ptrs so I don't think these clear() are needed. When the Cluster memory management rework (the subsequent PR which will be opened) is merged. This clear() function will no longer exist and the eraseDrawing  will be moved to HierRTLMP::run();

maliberty: That sounds reasonable

AcKoucher:
Ready for another review.

maliberty (approved):

PR #5495, opened 2024-08-02
pad: handle IO sites with rotations

gadfort:
Fixes:
- IO row creation when the sites are correctly aligned with the rows, but causing the width of a row to appear as the height in dbSite.


maliberty (approved):

src/pad/test/make_io_sites_different_sites.ok
```diff
@@ -0,0 +1,9 @@
+[INFO ODB-0227] LEF file: Nangate45/Nangate45.lef, created 22 layers, 27 vias, 135 library cells
+[INFO ODB-0227] LEF file: Nangate45_io/dummy_pads.lef, created 24 library cells
+[INFO ODB-0128] Design: soc_bsg_black_parrot
+[INFO ODB-0130]     Created 135 pins.
+[INFO ODB-0131]     Created 267 components and 2277 component-terminals.
+[INFO ODB-0133]     Created 350 nets and 390 connections.
+Differences found at line 7.
```
rovinski, line 7: Differences

gadfort: Thanks for pointing that out. Should be fixed in #5498

Issue #5494, opened 2024-08-01
GRT: Check if the input route segments are valid


eder-matheus:
### Description

In PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/5482, the ability to write and read the global route segments was added in GRT. The global route segments are similar to the route guides, but the main differences are:

In the route guides, each line represents a rectangle in a specific routing layer: `(lower_left_x, lower_left_y, upper_right_x, upper_right_y, routing_layer)`.

In the route segments, each line describes a vertical or horizontal line, or a via, going from one point to another in a given routing layer: `(x1, y1, routing_layer1, x2, y2, routing_layer2)`.

Some examples:
1) `18000 154800 met4 18000 205200 met4`: this segment represents a vertical line in the routing layer met4.
2) `18000 154800 met3 133200 154800 met3`: this segment represents a horizontal line in the routing layer met3.
3) `18000 205200 met3 18000 205200 met4`: this segment represents a via going from met3 to met4.
4) `133200 154800 met3 133200 154800 met2`: this segment represents a via going from met3 to met2.
Notice that the via segments can go from lower layers to upper layers (like in example 3) or from upper layers to lower layers (like in example 4).

When reading the global route segments from a file, GRT should check if they're valid:
1. Check if the segments are connected (i.e., there are no "floating" segments)
2. Check if the pins of the net are all covered by the global route segments.
   (Note: a reference on how GRT understands the pin positions of the nets can be found here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/2197#discussioncomment-10203352)

### Additional Context

Check Discussion https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/2197 for more information and background on this task.

PR #5496, opened 2024-08-02
grt: add -show_segments option to draw_route_guides command

eder-matheus:
Route segments:
image

Route guides:
image


eder-matheus:
@bnmfw This might help you when working in https://github.com/The-OpenROAD-Project/OpenROAD/issues/5494

PR #5497, opened 2024-08-02
gpl: mbff updates from recent paper

therealshreyas:
Updates to the MBFF implementation from the recent paper:
(1) Remove constant knn value and use the Silhouette metric to find a good knn value.
(2) Fix AR-based distance bug.
(3) Updated default alpha, beta, and max_split_size default values to match experiments using ASAP7.
(4) Revert ILP to match implementation from old reference.
(5) Add the new paper reference.


src/stt/include/stt/SteinerTreeBuilder.h
```diff
@@ -100,6 +100,12 @@ class SteinerTreeBuilder
                        const std::vector<int>& y,
                        const std::vector<int>& s,
                        int acc);
+
+  // run FOARS
+  Tree makeFoarsTree(const std::vector<int>& x_pin,
```
maliberty, line 105: Did you intend to include this in this PR?  It sems unrelated.

src/gpl/src/mbff.cpp
```diff
@@ -137,6 +137,11 @@ float MBFF::GetDist(const Point& a, const Point& b)
   return (abs(a.x - b.x) + abs(a.y - b.y));
 }
 
+float MBFF::GetDistAR(const Point& a, const Point& b, const float AR)
```
maliberty, line 140: What is AR?

src/stt/include/stt/SteinerTreeBuilder.h
therealshreyas: not sure why this happened again ... removed foars content in latest PR

src/gpl/src/mbff.cpp
therealshreyas: aspect ratio-based distance for initial mbff generation

maliberty (approved):

PR #5502, opened 2024-08-02
mpl2: finalization of memory management rework

AcKoucher:
Clusters now own their children in such a way that destroying the root will destroy all other clusters up to the leaves.


AcKoucher:
@maliberty I'm running some tests with valgrind and will also start a secure-CI. With that in mind, this is ready for review.

AcKoucher:
I force pushed just to fix the last commit message.

src/mpl2/src/object.cpp
```diff
@@ -479,32 +479,50 @@ void Cluster::setParent(Cluster* parent)
...
 
-void Cluster::addChild(Cluster* child)
+void Cluster::addChild(std::unique_ptr<Cluster> child)
 {
-  children_.push_back(child);
+  children_.push_back(std::move(child));
 }
 
-void Cluster::removeChild(const Cluster* child)
+std::unique_ptr<Cluster> Cluster::releaseChild(const Cluster* candidate)
```
maliberty, line 487: Is there a reason you need to use child_index and not a regular iterator from find_if?  It feels like this could be simpler.

AcKoucher: It could be simpler, I'm fixing it.

AcKoucher:
Done

maliberty (approved):

PR #5514, opened 2024-08-05
gui: use a buffer to store report text to avoid slowing down OR when logging in Scripting

### LINKED ISSUE #5340: script output display is very slow when the gui is on
### Describe the bug

I've noticed a slowness in gui which is very annoying. It seems this comes from the console input & output.

Using the following procedure:

    proc measure_time {} {
        set TIME_start [clock clicks -milliseconds]
        for {set i 0} {$i < 100} {incr i} {
            puts "Iteration $i"
        }
        set TIME_taken [expr [clock clicks -milliseconds] - $TIME_start]
        puts "Duration $TIME_taken ms"
    }

Without gui the script takes less that 1 ms.

With gui (-gui option or gui::show), it takes 40 ms.
    - source the script takes less that 1 ms whenever it is called.

This is another specific condition in gui that show the slowness with the procedure taking 8 ms.

Basically after link_design is called, all subsequent commands before you get the prompt again are horribly slow.

So basically a script test.tcl ending by run by openroad -gui test.tcl :

    measure_time => 40 ms
    link_design test
    measure_time => 8 s

Once the prompt is available calling again measure_time will takes 40 ms

Now commenting these 3 lines and start again, executing these 3 commands one by one => 40 ms

Same with 3 commands called in a row will give 40 ms followed by 8 s.

### Expected Behavior

Should be few.

### Environment

```shell
OpenROAD 74da93a32ce0cb78d8d93ca473121b8cdd52ff8d from openroad-2024-06-11 package
```

### To Reproduce

openroad -gui test.tcl

test.tcl:

    proc measure_time {} {
        set TIME_start [clock clicks -milliseconds]
        for {set i 0} {$i < 100} {incr i} {
            puts "Iteration $i"
        }
        set TIME_taken [expr [clock clicks -milliseconds] - $TIME_start]
        puts "Duration $TIME_taken ms"
    }
    ....
    measure_time
    link_design test
    measure_time

I'm trying to make a test case. It does not seem easy as using a simpler design does not show the issue.

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
Resolve #5340

### DIFF
--- a/src/gui/src/scriptWidget.cpp
+++ b/src/gui/src/scriptWidget.cpp
@@ -113,7 +113,7 @@ ScriptWidget::ScriptWidget(QWidget* parent)
   connect(input_,
           &TclCmdInputWidget::commandFinishedExecuting,
           this,
-          &ScriptWidget::flushReportBufferToOuput);
+          &ScriptWidget::flushReportBufferToOutput);
   connect(output_,
           &QPlainTextEdit::textChanged,
           this,
@@ -123,7 +123,7 @@ ScriptWidget::ScriptWidget(QWidget* parent)
   connect(report_timer_.get(),
           &QTimer::timeout,
           this,
-          &ScriptWidget::flushReportBufferToOuput);
+          &ScriptWidget::flushReportBufferToOutput);
 
   connect(this,
           &ScriptWidget::addToOutput,
@@ -137,9 +137,11 @@ ScriptWidget::ScriptWidget(QWidget* parent)
 // When displaying text in Scripting, processing events in order to
 // ensure the text is visible to the user can considerably slow down
 // the logging, so, for reports, we use a buffer.
-void ScriptWidget::flushReportBufferToOuput()
+void ScriptWidget::flushReportBufferToOutput()
 {
-  report_timer_->stop();
+  if (report_buffer_.isEmpty()) {
+    return;
+  }
 
   // this comes from a ->report
   addTextToOutput(report_buffer_, buffer_msg_);
@@ -227,11 +229,6 @@ void ScriptWidget::addResultToOutput(const QString& result, bool is_ok)
 void ScriptWidget::addLogToOutput(const QString& text, const QColor& color)
 {
   addToOutput(text, color);
-}
-
-bool ScriptWidget::reportTimerIsActive()
-{
-  return report_timer_->isActive();
 }
 
 void ScriptWidget::startReportTimer()
@@ -416,7 +413,7 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
         std::string(formatted.data(), formatted.size()));
 
     if (msg.level == spdlog::level::level_enum::off) {
-      if (!widget_->reportTimerIsActive()) {
+      if (!widget_->report_timer_->isActive()) {
         widget_->startReportTimer();
       }
 
--- a/src/gui/src/scriptWidget.cpp
+++ b/src/gui/src/scriptWidget.cpp
@@ -233,7 +233,7 @@ void ScriptWidget::addLogToOutput(const QString& text, const QColor& color)
 
 void ScriptWidget::startReportTimer()
 {
-  report_timer_->start(20 /*ms*/);
+  report_timer_->start(report_display_interval);
 }
 
 void ScriptWidget::addMsgToReportBuffer(const QString& text)
--- a/src/gui/src/scriptWidget.cpp
+++ b/src/gui/src/scriptWidget.cpp
@@ -56,6 +56,7 @@ ScriptWidget::ScriptWidget(QWidget* parent)
       input_(new TclCmdInputWidget(this)),
       pauser_(new QPushButton("Idle", this)),
       pause_timer_(std::make_unique<QTimer>()),
+      report_timer_(std::make_unique<QTimer>()),
       paused_(false),
       logger_(nullptr),
       buffer_outputs_(false),
@@ -115,6 +116,10 @@ ScriptWidget::ScriptWidget(QWidget* parent)
           &ScriptWidget::outputChanged);
   connect(pauser_, &QPushButton::pressed, this, &ScriptWidget::pauserClicked);
   connect(pause_timer_.get(), &QTimer::timeout, this, &ScriptWidget::unpause);
+  connect(report_timer_.get(),
+          &QTimer::timeout,
+          this,
+          &ScriptWidget::flushReportBufferToOuput);
 
   connect(this,
           &ScriptWidget::addToOutput,
@@ -123,6 +128,15 @@ ScriptWidget::ScriptWidget(QWidget* parent)
           Qt::QueuedConnection);
 
   setWidget(container);
+}
+
+void ScriptWidget::flushReportBufferToOuput()
+{
+  report_timer_->stop();
+
+  // this comes from a ->report
+  addTextToOutput(report_buffer_, buffer_msg_);
+  report_buffer_.clear();
 }
 
 ScriptWidget::~ScriptWidget()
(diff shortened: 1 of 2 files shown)


maliberty:
@gadfort this is an attempt to speed up the GUI output window when there is a lot of fast scrolling output.  I believe you've used the timer trick elsewhere and would like your input.

oharboe:
@jeffng-or Perhaps this will speed up loading `make gui_route` on megaboom too?

src/gui/src/scriptWidget.cpp
```diff
@@ -115,6 +116,10 @@ ScriptWidget::ScriptWidget(QWidget* parent)
           &ScriptWidget::outputChanged);
   connect(pauser_, &QPushButton::pressed, this, &ScriptWidget::pauserClicked);
   connect(pause_timer_.get(), &QTimer::timeout, this, &ScriptWidget::unpause);
+  connect(report_timer_.get(),
+          &QTimer::timeout,
+          this,
+          &ScriptWidget::flushReportBufferToOuput);
```
gadfort, line 122: you should also connect this to `commandFinishedExecuting` to avoid a delay in printing the output.

src/gui/src/scriptWidget.cpp
```diff
@@ -208,9 +225,19 @@ void ScriptWidget::addLogToOutput(const QString& text, const QColor& color)
   addToOutput(text, color);
 }
 
-void ScriptWidget::addReportToOutput(const QString& text)
+bool ScriptWidget::reportTimerIsActive()
+{
+  return report_timer_->isActive();
+}
```
gadfort, line 231: this might be a overkill since the single line of code it wraps is fairly self-explanatory. (this is a stylistic thing, so nothing major here)

src/gui/src/scriptWidget.cpp
```diff
@@ -385,8 +412,15 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
         std::string(formatted.data(), formatted.size()));
 
     if (msg.level == spdlog::level::level_enum::off) {
-      // this comes from a ->report
-      widget_->addReportToOutput(formatted_msg);
+      if (!widget_->reportTimerIsActive()) {
+        widget_->startReportTimer();
```
gadfort, line 416: this can only be started from the main thread, you need to guard against this. otherwise you will get a bunch of errors about this timer.

src/gui/src/scriptWidget.cpp
```diff
@@ -125,6 +130,18 @@ ScriptWidget::ScriptWidget(QWidget* parent)
   setWidget(container);
 }
 
+// When displaying text in Scripting, processing events in order to
+// ensure the text is visible to the user can considerably slow down
+// the logging, so, for reports, we use a buffer.
+void ScriptWidget::flushReportBufferToOuput()
+{
+  report_timer_->stop();
```
gadfort, line 138: not sure if you need to do this.

src/gui/src/scriptWidget.cpp
```diff
@@ -125,6 +130,18 @@ ScriptWidget::ScriptWidget(QWidget* parent)
...
+
+  // this comes from a ->report
+  addTextToOutput(report_buffer_, buffer_msg_);
```
gadfort, line 141: you might want to check if the buffer has anything before calling this

src/gui/src/scriptWidget.cpp
```diff
@@ -125,6 +130,18 @@ ScriptWidget::ScriptWidget(QWidget* parent)
...
+  report_buffer_.clear();
```
gadfort, line 142: thread safety?

gadfort (changes requested):
@maliberty I think this is alright, but I have some concerns about the timer and thread access. The timers can only be triggered from main thread, failing to do so will cause Qt to generate a bunch of warnings. Additionally , the buffer can be added to and cleared from any number of threads (DRT tends to trigger this a lot), so the buffer needs a mutex to avoid clobbering the information.

Did anyone profile to see what the slow-ness comes from? If I recall correctly, adding HTML vs. plain text to the screen takes a lot longer. Right now all text is added as HTML, it might be worth looking at that before using a buffering solution.

maliberty:
> Did anyone profile to see what the slow-ness comes from? If I recall correctly, adding HTML vs. plain text to the screen takes a lot longer. Right now all text is added as HTML, it might be worth looking at that before using a buffering solution.

@AcKoucher did profile and the time is in event processing as I recall.  I think it will be hard to tell from a profile if plain vs html is the culprit.  We could try it the other way though I suspect that high volume updates will still be an issue if we process them line by line.

AcKoucher:
@gadfort FYI profiling showed that some part of the runtime is due to the html part. However as Matt pointed out, the main problem is processing the events for every single line of text to refresh the GUI. Here's the flame graph for logging 10k lines of text using a .tcl script

image

Without using the GUI it takes around 100ms

src/gui/src/scriptWidget.cpp
```diff
@@ -125,6 +130,18 @@ ScriptWidget::ScriptWidget(QWidget* parent)
   setWidget(container);
 }
 
+// When displaying text in Scripting, processing events in order to
+// ensure the text is visible to the user can considerably slow down
+// the logging, so, for reports, we use a buffer.
+void ScriptWidget::flushReportBufferToOuput()
```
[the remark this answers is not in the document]
gadfort: rename function, it's misspelled.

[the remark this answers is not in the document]
AcKoucher: I see now. Thanks.

gadfort:
@AcKoucher thanks. I think the approach will be work, but needs to be modified to be thread safe for Qt.

AcKoucher:
@gadfort Ok! I'll address the review.

src/gui/src/scriptWidget.cpp
```diff
@@ -208,9 +225,19 @@ void ScriptWidget::addLogToOutput(const QString& text, const QColor& color)
...
 
-void ScriptWidget::addReportToOutput(const QString& text)
+bool ScriptWidget::reportTimerIsActive()
+{
+  return report_timer_->isActive();
+}
+
+void ScriptWidget::startReportTimer()
+{
+  report_timer_->start(20 /*ms*/);
```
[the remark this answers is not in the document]
AcKoucher: I just at 20Hz, it looks lowering more don't have much more impact

src/gui/src/scriptWidget.cpp
```diff
@@ -208,9 +232,14 @@ void ScriptWidget::addLogToOutput(const QString& text, const QColor& color)
...
-void ScriptWidget::addReportToOutput(const QString& text)
+void ScriptWidget::startReportTimer()
+{
+  report_timer_->start(report_display_interval);
+}
+
+void ScriptWidget::addMsgToReportBuffer(const QString& text)
 {
-  addToOutput(text, buffer_msg_);
+  report_buffer_ += text;
```
gadfort, line 242: this will also need a mutex here so different threads don't try to add at the same time.

src/gui/src/scriptWidget.cpp
```diff
@@ -385,8 +414,17 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
...
 
     if (msg.level == spdlog::level::level_enum::off) {
-      // this comes from a ->report
-      widget_->addReportToOutput(formatted_msg);
+      if (QThread::currentThread() == widget_->thread()) {
+        if (!widget_->report_timer_->isActive()) {
+          widget_->startReportTimer();
+        }
+
+        widget_->addMsgToReportBuffer(formatted_msg);
```
gadfort, line 422: you still need to add the message, you just can't trigger the timer or the event loop. Just move this line up to before the thread check.

gadfort (changes requested):
Looks good, just needs the additional mutex lock and fix the handling of the message when not in the main thread.

AcKoucher:
Fixed. Thanks for the feedback @gadfort !

src/gui/src/scriptWidget.cpp
```diff
@@ -385,8 +415,17 @@ class ScriptWidget::GuiSink : public spdlog::sinks::base_sink<Mutex>
...
+      widget_->addMsgToReportBuffer(formatted_msg);
+
+      if (QThread::currentThread() == widget_->thread()) {
+        if (!widget_->report_timer_->isActive()) {
+          widget_->startReportTimer();
+        }
+
+        // the event loop is blocked so we need to manually process
+        // events so that the timer keeps going
+        QCoreApplication::processEvents(QEventLoop::ExcludeUserInputEvents);
```
maliberty, line 427: Does this mean that the event won't be processed until the main thread get a message?

gadfort: yes, or the next time the event loop is processed which would be when the main thread is done executing something

gadfort (approved):
LGTM

maliberty (approved):

Issue #5002, opened 2024-04-26
report_clock_skew does not find the worst skew of any constrained path


oharboe:
### Description

Run:

1. untar [skew-consistency.tar.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/15127013/skew-consistency.tar.gz). based on `make DESIGN_CONFIG=designs/asap7/mock-array/config.mk cts`
2. run `./run-me-mock-array-asap7-base.sh`

Look at the timing path that `report_clock_skew` reports, I would have expected the "Skew" column to match up with `report_clock_skew`

![image](https://github.com/The-OpenROAD-Project/OpenROAD/assets/2798822/05faf017-9f81-4866-8d5a-37f9df79ddc6)

### Suggested Solution

Skew column should match up with report_clock_skew

maliberty:
@tspyrou I don't know what's going on here.  Please look at it or discuss it with Cherry.  I can't relate

```
Clock clock
 313.20 source latency ces_4_2/clock ^
  96.54 source clock tree delay
-283.39 target latency ces_3_2/clock ^
 -77.04 target clock tree delay
  -7.42 CRPR
--------------
  41.89 setup skew
```

to

```
>>> rc -through ces_4_2/clock -through ces_3_2/*
Startpoint: ces_4_2 (rising edge-triggered flip-flop clocked by clock)
Endpoint: ces_3_2 (rising edge-triggered flip-flop clocked by clock)
Path Group: clock
Path Type: max

Fanout     Cap    Slew   Delay    Time   Description
-----------------------------------------------------------------------------
                          0.00    0.00   clock clock (rise edge)
                          0.00    0.00   clock source latency
     1   13.08    0.00    0.00    0.00 ^ clock (in)
                                         clock (net)
                 33.86   10.68   10.68 ^ wire2/A (BUFx16f_ASAP7_75t_R)
     1   18.19    9.80   21.51   32.19 ^ wire2/Y (BUFx16f_ASAP7_75t_R)
                                         net4218 (net)
                 71.11   22.20   54.40 ^ wire1/A (BUFx16f_ASAP7_75t_R)
     2   29.31   11.48   27.19   81.58 ^ wire1/Y (BUFx16f_ASAP7_75t_R)
                                         net4217 (net)
                155.26   48.77  130.35 ^ clkbuf_0_clock/A (BUFx24_ASAP7_75t_R)
[...]

PR #5515, opened 2024-08-06
timing report: add a more elaborate tooltip

oharboe:
If there was a f1/help button for all menus, then
some of this tooltip would be moved in there.

Meanwhile, best that can be done is a slightly verbose tooltip.

Doesn't look too bad. Tolerable...

image


maliberty:
Where did "subtracting a clock period" come from?

oharboe:
> Where did "subtracting a clock period" come from?

https://github.com/The-OpenROAD-Project/OpenROAD/issues/5002#issuecomment-2088063107

Issue #5306, opened 2024-06-29
cts not honoring fanout limits


gadfort:
### Describe the bug

CTS does not appear to be honoring fanout limits in ihp130.
```
Pin                                   Limit Fanout  Slack
---------------------------------------------------------
clkbuf_leaf_11_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_12_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_16_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_29_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_41_clk/X                      8     14     -6 (VIOLATED)
```

### Expected Behavior

No DRV violations from CTS

### Environment

```shell
Jenkins CI runner
```

### To Reproduce

aes with ihp130 from ORFS (in jenkins CI so should be reprodusable)

### Relevant log output

```shell
Pin                                   Limit Fanout  Slack
---------------------------------------------------------
clkbuf_leaf_11_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_12_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_16_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_29_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_41_clk/X                      8     14     -6 (VIOLATED)
clkbuf_leaf_9_clk/X                       8     14     -6 (VIOLATED)
clkbuf_leaf_13_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_17_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_20_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_28_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_40_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_48_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_53_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_55_clk/X                      8     13     -5 (VIOLATED)
clkbuf_leaf_5_clk/X                       8     13     -5 (VIOLATED)
clkbuf_leaf_19_clk/X                      8     12     -4 (VIOLATED)
clkbuf_leaf_22_clk/X                      8     12     -4 (VIOLATED)
[...]

PR #5516, opened 2024-08-06
Cts obey max fanout

arthurjolo:
As reported on [#5306](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5306), CTS is not honoring the fanout limit set. Even with the sinks cluster and max sink per leaf buffer also being limited.

This was happening because after CTS builds the HTree it does some optimizations on the sinks distribution on the leaf buffers, shifting some sinks from one buffer to another violating the fanout limit. This PR adds a flag to set a fanout limit on CTS, making sure that it will respect this limit.

This flag must be added on the designs that have a fanout limit set so CTS can undestand this limit.


arthurjolo:
This PR will only created the flag `-max_fanout,` to fix the issue [#5306](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5306) the flag must be added the ORFS designs that will be done on a separate PR.

src/cts/src/TritonCTS.tcl
```diff
@@ -80,6 +80,7 @@ sta::define_cmd_args "clock_tree_synthesis" {[-wire_unit unit]
                                              [-sink_clustering_size] \
                                              [-sink_clustering_max_diameter] \
                                              [-sink_clustering_enable] \
+                                             [-max_fanout] \
```
precisionmoon, line 83: We want to honor SDC set_max_fanout constraint by default.   Let's add -ignore_max_fanout If user wants to ignore max fanout limit,

src/cts/src/HTreeBuilder.cpp
```diff
@@ -1209,6 +1209,14 @@ void HTreeBuilder::run()
   // clang-format on
 }
 
+bool HTreeBuilder::isNumberOfSinksTooSmall(unsigned numSinksPerSubRegion) const
```
precisionmoon, line 1212: Shouldn't this API return if either of the conditions is met?   < numMaxLeafSinks_  || < options_->getMaxFanout()?

src/cts/src/TritonCTS.tcl
QuantamHD: That seems ultra confusing. If a user sets this flag and the SDC flag it seems like we could offer a warning, but principle of least surprise seems to be we should honor the more specific constraint in the CTS call.

maliberty: We discussed today not needing the flag.  Please wait for the next update.

arthurjolo:
I used the key option because I couldn't find a way for CTS to check if it should respect a max fanout limit. Looking deeper I found a sdc API that I can get the information needed, I am already working on removing the key.

src/cts/src/TritonCTS.cpp
```diff
@@ -109,14 +109,72 @@ void TritonCTS::addBuilder(TreeBuilder* builder)
...
+    auto io_type = mterm->getIoType();
+    if (io_type == odb::dbIoType::OUTPUT) {
+      buffer_port = network_->dbToSta(mterm);
+      break;
+    }
+  }
+  if (buffer_port == nullptr) {
+    return 0;
+  }
+  openSta_->sdc()->fanoutLimit(
```
precisionmoon, line 134: Just a minor suggestion.  You can also start with a large fanout limit like INF. 
Then you can start taking, sequentially one by one, min (fanout_limit, new_fanout_limit) for port, cell, liberty port, etc.
fanout_limit = INF;
if (port has fanout limit) {
    fanout_limit = MIN (fanout_limit, new fanout limit);
}
If (cell has fanout limt) {
   fanout_limit = MIN(fanout_limit, new fanout limit);
}
If (lib port has fanout limit) {
   fanout_limit = MIN(fanout_limit, new fanout limit);
}

arthurjolo:
The secure-CI is passing for this PR. There is only 1 design that still is having 1 max fanout violation: ihp-sg13g2/riscvi,  this violation is related to the tech characterization step, where CTS is deciding to not use buffers on some levels of the clock tree. Talking with @precisionmoon we believe it is best to merge this PR to fix the max fanout violations related with the sinks and look at the characterization as a separate issue.

arthurjolo:
> The secure-CI is passing for this PR. There is only 1 design that still is having 1 max fanout violation: ihp-sg13g2/riscvi, this violation is related to the tech characterization step, where CTS is deciding to not use buffers on some levels of the clock tree. Talking with @precisionmoon we believe it is best to merge this PR to fix the max fanout violations related with the sinks and look at the characterization as a separate issue.

Using `-distance_between_buffers 60` on ihp-sg13g2/riscvi it was possible to solve this problem.

maliberty:
Please make a companion ORFS PR with the necessary change so they can be merged together.

arthurjolo:
Ok, I am going to create the PR.

src/cts/src/HTreeBuilder.cpp
```diff
@@ -1209,6 +1210,15 @@ void HTreeBuilder::run()
...
+{
+  if (options_->getMaxFanout()) {
+    logger_->report("Ta verificando por fanout");
```
maliberty, line 1216: rm - Look like debug code

src/cts/src/TritonCTS.cpp
```diff
@@ -1493,6 +1554,19 @@ odb::dbITerm* TritonCTS::getFirstInput(odb::dbInst* inst) const
...
+  odb::dbSet<odb::dbITerm> iterms = inst->getITerms();
+  for (odb::dbITerm* iterm : iterms) {
+    if (iterm->isInputSignal()) {
+      return iterm;
+    }
+  }
+
+  return nullptr;
+}
+*/
```
maliberty, line 1568: rm

maliberty (changes requested):

maliberty (approved):

PR #5518, opened 2024-08-06
grt: read_segments conectivity, dimensionality and pin coverage validation

bnmfw:
Fixes [#5494](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5494).
### Conectivity Check
To check if all segments in a net connect the method **GlobalRouter::isConnected** was created. It implements the union-find algorithm to check if all segments in a net connect, the check between two segments is done with the new method **GlobalRouter::segmentsConnect**. The time complexity is O(s^2\*log(s)) where s is the number of segments. If the received segments list is ordered in a way that a segment always connects to the previous the complexity falls to O(s*log(s)).
A failure in the conectivity check raises the error GRT 262.
### Deminsionality Check
During the conectivity check every segments has its dimensions checked to see if its a horizontal line, vertical line or via. This is done by the new method **GlobalRouter::segmentIsLine**.
A dimensionality violation will raise the error GRT 264 or GRT 265.
### Pin Coverage Check
After the connectivity check, for every net all its pins are checked against all its segments to see if the pin is covered by a segment. The time complexity is O(s*p) where s is the number of segments and p the number of pins.
A failure in the pin coverage check raises the error GRT 262


src/grt/src/GlobalRouter.cpp
```diff
@@ -2227,6 +2227,131 @@ void GlobalRouter::readSegments(const char* file_name)
...
+      }
+      if (!pin_is_covered) {
+        logger_->error(GRT,
+                       263,
+                       "Pin {} is not covered by net {}",
+                       pin.getName(),
+                       net->getName());
+      }
+    }
+  }
```
eder-matheus, line 2264: Turn this code into a separate _netIsCovered_ function. The function can receive the grt::Net* and the GRoute& as parameters.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2227,6 +2227,131 @@ void GlobalRouter::readSegments(const char* file_name)
...
+  int dimensionality = (segment.init_x != segment.final_x)
+                       + (segment.init_y != segment.final_y)
+                       + (segment.init_layer != segment.final_layer);
+  return (dimensionality == 1);
+}
+
+// Implements Union-Find algorithm to determine connectivity
+bool GlobalRouter::isConnected(odb::dbNet* net)
+{
+  std::vector<int> parent(routes_[net].size()), rank(routes_[net].size(), 0);
```
eder-matheus, line 2280: Refactor `routes_[net].size()` into a variable and use it instead of calling the size function multiple times.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2227,6 +2227,131 @@ void GlobalRouter::readSegments(const char* file_name)
...
+            && pin.getConnectionLayer()
+                   <= std::max(seg.init_layer, seg.final_layer)) {
+          pin_is_covered = true;
+          break;
+        }
+      }
+      if (!pin_is_covered) {
+        logger_->error(GRT,
+                       263,
+                       "Pin {} is not covered by net {}",
```
eder-matheus, line 2259: Add a period at the end of messages. Here and below.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2227,6 +2227,131 @@ void GlobalRouter::readSegments(const char* file_name)
...
+
+  std::function<int(int)> find = [&](int x) -> int {
+    if (parent[x] != x) {
+      parent[x] = find(parent[x]);
+    }
+    return parent[x];
+  };
+
+  std::function<void(int, int)> uniteGroups = [&](int u, int v) {
+    int rootU = find(u), rootV = find(v);
```
eder-matheus, line 2303: Use the same convention for variable names. Instead of rootU/rootV, use root_u/root_v.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2227,6 +2227,131 @@ void GlobalRouter::readSegments(const char* file_name)
...
+      }
+    }
+  }
+  return (initialized_groups == 1);
+}
+
+bool GlobalRouter::segmentsConnect(const GSegment& segment1,
+                                   const GSegment& segment2)
+{
+  std::function<std::pair<int, int>(int, int)> minMax
```
eder-matheus, line 2342: You can use std::minmax: https://cplusplus.com/reference/algorithm/minmax/

eder-matheus (changes requested):
Please, add unit tests for the checks. You can add at least 2 unit tests:
1. An input segs file that is not connected
2. An input segs file that is not covering one of the pins

You can use the read_segments1 and read_segments2 unit tests as a base.

eder-matheus (approved):

PR #5519, opened 2024-08-06
odb: add polygon writing to lefout

gadfort:
Adds:
- POLYGON support to LEFout for libraries


src/odb/test/write_lef_polygon.lefok
```diff
@@ -0,0 +1,2921 @@
...
+        DIRECTION INOUT ; 
+        USE SIGNAL ; 
+        PORT
+            LAYER metal6 ;
+              POLYGON  14.5 -5.42 5.42 -14.5 -5.42 -14.5 -14.5 -5.42 -14.5 5.42 -5.42 14.5 5.42 14.5 14.5 5.42 ;
+        END
+    END PAD
+    OBS
+      LAYER via5 ;
+        POLYGON  14.5 -5.42 5.42 -14.5 -5.42 -14.5 -14.5 -5.42 -14.5 5.42 -5.42 14.5 5.42 14.5 14.5 5.42 ;
```
maliberty, line 2918: The input LEF has
```
OBS
    LAYER via5 ;
      POLYGON 14.5 5.42 5.42 14.5 -5.42 14.5 -14.5 5.42 -14.5 -5.42 -5.42 -14.5 5.42 -14.5 14.5 -5.42 ;
  END
```
which doesn't match this polygon.

src/odb/test/write_lef_polygon.lefok
```diff
@@ -0,0 +1,2921 @@
...
+MACRO BUMP
+    CLASS COVER BUMP ;
+    ORIGIN 14.5 14.5 ;
+    SIZE 29 BY 29 ;
+    PIN PAD
+        DIRECTION INOUT ; 
+        USE SIGNAL ; 
+        PORT
+            LAYER metal6 ;
+              POLYGON  14.5 -5.42 5.42 -14.5 -5.42 -14.5 -14.5 -5.42 -14.5 5.42 -5.42 14.5 5.42 14.5 14.5 5.42 ;
```
maliberty, line 2913: The input LEF has
```
 POLYGON 14.5 5.42 5.42 14.5 -5.42 14.5 -14.5 5.42 -14.5 -5.42 -5.42 -14.5 5.42 -14.5 14.5 -5.42 ;
```
which doesn't match

src/odb/test/write_lef_polygon.lefok
gadfort: In the decomp come we reorder the points to be "clockwise", I think it gets applied to the polygon, but shouldn't (I don't know if its needed any more)

gadfort: thanks for catching that, I moved the clockwise thing into the decompose code to avoid modifying the polygon points

maliberty:
Similarly
```
  if (polygon[0] == polygon[polygon.size() - 1]) {
    polygon.pop_back();
  }
```
should be moved to decomp.

src/odb/src/db/dbPolygon.cpp
```diff
@@ -251,14 +251,6 @@ Polygon _dbPolygon::checkPolygon(std::vector<Point> polygon)
     return {};
   }
 
-  if (polygon[0] == polygon[polygon.size() - 1]) {
-    polygon.pop_back();
-  }
```
maliberty, line 256: I don't know if the decomposer assumes this has been done.

gadfort: good point. I added it back.

maliberty (approved):

PR #5521, opened 2024-08-07
spef: writing dates in output files is an anachronism

oharboe:
@QuantamHD Thoughts? 

See also CI test https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2223

I saw global routing failing after I set SKIP_REPORT_METRICS=1, though I suspect it can also come from some other change I was not aware I did that changed the initial conditions.

This got me thinking that it would be nice for options that should have no effect on the result to have a CI smoke test...

Documenting non-functional behaviors is pretty much worthless, if there's no CI test, don't expect some desired non-functional behavior to be conserved over time... You become what you test and measure...

### DIFF
--- a/src/rcx/src/extSpef.cpp
+++ b/src/rcx/src/extSpef.cpp
@@ -1648,8 +1648,6 @@ void extSpef::writeHeaderInfo()
                 "%H:%M:%S %A %B %d, %Y",
                 std::localtime(&currentTime));
 
-  fprintf(_outFP, "*DATE \"%s\"\n", buffer);
-
   fprintf(_outFP, "*VENDOR \"The OpenROAD Project\"\n");
   fprintf(_outFP, "*PROGRAM \"OpenROAD\"\n");
   fprintf(_outFP, "*VERSION \"%s\"\n", _version);
--- a/src/rcx/src/extSpef.cpp
+++ b/src/rcx/src/extSpef.cpp
@@ -1641,12 +1641,9 @@ void extSpef::writeHeaderInfo()
 
   std::time_t currentTime = std::time(nullptr);
 
-  // Format the current time as a string
-  char buffer[128];
-  std::strftime(buffer,
-                sizeof(buffer),
-                "%H:%M:%S %A %B %d, %Y",
-                std::localtime(&currentTime));
+  // With repeatable builds being the norm now, date in output files is
+  // an anachronism. Leave an N/A string as this is a required field.
+  fprintf(_outFP, "*DATE \"N/A\"\n");
 
   fprintf(_outFP, "*VENDOR \"The OpenROAD Project\"\n");
   fprintf(_outFP, "*PROGRAM \"OpenROAD\"\n");


gadfort:
@oharboe I like the model we have for write_sdc with a flag to control the date time writing:
https://github.com/parallaxsw/OpenSTA/blob/1a3a26bf3a1c397b763fc1f8abd9c2e2f6514424/sdc/Sdc.tcl#L153

oharboe:
@gadfort I had a quick stab at implementing such option, but it was a bit hairy. I think that fewer options is generally better. To my mind a date in a result file like this is simply an anachronism and I'm fine with simply removing it and adding an option if it proves necessary(which is not going to be any harder than adding it now).

I think the only thing we achieve by adding the option now, is that we are guaranteed to spend that engineering time, even if it proves a waste of time.

Unless we know now that it is necessary?

maliberty:
*DATE is not optional in SPEF so this would produce an invalid file.  SDF does allow an optional date so that flag makes more sense.

oharboe:
> *DATE is not optional in SPEF so this would produce an invalid file. SDF does allow an optional date so that flag makes more sense.

Can we use a dummy fixed date?

maliberty:
> Can we use a dummy fixed date?

That seems likely to lead to confusion.  What problem are we solving?

maliberty:
Note that date is just a quoted string.  We could write "N/A" or the like.

oharboe:
> Note that date is just a quoted string. We could write "N/A" or the like.

Sounds good.

oharboe:
> Note that date is just a quoted string. We could write "N/A" or the like.

Make it easy to check for idempotency of results. Repeatable builds are in the zeitgeist...

gadfort:
@maliberty would it be possible to have a flag to trigger the N/A over just removing the date all together? it's helpful when reviewing/debugging tapeouts to have the dates in the files when possible.

oharboe:
@maliberty Where does it say that DATE is mandatory? ChatGPT says that it is not...

oharboe:
@gadfort Dates for files is a generic problem that has to be solved for all file types and it should be solved out of bands(not in the file format) for output files so as to support idempotency. Idempotency is not a luxury, it is needed to avoid re-running subsequenty actions unecessarily when calculating hashes for dependencies(bazel being only one example, this is the way build systems are moving)

gadfort:
@oharboe I don't want to argue with you on which way is correct. All I'm suggesting is that we keep the default behavior and provide a way to control this behavior so we can both get what we want.

maliberty:
> @maliberty Where does it say that DATE is mandatory? ChatGPT says that it is not...

The standard - chatgpt is wrong

maliberty:
I think you need to add the option as there is already an objection.

oharboe:
> I think you need to add the option as there is already an objection.

Do you have a link to the standard?

QuantamHD:
The default at Google for dealing with dates is to set them at a fixed value usually unix time zero. Unless you include a --stamp flag. I wouldn't be opposed to such a solution here.

maliberty:
Its IEEE 1481-2019 but behind a paywall (I have access through UCSD).  I think it is better to match sta for consistency sake.

PR #5532, opened 2024-08-08
pad: remove terminal edges on success and failure to avoid shorting later routes into shape

gadfort:
Fixes:
- on a failed RDL route the graph edges were left behind opening the possibility to shorting routes because the cleanup did not occur. This ensures the edges are removed in either case to prevent routing shorts in RDL.


maliberty (approved):

maliberty:
conflict to resolve

gadfort:
@maliberty resolved.  Thanks

PR #5536, opened 2024-08-08
gui: use try_lock for flushing buffer

gadfort:
Why:
- I noticed the GUI would periodically lock up on me when the GUI was active and reports were printing.
- When checking in GDB it was this lock that kept coming up as waiting to lock, but not able to.

Change:
- changes this to a `try_to_lock` (non blocking), if it fails to aquire the lock it will just restart a timer to call this later and return without flushing the buffer. This might cause the GUI to print things out of order compared to the cmdline, but avoids the lock up because the flush was unable to get the mutex.

I'm open to suggestions on how to better resolve this issue, but this "lazy"-lock approach was the simplest I could think of.

### DIFF
--- a/src/gui/src/scriptWidget.cpp
+++ b/src/gui/src/scriptWidget.cpp
@@ -139,7 +139,15 @@ ScriptWidget::ScriptWidget(QWidget* parent)
 // the logging, so, for reports, we use a buffer.
 void ScriptWidget::flushReportBufferToOutput()
 {
-  std::lock_guard guard(reporting_);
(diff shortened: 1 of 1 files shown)


maliberty (approved):
Ok though it doesn't feel like an ideal solution.

maliberty:
@AcKoucher any thoughts on this?

gadfort:
@maliberty I agree that this is not perfect, but for now we can continue to use the GUI, but long term the printing order should not be changed between the cmdline and the GUI

AcKoucher:
I can't think of a better alternative now

Issue #5486, opened 2024-07-31
Error when calling detailed_route


titan73:
### Describe the bug

Here is the error meesage:

Error: test.tcl, 25 vector::_M_range_check: __n (which is 4) >= this->size() (which is 0)

### Expected Behavior

Should succeeded

### Environment

```shell
Latest git.
```

### To Reproduce

openroad test.tcl with test case provided.

[pin_access2.zip](https://github.com/user-attachments/files/16438851/pin_access2.zip)

maliberty:
@osamahammad21 this is a very small test case so hopefully this is quick to resolve.

osamahammad21:
@titan73 the verilog code doesn't make any sense.
```
module test(a, b, z);
  input a;
  wire a;
  input b;
  wire b;
  output z;
  wire z;
  STD_CELL _00_ (
    .A(flops[1]),
    .B(in_ports[1]),
    .Z(out_ports[1])
  );
endmodule
```
 There is no nets to route that's why DRT breaks and GRT does not do anything.

maliberty:
@osamahammad21 it would be nice to give a clear error in that case and stop early.

titan73:
Ah indeed. Sorry. I missed some renaming when making the test case. This should have been:

module test(a, b, z);
  input a;
  wire a;
  input b;
  wire b;
  output z;
  wire z;
  STD_CELL _00_ (
    .A(a),
    .B(b),
    .Z(z)
  );
endmodule

It's curious read_verilog didn't complain at all.

osamahammad21:
> @osamahammad21 it would be nice to give a clear error in that case and stop early.

the issue originates with DRT's global router and I don't think it is a priority at this point. maybe GRT should have an error or warning if there are no nets to route!

osamahammad21:
I am closing this issue since there is no reply for 4 days. If needed, please reopen.

titan73:
> I am closing this issue since there is no reply for 4 days. If needed, please reopen.

@osamahammad21 I was in holidays. There was no question for me but for @maliberty (see just below)

> @titan73 you might as well see it ending with one offgrid violation that cannot be solved. this is because metal1's half width is not divisible by the manufacturing grid. @maliberty not sure if there should be an error or a warning for this case?

How did you see that offgrid violations? I didn't see anything and I'd like to take a look.

> > @osamahammad21 it would be nice to give a clear error in that case and stop early.
> 
[...]

PR #5540, opened 2024-08-09
Issue an error if design does not have any routable net (with at least 2 terms) for grt and drt

titan73:
For https://github.com/The-OpenROAD-Project/OpenROAD/issues/5486.

### DIFF
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -942,6 +942,20 @@ int TritonRoute::main()
     }
   }
   initDesign();
+  bool has_routable_nets = false;
+  for (auto net : db_->getChip()->getBlock()->getNets()) {
+    if (net->getITerms().size() + net->getBTerms().size() > 1) {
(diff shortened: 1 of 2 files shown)


maliberty (approved):

src/drt/src/TritonRoute.cpp
```diff
@@ -942,6 +942,19 @@ int TritonRoute::main()
...
   initDesign();
+  bool has_routable_nets = false;
+  for (auto net : db_->getChip()->getBlock()->getNets()) {
+    if (net->getITerms().size() + net->getBTerms().size() > 1) {
+      has_routable_nets = true;
+      break;
+    }
+  }
+  if (!has_routable_nets) {
+    logger_->error(DRT,
```
maliberty, line 953: I think it might be better to give a warning and return.  It isn't strictly illegal.  Same in grt

titan73: ok I'll do the change. Thanks.

PR #5547, opened 2024-08-11
drt: coverity fixed (1)

eder-matheus:
There are many other coverity issues to be fixed, but I decided to start with these ones to avoid a huge PR.

### DIFF
--- a/src/drt/src/io/io.cpp
+++ b/src/drt/src/io/io.cpp
@@ -831,7 +831,7 @@ void io::Parser::updateNetRouting(frNet* netIn, odb::dbNet* net)
             styleWidth = layer->getWidth();
           }
         }
-        width = (width) ? width : tech_->name2layer_[layerName]->getWidth();
+        width = tech_->name2layer_[layerName]->getWidth();
         auto defaultBeginExt = width / 2;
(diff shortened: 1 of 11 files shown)


maliberty:
Please fix the errors before I review

eder-matheus:
> Please fix the errors before I review

Fixed.

src/drt/src/io/io.cpp
```diff
@@ -831,7 +831,7 @@ void io::Parser::updateNetRouting(frNet* netIn, odb::dbNet* net)
             styleWidth = layer->getWidth();
           }
         }
-        width = (width) ? width : tech_->name2layer_[layerName]->getWidth();
+        width = tech_->name2layer_[layerName]->getWidth();
```
maliberty, line 834: Please explain this change.

eder-matheus: The `width` var is assigned with zero here: https://github.com/eder-matheus/OpenROAD/blob/drt_coverity/src/drt/src/io/io.cpp#L711. This is the beginning of the loop, and its value is not changed anywhere else before reaching line 834. The check is redundant because `width` will always be zero at this point.

maliberty (approved):

PR #5550, opened 2024-08-11
rdl: fix accounting for turns to avoid DRCs

gadfort:
Fixes:
- the RDL router was not correctly removing routing edges from the grid it uses, causing 90 degree turns and 45 degree routes to have possible min-spacing violations. This is resolved by using check when committing the route to remove the unusable edges.
- The first row and column on the routing grid was not valid, this is corrected by accounting for the pitch near the edge.

Changes:
- adds GUI access to routes and routing obstruction view for better debugging.
- Adjusts RDL test defs to account for the changes + fix the two 45 degree routing tests as they has min-spacing violations.


maliberty:
If you do rip-up and re-route do you have to reinstate the edges?

maliberty (approved):

gadfort:
@maliberty yes, at the moment I have the hooks needed to track the edges removed (so they could be added back, but it's not in use).

PR #5561, opened 2024-08-12
drt: coverity fixes (2)

eder-matheus:
This PR fixes most of the remaining coverity issues. The issues related to FlexGR were left untouched as they are low priority. A couple of false positives regarding serialization were left too.

### DIFF
--- a/src/drt/src/io/io_guide.cpp
+++ b/src/drt/src/io/io_guide.cpp
@@ -676,11 +676,14 @@ bool io::Parser::genGuides_gCell2APInstTermMap(
   frMTerm* trueTerm = instTerm->getTerm();
   std::string name;
   frInst* inst = instTerm->getInst();
-  dbTransform shiftXform = inst->getTransform();
-  shiftXform.setOrient(dbOrientType(dbOrientType::R0));
+  dbTransform shiftXform;
 
   int pinIdx = 0;
   int pinAccessIdx = (inst) ? inst->getPinAccessIdx() : -1;
+  if (inst != nullptr) {
+    dbTransform shiftXform = inst->getTransform();
+    shiftXform.setOrient(dbOrientType(dbOrientType::R0));
+  }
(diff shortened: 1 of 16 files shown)


src/drt/src/io/io_guide.cpp
```diff
@@ -676,11 +676,14 @@ bool io::Parser::genGuides_gCell2APInstTermMap(
   frMTerm* trueTerm = instTerm->getTerm();
   std::string name;
   frInst* inst = instTerm->getInst();
-  dbTransform shiftXform = inst->getTransform();
-  shiftXform.setOrient(dbOrientType(dbOrientType::R0));
+  dbTransform shiftXform;
 
   int pinIdx = 0;
   int pinAccessIdx = (inst) ? inst->getPinAccessIdx() : -1;
+  if (inst != nullptr) {
```
maliberty, line 683: This doesn't hurt but if instTerm->getInst() returns null something is wrong with odb.

maliberty (approved):

src/drt/src/io/io_guide.cpp
eder-matheus: Perhaps we could have a critical here? We don't have a lot of criticals in the code, but it seems like a good case.

maliberty: it would make sense to have it in dbITerm::getInst if Coverity will then use that inter-procedurally.

eder-matheus: Like this: https://github.com/The-OpenROAD-Project/OpenROAD/pull/5561/files#diff-1bfa0367ed9744f752dcda88d6c1557285e18e9351bba5fcf6d4d67d4572129fR244-R247?

maliberty: yes

maliberty (approved):

Issue #5254, opened 2024-06-17
GRT: avoid rerouting nets that were changed, but still covered by the guides


eder-matheus:
During incremental GRT, nets can change their pins (by swapping the master of instances, moving the instances, connecting a new pin, etc.). Don't reroute the whole net if the existing route guides already cover the modified/new pin.

The branch https://github.com/eder-matheus/OpenROAD/tree/grt_incremental already implements the detection of when a net doesn't need to be rerouted. However, the previous routing might have useless segments that were not connecting to any pin. See the examples below:

<img width="667" alt="Pasted Graphic" src="https://github.com/user-attachments/assets/70a3ad3c-f954-43ba-a6c0-2b417bb29ecc">
<img width="498" alt="Pasted Graphic 1" src="https://github.com/user-attachments/assets/72751a0a-eb4a-4e99-8813-eea747aa5c5a">

For the attached test case, nets `_031_` and `_036_` had their pins modified yet still covered by the guides, but some segments are connecting to nothing.

The next step is to prune the extra guides that don't connect with anything.

Test case: https://drive.google.com/file/d/1VPn_PvqXPEPM8T5i2GqL-R76b5GaabzP/view?usp=drive_link

PR #5562, opened 2024-08-12
Grt delete segments

bnmfw:
Fixes [#5254](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5254).
### Segment Cover Pin Refactoring
Function **GlobalRouter::segmentCoversPin()** to check if a given segment covers a given pin. **GlobalRouter::netIsCovered()** was refactored to include this function
### Net Graph Support and Delete Route
For this task it was pertinent to have segments of a net be represented as a graph assigning which connects to which. For this purpose a parent_segment_ vector was created. This models the net as a tree rooted on the net driver, the vector gives the parent of each segment in this tree. The root is its own parent. The indexing follows the same indexing found on the routes_[dbnet] vector. 
A function **Net::getSegmentGraph()** was added to return the graph formatted as a vector of vectors where graph[node] = {all nodes connected to it}. This function is not currently in use but might be useful in other tasks.
A **Net::deleteRoute()** function was added that receives an index and a GRoute. It erases the element in the corresponding index from the GRoute and updated the parent_segment_ vector.
### Useless Segment Deletion
This is the main part of the task, all functions before are auxiliary to this part.
In time complexity s is for number of segments, p for number of pins and d for deleted segments.
The function **GlobalRouter::shrinkNetRoute()** was implemented to delete this segments. Following are the steps taken to do so:
#### Pin Coverage and root detection O(p*s)
Every pair of segment and pin is iterated through, given a complexity of O(p*s).
Every segment is assigned a boolean value indicating whether it contains a pin.
The segment containing the driver is assigned as the root.
#### Build Net Graph O(s²)
A function **GlobalRouter::buildNetGraph()** as implemented to build a graph of the segments. It iterates through every pair of segments checking if they connect. If so and edge in the graph is created. The time complexity is O(s²).
#### BFS to create a tree O(s)
With the graph a breadth first search is done to create a tree rooted in the driver representing the net.
A vector containing the number of children a node has in the tree was also created.
At the end of the process the Net parent_segments_ is set to the tree created.
#### Useless Routes Pruning O(s)
A set containing segments that will be deleted is created
A queue with all the leafs is created and iterated though.
if a leaf (no children) is marked as not covering a pin it is added to the segments for deletion. The number of children of its parent is decreased and if this turns it into a leaf the parent is added to the queue.
At the end of the process every leaf of the tree will cover a pin
#### Deleting Useless Segments O(s*d)
Every element in the set is deleted from the Net. Since it is a vector deletion it costs O(n) for every element deleted, resulting in a O(s*d) time complexity.


src/grt/src/Net.h
```diff
@@ -52,11 +54,18 @@ class Net
...
   int getNumPins() const { return pins_.size(); }
   float getSlack() const { return slack_; }
   void setSlack(float slack) { slack_ = slack; }
   void setHasWires(bool in) { has_wires_ = in; }
+  void setSegmentParent(std::vector<uint16_t> segment_parent)
+  {
+    segment_parent_ = std::move(segment_parent);
+  }
+  std::vector<uint16_t> getSegmentParent() const { return segment_parent_; }
+  std::vector<std::vector<uint16_t>> getSegmentGraph();
```
eder-matheus, line 68: Maybe `buildSegmentsGraph`? It makes it clearer that the graph is related to all segments of the net and that this function is creating the graph, not just returning the content of a class member.

src/grt/include/grt/GlobalRouter.h
```diff
@@ -190,6 +190,8 @@ class GlobalRouter : public ant::GlobalRouteSource
   void readSegments(const char* file_name);
   bool netIsCovered(odb::dbNet* db_net, std::string& pins_not_covered);
   bool segmentIsLine(const GSegment& segment);
+  bool segmentCoversPin(const GSegment& segment, const Pin& pin);
+  std::vector<std::vector<int>> buildNetGraph(odb::dbNet* net);
```
eder-matheus, line 194: Create a typedef for the `std::vector<std::vector<int>>`.

src/grt/src/GlobalRouter.cpp
```diff
@@ -620,13 +622,84 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
         logger_->error(
             GRT, 267, "Net {} has disconnected segments.", net->getName());
       }
+    } else if (net->getPins().size() >= 2) {
```
eder-matheus, line 625: I think you'll need to check the function `newPinOnGrid` in the first if statement too. The way it is now, your code is probably never being called.

eder-matheus: You'll also need to check it again in this if. So your clause will be something like `else if (!newPinOnGrid() && net->getPins().size() >= 2)`

src/grt/src/GlobalRouter.cpp
```diff
@@ -2269,6 +2346,39 @@ bool GlobalRouter::segmentIsLine(const GSegment& segment)
...
+          && pin.getOnGridPosition().getY() <= max_y
+          // Pin and segment share a layer
+          && pin.getConnectionLayer() >= min_layer
+          && pin.getConnectionLayer() <= max_layer);
+}
+
+// Builds the Net Graph in O(N²)
+std::vector<std::vector<int>> GlobalRouter::buildNetGraph(odb::dbNet* net)
+{
+  int total_segments = routes_[net].size();
```
eder-matheus, line 2368: const int

src/grt/src/GlobalRouter.cpp
```diff
@@ -2269,6 +2346,39 @@ bool GlobalRouter::segmentIsLine(const GSegment& segment)
...
+  std::vector<std::vector<int>> graph(total_segments, std::vector<int>());
+  for (int i = 0; i < total_segments; i++) {
+    for (int j = i - 1; j >= 0; j--) {
+      if (!segmentsConnect(routes_[net][i], routes_[net][j])) {
```
eder-matheus, line 2372: You can create a const& for routes_[net], avoiding multiple access to the routes_ map.

src/grt/src/GlobalRouter.cpp
```diff
@@ -946,6 +1019,20 @@ bool GlobalRouter::pinPositionsChanged(Net* net,
   return is_diferent;
 }
 
+bool GlobalRouter::newPinOnGrid(Net* net, std::set<RoutePt>& last_pos)
+{
+  for (const Pin& pin : net->getPins()) {
+    if (last_pos.find(RoutePt(pin.getOnGridPosition().getX(),
```
eder-matheus, line 1025: Move the creation of the RoutePt outside the for loop.

src/grt/src/GlobalRouter.cpp
```diff
@@ -620,13 +622,84 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
...
 
+void GlobalRouter::shrinkNetRoute(odb::dbNet* db_net)
+{
+  Net* net = db_net_map_[db_net];
+  GRoute segments = routes_[db_net];
+  int total_segments = segments.size();
+  std::vector<Pin> pins = net->getPins();
+
+  std::string dump;
+  netIsCovered(db_net, dump);
```
eder-matheus, line 642: You should check if this is true and error out if it's not.

src/grt/src/GlobalRouter.cpp
```diff
@@ -620,13 +622,84 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
...
     net->setDirtyNet(false);
   }
   dirty_nets_.clear();
 }
 
+void GlobalRouter::shrinkNetRoute(odb::dbNet* db_net)
+{
+  Net* net = db_net_map_[db_net];
+  GRoute segments = routes_[db_net];
+  int total_segments = segments.size();
```
eder-matheus, line 638: const

src/grt/src/GlobalRouter.cpp
```diff
@@ -620,13 +622,84 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
...
     net->setMergedNet(false);
     net->setDirtyNet(false);
   }
   dirty_nets_.clear();
 }
 
+void GlobalRouter::shrinkNetRoute(odb::dbNet* db_net)
+{
+  Net* net = db_net_map_[db_net];
+  GRoute segments = routes_[db_net];
```
eder-matheus, line 637: const&

src/grt/src/GlobalRouter.cpp
```diff
@@ -620,13 +622,84 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
...
+  int total_segments = segments.size();
+  std::vector<Pin> pins = net->getPins();
```
eder-matheus, line 639: const&

src/grt/src/GlobalRouter.cpp
```diff
@@ -620,13 +622,84 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
...
+
+  std::string dump;
+  netIsCovered(db_net, dump);
+  int root = -1;
+  std::vector<bool> coversPin(total_segments, false);
```
eder-matheus, line 644: Use snake case for variable names.

src/grt/src/GlobalRouter.cpp
```diff
@@ -620,13 +622,84 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
...
+        }
+      }
+    }
+  }
+
+  std::vector<std::vector<int>> graph = buildNetGraph(db_net);
+
+  // Runs a BFS trough the graph
+  std::vector<uint16_t> parent(total_segments, UINT16_MAX),
+      total_children(total_segments, 0);
```
eder-matheus, line 661: Separate the vector declarations into two individual declarations. I find it a bit confusing the way it is right now.

src/grt/src/Net.h
```diff
@@ -73,6 +82,7 @@ class Net
   std::vector<Pin> pins_;
   float slack_;
   bool has_wires_;
+  std::vector<uint16_t> segment_parent_;
```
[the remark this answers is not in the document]
maliberty: What does clang-tidy say?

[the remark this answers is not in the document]
eder-matheus: clang-tidy recommends to use uint16_t instead of unsigned short. But Bernardo's question was a misunderstanding of my request.

src/grt/src/Net.h
```diff
@@ -52,11 +54,21 @@ class Net
   const char* getConstName() const;
   odb::dbSigType getSignalType() const;
   void addPin(Pin& pin);
+  void deleteSegment(int seg_id, GRoute& routes);
```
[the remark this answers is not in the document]
bnmfw: Interesting

src/grt/src/Net.h
```diff
@@ -52,11 +54,21 @@ class Net
...
   std::vector<Pin>& getPins() { return pins_; }
   int getNumPins() const { return pins_.size(); }
   float getSlack() const { return slack_; }
   void setSlack(float slack) { slack_ = slack; }
   void setHasWires(bool in) { has_wires_ = in; }
+  void setSegmentParent(std::vector<SegmentIndex> segment_parent)
+  {
+    parent_segment_indices_ = std::move(segment_parent);
+  }
+  std::vector<SegmentIndex> getSegmentParent() const
```
eder-matheus, line 67: The function names should match the class member name.

eder-matheus (approved):
Please address the last comment before merging. Other than that, the code looks good!

PR #5569, opened 2024-08-13
pad: fix coverity identified overflow

gadfort:


src/pad/src/RDLRouter.cpp
```diff
@@ -327,7 +327,7 @@ std::vector<RDLRouter::Edge> RDLRouter::insertTerminalVertex(
       if (new_dist < dist) {
         dist = new_dist;
       } else {
-        return {grid[i - 1], static_cast<int>(i - 1)};
+        return {grid[i - 1], static_cast<int>(i) - 1};
```
maliberty, line 330: It seems simpler to just do
```
 for (int i = 0; i < grid.size(); i++) {
```
and drop the cast.

maliberty (approved):

PR #5575, opened 2024-08-14
odb: add boost to Oct and Polygon

gadfort:
Adds:
- boost types for `polygon_type` for `odb::Oct` and `odb::Polygon`
- this enablement is suitable for rtrees and intersection checks, but not for cases where boost modifies the underlying data. I tied to get this working, but wasn't able to. Open to suggestions for how.
- Adds `getRect` to `Oct` to mirror `getRect` in `Polygon`
- Adds `bloat` to `Oct` and `Polygon` to provide similar feature to `bloat` in `Rect`

Future PR adds this enablement to the RDL router.


src/odb/include/odb/geom.h
```diff
@@ -176,6 +177,9 @@ class Oct
   int yMax() const;
   std::vector<Point> getPoints() const;
 
+  Oct bloat(int margin) const;
+  Rect getRect() const;
```
maliberty, line 181: I think `getEnclosingRect` is clearer as the rect is not equivalent to the oct.

src/odb/include/odb/geom.h
```diff
@@ -339,6 +343,10 @@ class Polygon
   int dx() const { return getRect().dx(); }
   int dy() const { return getRect().dy(); }
 
+  // returns a corrected Polygon with a closed form and counter-clockwise points
+  Polygon correct() const;
```
maliberty, line 347: Is there a reason to not make this automatic in the ctor / setPoints ?  I'm not sure why we would want an 'incorrect' one.

src/odb/src/db/geom.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  boost::geometry::assign_points(polygon_in, polygon.getPoints());
+
+  using dist = boost::geometry::strategy::buffer::distance_symmetric<int>;
+  boost::geometry::strategy::buffer::side_straight side_strategy;
+  boost::geometry::strategy::buffer::join_round join_strategy;
+  boost::geometry::strategy::buffer::end_flat end_strategy;
+  boost::geometry::strategy::buffer::point_circle point_strategy;
+
+  boost::geometry::buffer(polygon_in,
```
maliberty, line 66: I'm curious why not use `T1 operator+(const T1&, coordinate_type bloating)` from https://www.boost.org/doc/libs/1_79_0/libs/polygon/doc/gtl_polygon_set_concept.htm

maliberty (changes requested):

src/odb/include/odb/geom.h
gadfort: I agree, but that would mean that dbPolygon would need to track the "original" points vs the correct points to allow the lefout to reflect back the points from the lefin (or alternatively, accept that the LEFout will always be a correct polygon). Any thoughts?

src/odb/src/db/geom.cpp
gadfort: I wasn't able to get that working for non-45, 90 degree shapes, but happy to give it ago again.
`::buffer` provides a generic way to bloat / shrink a shape.

src/odb/include/odb/geom.h
gadfort: Probably makes sense to correct name in Polygon as well

src/odb/include/odb/geom.h
maliberty: So it is more of "make boost friendly" than "correct".  In that case perhaps it should be private as it seems to be specific to the bloat case.

src/odb/src/db/geom.cpp
maliberty: I think it would be nicer to get it working.  What problem did you get stuck on?

src/odb/include/odb/geom.h
gadfort: yes and no, to properly do `boost::geometry::intersect` the polygon has to be correct. Without it you can't rely on it. Personally, I think the `Polygon` should correct the points to forms a complete polygon otherwise all code that uses a polygon needs to first get the "correct" variant of it (so it could be private and handled in `setPoints`).
I use this function for that very reason in the RDL code, so it can't be private as long as the polygon may be incomplete.

src/odb/src/db/geom.cpp
gadfort: I think what I was stuck on was an earlier attempt to do this, it seems to work just fine now.

maliberty (approved):

PR #5582, opened 2024-08-14
grt: support remove buffers

eder-matheus:
When a buffer is removed, the nets previously connected are merged into a single net. Instead of rerouting this single net from scratch, GRT now merges the route segments of the nets, preserving the topology of the nets in the new one.

Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5625


eder-matheus:
@maliberty I have a secure ci running for it, and I intend to perform a few more metrics evaluations before merging, but I decided to create the PR to receive some feedback on what can be changed.

src/rsz/test/repair_setup6.ok
```diff
@@ -17,8 +17,9 @@ delta HPWL                    0 %
...
 [INFO RSZ-0099] Repairing 4 out of 4 (100.00%) violating endpoints...
-[INFO RSZ-0059] Removed 5 buffers.
-[INFO RSZ-0041] Resized 1 instances.
+[INFO RSZ-0059] Removed 3 buffers.
+[INFO RSZ-0040] Inserted 3 buffers.
+[INFO RSZ-0041] Resized 2 instances.
 Repair timing output passed/skipped equivalence test
 worst slack 0.01
-worst slack 0.01
+worst slack -0.10
```
maliberty, line 25: Any idea why this degraded slightly?

src/rsz/src/Resizer.cc
```diff
@@ -375,6 +375,13 @@ bool Resizer::removeBuffer(Instance* buffer,
                "remove {}",
                db_network_->name(buffer));
     buffer_removed = true;
+
+    odb::dbNet* db_survivor = db_network_->staToDb(survivor);
+    odb::dbNet* db_removed = db_network_->staToDb(removed);
+    if (removed) {
+      db_survivor->mergeNet(db_removed);
```
maliberty, line 382: The db should do the merge and the sta merged related code here should become unnecessary (aside from parasitics).

src/odb/src/db/dbNet.cpp
```diff
@@ -3201,6 +3201,15 @@ dbNet* dbNet::getValidNet(dbBlock* block_, uint dbid_)
   return (dbNet*) block->_net_tbl->getPtr(dbid_);
 }
 
+void dbNet::mergeNet(dbNet* in_net)
```
maliberty, line 3204: This needs to actually do the merge and not just issue a callback.  Its is quite confusing.

src/odb/include/odb/dbBlockCallBackObj.h
```diff
@@ -85,6 +85,7 @@ class dbBlockCallBackObj
   // dbNet Start
   virtual void inDbNetCreate(dbNet*) {}
   virtual void inDbNetDestroy(dbNet*) {}
+  virtual void inDbNetMerge(dbNet*, dbNet*) {}
```
maliberty, line 88: The should be inDbNetPreMerge or inDbNetPostMerge depending on where the callback happens relative to the edits (or both if needed).

src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -117,7 +117,7 @@ class FastRouteCore
                 int max_layer,
                 float slack,
                 std::vector<int>* edge_cost_per_layer);
-  void removeNet(odb::dbNet* db_net);
+  void removeNet(odb::dbNet* db_net, bool release_resources);
```
maliberty, line 120: It might be better to have a mergeNet here rather than remove but keep the resources which feels semantically odd.

src/grt/src/GlobalRouter.cpp
```diff
@@ -3776,6 +3802,139 @@ std::vector<odb::dbNet*> GlobalRouter::getNetsToRoute()
...
+
+void GlobalRouter::findBufferPinPostions(Net* net1,
+                                         Net* net2,
+                                         odb::Point& pin_pos1,
+                                         odb::Point& pin_pos2)
+{
+  for (const Pin& pin1 : net1->getPins()) {
+    if (!pin1.isPort()) {
+      for (const Pin& pin2 : net2->getPins()) {
+        if (!pin2.isPort()) {
```
maliberty, line 3849: This could be more efficient though I'm not sure it matters.

src/rsz/test/repair_setup6.ok
eder-matheus: Resizer is removing fewer buffers with my update. I guess preserving the net topologies helps avoid extra buffer removals, but the QoR is affected.

src/grt/src/GlobalRouter.cpp
eder-matheus: I found that way easier because I already have access to both pin positions on the grid. I could iterate only through the pins of one net and find the buffer that connects them, but I would need to compute the position on grid of the pins.

eder-matheus:
@maliberty This is ready for review. Public and private CI are passing, and we have made good improvements in QoR, especially in TNS. I see some small degradations, but we don't have failures in the CI.

maliberty (approved):

Issue #5530, opened 2024-08-08
cmake try to install dependency googletest/archive/refs/tags/v1.13.0.zip


titan73:
### Describe the bug

As I understand, etc/DependencyInstaller.sh is supposed to do this job.

Due to IO firewall policies it is easier to deal with dependencies in one step.

### Expected Behavior

Should not deal with dependencies installation.

### Environment

```shell
Latest git.
```

### To Reproduce

cd build && cmake ..

### Relevant log output

```shell
CMake Error at googletest-subbuild/googletest-populate-prefix/src/googletest-populate-stamp/download-googletest-populate.cmake:170 (message):
  Each download failed!
    error: downloading 'https://github.com/google/googletest/archive/refs/tags/v1.13.0.zip' failed
          status_code: 6
          status_string: "Couldn't resolve host name"
          log:
          --- LOG BEGIN ---
          getaddrinfo(3) failed for github.com:443
  Could not resolve host: github.com
```

maliberty:
Looks like you have a dns problem or a network issue as there is nothing we can do if github.com can't be resolved.  I don't see this as an OR issue.  Is it a transient problem?

titan73:
No it's not transient. I can't install dependencies on my laptop from my company network so I use my mobile phone to avoid the firewall. I generally do it when running etc/DependencyInstaller.sh then switch back to company network for all the rest.
I can do it also when running cmake but what is the reason some dependencies are install during cmake. Should be done with etc/DependencyInstaller.sh? Is there a particular reason?

maliberty:
@QuantamHD I believe you set this up so perhaps you can answer.

QuantamHD:
The reason is that relying on the system Gtest results in a lot of incompatibility, and the prefered installation path is to dynamically load in the project the way we've done so. 

@titan73 We could probably offer an override var to the URL so you could set it to some file:/// path on your local machine.  

Would that be sufficient?

gadfort:
@titan73 could you disable the test builds with `-DENABLE_TESTS=OFF` that should prevent the GoogleTest stuff from triggering?

titan73:
@gadfort Yes indeed. This prevent the issue.

titan73:
@QuantamHD Maybe my question was not clear. I understand the needs to use custom version of tools rather than using the system version. The question was about why it's done during cmake rather the DependencyInstaller.sh

[...]

PR #5592, opened 2024-08-15
Move install of gtest to DependencyInstaller.sh.

titan73:
For https://github.com/The-OpenROAD-Project/OpenROAD/issues/5530.

Tested with clean build and make test.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -79,6 +79,9 @@ _installCommonDev() {
     cuddVersion=3.0.0
     lemonVersion=1.3.1
     spdlogVersion=1.8.1
+    gtestVersion=1.13.0
+    gtestChecksum="a1279c6fb5bf7d4a5e0d0b2a4adb39ac"
+
 
     rm -rf "${baseDir}"
     mkdir -p "${baseDir}"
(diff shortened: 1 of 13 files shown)


QuantamHD:
Yeah. I'm personally opposed to this. Sorry I forgot to respond to your original issue, but pull the system GTest is exactly what we want to avoid. Most of the time Gtest is just directly included in a repo this is a nicer alternative to committing the files to the repo.

This is just generally the way Gtest asks you to install into your repo. https://github.com/google/googletest/blob/main/googletest/README.md#build-with-cmake
image

I understand your company has firewall rules that's blocking this stuff, and I think a better approach for you is to be able to override the URL to a local file. We could dependency install a tar file somewhere on the system and then point Fetch content at that, but I just fear a super super old version of gtest that just happens to be installed on someones machine.

titan73:
@QuantamHD I don't know what you mean by "pulling the system GTest". I do exactly things the same way as in DependencyInstaller.sh.

Who are the "we" in "what we want to avoid"?

AFAIK:
- OpenSTA and abc are submodules
- DependencyInstaller.sh manages over 20 dependencies
- Gtest is the single dependency manage by the build system

So can you explain why Gtest have an exception?

I noticed cudd has been recently added into DependencyInstaller.sh. Why not integrated into the build system as got gtest?

QuantamHD:
By the system GTest I mean the version the user happened to install.

We is in general the project, but I don't speak for everyone. What I worry about is getting a really old version of Gtest where the tests don't work because the user has an out of date version of Gtest. Gtest in particular is finicky about versions and changes frequently which is why most projects just include it in their build tree. 

If we can avoid relying on a system dependency that's more ideal since we build it in tree rather than relying on something on the system. Ideally we could just build and statically link our transitive dependency closure from source and do away with dependency installer entirely. But for now every dependency that doesn't have to be in the dependency installer is one less platform issue we run into.

QuantamHD:
I'm totally fine if you want to download the gtest zip file somewhere on the system from the dependency installer and point fetch content at that file. It achieves the same thing as pulling it from github.

titan73:
I clearly understand that we want to use our own installed version of gtest (1.13.0) and not the system one installed with apt. The PR does this and I tested version 1.13.0 is used whatever the system one is installed or not. But indeed it is installed in /usr/local. If /usr/local should be avoided, all the prefixes in deps installer should be modfied to a local directory then.

maliberty:
We assume that we can install to /usr/local so I don't see this as an issue.  We do allow a prefix to be specified for those that need it.

QuantamHD:
This PR is fine. I personally dislike /usr/local install stuff and would happily remove it all if we could, but that's a personal issue, and if this helps you use OpenROAD we should move ahead with it.

titan73:
Rebasing.
clang-tidy-review fails because the CI is not update with new deps.
I guess the 2 builds too although oddly the details show all steps green.

titan73:
Is there anything more to do in this PR?
The CI needs to be updated to make the checks pass. I don't know how to do it.

vvbandeira (approved):

Issue #5577, opened 2024-08-14
Resizing Problems


AcKoucher:
### Description

Currently, RepairDesign uses a strategy of resizing cells to a target slew in order to perform Post Placement Optimization. With the current methodology, the Resizer ends up "fixing" things that are not actually violations of any rule leading to over-sizing.

Note: The API that resizes to the target slew is also used in RepairSetup and RepairHold.

### Suggested Solution

(Some) Current problems:
  - [ ] Cells are resized regardless of violations (No DRC, Setup/Hold violations)
  - [ ] FFs are being touched.
  - [ ] Special cells are also touched. They should be checked with the cell footprint.
  - [ ] Clock buffers are used for data network.

Add fanout-of-4 (FO4) Rule
  - [ ] Gates that don't respect FO4 should be resized.
  - [ ] Gates with cap ratio that don't violate FO4 shouldn't be further sized

maliberty:
@povik you will take over the remaining tasks

maliberty:
@povik how much of phases 2 & 3 remains?  We should make sub-issues for anything yet to do as this is rather too large an item to track.

povik:
From phase 2:

> Do sequential sizing based on area increase (nominal buffer area threshold)

is missing

From phase 3, all is missing.

povik:
I'll discuss "Do sequential sizing based on area increase" with Cho.

Ad phase 3:

* https://github.com/The-OpenROAD-Project/OpenROAD/issues/8013
* https://github.com/The-OpenROAD-Project/OpenROAD/issues/8014

maliberty:
So "Add FO4" and children are done?

povik:
That's right, though we still want to adjust when in the flow we invoke it.

povik:
All items are implemented or tracked by a narrower ticket except for

> Do sequential sizing based on area increase (nominal buffer area threshold)

We are not using FO4 sizing in the flow so I think it can be closed as completed/stale.

PR #5601, opened 2024-08-16
rsz: avoid resizing registers

AcKoucher:
Beginning of the work for #5577.


AcKoucher:
Running Secure-CI.

AcKoucher:
QoR study showed that different PDKs may have different behaviors when it comes to area efficiency for FF resizing. This will be revisited when rewriting the core sizer.

PR #5620, opened 2024-08-19
gui: Issue an error instead of crashing when loading a drc file without a database opened.

titan73:


src/gui/src/drcWidget.cpp
```diff
@@ -308,6 +308,14 @@ void DRCWidget::setLogger(utl::Logger* logger)
 
 void DRCWidget::selectReport()
 {
+  if (!block_) {
+    try {
+      logger_->error(utl::GUI, 104, "No database has been loaded");
+    } catch (const std::runtime_error& e) {
+    }
```
maliberty, line 315: This makes no sense.  Drop the try-catch.  As written this is equivalent to a warning.  Why shouldn't it be an exception that throws?

titan73: I put a try-catch because, contrary to a warn, it triggers an exception and crashes:

```
[ERROR GUI-0104] No database has been loaded
terminate called after throwing an instance of 'std::runtime_error'
  what():  GUI-0104
Signal 6 received
```

This is how errors works no?

gadfort: @maliberty the GUI doesn't naturally handle the exceptions like the TCL interface does. What is there is equivalent to other places in the GUI code where Openroad might throw and exception and it needs to be handled properly.

maliberty: See #5624

maliberty:
I've merged the other PR so you should be able to drop the try-catch here.

titan73:
Ok. Thanks. I've just updated the PR.

src/gui/src/drcWidget.cpp
```diff
@@ -308,6 +308,11 @@ void DRCWidget::setLogger(utl::Logger* logger)
 
 void DRCWidget::selectReport()
 {
+  if (!block_) {
+    logger_->error(utl::GUI, 104, "No database has been loaded");
+    return;
```
maliberty, line 313: no need for a return as error will throw and return will be unreachable

maliberty (approved):

PR #5621, opened 2024-08-19
gui: Demangle type name when printing an warning.

titan73:
The type names are mangled.

```
$ echo "PN3gui12DRCViolationE" | c++filt -t
gui::DRCViolation*
$ echo "Dn" | c++filt -t
decltype(nullptr)
```

The type name "decltype(nullptr)" is more straightforwards than "Dn".

### DIFF
--- a/src/gui/src/gui.cpp
+++ b/src/gui/src/gui.cpp
@@ -287,10 +287,11 @@ Selected Gui::makeSelected(const std::any& object)
   if (it != descriptors_.end()) {
     return it->second->makeSelected(object);
   }
-  logger_->warn(utl::GUI,
-                33,
-                "No descriptor is registered for {}.",
-                object.type().name());
+  char* type_name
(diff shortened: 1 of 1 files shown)


src/gui/src/gui.cpp
```diff
@@ -287,10 +287,11 @@ Selected Gui::makeSelected(const std::any& object)
...
   }
-  logger_->warn(utl::GUI,
-                33,
-                "No descriptor is registered for {}.",
-                object.type().name());
+  logger_->warn(
+      utl::GUI,
+      33,
+      "No descriptor is registered for type {}.",
+      abi::__cxa_demangle(object.type().name(), nullptr, nullptr, nullptr));
```
maliberty, line 294: The memory from this call is leaked.

titan73: Ok. Thanks. Will update.

maliberty (approved):

PR #5623, opened 2024-08-19
gui: Do not init std::any item var if we intend to use has_value() on it.

titan73:
The var should no be inited otherwise has_value() will always be true. That way the condition of the if will decide if the var has been set or not.

Fixes warning:
    [WARNING GUI-0033] No descriptor is registered for Dn.
and now
    [WARNING GUI-0033] No descriptor is registered for type
decltype(nullptr).

Caused by frFakeVSS and frFakeGND:
    [WARNING GUI-0044] Unable to find net (line: 104375): frFakeVSS
    [WARNING GUI-0044] Unable to find net (line: 104378): frFakeVDD


maliberty (approved):
Thanks for the fix!

PR #5630, opened 2024-08-22
etc: add gtest to DependencyInstaller

vvbandeira:
cherry-picked from #5592

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -79,6 +79,9 @@ _installCommonDev() {
     cuddVersion=3.0.0
     lemonVersion=1.3.1
     spdlogVersion=1.8.1
+    gtestVersion=1.13.0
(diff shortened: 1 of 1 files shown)


stefanottili:
This comit broke compling on MacOS.
"brew install googletest" is not suficcient, linking steps fail with
```
[ 73%] Built target odb_test_helper
ld: warning: ignoring duplicate libraries: '../../src/db/libdb.a', '../../src/zutil/libzutil.a'
ld: library 'gtest' not found
```

maliberty:
@eder-matheus would you try?

gadfort:
@maliberty this introduced a new requirement which is not captured in the dependency installer.
I need to install `unzip` which is not available in the minimal ubuntu docker images, if I want to install this. This used to not be required.

maliberty:
@vvbandeira please add unzip

PR #5646, opened 2024-08-25
CTS fix copy paste bug on max fanout

arthurjolo:
This PR fixes a copy paste error on PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/5516.


maliberty (commented):
Do we have a unit test for this?  It seems like something should be affected by this change.

PR #5651, opened 2024-08-26
nix: migrate to flake

donn:
Replacing `etc/shell.nix`, this is a [Flake](https://nixos.wiki/wiki/Flakes) based on the OpenROAD Nix derivation used for OpenLane 2. This enhances reproducibility, and with Nix installed with flakes enabled, getting an OpenROAD build environment is as simple as typing `nix develop`.

Also fixes issue with versions of OpenROAD built in the dev environments failing to launch the GUI. 

Should support both Linux and macOS on x86-64 and arm64. Tested on Sonoma/Apple Silicon.

---

Some notes:

### Submodule dependence

Unlike the OpenLane 2 derivation set, this relies on the submodules to be cloned. Without adding `flake.nix` files or basically 
* You will have to clone the git repository and type `nix develop` from there
* `nix {run,profile install} github:The-OpenROAD-Project/OpenROAD?submodules=1` must include the argument at the end and will do all of the following:
   * Clone with submodules (and thus, the full history of all of openroad, opensta and abc)
   * Build OpenROAD with libabc and libsta as part of the same compile unit

This will take forever and I did not test it. Given that most people will elect to clone the repo, I am not losing much sleep over this.

### On the GUI issue

The long explanation for the GUI issue is that Nix applies a wrapper to Qt apps to set the environment for plugin search paths. This unfortunately though does not propagate to development environments such as `nix-shell`/`nix develop` unless set explicitly. (cc @widlarizer)


gadfort:
Shouldn't this have a CI build associated with it to ensure it remains functional?

maliberty (approved):

maliberty:
> Shouldn't this have a CI build associated with it to ensure it remains functional?

I've been assuming that efabless test and maintain this.

widlarizer:
Unfortunately `nix run "github:The-OpenROAD-Project/OpenROAD?submodules=1"` fails to build (quotes added only because fish shell interprets the question mark as a keyword). It's only with a local checkout that `nix run ".?submodules=1"` works. Which is good enough for me

widlarizer:
Seems that flake built OpenROAD doesn't see .git at build time and therefore `cmake/GetGitRevisionDescription.cmake` sets the version to `GITDIR-NOTFOUND`. This makes my benchmarking workflow risky with unreported differences in OR versions

maliberty:
@donn any comments on the above issues?

donn:
Inherent to Nix tbh so not much I can do, `.git` is highly stateful and cannot be included in a reproducible build (even without flakes, it should not be included.) 

We do an out-of-tree build and so… this is a viable option for us:

https://github.com/efabless/openlane2/blob/8871ff3ecc0e9938ffd74d32005271c2795a8ee8/nix/openroad.nix#L87

PR #5652, opened 2024-08-26
nix: add missing 'gtest' dependency

donn:
This PR's purpose was changed after discussion with maintainers. Original text follows:

---
Per @gadfort's suggestion

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -33,6 +33,7 @@
 ##
 ###############################################################################
 {
+  flake,
   lib,
   clangStdenv,
   fetchFromGitHub,
@@ -62,6 +63,7 @@
   flex,
   bison,
   clang-tools_14,
+  gtest,
   # or-tools
   stdenv,
   overrideSDK,
@@ -87,7 +89,7 @@
   self = clangStdenv.mkDerivation (finalAttrs: {
     name = "openroad";
 
-    src = ./.;
+    src = flake;
 
     cmakeFlags = [
       "-DTCL_LIBRARY=${tcl}/lib/libtcl${clangStdenv.hostPlatform.extensions.sharedLibrary}"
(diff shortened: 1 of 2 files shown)


maliberty:
Over 11m in and it hasn't even started to compile.  Nix doesn't seem to be a great match for transient workers.

donn:
Yeah, I'm not sure if you want to wait a good hour on every pull request or commit. Let me keep working on this PR though- I'll get rid of the action but it appears I need to do some further fixes

maliberty:
13.5m to end with "Could NOT find GTest" (a new dependency)

donn:
It should be mentioned that most of this time is spent building `or-tools` to patch in https://github.com/google/or-tools/commit/af44f98dbeb905656b5a9fc664b5fdcffcbe1f60 - that's a good 5-10 minutes on a CI machine on its own. Although on a local machine it would be cached after the first build.

vvbandeira:
If you can give me a Dockerfile with the dependencies needed to run and test `nix,` I can add it to our Jenkins pipeline. This way, the build time should be reduced, and we can make it conditional only when the `*.nix` files change.

donn:
The downloaded dependencies themselves don't take that long- it's building `or-tools` that's annoyingly long. We use [cachix](cachix.org) to upload dependencies that take forever to build for our own purposes, not sure if you want to consider that

Triggering only on Nix file changes would miss changes to the OpenROAD build process that may require attention so I wouldn't recommend that.

vvbandeira:
> Triggering only on Nix file changes would miss changes to the OpenROAD build process that may require attention so I wouldn't recommend that.

Yes, that is true. I was thinking more about a subset of files (e.g., nix + `DependencyInstaller.sh`/Docker/...), etc. But we can get back to this later.

TBH, I would prefer to use more Jenkins and less GHA where possible that's the main reason I suggested the approach. But if you can make a GHA that has an acceptable runtime, that is fine as well.

donn:
@vvbandeira We can work through standing this up on Jenkins on Slack if you'd like.

donn:
@gadfort Yeah, it was considered too slow. :(

donn:
Ready to merge. Can contact @vvbandeira offline about getting the integration test in Jenkins somehow.

vvbandeira (approved):

Issue #4441, opened 2023-12-25
Dumping estimated parasitics as SPEF


donn:
### Description

You can run `estimate_parasitics` prior to RCX extraction with OpenROAD, however, there isn't a way to dump this data to a SPEF file as far as we can tell. We're trying to bring up a PDK internally and it would be helpful for calibrating the RC values. 

### Suggested Solution

A way to dump the estimated parasitics would be ideal.

maliberty:
There isn't a method currently.  How would you use the spef?  If you just need something like total C that's simpler.

QuantamHD:
I'd also like this

donn:
> There isn't a method currently. How would you use the spef? If you just need something like total C that's simpler.

We're trying to compare the values pre and post-routing (and RCX) to try and calibrate the RC values to try to get the pre-routed values a bit closer to post-route extracted parasitics.

If total C (presuming you meant per-net) can also be dumped the same format pre and post routing and RCX, that would also be helpful as a starting point.

rovinski:
@donn take a look in ORFS where we have a script that does linear correlation between GPL, GRT, and RCX parasitics on both total cap and total res.

[`write_net_rc.tcl`](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/6b53853149d273909d7458f3f6dcba1397767454/flow/util/write_net_rc.tcl) contains a bunch of procs for extracting cap and res for each net.
[`write_net_rc_script.tcl`](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/6b53853149d273909d7458f3f6dcba1397767454/flow/util/write_net_rc_script.tcl) uses those procs to get the GPL, GRT, and RCX cap and res into CSV files.
[`correlateRC.py`](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/6b53853149d273909d7458f3f6dcba1397767454/flow/util/correlateRC.py) does a simple linear regression on the estimated cap and the extracted cap to generate a correlated `setRC.tcl`

[`make correlate_rc`](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/6b53853149d273909d7458f3f6dcba1397767454/flow/util/utils.mk#L46-L48) will run these scripts on a specific design
[...]

PR #5656, opened 2024-08-26
grt/rsz: write spef for estimated parasitics

joaomai:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/4441.

Added as an option to the estimate_parasitics command, using -spef_file to pass the file path as an argument, writing one or more files based on how many corners there are. 
The information is written to the file during estimation, writing the net data before RC reduction is performed.

I'm unsure about the units in the header, using the same ones from the rcx header currently.


src/rsz/src/Resizer.tcl
```diff
@@ -313,21 +313,26 @@ proc set_wire_rc { args } {
   }
 }
 
-sta::define_cmd_args "estimate_parasitics" { -placement|-global_routing }
+sta::define_cmd_args "estimate_parasitics" { -placement|-global_routing \
+                                            [-spef_file]}
```
maliberty, line 317: [-spef_file filename]

src/rsz/include/rsz/Resizer.hh
```diff
@@ -234,8 +234,8 @@ class Resizer : public dbStaState
   double wireClkCapacitance(const Corner* corner) const;
   double wireClkHCapacitance(const Corner* corner) const;
   double wireClkVCapacitance(const Corner* corner) const;
-  void estimateParasitics(ParasiticsSrc src);
-  void estimateWireParasitics();
+  void estimateParasitics(ParasiticsSrc src, const char* file_path = "");
+  void estimateWireParasitics(const char* path = "");
```
maliberty, line 238: std::string_view is more flexible than char*

src/rsz/include/rsz/Resizer.hh
```diff
@@ -404,6 +404,16 @@ class Resizer : public dbStaState
...
 
+  void openSpefFile(const char* path);
+  void closeSpefFile();
+  void writeSpefHeader(Corner* corner, dbNetwork* network);
+  void writeSpefPorts(Corner* corner, dbNetwork* network);
+  void writeSpefNet(Corner* corner,
+                    const Net* net,
+                    Parasitic* parasitic,
+                    dbNetwork* network,
+                    Parasitics* parasitics);
```
maliberty, line 415: Can the writer be split into its own class?  I don't see a dependency on the rest of resizer in this code and it is shared with grt.

src/rsz/include/rsz/Resizer.hh
```diff
@@ -739,6 +749,9 @@ class Resizer : public dbStaState
 
   dpl::Opendp* opendp_ = nullptr;
 
+  std::map<Corner*, std::ofstream> spef_file;
+  bool write_spef_file;
```
maliberty, line 753: Data members should have a trailing underscore (https://google.github.io/styleguide/cppguide.html#Variable_Names)

maliberty, line 753: Give a default value, eg `bool write_spef_file = false;`.   Is the necessary or is it equivalent to spef_files.empty()?

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -736,4 +742,184 @@ void Resizer::parasiticsInvalid(const dbNet* net)
...
+      std::string suffix("_");
+      suffix.append(corner->name());
+      if (file_path.find(".spef") != std::string::npos
+          || file_path.find(".SPEF") != std::string::npos) {
+        file_path.insert(file_path.size() - 5, suffix);
+      } else {
+        file_path.append(suffix);
+      }
+
+      spef_file[corner].open(file_path, std::ofstream::out);
```
maliberty, line 766: You should check that open succeeded

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -736,4 +742,184 @@ void Resizer::parasiticsInvalid(const dbNet* net)
...
+  spef_file[corner] << "*PROGRAM \"OpenROAD\"" << '\n';
+  spef_file[corner] << "*VERSION \"1.0\"" << '\n';
+  spef_file[corner] << "*DESIGN_FLOW \"NAME_SCOPE LOCAL\" \"PIN_CAP NONE\""
+                    << '\n';
+  spef_file[corner] << "*DIVIDER /" << '\n';
+  spef_file[corner] << "*DELIMITER :" << '\n';
+  spef_file[corner] << "*BUS_DELIMITER []" << '\n';
+  spef_file[corner] << "*T_UNIT 1 NS" << '\n';
+  spef_file[corner] << "*C_UNIT 1 PF" << '\n';
+  spef_file[corner] << "*R_UNIT 1 OHM" << '\n';
```
maliberty, line 799: These should come from sta

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -736,4 +742,184 @@ void Resizer::parasiticsInvalid(const dbNet* net)
...
+  spef_file[corner] << "*L_UNIT 1 HENRY" << '\n';
+  spef_file[corner] << '\n';
+}
+
+void Resizer::writeSpefPorts(Corner* corner, dbNetwork* network)
+{
+  auto pin_iter = network->pinIterator(network->topInstance());
```
maliberty, line 806: sta is old fashioned and this pointer must be freed by the caller.  The easiest is to store the result value in a unique_ptr.

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -736,4 +742,184 @@ void Resizer::parasiticsInvalid(const dbNet* net)
...
+  while (pin_iter->hasNext()) {
+    sta::Pin* pin = pin_iter->next();
+    odb::dbITerm* iterm = nullptr;
+    odb::dbBTerm* bterm = nullptr;
+    odb::dbModITerm* moditerm = nullptr;
+    odb::dbModBTerm* modbterm = nullptr;
+    network->staToDb(pin, iterm, bterm, moditerm, modbterm);
+
+    spef_file[corner] << bterm->getName() << " ";
+    if (bterm->getIoType() == odb::dbIoType::INPUT) {
```
maliberty, line 817: What if this isn't a bterm?  An iterm is possible and a mod* should be an error.

maliberty (changes requested):

src/grt/include/grt/GlobalRouter.h
```diff
@@ -195,7 +195,7 @@ class GlobalRouter : public ant::GlobalRouteSource
   bool isCoveringPin(Net* net, GSegment& segment);
   std::vector<Net*> initFastRoute(int min_routing_layer, int max_routing_layer);
   void initFastRouteIncr(std::vector<Net*>& nets);
-  void estimateRC();
+  void estimateRC(const char* path = "");
```
QuantamHD, line 198: Can we change this to be std::ostream instead of a path? It makes it way more versatile, and easier to test.

src/rsz/include/rsz/Resizer.hh
maliberty: There is a request to take 
```
 std::map<Corner*, std::ostream*>& spef_files
```
instead of a file_path so that the client can control the output (to a ofstream or a ststream).  The .i could handle setting these up for file based output.

maliberty: Note that  corners could be a subset of all available corners

maliberty: You could keep the write to file version an an overload that uses the other API as its implementation.

rovinski:
Not sure I'm fond of -spef as a flag to `estimate_parasitics`. Couldn't this instead use the existing API by doing `write_spef` after `estimate_parasitics`? You could print a message during `write_spef` indicating what kind of parasitics it's writing (global vs. detailed).

maliberty:
sta reduces the parasitics so you can't write them afterwards except in reduced form which isn't that helpful.

rovinski:
But could you, e.g., just add a Tcl hook or a SWIG hook which determines which function to call? Or are you saying that to write the SPEF without reduced parasitics then it has to be written out immediately?

maliberty:
Yes they have to be written immediately as the unreduced form is not stored anywhere.  Storing it would be quite expensive on the off change someone wants to write them out.

rovinski:
That's fair then.

maliberty:
@joaomai there is some feedback that still needs addressing before this can be merged.

src/grt/src/GlobalRouter.cpp
```diff
@@ -448,18 +450,25 @@ NetRouteMap GlobalRouter::findRouting(std::vector<Net*>& nets,
...
+
+void GlobalRouter::estimateRC(std::map<sta::Corner*, std::ostream*>& spef_files)
 {
   // Remove any existing parasitics.
   sta_->deleteParasitics();
 
   // Make separate parasitics for each corner.
   sta_->setParasiticAnalysisPts(true);
 
+  rsz::SpefWriter spef_writer(logger_, sta_, spef_files);
```
eder-matheus, line 466: Instead of instantiate the SpefWriter here, you could have a single spef_writer on resizer and pass it as a pointer to estimateRC. This way, estimateRC don't need to know about the map of Corner to ostream, since it will be encapsulated on resizer.

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -272,13 +273,20 @@ void Resizer::ensureParasitics()
 }
 
 void Resizer::estimateParasitics(ParasiticsSrc src)
+{
+  std::map<Corner*, std::ostream*> spef_files;
+  estimateParasitics(src, spef_files);
+}
+
+void Resizer::estimateParasitics(ParasiticsSrc src,
```
eder-matheus, line 281: You can initialize the spef_writer_ here and pass it to the functions, instead of passing the map.

src/grt/src/MakeWireParasitics.cpp
```diff
@@ -74,6 +75,32 @@ MakeWireParasitics::MakeWireParasitics(utl::Logger* logger,
...
+      resizer_(resizer),
+      spef_writer_(spef_writer),
+      sta_(sta),
+      network_(sta_->getDbNetwork()),
+      parasitics_(sta_->parasitics()),
+      arc_delay_calc_(sta_->arcDelayCalc()),
+      min_max_(sta::MinMax::max()),
+      resistor_id_(1)
+{
+  for (sta::Corner* corner : *sta_->corners()) {
```
eder-matheus, line 98: Why creating a new constructor? Also, I don't think the class constructor should be writing anything related to the spef file.

src/rsz/include/rsz/Resizer.hh
```diff
@@ -739,6 +745,9 @@ class Resizer : public dbStaState
 
   dpl::Opendp* opendp_ = nullptr;
 
+  std::map<Corner*, std::ofstream> spef_file;
```
eder-matheus, line 748: Should these still be here? They seem to be owned by SpefWriter class now.

eder-matheus (changes requested):

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -457,6 +474,9 @@ void Resizer::makePadParasitic(const Net* net)
 
     // Use a small resistor to keep the connectivity intact.
     parasitics_->makeResistor(parasitic, 1, .001, n1, n2);
+    if (spef_writer_) {
```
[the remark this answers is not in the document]
joaomai: I don't have access to the spef_files_ map outside the rsz::estimateParasitics function. I think it's simpler to not initialize the spef_writer on line 284 if the received map is empty.

Also, maybe I should change the name from spef_files_ to spef_streams_ since it will not be necessarily written to a file now that it's using std::ostream*?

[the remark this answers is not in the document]
eder-matheus: Sounds good.

src/rsz/src/Resizer.i
```diff
@@ -331,12 +332,42 @@ wire_clk_capacitance(const Corner *corner)
...
+        if (file_path.find(".spef") != std::string::npos
+            || file_path.find(".SPEF") != std::string::npos) {
+          file_path.insert(file_path.size() - 5, suffix);
+        } else {
+          file_path.append(suffix);
+        }
+
+        std::ofstream* file = new std::ofstream(file_path);
+
+        if (file->is_open()) {
```
eder-matheus, line 357: Should it be an error if the file isn't open?

src/rsz/src/SpefWriter.cc
```diff
@@ -0,0 +1,246 @@
...
+
+void SpefWriter::writeSpefPorts(Corner* corner)
+{
+  auto it = spef_files_.find(corner);
+  if (it == spef_files_.end()) {
+    return;
+  }
+  std::ostream& stream = *it->second;
+
+  auto pin_iter = network_->pinIterator(network_->topInstance());
```
eder-matheus, line 104: Using a std::unique_ptr here, as suggested by @maliberty, is clearer and remove the need for the `delete` in the end of the function.

eder-matheus (changes requested):

src/rsz/src/EstimateWireParasitics.cc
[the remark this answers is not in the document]
eder-matheus: This comment wasn't addressed yet.

eder-matheus (approved):

eder-matheus:
@maliberty Any more comments? I believe we need your approval to merge this PR.

src/rsz/include/rsz/SpefWriter.hh
```diff
@@ -0,0 +1,81 @@
...
+
+class SpefWriter
+{
+ public:
+  SpefWriter(Logger* logger,
+             dbSta* sta,
+             std::map<Corner*, std::ostream*>& spef_streams);
+  ~SpefWriter() = default;
+  void writeSpefHeader(Corner* corner);
+  void writeSpefPorts(Corner* corner);
```
maliberty, line 69: These two only use the corner to look up the stream and force the caller to loop over all corners.  It seems simpler to have writeSpefHeaders() & writeSpefPorts() and handle the iteration internally.

Is there a use case where we don't want to write these?  If not they should be automatically handled (eg on first writeSpefNet or construction).

src/rsz/src/SpefWriter.cc
```diff
@@ -0,0 +1,246 @@
...
+
+    if (iterm != nullptr) {
+      stream << iterm->getName() << " ";
+      if (iterm->getIoType() == odb::dbIoType::INPUT) {
+        stream << "I";
+      } else if (iterm->getIoType() == odb::dbIoType::OUTPUT) {
+        stream << "O";
+      } else {
+        stream << "B";
+      }
```
maliberty, line 124: Stylistically I would avoid repeating this block and just get the io type in the iterm/bterm blocks and then convert it to text once afterwards.

src/rsz/src/SpefWriter.cc
```diff
@@ -0,0 +1,246 @@
...
+  stream << '\n';
+}
+
+void SpefWriter::writeSpefNet(Corner* corner,
+                              const Net* net,
+                              Parasitic* parasitic)
+{
+  auto it = spef_streams_.find(corner);
+  if (it == spef_streams_.end()) {
+    return;
```
maliberty, line 152: This seems like an error.

src/rsz/src/SpefWriter.cc
```diff
@@ -0,0 +1,246 @@
...
+      network_->staToDb(pin, iterm, bterm, moditerm, modbterm);
+
+      if (iterm != nullptr) {
+        stream << "*I " << parasitics_->name(node) << " ";
+        if (iterm->getIoType() == odb::dbIoType::INPUT) {
+          stream << "I";
+        } else if (iterm->getIoType() == odb::dbIoType::OUTPUT) {
+          stream << "O";
+        } else {
+          stream << "B";
```
maliberty, line 176: I see this repeats 2X again - it can be factored to a common function.

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -273,17 +274,37 @@ void Resizer::ensureParasitics()
...
+    spef_writer_ = new SpefWriter(logger_, sta_, spef_streams);
+    for (auto& [corner, _] : spef_streams) {
+      spef_writer_->writeSpefHeader(corner);
+      spef_writer_->writeSpefPorts(corner);
+    }
+  }
+
   switch (src) {
     case ParasiticsSrc::placement:
       estimateWireParasitics();
```
maliberty, line 294: In a future PR it would be good to handle placement based parasitics too.

src/rsz/include/rsz/SpefWriter.hh
```diff
@@ -0,0 +1,81 @@
...
+  void writeSpefNet(Corner* corner, const Net* net, Parasitic* parasitic);
+
+ private:
+  Logger* logger_;
+  dbSta* sta_;
+  dbNetwork* network_;
+  Parasitics* parasitics_;
+
+  std::map<Corner*, std::ostream*>& spef_streams_;
```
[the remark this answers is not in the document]
joaomai: In an earlier comment it was requested to make it a  `std::map<Corner*, std::ostream*>& spef_files` so that the client can control the output.

src/rsz/src/SpefWriter.cc
joaomai: This is here for the case where the caller wants to write only a subset of the available corners using the rsz::estimateWireParastics or grt::estimateRC functions.

src/rsz/src/EstimateWireParasitics.cc
joaomai: It was done by setting a Resizer spef_writer_ member instead of passing the SpefWriter* through functions like in GlobalRouter. I've changed it to be similar to GlobalRouter now.

I'll make a new commit after we end the discussion on the other comments.

src/rsz/src/SpefWriter.cc
maliberty: This seems like the opposite - they are writing a corner they didn't declare any ostream for.  A corner in a subset would be found.

src/rsz/include/rsz/SpefWriter.hh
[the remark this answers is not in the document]
maliberty: I agree to the map of ostreams, just debating the '&'.  Keeping it by reference makes this class vulnerable to the caller deallocating the map before writing is done.

src/rsz/src/SpefWriter.cc
```diff
@@ -58,98 +58,101 @@ SpefWriter::SpefWriter(Logger* logger,
...
-  std::transform(res_unit.begin(), res_unit.end(), res_unit.begin(), ::toupper);
-
-  stream << "*T_UNIT 1 " << time_unit << '\n';
-  stream << "*C_UNIT 1 " << cap_unit << '\n';
-  stream << "*R_UNIT 1 " << res_unit << '\n';
-  stream << '\n';
 }
 
-void SpefWriter::writeSpefPorts(Corner* corner)
+std::string getIoDirectionText(const odb::dbIoType& ioType)
```
maliberty, line 103: char would do for a return type.

maliberty (changes requested):
Nearly there

maliberty (approved):

donn:
Great work!

PR #5660, opened 2024-08-27
Add option to generate CMake parameters pointing to installed dependencies

eszpotanski:
This PR adds `-save-deps-prefixes=FILE` argument to `DependencyInstaller.sh`, which dumps additional CMake parameters in form of `-D <PackageName>_ROOT=<path>`.

Those parameters can be used for OpenROAD built, to make sure the correct dependencies' versions are used.

Moreover, `GTest::` was added to Google Test target link, fixing build errors (on Ubuntu 24.04):
```
/usr/bin/ld: cannot find -lgtest: No such file or directory
/usr/bin/ld: cannot find -lgmock: No such file or directory
/usr/bin/ld: cannot find -lgtest_main: No such file or directory
```

### DIFF
--- a/etc/Build.sh
+++ b/etc/Build.sh
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-cd "$(dirname $(readlink -f $0))/../"
+DIR="$(dirname $(readlink -f $0))"
+cd "$DIR/../"
 
 # default values, can be overwritten by cmdline args
 buildDir="build"
@@ -19,6 +20,7 @@ EOF
 fi
 cmakeOptions=""
 cleanBefore=no
+depsPrefixesFile=""
 keepLog=no
 compiler=gcc
 
@@ -30,7 +32,7 @@ OPTIONS:
   -cmake='-<key>=<value> [-<key>=<value> ...]'  User defined cmake options
                                                   Note: use single quote after
                                                   -cmake= and double quotes if
-                                                  <key> has muliple <values>
+                                                  <key> has multiple <values>
                                                   e.g.: -cmake='-DFLAGS="-a -b"'
   -compiler=COMPILER_NAME                       Compiler name: gcc or clang
                                                   Default: gcc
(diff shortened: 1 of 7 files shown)


etc/DependencyInstaller.sh
```diff
@@ -760,6 +778,9 @@ while [ "$#" -gt 0 ]; do
             fi
             export PREFIX="$(realpath $(echo $1 | sed -e 's/^[^=]*=//g'))"
             ;;
+        -save-deps-prefixes=*)
+            saveDepsPrefixes=$(realpath ${1#-save-deps-prefixes=})
+            ;;
```
vvbandeira, line 783: Would it make more sense to always save such file using a reasonable default name and plugging the file into `etc/Build.sh` so that it is always used?

eszpotanski: That was the final goal to use it by default, but I implemented it in ORFS: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2295

As it makes more sense to have it here, I'll add sane default and just reuse it from ORFS

eszpotanski: I've set the default path for the file to `openroad_deps_prefixes.txt` in the same directory as `DependencyInstaller` (usually `etc/openroad_deps_prefixes.txt`).

As building Docker image is multi-stage and `DependencyInstaller` is run from `/tmp`, different default path was chosen for Docker images - `/etc/openroad_deps_prefixes.txt`, which can be changed with argument from `DockerHelper` script.

`Build.sh` search for the file in OpenROAD's `etc` or system `/etc` directory or path to the file can be specify with `-deps-prefixes-file`

oharboe:
@vvbandeira Review?

etc/Build.sh
```diff
@@ -117,6 +132,20 @@ while [ "$#" -gt 0 ]; do
...
+        depsPrefixesFile="$DIR/openroad_deps_prefixes.txt"
+    elif [[ -f "/etc/openroad_deps_prefixes.txt" ]]; then
+        depsPrefixesFile="/etc/openroad_deps_prefixes.txt"
+    fi
+fi
+if [[ -f "$depsPrefixesFile" ]]; then
+    cmakeOptions+=" $(cat "$depsPrefixesFile")"
+    echo "[INFO] Using additional CMake parameters from $depsPrefixesFile"
+else
+    echo "[INFO] $depsPrefixesFile does not exist - CMake will choose the dependencies automatically"
```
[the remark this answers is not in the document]
vvbandeira: This or something to the same effect would be better. The default value for the variable is empty and thus can lead to the following message:
```
[INFO]  does not exist - CMake will choose the dependencies automatically
```

vvbandeira (changes requested):

etc/Build.sh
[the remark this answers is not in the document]
eszpotanski: Changed

eszpotanski:
I've applied suggestion and rebased the PR

PR #5665, opened 2024-08-27
gui: Add find by name to CTS viewer 

### LINKED ISSUE #5463: Add find by name to CTS viewer
### Description

To track:

From @precisionmoon `We want to find a register by name in the main GUI and see the matching sink in the clock tree viewer.`

### Suggested Solution

From @gadfort `It should be fairly easy to implement since you can piggy-back on the inspect connection to emit a second call to a sink in the viewer that can highlight it.`

### Additional Context

_No response_

LucasYuki:
Closes #5463.

Changes:
- Makes the find dialog search instances in the CTS Viewer. (Implements #5463)
- Find dialog focuses on the find text box when the popup opens

If instances are found, the CTS Viewer zooms to fit all found objects. 
When the design has multiple clock signals, the current tree changes to where the instances are found.


src/gui/src/clockWidget.cpp
```diff
@@ -1544,4 +1615,51 @@ void ClockWidget::fit()
...
+  }
+
+  std::set<int> found;
+  std::set<int> not_found;
+  int min_found = -1;
+  // ClockTreeView* finded_view;
+  for (int i = 0; i < views_.size(); i++) {
+    int n_found = views_[i]->findInCts(selections);
+
+    if (n_found) {
```
AcKoucher, line 1644: Avoid checking integers as if they were bools. Also, these names are not descriptive:
n_found --> `number_of_nodes_found` or `found_nodes_count` ..

src/gui/src/clockWidget.cpp
```diff
@@ -809,6 +817,58 @@ void ClockTreeView::fit()
...
+  std::set<ClockNodeGraphicsViewItem*> nodes = getNodes(selections);
+  if (!nodes.empty()) {
+    // remove old selection
+    clearSelection();
+    for (auto node : nodes) {
+      node->setSelected(true);
+    }
+    fitSelection();
+  }
+  return (int) nodes.size();
```
AcKoucher, line 855: This function is doing a lot of things: changing the current selection, fitting the new selection and returning the number of selected nodes. The name findInCts doesn't say much about what is being done.

One alternative could be to use a name like changeAndFitSelection() and you could keep the number of selected nodes in a member variable such as `number_of_selected_nodes_` to be used elsewhere.

src/gui/src/clockWidget.cpp
```diff
@@ -1544,4 +1615,51 @@ void ClockWidget::fit()
...
+{
+  if (views_.empty()) {
+    return;
+  }
+  if (selections.empty()) {
+    return;
+  }
+
+  std::set<int> found;
+  std::set<int> not_found;
```
AcKoucher, line 1638: These could have better names or at least a comment to make it clear that they're sets of indices of the views in which there are/aren't selected instances.

AcKoucher (changes requested):
@maliberty I'm not very familiar with the CTS Viewer code, so I made some more general review.

AcKoucher:
@LucasYuki FYI you can use the [special keywords in GH to link a PR to an issue(s)](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue), that way when the PR gets merged, the issue will be closed. This needs to be done in the PR description.

src/gui/src/clockWidget.cpp
```diff
@@ -809,6 +817,58 @@ void ClockTreeView::fit()
...
 }
 
+ClockNodeGraphicsViewItem* ClockTreeView::getItemFromName(
+    const std::string& name)
+{
+  if (items_.find(name) == items_.end()) {
+    return nullptr;
+  }
+
+  return items_[name];
```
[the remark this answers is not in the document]
LucasYuki: Fixed.
Now I'm using the iterator returned by the find method to return the pointer to the found item.

src/gui/src/clockWidget.cpp
```diff
@@ -1282,11 +1339,25 @@ ClockNodeGraphicsViewItem* ClockTreeView::addCellToScene(
...
+                            ClockNodeGraphicsViewItem* node,
+                            const QString& tooltip,
+                            sta::Delay delay)
+{
+  node->setPos({x, convertDelayToY(delay)});
+  node->setExtraToolTip(tooltip);
+  scene_->addItem(node);
+
+  items_[node->getInstName().toStdString()] = node;
+  // getInst()->getName()
```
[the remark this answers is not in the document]
LucasYuki: Unused code removed.

src/gui/src/clockWidget.cpp
```diff
@@ -1544,4 +1615,51 @@ void ClockWidget::fit()
...
+    return;
+  }
+  if (selections.empty()) {
+    return;
+  }
+
+  std::set<int> found;
+  std::set<int> not_found;
+  int min_found = -1;
+  // ClockTreeView* finded_view;
```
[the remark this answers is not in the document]
LucasYuki: Unused code removed.

src/gui/src/clockWidget.cpp
LucasYuki: Fixed.
I changed the function ClockTreeView::findInCts to ClockTreeView::changeSelection.
Now the function returns a boolean indicating if the selection changed.
The variable name was changed to selection_changed.

src/gui/src/clockWidget.cpp
LucasYuki: Fixed.
I changed the method `ClockTreeView::findInCts` to `ClockTreeView::changeSelection`.
Now the method returns a boolean indicating if the selection changed.

The method `ClockWidget::findInCts` calls `ClockTreeView::fitSelection` when the selection changes.

Since the number of currently selected nodes isn't used yet, I didn't implement a member variable, but it can be easily implemented as a method:
`int ClockTreeView::numberOfSelectedNodes() { (int) scene_->selectedItems().size(); };`

src/gui/src/clockWidget.cpp
LucasYuki: Fixed.

Changed `found` to `changed_views` and `not_found` to `not_changed_views`.
Also removed the variable `min_found`, now if the current view is not in the `changed_views` set, the current view is changed to `*(changed_views.begin())`.

src/gui/src/gui.i
```diff
@@ -561,14 +561,15 @@ int select(const std::string& type,
            const std::string& attribute = "",
            const std::string& value = "",
            bool case_sensitive = true,
-           int highlight_group = -1)
+           int highlight_group = -1,
+           bool find_in_cts = true)
```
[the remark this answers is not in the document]
LucasYuki: I changed the default to false, to have the same default value as the `Gui::select method`.

maliberty:
What is the thinking behind the option find_in_cts?  In what scenarios do we not want to find in cts?  The select command doesn't expose the option in tcl from what I see.  The inspector always passes true for instances.

LucasYuki:
The way it is implemented now it isn't possible to search nets in the Clock Tree Viewer, so when searching for nets we don't need to emit the `findInCts` signal.

I can remove this flag from the `Gui::select` and always pass as true the argument `find_in_cts` to the slot `MainWindow::addSelected`, but since the `MainWindow::addSelected` slot is responsible for emitting the `findInCts` signal and this slot is called in other places, on this slot the flag is necessary.

Another alternative is to implement a checkbox in the find dialog, something like "Find the Clock Tree Viewer" to enable the search, so the user can choose to use or not this feature.

maliberty:
We don't have such an option for find and the other displays (layout, inspector).  @precisionmoon - you originally asked for this feature - do you need to have a find but not in CTS option?

precisionmoon:
> We don't have such an option for find and the other displays (layout, inspector). @precisionmoon - you originally asked for this feature - do you need to have a find but not in CTS option?

Sorry for the delayed response.  No, we don't want an option to enable/disable search.   Let's enable search all the time.

maliberty:
You missed signing your last commit.

maliberty (approved):

Issue #5655, opened 2024-08-26
[ERROR MPL-0040] Failed on cluster 


oharboe:
13 minutes to reproduce:

untar https://drive.google.com/file/d/18n0z4_Bk9Gscy3RRCiU6FiNvghIb6zIG/view?usp=drive_link

```
$ time ./run-me-BoomTile-asap7-base.sh
OpenROAD v2.0-15340-g7ebef4425
Features included (+) or not (-): +Charts +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
HierRTLMP Flow enabled...
rtl_macro_placer -halo_width 20 -halo_height 20 -report_directory .//objects/asap7/BoomTile/base/rtlmp -target_util 0.60
Floorplan Outline: (0.0, 0.0) (2160.73, 2160.73),  Core Outline: (1.026, 1.08) (2159.73, 2159.73)
        Number of std cell instances: 1743858
        Area of std cell instances: 220985.73
        Number of macros: 72
        Area of macros: 691249.12
        Halo width: 20.00
        Halo height: 20.00
        Area of macros with halos: 1292620.62
        Area of std cell instances + Area of macros: 912234.88
        Core area: 4659886.50
        Design Utilization: 0.20
        Core Utilization: 0.06
        Manufacturing Grid: 1

[ERROR MPL-0040] Failed on cluster frontend/bpd/banked_predictors_1/btb
Error: macro_place_util.tcl, 143 MPL-0040
```

_Originally posted by @oharboe in https://github.com/The-OpenROAD-Project/megaboom/issues/97#issuecomment-2308988697_

oharboe:
@AcKoucher @maliberty Found a workaround, [tweak initial conditions](https://github.com/The-OpenROAD-Project/megaboom/pull/97/commits/26045445f91044ece61811badf1acd486225a041#r1731658195)

maliberty:
> @AcKoucher @maliberty Found a workaround, [tweak initial conditions](https://github.com/The-OpenROAD-Project/megaboom/pull/97/commits/26045445f91044ece61811badf1acd486225a041#r1731658195)

From the megaboom PR:
```
I would like to see an initial diagnosis from @AcKoucher first... but yes. I hope the problem is just some existing rare problem that presents itself with some unfortunate initial conditions and that it can be solved in due course but without urgency.
```

[...]

PR #5666, opened 2024-08-27
mpl2: avoid creating clusters for empty modules

AcKoucher:
This PR addresses one of the problems showed in #5655, that is: the clustering engine leaves the physical hierarchy with empty clusters created from empty modules:
image

Subsequent work is needed in order to address other things.

_I also made some renaming, because it was a bit confusing._

### DIFF
--- a/src/mpl2/src/clusterEngine.cpp
+++ b/src/mpl2/src/clusterEngine.cpp
@@ -1196,7 +1196,7 @@ void ClusteringEngine::breakCluster(Cluster* parent)
       if (parent == tree_->root.get()) {
         createFlatCluster(module, parent);
       } else {
-        addModuleInstsToCluster(parent, module);
+        addModuleLeafInstsToCluster(parent, module);
         parent->clearDbModules();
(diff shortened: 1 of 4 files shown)


maliberty (approved):

maliberty:
@oharboe any idea where there are empty Verilog modules?  Just curious if there is an upstream problem as well.

oharboe:
> @oharboe any idea where there are empty Verilog modules?  Just curious if there is an upstream problem as well.

No, makes no sense... A problem earlier in the flow?

oharboe:
@maliberty Aha! I used deltaDebug.py to reduce turnaround time from 1 hr to 13 mins, so yes, almost all modules should be empty....

maliberty:
Ok so a side issue

oharboe:
> Ok so a side issue

Possibly a red herring: deltadebug can gut modules so they are empty while still reproducing the mpl2 problem...

maliberty:
Its still good to fix up these little corner cases

PR #5674, opened 2024-08-30
cts: delete leaked iterators

AcKoucher:
As I was studying the code to see how to avoid using clock buffers during post-placement optimization I realized that the Lib iterators are not being freed after used in CTS.


maliberty (approved):

precisionmoon:
Good catch!

PR #5676, opened 2024-08-31
RSZ: remove max_steiner_pin_count_ limit on pins per net handled by rsz

osamahammad21:
RSZ skips nets with more pins than 200000. This PR removes this limit.


maliberty (approved):

maliberty:
@jeffng-or please test on megaboom

jeffng-or:
Running route step locally using PR5676 PR. Will let you know when it finishes.

PR #5681, opened 2024-09-03
etc: Add option to disable certificate checks

titan73:
Every now and then I get issue with certificates so add an option to workaround it.
"eval" is added before wget as alias are expanded when reading the commands and not executed.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -94,7 +94,7 @@ _installCommonDev() {
     cmakeBin=${cmakePrefix}/bin/cmake
     if [[ ! -f ${cmakeBin} || -z $(${cmakeBin} --version | grep ${cmakeVersionBig}) ]]; then
         cd "${baseDir}"
-        wget https://cmake.org/files/v${cmakeVersionBig}/cmake-${cmakeVersionSmall}-${osName}-${arch}.sh
+        eval wget https://cmake.org/files/v${cmakeVersionBig}/cmake-${cmakeVersionSmall}-${osName}-${arch}.sh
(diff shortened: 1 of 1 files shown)


maliberty (approved):

rovinski:
Seems rather dangerous from a security perspective...

maliberty:
> Seems rather dangerous from a security perspective...

Its only for use by user opt-in.  OR will never use it by default or in the CI.

rovinski:
Yeah, I mean even from a user standpoint. If someone is having this problem then they should figure out the issue themselves. They could be victim of a MITM attack.

maliberty:
@titan73 can you describe the problem further?

titan73:
An example of error:
```
--2024-09-08 12:08:01--  https://cmake.org/files/v3.24/cmake-3.24.2-linux-x86_64.sh
Resolving cmake.org (cmake.org)... 66.194.253.25
Connecting to cmake.org (cmake.org)|66.194.253.25|:443... connected.
ERROR: cannot verify cmake.org's certificate, issued by ‘CN=Zscaler Intermediate Root CA (zscaler.net) (t)\\ ,OU=Zscaler Inc.,O=Zscaler Inc.,ST=California,C=US’:
  Unable to locally verify the issuer's authority.
To connect to cmake.org insecurely, use `--no-check-certificate'.
```

Zscaler is the firewall which is complaining. This firewall causes a lot of troubles. Nothing to do with any osrt of attacks.

rovinski:
So it is an MITM attack... just one that you intend. There should be a warning attached to this option so that users know that under normal circumstances they should not use it.

PR #5682, opened 2024-09-03
DependencyInstaller: if docker is present, do not reinstall

oharboe:
The user can uninstall to have it reinstalled. The thinking is that reinstalling on a docker can cause more problems than it solves, so better to do nothing until the user explicitly uninstalls.

Also, this allows the user to choose podman-docker, if that is his preference.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -642,6 +642,14 @@ _installCI() {
         parallel \
         software-properties-common
 
+    if command -v docker &> /dev/null; then
+        # The user can uninstall docker if they want to reinstall it,
+        # and also this allows the user to choose drop in replacements
+        # for docker, such as podman-docker
+        echo "Docker is already installed, skip docker reinstall."
+        return 0
+    fi
+
     # Add Docker's official GPG key:
     install -m 0755 -d /etc/apt/keyrings
     curl -fsSL https://download.docker.com/linux/ubuntu/gpg \


jeffng-or (commented):
repladements -> replacements

Otherwise, I think it's OK since it's limited to the CI configuration method. In the normal developer mode, docker would have already been installed.

Adding @vvbandeira to verify that it's OK to merge.

oharboe:
> repladements -> replacements

Fixed.
 
> Otherwise, I think it's OK since it's limited to the CI configuration method. In the normal developer mode, docker would have already been installed.

I'm unsure what you mean when you say "already have been installed"?

There are no other DependencyInstaller.sh scripts in ORFS that install docker, you mean that the develper would/could have installed it for other purposes?

jeffng-or:
> > Otherwise, I think it's OK since it's limited to the CI configuration method. In the normal developer mode, docker would have already been installed.
> 
> I'm unsure what you mean when you say "already have been installed"?
> 
> There are no other DependencyInstaller.sh scripts in ORFS that install docker, you mean that the develper would/could have installed it for other purposes?

The ORFS DependencyInstaller.sh installs docker and it's run before the OR DependencyInstaller.sh in setup.sh.

vvbandeira:
My concern with the change is that the apt commands after the new if-statement will never be run after docker is installed.
A couple of scenarios would not be covered:
1. If a user already has docker installed in their system and decides to start using OR. The `buildx` plugging might not be installed already, it will not be installed by the dependency installer and will lead to errors.
2. If we add another plugin dependency, such as with `buildx`, every user must manually install that dependency.
3. This would prevent the docker installation from being updated during script re-runs. 

I am ok with the goal of the PR, but I am worried about side effects.

oharboe:
@maliberty @jeffng-or @vvbandeira Reinstalling on top of an existing docker installation without questions is introducing more problems than it solves. Better that the user uninstalls docker himself to retrigger a re-install. The user is less likely to get his system messed up in this way.

oharboe:
@maliberty @vvbandeira @jeffng-or this is a copy and paste of [merged](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2319)

maliberty:
> @maliberty @vvbandeira @jeffng-or this is a copy and paste of [merged](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2319)

so close?

oharboe:
> > @maliberty @vvbandeira @jeffng-or this is a copy and paste of [merged](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2319)
> 
> so close?

No, it needs to be fixed in both repositories...

There's copy & paste code between DependencyInstaller in OR and ORFS, but that is a concern beyond this PR...

maliberty (approved):

PR #5683, opened 2024-09-03
test: add post-DRT repair_antennas to regression tests

eder-matheus:
Besides adding post-DRT repair_antennas, add a metric regarding the total number of diodes inserted by repair_antennas.


test/tinyRocket_nangate45.metrics
```diff
@@ -3,67 +3,69 @@
...
+	"route__wirelength": 598385,
+	"route__vias": 181888,
+	"route__vias__singlecut": 181888,
 	"route__vias__multicut": 0,
 	"DRT::drv": "0",
 	"antenna__violating__nets": 0,
 	"antenna__violating__pins": 0,
+	"antenna__violating__nets": 0,
+	"antenna__violating__pins": 0,
 	"DRT::ANT::errors": "0",
```
maliberty, line 58: Why are there duplication metrics?

eder-matheus: I didn't notice that. I thought metrics with the same name would be overwritten when writing it multiple times, but I guess it just creates multiple entries now that we call check_antennas in the loop.

maliberty: No it just uses a list of metrics.  You could append the iteration number to make them unique.

eder-matheus: Can this be done in the flow.tcl script? I don't have access to the iteration number in the C++ implementation.

maliberty: Why do we have two iterations if the violations are solved in the first one?

eder-matheus: In this case, I believe the problem is because we're calling check_antennas in three different places:
1. After global_route
2. During the iterations to repair_antennas post-DRT
3. After the iterations, for the final report.

In the first case, we call it to generate the report for post-GRT violations.

The third case might be redundant in cases where all the violations are solved in less than 5 iterations, but it is necessary for the cases where some iterations remains.

maliberty: Perhaps can be used utl::set_metrics_stage to solve this?

eder-matheus: Thanks! I will give it a try.

maliberty (approved):

PR #5684, opened 2024-09-03
odb: Convert old error message to logger.

titan73:
I'm marking this PR as draft as generally in the build, duplicate message numbers generate an error.
In this case, there are 2 errors with number 250 and  I don't get any an error. Something is wrong with the build.

Can someone take a look at it?

  if (_mode != defin::DEFAULT) {
    if (chip == nullptr) {
      _logger->error(utl::ODB, **250**, "Chip does not exist");
    }
  } else if (chip != nullptr) {
    _logger->error(utl::ODB, **250**, "Chip already exists");
  } else {
    chip = dbChip::create(_db);
  }

### DIFF
--- a/src/odb/src/defin/definReader.cpp
+++ b/src/odb/src/defin/definReader.cpp
@@ -1890,8 +1890,7 @@ dbChip* definReader::createChip(std::vector<dbLib*>& libs,
       _logger->error(utl::ODB, 250, "Chip does not exist");
     }
   } else if (chip != nullptr) {
-    fprintf(stderr, "Error: Chip already exists\n");
-    return nullptr;
+    _logger->error(utl::ODB, 251, "Chip already exists");
   } else {
     chip = dbChip::create(_db);
   }


maliberty:
I see it failing in this PR: https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-5684-head/2/pipeline-console/?selected-node=33

src/odb/src/defin/definReader.cpp
```diff
@@ -1890,8 +1890,7 @@ dbChip* definReader::createChip(std::vector<dbLib*>& libs,
       _logger->error(utl::ODB, 250, "Chip does not exist");
     }
   } else if (chip != nullptr) {
-    fprintf(stderr, "Error: Chip already exists\n");
-    return nullptr;
+    _logger->error(utl::ODB, 250, "Chip already exists");
```
eder-matheus, line 1893: @titan73 If you look at line 1890, you'll see that the message ID 250 is already used. You need to set the ID to a different value. 251 should work, as mentioned in the build error message.

titan73:
I can't see it locally. Don't know why.
I updated the number to 251 as mentioned in the CI.
I'll take a look at it. Where is the script that check the IDs?

titan73:
Ah here:
etc/find_messages.py:                "Error: {} used {} times, next free message id is {}".format

I'll take a look.

maliberty:
When I tried it locally I observed the problem.   https://github.com/The-OpenROAD-Project/OpenROAD/pull/5685 should fix it

titan73:
find_message.py catches the duplicate ids but it is not called. Touching src/odb/src/db/CMakeFiles/db.dir/build.make makes find_message.py to be called and trigger the errors.
I've tested your fix and  it is not better. I'd say it's worst as it does not fix the issue and touch src/odb/src/db/CMakeFiles/db.dir/build.make does not work anymore.

I've removed the DRAFT prefix so you can merge the PR. I'll file an issue for this particular bug.

titan73:
Ok prefix removed. for merging.

maliberty:
The trigger is in 'db' and if you are editing 'defin' or another module then it won't trigger.  It would be nice to trigger on any part of odb.

maliberty (approved):

PR #5691, opened 2024-09-04
ant: avoid inserting diodes for drivers

eder-matheus:


eder-matheus:
@maliberty I've started a CI with the public designs to check the impact of it in the final violation count. I expect no changes in it, but lets see how it goes.

maliberty (approved):

eder-matheus:
@maliberty I've updated this PR to return only the pin with violation in each ant::Violation object. This way, drivers will not have diodes near them.
The crash we had was due to a clear in a std::map in multithreading. I moved this clear to a better place now.

src/ant/src/AntennaChecker.cc
```diff
@@ -772,7 +774,12 @@ int AntennaChecker::checkGates(odb::dbNet* db_net,
   for (const auto& [node, layer_to_node] : gate_info) {
     bool pin_has_violation = false;
 
-    std::string pin_name = fmt::format("  Pin: {}", node);
+    odb::dbMTerm* mterm = node->getMTerm();
+    std::string node_name = fmt::format("  {}/{} ({})",
+                                        node->getInst()->getConstName(),
+                                        mterm->getConstName(),
+                                        mterm->getMaster()->getConstName());
+    std::string pin_name = fmt::format("  Pin: {}", node_name);
```
maliberty, line 782: Why not combine these into "  Pin:   {}/{} ({})" ...

maliberty (approved):
Does this need a secure CI?

eder-matheus:
> Does this need a secure CI?

CI is passing.

PR #5693, opened 2024-09-04
rsz: use only data buffers

AcKoucher:
For #5577.

### DIFF
--- a/src/rsz/src/Resizer.cc
+++ b/src/rsz/src/Resizer.cc
@@ -58,6 +58,7 @@
 #include "sta/Liberty.hh"
 #include "sta/Network.hh"
 #include "sta/Parasitics.hh"
+#include "sta/PatternMatch.hh"
 #include "sta/PortDirection.hh"
 #include "sta/Sdc.hh"
 #include "sta/Search.hh"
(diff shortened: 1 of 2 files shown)


src/rsz/src/Resizer.cc
```diff
@@ -497,18 +498,31 @@ void Resizer::balanceRowUsage()
...
 
-void Resizer::findBuffers()
+void Resizer::findDataBuffers()
 {
   if (buffer_cells_.empty()) {
     LibertyLibraryIterator* lib_iter = network_->libertyLibraryIterator();
+    sta::PatternMatch clkbuf_pattern(".*CLKBUF.*",
+                                     /* is_regexp */ true,
+                                     /* nocase */ true,
+                                     /* Tcl_interp* */ nullptr);
```
QuantamHD, line 508: It feels like we should factor this out into a separate function. There's lots and lots of different names for these clock buffers.

In general though it feels like we should be able to identify them based on the difference between their rise and fall times. We should have one class of buffers where the delta is much bigger and one class where it's nearly zero. If we run a k-means of 2 I suspect we'll get what we want.

AcKoucher: I used as reference the strategy used in TritonCTS when we select the buffers that will be used for the clock tree.
See `TritonCTS::inferBufferList`. I opened this as a draft to propose a discussion as I don't feel that this change is close to an actual solid solution. @maliberty @precisionmoon Do you have any input?

precisionmoon: @QuantamHD, @AcKoucher, Yes it makes sense to look for buffers with similar rise/fall delays in the absence of special buffer names or attributes.   We want to do this for both CTS and RSZ engines.  For now, I suggest that we proceed with cell names and attributes and enhance it for rise/fall delays in the future.

maliberty: I looked and all the public PDKs would work with this (not all have clock buffers).  There are private PDKs that don't follow this pattern.  I agree it is ok for a first step.

maliberty: We could centralize the decision about what is a clock buffer in dbSta so it can be shared with CTS.

maliberty:
The code looks fine but this will need a secure CI.

AcKoucher:
Running secure-CI. I'm studying the failures in the regression tests.

titan73:
What is purpose of the secure CI?

maliberty:
> What is purpose of the secure CI?

To test on proprietary PDKs that PII has access to under NDA.

maliberty:
Precision Innovation Inc.

src/rsz/src/Resizer.cc
gadfort: It seems like proceeding with a half solution like the clockbuf pattern thing would be a poor idea in my mind. If the better solution is the look at rise and fall times, etc then this should wait until a centralize solution is available

Issue #5696, opened 2024-09-05
GRT: Bad GRT guides results in DRT break


osamahammad21:
### Describe the bug

https://github.com/The-OpenROAD-Project/OpenROAD/issues/5637 triggered an issue in DRT that originates in GRT guides. GRT chooses to create a 2-gcells route segment on Metal1 for net clknet_4_6_0_clk to access pin clkbuf_4_6_0_clk/Z. It makes more sense if GRT chooses one gcell to access the pin. The available tracks on Metal1 should be 0 in that area.
![Screenshot from 2024-09-05 19-55-03](https://github.com/user-attachments/assets/1de841b9-2e0f-4a89-97b3-5ba28015435d)

### Expected Behavior

GRT accesses the pin using 1 gcell only on Metal1

### Environment

```shell
refer to https://github.com/The-OpenROAD-Project/OpenROAD/issues/5637
```

### To Reproduce

refer to https://github.com/The-OpenROAD-Project/OpenROAD/issues/5637

eder-matheus:
You are right. Even if we allow Metal1 routing (as in this design), all tracks are completely obstructed by the instance pins and obstructions. I will see what's wrong in GRT. Probably I'm not iterating over a list of obstructions.
<img width="589" alt="image" src="https://github.com/user-attachments/assets/a1882575-4cd7-4e9e-99ae-022fc1df3567">

maliberty:
My guess is that since the AP is on the border that confuses grt as well about which gcell the ap is in.

eder-matheus:
> My guess is that since the AP is on the border that confuses grt as well about which gcell the ap is in.

I think the problem is simpler than that. GRT understands that the pin is on the gcell shown in the picture. However, when I look at the resources available at Metal1 in this gcell, it shows much more than I would expect (9 tracks of 15, while I would expect at most 3 tracks available). Since ew allow Metal1 routing in this design, it decided to create the segment in Metal1.

I need to understand why it is not computing the resources properly.

maliberty:
I would expect 0 - why 3?

osamahammad21:
@eder-matheus Any update on this?

eder-matheus:
> @eder-matheus Any update on this?

I'm trying to solve congestion issues in 3 designs of the public CI. I believe we are having problem with pin access in GRT that is causing the congestion, because the values are large.

PR #5703, opened 2024-09-06
grt: fix handle of obstructions in single gcells

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5696.

Secure-ci is running.


src/grt/src/GlobalRouter.cpp
```diff
@@ -1610,13 +1610,28 @@ void GlobalRouter::applyObstructionAdjustment(const odb::Rect& obstruction,
...
 
+  int grid_limit = vertical ? grid_->getYGrids() : grid_->getXGrids();
   if (!vertical) {
+    // if obstruction is inside a single gcell, block the edge between current
+    // gcell and the adjacent gcell
+    if (first_tile.getX() == last_tile.getX()
+        && last_tile.getX() + 1 < grid_limit) {
+      int last_tile_x = last_tile.getX() + 1;
+      last_tile.setX(last_tile_x);
+    }
```
osamahammad21, line 1621: I think generally not just for first_tile.getX() == last_tile.getX():
```
last_tile.setX(last_tile.getX() + 1);
first_tile.setX(first_tile.getX() - 1);
```

eder-matheus: I've tried this approach, but it is too agressive. A bunch of designs failed with high congestion. We are probably blocking pin access for macro pins and instance pins that are above layer 1.

My understanding is that blocking to the right/top is enought because the way FastRoute algorithm works, creating the edges in these two directions. Also, the test case used to validate this update is showing the correct congestion in layer 1 and we don't have guides in this layer anymore.

maliberty:
@eder-matheus what's the status of this PR?

eder-matheus:
> @eder-matheus what's the status of this PR?

I have congestion failures in a few public designs that I need to investigate. I didn't prioritized it yet because I was looking at repair antennas and the new yosys version, but I'll got back to it soon.

src/grt/src/GlobalRouter.cpp
```diff
@@ -689,7 +689,6 @@ void GlobalRouter::updateDirtyNets(std::vector<Net*>& dirty_nets)
       dirty_nets.push_back(db_net_map_[db_net]);
       routes_[db_net].clear();
       db_net->clearGuides();
-      fastroute_->clearNetRoute(db_net);
```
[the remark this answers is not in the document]
eder-matheus: I removed it because when we add the net to the dirty list, the route of it will be removed later when we add it to the FastRouteCore. Tho I believe I'll restore it, just to ensure the resources will be released.

maliberty (approved):

eder-matheus:
@maliberty Thanks for approving it. Before the merge, I have to update some metrics in ORFS, so it will take a bit of time.

PR #5704, opened 2024-09-08
odb: Move messages checking to odb to fix deps.

titan73:
The messages checking was set at db level as odb is an interface and doesn't not have a post-build trigger.
Unfortunately there were no dependencies on other libs besides db such as defin.
So since add_dependencies can only add targets but not files, create a custom target which depends on a generated files which
triggers the call to find_messages.py.

### DIFF
--- a/src/odb/CMakeLists.txt
+++ b/src/odb/CMakeLists.txt
@@ -81,3 +81,22 @@ if(ENABLE_TESTS)
     odb
   )
 endif()
+
+add_custom_command(
+  OUTPUT messages_checked
+  COMMAND ${CMAKE_SOURCE_DIR}/etc/find_messages.py > messages.txt && touch ${CMAKE_CURRENT_BINARY_DIR}/messages_checked
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  DEPENDS
+    db
+    cdl
+    defin
+    defout
+    lefin
+    lefout
+    zutil
+    gdsin
+)
+
+add_custom_target(odb_messages DEPENDS messages_checked)
+
+add_dependencies(odb odb_messages)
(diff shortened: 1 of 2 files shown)


maliberty:
I assume this is to fix #5681 

add_custom_target docs say

>The target has no output file and is always considered out of date even if the commands try to create a file with the name of the target. Use the [add_custom_command()](https://cmake.org/cmake/help/latest/command/add_custom_command.html#command:add_custom_command) command to generate a file with dependencies.

I think this should only run when there is a change to a dependency.  Can this be accomplished with add_custom_command ?

titan73:
You're right. PR updated as suggested.
We still need a custom target as add_dependencies accepts only targets.
But this time the custom target depends on a file generated only when the deps have change.

src/odb/CMakeLists.txt
```diff
@@ -81,3 +81,22 @@ if(ENABLE_TESTS)
     odb
   )
 endif()
+
+add_custom_command(
+  OUTPUT messages.txt
+  COMMAND ${CMAKE_SOURCE_DIR}/etc/find_messages.py > messages.txt && touch ${CMAKE_CURRENT_BINARY_DIR}/messages.txt
```
maliberty, line 87: What is the purpose of the touch command?  That file doesn't seem to be referenced.

titan73: This file is referenced by add_custom_target(odb_messages DEPENDS messages.txt). This file is in build/src/odb and is only generated when find_messages.py succeeds. This way invoking make will rerun find_messages.py if it's missing or if a dependencies is changed.
The output of find_messages.py is in src/odb and is always generated whenever find_messages.py succeeds or not which is the current behaviour.

PR #5715, opened 2024-09-09
Grt: Removing static variables from maze3D

luis201420:
Avoiding the need for static variables in maze3D functions in order to support multithreading. 
The variables are added as a member of the class and initialized along with auxiliary variables.


maliberty (approved):

src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -93,6 +93,12 @@ struct DebugSetting
 
 using stt::Tree;
 
+struct parent3D
+{
+  short l;
```
eder-matheus, line 98: Maybe rename it to layer to avoid confusion.

PR #5726, opened 2024-09-10
DependencyInstaller: Add option for constant build path

glatosinski:
This PR adds `-constant-build-dir` flag that sets `baseDir` for dependency building to a fixed directory instead of `mktemp`-based directory.

Dependencies like Yosys and abc use `__FILE__` macro, which is different on every build.

Such changes can lead to cache invalidation in e.g. https://github.com/The-OpenROAD-Project/bazel-orfs/

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -738,6 +738,9 @@ Usage: $0
        $0 -save-deps-prefixes=FILE
                                 # Dumps OpenROAD build arguments and variables
                                 # to FILE
+       $0 -constant-build-dir
+                                # Use constant build directory, instead of
+                                #    random one.
 
 EOF
     exit "${1:-1}"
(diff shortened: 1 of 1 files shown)


etc/DependencyInstaller.sh
```diff
@@ -776,6 +779,14 @@ while [ "$#" -gt 0 ]; do
             export PREFIX="${HOME}/.local"
             export isLocal="true"
             ;;
+        -constant-build-dir)
+            baseDir="/tmp/DependencyInstaller-OpenROAD"
+            if [[ -d "$baseDir" ]]; then
+                echo "INFO: removing old building directory $baseDir"
+                rm -r "$baseDir"
+            fi
+            mkdir -p "$baseDir"
```
vvbandeira, line 788:
```suggestion
            if [[ -d "$baseDir" ]]; then
                echo "INFO: removing old building directory $baseDir"
                rm -r "$baseDir"
            fi
            baseDir="/tmp/DependencyInstaller-OpenROAD"
            mkdir -p "$baseDir"
```

vvbandeira: Would this be the correct order? First, remove the folder created by mktmp, then reassign the value and make the folder. Otherwise, the mktmp folder will still exist.

vvbandeira (changes requested):

etc/DependencyInstaller.sh
eszpotanski: Yes, I've changed the order, so now a previous `baseDir` is removed

vvbandeira (approved):

PR #5740, opened 2024-09-12
grt/ant: fix instability on repair_antennas

eder-matheus:


src/ant/src/PinType.hh
```diff
@@ -64,9 +64,9 @@ struct PinType
 class PinTypeHash
 {
  public:
-  size_t operator()(const PinType& t) const
+  size_t operator()(const PinType& t1, const PinType& t2) const
   {
-    return std::hash<std::string>{}(t.name);
+    return t1.name < t2.name;
```
maliberty, line 69: Looks like it is no longer a hash functor so the class name needs updating.

maliberty (approved):
Approved - just fix the name before merging.

PR #5741, opened 2024-09-12
Grt: Updates on Fastroute

luis201420:
Replacing whiles with `std::find` to find an element in the heap in maze2D and maze3D functions. Handling the case when the element is not found.

Avoiding initializing maze3D variables in the `FastRouteCore ::initAuxVar` function, since that function is called in each incremental GRT iteration. Now the variables are initialized in the `FastRouteCore ::setGridsAndLayers` function which is used once when FastRoute is initialized.

Adding minor modifications.


eder-matheus (approved):
Did you run a secure-ci for it?

luis201420:
> Did you run a secure-ci for it?

Yes, I ran a secure-ci and all designs passed without problems.

PR #5749, opened 2024-09-15
dft: fix lib-related crashes

donn:
Fixes two null pointer dereferencing crashes:
* One where no valid lib files have not been read prior to DFT function invocation
* One where a cell might be missing entirely from the lib files in a PDK (such as `sky130_fd_sc_hd__tapvpwrvgnd_1` in sky130)

Warnings are printed in both situations.

---

@maliberty i just incremented the numbers for the new messages- not sure if that's okay or not

### DIFF
--- a/src/dft/src/replace/ScanReplace.cpp
+++ b/src/dft/src/replace/ScanReplace.cpp
@@ -368,7 +368,8 @@ void ScanReplace::scanReplace(odb::dbBlock* block)
 
     if (from_liberty_cell == nullptr) {
       // cell doesn't exist in lib, no timing info
-      if (no_lib_warned.find(master) == no_lib_warned.end()) {
+      if (master->getType() == odb::dbMasterType::CORE
+          && no_lib_warned.find(master) == no_lib_warned.end()) {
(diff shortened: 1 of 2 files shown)


src/dft/src/replace/ScanReplace.cpp
```diff
@@ -360,15 +362,28 @@ void ScanReplace::scanReplace(odb::dbBlock* block)
...
-      continue;
-    }
-
     odb::dbMaster* master = inst->getMaster();
     sta::Cell* master_cell = db_network_->dbToSta(master);
     sta::LibertyCell* from_liberty_cell = db_network_->libertyCell(master_cell);
 
+    if (from_liberty_cell == nullptr) {
+      // cell doesn't exist in lib, no timing info
+      if (no_lib_warned.find(master) == no_lib_warned.end()) {
```
maliberty, line 371: You could filter useless warning by `master->getType() == dbMasterType::CORE` as no other types make sense for DFT swap.

maliberty (approved):

PR #5752, opened 2024-09-16
rsz: use adequate buffers for each operation

AcKoucher:
For #5577,

Replaces #5693.

Changes:

- Makes RSZ default behavior to not include clock buffers for its data sizing operations. There's one exception: As some technologies such as nangate45 don't have delay cells, until we have a better approach, it's better to consider clock buffers for _hold violation repairing_ as these buffers' delay may be slighty higher and we'll need fewer insertions.
- Keep the list of buffers generated by CTS for _clock nets repairing_ function to use it.
- Create new class in dbSta for distinguishing cell function type.

 
Note: On the previous PR the increase in the number of hold buffers inserted was leading gcd_resize rsz regression test to ERROR, with the changes here we have a subtle increase in the number of hold buffers inserted (99 --> 105), however it doesn't blow up.


src/rsz/src/Resizer.cc
```diff
@@ -501,14 +502,34 @@ void Resizer::findBuffers()
...
     LibertyLibraryIterator* lib_iter = network_->libertyLibraryIterator();
+    std::unique_ptr<sta::PatternMatch> clkbuf_pattern;
+
+    if (exclude_clock_buffers_) {
+      clkbuf_pattern
+          = std::make_unique<sta::PatternMatch>(".*CLKBUF.*",
+                                                /* is_regexp */ true,
+                                                /* nocase */ true,
+                                                /* Tcl_interp* */ nullptr);
+    }
```
QuantamHD, line 513: I really feel strongly that this should be refactored into it's own method that returns something that can be queried for function.

Something like 

```c++
CellFunction GetCellFunction(sta::LibertyCell* | dbMaster*);

CellFunction function = GetCellFunction(buffer);

if (function->isClockBuffer(buffers)) {
 ...
}
```

I think otherwise we risk spreading around this kind of adhoc code everywhere and making a mess.

QuantamHD: Honestly, I'd prefer if we didn't have this regex thing here at all, but since it seems like the consensus is the project would like it add it, I strongly thing we should do the above to limit the blast radius.

AcKoucher: @QuantamHD Within this logic would we want to keep this code in the RSZ or move it to dbSTA? It seems like it would more sense for it to stay in the latter. I don't mind making the change and it does looks cleaner to me. @maliberty Any thoughts?

QuantamHD: Making it maximally accessible in dbSta seems like a good change since it would reduce friction to using it.

maliberty: I view the regex as temporary means of determining what is a clock buffer.  We could take it from the user or some internal analysis in the future.  Encapsulating it seems fine.

QuantamHD: "Nothing more permanent than a temporary fix."

gadfort: Will there be an issue opened to track the replacement of the regex? I would worry that without some kind of tracking for this, this regex will just linger as an "okay" solution so no need to go back and fix it.

maliberty: @AcKoucher make an issue please

AcKoucher: Opened https://github.com/The-OpenROAD-Project/OpenROAD/issues/5753

AcKoucher:
@maliberty Could you take a look if the approach here seems reasonable?

src/dbSta/include/db_sta/dbSta.hh
```diff
@@ -83,6 +84,23 @@ class dbStaState : public sta::StaState
   dbSta* sta_ = nullptr;
 };
 
+enum CellFunction
+{
+  DATA,
+  CLOCK
+};
+
+class CellFunctionAnalyser
```
maliberty, line 93: Function is confusing as Liberty defines the cells function (eg Y=!A).  This doesn't apply to all cells, just buffers.  How about BufferUse ?

src/rsz/src/Resizer.cc
```diff
@@ -58,6 +58,7 @@
 #include "sta/Liberty.hh"
 #include "sta/Network.hh"
 #include "sta/Parasitics.hh"
+#include "sta/PatternMatch.hh"
```
maliberty, line 61: not needed here

maliberty (changes requested):

src/rsz/src/Resizer.cc
AcKoucher: I forgot to remove it.

AcKoucher:
@maliberty @precisionmoon gdb shows that the failures in the Cpp tests happens, because some tests need to create a dbSta object. It blows up when it tries to create the sta::PatternMacth from within the instantiation of the new BufferUseAnalyser member of dbSta. However, it looks like the segfault comes from within the Tcl library (?):

```
#0  0x00007ffff7f12c12 in Tcl_GetThreadData () from /lib/x86_64-linux-gnu/libtcl8.6.so
#1  0x00007ffff7f04147 in ?? () from /lib/x86_64-linux-gnu/libtcl8.6.so
#2  0x00007ffff7f03e2f in Tcl_GetRegExpFromObj () from /lib/x86_64-linux-gnu/libtcl8.6.so
#3  0x000055555572c8ab in sta::PatternMatch::compileRegexp (this=0x555558607630) at /home/arthur/OpenROAD-flow-scripts/tools/OpenROAD/src/sta/util/PatternMatch.cc:84
#4  0x000055555572c6ab in sta::PatternMatch::PatternMatch (this=0x555558607630, pattern=0x55555748d55d ".*CLKBUF.*", is_regexp=true, nocase=true, interp=0x0)
    at /home/arthur/OpenROAD-flow-scripts/tools/OpenROAD/src/sta/util/PatternMatch.cc:36
#5  0x0000555555af2b94 in std::make_unique<sta::PatternMatch, char const (&) [11], bool, bool, decltype(nullptr)>(char const (&) [11], bool&&, bool&&, decltype(nullptr)&&) ()
    at /usr/include/c++/11/bits/unique_ptr.h:962
#6  0x0000555555af0284 in sta::BufferUseAnalyser::BufferUseAnalyser (this=0x555558607610) at /home/arthur/OpenROAD-flow-scripts/tools/OpenROAD/src/dbSta/src/dbSta.cc:883
#7  0x0000555555af0e45 in sta::dbSta::dbSta (this=0x555558607460) at /home/arthur/OpenROAD-flow-scripts/tools/OpenROAD/src/dbSta/src/../include/db_sta/dbSta.hh:104
#8  0x0000555555aec0ff in ord::makeDbSta () at /home/arthur/OpenROAD-flow-scripts/tools/OpenROAD/src/dbSta/src/dbSta.cc:78
#9  0x00005555555e1733 in rmp::AbcTest::SetUp (this=0x7ffff7c3e010) at /home/arthur/OpenROAD-flow-scripts/tools/OpenROAD/src/rmp/test/cpp/TestAbc.cc:63
#10 0x0000555555abd263 in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#11 0x0000555555ab6511 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#12 0x0000555555a925da in testing::Test::Run() ()
#13 0x0000555555a9316f in testing::TestInfo::Run() ()
#14 0x0000555555a93aad in testing::TestSuite::Run() ()
#15 0x0000555555aa3e5d in testing::internal::UnitTestImpl::RunAllTests() ()
#16 0x0000555555abe21c in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#17 0x0000555555ab762b in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#18 0x0000555555aa247f in testing::UnitTest::Run() ()
#19 0x0000555555ace189 in RUN_ALL_TESTS() ()
#20 0x0000555555ace102 in main ()
```

Is there something that need to be manually initialized before we can access the Tcl libraries from sta?

maliberty:
Is interp_ null?

maliberty:
The problem is you are initializing in the ctor but the call sequence is
```
    sta_ = std::unique_ptr<sta::dbSta>(ord::makeDbSta());
    sta_->initVars(Tcl_CreateInterp(), db_.get(), &logger_);
```
So initVars hasn't happened yet.  I suggest dbSta contain by unique_ptr rather than by value and initialization happens in initVars.

AcKoucher:
@precisionmoon @maliberty After updating the FlowTests results and the last changes to avoid the rmp cpp tests segfaults all tests are passing.

I need to run a secure-ci and there'll be metrics to update. I'll wait for #5790 to be merged before I run it.

maliberty:
@AcKoucher 5790 was merged

AcKoucher:
There are .ok files which haven't been properly updated on master. On src/gpl/test/simple01-td-tune.ok I see:

```
Differences found at line 63.
    - _276_ NOR2_X2 + PLACED ( 41514 9431 ) N ;
    - _276_ NOR2_X2 + PLACED ( 41516 9447 ) N ;
```

I'm fixing the ones from this PR.

AcKoucher:
Running CI

maliberty (approved):

PR #5754, opened 2024-09-17
pdn/via: Fix debug print units

gatecat:
Spotted this while debugging some other arrayspacing related via issues, seems like it only affects the debug prints but avoids confusion with incorrect units in the prints.


gadfort (approved):
Lgtm

PR #5758, opened 2024-09-17
grt: allow estimate parasitics from guides stored in odb

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5719


src/odb/src/db/dbGuide.cpp
```diff
@@ -110,6 +116,7 @@ dbIStream& operator>>(dbIStream& stream, _dbGuide& obj)
   stream >> obj.net_;
   stream >> obj.box_;
   stream >> obj.layer_;
+  stream >> obj.via_layer_;
```
eder-matheus, line 119: @maliberty Should I add the check for the schema version here? I believe the codeGenerator would remove manual changes here, right?

gadfort: I think you just need to add:
```
            "schema":"db_schema_db_guide_via_layer"
```
to the json file

eder-matheus: It worked, thanks very much!

src/grt/src/GlobalRouter.cpp
```diff
@@ -1956,7 +1957,8 @@ void GlobalRouter::readGuides(const char* file_name)
       odb::Rect rect(
           stoi(tokens[0]), stoi(tokens[1]), stoi(tokens[2]), stoi(tokens[3]));
       guides[net].push_back(std::make_pair(layer->getRoutingLevel(), rect));
-      boxToGlobalRouting(rect, layer->getRoutingLevel(), routes_[net]);
+      int layer_idx = layer->getRoutingLevel();
+      boxToGlobalRouting(rect, layer_idx, layer_idx, routes_[net]);
```
maliberty, line 1961: Since the guides from a file won't have the via information to enable parasitic estimation we should give a message about that to the user.

src/grt/src/GlobalRouter.cpp
```diff
@@ -2638,63 +2647,17 @@ void GlobalRouter::connectPadPins(NetRouteMap& routes)
...
-    init_y = route.init_y;
-    final_y = route.final_y;
-  } else {
-    init_y = route.final_y;
-    final_y = route.init_y;
-  }
+  const int init_x = std::min(route.init_x, route.final_x);
+  const int final_x = std::max(route.init_x, route.final_x);
+  const int init_y = std::min(route.init_y, route.final_y);
+  const int final_y = std::max(route.init_y, route.final_y);
```
maliberty, line 2657: Use std::minmax
```
const auto [init_x, final_x] = std::minmax(route.init_x, route.final_x);
```
likewise y

maliberty (changes requested):

maliberty (approved):

maliberty:
"Lint Tcl code" needs fixing but otherwise looks good.

PR #5760, opened 2024-09-18
drt: PA name changes

bnmfw:
This PR contains only name changes to the PA module. There is no change to functionality or logic at all, this PR can be considered aesthetic. (Large diffs with no variable highlights are just git getting lost due to too many changes in a block)
### Variable Name Changes:
There are too many variable changes to list. Here are the most common changes:
- Variables now adhere to snake_case
- In some variables `ap` is now fully written as `access_point`
- Some variables had `V` to mean vertical, now the abbreviation used is `vert`
- In tome maps and such the formatting `key2Value` is now `key_to_value`

Here are all other variable names changes that do not fit the above patterns:
- `bp` → `begin_point`
- `ep` → `end_point`
- `sol` → `no_drv` (this variable clearly checks if there are no Design Rule Violations)
- `gcWorker` → `design_rule_checker` (more explicit role for the worker)
- in `prepPoint_pin_genPoints_rect_genCenter`: `cnt`, `c1`, `c2` → `candidates_on_grid`, `coordinate`, `cost`
- `isLRBound` → `is_side_bound` (LR just means left right, i think 'side' expresses it well enough
#### Variables Changed to Existing Names:
For every name changed It was checked if the new name already was used in the module. This happened in some cases, but none cause errors, since all cases fall on either:
1. `new_name` present in `FlexPA_graphics.*` file but `oldName` not
2. `new_name` present as a local variable in some `function` but `oldName` not

Here are all cases listed:
- `viaDef` → `via_def` Case 1
- `instTerm` → `inst_term` Case 1
- `layerNum` → `layer_num` Case 2 for `getPrefTrackPatterns`
- `lowerType` → `lower_type` Case 1 and 2 for `prepPoint_pin_checkPoint_via`
- `upperType` → `upper_type` Case 1 and 2 for `prepPoint_pin_checkPoint_via`
- `instRows` → `inst_rows` Case 2 for `prepPattern`

### DIFF
--- a/src/drt/src/pa/FlexPA_graphics.cpp
+++ b/src/drt/src/pa/FlexPA_graphics.cpp
@@ -94,11 +94,11 @@ FlexPAGraphics::FlexPAGraphics(frDebugSettings* settings,
 
 void FlexPAGraphics::drawLayer(odb::dbTechLayer* layer, gui::Painter& painter)
 {
-  frLayerNum layerNum;
+  frLayerNum layer_num;
   if (!shapes_.empty()) {
-    layerNum = layer_map_.at(layer->getNumber()).first;
+    layer_num = layer_map_.at(layer->getNumber()).first;
     for (auto& b : shapes_) {
-      if (b.second != layerNum) {
+      if (b.second != layer_num) {
         continue;
       }
       painter.drawRect(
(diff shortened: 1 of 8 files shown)


src/drt/src/pa/FlexPA_unique.h
```diff
@@ -96,18 +96,18 @@ class UniqueInsts
...
+  std::map<frInst*, int, frBlockObjectComp> unique_to_pa_idx_;
   // Maps a unique instance to its index in unique_
-  std::map<frInst*, int, frBlockObjectComp> unique2Idx_;
+  std::map<frInst*, int, frBlockObjectComp> unique_to_idx_;
   // master orient track-offset to instances
   std::map<frMaster*,
            std::map<dbOrientType, std::map<std::vector<frCoord>, InstSet>>,
            frBlockObjectComp>
-      masterOT2Insts_;
+      master_OT_to_insts_;
```
maliberty, line 110: What is OT?

src/drt/src/pa/FlexPA_graphics.cpp
```diff
@@ -110,50 +110,56 @@ void FlexPAGraphics::drawLayer(odb::dbTechLayer* layer, gui::Painter& painter)
...
-  if (layerNum < 0) {
+  layer_num = layer_map_.at(layer->getNumber()).first;
+  if (layer_num < 0) {
     return;
   }
 
   for (auto via : pa_vias_) {
     auto* via_def = via->getViaDef();
-    Rect bbox;
+    Rect boundary_box;
```
maliberty, line 120: bbox is a common term in EDA but if you spell it out it would be bounding_box

bnmfw: Okay, revert boundary_box to bbox then

src/drt/src/pa/FlexPA_unique.h
bnmfw: I think its Orient Type (`dbOrientType`) but I'm not completely sure what its supposed to mean

maliberty: From `masterOT2Insts_[inst->getMaster()][orient][offset].insert(inst.get());`  I would guess master_orient_trackoffset_to_insts_ is intended (maybe O=orient T=track offset?).

bnmfw: Oh, ok, that makes sense

PR #5766, opened 2024-09-18
pdn: fix lef for powerswitch to align with site

gadfort:
This ensures the lef matches the row pattern and power grid pattern.


maliberty:
Is this a problem in your unit test or in the skywater cells?

PR #5744, opened 2024-09-13
drt: PA minor refactoring

bnmfw:
[Context] This used to be a huge, overwhelming PR, Its contents were splits in PRs #5760 #5770 #5778 #5783 and #5788. As those were merged this PR was updated to not included their changes ('drt: master merge' commits). What remains can be reviewed as a single PR.

This PR contains minor refactoring changes do the PA module
### Straight Forward Refactoring Decisions:
- some variables still needed snake_case.
- some spaces were added
- some comments were changed
- Doxygen was added to some functions missing it
- `genAPEnclosedBoundary` was moved to be closer to `genAPCentered` and `genAPOnTrack` which are very similar functions
- in `mergePinShapes` and auxilary variable `tech` was created to replace `getDesign()->getTech()` which was called some times
### Other Refactoring Decisions:
Some minor parts of code were changed for a less verbose equivalent, here are they:
#### Initialize a vector with set size:
```
std::vector<gtl::polygon_90_set_data<frCoord>> pin_shapes;
...
pin_shapes.clear();
pin_shapes.resize(getDesign()->getTech()->getLayers().size());
```
changed to
```
std::vector<gtl::polygon_90_set_data<frCoord>> pin_shapes(num_layers);
```
#### Using find directly in if statement:
```
auto it = coords.find(coord);
if (it == coords.end()) {
```
changed to
```
if (coords.find(coord) == coords.end()) {
```
#### Cutting unnecessary if statement
  ```
if (valid_via_defs.empty()) {
    ap->setAccess(dir, false);
  } else {
    ap->setAccess(dir, true);
  }
```
changed to
```
ap->setAccess(dir, !valid_via_defs.empty());
```

PR #5770, opened 2024-09-18
drt: function name changes and doxygen documentation

bnmfw:
This PR contains function name changes and doxygen documentation and some variable name changes. This PR does not change functionality and can be considered aesthetic.
### Function Name Changes:
Functions with the prefix `gen_`and `check_` are temporary names and should be removed after class splitting. Some functions  have `gen` and `check` without `_`, this also identifies grouping but are final names as said functions actually generate or check for something.
- `prepPoint` → `initAllAccessPoints` 
- `prepPoint_pin` → `initPinAccess`
- `prepPoint_pin_mergePinShapes` → `mergePinShapes`
- `prepPoint_pin_genPoints` → `genAPsFromPinShapes`
- `prepPoint_pin_genPoints_layerShapes` → `genAPsFromLayerShapes`
- `prepPoint_pin_genPoints_rect` → `genAPsFromRect`
- `prepPoint_pin_genPoints_rect_genGrid` → `genAPOnTrack`
- `prepPoint_pin_genPoints_rect_genCenter` → `genAPCentered`
- `prepPoint_pin_genPoints_rect_genEnc`→ `genAPEnclosedBoundary`
- `prepPoint_pin_genPoints_rect_ap` → `gen_initializeAccessPoints`
- `prepPoint_pin_genPoints_rect_ap_helper` → `gen_createAccessPoint`
- `prepPoint_pin_checkPoints` → `check_setAPsAccesses`
- `prepPoint_pin_checkPoint` → `check_addAccess`
- `prepPoint_pin_checkPoint_planar` → `check_addPlanarAccess`
- `prepPoint_pin_checkPoint_planar_ep` → `check_endPointIsOutside`
    - This function has two responsibilities and hereafter will be split into `genEndPoint` and `isPointOutsideShapes`  
- `prepPoint_pin_checkPoint_via` → `check_addViaAccess`
- `prepPoint_pin_checkPoint_via_helper` → `checkViaAccess`
- `prepPoint_pin_checkPoint_viaDir_helper` → `checkDirectionalViaAccess`
- `prepPoint_pin_updateStat` → `updatePinStats`
- `prepPoint_pin_helper` → `initPinAccessCostBounded`
### Other Changes
Some variable names were missing from the last PR, here are them:
- `yCoord` → `y_coord`
- `xCoord` → `x_coord`
- `upperlayer` → `upper_layer`
- `reversedPins` → `reversed_pins`


maliberty:
FlexPA is an overly large monolith.  _genPoints_ and _checkPoint_ grouping suggest that these methods really should belong to separate classes that deal with those tasks.  

The renaming / documenting is a good first step but it does hide this relationship.  I'm ok with this renaming PR but I think you should look to the original names as you think about further decomposing FlexPA.

bnmfw:
> FlexPA is an overly large monolith. _genPoints_ and _checkPoint_ grouping suggest that these methods really should belong to separate classes that deal with those tasks.
> 
> The renaming / documenting is a good first step but it does hide this relationship. I'm ok with this renaming PR but I think you should look to the original names as you think about further decomposing FlexPA.

From what I understand there are three main different groups of functions present in `FlexPA.cpp`, this PR only tackled two of them.

#### Gen Point Group:
Functions that come after `prepPoint_pin_genPoints`. This Group only responsibility is to generate access points for each present pin. It checks all the kinds of access in order and atributes a list of access to each pin
#### Check Point Group:
Functions that come after `prepPoint_pin_checkPoint`. This Group already has the access points given and is responsible only for setting the access atributes to each one of them, e.g. setting that an access point can be reached from West.
#### Access Pattern Group:
Functions containing `*Pattern*` in their name. I still have to take a look into them. From my understanding these are the functions responsible to generating Access Patterns given valid access points for each pin.

If you are suggesting the splitting of `FlexPA_prep.cpp` (The file containing all there functions) I suggest separating them in `FlexPA_access_point.cpp` and `Flex_access_pattern.cpp` or any other names (unfortunate that both term abbreviate to AP). I think It is coherent to keep Gen Point and Check Point function together in the same file.

Also there is an argument to be made that access should be set as access points are generated, so there would not be such a big separation. In the current version all access points are generated before the first one has its accesses set.

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -74,19 +76,73 @@ class UniqueInsts
 
   frDesign* getDesign() const { return design_; }
   frTechObject* getTech() const { return design_->getTech(); }
+
+  /**
+   * @brief Checks if any terminal has a NonDefaultRule.
```
osamahammad21, line 81: terminals do not have NDRs themselves, but their nets do.

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -74,19 +76,73 @@ class UniqueInsts
...
   bool isNDRInst(frInst& inst);
   bool hasTrackPattern(frTrackPattern* tp, const Rect& box) const;
 
+  /**
+   * @brief Creates a vector of preferred track patterns.
+   *
+   * Not every track pattern is a preferred one,
+   * this function acts as filter of design_->getTopBlock()->getTrackPatterns()
+   * to only take the preferred ones.
+   *
```
osamahammad21, line 97: Preferred tracks are the tracks along the routing direction of the layer. Could you clarify that in this doc.

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -74,19 +76,73 @@ class UniqueInsts
...
+   * @return A vector of track patterns objects.
```
osamahammad21, line 98: returns void, I assume you are planning to make it return the vector directly. if so, just add a TODO comment.

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -74,19 +76,73 @@ class UniqueInsts
...
+   */
   void getPrefTrackPatterns(std::vector<frTrackPattern*>& pref_track_patterns);
   void applyPatternsFile(const char* file_path);
 
+  /**
+   * @brief Computes all unique instances data structures
+   *
+   * Proxies computeUnique, only starting the input data strcutures before.
```
osamahammad21, line 106: starting->prepping or initializing.

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -74,19 +76,73 @@ class UniqueInsts
...
+   * @param master_to_pin_layer_range Map from a master instance to layerRange.
+   * @param pref_track_patterns Vector of preffered track patterns.
+   * Fills: master_OT_to_insts_, inst_to_unique_, inst_to_class_ and
+   * unique_to_idx_.
+   */
   void computeUnique(const MasterLayerRange& master_to_pin_layer_range,
                      const std::vector<frTrackPattern*>& pref_track_patterns);
+
+  /**
+   * @brief Error Raiser for pin shape
```
osamahammad21, line 139: Raises an error if pin shape is illegal. We should try to stick with one style of commenting (starting with a verb)

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_unique.h
bnmfw: Yeah, sorry for the confusion. I am taking the parts of PR [#5744](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5744/files) and making separate PRs for each for easier reviewing. This comment assumes that the changes from PR [#5778](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5778) are already in place, that's why the comment refers to a return and does not mentions an input. As both PRs are still unmerged I will transfer these comments to that PR so its more coherent. Sorry for this oversight

maliberty:
@osamahammad21 please merge when you are satisfied.

osamahammad21 (approved):

PR #5778, opened 2024-09-19
drt: functions changed to have more explicit output

bnmfw:
This PR contain minor changes to some functions in PA. Functions changed take a reference to a container and fill said container. This PR changes it so instead it returns it.

This makes the functions role easier to understand, as its more obvious that its role is to return that structure and that it is an output, not an input.

This should not affect performance as due to **move semantics** instead of copying the returned structure its ownership is transferred to the caller.

Some functions have name changes:
- `prepPoint_pin_mergePinShapes` → `mergePinShapes` (Conform to PR [#5770](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5770))
- `initMaster2PinLayerRange` → `initMasterToPinLayerRange`

Also `LayerRange` was changed to a pair instead of a tuple, as its always a tuple of 2 elements. This does not affect its use in code as it is accessed using `auto [a, b] = LayerRange`.

### DIFF
--- a/src/drt/src/pa/FlexPA_init.cpp
+++ b/src/drt/src/pa/FlexPA_init.cpp
@@ -48,7 +48,7 @@ void FlexPA::initViaRawPriority()
     for (auto& via_def :
          design_->getTech()->getLayer(layer_num)->getViaDefs()) {
       const int cutNum = int(via_def->getCutFigs().size());
-      ViaRawPriorityTuple priority = getViaRawPriority(via_def);
+      const ViaRawPriorityTuple priority = getViaRawPriority(via_def);
       layer_num_to_via_defs_[layer_num][cutNum][priority] = via_def;
     }
   }
--- a/src/drt/src/pa/FlexPA_init.cpp
+++ b/src/drt/src/pa/FlexPA_init.cpp
@@ -48,14 +48,13 @@ void FlexPA::initViaRawPriority()
     for (auto& via_def :
          design_->getTech()->getLayer(layer_num)->getViaDefs()) {
       const int cutNum = int(via_def->getCutFigs().size());
-      ViaRawPriorityTuple priority;
-      getViaRawPriority(via_def, priority);
+      ViaRawPriorityTuple priority = getViaRawPriority(via_def);
       layer_num_to_via_defs_[layer_num][cutNum][priority] = via_def;
     }
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/drt/src/pa/FlexPA_init.cpp
```diff
@@ -48,14 +48,13 @@ void FlexPA::initViaRawPriority()
     for (auto& via_def :
          design_->getTech()->getLayer(layer_num)->getViaDefs()) {
       const int cutNum = int(via_def->getCutFigs().size());
-      ViaRawPriorityTuple priority;
-      getViaRawPriority(via_def, priority);
+      ViaRawPriorityTuple priority = getViaRawPriority(via_def);
```
maliberty, line 51: can be const now

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -113,11 +111,11 @@ void FlexPA::prepPoint_pin_mergePinShapes(
       using boost::polygon::operators::operator+=;
       pin_shapes[layer_num] += poly;
     } else {
-      logger_->error(
-          DRT, 67, "FlexPA prepPoint_pin_mergePinShapes unsupported shape.");
+      logger_->error(DRT, 67, "FlexPA mergePinShapes unsupported shape.");
       exit(1);
```
maliberty, line 115: Please remove the `exit(1)` it is deadcode (and we should never call exit).

maliberty (dismissed):

maliberty:
I'm satisfied but want to leave a chance for @osamahammad21 to comment before merging.

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -69,21 +69,21 @@ class UniqueInsts
   void setDesign(frDesign* design) { design_ = design; }
 
  private:
-  using LayerRange = std::tuple<frLayerNum, frLayerNum>;
+  using LayerRange = std::pair<frLayerNum, frLayerNum>;
   using MasterLayerRange = std::map<frMaster*, LayerRange, frBlockObjectComp>;
```
osamahammad21, line 73: Use frBlockObjectMap instead, not necessary in this PR.

osamahammad21 (approved):

osamahammad21 (approved):

PR #5780, opened 2024-09-19
rsz: correct units and escaping $ in SpefWriter

### LINKED ISSUE #5777: read_spef fails reading a file generated by estimate_parasitics -spef_file
### Describe the bug

SPEF files generated using the estimate_parastics command with the -spef_file argument can't be read by the read_spef command. The issue lies in the units written by estimate_parasitics, where it is missing the scaling factor and the unit for inductance.

### Expected Behavior

read_spef should't fail when reading a file generated by estimate_parasitics -spef_file.

### Environment

```shell
Git commit: 9136ba6993684ed613492ea3ecc091ddb13bdbb5
kernel: Linux 6.1.0-23-amd64
os: Debian GNU/Linux 12 (bookworm)
cmake version 3.24.2
```

### To Reproduce

Generate a SPEF file using estimate_parasitics -spef_file and read it using read_spef.

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

joaomai:
Fixes #5777.

Using the correct method when extracting the units from STA. STA doesn't have a unit for inductance, so I defaulted it to HENRY since its the one used by RCX. 

When testing I found out that read_spef also fails to read a file with the $ character not escaped, so I've changed the write methods from SpefWriter to escape the character when writing the name of nets and pins.


src/rsz/test/make_parasitics6_ss.spefok
```diff
@@ -8,9 +8,10 @@
 *DIVIDER /
 *DELIMITER :
 *BUS_DELIMITER []
-*T_UNIT 1 S
-*C_UNIT 1 F
-*R_UNIT 1 OHM
+*T_UNIT 1 NS
+*C_UNIT 1 FF
```
maliberty, line 12: You haven't scaled the values to match, eg
```
*CAP
1 in1:0 7.2699e-15
2 in1:1 3.6078e-15
```
is still in F not FF.

maliberty (approved):

PR #5783, opened 2024-09-20
drt: explicit_wrong_dir, no_drv and pin_term/pin_net

bnmfw:
This PR has some minor changes to `prepPoint_pin_checkPoint_planar` and `prepPoint_pin_checkPoint_viaDir_helper`.
There are essentially three changes:
#### `vert_dir`, `horz_dir` and `wrong_dir` made explicit
Throughout both functions an at least on if statement has to check whether the given dir is in the preferred direction of the layer. This check is verbose and makes it difficult to follow the part of the code where those checks exists. `wrong_dir` was created to facilitate legibility.
#### `pin_term` and `pin_net` variables
Multiple getters for this variables though the code make it more bloated. Accessing them once and using them further make the code smaller.
#### `no_drv`
Both `no_drv` changes are very simple. In both cases the variable is checked, if `true` something happens using `true` as an argument (either setting the variable itself or `ap->setAccess()`), if `false` the same happens using `false`. `no_drv` is now used directly. There is no loss in readability since `no_drv` is defined right before both cases. 
One could argue that `no_drv` is unnecessary, and its condition could be used in its place, but I'd argue that the `no_drv` labeling makes its meaning more clear

### DIFF
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -856,6 +856,11 @@ void FlexPA::prepPoint_pin_checkPoint_planar(
   auto layer = getDesign()->getTech()->getLayer(ap->getLayerNum());
   auto ps = std::make_unique<frPathSeg>();
   auto style = layer->getDefaultSegStyle();
+  const bool vert_dir = (dir == frDirEnum::S || dir == frDirEnum::N);
+  const bool horz_dir = (dir == frDirEnum::W || dir == frDirEnum::E);
+  const bool wrong_dir
+      = (layer->getDir() == dbTechLayerDir::HORIZONTAL && vert_dir)
+        || (layer->getDir() == dbTechLayerDir::VERTICAL && horz_dir);
   if (dir == frDirEnum::W || dir == frDirEnum::S) {
     ps->setPoints(end_point, begin_point);
     style.setEndStyle(frcTruncateEndStyle, 0);
@@ -863,14 +868,8 @@ void FlexPA::prepPoint_pin_checkPoint_planar(
     ps->setPoints(begin_point, end_point);
     style.setBeginStyle(frcTruncateEndStyle, 0);
   }
-  if (layer->getDir() == dbTechLayerDir::VERTICAL) {
-    if (dir == frDirEnum::W || dir == frDirEnum::E) {
-      style.setWidth(layer->getWrongDirWidth());
-    }
-  } else {
-    if (dir == frDirEnum::S || dir == frDirEnum::N) {
-      style.setWidth(layer->getWrongDirWidth());
-    }
+  if (wrong_dir) {
+    style.setWidth(layer->getWrongDirWidth());
   }
   ps->setLayerNum(ap->getLayerNum());
   ps->setStyle(style);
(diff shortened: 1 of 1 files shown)


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -856,21 +856,20 @@ void FlexPA::prepPoint_pin_checkPoint_planar(
   auto layer = getDesign()->getTech()->getLayer(ap->getLayerNum());
   auto ps = std::make_unique<frPathSeg>();
   auto style = layer->getDefaultSegStyle();
+  const bool vert_dir = (dir == frDirEnum::S || dir == frDirEnum::N);
+  const bool horz_dir = (dir == frDirEnum::W || dir == frDirEnum::E);
```
osamahammad21, line 860: One is sufficient and implies the other. vert_dir=!horz_dir

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -919,11 +919,8 @@ void FlexPA::prepPoint_pin_checkPoint_planar(
   design_rule_checker.main();
   design_rule_checker.end();
 
-  if (design_rule_checker.getMarkers().empty()) {
-    ap->setAccess(dir, true);
-  } else {
-    ap->setAccess(dir, false);
-  }
+  const bool no_drv = design_rule_checker.getMarkers().empty();
```
osamahammad21, line 922: When did gc_worker change to design_rule_checker? I am ok with changing the classname but the object should follow the class name untill this is done.

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -1136,17 +1133,20 @@ bool FlexPA::prepPoint_pin_checkPoint_viaDir_helper(
...
-  if (!USENONPREFTRACKS || upperlayer->isUnidirectional()) {
-    if (upperlayer->isHorizontal()
-        && (dir == frDirEnum::S || dir == frDirEnum::N)) {
-      return false;
-    }
-    if (!upperlayer->isHorizontal()
-        && (dir == frDirEnum::W || dir == frDirEnum::E)) {
+  auto upper_layer = getTech()->getLayer(via->getViaDef()->getLayer2Num());
+  const bool horz_dir = (dir == frDirEnum::W || dir == frDirEnum::E);
+  const bool vert_dir = (dir == frDirEnum::S || dir == frDirEnum::N);
```
osamahammad21, line 1138: Ditto

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_prep.cpp
bnmfw: Ok, I was worried here that because dir can assume U (up) and D (down) maybe just setting vert_dir != horz_dir was not enough. Although the function name imples **planar** access (this is kept in the function name changes) I was worried for such corner cases

src/drt/src/pa/FlexPA_prep.cpp
bnmfw: PR [#5760](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5760)
The reasoning is that in this case the name "design_rule_checker" makes it clear what this object is doing even if you are not familiar with the code base.

osamahammad21: In future changes, I prefer keeping the name of the object matching the name of the class until the class name is changed.

src/drt/src/pa/FlexPA_prep.cpp
osamahammad21: When in doubt, look for the function usage in the code. You can see that dir is never passed as U or D. So it's safe to assume that dir is always planar.

osamahammad21 (approved):

PR #5784, opened 2024-09-20
Grt: Jumper insertion for repair antenna violations

### LINKED ISSUE #5332: ANT/GRT: Break long segments that cause antenna violations
In addition to the existent diode insertion, repair_antennas needs other options to solve the antenna violations.
Breaking the long segments in the layers with antenna violations is a well-known solution during the global route and seems a good option for us.

Note: The detailed router will also need updates to follow the changes in the guides to avoid violations.

luis201420:
### Jumper insertion for repair antenna violations
The algorithm adds jumpers with size = 3GCells in the long segments with antenna violations.

- First it verifies that the layer has available layers above it with the same orientation (layer level + 2)
- The net segments are structured to be able to obtain the segments involved in each antenna violation. For this, the DSU algorithm is used.
- With the segments found, the bounding box is found and the number of pins near the min and max points of the bounding box is calculated. This will help to find the segment to insert the jumpers.
- To o find the position where a jumper should be added, the algorithm will calculate a ratio of the current PAR to the required PAR. To use the ratio it divides the total wirelength of the segments by the ratio, to get the required wirelength which will not have a violation.
- To insert the jumpers, the segments are iterated looking for the position that can contain the jumper size avoiding adding over vias. Before inserting the jumper, it is verified that the layer has the necessary resources for its jumper routing.
- When an antenna violation has pins near both sides, two jumpers will be inserted near both ends.

Fixes #5332


src/grt/src/RepairAntennas.cpp
```diff
@@ -855,4 +857,623 @@ double RepairAntennas::diffArea(odb::dbMTerm* mterm)
...
+  int best_req_space = 0;
+  bool is_reversed = false;
+  // Reverse the segments when jumper is added near the end (max)
+  if (!in_start) {
+    is_reversed = true;
+    std::reverse(segment_ids.begin(), segment_ids.end());
+  }
+  for (int pos = 0; pos < segment_ids.size(); pos++) {
+    const auto& seg = route[segment_ids[pos]];
+    // if seg could have jumper, get via position
```
eder-matheus, line 997: This comment seem out of place. I can't find in the code where you're getting the via position.

src/grt/src/RepairAntennas.cpp
```diff
@@ -855,4 +857,623 @@ double RepairAntennas::diffArea(odb::dbMTerm* mterm)
...
+    if (seg.length() >= bridge_size + 2 * tile_size) {
+      // get the jumper position via aware
+      const int req_space = searchViaAware(
+          is_reversed, is_horizontal, seg, vias_pos, bridge_size, tile_size);
+      // if is valid segment, save the segment pos in route
```
eder-matheus, line 1002: Same here. I don't understand the "save the segment pos in route".

src/grt/src/RepairAntennas.cpp
```diff
@@ -855,4 +857,623 @@ double RepairAntennas::diffArea(odb::dbMTerm* mterm)
...
+      } else {
+        // Get jumper position
+        const int bridge_init_y = seg.final_y - req_size - bridge_size;
+        const int bridge_final_y = bridge_init_y + bridge_size;
+
+        jumper_count++;
+        addJumperVertical(
+            seg_pos, route, bridge_init_y, bridge_final_y, layer_level);
+      }
+    }
```
eder-matheus, line 1164: This block of code is almost identical to the block above. Could you refactor it into a single function?

src/grt/src/RepairAntennas.cpp
```diff
@@ -855,4 +857,623 @@ double RepairAntennas::diffArea(odb::dbMTerm* mterm)
...
+  }
+  return jumper_count;
+}
+
+void RepairAntennas::getPinNumberNearEndPoint(
```
eder-matheus, line 1169: getPinNumberNearEndPoint --> getPinCountNearEndPoints

src/grt/src/RepairAntennas.cpp
```diff
@@ -855,4 +857,623 @@ double RepairAntennas::diffArea(odb::dbMTerm* mterm)
...
+                 odb::Point::manhattanDistance(gate_maxX_maxY, seg_max)));
+    if (dist_to_min < dist_to_max) {
+      info.pin_num_near_to_start_++;
+    } else {
+      info.pin_num_near_to_end_++;
+    }
+  }
+}
+
+struct SegInfo
```
eder-matheus, line 1219: Why do you have a struct definition here? Should this be moved to the header?

eder-matheus (changes requested):
This is the first batch of reviews. I'll wait for the changes before finishing the review of the last functions.

src/grt/src/RepairAntennas.h
```diff
@@ -80,6 +80,13 @@ struct RoutePtPins
 };
 
 using RoutePtPinsMap = std::map<RoutePt, RoutePtPins>;
+using SegmentIdByViolation = std::unordered_map<int, std::vector<int>>;
+
+struct ViolationInfo
```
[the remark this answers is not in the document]
luis201420: Yes, the members refer to the pin count near the start and the end of the segments

src/grt/src/RepairAntennas.cpp
```diff
@@ -855,4 +857,623 @@ double RepairAntennas::diffArea(odb::dbMTerm* mterm)
...
+      init_x, init_y, layer_level + 2, final_x, final_y, layer_level + 2));
+  // Reduce resources in layer level
+  grouter_->updateReources(init_x, init_y, final_x, final_y, layer_level, -1);
+  // Increase resources in layer level + 2
+  grouter_->updateReources(
+      init_x, init_y, final_x, final_y, layer_level + 2, 1);
+}
+
+// validate position when iterating from min to max position and vice versa
+bool validatePos(bool is_reversed,
```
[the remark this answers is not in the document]
luis201420: Yes, this function helps me verify that the position is within the segment. It helps me validate the position in the while loop.

src/grt/src/RepairAntennas.h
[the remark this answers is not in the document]
eder-matheus: Please rename the struct name to be more related to its members. `ViolationInfo` says nothing about it.

eder-matheus (changes requested):

eder-matheus (approved):

src/grt/src/RepairAntennas.cpp
```diff
@@ -1208,14 +1208,10 @@ void RepairAntennas::getPinCountNearEndPoints(
...
-    const odb::Point gate_minX_maxY
-        = odb::Point(sink_bbox.xMin(), sink_bbox.yMax());
-    const odb::Point gate_maxX_minY
-        = odb::Point(sink_bbox.xMax(), sink_bbox.yMin());
-    const odb::Point gate_maxX_maxY
-        = odb::Point(sink_bbox.xMax(), sink_bbox.yMax());
+    const odb::Point gate_minX_minY(sink_bbox.xMin(), sink_bbox.yMin());
+    const odb::Point gate_minX_maxY(sink_bbox.xMin(), sink_bbox.yMax());
+    const odb::Point gate_maxX_minY(sink_bbox.xMax(), sink_bbox.yMin());
+    const odb::Point gate_maxX_maxY(sink_bbox.xMax(), sink_bbox.yMax());
```
rovinski, line 1214: `Rect.ll()`, `Rect.ur()`, etc?

PR #5786, opened 2024-09-21
pad: add initial autoplacement capability for io pads

gadfort:
Adds:
- `place_pads` to place a selected set of pads into an IO ROW in order
- if bump are not available the pads are placed uniformly
- if bumps are present and ports placed, the pads will be placed to attempt to minimize the rdl routing lengths (supply nets are excluded from this)

Possible future improvements:
- allow for pads to be reordered, this requires proper handling of power domains and ensuring that power pads are placed at the proper intervals to support the pads.


src/pad/src/ICeWall.cpp
```diff
@@ -589,6 +589,320 @@ void ICeWall::placePad(odb::dbMaster* master,
   placeInstance(row, snapToRowSite(row, location), inst, orient.getOrient());
 }
 
+void ICeWall::placePads(const std::vector<odb::dbInst*>& insts, odb::dbRow* row)
```
maliberty, line 592: This is overly long.  I think the lambdas here don't improve the readability and should be broken out.

maliberty (changes requested):

src/pad/README.md
```diff
@@ -204,7 +204,31 @@ Example usage:
 place_corners sky130_fd_io__corner_bus_overlay
 ```
 
-### Placing Pads
+### Placing Pads Automatically
+
+To place a set pads into the pad ring. If bumps have been placed and ports placed, the placer will attempt to align the pads with the bumps, while still preserving the order of the pads.
```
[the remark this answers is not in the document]
gadfort: Updated

src/pad/src/ICeWall.cpp
gadfort: removed the lambdas and split the code into 2 functions, `placePadsUniform` (very short, but to be consistent) and `placePadsBumpAligned`

maliberty:
Would you summarize the method for the bumps case?  I'm finding the code a bit obscure.

gadfort:
@maliberty I tried breakout out a few more blocks of code + added some comments:
Summary:
1. determine if an a pad is part of a group of pads connected to bumps which are in the same row or column (depending on the edge of the row).
2. build the proper locations for each pad instance in that group (or assign a dummy location to pads not part of any group).
3. place the pads as close as possible to the computed locations
4. check if flipping the pad helps with wirelengths (this really only impacts pads with multiple bump connections such as differential pair etc).

maliberty (approved):

PR #5788, opened 2024-09-23
drt: cardinal directions and allow via simplification

bnmfw:
This PR simply adds the use of `frDirEnumPlanar` when appropriate.
It also simplifies
```
if (allow_via) {
    ap->setAccess(frDirEnum::U, true);
  } else {
    ap->setAccess(frDirEnum::U, false);
  }
```
to
```
ap->setAccess(frDirEnum::U, allow_via);
```

### DIFF
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -198,10 +198,9 @@ void FlexPA::prepPoint_pin_genPoints_rect_ap_helper(
   auto ap = std::make_unique<frAccessPoint>(fpt, layer_num);
   if (allow_planar) {
     const auto lower_layer = getDesign()->getTech()->getLayer(layer_num);
-    ap->setAccess(frDirEnum::W, true);
-    ap->setAccess(frDirEnum::E, true);
-    ap->setAccess(frDirEnum::S, true);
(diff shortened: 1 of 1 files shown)


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -1113,14 +1108,12 @@ bool FlexPA::prepPoint_pin_checkPoint_via_helper(
...
-             ap, via, pin, inst_term, layer_polys, frDirEnum::W)
-         || prepPoint_pin_checkPoint_viaDir_helper(
-             ap, via, pin, inst_term, layer_polys, frDirEnum::S)
-         || prepPoint_pin_checkPoint_viaDir_helper(
-             ap, via, pin, inst_term, layer_polys, frDirEnum::N);
+  bool some_dir_is_legal = false;
+  for (const frDirEnum dir : frDirEnumPlanar) {
+    some_dir_is_legal |= prepPoint_pin_checkPoint_viaDir_helper(
+        ap, via, pin, inst_term, layer_polys, dir);
+  }
```
maliberty, line 1115: This is not equivalent in terms of efficiency as the || operator short-circuits and the loop doesn't.

maliberty (approved):

PR #5791, opened 2024-09-23
tap: avoid iterating over insts multiple times when placing tapcells

gadfort:
No functional changes.

Changes:
- over iterate over fixed instances once in when placing tapcells

On a 16mm^2 floorplan this takes the tapcell insertion from 158s to 1.9s, most of that extra time was repeatedly iterating over the number of instances.


maliberty (approved):

maliberty:
@eder-matheus FYI

PR #5796, opened 2024-09-24
drt: endPointIsOutside split

bnmfw:
This PR is waiting for PR #5770 to merge.
`endPointIsOutside` if a function that clearly has two responsibilities:
1. Generate an end point from a begin point in some direction in some layer
2. Check if this point is contained in the pin polygons on that layer.

The function is called twice:
In `prepPoint_pin_checkPoint_planar`:
```
Point end_point;
const bool is_outside = endPointIsOutside(end_point, layer_polys, begin_point, ap->getLayerNum(), dir, is_block);
```
And `prepPoint_pin_checkPoint_viaDir_helper`:
```
Point end_point;
endPointIsOutside(end_point, layer_polys, getViaDef()->getLayer2Num(), dir, is_block);
```
In both cases the caller is interested in generating an end point, but the check for the point being outside is only pertinent for the first, as the second call does not check the returned value.
Therefore the function was split in two different, one for each responsability: `genEndPoint()`and `isPointOutsideShapes()` respectively.

### DIFF
--- a/src/drt/src/pa/FlexPA.h
+++ b/src/drt/src/pa/FlexPA.h
@@ -237,13 +237,35 @@ class FlexPA
       frDirEnum dir,
       T* pin,
       frInstTerm* inst_term);
-  bool prepPoint_pin_checkPoint_planar_ep(
-      Point& end_point,
+
+  /**
+   * @brief Generates an end_point given an begin_point in the direction
+   *
+   * @param layer_polys Pin Polygons on the layer (used for a check)
+   * TODO: maybe the check can be moves to isPointOusideShapes, but not sure
+   * @param begin_point The begin reference point
+   * @param layer_num layer where the point is being created
+   * @param dir direction where the layer will be created
+   * @param is_block wether the begin_point is from a macro block
+   *
+   * @returns the generated end point
+   */
+  Point genEndPoint(
       const std::vector<gtl::polygon_90_data<frCoord>>& layer_polys,
       const Point& begin_point,
-      frLayerNum layer_num,
-      frDirEnum dir,
-      bool is_block);
+      const frLayerNum layer_num,
+      const frDirEnum dir,
+      const bool is_block);
+
+  /**
+   * @brief Checks if a point is outside the layer_polygons
+   *
+   * @return if the point is outside the pin shapes
+   */
+  bool isPointOutsideShapes(
+      Point& point,
+      const std::vector<gtl::polygon_90_data<frCoord>>& layer_polys);
+
   template <typename T>
   void prepPoint_pin_checkPoint_via(
       frAccessPoint* ap,
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -754,8 +754,7 @@ void FlexPA::prepPoint_pin_genPoints(
   }
 }
 
-bool FlexPA::prepPoint_pin_checkPoint_planar_ep(
-    Point& end_point,
+Point FlexPA::genEndPoint(
     const std::vector<gtl::polygon_90_data<frCoord>>& layer_polys,
     const Point& begin_point,
     const frLayerNum layer_num,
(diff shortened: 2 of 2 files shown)


src/drt/src/pa/FlexPA.h
```diff
@@ -371,24 +371,33 @@ class FlexPA
...
+   * @brief Generates an end_point given an begin_point in the direction
    *
-   * @param end_point the End Point to be filled
-   * @param layer_polys a vector with all the pin polygons
-   * @param begin_point the Begin Point
-   * @param layer_num the number of the layer where begin_point is
-   * @param dir the direction the End Point is from the Begin Point
-   * @param is_block if the instance is a macro block.
+   * @param layer_polys Pin Polygons on the layer (used for a check)
+   * TODO: maybe the check can be moves to isPointOusideShapes, but not sure
```
osamahammad21, line 377: I see that we now use only the first polygon in that vector. Maybe pass it directly instead of passing the vector?
Also I don't understand the TODO comment.

src/drt/src/pa/FlexPA.h
```diff
@@ -371,24 +371,33 @@ class FlexPA
...
+   * @param begin_point The begin reference point
+   * @param layer_num layer where the point is being created
+   * @param dir direction where the layer will be created
```
osamahammad21, line 380: what layer?!

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -808,17 +807,21 @@ bool FlexPA::check_endPointIsOutside(
...
       logger_->error(DRT, 70, "Unexpected direction in getPlanarEP.");
   }
-  end_point = {x, y};
-  const gtl::point_data<frCoord> pt(x, y);
-  bool outside = true;
+  return {x, y};
+}
+
+bool FlexPA::isPointOutsideShapes(
+    Point& point,
```
osamahammad21, line 814: const

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -837,9 +840,9 @@ void FlexPA::check_addPlanarAccess(
   const bool is_block
       = inst_term
         && inst_term->getInst()->getMaster()->getMasterType().isBlock();
-  Point end_point;
-  const bool is_outside = check_endPointIsOutside(
-      end_point, layer_polys, begin_point, ap->getLayerNum(), dir, is_block);
+  Point end_point
```
osamahammad21, line 843: const

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -1139,13 +1142,11 @@ bool FlexPA::checkDirectionalViaAccess(
...
       = inst_term
         && inst_term->getInst()->getMaster()->getMasterType().isBlock();
-  Point end_point;
-  check_endPointIsOutside(end_point,
-                          layer_polys,
-                          begin_point,
-                          via->getViaDef()->getLayer2Num(),
-                          dir,
-                          is_block);
+  Point end_point = genEndPoint(layer_polys,
```
osamahammad21, line 1145: ditto

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA.h
bnmfw: The functions still checks for size, so I'm not sure only the first element can be passed.
The TODO comment is about the fact that now that the functions are split the `layer_polys` arguments only exists in `genEndPoint` to raise a warning. I though to moving this warning to `isPointOusideShapes` as it actually uses `layer_polys` for something. The problem is that `isPointOusideShapes` is only used in one of the previous cases, so I don't yet know if moving to it would matter, as then when `genEndPoint` is called alone it will not do this check

osamahammad21:
Check the CI failure and merge conflicts

osamahammad21 (approved):

PR #5800, opened 2024-09-25
pad: fix make_io_site rotations

gadfort:
Changes:
- the order of operations were wrong for making the io sites causing pads to get placed in the wrong orientation.
- adds test for row orientations


src/pad/README.md
```diff
@@ -167,9 +167,9 @@ make_io_sites
 | `-vertical_site` | Name of the site for the vertical pads (north and south). |
 | `-corner_site` | Name of the site for the corner cells. |
 | `-offset` | Offset from the die edge to place the rows. |
-| `-rotation_horizontal` | Rotation to apply to the horizontal sites to ensure pads are placed correctly. The default value is `R0`. |
-| `-rotation_vertical` | Rotation to apply to the vertical sites to ensure pads are placed correctly. The default value is `R0`. |
-| `-rotation_corner` | Rotation to apply to the corner sites to ensure pads are placed correctly. The default value is `R0`. |
+| `-rotation_horizontal` | Rotation to apply to the horizontal sites to ensure pads are placed correctly. The default value is `R0` for the eastern (right) row. |
+| `-rotation_vertical` | Rotation to apply to the vertical sites to ensure pads are placed correctly. The default value is `R0` for the southern (bottom) row. |
+| `-rotation_corner` | Rotation to apply to the corner sites to ensure pads are placed correctly. The default value is `R0` for the south west (lower left) corner. |
```
maliberty, line 172: It feels odd to use right/bottom/lower-left; left/bottom/lower-left feels more consistent.

gadfort: @maliberty I agree, I can update the code to reflect that as well (I wanted to but wasn't sure if that would be worth it). it will break any `make_io_sites` calls that use the rotation_vertical flag, but luckily I think it's a very small number of users (ie. me and possibly one other I know of).

gadfort:
@maliberty I changes the "reference" edges to be bottom/left and updated that text in the README to reflect that

maliberty (approved):

PR #5802, opened 2024-09-25
drt: check for inst_term nullptr

bnmfw:
This PR patches a Coverity scan defect. `inst_term->getInst()` could be called even if `inst_term` was a `nullprt`. Now it can't.

### DIFF
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -1499,7 +1499,7 @@ int FlexPA::initPinAccess(T* pin, frInstTerm* inst_term)
     if (is_macro_cell_pin) {
       macro_cell_pin_no_ap_cnt_++;
     }
-  } else {
+  } else if (inst_term) {
     // write to pa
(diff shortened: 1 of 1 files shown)


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -1499,7 +1499,7 @@ int FlexPA::initPinAccess(T* pin, frInstTerm* inst_term)
     if (is_macro_cell_pin) {
       macro_cell_pin_no_ap_cnt_++;
     }
-  } else {
+  } else if (inst_term) {
```
eder-matheus, line 1502: What if the inst_temr is nullptr? Should it be an error?

maliberty: The comment 
```
  // inst_term aps are written back here if not early stopped
  // IO term aps are are written back in prepPoint_pin_helper and always
  // early stopped
```
suggests that it is expected to be non-null here as the IO term (bterm) are handled separately

maliberty (approved):

PR #5803, opened 2024-09-25
drt: genAPsFromRect code duplication elimination

bnmfw:
genAPsFromRect has a lot of code duplication. 
### Problem:
When generating access points for a given rectangle shape there are two relevant layer, a layer 1 and layer 2. Currently, the code verifies if layer 1 was horizontal. If true it made a bunch of calls to generate access points. In some cases Y/vertical coordinates were used and other X/horizontal coordinates were used. If false it made the exact functions calls with the same patterns, except Y/vertical and X/horizontal were switched. This happened twice.
### Solution:
Auxiliary variables `layer1_rect_min`, `layer1_rect_max`, `layer2_rect_min`, `layer2_rect_max`, `layer1_coords` and `layer2_coords` were created. The logic to check if layer 1 is horizontal is done in them through ternary operations and they are used in posterior calls, so code does not have to be duplicated
### Others:
An auxilary `offset` variable was created to substitute the ternary `is_macro_cell_pin ? hwidth : 0` as it appeared 6 times in the code

### DIFF
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -502,148 +502,87 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
   }
 
   // gen all full/half grid coords
+  int offset = is_macro_cell_pin ? hwidth : 0;
+  int layer1_rect_min = is_layer1_horz ? gtl::yl(rect) : gtl::xl(rect);
+  int layer1_rect_max = is_layer1_horz ? gtl::yh(rect) : gtl::xh(rect);
+  int layer2_rect_min = is_layer1_horz ? gtl::xl(rect) : gtl::yl(rect);
+  int layer2_rect_max = is_layer1_horz ? gtl::xh(rect) : gtl::yh(rect);
+  std::map<frCoord, frAccessPointEnum>& layer1_coords
+      = is_layer1_horz ? y_coords : x_coords;
+  std::map<frCoord, frAccessPointEnum>& layer2_coords
+      = is_layer1_horz ? x_coords : y_coords;
+
   if (!is_macro_cell_pin || !use_center_line) {
-    if (is_layer1_horz) {
-      genAPOnTrack(y_coords, layer1_track_coords, gtl::yl(rect), gtl::yh(rect));
-      genAPOnTrack(x_coords,
+    genAPOnTrack(
+        layer1_coords, layer1_track_coords, layer1_rect_min, layer1_rect_max);
+    genAPOnTrack(layer2_coords,
+                 layer2_track_coords,
+                 layer2_rect_min + offset,
+                 layer2_rect_max - offset);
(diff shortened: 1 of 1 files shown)


bnmfw:
@joaomai this is the PR Eder just mentioned on today's meeting

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -499,148 +499,87 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
   }
 
   // gen all full/half grid coords
+  int offset = is_macro_cell_pin ? hwidth : 0;
+  int layer1_rect_min = is_layer1_horz ? gtl::yl(rect) : gtl::xl(rect);
+  int layer1_rect_max = is_layer1_horz ? gtl::yh(rect) : gtl::xh(rect);
+  int layer2_rect_min = is_layer1_horz ? gtl::xl(rect) : gtl::yl(rect);
+  int layer2_rect_max = is_layer1_horz ? gtl::xh(rect) : gtl::yh(rect);
```
maliberty, line 506: const

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -499,148 +499,87 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
...
+  std::map<frCoord, frAccessPointEnum>& layer1_coords
+      = is_layer1_horz ? y_coords : x_coords;
+  std::map<frCoord, frAccessPointEnum>& layer2_coords
+      = is_layer1_horz ? x_coords : y_coords;
```
maliberty, line 510: I think `auto& layer1_coords = is_layer1_horz ? y_coords : x_coords;` reads easier here as the type is not what's important.  Likewise layer2.

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -499,148 +499,87 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
...
-                     layer2_track_coords,
-                     gtl::yl(rect) - min_width_layer2,
-                     gtl::yl(rect),
-                     true);
-      }
-      genAPCentered(x_coords, layer_num, gtl::xl(rect), gtl::xh(rect));
-      for (auto& [x_coord, cost] : x_coords) {
-        x_coords[x_coord] = frAccessPointEnum::OnGrid;
-      }
+    lower_type = frAccessPointEnum::OnGrid;
```
maliberty, line 559: There is an odd asymmetry to this line.  It only appears in one of the four branches.  I think this PR moves it into two of the branches.  I'm not sure it belongs in any.

maliberty (changes requested):

maliberty:
Ok with me; I'll leave a chance for @osamahammad21 to comment (and merge when satisfied).

PR #5804, opened 2024-09-26
grt: restore crash fix when merging nets

eder-matheus:
Instead of removing the pin from the pin vector of a net, mark the pin as deleted and check this flag. This is better as we don't have the cost of removing the pin anymore. Also, the pin list of the dirty nets is updated later, so removing it from the list would be redundant.

The errors in the CI happened because when removing the pin, the check for the necessity of rerouting a net was wrong. A new pin could be inside the same GCell of the removed pin.


maliberty:
Deleted flags tend tend to be error prone.  Somebody will forget to check the flag.  I'd like to get a better sense of problem and if this is the best solution.

src/grt/src/GlobalRouter.cpp
```diff
@@ -3982,9 +3982,9 @@ void GlobalRouter::findBufferPinPostions(Net* net1,
                                          odb::Point& pin_pos2)
 {
   for (const Pin& pin1 : net1->getPins()) {
-    if (!pin1.isPort()) {
+    if (!pin1.isPort() && !pin1.isDeleted()) {
       for (const Pin& pin2 : net2->getPins()) {
-        if (!pin2.isPort()) {
+        if (!pin2.isPort() && !pin2.isDeleted()) {
           if (pin1.getITerm()->getInst() == pin2.getITerm()->getInst()) {
```
eder-matheus, line 3988: @maliberty The problem resides here. When we delete an instance, the grt pin is not updated and tries to access an invalid pointer.

Maybe we can set the iterm_/bterm_ to nullptr when the pin is deleted and check it here. We have few places that calls the getITerm function and it would be easy to update them.

src/grt/src/GlobalRouter.cpp
```diff
@@ -3982,9 +3982,9 @@ void GlobalRouter::findBufferPinPostions(Net* net1,
                                          odb::Point& pin_pos2)
 {
   for (const Pin& pin1 : net1->getPins()) {
-    if (!pin1.isPort()) {
+    if (!pin1.isPort() && pin1.getITerm() != nullptr) {
       for (const Pin& pin2 : net2->getPins()) {
-        if (!pin2.isPort()) {
+        if (!pin2.isPort() && pin2.getITerm() != nullptr) {
```
eder-matheus, line 3987: @maliberty In the current code, here's is the only part where the check is needed, since the other parts run before the incremental grt.

maliberty:
re "The errors in the CI happened because when removing the pin, the check for the necessity of rerouting a net was wrong. A new pin could be inside the same GCell of the removed pin."

Could we fix this check on rerouting while still deleting the pin as you originally did?

eder-matheus:
> re "The errors in the CI happened because when removing the pin, the check for the necessity of rerouting a net was wrong. A new pin could be inside the same GCell of the removed pin."
> 
> Could we fix this check on rerouting while still deleting the pin as you originally did?

I was able to achieve it by storing the old pin positions inside the Net object right before adding it to the dirty nets list. This way, we have the old positions and can delete the pin safely, without affecting the check of nets to reroute.

maliberty (approved):

maliberty:
Thanks, this seems like a cleaner solution.

PR #5809, opened 2024-09-26
 mpl2: IOs abstraction adaptation to ORFS flow - first version (simplified)

### LINKED ISSUE #4867: Improve mpl2 so that sky130hd-microwatt does not have to rely on enhanced measures
### Description

Currently, mpl2 cannot find a solution for sky130hd-microwatt without relying on using extreme values for target util. 

### Suggested Solution

1. Not relying on enhancedHierarchicalMacroPlacement function measures.
2. Being able to use the boundary penalty

### Additional Context

_No response_

### LINKED ISSUE #5669: Use IO constraints kept by PPL inside MPL2 to improve QoR
### Description

MPL2 uses the IOs' locations to structure the physical hierarchy in a particular way:
we bundle the pins into clusters which will be fixed on the boundaries that they were placed.

With this mechanism, we can benefit from:
- The connections between the other clusters and bundled IOs;
- The data flow of the nets that have IOs. 

However in ORFS, MPL2 uses the IOs' locations that come from random PPL. And so, there are a couple of problematic things:
1. These locations are random, so there's no guarantee that we're actually having some sort of benefit.
2. In many cases, bundled IOs cover all four boundaries and, due to the way they're modeled in sequence pair, SA can have a very hard time finding a good solution.

### Suggested Solution

MPL2 should not rely on fixed locations for the pins (or have a new mode for it).

Instead of having bundled IOs, we can use the constraints data that is now stored by PPL to have some idea about the "preferred" boundary of the clusters during SA.

During cluster placement, instead of having a connection with each bundled IO, each cluster would have a min_dist connection the boundaries that it has IOs connected to. We could base the weight of this connection on:

### Additional Context

This can be the solution for #4867 and [#2079](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2079)

### LINKED ISSUE #6050: routing congestion on ihp with rtlmp enabled
### Describe the bug

When enabling rtlmp over the previous macro placer, the resultant design is not able to be routed.

Relates to proposed removal of previous macro placer: https://github.com/The-OpenROAD-Project/OpenROAD/pull/6047

### Expected Behavior

Design is routable

### Environment

```shell
OpenROAD v2.0-16787-gcd519bb5e 
```

### To Reproduce

Placement:
[sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_place0_20241029-121603.tar.gz](https://github.com/user-attachments/files/17560020/sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_place0_20241029-121603.tar.gz)
```
tar xvf sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_place0_20241029-121603.tar.gz
cd sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_place0_20241029-121603
./run.sh
```
Routing:
[sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_route0_20241029-120252.tar.gz](https://github.com/user-attachments/files/17560031/sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_route0_20241029-120252.tar.gz)
```
tar xvf sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_route0_20241029-120252.tar.gz
cd sc_issue_ethmac_job0_ihp130_sg13g2_stdcell_route0_20241029-120252
./run.sh
```

### Relevant log output

_No response_

### Screenshots

place_macro placement:
image

rtlmp placement:
image

congresion markers:
image

### Additional Context

_No response_

AcKoucher:
Resolve the first top level task in #5669.
Resolve #6050.
Resolve #4867.

### Introductory Considerations
- This is the first step of the enhancements to adapt the IOs abstractions inside mpl2 to ORFS flow.
- This first version will not consider the dimensions of the constraints whether they are -exclude constraints or pins' constraint regions.

### Goals
1. Improve QoR in terms of WL.
2. Ease the life of the annealer for some cases (e.g., sky130hd/uW) that require extreme virtual std cell utilization configurations to converge.
3. Further steps into switching from manual mpl to mpl2 for cases such as sky130hd/chameleon (This will still require clustering work)

#### Modify IO Clusters
IO clusters that represent bundledIOs created from the pins placed by random PPL are now groups of unplaced IOs with the same constraints based on the following:
- Individual Pin Constraints:
  - If an IO has a constraint region in a certain boundary, it is constrained to that entire boundary.
  - If an IO has no constraint region, it is constrained to all boundaries.
 
- Global Pins' Constraints (-exclude)
  - If a boundary has more than 70% of its extension blocked, it is considered a blocked boundary for IOs.

#### What Annealing Sees
- WL between a cluster/macro (soft/hard) M and an IO cluster IO is represented as follow:
  - If IO represents pins constrained to a certain boundary A, WL is the minimum distance between M's bundled center pin and A.
  - If IO represents unconstrained pins, WL is the minimum distance between M's bundled center pin and the closest non-blocked boundary.

**Obs: This distance only reduces cost when M is inside the outline.**

#### Needed for the Approach to Work
1. Add Cluster support to HardMacro so that SA can check if they are IO clusters' fixed terminals.
2. Make both Soft/HardMacroSACores have access to the PhisycalHierarchical data where the global constraints are now stored.
3. Make fixed terminals that represent IO clusters have the cluster data inside them so that this data is accessible inside SA for both Soft and Hard. I.e., the cluster of a fixed terminal that represents an IO cluster will be != nullptr

#### Blockages for Pin Access
Generated based on a combination of both global and individual pins' constraints:
1. If there are pins constrained to a certain boundary, we'll generate a blockage for that boundary.
2. If there are no pins constrained a boundary A but there are pins constrained to all boundaries, A won't have a pin access blockage. (Specially important so that SA has more freedom).
3. The depth of the blockage is a combination of the extension of IO clusters and std cell area (_this can definitely be further improved to consider the IOs density_)

#### Orientation Improvement
When iterating a net, compute HPWL based on:
1. If a pin has a constraint region, use its center.
2. If a pin doesn't have a constraint region, use the closest point of the closest boundary to that macro pin.

#### Debug Mode
1. Blocked boundaries are marked with a red X in the center of their extension
2. Draw distance to constraint || closest boundary only when the cluster/macro is inside the outline


src/mpl2/src/SimulatedAnnealingCore.cpp
```diff
@@ -272,10 +295,18 @@ void SimulatedAnnealingCore<T>::calWirelength()
...
 
   for (const auto& net : nets_) {
-    const float x1 = macros_[net.terminals.first].getPinX();
-    const float y1 = macros_[net.terminals.first].getPinY();
-    const float x2 = macros_[net.terminals.second].getPinX();
-    const float y2 = macros_[net.terminals.second].getPinY();
+    T& source = macros_[net.terminals.first];
+    T& target = macros_[net.terminals.second];
+
+    if (target.isIOCluster()) {
```
eder-matheus, line 301: Will `source` never be a macro?

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -914,142 +912,119 @@ void HierRTLMP::setTightPackingTilings(Cluster* macro_array)
...
+    }
+  }
+}
+
+void HierRTLMP::createPinAccessBlockage(Boundary constraint_boundary,
+                                        const float depth,
+                                        const Rect& die)
+{
+  if (constraint_boundary == L) {
+    const Rect left_io_blockage(
```
eder-matheus, line 965: Similar to another comment I made, in the future we should work on enhancing this process to consider the actual constraint region. Blocking the whole edge might be to extreme for cases where the constraint region is small.

src/mpl2/src/SimulatedAnnealingCore.cpp
AcKoucher: Afaik, no. Perhaps add a critical for that case?

eder-matheus: I don\t think it is necessary. I was just curious why you're only checking if the `target` is an IO cluster, but if source is never an IO cluster, that's fine.
(My original comment was wrong, btw)

src/mpl2/src/clusterEngine.cpp
```diff
@@ -327,165 +329,155 @@ void ClusteringEngine::setBaseThresholds()
...
+      blockage_extension = blocked_region.dx() / static_cast<float>(die.dx());
+    }
+
+    blockage_extension_map[blocked_region_boundary] += blockage_extension;
+  }
+
+  return blockage_extension_map;
+}
+
+void ClusteringEngine::setIOClusterDimensions(const odb::Rect& die,
```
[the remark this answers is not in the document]
AcKoucher: wouldn't "find" fit better for a getter? It looks to me that we're literally setting the dimension inside this method.

src/mpl2/src/hier_rtlmp.cpp
AcKoucher: I''ll open an issue for that.

src/mpl2/src/SimulatedAnnealingCore.cpp
```diff
@@ -288,6 +319,70 @@ void SimulatedAnnealingCore<T>::calWirelength()
...
+    float dist_to_top = die_hpwl;
+    if (!top_is_blocked_) {
+      dist_to_top = std::abs(y1 - die.yMax());
+    }
+
+    wirelength_
+        += net_weight
+           * std::min(
+               {dist_to_left, dist_to_right, dist_to_bottom, dist_to_top});
+  } else if (constraint_boundary == Boundary::L
```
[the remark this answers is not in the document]
AcKoucher: It's the first option and this is wrong and should be fixed.

src/mpl2/src/hier_rtlmp.cpp
AcKoucher: Opened #5816

src/mpl2/src/clusterEngine.cpp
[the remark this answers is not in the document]
eder-matheus: I was thinking on only using `set` for class member setters, but it seems a bit restrictive. I'm fine with either set or find.

maliberty:
"Lint Tcl code" needs fixing.

If a pin has been manually placed with place_pin (status=FIRM), will this code consider its location rather than its constraint?

AcKoucher:
@maliberty I edited the description of the PR/Issue to make it clear that this is a simplified version of these enhancements.
I also pushed the change regarding the blocked boundaries as discussed.

When addressing the review I'll fix the Lint Tcl.

AcKoucher:
Resolved.

maliberty:
Some unit test need fixing:
```
[2024-11-14T19:15:30.551Z] mpl2
[2024-11-14T19:15:37.996Z] io_constraints1 (tcl) *FAIL*
[2024-11-14T19:15:46.470Z] io_constraints2 (tcl) *FAIL*
```

maliberty:
It would be good to merge master and do a fresh secure CI as this PR is a bit old

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -923,142 +921,119 @@ void HierRTLMP::setTightPackingTilings(Cluster* macro_array)
...
+        die.xMin(), die.yMin(), die.xMin() + depth, die.yMax());
     boundary_to_io_blockage_[L] = left_io_blockage;
     macro_blockages_.push_back(left_io_blockage);
   }
```
maliberty, line 978: This would be more succinct as
```
Rect blockage = die;
if (constraint_boundary == L) {
  blockage.set_xhi(die.xMin() + depth);
} else if (..) {...}...

 boundary_to_io_blockage_[constraint_boundary] = blockage;
 macro_blockages_.push_back(blockage);
```

maliberty (commented):
Pending updated secure CI.

src/mpl2/src/hier_rtlmp.cpp
AcKoucher: This is mpl2::Rect so it doesn't have set_xhi. I'll add it.

AcKoucher:
I ran an updated Secure-CI in the weekend and the only concerning thing is a WL degradation in gf180/uartblocks:

gf180 uart-blocks
      [ERROR] globalroute__antenna_diodes_count fail test: 11.0 <= 10.0
      [ERROR] detailedroute__route__wirelength fail test: 23164.0 <= 19722.0
      
I'll investigate it.

AcKoucher:
The problem with uart-blocks was due to the fact that I wasn't considering the case in which a pin was not constrained to any region in the orientation improvement step. Now we compute WL with a similar technique used in SA: we get the closest point in the closest boundary and merge it's rect in the net_box.

Before:


After:


Also:
While updating the regression tests I found a bug in the pin access blockage creation which was fixed.
The same for a bug in graphics.

AcKoucher:
I'm running a new secure-CI with my latest changes and master.

eder-matheus (approved):
I don't have further comments, but I'll wait for @maliberty's final review before merging.

maliberty:
> There a couple designs that need metrics update. 
Is there a PR for this?

maliberty (approved):

AcKoucher:
> Is there a PR for this?

I opened [#2582](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2582). However, a metrics conflict emerged from another PR. I'm re-running Secure-CI to update the metrics correctly.

maliberty:
@AcKoucher what remains to merge this?

AcKoucher:
@maliberty This PR is blocked by a gpl divergence in uW (See #6235).

AcKoucher:
I resolved the conflicts here.

With the updates from master, GPL no longer diverges in uW. However, now GRT is throwing an error due to high congestion in post-drt. I'm investigating.

AcKoucher:
Updating..

I ran a new Secure-CI with a custom GRT routing layer adjustment and it solved uW's congestion problem.
However, now g12/ca53 is failing in DPL.

AcKoucher:
Settings changes and metrics update in [#2706](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2706)

PR #5812, opened 2024-09-27
Format Tcl code using `tclfmt`

nmoroze:
Now that OpenROAD's on tclint 0.4.x (as of #5805), you can use tclfmt to enforce and maintain the Tcl coding style. This PR uses tcflmt to reformat all Tcl code in the codebase (that was previously being checked by tclint), and updates CI to use tclfmt for format checks. (Note I plan for this to be the supported way going forward - the individual tclint style checks will be deprecated. I'm open to more feedback on tclfmt before that happens though, so please let me know what you think!)

The bulk of changes fix spacing before and after enclosing braces, since the `spaces-in-braces` tclint rule was previously ignored. There ended up being a lot of changes, so I'm happy to split this PR up into smaller chunks for ease of review if switching to use of the formatter is generally something you want to do.


maliberty:
Would it be less of a change to use `spaces-in-braces = true`?  sta also has spaces and I don't see it changing.

nmoroze:
The PR currently uses `spaces-in-braces = true` actually. I did an experiment and counted 418 violations that needed to be fixed for `spaces-in-braces = true`, but 767 for `spaces-in-braces = false`, so opted to go for the smaller diff.

maliberty (approved):

maliberty:
Is there a reason we exclude more than src/sta?

PR #5814, opened 2024-09-27
ppl: a group is on a single metal layer

oharboe:
Improve error message for non-obvious problem.

I have a macro that is routed by abutment and temporarily during development I need to have a LOT of pins which can't fit on a single metal layer. I therefore increase the number of metal layers via IO_LAYER_H, but it still failed because I had a `set_io_pin_constraint -group -order` constraint which forced all pins to be put on a single metal layer.

It was not obvious to me that a group can't span metal layers... I'm hoping that mentioning "metal layer" here will key the user onto the problem.

Knowing this, I was able to work around some global routing congestion errors by removing a constraint:

```
 foreach {direction direction2 names} $assignments {
     set mirrored [zip {*}$names]
     set_io_pin_constraint -region $direction2:* -pin_names [lindex $names 1]
-    set_io_pin_constraint -group -order -pin_names [lindex $names 1]
+    #set_io_pin_constraint -group -order -pin_names [lindex $names 1]
     set_io_pin_constraint -mirrored_pins $mirrored
 }
```

### DIFF
--- a/src/ppl/src/IOPlacer.cpp
+++ b/src/ppl/src/IOPlacer.cpp
@@ -385,7 +385,8 @@ int IOPlacer::placeFallbackPins(bool random)
             logger_->error(PPL,
                            93,
                            "Pin group of size {} does not fit in the "
-                           "constrained region {:.2f}-{:.2f} at {} edge. "
+                           "constrained region {:.2f}-{:.2f} at {} edge on a "
+                           "single metal layer. "
                            "First pin of the group is {}.",
                            group.first.size(),
                            getBlock()->dbuToMicrons(interval.getBegin()),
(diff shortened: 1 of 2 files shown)


eder-matheus (approved):
@oharboe Thanks for updating the error message. I will work on supporting multiple layers for pin groups and constraints.

oharboe:
> @oharboe Thanks for updating the error message. I will work on supporting multiple layers for pin groups and constraints.

Thanks!

Note that the important thing for us is to have good error messages here that clue us onto the problem, the actual feature set of openroad is sufficient for our part here.

The solution in our case was simply to add some more pin layers and remove a group constraint for pins.

PR #5815, opened 2024-09-27
gui: fix pauser to stop repeated triggering

gadfort:
Changes:
- add single shot to pauser timer in GUI, this prevents it from continuing run and trigger when pausing is not needed
- avoids formatting the idle/pause label when timer is not running, but update is still called.


maliberty:
I don't understand the goal.  If a pause is requested shouldn't we always pause?

gadfort:
@maliberty if the pause is requested with a timeout (then the timer is started), since this timer is not set as singleshot, it will repeat over and over again, this ensures the timer only triggers once (per pause).
The issue that exposes this would be a pause with a timeout followed by a pause without
```
gui::pause 1000
gui::pause <- this will get unpaused by the timer that is continuing to run.
```

maliberty (approved):

maliberty:
I didn't even know we had pause with a timeout :)

PR #5817, opened 2024-09-27
deltaDebug.py: fewer red lines in editor, no change

oharboe:
no change, just fewer red lines in python editors while wandering through the code.

### DIFF
--- a/etc/deltaDebug.py
+++ b/etc/deltaDebug.py
@@ -1,21 +1,28 @@
 ################################
 # You need to provide the script with 3 main arguments
 # --base_db_path <the relative path to the db file to perform the step on>
-# --error_string <the output that indicates a target error has occured>
+# --error_string <the output that indicates a target error has occurred>
 # --step <Command used to perform a step on the base_db file>
 # You also have 1 additional argument
-# --persistence <a value in [1,6] indicating maximum granularity where maximum granularity = 2^persistence>
-# --use_stdout <a flag to either enable or disable detecting the error string from stdout in addition to stderr>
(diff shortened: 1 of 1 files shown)


gadfort:
@oharboe I think a python lint spec is needed since your "red lines" might be different from "mine" (I'm not disagreeing this PR, just observing).

maliberty (approved):

maliberty:
@gadfort do you think we should use the same as OR (https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/.github/workflows/black.yaml)?

oharboe:
> @gadfort do you think we should use the same as OR (https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/.github/workflows/black.yaml)?

I'm confused, aren't we doing that already?

I just fixed some additional line-length issues and re-ran black; slightly stricter than black for some default visual studio config.

Ideally black would just fix everything so we never have to be distracted in IDEs again :-)

maliberty:
I see so the request is for a black settings file?

gadfort:
@maliberty yeah, I didn't realize there was a formatter running since the files don't appear to be formatted

maliberty:
@gadfort what seems unformatted?  I ran black and didn't find anything.

gadfort:
@maliberty I'm mistaken, I was thinking of the test files, but those were formatted.

maliberty:
@oharboe it sounds like your setting in VS are the issue.  The files are all formatted according to black and you wouldn't see any issue with compatible settings.

PR #2305, opened 2022-09-26
Grt: improve repair antennas

luis201420:

PR #5819, opened 2024-09-27
CTS-Avoid level balance buffers overlap.

arthurjolo:
The current Level Balancer of CTS tends to insert buffers one on top of the other and when they are legalized a cluster of buffers is formed, issue [#2305](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2305) shows an example of this. This happens specially if the clock buffers are large and it creates a high congested area that can lead to the global router finishing with overflow.

This PR changes the level balancer to avoid overlapping the inserted buffers and preventing the creation of the clusters of buffers. The next images show the result for the regressions test of CTS _balance_levels_:
Current:
image

New:
image

### DIFF
--- a/src/cts/src/LevelBalancer.cpp
+++ b/src/cts/src/LevelBalancer.cpp
@@ -101,16 +101,16 @@ void LevelBalancer::addBufferLevels(TreeBuilder* builder,
   const double centroidY = totalY / cluster.size();
   int x = prevLevelSubNet->getDriver()->getX();
   int y = prevLevelSubNet->getDriver()->getY();
-  
-  int steps = std::max((int)((centroidY - y) / height), 1);
-  int buffPerStep = std::ceil((float)bufLevels / (float)steps);
+
+  int steps = std::max((int) ((centroidY - y) / height), 1);
+  int buffPerStep = std::ceil((float) bufLevels / (float) steps);
 
   for (unsigned level = 0; level < bufLevels; level++) {
     // Add buffer
-    if(level % buffPerStep) {
-      x = (x + width) ;
+    if (level % buffPerStep) {
+      x = (x + width);
     } else {
-      y = (y + height) ;
+      y = (y + height);
     }
     Point<double> bufferLoc(x / wireSegmentUnit_, y / wireSegmentUnit_);
     Point<double> legalBufferLoc
--- a/src/cts/src/LevelBalancer.cpp
+++ b/src/cts/src/LevelBalancer.cpp
@@ -94,22 +94,25 @@ void LevelBalancer::addBufferLevels(TreeBuilder* builder,
     totalX += clockInstObj->getX();
     totalY += clockInstObj->getY();
   }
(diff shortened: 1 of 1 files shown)


src/cts/src/LevelBalancer.cpp
```diff
@@ -94,22 +94,26 @@ void LevelBalancer::addBufferLevels(TreeBuilder* builder,
...
-        = (driverX
-           + (centroidX - driverX) * (double) (level + 1) / (bufLevels + 1))
-          / wireSegmentUnit_;
-    double y
-        = (driverY
-           + (centroidY - driverY) * (double) (level + 1) / (bufLevels + 1))
-          / wireSegmentUnit_;
-    Point<double> bufferLoc(x, y);
+    if (level % buffPerStep) {
+      x = (x + width);
```
maliberty, line 112: `x += width;` is more idiomatic.  (likewise y)

src/cts/src/LevelBalancer.cpp
```diff
@@ -94,22 +94,26 @@ void LevelBalancer::addBufferLevels(TreeBuilder* builder,
...
-  const int driverY = prevLevelSubNet->getDriver()->getY();
+  int x = prevLevelSubNet->getDriver()->getX();
+  int y = prevLevelSubNet->getDriver()->getY();
+
+  int dir = (centroidY - y) / std::abs(centroidY - y);
+  double height = builder->getBufferHeight() * wireSegmentUnit_ * dir;
+  double width = builder->getBufferWidth() * wireSegmentUnit_ * dir;
+
+  int steps = std::max((int) (std::abs((centroidY - y) / height)), 1);
+  int buffPerStep = std::ceil((float) bufLevels / (float) steps);
```
maliberty, line 107: const

src/cts/src/LevelBalancer.cpp
```diff
@@ -94,22 +94,26 @@ void LevelBalancer::addBufferLevels(TreeBuilder* builder,
...
   }
-  const double centroidX = totalX / cluster.size();
+
   const double centroidY = totalY / cluster.size();
-  const int driverX = prevLevelSubNet->getDriver()->getX();
-  const int driverY = prevLevelSubNet->getDriver()->getY();
+  int x = prevLevelSubNet->getDriver()->getX();
+  int y = prevLevelSubNet->getDriver()->getY();
+
+  int dir = (centroidY - y) / std::abs(centroidY - y);
```
maliberty, line 102: Why use the same direction for both x & y?  You might be moving away from the centroid in x.

maliberty (changes requested):

arthurjolo:
Secure Ci is passing, only need to update a antenna diode count metric for 1 design. I am going to create a PR for this update.

src/cts/src/LevelBalancer.cpp
arthurjolo: Yes, I missed that. It was already fixed, thanks.

arthurjolo:
Close as balance levels is now obsolete.

PR #5821, opened 2024-09-28
drt: genPattern code duplication elimination

bnmfw:
[Context] I'm starting to take a look at the Access Pattern generation in the PA module, as until now I only looked at the Access Point generation. I just started, but this code duplication case stood out like a sore thumb. I don't like naming functions "helper" but I intend to change the names of Access Patterns generation functions later so I created a temporary name.

Eliminates code duplication in `genPatterns`. Creates an auxiliary function `genPatterns_helper` to execute the duplicated code so the helper is called twice instead of repeating the code.

Git diff is making it hard to see that the code was in fact duplicated. If its hard to see I recommend looking at `genPatterns` in master to see that there is a clear repetition just changing `pins` to `reversed_pins`.

Also adds spaces between functions declarations on the header file. This will make future git diffs adding doxygen and changing names more clear.

### DIFF
--- a/src/drt/src/pa/FlexPA.h
+++ b/src/drt/src/pa/FlexPA.h
@@ -485,8 +485,8 @@ class FlexPA
       std::set<std::vector<int>>& inst_access_patterns,
       std::set<std::pair<int, int>>& used_access_points,
       std::set<std::pair<int, int>>& viol_access_points,
-      const int curr_unique_inst_idx,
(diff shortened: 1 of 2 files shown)


osamahammad21 (changes requested):
address clang-tidy requested changes

osamahammad21 (approved):

PR #5823, opened 2024-09-28
drt: genApCosted

bnmfw:
A secure CI associated with this branch passed.

The latter part of `genAPsFromRect` is a bit verbose for what it is doing. In essence, it goes through 4 access point costs (`OnGrid`, `Center`, `EncOpt` and `NearbyGrid`), checks if the layer cost limit allows for these costs and creates them with their respective functions. This takes a lot of lines and is verbose for a simple ideia. 

To encapsulate this simple ideia the function `genAPCosted` was created. It takes a cost and the necessary arguments to create any kind of access point and creates the according access point. I very much like the fact that the association between cost and function is made very clear, although it makes the use of some arguments (like `offset`) a little harder to understand.

More importantly, the use of the function makes `genAPsFromRect` smaller and much easier to understand its general flow, although it might make it harder to understand in detail.

I would like to hear opinions in this solution, although I like most parts of it and makes parts of the code more readable it can make other parts less.

### DIFF
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -527,6 +527,10 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
   }
 
   // gen all full/half grid coords
+  /** offset used to only be used after an if (!is_macro_cell_pin ||
+   * !use_center_line), so this logic was combined with offset is_macro_cell_pin
+   * ? hwidth : 0;
+   */
   const int offset = is_macro_cell_pin && !use_center_line ? hwidth : 0;
   const int layer1_rect_min = is_layer1_horz ? gtl::yl(rect) : gtl::xl(rect);
   const int layer1_rect_max = is_layer1_horz ? gtl::yh(rect) : gtl::xh(rect);
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -263,9 +263,7 @@ void FlexPA::genAPCosted(
       genAPOnTrack(coords, track_coords, rect_min + offset, rect_max - offset);
       break;
 
-    case (frAccessPointEnum::HalfGrid):
-      genAPOnTrack(coords, track_coords, rect_min + offset, rect_max - offset);
-      break;
+      // frAccessPointEnum::Halfgrid not defined
 
     case (frAccessPointEnum::Center):
       genAPCentered(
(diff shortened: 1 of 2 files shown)


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -243,6 +243,51 @@ void FlexPA::genAPEnclosedBoundary(std::map<frCoord, frAccessPointEnum>& coords,
...
+
+  switch (cost) {
+    case (frAccessPointEnum::OnGrid):
+      genAPOnTrack(coords, track_coords, rect_min + offset, rect_max - offset);
+      break;
+
+    case (frAccessPointEnum::HalfGrid):
+      genAPOnTrack(coords, track_coords, rect_min + offset, rect_max - offset);
+      break;
+
```
osamahammad21, line 269: HalfGrid is not one of expected values

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -486,81 +529,43 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
   }
 
   // gen all full/half grid coords
-  const int offset = is_macro_cell_pin ? hwidth : 0;
+  int offset = is_macro_cell_pin ? hwidth : 0;
+  offset = !is_macro_cell_pin || !use_center_line ? offset : 0;
```
osamahammad21, line 533: I think `!(is_macro_cell_pin && use_center_line) ` is simpler to read

osamahammad21: better even if switch the values and make it
`offset = (is_macro_cell_pin && use_center_line) ? 0 : offset`

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -486,81 +529,43 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
...
+                  layer2_coords,
+                  layer2_track_coords,
+                  layer_num,
+                  second_layer_num,
+                  rect,
+                  !is_layer1_horz,
+                  offset);
+    }
+  }
   if (!is_macro_cell_pin || !use_center_line) {
```
osamahammad21, line 556: ditto

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -486,81 +527,42 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
   }
 
   // gen all full/half grid coords
-  const int offset = is_macro_cell_pin ? hwidth : 0;
+  const int offset = is_macro_cell_pin && !use_center_line ? hwidth : 0;
```
osamahammad21, line 530: That's not equivalent

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_prep.cpp
bnmfw: This comes from the fact that originally `offset` only is used inside the 
```
if (!is_macro_cell_pin || !use_center_line)
```
if statement.
This allows us to reduce offset directly to:
```
offset = is_macro_cell_pin && (!is_macro_cell_pin || !use_center_line) ? hwidth : 0;
```
as the statement already requires ```is_macro_cell_pin``` to be true to even consider the second part it is equivalent to
```
offset = is_macro_cell_pin && !use_center_line ? hwidth : 0;
```

osamahammad21: I understand, Could you add a comment describing what's happening here so that this doesn't become confusing later.

PR #5825, opened 2024-09-28
gui: add to sta descriptors

gadfort:
Adds:
- sta::Corner descriptor
- sta::Instance descriptor
- adds fields to the properties in allow access to more information
- add highlighting where possible
- sta::Instance can be slow to load the first time since it needs to get power information, but I think it's worth having access to that, especially when debugging things like clock gating power @QuantamHD (the added loading time is also why I think it makes sense to keep it separate the DB descriptor).

Discussion:
- I don't like the names I've chosen for `STA Instance` and `DB ???`, so I'm open to suggestions for what to rename those to make it better

### DIFF
--- a/src/gui/include/gui/gui.h
+++ b/src/gui/include/gui/gui.h
@@ -344,7 +344,9 @@ class Descriptor
   virtual void highlight(std::any object, Painter& painter) const = 0;
   virtual bool isSlowHighlight(std::any /* object */) const { return false; }
 
-  static std::string convertUnits(double value, bool area = false);
+  static std::string convertUnits(double value,
+                                  bool area = false,
+                                  int digits = 3);
 };
 
 // An object selected in the gui.  The object is stored as a
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -99,7 +99,9 @@ static void populateODBProperties(Descriptor::Properties& props,
   }
 }
 
-std::string Descriptor::convertUnits(const double value, const bool area)
+std::string Descriptor::convertUnits(const double value,
+                                     const bool area,
+                                     int digits)
 {
   double log_value = value;
   if (area) {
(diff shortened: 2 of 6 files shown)


maliberty:
Why not just put the timing data on dbInst?  The split between db and sta is mostly relevant to developers and not users.

maliberty:
I'm glad I lured you into adding more timing descriptors :)

gadfort:
@maliberty because the STA instance takes time to load, I think it makes sense to separate it out (on a large design is can be a long time for STA to compute the information so it would be annoying (I think) to force that on the dbInst descriptor. But we could "rename" the STA instance descriptor to something like "Timing/Power"?

gadfort:
@maliberty I added a clock descriptor and added some handling of port busses as well

maliberty:
The code looks good.  Thanks for filling in the highlighting.

If the timing/power is going to take time then we should have some indication (eg wait cursor) that processing is happening.

maliberty:
Do you have an example of what the timing/power inspector looks like?

gadfort:
@maliberty 
image

I think the power activity formatting could use some work.
The setup and hold times are actually arrival times, so I'm open to rename them, if you have any suggestions.
Additionally, if there are other information that would be accessible that is useful I can add it now (if you have something in mind). This was largely done to take a look at the clockgating power activities.

gadfort:
@maliberty I added a wait cursor to the data loading in the inspector

maliberty:
I think we can skip activity on power/ground pins.  I think we should suppress "INF ps"/"-INF ps" or replace them with "None"

How about "Max arrival times" for "Setup times" and min for hold?  If it is an timing endpoint you could add the actual setup/hold times separately where applicable.

maliberty:
It also too many digits, eg "50.0000%".  Two is the default in sta_report_default_digits

gadfort:
@maliberty 
image

This is what it looks like now + a segfault fix on internal ports.

maliberty:
Much nicer.  I gave 50.0000% as an example but I really mean all the values in the form not just the percentage.

gadfort:
> Much nicer. I gave 50.0000% as an example but I really mean all the values in the form not just the percentage.

image

How about now?

maliberty (approved):

rovinski:
Maybe `0.00% at 0 aHz from constant` could just be `constant`

PR #5826, opened 2024-09-29
pad: update rdl router

gadfort:
Changes:
- change RDL router to use priority queue for routing.
- instead of only attempting a single target pair, this now attempts to route all available (unless a route is successful). the order will be based on distance (shortest first and going to the longest route at the end).
- if a route is not possible, the intersecting routes will be ripped up and routed after the failed route is attempted. This ensure that trivial failed routes are possible to complete.
- when working on an octagon shaped bump, 4 targets will be created at the north, south, east, and west edges to ensure legal routes are generated to these bumps.
- add max iterations option to avoid running forever when a given bump / pad layout is not possible to route. (default set to 10 iterations), however, the more iterations do not ensure a better result since this would indicate a poor layout.
- Added tests to check for proper new behavior and updated defs for new routing order.

I realize this ended up being a fairly large PR, the bulk of the file changes are related to adding an octagon bump and the new tests, the code changes are mainly in the `RDLRouter.??`. The fundamental routing mechanism is not modified, just how routes are handled before/after routing.


src/pad/src/RDLRouter.cpp
```diff
@@ -128,9 +129,141 @@ class RDLRouterGoalVisitor : public boost::default_astar_visitor
...
+
+  next_ = terminals_.begin();
+}
+
+bool RDLRoute::isIntersecting(RDLRoute* other, int extent) const
+{
+  // check current next_
+  // if at end of next, select begin + priority offset (modulo)
+  const int dst_idx = priority_ % terminals_.size();
+  odb::dbITerm* dst = terminals_[dst_idx];
```
maliberty, line 234: I don't follow this.

src/pad/src/RDLRouter.h
```diff
@@ -166,15 +249,12 @@ class RDLRouter
...
   {
-    return routing_terminals_;
-  }
-
-  const std::map<odb::dbNet*, std::vector<NetRoute>>& getRoutes() const
-  {
-    return routes_;
+    return routing_targets_;
   }
+  const std::vector<RDLRoutePtr>& getRoutes() const { return routes_; }
```
maliberty, line 256: No caller?

Do all these methods need to be public?  It seems like many are just called internally (eg getFailedRoutes).

src/pad/src/RDLRouter.cpp
```diff
@@ -128,9 +129,138 @@ class RDLRouterGoalVisitor : public boost::default_astar_visitor
...
+
+  next_++;
+
+  return iterm;
+}
+
+bool RDLRoute::compare(const std::shared_ptr<RDLRoute>& other) const
+{
+  const auto lhs_priority = -getPriority();
+  const auto rhs_priority = -other->getPriority();
```
maliberty, line 189: Priority is always first so why not just handle it up front
```
if (lhs_priority != rhs_priority) {
  return lhs_priority < rhs_priority ;
}
```
and let the rest of the code deal with distance alone if needed.

src/pad/src/RDLRouter.cpp
```diff
@@ -190,113 +322,439 @@ RDLRouter::~RDLRouter()
...
+    route_queue.pop();
+
+    route->markRouting();
+
+    odb::dbNet* net = src->getNet();
+    auto& net_targets = routing_targets_[net];
+
+    if (routed_covers.find(src->getInst()) != routed_covers.end()) {
+      // we've already routed this cover once (indicates the cover has multiple
+      // iterms)
```
maliberty, line 518: The subsequent code is a bit unclear as to the intention.  Is this skipped or delayed or ??

maliberty (changes requested):

src/pad/src/RDLRouter.cpp
```diff
@@ -1508,6 +1937,7 @@ RDLGui::RDLGui()
   addDisplayControl(draw_vertex_, true);
   addDisplayControl(draw_edge_, true);
   addDisplayControl(draw_obs_, true);
+  addDisplayControl(draw_targets_, true);
```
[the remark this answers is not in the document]
gadfort: once this lands, I can do that and the same for the RDLRoute class (that was laziness on my part but now the file has grown too large to keep together).

src/pad/src/RDLRouter.cpp
gadfort: This is just a simple way to "perturb" the routing calculations. The idea is just to go to a new destination iterm, if possible. I'm sure there are many improvements possible to the "what to ripup"-code, but this avoids the ripup from being identically calculated on a route that fails more than once.

src/pad/src/RDLRouter.h
gadfort: I believe they are there to be available to the GUI. If you have suggestions for how to structure it such that they can be private I can fix that.

src/pad/src/RDLRouter.cpp
gadfort: skipped, updated the comment to clarify

src/pad/src/RDLRouter.cpp
maliberty: Ok but please add a more descriptive comment.

src/pad/src/RDLRouter.h
maliberty: getFailedRoutes is only called from within RDLRouter.  getRoutes has no callers at all.  I haven't checked the rest.

maliberty:
I see two comments are resolved but I don't see an update.  Did you forgot to push?

gadfort:
> I see two comments are resolved but I don't see an update. Did you forget to push?

@maliberty I was marking them resolved as I fixed them locally, but I got pulled away so I haven't finished the last few issues. I'll push it this evening when I'm home

src/pad/src/RDLRouter.h
gadfort: https://github.com/The-OpenROAD-Project/OpenROAD/blob/a68afa069d600afdc0d5037b11e8a08dff96ecfa/src/pad/src/RDLRouter.cpp#L2039
https://github.com/The-OpenROAD-Project/OpenROAD/blob/a68afa069d600afdc0d5037b11e8a08dff96ecfa/src/pad/src/RDLRouter.cpp#L1962

src/pad/src/RDLRouter.cpp
```diff
@@ -128,9 +129,141 @@ class RDLRouterGoalVisitor : public boost::default_astar_visitor
...
+                   terminals_.end());
+
+  const odb::Point iterm_center = iterm_->getBBox().center();
+  std::stable_sort(terminals_.begin(),
+                   terminals_.end(),
+                   [&iterm_center](odb::dbITerm* lhs, odb::dbITerm* rhs) {
+                     const bool lhs_cover = RDLRouter::isCoverTerm(lhs);
+                     const bool rhs_cover = RDLRouter::isCoverTerm(rhs);
+                     // sort non-cover terms first
+                     if (lhs_cover == rhs_cover) {
```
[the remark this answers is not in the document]
gadfort: this was wrong, I fixed it.

src/pad/src/RDLRouter.cpp
```diff
@@ -154,16 +154,16 @@ RDLRoute::RDLRoute(odb::dbITerm* source,
   std::stable_sort(terminals_.begin(),
                    terminals_.end(),
                    [&iterm_center](odb::dbITerm* lhs, odb::dbITerm* rhs) {
-                     const bool lhs_cover = RDLRouter::isCoverTerm(lhs);
-                     const bool rhs_cover = RDLRouter::isCoverTerm(rhs);
+                     const int lhs_cover = RDLRouter::isCoverTerm(lhs) ? 1 : 0;
+                     const int rhs_cover = RDLRouter::isCoverTerm(rhs) ? 1 : 0;
```
maliberty, line 158: why convert from bool to int?

gadfort: It wasn't 100% clear to me how the bool comparison would, it wasn't clear from the cpp docs if true or false was smaller

maliberty: std::cout << (true < false);
    std::cout << (false < true);

01

gadfort: thanks for clarifying

maliberty (approved):

PR #5832, opened 2024-09-30
grt: FastRoute updateCongestionHistory refactor

joaomai:
The edge update that each up_type did was very similar, with small adjustments in the `else if (!stop_decreasing)` based on the up_type. 

Since the code is very particular to updateCongestionHistory and can hardly be used elsewhere, I've decided to keep the updateEdge function inside updateCongestionHistory.


src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -269,7 +269,9 @@ class FastRouteCore
                      const int L,
                      float& slack_th);
   void convertToMazeroute();
-  void updateCongestionHistory(const int upType, bool stopDEC, int& max_adj);
```
eder-matheus, line 272: Any particular reason to change the `updateCongestionHistory` to return an int instead of updating the `max_adj` var in the reference? This is one of the reasons the build is failing: max_adj is not initialized the first time it is used here.

joaomai: I've changed it to return an int instead of updating `max_adj` via reference to make it more explicit that the function is updating `max_adj` in the caller's code. Should I revert it to update via reference?

eder-matheus: I think it would be better, because it doesn't make much sense to me an update function returning a value. Keeping the reference will also fix the build issue.

eder-matheus (approved):

PR #5833, opened 2024-09-30
grt: disallow 1-track gcell near macro obstructions

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2382.

The two DRCs in asap7/swerv_wrapper are caused because GRT is using the only free track in M3 inside of a GCell that overlaps with a macro. When DRT uses this track, a violation happens in a M2-M3 via.

This PR disallows GRT to use single-track gcells that overlaps with macros to avoid via violations with the macro obstruction.


eder-matheus:
Metrics update here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2404

maliberty (approved):

PR #5834, opened 2024-09-30
gpl: mbff clock pin issue fix

therealshreyas:
IsClockPin() returned true for various pins in the synthetic ASAP7 MBFFs due to sta::Pin* -- removed for now.


maliberty:
I think this is just papering over whatever the real problem is.  I think the connectivity is so corrupted that clock appears to propagate to nets that make no sense.  For example
image

clk goes CLK, D0, QN1.  I would rather find the root cause then take this and have little confidence that there aren't other less obvious connectivity bugs.

maliberty (approved):

PR #5840, opened 2024-10-01
gui: fix coverity issues

gadfort:


src/gui/src/staDescriptors.cpp
```diff
@@ -204,7 +204,8 @@ Descriptor::Properties LibertyLibraryDescriptor::getProperties(
 
   SelectionSet corners;
   for (auto* corner : *sta_->corners()) {
-    for (const auto min_max : {sta::MinMax::min(), sta::MinMax::max()}) {
+    for (const sta::MinMax* min_max :
+         {sta::MinMax::min(), sta::MinMax::max()}) {
```
maliberty, line 208: What was the issue here?

maliberty (approved):

src/gui/src/staDescriptors.cpp
gadfort: auto vs. auto&, I just made it explicit that these are pointers.

maliberty: Weird I don't see why you need a reference to the pointer.

PR #5842, opened 2024-10-01
Add metrics to report_cell_usage

kareefardi:


src/dbSta/src/dbSta.cc
```diff
@@ -520,6 +529,11 @@ void dbSta::report_cell_usage(const bool verbose)
     logger_->report(
         format, type_name, stats.count, stats.area / area_to_microns);
     total_area += stats.area;
+
+    std::regex regexp(" |/|-");
+    logger_->metric("design__instance__count__class:"
+                        + toLowerCase(regex_replace(type_name, regexp, "_")),
```
maliberty, line 535: Why transform the type_name ?

kareefardi: I never saw any metric which is not using snake case. Is the non transformed name a valid metric name, like `Input/output pad`?

maliberty: It is legal in json though it could be confusing.

maliberty (approved):

PR #5858, opened 2024-10-02
drt: PA access pattern variable name changes

bnmfw:
As I was reading the Access Pattern of the PA module I noticed some variables still needed name changes. Here they are:

- `numNode` → `num_node`
- `numEdge` → `num_edge`
- `vioEdge` → `vio_edge`
- `apIdx` → `ap_idx`
- `via_2` → `via2`
- `via_1` → `via1`
- `leftAP` → `left_access_point` (In Access Pattern it is important to differentiate Access Point from Access Patterns as both abbreviate to AP)
- `rightAP` → `right_access_point`
- `leftPt` → `left_pt`
- `rightPt` → `right_pt`
- `tmpPt` → `tmp_pt`

### DIFF
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -1842,9 +1842,9 @@ void FlexPA::genInstRowPattern(std::vector<frInst*>& insts)
     return;
   }
 
-  const int numNode = (insts.size() + 2) * ACCESS_PATTERN_END_ITERATION_NUM;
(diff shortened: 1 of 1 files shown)


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2700,10 +2701,10 @@ bool FlexPA::genPatterns_commit(
       }
     }
 
-    frAccessPoint* leftAP = nullptr;
```
osamahammad21, line 2703: ap is fine by the way. it abbreviates accesspoint and is pretty standard so it's understandable.

osamahammad21 (approved):

PR #5862, opened 2024-10-02
drt: curr_idx name changes

bnmfw:
This PR used to change how getNestedIdx inputs and outputs data, but since the function will be eliminated further down the line I reverted it. Now this PR only contains variable changes.

Also changes instances of 
`curr_idx_1`, `curr_idx_2` to either  
`curr_pin_idx`, `curr_acc_point_idx` or
`curr_inst_idx`, `curr_acc_pattern_idx`
depending on the appropriate use.

Similar changes were applied to variables with the same format that have `prev` or `prev_prev` instead of `curr`.


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2863,13 +2851,12 @@ int FlexPA::getFlatIdx(const int idx_1, const int idx_2, const int idx_2_dim)
...
-                          int& idx_2,
-                          const int idx_2_dim)
+std::pair<int, int> FlexPA::getNestedIdx(const int flat_idx,
+                                         const int idx_2_dim)
 {
-  idx_1 = flat_idx / idx_2_dim - 1;
-  idx_2 = flat_idx % idx_2_dim;
+  const int idx_1 = flat_idx / idx_2_dim - 1;
+  const int idx_2 = flat_idx % idx_2_dim;
+  return {idx_1, idx_2};
```
maliberty, line 2859: Any idea where there is a '-1' here?  It seems non-idiomatic.

bnmfw: Yes. This function is the reciprocal to getFlatIdx:
```
int FlexPA::getFlatIdx(const int idx_1, const int idx_2, const int idx_2_dim)
{
  return ((idx_1 + 1) * idx_2_dim + idx_2);
}
```
The `-1` inverts the `+1` in the `idx_1`. The `-1`, in turn, is there because the minimal value for `idx_1` is `-1`, not `0`. So the returned value will be always positive. 

`-1` is used as an argument for this function at `genPatterns_reset`, `genPatternsInit`and `genInstRowPatternInit`

bnmfw: I actually don't like how the code keeps flattening and nesting these indexes, it happens a lot and some times one right after the other. If I find a way to eliminate it and keep a single representation for these indexes I will implement it.

bnmfw:
Just understood what `curr_idx_1` and `curr_idx_2` are and they are terrible names so I'm changing them and including the change in this PR

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -1881,20 +1881,22 @@ void FlexPA::genInstRowPatternInit(std::vector<FlexDPNode>& nodes,
 void FlexPA::genInstRowPatternPerform(std::vector<FlexDPNode>& nodes,
                                       const std::vector<frInst*>& insts)
 {
-  for (int curr_idx_1 = 0; curr_idx_1 <= (int) insts.size(); curr_idx_1++) {
-    for (int curr_idx_2 = 0; curr_idx_2 < ACCESS_PATTERN_END_ITERATION_NUM;
-         curr_idx_2++) {
+  for (int curr_pin_idx = 0; curr_pin_idx <= (int) insts.size();
```
osamahammad21, line 1884: why not curr _inst_idx?

osamahammad21 (approved):
I am running an ISPD CI and a secure CI. merging when they pass.

src/drt/src/pa/FlexPA_prep.cpp
bnmfw: Oversight. I actually have already caught that on a further PR I'm working on, but did not think to fix it here. Nice catch. I'm fixing this an re requesting review.

bnmfw:
> I am running an ISPD CI and a secure CI. merging when they pass.

Did the CIs pass?

osamahammad21 (approved):

Issue #628, opened 2021-05-04
'Not enough memory' error at detailed routing step


stevvik:
I'm trying to run test design through OpenROAD flow. It is a simple combinational scheme that contains ~4K gates. For this number of gates I have to increase die area up to 1000x1000 um. With this area design passes successfully. Buf if I try to increase die area up to ~2000x2000um, the openroad flow fails at the begining of detailed routing step with the following error:

set routed_def [make_resultfile ${design}${platform}_route.def]
set tr_lef [make_tr_lef]
Error: flow.tcl, 149 couldn't fork child process: not enough memory

I run openroad in container at host with 8G memory.

According to Matt Liberty recomendatin in gitter community I opened this issue and created a link to design.
Link below contains tuned dir structure with failed design, runfiles and logs. RUN.log contains log of failed run.
Design is created from gcd test example from openroad flow. I had to comment CTS step in flow.tcl to avoid errors that clock is absent in design.

[https://drive.google.com/file/d/1_Yz6AcxzRaMImOt5hrs5_20ks-IXL2uX/view?usp=sharing](url)

maliberty:
I tweaked your test case to run with the latest openroad and I can see the virtual size of openroad exceed 8Gb.  However I would expect that would just lead to swapping no an out of memory error.  Do you have swapping enable / have sufficient swap inside your container?

maliberty:
You do not have 4k instances, you have 116,189 instances.  There are 127 instances of crc32_table * 883 std cell per instace = 112,141 instances plus some 4k top level std cells.

Resizer will add some buffers too.  Increasing the die area increases the fill cell count so you get to 809,675 instances by the time detailed routing starts with 126k nets to route.

For this scenario the memory usage seems reasonable.

stevvik:
Matt, thank you for answer! Where did you take this statistic for instances? I got 4K by summing all gates after yosys (near the end of its log). I sow a big number of instances in openroad log, but decided that it is not a number of gates, but a total number of placed instances, for example, including vias. 

— Евгений Столбиков



среда, 5 мая 2021 г., 0:55 +0300 от ***@***.***  ***@***.***>:

maliberty:
We print the number at various points in the flow but you can always do:

set blk [ord::get_db_block]
puts "Num insts = [llen [$blk getInsts]]"

stevvik:
Thank you. And what are the ‘instances’ in your case? Are they a standard cells only or another topology figures too?

[...]

PR #5863, opened 2024-10-03
ppl: update write_pin_placement to write pins from odb

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5851.


eder-matheus:
@maliberty Lint Tcl failed, but it doesn't show what is wrong. Running tclfmt in the file change a lot of things completely unrelated to this PR. Do you know how to proceed?

maliberty:
Looks like you'll need to run tclfmt on IOPlacer.tcl
image

@nmoroze it is possible to get a more descriptive output?

maliberty (approved):

gadfort:
@eder-matheus you might need to make sure you have the right version. I ran into that issue before.

nmoroze:
@eder-matheus what tclfmt version/command are you running and what are some examples of the changes you're seeing? @gadfort is right, although I think in this case using any released version that includes `tclfmt` should give the correct result. Looking at things locally, the only change that needs to happen is:

Screenshot 2024-10-03 at 8 52 16 AM

@maliberty good question, it would be simple for the tool to at least print out what command to run to fix the issue. Another thing I can try is updating the CI job to this:

```
tclfmt --in-place .
git diff --exit-code
```

which is a pretty simple way to achieve the equivalent of having something like a diff mode.

eder-matheus:
> tclfmt --in-place .

@nmoroze I'm using tclfmt 0.4.0. When running `tclfmt IOPlacer.tcl`, there are changes all over the code, similar to the figure below:
image

But also some larger changes:
image

gadfort:
@nmoroze it might be helpful to also print the version of tclfmt at the start of the check to make this easier to debug as well.

eder-matheus:
@maliberty This unit test is failing in the head pipeline, and has nothing to do with this PR:
`614/799 Test #628: psm.gcd_em_test_vdd ...........................................***Failed    2.89 sec`

@bnmfw reported something similar in one of his DRT PRs.

maliberty:
The CI uses 0.4.1 (https://github.com/The-OpenROAD-Project/OpenROAD/blob/edf00dff99f6c40d67a30c0e22a8191c5d2ed9d6/.github/workflows/github-actions-lint-tcl.yml#L15)

maliberty:
I sounds like psm.gcd_em_test_vdd might have some non-determinism that should be investigated.  Running it with valgrind and/or ASAN would be good.

eder-matheus:
> I sounds like psm.gcd_em_test_vdd might have some non-determinism that should be investigated. Running it with valgrind and/or ASAN would be good.

I'm on it.

nmoroze:
@eder-matheus I looked more closely and I think the large errors you saw with tclfmt 0.4.0 probably came from running tclfmt outside of the project root directory (which would have prevented it from picking up the configuration). So if you see that again, that's something to double check.

I plan to fix this behavior as well - there's already an old issue for it (https://github.com/nmoroze/tclint/issues/30), but I'll make it a higher priority to get around to.

eder-matheus:
> @eder-matheus I looked more closely and I think the huge changes you saw with tclfmt 0.4.0 probably came from running tclfmt outside of the project root directory (which would have prevented it from picking up the configuration). So if you see that again, that's something to double check.
> 
> I plan to fix this behavior as well - there's already an old issue for it ([nmoroze/tclint#30](https://github.com/nmoroze/tclint/issues/30)), but I'll make it a higher priority to get around to.

That's it! I run it inside `OpenROAD/src/ppl/src` directory. Running it in the project root behaves as expected.

andyfox-rushc:
I struggled a bit with the tcl formatting. Here are some notes (doubtless to myself!).

I found that I can just run it on the file I changed (eg OpenRoad/src/OpenRoad.tcl)  from OpenRoad project root 

To get the linter:
 python -m pip install tclint==0.4.1

tclfmt --in-place ./src/OpenRoad.tcl  <-- format the file.  Seems to use configuration in  tclint.toml
tclint --check ./src/OpenRoad.tcl
tclint --no-check-style ./src/OpenRoad.tcl

However there are some issues if you run in the local directory because it does not pick up the configuration in the tclint.toml. The worst of which is the line size (tclint defaults to 80  and ident 8).  To run locally in the directory I used:  tclfmt --indent 2 --spaces-in-braces --in-place source OpenRoad.tcl
(I could not find a line option to set the width).

Running the commands standalone shows the offending lines in the code.

maliberty (approved):

andyfox-rushc:
Might be worth reformatting the tcl file so that: 

(a) Expressions use { }. eg what is the difference between: 
set make_tech [expr { ![ord::db_has_tech] }] and
set make_tech [expr ![ord::db_has_tech]]
(b) No lines > 80 columns

I ran some slightly stricter linting and found a few of these kinds of things. Not sure how important it is ! I had put the changes in pull request 5836 (now backed out)

nmoroze:
@andyfox-rushc to address some of your comments:

Yes, the observation about different results depending on where you run `tclint` or `tclfmt` is the same issue that @eder-matheus ran into (and is tracked as something to be fixed: https://github.com/nmoroze/tclint/issues/30). A good workaround that I just thought of is to use the `-c`/`--config` flag, for example:

```
export TCLINT_CFG=/path/to/openroad/tclint.toml

# prepend `-c $TCLINT_CFG` to any calls within the repository to always use the config, e.g.:
tclint -c $TCLINT_CFG ...
tclfmt -c $TCLINT_CFG ...
```

The stricter linting you noticed is also related to the config file not being applied. I don't think there's any need to change specific files to match the repo's tclint configuration, unless that configuration itself is going to be changed.

The particular note about braces around expressions is interesting, since that may be something worth eventually fixing and un-ignoring in the config. The braces actually make a subtle difference in how the expression is interpreted (and I figure in 99% of cases the braced version is what the user actually intends). Without the braces, substitution is performed first and interpreted as part of the expression itself, otherwise it acts like a variable within the expression. Better explained with an example:

```tcl
set foo "1 + 2"
puts [expr $foo]   ;# outputs "3"
puts [expr {$foo}] ;# outputs "1 + 2"
```

The Tcl language reference itself suggests adding braces as a performance consideration as well, see "Performance Considerations" under https://tcl.tk/man/tcl/TclCmd/expr.htm. 

Anyways, I think this isn't directly related to this PR at this point - happy to clarify more but maybe this discussion should be moved to #5836 or #5812.

maliberty:
@nmoroze please see the comments about braces above.

nmoroze:
@maliberty I think I addressed it above, but let me know if there's something else I'm missing!

PR #5869, opened 2024-10-03
grt: return the diodes count on repair_antennas command

eder-matheus:
Fix necessary for ORFS post-DRT repair_antennas.


maliberty (approved):
Will you need something like this for jumpers?

eder-matheus:
> Will you need something like this for jumpers?

The way I think about jumpers, no. Post-DRT repair will only rely on diode insertion, so we won't need this information in the flow.

On the other hand, it can be nice to have at least a metric of jumpers count to understand how much we are changing the original groute.

PR #5871, opened 2024-10-04
Added API isInClock to check if an inst ITerm is connected to clock

vikrg23:
Modified Design.cc and Design.h to add function for isInClock with iTerm as the argument.
Also modified timing_api_3.py and timing_api_3.ok to test this new feature.


maliberty (approved):

maliberty:
Thanks for the PR.  You have missed signing your commit so DCO is failing.  Please click the details link to see info on fixing that.

PR #5872, opened 2024-10-04
drt: genPatterns_commit refactoring

bnmfw:
Some refactoring on `genPatterns_commit` is being done to improve readability and reduce code.
### Access Patterns Graph Backwards Path Function and Readability
The first part of `genPatterns_commit` is a loop that traces a path through access points of an instance that represent a chosen access pattern.

Right now the loop is a little over complicated for what it does, so I refactored it a bit to make more obvious.

Added the variables `source_node_idx` and `drain_node_idx` to make the start and end points of the path search more clear, instead decreasing `pin_cnt`.

This was transformed in a function called `extractAccessPatternFromNodes` as it has a pretty distinct role and function.

### Used Pattern Guard
The later part of the had an if statement to check is the generated pattern was unused, if true it did more checks, later it essentially had an else with `return false`. This as transformed into a guard.

No further changes on this function will be done in this PR as the git diff shows everything past the guard as changed due to the indentation level change. Further changes will be at a separate PR.


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2634,139 +2665,113 @@ bool FlexPA::genPatterns_commit(
...
+        n_no_ap_pins++;
+        pin_access_pattern->addAccessPoint(nullptr);
+      } else {
+        const auto& ap = pin_to_access_point[pin.get()];
+        const Point tmpPt = ap->getPoint();
+        if (tmpPt.x() < leftPt) {
+          leftAP = ap;
+          leftPt = tmpPt.x();
+        }
+        if (tmpPt.x() > rightPt) {
```
rovinski, line 2735: Should this be `else if`?

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2634,139 +2665,113 @@ bool FlexPA::genPatterns_commit(
...
+          rightAP = ap;
+          rightPt = tmpPt.x();
         }
+        pin_access_pattern->addAccessPoint(ap);
       }
     }
+    if (n_no_ap_pins == inst_term->getTerm()->getPins().size()) {
+      logger_->error(DRT, 91, "Pin does not have valid ap.");
```
rovinski, line 2743: Capitalize AP. Also, can you add the term name to the print?

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2624,6 +2624,37 @@ int FlexPA::getEdgeCost(
...
+    const std::vector<std::pair<frMPin*, frInstTerm*>>& pins,
+    std::set<std::pair<int, int>>& used_access_points,
+    const int max_access_point_size)
+{
+  std::vector<int> access_pattern(pins.size(), -1);
+
+  const int source_node_idx = getFlatIdx(-1, 0, max_access_point_size);
+  const int drain_node_idx = getFlatIdx(pins.size(), 0, max_access_point_size);
+  auto drain_node = &(nodes[drain_node_idx]);
+  int curr_node_idx = drain_node->getPrevNodeIdx();
```
rovinski, line 2638: Does "drain" have a special meeting here? I the opposite of source is usually "sink" in these contexts

src/drt/src/pa/FlexPA.h
```diff
@@ -568,6 +568,25 @@ class FlexPA
                   int curr_unique_inst_idx,
                   int max_access_point_size);
 
+  /**
+   * @brief Extracts the access patterns given the graph nodes composing the
+   * access points relatioship
```
rovinski, line 573:
```suggestion
   * access points relationship
```

src/drt/src/pa/FlexPA.h
```diff
@@ -568,6 +568,25 @@ class FlexPA
...
+   *
+   * @param nodes {pin,access_point} nodes of the access pattern graph
+   * @param pins vector os pins of the unique instance
```
rovinski, line 576: vector of pins?

src/drt/src/pa/FlexPA.h
```diff
@@ -568,6 +568,25 @@ class FlexPA
...
+   * @param used_access_points a set of all used access points
+   * @param max_access_point_size number of acc points the instance with most
+   * acc points has
```
rovinski, line 579:
```suggestion
   * @param max_access_point_size the maximum number of access points across
   * all instances
```

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2634,139 +2665,113 @@ bool FlexPA::genPatterns_commit(
...
-        if (inst_term->hasNet()) {
-          if (owners.find(inst_term->getNet()) != owners.end()) {
-            viol_access_points.insert(std::make_pair(idx_1, idx_2));  // idx ;
-          }
-        } else {
-          if (owners.find(inst_term) != owners.end()) {
-            viol_access_points.insert(std::make_pair(idx_1, idx_2));  // idx ;
-          }
+      } else {
+        if (owners.find(inst_term) != owners.end()) {
```
rovinski, line 2766: I wonder if there's a more efficient way to do this because this will traverse `owners` twice. It might not matter if the `else` is rarely exercised or if `owners` is small.

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2624,6 +2624,37 @@ int FlexPA::getEdgeCost(
...
+    const std::vector<FlexDPNode>& nodes,
+    const std::vector<std::pair<frMPin*, frInstTerm*>>& pins,
+    std::set<std::pair<int, int>>& used_access_points,
+    const int max_access_point_size)
+{
+  std::vector<int> access_pattern(pins.size(), -1);
+
+  const int source_node_idx = getFlatIdx(-1, 0, max_access_point_size);
+  const int drain_node_idx = getFlatIdx(pins.size(), 0, max_access_point_size);
+  auto drain_node = &(nodes[drain_node_idx]);
```
osamahammad21, line 2637:
```suggestion
  const auto& drain_node = nodes[drain_node_idx];
```
use const reference to the object instead of pointer.

src/drt/src/pa/FlexPA_prep.cpp
osamahammad21: I guess sink is the correct word in this context.

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2624,6 +2624,37 @@ int FlexPA::getEdgeCost(
...
+  int curr_node_idx = drain_node->getPrevNodeIdx();
+
+  while (curr_node_idx != source_node_idx) {
+    if (curr_node_idx == -1) {
+      logger_->error(DRT, 90, "Valid access pattern not found.");
+    }
+
+    auto curr_node = &(nodes[curr_node_idx]);
```
osamahammad21, line 2645: ditto

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2634,139 +2665,113 @@ bool FlexPA::genPatterns_commit(
...
-      // add objs
-      std::unique_ptr<frVia> via;
-      if (access_point->hasAccess(frDirEnum::U)) {
-        via = std::make_unique<frVia>(access_point->getViaDef());
-        auto rvia = via.get();
-        temp_vias.push_back(std::move(via));
+  frAccessPoint* leftAP = nullptr;
+  frAccessPoint* rightAP = nullptr;
+  frCoord leftPt = std::numeric_limits<frCoord>::max();
+  frCoord rightPt = std::numeric_limits<frCoord>::min();
```
osamahammad21, line 2715: why the renaming? this is not the followed style.

src/drt/src/pa/FlexPA_prep.cpp
osamahammad21: nope. Consider the first iteration. we need to set both; leftPt and rightPt, to tmpPt

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA.h
osamahammad21: @bnmfw

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2643,139 +2674,116 @@ bool FlexPA::genPatterns_commit(
...
-            right_pt = tmp_pt.x();
-          }
-          pin_access_pattern->addAccessPoint(ap);
+  const auto& [pin, inst_term] = pins[0];
+  const auto inst = inst_term->getInst();
+  for (auto& inst_term : inst->getInstTerms()) {
+    if (isSkipInstTerm(inst_term.get())) {
+      continue;
+    }
+    uint64_t n_no_AP_pins = 0;
```
osamahammad21, line 2732: why change variable name? That's not the correct style.

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2643,139 +2674,116 @@ bool FlexPA::genPatterns_commit(
...
+    for (auto& pin : inst_term->getTerm()->getPins()) {
+      if (pin_to_access_point.find(pin.get()) == pin_to_access_point.end()) {
+        n_no_AP_pins++;
+        pin_access_pattern->addAccessPoint(nullptr);
+      } else {
+        const auto& ap = pin_to_access_point[pin.get()];
+        const Point tmpPt = ap->getPoint();
```
osamahammad21, line 2739: ditto

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2643,139 +2674,116 @@ bool FlexPA::genPatterns_commit(
...
-            viol_access_points.insert(std::make_pair(idx_1, idx_2));  // idx ;
-          }
-        } else {
-          if (owners.find(inst_term) != owners.end()) {
-            viol_access_points.insert(std::make_pair(idx_1, idx_2));  // idx ;
-          }
+    if (n_no_AP_pins == inst_term->getTerm()->getPins().size()) {
+      logger_->error(DRT,
+                     91,
+                     "{} Pins do not have valid access points.",
```
osamahammad21, line 2754:
```suggestion
                     "{} does not have valid access points.",
```

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -2633,6 +2633,37 @@ int FlexPA::getEdgeCost(
...
+
+  const int source_node_idx = getFlatIdx(-1, 0, max_access_point_size);
+  const int sink_node_idx = getFlatIdx(pins.size(), 0, max_access_point_size);
+  const FlexDPNode& sink_node = nodes[sink_node_idx];
+  int curr_node_idx = sink_node.getPrevNodeIdx();
+
+  while (curr_node_idx != source_node_idx) {
+    if (curr_node_idx == -1) {
+      logger_->error(DRT, 90, "Valid access pattern not found.");
+    }
```
osamahammad21, line 2652: Is this equivalent to counting the number of pins visited?

osamahammad21 (changes requested):
Address the comments please.

maliberty:
@bnmfw I see various questions that need a response.

src/drt/src/pa/FlexPA_prep.cpp
bnmfw: curr_node_idx == -1 means we reached a DP node that does not have an ancestor. For a valid access patterns this should only happen at the source node. If this happens that means that the access pattern has not been solved for this unique instance. I don't see the how it could correlate to the number of pins visited. If you can reproduce the issue I would be glad to address it directly

src/drt/src/pa/FlexPA.h
```diff
@@ -595,6 +595,25 @@ class FlexPA
                   int curr_unique_inst_idx,
                   int max_access_point_size);
 
+  /**
+   * @brief Extracts the access patterns given the graph nodes composing the
+   * access points relationship
+   *
+   * @param nodes {pin,access_point} nodes of the access pattern graph
+   * @param pins vector os pins of the unique instance
```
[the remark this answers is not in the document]
bnmfw: 💀

PR #5879, opened 2024-10-04
odb: add command to check if design is routed

eder-matheus:


src/odb/src/db/odb.tcl
```diff
@@ -648,3 +648,18 @@ proc set_max_layer { args } {
   set block [$chip getBlock]
   $block setMaxRoutingLayer $maxLayer
 }
+
+sta::define_cmd_args "design_is_routed" {}
+
+proc design_is_routed { args } {
+  sta::parse_key_args "set_max_layer" args keys {} flags {}
```
gadfort, line 655: design_is_routed

gadfort (commented):
Shouldn't this have a set of CI tests?

eder-matheus:
> Shouldn't this have a set of CI tests?

Yes, I'm working on it. I'm also testing it in the CI to check if I'm missing some check.

src/odb/src/db/dbBlock.cpp
```diff
@@ -4097,6 +4097,18 @@ void dbBlock::preExttreeMergeRC(double max_cap, uint corner)
   }
 }
 
+bool dbBlock::designIsRouted()
+{
+  for (dbNet* net : getNets()) {
+    const int pin_count = net->getBTermCount() + net->getITerms().size();
+    if (pin_count > 1 && net->getSWires().empty() && net->getWire() == nullptr
+        && !net->isConnectedByAbutment()) {
```
gadfort, line 4105: if the SWires is not empty, you should probably verify that they actually have routing in them.
Same for getWire (ensure it actually contains something).

eder-matheus: I was able to simplify it by getting the wire count of each net. This function already cover both special and non-special nets.

eder-matheus:

maliberty (approved):

PR #5892, opened 2024-10-07
drt: pin type check

bnmfw:
This secure CI associated with this PR passed with no changes.

In Multiple parts of the code the Master of an instance was deferenced to get its type and determine the type of the instance. In multiple parts of the code the same check was done either to verify if the instance was a Standard Cell or a Macro Cell. Two respective functions were created to make this check given an instance.

`isStdCell()` checks for as the `isCore()` instead of directly looking for various typing. This caused no changes in the secure CI Run

### DIFF
--- a/src/drt/src/pa/FlexPA_prep.cpp
+++ b/src/drt/src/pa/FlexPA_prep.cpp
@@ -53,11 +53,12 @@ using utl::ThreadException;
 // terms from OpenDB
 bool FlexPA::isStdCellPin(frInst* inst)
 {
-  dbMasterType masterType = inst->getMaster()->getMasterType();
-  return (masterType == dbMasterType::CORE
-          || masterType == dbMasterType::CORE_TIEHIGH
-          || masterType == dbMasterType::CORE_TIELOW
-          || masterType == dbMasterType::CORE_ANTENNACELL);
+  // dbMasterType masterType = inst->getMaster()->getMasterType();
+  // return (masterType == dbMasterType::CORE
+  //         || masterType == dbMasterType::CORE_TIEHIGH
+  //         || masterType == dbMasterType::CORE_TIELOW
+  //         || masterType == dbMasterType::CORE_ANTENNACELL);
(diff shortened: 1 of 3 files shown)


src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -49,6 +49,25 @@ namespace drt {
 
 using utl::ThreadException;
 
+// TODO there should be a better way to get this info by getting the master
```
eder-matheus, line 52: It seems this comment is outdated now, since you are already using odb to get the info.

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -49,6 +49,20 @@ namespace drt {
...
+bool FlexPA::isStdCell(frInst* inst)
+{
+  return inst->getMaster()->getMasterType().isCore();
+}
+
+bool FlexPA::isMacroCell(frInst* inst)
+{
+  dbMasterType masterType = inst->getMaster()->getMasterType();
+  return (masterType.isBlock() || masterType.isPad()
+          || masterType == dbMasterType::RING);
```
eder-matheus, line 63: The only comment I have is this `dbMasterType::RING` inside a function called `isMacroCell`. It makes sense in the logic of DRT code, but I'm not sure how accurate is that in terms of the database. @maliberty any thoughts on this?

maliberty: From the manual:
```
Large macro that has an internal power mesh, and only exposes powerpin shapes that form a ring along the macro boundary
```
so a RING has no signal pins that could matter to drt/pa.    I doubt it matters whether is included or not but it is more like a macro than a stdcell.

eder-matheus: Thanks for the clarification! I think it's fine to keep it for now. The rest of the changes looks good too.

eder-matheus (approved):

PR #5895, opened 2024-10-08
Improved output in lint Tcl job

nmoroze:
Some small tweaks based on the feedback in #5863. 

Also updates to the new tclint, which gets some bugfixes.

### DIFF
--- a/.github/workflows/github-actions-lint-tcl.yml
+++ b/.github/workflows/github-actions-lint-tcl.yml
@@ -16,5 +16,7 @@ jobs:
 
       - name: Lint
         run: |
+          tclfmt --version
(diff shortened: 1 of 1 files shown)


maliberty:
I suppose from "DO NOT MERGE example bad change](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5895/commits/a7409a6ad5c17549e513fcc9e7959dce3ed21819)" that this is still in testing and not ready to go.

nmoroze:
Sorry, should have started this in a draft state... but just lopped that commit off and should be good to go now!

maliberty (approved):

PR #5898, opened 2024-10-08
mbff: Fix missing fmt include

gatecat:
On my system this was causing a `no member named 'join' in namespace 'fmt'` error.


maliberty:
You must have a different version of fmt installed:
```
/tmp/workspace/OpenROAD-Public_PR-5898-head/src/gpl/src/mbff.cpp:49:10: fatal error: spdlog/fmt/ranges.h: No such file or directory
   49 | #include "spdlog/fmt/ranges.h"
```

gatecat:
Yeah, looks like things were using system spdlog/fmt.

PR #5900, opened 2024-10-08
Cts enable obstruction aware

arthurjolo:
This PR removes the CTS option -obstruction_aware and enables it by default, and adds improvements to prevent metrics degradation:

- Remove CTS option -obstruction_aware
- Insert delay to balance clock gater tree
- Improve clock gater handling to correctly inset delay.


arthurjolo:
Secure-CI is has few metrics failures and a PR has been created to update them, a7/mock_array and ng45/bp_fe have degradation on timing metrics due to a bad placement decision from obstruction aware, I am already working on fixing this issue.

PR #5906, opened 2024-10-09
grt: getCost refactor and mazeRouteMSMD relax grids refactor

joaomai:
getCost function refactoring to remove the need of checking if the cost was already calculated for a given position.

relaxAdjacent and updateAdjacent were made as lambdas since both need to access and modify local variables to the mazeRouteMSMD function.

When relaxing the left grid the cost was stored using float, when the other directions were using double. The change to double caused changes in metrics such as total usage and via count for some tests. Similarly, both cost tables were using double while getCost returned a float, which was updated to use double too.


src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -99,6 +99,25 @@ struct parent3D
...
+  const int cost_type;
+
+  CostParams(const float logis_cof,
+             const float cost_height,
+             const int slope,
+             const int cost_type)
+      : logis_cof(logis_cof),
+        cost_height(cost_height),
+        slope(slope),
+        cost_type(cost_type)
```
eder-matheus, line 116: It would be nice to understand what each cost type means and change it from `int` to an `enum class`, with descritive names. Today we only compare with numbers, which is very unclear.

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -1097,59 +1098,51 @@ void FastRouteCore::reInitTree(const int netID)
   checkAndFixEmbeddedTree(netID);
 }
 
-float getCost(const int i,
-              const float logis_cof,
-              const float cost_height,
-              const int slope,
-              const int capacity,
-              const int cost_type)
+double FastRouteCore::getCost(const int i,
```
eder-matheus, line 1101: The parameter `i` could have a more descriptive name. Looking at the usage of this functions, the code is passing a `pos`, but it doesn't seem to be a position relative to the grid.

src/grt/src/fastroute/src/maze.cpp
```diff
@@ -1174,6 +1167,91 @@ void FastRouteCore::mazeRouteMSMD(const int iter,
 
   std::vector<bool> pop_heap2(y_grid_ * x_range_, false);
 
+  auto updateAdjacent = [&](const int cur_x,
```
eder-matheus, line 1170: Add a small comment with the summary of what this lambda function is doing. Here and below.

src/grt/src/fastroute/src/maze.cpp
eder-matheus: It is probably a position related to the heap.

eder-matheus (changes requested):

eder-matheus (approved):

eder-matheus:
@joaomai The changes look good, but you will need to update the metrics for the failing designs in the secure-ci. Only then we can merge this PR.

maliberty:
conflicts to resolve

PR #5913, opened 2024-10-10
drt: Push node index computation up

kbieganski:
No functional change intended. This speeds up DRT by 3-8%, at least for `asap7` (varies depending on the design). Baseline is a recent commit from Oct 6.

## `asap7/aes` - stage `5_2_route` - 8 thread(s)

| Branch | Min [s] | Median [s] | Relative |
|--------|---------|------------|----------|
| Baseline | 221.96 | 222.23 | 1.03 |
| This PR | 215.48 | 215.91 | 1.00 |

## `asap7/ibex` - stage `5_2_route` - 8 thread(s)

| Branch | Min [s] | Median [s] | Relative |
|--------|---------|------------|----------|
| Baseline | 337.61 | 338.30 | 1.08 |
| This PR | 312.99 | 313.16 | 1.00 |

### DIFF
--- a/src/drt/src/dr/FlexDR.h
+++ b/src/drt/src/dr/FlexDR.h
@@ -867,19 +867,13 @@ class FlexDRWorker
   void modMinSpacingCostVia_eol(const Rect& box,
                                 const Rect& tmpBx,
                                 ModCostType type,
-                                bool isUpperVia,
                                 const drEolSpacingConstraint& drCon,
-                                frMIdx i,
-                                frMIdx j,
-                                frMIdx z,
+                                frMIdx idx,
                                 bool ndr = false);
   void modMinSpacingCostVia_eol_helper(const Rect& box,
                                        const Rect& testBox,
                                        ModCostType type,
-                                       bool isUpperVia,
-                                       frMIdx i,
-                                       frMIdx j,
-                                       frMIdx z,
+                                       frMIdx idx,
                                        bool ndr = false);
   // eolSpc
   void modEolSpacingCost_helper(const Rect& testbox,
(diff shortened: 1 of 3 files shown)


kbieganski:
@osamahammad21 Have you had a change to look at this?

kbieganski:
Can this be reviewed?

maliberty:
@osamahammad21 please review

osamahammad21 (changes requested):
In this PR, I see that checking for an index validity is completely removed from all APIs. Has it been verified that the passed index to getIdx is valid? For example, bi.z() - 1 could be -1 in some cases, right?

kbieganski:
I think each `switch` is a mix of checked and unchecked accesses (the `*Fixed*` ones are checked, the `*Route*` ones aren't). For example, see [this part](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5913/files#diff-3b0ee5af511f4838323555d94ad93cc7da9df71a64bf049fa6b634da77978e64L84-R85) which calls [this function](https://github.com/The-OpenROAD-Project/OpenROAD/pull/5913/files#diff-a62f2dab4026966036f9855039c60b6de4b6f462499a04b0043890390288d668L555-R558). So I was confused by the "safe access" comment, thought it was about something else. Anyway, it seems the checks aren't actually needed?

kbieganski:
@osamahammad21 In case you missed it

osamahammad21:
Sorry I didn't notice your comment. It seems that as you said we do not always check the validity of the index which is kind doesn't seem very safe. However, this PR doesn't introduce new problems, So I'll just merge and get back to the index validity later on a separate PR.

osamahammad21 (approved):

PR #5920, opened 2024-10-11
gui: include pathgroup name information when searching for pathends

### LINKED ISSUE #5882: Clicking on reg2reg bucket lists 0 paths in timing port
### Describe the bug

untar 
[reg2reg-bucket.tar.gz](https://github.com/user-attachments/files/17266840/reg2reg-bucket.tar.gz)

```
 ./run-me-ebtb_128x40-asap7-base.sh 
```

Navigate to reg2reg, there should be no paths in reg2reg, but there is a green bucket.

Clicking on the green bucket shows no timing paths(which I believe to be correct).

image

### Expected Behavior

If there are no paths, show no bucket, if there is a bucket and there are paths in it list only those paths

### Environment

```shell
OpenROAD v2.0-16257-g81b545063
```

### To Reproduce

See above

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

AcKoucher:
Resolve #5882

It seems like the manual check we were doing was wrong. There's an option to include the PathGroup(s) name(s) on the search so we can get only the paths that belong to that group.

### DIFF
--- a/src/gui/src/staGuiInterface.cpp
+++ b/src/gui/src/staGuiInterface.cpp
@@ -1036,8 +1036,13 @@ TimingPathList STAGuiInterface::getTimingPaths(
                                  sta::RiseFallBoth::riseFall());
   }
 
-  sta::Search* search = sta_->search();
+  std::unique_ptr<sta::PathGroupNameSet> group_names;
+  if (!path_group_name.empty()) {
+    group_names = std::make_unique<sta::PathGroupNameSet>();
+    group_names->insert(path_group_name.c_str());
+  }
 
+  sta::Search* search = sta_->search();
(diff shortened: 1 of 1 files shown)


eder-matheus:
Will it change the behavior of cases that didn't have issues before?

maliberty (approved):

maliberty:
Looks ok to me but I'll let @eder-matheus merge once his question is answered.

AcKoucher:
@eder-matheus @maliberty 
>Will it change the behavior of cases that didn't have issues before?

It shouldn't. `group_names == nullptr` for all cases just like before. We only use this option when generating the report from Charts.
I can add a comment to make it clearer if desired.

PR #5929, opened 2024-10-12
gui: if no_init is specified on commandline previous settings will not be loaded

gadfort:
This provides a way to pass in the `no_init` information to the GUI on load to start the GUI in a clean state.
If we dont want to bring in the commandline flag, I suggest we still have this capability since it would be useful when debugging GUI issues and dealing with sometimes corrupted settings files.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -117,6 +117,7 @@ int cmd_argc;
 char** cmd_argv;
 const char* log_filename = nullptr;
 const char* metrics_filename = nullptr;
+bool no_init = false;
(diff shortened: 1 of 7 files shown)


maliberty:
I'm not sure it is good to overload -no_init.  Perhaps a separate -no_settings would be clearer?

gadfort:
I'll add a new flag for this one.

gadfort:
@maliberty done

maliberty (approved):

PR #5932, opened 2024-10-12
rmp: Moves Logic Cut to net based primary IO

QuantamHD:
Before this change logic cuts were sta::Pin* terminated this lead to weird effects where ABC would see two pins with the same driving net, and do weird things to its internal datastructure.

It also has the nice property that a logic cut based on nets can be reused multiple times if a user wants to run two versions of a script on a Abc_Ntk*.

### DIFF
--- a/src/rmp/src/logic_cut.cpp
+++ b/src/rmp/src/logic_cut.cpp
@@ -19,35 +19,40 @@
 
 namespace rmp {
 
-std::unordered_map<sta::Pin*, abc::Abc_Obj_t*> CreateAbcPrimaryInputs(
-    const std::vector<sta::Pin*>& primary_inputs,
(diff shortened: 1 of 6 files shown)


maliberty:
I don't understand the description "ABC would see two pins with the same driving net".  Why would cutting at pins cause this?

QuantamHD:
ABC expects that each primary output pin is driven by a unique logic function. It gets a bit weird when you have two primary outputs that are really the same output driven by a net. Let's say for some reason you had one net driving two flops, in the old code you would get two primary output pins, one for each terminating output pin.

In this new code primary inputs and outputs are derived from the nets connected to terminating pins. This makes it so that you don't have a single AIG instance fanning out to two primary outputs in ABC, which is when things get a bit weird.

Issue #5926, opened 2024-10-12
`io_outs_down[9]` should be at the bottom.


oharboe:
`io_outs_down[9]` should be at the bottom.

![image](https://github.com/user-attachments/assets/022e7f49-ce09-4542-9d7a-2cc9e5747575)

to reproduce with latest OpenROAD master:

[io_placement.tar.gz](https://github.com/user-attachments/files/17349220/io_placement.tar.gz)

_Originally posted by @oharboe in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2412#issuecomment-2408445062_

PR #5948, opened 2024-10-14
ppl: fix random assignment to respect constraints

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/5926


eder-matheus:
@oharboe Could you give it a try? I tested this fix with two of your test cases and it seems to work now.

eder-matheus:
@AcKoucher FYI this PR also fixes the bug related to the -exclude option that you reported earlier today.

maliberty (approved):

PR #5950, opened 2024-10-15
tcl: add a fallback from hardcoded TCLRL_LIBRARY path

oharboe:
TL;DR This fixes the last niggling little deployment problem with non-Ubuntu systems and bazel-orfs GUI.

Not pretty, but good 'nuf to use around the house.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -343,17 +343,17 @@ std::string findPathToTclreadlineInit(Tcl_Interp* interp)
   }
 
   const char* tclScript = R"(
-        proc find_path {} {
-            foreach dir $::auto_path {
-                set folder [file join $dir]
-                set path [file join $folder "tclreadline2.3.8" "tclreadlineInit.tcl"]
-                if {[file exists $path]} {
-                    return $path
-                }
+      namespace eval temp {
+        foreach dir $::auto_path {
+            set folder [file join $dir]
+            set path [file join $folder "tclreadline)" TCLRL_VERSION_STR
+                          R"(" "tclreadlineInit.tcl"]
+            if {[file exists $path]} {
+                return $path
             }
-            error "tclreadlineInit.tcl not found in any of the directories in auto_path"
         }
-        find_path
+        error "tclreadlineInit.tcl not found in any of the directories in auto_path"
+      }
     )";
 
   if (Tcl_Eval(interp, tclScript) == TCL_ERROR) {
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -342,6 +342,24 @@ std::string findPathToTclreadlineInit(Tcl_Interp* interp)
     return path;
   }
 
+  // TL;DR it is possible to run the OpenROAD binary from within the
+  // official Docker image on a different distribution than the
+  // distribution within the Docker image.
+  //
+  // In this case we have to look up
+  // the location of the tclreadline scripts instead of using the hardcoded
+  // path.
+  //
+  // It is helpful to use the official Docker image as CI infrastructure and
+  // also because it is a good way to have as similar an environment as possible
+  // during testing and deployment.
+  //
+  // See
+  // https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/docker.BUILD.bazel
+  // for the details on how this is done.
+  //
+  // Running Docker within a bazel isolated environment introduces lots of
+  // problems and is not really done.
   const char* tclScript = R"(
       namespace eval temp {
         foreach dir $::auto_path {
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -334,14 +334,10 @@ static int tclReadlineInit(Tcl_Interp* interp)
 #endif
 
 namespace {
-// A fallback from the hardcoded TCLRL_LIBRARY path.
+// A stopgap fallback from the hardcoded TCLRL_LIBRARY path for OpenROAD,
+// not essential for OpenSTA
 std::string findPathToTclreadlineInit(Tcl_Interp* interp)
 {
-  std::string path(TCLRL_LIBRARY "/tclreadlineInit.tcl");
-  if (is_regular_file(path.c_str())) {
-    return path;
-  }
-
   // TL;DR it is possible to run the OpenROAD binary from within the
   // official Docker image on a different distribution than the
   // distribution within the Docker image.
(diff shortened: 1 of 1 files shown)


gadfort:
@oharboe why is this needed? If TCL readline is installed correctly the TCLRL_LIBRARY should be defined in the header:
https://github.com/flightaware/tclreadline/blob/20f1425cec34eeef8f24f1e942e0be4038b8df3a/tclreadline.h.in#L16
Also not a huge fan of hardcoding the version number into the embedded tcl script, second, you should either define this in a namespace for tcl do "find_path" isn't just a random function in the tcl environment or you need to cleanup after yourself.

I would like to know better what the actual issue was that this is addressing, since in a headless build for bazel, why add tclreadline?

gadfort:
> > @oharboe why is this needed?
> 
> It is the final hack at the end of a long story...

If this cruft is to be "acceptable" I think it's worth knowing what the story is. Keep in mind there are no comments or anything describing the issue, so the next person who encounters this code will wonder why this is needed.

This feels like there is an issue with your tclreadline install and it would be better to fix that than have some code that isn't in opensta or the tclshrl builds.

oharboe:
@gadfort Added some gory details on why we have to look up the path dynamically.

oharboe:
@maliberty I don't understand the pr-merge failure, pr-head passes and this PR is on top of the most recent master, so pr-merge and pr-head should be identical.

src/Main.cc
```diff
@@ -333,6 +333,57 @@ static int tclReadlineInit(Tcl_Interp* interp)
...
+  // the location of the tclreadline scripts instead of using the hardcoded
+  // path.
+  //
+  // It is helpful to use the official Docker image as CI infrastructure and
+  // also because it is a good way to have as similar an environment as possible
+  // during testing and deployment.
+  //
+  // See
+  // https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/docker.BUILD.bazel
+  // for the details on how this is done.
```
maliberty, line 359: I don't see any details there that explain how it is done.  Its just a list of files and there is no mention of docker.

src/Main.cc
```diff
@@ -333,6 +333,57 @@ static int tclReadlineInit(Tcl_Interp* interp)
...
+std::string findPathToTclreadlineInit(Tcl_Interp* interp)
+{
+  std::string path(TCLRL_LIBRARY "/tclreadlineInit.tcl");
+  if (is_regular_file(path.c_str())) {
+    return path;
+  }
+
+  // TL;DR it is possible to run the OpenROAD binary from within the
+  // official Docker image on a different distribution than the
+  // distribution within the Docker image.
```
maliberty, line 347: Why do you want to do this?

src/Main.cc
oharboe: Documentation on **how** this is done belongs in bazel-orfs. Admittedly it could be better.

src/Main.cc
gadfort: it seems like there is something not right about the usecase, which is leading to this solution. Personally, I would prefer not to bring this in, if the solution is to properly build OpenROAD for bazel (or whatever platform) instead of building on one platform and moving the binary around.

oharboe: It is not one or the other; there's no conflict here.

This PR is boarding up a broken window (lack of resources for CI + programming), longer term I hope to clean up bazel-orfs.

src/Main.cc
```diff
@@ -373,8 +424,9 @@ static int tclAppInit(int& argc,
       // script is done.
       Tcl_StaticPackage(
           interp, "tclreadline", Tclreadline_Init, Tclreadline_SafeInit);
-      if (Tcl_EvalFile(interp, TCLRL_LIBRARY "/tclreadlineInit.tcl")
-          != TCL_OK) {
+
+      std::string path = findPathToTclreadlineInit(interp);
+      if (path.empty() || Tcl_EvalFile(interp, path.c_str()) != TCL_OK) {
```
maliberty, line 429: Rather than always taking this path, can we first do what was already there
```
    if (Tcl_EvalFile(interp, TCLRL_LIBRARY "/tclreadlineInit.tcl")
          != TCL_OK) {
```
and treat your solution as a fallback.  It seems less likely to be disruptive to standard use case.

oharboe: Then

maliberty: ?

oharboe: fat fingers, hit send early, response corrected.

maliberty:
> This feels like there is an issue with your tclreadline install and it would be better to fix that than have some code that isn't in opensta or the tclshrl builds.

This is a good point.  The sta app will not work with this approach.  I believe you are sta for power reporting.

src/Main.cc
oharboe: gotit. Fixed.

oharboe:
@maliberty mystery pr-merge error?

maliberty:
I see the same two tests failed in many builds:
	626 - psm.gcd_write_sp_test_vdd (Failed)
	628 - psm.gcd_em_test_vdd (Failed)

I think it worth taking a look at it doesn't seem to be some random flakiness.

src/Main.cc
gadfort: How does @QuantamHD handle the OpenROAD build for his bazel setup? He must have encountered a similar issue

maliberty: I believe they use a monorepo and build from source.

oharboe: I also saw some patching of OpenROAD going on, so my short answer is that it would take a bit of time to investigate what they are doing.

In the future, I'd like to see an officially maintained bazel build of OpenROAD with an artifact repository that I can use, just like Docker images. Would save tons of building time for anyone involved + be single source of truth + better chance of everyone running the exact same version w.r.t. bug reports and communicating in the community...

Perhaps bazel_hdl could use it too?

That's not happening tomorrow though...

oharboe:
@maliberty `ctest --test-dir build -j 32` had failures locally without https://github.com/The-OpenROAD-Project/OpenROAD/pull/5961

I propose to merge this as-is or that I wait for https://github.com/The-OpenROAD-Project/OpenROAD/pull/5961 to be merged, then merge with master and do a rebuild of this PR.

oharboe:
@maliberty Fixed!

maliberty (approved):

PR #5961, opened 2024-10-16
ppl: escape backslashes in python tests

htfab:
Python 3.12 generates a SynthesisWarning for strings containing unescaped `\[` or `\]`. This causes the test suite to fail, and indirectly blocks building .deb packages for Ubuntu 24.04 (which comes with Python 3.12 as the default).

### DIFF
--- a/src/ppl/test/group_pins7.py
+++ b/src/ppl/test/group_pins7.py
@@ -11,36 +11,36 @@ design.readDef("group_pins3.def")
 
 ppl_aux.set_io_pin_constraint(
     design,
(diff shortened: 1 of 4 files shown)


maliberty:
Thanks for the fix.  Note that commits need to signed with `git commit -s` to pass DCO.  Please see the details link next to DCO for instruction of fixing it for this PR.

maliberty (approved):

oharboe:
Thanks! When I run `ctest --test-dir build -j 32`, it fails without this fix.

PR #5968, opened 2024-10-16
odb: write int instead of size_t

gadfort:


stefanottili:
line 243 needs to be fixed too

gadfort:
@stefanottili done

gadfort:
@stefanottili @maliberty anything else let to do here?

maliberty (approved):

PR #5979, opened 2024-10-17
pad: ensure sorting order is stable when routing distances are equal

gadfort:
Fixes:
- platform instability cause by sorting changes


maliberty (approved):

maliberty:
@vvbandeira this solves the Ubuntu20/Debian11 issue.

PR #5982, opened 2024-10-18
colors: tweak default colors to work better on black background

oharboe:
Made some darker colors brighter and easier to see. A side-by-side review of all the colors needs to be done, I don't know exactly where to find them in the GUI though.

Before:

image

After:

image

### DIFF
--- a/src/gui/src/displayControls.cpp
+++ b/src/gui/src/displayControls.cpp
@@ -1914,16 +1914,16 @@ void DisplayControls::techInit(odb::dbTech* tech)
                            QColor(254, 0, 0),
                            QColor(9, 221, 0),
                            QColor(190, 244, 81),
-                           QColor(255, 69, 100),
+                           QColor(255, 165, 0),  // Metal 5
                            QColor(32, 216, 253),
                            QColor(253, 108, 160),
-                           QColor(150, 100, 255),
-                           QColor(200, 255, 100),
+                           QColor(117, 63, 194),
+                           QColor(128, 155, 49),
                            QColor(234, 63, 252),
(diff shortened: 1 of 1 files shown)


gadfort:
@oharboe it might be nice to have a side by side of all the color changes you are proposing

maliberty:
Just brightening it will make it confusing with metal2

oharboe:
Will have a look. Is that the only conflict?

maliberty:
How were these colors derived?  Please make a before and after comparison images.

oharboe:
> How were these colors derived? Please make a before and after comparison images.

I'm going to revert all color changes but the one that is most problematic. The others looked a bit darker to me too, but it is a bridge to far. The problem is to create lots of easy on the eye high contrast colors... Which is enormously tricky. :-)

oharboe:
Instead of changing multiple colors, I've changed only metal 5 and tried to introduce an additional different base color, bright orange, that works on black background.

Before:

image



After:

image

oharboe:
I'm going to try on more monitors, but on one either was OK, on my laptop the new scheme was definitely better.

gadfort:
Attempting to keep the color similar with higher contract with the adjacent layers: `QColor(222, 33, 96)`

image

This preserves the relative contract between M4 / M6 and increases the contrast with the background (although not as much as the orange).

oharboe:
Thanks! Will try on some more monitors.

oharboe:
@gadfort Used your proposed color. Works better on the monitors I tried. :+1:

maliberty (approved):

PR #5993, opened 2024-10-21
ppl: check if slot location is used by a fixed bterm

eder-matheus:
Fix overlaps between FIXED pins placed with the `place_pin` command and the remaining pins placed with `place_pins` command.


gadfort (approved):
@eder-matheus thanks. Fixes my testcase.

PR #5994, opened 2024-10-22
pdn: Fix flags documentation.

titan73:


titan73:
Actually -verbose flag of pdngen is deprecated.

src/pdn/README.md
```diff
@@ -168,7 +167,7 @@ define_power_switch_cell
     -ground ground_pin
     -name name
     -power unswitched_power_pin
-    -power_switchable power_switchable_pin
```
gadfort, line 171: the tcl code has this as `power_switchable`

gadfort (changes requested):

src/pdn/README.md
titan73: Fixed. Thanks.

gadfort (approved):
@titan73 try not to force-push after opening a PR it makes reviewing requested changed harder.

titan73:
> @titan73 try not to force-push after opening a PR it makes reviewing requested changed harder.

@gadfort What should I do? A commit for each update?

gadfort:
> > @titan73 try not to force-push after opening a PR it makes reviewing requested changed harder.
> 
> @gadfort What should I do? A commit for each update?

yeah, that preserves the diff history

titan73:
> yeah, that preserves the diff history
But this will pollute master history once PR is merged, right?

titan73:
> 
> yeah, that preserves the diff history

But this will pollute master history once PR is merged, right?

gadfort:
The PR could always be merged by squashing it. But in general it doesn't really matter. @maliberty can chime in if he has anything to add.

titan73:
Ah. I personally find the history important to follow the development of OR and learn from that but it's hard to read due to merge commits (https://github.com/The-OpenROAD-Project/OpenROAD/discussions/5830) and request change commits.

maliberty:
I agree with @gadfort about not force pushing as it disabled the "changes since last review" option in GH.

There are a bunch of constraints that have pushed us towards merge commits but I agree it makes the history harder to follow.  Perhaps we can resolve them and move to squash commits in the future.

titan73:
> I agree with @gadfort about not force pushing as it disabled the "changes since last review" option in GH.

Indeed. No very good.

> 
> There are a bunch of constraints that have pushed us towards merge commits but I agree it makes the history harder to follow. Perhaps we can resolve them and move to squash commits in the future.

No sure automatic squashing is good in all cases. This can be convenient for PR with big changes decomposed into several commits to keep them as is.

Maybe I can instead squash commits and force pushed after approval if the PR becomes a bit messy due to several updates.

maliberty:
If you want a squash commit you can request it and GH allows it as an option at PR merge.  Ideally squash would be the default and non-squash the exception.

PR #5995, opened 2024-10-22
utl: add mechanism to make it easier to handle operation-specific settings

AcKoucher:
Generic functionality that will be useful for the match footprint/user_function_class task in #5577.

### DIFF
--- a/src/rsz/include/rsz/Resizer.hh
+++ b/src/rsz/include/rsz/Resizer.hh
@@ -182,32 +182,17 @@ struct BufferData
 
 class OdbCallBack;
 
-// RAII mechanism to facilitate using operation specific settings, e.g.,
-// use only the clock buffers from cts when doing repairClkNets, don't
-// use buffers marked as "clock" in hold repairing, etc.
-// Can be instantiated in two ways:
-// - Before resizePreamble(): set the "filters" used by the preamble.
-// - After resizePreamble(): enforce the variables we want.
-class OperationSettings
+// RAII mechanism to facilitate using operation-specific settings.
+template <typename T>
+class SetAndRestore
 {
  public:
-  // Call before preamble:
-  OperationSettings(bool* exclude_clock_buffers,
-                    bool new_exclude_clock_buffers,
-                    sta::LibertyCellSeq* buffer_cells);
-
-  // Call after preamble:
-  OperationSettings(sta::LibertyCellSeq* buffer_cells,
-                    const sta::LibertyCellSeq& new_buffer_cells);
-
-  ~OperationSettings();
+  SetAndRestore(T& storage, const T& new_value);
+  ~SetAndRestore();
 
  private:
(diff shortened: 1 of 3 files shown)


src/rsz/src/Resizer.cc
```diff
@@ -2855,6 +2850,43 @@ void Resizer::rebufferNet(const Pin* drvr_pin)
...
+{
+  if (exclude_clock_buffers) {
+    exclude_clock_buffers_ = exclude_clock_buffers;
+    prev_exclude_clock_buffers_ = *exclude_clock_buffers_;
+    *exclude_clock_buffers_ = new_exclude_clock_buffers;
+  }
+
+  if (buffer_cells) {
+    buffer_cells_ = buffer_cells;
+    buffer_cells_->clear();
```
maliberty, line 2865: Why set and then clear?

src/rsz/include/rsz/Resizer.hh
```diff
@@ -182,6 +182,34 @@ struct BufferData
 
 class OdbCallBack;
 
+// RAII mechanism to facilitate using operation specific settings, e.g.,
+// use only the clock buffers from cts when doing repairClkNets, don't
+// use buffers marked as "clock" in hold repairing, etc.
+// Can be instantiated in two ways:
+// - Before resizePreamble(): set the "filters" used by the preamble.
+// - After resizePreamble(): enforce the variables we want.
+class OperationSettings
```
maliberty, line 191: Can this be written in a more generic fashion like
```
template<typename T>
class SetAndRestore
{
   SetAndRestore(const T& new_value, const T& storage); // store current value in storage and assign it to new_value
  ~SetAndRestore(); // assign storage back to its previous value
};
```
Then apply it to the specific cases of exclude_clock_buffers and buffer_cells.

src/rsz/src/Resizer.cc
AcKoucher: The idea would be that for this use case we only clear the RSZ buffer list for the subsequent preamble to populate it.
We need to set the pointer so that the destructor can clear it again for the next operation.

src/rsz/include/rsz/Resizer.hh
```diff
@@ -182,6 +182,19 @@ struct BufferData
 
 class OdbCallBack;
 
+// RAII mechanism to facilitate using operation-specific settings.
```
maliberty, line 185: This seems generic enough to put in utl.  You should put the method definitions in the header as well.

src/utl/include/utl/scope.h
```diff
@@ -0,0 +1,58 @@
+/////////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2020, The Regents of the University of California
```
maliberty, line 3: 2024 PII

maliberty (approved):

PR #6002, opened 2024-10-22
psm: sort connection maps

gadfort:
@maliberty this is an attempt to stabilize the PSM em tests.

Changes:
- use Compare for all maps keyed on Connection*


src/psm/test/aes_asap7_vdd.ok
```diff
@@ -23,7 +23,7 @@ Corner           : default
 Supply voltage   : 7.00e-01 V
 Worstcase voltage: 7.00e-01 V
 Average voltage  : 7.00e-01 V
-Average IR drop  : 3.02e-06 V
-Worstcase IR drop: 7.57e-06 V
+Average IR drop  : 3.11e-06 V
+Worstcase IR drop: 7.80e-06 V
```
maliberty, line 27: Is a change in results expected?

gadfort: I believe this is just a result of changing summation order slightly (this test consumes almost no power so a lot of the numbers are tiny which seems to introduce some strange behavior).

maliberty (approved):

PR #6009, opened 2024-10-23
psm: limit spice digits

gadfort:


maliberty (approved):
@vvbandeira  @eder-matheus hopefully this make the psm tests stable 🤞

PR #6015, opened 2024-10-24
pdn: Issue an error when -starts_with is used with -followpin in add_pdn_strip.

titan73:
Currently it is ignored when -followpin is set. Better issue an error and point to the correct way.

### DIFF
--- a/src/pdn/src/pdn.tcl
+++ b/src/pdn/src/pdn.tcl
@@ -377,8 +377,7 @@ proc add_pdn_stripe { args } {
 
   if { [info exists flags(-followpins)] } {
     if { [info exists keys(-starts_with)] } {
-      utl::error PDN 211 "Option -starts_with cannot be used with -followpins.\
-        Flip sites when initializing floorplan to change followpin power/ground order."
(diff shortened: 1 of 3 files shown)


src/pdn/src/pdn.tcl
```diff
@@ -376,6 +376,10 @@ proc add_pdn_stripe { args } {
   }
 
   if { [info exists flags(-followpins)] } {
+    if { [info exists keys(-starts_with)] } {
+      utl::error PDN 211 "Option -starts_with cannot be used with -followpins.\
+        Flip sites when initializing floorplan to change followpin power/ground order."
+    }
```
gadfort, line 382: I think a warning is sufficient. No need to add the "Flip sites when initializing floorplan to change followpin power/ground order."

titan73: I remove "Flip sites..." also in the doc?

gadfort: I think it's fine there.

gadfort: sure, that would be fine

gadfort (approved):
LGTM

PR #6018, opened 2024-10-24
rsz: add option to honor cell footprint in repair_design and repair_timing

AcKoucher:
Resolve one of the tasks in #5577.

New flags in repair_design and repair_timing needed for @precisionmoon proposal.
We'll have a new ORFS varialbe HONOR_CELL_FOOTPRINT that will turn on these flags.

### DIFF
--- a/src/rsz/src/Resizer.cc
+++ b/src/rsz/src/Resizer.cc
@@ -450,11 +450,8 @@ void Resizer::balanceBin(const vector<odb::dbInst*>& bin,
       }
       Instance* sta_inst = db_network_->dbToSta(inst);
       LibertyCell* cell = network_->libertyCell(sta_inst);
-      LibertyCellSeq* equiv_cells = sta_->equivCells(cell);
-      if (!equiv_cells) {
-        continue;
-      }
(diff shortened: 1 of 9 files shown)


src/rsz/src/Resizer.cc
```diff
@@ -2911,8 +2942,11 @@ int Resizer::holdBufferCount() const
 }
 
 ////////////////////////////////////////////////////////////////
-void Resizer::recoverPower(float recover_power_percent)
+void Resizer::recoverPower(float recover_power_percent,
+                           bool match_cell_footprint)
 {
+  utl::SetAndRestore<bool> set_match_footprint(match_cell_footprint_,
+                                               match_cell_footprint);
```
QuantamHD, line 2949: nit: I think this SetAndRestore utility is actually harder to read than the equivalent tmp, set and restore stuff.

maliberty: It is more exception safe (and early return).

QuantamHD: Fair on both points. :)

maliberty (approved):

PR #6022, opened 2024-10-24
gui: use base descriptor class to remove duplicated code for lessThan, makeSelected, and populateODBProps

gadfort:

The last time I was adding a ODB descriptor I noticed that I was just copying and pasting the makeSelected, lessThan, and forgot to add the ODB properties to the getProperties.

So this is an attempt to create a base class that can do these things and remove that duplicated code.


maliberty:
Nice removal of duplicated code.  I'll approve once the clang-tidy is cleaned up.

gadfort:
@maliberty done

PR #6028, opened 2024-10-26
ppl/test: Suppress messages that are only in tcl.

titan73:
So no need to suppress them in python. Improve comment to be clearer.
Id 48 appears in both sides so no need to suppress it either.  Update ok files accordingly.

### DIFF
--- a/src/ppl/src/IOPlacer.cpp
+++ b/src/ppl/src/IOPlacer.cpp
@@ -1691,17 +1691,16 @@ void IOPlacer::addTopLayerConstraint(PinSet* pins, const odb::Rect& region)
     }
   }
 
-  logger_->info(
-      utl::PPL,
-      60,
-      "Restrict pins [ {}] to region ({:.2f}u, {:.2f}u)-({:.2f}u, {:.2f}u) at routing layer {}.",
-      pin_names,
-      getBlock()->dbuToMicrons(region.xMin()),
-      getBlock()->dbuToMicrons(region.yMin()),
-      getBlock()->dbuToMicrons(region.xMax()),
-      getBlock()->dbuToMicrons(region.yMax()),
-      getTopLayer()->getConstName()
-  );
+  logger_->info(utl::PPL,
+                60,
+                "Restrict pins [ {}] to region ({:.2f}u, {:.2f}u)-({:.2f}u, "
+                "{:.2f}u) at routing layer {}.",
+                pin_names,
+                getBlock()->dbuToMicrons(region.xMin()),
+                getBlock()->dbuToMicrons(region.yMin()),
+                getBlock()->dbuToMicrons(region.xMax()),
+                getBlock()->dbuToMicrons(region.yMax()),
+                getTopLayer()->getConstName());
 
   constraints_.push_back(constraint);
   for (odb::dbBTerm* bterm : *pins) {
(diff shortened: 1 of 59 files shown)


test/helpers.tcl
```diff
@@ -203,18 +203,16 @@ suppress_message ODB 127
...
 suppress_message ODB 134
 
-# suppress ppl info messages. The ones defined in tcl can never
-# match between tcl and Python
+# suppress ppl info messages which are only present in tcl
 suppress_message PPL 41
-suppress_message PPL 48
-suppress_message PPL 67
 suppress_message PPL 60
+suppress_message PPL 67
```
maliberty, line 209: There appears to be no message 67.

test/helpers.tcl
```diff
@@ -203,18 +203,16 @@ suppress_message ODB 127
 # Finished DEF file
 suppress_message ODB 134
 
-# suppress ppl info messages. The ones defined in tcl can never
-# match between tcl and Python
+# suppress ppl info messages which are only present in tcl
 suppress_message PPL 41
```
maliberty, line 207: 41 doesn't seem like it would help any user and should be removed

test/helpers.tcl
```diff
@@ -203,18 +203,16 @@ suppress_message ODB 127
...
-suppress_message PPL 48
-suppress_message PPL 67
 suppress_message PPL 60
```
maliberty, line 208: 60 should be moved to c++ like 48.

maliberty (changes requested):
There should be no difference in messaging between tcl & python.  I've suggested some things to equalize them.

test/helpers.tcl
titan73: Seems there are when removing the suppression, here an example:
```
** /home/titan/OpenROAD/src/ppl/test/add_constraint1.ok	2024-10-12 13:35:34.649208359 +0200
--- /home/titan/OpenROAD/src/ppl/test/results/add_constraint1-tcl.log	2024-10-26 17:10:29.250298762 +0200
***************
*** 3,8 ****
--- 3,9 ----
  [INFO ODB-0131]     Created 88 components and 422 component-terminals.
  [INFO ODB-0133]     Created 54 nets and 88 connections.
+ [INFO PPL-0067] Restrict INPUT pins to region 0.0u-100.13u, in the top edge.
  Found 0 macro blocks.
  [INFO PPL-0001] Number of slots           2494
  [INFO PPL-0002] Number of I/O             54
```

titan73:
Sorry for the force push. I addressed msg 41 & 60.
Let me know for 67.

src/ppl/src/IOPlacer.cpp
```diff
@@ -1672,13 +1690,44 @@ void IOPlacer::addDirectionConstraint(Direction direction,
...
+  int pin_cnt = 0;
+  for (odb::dbBTerm* pin : *pins) {
+    pin_names += pin->getName() + " ";
+    pin_cnt++;
+    if (pin_cnt >= pins_per_report_
+        && !logger_->debugCheck(utl::PPL, "pin_groups", 1)) {
+      pin_names += "... ";
+      break;
+    }
+  }
```
maliberty, line 1718: This same block of code already occurs twice.  Please make a method rather than add a third copy.

maliberty (approved):

PR #6040, opened 2024-10-28
odb/cdl: Handle top-level pins wired to nets of different name

smunaut:
ATM the code seem to assume that for each top level pin, the net name is equal to the pin name.

That's not always the case and this produces an incorect netlist.

To solve that, during pin name enumeration when creating the subckt line, we create a mapping from "net name" -> "pin name" for each pin connected to a net of a different name. Then when outputting the connections between sub blocks, we use the pin name instead of the DB net name.

### DIFF
--- a/src/odb/src/cdl/cdl.cpp
+++ b/src/odb/src/cdl/cdl.cpp
@@ -105,6 +105,11 @@ std::string getNetName(dbBlock* block,
     return getUnconnectedNet(block, unconnectedNets);
   }
 
+  dbBTerm* bterm = net->get1stBTerm();
+  if (bterm) {
+    return bterm->getName();
+  }
+
   return net->getName();
 }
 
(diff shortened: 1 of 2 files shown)


maliberty:
Please add a unit test that demonstrates this use case.

maliberty:
I don't see a full test, just

image

If no ok needs updating it means this change has no effect on any test.

maliberty (approved):

src/odb/src/cdl/cdl.cpp
```diff
@@ -105,7 +106,14 @@ std::string getNetName(dbBlock* block,
     return getUnconnectedNet(block, unconnectedNets);
   }
 
-  return net->getName();
+  std::string net_name = net->getName();
+
+  auto pin_name_match = net2pin.find(net_name);
+  if (pin_name_match != net2pin.end()) {
+    return pin_name_match->second;
+  }
```
maliberty, line 114: Rather than build net2pin and lookup, you could just handle the bterm here, eg
```
dbBTerm* bterm = net->get1stBTerm();
if (bterm) {
  return bterm->getName();
}
return net->getName();
```

smunaut:
* Rebased to master
* Implemented the more elegant method above
* Checked that the new test doesn't pass without the patch and passes with the patch.

PR #6046, opened 2024-10-29
report_design_area: allow redirection to file

oharboe:
same as all other commands, don't know why this one slipped between the cracks

### DIFF
--- a/src/rsz/src/Resizer.tcl
+++ b/src/rsz/src/Resizer.tcl
@@ -637,7 +637,7 @@ proc repair_timing { args } {
 
 sta::define_cmd_args "report_design_area" {}
 
-proc report_design_area { args } {
+sta::proc_redirect report_design_area { args } {
   sta::parse_key_args "report_design_area" args keys {} flags {}
   set util [format %.0f [expr [rsz::utilization] * 100]]
   set area [sta::format_area [rsz::design_area] 0]
--- a/src/rsz/src/Resizer.tcl
+++ b/src/rsz/src/Resizer.tcl
@@ -637,7 +637,7 @@ proc repair_timing { args } {
 
 sta::define_cmd_args "report_design_area" {[> filename] [>> filename]} ;# checker off
 
-sta::proc_redirect report_design_area { args } {
+sta::proc_redirect report_design_area {
   sta::parse_key_args "report_design_area" args keys {} flags {}
   set util [format %.0f [expr [rsz::utilization] * 100]]
   set area [sta::format_area [rsz::design_area] 0]
--- a/src/rsz/src/Resizer.tcl
+++ b/src/rsz/src/Resizer.tcl
@@ -635,7 +635,7 @@ proc repair_timing { args } {
 
 ################################################################
 
-sta::define_cmd_args "report_design_area" {}
+sta::define_cmd_args "report_design_area" {[> filename] [>> filename]} ;# checker off
 
 sta::proc_redirect report_design_area { args } {
   sta::parse_key_args "report_design_area" args keys {} flags {}


src/rsz/src/Resizer.tcl
```diff
@@ -637,7 +637,7 @@ proc repair_timing { args } {
 
 sta::define_cmd_args "report_design_area" {}
 
-proc report_design_area { args } {
```
gadfort, line 640: you need to update the define_cmd_args to reflect this change.
See: https://github.com/The-OpenROAD-Project/OpenROAD/blob/0fc26a3a3a4db859d6fca00af71ca8d132003a2f/src/rsz/src/Resizer.tcl#L647

oharboe: Getting some errors.

```
Running tapcell.tcl, stage 2_4_floorplan_tapcell
Error: TCL init script: invalid command name "wrong # args: should be "sta::proc_redirect proc_name body"
    while executing
"sta::proc_redirect report_design_area { args } {
```

gadfort: looks like the proc_redirect indirectly defined `args`:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/0fc26a3a3a4db859d6fca00af71ca8d132003a2f/src/rsz/src/Resizer.tcl#L649

I think you just need to change it to `sta::proc_redirect report_design_area {`

oharboe: After a bit more hacking and 

```
find tools/OpenROAD/ -regex .*tclInitVar.*.cc -exec rm {} \;
./build_openroad.sh --local --no_init
```

It compiles.

However "make do-2_1_floorplan" does not redirect yet...

```
$ make open_floorplan
ODB_FILE=./results/nangate45/gcd/base/2_floorplan.odb /home/oyvind/OpenROAD-flow-scripts/tools/install/OpenROAD/bin/openroad -no_init -threads 48  /home/oyvind/OpenROAD-flow-scripts/flow/scripts/open.tcl
OpenROAD v2.0-16809-g23f5678bc 
Features included (+) or not (-): +Charts +GPU +GUI +Python : Debug
[INFO ORD-0030] Using 48 thread(s).
GUI_TIMING=1 reading timing, takes a little while for large designs...
Populating timing paths...
openroad> report_design_area >ll.txt
Design area 663 u^2 58% utilization.
openroad> report_design_area >>ll.txt
Design area 663 u^2 58% utilization.
openroad> 

$ cat ll.txt 
```

gadfort: The issue seems to be that using `utl::report` bypasses the file redirect.
The same issue appears in the `report_floating_nets` it's possible that the redirect doesn't work correctly with the logger.

gadfort: Yeah, It would seem that this would be an issue in several places. If you want this function fixed you could change it to a `puts` instead of `utl::report` that ensures the text goes to the file redirect.

gadfort: For the record: `report_long_wires`, `report_floating_nets` are the only two currently effected by this.

oharboe: I'm sitting tight for advice here, I have no idea what the intention is or what would be consistent in this code.

gadfort: I think the fix needs to be in a different PR (you might want to open an issue for tracking).
While this code doesn't work as intended until the underlying issue is resolved I don't think there is any harm in merging this in (just means to be able to use it it might take some time)

oharboe: One concern at the time, yes.

oharboe: @maliberty Can you check if it is a quick fix to address what @gadfort found or articulate a github issue? I'm a little bit vague on exactly what the problem is.

gadfort: it's somewhat involved...
The underlying issue is here: https://github.com/The-OpenROAD-Project/OpenROAD/blob/a5e71280688962b43cdee4a24d38da08089594b2/src/dbSta/src/dbSta.cc#L622
Because the tcl code and STA directs all it's prints to this function we correctly redirect the code, in openroad, we use the logger directly and thus this is broken.

oharboe:
@maliberty Let me know if you want to move forward with this

PR #6049, opened 2024-10-29
odb: In dbITerm::getGeometries, return layer + rect

smunaut:
Currently only the rect is returned but that's not very useful. So instead we return pairs of layer + rect.

The appropriate SWIG mapping is also added so the result is usable both in Python and in TCL.

Note that someone should definitely have a close look at the SWIG mapping because I have very little experience with those.
I did test that it seems to work as expected in both TCL and Python though.

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -3636,7 +3636,8 @@ class dbITerm : public dbObject
   ///
   /// Returns all geometries of all dbMPin associated with
   /// the dbMTerm.
-  std::vector<Rect> getGeometries() const;
+  ///
+  std::vector<std::pair<dbTechLayer*, Rect>> getGeometries() const;
 
   void setAccessPoint(dbMPin* pin, dbAccessPoint* ap);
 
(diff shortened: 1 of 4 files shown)


gadfort:
Could this return a set of dbShape instead? https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/odb/include/odb/dbShape.h

maliberty:
Looking your suggestion I noticed dbITermShapeItr which might be what is wanted here.

maliberty:
@smunaut this has compilation and clang-tidy errors?  Do you plan to fix this up?

smunaut:
Ok, I think I fixed it actually.  For some reason it says `pr-hread` failed but if you click on it, all build succeeded. I think it got a bit confused because I pushed a new version before the previous build was done ?

maliberty:
You need to merge master to resolve the c++20 compiler issue.

smunaut:
Rebased to master, all checks passed.

maliberty (approved):

PR #6052, opened 2024-10-29
rsz: add detailed_routing as parasitics source

eder-matheus:
This first step only adds `detailed_routing` as a new source of parasitics, and is used mainly to avoid calculating the parasitics from wire load models.

In the first iteration on each net, the droute parasitics are used, but once the net is changed, later iterations will use the global routing parasitics for estimation. At some point we should be able to detail route again specific nets inside the repair_timing flow.

Last, I'm checking if the design is routed to define the parasitics source as detailed_routing. It doesn't seem the ideal way, but I didn't find any functions in STA to detect there are parasitics present in its internal structures. Another option would be have a command that the user explicitly calls to define detailed_routing as the parasitics source.

I'm open to suggestions on how to improve this part of the code.

### DIFF
--- a/src/rsz/src/EstimateWireParasitics.cc
+++ b/src/rsz/src/EstimateWireParasitics.cc
@@ -349,7 +349,8 @@ void Resizer::updateParasitics(bool save_guides)
       }
       parasitics_invalid_.clear();
       break;
-    case ParasiticsSrc::global_routing: {
+    case ParasiticsSrc::global_routing:
+    case ParasiticsSrc::detailed_routing: {
       incr_groute_->updateRoutes(save_guides);
       for (const Net* net : parasitics_invalid_) {
         global_router_->estimateRC(db_network_->staToDb(net));
@@ -357,7 +358,6 @@ void Resizer::updateParasitics(bool save_guides)
       parasitics_invalid_.clear();
       break;
     }
-    case ParasiticsSrc::detailed_routing:
     case ParasiticsSrc::none:
       break;
   }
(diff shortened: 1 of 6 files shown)


src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -294,6 +294,9 @@ void Resizer::estimateParasitics(ParasiticsSrc src,
       global_router_->estimateRC(spef_writer.get());
       parasitics_src_ = ParasiticsSrc::global_routing;
       break;
+    case ParasiticsSrc::detailed_routing:
+      parasitics_src_ = ParasiticsSrc::detailed_routing;
+      break;
```
maliberty, line 299: Is it right that this does no estimation?

src/rsz/src/EstimateWireParasitics.cc
```diff
@@ -391,6 +397,8 @@ void Resizer::ensureWireParasitic(const Pin* drvr_pin, const Net* net)
         parasitics_invalid_.erase(net);
         break;
       }
+      case ParasiticsSrc::detailed_routing:
+        break;
```
maliberty, line 401: no action?

maliberty:
Even if you could ask sta if it has parasitics, it wouldn't know the source of them.

src/rsz/src/EstimateWireParasitics.cc
eder-matheus: The way I'm implementing it on ORFS, we already load the SPEF file after writing it with RCX, so no estimation is needed. Perhaps in the future we want to call rcx and load the spef data only in C++, and here would be a correct place for it.

Maybe I can add a comment about these future plans here?

src/rsz/src/EstimateWireParasitics.cc
eder-matheus: Similar comment here. The idea is that the parasitics are already loaded in sta.

PR #6058, opened 2024-10-30
drt: Fix pin double count and not appearing info.

titan73:
Regression from 58b46bbc5ac5a206af201b991e0fcd5c8e7f0ade.
Info message was never shown as pin count was always odd when performing the modulo.
Also simplify code even more.


titan73:
Note that before https://github.com/The-OpenROAD-Project/OpenROAD/commit/58b46bbc5ac5a206af201b991e0fcd5c8e7f0ade. 
pin count was 20095 and after 29836 (double count printed 59672). Hope that was expected.

titan73:
There are similar code I'll simplify in another PR.

osamahammad21 (approved):

osamahammad21:
@bnmfw FYI

Issue #6064, opened 2024-10-30
RSZ: extract parasitics for repair timing and repair design inside rsz code


eder-matheus:
### Description

Follow-up on PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6052.

The current implementation assumes that the parasitics will be extracted before running the repair command, so the estimation part of the flow is doing nothing when the design is routed.

### Suggested Solution

Call parasitcs extraction inside the estimation code, when the parasitics source is detailed routing. This includes extracting the parasitics and feeding them to sta.

PR #6068, opened 2024-10-30
rsz: extract detailed routing parasitics inside estimate_parasitics code

eder-matheus:
Part one to fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/6064.


src/rsz/README.md
```diff
@@ -115,15 +115,17 @@ Standard Parasitic Exchange Format.
...
     [-spef_file spef_file]
+    [-ext_model_file]
 ```
 
 #### Options
 
 | Switch Name | Description |
 | ----- | ----- |
-| `-placement` or `-global_routing` | Either of these flags must be set. Parasitics are estimated based after placement stage versus after global routing stage. |
+| `-placement` or `-global_routing` or `-detailed_routing` | Either of these flags must be set. Parasitics are estimated based after placement stage versus after global routing stage, or extracted from the detailed routing. |
```
maliberty, line 127:
```suggestion
| `-placement` or `-global_routing` or `-detailed_routing` | One of these flags must be set. Parasitics are estimated based on placement, global routing, or detailed routing. |
```

src/rsz/README.md
```diff
@@ -115,15 +115,17 @@ Standard Parasitic Exchange Format.
 
 ```tcl
 estimate_parasitics
-    -placement|-global_routing
+    -placement|-global_routing|-detailed_routing
     [-spef_file spef_file]
+    [-ext_model_file]
```
maliberty, line 120:
```suggestion
    [-ext_model_file model_file]
```

src/rsz/README.md
```diff
@@ -115,15 +115,17 @@ Standard Parasitic Exchange Format.
...
 ```
 
 #### Options
 
 | Switch Name | Description |
 | ----- | ----- |
-| `-placement` or `-global_routing` | Either of these flags must be set. Parasitics are estimated based after placement stage versus after global routing stage. |
+| `-placement` or `-global_routing` or `-detailed_routing` | Either of these flags must be set. Parasitics are estimated based after placement stage versus after global routing stage, or extracted from the detailed routing. |
+| `-ext_model_file` | Optional. File name to load the extraction model file for parasitics extraction from detailed routing. |
```
maliberty, line 128:
```suggestion
| `-ext_model_file model_file` | Optional. File name to load the extraction model file for parasitics extraction from detailed routing. |
```

maliberty (changes requested):

src/rsz/README.md
gadfort: how does this handle corners? they have separate model files.

eder-matheus: In my tests, I'm using only the public PDKs available in ORFS, which use only on model file and one corner for parasitics extraction.

How are you using multiple corners in your flow? Do you loop over each corner to extract the parasitics for them?

gadfort: Yes, we loop over them to generate a different spef file per corner and read them in per corner to be able to get timing across corners (it's clumsy, but it's the interface we have in OpenROAD at the moment). I think corner support is needed for this to be a complete/correct solution.

gadfort: If rcx was augmented to allow us to set the model files per corner that would also allow us to avoid adding this argument and just ask rcx to generate the parasitics

maliberty:
@gadfort do you have a snippet of code that shows how you set it up today for rcx & sta?

gadfort:
@maliberty https://github.com/siliconcompiler/siliconcompiler/blob/adb451bc72eb0f9b8a863ddf192a164b7f0b0e25/siliconcompiler/tools/openroad/scripts/sc_export.tcl#L40-L69
First few lines handle generate unique SPEFs for each corner and then the last bit just reads them in per corner.
In most other tools, you just define a pexcorner and associate the extraction file with that and then when you call for parasitics these are used. usually write_spef is available and takes a corner input to allow you to dump any corner you want.

gadfort:
I should note that it looks like rcx could "handle" corners if they are in the RCX file, but the current conversion flows we have don't support creating a single file with multiple corners (no clue what that would look like anyways), most PDKs separate the corners for PEX

src/rsz/README.md
eder-matheus: Maybe this option can be updated to take multiple files, and them the implementation will perform parasitics extraction for each of them. I believe sta have the support to have the parasitics stored for multiple corners, so when running repair_timing it would look at all of them.

gadfort: @eder-matheus it seems to me like modifying RCX to store the model files would require a much less complex argument to this command. and yes, STA fully supports corners.

eder-matheus:
The current implementation of RCX doesn't allow reading multiple model files at once. Since we have a big update ongoing on RCX, I think we should not implement new features in this tool until they are merged.

Regarding my implementation here, I believe we could merge what we have now but keep track of the required updates in RCX regarding multiple model files. I will add comments in my code to track what parts needs to be updated once RCX has full support for multiple model files.

dimitrisfotakis:
There is no need to support multiple model files. After the latest merged PR, rcx can create a single model file with multiple corners in it. The "extract" command can define which corners need to be extracted and stored in db. "write_spef" can write a single SPEF file with all extracted corners or a specific corner. See relevant documentation in README files in src/rcx/docs dirs. I am in the process of a PR providing regression tests of corners.

maliberty:
@eder-matheus still relevant?

eder-matheus:
> @eder-matheus still relevant?

This would require a large rework. I'll close this and evaluate the best path to bring this project back.

Issue #6073, opened 2024-10-31
GRT: store route guides even when it fails with congestion


eder-matheus:
### Description

Today we don't have the option to see route guides of nets when the design has congestion. This makes debug harder, since having only the markers with the congestion regions don't say much about the routing each net is taking.

### Suggested Solution

Complete the global route flow even when routing congestion is detected, and store the guides in odb.
The detailed router needs to be updated too, in order to avoid routing a congested design.

gadfort:
Wouldn't it be better to allow GRT to mark the guides as "dirty" or "unusable" that way DRT only needs to check a simple flag instead of needing to check for congestion (since we have flags that allow for GRT to finish with congestion and DRT is capable of handling a little bit of congestion)

eder-matheus:
> Wouldn't it be better to allow GRT to mark the guides as "dirty" or "unusable" that way DRT only needs to check a simple flag instead of needing to check for congestion (since we have flags that allow for GRT to finish with congestion and DRT is capable of handling a little bit of congestion)

I'm planning to use the congestion markers stored in odb to detect that. In the final implementation, DRT would need only a single function call to check that.

PR #6076, opened 2024-10-31
grt: complete route flow before failing due to congestion

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/6073.

* When the design has congestion, grt will complete its flow and save the guides in odb before the error.
* The -allow_congestion flag is preserved for cases where the user wants to go through drt even with congestion in grt.
* odb::dbGuide was updated with the is_congested flag. This is used inside drt to ensure that it will not route a design that has global routing congestion. However, when the -allow_congestion flag is used in grt, this flag is set to false.


eder-matheus:

src/odb/src/db/dbJournal.cpp
```diff
@@ -1544,18 +1544,21 @@ void dbJournal::undo_deleteObject()
...
       uint via_layer_id;
+      bool is_congested;
       _log.pop(net_id);
       _log.pop(x_min);
       _log.pop(y_min);
       _log.pop(x_max);
       _log.pop(y_max);
       _log.pop(layer_id);
       _log.pop(via_layer_id);
+      _log.pop(is_congested);
```
maliberty, line 1555: I don't see this being pushed in the destroy method

src/grt/src/GlobalRouter.cpp
```diff
@@ -4771,7 +4759,7 @@ std::vector<Net*> GlobalRouter::updateDirtyRoutes(bool save_guides)
         saveCongestion();
         logger_->error(GRT,
                        232,
-                       "Routing congestion too high. Check the congestion "
+                       "Routing congestion too high. Check the congested "
```
maliberty, line 4762: I think `congestion` is better

maliberty (changes requested):

src/odb/src/db/dbJournal.cpp
eder-matheus: Where exactly it should be pushed? I was following the same logic I did with via_layer_id and I'm not sure where whould I do that.

maliberty: dbGuide::destroy

maliberty: You did add 
```
    block->_journal->pushParam(_guide->via_layer_);
```
previously.

eder-matheus: Thanks! Fixed it now.

maliberty:
Fix "Check That ODB Files Are Generated "

eder-matheus:
> Fix "Check That ODB Files Are Generated "

Fixed.

maliberty (approved):

Issue #6102, opened 2024-11-05
MPL2: Remove flip perturbations from Hard annealing


AcKoucher:
### Description

Now that we have the orientation improvement step, it doesn't make sense to spend perturbations with this. I.e., is deprecated.

### Suggested Solution

Remove flip perturbations and adapt the probabilities.
Extra credits: Group probabilities in a struct.

joaomai:
Closed by PR #8082.

PR #6103, opened 2024-11-05
mpl2: use best valid result for both Hard/Soft

AcKoucher:
Make the Result live inside the Core so that we can use the same struct for both types of Macro.
For each type there's a different type of result now:

Soft -> Sequence Pair + Shapes
Hard -> Sequence Pair (I'm not including the orientation, because the flip perturbations in SA are now deprecated and should be removed. See #6102 )


AcKoucher:
Running Secure-CI.

AcKoucher:
Secure-CI is fine.

maliberty (approved):

PR #6106, opened 2024-11-06
etc: Major runtime impact due to find at system root for large and ne…

rpriyadarshi:
Major runtime impact due to find at system root for large and network mounts took hours to complete if left alone. More likely, the user would terminate the script thinking something is wrong.

1. Added a wait cursor to give visual feedback to the user.
2. Now mitigates runtime using locate if available.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -4,6 +4,143 @@ set -euo pipefail
 
 CMAKE_PACKAGE_ROOT_ARGS=""
 
+# Function to display usage instructions
+_searchUsage() {
+    echo "Usage: $0 [--use-find] [--update-db] --file-pattern <filename-pattern>"
+    echo "Options:"
+    echo "  --use-find      Force the use of 'find' even if 'locate' or 'plocate' is available"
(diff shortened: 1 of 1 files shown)


etc/DependencyInstaller.sh
```diff
@@ -274,7 +411,7 @@ _installOrTools() {
 
     # Disable exit on error for 'find' command, as it might return non zero
     set +euo pipefail
-    LIST=($(find / -type f -name "libortools.so*" 2>/dev/null))
+    _searchEverywhere --update-db --file-pattern "libortools.so*"
```
maliberty, line 414: --update-db will trigger `sudo updatedb` even if locate isn't installed.  This should not be required in that case.

vvbandeira:
@rpriyadarshi
Does the current line that searches fewer places works for you? Or would this still be too onerous?
https://github.com/The-OpenROAD-Project/OpenROAD/blob/a634a79ab89af38fb736c5c41e67f9e6fe2c63d8/etc/DependencyInstaller.sh#L294

maliberty:
@rpriyadarshi ping

maliberty:
no activity

PR #6122, opened 2024-11-08
rsz: Add command for eliminating dead logic

povik:
Removes instances which are all:

 * standard logic cells
 * not marked dont touch, not connected to a net marked dont touch
 * not in the transitive input cone of a top-level port, or a dont touch cell, dont touch net, or a cell which is not a standard logic cell

In ORFS on nangate45/black_parrot, which has hierarchical synthesis enabled, this finds plenty of cells to eliminate.

At floorplan stage, pre-elimination

```
Cell type report:                       Count       Area
  Macro                                    24  328372.48
  Tap cell                               9417    2504.92
  Timing Repair Buffer                  26827   39861.70
  Inverter                              33586   19516.42
  Sequential cell                       38477  174264.85
  Multi-Input combinational cell       200734  208349.55
  Total                                309065  772869.92
```
post-elimination
```
Removed 40566 instances
Cell type report:                       Count       Area
  Macro                                    24  328372.48
  Tap cell                               9417    2504.92
  Timing Repair Buffer                  24933   36703.21
  Inverter                              30144   17674.37
  Sequential cell                       34168  154779.55
  Multi-Input combinational cell       169813  177765.94
  Total                                268499  717800.47
```


src/rsz/src/Resizer.i
```diff
@@ -825,6 +825,14 @@ set_parasitics_src(ParasiticsSrc src)
   resizer->setParasiticsSrc(src);
 }
 
+void
+eliminate_dead_logic_cmd()
```
maliberty, line 829: A tcl wrapper should be provided.

src/rsz/src/Resizer.cc
```diff
@@ -3648,4 +3648,79 @@ void Resizer::resetInputSlews()
   input_slew_map_.clear();
 }
 
+void Resizer::eliminateDeadLogic()
+{
+  std::vector<const Instance*> queue;
```
maliberty, line 3653: std::queue ?

src/rsz/src/Resizer.cc
```diff
@@ -3648,4 +3648,79 @@ void Resizer::resetInputSlews()
...
+        Net* net;
+        if ((net = network_->net(pin)) && dontTouch(net)) {
+          keepInst(inst);
+        }
+      }
+      delete iter;
+    }
+  }
+
+  for (int i = 0; i < queue.size(); i++) {
```
maliberty, line 3708: With a queue this would be a more natural while !empty() and pop()

src/rsz/src/Resizer.cc
```diff
@@ -3648,4 +3648,79 @@ void Resizer::resetInputSlews()
...
+  }
+
+  int remove_count = 0;
+  for (auto inst : network_->leafInstances()) {
+    if (!kept_instances.count(inst)) {
+      sta_->deleteInstance((Instance*) inst);
+      remove_count++;
+    }
+  }
+  logger_->report("Removed {} instances", remove_count);
```
maliberty, line 3723: You may have floating nets left behind that should also be cleaned up.

maliberty (changes requested):

src/rsz/src/Resizer.cc
povik: Needlessly complicated to me, we don't need FIFO here

oharboe:
It would be interesting to run this on megaboom. megaboom has a 50% reduction in instances after hierarchical synthesis if the same(?) is done by naja edit.

@jeffng-or @maliberty FYI

maliberty:
I think naja also does constant propagation which could be another item of interest in OR.

src/rsz/src/Resizer.cc
maliberty: Leaving a load with no driver is an error.  Under what circumstance would that happen?

povik: What I am saying is: `find_floating_nets` only detects nets which have loads but no drivers, so the kind of nets this new elimination command is leaving (which have both no drivers and no loads) are not picked up by it.

maliberty: `find_floating_nets` is just for reporting and doesn't clean anything up.  I suppose it is a matter of definition but I would include a net with no connections at all as floating.  It might not strictly be a violation but it is useless and makes little sense.

Regardless I think we should remove such nets.

povik: > find_floating_nets is just for reporting and doesn't clean anything up.

I am not claiming otherwise.

> Regardless I think we should remove such nets.

OK

povik:
@maliberty tooling is happy now (expect for CI which seems to fail for an external reason)

povik:
> constant propagation which could be another item of interest in OR.

I see OpenSTA does some constant propagation for its own purposes, the result of which is available under `sta_->simLogicValue(Pin*)`, but I don't think we can reuse it directly

Some of the constants originate from tie-low/tie-high cells, but there are also constants sourced from SDC:
```
void
Sim::seedConstants()
{
  // Propagate constants from inputs tied hi/low in the network.
  enqueueConstantPinInputs();
  // Propagate set_LogicValue::zero, set_LogicValue::one, set_logic_dc constants.
  setConstraintConstPins(sdc_->logicValues());
  // Propagate set_case_analysis constants.
  setConstraintConstPins(sdc_->caseLogicValues());
  // Propagate 0/1 constant functions.
  setConstFuncPins();
}
```

Especially `set_case_analysis` looks like something the design tool should *not* propagate into the logic and replace the logic with

maliberty:
set_case_analysis is more for modelling a particular operating mode and we wouldn't want to remove logic based on it (it could be used in another mode).  Netlist constants should be the basis for logic removal.

povik:
That's what I thought

src/rsz/src/Resizer.cc
maliberty: Are you wanting to defer this cleanup to another PR?

src/rsz/src/Resizer.cc
```diff
@@ -3648,4 +3648,106 @@ void Resizer::resetInputSlews()
...
+
+  int remove_inst_count = 0, remove_net_count = 0;
+  for (auto inst : network_->leafInstances()) {
+    if (!kept_instances.count(inst)) {
+      sta_->deleteInstance((Instance*) inst);
+      remove_inst_count++;
+    }
+  }
+
+  if (clean_nets) {
```
povik, line 3726: @maliberty I implemented the cleanup, see here

maliberty: This would be easier to implement directly on odb where you could just  check `net->getITerms().empty() && net->getBTerms().empty()` without having to build PinSeq.  Also dbNet::destory returns an updated iterator so you don't have to delay destruction.

povik: So far I got by without needing to learn much about odb but clearly some operations would be simpler if I access it directly. Should I revisit the code here or do you want to merge it as is so we can experiment with the new command?

maliberty: We can merge this but it is worth learning more of odb

maliberty:
It would be good to have at least one unit test.

src/rsz/test/eliminate_dead_logic1.tcl
```diff
@@ -0,0 +1,10 @@
+read_lef sky130hd/sky130hd.tlef
+read_lef sky130hd/sky130hd_std_cell.lef
+read_liberty sky130hd/sky130hd_tt.lib
+
+read_verilog eliminate_dead_logic1.v
+link_design top
+
+set_dont_touch \q2[0]
+eliminate_dead_logic
+write_verilog /dev/stdout
```
povik, line 10: We don't support non-Unix so this is hopefully fine

maliberty: Its non-idiomatic.  If you look at most other tests we would write to a .v file and have a .vok to diff against.

povik: Fixed

maliberty (approved):

Issue #6120, opened 2024-11-07
odb: write_odb fails with not so pretty iostream message


titan73:
### Describe the bug

See title.

### Expected Behavior

write_db should succeed.

### Environment

```shell
v2.0-16855-ga15114a95
```

### To Reproduce

openroad test.tcl

[write_db.zip](https://github.com/user-attachments/files/17669683/write_db.zip)

eder-matheus:
@titan73 If I manually create the dbs directory, the command don't fail. I'm not sure if any of our write commands should create directories. @maliberty any comments?

titan73:
@eder-matheus yeah indeed I've just figured out that dbs folder was removed but the error message was not very helpful and I couldn't pinpoint the problem. I don't expect write_db to create the folder that that would be good to have a error message complaining about the missing "dbs" folder.

maliberty:
What was the message?

maliberty:
I don't think there is any change here so "anymore" seems inaccurate.

eder-matheus:
> What was the message?

@maliberty This: `Error: test.tcl, 8 basic_ios::clear: iostream error (failed to open 'dbs/floorplan.odb.1'): iostream error`.

My guess is that we are not checking if the file was properly created before trying to write to it.

maliberty:
"failed to open" is pretty clear that it is a file system error but we could make it prettier.

titan73:
Updated title.

Indeed the message says what's wrong. I got confused with its complicated message (like an assert) and thought there was a bug.
So nothing wrong actually but yes a better message could be nice.
Btw is this message sent to the logger?
And I don't understand. The message refers to file 'dbs/floorplan.odb.1' but the file is 'dbs/floorplan.odb'. What is this '.1'?

titan73:
Fixed  by merged PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6133.

PR #6123, opened 2024-11-08
ord: Print a simpler message when write_db fails to open file.

titan73:


titan73:
For https://github.com/The-OpenROAD-Project/OpenROAD/issues/6120.

src/OpenRoad.cc
```diff
@@ -507,9 +507,12 @@ void OpenRoad::writeDb(std::ostream& stream)
...
 void OpenRoad::writeDb(const char* filename)
 {
-  utl::StreamHandler stream_handler(filename, true);
-
-  db_->write(stream_handler.getStream());
+  try {
+    utl::StreamHandler stream_handler(filename, true);
+    db_->write(stream_handler.getStream());
+  } catch (const std::ios_base::failure& f) {
+    logger_->error(ORD, 55, "Cannot write database to {}", filename);
```
maliberty, line 514: Please include the f.what string in the message as it may give more detail on the cause of the failure.

titan73: Yes for write_lef. write_def is different with the '.1'. Where does it come from?
```
>>> write_def foo/bar
[ERROR GUI-0070] No such file or directory: foo/bar.1
>>> write_lef foo/bar
[ERROR GUI-0070] basic_ios::clear: iostream error (failed to open 'foo/bar.1'): iostream error
```

maliberty: write_lef might need to write multiple files so they get .<num> suffixes.

write_lef/def are not part of GUI.

titan73: What is the case where write_lef write multiple files?
Regarding write_db & write_def, why using a suffix then? Should we use basic iostream instead?

oharboe (changes requested):
If we are going to do this, then the util::StreamHandler would be the place to do it.

As I recall util::StreamHandler went through a number of iterations and lots of review feedback, a lots of tripwires...

titan73:
> If we are going to do this, then the util::StreamHandler would be the place to do it.
> 

I don't understand your suggestion.

> As I recall util::StreamHandler went through a number of iterations and lots of review feedback, a lots of tripwires...

What is the purpose of util::StreamHandler? What problem is it supposed to solve? OR uses std::ifstream for reading.

oharboe:
@titan73 master for the code you propose to change is using utl::StreamHandler today. I think the  code reads pretty well and the purpose is pretty clear from the code.

If you ask ChatGPT/copilot to explain what it does and why it exists instead of your proposed change, I am pretty sure you will get a satisfactory answer.

In this day and age, I consider simple code that ChatGPT/copilot can adequately explain purpose and implementation of sufficiently documented.

titan73:
No need for ChapGPT. Seems the code prevents the existing file to be overwritten in case of failure.
I close this PR for now. I'll come up with something else later.

PR #6126, opened 2024-11-09
gui: dont check hidden children states when toggling parent checkbox

gadfort:
I noticed this when debugging https://github.com/The-OpenROAD-Project/OpenROAD/pull/6121.

Before the layers checkbox would be partially filled in (even though all visible boxes are checked):
Pasted image

After the layers checkbox is behaving correctly:
Pasted image (2)


maliberty (approved):
When do we have hidden rows?

gadfort:
> When do we have hidden rows?

When a technology doesn't have overlay and implant layers, those rows are hidden from view.

PR #6128, opened 2024-11-09
dbsta: add modinst optional argument to report_cell_usage

gadfort:
Changes:
- adds optional module inst argument to `report_cell_usage` to report the usage of cells in submodules. Default stays the top module.

### DIFF
--- a/src/dbSta/src/dbSta.tcl
+++ b/src/dbSta/src/dbSta.tcl
@@ -79,16 +79,28 @@ proc highlight_path { args } {
   }
 }
 
-define_cmd_args "report_cell_usage" {[-verbose]}
+define_cmd_args "report_cell_usage" {[-verbose] [module_inst]}
 
 proc report_cell_usage { args } {
   parse_key_args "highlight_path" args keys {} \
     flags {-verbose} 0
 
+  check_argc_eq0or1 "report_cell_usage" $args
+
   if { [ord::get_db_block] == "NULL" } {
-    sta_error "No design block found."
+    sta_error 1001 "No design block found."
   }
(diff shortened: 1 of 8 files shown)


src/dbSta/src/dbSta.tcl
```diff
@@ -79,16 +79,28 @@ proc highlight_path { args } {
   }
 }
 
-define_cmd_args "report_cell_usage" {[-verbose]}
+define_cmd_args "report_cell_usage" {[-verbose] [module_inst]}
```
maliberty, line 82: Doc update?

src/dbSta/test/report_cell_usage_modinsts.ok
```diff
@@ -0,0 +1,25 @@
+[INFO ODB-0227] LEF file: liberty1.lef, created 2 layers, 6 library cells
+Cell type report:                       Count       Area
+  Buffer                                    4    4000.00
+  Sequential cell                           2    2000.00
+  Total                                     6    6000.00
+
+Cell instance report:
+  snl_bufx1                                 4    4000.00
+  snl_ffqx1                                 2    2000.00
+Cell type report:                       Count       Area
```
maliberty, line 10: If the report is not for the top module I think it would be good to indicate that in the report.  Its confusing to see these three reports and only by looking at the tcl can you see what they mean.

maliberty (changes requested):

src/dbSta/src/dbSta.tcl
gadfort: it's a strange location for the docs, but updated.

maliberty (approved):

PR #6130, opened 2024-11-10
dbsta/utl: implement file and string redirection in logger

gadfort:
Changes:
- implements an interface to the OpenSTA file and string redirection to the logger. This change allows openroad tools to properly implement file redirection such as is needed for https://github.com/The-OpenROAD-Project/OpenROAD/pull/6046
- Add two units tests to check this behavior through both the logger and tcl commands.
- Change from using the default spglog formatter to a custom instance to allow for control of the newline that gets added automatically. This is now controlled in the calls to `log`.


gadfort:
@oharboe FYI, this should allow you to reopen the PR you had and get it working.

maliberty (approved):

maliberty:
I'm hoping you tested this with the GUI as well as we lack automated testing.

PR #6131, opened 2024-11-10
ord: Redirect writeDb file opening exception to ORD.

titan73:
Also add open method to strean handler so it can be instanciated outside the try block where file opening is done.

### DIFF
--- a/src/OpenRoad.cc
+++ b/src/OpenRoad.cc
@@ -507,8 +507,12 @@ void OpenRoad::writeDb(std::ostream& stream)
 
 void OpenRoad::writeDb(const char* filename)
 {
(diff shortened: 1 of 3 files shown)


oharboe (changes requested):
Keep the RAII interface unchanged and improve only error messages.

This PR requires all invocations to become substantially messier.

It is unclear to me why the API needs to change to fix the error messages and why we want to fix the error messages for only one case.

titan73:
Created https://github.com/The-OpenROAD-Project/OpenROAD/pull/6133 for error message in stream handler.

maliberty:
Is this still relevant after #6133 ?

titan73:
No. Closing.

Issue #5712, opened 2024-09-09
DRT: resolve conflicts between frInst::getTransform and frInst::getUpdatedXform


osamahammad21:
### Description

The code in DRT keeps fluctuating between the use of getTransform and getUpdatedXform. getTransform cannot be applied directly to the master shapes as its offset is not the correct transform offset. In my opinion, getTransform is always wrong and useless, however some parts of the code get away with using it.

titan73:
So the idea is to convert the code to always use getUpdatedXform and remove getTransform at the end?

maliberty:
What is the goal of getUpdatedXform?  I'm used to seeing people use getTransform across hierarchies.  Why do you think it is wrong?

osamahammad21:
getUpdatedXform returns the correct transform that should be applied to the master shapes. However, getTransform returns a wrong one, because its origin point is not the actual origin point but the lower left point of the instance. I am still managing to understand how it is used correctly.

maliberty:
Is the transform different from what you would get in DEF for the instance?

osamahammad21:
yes, thus the need for getUpdateXform.

maliberty:
getUpdatedXform is pretty ugly.  Can we just use the normal transform and avoid all this?  I'm concerned about how much of an impact it might have but haven't scoped it myself.

PR #6141, opened 2024-11-11
drt: getUpdatedXform(false) elimination

bnmfw:
This PR is ISPD and Secure CI Safe.

### The problem
This PR solves part of the conflicts presented in issue #5712. In `io::Parser::setInst()` the Transform object from `odb::dbInst` is copied and held as a private attribute of `drt::frInst`. This new Transform is also modified to have its origin at the lower left corner, which is not always accurate. Furthermore, an auxiliary function, `getUpdatedXform`, exists to undo this origin transformation, trying to get back to the original odb Transform. Unfortunately, even this strategy is not error proof, as this patch to get back to the odb inst in done on the private `drt::frInst` which might become obsolete if the instance is moved, as the update in the Transform to get the new cell location is not precise.

### Solutions
- New, temporary functions, called `getDBTransform` and `setDBInst` were created. Whenever `getUpdatedXform()` is called, `getDBTransform` is used as that is what the original function is trying to do. Once the private `drt::frInst` Transform is eliminted `getDBTransform` will be renamed to become simply `getTransform`. `getUpdatedXform(true)` use cases are not eliminated in this PR.
- `DesignCallBack::inDbPreMoveInst` was included to correctly "move" the instance in the R Tree strucute.

The elimination of `getUpdatedXform()` cases allowed for the elimination of `updateXform()` and simplifications of functions that used it to simply use `getDBTransform()`.

Furthermore, `getUpdatedXform()` was refactored to eliminate the now unused functionality. This reduced the function to one that simply gets the Transformation and sets the rotation as R0. `getUpdatedXform()` was renamed to `getNoRotationTransform()` accordingly (this is temporary, the objective is to eliminate the function altogether).

In a lot of cases I noticed that the code manually got the Transformation and set its rotation to R0. All this cases were substituted by `getNoRotationTransform()`. This might help eliminating the `drt::Inst` Transform, as it reduces many uses to a single function.

`frInstTerm::getShapes` and `frInstTerm::getBBox()` had dead code eliminated. Both functions had a mode in which they used the `drt::Inst` Transform, and a mode where they used the old `getUpdatedXform()` but only used the latter. They were refactored to only have the second case and no alternative mode.

Finally some redundant code was eliminated regarding Transform manipulation. When a transform is created it always defaults to having the origin at 0,0 and rotation as R0, specifying that upon creation of the Transform is redundant. Also, ideally there wont even be a way to set this parameters in the future, as odb will handle the Transform, so the less the drt sets this parameters the better. `Fixture::makeInst` was simplified to avoid these redundancies.

### gc_Test Patch
Now that the creation of an inst requires the association to a `odb::Inst` some of the tests on gc_test had to be patched to accommodate this initialization. `Fixture::createDummyInst` was created to accommodate for this requirement. The only required part of the instance is the transform which is initialized,


src/drt/src/db/obj/frInst.h
```diff
@@ -94,6 +95,15 @@ class frInst : public frRef
   void setOrigin(const Point& tmpPoint) override { xform_.setOffset(tmpPoint); }
   dbTransform getTransform() const override { return xform_; }
   void setTransform(const dbTransform& xformIn) override { xform_ = xformIn; }
+  void setDBInst(odb::dbInst* db_inst_in) { db_inst_ = db_inst_in; }
```
osamahammad21, line 98: make it a parameter in the constructor. There should never be an frInst that is not linked to dbInst

src/drt/src/db/obj/frInst.h
```diff
@@ -94,6 +95,15 @@ class frInst : public frRef
...
+  odb::dbInst* getDBInst() const { return db_inst_; }
+  dbTransform getDBTransform() const
+  {
+    if (!db_inst_) {
+      return dbTransform();
+    }
```
osamahammad21, line 104: unneeded if previous comment is addressed

src/drt/test/fixture.cpp
```diff
@@ -174,16 +174,11 @@ frTerm* Fixture::makeMacroPin(frMaster* master,
   return term;
 }
 
-frInst* Fixture::makeInst(const char* name,
-                          frMaster* master,
-                          frCoord x,
-                          frCoord y)
+frInst* Fixture::makeInst(const char* name, frMaster* master)
```
osamahammad21, line 177: This should create a dbInst and link the frInst to it.

src/drt/src/db/obj/frInst.h
```diff
@@ -94,6 +95,15 @@ class frInst : public frRef
   void setOrigin(const Point& tmpPoint) override { xform_.setOffset(tmpPoint); }
   dbTransform getTransform() const override { return xform_; }
```
osamahammad21, line 96: Is this used anywhere now?

osamahammad21 (changes requested):

src/drt/src/db/obj/frInst.h
bnmfw: Unfortunatly, yes. `getTransform()` itself is used in the now renamed `getNoRotationTransform()` and in some other cases. Both `setOrigin()` and `setOrient()` still se some uses. I intend to completely eliminate these two functions, but I'm not sure it is possible. (I'm not 100% sure about what I'm about to say) During Pin Access virtual vias have to be created to check for DRVs when considering if a Access Point is valid. These vias exists only on the drt, and don't have an odb counterpart during that checking so they currently need to use and configure the drt Transform.

src/drt/src/db/obj/frInst.cpp
```diff
@@ -35,77 +35,28 @@ namespace drt {
...
-  dbTransform xform = getTransform();
-  Point s(box.xMax(), box.yMax());
-  updateXform(xform, s);
+  dbTransform xform = getDBTransform();
   xform.apply(box);
   return box;
 }
 
-dbTransform frInst::getUpdatedXform(bool noOrient) const
+#include <iostream>
```
maliberty, line 51: Should be at the top of the file with the other includes

src/drt/src/db/obj/frInst.h
bnmfw: This exists only as a patch for the gc_test unit test. I was having a lot of trouble with it, as the odb Transform held trash when called. The old version worked because the drt Transform created a default Transform upon initialization 0,0,R0. This piece of code is to replicate the behavior as I thought I was loosing too much time trying to fix the unit test trying to make it work correctly. I will comply with your previous feedback, if it fixes the problem I will delete this piece of code.

src/drt/src/db/obj/frInst.h
maliberty: Does it need to be public?

src/drt/src/db/obj/frInst.cpp
bnmfw: Oh sorry, I used iostream for debugging, I missed it when preparing the PR

src/drt/test/fixture.cpp
```diff
@@ -83,6 +83,31 @@ void Fixture::addLayer(frTechObject* tech,
...
+  odb::dbMTerm::create(master, "a", dbIoType::INPUT, dbSigType::SIGNAL);
+  odb::dbMTerm::create(master, "b", dbIoType::INPUT, dbSigType::SIGNAL);
+  odb::dbMTerm::create(master, "c", dbIoType::OUTPUT, dbSigType::SIGNAL);
+  master->setFrozen();
+  auto ptr_db_inst = std::make_unique<odb::dbInst>();
+  odb::dbInst* db_inst = ptr_db_inst->create(chip->getBlock(), master, "dummy");
+  dbTransform trans = dbTransform();
+  db_inst->setTransform(trans);
+  return db_inst;
+}
```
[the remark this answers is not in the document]
bnmfw: ❓

bnmfw:
@osamahammad21 this is ISPD and secure CI Safe

src/drt/test/fixture.cpp
osamahammad21, line 109: This looks weird. I see that in this function you're creating a database, library, chip, block, master and inst. They are all separate from anything related to the actual design just to set a correct transform. The correct approach would be to modify the infrastructure for the test suite altogether. The setup for the test suite is done at Fixture::makeDesign. in that function you could initialize a database, library, chip and a block. in Fixture::makeMacro, you could create the corresponding dbMaster. Same would apply in makeMacroPin. finally, in makeInst create the corresponding dbInst as well. 
In the end, you should have only one db, tech, chip, block, and library for the design. and they all should be meaningful.

osamahammad21 (changes requested):

src/drt/test/fixture.cpp
bnmfw: I'll be sure to make fix this. Most of the code was taken from `dbDatabase* createSimpleDB`.

osamahammad21 (approved):

PR #6148, opened 2024-11-13
ci: drop cpp20 docker image

vvbandeira:


vvbandeira:
@maliberty @gadfort FYI

etc/DependencyInstaller.sh
```diff
@@ -696,6 +698,12 @@ _installCI() {
         docker-ce-cli \
         containerd.io \
         docker-buildx-plugin
+
+    # Install clang for C++20 support
+    wget https://apt.llvm.org/llvm.sh
+    chmod +x llvm.sh
+    ./llvm.sh 16 all
```
maliberty, line 705: Will this cause a change in compiler version for non-c++20 builds?

maliberty (approved):

etc/DependencyInstaller.sh
vvbandeira: No, by default we compile with gcc you need to ask cmake or the `Build.sh` script to use `clang-16` or other compilers.

maliberty: ok - ready to merge?

vvbandeira: On my side yes, not sure if Peter depended on this Dockerfile for anything or if it was just for your internal testing.

Issue #6149, opened 2024-11-13
gui: drc violation blinking disappears after selecting violation in drc viewer


titan73:
### Describe the bug

In drc viewer, mousing over a violation make it highlighted with blinking yellow in the renderer but when clicking on this very violation the highlighting disappears. That would be more natural to keep it highlighted.

In addition to this issue a possible enhancement could be to keep the selected violation highlighted when mouse cursor is not on any violations.

### Expected Behavior

Highlight is kept when clicking on a violation.

### Environment

```shell
v2.0-16941-g1452788e1
```

### To Reproduce

Whatever design with drc violations.

titan73:
You can assigned it to me. I have make it working and will send a PR.

gadfort:
@titan73 do you have `highlight selected` turned off?

titan73:
Yes it was was off. I turned it on. The selection becomes yellow but there is not blinking like we have with the mouse over.

titan73:
@gadfort Should we still keep this issue opened?
As you said in related PR:
`It seems like this exposes and masks some odd behavior of the GUI, but in the short run it seems fine. I think we should keep the issue open to track fixing the inspector animation behavior.`
Shouldn't we close this issue and open a new one for other issues?

maliberty:
What was the related PR?  I see nothing to fix here.  If there is another issue please open a new one.

titan73:
> What was the related PR? I see nothing to fix here. If there is another issue please open a new one.

The PR is https://github.com/The-OpenROAD-Project/OpenROAD/pull/6150 which has been merged.

PR #6150, opened 2024-11-13
gui: Make drc marker blink when selected as for mouse over.

titan73:
Mouse over makes the marker blink and then clicking stops it. Make it continue blinking after selection.

### DIFF
--- a/src/gui/src/drcWidget.cpp
+++ b/src/gui/src/drcWidget.cpp
@@ -225,6 +225,7 @@ void DRCWidget::clicked(const QModelIndex& index)
     } else {
       Selected t = Gui::get()->makeSelected(marker);
       emit selectDRC(t);
+      focusIndex(index);
     }
   } else {
     if (item->hasChildren()) {


titan73:
For https://github.com/The-OpenROAD-Project/OpenROAD/issues/6149.

gadfort:
@titan73 I'm unable to get a different behavior before or after this change. What exactly is the sequence of events that you are expecting this to cause? and how to test it?

titan73:
To reproduce
- extract test case below [drc_blink.zip](https://github.com/user-attachments/files/17738308/drc_blink.zip)
- cd to folder
- openroad -gui
- open test.odb from menu
- load violations.drc
- expand "short" tree
- click on first violation
- move mouse out of violation entry
- move mouse back to violation, the marker starts blinking
- click on violation entry and the marker stops blinking
With the change the last step does not make blinking stop

gadfort (approved):
@titan73 I'm able to replicate it. It doesn't actually "keep" the animation going, but instead its getting restarted.
I seems like this exposes and masks some odd behavior of the GUI, but in the short run it seems fine. I think we should keep the issue open to track fixing the inspector animation behavior.

titan73:
Yes indeed, it is restarted.
I'll keep the issue open. What behavior did you see?

PR #6156, opened 2024-11-13
dbsta: add cell area to metrics to mirror report information and totals

gadfort:
Adds area to metrics recorded with `report_cell_usage` and the totals for the module.


gadfort:
@maliberty could you take a look at this?

src/dbSta/src/dbSta.cc
```diff
@@ -536,17 +536,25 @@ void dbSta::report_cell_usage(odb::dbModule* module, const bool verbose)
...
+        = toLowerCase(regex_replace(type_name, regexp, "_"));
+
+    logger_->metric("design__instance__count__class:" + type_class,
                     stats.count);
+    logger_->metric("design__instance__area__class:" + type_class,
+                    stats.area / area_to_microns);
   }
+  logger_->metric("design__instance__count__class:total", total_usage);
+  logger_->metric("design__instance__area__class:total",
+                  total_area / area_to_microns);
```
maliberty, line 557: We already have design__instance__count, do we need design__instance__count__class:total?  Likewise design__instance__area and design__instance__area__class:total

gadfort: This provides the number for the selected modinst (design__instance__area is for the entire design)

maliberty: If you run report_cell_usage on two different modules then the metrics will collide.

gadfort: this is true to many commands, I change the metrics stage in my script before each call, but we could change the metrics prefix if you have a suggestion.

gadfort: for the top module would you drop the `__in_module`?

maliberty: The top module is the same as whole design, ie design__instance__count.

maliberty:
It would be nice to have a test for this.  Do you mind adding a small one?

gadfort:
@maliberty I figured it out, test added.

maliberty (approved):

PR #6158, opened 2024-11-13
drt: globals variables owned by TritonRoute instead of being extern

joaomai:
This PR encapsulates all global variables used in DRT into a struct that's owned by the TritonRoute class. Extern variables can cause problems with multi-threading, and all globals were extern before the changes. 

In general, most of them are written only when TritonRoute is instantiated and are set by TritonRoute::setParams. The exceptions being (from what I was able to gather):
```
- REPAIR_PDN_LAYER_NUM: TritonRoute::repairPDNVias and TritonRoute::checkDRC
- GC_IGNORE_PDN_LAYER_NUM: TritonRoute::repairPDNVias, TritonRoute::initDesign and TritonRoute::checkDRC
- ENABLE_VIA_GEN: TritonRoute::main and TritonRoute::pinAccess
- VIAINPIN_BOTTOMLAYERNUM and VIAINPIN_TOPLAYERNUM: TritonRoute::initGuide
- USENONPREFTRACKS: TritonRoute::main (but its value is based on another global, DBPROCESSNODE)
- MAX_THREADS: TritonRoute::main, TritonRoute::pinAccess, TritonRoute::getDRCMarkers, FlexDRGraphics::startIter and FlexPAGraphics::FlexPAGraphics
- MTSAFEDIST: io::Parser::setNDRs
- BOTTOM_ROUTING_LAYER and TOP_ROUTING_LAYER: io::Parser::readTechAndLibs
```

Further analysis on what really needs or not to be global might be interesting.


QuantamHD:
Just a general comment I think the struct name should be some something like RouterConfiguration rather than Global.

I think it will generally help with readability

maliberty:
In a future PR it would also be nice to make the member variables conform to standard naming.

maliberty (approved):

maliberty:
@osamahammad21 do you have any concerns before merging?  I'd like to get this finished.

PR #6160, opened 2024-11-14
utl: add tee TCL command

gadfort:
Adds:
- `tee` as a tcl command modeled after: https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/tee.html


maliberty:
This seems to make the previous PR for redirection redundant as `tee -quiet` does the same.

src/utl/README.md
```diff
@@ -51,6 +51,33 @@ man clock_tree_synthesis
...
+    [-append]
+    [-quiet]
+    command
+```
+
+#### Options
+
+| Switch Name | Description | 
+| ----- | ----- |
+| `filename` | File to redirect output into. |
```
maliberty, line 69: -file filename

src/utl/src/Logger.cpp
```diff
@@ -239,6 +239,24 @@ void Logger::redirectFileEnd()
...
+void Logger::teeFileBegin(const std::string& filename)
+{
+  file_redirect_ = std::make_unique<std::ofstream>(filename.c_str());
+  setRedirectSink(*file_redirect_, true);
+}
+
+void Logger::teeFileAppendBegin(const std::string& filename)
+{
+  file_redirect_
+      = std::make_unique<std::ofstream>(filename.c_str(), std::ofstream::app);
```
maliberty, line 251: `.c_str()` is unnecessary

maliberty:
I know its a bit odd but what happens with `tee -file foo { report_checks > bar }`?

maliberty (changes requested):

gadfort:
> This seems to make the previous PR for redirection redundant as `tee -quiet` does the same.

@maliberty yes, this provides a consistent interface to use.


> I know its a bit odd but what happens with `tee -file foo { report_checks > bar }`?

I don't think we want to support that, so it should throw an error.

maliberty:
> should throw an error.

Does it?

maliberty (approved):

PR #6166, opened 2024-11-14
Format max message reached log messages.

mikesinouye:
Formatting of the max message reached log may have been accidentally omitted.


stefanottili:
it would be very helpful when this message would tell the novice user
1) what the limit is and
2) which variable to set to change it

src/utl/include/utl/Logger.h
```diff
@@ -289,9 +289,9 @@ class Logger
 
     if (count == max_message_print) {
       logger_->log(level,
-                   "[{} {}-{:04d}] message limit reached, "
-                   "this message will no longer print"
-                       + std::string(spdlog::details::os::default_eol),
+                   FMT_RUNTIME("[{} {}-{:04d}] message limit reached, "
+                               "this message will no longer print"
+                               + std::string(spdlog::details::os::default_eol)),
```
maliberty, line 294: @gadfort could we just use \n here?  Are you changing the EOL?

maliberty:
@stefanottili I agree though not strictly related to the problem at hand.  @mikesinouye do you want to tackle that request?

mikesinouye:
> @stefanottili I agree though not strictly related to the problem at hand. @mikesinouye do you want to tackle that request?

I added printing of the limit value but only the file the limit is in; the limit is hardcoded and piping it in as a flag or similar will be slightly more involved so I'll leave that as a future improvement.

src/utl/include/utl/Logger.h
gadfort: This is consistent with all the logging commands have that added except the one that doesn't need it. I think it evaluates to \n, but I don't see any reason to change it since it should then be changed everywhere.

src/utl/include/utl/Logger.h
```diff
@@ -289,9 +289,11 @@ class Logger
 
     if (count == max_message_print) {
       logger_->log(level,
-                   "[{} {}-{:04d}] message limit reached, "
-                   "this message will no longer print"
-                       + std::string(spdlog::details::os::default_eol),
+                   FMT_RUNTIME("[{} {}-{:04d}] message limit ("
+                               + std::to_string(max_message_print)
```
maliberty, line 293: Just put {} in the message, pass default_eol as an arg, and skip FMT_RUNTIME.  The is no user provided string here that requires this approach.

maliberty (approved):

mikesinouye:
@maliberty I thought `FMT_RUNTIME ` was required to properly format the error category and number (`[{} {}-{:04d}]`) and not the newline?

I see this error in the build : 

```
Error: UTL 0001 used 2 times, next free message id is 10
[2024-11-19T04:22:28.415Z]   Appears in CFileUtils.cpp on line 8 
[2024-11-19T04:22:28.415Z]   Appears in logger_max_messages.tcl on line 3 
[2024-11-19T04:22:28.415Z] gmake[2]: *** [src/utl/CMakeFiles/utl.dir/build.make:157: src/utl/utl.a] Error 1
[2024-11-19T04:22:28.415Z] gmake[2]: *** Deleting file 'src/utl/utl.a'
[2024-11-19T04:22:28.415Z] gmake[1]: *** [CMakeFiles/Makefile2:6253: src/utl/CMakeFiles/utl.dir/all] Error 2
[2024-11-19T04:22:28.415Z] gmake[1]: *** Waiting for unfinished jobs....
```

maliberty:
FMT_RUNTIME is only required if the message string is dynamic.  Having placeholders does not require FMT_RUNTIME .

PR #6188, opened 2024-11-18
utl/man: ensure man pages get installed in openroad share directory to avoid clobbering other installs

gadfort:
Changes:
- installs the pages for `man` in the `OPENROAD_SHARE` share directory to avoid interfering with other installs


luarss (approved):
LGTM

PR #6196, opened 2024-11-18
Rename deleted_unique_ptr to UniquePtrWithDeleter

allight:
`deleted_unique_ptr` implies that the contents are already deleted or something which is both confusing and incorrect. `UniquePtrWithDeleter` more accurately captures the intent of a unique_ptr with a custom Deleter.

### DIFF
--- a/src/utl/include/utl/deleter.h
+++ b/src/utl/include/utl/deleter.h
@@ -10,6 +10,6 @@
 namespace utl {
 
 template <typename T>
-using deleter_unique_ptr = std::unique_ptr<T, std::function<void(T*)>>;
+using UniquePtrWithDeleter = std::unique_ptr<T, std::function<void(T*)>>;
 
 }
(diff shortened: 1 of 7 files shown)


allight:
@QuantamHD

QuantamHD:
LGTM to me.

src/utl/include/utl/deleter.h
```diff
@@ -10,6 +10,6 @@
 namespace utl {
 
 template <typename T>
-using deleted_unique_ptr = std::unique_ptr<T, std::function<void(T*)>>;
+using deleter_unique_ptr = std::unique_ptr<T, std::function<void(T*)>>;
```
maliberty, line 13: If we are fixing the name it would also be good to make it adhere to Google style where type names are PascalCase: DeleterUniquePtr.  Personally I would choose UniquePtrWithDeleter even though its a bit more verbose.

allight: Done.

PR #6199, opened 2024-11-19
grt: check if edges are initialized before get congestion data

eder-matheus:
Fix Coverage pipeline issue.


eder-matheus:
> When does it happen that these are not initialized when calling saveCongestion?

It happens when we initialize the nets. This process happens before initializing the edges of FastRouteCore.

eder-matheus:
> Why do we call saveCongestion that early?

If I understand correctly, the code is catching the error and try to save the congestion here:
image

PR #6215, opened 2024-11-21
stt: converting flute to class and removing global variables

LucasYuki:
Converted flute heuristic algorithm into the class stt::flt::Flute.

Global variables removed  (issue #5981)
Variables changed:
- Moved into Flute class:
  - stt::flt::LUT
  - stt::flt::lut_valid_d
  - stt::flt::numsoln 
- Changed to static const:
  - stt::flt::post9[abi:cxx11]
  - stt::flt::powv9[abi:cxx11]
- Was already static const
  - stt::flt::base64_chars


src/stt/include/stt/flute.h
```diff
@@ -56,116 +55,166 @@ using stt::Tree;
...
+ public:
+  struct csoln
+  {
+    unsigned char parent;
+    unsigned char seg[11];  // Add: 0..i, Sub: j..10; seg[i+1]=seg[j-1]=0
+    unsigned char rowcol[FLUTE_D - 2];  // row = rowcol[]/16, col = rowcol[]%16,
+    unsigned char neighbor[2 * FLUTE_D - 2];
+  };
+  using LUT_TYPE = struct csoln***;
+  using NUMSOLN_TYPE = int**;
```
maliberty, line 80: do these need to be public?

src/stt/include/stt/flute.h
```diff
@@ -56,116 +55,166 @@ using stt::Tree;
...
+  ~Flute() { deleteLUT(); }
+
+  // User-Callable Functions
+  // Delete LUT tables for exit so they are not leaked.
+  void readLUT();
+  void makeLUT(LUT_TYPE& LUT, NUMSOLN_TYPE& numsoln);
+  void initLUT(int to_d, LUT_TYPE LUT, NUMSOLN_TYPE numsoln);
+  void ensureLUT(int d);
+  void deleteLUT();
+  void deleteLUT(LUT_TYPE& LUT, NUMSOLN_TYPE& numsoln);
```
maliberty, line 93: It looks like only ensureLUT needs to be public.

src/mpl/src/ParquetFP/src/DB.cxx
```diff
@@ -1800,6 +1800,7 @@ float DB::evalSteiner(bool useWts, bool scaleTerms)
 
   float total = 0.f;
   vector<Point> pointsOnNet;
+  Flute f = Flute();
```
eder-matheus, line 1803: I'm not a big fan of single char names for variables. I think you can just use `Flute flute;` for clarity.

src/stt/include/stt/flute.h
```diff
@@ -56,116 +55,166 @@ using stt::Tree;
...
+#if FLUTE_D <= 7
+#define MGROUP 5040 / 4  // Max. # of groups, 7! = 5040
+#define MPOWV 15         // Max. # of POWVs per group
+#elif FLUTE_D == 8
+#define MGROUP 40320 / 4  // Max. # of groups, 8! = 40320
+#define MPOWV 33          // Max. # of POWVs per group
+#elif FLUTE_D == 9
+#define MGROUP 362880 / 4  // Max. # of groups, 9! = 362880
+#define MPOWV 79           // Max. # of POWVs per group
+#endif
```
eder-matheus, line 67: I would like to see these defines moved to const members of the class. Also, since FLUTE_D is always 9, I don't see the point on keeping the checks for other values. This can be made in a future PR tho.

eder-matheus (changes requested):

src/stt/include/stt/flute.h
LucasYuki: Only the methods flute, plottree, write_svg, and flutes need to be public

src/stt/include/stt/flute.h
LucasYuki: no

src/mpl/src/ParquetFP/src/DB.cxx
```diff
@@ -1895,7 +1896,8 @@ float DB::evalSteiner(bool useWts, bool scaleTerms)
         flute_x[i] = static_cast<double>(pointsOnNet[i].x);
         flute_y[i] = static_cast<double>(pointsOnNet[i].y);
       }
-      Tree flutetree = flute(pointsOnNet.size(), flute_x, flute_y, ACCURACY);
+      Tree flutetree
+          = flute.flute(pointsOnNet.size(), flute_x, flute_y, ACCURACY);
```
eder-matheus, line 1900: Are you initializing this flute object? I believe you need to call the flute() function from an SteinerTreeBuilder object.

eder-matheus (changes requested):

src/mpl/src/ParquetFP/src/DB.cxx
LucasYuki: As we discussed, yes, the flute object is not initialized this way, but this code was inside a conditional inclusion that's not used anymore (if USEFLUTE is defined).

So I removed all conditional includes that have this condition

eder-matheus (approved):
@LucasYuki The last changes looks good to me. But please, fix the clang-tidy warnings on the mpl code.

maliberty (approved):

PR #6226, opened 2024-11-24
Implement a tool to query ORAssistant

palaniappan-r:
Implemented the `askbot` and `set_bothost` commands under the `ora` tool. They can be used to make queries to [ORAssistant](https://github.com/The-OpenROAD-Project/ORAssistant), a LLM based conversational assistant for OpenROAD

### `askbot`

This command takes an argument and passes it as a query to the hosted ORAssistant backend, by making a POST request using the `libcurl` library.

### `set_bothost`

This command modifies ORAssistant backend's URL that's being queried by `askbot`. The modified URL is then saved to a file locally, for subsequent reference.


rovinski:
Why can't this run locally as an add-on rather than sending requests over the internet?

palaniappan-r:
> Why can't this run locally as an add-on rather than sending requests over the internet?

If the user chooses to the application locally, they'll have to follow instructions on [ORAssistant's README](https://github.com/The-OpenROAD-Project/ORAssistant?tab=readme-ov-file#backend-setup), spin the backend server up, and use the `set_bothost` command to have the `askbot` command query the local server instead.

Currently, `askbot` queries a hosted version of ORAssistant by default, but I can change this if necessary.

palaniappan-r:
> Is there a data control plan or disclaimer if users are sending data over the internet? [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) may even come into play here.
The only data being transmitted over the internet is the query entered by the user. 

Currently, there is no formal disclaimer in place. I could draft a suitable disclaimer to address this and ensure compliance with relevant regulations such as GDPR.

docker/Dockerfile.builder
```diff
@@ -18,4 +18,4 @@ WORKDIR /OpenROAD
 
 ENV PATH=${LOCAL_PATH}:${PATH}
 
-RUN ./etc/Build.sh -compiler=${compiler} -threads=${numThreads} -deps-prefixes-file=${depsPrefixFile}
+RUN ./etc/Build.sh -compiler=${compiler} -threads=${numThreads} -deps-prefixes-file=${depsPrefixFile}
```
luarss, line 21: why is there a change here?

etc/DependencyInstaller.sh
```diff
@@ -579,7 +583,7 @@ _installDebianPackages() {
         unzip \
         wget \
         zlib1g-dev
-
+
```
luarss, line 586: stray change?

src/ora/src/Ora.tcl
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2024, The Regents of the University of California
```
luarss, line 1: Maybe Copyright 2025?

luarss (commented):
Just some nits, no comments on code end

src/ora/src/Ora.tcl
palaniappan-r: I've updated it now.

docker/Dockerfile.builder
palaniappan-r: Reverted.

luarss:
@palaniappan-r  no further comments from me, pending @maliberty and @vvbandeira review.

luarss:
On second thought, I think we need to mark this as draft because the permalink is not available. I will discuss more with @dralabeing and @vvbandeira.

src/ora/src/Ora.cc
```diff
@@ -0,0 +1,276 @@
...
+  db_ = db;
+  logger_ = logger;
+
+  Ora_Init(tcl_interp);
+  sta::evalTclInit(tcl_interp, sta::ora_tcl_inits);
+
+  hostUrl
+      = "https://bursting-stallion-friendly.ngrok-free.app/graphs/"
+        "agent-retriever";
+  localDirPath = std::string(getenv("HOME")) + "/.local/share/openroad";
```
gadfort, line 128: why not make a helper function to generate this path since it's used in multiple locations

src/ora/src/Ora.cc
```diff
@@ -0,0 +1,276 @@
...
+  sourceFlag_ = true;
+}
+
+void Ora::checkLocalDir()
+{
+  std::ifstream localDir(localDirPath);
+  if (!localDir) {
+    logger_->info(utl::ORA, 112, "Creating ~/.local/share/openroad directory.");
+    std::string mkdirCmd = "mkdir -p " + localDirPath;
+    int ret = system(mkdirCmd.c_str());
```
gadfort, line 149: wouldn't it be better to use: https://en.cppreference.com/w/cpp/filesystem/create_directory instead of generating a system call (something I think we could avoid)

src/ora/src/Ora.cc
```diff
@@ -0,0 +1,276 @@
...
+    logger_->info(utl::ORA,
+                  109,
+                  "ORAssistant host saved to "
+                  "~/.local/share/openroad/orassistant_host.txt");
+  } else {
+    logger_->warn(utl::ORA, 108, "Failed to write ORAssistant host to file.");
+  }
+}
+
+void Ora::setConsent(const char* consent)
```
gadfort, line 258: since you are checking for just a "y" for yes, I think you need to verify that this consent string matches your expectations.

src/ora/src/Ora.tcl
```diff
@@ -0,0 +1,43 @@
...
+proc set_bothost { hostUrl } {
+  puts "Setting ORAssistant host to $hostUrl"
+
+  sta::check_argc_eq1 "set_bothost" $hostUrl
+
+  ora::set_bothost $hostUrl
+}
+
+proc set_consent { consent } {
+  puts "Setting ORAssistant consent to yes."
```
gadfort, line 37: what is I said "no"?

src/ora/test/regression_tests.tcl
```diff
@@ -0,0 +1,5 @@
+record_tests {
+  set_bothost1
+  set_consent1
+  askbot1
+}
```
gadfort, line 5: all the testing needs to be moved to the CTest as this was removed.

src/ora/src/Ora.tcl
```diff
@@ -0,0 +1,43 @@
...
+
+proc set_bothost { hostUrl } {
+  puts "Setting ORAssistant host to $hostUrl"
+
+  sta::check_argc_eq1 "set_bothost" $hostUrl
+
+  ora::set_bothost $hostUrl
+}
+
+proc set_consent { consent } {
```
gadfort, line 36: `set_consent` is an awfully generic name, it might be better as `set_ora_consent`

gadfort (changes requested):
There is no README for this

rovinski:
The request for consent is a bit better than the state before, but there are still a few things I would want to see change:

1. Default to a locally hosted copy first, or at least explicitly ask the user if it's okay to use the web-hosted version
2. Make it explicit that only the data the user types will be sent outside the application, no other user or design data.
3. Point to directions on how to set up a locally hosted copy on first use
4. Don't require consent for anything other than the web-hosted copy (not only because it's not needed, but because safety guarantees can't be provided)
5. There needs to be some kind of versioning for the consent. If there were to be a new version in the future which has a different kind of privacy policy, then you can't rely on an old consent file with a new version of OpenROAD.

palaniappan-r:
> The request for consent is a bit better than the state before, but there are still a few things I would want to see change:
> 
>     1. Default to a locally hosted copy first, or at least explicitly ask the user if it's okay to use the web-hosted version
> 
>     2. Make it explicit that only the data the user types will be sent outside the application, no other user or design data.
> 
>     3. Point to directions on how to set up a locally hosted copy on first use
> 
>     4. Don't require consent for anything other than the web-hosted copy (not only because it's not needed, but because safety guarantees can't be provided)
> 
>     5. There needs to be some kind of versioning for the consent. If there were to be a new version in the future which has a different kind of privacy policy, then you can't rely on an old consent file with a new version of OpenROAD.

The `askbot` command now requires a (y/n) consent, where `y` is for the web-hosted version and `n` is for the locally hosted version. Both the consent and the local host URL can be configured using the `ora_init` command.

src/ora/src/Ora.tcl
palaniappan-r: I've moved this to `ora_init`.

src/ora/test/regression_tests.tcl
palaniappan-r: I've moved the tests, but still have to figure out a way to start the python server (to mock API responses for tests).  Previously, I started the Python server within the regression bash script itself. Could you suggest ways to handle this directly in CTest?

rovinski:
That doesn't really make sense - it shouldn't need to ask for consent if not connecting to the OpenROAD web-hosted version. It could be connected to another web-hosted version, in which case, it shouldn't display the message. Not only because the user wouldn't need to consent, but also the other web-hosted version may not have a privacy policy and so the consent message shouldn't claim so.

palaniappan-r:
> That doesn't really make sense - it shouldn't need to ask for consent if not connecting to the OpenROAD web-hosted version. It could be connected to another web-hosted version, in which case, it shouldn't display the message. Not only because the user wouldn't need to consent, but also the other web-hosted version may not have a privacy policy and so the consent message shouldn't claim so.
> 
> What about the other concerns?

When the `askbot`  command is run for the first time, it displays instructions pointing to a setting up a local version. It then checks if a local url has been provided, subsequent pointing to the `ora_init` command to configure this. If my earlier comment wasn't clear, only this message is displayed for `askbot`'s first use, not the consent message.

As for the consent management, it is being handed with the `ora_init` command, with `y` required for using the official web hosted version (the consent message is only displayed for this). The same command is used to connect the app to a local version. This currently requires the consent to be explicitly set to `n`, but I can make this automatic if need be.

I've explicitly mentioned that only the data typed-in will be sent outside the application in the consent message as well. I hope this clears your concerns, please let know if there's anything else to be addressed.

rovinski:
Ok thanks, that's much clearer. I don't think using a local copy should require consent to be set to "n". If a user wants to switch between web and local they should be able to without having to constantly grant and revoke consent.

My only other concern would be a version number for the consent to make sure that in the future if the requirements ever change, the user can be reprompted. Something like
```
version=1
consent=y
```
then if version 2 comes along which requires different permissions, the user will need to re-consent.

src/ora/test/regression_tests.tcl
palaniappan-r: I've rewritten the tests to use CTest and have implemented the requested changes. Please let me know if any further modifications are needed.

palaniappan-r:
> Ok thanks, that's much clearer. I don't think using a local copy should require consent to be set to "n". If a user wants to switch between web and local they should be able to without having to constantly grant and revoke consent.
> 
> My only other concern would be a version number for the consent to make sure that in the future if the requirements ever change, the user can be reprompted. Something like
> 
> ```
> version=1
> consent=y
> ```
> 
> then if version 2 comes along which requires different permissions, the user will need to re-consent.

I've updated the consent management system to track the version for which consent was provided. If the version changes, the user will need to provide consent again.  

Users can now switch between the local and cloud versions without needing to provide or revoke consent. To use the cloud version, running `ora_init cloud y` grants consent and tracks the version. For the local version, running `ora_init local hostUrl` will do, with no need to revoke consent, if it was previously provided.

Please let me know if any further modifications are needed.

src/ora/src/Ora.cc
palaniappan-r: I'm checking this (along with the mode string) and printing appropriate error msgs in the `Ora.i` file.

vvbandeira:
@luarss 
is this still relevant?

luarss:
We can close it for the time being. We will probably create a new PR for this

PR #6232, opened 2024-11-25
grt: MakeWireParasitics consider segments out of range when the design is congested

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/6228

### DIFF
--- a/src/grt/src/MakeWireParasitics.cpp
+++ b/src/grt/src/MakeWireParasitics.cpp
@@ -184,13 +184,18 @@ void MakeWireParasitics::makeRouteParasitics(
     NodeRoutePtMap& node_map)
 {
   const int min_routing_layer = grouter_->getMinRoutingLayer();
+  const bool is_congested = grouter_->isCongested();
 
   size_t resistor_id_ = 1;
   for (GSegment& segment : route) {
     const int wire_length_dbu = segment.length();
 
     const int init_layer = segment.init_layer;
-    sta::ParasiticNode* n1 = (init_layer >= min_routing_layer)
+    // allow segments out of the layer range only when the design have
+    // congestion, becaus GRT may produce segments out of the layer range in
+    // this scenario.
+    bool is_valid_layer = init_layer >= min_routing_layer || is_congested;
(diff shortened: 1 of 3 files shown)


src/grt/src/MakeWireParasitics.cpp
```diff
@@ -184,13 +184,18 @@ void MakeWireParasitics::makeRouteParasitics(
...
 
   size_t resistor_id_ = 1;
   for (GSegment& segment : route) {
     const int wire_length_dbu = segment.length();
 
     const int init_layer = segment.init_layer;
-    sta::ParasiticNode* n1 = (init_layer >= min_routing_layer)
+    // allow segments out of the layer range only when the design have
+    // congestion, becaus GRT may produce segments out of the layer range in
+    // this scenario.
```
maliberty, line 196: I'm surprised by this - I would have though the overflow would still be confined to the valid layer range.  Please say more about this.

eder-matheus: A layer out of the layer range just have all of its edges with zero capacity. During layer assignment, it will try to assign each segment to an edge with non-zero capacity, but when everything is congested, it will assign to any layer that has the best cost. This usually happens in segments that are close to the pins, so it is unlikely to have a long segment assigned to a layer out of range.

I guess we could update layer assignment to respect the layer range, but it would require a larger PR.

maliberty (approved):

maliberty:
Needs a format

src/grt/src/MakeWireParasitics.cpp
maliberty: We could just not build gcells for out of range layers.  In any case it can wait.

eder-matheus: We need them for pin access. We don't create segments in layers out of range in a regular case (i.e., no congestion), but we do create vias to these layers, specially for the first layer.

In any case, I think there is a way of doing that. I will create an issue for this to keep track of this enhancement.

PR #6236, opened 2024-11-25
mpl: removing parquetfp::BTreeAreaWireAnnealer::makeMoveSlacks()

LucasYuki:
Removing unused makeMoveSlacks function from parquetfp::BTreeAreaWireAnnealer.

This removes 2 static variables from mpl (issue #5981):
- parquetfp::BTreeAreaWireAnnealer::makeMoveSlacks()::total
- parquetfp::BTreeAreaWireAnnealer::makeMoveSlacks()::numHoriz


maliberty:
@LucasYuki you don't need to invest time into mpl as it is going to be replaced by mpl2.  It should be considered in sunset.

PR #6237, opened 2024-11-26
grt: update grt::have_routes to check if guides are congested

eder-matheus:


src/grt/src/GlobalRouter.cpp
```diff
@@ -229,7 +229,8 @@ bool GlobalRouter::haveRoutes()
     logger_->warn(GRT, 97, "No global routing found for nets.");
   }
 
-  return !routes_.empty();
+  bool congested_routes = is_congested_ && !allow_congestion_;
```
maliberty, line 232: Is allow_congestion_ going to be set on reload?

PR #6239, opened 2024-11-26
gui: help browser

gadfort:
Adds:
- `HelpBrowser` widget to allow for interactive / offline viewing of help if it's available.

Changes:
- Adds two functions to OpenROAD to allow for a centralized location to determine the location of the docs
- `man` switched to use new path access.

Notes:
- ~Due to the exclusion of `std::filesystem` I had to do path and directory handling manually, this can be simplified if `std::filesystem` is allowed.~
- `man` could be configured to open the gui help if we want.
- The HTML help is not particularly pretty and I don't find it helpful to know the compile date or header, but these issues can be handled separately.
- The script widget could be modified to allow for hyperlinks to this widget got commands and error messages (that would be best handled in a separate PR).
- there are lots of missing functions from this (outside the scope of this PR), but anything in ODB, ORD, and STA are completely missing.
- the GUI commands probably just need to be promoted out of the namespace from `gui::show` -> `gui_show` or `show_gui` since the man page generation clobbers its name (should be done in a separate PR).

Preview:
image

### DIFF
--- a/src/OpenRoad.cc
+++ b/src/OpenRoad.cc
@@ -35,6 +35,7 @@
 
 #include "ord/OpenRoad.hh"
 
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <thread>
@@ -660,22 +661,18 @@ std::string OpenRoad::getDocsPath() const
     return "";
   }
 
+  std::filesystem::path path(exe);
+
   // remove binary name
-  std::string path = exe.substr(0, exe.find_last_of('/'));
+  path = path.parent_path();
 
-  const std::size_t build_idx = exe.length() - 13;
-  if (exe.find("/src/openroad") != build_idx - 1) {
-    path = path.substr(0, build_idx);
-
+  if (path.stem() == "src") {
     // remove build
-    path = path.substr(0, path.find_last_of('/'));
-    return path + "/docs";
+    return path.parent_path().parent_path() / "docs";
   }
 
   // remove bin
-  path = path.substr(0, path.find_last_of('/'));
-
-  return path + "/share/openroad/man";
+  return path.parent_path() / "share" / "openroad" / "man";
 }
 
 const char* OpenRoad::getVersion()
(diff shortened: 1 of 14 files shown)


maliberty:
Could you include a screen shot or two?  

Now that we are off centos7 I think we can try to use std filesystem and hopefully not having linking headaches.

src/OpenRoad.cc
```diff
@@ -640,6 +640,44 @@ int OpenRoad::getThreadCount()
...
+  const std::string exe = getExePath();
+
+  if (exe.empty()) {
+    return "";
+  }
+
+  // remove binary name
+  std::string path = exe.substr(0, exe.find_last_of('/'));
+
+  const std::size_t build_idx = exe.length() - 13;
```
maliberty, line 666: why 13?

src/gui/src/gui.i
```diff
@@ -741,4 +741,13 @@ void unminimize()
   gui->unminimize();
 }
 
+void show_help(const std::string& item)
```
maliberty, line 744: Add to README?

maliberty:
Based on #6241 you can use std filesystem

gadfort:
@maliberty I'll rework the file handling then to use std::filesystem

src/OpenRoad.cc
gadfort: It's len("/src/OpenROAD"), but should be moot with std::filesystem

src/OpenRoad.cc
```diff
@@ -640,6 +641,40 @@ int OpenRoad::getThreadCount()
...
+  if (exe.empty()) {
+    return "";
+  }
+
+  std::filesystem::path path(exe);
+
+  // remove binary name
+  path = path.parent_path();
+
+  if (path.stem() == "src") {
```
maliberty, line 669: When does this occur?

gadfort: this happens when executing from the build directory instead of an installed version of openroad

maliberty (approved):

PR #6245, opened 2024-11-26
odb: protecting dbModule::makeUniqueDbModule()::name_id_map with a lock

LucasYuki:
Protects the static variable name_id_map from the dbModule::makeUniqueDbModule method with a lock (issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/5981)

### DIFF
--- a/src/odb/src/db/dbModule.cpp
+++ b/src/odb/src/db/dbModule.cpp
@@ -561,13 +561,13 @@ dbModule* dbModule::makeUniqueDbModule(const char* cell_name,
   do {
     std::string full_name = module_name;
 
-    name_id_map_mutex.lock();
+    std::unique_lock<std::mutex> lock(name_id_map_mutex);
     int& id = name_id_map[module_name];
     if (id > 0) {
       full_name += "_" + std::to_string(id);
     }
     ++id;
-    name_id_map_mutex.unlock();
+    lock.unlock();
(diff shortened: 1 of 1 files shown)


src/odb/src/db/dbModule.cpp
```diff
@@ -559,11 +560,15 @@ dbModule* dbModule::makeUniqueDbModule(const char* cell_name,
   std::string module_name = orig_cell_name + '_' + std::string(inst_name);
   do {
     std::string full_name = module_name;
+
+    name_id_map_mutex.lock();
```
maliberty, line 564: Its better to use std::unique_lock than manual lock and unlock calls (eg exception safety).

QuantamHD:
This does not seem like an acceptable solution. If you have multiple odb's won't calls to makeUniqueDbModule interfere with each other?

maliberty:
Interfer in the sense that the numbering will depend on the order of the calls.  It won't be a thread safety issue but it could cause some non-determinism in the names.

maliberty:
I think the easiest thing to do is move name_id_map to be a non-persistent member of dbBlock.  No lock is then needed and the extra memory will disappear on close.

LucasYuki:
Implemented as @maliberty suggested in the PR #6300.

Issue #6234, opened 2024-11-25
GRT: should not create segments out of the layer range when design has congestion


eder-matheus:
### Describe the bug

During layer assignment in a congested design, GRT assigns edges to layers out of the defined layer range.

### Expected Behavior

Never create segments out of the layer range.

### Environment

```shell
See CI.
```

### To Reproduce

Any design with high congestion

PR #6248, opened 2024-11-26
grt: avoid segments out of layer range in congested routing

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/6234


maliberty (approved):

maliberty:
needs a format

PR #6269, opened 2024-11-29
gpl routability log change

gudeh:
Minor modification on routability log message from "numCall" variable name to "iteration", which is exactly what the internal variable represents.

Noticed that the inflation and bloat counters are unused in practice. The counters exceed their maximum values without triggering any functionality.

Bloat is expected to occur when inflation reaches its maximum iterations, but no action is associated with this condition. This appears to be a remnant of an earlier idea that was not fully implemented or was removed, leaving behind unused code.


gudeh:
This PR should cover the request in issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/5201

eder-matheus (approved):

PR #6272, opened 2024-11-29
pad: add -mirror to place_pad to ensure proper functionality

gadfort:


gadfort:
@maliberty this is correct mirror behavior in this test. Not #6270

maliberty (approved):

PR #6284, opened 2024-12-03
DRT: fix distributed routing and cleanup code

osamahammad21:


src/drt/doc/Distributed.md
```diff
@@ -56,6 +56,15 @@ A shared folder on a NFS is used to share routing updates between the leader and
 
 [https://www.digitalocean.com/community/tutorials/how-to-set-up-an-nfs-mount-on-ubuntu-20-04](https://www.digitalocean.com/community/tutorials/how-to-set-up-an-nfs-mount-on-ubuntu-20-04)
 
+The main steps on the NFS server machine are:
+
+1. run ` sudo apt install nfs-kernel-server ` and ` sudo apt install nfs-common `
+1. In the etc/exports file, add the following line replacing \${PATH} with the actual shared directory path you choose: ` ${PATH} 10.128.0.0/255.255.0.0(rw,no_subtree_check,no_root_squash) `
```
maliberty, line 62: Is `10.128.0.0` generic for all users?

src/drt/test/Distributed/k8s-drt.yaml
```diff
@@ -35,12 +35,13 @@ spec:
       hostNetwork: true
       containers:
         - name: openroad
-          image: openroad/centos-binary:latest
+          image: openroad/orfs:v3.0-1871-g4e29c449
```
maliberty, line 38: Will this image always exist?

src/drt/doc/Distributed.md
osamahammad21: not really. it depends on the private network local ip range.

src/drt/test/Distributed/k8s-drt.yaml
maliberty: @vvbandeira @sombraSoft  what do you recommend here for the tag?

vvbandeira: The image will exist for as long as DockerHub maintains the current policy of allowing unlimited images for open-source projects.

One case would be to use the base ORFS image and compile OR on the fly to guarantee that the code being tested is the current checkout code. The other case might be to use `orfs:latest` and include this in the CI for ORFS and OR to ensure it will always work.

maliberty: @vvbandeira  `orfs:latest` does not appear to be updating as Osama commented.  See https://hub.docker.com/r/openroad/orfs/tags?name=latest

image

src/drt/doc/Distributed.md
maliberty: Then the docs should explain that the users needs to update this as well as PATH.

osamahammad21:
@maliberty re-review please.

maliberty (approved):

PR #6285, opened 2024-12-03
Gpl update large tests

gudeh:
I believe this update was required since the change in commit: https://github.com/The-OpenROAD-Project/OpenROAD/commit/d578d454be4b074dd6b42fb1fa39352a4dd264ab.

The pipeline of [OpenROAD-Nightly-Public](https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-Nightly-Public/activity) started failing from ID 1847, which had the following changes, including the mentioned gpl commit: 
image

### DIFF
--- a/src/drt/src/dr/FlexDR.h
+++ b/src/drt/src/dr/FlexDR.h
@@ -409,6 +409,9 @@ class FlexDRWorker
   std::vector<std::unique_ptr<drNet>>& getNets() { return nets_; }
   const std::vector<drNet*>* getDRNets(frNet* net) const
   {
+    if (net == nullptr || net->isSpecial()) {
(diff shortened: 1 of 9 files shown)


gudeh:
I see a "ERROR: File src/gpl/test/large02.defok is big but not whitelisted (hash 61bc5fe661f07ec1916bec73174d8352)". Should I do something special to such large def file?

maliberty:
> I see a "ERROR: File src/gpl/test/large02.defok is big but not whitelisted (hash 61bc5fe661f07ec1916bec73174d8352)". Should I do something special to such large def file?

I'll take care of that.

maliberty (approved):

PR #6289, opened 2024-12-03
rsz: dont touch special nets

gadfort:
Changes:
- dont touch special nets in resizer, during eliminate_dead_logic the power nets get deleted without this.

### DIFF
--- a/src/rsz/src/Resizer.cc
+++ b/src/rsz/src/Resizer.cc
@@ -1675,7 +1675,7 @@ void Resizer::setDontTouch(const Net* net, bool dont_touch)
 bool Resizer::dontTouch(const Net* net)
 {
   dbNet* db_net = db_network_->staToDb(net);
-  return db_net->isDoNotTouch();
+  return db_net->isDoNotTouch() || db_net->isSpecial();
 }
 
 ////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


gadfort:
@QuantamHD there are plenty of cases were we want to be able to delete and modify special nets, but resizer is not one of those tools

QuantamHD:
You know what we should have... https://man.openbsd.org/pledge.2

We should add something to ODB that let's us pledge away certain thing like deleting special nets in resizer. Then remove the pledges once rsz is done.

gadfort:
@QuantamHD I think that is beyond the scope of this PR. I would suggest creating an issue since that would be a larger task

maliberty:
This sounds like a bug in eliminate dead logic.  The power nets shouldn't be fully disconnected and so shouldn't be deleted.  I'm wondering if this fixes a symptom.  @povik please comment.

povik:
@maliberty Yes, I think the root cause is eliminate_dead_logic ignoring pins if their direction is none of input, output, tristate, bidirectional, but there are also directions ground/power. I can write a fix later.

Not sure if we want to conflate special nets with dont touch for the following reason: if I remember right if an instance is connected to a dont touch net on any pin, that instance cannot be changed (other than maybe sizing?), special nets don't have the same restriction

gadfort:
Closed in favor of: #6314

PR #6300, opened 2024-12-04
odb: moving dbModule::makeUniqueDbModule()::name_id_map to dbBlock 

LucasYuki:
Removes the static variable name_id_map from the dbModule::makeUniqueDbModule method (issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/5981)

Variable name_id_map moved to dbBlock, as suggested by @maliberty in https://github.com/The-OpenROAD-Project/OpenROAD/pull/6245#issuecomment-2501902499, and is now accessed via the dbBlock::getModuleNameIdMap() method.

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -1694,11 +1694,6 @@ class dbBlock : public dbObject
 
   void clearUserInstFlags();
 
-  ///
-  /// return the module name id map
-  ///
-  std::map<std::string, int>* getModuleNameIdMap();
-
  public:
   ///
   /// This method copies the via-table from the src block to the destination
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -1694,6 +1694,11 @@ class dbBlock : public dbObject
 
   void clearUserInstFlags();
 
+  ///
+  /// return the module name id map
+  ///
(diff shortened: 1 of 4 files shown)


src/odb/include/odb/db.h
```diff
@@ -1694,6 +1694,11 @@ class dbBlock : public dbObject
 
   void clearUserInstFlags();
 
+  ///
+  /// return the module name id map
+  ///
+  std::map<std::string, int>* getModuleNameIdMap();
```
maliberty, line 1700: This is an internal implementation detail and doesn't need to be in the public interface.

src/odb/src/db/dbModule.cpp
```diff
@@ -549,12 +549,12 @@ dbModule* dbModule::makeUniqueDbModule(const char* cell_name,
                                        dbBlock* block)
 
 {
-  static std::map<std::string, int> name_id_map;
   dbModule* module = dbModule::create(block, cell_name);
   if (module != nullptr) {
     return module;
   }
 
+  std::map<std::string, int>& name_id_map = *(block->getModuleNameIdMap());
```
maliberty, line 557: Just cast block to its private class and access the field directly

maliberty (approved):

PR #6307, opened 2024-12-05
clock tree: default to never show

oharboe:
Problems addressed:

- I have seen users asking how to disable the clock tree. Now they need to discover how to enable it, where right clicking isn't such a far stretch of the imagination, at which point they also learn how to disable it
- save_clocktree_image saves the same thing regardless of GUI state
- save_clocktree_image still saves the clock tree, but restores the clock tree show status state afterwards
- no need to extend the .tcl interface


gadfort:
@oharboe this does not work as intended. The rendered clocktree is not saved during that call, so its pointless to change it's state.

I prefer the default to be on.

oharboe:
> @oharboe this does not work as intended. The rendered clocktree is not saved during that call, so its pointless to change it's state.

Bugs can be fixed, not worried about that.

> I prefer the default to be on.

Do you have any insights into how I can save the clock tree without levaing it on?

I do think that it is less confusing to have it off by default and let the user enable it, so that he knows how to disable it.

gadfort:
@oharboe to preserve the capability to saving the rendered clock tree, you will need to expose the knob to enable/disable rendering to tcl. Otherwise this is loosing functionality

oharboe:
Thanks! Independently of that, IMO, and there is some evidence from users, that the current behavior is confusing. I think for the aforementioned reasions that default off is best.

gadfort (changes requested):
At a minimum this needs to be controllable from TCL. And there is no need to turn it on or off during `saveImage` as this does not control anything this image is generating.

oharboe:
Something else is needed to solve these problems, but I'd still like to hear thoughts from @maliberty before I close this PR.

oharboe:
> Would having a button to control the visibility be more obvious? Something like we have in the hierarchy view?

I still think that not having the tree enabled is better as a default.

PR #6308, opened 2024-12-05
nix: update to nixos 24.11, fix build

donn:
* updated nix package set to nixos 24.11
  * changed compiler to clang 17
  * bumped boost to 1.86
* src/gui/src/tclCmdInputWidget.h: add workaround for swig issue https://github.com/swig/swig/issues/2887
* work around a number of fmt peculiarities
  * src/odb/src/db/dbMarker.cpp: use .value() for two logging instances where the optional variable has already been unwrapped as some combinations of boost and fmt cannot format boost::optional
  * src/drt/src/dr/FlexGridGraph_maze.cpp: explicitly cast enum to int because fmt apparently stopped doing so started fmt 10

--

I'm not sure why the fmt issues are happening. I tried downgrading and upgrading fmt to 9 and 11 (Nix's default is 10) and neither helped.

(cc @widlarizer / thanks for tracking down the SWIG patch from freebsd)

### DIFF
--- a/src/drt/src/dr/FlexGridGraph_maze.cpp
+++ b/src/drt/src/dr/FlexGridGraph_maze.cpp
@@ -58,7 +58,7 @@ void FlexGridGraph::printExpansion(const FlexWavefrontGrid& currGrid,
       pt,
       currGrid.getCost(),
       currGrid.getPathCost(),
-      currGrid.getLastDir(),
+      int(currGrid.getLastDir()),
       currGrid.getCost() - currGrid.getPathCost(),
       gridX,
       gridY);
--- a/src/gui/src/tclCmdInputWidget.h
+++ b/src/gui/src/tclCmdInputWidget.h
@@ -33,6 +33,10 @@
 #pragma once
 
 #include <tcl.h>
+// Workaround for https://github.com/swig/swig/issues/2887
+#ifndef TCL_SIZE_MAX
+using Tcl_Size = int;
+#endif
 
 #include <QCompleter>
 #include <QMenu>
--- a/default.nix
+++ b/default.nix
@@ -35,10 +35,10 @@
 {
   flake,
   lib,
-  clangStdenv,
+  llvmPackages_17,
   fetchFromGitHub,
   libsForQt5,
-  boost183,
+  boost186,
   eigen,
   cudd,
   ninja,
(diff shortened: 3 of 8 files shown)


src/gui/src/tclCmdInputWidget.h
```diff
@@ -33,6 +33,10 @@
 #pragma once
 
 #include <tcl.h>
+// Workaround for https://github.com/swig/swig/issues/2887
+#ifndef TCL_SIZE_MAX
+typedef int Tcl_Size;
```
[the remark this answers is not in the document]
donn: sure

src/drt/src/dr/FlexGridGraph_maze.cpp
```diff
@@ -58,7 +58,7 @@ void FlexGridGraph::printExpansion(const FlexWavefrontGrid& currGrid,
       pt,
       currGrid.getCost(),
       currGrid.getPathCost(),
-      currGrid.getLastDir(),
+      int(currGrid.getLastDir()),
```
maliberty, line 61: I'm in the process of updating spdlog which should obviate this need.  Are you updating spdlog in this PR?  With the current 1.8.1 this shouldn't be needed.  Likewise elsewhere.

donn: Ah, heck, why not. Let me try and get back to you.

donn: Oh, I misunderstood the message. Yeah, spdlog is part of NixOS so we're not explicitly controlling its version (It's 1.14.something). I will try a build with 1.8.1. Even then however, I would need a matching `fmt`, so it quickly starts spiraling.

widlarizer:
With this PR, I'm seeing this:

```
openroad> /nix/store/bwkb907myixfzzykp21m9iczkhrq5pfy-binutils-2.43.1/bin/ld: warning: libabsl_time_zone.so.2407.0.0, needed by /nix/store/rf913n1a5qwj5pa16b0fr0368430pzb0-re2-2024-07-02/lib/libre2.so.11.0.0, may conflict with libabsl_time_zone.so.2301.0.0
openroad> ERROR: Something is wrong with flag 'flagfile' in file 'build/source/absl/flags/parse.cc'. One possibility: file 'build/source/absl/flags/parse.cc' is being linked both statically and dynamically into this executable. e.g. some files listed as srcs to a test and also listed as srcs of some shared lib deps of the same test.
```
when I run `nix build ".?submodules=1#" -L` and there's a lot of those ld warnings about abseil components

donn:
I'll investigate

donn:
So the funny thing is or-tools in Nix is a bit messy and used an old version of protobuf and abseil which conflicted with the one used by re2 (despite or-tools itself depending on re2.)

I just updated it to v9.11 and removed all of the Python binding builds because they're a PITA to build and are not needed for OpenROAD.

donn:
`master` builds now. Older NixOS, but with the current `or-tools` fiasco (out-of-date and not building with Clang on Linux) I'm just gonna not worry about this PR for now.

PR #6310, opened 2024-12-05
psm: add ability to assign power to instances

gadfort:
Adds:
- capability to assign power to specific instances for analysis

Notes:
- this is currently not a user facing function, I think we can debate if we want it to be one, but it's mainly there for debugging.

### DIFF
--- a/src/psm/src/pdnsim.tcl
+++ b/src/psm/src/pdnsim.tcl
@@ -226,6 +226,25 @@ proc set_pdnsim_net_voltage { args } {
   }
 }
 
+sta::define_cmd_args "set_pdnsim_inst_power" {
+  -inst instance
+  -power power
+  [-corner corner]}
+
+proc set_pdnsim_inst_power { args } {
+  sta::parse_key_args "set_pdnsim_inst_power" args \
+    keys {-inst -corner -power} flags {}
+  if { [info exists keys(-inst)] && [info exists keys(-power)] } {
+    set inst [psm::find_inst $keys(-inst)]
+    set power $keys(-power)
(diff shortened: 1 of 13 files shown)


maliberty:
What use model motivates this change?  Is this some sort of what-if?

gadfort:
@maliberty it's a what-if kind of check at the moment, either because the is no liberty model available for sta or because we're trying to stress something to an extreme.
If you think it would be worth having a "proper" interface for it I can certainly add one (if you have suggestions for the name of the command that would be helpful).

maliberty:
set_inst_power is fine.  Please add a tcl wrapper and docs

gadfort:
@maliberty command added

src/psm/src/pdnsim.tcl
```diff
@@ -226,6 +226,25 @@ proc set_pdnsim_net_voltage { args } {
...
+  [-corner corner]}
+
+proc set_pdnsim_inst_power { args } {
+  sta::parse_key_args "set_pdnsim_inst_power" args \
+    keys {-inst -corner -power} flags {}
+  if { [info exists keys(-inst)] && [info exists keys(-power)] } {
+    set inst [psm::find_inst $keys(-inst)]
+    set power $keys(-power)
+    set corner [sta::parse_corner_or_all keys]
+    psm::set_inst_power $inst $corner $power
```
maliberty, line 241: What if the instance or corner are not found?

gadfort: instance not found is handled in `psm::find_inst` and corner is handled in `sta::parse_corner_or_all`

maliberty (approved):

PR #6313, opened 2024-12-05
ctest: move all tools to ctest regression

gadfort:
Changes:
- removes the use of the previous `regression_tests.tcl`
- Switches ctests generation to build both tcl and python independently.

Notes:
- flow tests are not converted, but should not require much additional effort to migrate (this would be needed to retire the entire tcl based system).
- rmp needed some modifications since both tcl and python scripts edit the same files.


maliberty:
@gadfort conflicts to resolve.

gadfort:
@maliberty conflict resolved

Jenkinsfile
maliberty, line 1: @vvbandeira any concerns here?

maliberty (approved):

Jenkinsfile
vvbandeira: LGTM

PR #6319, opened 2024-12-06
ord: allow for location placement in place_inst

gadfort:
Changes:
- adds `-location` to `place_inst` to allow for placement against the LL instead of origin only
- Unit test for functionality


src/OpenRoad.tcl
```diff
@@ -426,8 +441,12 @@ proc place_inst { args } {
     utl::error ORD 65 "Cannot create instance $inst_name of $cell_name."
   }
 
-  $inst setOrigin $x $y
   $inst setOrient $orient
+  if { [info exists keys(-origin)] } {
+    $inst setOrigin $x $y
```
maliberty, line 446: setOrient & setOrigin are not commutative

gadfort: What do you mean? Looking at the code for these two functions it doesn't look like either depends on the other. setLocation uses Origin which means order matters for that.

maliberty: You are correct.

maliberty (approved):

PR #6322, opened 2024-12-06
ord: ensure global connects are updated when nets are destroyed

gadfort:
I noticed that the global connects hold on to the old information after a net is destroyed (usually causing a segfault).
This fixes that by ensuring that the global connects are checked during dbNet destroy.


maliberty (approved):

maliberty:
When do we destroy such a net?

gadfort:
> When do we destroy such a net?

I noticed this because of the `eleminate_dead_logic` bug. Otherwise we don't.

PR #6326, opened 2024-12-08
drt: elimination of getNoRotationTransform and drt::dbTransform

### LINKED ISSUE #5712: DRT: resolve conflicts between frInst::getTransform and frInst::getUpdatedXform

bnmfw:
This PR has an ISPD and Secure CI associated and will be opened when they pass.
Fixes #5712.

### Elimination of getNoRotationTransform
Previously on #6141 most use cases of `getUpdatedXform` were refactored to use the `odb::dbTransform` object. This allowed for the refactoring of `getUpdatedXform` into `getNoRotationTransform` which was the use case remaining. The refactored function was only using when recalculation the access points coordinates.

During `revertAccessPoints` chip coordinates are transformed to take the origin of their instance as the 0,0. However, the function was taking the incorrect `frInst::dbTransform` as a basis and not taking rotation into consideration. This mistake was compensated by `getNoRotationTransform`. Now `revertAccessPoints` takes the DB transformation as a basis, which allows the substitution of `getNoRotationTransform` to `getDBTransform` which is correct and inline with the db.

The remaining use of `frInst::dbTranform` was at `frInstTerm::getAccessPoint` where an inverse transformation was done manually. The adequate transformation was used instead.

After this modification `frInst::dbTransform` was never accessed, so it was eliminated from the class. This allowed the renaming of `frInst::getDBTransform` to `frInst::getTransform` since it now is the only transform `frInst` has access to. `frInst::setOrient` and `frInst::setOrigin` were deleted since at no part of the code the DRT actually has to change the `odb::dbTransform` anymore. This deletion required deleting the methods from the abstract base class `frRef`, which in part required the same methods on `frVia` to not be overrides anymore.

Finally at some parts of the code were previously both `getUpdatedXform` and `getTransform` were called, only the latter is necessary since their uses are the same, thus only one transformation is used and substitutes the deleted. Some variable name changes were done (Mainly `shiftXform` -> `transform`). The way the global router orients access points was also changed to conform with these changes.


src/drt/src/db/obj/frVia.h
```diff
@@ -102,11 +102,11 @@ class frVia : public frRef
    */
 
   dbOrientType getOrient() const override { return dbOrientType(); }
-  void setOrient(const dbOrientType& tmpOrient) override { ; }
+  void setOrient(const dbOrientType& tmpOrient) { ; }
```
osamahammad21, line 105: why keep it?

src/drt/src/db/obj/frVia.h
```diff
@@ -102,11 +102,11 @@ class frVia : public frRef
...
   Point getOrigin() const override { return origin_; }
-  void setOrigin(const Point& tmpPoint) override { origin_ = tmpPoint; }
+  void setOrigin(const Point& tmpPoint) { origin_ = tmpPoint; }
   dbTransform getTransform() const override { return origin_; }
-  void setTransform(const dbTransform& xformIn) override {}
+  void setTransform(const dbTransform& xformIn) {}
```
osamahammad21, line 109: ditto

src/drt/src/db/obj/frRef.h
```diff
@@ -39,10 +39,6 @@ class frRef : public frPinFig
   virtual dbOrientType getOrient() const = 0;
   virtual Point getOrigin() const = 0;
   virtual dbTransform getTransform() const = 0;
-  // setters
-  virtual void setOrient(const dbOrientType& tmpOrient) = 0;
-  virtual void setOrigin(const Point& tmpPoint) = 0;
-  virtual void setTransform(const dbTransform& xform) = 0;
```
osamahammad21, line 45: why remove this

src/drt/src/pa/FlexPA_prep.cpp
```diff
@@ -1789,9 +1789,8 @@ void FlexPA::revertAccessPoints()
 {
   const auto& unique = unique_insts_.getUnique();
   for (auto& inst : unique) {
-    const dbTransform xform = inst->getTransform();
-    const Point offset(xform.getOffset());
-    dbTransform revertXform(Point(-offset.getX(), -offset.getY()));
+    dbTransform revertXform;
```
osamahammad21, line 1792: follow naming standards
```suggestion
    dbTransform revert_transform;
```

osamahammad21 (changes requested):

src/drt/src/db/obj/frVia.h
bnmfw: Great question, I kept them just because they were already there and I wanted to make minimal changes to `frVia`, but you are right, I should remove them

Issue #6316, opened 2024-12-06
make grt::have_routes return silent 0 when elements are not placed


oharboe:
### Description

```
make gui_floorplan
```

Results in:

```
>>> grt::have_routes
[ERROR GRT-0010] Instance ctrl.state.out\[0\]$_DFF_P_ is not placed.
[ERROR GUI-0070] GRT-0010
```

### Suggested Solution

silently return 0

### Additional Context

Useful in scripting

maliberty:
Didn't you already make a PR for this?

oharboe:
> Didn't you already make a PR for this?

This is the next bug in the same vein

maliberty:
Are you planning to fix this?  I assigned the last one only to have you fix it.

oharboe:
> Are you planning to fix this?  I assigned the last one only to have you fix it.

No, I have a workaround. The last one turned out to be trivial to fix and it was blocking me.

This one is more about consistency, hopefully trivial too though.

PR #6333, opened 2024-12-09
grt: check if design is placed for grt::have_routes

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/6316


precisionmoon (approved):
Approved

tspyrou (approved):

PR #6336, opened 2024-12-11
rsz: add report_dont_* commands

gadfort:
Adds:
- `report_dont_use` and `report_dont_touch` to provide feedback about the instance, cells, and nets that have been marked by resizer.

### DIFF
--- a/src/rsz/src/Resizer.i
+++ b/src/rsz/src/Resizer.i
@@ -467,6 +467,7 @@ report_dont_use()
 void
 report_dont_touch()
 {
+  ensureLinked();
   Resizer *resizer = getResizer();
   resizer->reportDontTouch();
 }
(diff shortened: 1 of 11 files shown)


src/rsz/src/Resizer.i
```diff
@@ -457,6 +457,20 @@ set_dont_touch_instance(Instance *inst,
...
+report_dont_use()
+{
+  Resizer *resizer = getResizer();
+  resizer->reportDontUse();
+}
+
+void
+report_dont_touch()
+{
+  Resizer *resizer = getResizer();
```
povik, line 470: I think this is missing `ensureLinked();` and that's the reason for the CI failure (it crashes in `commands_without_load.tcl`)

gadfort: thanks

PR #6343, opened 2024-12-11
rsz: use masters for report_dont_use to avoid listing cells by corner

gadfort:
Changes:
- use masters to uniquely report dont use cells, otherwise one per corner appears in the report


src/rsz/src/Resizer.cc
```diff
@@ -1810,8 +1810,14 @@ void Resizer::reportDontUse() const
   if (dont_use_.empty()) {
     logger_->report("  none");
   } else {
+    std::set<dbMaster*> cells;
+
     for (auto* cell : dont_use_) {
-      logger_->report("  {}", cell->name());
+      cells.insert(db_network_->staToDb(cell));
```
povik, line 1816: I think `if(isLinkCell(cell))` also filters out the per-corner copies

maliberty: @gadfort this seems correct

gadfort: done

src/rsz/src/Resizer.cc
```diff
@@ -1803,14 +1803,17 @@ bool Resizer::dontUse(LibertyCell* cell)
   return cell->dontUse() || dont_use_.hasKey(cell);
 }
 
-void Resizer::reportDontUse() const
+void Resizer::reportDontUse()
```
maliberty, line 1806: I think you can make Resizer::isLinkCell const rather than drop const here.

maliberty (approved):

PR #6349, opened 2024-12-13
etc: Update spdlog also when installed version does not match.

titan73:


titan73:
Any feedback on this PR?

PR #6351, opened 2024-12-13
gpl: debug options for overflow

gudeh:
These modifications does not change any functionality.

- Add debug prints for overflow calculation values.
- Show microns instead of dbu for bins in debug mode on gui.

### DIFF
--- a/src/gpl/src/graphics.cpp
+++ b/src/gpl/src/graphics.cpp
@@ -454,7 +454,8 @@ bool Graphics::populateMap()
           0.0f,
           static_cast<float>(bin.instPlacedAreaUnscaled())
               + static_cast<float>(bin.nonPlaceAreaUnscaled()) - scaledBinArea);
-      addToMap(box, value);
+      odb::dbBlock* block = pbc_->db()->getChip()->getBlock();
+      addToMap(box, block->dbuAreaToMicrons(value));
     }
   }
 
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -667,12 +667,12 @@ int BinGrid::binSizeY() const
 
 int64_t BinGrid::overflowArea() const
 {
-  return overflowArea_;
+  return sumOverflowArea_;
 }
 
 int64_t BinGrid::overflowAreaUnscaled() const
 {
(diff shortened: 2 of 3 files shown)


eder-matheus (approved):

src/gpl/src/graphics.cpp
```diff
@@ -454,7 +454,8 @@ bool Graphics::populateMap()
           0.0f,
           static_cast<float>(bin.instPlacedAreaUnscaled())
               + static_cast<float>(bin.nonPlaceAreaUnscaled()) - scaledBinArea);
-      addToMap(box, value);
+      odb::dbBlock* block = pbc_->db()->getChip()->getBlock();
+      addToMap(box, block->dbuAreaToMicrons(value));
```
maliberty, line 458: @gadfort is this right or should use_dbu_ from the base HeatMapDataSource come into play?

andyfox-rushc:
I sincerely hope this is (my) pilot error. But in debug build of master I am seeing lots of failures in formatting.

For example ppl add_constraint1:
[INFO PPL-0067] Restrict INPUT pins to region 0u-100.13u, in the TOP edge.
is printed
whereas before it was:
! [INFO PPL-0067] Restrict INPUT pins to region 0.0u-100.13u, in the TOP edge.

I am putting this comment here because this pr seems to relate to some printing change. The precision change from 0.0u <-> 0u seems to be some minor formatting issue. No big deal for me, but it is causing some noise on my local regression. Perhaps this is some kind of setup/package printing issue.

rovinski:
@andyfox-rushc the default version of spdlog was updated which changes the way some values are printed.

andyfox-rushc:
Thank you for the heads up. I found that running ./etc/DependencyInstaller.sh did not do it for me and I had to manually remove spdlog (from usr/local)  and then re-run ./etc/DependencyInstaller.sh
Thank you again for the helpful pointer, am now back to running my regressions without noise !

src/gpl/src/graphics.cpp
gadfort: I think that is correct. It's just changing that information is stored in the heatmap. Which should not depend on use_dbu, however, the display and scaling code could be modified to honor that, in which case it would be easier to store the unscaled data and scale it at the end.

PR #6369, opened 2024-12-17
mpl2: guidance regions

### LINKED ISSUE #6200: MPL2: Allow guidance regions to be used
### Description

We have internal cpp support for guidance regions but it's not available to the user in .TCL.
There are bugs that need fixing for it to work as well.
The idea is: the user should be able to specify a guidance region for each macro in order to help guiding the macro placer.

### Suggested Solution

Complete the unfinished support for the feature.

### Additional Context

This could be an alternative to #6033. _To be tested and verified._

joaomai:
Fixes #6200.
This PR updates and exposes the existing guidance regions API that existed inside MPL2 to be used through the Tcl interface.

The user has to set the guides using the command `set_macro_guidance_region`, along with the macro name and region:
`set_macro_guidance_region -macro_name MACRO_1 -region {0 0 100 100}`

The default weight for guidance regions is 10, but it can be adjusted using `-guidance_weight` when calling `rtl_macro_placer`:
`rtl_macro_placer ... -guidance_weight 20.0`


src/mpl2/src/rtl_mp.cpp
```diff
@@ -190,6 +191,39 @@ void MacroPlacer2::placeMacro(odb::dbInst* inst,
                 orientation.getString());
 }
 
+void MacroPlacer2::addGuidanceRegion(odb::dbInst* macro, const Rect& region)
+{
+  odb::dbBlock* block = db_->getChip()->getBlock();
+  const odb::Rect& core = block->getCoreArea();
+  odb::Rect dbu_region(block->micronsToDbu(region.xMin()),
```
AcKoucher, line 198: const

src/mpl2/test/guides1.tcl
```diff
@@ -0,0 +1,25 @@
+# Test if pin access blockage is generated correctly for a case
+# with pins in a single boundary.
```
AcKoucher, line 2: This comment is not related to what is being tested here. We should add a similar comment, but wrt what's the intention of this test.

src/mpl2/test/guides1.tcl
```diff
@@ -0,0 +1,25 @@
...
+
+read_lef "./testcases/macro_only.lef"
+read_liberty "./testcases/macro_only.lib"
+
+read_verilog "./testcases/io_constraints1.v"
+link_design "io_constraints1"
+read_def "./testcases/io_constraints1.def" -floorplan_initialize
+
+set_io_pin_constraint -direction INPUT -region left:*
+place_pins -annealing -random -hor_layers metal5 -ver_layer metal6
```
AcKoucher, line 16: Unneeded wrt what's being tested.

src/mpl2/test/guides2.tcl
```diff
@@ -0,0 +1,25 @@
+# Test if pin access blockage is generated correctly for a case
+# with pins in a single boundary.
```
AcKoucher, line 2: Same as the other test. Add a meaningful comment wrt this test.

src/mpl2/test/guides2.tcl
```diff
@@ -0,0 +1,25 @@
...
+
+read_lef "./testcases/macro_only.lef"
+read_liberty "./testcases/macro_only.lib"
+
+read_verilog "./testcases/io_constraints1.v"
+link_design "io_constraints1"
+read_def "./testcases/io_constraints1.def" -floorplan_initialize
+
+set_io_pin_constraint -direction INPUT -region left:*
+place_pins -annealing -random -hor_layers metal5 -ver_layer metal6
```
AcKoucher, line 16: Unneeded as in the other test.

AcKoucher (changes requested):
The only difference between the two tests looks to be regarding the guidance region set for the macro. I think guides2 would be better of as a test for macro placement - isolated from cluster placement (see macro_only.tcl and the special case in ClusteringEngine::run() in which we skip cluster placement).

AcKoucher:
@eder-matheus any idea of what's causing the Docs failure here?

eder-matheus:
> @eder-matheus any idea of what's causing the Docs failure here?

I was looking at it before, the error is pretty odd:
```
set_macro_guidance_region
command not parsed in proc_dict
Failed.
Help Dict: { [-macro_name macro_name] [-region region] }
Proc Dict: Not found
Readme Dict: -macro_name macro_name -region region
```

When I look at the .tcl and the readme file, they seem to be matching. Not sure if it's a bug in the Docs Tester or if we're missing something.

eder-matheus:
> ```
> set_macro_guidance_region
> command not parsed in proc_dict
> Failed.
> Help Dict: { [-macro_name macro_name] [-region region] }
> Proc Dict: Not found
> Readme Dict: -macro_name macro_name -region region
> ```

@AcKoucher @joaomai Now I understand what's happening. In the Tcl file, you are defining the keywords between **[ ]**, while in the readme they are not inside it. You should not use this in the Tcl file, unless the keyword is optional.

joaomai:
@eder-matheus still complains after removing the square brackets from the .tcl file.

src/mpl2/src/mpl.tcl
```diff
@@ -294,6 +294,45 @@ proc place_macro { args } {
   mpl2::place_macro $macro $x_origin $y_origin $orientation
 }
 
+sta::define_cmd_args "set_macro_guidance_region" { -macro_name macro_name \
+                                                   -region region }
+proc set_macro_guidance_region { args } {
+  sta::parse_key_args "set_macro_guidance_region" args \
+    keys { -macro_name -region } \
+    sta::check_argc_eq0 "set_macro_guidance_region" $args
```
eder-matheus, line 302: I think you need to add `flags {}` to indicate there are not flags in this command. At least this will follow the pattern from other commands.

Also, the `sta::check_argc_eq0` should not be on the same line of the `keys`. You should remove the `\` from the end of the line 301.

AcKoucher (changes requested):
Please, as you resolve the reviews' items code wise, also remember to resolve the GH comments. It makes the PR conversation more concise.

src/mpl2/test/guides1.tcl
```diff
@@ -0,0 +1,25 @@
...
+
+read_lef "./testcases/macro_only.lef"
+read_liberty "./testcases/macro_only.lib"
+
+read_verilog "./testcases/io_constraints1.v"
+link_design "io_constraints1"
+read_def "./testcases/io_constraints1.def" -floorplan_initialize
+
+#set_io_pin_constraint -direction INPUT -region left:*
+#place_pins -annealing -random -hor_layers metal5 -ver_layer metal6
```
AcKoucher, line 16: Sorry, it looks like the test will error, because mpl2 will be unhappy with unplaced pins. Please, leave the PPL cmd as it was but adapt the guide region to occupy the lower right corner of the core.

AcKoucher (changes requested):
I think we should also change the names of the test and its files to guides1. ioconstraint1 doesn't make sense as it refers to a different PR changes.

src/mpl2/src/object.h
```diff
@@ -778,17 +778,18 @@ struct Rect
...
 
-  bool isValid() const
+  bool isValid() const { return (lx < ux) && (ly < uy); }
+
+  void mergeInit()
   {
-    return (lx > 0.0) && (ly > 0.0) && (ux > 0.0) && (uy > 0.0);
+    lx = std::numeric_limits<float>::max();
+    ly = lx;
+    ux = std::numeric_limits<float>::min();
```
maliberty, line 787: This is a classic C++ trap.  You want ::lowest not ::min.  More annoyingly ::min works fine for integral types.

src/mpl2/src/object.h
```diff
@@ -778,17 +778,18 @@ struct Rect
     f_y = f_y_;
   }
 
-  bool isValid() const
+  bool isValid() const { return (lx < ux) && (ly < uy); }
```
maliberty, line 781: Is zero width or height invalid?

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -1257,15 +1262,18 @@ void HierRTLMP::runHierarchicalMacroPlacement(Cluster* parent)
     if (cluster->getClusterType() == StdCellCluster) {
       continue;
     }
-    Rect fence(-1.0, -1.0, -1.0, -1.0);
-    Rect guide(-1.0, -1.0, -1.0, -1.0);
+    Rect fence, guide;
+    fence.mergeInit();
+    guide.mergeInit();
```
maliberty, line 1267: If the guide remains in the init state with no merges, does the subsequent relocate call confuse isValid?

maliberty (changes requested):

src/mpl2/test/regression_tests.tcl
```diff
@@ -1,6 +1,8 @@
 record_tests {
     macro_only
     no_unfixed_macros
+    guides1
+    guides2
```
gadfort, line 5: add to ctest list too

src/mpl2/src/mpl.i
```diff
@@ -144,6 +144,16 @@ void set_debug_cmd(odb::dbBlock* block,
...
 
+void
+add_guidance_region(odb::dbInst* macro,
+                    float x1,
+                    float y1,
+                    float x2,
+                    float y2)
+{
+  getMacroPlacer2()->addGuidanceRegion(macro, Rect(x1, y1, x2, y2));
+}
```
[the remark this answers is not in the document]
joaomai: The guide region is stored using microns though, the conversion to dbu is only to check if the guide is inside the core area.

src/mpl2/src/object.h
joaomai: Considering it's used for guidance and fence regions only (for now), I think it's reasonable to make a zero height/width region invalid.

src/mpl2/src/hier_rtlmp.cpp
joaomai: Relocate checks if the rect is valid before relocating, so if no fence/guide was merged, it will do nothing.

src/mpl2/README.md
```diff
@@ -109,6 +109,23 @@ place_macro
...
+    -macro_name macro_name
+    -region region
+```
+
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-macro_name` | The name of a macro of the design. |
+| `-region` | The lower left corner and upper right corner {lx ly ux uy} of the region in microns. |
```
oharboe, line 127: Clarify: within which the macro must fit.

oharboe (commented):
There is a check in macro placement if there is a *chance* if the macros can fit(area + halo), is this check done for regions too?

src/mpl2/README.md
AcKoucher: https://github.com/joaomai/OpenROAD/blob/49ce3f117a13f28a3ae92fe4a214735927ba3bfa/src/mpl2/src/rtl_mp.cpp#L203-L217

@oharboe Is this what your asking about?

oharboe: No, if the macros for a guidance region cant possibly fit, it would be good with an early clear error, rather than a more obscure error much later.

src/mpl2/src/mpl.i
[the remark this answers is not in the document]
maliberty: Ok, I forget that mpl2 needs a total rework of its units.

src/mpl2/README.md
maliberty: The placer tries to maximize overlap with the region.  If the region is smaller then it says you are willing to accept any overlapping placement (ie there is some slop).

oharboe: > The placer tries to maximize overlap with the region. If the region is smaller then it says you are willing to accept any overlapping placement (ie there is some slop).

Seems useful to know... Clarify in docs?

joaomai: I'll update the command description.

maliberty (approved):

maliberty:
`Error: MPL 0037 used 2 times, next free message id is 43`

PR #6377, opened 2024-12-18
mpl2: update .ok to new logger behavior

AcKoucher:


AcKoucher:
I had the older version of spdlog in my machine. Master is green so this is probably the older behavior.

PR #6384, opened 2024-12-19
ord: add option to print stacktrace on exception

gadfort:
Changes:
- use `abort` instead of `exit(1)` to ensure memory errors print the stacktrace.
- print the stacktrace if "ORD trace" is set to >= 1.

Reason:
- since utl throws a `std::runtime_error` when `utl::error` is called this can mask real errors like `std::map::at` which also throws an `std::runtime_error` when accessing an incorrect element in the map. This change will allow us to still print the stacktrace making debugging of these issues possible.


src/odb/src/swig/tcl/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ swig_lib(NAME          odbtcl
          I_FILE        ../common/odb.i
          SWIG_INCLUDES ${PROJECT_SOURCE_DIR}/include
                        ${PROJECT_SOURCE_DIR}/src/swig/tcl
+         DEFINES       TEST_BUILD
```
maliberty, line 8: What is the purpose of TEST_BUILD?

gadfort: I wasn't able to generate a stub correctly, but I just needed a nights sleep. This parameter is gone now.

src/Exception.i
maliberty, line 1: Any reason not to do Exception-py.i ?

maliberty:
Would you explain "use abort instead of exit(1) to ensure memory errors print the stacktrace." more?  I don't get it.

gadfort:
> Would you explain "use abort instead of exit(1) to ensure memory errors print the stacktrace." more? I don't get it.

If we call exit(1) then the program just exits, if we call abort, the signal handler we have will capture that can print the stacktrace and then exit.

src/Exception.i
gadfort: not really, I didn't look at it, but I can make the same changes there

maliberty (approved):

Issue #5610, opened 2024-08-17
DRT Refactor


osamahammad21:
### Description

The refactor should achieve the following:
- Clear and Easy code to understand and maintain
- Less number of lines per method
- Cyclomatic complexity less than 20 (preferably less than 10)
- Doxygen Documentation for all functions/classes
- Following google code style c++
- Eliminate code duplication
- Unit testing
- clang-tidy clean
- Coverity clean

DRT follows modular architecture. The refactor should include the following modules:
- [ ] [io](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5611)
- [ ] pa
- [ ] [ta](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5614)
- [ ] rp
- [ ] dr

PR #6389, opened 2024-12-19
drt: slight routeNet_postAstarPatchMinAreaVio refactor

bnmfw:
This PR is ISPD and Secure CI safe. Tackles #5610.

While debugging #6326 I had to take a look at `routeNet_postAstarPatchMinAreaVio` where I noticed a weirdly verbose chunk of boolean logic that was also duplicated. I made a helper function with the condensed logic to eliminate duplication.

### DIFF
--- a/src/drt/src/dr/FlexDR_maze.cpp
+++ b/src/drt/src/dr/FlexDR_maze.cpp
@@ -3479,15 +3479,15 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio_helper(
   frArea gapArea = reqArea
                    - (currArea - startViaHalfEncArea - endViaHalfEncArea)
                    - std::min(startViaHalfEncArea, endViaHalfEncArea);
-  // new
-  bool bpPatchStyle = true;  // style 1: left only; 0: right only
-  bool epPatchStyle = false;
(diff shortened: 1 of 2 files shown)


src/drt/src/dr/FlexDR_maze.cpp
```diff
@@ -3523,87 +3451,75 @@ void FlexDRWorker::routeNet_postAstarPatchMinAreaVio(
...
+    int point_idx,
+    int prev_point_idx)
+{
+  FlexMazeIdx begin_point, end_point;
+  frArea gapArea = reqArea
+                   - (currArea - startViaHalfEncArea - endViaHalfEncArea)
+                   - std::min(startViaHalfEncArea, endViaHalfEncArea);
+  // new
+  bool bpPatchStyle = true;  // style 1: left only; 0: right only
+  bool epPatchStyle = false;
```
rovinski, line 3484: Consider using an enum for clarity?

bnmfw: enum made some parts of the code overly verbose, specially with multiple binary checks to set something to either enum. I did, however, change the variable name so its more like a boolean and specifies which is the true value.

osamahammad21 (approved):

Issue #6386, opened 2024-12-19
in global route, cap at 10000 random DRC errors instead of the first 10000 


oharboe:
### Description

10000 random DRC errors gives an overall impression of where the problem is, whereas 10000 will mislead the user to think that the problem is only in one particular part of the design.

Also, in the DRC errors in global routing there's can be DRC errors that could have been fixed, if it wasn't for some impossible problems.

I recently ran into a case where I had placed macros with very slight overlap(the intention was abutment), in this case the 10000 DRC errors did not include the errors in the macro overlap area until after I had run deltaDebug.py to whittle down the problem.

Here is an example from MegaBoom, we can clearly see that the global routing DRC failures stop at a vertical line in one of the DRC blobs. I thought that this was because of the 10000 DRC error cutoff and that there is in fact more global routing congestion, but then again, the placement heatmap has a similar vertical line.

![image](https://github.com/user-attachments/assets/894b2b09-193f-46d4-be2b-e745639c7598)

![image](https://github.com/user-attachments/assets/4c3f42e8-7485-4b6a-8b13-1fb88c7705c2)

![image](https://github.com/user-attachments/assets/43a356d6-747b-4c5b-b52e-7f9feb6cf7e7)

![image](https://github.com/user-attachments/assets/54ecf458-7919-4e48-9333-b81e6dc3d085)

### Suggested Solution

Ideally limit at the 10000 most interesting errors, but in lieu of that I'll take a cap of 10000 random DRC errors that gives me an overall impression of what is wrong.

PR #6402, opened 2024-12-20
grt: shuffle congestion markers for better visualization on gui

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/6386

### DIFF
--- a/src/grt/test/congestion7-20.rptok
+++ b/src/grt/test/congestion7-20.rptok
@@ -1,1288 +1,4 @@
 violation type: Horizontal congestion
-	srcs: net:_381_ net:_421_
-	comment: capacity:1 usage:2 overflow:1
-	bbox = (25.2000, 27.3000) - (27.3000, 29.4000) on Layer -
-violation type: Horizontal congestion
-	srcs: net:clk net:_370_
-	comment: capacity:1 usage:2 overflow:1
-	bbox = (25.2000, 35.7000) - (27.3000, 37.8000) on Layer -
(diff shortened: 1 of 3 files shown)


src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2488,6 +2490,13 @@ void FastRouteCore::saveCongestion(const int iter)
   std::vector<CongestionInformation> congestionGridsV, congestionGridsH;
   if (!h_edges_.empty() && !v_edges_.empty()) {
     getCongestionGrid(congestionGridsV, congestionGridsH);
+
+    std::mt19937 g;
+    const int seed = 42;
+    g.seed(seed);
+
+    utl::shuffle(congestionGridsV.begin(), congestionGridsV.end(), g);
+    utl::shuffle(congestionGridsH.begin(), congestionGridsH.end(), g);
```
maliberty, line 2499: There is still an ordering for V markers then H markers.  With many markers the V will hit the limit and block the H.

eder-matheus: Is that right? When I open the GUI, I see two different tabs for V and H, and both of them have 10000 occurrences.

maliberty: I didn't realize they are separate categories

maliberty:
grt.congestion7.tcl (Failed)

maliberty (approved):

PR #6403, opened 2024-12-20
mpl2: new flags for individual manual macro placement

### LINKED ISSUE #6385: manual macro placement with snapping to legal coordinates should complain about suspicious macro placements
### Description

place_macro will by default snap to legal coordinates, but it does not complain on overlapping macros.

Such an error on the user's part is annoying, because it will work fine all the way up until global routing, which will fail to converge. It can take hours to generate an error message about failure do global routing.

What is worse is that grt will fail in a very unobvious fashion: it can have tens of thousands of DRC errors as grt will stop trying early and what is shown as DRC errors are only the first 10000 horizontal and vertical DRC errors in global routing. These first 10000 DRC errors could have been easily resolved if it wasn't for the overlap, but the overlap DRC errors are not shown in the GUI as the DRC errors are capped at 10000. I was only able to figure this out by doing a deltaDebug.py run on grt + help from @maliberty 

Overlapping macros to me sounds suspicous and if is probably not what the user intended if they asked to have macros placed at legal coordinates.

I mention overlapping macros here as "supicious macro placement", but perhaps there are other things that can also be categorized as "supicious"?

### Suggested Solution

Add a `-literal` option that disables snapping and checks for sensible macro placements.

### Additional Context

_No response_

AcKoucher:
Resolve #6385 

Added:

| Flag | Behavior |
| ---------------- | ------------ |
|  -exact  | Enforce exact location set. Default is to snap. |
|  -allow_overlap  | Places macro if even if it overlaps with other macros. Default is to not allow the macro to be place if it generates overlap with other macros. |


AcKoucher:
@oharboe I thought perhaps `-literal` could be a bit confusing when using both flags at the same time, so I decided to go with more somewhat straightforward names.

src/mpl2/src/mpl.tcl
```diff
@@ -259,11 +259,13 @@ proc rtl_macro_placer { args } {
 sta::define_cmd_args "place_macro" {-macro_name macro_name \
                                     -location location \
                                     [-orientation orientation] \
+                                    [-dont_snap]
```
oharboe: I find negative names to cause more confusion and be harder to discuss in documentation and harder to remember. I would prefer a positive name, like `-exact` as in exact coordinates.

src/mpl2/README.md
```diff
@@ -108,6 +110,8 @@ place_macro
 | `-macro_name` | The name of a macro of the design. |
 | `-location` | The lower left corner of the macro in microns. |
 | `-orientation` | The orientation according to odb. If nothing is specified, defaults to `R0`.  We only allow `R0`, `MY`, `MX` and `R180`.  |
+| `-dont_snap` | Enforce the exact `-location`. Note that the default behavior of `place_macro` is to snap the macro, that is, tune the entered `-location` so that the signal pins get aligned with the track-grid. |
+| `-check_overlap` | Place macro only if no overlap with other macros is found. |
```
oharboe, line 114: I think this should be default and also to have a positve name, so how about `-allow_overlap`?

oharboe (changes requested):
I suggest using positive names, easier to document and reason about and to check default to check for overlap

oharboe (approved):
😌

eder-matheus (approved):

Issue #6406, opened 2024-12-21
cts: modifying cts_lib files does not rerun messages checking when building


titan73:
### Describe the bug

Modifying cts_lib files regenerates cts_lib.a but cts.a which depends on cts_lib.a is not which causes messages not to be checked.
Basically modifying TritonCTS.cpp with duplicated message ids will not be catched while the same in TritonCTS.tcl will.

### Expected Behavior

Duplicated message ids catches in cts_lib files.

### Environment

```shell
Latest OR
```

### To Reproduce

Make duplicate ids in TritonCTS.cpp and run make.

titan73:
Fixed by https://github.com/The-OpenROAD-Project/OpenROAD/pull/6468. Closing.

PR #6407, opened 2024-12-21
cts: Fix messages check.

titan73:
For https://github.com/The-OpenROAD-Project/OpenROAD/issues/6406.

This is similar to what is done for odb but in a generic way that can be reused.
The new messages_multi cmake function can probably be merged in the messages one.

### DIFF
--- a/src/cmake/messages.cmake
+++ b/src/cmake/messages.cmake
@@ -41,13 +41,14 @@
 #   SOURCE_DIR <dir>: the directory to search for sources from
 #                     [defaults to OUTPUT_DIR]
 #   LOCAL: don't recurse [defaults to false]
+#   DEPENDS: <dep1 ...>: Add extra dependencies [optional]
 
 function(messages)
 
   # Parse args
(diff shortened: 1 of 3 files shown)


src/cmake/messages.cmake
```diff
@@ -95,3 +95,65 @@ function(messages)
     WORKING_DIRECTORY ${SOURCE_DIR}
   )
 endfunction()
+
+function(messages_multi)
```
maliberty, line 99: This appears to copy the `messages` function just to add a new DEPENDS argument.  Could you just make take an option on the current one and handle if it is empty?

titan73: Resent PR which extends messages function, convert odb and fix cts. Updated PR title. Thanks.

titan73:
I created another PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6468 as I accidentally trashed my branch and couldn't push a typo fix.

PR #6415, opened 2024-12-22
Add a script to run clang-tidy over all files in parallel, cache the results

hzeller:
Running clang-tidy is very slow and time consuming. We're already running it as part of pull requests to incrementally point out issues, but sometimes it would be good to see an overall summary or simply a way to figure out which clang-tidy checks make sense to keep.

This adds a new script to etc/ to allow running clang-tidy in parallel, while also caching the results. So subsequent runs only need to process files that have changed.

For simplicity, this script is written in C++, but it can be invoked as a script as it is self-compiling. So it is as simple as running

```
  /bin/sh etc/run-clang-tidy-cached.cc
```

(or, with the executable bit set, you can invoke it
 even simpler as `etc/run-clang-tidy-cached.cc`)

Note: The first time, before there is a cache built, this will take a while.

This will generate two outputs

  * `OpenRoad_clang-tidy.summary` tallies up all findings and sorts them top first. This also allows to also easily see if there are clang-tidy rules that we might not be so interested in.
  * `OpenRoad_clang-tidy.out` contains all the findings in one file. This is neat to just put in the editor as 'compile output' and step through the results. (Personally, I use 'cat OpenRoad_clang-tidy.out' as `compile-command` in emacs, then it is simple to step from issue to issue).

Since clang-tidy needs the compilation database, it is first necessary to run `cmake` and make the comnpilation database available in the toplevel directory (it not already):

```
  ln -s build/compile_commands.json .
```

Here is a sample of the summary output running this script over all files.

```
Details: OpenRoad_clang-tidy.out
Summary: OpenRoad_clang-tidy.summary
---- Summary ----
 2339 [modernize-use-nullptr]
 1360 [modernize-use-default-member-init]
  609 [readability-else-after-return]
  549 [performance-avoid-endl]
  464 [modernize-use-override]
  451 [modernize-macro-to-enum]
  391 [readability-inconsistent-declaration-parameter-name]
  356 [modernize-use-using]
  351 [readability-redundant-casting]
  343 [clang-analyzer-security.insecureAPI.strcpy]
  327 [modernize-deprecated-headers]
  304 [clang-diagnostic-error]
  262 [performance-enum-size]
  252 [performance-unnecessary-value-param]
  242 [bugprone-suspicious-string-compare]
  211 [google-runtime-int]
  176 [readability-avoid-const-params-in-decls]
  153 [modernize-use-emplace]
  152 [google-readability-namespace-comments]
  129 [modernize-use-equals-default]
  117 [clang-analyzer-deadcode.DeadStores]
  108 [google-build-using-namespace]
   91 [modernize-use-bool-literals]
   88 [readability-braces-around-statements]
   87 [readability-const-return-type]
   84 [bugprone-multi-level-implicit-pointer-conversion]
   82 [readability-redundant-member-init]
   78 [bugprone-implicit-widening-of-multiplication-result]
   73 [bugprone-switch-missing-default-case]
   70 [bugprone-reserved-identifier]
   67 [misc-unused-using-decls]
   64 [google-default-arguments]
   62 [readability-redundant-declaration]
   60 [bugprone-assignment-in-if-condition]
   60 [modernize-loop-convert]
   48 [readability-suspicious-call-argument]
   46 [clang-analyzer-unix.Malloc]
 ...long tail...
```

The script is maintained in https://github.com/hzeller/dev-tools and just copied as-is to etc/ with the only change to run the local clang-format and change the start directory in the configuration direction to be `src/`.

### DIFF
--- a/docs/contrib/DeveloperGuide.md
+++ b/docs/contrib/DeveloperGuide.md
@@ -20,7 +20,7 @@ and headers together.
 
 - `src/`
    This folder contains the source files for individual tools.
- 
+
 | `src`           | Purpose |
 |-----------------|--------------|
 | `CMakeLists.txt`  | `add_subdirectory` for each tool|
@@ -245,12 +245,12 @@ file. Detailed documentation should be the `tool/README.md` file.
 Please refer to the README formatting [guide](ReadmeFormat.md).
 Our top-level READMEs, in particular, have to be formatted in this specific
 manner because of the automatic parsing used to convert the READMEs into
-manpages. 
+manpages.
 :::
 
 ## Tool Flow Namespace
 
-Tool namespaces are usually three-lettered lowercase letters. 
+Tool namespaces are usually three-lettered lowercase letters.
 
 - Verilog to DB (dbSTA)
 - OpenDB: Open Database ([odb](../main/src/odb/README.md))
@@ -305,7 +305,7 @@ dependencies make this vastly more complicated.
 1. `regression` script should only write files in a directory that is in the tool's `.gitignore` so the hierarchy does not have modified files in it as a result or running the regressions.
 1. Regressions report no memory errors with `valgrind` (stretch goal).
 1. Regressions report no memory leaks with `valgrind` (difficult).
-1. Ensure the top-level README and Tcl format are compliant. 
+1. Ensure the top-level README and Tcl format are compliant.
 
 ## Code Linting and Formatting
 
@@ -316,6 +316,16 @@ cmake . -B build  # generate build files
 # typically only run these commands on files you changed.
 clang-tidy -p ./build source_file.cpp
 clang-format -i -style=file:.clang-format source_file.cpp
+```
+
+To run `clang-tidy` on all files, you can use the following script that runs
+`clang-tidy` in parallel and also caches the results, so subsequent runs
+only have to operate on changed files.
+
+```shell
+cmake . -B build  # generate build files
+ln -sf build/compile_commands.json .  # make compilation db visible
+/bin/sh etc/run-clang-tidy-cached.cc
 ```
 
 ## Doxygen
--- a/etc/run-clang-tidy-cached.cc
+++ b/etc/run-clang-tidy-cached.cc
@@ -1,0 +1,628 @@
+#if 0  // Invoke with /bin/sh or simply add executable bit on this file on Unix.
+B=${0%%.cc}; [ "$B" -nt "$0" ] || c++ -std=c++17 -o"$B" "$0" && exec "$B" "$@";
+#endif
+// Copyright 2023 Henner Zeller <h.zeller@acm.org>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Location: https://github.com/hzeller/dev-tools (2024-12-23)
+
+// Script to run clang-tidy on files in a bazel project while caching the
+// results as clang-tidy can be pretty slow. The clang-tidy output messages
+// are content-addressed in a hash(cc-file-content) cache file.
+// Should run on any system with a shell that provides 2>/dev/null redirect.
+//
+// Invocation without parameters simply uses the .clang-tidy config to run on
+// all *.{cc,h} files. Additional parameters passed to this script are passed
+// to clang-tidy as-is. Typical use could be for instance
+//   run-clang-tidy-cached.cc --checks="-*,modernize-use-override" --fix
+//
+// Note: useful environment variables to configure are
+//  CLANG_TIDY         = binary to run; default would just be clang-tidy.
+//  CLANG_TIDY_CONFIG  = override configuration file in kConfig.clang_tidy_file
+//  CACHE_DIR          = where to put the cached content; default ~/.cache
+//  CLANG_TIDY_JOBS    = Number of tasks to run in parallel.
+
+// This file shall be c++17 self-contained; not using any re2 or absl niceties.
+#include <unistd.h>
+
+#include <algorithm>
+#include <cerrno>
+#include <cinttypes>
+#include <csignal>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <filesystem>
+#include <fstream>
+#include <functional>
+#include <initializer_list>
+#include <iostream>
+#include <list>
+#include <map>
+#include <mutex>
+#include <regex>
+#include <string>
+#include <string_view>
+#include <system_error>
+#include <thread>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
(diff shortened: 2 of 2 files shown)


hzeller:
Updates: Given the duplicates when clang-tidy follows header files, now also made it de-duplicate these in the summary, making the numbers there count unique findings.

hzeller:
the CI failure looks unrelated.

docs/contrib/DeveloperGuide.md
```diff
@@ -318,6 +318,16 @@ clang-tidy -p ./build source_file.cpp
 clang-format -i -style=file:.clang-format source_file.cpp
 ```
 
+To run `clang-tidy` on all files, you can use the following script that runs
+`clang-tidy` in parallel and also caches the results, so subsequent runs
+only have to operate on changed files.
```
maliberty, line 323: Could this be accomplished through cmake?  It sounds very similar to what is already does in terms of parallelism and updating.

hzeller: I don't know, but cmake is a somewhat black box to me. 
Maybe it can be done to run `clang-tidy` in parallel, but I suspect the sophisticated persistent caching can not be easily achieved.

etc/run-clang-tidy-cached.cc
```diff
@@ -0,0 +1,627 @@
...
+std::string GetContent(FILE* f)
+{
+  std::string result;
+  if (!f) {
+    return result;  // ¯\_(ツ)_/¯ best effort.
+  }
+  char buf[4096];
+  while (const size_t r = fread(buf, 1, sizeof(buf), f)) {
+    result.append(buf, r);
+  }
```
maliberty, line 176: Why old style C io ? GPT gives
```
    std::ostringstream buffer;
    buffer << file.rdbuf();  // Read the file into the stream buffer
    std::string content = buffer.str();  // Convert the buffer to a string
```

etc/run-clang-tidy-cached.cc
```diff
@@ -0,0 +1,627 @@
...
+  // Fix filename paths found in logfiles that are not emitted relative to
+  // project root in the log - remove that prefix.
+  // (bazel has its own, so if this is bazel, also bazel-specific fix up that).
+  static void RepairFilenameOccurences(const fs::path& infile,
+                                       const fs::path& outfile)
+  {
+    static const std::regex sFixPathsRe = []() {
+      std::string canonicalize_expr = "(^|\\n)(";  // fix names at start of line
+      if (kConfig.is_bazel_project) {
+        auto bzroot = GetCommandOutput("bazel info execution_root 2>/dev/null");
```
maliberty, line 407: This will not work in OR

etc/run-clang-tidy-cached.cc
```diff
@@ -0,0 +1,627 @@
...
+  static fs::path GetCacheBaseDir()
+  {
+    if (const char* from_env = getenv("CACHE_DIR")) {
+      return fs::path{from_env};
+    }
+    if (const char* home = getenv("HOME")) {
+      if (auto cdir = fs::path(home) / ".cache/"; fs::exists(cdir)) {
+        return cdir;
+      }
+    }
```
maliberty, line 356: If you cache in the home dir you may run into conflicts between runs in different work areas.  Is there a policy to remove old cache entries or does it grow indefinately?

etc/run-clang-tidy-cached.cc
hzeller: std::stream stuff is significantly slower than the `fread()` and stuff. that is why I prefer that.

etc/run-clang-tidy-cached.cc
hzeller: `is_bazel_project` is disabled (this is a generic script that can be used for all kinds of situations)

etc/run-clang-tidy-cached.cc
hzeller: the cache is separated by project and hash of the clang-tidy configuration used, so it will not clash with other projects. It will share the cache between different checkouts of OpenRoad, but that is desired, as it allows to quickly switch between branches and re-create the output almost instantaneously.

Old cache entries are currently not automatically removed (it could be added but for now this is not a large problem in particular if the project is mostly clang-tidy clean (which we're not quite yet :) )).

maliberty: What guards against two different concurrent runs of OR in different areas colliding in the shared cache?

hzeller: The outputs are separate files that are content-addressed, so their name depends on the hash of its content and the headers it depends on (and the clang-tidy configuration). Which means that even if there are two instances in parallel running on exactly the same file they will create the same content anyway. That writes into a temporary file.

After that, if the file is fully created without issues, it is atomically moved to its final place. So even if there were two instances running, there is only one 'winner' (which then serves both).

maliberty:
Is this ready to merge or are you still working on anything?

maliberty (approved):

maliberty:
If there is a change to a header, are all includers re-run?  If yes, does this apply transitively though a change of includes?

hzeller:
Includes: right now, I only go one level down, so all header that a particular header includes will be checked (as sometimes clang-tidy might find something in them that changes the outcome). But not recursively all the way down.

Triggered by your last question, I have now also added a uniqifier to the temporary file so that it really really will be not getting in each others way. So that was the last push, but it is ready to merge now.

hzeller:
But yes to answer the question more specifically: if a header changed, not only the header but also all the files that include that header are re-run.
But it stops there and the includers of a header header that includes a changed header are not re-run (they should not be needed anyway, as everyone should adhere to IWYU but that also might be a bit of work until we're there).

maliberty:
We don't run or adhere to IWYU today.  I'm not saying we shouldn't but it isn't a good assumption.

hzeller:
I noticed :) I might use some free cycles here and there to help sending some clean-up changes.

Speaking of which, what would be a preferred way of doing this ? Sending a missing include of one particular header (say `#include <vector>`) for all files that are missing them, or per-directory in `src/*` ? (and: are they appreciated at all or just annoying noise ?)

docs/contrib/DeveloperGuide.md
```diff
@@ -318,6 +318,16 @@ clang-tidy -p ./build source_file.cpp
...
+
+```shell
+cmake . -B build  # generate build files
+ln -sf build/compile_commands.json .  # make compilation db visible
```
maliberty, line 327: Btw, it would be nice to eliminate the need for the link and just take it as an arg or envar

maliberty:
Per directory is easier as it less likely to generate conflicts than a touch everything PR.

For IWYU, it would be good start with a way to run it that is standardized.  If it is isn't in the CI your efforts will just rot.

hzeller:
There is a check `misc-include-cleaner` in `clang-tidy` that will help find these cases (stuff that is included that is not needed, and symbols that are defined but not with the necessary header).

So if we add `misc-include-cleaner` to the `.clang-tidy`, these issues should issue in code-reviews by the automatic clang-tidy run.

maliberty:
Great - would you make a PR to add `misc-include-cleaner`?

maliberty:
I tried it out and it works nicely initially.  However after I fixed a bunch of issues and re-run it I find many false reports of issues I already fixed.  I suspect this is the lack of transitive header rebuilds as too few files seemed to need re-running.  It definitely limits its usefulness in its current form.

hzeller:
Interesting, I'll play with that.
Maybe it is indeed needed to go further down the dependency chain until we have IWYU and all direct needed includes are actually included and not rely on transitive includes to work. So getting includes fixed as much as possible should probably be high on the list of relevant things to look at.

PR #6418, opened 2024-12-23
Cleanup: Fix missing header in parsers.h and some clang-tidy issues.

hzeller:
The std::ifstream is used, but necessary header was missing. Also `<cstring>` was missing for the strcmp().
While at it, remove superfluous headers.

Also: simplify a boolean return (there was a else after return issue, and the whole expression could be written more concise and readable anyway).

If there is a single statement in a block in a new line, add the necessary curly braces.

### DIFF
--- a/src/mpl/src/ParquetFP/src/parsers.h
+++ b/src/mpl/src/ParquetFP/src/parsers.h
@@ -33,15 +33,14 @@
 #ifndef PARSERS_H
 #define PARSERS_H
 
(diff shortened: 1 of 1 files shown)


maliberty:
I wouldn't spend any time on mpl as mpl2 will replace it.  We are just working through some issues before https://github.com/The-OpenROAD-Project/OpenROAD/pull/6047

maliberty (approved):

PR #6421, opened 2024-12-23
Enable misc-include-cleaner in .clang-tidy

hzeller:
This will help finding missing and superfluous includes. Right now, over the project, there are a lot of findings, so let's see if this will be good or annoying until most of it is fixed.


maliberty (approved):

hzeller:
Follow-up to https://github.com/The-OpenROAD-Project/OpenROAD/pull/6415#issuecomment-2560315750

Turns out, it was explicitly disabled, probably due to the high count currently.

PR #6422, opened 2024-12-23
Add missing `#include <vector>` in src/odb/src/db/*

hzeller:
Files that use std::vector, but haven't included the corresponding header yet.

### DIFF
--- a/src/odb/src/db/dbAccessPoint.cpp
+++ b/src/odb/src/db/dbAccessPoint.cpp
@@ -41,6 +41,7 @@
 #include "odb/dbTypes.h"
 // User Code Begin Includes
 #include <algorithm>
+#include <vector>
 
 #include "dbBPin.h"
 #include "dbBTerm.h"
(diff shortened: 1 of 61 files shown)


hzeller:
mmh, looks like the clang-format version 17.0.6 I am using produces slightly different result than the CI one. Let me manually fix that.

hzeller:
alright, formatting hurdle fixed by getting an old 14.0.6.
Looks like some of these files are actually generated, so I'll have to fix the generator templates instead...

maliberty:
No need to touch the templates as we clang-format the output of code generation.  Probably it is still picking up the wrong version.

hzeller:
what I mean with fix the templates is, that the `#include <vector>` needs to go into these or the generator script. Currently exploring how that is done.

maliberty:
Yes in that case you'll need to adjust the templates to add the include.

hzeller:
closing this for now, will have a closer look if I find time during the holidays.

PR #6423, opened 2024-12-24
Fix missing includes in tapcell.{h,cpp}

hzeller:
Fixed 20 misc-include-cleaner issues. Only 12199 to go :)

### DIFF
--- a/src/tap/include/tap/tapcell.h
+++ b/src/tap/include/tap/tapcell.h
@@ -34,8 +34,16 @@
 ///////////////////////////////////////////////////////////////////////////////
 
 #include <boost/polygon/polygon.hpp>
+#include <map>
+#include <optional>
+#include <set>
+#include <string>
+#include <utility>
+#include <vector>
 
 #include "odb/db.h"
+#include "odb/dbTypes.h"
+#include "odb/geom.h"
 #include "odb/geom_boost.h"
 
 namespace ord {
--- a/src/tap/src/tapcell.cpp
+++ b/src/tap/src/tapcell.cpp
@@ -36,17 +36,20 @@
 #include "tap/tapcell.h"
 
 #include <algorithm>
+#include <array>
+#include <cmath>
+#include <cstddef>
 #include <map>
 #include <optional>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "odb/db.h"
+#include "odb/dbTypes.h"
 #include "odb/util.h"
 #include "ord/OpenRoad.hh"
-#include "sta/StaMain.hh"
 #include "utl/Logger.h"
-#include "utl/algorithms.h"
 
 namespace tap {
 


maliberty (approved):

maliberty:
IWYU says "for every symbol (type, function, variable, or macro) that you use in foo.cc (or foo.cpp), **_either_** foo.cc or foo.h should include a .h file that exports the declaration of that symbol.".  So it wouldn't be necessary to add vector to both the h and cpp file but that what happens in this PR.  Is the clang-tidy check stricter than IWYU ?

hzeller:
Correct, `misc-include-cleaner` makes sure that each file can be considered stand-alone it itself, importing all the things it needs. So if the header has something that needs `std::vector` it needs to include `<vector>`, but then also the corresponding compilation unit also uses `std::vector`, then it _also_ needs to include `<vector>`. Μaybe the older IWYU rules were more lenient, but this stand-alone-ness makes it very easy not only for tools, but also humans that look at a file.

There are ways to allow for one level of transitive export. Sometimes there are headers that include others headers whose symbols we want to make look like as if they had been directly defined in that header, in that case we can use the `IWYU pragma: export` to tell it that this is the intent, and misc-include-cleaner will not warn about that header directly included anymore.
Should of course be used rarely and not for things such as `<vector>`, but sometimes there are 'collection' headers that combine some key headers where this might be useful.

```
#include "include/header/that/whose/symbols/we/want/to/export.h"  // IWYU pragma: export
```

The direct documentation of the check is a bit limited 
https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html

... but the linked document in there 
https://clangd.llvm.org/design/include-cleaner

maliberty:
The quote is directly from https://github.com/include-what-you-use/include-what-you-use so I imagine it to be the current rule.  Given the rule that the .h should always be included by the .cpp it is a bit redundant but not worth a lot of effort to resolve.

hzeller:
IIRC, IWYU is essentially a 15y old project that emerged from avoiding forward declares in the Google codebase back then to actually including most of the things, so they might've settled for a simpler goal.

If we could configure `misc-include-cleaner` to not worry about it, that would also be good ('complain about things that are not included unless they are included in a file whose basename looks like mine' :) ).
But even if not, the additional readability of the code is also a plus.

Using the `clangd` language server in emacs/vi/vscode is very helpful to fix these things while developing.

I think it is doable to get in a better state; I have gotten other projects from tens of thousand `misc-include-cleaner` warnings to zero fairly quickly with an overall improvement in readability.  I'll keep chipping away on it for OR.

Different topic: One questiion w.r.t. to rules to consider: There is a rule that seems to be somewhat noisy and probably not too interesting (at least at this point): `performance-enum-size` ... it suggests if an enum does not have too many values to change it ti `uint8_t` for instance instead of the default `uint32_t`. I think for now, we should probably disable it as it is somewhat noisy, wdyt ?

maliberty:
> Using the clangd language server in emacs/vi/vscode is very helpful to fix these things while developing.

How does it help?

hzeller:
> > Using the clangd language server in emacs/vi/vscode is very helpful to fix these things while developing.
> 
> How does it help?

If you have the `clangd` language server enabled and linked the `compile_commands.json` in the toplevel directory for the language server to find, it will show wiggly lines on symbols that are not directly mentioned in included files and allows wit one click to add the the corresponding header at the top of the file, already properly sorted according to `.clang-format`.
Also it will highlight headers that are not actually needed in a file that can be removed.

PR #6426, opened 2024-12-24
Switch off performance-enum-size clang-tidy check for now.

hzeller:
It warns on every enum if it is smaller than fitting into a uint32_t to use a smaller type. While it might be an interesting exercise at some point to look at these, right now, this is a somewhat noisy check.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -31,6 +31,7 @@ Checks: >
   -google-runtime-references,
   -google-explicit-constructor,
   performance-*,
+  -performance-enum-size,
   bugprone-*,
   -bugprone-branch-clone,
   -bugprone-easily-swappable-parameters,


hzeller:
About 1/6 of the lines from clang-tidy over all code in `OpenRoad_clang-tidy.out` are due to this warning (with a lot of duplicates as it creates a new warning in each header it finds an `enum` with that condition)

maliberty:
Are these in face many unique problems or just many instances of a few problems that fall in headers?  Given the potential memory savings I'd like to analyze it more before turning it off.

maliberty:
Missing DCO

hzeller:
> Are these in face many unique problems or just many instances of a few problems that fall in headers? Given the potential memory savings I'd like to analyze it more before turning it off.

There are many duplicates, I think unique mentions are about 200ish. In particular if enums are stored in tight data structures, that can be super-useful indeed; if this is data at rest somewhere (if it is serialized in some form), it is probably good to double-check that this does not mess up anything in the storage format.

hzeller:
You can find all the instances by looking through the `clang-tidy` output:

```
grep performance-enum-size OpenRoad_clang-tidy.out | sort | uniq
```

maliberty:
What version of clang-tidy are you using?  I get 0 results from that grep.

maliberty:
I'm using `Ubuntu LLVM version 14.0.0`

maliberty:
I think that check was introduced in 18.  It seems we should start with an agreement on version.

The CI action uses ubuntu-latest which is 22.04 (https://github.com/actions/runner-images) which uses 14.0 by default.  ubuntu-24.04 is on 18

maliberty:
Looks like it will switch to 24 soon https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/

hzeller:
I am using clang-tidy 18.1.8 (nixos unstable)

maliberty:
With 18 I still only see `  240 [performance-enum-size]` in the summary

hzeller:
yep, 240 sounds about right (in the *.out file you see a lot of duplicates because each file that includes a header with an enum, it reports it anew. Maybe I should pre-filter in each of the out files first).

maliberty:
After playing with it for a quite a while I agree to turn this off.  The main problem is working with enums with specified types and bitfields in the db becomes quite a headache.

maliberty (approved):

Issue #6420, opened 2024-12-23
Make  `estimate_parasitics -spef_file foo.spef` write to foo.spef, not foo_default.spef


oharboe:
### Description

It is surprising that the _default suffix is added to the filename.

To work around this, I have to add a mv command:

```tcl
estimate_parasitics -placement -spef_file $::env(RESULTS_DIR)/4_cts.spef
exec mv $::env(RESULTS_DIR)/4_cts_default.spef $::env(RESULTS_DIR)/4_cts.spef
```

### Suggested Solution

Write to the exact file that was specified on the command instead of writing to some other file.

maliberty:
If you have multiple corners then multiple files need to be written with unique names.  With just one corner we can do as you suggest.

PR #6434, opened 2024-12-26
rsz: add corner prefix when design has more than one corner

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/6420


src/rsz/test/make_parasitics7_ss.spefok
```diff
@@ -0,0 +1,122 @@
+*SPEF "ieee 1481-1999"
```
gadfort, line 1: How come there are two spefok files but you are only checking one of them.
Second why would you write the ss file if there is only one corner defined?

eder-matheus: It was not intended to be in this PR, I added this file by accident. Thanks for noting that!

gadfort: Ahh, okay.

maliberty (approved):

PR #6446, opened 2024-12-30
utl:: check if container is empty on utl::shuffle

eder-matheus:
Fix Coverage errors

### DIFF
--- a/src/grt/src/fastroute/src/utility.cpp
+++ b/src/grt/src/fastroute/src/utility.cpp
@@ -2495,8 +2495,13 @@ void FastRouteCore::saveCongestion(const int iter)
     const int seed = 42;
     g.seed(seed);
 
-    utl::shuffle(congestionGridsV.begin(), congestionGridsV.end(), g);
-    utl::shuffle(congestionGridsH.begin(), congestionGridsH.end(), g);
+    if (!congestionGridsH.empty()) {
+      utl::shuffle(congestionGridsH.begin(), congestionGridsH.end(), g);
+    }
+
+    if (!congestionGridsV.empty()) {
+      utl::shuffle(congestionGridsV.begin(), congestionGridsV.end(), g);
+    }
   }
 
   const std::string marker_group_name = fmt::format(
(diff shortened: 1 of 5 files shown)


src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2495,8 +2495,13 @@ void FastRouteCore::saveCongestion(const int iter)
     const int seed = 42;
     g.seed(seed);
 
-    utl::shuffle(congestionGridsV.begin(), congestionGridsV.end(), g);
-    utl::shuffle(congestionGridsH.begin(), congestionGridsH.end(), g);
+    if (!congestionGridsH.empty()) {
+      utl::shuffle(congestionGridsH.begin(), congestionGridsH.end(), g);
+    }
```
maliberty, line 2500: It would be better to make shuffle handle the empty vector correctly so that all users don't need this  check.

maliberty: `if (n == 0) return;` should do it

eder-matheus: Thanks! I'm testing it right now and will add it to the PR soon.

src/grt/src/fastroute/src/utility.cpp
```diff
@@ -2495,8 +2495,8 @@ void FastRouteCore::saveCongestion(const int iter)
     const int seed = 42;
     g.seed(seed);
 
-    utl::shuffle(congestionGridsV.begin(), congestionGridsV.end(), g);
     utl::shuffle(congestionGridsH.begin(), congestionGridsH.end(), g);
+    utl::shuffle(congestionGridsV.begin(), congestionGridsV.end(), g);
```
maliberty, line 2499: Is this just a leftover or is there some significance to the order of these lines?

maliberty (approved):

src/grt/src/fastroute/src/utility.cpp
eder-matheus: Just a leftover.

PR #6469, opened 2025-01-04
dbSta: Add create_cell_usage_snapshot command

bangqixu:
Add create_cell_usage_snapshot command to create a cell usage snapshot in JSON format.

create_cell_usage_snapshot -path <path> -stage <stage_name> will create a JSON file named as cell_usage_snapshot-<module_name>-<stage_name>.json under <path>. The snapshot reflects (i) the name of the stage and (ii) cell usage statistics. Currently cell usage statistics consist of (a) cell name, (b) number of instance of the cell and (c) the cell area in um^2.

As compared to current implementation of report_cell_usage -v, this is more structured and more extensible and testable. We have planned work to support diff of such snapshots such that we can better evaluate the performance of each flow step / stage by capturing snapshots as often as needed.

### DIFF
--- a/src/dbSta/include/db_sta/dbSta.hh
+++ b/src/dbSta/include/db_sta/dbSta.hh
@@ -35,7 +35,11 @@
 
 #pragma once
 
+#include <cstdint>
 #include <memory>
+#include <string>
+#include <string_view>
+#include <vector>
 
 #include "odb/db.h"
 #include "odb/dbBlockCallBackObj.h"
@@ -190,6 +194,30 @@ class dbSta : public Sta, public ord::OpenRoadObserver
   std::string getInstanceTypeText(InstType type);
   InstType getInstanceType(odb::dbInst* inst);
   void report_cell_usage(odb::dbModule* module, bool verbose);
+
+  // Creates a cell usage snapshot for the given module at the given stage.
+  // The snapshot is written to the given path as a JSON file.
+  void CreateCellUsageSnapshot(
+      odb::dbModule* module, std::string_view path, std::string_view stage);
+
+  // Holds the usage information of a specific cell.
+  struct CellUsageInfo {
+    // Name of the cell.
+    std::string name;
+
+    // Number of instances of the cell.
+    int count = 0;
+
+    // Area of the cell in microns^2. Total area of the cell instances can be
+    // calculated by multiplying the count by this value.
+    double area = 0.0;
+  };
+
+  // Holds a snapshot of cell usage information at a given stage.
+  struct CellUsageSnapshot {
+    std::string stage;
+    std::vector<CellUsageInfo> cells_usage_info;
+  };
 
   BufferUse getBufferUse(sta::LibertyCell* buffer);
 
--- a/src/dbSta/src/dbSta.cc
+++ b/src/dbSta/src/dbSta.cc
@@ -44,11 +44,21 @@
 #include <tcl.h>
 
 #include <algorithm>  // min
+#include <cmath>
+#include <cstdint>
+#include <fstream>
+#include <map>
 #include <mutex>
 #include <regex>
+#include <string>
+#include <string_view>
+#include <vector>
 
 #include "AbstractPathRenderer.h"
 #include "AbstractPowerDensityDataSource.h"
+#include "boost/algorithm/string/join.hpp"
+#include "boost/json.hpp"
+#include "boost/json/src.hpp"
 #include "dbSdcNetwork.hh"
 #include "db_sta/MakeDbSta.hh"
 #include "db_sta/dbNetwork.hh"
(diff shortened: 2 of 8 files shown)


maliberty:
We already have report_cell_usage so it would be better to enhance that than create a similar command.

bangqixu:
Thanks for the feedback! I thought about that option and then realized that report_cell_usage at the moment is already a bit convoluted (the command itself reports a few different things and "-verbose" is not very expressive / self-explanatory wrt what is reported), so I think it is more readable and testable to create a separate command for this.

Besides, this is the first step of a series of create_.*_snapshot(s). We have plan to introduce timing histogram snapshot (and potentially a power snapshot as well) to capture the snapshot of PPA at a given moment. The general idea is to dump snapshots as JSON and separate utilities can then diff between snapshots to get the delta which will enable evaluation of flow steps.

IMHO, it would be cleaner to have create_.*_snapshots that creates snapshots as JSON files while the other commands can still report it as part of the tool (log) output.

If you have more constructive suggestions on how to achieve this without degrading readability and testability, I would be happy to help!

Let me know what you think.

maliberty:
All reports are inherently snapshots so having both create_.*_snapshots and report_.* feels redundant and confusing to me, especially if they capture different information.  

Having the option to generate a json report makes sense to me.  I don't see it requiring a different command but just being an option to existing reporting.

If you are not fond of -verbose please explain what options you intend to create to replace it.  We can always update the options for reporting to be more fine grained.

We also have the metrics format in json for dumping QOR information.  Can what you need to represented as a metric?

bangqixu:
Thanks for the prompt response on the weekend! I don't disagree that all
reports are inherently snapshots, yet if the output goes to log, then it
requires extra efforts / hurdles to parse the log and then convert it to be
machine readable.

So far the options really are (i) the directory and (ii) name of the stage
when the snapshot is taken. The command is designed to do one thing and one
thing only, so it is not flexible in terms of output format (i.e., JSON
with the structure mapped to the C++ struct in the code) or destination
(i.e., it only outputs to JSON file and won't print in log).

The command currently can be called as "create_cell_usage_snapshot -path
<path> -stage <stage_name>". Are you fine with something like
"report_cell_usage -snapshot -path <path> -stage <stage_name>"? I was a bit
concerned whether this would complicate option validation (e.g., if
-snapshot is present, then -path and -stage must be present).

Is there a doc for QoR (or C++ implementation)? In general we would prefer
something more structured than key-value pairs. For example, we can have
some hierarchy in the snapshot like
"{"stage":"test_stage","cell_usage_info":[{"name":"snl_bufx1","count":4,"area":1E3},{"name":"snl_ffqx1","count":2,"area":1E3}]}".
If you have a pointer to some doc of metrics, I would be more than happy to
take a look. Thank you!

maliberty:
The metrics format is from https://vlsicad.ucsd.edu/Publications/Conferences/388/c388.pdf.  You can run any design in ORFS to see the metrics reported by stage.  It encodes information in the key name rather than introduce extra hierarchy in the json.  I'd like to avoid having to support multiple formats unless there is a compelling reason.

What does `-snapshot` do?  Is it just `-json` or `-format json`?

We already have `utl::set_metrics_stage` so `-stage` is duplicative.

Would `-file` make more sense than `path`?  Are you writing more than one file?

bangqixu:
Thanks for getting back to me!

I would argue that encoding information in the key name is not a great way to do this and explicitly having hierarchy in the data structure is a cleaner way. Plus if information is all encoded in the key, backward compatibility would be a pain as all entries need to be updated whereas if you have hierarchy and the extra attribute only exists on a low level, you can leave most of the information untouched. If I have to pick one, I would drop the support of flat / pure key-value pair one. It is similar spirit to that for liberty / LEF cell name, much information is encoded in the name (e.g., VT, cell height, etc), yet it is still more usable if such information is explicitly written down for easier query.

```-snapshot``` is to indicate that the command is going to create a snapshot that matches the snapshot definition (again, with the same hierarchy between C++ code and JSON format for better readabilty / portability). If we do ```-json``` or ```-format json```, that would not be less expressive what the command is going to output.

I would argue that ```-stage``` is more expressive as we don't need to introduce dependency on internal state set by ```utl::set_metrics_stage``` (plus this seems to be tied to the metrics concept).

I was thinking about creating snapshots for each module if there are hierarchy, so I put ```-path``` instead of ```-file``` there so that the format of the file name is coded and not arbitrary.

Thank you!

bangqixu:
I will write a one-pager later this week.

bangqixu:
Given the amount of time I can spend on this, I am going to change it to "report_cell_usage -format json -file".

bangqixu:
What would be the proper way to handle the module_inst argument of report_cell_usage? Currently it seems that the command assumes anything other than -verbose would be the name of the targeted module. Does it make sense to introduce -module key for the command (along with -file and -format). I feel like it would be cleaner to first introduce -module key (and update tests) before introducing -file and -format. Would breaking other's script a concern since this is effectively an API change?

maliberty:
You could add a -module option.  You can also keep backward compatibility by treating any non-flag argument as the module name.

bangqixu:
Thanks for the suggestion! I will close this one and start over with adding -module option first.

PR #6475, opened 2025-01-07
odb: protecting LefParser and DefParser namespaces

LucasYuki:
Protects the global and static variables in the LefParser and DefParser namespaces (issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/5981)

Changes:
- lefin class splitted into lefin and lefinReader:
  - lefin: interface used by OR;
  - lefinReader: interface used by the Lef module.
- lefin and defin methods protected with a lock;


src/odb/src/defin/defin.cpp
```diff
@@ -37,70 +37,85 @@
...
+
 defin::defin(dbDatabase* db, utl::Logger* logger, MODE mode)
 {
+  std::lock_guard<std::mutex> lock(_def_mutex);
   _reader = new definReader(db, logger, mode);
 }
 
 defin::~defin()
 {
+  std::lock_guard<std::mutex> lock(_def_mutex);
```
QuantamHD, line 50: I'm not clear the locking is doing anything here. It kind of looks like this is trying to protect two defin's from running at the same time on the same database, but what we're really concerned about is two different defin's running on two different databases.

What are these locks here to do?

QuantamHD (changes requested):

src/odb/src/defin/defin.cpp
maliberty: I think he has just locked every method.  The key ones are ::create*/replace* methods which will block simultaneous parsing.

@LucasYuki I think some of these do nothing more than set a manipulate local variables and not the db and therefore do not need locks (eg ctor, dtor, skip).  Please analyze it a bit more.

maliberty (approved):

QuantamHD:
I am, but could we add a comment to the mutexes that explains why they exist. Specifically mentioning that the defin and lefin parsers rely on static vars.

maliberty:
@LucasYuki please add the comment to this PR.

maliberty (approved):

PR #6477, opened 2025-01-07
rmp: Adds DelayOptimization pass to remapper

QuantamHD:
Adds new pass based infrastructure to add new Abc optimization passes. The first delay based optimization is added here.

General intention is to build lots of interfaces that can be extended by both the public and private clients.

### DIFF
--- a/src/rmp/src/delay_optimization_strategy.cpp
+++ b/src/rmp/src/delay_optimization_strategy.cpp
@@ -1,0 +1,134 @@
+// Copyright 2024 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+#include "delay_optimization_strategy.h"
+
+#include "abc_library_factory.h"
+#include "base/abc/abc.h"
+#include "map/mio/mio.h"
+#include "map/scl/sclLib.h"
+#include "map/scl/sclSize.h"
+#include "utl/deleter.h"
+
+namespace abc {
+extern Abc_Ntk_t* Abc_NtkMap(Abc_Ntk_t* pNtk,
+                             Mio_Library_t* userLib,
+                             double DelayTarget,
+                             double AreaMulti,
+                             double DelayMulti,
+                             float LogFan,
+                             float Slew,
+                             float Gain,
+                             int nGatesMin,
+                             int fRecovery,
+                             int fSwitching,
+                             int fSkipFanout,
+                             int fUseProfile,
+                             int fUseBuffs,
+                             int fVerbose);
+extern void Abc_FrameSetLibGen(void* pLib);
+}  // namespace abc
+
+namespace rmp {
+
+utl::UniquePtrWithDeleter<abc::Abc_Ntk_t> WrapUnique(abc::Abc_Ntk_t* ntk)
+{
+  return utl::UniquePtrWithDeleter<abc::Abc_Ntk_t>(ntk, &abc::Abc_NtkDelete);
+}
+
+void AbcPrintStats(const abc::Abc_Ntk_t* ntk)
+{
+  abc::Abc_NtkPrintStats(const_cast<abc::Abc_Ntk_t*>(ntk),
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false);
+}
+
+utl::UniquePtrWithDeleter<abc::Abc_Ntk_t> BufferNetwork(
+    abc::Abc_Ntk_t* ntk,
+    AbcLibrary& abc_sc_library)
+{
+  abc::SC_BusPars buffer_parameters;
+  memset(&buffer_parameters, 0, sizeof(abc::SC_BusPars));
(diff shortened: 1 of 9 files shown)


src/rmp/test/aes_nangate45.tcl
```diff
@@ -0,0 +1,8 @@
+# For debugging optimizations
```
maliberty, line 1: Does this need to be committed?  Likewise the sdc

src/rmp/test/cpp/TestAbc.cc
```diff
@@ -457,7 +462,7 @@ TEST_F(AbcTest, InsertingMappedLogicCutDoesNotThrow)
   sta::dbNetwork* network = sta_->getDbNetwork();
   sta::Vertex* flop_input_vertex = nullptr;
   for (sta::Vertex* vertex : *sta_->endpoints()) {
-    if (std::string(vertex->name(network)) == "_32989_/D") {
+    if (std::string(vertex->name(network)) == "_33122_/D") {
```
maliberty, line 465: Why does this change?  It seems brittle.

src/rmp/src/delay_optimization_strategy.h
```diff
@@ -0,0 +1,32 @@
...
+#include "utl/deleter.h"
+
+namespace rmp {
+
+class DelayOptimizationStrategy : public LogicOptimizationStrategy
+{
+ public:
+  ~DelayOptimizationStrategy() override = default;
+
+  DelayOptimizationStrategy(sta::dbSta* sta) : sta_(sta){};
```
maliberty, line 22: stray ;

src/rmp/src/delay_optimization_strategy.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include "abc_library_factory.h"
+#include "base/abc/abc.h"
+#include "map/mio/mio.h"
+#include "map/scl/sclLib.h"
+#include "map/scl/sclSize.h"
+#include "utl/deleter.h"
+
+namespace abc {
+extern Abc_Ntk_t* Abc_NtkMap(Abc_Ntk_t* pNtk,
```
maliberty, line 17: Is this not defined in an abc header than you can include?

src/rmp/src/delay_optimization_strategy.cpp
```diff
@@ -0,0 +1,134 @@
...
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false,
+                         false);
```
maliberty, line 54: It would be nice to include the parameter names like you did Abc_NtkMap below

src/rmp/src/delay_optimization_strategy.cpp
```diff
@@ -0,0 +1,134 @@
...
+  library_factory.AddDbSta(sta_);
+  AbcLibrary abc_sc_library = library_factory.Build();
+  abc::Abc_SclHashCells(abc_sc_library.abc_library());
+  abc::Abc_SclLinkCells(abc_sc_library.abc_library());
+
+  AbcPrintStats(ntk);
+
+  utl::UniquePtrWithDeleter<abc::Abc_Ntk_t> current_network(
+      abc::Abc_NtkToLogic(const_cast<abc::Abc_Ntk_t*>(ntk)),
+      &abc::Abc_NtkDelete);
```
maliberty, line 100: Does Abc_NtkToLogic return a new Abc_Ntk_t?  If so it shouldn't require a non-const input.

src/rmp/test/cpp/TestAbc.cc
QuantamHD: It's just a different path. I changed it because it's the critical endpoint.

src/rmp/test/aes_nangate45.tcl
QuantamHD: I would find it helpful if they were committed, but I'll remove them if you feel strongly

src/rmp/src/delay_optimization_strategy.cpp
QuantamHD: No it's not, I can send a PR to ABC to fix that, but would rather not block this PR.

src/rmp/src/delay_optimization_strategy.cpp
QuantamHD: Abc_NtkToLogic and most ABC passes are const-able, but ABC never declares them as const. I set const on the Optimize input to at least let OpenROAD devs that the input network isn't modified. I can try to send a PR to make it const as well, but once again I don't think this should block the PR.

src/rmp/src/delay_optimization_strategy.cpp
maliberty: I won't block it but it would be nice to fix this and the const issue in abc.

maliberty (approved):

maliberty:
Please fix the DCO

maliberty:
Is this a progress step in part of larger plan?  I shared Peter's confusion.

maliberty:
@andyfox-rushc in case you have any comments from your experience in this area.

PR #6478, opened 2025-01-07
[drt] Only spawn new threads when distributed_ is true

jmolloy:
Before this patch the constructor of boost::asio::thread_pool{1} created a thread, regardless of if distributed_ was true.

Now it is wrapped in an optional so that it is only created if or when distributed_ becomes true.

This means that OpenROAD can be initialized without spawning any new threads (a niche usecase for sure, but useful in certain environments).


QuantamHD:
LGTM. This change looks good and is required for some Google workflows.

maliberty (approved):

Issue #6152, opened 2024-11-13
[IHP sg13g2] IO pin access problem on the chip right side only


lanserge:
### Describe the bug

Global routing failed reporting congestion at the right side IO cells only. However there is no real congestion.
With some experimentation I could say that removing OBS from IO cells in LEF results in successful routing, but those OBS are actually on opposite side of IO cell and are not obstacles for routing. And actually other sides of the chip have same IO and routed without issue. Looks like some issue related to the cells orientation in the OpenROAD with determining pin access.

### Expected Behavior

All sides of the chip routed.

### Environment

```shell
kernel: Linux 6.10.11+bpo-amd64
os: Debian GNU/Linux 12 (bookworm)
cmake version 3.30.5
-- The CXX compiler identification is GNU 12.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-16972-g32d202458
-- System name: Linux
-- Compiler: GNU 12.2.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull - Success
[...]

PR #6481, opened 2025-01-07
grt: ensure on grid position near instance edge for macro and pad pins

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2656
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/6152

### DIFF
--- a/src/grt/src/Pin.cpp
+++ b/src/grt/src/Pin.cpp
@@ -189,6 +189,18 @@ bool Pin::isDriver()
   return type == odb::dbIoType::OUTPUT || type == odb::dbIoType::INOUT;
 }
 
+odb::Point Pin::getPositionNearInstEdge(const odb::Rect& pin_box) const
+{
+  odb::Point pin_pos;
+  if (getEdge() == PinEdge::north || getEdge() == PinEdge::east) {
+    pin_pos = odb::Point(pin_box.xMax(), pin_box.yMax());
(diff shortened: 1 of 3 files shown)


src/grt/src/Pin.cpp
```diff
@@ -189,6 +189,18 @@ bool Pin::isDriver()
...
 }
 
+odb::Point Pin::getPositionNearInstEdge(const odb::Rect& pin_box) const
+{
+  odb::Point pin_pos;
+  if (getEdge() == PinEdge::north || getEdge() == PinEdge::east) {
+    pin_pos = odb::Point(pin_box.xMax(), pin_box.yMax());
+  } else if (getEdge() == PinEdge::south || getEdge() == PinEdge::west) {
+    pin_pos = odb::Point(pin_box.xMin(), pin_box.yMin());
+  }
```
maliberty, line 199: Why does it snap in both directions?  For example on north wouldn't you just snap Y but not X?

eder-matheus: I made it this way to simplify the code. Since we're working only with one pin geometry at a time, I decided to get the extremes of the geometry for north/east and south/west.

I didn't think in a better way to initialize this position before, but I now noticed that we can use the pin center as the first value, and then snap only in the necessary direction. I will update it.

maliberty (approved):

PR #6483, opened 2025-01-07
gpl: ensure call to BiCGSTAB solver honors ord thread limit

gadfort:
Changes:
- Eigen will use all cores when possible, which can go beyond the number of threads requested.
- This change ensures that the call to BiCGSTAB during initial placement is limited to the thread count.

### DIFF
--- a/src/gpl/src/initialPlace.cpp
+++ b/src/gpl/src/initialPlace.cpp
@@ -79,7 +79,6 @@ void InitialPlace::doBicgstabPlace(int threads)
   // set ExtId for idx reference // easy recovery
   setPlaceInstExtId();
 
-  // omp_set_num_threads(threads);
   for (size_t iter = 1; iter <= ipVars_.maxIter; iter++) {
     updatePinInfo();
     createSparseMatrix();


src/gpl/src/initialPlace.cpp
```diff
@@ -79,6 +79,7 @@ void InitialPlace::doBicgstabPlace()
   // set ExtId for idx reference // easy recovery
   setPlaceInstExtId();
 
+  // omp_set_num_threads(threads);
```
maliberty, line 82: rm

src/gpl/test/gpl_aux.py
```diff
@@ -160,7 +160,7 @@ def global_placement(
         if incremental:
             gpl.doIncrementalPlace(1)
         else:
-            gpl.doInitialPlace()
+            gpl.doInitialPlace(1)
```
maliberty, line 163: Why 1 and not the default like tcl?

maliberty (changes requested):

src/gpl/test/gpl_aux.py
gadfort: because it's `1` in `doIncrementalPlace` and `doNesterovPlace` (above and below). I think it can be a separate PR to change this, but it only effects the testing which already defaults to 1 anyways.

maliberty (approved):

src/gpl/test/gpl_aux.py
gadfort: @maliberty I wasn't planning on making that PR

PR #6486, opened 2025-01-07
drt: Bidirectional A*

kbieganski:
Implements bidirectional A* for DRT.

Unfortunately, usually this doesn't improve run time.  I think the main problem performance-wise is the cost map which wasn't there in the original A*. But the approach seems promising:

| Design                                     | Branch           | RSS [GB] | Time [s] |   Wirelength |
|:-------------------------------------------|:-----------------|---------:|---------:|-------------:|
| `asap7/aes`                                | Baseline         |  4.3527  |  497.538 |        76299 |
| `asap7/aes`                                | Bidirectional A* |  3.53801 |  528.685 |        76085 |
| `asap7/ibex`                               | Baseline         |  5.46439 |  685.251 |       116574 |
| `asap7/ibex`                               | Bidirectional A* |  4.39161 |  716.15  |       116305 |
| `nangate45/aes`                            | Baseline         |  1.52102 |  427.31  |       311438 |
| `nangate45/aes`                            | Bidirectional A* |  1.71426 |  449.323 |       311349 |
| `nangate45/black_parrot`                   | Baseline         |  9.07848 | 2843.38  |              |
| `nangate45/black_parrot`                   | Bidirectional A* |  9.76452 | 2815.57  |              |
| `nangate45/ibex`                           | Baseline         |  1.54163 |  264.78  |       331945 |
| `nangate45/ibex`                           | Bidirectional A* |  1.57943 |  262.21  |       332115 | 

What's interesting is that it often produces slightly shorter wirelength.

I probably won't be working on this for now, but putting it out there so that perhaps other people could expand on it, or share some ideas.


rovinski:
I believe bidirectional A* was tried several years ago and dropped also because there was no performance benefit. Same for a line search approach that was tried.

bangqixu:
Just some drive-by comments. One of the major challenges of applying bidirectional A* search is that when the two waves meet, they don't necessarily meet in a DRC-free way. One simple example is that, they could meet where they haven't travelled long enough on the layer thus causing minimum area violation (this is just the simplest example I can think of and there can be many other such design rules). With unidirectional A* search, there are heuristics to help avoid such violations. With bidirectional A*, we  probably need some new heuristics (sorry if you already have it in the code because I have not read in details).

That being said, the runtime comparison here may not reflect the performance of uni/bidirectional A* searches themselves because bidirectional A* may be dealing with more DRCs for the reason above. Without a search algorithm benchmarking utility, the closest thing I can think of is to run one iteration of detail route and only allow each net to be routed once. If the DRC numbers do not differ too much, then the comparison probably can better reflect the raw performance between uni/bidirectional.

maliberty:
reopen if you plan to work on it further

PR #6487, opened 2025-01-07
Cts cluster macros with different parameter than registers

arthurjolo:
Currently the macro tree uses the same parameters to cluster the its sinks (cluster size, cluster diameter and minimum number of sinks to enable clustering), however due to the size of the macros CTS almost never finds a clustering solution. In this PR new parameters for clustering the macro tree sinks are used, allowing the macros to be clustered.

This PR uses default values for clustering parameters:

- cluster size = 4
- cluster diameter = 2 * largest macro
- minimum number of macros = 10

Eg. MockArray:

Not clustering macros
mock_array_unclustered

Clustering macros:
mock_array_clustered


maliberty:
@arthurjolo @precisionmoon what's the status of this PR?

PR #6492, opened 2025-01-08
gui: Reduce display controls sections min size so 3 last columns can be resize to their content.

titan73:
The default min size is 60 while 3 last colums are between 26 and 28.

Before:
before

After:
after


maliberty (approved):

maliberty:
It must be somewhat platform dependent as I see on master
image

PR #6498, opened 2025-01-09
grt: use boost polygon90set in layerIsBlocked to speedup init

gadfort:
Changes:
- removes the nested loops and replaces it with a "&" operation on two Boost Polygon90set.

On zerosoc sky130:
Previous implementation: `layerIsBlocked` takes 187.15s
New implementation: `layerIsBlocked` takes 0.28s
Speedup of 99.8%


QuantamHD:
Wow!

maliberty (approved):

PR #6501, opened 2025-01-10
Tcl versions of Python unit tests

jeffng-or:
1. Created Tcl versions of ODB Python unit tests
2. SWIG modifications for dbWireDecoder::getPoint_ext on both Python and Tcl

One note about the test_inst.py and test_iterm.tcl. They both do negative testing, which causes some problems because the DB doesn't have a logger. For the time being, a include_neg_tests parameter is added to the test, which controls whether the expected failure tests should be executed or not. They are disabled by default.


src/odb/test/helper.tcl
```diff
@@ -0,0 +1,207 @@
+# Converted from helper.py
+
+proc createSimpleDB {} {
+    set db [odb::dbDatabase_create]
```
maliberty, line 4: You can use ord::get_db to get the default database with a logger.

jeffng-or: Thanks for the pointer. I've switched to use the default DB for the two cases that need the logger and removed the option to include/exclude the negative tests.

maliberty (approved):

PR #6371, opened 2024-12-17
Hier-RTLMP Fix

ZhiangWang033:
fix the TP invalid partitioning issue and add the real_blockage_flag which converts the placement blockages into macros with fence constraints.   Also, the IO blockage will be treated as placement blockages.

PR #6502, opened 2025-01-10
mpl: convert pin access blockages to placement blockages

AcKoucher:
Based on #6371.

Ideas here:
1. Make mpl2 placement blockages and macro blockages be treated differently. Macro blockages will keep being treated as regions that should not be occupied by clusters with macros - penalty based on overlap area. Placement blockages, on the other hand, will be treated as macro clusters constrained to fences - the idea of this representation is to allow the generation of more intelligent results by the annealer: If we use the former approach, the only way to reduce  overlap penalty is to fill the macro blockage area with either std cell clusters or leave it empty (the latter is very challenging because of the sequence pair representation), so treating the placement blockages as macro clusters works as if we could leave the area of that blockage empty using the macro blockage approach.
2. Model pin access as placement blockages instead of macro blockages.

Collateral:
1. Increase fence penalty so that if effectively places the blockages where they belong.

If we prove this approach to be more beneficial in terms of QoR and these changes are incorporated, I wonder what would be the usage of the macro blockages and if we couldn't use this same strategy to handle pre-placed macros.

### DIFF
--- a/src/mpl2/src/Mpl2Observer.h
+++ b/src/mpl2/src/Mpl2Observer.h
@@ -74,10 +74,6 @@ class Mpl2Observer
   virtual void setMacroBlockages(const std::vector<mpl2::Rect>& macro_blockages)
   {
   }
-  virtual void setPlacementBlockages(
-      const std::vector<mpl2::Rect>& placement_blockages)
-  {
-  }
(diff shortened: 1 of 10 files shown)


AcKoucher:
Too many things happened since this draft was opened. It was easier to move the work to a fresh branch.

Replaced by #7509.

PR #6503, opened 2025-01-10
psm: allow reuse of previous solution

gadfort:
Changes:
- if a previous solution is available, this allows the user to specify they want to use that instead of recomputing.


src/psm/src/pdnsim.cpp
```diff
@@ -116,7 +117,9 @@ void PDNSim::analyzePowerGrid(odb::dbNet* net,
 
   last_corner_ = corner;
   auto* solver = getIRSolver(net, false);
-  solver->solve(corner, source_type, voltage_source_file);
+  if (!use_prev_solution || !solver->hasSolution(corner)) {
```
maliberty, line 120: I think it would be good to have an else with an info msg so that people know reuse actually happened.

maliberty (approved):

PR #6507, opened 2025-01-11
mpl2: increase snapping search range based on the layers pitch

joaomai:
This PR changes the search attempts from a fixed number to using the lowest common multiple from all the layers that can be snapped as basis in order the explore more possibilities.

### DIFF
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -4546,8 +4546,6 @@ SameDirectionLayersData Snapper::computeSameDirectionLayersData(
 
           if (data.layer_to_pin.empty()) {
             data.snap_layer = layer;
-            // } else if (layer->getNumber() < data.snap_layer->getNumber()) {
-            //   data.snap_layer = layer;
           }
 
           data.layer_to_pin[layer] = iterm;
--- a/src/mpl2/src/hier_rtlmp.cpp
+++ b/src/mpl2/src/hier_rtlmp.cpp
@@ -4546,6 +4546,8 @@ SameDirectionLayersData Snapper::computeSameDirectionLayersData(
 
           if (data.layer_to_pin.empty()) {
             data.snap_layer = layer;
+            // } else if (layer->getNumber() < data.snap_layer->getNumber()) {
+            //   data.snap_layer = layer;
           }
 
           data.layer_to_pin[layer] = iterm;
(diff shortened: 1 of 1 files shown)


src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -4653,56 +4653,42 @@ void Snapper::attemptSnapToExtraLayers(
...
 
-      if (curr_number_of_snapped_layers == total_number_of_layers) {
-        all_layers_snapped = true;
-        break;
-      }
+    setOrigin(origin + (pitch * steps), target_direction);
+    int snapped_layers = 0;
+    for (const auto& [layer, pin] : layers_data.layer_to_pin) {
+      snapped_layers += pinsAreAlignedWithTrackGrid(
+          pin, layers_data.layer_to_params.at(layer), target_direction);
```
AcKoucher, line 4677: Avoid implicit casting:

```
if (pinsAreAlignedWithTrackGrid()) {
  ++snapped_layers;
}
```

I also find it to be slightly easier to read.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -4653,56 +4653,42 @@ void Snapper::attemptSnapToExtraLayers(
...
-        continue;
-      }
+  int best_snapped_layers = 1;
 
-      if (pinsAreAlignedWithTrackGrid(
-              pin, layers_data.layer_to_params.at(layer), target_direction)) {
-        ++curr_number_of_snapped_layers;
-      }
+  for (int i = 0; i <= total_attempts; i++) {
+    int steps = (i % 2 == 1) ? (i + 1) / 2 : -(i / 2);
```
AcKoucher, line 4671: Perhaps a small comment saying that we're doing both positive and negative steps one at the time.

src/mpl2/src/hier_rtlmp.cpp
```diff
@@ -4653,56 +4653,42 @@ void Snapper::attemptSnapToExtraLayers(
...
+      // Stop search if all layers are snapped
+      if (total_layers == best_snapped_layers) {
+        break;
+      }
     }
-
-    --remaining_attempts;
   }
 
+  // Set the origin to the best found origin
```
AcKoucher, line 4691: A bit redundant.

AcKoucher (changes requested):
Just some nits.

AcKoucher:
@joaomai Please, include in the description that we're now also doing the search in both positive and negative directions.

eder-matheus (approved):

PR #6514, opened 2025-01-13
gui: optimizing find in clock tree viewer

### LINKED ISSUE #6400: Stop search instance "*FF*" from locking up the GUI
### Description

For a large design, it is easy to accidentally launch a search that is too broad, this will then lock up the GUI.

image

### Suggested Solution

Press "ESC"  to stop search

### Additional Context

_No response_

LucasYuki:
Closes #6400.

Changes:
- Optimizes instance search in CTS Viewer by disabling rendering when setting nodes to selected
  - Previously, the CTS Viewer was rendering everything after each node was set to selected
- Changes items name hash table to unordered_map

### DIFF
--- a/src/gui/src/clockWidget.cpp
+++ b/src/gui/src/clockWidget.cpp
@@ -847,9 +847,13 @@ bool ClockTreeView::changeSelection(const SelectionSet& selections)
   if (!nodes.empty()) {
     // remove old selection
     clearSelection();
+
+    // prevent ClockTreeView to draw until unlock
+    lockRender();
     for (auto node : nodes) {
       node->setSelected(true);
     }
+    unlockRender();
     return true;
   }
   return false;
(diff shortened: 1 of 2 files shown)


maliberty:
@LucasYuki what actually triggers the update after you unlock?

@gadfort I'd like your input as the original author.

src/gui/src/clockWidget.cpp
```diff
@@ -847,9 +847,13 @@ bool ClockTreeView::changeSelection(const SelectionSet& selections)
   if (!nodes.empty()) {
     // remove old selection
     clearSelection();
+
+    // prevent ClockTreeView to draw until unlock
+    lockRender();
     for (auto node : nodes) {
       node->setSelected(true);
     }
+    unlockRender();
```
gadfort, line 856: when this is done, you need to call `selectionChanges()` otherwise all the selection emits from `canConvertAndEmit` will not be processed.

gadfort (changes requested):

maliberty:
> > @LucasYuki what actually triggers the update after you unlock?
> 
> After the change in the selection, the view is fitted to accommodate all selected instances. The view change triggers an update.

That happens in findInCts but if you came to changeSelection by another path it wouldn't.  I don't see such in the code currently but I think @gadfort suggestion is safer.

gadfort (approved):

PR #6521, opened 2025-01-13
gpl: ensure cleanup occurs before exiting run

gadfort:
Moves the cleanup code to before exiting function to avoid leaving behind a bunch of instances.

### DIFF
--- a/src/gpl/src/mbff.cpp
+++ b/src/gpl/src/mbff.cpp
@@ -2191,6 +2191,13 @@ void MBFF::Run(const int mx_sz, const float alpha, const float beta)
     tot_ilp += RunClustering(FFs[i], mx_sz, alpha, beta, array_mask);
   }
 
+  // delete test_trays
+  for (int i = 0; i < test_idx_; i++) {
+    const std::string test_tray_name = "test_tray_" + std::to_string(i);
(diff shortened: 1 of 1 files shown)


src/gpl/src/mbff.cpp
```diff
@@ -2191,6 +2191,13 @@ void MBFF::Run(const int mx_sz, const float alpha, const float beta)
     tot_ilp += RunClustering(FFs[i], mx_sz, alpha, beta, array_mask);
   }
 
+  // delete test_trays
+  for (int i = 0; i < test_idx_; i++) {
+    const std::string test_tray_name = "test_tray_" + std::to_string(i);
+    dbInst* inst = block_->findInst(test_tray_name.c_str());
+    dbInst::destroy(inst);
+  }
```
QuantamHD, line 2199: @maliberty Not necessarily for this PR, but I think it would be good to encapsulate the test trays into a parent object where this code runs in its destructor.

maliberty (approved):

src/gpl/src/mbff.cpp
maliberty: I think this and cts' TechChar should be refactored to have a common scratch pad class for this purpose.

Issue #5867, opened 2024-10-03
DRT: incremental Pin Access with odb callbacks


eder-matheus:
### Description

During incremental grt, the APs stored in odb can be invalidated when moving instances or changing their orientation. DRT should be able to update the APs for the iterms from modified instances incrementally, using the odb callbacks to understand what should be updated.

maliberty:
I think this could be good for @bnmfw as he has been learning the PA code.

bnmfw:
I don't yet have the full picture for the PA module, but from what I understand right now there is no way to incrementally change access to instances. The starting point of the PA is `FlexPA::prep()` where every pin access is determined at once. Today me an @eder-matheus had a meeting where we saw that **apparently** during incremental DRT the whole PA flow is done from scratch.  

### How It could be done
Determining the Access Points of an instance could be done in a vacuum. `initAllAccessPoints()` could be refactor to have a function responsible determining the Access Points of an instance. (Right now it iterates over instances and then pins directly).

Determining the Access Patterns of an instance could **probably** be done in a vacuum. `prepPatternInst()` already determines Access Patterns for a single instance. I'm still refactoring this part of code so I'm not 100% sure if it would be safe to call this function alone.

The problem starts in the instance rows. I still haven't got to this part of code so I'm speculating, once I'm more firm about this I will update this comment. `genInstRowPattern()` is responsible for determining the used Access Patterns of a single row. Changing an instance and its Access Patterns might change the Access Patters used for the other instances in the row, therefore most probably major changes would have to be done to allow an instance to change in a row with minimal changes to other instances in the row.

bnmfw:
I'm using this message to mark the point where a prototype of incremental PA is working. It currently set to only work in bulk, which needs to be changed. The next few PRs will set the structures for incremental to work instance by instance.

PR #6522, opened 2025-01-14
dpl: add API to get adjacent cells cluster

eder-matheus:
This PR adds a public API for DPL that returns a cluster of instances, where every instances is adjacent to at least one instance inside the cluster.

This was requested by @bnmfw for his incremental pin access update.

This cluster should only contain instances that the pin access will do something, so tapcells and endcaps are ignored. Also, since macros are handled separately, they're also ignored.

### DIFF
--- a/src/dpl/src/Opendp.cpp
+++ b/src/dpl/src/Opendp.cpp
@@ -377,6 +377,40 @@ void Opendp::groupInitPixels2()
   }
 }
 
+std::pair<dbInst*, dbInst*> Opendp::getAdjacentInstances(dbInst* inst) const
+{
+  const Rect inst_rect = inst->getBBox()->getBox();
+  Point left(inst_rect.xMin() - 1, inst_rect.center().getY());
+  Point right(inst_rect.xMax() + 1, inst_rect.center().getY());
+
+  const Rect core = grid_->getCore();
+  GridX left_x = grid_->gridX(DbuX{left.getX() - core.xMin()});
+  GridY left_y = grid_->gridSnapDownY(DbuY{left.getY() - core.yMin()});
+
+  GridX right_x = grid_->gridX(DbuX{right.getX() - core.xMin()});
+  GridY right_y = grid_->gridSnapDownY(DbuY{right.getY() - core.yMin()});
+
+  Pixel* left_pixel = grid_->gridPixel(left_x, left_y);
+  Pixel* right_pixel = grid_->gridPixel(right_x, right_y);
+
+  dbInst* left_inst = nullptr;
+  dbInst* right_inst = nullptr;
+
+  if (left_pixel != nullptr) {
+    if (left_pixel->cell) {
+      left_inst = left_pixel->cell->db_inst_;
+    }
+  }
+
+  if (right_pixel != nullptr) {
+    if (right_pixel->cell) {
+      right_inst = right_pixel->cell->db_inst_;
+    }
+  }
+
+  return {left_inst, right_inst};
+}
+
 /* static */
 bool Opendp::isInside(const Rect& cell, const Rect& box)
 {
--- a/src/dpl/src/Opendp.cpp
+++ b/src/dpl/src/Opendp.cpp
@@ -381,21 +381,19 @@ dbInst* Opendp::getAdjacentInstance(dbInst* inst, bool left) const
 {
   const Rect core = grid_->getCore();
   const Rect inst_rect = inst->getBBox()->getBox();
-  GridX left_x = grid_->gridX(DbuX{inst_rect.xMin() - 1 - core.xMin()});
-  GridX right_x = grid_->gridX(DbuX{inst_rect.xMax() + 1 - core.xMin()});
+  DbuX x_dbu = left ? DbuX{inst_rect.xMin() - 1} : DbuX{inst_rect.xMax() + 1};
+  x_dbu -= core.xMin();
+  GridX x = grid_->gridX(x_dbu);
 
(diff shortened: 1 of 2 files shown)


src/dpl/src/Opendp.cpp
```diff
@@ -377,6 +377,69 @@ void Opendp::groupInitPixels2()
   }
 }
 
+std::pair<dbInst*, dbInst*> Opendp::getAdjacentInstances(dbInst* inst) const
+{
+  const Rect inst_rect = inst->getBBox()->getBox();
+  Point left(inst_rect.xMin() - 1, inst_rect.center().getY());
+  Point right(inst_rect.xMax() + 1, inst_rect.center().getY());
```
maliberty, line 384: Note that this is ok for single-height cells but not multi-height.  Pin access needs work for multi-height cells but we need to at least ensure compatibility with the current behavior.  Please check with @bnmfw is to what is expected.

src/dpl/src/Opendp.cpp
```diff
@@ -377,6 +377,69 @@ void Opendp::groupInitPixels2()
...
+
+  const Rect core = grid_->getCore();
+  GridX left_x = grid_->gridX(DbuX{left.getX() - core.xMin()});
+  GridY left_y = grid_->gridSnapDownY(DbuY{left.getY() - core.yMin()});
```
maliberty, line 388: The y is the same so no need for left_y and right_y.

src/dpl/src/Opendp.cpp
maliberty, line 384: Do `inst_rect.center().getY()` once

maliberty (changes requested):

src/dpl/src/Opendp.cpp
bnmfw: Actually, to ensure compatibility with the current behavior the left and right points should have a yMin() instead of center().getY().

src/dpl/src/Opendp.cpp
```diff
@@ -377,6 +377,69 @@ void Opendp::groupInitPixels2()
...
+    auto [l, r] = getAdjacentInstances(left);
+    left = l;
+  }
+
+  while (right != nullptr) {
+    adj_inst_cluster.push_back(right);
+    // the left instance can be ignored, since it was added in the line above
+    auto [l, r] = getAdjacentInstances(right);
+    right = r;
+  }
```
bnmfw, line 434: What is the complexity of finding an dbInst given its coordinates on `getAdjacentInstances()`? In this implementation `getAdjacentInstancesCluster()` searches each instance in the cluster two times, except the original and borders. This implies a complexity of `O(|clusters| * complexity of finding an instance)` of redundant work. Although the number of instances in a cluster is likely small, if the search complexity is worst than `O(log n)` I would consider optimizing the dead searches out. Probably a `getLeftInstance()` and `getRightInstance()` functions that are auxiliary to the others.

maliberty: I think it is an easy and worthwhile optimization.

src/dpl/src/Opendp.cpp
```diff
@@ -377,6 +377,58 @@ void Opendp::groupInitPixels2()
   }
 }
 
+dbInst* Opendp::getAdjacentInstance(dbInst* inst, bool left) const
+{
+  const Rect core = grid_->getCore();
+  const Rect inst_rect = inst->getBBox()->getBox();
+  GridX left_x = grid_->gridX(DbuX{inst_rect.xMin() - 1 - core.xMin()});
+  GridX right_x = grid_->gridX(DbuX{inst_rect.xMax() + 1 - core.xMin()});
```
maliberty, line 385: You will only need one of these so it is better to do
```
DbuX x_dbu = left ? inst_rect.xMin() - 1 : inst_rect.xMax() + 1;
x_dbu -= core.xMin();
GridX x = grid_->gridX(x_dbu);
Pixel* pixel= grid_->gridPixel(x, y);
```

src/dpl/src/Opendp.cpp
```diff
@@ -377,6 +377,58 @@ void Opendp::groupInitPixels2()
...
+  GridY y = grid_->gridSnapDownY(DbuY{inst_rect.yMin() - core.yMin()});
+
+  Pixel* pixel
+      = left ? grid_->gridPixel(left_x, y) : grid_->gridPixel(right_x, y);
+
+  dbInst* adjacent_inst = nullptr;
+
+  if (pixel != nullptr) {
+    // do not return macros, endcaps and tapcells
+    if (pixel->cell && pixel->cell->db_inst_->isCore()) {
```
maliberty, line 396: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/docs/contrib/CodingPractices.md#practice-20

src/dpl/src/Opendp.cpp
```diff
@@ -377,6 +377,58 @@ void Opendp::groupInitPixels2()
...
+  while (right_inst != nullptr) {
+    adj_inst_cluster.push_back(right_inst);
+    // the left instance can be ignored, since it was added in the line above
+    right_inst = getAdjacentInstance(right_inst, right);
+  }
+
+  auto cmp = [](const dbInst* inst1, const dbInst* inst2) {
+    return inst1->getLocation().getX() < inst2->getLocation().getX();
+  };
+  std::sort(adj_inst_cluster.begin(), adj_inst_cluster.end(), cmp);
```
maliberty, line 428: No need to sort - push the left elements, reverse them, push the middle, then push the right elements.

maliberty (approved):

bnmfw:
I'm just mentioning #5867 so I can find this faster

PR #6523, opened 2025-01-14
added new assert helper functions to avoid ambiguity/bad behavior

jeffng-or:
Found some issues with the current asserts, so creating a set of situation-specific asserts to resolve any ambiguity.

### DIFF
--- a/src/odb/test/helper.tcl
+++ b/src/odb/test/helper.tcl
@@ -193,6 +193,42 @@ proc assert { condition { message "assertion failed" } } {
 }
 
 #
+# Assertion helper for strings
+#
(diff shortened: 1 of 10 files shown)


maliberty (approved):

maliberty:
What was the issue/ambiguity you fixed?

jeffng-or:
> What was the issue/ambiguity you fixed?

I was seeing inconsistent behavior with the string equal and whether an object was not “NULL” as I was passing it to my assert method. So, I decided to create situation specific asserts:

- String is equal
- String is not equal
- Object is null
- Object is not null

Then, I fixed some of my test bugs that the current assert was letting past, so I think it’s more correct now.

PR #6533, opened 2025-01-15
dft: implementing interface to support different test modes

fgaray:
This is a refactoring change + public API change where we start supporting different tests modes. The test mode currently does nothing except for independendly architect and stitch each test mode.

In the future, different tests modes may have different types of scan (ex: internal scan, scan compression, boundary scan, etc) and may support to include/exclude cells in certain modes.

### DIFF
--- a/src/dft/src/config/ScanArchitectConfig.hh
+++ b/src/dft/src/config/ScanArchitectConfig.hh
@@ -48,6 +48,8 @@ class ScanArchitectConfig
     ClockMix  // We architect the flops of different clock and edge together
   };
 
+  ScanArchitectConfig() : clock_mixing_(ClockMixing::NoMix) {}
+
   void setClockMixing(ClockMixing clock_mixing);
 
   // The max length in bits that a scan chain can have
--- a/src/dft/src/config/TestModeConfig.hh
+++ b/src/dft/src/config/TestModeConfig.hh
@@ -1,7 +1,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 // BSD 3-Clause License
 //
-// Copyright (c) 2023, Google LLC
+// Copyright (c) 2025, Google LLC
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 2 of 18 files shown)


src/dft/src/dft.tcl
```diff
@@ -124,3 +133,19 @@ proc report_dft_config { args } {
   sta::parse_key_args "report_dft_config" args keys {} flags {}
   dft::report_dft_config
 }
+
+
+sta::define_cmd_args "create_dft_test_mode" { [-test_mode test_mode] }
```
maliberty, line 138: Is -test_mode really optional?  If not provided the command is a silent no-op which seems unexpected.

src/dft/test/sub_modules_sky130.tcl
```diff
@@ -9,6 +9,8 @@ link_design sub_modules
 
 create_clock -name main_clock -period 2.0000 -waveform {0.0000 1.0000} [get_ports {clock}]
 
+set_dft_config -max_chains 1
```
maliberty, line 12: This has no effect on the ok file and doesn't relate to test modes - what is the purpose?

src/dft/src/config/TestModeConfig.hh
```diff
@@ -0,0 +1,64 @@
...
+class TestModeConfig
+{
+ public:
+  TestModeConfig(const std::string& name) : name_(name) {}
+  // Not copyable or movable.
+  TestModeConfig(const TestModeConfig&) = delete;
+  TestModeConfig& operator=(const TestModeConfig&) = delete;
+
+  ScanArchitectConfig* getMutableScanArchitectConfig();
+  const ScanArchitectConfig& getScanArchitectConfig() const;
```
maliberty, line 49: Is there a reason one is by pointer and the other by reference?

src/dft/src/config/TestModeConfig.hh
```diff
@@ -0,0 +1,64 @@
...
+namespace dft {
+
+// Holds the config of this particular test mode.
+class TestModeConfig
+{
+ public:
+  TestModeConfig(const std::string& name) : name_(name) {}
+  // Not copyable or movable.
+  TestModeConfig(const TestModeConfig&) = delete;
+  TestModeConfig& operator=(const TestModeConfig&) = delete;
```
maliberty, line 46: You have no prevented moving, only copying.

src/dft/src/config/ScanArchitectConfig.hh
```diff
@@ -48,6 +48,8 @@ class ScanArchitectConfig
     ClockMix  // We architect the flops of different clock and edge together
   };
 
+  ScanArchitectConfig() : clock_mixing_(ClockMixing::NoMix) {}
```
maliberty, line 51: You can just do
```
ClockMixing clock_mixing_{NoMix};
```
and skip the ctor.  Does this relate to test mode?

src/dft/src/config/DftConfig.cpp
```diff
@@ -34,33 +34,46 @@
...
+TestModeConfig* DftConfig::getOrDefaultMutableTestModeConfig(
+    const std::string& name)
 {
-  return scan_stitch_config_;
+  auto found = test_modes_config_.find(name);
+  if (found == test_modes_config_.end()) {
+    if (name == kDefaultTestModeName) {
+      return createTestMode(name);
+    }
+    return nullptr;
```
maliberty, line 68: getOrDefaultMutableTestModeConfig sounds like it should return default if not found.

maliberty (changes requested):

src/dft/src/dft.tcl
fgaray: Ah, yeah, I will fix this, good point. I will revisit this PR on Friday

maliberty:
It would be nice to add a test with multiple test modes even if it doesn't do much yet.

src/dft/test/sub_modules_sky130.tcl
fgaray: Yes. The user should always set at leaast a max_length or a chain_count. I added now some validation logic to force this.

fgaray:
>It would be nice to add a test with multiple test modes even if it doesn't do much yet.

There is not much to test now as I just wanted to finish the config hierarchy for other developers.

Test modes need more work to be implements. For one we need to add new ports to select the current active mode, ways to include/exclude cells in each test mode and add muxes in the scan chain signals depending on what cells are part of what test mode.

For sure more work than can be done in just one PR :)

I added an error message notifying that we don't currently support more than one test mode.

src/dft/src/config/TestModeConfig.hh
maliberty: Yes but why the difference in & vs * ?

maliberty:
@fgaray this seems to have gotten lost and now has a bunch of merge conflicts.  Is this something we should try to cleanup and get merged?

(@AcKoucher FYI)

PR #6541, opened 2025-01-16
Cts enable obstruction aware

arthurjolo:
This pull request enables the CTS option obstruction_aware by default.
Also this PR enhances the obstruction_aware CTS to consider the tree topology during its legalization, trying to maintain the tree as similar to as it was originally build.


src/cts/src/TritonCTS.tcl
```diff
@@ -102,7 +101,7 @@ proc clock_tree_synthesis { args } {
           -sink_clustering_levels -tree_buf \
           -sink_buffer_max_cap_derate -delay_buffer_derate -library} \
     flags {-post_cts_disable -sink_clustering_enable -balance_levels \
-           -obstruction_aware -apply_ndr -dont_use_dummy_load
```
gadfort, line 105: Could we keep the flag and add a deprecation warning? That way we preserve some backwards compatibility at least for a while.

precisionmoon: Yes, Peter's suggestion is good.   Also, we need an option to disable obstruction aware like -no_obstruction_aware.

precisionmoon (approved):
I reviewed the code changes with Arthur.   They are fine.

maliberty:
The code looks fine.  Is there an ORFS PR that needs coordinated merging?

arthurjolo:
I am waiting for a run with the new sta to finish.

arthurjolo:
> The code looks fine. Is there an ORFS PR that needs coordinated merging?

There is one design that needs antenna diode count metrics update I created a PR for it.

PR #6546, opened 2025-01-17
gha: no longer need to upload these files

vvbandeira:


QuantamHD:
I'm pretty sure this broke clang tidy. @maliberty

maliberty:
@QuantamHD I believe this was fixing the prior break in clang tidy.

vvbandeira:
@QuantamHD I made this change because we were seeing errors in the GHA saying that there were attempts to upload multiple artifacts with the same name. From the docs of the clang-tidy-review action the upload of these files I removed are already done by the action itself automatically:
https://github.com/The-OpenROAD-Project/clang-tidy-review?tab=readme-ov-file#usage-in-fork-environments-split-workflow

PR #6543, opened 2025-01-16
mpl2: ignore fixed instances

AcKoucher:
Needs #6547
Resolve #6092 

Needed to switch from manual macro placement to automated macro placement in gf12/bp_single.
Note that I also needed to include the ignore check in the Orientation Improvement step.

PR #6547, opened 2025-01-17
mpl2: do not allow fixed instances inside floorplan area

AcKoucher:
Based on the discussion in #6543 so that we can ignore fixed instances.

Also changing the feasibility of the macro placement to be based on the area computed using the global fence instead of directly the core.


AcKoucher:
I ran a Secure-CI for this and it's fine.

eder-matheus (approved):

Issue #6532, opened 2025-01-15
DRT integration with GRT include issue


bnmfw:
### Description

For #5867 to be possible, it is essential for the GRT to have access to the DRT interface. During the attempts to make all the includes to allow for this to happen many issues arose. @eder-matheus mentioned that there have been previous attempts to implement this integration which lead to the same problems. The goal is, in any function of GlobalRouter.cpp, to be able to call drt_->anyPublicFunction().

### Main Attempted Changes Diff:
drt/CMakeLists.txt:
```
--- a/src/grt/CMakeLists.txt
+++ b/src/grt/CMakeLists.txt
@@ -88,6 +88,7 @@ target_link_libraries(grt_lib
     dbSta_lib
     stt_lib
     rsz_lib
+    drt
     OpenSTA
     Boost::boost
     OpenMP::OpenMP_CXX
@@ -103,6 +104,7 @@ target_link_libraries(grt
     gui
     stt
     rsz
+    drt
     OpenSTA
     Boost::boost
     OpenMP::OpenMP_CXX
@@ -131,6 +133,7 @@ if (Python3_FOUND AND BUILD_PYTHON)
       dbSta
       gui
       stt
+      drt
       OpenSTA
       Boost::boost
   )
```
GlobalRouter.cpp:

```
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -69,6 +69,7 @@
 #include "odb/wOrder.h"
 #include "rsz/Resizer.hh"
 #include "rsz/SpefWriter.hh"
+#include "triton_route/TritonRoute.h"
 #include "sta/Clock.hh"
 #include "sta/MinMax.hh"
 #include "sta/Parasitics.hh"
```

bnmfw:
@maliberty eder mentioned that you might have some ideas on how to tackle this.

maliberty:
What are the issues you are facing?

maliberty:
I tried the above and it looks like a simple linking issue.  I've got a PR coming.

maliberty:
Please make a small reproducer on a branch or PR

bnmfw:
I have the [branch here](https://github.com/bnmfw/OpenROAD_Bernardo_Sandoval/tree/drt_grt_integration), it uses a dummy function to call drt

QuantamHD:
Largely the thing you need to do is decouple the openroad singleton from drt

[...]

PR #6550, opened 2025-01-18
drt: decoupling of the Graphics classes and stable drt_lib

bnmfw:
Tackles #6532

This PR has a secure CI run associated with it and will be made open once it passes.

### Abstract Classes
This PR makes so the core code of the `DRT` module is not dependent on the `GUI`. Abstract classes for the `PA`, `TA` and `DR` submodules were created so these are the classes the `DRT` module knows at compile time, while the concrete classes, the ones dependent on the `GUI` are passed during run time.

### Graphic Classes as Class Attributes
The graphic classes used to be instantiated during the `DRT` execution and passed around. Now, all three are instantiated during `initTritonRoute()` and passed to the `DRT`. Then, they become class objects that are held by TritonRoute during it whole execution or until are moved to the respective submodule manager.

### CMakeLists.txt modification
Now there is a clear separation between `drt` and `drt_lib`. I'm making the PR at a point where this separation can be done without any crashes, so it is stable. However it is still not possible to link `drt_lib` to the `GRT` without running into problems regarding the OR singleton dependency. This PR is becoming big, so I'm submitting it now, further linking problems will be tackled in a separate PR.


bnmfw:
I'm working on creating the abstract classes to invert said dependencies and will probably make a PR for it tomorrow.

QuantamHD:
You need to make sure you're using the exact same clang format version. Ubuntu clang-format version 18.1.3

I've noticed there's subtle difference between local clang-format and the one installed in the CI lately.

src/drt/src/dr/AbstractDRGraphics.h
```diff
@@ -0,0 +1,78 @@
+// Copyright 2025 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
```
QuantamHD, line 5: Should be Precision copyright.

src/drt/include/triton_route/TritonRoute.h
```diff
@@ -221,6 +228,11 @@ class TritonRoute
   int results_sz_{0};
   unsigned int cloud_sz_{0};
   std::optional<boost::asio::thread_pool> dist_pool_;
+  std::unique_ptr<drt::FlexPA> pa_{nullptr};
+  std::unique_ptr<drt::FlexTA> ta_{nullptr};
```
maliberty, line 232: Why do these need to be persistent now?

src/drt/src/dr/AbstractDRGraphics.h
```diff
@@ -0,0 +1,78 @@
...
+
+class FlexGridGraph;
+class FlexWavefrontGrid;
+class FlexDRWorker;
+class drNet;
+class frDesign;
+class frBlockObject;
+struct RouterConfiguration;
+
+class AbstractDRGraphics
```
maliberty, line 34: I believe many of these methods don't need to be defined in this interface.  Only methods called from the rest of dr need to be here.  For example status() and methods from Renderer API.  Likewise with the other abstract base classes for graphics.

src/drt/src/pa/AbstractPAGraphics.h
```diff
@@ -0,0 +1,67 @@
+// Copyright 2025 Google LLC
```
maliberty, line 1: Same comment on PII

src/drt/src/ta/AbstractTAGraphics.h
```diff
@@ -0,0 +1,43 @@
+// Copyright 2025 Google LLC
```
maliberty, line 1: Same comment on PII

src/grt/CMakeLists.txt
maliberty, line 1: What motivates the rearranging of this file?

src/drt/include/triton_route/TritonRoute.h
bnmfw: They don't need now, but at least PA needs to be persistent to allow for PA with ODB callbacks, I was just starting to prepare things

maliberty: I think incremental PA will capture a dirty list during the callbacks and then re-run in a single invocation for efficiency.  We would still have an entry point in which to setup

maliberty (approved):

PR #6562, opened 2025-01-20
dbSta: Add -module option to report_cell_usage

bangqixu:
As discussed before, we will introduce another few options to report_cell_usage. As a prerequisite, we need to add a "-module" option to the current report_cell_usage.

### DIFF
--- a/src/dbSta/src/dbSta.tcl
+++ b/src/dbSta/src/dbSta.tcl
@@ -79,25 +79,27 @@ proc highlight_path { args } {
   }
 }
 
-define_cmd_args "report_cell_usage" {[-verbose] [module_inst]}
+define_cmd_args "report_cell_usage" {\
+	                             [-verbose] \
+	                             [-module]
+	                            }
 
 proc report_cell_usage { args } {
-  parse_key_args "highlight_path" args keys {} \
+  parse_key_args "highlight_path" args keys {-module} \
     flags {-verbose} 0
-
-  check_argc_eq0or1 "report_cell_usage" $args
 
   if { [ord::get_db_block] == "NULL" } {
     sta_error 1001 "No design block found."
   }
 
   set module [[ord::get_db_block] getTopModule]
-  if { $args != "" } {
-    set modinst [[ord::get_db_block] findModInst $args]
-    if { $modinst == "NULL" } {
-      sta_error 1002 "Unable to find $args"
+  if { [info exists keys(-module)] } {
+    set module_inst_name $keys(-module)
+    set module_inst [[ord::get_db_block] findModInst $module_inst_name ]
+    if { $module_inst == "NULL" } {
+      sta_error 1002 "Unable to find $module_inst_name"
     }
-    set module [$modinst getMaster]
+    set module [$module_inst getMaster]
   }
 
   report_cell_usage_cmd $module [info exists flags(-verbose)]
--- a/src/dbSta/src/dbSta.tcl
+++ b/src/dbSta/src/dbSta.tcl
@@ -88,6 +88,8 @@ proc report_cell_usage { args } {
   parse_key_args "highlight_path" args keys {-module} \
     flags {-verbose} 0
 
+  check_argc_eq0or1 "report_cell_usage" $args
+
   if { [ord::get_db_block] == "NULL" } {
     sta_error 1001 "No design block found."
   }
--- a/src/dbSta/test/report_cell_usage_modinsts.tcl
+++ b/src/dbSta/test/report_cell_usage_modinsts.tcl
@@ -8,5 +8,5 @@ link_design top
 
 report_cell_usage -verbose
 
-report_cell_usage -verbose b1
-report_cell_usage -verbose b2
+report_cell_usage -verbose -module b1
+report_cell_usage -verbose -module b2
--- a/src/dbSta/test/report_cell_usage_modinsts_metrics.tcl
+++ b/src/dbSta/test/report_cell_usage_modinsts_metrics.tcl
@@ -12,8 +12,8 @@ utl::open_metrics $metrics
 
 report_cell_usage -verbose
 
-report_cell_usage -verbose b1
-report_cell_usage -verbose b2
+report_cell_usage -verbose -module b1
+report_cell_usage -verbose -module b2
 
 utl::close_metrics $metrics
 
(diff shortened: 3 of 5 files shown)


src/dbSta/src/dbSta.tcl
```diff
@@ -79,25 +79,27 @@ proc highlight_path { args } {
...
+	                             [-verbose] \
+	                             [-module]
+	                            }
 
 proc report_cell_usage { args } {
-  parse_key_args "highlight_path" args keys {} \
+  parse_key_args "highlight_path" args keys {-module} \
     flags {-verbose} 0
 
-  check_argc_eq0or1 "report_cell_usage" $args
```
gadfort, line 88: Why remove this?

src/dbSta/src/dbSta.tcl
```diff
@@ -79,25 +79,27 @@ proc highlight_path { args } {
...
   if { [ord::get_db_block] == "NULL" } {
     sta_error 1001 "No design block found."
   }
 
   set module [[ord::get_db_block] getTopModule]
-  if { $args != "" } {
-    set modinst [[ord::get_db_block] findModInst $args]
-    if { $modinst == "NULL" } {
-      sta_error 1002 "Unable to find $args"
+  if { [info exists keys(-module)] } {
```
gadfort, line 96: This change does not appear to be backwards compatible, can you please add a test or revert one of the tests to ensure backwards compatibility

gadfort (changes requested):

src/dbSta/src/dbSta.tcl
bangqixu: It is true that this is not backward compatible. When I was discussing with @maliberty in a previous pull request, he suggested introducing such an option and backward compatibility did not seem to be a concern. See https://github.com/The-OpenROAD-Project/OpenROAD/pull/6469#issuecomment-2585127694

src/dbSta/src/dbSta.tcl
bangqixu: The reason is that the next CL will introduce more arguments. You can argue that this is not needed now, but since the next CL will come in soon, I think it is fine to remove this here. I don't mind reverting this change if you think it is better to remove it later.

src/dbSta/src/dbSta.tcl
gadfort: https://github.com/The-OpenROAD-Project/OpenROAD/pull/6469#issuecomment-2585127694

Since I rely on the current API I would prefer maintaining backwards compatibility (you can add a deprecation warning if you would like)

src/dbSta/src/dbSta.tcl
gadfort: This is a general error check that should be kept.

src/dbSta/src/dbSta.tcl
bangqixu: How can I add a new option while keeping backward compatibility? For example, if I want to add "-format" and "-file" options, is it doable while keeping backward compatibility? Currently the command assumes there is only one option (not flag) that refers to the target module.

src/dbSta/src/dbSta.tcl
gadfort: This is checking how many remaining arguments are left after parsing the arguments (is. Only checking the positional arguments). So I would assume that for backwards compatibility you should keep it 0 or 1.

src/dbSta/src/dbSta.tcl
gadfort: The flags are handled separately from the positional arguments, so adding new arguments doesn't have to change anything else.

src/dbSta/src/dbSta.tcl
bangqixu: Well. I am more than happy to introduce a new command without touching the existing API, but @maliberty suggested that it would be better to modify this command. See the discussion in the previous PR.

src/dbSta/src/dbSta.tcl
bangqixu: Can you give me a pointer to an example which I can follow?

bangqixu:
@maliberty can you help chime in here?

src/dbSta/src/dbSta.tcl
maliberty: We are not introducing a new command.  I don't see any motivation in this PR to remove this check.  Adding more arguments has nothing to do with this check which is for any extraneous arguments.

maliberty: In your last PR you were going to write up a description of your end state but I haven't seen it.  Trying to guess what you are planning to do and how to resolve your concerns isn't very productive.   My suggestion was to add new options as needed, not to break the existing use model.

bangqixu: That is not true. In https://github.com/The-OpenROAD-Project/OpenROAD/pull/6469#issuecomment-2579406586, I mentioned that I was going to do report_cell_usage -format json -file <path_to_file>" and you gave a thumb up to that.

I brought up the concern of backward compatibility and you suggested with "You could add a -module option". See https://github.com/The-OpenROAD-Project/OpenROAD/pull/6469#issuecomment-2585127694 .

Am I not doing what you agreed on?

bangqixu: Just to be fair. I added the check back. We still good with "-module" option?

bangqixu:
> @maliberty can you help chime in here? It would be good if you and @gadfort can reach a consensus before I make further changes.

Have you got a chance to discuss this? Thanks.

bangqixu:
Had discussion with @QuantamHD and I didn't know that $args is whatever is not matched. I will add "-module" then.

maliberty:
> Had discussion with @QuantamHD and I didn't know that $args is whatever is not matched. I will add "-module" then.

It sounds like you have a next step then.  Let me know when you are ready for a next review.

Btw, you'll need to fix the DCO before this can be merged.

bangqixu:
> > Had discussion with @QuantamHD and I didn't know that $args is whatever is not matched. I will add "-module" then.
> 
> It sounds like you have a next step then. Let me know when you are ready for a next review.
> 
> Btw, you'll need to fix the DCO before this can be merged.

Will do. Thank you!

PR #6566, opened 2025-01-22
utl: provide ability to convert NaN from tcl

gadfort:
Add a string pass through to allow cpp to handle casting of NaN from tcl.

Why? The new Sta code seems to sometimes respond with NaN for the power which is causing `report_power_metrics` to fail
```
Error: reports.tcl, 118 TypeError in method 'metric_float', argument 2 of type 'double'
```


maliberty (approved):

gadfort:
> I think NaN from sta is probably worth some investigation as well.

Issue created: https://github.com/The-OpenROAD-Project/OpenROAD/issues/6574

PR #6571, opened 2025-01-22
rmp: First resynthesis strategy ZeroSlackStrategy

QuantamHD:
This is the first working resynthesis strategy. I haven't added it to the restructure command yet, but would like to do that in a follow up PR. Right now the resynth leans a bit area heavy.

see TestAbc.cc changes for example usage.

### DIFF
--- a/src/rmp/src/resynthesis_strategy.h
+++ b/src/rmp/src/resynthesis_strategy.h
@@ -11,22 +11,10 @@
 
 namespace rmp {
 
-struct OptimizationResult
-{
-  double initial_area;
-  double final_area;
-  double initial_wns;
-  double final_wns;
-  double initial_tns;
-  double final_tns;
-};
-
 class ResynthesisStrategy
 {
  public:
   virtual ~ResynthesisStrategy() = default;
-  virtual OptimizationResult OptimizeDesign(sta::dbSta* sta,
-                                            utl::Logger* logger)
-      = 0;
+  virtual void OptimizeDesign(sta::dbSta* sta, utl::Logger* logger) = 0;
 };
 }  // namespace rmp
--- a/src/rmp/src/resynthesis_strategy.h
+++ b/src/rmp/src/resynthesis_strategy.h
@@ -1,0 +1,32 @@
+// Copyright 2025 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+#pragma once
+
+#include "db_sta/dbSta.hh"
+#include "utl/Logger.h"
+
+namespace rmp {
+
+struct OptimizationResult
+{
+  double initial_area;
+  double final_area;
+  double initial_wns;
+  double final_wns;
+  double initial_tns;
+  double final_tns;
+};
+
+class ResynthesisStrategy
+{
+ public:
+  virtual ~ResynthesisStrategy() = default;
+  virtual OptimizationResult OptimizeDesign(sta::dbSta* sta,
+                                            utl::Logger* logger)
+      = 0;
+};
+}  // namespace rmp
--- a/src/rmp/src/zero_slack_strategy.cpp
+++ b/src/rmp/src/zero_slack_strategy.cpp
@@ -6,7 +6,7 @@
 
 #include "zero_slack_strategy.h"
 
-#include <iostream>
+#include <vector>
 
 #include "abc_library_factory.h"
 #include "db_sta/dbNetwork.hh"
@@ -24,7 +24,7 @@
 
 namespace rmp {
 
-std::vector<sta::Vertex*> GetNegativeVertices(sta::dbSta* sta)
+std::vector<sta::Vertex*> GetNegativeEndpoints(sta::dbSta* sta)
 {
   std::vector<sta::Vertex*> result;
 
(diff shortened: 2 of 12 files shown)


src/rmp/src/resynthesis_strategy.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include "db_sta/dbSta.hh"
+#include "utl/Logger.h"
+
+namespace rmp {
+
+struct OptimizationResult
+{
+  double initial_area;
+  double final_area;
```
maliberty, line 17: What are the units of area?

src/rmp/src/zero_slack_strategy.h
```diff
@@ -0,0 +1,24 @@
...
+#include "db_sta/dbSta.hh"
+#include "resynthesis_strategy.h"
+#include "utl/Logger.h"
+
+namespace rmp {
+
+class ZeroSlackStrategy : public ResynthesisStrategy
+{
+ public:
+  ZeroSlackStrategy() {};
```
maliberty, line 18: Why write a ctor & dtor if they aren't doing anything?

src/rmp/src/zero_slack_strategy.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    DelayOptimizationStrategy strat(sta);
+    utl::UniquePtrWithDeleter<abc::Abc_Ntk_t> remapped
+        = strat.Optimize(mapped_abc_network.get(), logger);
+
+    cut.InsertMappedAbcNetwork(
+        remapped.get(), abc_library, network, unique_name, logger);
+  }
+
+  return {};
```
maliberty, line 96: return a meaningful value?

src/rmp/src/zero_slack_strategy.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include "sta/PortDirection.hh"
+#include "sta/Units.hh"
+#include "sta/VerilogWriter.hh"
+#include "unique_name.h"
+#include "utl/Logger.h"
+#include "utl/deleter.h"
+
+namespace rmp {
+
+std::vector<sta::Vertex*> GetNegativeVertices(sta::dbSta* sta)
```
maliberty, line 27: GetNegativeEndpoints is more descriptive

src/rmp/src/zero_slack_strategy.cpp
```diff
@@ -0,0 +1,98 @@
...
+  std::vector<sta::Vertex*> result;
+
+  sta::dbNetwork* network = sta->getDbNetwork();
+  for (sta::Vertex* vertex : *sta->endpoints()) {
+    sta::PortDirection* direction = network->direction(vertex->pin());
+    if (!direction->isInput()) {
+      continue;
+    }
+
+    sta::Slack slack = sta->vertexSlack(vertex, sta::MinMax::max());
```
maliberty, line 38: Use const where possible throughout

src/rmp/src/zero_slack_strategy.cpp
```diff
@@ -0,0 +1,98 @@
...
+  sta::dbNetwork* network = sta->getDbNetwork();
+
+  AbcLibraryFactory factory(logger);
+  factory.AddDbSta(sta);
+  AbcLibrary abc_library = factory.Build();
+
+  std::vector<sta::Vertex*> candidate_vertices = GetNegativeVertices(sta);
+
+  rmp::UniqueName unique_name;
+  for (sta::Vertex* negative_vertices : candidate_vertices) {
```
maliberty, line 66: negative_vertex  (its is just one).  Better negative_endpoint

maliberty (changes requested):

src/rmp/src/zero_slack_strategy.cpp
macd: I'm probably misunderstanding something, but it looks like you are collecting _input_ vertices with negative slack? Is that what you want?

QuantamHD: The endpoints are generally the D and Q ports of flip flops. We're trying to grab a cone of logic which we want to start at D ports. So in this case we do want the inputs ports of flops.

maliberty: What about cones from primary outputs?

QuantamHD: I think there's still an input pin there if I'm not mistaken.

maliberty: What would it be an input to?  I would verify that.

macd: Ah, yes, of course. It's just that I had the default Yosys behavior in mind. They do not send any sequentials to ABC.

src/rmp/src/resynthesis_strategy.h
QuantamHD: Done

src/rmp/src/zero_slack_strategy.h
QuantamHD: Done

src/rmp/src/zero_slack_strategy.cpp
QuantamHD: done

src/rmp/src/zero_slack_strategy.cpp
QuantamHD: Done

src/rmp/src/zero_slack_strategy.cpp
QuantamHD: Done

src/rmp/src/zero_slack_strategy.cpp
QuantamHD: Done

QuantamHD:
@maliberty mind taking another look?

maliberty (approved):

PR #6584, opened 2025-01-23
psm: add check for bterms option

gadfort:
Adds:
- flag to enforce bterms are present for a net during `check_power_grid`


src/psm/README.md
```diff
@@ -69,6 +70,7 @@ check_power_grid
 | `-net` | Name of the net to analyze. Must be a power or ground net name. |
 | `-floorplanning` | Ignore non-fixed instances in the power grid, this is useful during floorplanning analysis when instances may not be properly placed. |
 | `-error_file` | File to write power grid errors to. |
+| `-require_terminals` | If specified, this will check of there are terminals on the net. |
```
maliberty, line 73: Do you think this should be off by default or on?

gadfort: I would suggest on by default, but that might have some short term implications on CI tests that run this command

maliberty: Then let's flip the option to on by default with an off option.  We might as well sort out the CI now as in another PR.

gadfort:
@maliberty I switched the polarity of the flag so it will fail by default if bpins are not found

src/psm/test/gcd_test_assign_power.tcl
```diff
@@ -10,5 +10,5 @@ set voltage_file [make_result_file gcd_test_vdd-voltage.rpt]
 set_pdnsim_inst_power -inst _440_ -power 1e-5
 set_pdnsim_inst_power -inst _829_ -power 1e-5
 
-check_power_grid -net VDD
+check_power_grid -net VDD -dont_require_terminals
```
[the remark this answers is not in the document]
gadfort: pdngen can generate them if it's setup to do so, that is the correct place to do it, this is only a checker.

maliberty (approved):

PR #6591, opened 2025-01-24
Cts don't insert delay for clustering buffers

arthurjolo:
This PR removes the insertion delay added to the clustering buffers.

For designs with macros clustered together the delay inserted caused the clustering buffers to be placed outside of the die area causing problems during the legalization.


maliberty:
The code change is simple enough but does this need a secure CI for QOR impact?

arthurjolo:
I did ran a secure CI and it didn't impact any designs, just ca53 (that fixes the crash it was having until yesterday) and it is passing.

maliberty (approved):

Issue #3899, opened 2023-08-24
Support LEF58_CELLEDGESPACINGTABLE & LEF58_EDGETYPE


maliberty:
### Description

Add support for LEF58_EDGETYPE & LEF58_CELLEDGESPACINGTABLE attributes in odb, dpl, and dpo.  

An abstracted example of the required syntax:
```
    MACRO LEF58_EDGETYPE STRING ;
    LIBRARY LEF58_CELLEDGESPACINGTABLE STRING
    "CELLEDGESPACINGTABLE
EDGETYPE TYPE3 TYPE4 0.ZZ
EDGETYPE TYPE1       TYPE1    EXCEPTABUTTED   0.XXX
EDGETYPE TYPE2        TYPE1    EXCEPTABUTTED   0.XXX
EDGETYPE TYPE2        TYPE2                    0.YYY
EDGETYPE TYPE12CPP   TYPE12CPP   EXCEPTABUTTED   0.XXX
EDGETYPE TYPE1       TYPE12CPP   EXCEPTABUTTED   0.XXX
EDGETYPE TYPE12CPP   TYPE2                    0.XXX
      ;" ;

  PROPERTY LEF58_EDGETYPE "
    EDGETYPE LEFT TYPE1 ;
    EDGETYPE RIGHT TYPE2 ; " ;

  PROPERTY LEF58_EDGETYPE "
    EDGETYPE LEFT TYPE1 ;
    EDGETYPE RIGHT TYPE2 ;
    EDGETYPE TOP TYPE4 RANGE 0 0.VVV ;
    EDGETYPE BOTTOM TYPE4 RANGE 0 0.VVV ;
    EDGETYPE TOP TYPE3 RANGE 0.VVV 0.QQQ ;
    EDGETYPE BOTTOM TYPE3 RANGE 0.VVV 0.QQQ ;
    EDGETYPE TOP TYPE4 RANGE 0.QQQ T.TTT ;
    EDGETYPE BOTTOM TYPE4 RANGE 0.QQQ T.TTT ; " ;
```

lefin & odb should support the full LEF syntax.  In dpl/dpo, priority should be given to the syntax above as the full property definitions contain more options that are not immediately required.

### Suggested Solution

The odb work should follow the template of other LEF properties.  In dpl this should extend the set of checks made when detecting if a placement is valid during the diamond search.

### Additional Context

@gadfort FYI

PR #6595, opened 2025-01-26
dpl: support LEF58_CELLEDGESPACINGTABLE

osamahammad21:
Linked to https://github.com/The-OpenROAD-Project/OpenROAD/issues/3899


osamahammad21:
There will be a followup PR for optimizing performance of the checkEdgeSpacing function. This is just an initial attempt.

src/dpl/src/CheckPlacement.cpp
```diff
@@ -247,6 +259,13 @@ void Opendp::saveFailures(const vector<Cell*>& placed_failures,
     category->setDescription("Cells that DPL failed to place.");
     saveViolations(placement_failures, category);
   }
+  if (!edge_spacing_failures.empty()) {
+    auto category = odb::dbMarkerCategory::createOrReplace(
+        tool_category, "Cell_edge_spacing_failyres");
```
maliberty, line 264: failyres -> failures

src/dpl/src/Place.cpp
```diff
@@ -932,6 +933,112 @@ bool Opendp::checkRegionOverlap(const Cell* cell,
...
+                              const odb::dbOrientType& orient) const
+{
+  auto spacing_rules = db_->getTech()->getCellEdgeSpacingTable();
+  if (spacing_rules.empty()) {
+    return true;
+  }
+  const auto& master = db_master_map_.at(cell->db_inst_->getMaster());
+  auto x_real = gridToDbu(x, grid_->getSiteWidth());
+  auto y_real = grid_->gridYToDbu(y);
+  for (const auto& edge1 : master.edges_) {
```
maliberty, line 967: How is the DEFAULT case handled?

src/dpl/src/Place.cpp
```diff
@@ -932,6 +933,112 @@ bool Opendp::checkRegionOverlap(const Cell* cell,
...
+    for (auto spacing_rule : db_->getTech()->getCellEdgeSpacingTable()) {
+      if (spacing_rule->isOptional() || spacing_rule->isSoft()) {
+        continue;
+      }
+      std::string second_edge_type;
+      if (spacing_rule->getFirstEdgeType() == edge1.getEdgeType()) {
+        second_edge_type = spacing_rule->getSecondEdgeType();
```
maliberty, line 974: Having string comparison in this inner loop feels like a efficiency issue.  I think it would be better to intern the edge type strings in odb and have a getEdgeTypeIndex() method.

maliberty:
> I can create an rtree for edges to use for the search instead, but this would have a memory impact as each cell would have its associated edges. I will hold on this until I try with a real design test case.

I don't think this is necessary.  The search on the pixel grid is already fairly efficient.  The extra memory cost versus performance feels heavy.

src/dpl/src/Place.cpp
```diff
@@ -932,6 +933,112 @@ bool Opendp::checkRegionOverlap(const Cell* cell,
...
+        // vertical edge
+        query_rect = query_rect.bloat(spc, odb::Orientation2D::Horizontal);
+      } else {
+        // horizontal edge
+        query_rect = query_rect.bloat(spc, odb::Orientation2D::Vertical);
+      }
+      auto xMin = grid_->gridX(DbuX(query_rect.xMin()));
+      auto xMax = grid_->gridEndX(DbuX(query_rect.xMax()));
+      auto yMin = grid_->gridEndY(DbuY(query_rect.yMin())) - 1;
+      auto yMax = grid_->gridEndY(DbuY(query_rect.yMax()));
```
QuantamHD, line 999: Can you remove most of the autos from this routine? It would help readability greatly to get a sense of the types.

src/dpl/src/Place.cpp
```diff
@@ -932,6 +933,112 @@ bool Opendp::checkRegionOverlap(const Cell* cell,
...
+          edge1.getBBox(), cell, x_real, y_real, orient);
+      Rect query_rect(edge1_box);
+      bool is_vertical_edge = edge1_box.getDir() == 0;
+      if (is_vertical_edge) {
+        // vertical edge
+        query_rect = query_rect.bloat(spc, odb::Orientation2D::Horizontal);
+      } else {
+        // horizontal edge
+        query_rect = query_rect.bloat(spc, odb::Orientation2D::Vertical);
+      }
```
QuantamHD, line 995: Can you refactor this into a helper function?

src/dpl/src/Place.cpp
```diff
@@ -932,6 +933,112 @@ bool Opendp::checkRegionOverlap(const Cell* cell,
...
+        second_edge_type = spacing_rule->getSecondEdgeType();
+      } else if (spacing_rule->getSecondEdgeType() == edge1.getEdgeType()) {
+        second_edge_type = spacing_rule->getFirstEdgeType();
+      } else if (spacing_rule->getFirstEdgeType() == "DEFAULT") {
+        second_edge_type = spacing_rule->getSecondEdgeType();
+      } else if (spacing_rule->getSecondEdgeType() == "DEFAULT") {
+        second_edge_type = spacing_rule->getFirstEdgeType();
+      } else {
+        continue;
+      }
```
QuantamHD, line 983: Can this be extracted into a helper that returns std::optional<second_edge_type> and then you can continue if the optional is empty.

src/dpl/src/Place.cpp
```diff
@@ -932,6 +933,112 @@ bool Opendp::checkRegionOverlap(const Cell* cell,
...
+      } else {
+        // horizontal edge
+        query_rect = query_rect.bloat(spc, odb::Orientation2D::Vertical);
+      }
+      auto xMin = grid_->gridX(DbuX(query_rect.xMin()));
+      auto xMax = grid_->gridEndX(DbuX(query_rect.xMax()));
+      auto yMin = grid_->gridEndY(DbuY(query_rect.yMin())) - 1;
+      auto yMax = grid_->gridEndY(DbuY(query_rect.yMax()));
+      for (GridY y1 = yMin; y1 <= yMax; y1++) {
+        for (GridX x1 = xMin; x1 <= xMax; x1++) {
```
QuantamHD, line 1001: Please add comment above the for loop that explains at a high level what the intent of the following lines is.

QuantamHD (changes requested):

src/dpl/src/dbToOpendp.cpp
```diff
@@ -107,9 +110,199 @@ void Opendp::makeMacros()
...
+  const int start
+      = is_horizontal ? parent_segment.xMin() : parent_segment.yMin();
+  const int end = is_horizontal ? parent_segment.xMax() : parent_segment.yMax();
+  int current_pos = start;
+  std::vector<Rect> result;
+  for (const Rect& seg : sorted_segs) {
+    int seg_start = is_horizontal ? seg.xMin() : seg.yMin();
+    int seg_end = is_horizontal ? seg.xMax() : seg.yMax();
+    if (seg_end <= current_pos) {
+      continue;
```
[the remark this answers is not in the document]
osamahammad21: I added this line before merging segments. Will remove it

src/dpl/src/Objects.h
```diff
@@ -41,9 +41,38 @@ namespace dpl {
...
+
+struct EdgeSpacingEntry
+{
+  EdgeSpacingEntry(const int spc_in,
+                   const bool is_exact_in,
+                   const bool except_abutted_in)
+      : spc(spc_in), is_exact(is_exact_in), except_abutted(except_abutted_in)
+  {
+  }
+  bool operator<(const EdgeSpacingEntry& rhs) const { return spc < rhs.spc; }
```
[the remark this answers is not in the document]
osamahammad21: I only use this for getting the maximum spacing value. It doesn't matter if a < b or b < a if they both end up with the same spacing value.

src/dpl/src/Place.cpp
```diff
@@ -932,6 +933,113 @@ bool Opendp::checkRegionOverlap(const Cell* cell,
   // be fully contained by the cell's bounding box.
   return result.empty();
 }
+namespace cell_edges {
+Rect transformEdgeRect(const Rect& edge_rect,
```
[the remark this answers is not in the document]
osamahammad21: because x and y are the instance location (lower left corner point of the instance bounding box) not the origin point. applying the inst transform directly is not correct.

src/dpl/src/dbToOpendp.cpp
```diff
@@ -107,9 +110,199 @@ void Opendp::makeMacros()
...
+      }
+    }
+    typed_segs[dir].push_back(edge_rect);
+    edge_types_indices_.try_emplace(edge->getEdgeType(),
+                                    edge_types_indices_.size());
+    master->edges_.emplace_back(edge_types_indices_[edge->getEdgeType()],
+                                edge_rect);
+  }
+  // Add the remaining DEFAULT un-typed segments
+  for (size_t dir_idx = 0; dir_idx <= 3; dir_idx++) {
```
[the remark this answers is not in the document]
osamahammad21: makes sense :+1:

maliberty (approved):

PR #6600, opened 2025-01-27
mpl: add support for IO Pads

### LINKED ISSUE #6585: MPL can't see the connections between IO PADs and the core
### Description

Nowadays mpl2 can't see the connections between IO Pads and elements within the core.

### Suggested Solution

Add the support accordingly.

### Additional Context

_No response_

AcKoucher:
Resolve #6585 

Changes:
- Divide the IO Clusters into two types:
  1) Group of Unplaced IO Pins
  2) IO Pad
- When creating the hypergraph that will be used to build the dataflow maps and iterating over the terminals of an IO PAD net, consider the start of the path to be the bterm mapped to that PAD instead of the PAD itself so that MPL can see the connection properly inside SA.
- Do data flow before creating IO Clusters so that the IO Clusters creation can be based on the dataflow. With this, we'll only create IO Pad Clusters for the PADs that have useful connections macro placement wise.
     Obs: Large nets are excluded from the hypergraph and not used by SA, so we don't want to look at those.
- Use the distance to boundary technique inside SA only for the first type of IO Clusters.

### DIFF
--- a/src/mpl/src/SimulatedAnnealingCore.cpp
+++ b/src/mpl/src/SimulatedAnnealingCore.cpp
@@ -302,7 +302,7 @@ void SimulatedAnnealingCore<T>::calWirelength()
     T& source = macros_[net.terminals.first];
     T& target = macros_[net.terminals.second];
 
-    if (target.isIOCluster()) {
+    if (target.isClusterOfUnplacedIOPins()) {
       addBoundaryDistToWirelength(source, target, net.weight);
       continue;
     }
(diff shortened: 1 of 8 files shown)


AcKoucher:
Running locally with the sta fix to solve the timeout showed significantly better metrics. Even though we can´t merge this due to the timeout, it's ready for review.

maliberty (approved):

AcKoucher:
@eder-matheus I had a chat with @osamahammad21 yesterday and this PR comes with three DRCs which are not too challenging to solve. The chat was based on the results I got from my run from 4 days ago. 

I re-ran Secure-CI last night and the result is stable, we still see the same violations.
That said, this is good to merge.

PR #6621, opened 2025-01-30
odb/pdn: avoid double registration of identical connect rules

gadfort:
Changes:
- I noticed that repeated calls to `add_global_connection` with identical parameters results in multiple identical rules.
- This adds a check for existing rules and removes the duplicate (this should also allow calls to global_connect to process faster since the duplicate rules would just double the runtime).


maliberty (approved):
Is this happening in some PDK or just an observation?

gadfort:
@maliberty observation when I was debugging things

PR #6624, opened 2025-01-31
dbSta: fix modinst lookup

gadfort:
@maliberty this seems to be enough to resolve the issue I was having.
The issue was the list -> string conversion with escape characters.

### DIFF
--- a/src/dbSta/src/dbSta.tcl
+++ b/src/dbSta/src/dbSta.tcl
@@ -93,7 +93,7 @@ proc report_cell_usage { args } {
 
   set module [[ord::get_db_block] getTopModule]
   if { $args != "" } {
-    set modinst [[ord::get_db_block] findModInst $args]
+    set modinst [[ord::get_db_block] findModInst [lindex $args 0]]
     if { $modinst == "NULL" } {
       sta_error 1002 "Unable to find $args"
     }


src/dbSta/src/dbSta.tcl
```diff
@@ -93,7 +93,7 @@ proc report_cell_usage { args } {
 
   set module [[ord::get_db_block] getTopModule]
   if { $args != "" } {
-    set modinst [[ord::get_db_block] findModInst $args]
+    set modinst [[ord::get_db_block] findModInst [lindex $args 0]]
```
maliberty, line 96: The command doesn't take a list.  It might be surprising to someone who tries that to get only the first element.  Is an error better?

gadfort: we already check for this: `  check_argc_eq0or1 "report_cell_usage" $args` above (`args` must be length 0 or 1)

maliberty: Not really:
```
>>> report_cell_usage [list 1 2 3]
[ERROR STA-1002] Unable to find {1 2 3}
[ERROR GUI-0070] STA-1002
>>> report_cell_usage  1 2 3
[ERROR STA-0566] report_cell_usage requires zero or one positional arguments.
[ERROR GUI-0070] STA-0566
```

gadfort: in the first case, you are passing in an list, which gets converted to "{1 2 3}", not sure what we would check then, this is true for everything, if you pass in a list when a list is not expected it's wrong.

maliberty (approved):

PR #6632, opened 2025-02-03
mpl: fix truncated std::round calls

joaomai:
While writing the tests for MPL I found some std::round calls that were always being truncated, causing snapping to suboptimal positions.


AcKoucher (changes requested):
Good catch.

The code seems fine. The regression tests are failing probably due to subtle changes in the position. @joaomai please, take a look.

eder-matheus (approved):
@joaomai @AcKoucher I will approve the PR, and merge it after the unit test is fixed.

PR #6633, opened 2025-02-03
rsz: hold check dont touch instances

gadfort:
Adds:
- check for dont_touch on hold repair for instances to prevent resizer from attempting to change instance connections.


src/rsz/src/RepairHold.cc
```diff
@@ -440,7 +442,16 @@ void RepairHold::repairEndHold(Vertex* end_vertex,
               }
             }
           }
-          if (!load_pins.empty()) {
+          // check for dont touch loads
```
maliberty, line 445: Rather than prevent buffering if there is a dont-touch fanout you can just leave the fanouts attached and add the new net to the buffer's input.  I believe RepairDesign::makeRepeater does this and perhaps could be generalized?

gadfort:
replaced with #6660

PR #6642, opened 2025-02-05
drt: decoupling of the Graphics classes and stable drt_lib & graphic Factory

bnmfw:
Tackles [6532](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6532).

This is PR #6550 but better.

The branch is the same and every new addition can be found starting from the "graphic factory" commit.

### Graphics Factory
Instead of passing the three DR classes during the initialization of `TritonRoute`, now a new class, `GraphicsFactory` is responsible for providing both unique pointers and simple pointers for all the three classes. Whenever `initDesign` is called, `initGraphics` which passes the necessary arguments for these classes is called, ensuring they are up to date.


maliberty:
ready for review?

src/drt/src/TritonRoute.cpp
```diff
@@ -243,26 +232,22 @@ void TritonRoute::debugSingleWorker(const std::string& dumpDir,
...
                            std::ios::binary);
   std::string workerStr((std::istreambuf_iterator<char>(workerFile)),
                         std::istreambuf_iterator<char>());
   workerFile.close();
   auto worker = FlexDRWorker::load(
       workerStr, &viaData, design_.get(), logger_, router_cfg_.get());
-  worker->setGraphics(graphics.get());
+  AbstractDRGraphics* graphics
+      = debug_->debugDR ? graphics_factory_->getDRGraphics() : nullptr;
+  worker->setGraphics(graphics);
```
maliberty, line 250: Why not create and give ownership to the worker as before?

src/drt/src/MakeTritonRoute.cpp
```diff
@@ -50,11 +64,20 @@ void deleteTritonRoute(drt::TritonRoute* router)
...
 {
-  openroad->getTritonRoute()->init(openroad->tclInterp(),
-                                   openroad->getDb(),
-                                   openroad->getLogger(),
-                                   openroad->getDistributed(),
-                                   openroad->getSteinerTreeBuilder());
+  // Define swig TCL commands.
+  auto tcl_interp = openroad->tclInterp();
+  Drt_Init(tcl_interp);
+  sta::evalTclInit(tcl_interp, sta::drt_tcl_inits);
```
maliberty, line 70: Did this need to be moved?  Does the factory ctor depend on tcl somehow?

src/drt/src/TritonRoute.cpp
```diff
@@ -630,17 +620,22 @@ void TritonRoute::gr()
...
 {
-  FlexTA ta(getDesign(), logger_, router_cfg_.get(), distributed_);
-  ta.setDebug(debug_.get(), db_);
-  ta.main();
+  std::unique_ptr<FlexTA> ta = std::make_unique<FlexTA>(
+      getDesign(), logger_, router_cfg_.get(), distributed_);
+  if (debug_->debugTA) {
+    ta->setDebug(graphics_factory_->makeUniqueTAGraphics());
+  }
+  ta->main();
```
maliberty, line 628: What motivates the change from stack to unique_ptr?

maliberty (commented):
Partial review as I'd like to get clarity on the change in ownership semantics.

src/drt/src/MakeTritonRoute.cpp
bnmfw: This came before the factory, it was part of the old PR. tcl_interp being on TritonRoute.cpp creates linking issues. In most other tools on the project tcl_interp is always on the Make<toolName>.cpp file. This is both for keeping this patterns and to allow for the drt_lib

src/drt/src/TritonRoute.cpp
bnmfw: Kinda arbitrary tbh. PA will have to be a unique_ptr created at initialization at some point for the GRT with PA to be possible. I planned to make this with DR, TA and PA, started it but reverted for this PR. This change is residual, I can revert it if you want, but at some point I think it will become a unique_ptr anyway.

src/drt/src/AbstractGraphicsFactory.h
```diff
@@ -0,0 +1,64 @@
...
+  virtual void reset(frDebugSettings* settings,
+                     frDesign* design,
+                     odb::dbDatabase* db,
+                     Logger* logger,
+                     RouterConfiguration* router_cfg)
+      = 0;
+  virtual bool guiActive() = 0;
+  virtual AbstractDRGraphics* getDRGraphics() = 0;
+  virtual AbstractTAGraphics* getTAGraphics() = 0;
+  virtual AbstractPAGraphics* getPAGraphics() = 0;
```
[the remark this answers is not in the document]
bnmfw: Answered bellow

src/drt/src/TritonRoute.cpp
bnmfw: the worker actually did not get the ownership of the pointer before. It just got a raw pointer and ownership was kept to the "scope of the function". I made the `getXXGraphics` to accommodate for this. Besides that I made every method for all three tools for the sake of keeping the pattern, even though I do not use all.

maliberty: Now the ownership is not with the "scope of the function" but factory class itself.  What is the advantage of that?  I don't expect a factory to own its products.

bnmfw: fixed

maliberty (approved):

PR #6644, opened 2025-02-05
DependencyInstaller defaults to no option

vvbandeira:


docs/user/Build.md
```diff
@@ -74,53 +74,47 @@ it can be uploaded in the "Relevant log output" section of OpenROAD
...
 
 ### Compiling OpenROAD Manually
 
 ``` shell
-mkdir build && cd build
-cmake ..
-make
-sudo make install 
+./etc/Build.sh
+sudo make install
```
oharboe, line 91: I would leave out all instructions in the manual to install OpenROAD into /usr/local, this generally harms the system and should be discouraged.

Installing into /usr/local is really, to my mind, a use-case for Docker only.

etc/DependencyInstaller.sh
```diff
@@ -808,6 +814,11 @@ while [ "$#" -gt 0 ]; do
     shift 1
 done
 
+if [[ "${option}" == "none" ]]; then
+    echo "You must use one of: -all|-base|-common" >&2
+    _help
```
oharboe, line 819: nit: add exit 1

oharboe (changes requested):
delete all instructions in README.md to install into /usr/local, which mangles the system. It is to be discouraged and only done by those who really understands the ramifications.

To my mind installing in /usr/local is for the Docker use-case only.

docs/user/Build.md
vvbandeira: I will split this into a separate section since this `make install` only targets the `openroad` bin. Installing `openroad` at the system level has helped when running tutorials or allowing sysadmins to maintain one bin for all users of a shared machine.

maliberty: It is also helpful if you maintain many copies of OR and don't have to update the dependencies for each.  It can be a non-default option but it should be an option.

oharboe: Yep. No sucker punches, but if you ask for it explicitly, then you will get it and you deserve the result.

oharboe (changes requested):
I find the three -all, -common, -base, -prefix option names very confusing.

They appear to be something between an option and something mutually exclusive.

I dont have an immediate suggestion for alternative interface or names.

oharboe (approved):

oharboe:
A thought: most users should run the no options, safe, "sudo ./setup.sh".

DependencyInstaller.sh is an implementation detail for special cases.

If DependencyInstaller.sh is invoked without arguments, point the user to "you probably want sudo ./setup.sh" by printing a message.

At that point, it hardly matters that the arguments are obtuse as long as nothing bad happens upon sudo etc/DependencyInstaller.sh

vvbandeira:
I will add the suggestion of using `sudo setup.sh` in ORFS, but here, for OpenROAD, we do not have a `setup.sh` yet.

oharboe:
> I will add the suggestion of using `sudo setup.sh` in ORFS, but here, for OpenROAD, we do not have a `setup.sh` yet.

Add same message? "If you don't know exactly what you are doing, then you probably want to run sudo ORFS/setup.sh"

PR #6646, opened 2025-02-05
gpl: modify gpl debug mode

gudeh:
This PR introduces two additions to gpl debug mode, and does not modify any functionality.

- include circles on the tip of electro forces to better indicated their direction.
Previous version:
image
New version:
image

- include a new gpl debug mode heatmap for overflow based on min and max normalization, useful for checking the least and most overflowed bins. Previous existing heatmaps are maintained as is.
image


src/gpl/src/graphics.cpp
```diff
@@ -451,26 +466,55 @@ odb::Rect Graphics::getBounds() const
...
+          0.0f,
+          static_cast<float>(bin.instPlacedAreaUnscaled())
+              + static_cast<float>(bin.nonPlaceAreaUnscaled()) - scaledBinArea);
+      value = block->dbuAreaToMicrons(value);
+
+      min_value = std::min(min_value, value);
+      max_value = std::max(max_value, value);
+    }
+
+    logger_->report("Overflow min: {}, max: {}", min_value, max_value);
```
gadfort, line 490: I would suggest removing this report

eder-matheus:
@gudeh Could you share a figure of this new heatmap you added, comparing with the previous heatmaps?

eder-matheus:
@gudeh Thanks for the figures! That's a nice addition indeed.

gudeh:
image

PR #6657, opened 2025-02-06
etc: only download or-tools if needed

vvbandeira:


maliberty (approved):

etc/DependencyInstaller.sh
```diff
@@ -303,9 +303,19 @@ _installOrTools() {
...
     orToolsPath=${PREFIX:-"/opt/or-tools"}
+    orToolsVersion="none"
+    if [ -f ${orToolsPath}/Makefile ]; then
+        orToolsVersion=$(make -C ${orToolsPath} detect | grep VERSION | awk '{print $3}')
+    fi
+    CMAKE_PACKAGE_ROOT_ARGS+=" -D ortools_ROOT=$(realpath $orToolsPath) "
+    if [ ${orToolsVersion} == ${orToolsVersionSmall} ]; then
+        return
+    else
+        rm -rf ${orToolsPath}
```
gadfort, line 314: you cannot remove this path if it's prefix path. This script runs as sudo and if PREFIX is set to any system directory you will delete all files

PR #6658, opened 2025-02-06
drt: decoupling of the OR singleton from DRT core code

bnmfw:
Tackles #6532.

This PR is branched from #6642 and is waiting for it to merge. (I do not recommend reviewing this just yet)
A Secure CI is associated with it, once it resolves the result will be updated.

### DRT Core Dependencies on the ORD singleton
There are two main dependencies DRT has:
1. Getting the Number of Threads.
2. Making reads and writes to the ODB.

#### Number of Threads
This was simply solved by getting the number of threads in during the `initTritonRoute` execution in the `MakeTritonRoute` file, which is not part of the core DRT code and therefore can depend on the ODB. This information becomes an attribute to `initTritonRoute` which needs the value and is passed to `FlexDR` which also needs it.

#### Reads and Writes
This is controversial. The DRT makes 4 separate calls to the ODB, which was pointed to me as possibly being wrong, as the DRT normally should not have this kind of access to the ODB. The usual abstract interface strategy was used to solve this dependency. A separate `ORDBInterface` class was created to handle the communication with ODB. The DRT core depends on the Abstract version of this new class and the actual concrete implementation is passed in Run Time during `TritonRoute::init()`.

There might be a case for all these dependencies to be eliminated in a way that DRT loses access to writing and reading from the ODB.

### Next Step
This PR still does not solve the Issue associated. When trying to link the GRT to the DRT no more linking errors are raised due to DRT code, however, DRT_LIB depends on DST which, in turn, depends on the ORD singleton. My next PR associated with this issue will tackle the DST dependency on the ODB, probably making a separation in DST and DST_LIB in the same manner it was done to the DRT.


src/drt/src/MakeTritonRoute.cpp
```diff
@@ -50,11 +65,26 @@ void deleteTritonRoute(drt::TritonRoute* router)
...
+  std::unique_ptr<drt::ORDBInterface> or_db_interface
+      = std::make_unique<drt::ORDBInterface>();
+
+  router->init(openroad->getDb(),
+               openroad->getLogger(),
+               openroad->getDistributed(),
+               openroad->getSteinerTreeBuilder(),
+               std::move(graphics_factory),
+               std::move(or_db_interface),
+               ord::OpenRoad::openRoad()->getThreadCount());
```
gadfort, line 87: this may be modified by the user after initialization, but before calling for routing. This cannot be set here.

maliberty:
@bnmfw I thought we merged a PR for this.  Is this stale?

bnmfw:
@maliberty I think you are referring to #6642. They both tackle the same issue (this closes it), the first solves coupling to the GUI and this solves coupling to the ORD singleton. If follows a similar decoupling pattern as the aforementioned PR, but it solves something different. I also wanted to make this a separate PR due to the debate that could arise from the Reads and Writes part.

maliberty:
I think the read/write_(lef/def/db) could all be moved to odb.  The only tricky case seems to be readDb's callback to sta but that could be handled by an observer.  If you make that change in a separate PR it will simplify this one.

Is just inconvenient to plumb the thread count to where it is needed?  If so the router config avoid that though I would tend to just pass it through in a more explicit style.

bnmfw:
@QuantamHD Thank you very much!

bnmfw:
@maliberty This is ready for a new review. I have, however three points I'm not 100% on:

1. The three uses of `router_->getRouterConfiguration()->MAX_THREADS` on `RoutingCallBack.h`. I'm not 100% sure if the thread count will be updated when the callbacks run as I'm not fully sure when they will run.
2. I had the python problem we discussed briefly. One of the python test now needs to send the thread count as a parameter. I defaulted it to 1 for now, but it needs to be fixed to actually retrieve the thread count.
3. The solution for the readDb does not seem to be the best one. The call actually does not need to interface with STA, as it does not sets "hierarchical" as true, so I could just transport the code to the DRT. I left the observer callbacks out, however, as it seem to not impact the regression tests I ran. I'd like to discuss any better solutions as I don't like very much how that part of the code turned out. I'm also running the ISPD and Secure CIs again to see if anything change, once they pass I will reopen the PR.

maliberty:
I fixed the python thread count issue.

maliberty:
Concerns 1&3 only occur in the distributed routing flow which is not in production so I'm less concerned.  I think the read_db call only happens in the worker where we are not using timing and so not concerned about hierarchy.

PR #6659, opened 2025-02-06
tap: expose 1 site gap in place_tapcells

gadfort:
This fixes the tapcell issue I mentioned


maliberty:
I think we should move in the opposite direction to making this more automatic rather than more manual.

gadfort:
@maliberty I agree, but this is just exposing the flag that was missing before. If the automatic detection is going to take a while before someone can work on it and implement it, then I would suggest we merge this and then when its ready we can deprecate the issue (this is not the same issue as the antenna repair one).

src/tap/test/disallow_one_site_gaps2.tcl
```diff
@@ -0,0 +1,15 @@
...
+set def_file [make_result_file disallow_one_site_gaps2.def]
+
+place_endcaps -endcap_vertical "TAPCELL_X1"
+place_tapcells -distance 20 \
+    -master "TAPCELL_X1" \
+    -disallow_one_site_gaps
+
+write_def $def_file
+
+diff_file disallow_one_site_gaps.defok $def_file
```
maliberty, line 15: Why does this use the defok from another test?

maliberty:
Rather than expose the flag can you use the API in https://github.com/The-OpenROAD-Project/OpenROAD/pull/6678 to set it internally?

gadfort:
> Rather than expose the flag can you use the API in #6678 to set it internally?

Done, probably needs a full run on your side

src/tap/test/disallow_one_site_gaps2.tcl
gadfort: since these two tests need to generate the same result, so I just used this file again, if you prefer to split defok, I can do that

maliberty:
The code looks fine - I've started a full run

maliberty:
Fixed a bug and restarted CI

maliberty (approved):

rovinski:
@maliberty I have one small nit that I think would be worth addressing. I have seen cell libraries with x1 fillers, but you aren't allowed to use them unless it's used to fill an x1 gap between two logic cells.

My proposal is to also check if the master is marked as `dont_use` in `hasOneSiteMaster()`. A user can then mark the filler as `dont_use` to still enable the x1 gap avoidance.

maliberty:
That sounds reasonable to me  @gadfort any opinion?

gadfort:
@maliberty sure, but this will require moving the dont_use flag from resizer to ODB (probably the right place for it anyways)

maliberty:
I think that makes sense too

PR #6661, opened 2025-02-07
cts: check if clock edge is nullptr

gadfort:
This might indicate an issue with the design, but it looks like we need to check the return to avoid segfaulting


maliberty (approved):

maliberty:
It would be interesting to understand what triggers this condition and flag it more clearly.

PR #6662, opened 2025-02-07
rsz: if no layers or orientation are specifeid consider bot H & V resistance for clk wires

arthurjolo:
As mentioned in [#6651]( https://github.com/The-OpenROAD-Project/OpenROAD/discussions/6651#discussioncomment-12083425), the resistance for clk wires when there is no layer or orientation provided always uses horizontal resistance. It should consider also the vertical resistance, like it is done for clk wire capacitance.


maliberty:
The code change is fine but its worth a secure CI

arthurjolo:
CI is passing, only the calibre run was aborted but I believe it was the same errors we've been having for a while.

maliberty (approved):

PR #6663, opened 2025-02-07
meta: Fixes memory leaks in odb and dbsta

QuantamHD:


maliberty:
Ready for review?  If so, @eder-matheus please review

eder-matheus:
@QuantamHD We have failures in the unit tests, but something is off. When running the tests with the `./regression` script of each tool, the tests fail. But when running the specific test using the openroad binary, like `openroad constant1.tcl`, the test works properly.

QuantamHD:
@eder-matheus The memory management was super weird, and tbh I really don't know how it was working before. I removed dbReadVerilog in favor of giving the OpenROAD object the sole lifetime control of the VerilogReader. I think the code is better now, and it should be leak free.

eder-matheus (approved):

eder-matheus:
> @eder-matheus The memory management was super weird, and tbh I really don't know how it was working before. I removed dbReadVerilog in favor of giving the OpenROAD object the sole lifetime control of the VerilogReader. I think the code is better now, and it should be leak free.

@QuantamHD thanks for this update. It's really much better now.

PR #6666, opened 2025-02-07
rsz/odb: check for mergeNet success before removing buffer

gadfort:
Fixes remove_buffers failing with dont_touch set

Changes:
- mergeNet checks if iterm insts are marked dontTouch and return before making modifications to the design if any are set.
- removeBuffers will not remove a buffer is the net fails to be merged

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -2867,7 +2867,7 @@ class dbNet : public dbObject
   ///
   /// Merge the iterms and bterms of the in_net with this net
   ///
-  void mergeNet(dbNet* in_net);
+  bool mergeNet(dbNet* in_net);
 
   dbSet<dbGuide> getGuides() const;
 
--- a/src/odb/src/db/dbNet.cpp
+++ b/src/odb/src/db/dbNet.cpp
@@ -3238,21 +3238,29 @@ dbNet* dbNet::getValidNet(dbBlock* block_, uint dbid_)
   return (dbNet*) block->_net_tbl->getPtr(dbid_);
 }
 
-void dbNet::mergeNet(dbNet* in_net)
+bool dbNet::mergeNet(dbNet* in_net)
 {
   _dbNet* net = (_dbNet*) this;
   _dbBlock* block = (_dbBlock*) net->getOwner();
-  for (auto callback : block->_callbacks) {
-    callback->inDbNetPreMerge(this, in_net);
+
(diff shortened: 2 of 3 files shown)


src/odb/include/odb/db.h
```diff
@@ -2867,7 +2867,7 @@ class dbNet : public dbObject
   ///
   /// Merge the iterms and bterms of the in_net with this net
   ///
-  void mergeNet(dbNet* in_net);
+  bool mergeNet(dbNet* in_net);
```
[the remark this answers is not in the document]
gadfort: unrelated

src/odb/src/db/dbNet.cpp
```diff
@@ -3238,21 +3238,29 @@ dbNet* dbNet::getValidNet(dbBlock* block_, uint dbid_)
...
+      return false;
+    }
+  }
+
+  for (auto callback : block->_callbacks) {
+    callback->inDbNetPreMerge(this, in_net);
   }
 
   for (dbITerm* iterm : iterms) {
+    iterm->disconnect();
```
maliberty, line 3263: not necessary as connect auto-disconnects

maliberty (approved):

src/odb/src/db/dbNet.cpp
gadfort: both iterm and bterm just call `disconnect()`:
```
  if (iterm->_mnet != 0) {
    disconnect();
  }
```
the mergeNet code is the same with:
```
disconnect()
connect(this)
```
or
```
connect(this)
```

andyfox-rushc: Correct.  But note that connect is subtly killing the hierarchical net (which won't show up in your test) and might be a source of bugs. 

Perhaps it is better to use your original pattern (1) disconnect then (2) connect. 

That way it is clear in the code that the hierarchical net needs to be saved and reconnected. So the pattern becomes:

dbModNet* hier_net = iterm -> getModNet()...
iterm -> disconnect() <-- explicity disconnect the pin (nothing on iterm)
iterm -> connect(new_db_net) <-- explicitly connect the flat net
-- if desired to preserve the hierarchical net (usually we move them around).
iterm -> connect(hier_net)...

maliberty: Disconnect needs to handle hierarchy - it doesn't matter whether it is called explicitly or implicitly.

PR #6667, opened 2025-02-07
Remove odb's isSequential for DFT code

fgaray:
I am also removing one_cell_nangate45.tcl test as it seems that it was not working as the golden logs show that no scan replacement was performed.


maliberty (approved):

maliberty:
Is the fact that the test wasn't working not indicative of a problem?

PR #6672, opened 2025-02-08
Fix incorrect and ambiguous help strings

nmoroze:
This PR updates some help strings that are incorrect or ambiguous. These were caught by some old work I was doing on parsing OpenROAD help strings.

`clock_tree_synthesis`: a bunch of the switches were missing arguments, making them look like flags
- `detailed_route_debug`: putting braces around list items is more consistent with e.g. `initialize_floorplan`, and makes it explicit that these items need to be part of a list (rather than 4 separate words)
- `repair_antennas`: the `diode_cell` argument is optional
- `place_pad`: this one feels the least important since I think most people could infer this, but putting the positional argument makes it extra clear that `-mirror` doesn't accept a value "name". I think this is also consistent with how some other help strings are defined.

### DIFF
--- a/src/grt/src/GlobalRouter.tcl
+++ b/src/grt/src/GlobalRouter.tcl
@@ -255,7 +255,7 @@ proc global_route { args } {
   }
 }
 
-sta::define_cmd_args "repair_antennas" { diode_cell \
+sta::define_cmd_args "repair_antennas" { [diode_cell] \
                                          [-iterations iterations] \
                                          [-ratio_margin ratio_margin]}
 
--- a/src/pad/src/pad.tcl
+++ b/src/pad/src/pad.tcl
@@ -230,11 +230,11 @@ proc place_corners { args } {
   pad::place_corner $master $index
 }
 
-sta::define_cmd_args "place_pad" {[-master master] \
+sta::define_cmd_args "place_pad" {name \
+                                  [-master master] \
                                   -row row_name \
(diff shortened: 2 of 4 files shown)


src/grt/src/GlobalRouter.tcl
```diff
@@ -255,7 +255,7 @@ proc global_route { args } {
   }
 }
 
-sta::define_cmd_args "repair_antennas" { diode_cell \
+sta::define_cmd_args "repair_antennas" { [diode_cell] \
                                          [-iterations iterations] \
                                          [-ratio_margin ratio_margin]}
```
maliberty, line 260: As diode_cell is a positional argument it should come after the flags

src/pad/src/pad.tcl
```diff
@@ -230,11 +230,11 @@ proc place_corners { args } {
   pad::place_corner $master $index
 }
 
-sta::define_cmd_args "place_pad" {[-master master] \
+sta::define_cmd_args "place_pad" {name \
```
maliberty, line 233: name is positional and should remain at the end

maliberty (changes requested):

src/grt/src/GlobalRouter.tcl
nmoroze: Got it, didn't realize that's the convention - updated it.

PR #6688, opened 2025-02-11
GRT: Pad pins on left and bottom borders are unreachable

osamahammad21:


oharboe (commented):
isnt this broken for lower left case?

osamahammad21:
> isnt this broken for lower left case?

It was, but now it is just going to assume that the edge_cap is 0 for that case.

oharboe:
I was concerned about the case where pin_x == 0 && pin_y == 0

osamahammad21:
> I was concerned about the case where pin_x == 0 && pin_y == 0

In that case, the edge_cap would be 0 no matter which layer the pin is on. The whole point of this PR is to avoid passing a negative value to getEdgeCapacity function.

eder-matheus (approved):

PR #6689, opened 2025-02-11
mpl: fix fixed terminals for cluster of unplaced IOs and max cost

### LINKED ISSUE #6668: mpl: the distance to boundary is wrongly computed from level != 0 inside SA
### Description

There's an offset compensation missing. This compensation happens "automatically" for all fixed terminals, but, for the unconstrained ios cluster we compute the distance to the boundaries based on the shape of die - which needs to be relocated according to the current parent offset.

### Suggested Solution

Apply the correct compensation.

### Additional Context

_No response_

AcKoucher:
Fix #6668

### What is Wrong
In the context of the WL computation inside SA for when the target cluster is a cluster of unplaced IOs:
1. The distance to boundary is being wrongly computed when the cluster of unplaced IOs is the _the cluster of unconstrained pins_, because we're using the shape of its Cluster object which does not take into account the offset w.r.t. the current outline.
2. The max/limit cost which should be the die area HPWL is also being wrongly computed, because we always use the shape of the target io cluster that may not be the die area. _The only situation in which that is true is if the cluster of unplaced IOs represents pins without constraints._

### Context and Changes
Originally, a fixed terminal in HierRTLMP was just a Soft/Hard Macro with the shape of a point (area = 0).
With the new IOs' abstraction implemented in #5809, for clusters of unplaced IOs, we need to have a Soft/Hard Macro object with a proper shape inside SA in order to compute the distance to the closest boundary correctly.

As this needs to happen for both cluster placement (Soft) and macro placement (Hard) I made the constructors used to create fixed terminals in both Hard/Soft Macro have the same arguments to be able to template the function that uses these constructors.

I.e., this PR ensures that:
- Inside SACore, for a WL computation in which one of the objects is a cluster of unplaced IOs, the Cluster object is used to identify to which boundary the cluster of unplaced IOs is constrained to.
- We now cache the dimensions of the die area inside the PhysicalHierarchy struct during the ClusteringEngine initialization. This cache is passed on to the SA Core, which during its initialization will offset it to the current outline.  The cache will be, then, used to compute the dist to the closest boundary in the case of the cluster of unconstrained IOs.
- The cached die area shape will also be used to compute the max distance to generate the max cost.


AcKoucher:
Running Secure-CI

maliberty (approved):

maliberty:
Code is ok; waiting on CI to merge

AcKoucher:
The failure in the regression test seems to be related to a bug that is not related to the changes here. However, it needs fixing before this PR gets merged. I'm working on it.

AcKoucher:
@maliberty I found another bug in the same scope of the changes here. I pushed the fix and updated the description.

AcKoucher:
I pushed one more change that will help me rebasing my branch with the second phase of the IO constraints projects with the changes here.

AcKoucher:
I pushed new master changes to fix the regression tests results.

Running new Secure-CI now that #6697 was merged.

AcKoucher:
Re-running Secure-CI.

PR #6690, opened 2025-02-11
dpl: remove use of disallow_one_site_gaps

gadfort:
Changes:
- removes use of -disallow_one_site_gaps from DPL in favor of database implementation
- corrects segfault on macros without a site, probably should be excluding pads and macros from this check to avoid a pad filler allowing 1 site gaps when the standard cells dont


maliberty:
I see we fixed the same bug; I did in your tap PR.

src/dpo/test/gcd_no_one_site_gaps.ok
```diff
@@ -1,8 +1,9 @@
-[INFO ODB-0227] LEF file: Nangate45/Nangate45.lef, created 22 layers, 27 vias, 135 library cells
+[INFO ODB-0227] LEF file: Nangate45_data/Nangate45.lef, created 22 layers, 27 vias, 135 library cells
 [INFO ODB-0128] Design: gcd
 [INFO ODB-0130]     Created 54 pins.
 [INFO ODB-0131]     Created 549 components and 2166 component-terminals.
 [INFO ODB-0133]     Created 364 nets and 1068 connections.
+[WARNING DPO-0003] -disallow_one_site_gaps is deprecated
```
maliberty, line 6: Update the test to remove the deprecated flag

src/dpo/test/gcd_no_one_site_gaps.ok
```diff
@@ -58,7 +64,8 @@ Importing netlist into detailed improver.
 Detailed Improvement Results
 ------------------------------------------
 Original HPWL             7709.2 u
-Final HPWL                7326.7 u
-Delta HPWL                  -5.0 %
+Final HPWL                7257.2 u
+Delta HPWL                  -5.9 %
 
+[WARNING DPL-0004] -disallow_one_site_gaps is deprecated
```
maliberty, line 70: Update the test to remove the deprecated flag

maliberty (changes requested):

maliberty:
I agree on the pads.  Do you want to add it to this PR?

gadfort:
@maliberty added the checks to this PR and merged with master, so this should be good to check now.

src/dpo/test/Nangate45_data/Nangate45.lef
maliberty, line 1: Please use a symlink rather than copying

src/odb/src/zutil/util.cpp
```diff
@@ -279,8 +279,29 @@ bool hasOneSiteMaster(dbDatabase* db)
...
+        continue;
+      }
+
+      // Ignore IO corner cells
+      dbMasterType type = master->getType();
+      if (type == dbMasterType::ENDCAP_TOPLEFT
+          || type == dbMasterType::ENDCAP_TOPRIGHT
+          || type == dbMasterType::ENDCAP_BOTTOMLEFT
+          || type == dbMasterType::ENDCAP_BOTTOMRIGHT) {
+        continue;
```
maliberty, line 292: Use `mater->getType().isEndCap()` (there are quite a few more types for endcaps)

maliberty (changes requested):

src/odb/src/zutil/util.cpp
gadfort: only the 4 listed refer to PAD corners, the remainder refer to standard cell end caps. It's a dumb LEF naming thing, but isEndCap captures both standard cells (which we want to check) and IO corners (which we dont)

src/dpo/test/Nangate45_data/Nangate45.lef
gadfort: symlink from where? This is not a copy but a copy+modification.

maliberty: In that case it is confusingly named.  Something should indicate the modifications.

gadfort: I will add some comments to the top of the file to clarify

maliberty (approved):

eder-matheus:
@maliberty @gadfort Our private CI is broke after this update. All designs are failing in DPL with the following error:
```
[WARNING DPL-0007] One site gap check failed (16).
...
...
[ERROR DPL-0033] detailed placement checks failed.
```

maliberty:
We are going to have to revert this while we figure out what's wrong.  It appears that we didn't properly use the flag before so now we are activating this code automatically and seeing a change.

gadfort:
@maliberty I mentioned to @eder-matheus I found before that DPO does some things that result in an illegal placement, to solve it I just follow it with a call to detail placement (I know not great). Maybe it's possible to just revert out the changes to DPO? since that seems to be where the issue is currently present. Unless you found a different issue as well.

maliberty:
We weren't passing -disallow to either dpl or dpo in ORFS!  For now I want to take both out and then we can look at a partial restore of dpl.  I'm not sure why this didn't produce gap problems before, perhaps just luck?

gadfort:
Ahh, I see, yes, that is an oversight

eder-matheus:
@maliberty @gadfort After some investigation, I found that DPO is the  source of the problems. Let me give some context:

* I'm running my tests in a branch where I restored this PR, so the auto detection of disallow_one_site_gaps is on.
* After the `detailed_placement` call, `check_placement` don't report violations. So the DPL side is fine.
* However, after running `improve_placement -max_displacement`, `check_placement` reports one site gap check fails.
* Looking at the violations in ODB, we don't have one-site gaps, but three-site gaps the `check_placement` considers as violations. Check the figure below:
image
* This also happens in our master branch! If I change the ORFS scripts to use the `-disallow_one_site_gaps` flag in both `detail_placement` and `improve_placement` calls, the results are exactly the same from my tests with this PR restored. However, we also need to call `check_placement` with the disallow flag.
* What triggered this error in this PR is that now `check_placement` automatically detects when it's needed to check for one-site gaps.

I believe the problem is that DPL has a different understanding from DPO of what is a one site gap violation. What tool should be fixed? I believe DPL is doing the right thing here, but I just want to confirm.

gadfort:
@eder-matheus why would a 3 site gap be considered a 1 site violation?

eder-matheus:
> @eder-matheus why would a 3 site gap be considered a 1 site violation?

I believe it's caused by the `set_placement_padding` call. We add a PAD of 1 in this design in the beginning of the `detail_place.tcl` script, and it is applied to the `check_placement` at the end of this script. Setting the PAD to zero before checking the placement results in no errors.

gadfort:
It seems to me that `check_placement` should not be using the padding during the 1 site gap checks.

eder-matheus:
> It seems to me that `check_placement` should not be using the padding during the 1 site gap checks.

I've created this PR doing this: https://github.com/The-OpenROAD-Project/OpenROAD/pull/6718.

I also started a secure-ci to ensure it will work for every design.

maliberty:
I agree with @gadfort that check_placement should not consider padding.  You can fill a 3-site gap.

PR #6691, opened 2025-02-11
docs: clone submodules to generate sta docs

vvbandeira:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2818

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,6 +4,10 @@
 # Required
 version: 2
 
+submodules:
+    include:
+      - src/sta
+
 # Set the OS, Python version and other tools you might need
 build:
   os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


maliberty (approved):

docs/conf.py
```diff
@@ -163,6 +163,8 @@ def swap_prefix(file, old, new):
 def setup(app):
     import os
 
+    os.popen("git submodule update --init --recursive")
```
oharboe, line 166: ouch! This will run asynchronously? I think the concern of setting up the source folder should be kept out of the Sphinx configuration file. Do it before running sphinx.

oharboe (changes requested):
Keep the concern of updating git and configuring sphinx separate.

oharboe (commented):
No experience with these .yaml files, but glad to see .git operations are not taking place in sphinx conf.py

PR #6692, opened 2025-02-11
ppl: save constraints in odb at each call of set_io_pin_constraint command

eder-matheus:


oharboe (changes requested):
What are you asking the user to do here and what happens if they don't heed this advice?

I think an error message is better...

"It is **strongly recommended** to set the mirrored pins before any other constraint, in order to have them properly stored into the database. "

gadfort:
@eder-matheus wouldn't it be better to process the mirror constraints first and then then remainder if order matters?

eder-matheus:
@gadfort @oharboe I've updated the code to not depend on the order we call the commands. It should be able to properly handle multiple mirrored pins and region constraint calls, independently of what is called first.

Some unit tests were added to depict some different scenarios, and the result is the expected in all of them.

oharboe:
> @gadfort @oharboe I've updated the code to not depend on the order we call the commands. It should be able to properly handle multiple mirrored pins and region constraint calls, independently of what is called first.
> 
> Some unit tests were added to depict some different scenarios, and the result is the expected in all of them.

This sounds great! Sounds like I don't have to think about this anymore :-)

It is easier to write code and automated tests than documentation that builds a strong mental model with the user so that they can follow an exact recipe.

eder-matheus:

maliberty:
To be replaced

PR #6699, opened 2025-02-12
rsz: ensure hold verbose actually prints progress

gadfort:
Changes:
- in the current implementation the hold fixing only printed the first and final result, but no progress.
- This ensures that the process is printed as expected.


maliberty (approved):
Thanks!

PR #6702, opened 2025-02-12
rsz: add support for verbose in recover power

gadfort:
Adds:
- support for -verbose to -recover_power
- a single test that checks -recover_power (this functionality has no unit tests that I can find)

Output:
```
Iteration |  Resized |   WNS    | Endpt
---------------------------------------
        0 |        0 |    1.198 | r2/D
    final |        2 |    0.616 | r7/D
---------------------------------------
[INFO RSZ-0141] Resized 2 instances.
```


src/rsz/src/RecoverPower.cc
```diff
@@ -468,4 +487,36 @@ int RecoverPower::fanout(Vertex* vertex)
   return fanout;
 }
 
+void RecoverPower::printProgress(int iteration, bool force, bool end) const
+{
+  const bool start = iteration == 0;
+
+  if (start && !end) {
+    logger_->report("Iteration |  Resized |   WNS    | Endpt");
```
maliberty, line 495: I think it would be nice to have area percentage here as well (presumably negative).

gadfort: easy enough to add

maliberty (approved):

rovinski:
What does it mean to have -129.6% area?

PR #6708, opened 2025-02-13
rsz: makeRepeater check IO dont touch

gadfort:
Fixes:
- previously, the dont touches would result in floating nets and double driven nets
- this avoid this by rejecting the repeater when its not possible to add it and cleanup afterwards if the buffer ended up being left floating due to dont touches.


andyfox-rushc:
I respectfully suggest checking the load_net driver pin(s) and load pins for sanity and rejecting the makeRepeater move on entry to make Repeater if it is something we cannot handle. The tricky part is defining the "axiom" for what makes a net we cannot rebuffer. The pr author has identified that the driving pin being connected to a dont touch primary input as being one such case. There are likely others. Therefore having a "legallityCheck" on the load_net and the load_pins input to the function seems to me to be a good idea so that part of the code is nicely isolated:
 
To help:
 
 
To see how to get the load_net driver starting from the load_net use: db_network -> getNetDriverParentModule. (as a side effect this returns the driver pin through its second argument)
 
Iterate through the load_pins and get the load instances.
 
Then do some case analysis to see if makeRepeater makes sense eg the following cases might not. Perhaps the following might apply:
 
1. The load_net is don't touch. (Interestingly I don't see much code that seems to check for dont touch nets)
2. The load net is hooked to a primary i/o that is don't touch.
3. All the load pins and the driver pin are don't touch.
4. There are multiple primary input drivers on the net (some kind of wired or or tri-state).
Of course 1-4 above might be junk and I am likely wrong but it seems best to me to sanity check the incoming data and not modifying the circuit rather than selectively rejecting the move after the event.
 
If the above strategy seems ok I am happy to provide the code to do it and then the "sanity check" routine can simply be rehacked as we find more amusing and interesting cases where makeRepeater is to be throttled.
 
andy

src/rsz/src/RepairDesign.cc
```diff
@@ -1850,6 +1854,30 @@ void RepairDesign::makeRepeater(
     }
   }
 
+  if (!hasInputPort(load_net) && preserve_outputs) {
```
maliberty, line 1857: If `hasInputPort(load_net) && preserve_outputs` is true then it seems you can't insert a buffer.

src/rsz/src/RepairDesign.cc
```diff
@@ -2019,6 +2051,27 @@ void RepairDesign::makeRepeater(
     sta_->connectPin(buffer, buffer_output_port, buffer_op_net);
   }  // case 2
 
+  if (!connections_modified) {
```
maliberty, line 2054: When would this happen?  It seems both cases modify connections.

src/rsz/src/RepairDesign.cc
gadfort: this is just the `not(hasInputPort(load_net) || !preserve_outputs)` from below since this is checking a specific case, I'll redo that logic to make it clear

src/rsz/src/RepairDesign.cc
maliberty: You can still insert a repeater in that case unless the driver is also dont-touch or a primary input

gadfort: you cannot, because you cannot change the net that the dont_touch buffers are attached to (thats what is failing). If you cannot break the net into two parts because all the instances on one or the other side of it are marked dont_touch then adding the buffer just results in a floating net somewhere.

maliberty: You just leave the loads on their current net and add a new net at the repeater input instead.

maliberty: In any case it is better to avoid this upstream impossible combinations

gadfort: I mode the check up to earlier, so the original code is not really modified.

maliberty (approved):

PR #6713, opened 2025-02-13
rsz: add command to report nets driven by multiple drivers

gadfort:
When debugging the dont_touch issues, I noticed several instances were resizer ended up with multiple drivers on a net (fixed by https://github.com/The-OpenROAD-Project/OpenROAD/pull/6708) but it didn't appear that we had any way of reporting these types of issues.

This adds:
- `report_overdriven_nets` (a mirror to `report_floating_nets`) which report nets with multiple drivers accounting for nets with tristate drivers.

### DIFF
--- a/src/rsz/README.md
+++ b/src/rsz/README.md
@@ -401,8 +401,16 @@ The `report_overdriven_nets` command reports nets with connected by multiple dri
 
 ```tcl
 report_overdriven_nets
+    [-include_parallel_driven]
     [-verbose]
 ```
+
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-include_parallel_driven` | Include nets that are driven by multiple parallel drivers. |
+| `-verbose` | Print the net names. |
 
 ### Eliminate Dead Logic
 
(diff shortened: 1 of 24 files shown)


gadfort:
@maliberty any feedback on this?

maliberty:
Another case to exclude is when you have drivers in parallel (usually inverters or buffers).

gadfort:
@maliberty that would be the cases I would want to know about, more than 1 driver on a net. It's a reporting command, so I would think you would want to know that, it doesn't generate an error, but simply lets you know.

maliberty:
What is the concern about a parallel driver?

gadfort:
are you suggesting checking the inputs of the drivers to ensure they are all identical (of course that will require the cells to be equivalent as well)?

maliberty:
I think it would be sufficient to cover inverters and buffers as it is less common for larger gates.

gadfort:
@maliberty I added the code to exclude the parallel driven with buffers or inverters, and I added a flag to report them anyways if requested.

maliberty (approved):

PR #6720, opened 2025-02-15
flake.nix: replace deprecated url literal with a string literal

pr2502:
URL Literals are legacy syntax which has been long disallowed in NixOS/nixpkgs, since Lix release 2.92 they have been deprecated in Lix. See <https://lix.systems/blog/2025-01-18-lix-2.92-release/>. Using the normal string syntax makes this flake compatible with both Nix implementations without any special flags.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -13,7 +13,7 @@
 
 {
   inputs = {
-    nixpkgs.url = github:nixos/nixpkgs/nixos-24.05;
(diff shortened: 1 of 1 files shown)


maliberty (approved):

maliberty:
Thanks Max

donn:
Thanks for the heads up @pr2502

PR #6721, opened 2025-02-16
drt: addUniqueInst

bnmfw:
Supports #5867.

This PR is ISPD and Secure CI safe.
I recommend looking at the commits separately while reviewing, specially the dead code ones. 

Like many other things in PA, for incremental PA to be possible, unique instances need to be a live structure, after being properly initialized, it needs to be possible to add new instances to the unique instances structures dynamically, this PR adds said support.

### Live `pref_track_patterns_` and `master_to_pin_layer_range_`
Both these variables became member member variables of `UniqueInsts`. This is necessary because during incremental PA these structures will need to be constantly consulted and updated, creating the necessity of making them more permanent. This change made the functions that create them, `computePrefTrackPatterns()` (name changed) and `initMasterToPinLayerRange()` to not return anything, since they now simply initialize the member variables. This change made the already small `initUniqueInstance()` obsolete, as the passing around of the older variables is not necessary anymore, so it was deleted, and the initialization of the variables was passed to `computeUnique()`.

### `addUniqueInst()`
This function will be used during incremental PA, it essentially refactors the bulk of the old `computeUnique()` code to receive a single `frInst*` pointer and initialize the unique instance data for that instance. Some of the code had to be refactored to accommodate for this change, specially what comes after the "// Fills data structure that relate a instance to its unique instance" comment.

### Dead Code elimination
This is a weird one, something I did not caught when refactoring PA, I recommend looking at both the dead code commits to see exactly what I'm talking about. In two separate parts of the code there was this patterns where a initial data structure was created with the target insts and masters of the run (`target_frinsts` and `masters`). Soon after, during iteration of masters and insts of the design there existed a continue guard. This guard would activate if the original data structure had something in it and the inst/master of the loop was present in it. The structure was always empty, so this guard never activated, and the creation of the original structures had no actual purpose, so I deleted them.


osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -136,92 +125,63 @@ bool UniqueInsts::isNDRInst(frInst& inst)
...
+void UniqueInsts::addUniqueInst(frInst* inst)
 {
-  std::set<frInst*> target_frinsts;
-  for (auto inst : target_insts_) {
-    target_frinsts.insert(design_->getTopBlock()->findInst(inst->getName()));
+  if (!router_cfg_->AUTO_TAPER_NDR_NETS && isNDRInst(*inst)) {
+    unique_.push_back(inst);
+    inst_to_unique_[inst] = inst;
+    inst_to_class_[inst] = nullptr;
   }
```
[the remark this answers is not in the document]
bnmfw: Fixed

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -136,92 +132,72 @@ bool UniqueInsts::isNDRInst(frInst& inst)
...
```
osamahammad21, line 141: @bnmfw You're not setting the unique_to_idx_ for ndr instances.

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_unique.cpp
bnmfw: great catch, fixed.

osamahammad21 (approved):

osamahammad21:
@bnmfw Make sure to rerun the ISPD CI and the secure CI. There is a major change in DRT happened yesterday.

bnmfw:
@osamahammad21 Both pass

PR #6731, opened 2025-02-18
drt: deleteUniqueInst

bnmfw:
Supports #5867.

This is ISPD and Secure CI safe. Although `deleteInst()` has not use on this PR, it is tested by, after creating all Unique Instances in `computeUnique()`, deleting them all and adding them again.

### `deleteInst()`
This function is responsible for deleting an unique inst and relocating all the unique insts data structures. Now that there needs to exist support for deleting unique insts, something that was not predicted by the original code, there might be a question to wether or not the current structures are the best. I've invited discussion about this structures, so here is a detailed explanation of how work, currently:

#### What is an Unique Instance?
Two instances of the same cell (same master) that have the same orientation and have the same offset from their preferred track pattern will have the same access pattern solution. This means that it is only necessary to solve it for one of those instances, and use that solution for the other, we say that they are the same **unique instance**.

A group of all instances that fall under the same unique instance is called an **unique class**. Currently, we elect the first instance of its class to represent the unique instance, this is referred as the **head** on the new code.

Here is a list of all current structures that manage unique instances:

- **unique_:** a simple vector of every unique instance, thus, a vector of the head of every unique class.
- **unique_to_idx_:** a map that inverts the above relation, maps an unique instance to its position on the unique_ vector.
- **inst_to_class_:** maps an inst to its class (the class is implemented as a set).
- **inst_to_unique_:** maps an inst to its unique inst (class head).
- **unique_to_pa_idx_**: maps an unique inst to the number of pin access it has.

#### Problems with the current structures:
Now that deleting is supported a few problem arise.

##### Deleting the last instance of its class:
This means that the class itself has to be deleted, implying that `unique_` will have a deletion (O(|uniques|) and that `unique_to_idx_` has to change to accommodate for that (O(|uniques|). There might be a better way to make this reciprocal data structures that are so unfriendly to deletion.

##### Deleting the head of a class:
This means that a new head has to be elected, and that every value of `inst_to_unique_` of instance on that class have to be adjusting (O(|class|)).

I intend to change most of the unique structures to be oriented by the class instead of by one instance representative. This will be a future PR, and be done at a cleanup step after incremental PA is mostly functional.


src/drt/src/pa/FlexPA_unique.h
```diff
@@ -157,7 +152,7 @@ class UniqueInsts
 
   // All the unique instances
   std::vector<frInst*> unique_;
-  // Mapp all instances to their representative unique instance
+  // Mapa all instances to their representative unique instance
```
rovinski, line 155:
```suggestion
  // Maps all instances to their representative unique instance
```

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -159,17 +160,26 @@ void UniqueInsts::addUniqueInst(frInst* inst)
...
+  if (unique_class.empty()) {
     int i = unique_.size();
     unique_.push_back(inst);
     unique_to_idx_[inst] = i;
+    unique_inst = inst;
+    new_unique_class = true;
+  } else {
+    // guarantees everyone on the family has the same unique_inst (the first
+    // that came)
+    unique_inst = inst_to_unique_[*unique_class.begin()];
```
LucasYuki, line 178: Is **unique_**, **unique_to_idx_**, **inst_to_unique_**, and **inst_to_unique_** necessary?

If unique inst (class head) is always unique_class.begin(), can't the others be easily inferred using the **inst_to_class_** map?

bnmfw: I do think using actual instances to represent classes to be weird. This is something I did not caught when refactoring PA some time ago. There is actually some functions, specially in flex_pa_acc_pattern.cpp that have both the unique instance itself, as well as their idx on the vector as a parameter which is redundant at best and harmfully confusing at worst. I'm thinking of revamping the whole unique_inst structures to use the class itself to represent unique instances as I do think it is more intuitive and less error prone.

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -287,6 +297,56 @@ bool UniqueInsts::hasUnique(frInst* inst) const
...
+      if (unique_to_idx_[unique_inst] >= i) {
+        unique_to_idx_[unique_inst]--;
+      }
+    }
+    unique_.erase(it);
+    unique_to_idx_.erase(inst);
+    unique_to_pa_idx_.erase(inst);
+    class_head = nullptr;
+
+  } else if (inst == inst_to_unique_[inst] && unique_class.size() > 1) {
```
maliberty, line 327: `unique_class.size() > 1` seems redundant (we don't expect it to be empty right?).

bnmfw: True

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -71,6 +71,24 @@ class UniqueInsts
   frInst* getUnique(int idx) const;
   bool hasUnique(frInst* inst) const;
 
+  /**
+   * @brief Adds the instance to the unique instances structures,
+   * inserting new data if it is actually a new unique instance.
+   *
+   * @returns if this instance if the first of its unique class.
```
osamahammad21, line 78:
```suggestion
   * @returns True if the instance is the first of its unique class.
```

src/drt/src/pa/FlexPA_unique.h
```diff
@@ -71,6 +71,24 @@ class UniqueInsts
...
+
+  /**
+   * @brief deletes an inst from the unique insts structures
+   *
+   * @param inst instance to be deleted
+   *
+   * @returns the unique inst that represents the unique class. If the class was
+   * deleted returns nullptr
+   */
+  frInst* deleteInst(frInst* inst);
```
osamahammad21, line 90: I have no problem with returning the pointer to the head. I just don't think it would ever be useful to the caller.

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -167,17 +168,26 @@ void UniqueInsts::addUniqueInst(frInst* inst)
...
   // Fills data structure that relate a instance to its unique instance
-  UniqueInsts::InstSet& unique_family
+  UniqueInsts::InstSet& unique_class
       = master_orient_trackoffset_to_insts_[inst->getMaster()][orient][offset];
-  if (unique_family.empty()) {
+  inst_to_class_[inst] = &unique_class;
+
+  frInst* unique_inst = nullptr;
+  bool new_unique_class = false;
+  if (unique_class.empty()) {
```
osamahammad21, line 177:
```suggestion
  bool new_unique_class = unique_class.empty();
  if (new_unique_class) {
```
Just a suggestion for readability. Feel free to ignore.

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -167,17 +168,26 @@ void UniqueInsts::addUniqueInst(frInst* inst)
...
     unique_to_idx_[inst] = i;
+    unique_inst = inst;
+    new_unique_class = true;
+  } else {
+    // guarantees everyone on the class has the same unique_inst (the first
+    // that came)
+    unique_inst = inst_to_unique_[*unique_class.begin()];
   }
-  unique_family.insert(inst);
-  frInst* unique_inst = *(unique_family.begin());
```
osamahammad21, line 178: How did that use to work before? the unique_family set ordering changes with each new addition. I think we got lucky because we were passing the instances in order so the IDs were always incremental. Good catch.

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -287,6 +297,56 @@ bool UniqueInsts::hasUnique(frInst* inst) const
...
+  frInst* class_head = inst;
+  if (unique_class.size() == 1) {
+    auto it = std::find(unique_.begin(), unique_.end(), inst);
+    if (it == unique_.end()) {
+      logger_->error(DRT,
+                     25,
+                     "{} not found on unique insts, although being the only "
+                     "one of its unique class");
+    }
+    int i = (int) std::distance(unique_.begin(), it);
```
osamahammad21, line 314: use unique_to_idx_ directly instead to get the index in the unique_ vector.

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -287,6 +297,56 @@ bool UniqueInsts::hasUnique(frInst* inst) const
...
+
+    // readjusts unique_to_idx_ to compensate for posterior unique deletion
+    for (frInst* unique_inst : unique_) {
+      if (unique_to_idx_[unique_inst] >= i) {
+        unique_to_idx_[unique_inst]--;
+      }
+    }
```
osamahammad21, line 321:
```suggestion
    for (i; i < unique_.size(); i++) {
      auto unique_inst = unique_[i];
      unique_to_idx_[unique_inst]--;
    }
```

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -287,6 +297,56 @@ bool UniqueInsts::hasUnique(frInst* inst) const
...
+    unique_to_idx_.erase(inst);
+    unique_to_pa_idx_.erase(inst);
+    class_head = nullptr;
+
+  } else if (inst == inst_to_unique_[inst]) {
+    // the inst does not belong to the class anymore, but is the reference
+    // unique_inst, so the reference has to be another inst
+    auto class_begin = inst_to_class_[inst]->begin();
+    // new class head
+    class_head = *class_begin != inst ? *class_begin : *(++class_begin);
```
osamahammad21, line 332: why recheck? you already know that the current inst is the class head.

src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -287,6 +297,56 @@ bool UniqueInsts::hasUnique(frInst* inst) const
...
+    unique_to_idx_.erase(inst);
+    if (unique_to_pa_idx_.find(inst) != unique_to_pa_idx_.end()) {
+      unique_to_pa_idx_[class_head] = unique_to_pa_idx_[inst];
+      unique_to_pa_idx_.erase(inst);
+    }
+  }
+  inst_to_class_[inst]->erase(inst);
+  inst_to_class_.erase(inst);
+  inst_to_unique_.erase(inst);
+  return class_head;
```
osamahammad21, line 347: class_head would be nullptr if the current instance being deleted is not the current class head. I don't think this is intended, right?

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_unique.h
bnmfw: Currently, I'm trying to avoid refactoring the structures, as I think making incremental PA run a priority. Some of the structures that manage unique insts exists on the `UniqueInst` class, others in `FlexPA`. Ideally, I think it would be better if they all existed under management of a single class, but that is beyond the point.

`deleteInst()` returning the new class head helps me, because I'm calling this function from `FlexPA` and need to also update the `FlexPA` strucutres (such as `unique_inst_patterns_`). It helps me to know it the deletion destroyed the class itself, or if it changed the class head, which inst is the new head.

src/drt/src/pa/FlexPA_unique.cpp
bnmfw: Yeah, the incremental way we iterate over the insts is what held things together. Unfortunately now we need support for doing it in a random order, I just forgot to check for that in the [addInst PR](https://github.com/The-OpenROAD-Project/OpenROAD/pull/6721) which was not a problem, since we still only inset it in order anyway.

src/drt/src/pa/FlexPA_unique.cpp
bnmfw: Banger. I tunneled into avoiding using the classic `for(int i =0; i < size(); ++i)` and did not see this way more elegant solution.

src/drt/src/pa/FlexPA_unique.cpp
bnmfw: Due to the support for inserting instances into the structure in a random order, the class head is not necessarily the first instance of the structure (although this is true after the first run of the PA). After some steps of incremental PA with things being moved and the class head might be anywhere in the structure. It is possible for it to not be on the 1st or 2nd place, this check just guarantees that the new candidate is not it.

src/drt/src/pa/FlexPA_unique.cpp
bnmfw: True, good catch.

maliberty:
What is the status of this PR?  Is it ready for Osama to re-review?

osamahammad21 (approved):

PR #6732, opened 2025-02-18
rsz: modify default verbose behavior of repair design and repair timing

gudeh:
always print first and last iteration, only print middle ones when verbose is active


precisionmoon (approved):
This looks fine.  Thanks.

maliberty:
Some tests need fixing

gudeh:
In ORFS, we opted to keep the default verbosity of RSZ repair design and repair timing to true (as in flow/scripts/util.tcl). However, if an ORFS user sets the verbosity to false, they will now see messages for the first and last iterations.

PR #6734, opened 2025-02-19
dbSta: Add snapshot creation to report_cell_usage

bangqixu:
Add snapshot creation functionality to report_cell_usage. This will only be active if -file flag is specified (and optionally a name can be given to the stage at which the snapshot is created)

### DIFF
--- a/src/dbSta/include/db_sta/dbSta.hh
+++ b/src/dbSta/include/db_sta/dbSta.hh
@@ -36,7 +36,6 @@
 #pragma once
 
 #include <memory>
-#include <optional>
 #include <string>
 #include <vector>
 
@@ -197,21 +196,6 @@ class dbSta : public Sta, public ord::OpenRoadObserver
                          bool verbose,
                          const char *file_name,
                          const char *stage_name);
-
-  // Holds the usage information of a specific cell which includes (i) name of
-  // the cell, (ii) number of instances of the cell, and (iii) area of the cell
-  // in microns^2.
-  struct CellUsageInfo {
-    std::string name;
-    int count = 0;
-    double area = 0.0;
-  };
-
-  // Holds a snapshot of cell usage information at a given stage.
-  struct CellUsageSnapshot {
-    std::string stage;
-    std::vector<CellUsageInfo> cells_usage_info;
-  };
 
   BufferUse getBufferUse(sta::LibertyCell* buffer);
 
--- a/src/dbSta/include/db_sta/dbSta.hh
+++ b/src/dbSta/include/db_sta/dbSta.hh
@@ -36,6 +36,9 @@
 #pragma once
 
 #include <memory>
+#include <optional>
+#include <string>
+#include <vector>
 
 #include "odb/db.h"
 #include "odb/dbBlockCallBackObj.h"
@@ -190,7 +193,25 @@ class dbSta : public Sta, public ord::OpenRoadObserver
 
   std::string getInstanceTypeText(InstType type) const;
   InstType getInstanceType(odb::dbInst* inst);
-  void report_cell_usage(odb::dbModule* module, bool verbose);
+  void report_cell_usage(odb::dbModule* module,
+                         bool verbose,
+                         const char *file_name,
+                         const char *stage_name);
+
+  // Holds the usage information of a specific cell which includes (i) name of
+  // the cell, (ii) number of instances of the cell, and (iii) area of the cell
+  // in microns^2.
+  struct CellUsageInfo {
+    std::string name;
+    int count = 0;
+    double area = 0.0;
+  };
+
+  // Holds a snapshot of cell usage information at a given stage.
+  struct CellUsageSnapshot {
+    std::string stage;
+    std::vector<CellUsageInfo> cells_usage_info;
+  };
 
   BufferUse getBufferUse(sta::LibertyCell* buffer);
 
--- a/src/dbSta/include/db_sta/dbSta.hh
+++ b/src/dbSta/include/db_sta/dbSta.hh
@@ -192,10 +192,8 @@ class dbSta : public Sta, public ord::OpenRoadObserver
 
   std::string getInstanceTypeText(InstType type) const;
   InstType getInstanceType(odb::dbInst* inst);
-  void report_cell_usage(odb::dbModule* module,
-                         bool verbose,
-                         const char *file_name,
-                         const char *stage_name);
+  void report_cell_usage(odb::dbModule* module, bool verbose,
+                         const char* file_name, const char* stage_name);
 
   BufferUse getBufferUse(sta::LibertyCell* buffer);
 
--- a/src/dbSta/include/db_sta/dbSta.hh
+++ b/src/dbSta/include/db_sta/dbSta.hh
@@ -192,8 +192,10 @@ class dbSta : public Sta, public ord::OpenRoadObserver
 
   std::string getInstanceTypeText(InstType type) const;
   InstType getInstanceType(odb::dbInst* inst);
-  void report_cell_usage(odb::dbModule* module, bool verbose,
-                         const char* file_name, const char* stage_name);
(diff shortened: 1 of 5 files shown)


src/dbSta/src/dbSta.cc
```diff
@@ -603,6 +645,39 @@ void dbSta::report_cell_usage(odb::dbModule* module, const bool verbose)
           format, master->getName(), stats.count, stats.area / area_to_microns);
     }
   }
+
+  std::string file(file_name);
+  if (!file.empty()) {
+    std::map<std::string, CellUsageInfo> name_to_cell_usage_info;
+    dbBlock* block = db_->getChip()->getBlock();
+    const std::vector<dbInst*> insts = module->getLeafInsts();
+    const double dbu_to_microns = std::pow(block->getDbUnitsPerMicron(), 2);
```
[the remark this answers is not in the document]
bangqixu: Done. Thanks for pointing this out!

src/dbSta/src/dbSta.cc
```diff
@@ -603,6 +645,39 @@ void dbSta::report_cell_usage(odb::dbModule* module, const bool verbose)
...
+        name_to_cell_usage_info.size());
+    for (const auto& [cell_name, cell_usage_info] : name_to_cell_usage_info) {
+      cell_usage_snapshot.cells_usage_info.push_back(cell_usage_info);
+    }
+    boost::json::value output = boost::json::value_from(cell_usage_snapshot);
+
+    std::ofstream snapshot;
+    snapshot.open(file);
+    snapshot << output.as_object();
+    snapshot.close();
```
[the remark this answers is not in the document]
bangqixu: Done. Thanks for pointing this out!

src/dbSta/src/dbSta.cc
```diff
@@ -603,6 +645,39 @@ void dbSta::report_cell_usage(odb::dbModule* module, const bool verbose)
...
+      const std::string& cell_name = inst->getMaster()->getName();
+      auto [it, inserted] = name_to_cell_usage_info.insert(
+          {cell_name, CellUsageInfo{
+                          .name = cell_name,
+                          .count = 1,
+                          .area = inst->getMaster()->getArea() / dbu_to_microns,
+                      }});
+      if (!inserted) {
+        it->second.count++;
+      }
```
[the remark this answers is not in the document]
bangqixu: These meet the minimal requirement for our use case. Since the struct can be easily extended, we can add other information needed at a later time when we have a use case.

bangqixu:
Thanks for the review! I have addressed the comment in a new commit. PTAL.

src/dbSta/include/db_sta/dbSta.hh
```diff
@@ -190,7 +193,25 @@ class dbSta : public Sta, public ord::OpenRoadObserver
...
+                         const char *file_name,
+                         const char *stage_name);
+
+  // Holds the usage information of a specific cell which includes (i) name of
+  // the cell, (ii) number of instances of the cell, and (iii) area of the cell
+  // in microns^2.
+  struct CellUsageInfo {
+    std::string name;
+    int count = 0;
+    double area = 0.0;
```
maliberty, line 207: Why not just store dbMaster rather than it name and area?

src/dbSta/src/dbSta.cc
```diff
@@ -603,6 +652,43 @@ void dbSta::report_cell_usage(odb::dbModule* module, const bool verbose)
...
+      const std::string& cell_name = inst->getMaster()->getName();
+      auto [it, inserted] = name_to_cell_usage_info.insert(
+          {cell_name,
+           CellUsageInfo{
+               .name = cell_name,
+               .count = 1,
+               .area = inst->getMaster()->getArea() / area_to_microns,
+           }});
+      if (!inserted) {
+        it->second.count++;
```
maliberty, line 669: This is a near duplicate of the code right above it.  Please sharing code rather than repeating it.

src/dbSta/src/dbSta.cc
```diff
@@ -80,6 +87,45 @@ dbSta* makeDbSta()
...
+                json::value& json_value,
+                ord::dbSta::CellUsageSnapshot const& cell_usage_snapshot)
+{
+  json_value
+      = {{"stage", cell_usage_snapshot.stage},
+         {"cell_usage_info",
+          boost::json::value_from(cell_usage_snapshot.cells_usage_info)}};
+}
+
+ord::dbSta::CellUsageInfo tag_invoke(value_to_tag<ord::dbSta::CellUsageInfo>,
```
maliberty, line 110: Is this necessary in order to write the data?  It seems to be parsing json rather than writing it.

src/dbSta/src/dbSta.cc
```diff
@@ -80,6 +87,45 @@ dbSta* makeDbSta()
 }
 }  // namespace ord
 
+namespace boost::json {
```
maliberty, line 90: Is it expected to add to a boost namespace?  It seems unusual.

src/dbSta/src/dbSta.tcl
```diff
@@ -79,10 +79,11 @@ proc highlight_path { args } {
   }
 }
 
-define_cmd_args "report_cell_usage" {[-verbose] [module_inst]}
+define_cmd_args "report_cell_usage" { \
+  [-verbose] [module_inst] [-file file] [-stage stage]}
```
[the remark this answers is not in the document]
bangqixu: Please give more specific instructions. Is it for README, or code, or something else.

src/dbSta/include/db_sta/dbSta.hh
```diff
@@ -190,7 +193,25 @@ class dbSta : public Sta, public ord::OpenRoadObserver
...
+  };
+
+  // Holds a snapshot of cell usage information at a given stage.
+  struct CellUsageSnapshot {
```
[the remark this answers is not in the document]
bangqixu: I am following exactly what "TypeStats" is defined and used. Why was it OK to do the same in https://github.com/The-OpenROAD-Project/OpenROAD/commit/69e83c424aa196cc92b4de09849ad19b9e158ada ?

src/dbSta/src/dbSta.cc
bangqixu: Yes.

src/dbSta/src/dbSta.cc
bangqixu: Other than what is being looped through, they look quite different, right? IMHO it would hurt readability if we merge these together. To put it another way, if the code snippets above these is put into a function, wouldn't "ReportCellUsage(); CreateCellUsageSnapshot()" be cleaner than "ReportCellUsageAndCreateSnapshot()"?

src/dbSta/src/dbSta.cc
bangqixu: Based on my research, this is the way to go. If you know alternatives, please advice.

src/dbSta/include/db_sta/dbSta.hh
bangqixu: Because we are writing using boost::json and I am not aware of a way to do the translation from dbMaster there? This is tied with the translation part below. I can only figure out how to translate from struct of kinda primitive type to JSON.

src/dbSta/src/dbSta.tcl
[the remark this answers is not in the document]
bangqixu: Did what you asked.

src/dbSta/include/db_sta/dbSta.hh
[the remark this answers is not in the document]
bangqixu: Did what you asked.

bangqixu:
PTAL.

maliberty (approved):

bangqixu:
Thank you both for the review! (And thanks Matt for the cleanup and adding a test!)

PR #6738, opened 2025-02-19
MPL: support for multi pattern snapping

### LINKED ISSUE #6360: MPL2: Add support for multi-pattern track-grid snap
### Description

After #6355 if we have multiple track patterns (e.g., m2 in asap7) we just use the first track pattern.

### Suggested Solution

Consider snapping to any of the patterns when dealing with a multi-pattern track.

### Additional Context

_No response_

joaomai:
Closes #6360.

This PR refactors the data structures used in snapping to enable snapping to patterns instead of snapping to layers. For most designs, it works in the same way.

The snapping logic is the same still, snapping to layers with multiple patterns simply assigns a pin to each pattern and tries to snap to the pattern. The pin is assigned by searching for a pair of pins that follow the pattern in question, as snapping to one of them means all pins in the same pattern are snapped.


src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3640,17 +3695,18 @@ int Snapper::getPinWidth(odb::dbITerm* pin,
   return pin_width;
 }
 
-void Snapper::getTrackGrid(odb::dbTrackGrid* track_grid,
-                           int& origin,
-                           int& step,
-                           const odb::dbTechLayerDir& target_direction)
+void Snapper::getTrackGridPattern(odb::dbTrackGrid* track_grid,
+                                  int grid_pattern,
```
AcKoucher, line 3699: const

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3582,38 +3597,77 @@ SameDirectionLayersData Snapper::computeSameDirectionLayersData(
   return data;
 }
 
-LayerParameters Snapper::computeLayerParameters(
-    odb::dbTechLayer* layer,
-    odb::dbITerm* pin,
+odb::dbITerm* Snapper::findPinForMultiPattern(
+    odb::dbTrackGrid* track_grid,
+    int grid_pattern,
```
AcKoucher, line 3602: const
grid_pattern ->pattern_idx

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3549,31 +3545,50 @@ void Snapper::setOrigin(const int origin,
...
+      }
+
+      seen.insert(layer);
+
+      odb::dbTrackGrid* track_grid = block->findTrackGrid(layer);
+      if (track_grid == nullptr) {
+        logger_->error(
+            MPL, 39, "No track-grid found for layer {}", layer->getName());
+      }
+      int grid_patterns = target_direction == odb::dbTechLayerDir::VERTICAL
```
AcKoucher, line 3575: const and I think it's better making explicit in the name that this is the number of patterns and not the patterns themselves: patterns_count or number_of_patterns, etc.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3640,17 +3695,18 @@ int Snapper::getPinWidth(odb::dbITerm* pin,
...
-                           int& origin,
-                           int& step,
-                           const odb::dbTechLayerDir& target_direction)
+void Snapper::getTrackGridPattern(odb::dbTrackGrid* track_grid,
+                                  int grid_pattern,
+                                  int& origin,
+                                  int& step,
+                                  const odb::dbTechLayerDir& target_direction)
 {
   // TODO: handle multiple patterns
```
AcKoucher, line 3704: Isn't this being addressed in this PR?

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3549,31 +3545,50 @@ void Snapper::setOrigin(const int origin,
   }
 }
 
-SameDirectionLayersData Snapper::computeSameDirectionLayersData(
+std::vector<PatternParameters> Snapper::computeSameDirectionPatternsData(
     const odb::dbTechLayerDir& target_direction)
 {
-  SameDirectionLayersData data;
+  std::vector<PatternParameters> data;
+  std::set<odb::dbTechLayer*> seen;
```
AcKoucher, line 3552: seen -> visited_layers

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3549,31 +3545,50 @@ void Snapper::setOrigin(const int origin,
...
+                              ? track_grid->getNumGridPatternsX()
+                              : track_grid->getNumGridPatternsY();
 
-          if (data.layer_to_pin.empty()) {
-            data.snap_layer = layer;
+      if (grid_patterns == 1) {
+        data.push_back(
+            computePatternParameters(track_grid, 0, iterm, target_direction));
+      } else {
+        for (int pattern = 0; pattern < grid_patterns; pattern++) {
```
AcKoucher, line 3583: pattern -> pattern_idx

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3549,31 +3545,50 @@ void Snapper::setOrigin(const int origin,
...
     }
 
     for (odb::dbMPin* mpin : iterm->getMTerm()->getMPins()) {
-      for (odb::dbBox* box : mpin->getGeometry()) {
-        odb::dbTechLayer* layer = box->getTechLayer();
-        if (layer->getDirection() == target_direction) {
-          if (data.layer_to_pin.find(layer) != data.layer_to_pin.end()) {
-            continue;
-          }
+      odb::dbTechLayer* layer = (*mpin->getGeometry().begin())->getTechLayer();
```
AcKoucher, line 3562: This can be a getLayer(dbMPin*) function.

AcKoucher (changes requested):
I'm curious to see the results with the multi pattern feature in PPL.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3522,31 +3518,52 @@ void Snapper::setOrigin(const int origin,
...
```
AcKoucher, line 3535: You can use the new function here.

AcKoucher (changes requested):

AcKoucher:
Is this ready for review?

joaomai:
There are some public designs failing that I'm still investigating.

AcKoucher:
I see you're still pushing changes. Is this ready for review?

AcKoucher (changes requested):
I'm curious about the changes in the regression tests results.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3267,183 +3269,198 @@ void Snapper::setOrigin(const int origin,
...
+            MPL, 39, "No track-grid found for layer {}", layer->getName());
       }
+
+      data[track_grid].push_back(iterm);
     }
   }
 
-  return data;
-}
+  auto cmp_pins = [&](odb::dbITerm* pin1, odb::dbITerm* pin2) {
```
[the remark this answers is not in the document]
maliberty: I disagree.  The lambda itself has no name and what you are storing it into is a variable and should be named as such.  The example in https://google.github.io/styleguide/cppguide.html#Formatting_Lambda_Expressions uses 
```
auto x_plus_n = [&x](int n) -> int { return x + n; }
```

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3220,41 +3220,43 @@ void Snapper::snapMacro()
...
-    // their center.
-    origin = std::round(origin / static_cast<double>(snap_layer_params.pitch))
-                 * snap_layer_params.pitch
-             + snap_layer_params.offset - snap_layer_params.pin_offset;
+  int lowest_pin_center_pos
+      = origin + getPinOffset(lowest_grid_pin, target_direction);
 
-    alignWithManufacturingGrid(origin);
-    setOrigin(origin, target_direction);
+  auto closest_pos = lower_bound(lowest_grid_positions.begin(),
```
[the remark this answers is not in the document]
joaomai: I'm not sure making the type explicit here helps with readability, the type being std::vector<int>::const_iterator. It's used in the following line only to calculate the variable that will actually be used.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3267,183 +3269,198 @@ void Snapper::setOrigin(const int origin,
   }
 }
 
-SameDirectionLayersData Snapper::computeSameDirectionLayersData(
+Snapper::SnapData Snapper::computeSameDirectionPatternsData(
     const odb::dbTechLayerDir& target_direction)
 {
-  SameDirectionLayersData data;
+  std::map<odb::dbTrackGrid*, std::vector<odb::dbITerm*>> data;
```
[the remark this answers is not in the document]
joaomai: Do you think that simply making the name more clear would help? I skipped making an alias for it since it's an auxiliary data structure used only inside this one function.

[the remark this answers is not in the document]
AcKoucher: I don't see a problem in creating the alias. I think that only changing the name will cause the formatter to move the name to the next line which could be a bit odd. Maybe I'm wrong.

src/mpl/src/hier_rtlmp.cpp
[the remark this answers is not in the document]
AcKoucher: I didn't realize that. Thanks!

src/mpl/src/hier_rtlmp.cpp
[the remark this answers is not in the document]
joaomai: I'm pretty sure it will move it to next line and make it look all weird. I'll add an alias then.

src/mpl/src/hier_rtlmp.cpp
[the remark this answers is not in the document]
AcKoucher: I agree. `auto` is definitely ok here.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3267,183 +3269,198 @@ void Snapper::setOrigin(const int origin,
...
+            MPL, 39, "No track-grid found for layer {}", layer->getName());
       }
+
+      track_grid_to_pin_list[track_grid].push_back(iterm);
     }
   }
 
-  return data;
-}
+  auto cmp_pins_center = [&](odb::dbITerm* pin1, odb::dbITerm* pin2) {
```
AcKoucher, line 3301: If we're going with this naming style then don't abbreviate, `cmp` can be either compare or compute which makes it a bit confusing.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3267,183 +3269,198 @@ void Snapper::setOrigin(const int origin,
...
-LayerParameters Snapper::computeLayerParameters(
-    odb::dbTechLayer* layer,
-    odb::dbITerm* pin,
-    const odb::dbTechLayerDir& target_direction)
-{
-  LayerParameters params;
-
-  odb::dbBlock* block = inst_->getBlock();
-  odb::dbTrackGrid* track_grid = block->findTrackGrid(layer);
+  auto cmp_layer_number = [](LayerData data1, LayerData data2) {
```
AcKoucher, line 3307: ditto.

Please, move this to after the loop.

AcKoucher (changes requested):
Just a nit.

AcKoucher (approved):
LGTM.

maliberty (approved):

joaomai:
Good to merge.

PR #6741, opened 2025-02-20
utl: Adds live prometheus monitoring to OpenROAD

QuantamHD:
OpenROAD now includes a metrics endpoint server that can track internal tool metrics over time.

To use this feature you need to do the following start the prometheus and grafana collectors

image

[Detailed instructions](/etc/monitoring/README.md):
```shell
$ cd etc/monitoring
$ docker compose up -d
```

This will start a grafana endpoint ready to collect from the OpenROAD application you would like to track. By default it's looking for an http server running on port 8080 on your localhost.

To start the metrics endpoint in OpenROAD, run:
```tcl
utl::startPrometheusEndpoint 8080
```

This is all configurable in the docker compose file, and you should be able to access grafana by going to http://localhost:3000 username: admin, password: grafana. Go to the dashboard tab and click service, then OpenROAD to see the pre-made dashboard.


maliberty:
This appears to have a copy of https://github.com/jupp0r/prometheus-cpp.  It should be in third-party.

QuantamHD:
This is a proper fork. I think it should live in utl

maliberty:
Is there a reason we want a fork?  Will we never want upstream updates?

QuantamHD:
The library hasn't been updated in two years and is arguably complete. At the end of the day it's essentially just a map with some keys and values and a way to serialize it into a text file.

I have more ideas on how to drive it forward that are good for OpenROAD, but arguably not in the interest of the original library. Like adding more pluggable inheritance and a way to stub in Google's monitoring backend Monarch.

QuantamHD:
Also this is the original implementation not the one you linked to. https://github.com/biaks/prometheus-cpp-lite

maliberty:
The one I pointed probably started as a copy but has more recent commits and many more stars and forks.  It seems like the original died but this one is still alive fwiw.

QuantamHD:
That's the official [otel](https://opentelemetry.io/) one. We just don't really need what it provides, namely a bunch of http server stuff, and unnecessary performance cruft. It's better to just have our own fork that we have full control over. Performance stuff matters in webservers where it could conceivably add to response latency, but we're talking 100s of milliseconds at worst, and when you're talking about OpenROAD we're talking hours of runtime against negligible metrics cost.

src/utl/README.md
```diff
@@ -36,10 +62,10 @@ man
...
 #### Options
 
-| Switch Name | Description | 
-| ----- | ----- |
-| `name` | Name of the command/message to query. |
-| `-manpath` | Include optional path to man pages (e.g. ~/OpenROAD/docs/cat). |
+| Switch Name | Description                                                                                                                          |
+| ----------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| `name`      | Name of the command/message to query.                                                                                                |
+| `-manpath`  | Include optional path to man pages (e.g. ~/OpenROAD/docs/cat).                                                                       |
```
maliberty, line 68: Why is this reformatted with extraneous whitespace?

src/utl/include/utl/Logger.h
```diff
@@ -59,6 +59,9 @@
 
 namespace utl {
 
+class PrometheusMetricsServer;
+class Registry;
```
maliberty, line 63: Registry is a very generic name.  I would prefer to see the code the prometheus subdir in namespace like utl::prometheus

src/utl/include/utl/prometheus/metrics_server.h
```diff
@@ -0,0 +1,58 @@
...
+#include "text_serializer.h"
+
+namespace utl {
+class Logger;
+}
+
+namespace utl {
+class PrometheusMetricsServer
+{
+  std::thread worker_thread_;
```
maliberty, line 27: The private section goes last https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/utl/include/utl/prometheus/metrics_server.h
```diff
@@ -0,0 +1,58 @@
...
+  std::atomic<utl::Logger*> logger_;
+  bool shutdown_ = false;
+  bool is_ready_ = false;
+
+  void RunServer();
+  void WorkerFunction();
+
+ public:
+  ~PrometheusMetricsServer();
+  PrometheusMetricsServer(std::shared_ptr<Registry>& registry_,
```
maliberty, line 39: Don't inline; ctor before dtor https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/utl/src/prometheus/metrics_server.cpp
```diff
@@ -0,0 +1,156 @@
...
+      boost::asio::ip::tcp::endpoint endpoint(
+          boost::asio::ip::address::from_string("127.0.0.1"), port_);
+      socket.connect(endpoint);         // This will unblock the accept().
+    } catch (const std::exception& e) { /*Do nothing, we're dying*/
+    }
+  }
+  worker_thread_.join();
+}
+
+std::string SnapshotPrometheusMetrics(Registry* registry)
```
maliberty, line 42: local functions should be in an anonymous namespace https://google.github.io/styleguide/cppguide.html#Internal_Linkage

src/utl/include/utl/Logger.h
QuantamHD: I think that's too much name spacing for this code, especially if it's going to be called in a lot of places. I renamed it to PrometheusRegistry, everything else feels uniquely named enough to live in utl.

QuantamHD:
> It appears the Prometheus just samples and doesn't necessarily record all changes. So if you have a gauge that is Set() many times, only some values may be Collect()'ed. This feel like a poor fit for investigating the tools behavior as you get non-deterministic sub-sampling (or over-sampling) of the interesting state and little insight into the transitions.

That's configurable on the prometheus side, and also you can write a custom collector that tracks each call if you would like. I think the initial focus should be looking at patterns and long term trends rather that super high frequency details. I suspect there's lot to harvest there.

maliberty:
We can go with this and see if you suspicions are correct.  Family, Counter, and Builder still feel overly generic.

maliberty (approved):

PR #6743, opened 2025-02-20
gui: fall back to type name hashing when not using libstdc++

donn:
There is currently an issue where clicking on any sta::Instance on macOS causes a crash. This is owing to an oddball RTTI issue that I've spent some time debugging but was ultimately unable to resolve.

My hypothesis as to why this issue does not surface on GNU/Linux is because of the specific type_info implementation in the GNU C++ standard library, libstdc++. This email chain from the libc message list may be insightful: https://lists.llvm.org/pipermail/llvm-dev/2014-June/073465.html

As a workaround, I've elected to use a compiler macro and only use type_index hashes if `__GLIBCXX__` is [EDIT: defined], otherwise, the mangled ABI names are used as the key for descriptors instead. With this fix, clicking on instances works on macOS. This, unfortunately, will incur a performance penalty, but for what it's worth it is quite literally unnoticeable (and certainly beats a crash.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,8 @@ venv/
 include/ord/Version.hh
 
 build
+build_linux
+build_mac
 test/results
 flow
 
--- a/src/gui/include/gui/gui.h
+++ b/src/gui/include/gui/gui.h
@@ -804,8 +804,13 @@ class Gui
   odb::dbDatabase* db_;
 
   // Maps types to descriptors
+#ifdef __GLIBCXX__
   std::unordered_map<std::type_index, std::unique_ptr<const Descriptor>>
       descriptors_;
+#else
+  std::unordered_map<std::string, std::unique_ptr<const Descriptor>>
+      descriptors_;
+#endif
   // Heatmaps
   std::set<HeatMapDataSource*> heat_maps_;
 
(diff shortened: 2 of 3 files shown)


gadfort:
@maliberty if the performance impact is not noticeable, would it make to just switch to using the name instead?

maliberty:
I'm not concerned about performance but from https://en.cppreference.com/w/cpp/types/type_info/name
```
No guarantees are given; in particular, the returned string can be identical for several types
```
so this is not a safe-solution.  We could localize the ifdef to a single function rather than putting it at all uses.

donn:
> No guarantees are given; in particular, the returned string can be identical for several types

Practically, it will only return the same name if there are two classes with an identical namespace and name that are only prevented from collision by hidden visibility. 

> We could localize the ifdef to a single function rather than putting it at all uses.

I suppose that would have to do with overriding the hash function on macOS to use the name?

maliberty:
> I suppose that would have to do with overriding the hash function on macOS to use the name?

I was thinking you would put the ifdef inside the custom hash function.  A comment explaining why it is needed would be helpful

donn:
@maliberty Done.

maliberty (approved):

maliberty:
Thanks @donn for the help

PR #6747, opened 2025-02-21
mpl: clean up / improvements of debug messages & some fixes

AcKoucher:
This will help a lot to see the results for each outline of cluster placement, especially when we need to know what happened at the level of each individual cost inside the annealer.

#### Colateral Changes / Fixes
1. Wrap utilities in mpl namespace;
2. Adapt debug docs.
3. Compute area penalty in base class to avoid duplicated code and allow the debug report approach used in this PR.

### Changes
1. Remove unused debugPrint messages
2. Report summary for each cluster/macro placement in table format.


AcKoucher:
Good to merge.

maliberty (approved):

Issue #6745, opened 2025-02-21
gpl: bug after non-virtual iteration with specific design on specific commit


gudeh:
### Description

All buffers inserted by rsz in the last non-virtual iteration of gpl timing driven get stuck on the origin (bottom left corner) of the die, later being placed by dpl on the corner of the core. Happens only with a private design with specific commit: 7add98675f24375541fc47d9386195af9621d078

after global placement, stage 3-3: 
![Image](https://github.com/user-attachments/assets/4642d041-22f7-428c-bf00-997850a92c96)

after detail place 3-5:
![Image](https://github.com/user-attachments/assets/d95e313a-95ca-421b-8be5-dadb6a9f4c71)

gudeh:
fixed with https://github.com/The-OpenROAD-Project/OpenROAD/pull/6756

Issue #6643, opened 2025-02-05
Fanout cells inserted during global placement


kareefardi:
### Describe the bug

This is new behavior and there is nothing in the documentation which suggests that this is intentional. I also believe that global placement shouldn't manipulate the design at all. 

### Expected Behavior

Clarification if this is a bug or intentional. 

### Environment

```shell
Git commit: 6f4c5c592f3f5f2e0ee5d4a91b86622deedb57b5
kernel: Linux 6.11.10-1-liquorix-amd64
os: Ubuntu 22.04.5 LTS (Jammy Jellyfish)
cmake version 3.25.1
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-18783-g6f4c5c592
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-nonnull
[...]

PR #6748, opened 2025-02-21
gpl: report rsz cells

gudeh:
The repair_design API for gpl in rsz did not report violation counters, unlike the other two instances of repair_design and repair_nets. To address this, I created a new function to avoid code duplication and to include this reporting feature during gpl timing-driven iteartions, as requested in issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/6643#issuecomment-2668973624.

This change also maintains a consistent order of violation counters across different instances of repair_design reports and improves code encapsulation.

Additionally, I introduced sanity checks in gpl non-virtual timing-driven iterations, regarding the number of inserted instances reported by rsz and gpl. The number of inserted instances should be consistent across regions (NesterovBase objects), the common object (NesterovBaseCommon), and should also match the count of inserted buffers provided by rsz. 

This new reports should help understand the issue happening in https://github.com/The-OpenROAD-Project/OpenROAD/issues/6745


gudeh:
I ran secure-CI for this branch and downloaded all the logs, including those for secure designs. I did not find any new sanity check warnings, which means all current designs have the same number of cells created between NesterovBase, NesterovBaseCommon, and as reported by rsz, as expected.

gudeh:
BTW, this has no change in functionality

src/rsz/src/RepairDesign.hh
```diff
@@ -107,6 +107,12 @@ class RepairDesign : dbStaState
                  double cap_margin);
   void repairClkNets(double max_wire_length);
   void repairClkInverters();
+  void printViolationCounters(bool invalidate_driver_vertices,
```
eder-matheus, line 110: I believe `reportViolationCounters` is a better name. Generally `report` over `print` is a better option.

src/rsz/src/RepairDesign.cc
```diff
@@ -2223,4 +2187,40 @@ void RepairDesign::printProgress(int iteration,
...
+  }
+  if (inserted_buffer_count_ > 0) {
+    logger_->info(utl::RSZ,
+                  invalidate_driver_vertices ? 55 : 38,
+                  "Inserted {} buffers in {} nets.",
+                  inserted_buffer_count_,
+                  repaired_net_count);
+    if (invalidate_driver_vertices) {
+      resizer_->level_drvr_vertices_valid_ = false;
+    }
```
eder-matheus, line 2222: You could change the flag name to something like `valid_driver_vertices` and just assign it to the `level_drvr_vertices_valid_` class member.

eder-matheus (changes requested):

src/rsz/src/RepairDesign.cc
gudeh: I mean, some in some instances of existing code, the violations are reported and `resizer_->level_drvr_vertices_valid_` is left as is. If we assign to true we could potentially overwrite the default of false. Also, I do not fully understand this variable.

eder-matheus: I see, I didn't notice it before. Then it's safer to leave it as it is.

eder-matheus (approved):

PR #6758, opened 2025-02-24
gui: add option to exclude buffering nodes when generating report_checks

AcKoucher:
This is to help when we want to check what happens to a path across different versions of OR.

Closest Match option now becomes a menu with two options:
- Exact (Same as the previous Closest Match)
- No Buffering (Exclude buffers and back-to-back inverters)

### DIFF
--- a/src/gui/src/timingWidget.cpp
+++ b/src/gui/src/timingWidget.cpp
@@ -45,6 +45,7 @@
 
 #include "db_sta/dbSta.hh"
 #include "gui_utils.h"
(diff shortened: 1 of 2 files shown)


AcKoucher:
@gudeh @eder-matheus FYI

AcKoucher:
@maliberty There is a problem with the implementation here. I'm skipping just the interface node between back-to-back inverters, we need to skip all the four nodes. I'll close, work on the fix and reopen when it's done.

maliberty (approved):

PR #6762, opened 2025-02-24
rsz: Cache result of getSwappableCells()

mikesinouye:
For large (private) PDKs that have many equivalent cells (~50 per logic type), large designs spend significant time recomputing the equivalent cell list during rsz operations. In typical OR flows, the liberty library does not change after init so this list should not have to be recomputed.

Also move checking of `dontUse()` and `isLinkCell() `to `getSwappableCells()` so it does not have to be repeatedly checked. This also eliminates the need to remember to conduct this check where it is called. The cache is reset if an additional liberty file is read or when the dont_use set is updated.

Baseline private design `global_placement -timing-driven` : 52s
With caching private design `global_placement -timing-driven`: 29s (-44%)

On a large private design, time attributed to these calls was approximately half of total flow runtime (>11hr) (
image
)

### DIFF
--- a/src/rsz/include/rsz/Resizer.hh
+++ b/src/rsz/include/rsz/Resizer.hh
@@ -749,6 +749,9 @@ class Resizer : public dbStaState, public dbNetworkObserver
   // exact same buffers when reparing clock nets.
   LibertyCellSeq clk_buffers_;
 
+  // Cache results of getSwappableCells() as this is expensive for large PDKs.
+  std::unordered_map<LibertyCell*, LibertyCellSeq> swappable_cells_cache_;
+
   CellTargetLoadMap* target_load_map_ = nullptr;
   VertexSeq level_drvr_vertices_;
   bool level_drvr_vertices_valid_ = false;
--- a/src/rsz/src/RepairDesign.cc
+++ b/src/rsz/src/RepairDesign.cc
@@ -2110,23 +2110,21 @@ LibertyCell* RepairDesign::findBufferUnderSlew(float max_slew, float load_cap)
                   > resizer_->bufferDriveResistance(buffer2);
          });
     for (LibertyCell* buffer : swappable_cells) {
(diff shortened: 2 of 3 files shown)


gadfort:
You need to add clearing of the new cache to:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/9193d43e5de89924b7e5839aae5325051b3383ee/src/rsz/src/Resizer.cc#L1887
and 
https://github.com/The-OpenROAD-Project/OpenROAD/blob/9193d43e5de89924b7e5839aae5325051b3383ee/src/rsz/src/Resizer.cc#L1900

otherwise changes to dont_use will get out of sync

mikesinouye:
> You need to add clearing of the new cache to:
> 
> https://github.com/The-OpenROAD-Project/OpenROAD/blob/9193d43e5de89924b7e5839aae5325051b3383ee/src/rsz/src/Resizer.cc#L1887
> 
> 
> and
> https://github.com/The-OpenROAD-Project/OpenROAD/blob/9193d43e5de89924b7e5839aae5325051b3383ee/src/rsz/src/Resizer.cc#L1900
> 
> otherwise changes to dont_use will get out of sync

Added those two changes, thanks

maliberty (approved):

povik:
Is there a piece missing? I don't see the cache getting filled other than in the trivial `swappable_cells_cache_[source_cell] = {};` case

mikesinouye:
> Is there a piece missing? I don't see the cache getting filled other than in the trivial `swappable_cells_cache_[source_cell] = {};` case

You're completely right, I must have omitted the final part when copying changes into the PR. I'll send a followup fix. Thanks for catching that!

povik:
Now that the cache can be the backing for it I think it would be great if we returned a LibertyCellSeq pointer rather than a value. Though a repeated getSwappableCells call could invalidate earlier pointers (once a new cell is added to the cache), so there's some pointer lifetime hazard. We probably don't have any users in the resizer for which this would be an issue but someone might misuse this at some point.

PR #6765, opened 2025-02-25
ifp: Adds make_rows command

QuantamHD:
Some floorplan DEFs lack row information. This PR adds a way to add rows to the floorplan without needing to redefine everything


src/ifp/README.md
```diff
@@ -95,6 +95,33 @@ make_tracks
 | `[-x_offset]`, `[-y_offset]` | If set, overrides the LEF technology x-/y- offset. Use the same unit as in the LEFT file. |
 | `[-x_pitch]`, `[-y_pitch]` | If set, overrides the LEF technology x-/y- pitch. Use the same unit as in the LEF file. |
 
+### Make Rows
+
+The `make_rows` command removes existing rows. 
+Use the `make_rows` command to add rows to an existing floorplan. Useful,
+if you're floorplan is stored as a floorplan def without rows.
```
gadfort, line 102: you're -> your

QuantamHD: Thanks Peter.

src/ifp/src/InitFloorplan.tcl
```diff
@@ -174,6 +174,114 @@ proc initialize_floorplan { args } {
   }
 }
 
+sta::define_cmd_args "make_rows" {
```
maliberty, line 177: This looks to be nearly a copy of initialize_floorplan.  Why not just make a -rows_only option rather than a whole new command?

QuantamHD: I thought about it, but it felt like the current solution is a better separation of responsibilities. Keeps the command and code more focused.

maliberty: It leads to a lot of duplication of code which is a negative.  If you prefer this path then please refactor the tcl to share more.

QuantamHD: I can just turn intialize_floorplan into a thing that calls make_die and make_rows. Is that your preference?

maliberty: @gadfort any opinion?

gadfort: I think because of the high number of shared flags between the two commands, it would make sense to just add a flag to init_floorplan, that will avoid the two commands getting out of sync, and avoid adding a new command

QuantamHD: But they really don't share that many flags. I think if I break this up into two commands it'll be cleaner for everyone. No one really needs to know make die even exists.

maliberty: These are all shared
```
  -site site_name
  [-additional_sites site_names]
  [-flip_sites site_names]
  [-row_parity NONE|EVEN|ODD]
```

They will know the make_rows exists and it is a lot of copying imo.  Any new flag added to make_rows will also have to be added to initialize_floorplan and be passed through.  I don't find that cleaner vs a simple flag to skip a step.

QuantamHD: Let me try anyways, and I think you'll change your mind.

QuantamHD: I think what I ended up with is pretty clean. I think it's a better organization for ifp, and shouldn't be too much harder to add new args to either command.

src/ifp/src/InitFloorplan.tcl
```diff
@@ -42,53 +42,49 @@ sta::define_cmd_args "initialize_floorplan" {[-utilization util]\
...
-  }
+  make_die_helper [array get keys]
+  make_rows_helper [array get keys]
+}
+
+sta::define_cmd_args "make_die" {
+                 [-utilization util]\
+					       [-aspect_ratio ratio]\
+					       [-core_space space | {bottom top left right}]\
+					       [-die_area {lx ly ux uy}]\
```
maliberty, line 71: The indentation is off

maliberty:
I can't say I like this better than a simple skip flag.
@gadfort opinion?

src/ifp/src/InitFloorplan.tcl
QuantamHD: Removed.

maliberty (approved):

PR #6766, opened 2025-02-25
rsz: fix area reporting

gadfort:


maliberty (approved):
D'oh!

@povik this is probably the issue in your PR as well

PR #6778, opened 2025-02-26
drt: prepPatternInstHelper and some refactors

bnmfw:
Supports #5867.

This PR has an ISPD and Secure CI associated and will be opened once they pass.

This PR has some refactoring of the access pattern code to support incremental PA.

### `prepPatternInstHelper`
Currently, there is no single function that _actually_ supports the passing of an unique inst to solve its access patterns, although `prepPatternInst` is almost it. The aforementioned function was renamed to `prepPatternInstHelper` and a new function that now has the name `prepPatternInst` was created in its place to simply receive the unique inst and solve its access pattern.

### `curr_unique_inst_idx` plumbing elimination.
Something very weird happens in the access pattern code as mentioned in the discussion of #6731. Many functions of the access patterns code expect both a pointer to an unique inst (`frInst* inst`) as well as the index this unique inst exists in the unique_ vector (`int curr_unique_inst_idx`). This is redundant and very error prone, so the passing of the index was eliminated, and any function that required it now gets the pointer to the instance itself. Some functions that used to get both now just get the instance. 

The appropriate refactoring was done to accommodate for this change. In some cases, deep into functions calls the inst was actually retrieved through some other way, such as getting the instance from the pin analyzed. Now that those functions get the instance, those methods are not needed.

### Other smaller refactors
The now `prepPatternInstHelper` had a weird parameter (`x_weight`) that acted as a boolean, but was a int that assumed wither 1 or 0 value and was used inside an equation. The parameter (now `use_x`) now actually acts as a boolean.

`genPatternsCommit` also had some slight refactoring at its end, two variable name changes (I don't know how those were not caught on the PA general refactor), and a slight owner refactor (that code repetition bothered me).

### DIFF
--- a/src/drt/src/pa/FlexPA.h
+++ b/src/drt/src/pa/FlexPA.h
@@ -593,7 +593,9 @@ class FlexPA
 
   void prepPattern();
 
-  int prepPatternInst(frInst* inst, int curr_unique_inst_idx, double x_weight);
+  int prepPatternInst(frInst* inst,
+                      int curr_unique_inst_idx,
+                      bool use_x = true);
 
   int genPatterns(const std::vector<std::pair<frMPin*, frInstTerm*>>& pins,
                   int curr_unique_inst_idx);
--- a/src/drt/src/pa/FlexPA.h
+++ b/src/drt/src/pa/FlexPA.h
@@ -593,9 +593,11 @@ class FlexPA
 
   void prepPattern();
 
-  int prepPatternInst(frInst* inst,
-                      int curr_unique_inst_idx,
-                      bool use_x = true);
+  void prepPatternInst(frInst* inst, int curr_unique_inst_idx);
+
+  int prepPatternInstHelper(frInst* inst,
+                            int curr_unique_inst_idx,
+                            bool use_x = true);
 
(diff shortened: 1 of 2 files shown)


src/drt/src/pa/FlexPA.h
```diff
@@ -593,17 +593,19 @@ class FlexPA
 
   void prepPattern();
 
-  int prepPatternInst(frInst* inst, int curr_unique_inst_idx, double x_weight);
+  void prepPatternInst(frInst* inst);
 
-  int genPatterns(const std::vector<std::pair<frMPin*, frInstTerm*>>& pins,
-                  int curr_unique_inst_idx);
+  int prepPatternInstHelper(frInst* inst, bool use_x = true);
```
osamahammad21, line 598: for better readability, I prefer not having this as a default valued argument. It brings attention to the one reading the code that there is something called use_x in here.
It would be also helpful if we have a documentation for this function stating this parameter as it is not easily interpreted by name. Not necessarily in this PR.

osamahammad21 (changes requested):

osamahammad21 (approved):

Issue #6786, opened 2025-02-27
Nix build fails


FlinkbaumFAU:
### Describe the bug

``nix build '.submodule=1#'`` fails because CMake tries to download googletest with ``FetchContent``, which is not allowed by the nix sandbox.

[openroad-build.log](https://github.com/user-attachments/files/19007350/openroad-build.log)

### Expected Behavior

``nix build '.submodule=1#'`` should build OpenROAD successfully

### Environment

```shell
[WARNING] Your current OpenROAD version is outdated.
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Linux 4.18.0-477.21.1.el8_8.x86_64
os: RHEL Linux 8.8 (Green Obsidian)
cmake version 3.20.2
-- The CXX compiler identification is GNU 8.5.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-19352-g98b7eda40
-- System name: Linux
-- Compiler: GNU 8.5.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- Older version of GCC detected. Linking against stdc++fs
-- The C compiler identification is GNU 8.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull
[...]

PR #6787, opened 2025-02-27
Prevent CMake from downloading content with FetchContent

FlinkbaumFAU:
Use nix instead of CMake to download all necessary deps.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -37,6 +37,7 @@
   lib,
   clangStdenv,
   fetchFromGitHub,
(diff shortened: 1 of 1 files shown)


FlinkbaumFAU:
Refering #6786

donn (approved):
I don't have the bandwidth to test it, but it looks fine. If you say it builds successfully, this can be merged.

FlinkbaumFAU:
Signature is added

PR #6790, opened 2025-02-27
rsz: use the swappable cells cache added in #6762.

mikesinouye:
This actually uses the cache. I had inadvertently omitted this part of the change when creating the PR. Thanks for catching that @povik.

I also added clearing of the cache in Resizer::init() so that we respect changes to the recently added opt configs.


precisionmoon (commented):
Resizer::init() is inside a preamble for many resizer functions like repair_design, repair_timing and buffer_inputs/ibuffer_outputs.  Do we want to reset the cache in Resizer::init(Logger* logger, dbDatabase* db, dbSta* sta, ....) instead?

mikesinouye:
> Resizer::init() is inside a preamble for many resizer functions like repair_design, repair_timing and buffer_inputs/ibuffer_outputs. Do we want to reset the cache in Resizer::init(Logger* logger, dbDatabase* db, dbSta* sta, ....) instead?

The main issue is that the equivalent cache is invalidated each time one of the various `set_opt_config`s such as `limit_sizing_area` is set. Currently, these settings are re-applied in each call to `Resizer::initBlock()`, not when they are actually set in TCL. Therefore it seemed I had to reset the cache for each call to the resizer. We still get the majority of the performance uplift by doing so.

If we could relocate the updating of the opt configs to when they are updated, then I think we could clear the cache right after.

mikesinouye:
Hi @precisionmoon , this PR is ready for review at your earliest convenience.

precisionmoon (approved):

precisionmoon:
> > Resizer::init() is inside a preamble for many resizer functions like repair_design, repair_timing and buffer_inputs/ibuffer_outputs. Do we want to reset the cache in Resizer::init(Logger* logger, dbDatabase* db, dbSta* sta, ....) instead?
> 
> The main issue is that the equivalent cache is invalidated each time one of the various `set_opt_config`s such as `limit_sizing_area` is set. Currently, these settings are re-applied in each call to `Resizer::initBlock()`, not when they are actually set in TCL. Therefore it seemed I had to reset the cache for each call to the resizer. We still get the majority of the performance uplift by doing so.
> 
> If we could relocate the updating of the opt configs to when they are updated, then I think we could clear the cache right after.

swappable cell cache is now cleared only when there is a change to set_opt_config in initBlock().

precisionmoon:
> Hi @precisionmoon , this PR is ready for review at your earliest convenience.

Looks good.  Thanks much for the update.

PR #6792, opened 2025-02-27
utl: add Progress reporting + gui implementation

gadfort:
Adds:
- `Progress` class to allow tools to announce their progress.
- `ProgressReporter` individual class used to track progress of a specific tool/process (this allow us to have multiple reporters in parallel when/if that occurs).
- Adds a reporter to DRT's main loop to allow us to exit early if an interrupt is received.
- Adds a capture of `control+C` to get the interrupts and (3x `control+C` in less than 10 seconds will terminate openroad).

Notes:
- We can go over the API to ensure it's contains what we want, but this was the minimum I found I needed to get the GUI working and capture interrupts.

GUI:
Adds a progress bar to the bottom right corner (this can display the progress of a single or multiple reporterts)
Pasted image
For example:
Pasted image (2)

Detailed view of reporters (this uses a set of reports that were interested for testing, but should not be included here):
Pasted image (3)

GUI interrupt:
Pasted image (4)

CLI:
Since there wasn't an easy tool to add reporting to, the logger is not getting exercised at the moment.


oharboe:
Nice! Some thoughts on use-case and user-experience.

I frequently run `make GUI_TIMING=0 gui_xxx`. However, this requires me to first start the GUI, discover that reading timing is taking too long, kill OpenROAD, restart with GUI_TIMING=0.

Ideally, I'd just launch the GUI and click "Skip" if something takes too long.

"Abort" or "Stop" sounds scary. I don't know what to expect of the state of the system is if I click "stop" or "abort". Sounds like the GUI is telling me: "You asked for it and you deserve the result".

However, if the GUI says "loading timing information" and there is a friendly "skip", then my expectation is that nothing scary will happen, it will be equivalent of `make GUI_TIMING=0 gui_xxx`

gadfort:
@oharboe the messages can be adjusted, at the moment your usecase isn't supported anyways, because this will require some interface with opensta. The tools can end with a `end(false)` to indicate that no database changes were performed so the DB is safe, but if the tool is aborted and made changes, the database may no longer be in a safe state beyond inspection.
I think in all cases, I called it `interrupt` and not `abort`. Once the basic functionality / API is agreed upon we can revisit the messaging.
One usecase that is not covered here, but might be nice to have is a tool that cannot be interrupted, but still wants to provide a progress report (there the interrupt button should probably be disabled since it's not useful).

maliberty:
I tried it out and one thing I noticed is that when interrupting detailed_route the command still finishes with success and the remainder to the script continues on.  I think if a command is interrupted we should stop further script execution as it is unlikely you want to subsequent commands to work on the incomplete result.

maliberty:
I was thinking that 
image
 would become the cancel button rather than adding another button for that purpose.

maliberty:
It feel redundant to have the iteration in the progress and status areas
image

gadfort:
@maliberty I enabled the run button to handle interrupts too, changed the end function to return true if an interrupt occurred and generate an error if there was an interrupt, and removed the status bar messages.

maliberty:
@gadfort ready for re-review?  Also a conflict needs fixing.

gadfort:
@maliberty yes, I'll fix the merge conflict here in a bit (shouldn't impact the review)

src/utl/src/Progress.cpp
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  if (signal_halt->interrupts.size()
+      >= Progress::ProgressHalt::max_interrupts) {
+    // halt the application
+    signal(sig, SIG_DFL);
+    raise(sig);
+  } else if (signal_halt->interrupts.size() + 1
+             >= Progress::ProgressHalt::max_interrupts) {
+    // let user know one more will terminate openroad
```
maliberty, line 65: If there is no process to interrupt then it would be good to have a message on the first ctrl-c.  Currently it just disappears with no response.

src/utl/include/utl/Progress.h
```diff
@@ -0,0 +1,193 @@
...
+    std::queue<std::time_t> interrupts;
+    constexpr static int max_interrupts = 3;
+    constexpr static int max_prune_delay = 10;
+  };
+
+  Progress(Logger* logger);
+  virtual ~Progress();
+
+  // Start an activity without any progress information
+  std::shared_ptr<ProgressReporter> start(const std::string& name);
```
maliberty, line 72: When would this be used?

maliberty:
Can calls to start* be nested in this PR?

gadfort:
> Can calls to start* be nested in this PR?

What do you mean?

src/utl/include/utl/Progress.h
gadfort: it wouldn't, I'll remove it

src/utl/src/Progress.cpp
gadfort: should be easy

maliberty (approved):

maliberty:
I feel there are a few improvements to be made in how drt reports progress but they can be handled later.  Thanks for taking this on!

PR #6793, opened 2025-02-28
mpl: ensure a single-macro macro cluster is not considered an array

AcKoucher:
This shouldn't impact any of our designs, but it's a valid fix in terms of consistency.
SA weights for an array of interconnected macros are different than the ones of a conventional macro placement, that's the reason why the orientations change in the regression tests.


AcKoucher:
Running Secure-CI

AcKoucher:
Secure-CI seems fine. I see a DRC in a private design which happens to be on master as well.

maliberty (approved):

PR #6795, opened 2025-03-01
ip::address::from_string is deprecated, use make_address instead.

hzeller:
ip::address::from_string is deprecated since 1.66.0 with an explanation to use make_address() instead.
https://beta.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__address/from_string.html

Last version it is available is in boost 1.86.0
https://beta.boost.org/doc/libs/1_86_0/doc/html/boost_asio/reference/ip__address/from_string.html

The current 1.87.0 has it removed.

### DIFF
--- a/src/utl/src/prometheus/metrics_server.cpp
+++ b/src/utl/src/prometheus/metrics_server.cpp
@@ -70,7 +70,7 @@ PrometheusMetricsServer::~PrometheusMetricsServer()
           io_context;  // Use a separate io_context for the connection.
       boost::asio::ip::tcp::socket socket(io_context);
       boost::asio::ip::tcp::endpoint endpoint(
-          boost::asio::ip::address::from_string("127.0.0.1"), port_);
+          boost::asio::ip::make_address("127.0.0.1"), port_);
       socket.connect(endpoint);         // This will unblock the accept().
     } catch (const std::exception& e) { /*Do nothing, we're dying*/
     }


hzeller:
cc @QuantamHD

maliberty (approved):

maliberty:
Please note that 1.87 failed for me due to https://github.com/boostorg/asio/issues/442

hzeller:
ah, thanks, good to know. I'll package OpenROAD with 1.86.0 in nix for now.

maliberty:
I didn't know you are using nix.  I'm interested to hear more about it as we are thinking of using it to replace the dependency installer for OR.

hzeller:
I like it. I'd say NixOS is the best Linux distribution I have played with in the last 32 years I am using Linux as my main driver. I think this is how packaging should look like.

I am still no expert in Nix internals, but as a user and developer it provides the wonderful clean set-up of a work environment one longs for when developing software. No accumulation of 'cruft' of libraries installed over the time on a system like in other distros - just exactly the dependencies visible that are needed. Independent of the machine it is run on. 

Sometimes it is a bit work if one in need is not packaged yet, so have to package yourself, but the effort is well worth it. And it happens rarely as Nix packages has one of the most comprehensive set of stuff packaged.

I am using native NixOS, so for me dependencies based on nix packages are great. But even for other Unixes: dependency definitions based on nix-packages should just work using the nix package manager when installed on these systems (downside of course: these users would have to install the nix package manager).

PR #6796, opened 2025-03-02
drt: unique_inst_patterns_ structure structure refactor

bnmfw:
Supports #5867.
ISPD and Secure CI associated, will be opened once safe.

### `unique_inst_patterns_` structure change.
One of the requirements of incremental PA is for the used structures to be dynamic. Once a complete run of PA is done, it is necessary to run it again for a set of instances while keeping data of unaltered instances. While #6778 introduces a function to allow for solving the access patterns of a single unique instance, the `unique_inst_patterns_` does not allow for new entries to be inserted. The structure used to be a vector of fixed sized, which made the managing of new entries very difficult. In that light the structure was changed to an unordered map, in which the unique instance pointer acts as a key. This allows for inserting and deleting entries without major disruptions of the whole structure.

#### Related minor changes.
- `FlexPA.h` now includes `#include <boost/serialization/unordered_map.hpp>` to support the unordered map serialization.
- `FrInst()` now has an empty default constructor, also required for serialization.
- `UniqueInsts` now has a function to retrieve the unique instance related to the instance.
- Modifications of `unique_inst_patterns_` are protected by `#pragma omp critical` to avoid race conditions.


src/drt/src/pa/FlexPA_acc_pattern.cpp
```diff
@@ -778,8 +783,8 @@ bool FlexPA::genPatternsCommit(
   if (target_obj != nullptr
       && genPatternsGC({target_obj}, objs, Commit, &owners)) {
     pin_access_pattern->updateCost();
-    const int inst_idx = unique_insts_.getIndex(inst);
-    unique_inst_patterns_[inst_idx].push_back(std::move(pin_access_pattern));
+#pragma omp critical
+    unique_inst_patterns_[unique_inst].push_back(std::move(pin_access_pattern));
```
osamahammad21, line 787: Not critical. No other thread would access the same entry for the current unique_inst. And we already created the entry beforehand, so the map would not be touched.

osamahammad21 (changes requested):
- unique_inst_patterns_ is not touched in deleteInst which doesn't seem right. The entry should be removed or edited if the class head is changed.
- If you're planning to use frInst* as the key, then you should remove all index related APIs and members.

bnmfw:
@osamahammad21 `UniqueInsts::deleteInst` does not own `unique_inst_patterns_`, that is why does not cover the deletion of the function. #6866, which is branched from this PR created `FlexPA::deleteInst` and covers this issue.

There is indeed an effort to use frInst* as keys. Using the index of unique_ as keys to other structures does not fit very well the attempt to make structures dynamic. There might be a PR later on that revamps the index usage as a key.

osamahammad21 (approved):

PR #6799, opened 2025-03-03
cts: add recording of masters used to report_cts

gadfort:
Changes:
- no functional changes to CTS, but adds tracking of the masters used during CTS to be able to report that as part of report_cts

### DIFF
--- a/src/cts/src/CtsOptions.h
+++ b/src/cts/src/CtsOptions.h
@@ -46,6 +46,7 @@
 #include "CtsObserver.h"
 #include "Util.h"
 #include "odb/db.h"
+#include "odb/dbBlockCallBackObj.h"
 #include "utl/Logger.h"
 
 namespace stt {
@@ -53,9 +54,16 @@ class SteinerTreeBuilder;
 }
 namespace cts {
 
-class CtsOptions
+class CtsOptions : public odb::dbBlockCallBackObj
 {
  public:
+  enum class MasterType
+  {
+    DUMMY,
+    TREE
+  };
+  using MasterCount = std::map<odb::dbMaster*, int>;
+
   CtsOptions(utl::Logger* logger, stt::SteinerTreeBuilder* sttBuildder)
       : logger_(logger), sttBuilder_(sttBuildder)
   {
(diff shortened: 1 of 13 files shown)


src/cts/src/CtsOptions.h
```diff
@@ -253,6 +254,9 @@ class CtsOptions
   const char* getCtsLibrary() { return ctsLibrary_.c_str(); }
   bool isCtsLibrarySet() { return !ctsLibrary_.empty(); }
 
+  void recordBuffer(odb::dbMaster* master) { buffer_count_[master]++; }
+  std::map<odb::dbMaster*, int> getBufferCount() const { return buffer_count_; }
```
maliberty, line 258: return by const&

src/cts/src/TritonCTS.cpp
```diff
@@ -1262,6 +1265,7 @@ Clock TritonCTS::forkRegisterClockNetwork(
   odb::dbMaster* master = db_->findMaster(options_->getRootBuffer().c_str());
   topBufferName = "clkbuf_regs_" + std::to_string(regTreeRootBufIndex_++) + "_"
                   + clockNet.getSdcName();
+  options_->recordBuffer(master);
```
maliberty, line 1268: It would be safer to use a db observer to capture instance creation rather the annotation each call site.

gadfort: good point. Might complicate the dummy load marking, but I'll take a look

src/cts/test/simple_test.ok
```diff
@@ -46,6 +46,7 @@ Total number of Clock Roots: 1.
 Total number of Buffers Inserted: 3.
 Total number of Clock Subnets: 3.
 Total number of Sinks: 16.
-Buffers used:
+Cells used:
   CLKBUF_X3: 3
+Dummys used:
```
maliberty, line 51: Suppress the header if there are no dummies

maliberty (changes requested):

maliberty (approved):

PR #6805, opened 2025-03-03
rsz: Fix a crash in row balancing for PDKs with macros w/out a SITE in the LEF.

mikesinouye:
Macros are still later skipped by nature of being FIXED, but some PDKs do not specify the SITE in the LEF for macros, so a crash can occur if site is nullptr in balanceRowUsage().

### DIFF
--- a/src/rsz/src/Resizer.cc
+++ b/src/rsz/src/Resizer.cc
@@ -607,8 +607,8 @@ void Resizer::balanceRowUsage()
   for (auto inst : block_->getInsts()) {
     auto master = inst->getMaster();
     auto site = master->getSite();
-    // Ignore macros, and multi-height cells for now
-    if (inst->isBlock() || site->hasRowPattern()) {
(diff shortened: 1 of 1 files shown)


src/rsz/src/Resizer.cc
```diff
@@ -607,8 +607,8 @@ void Resizer::balanceRowUsage()
   for (auto inst : block_->getInsts()) {
     auto master = inst->getMaster();
     auto site = master->getSite();
-    // Ignore multi-height cells for now
-    if (site->hasRowPattern()) {
+    // Ignore macros, and multi-height cells for now
+    if (inst->isBlock() || site->hasRowPattern()) {
```
maliberty, line 611:
```suggestion
    // Ignore multi-height cells for now
    if (site && site->hasRowPattern()) {
```

maliberty (approved):

PR #6811, opened 2025-03-04
Cts fix register top buffer position

arthurjolo:
Currently when CTS finds the position to place the top buffer for the register tree, that is the center of the register sink region, it considers the position of the buffer that was just created. Since the position of new instances is considered to be (0, 0) the sink region center is affected and the position of the root buffer is shifted to the bottom left.

This PR changes CTS to not consider the top buffer when finding its position avoiding this error.


precisionmoon:
@arthurjolo , please rebase flow test ibex_sky130hd after merging your branch with the master.

PR #6813, opened 2025-03-04
DPO: support edge spacing

osamahammad21:
- Add Journal tracking for easier redo/undo operations.
- Move Grid class from dpl to dpo as an initial step to merge both modules.
- Fix wrong interpretation of row site count limit.
- Add support to LEF58_EDGESPACINGTABLE rule


src/dpo/src/detailed_manager.cxx
```diff
@@ -902,6 +905,9 @@ void DetailedMgr::assignCellsToSegments(
   double movementX = 0.;
   double movementY = 0.;
   for (Node* nd : nodesToConsider) {
+    // place in grid
+
+    //
```
maliberty, line 910: Is something missing here or dead comments?

maliberty (approved):

osamahammad21:
This causes a non-related DRC in the secure CI that is fixed by PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6898
Should be merged after

PR #6815, opened 2025-03-04
mpl: fix orientation improvement and add tests

AcKoucher:
Obs: orientation_improvement.lef is the same as macro_only.lef but with more reasonable iterm positions in the macros (it's very easier to see the result of the tests).


AcKoucher:
Running Secure-CI

AcKoucher:
Secure-CI is fine.

PR #6816, opened 2025-03-06
DPL: public Grid class accessible to DPO

osamahammad21:


src/dpl/include/dpl/Grid.h
```diff
@@ -43,13 +43,35 @@
...
 
 namespace dpl {
 
 using odb::dbOrientType;
 using odb::dbSite;
 
+class GridNode
+{
+ public:
+  GridNode() = default;
```
maliberty, line 56: Should have a virtual dtor.  No need for `GridNode() = default;`

src/dpl/src/Opendp.cpp
```diff
@@ -381,8 +381,9 @@ dbInst* Opendp::getAdjacentInstance(dbInst* inst, bool left) const
   dbInst* adjacent_inst = nullptr;
 
   // do not return macros, endcaps and tapcells
-  if (pixel != nullptr && pixel->cell && pixel->cell->db_inst_->isCore()) {
-    adjacent_inst = pixel->cell->db_inst_;
+  if (pixel != nullptr && pixel->cell
+      && static_cast<Cell*>(pixel->cell)->db_inst_->isCore()) {
+    adjacent_inst = static_cast<Cell*>(pixel->cell)->db_inst_;
```
maliberty, line 386: Why not use getDbInst rather than static_cast?  Similar in a few other places

maliberty:
This PR represents an intermediate state along the way to unifying dpl & dpo.  Ultimately the grid will become private again and the abstract node class will not be needed.

src/dpl/src/Opendp.cpp
maliberty: The same occurs on the line above
image

maliberty (approved):

PR #6818, opened 2025-03-07
dbSta: add a report_timing_histogram command

mikesinouye:
We have the goal of reporting slack histogram data in batch/non-GUI runs of OpenROAD. Preferably, this is part of a text PPA report accessed over the terminal by a user in a non-desktop environment.

There is an existing GUI slack histogram in OR, and a GUI clocktree viewer which already supports exporting images via the command line using `save_clocktree_image`. However, the `save_clocktree_image` command requires the GUI to be open. So, we cannot perfectly replicate that code as our preferred use case is non-GUI.

Upon my initial investigation, updating the histogram chart logic to support exporting of the image without the GUI open does not seem trivial, as there is special considerations for the `LayoutViewer` in `save_image` that allow it to do this that do not apply to the clock tree and charts widgets. The `save_image` command, which does support this, works quite differently as it is connected to the LayoutViewer which is a different setup.

The logic required seemed lightweight enough (~60loc) to justify its existence as a standalone own utility. This PR adds a add a `report_timing_histogram` command that can accept the number of bins as a param.

Example of what this generates:

```
% report_timing_histogram -num_bins 25
[-0.438, -0.406): * (66)
[-0.406, -0.375): * (451)
[-0.375, -0.343): * (1213)
[-0.343, -0.311): * (1738)
[-0.311, -0.280): * (1769)
[-0.280, -0.248): * (1447)
[-0.248, -0.216): * (1107)
[-0.216, -0.185): * (1010)
[-0.185, -0.153): ** (2561)
[-0.153, -0.122): ****** (7154)
[-0.122, -0.090): ********* (10738)
[-0.090, -0.058): ************* (14195)
[-0.058, -0.027): ****************** (20405)
[-0.027,  0.005): ********************** (23728)
[ 0.005,  0.037): ************************* (27794)
[ 0.037,  0.068): ***************************** (32276)
[ 0.068,  0.100): ***************************************** (44733)
[ 0.100,  0.132): ************************************************** (53825)
[ 0.132,  0.163): ********************** (24669)
[ 0.163,  0.195): **************************** (30206)
[ 0.195,  0.227): ************************* (27583)
[ 0.227,  0.258): **************** (18242)
[ 0.258,  0.290): **** (4973)
[ 0.290,  0.322): * (621)
[ 0.322,  0.353): * (1580)
```


src/dbSta/src/dbSta.cc
```diff
@@ -690,6 +690,58 @@ void dbSta::reportCellUsage(odb::dbModule* module,
...
+    }
+    bins[bin]++;
+  }
+
+  // Print the histogram.
+  const int largest_bin = *std::max_element(bins.begin(), bins.end());
+  for (int bin = 0; bin < num_bins; ++bin) {
+    const float bin_start = min_slack + bin * bin_range;
+    const float bin_end = min_slack + (bin + 1) * bin_range;
+    int bar_length = max_bin_width * bins[bin] / largest_bin;
```
maliberty, line 733: Did you intend to round down rather than off by doing integral division?

src/dbSta/test/report_timing_histogram.ok
```diff
@@ -0,0 +1,11 @@
...
+[INFO ODB-0128] Design: top
+[INFO ODB-0130]     Created 6 pins.
+[INFO ODB-0131]     Created 5 components and 24 component-terminals.
+[INFO ODB-0132]     Created 2 special nets and 10 connections.
+[INFO ODB-0133]     Created 10 nets and 14 connections.
+[ 9.433,  9.516): ************************* (1)
+[ 9.516,  9.600):  (0)
+[ 9.600,  9.684):  (0)
+[ 9.684,  9.767): ************************* (1)
+[ 9.767,  9.851): ************************************************** (2)
```
maliberty, line 11: Nit: the last bin is closed on both ends

src/dbSta/src/dbSta.cc
mikesinouye: It wasn't intentional per se, but with a bin resolution of 50 the slight simplification of rounding down seemed fine, but it's correct to round it.

src/dbSta/test/report_timing_histogram.ok
mikesinouye: Done.

maliberty (approved):

PR #6820, opened 2025-03-07
mpl: IOs abstraction adaptation to ORFS flow - second version (enhanced)

### LINKED ISSUE #5669: Use IO constraints kept by PPL inside MPL2 to improve QoR

AcKoucher:
Resolve #5669

## Introductory Considerations
- This is the second and final step of the enhancements to adapt the IOs abstraction inside MPL to ORFS flow.
- Differently from the first version, in which we made simplifications w.r.t. dimensions of the constraints, this second version considers the actual dimensions of the constraints regions for pins, whether they are -exclude constraints or pins' constraints.

## Goals
1. Further improve QoR in terms of WL.
2. Improve accuracy of the locations that can be occupied by macros w.r.t. pin access.

### Modify Shapes of Clusters of Unplaced IO Pins

Instead of virtually constraining IOs to the edges of the die area, we now use the actual constraint region that comes from ODB. This results in:
  - A cluster of unplaced IOs have now the dimensions of the constraint region to which those pins are constrained to. I.e., pins are constrained to regions instead of boundaries.
  - The cluster of unplaced IOs without any constraint is a special case. The pins are constrained to all available regions for pins i.e., the regions that are not blocked. The shape of this cluster is the die area itself.

As we create these clusters, we populate a map that links them with their respective `BoundaryRegion` (the new object in MPL's framework to help handling constraint regions - for constrained pins' clusters - & available regions - for the unconstrained pins' cluster - inside SA and during orientation improvement).

### Computation of Available Regions For Pins
When there are no constrained pins, but there are -exclude constraints set by the user, we use the blocked regions for pins which are stored in ODB to compute a list of available regions for pins (this is made by doing regions' subtraction across the edges of the die area, see `HierRTLMP::computeAvailableRegions`).

### Pin Access

The alternatives for pin access blockages' generation are still the same as the first version. However, we now use the dimensions of constraint regions:
  - If there are constrained pins, we create blockages for the respective constraint regions.
  - If there are no constrained pins, but there are -exclude constraints set by the user, we use the available regions for pins to create the pin access blockages.
  - If there are no constraints at all - neither pins' constraints or -exclude constraints - we don't create pin access blockages to give more freedom to SA.

The depth now is limited by the dimensions of the die area. A horizontal blockage will have at max 20% of the die width; a vertical blockage will have at max 20% of the die height.

There is one important difference w.r.t the first version: _We now consider the density of IOs when computing the depth of the blockages for constrained pins. Denser regions are deeper, i.e., have thicker blockages._

### Cluster / Macro Placement (SA)

**In the first version of the new IOs**, WL **was** computed based on distance to the edges:
  - In the case of a cluster for constrained pins, the distance to its specific edge of constraint;
  - In the case of the cluster of unconstrained pins, the distance to the closest unblocked boundary.

**Now, as the clusters of unplaced IOs are constrained to actual regions,** we compute WL as:
  - In the case of a cluster of constrained pins, we compute the distance from the macro bundled pin to the nearest point inside the constraint region.
  - In the case of the cluster of unconstrained pins, we compute the distance from the macro bundled to the nearest point of the nearest available region for pins.

Obs: When initializing the SA Core, we create a cache of these regions offset w.r.t. current outline.
Obs2: The new WL computation is made in DBU, so the annealer needs a `dbBlock*` in order to allow us to perform unit conversion.

### Boundary Pushing

As we now store the pin access blockages in a vector instead of a map Boundary -> Blockage, because we don't need the boundary data anymore, the Pusher receives this vector and, when working, computes overlap iterating the blockages of this vector.

### Orientation Improvement

When computing HPWL for a certain orientation and iterating bterms of a net:
  - If the pin is constrained to a region, use the nearest point of this region.
  - If the pin is not constrained to any region, use the nearest point of the nearest available region for pins.


maliberty:
Nice description!

AcKoucher:
I adapted the implementation to avoid using the center of regions when computing WL and these changes are ready for a first round of review. I also updated the description accordingly so it should give a reasonable explanation of what is going on.

maliberty:
secure CI?

AcKoucher:
> secure CI?

Running.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -2852,6 +2947,84 @@ Rect HierRTLMP::dbuToMicrons(const odb::Rect& dbu_rect)
...
+//   |                  |
+//   |                  | "b"
+//   |                  |
+//   |        |
+//   |        |
+//   |        |
+//   |                  |
+//   |                  | "a"
+//   |                  |
+std::vector<odb::Rect> HierRTLMP::subtractOverlapRegion(
```
maliberty, line 2961: FYI, boost ICL will manage intervals for you.  Luis used it in grt for a similar purpose (determining tracks blocked by a set of obstructions).

src/mpl/src/SimulatedAnnealingCore.h
```diff
@@ -136,8 +137,8 @@ class SimulatedAnnealingCore
   Rect outline_;
   Rect die_area_;  // Offset to the current outline.
 
-  // Boundaries blocked for IO pins
-  std::set<Boundary> blocked_boundaries_;
+  BoundaryRegionList available_regions_for_pins_;
```
maliberty, line 140: Are these regions for unconstrained pins or all pins?  I think it is the former but is unclear.

src/mpl/test/CMakeLists.txt
```diff
@@ -7,6 +7,12 @@ or_integration_tests(
     guides2
     io_constraints1
     io_constraints2
+    io_constraints3
+    io_constraints4
+    io_constraints5
+    io_constraints6
+    io_constraints7
+    io_constraints8
```
maliberty, line 15: Please add to the BUILD file as well.

src/mpl/src/clusterEngine.cpp
```diff
@@ -349,162 +350,86 @@ void ClusteringEngine::setBaseThresholds()
...
-  } else if (boundary == R) {
-    x = die.xMax();
-    y = die.yMin();
-    width = 0;
-  } else {  // Bottom
-    x = die.xMin();
-    y = die.yMin();
-    height = 0;
+  auto cluster
+      = std::make_unique<Cluster>(id_, "ios_" + std::to_string(id_), logger_);
```
maliberty, line 405: Would there be an any advantage to having a cluster per constraint region vs per bterm?

maliberty (changes requested):

src/mpl/src/util.h
```diff
@@ -3,12 +3,62 @@
...
+struct BoundaryRegion;
+class Cluster;
+
+using ClusterToBoundaryRegionMap = std::map<Cluster*, BoundaryRegion>;
+using BoundaryRegionList = std::vector<BoundaryRegion>;
+
+// One of the edges of the die area.
+enum Boundary
+{
+  NONE,
```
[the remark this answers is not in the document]
AcKoucher: Literally, none. :-)
It's from the original implementation.
Removed.

src/mpl/src/util.h
```diff
@@ -35,6 +85,19 @@ struct PenaltyData
...
+struct BoundaryRegion
+{
+  BoundaryRegion() = default;
+  BoundaryRegion(const odb::Line& line, const Boundary boundary)
+      : line(line), boundary(boundary)
+  {
+  }
+
+  odb::Line line;
+  Boundary boundary;
```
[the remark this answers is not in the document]
AcKoucher: The default ctor is needed due to `std::map` usage.

src/mpl/src/SimulatedAnnealingCore.h
maliberty: Perhaps available_regions_for_unconstrained_pins_

src/mpl/src/util.h
```diff
@@ -52,4 +115,109 @@ inline bool isAreaSmaller(const Tiling& tiling_a, const Tiling& tiling_b)
...
+{
+  if (rect.dx() != 0 && rect.dy() != 0) {
+    logger->error(utl::MPL,
+                  60,
+                  "Coundn't convert rect {} to line. The region is not a line.",
+                  rect);
+  }
+
+  odb::Line line;
+  switch (getBoundary(block, rect)) {
```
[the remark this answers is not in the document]
AcKoucher: Indeed! That's such a nice simplification. Thanks.

src/mpl/src/util.h
```diff
@@ -52,4 +115,109 @@ inline bool isAreaSmaller(const Tiling& tiling_a, const Tiling& tiling_b)
...
+                                              const odb::Point& target)
+{
+  if (region.boundary == L || region.boundary == R) {
+    if (target.y() >= region.line.pt1().y()) {
+      return odb::Point(region.line.pt0().x(), region.line.pt1().y());
+    }
+    if (target.y() <= region.line.pt0().y()) {
+      return odb::Point(region.line.pt0().x(), region.line.pt0().y());
+    }
+    return odb::Point(region.line.pt0().x(), target.y());
```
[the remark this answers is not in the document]
AcKoucher: I also prefer your way.

src/mpl/src/hier_rtlmp.cpp
AcKoucher: I see. Should I use boost or keep this code?

src/mpl/src/clusterEngine.cpp
AcKoucher: Well what we currently have is a cluster per constraint and a cluster for the unconstrained pins.
Bterms are linked to their cluster by the map `PhysicalHierarchyMaps::bterm_to_cluster_id`.

src/mpl/src/SimulatedAnnealingCore.h
AcKoucher: Should I update the methods with that kind of name as well? It seems like it would get very verbose (that's why I didn't write  'unconstrained' in the first place):

`searchForAvailableRegionsForPins() -> searchForAvailableRegionsForUnconstrainedPins()`

src/mpl/src/hier_rtlmp.cpp
maliberty: It is simple enough for now but if this gets more complex we could consider switching.  Also if runtime becomes an issue though there are few enough regions that is unlikely.

src/mpl/src/SimulatedAnnealingCore.h
AcKoucher: Any ideas?

src/mpl/src/clusterEngine.cpp
AcKoucher: Any further comment?

AcKoucher:
Running new Secure-CI.

src/mpl/src/SimulatedAnnealingCore.h
eder-matheus: Maybe `searchAvailableRegionsForUnconstrainedPins`? It's a bit verbose, but I guess it's good for consistency with the member name.

AcKoucher:
As I was addressing the last comment of the review I found a bug which is now fixed. I'm running a new Secure-CI.

eder-matheus (approved):
Waiting on secure-CI to merge.

AcKoucher:
Secure-CI had some failures which are addressed in [#3175](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3175).

maliberty (approved):

maliberty:
I've approved and will look more closely tomorrow's

PR #6824, opened 2025-03-08
ppl: use constraints, mirrored pins and pin groups stored in odb

eder-matheus:
This PR updates PPL and ODB to have pin constraints, pin groups and mirrored pins stored in ODB. The changes are described below.

ODB changes:
* The `set_io_pin_constraint` command implementation was moved from PPL to ODB.
* Added auxiliar functions in `odb.tcl` to add the constraints, pin groups and mirrored pins.
* In a pair of pins {p1, p2}, p1 have a pointer to p2, and p2 has the flag `is_mirrored` set as true. This way, we can detect which pins needs to be placed by the PPL solvers, and have the information if a pin has its position mirrored from other pin.
* Set the constraint regions of each dbBTerm at the moment the command is called. Also, the mirrored regions for mirrored pins are computed during this process.

PPL changes:
* Removed the list of pin groups from its internal structures. Instead, get the pin groups directly from odb::dbBlock.
* Removed map of mirrored pins and use the odb::dbBTerm APIs to get this information
* Build the list of pins constrained to the same region by iterating over the pins and using odb::dbBTerm APIs to get this information.


eder-matheus:

maliberty (changes requested):

maliberty (approved):

PR #6825, opened 2025-03-08
gui: add save histogram image

gadfort:
Changes:
- reorganized the ChartsWidget to ensure the histogramwidget owns all the display data to allow for easier integration with the save image code.
- added a save option to the mouse menu
- added access from tcl to save the image via `save_histogram_image`
- added hold timing to to bar chart

### DIFF
--- a/src/gui/README.md
+++ b/src/gui/README.md
@@ -139,6 +139,28 @@ save_clocktree_image
 |`-height`| height of the image in pixels, defaults to the height of the GUI widget. |
 |`-width`| width of the image in pixels, defaults to the width of the GUI widget. |
 
+### Save Timing Histogram Image
+
+This command saves the screenshot of timing histogram given options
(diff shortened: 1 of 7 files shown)


maliberty:
The default size is quite small and when zoomed in shows the labels don't fit

image

gadfort:
@maliberty the default just takes the size of the widget from the GUI, I can update it enforce the height is `== width` when not specified. (unfortunately this size if dependent on the state of your gui, so it's a bit difficult to replicate).  This mirrors the behavior of the clock widget, but maybe it's better to select a default like (500x500 or 1000x1000) which tend to generate reasonable images (and then ignore the widget size altogether).

maliberty:
Thanks.  It works fine aside from the default size.  I think have some constant like you suggest is best.

maliberty (approved):

PR #6827, opened 2025-03-10
drt: better dbInst to frInst translation

bnmfw:
Supports #5867.

This PR contains some minor changes to support the translation of dbInst data to frInst data. During incremental PA the GRT will only request changed to dbInst objects, while the DRT works only with frInst data. This makes changes to make this communication smoother.

#### `findInst()`
`findInst()` requires the name of the inst to retrieve the respective `frInst*`, another function of the same name, that expects a `dbInst*` instead of a string was created. This is very minor (1 single line of code + 1 import), but helps clean some other parts of the code up.

#### `io::Parser::setInst()`
The main change is that now the function returns the raw pointer of the created `frInst*`. Incremental PA is being build to support the passing of an instance not already initialized in the DRT. If this is the case, this will be done by the `io::Parser::setInst()` function. Returning the raw pointer of the inst helps dynamically creating the inst. Some other minor refactors to the function were also done.

### DIFF
--- a/src/drt/src/db/obj/frBlock.h
+++ b/src/drt/src/db/obj/frBlock.h
@@ -41,6 +41,7 @@
 #include "db/obj/frNet.h"
 #include "db/obj/frTrackPattern.h"
 #include "frBaseTypes.h"
+#include "odb/db.h"
 
 namespace drt {
 namespace io {
@@ -122,6 +123,7 @@ class frBlock : public frBlockObject
     }
     return nullptr;
   }
+  frInst* findInst(odb::dbInst* inst) { return findInst(inst->getName()); }
   frNet* getNet(int id) const
   {
     if (id >= nets_.size()) {
--- a/src/drt/src/io/io.cpp
+++ b/src/drt/src/io/io.cpp
@@ -123,34 +123,34 @@ void io::Parser::setTracks(odb::dbBlock* block)
   }
 }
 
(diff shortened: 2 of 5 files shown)


eder-matheus (approved):
Looks good to me. @osamahammad21 do you have any comments?

osamahammad21:
The part related to setInst is ok, I just don't think the findInst by dbInst pointer is necessary.

osamahammad21 (approved):

eder-matheus:
> The part related to setInst is ok, I just don't think the findInst by dbInst pointer is necessary.

I guess it is just a way to have a cleaner API to get a frInst from a dbInst. @bnmfw could confirm that, but I imagine that it will have impact in his future PRs.

bnmfw:
@osamahammad21 @eder-matheus
The reason for the returning of the pointer is the use case where I want to set a `dbInst` to the block and immediately have access to the set `frInst`. The function was designed to run for all insts at once when initializing the DRT. During incremental PA there exists a necessity for setting a `dbInst` and having immediate access to its `frInst`. Since the dbInst itself does not hold a reference to the new `frInst` (which I think could be fine) I would have to use the `findInst` API, which actually involves a search on a hash table. This would mean that getting the `frInst` I just set would be O(n) (seems to be the implemented find algorithm complexity) when I can just return the `frInst` and have it for free.

PR #6829, opened 2025-03-10
mpl: avoid creating blockages with zero depth and improve debug messages

AcKoucher:
... so that we know which type of overlap happened when the boundary pushed is reverted.

#### Context
As I was creating the tests for boundary pushing I ran into a weird overlap push reverting which should not happen.

#### Additional
 - Remove top module `Metrics` pointer (which is accessible through the `PhysicalHierarchy` and doesn't need caching) from `HierRTLMP` class and its wrong setter API from the `ClusteringEngine` class.

### DIFF
--- a/src/mpl/src/clusterEngine.cpp
+++ b/src/mpl/src/clusterEngine.cpp
@@ -94,11 +94,6 @@ void ClusteringEngine::run()
   }
 }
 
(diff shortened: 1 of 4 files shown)


AcKoucher:
Running Secure-CI

PR #6830, opened 2025-03-10
drt: dpl connection to drt-pa

bnmfw:
Supports #5867.

This PR giver the PA module access to the DPL. The is needed, as incremental PA will need to call the function implemented in #6522.

I'm currently at a point in incremental PA where I majorly have all the basic interfaces I need implemented, I'm trying to tie everything together and see what is missing to make a initial test. This connection is one of the missing parts.

### DIFF
--- a/src/drt/CMakeLists.txt
+++ b/src/drt/CMakeLists.txt
@@ -148,7 +148,8 @@ target_link_libraries(drt_lib
     OpenSTA
     utl_lib
     dst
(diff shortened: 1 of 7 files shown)


osamahammad21:
@bnmfw Could you point out why would you need to use this API of getting the adjacent cell? We do have a regionQuery in the detailed router where you could any instance at any location you want.

PR #6831, opened 2025-03-11
rmp: Adds support for Sky130

QuantamHD:
SKY130's constant cell is actually just one cell with two outputs one high and one low.


QuantamHD:
> unit test?

Added.

maliberty (approved):

PR #6834, opened 2025-03-11
mpl: fix io blockages' micron -> dbu conversion and add tests

### LINKED ISSUE #6821: mpl: add regression tests for boundary pushing step
### Description

This will help the development in #6820 

### Suggested Solution

Three cases:
Push left -> right
Push bottom -> top
Push left -> right && bottom -> top

### Additional Context

_No response_

AcKoucher:
Resolve #6821 

During the initialization of the Pusher class - which is the responsible for the boundary pushing step - we're messing up the creation of the dbu blockage by using the wrong mpl::Rect APIs.


AcKoucher:
Running Secure-CI

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -3094,10 +3094,10 @@ void Pusher::setIOBlockages(
 {
   for (const auto& [boundary, box] : boundary_to_io_blockage) {
     boundary_to_io_blockage_[boundary]
-        = odb::Rect(block_->micronsToDbu(box.getX()),
-                    block_->micronsToDbu(box.getY()),
```
maliberty, line 3098: The is a confusing API.  Perhaps getCenterX/Y would make it more obvious.

maliberty (approved):
Looks fine; awaiting CI

AcKoucher:
I got one problem in Secure-CI:
The number of hold violations in `ngt45/bp_fe_top` went from `167 -> 1354` violating the limit which is `309`.

The macro placement is slightly different, because the  2x4 array is not being pushed to the right edge anymore, as it is correctly identifying the pin access blockage that lies there. (There are only unconstrained pins in this block and they're excluded from the top edge, meaning left/bottom/right have pin access blockages).

image


Overall, the paths look relatively similar when comparing base/branch (I didn't compare many). However, paths that had their required hold time being met now are being violated due to a subtle increase in the required time:

image

The problem seems to be related with the capture path and that could indicate a problem caused by CTS. _However, apparently there are no hold violations just after CTS._

I'll investigate the evolution of the hold violations along the flow.

PR #6840, opened 2025-03-12
Python API: Add helper functions for tech

erendn:
Python API allows easier parsing of output designs. However, some values were confusing since the technology unit scales or technology corners can change depending on the design configuration. Instead of digging into technology files, I implemented some Python helper functions.


include/ord/Tech.h
```diff
@@ -71,6 +71,18 @@ class Tech
   odb::dbTech* getTech();
   sta::dbSta* getSta();
 
+  std::string nominalProcess();
+  std::string nominalVoltage();
+  std::string nominalTemperature();
```
maliberty, line 76: Why do these return strings rather than the underlying float type?

maliberty (commented):
Thanks for contributing!

maliberty:
Please fix the DCO error (see the details link).  In general you should sign your commits with `git commit -s`.

maliberty:
It would be good to add to the test/timing_api.py some simple coverage for the new APIs

include/ord/Tech.h
erendn: I was initially confused with how they were used somewhere else in OpenROAD. I forgot to change them later. I pushed the change to return float.

test/timing_api.py
```diff
@@ -33,3 +33,15 @@
             print("  ", mterm.getName())
             for m in timing.getTimingFanoutFrom(mterm):
                 print("    ", m.getName())
+
+tech.nominalProcess()
```
maliberty, line 37: Please print these values so that we can see what they are returning and add them to the .ok file.

erendn: Done 👍

maliberty (approved):

maliberty:
Please clang-format src/Tech.cc

erendn:
Thanks for guiding me through it. This is my first time contributing to OpenROAD.

Issue #6845, opened 2025-03-13
The decision to split macros from regs should rely on .LEF instead of insertion delay


AcKoucher:
### Description

Currently the decision to split macros from registers when creating the clock tree - decision which makes a tree be created only for macros and should help with QoR - relies on the presence of insertion delay which may not always exist.

### Suggested Solution

This decision should rely on the .LEF data by verifying if a certain cell is a block (macro).

### Additional Context

Being able to create a clock tree for macros can help solving problems for some specific situations:

When running Secure-CI for #6834, `nangate45/bp_fe_top` had a huge increase in the number of hold violation at the end of the flow (see more details [here](https://github.com/The-OpenROAD-Project/OpenROAD/pull/6834#issuecomment-2718770527)).

Some investigation showed that these violations are generated during detail route stage. However, one hypothesis was that, having CTS done a better job, perhaps detail route would not end up producing a large amount of hold violations.

Some testing using @arthurjolo changes eliminated the violations at the end of the flow apparently confirming that hypothesis.

PR #6846, opened 2025-03-13
CTS separate macros from registers even if macro has no insertion delay

arthurjolo:
Fixes [#6845](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6845)

As described on the issue CTS decides to separate macros into a different tree based on the presence of insertion delay. This PR changes that decision to made based on the .LEF data of the cell being a macro.


precisionmoon (approved):

maliberty:
Does this need a secure CI?

precisionmoon:
> Does this need a secure CI?

Yes.   Arthur said like 3 designs will need tightening of metrics.

maliberty:
If there are no fails then we can tighten after merging.

arthurjolo:
There is one failure on metrics, otherwise it is all green. I am going to create the PR to solve the metrics failures. For tighten  I saw that there were a some metrics changes in latter runs than mine. Should we merge the PR and then tighten or do it now?

maliberty:
I think you can leave the tightening off.  I've been discussing this with Vitor and I think we will move to a model where the label action only updates failing metrics and tightening is done directly on master to avoid conflicts.

arthurjolo:
Ok I will create the PR and link it here

PR #6847, opened 2025-03-13
gha: fixes to work on private repos

vvbandeira:


.github/workflows/black.yaml
```diff
@@ -6,6 +6,13 @@ jobs:
...
+      - name: Checkout Code
+        uses: actions/checkout@v3
+      - name: Run black (public repo)
+        if: "!github.event.repository.private"
+        uses: psf/black@stable
+      - name: Run black (private repo)
+        uses: psf/black@stable
+        if: github.event.repository.private
+        with:
+          version: "24.8.0"
```
maliberty, line 18: Why do we need to pin the version here?  If the public and private get out of sync that could be an issue.

vvbandeira: We need to upgrade the OS in the private. The current version does not have the latest black version available.

maliberty (approved):

.github/workflows/black.yaml
maliberty: We should plan for that.  I have noticed black versions to generate slightly different results.

PR #6850, opened 2025-03-14
mpl: renaming confusing mpl::Rect API

AcKoucher:


maliberty:
are there no callers?

AcKoucher:
> are there no callers?
No.

It seemed a bit useful and harmless to keep these APIs that's why I renamed instead of removing them. Should I go on and remove them?

maliberty:
I don't mind if it useful

maliberty (approved):

rovinski:
I wonder, why not template `odb::Rect` with default type of `int` and then you can just use `odb::Rect<float>`? or at least extend `odb::Rect<float>` as the parent class.

maliberty:
@rovinski really this code should have been written in dbu not microns.  I'd rather fix that.

rovinski:
Oh interesting, I thought I saw some PR traffic about trying to move from int to float because of the algorithm, and I assumed this was a part of that. I guess that makes more sense then.

PR #6861, opened 2025-03-17
Cts use only hard blockages for obstruction aware

arthurjolo:
CTS obstruction aware is currently considering the soft blockages and only if there are no soft blockages it is using the macros as hard blockages, this PR makes CTS consider only the hard blockages. To enable this this PR has 2 main changes:
### 
**1. Ignore soft blockages and only consider hard blockages and macros.**

This change is necessary because clock buffers should be allowed to be place inside a soft blockage if there are rows in it. When considering the soft blockage usefull space between macros is lost and can lead to degradation. 

### 
**2. Merge blockages into one if there is no space to place a buffer between them.**

This second change fixes a problem of CTS obstruction aware. If two blockages are to close and there is no space between the macros to place a buffer the current obstruction aware does not understand that and can place a buffer between those macros which will be moved by the laglizer after CTS.

To fix this this PR also adds a feature to obstruction aware that merges the blockages into one if there is not enough space between them, thus avoiding buffers to placed between them. The  image shows how the merging of the blockages is done.

Obs_aware_merging_blockage


arthurjolo:
I am looking into using boost polygon, I just need to understand better the merging of the polygons.

precisionmoon:
@arthurjolo, please address CTS regression failures in contiuous-integration tests.   Both pr-head and pr-merge have failures.

arthurjolo:
It took me some time but I was able to do it using `boost::polygon` and in fact it is much simpler :smile:. I am going to run a new secure-Ci just to make sure that everything is the same.

arthurjolo:
Secure-CI turned out to be the same as the previous implementation, the PR is ready to merge I am just going to create a PR to update the necessary metrics.

maliberty:
Please be aware of https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2921 in case you are adjusting the same designs.  That PR should automerge later today.

src/cts/test/check_buffers_blockages.ok
```diff
@@ -12,7 +12,7 @@
 [INFO CTS-0010]  Clock net "clk" has 288 sinks.
 [INFO CTS-0008] TritonCTS found 1 clock nets.
 [INFO CTS-0097] Characterization used 1 buffer(s) types.
-[INFO CTS-0200] 5 placement blockages have been identified.
+[INFO CTS-0201] 0 placed hard macros will be treated like blockages.
```
maliberty, line 15: There are hard blockages that are not attached to a macro.  Are they considered?

arthurjolo:
There is only one design with failing metric that will not be affected by that PR, but for the Tighten I will have to see.

src/cts/test/check_buffers_blockages.ok
arthurjolo: I forgot about them, I am adding them back.

maliberty:
I pushed some changes as it was faster than trying to explain what I had in mind.  Please see the last commit.

"5 placed hard macros will be treated like blockages." is a bit misleading as we also include hard blockages not associated with a macro

arthurjolo:
I changed the message to specify that we use both hard placement blockages and placed macros. I didn't separate the exact number of macros and placement blockages because I think CTS is already verbose, but it is possible to do so.

arthurjolo:
FYI these changes fixes the metric failure we have for the public CI.

src/cts/test/array.ok
```diff
@@ -12,7 +12,7 @@
 [INFO CTS-0010]  Clock net "clk" has 2475 sinks.
 [INFO CTS-0008] TritonCTS found 1 clock nets.
 [INFO CTS-0097] Characterization used 1 buffer(s) types.
-[INFO CTS-0200] 225 placement blockages have been identified.
+[INFO CTS-0201] 225 blockages from hard placement blocakges and placed macros will be used.
```
maliberty, line 15: typo blocakges -> blockages

maliberty (approved):

src/cts/src/TreeBuilder.cpp
```diff
@@ -95,6 +112,14 @@ void TreeBuilder::initBlockages()
     logger_->error(
         CTS, 77, "No physical master cell found for cell {}.", buffer);
   }
+
+  mergeBlockages();
+
+  logger_->info(CTS,
+                201,
+                "{} blockages from hard placement blockages and placed macros "
```
precisionmoon, line 120: How about this?
"Identified {} blockages from placed macros and hard blockage constraints"

PR #6864, opened 2025-03-17
gha: bump tj-actions/changed-files

vvbandeira:


vvbandeira:
FYI @sombraSoft @maliberty

maliberty (approved):

rovinski:
I assume you already saw https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/ ?

Perhaps this should be adjusted to using a commit ID rather than a tag.

maliberty:
Yes that motivated this pr

vvbandeira:
> I assume you already saw https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/ ?
> 
> Perhaps this should be adjusted to using a commit ID rather than a tag.

Using the tag might be enough; GH and tj-action removed the exploit from their history, and we automatically got the fix. The update here was not strictly necessary. If we use a commit hash and something like this happens again, we would need to update all branches to ensure the bad version does not run, while if critical fixes are directly applied to the tags, we get the benefit without any work.

rovinski:
@vvbandeira but using a tag in the first place was a problem. The attacker updated all tags to point to a compromised commit. People who used a specific commit ID were not affected, because the attacker cannot forge a commit. This type of attack does not work against commit hashes because it would require compromising the SHA-1 algorithm, which is still not feasible at this time.

If you keep using tags, it does nothing to prevent this attack from happening in the future. If you use a commit hash, then it does stop this kind of attack.

PR #6870, opened 2025-03-18
ppl/odb: add function to define excluded regions for pin placement

eder-matheus:
Necessary to remove random pin placement from ORFS.


src/odb/src/db/odb.tcl
```diff
@@ -795,6 +795,54 @@ proc set_io_pin_constraint { args } {
   }
 }
 
+sta::define_cmd_args "exclude_io_pin_region" { [-region region] }
```
maliberty, line 798: region looks to be a mandatory argument otherwise the command is a no-op.

maliberty: To be fixed in a follow on to unblock

PR #6875, opened 2025-03-19
odb: make -region mandatory for exclude_io_pin_region command

eder-matheus:
Follow up on https://github.com/The-OpenROAD-Project/OpenROAD/pull/6870#discussion_r2001907073

### DIFF
--- a/src/odb/src/db/odb.tcl
+++ b/src/odb/src/db/odb.tcl
@@ -803,6 +803,10 @@ proc exclude_io_pin_region { args } {
 
   sta::check_argc_eq0 "exclude_io_pin_region" $args
 
+  if { ![info exists keys(-region)] } {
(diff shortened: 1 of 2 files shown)


src/odb/src/db/odb.tcl
```diff
@@ -803,6 +803,10 @@ proc exclude_io_pin_region { args } {
 
   sta::check_argc_eq0 "exclude_io_pin_region" $args
 
+  if { ![info exists keys(-region)] } {
+    utl::error PPL 10 "The -region keyword is required for exclude_io_pin_region command."
```
maliberty, line 807: Remove the square brackets from "[-region region]" which indicate it is optional.  Similarly in the README

maliberty (approved):

PR #6880, opened 2025-03-19
drt: updated PA error messages

bnmfw:
@jeffng-or mentioned that some PA error messages were not too helpful due to not including description of the Intance/Master or InstTerm. Just added a few of those. Also created warn 88, which appears whenever some minimal condition of access points (only minimal number of sparse access points currently) could not be met even when considering access points of every cost.


src/drt/src/pa/FlexPA_acc_pattern.cpp
jeffng-or, line 1: We might want to keep the formatting consistent with other DRT messages. For example, DRT-0076's message is inst/pin, not inst/master. Also, is the pin name important?

So maybe change the message to:
```
"Inst pin {}/{} ({}) does not have any access point.",
unique_inst->getName(), 
pin_name
unique_inst->getMaster()->getName());
```
The same could be used for DRT-0090.

src/drt/src/pa/FlexPA_acc_point.cpp
jeffng-or, line 1: Exhaustive (missing h)

jeffng-or (changes requested):

src/drt/src/pa/FlexPA_acc_point.cpp
bnmfw: Ooops, thanks.

src/drt/src/pa/FlexPA_acc_pattern.cpp
maliberty: The pin name is useful as we check pin access first pin by pin and knowing which one failed is useful.

@bnmfw please address

bnmfw: This error call is actually a bit misleading. Currently it is only called for an instance if there is no access point within all of their pins. For this case I'm gonna change the error to be raised if _any_ pin is missing an access point.

maliberty:
@bnmfw looks like there is still some pending feedback and a merge confict

bnmfw:
I complied with the suggestions, Some errors actually are raised only for full Instances like error 90.
inst_term->getName() already returns the name in the format "instance/pin". I added the master name to the revamped errors.

I'm not completely sure what is the build problem with the CIs

bnmfw:
@maliberty Sorry for the delay, the other merged PR actually made more regressions fail due to the new warning, I though it was a CI problem, but this is ready to merge

src/drt/src/pa/FlexPA_acc_pattern.cpp
```diff
@@ -244,7 +244,11 @@ int FlexPA::prepPatternInstHelper(frInst* unique_inst, const bool use_x)
       }
     }
     if (n_aps == 0 && !inst_term->getTerm()->getPins().empty()) {
-      logger_->error(DRT, 86, "Pin does not have an access point.");
+      logger_->error(DRT,
+                     86,
+                     "Term {} ({}) does not have any access point.",
+                     inst_term->getName(),
+                     unique_inst->getMaster()->getName());
```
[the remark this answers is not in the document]
jeffng-or: In my case, I was having to add additional masters to the DONT_USE_CELLS list to resolve the access point error. So, having the master name in the message saves me from having to pull up OR and find the instance's master. I can just pull the name from the message and try again.

Issue #6097, opened 2024-11-05
DRC violations due to routing to strict about following tracks when accessing pins


titan73:
### Describe the bug

After 18 iterations, there are 40 similar drc violations still not fixed. 

M2 route follow track when switching to M1 causing DRC. 

![or_pin](https://github.com/user-attachments/assets/70dcb907-ee27-4e7a-8776-01929ade9c98)

Pin access in Innovus

![inn_pin](https://github.com/user-attachments/assets/6ba57ace-b26b-4868-a22a-1527c9633469)

### Expected Behavior

M2 should not strictly follow track if possible to fix DRC.

### Environment

```shell
v2.0-16855-ga15114a95
```

### To Reproduce

openroad test.tcl with test case below:

[drc_pin_access.zip](https://github.com/user-attachments/files/17636330/drc_pin_access.zip)

maliberty:
Please don't include proprietary tool output.

stefanottili:
On the picture it looks as if your rows are not aligned with the M1/M2 tracks.
I that leads to the stdcell pins not being aligned either and should be the root cause of pin access issues.

osamahammad21:
@stefanottili I agree with you, but I don't understand why pin access did not complain about this earlier in the flow. I'll investigate further.

titan73:
> In the picture it looks as if your stdcell sites are not a multiple of the routing pitch. This is very unusual and leads to none of the stdcell pins being "on track".

Yes indeed, the sites for this library is not multiple of the pitch. This library is badly designed. All other sites for other libraries in this techno are correct.

maliberty:
Do you need to use the badly designed library?  I'm not saying we shoudn't look at it but it also seems like its making things unnecessarily hard.

titan73:
Yes. There is not replacement for this standard cells library (215 cells). It's a mature techno with a lot of products based on it. They unfortunately won't redesign these cells.

titan73:
Btw, apart from the site width not multiple of pitch, if a library has the correct site but the cells pins are not or badly aligned, that will cause the same issues in OR, right?

stefanottili:
Why don’t you use a pitch that matches the design of the stdcells ?

titan73:
> Why don’t you use a pitch that matches the design of the stdcells ?

[...]

PR #6885, opened 2025-03-20
drt: genPinAccessCostBounded refactor and EnoughAccessPoints

bnmfw:
Supports #6097.

This PR is a first of a pair, the second one actually solves the issue. This PR adds the `EnoughAccessPoints()` function, in which the condition for having at least a single access point far enough from the instance edge will be implemented.
I'm splitting this in two PRs because the part that adds the new requirement actually changes some unit test .DEFOK files. This PR is just refactoring to isolate this part of the code. I want it to be clear that none of these changes are what are changing the .DEF files.

### DIFF
--- a/src/drt/src/pa/FlexPA_acc_point.cpp
+++ b/src/drt/src/pa/FlexPA_acc_point.cpp
@@ -1110,7 +1110,7 @@ void FlexPA::filterMultipleAPAccesses(
 
 template <typename T>
 void FlexPA::updatePinStats(
-    const std::vector<std::unique_ptr<frAccessPoint>>& tmp_aps,
+    const std::vector<std::unique_ptr<frAccessPoint>>& new_aps,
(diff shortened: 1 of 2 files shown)


src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -1229,13 +1250,17 @@ bool FlexPA::genPinAccessCostBounded(
...
     // IO term pin always only have one access
     for (auto& ap : aps) {
       pin->getPinAccess(0)->addAccessPoint(std::move(ap));
     }
     return true;
   }
+
+  // weird edge case where pin is not from std_cell, macro or io, not sure it
+  // can even happen
   return false;
```
osamahammad21, line 1264: Not possible. through an error instead.

osamahammad21 (approved):

PR #6889, opened 2025-03-20
drt: pa far from edge

### LINKED ISSUE #6097: DRC violations due to routing to strict about following tracks when accessing pins

bnmfw:
Fixes #6097. Almost.
Secure and ISPD CI all passed, the only violations are on regression tests. Ibex_sky130hd has 1 metric failure:
```
Pin                                     Limit      Cap    Slack
---------------------------------------------------------------
_23420_/Y                               0.092    0.094   -0.001 (VIOLATED)
```
and jpeg_sky130hd as well as aes_sky130hd have 1 antenna violation each.

This PR introduces the soft requirement that a pin has an Access Point at least 2 times the layer width away from the instance edge. This seems to help the issue, but some investigation has to be done to find a less empirical value.


bnmfw:
There are actually some other metrics that differ on the failed designs that were updated.

maliberty:
Has this been through a secure CI?

maliberty:
The code changes look fine to me

bnmfw:
> Has this been through a secure CI?

Yes, no problem neither in Secure or ISPD CI

maliberty (approved):

PR #6891, opened 2025-03-21
Update mbff.cpp to use max threads in the solver.

mikesinouye:
The default value for `num_workers` is 0 which implies all threads on the machine, so this is a no-op. This declares this behavior explicitly.


maliberty (approved):

vvbandeira:
@mikesinouye 
For some reason, the `gpl` test `clust02.tcl` is getting into an infinite loop and not finishing after your change. Could you try to run the test locally to reproduce the error?

PR #6897, opened 2025-03-23
gpl: remove stay files that should not be there

gadfort:
No functional changes, just removing two test output files


maliberty (approved):

maliberty:
I'm not sure how these files got generated.  It might be good to add to the .gitignore

gadfort:
@gudeh committed the files, so maybe they would be best suited to put the gitignore together

gudeh:
I believe I did something like `ctest --test-dir ../../../build/ -R simple01-td` from `/src/gpl/test/`. I tried reproducing but I could not leave the files almost empty like in the mistaken commit.

maliberty:
FYI, you could do `regression -R simple01-td` as it will invoke ctest with the test-dir.

PR #6898, opened 2025-03-24
DRT: fix overflow

osamahammad21:
- Fix overflow in calculating the closest point in the maze search points to the patch origin added by GC. A very far away point would be considered the closest because it would have a -ve path length.

### DIFF
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -51,6 +51,7 @@
 #include "io/GuideProcessor.h"
 #include "io/io.h"
 #include "odb/dbShape.h"
+#include "ord/OpenRoad.hh"
 #include "pa/AbstractPAGraphics.h"
 #include "pa/FlexPA.h"
 #include "rp/FlexRP.h"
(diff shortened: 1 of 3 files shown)


src/drt/src/TritonRoute.cpp
```diff
@@ -307,17 +308,7 @@ void TritonRoute::resetDb(const char* file_name)
...
-          DRT,
-          9947,
-          "You can't load a new db file as the db is already populated");
-    }
-
-    stream.exceptions(std::ifstream::failbit | std::ifstream::badbit
-                      | std::ios::eofbit);
-
-    db_->read(stream);
+    ord::OpenRoad::openRoad()->readDb(stream);
```
QuantamHD, line 311: No new dependence on the OpenROAD object should be added. 

@maliberty

maliberty: I'll tackle this in a separate PR and it will be removed from this PR

maliberty (approved):

osamahammad21:
> grt.congestion1.py failed

This is probably a non-deterministic behavior in GRT. I am gonna rerun the job and it's probably gonna pass. @eder-matheus Could you take a look at this [run](https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-6898-head/2/pipeline-graph/)?

maliberty:
pr-merge passed so it does look to be a fluke

PR #6900, opened 2025-03-24
gpl: include Testing folder to gitignore

gudeh:
As suggested in https://github.com/The-OpenROAD-Project/OpenROAD/pull/6897

The folder was accidentally added on a [commit to update gpl tests](https://github.com/The-OpenROAD-Project/OpenROAD/commit/250e48e05383b78791663ae7105b572ae860acd8).

### DIFF
--- a/src/gpl/.gitignore
+++ b/src/gpl/.gitignore
@@ -1,3 +1,4 @@
 *~
 test/results
+test/Testing
 *.jpg


maliberty:
Do you know how to reproduce it?  I'm wondering if it could happen outside of gpl

gudeh:
Unfortunately I could not reproduce neither find it in my command history. I believe I did an incorrect ctest call.

maliberty (approved):

gudeh:
I just noticed. I was not able to run an upf test at `tools/OpenROAD/test/` with `./regression -R upf` neither with `./regression upf_test`. This one seems to work: `ctest --test-dir ../build -R upf_test`. After these commands I noticed a directory showed up as: `tools/OpenROAD/test/Testing/`. I removed the directory and tried re running the commands I previously did, but the folder did not show up again

gudeh:
Funny, I did an incorrect call: `ctest --test-dir ../build -R upf_aef` (aef instead of aes). Than I did: `ctest --test-dir ../build -R upf_aes`. It seems one of these two created the folder `/tools/OpenROAD/test/Testing`. I removed the folder, ran the commands again, but this time the folder did not show up.

PR #6901, opened 2025-03-24
Cts change blockage merging

arthurjolo:
PR [#6861](https://github.com/The-OpenROAD-Project/OpenROAD/pull/6861) introduced a blockage merging for obstruction aware, however with the API used the blockage merging was not done as expected.

The current code performs as showed by the example on the top of the image, with the merged blockage as 2 rectangles, this behavior could bring problems to CTS, we want a behavior like on the bottom of the image with the merge blockages as the bounding box of the original blockages. This PR fixes this by using the correct API to get the merged blockage.

Also this PR adds a unity test to test the blockage merging.

Obs_aware_merging_blockage_fix drawio


src/cts/test/CMakeLists.txt
```diff
@@ -7,12 +7,10 @@
 or_integration_tests(
   "cts"
   TESTS
-    array
-    array_ins_delay
-    array_no_blockages
```
gadfort, line 12: why are these removed?

arthurjolo: They should not be removed, I added them back.

maliberty (approved):

maliberty:
Have you run a secure CI?

arthurjolo:
I haven't run a secure-CI, because it shouldn´t alter any results. But I will run one just to be sure.

maliberty:
any update?

PR #6910, opened 2025-03-25
DPO: use DbuX and DbuY

osamahammad21:


src/dpo/src/Optdp.cpp
```diff
@@ -215,7 +215,7 @@ void Optdp::updateDbInstLocations()
       const Node* nd = it_n->second;
 
       const int y = nd->getBottom() + grid_->getCore().yMin();
-      const int x = nd->getLeft() + grid_->getCore().xMin();
+      const int x = nd->getLeft().v + grid_->getCore().xMin();
```
maliberty, line 218: This is a case where the variable `x` should be converted to use DbuX rather than downconverting from DbuX to integer.  I see a fair amount of use of .v() where it feels like the direction of change should be the opposite.

osamahammad21: This specific case is intended to be int because this x is going to be passed to the db object. in this function we are translating from DPO to ODB. So no need to make it DbuX and convert it back to int.
Other uses of .v are mostly when we are using doubles and multiplying by 0.5. Once that is resolved, it should bring down those occurences.

maliberty (approved):

osamahammad21:
@eder-matheus Another [occurence](https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-Public/detail/PR-6910-head/3/pipeline) for the non-deterministic grt behavior

PR #6913, opened 2025-03-26
odb/defin: Redirect defWarning to the logger.

VoltXd:
Also changed error callback ID from 1003 to 3, to avoid duplicate.
    As find message doesn't catch message without double quote.


maliberty (approved):

maliberty:
Thanks!

maliberty:
For the future, you need to format the code with `clang-format -i`.  I've done it for this PR.

PR #6916, opened 2025-03-26
drt: computeInstRows

bnmfw:
Supports #5867.

### `frInstLocationSet`
Creates a custom set of frInsts* that is sorted by origin. A member variable insts_set_ was created to hold these instances.

### `buildInstsSet`
`GetInsts` was renamed to `buildInstsSet` and now builds the insts_set_ member variable instead of build a frInst* vector it recieved as input.

### `computeInstsRows`
Moves `prepPattern()` and some dependencies back to FlexPA.cpp while moving parts of its code to `computeInstRows()` which will be used in incremental PA. The structure is built upon the insts_set_ instead of a sorted vector of insts.


bnmfw:
@maliberty I don't understand why this is failing. There is a "Process completed with exit code 1" on the build of clang-tidy

maliberty:
I reran clang-tidy - there seems to have been a short lived glitch in GH.

bnmfw:
Reviewing this commit by commit is easier than all at once

src/drt/src/pa/FlexPA.cpp
```diff
@@ -178,6 +196,69 @@ void FlexPA::prep()
   prepPattern();
 }
 
+void FlexPA::prepPattern()
```
osamahammad21, line 199: is there a reason you're moving this and serializePatterns here?

osamahammad21 (approved):

PR #6932, opened 2025-03-28
bazel: OpenROAD has the canonical Bazel setup

oharboe:
megaboom that started first will copy-cat the setup and the canonical home for the documentation will
be in OpenROAD.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,4 +7,8 @@ build --host_cxxopt "-ffp-contract=off"
 build --crosstool_top=@llvm_toolchain//:toolchain
 build --incompatible_strict_action_env
 
-try-import %workspace$/user.bazelrc
+# Read only remote cache, updated by OpenROAD team members and CI.
+build --remote_cache=https://storage.googleapis.com/megaboom-bazel-artifacts
+build --remote_cache_compression=true
+
+try-import %workspace$/.user.bazelrc
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ transceivers, OpenPower-based Microwatt etc.
 
 ## Build OpenROAD
 
-To build OpenROAD tools locally in your machine, follow steps
+To build OpenROAD tools locally on your machine, follow steps
 from [here](docs/user/Build.md).
 
 ## Regression Tests
--- a/docs/user/Bazel.md
+++ b/docs/user/Bazel.md
@@ -1,0 +1,51 @@
+# Building with Bazel
+
+
+## Using the OpenROAD project Bazel artifact server to download pre-built results
+
+A single read only artifact server is configured in `.bazelrc` OpenROAD hosted projects.
+
+This is a read only Bazel artifact server for anonymous access and is normally only updated by OpenROAD CI, though team OpenROAD team members can also update it directly.
+
+## OpenROAD team member and CI - configuring write access to artifact server
+
+If you only have a single Google account that you use for Google Cloud locally, you can use
+`--google_default_credentials`.
+
+If you are use multiple google accounts, using the default credentials can be cumbersome when
+switching between projects. To avoid this, you can use the `--credential_helper` option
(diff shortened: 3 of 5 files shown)


maliberty (approved):

.bazelrc
```diff
@@ -7,4 +7,8 @@ build --host_cxxopt "-ffp-contract=off"
 build --crosstool_top=@llvm_toolchain//:toolchain
 build --incompatible_strict_action_env
 
-try-import %workspace$/user.bazelrc
+# Read only remote cache, updated by OpenROAD team members and CI.
```
hzeller, line 10: so as a regular user I get error messages such as this:

```
WARNING: Remote Cache: 4 errors during bulk transfer:
com.google.devtools.build.lib.remote.http.HttpException: 403 Forbidden
<?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>Anonymous caller does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).</Details></Error>
com.google.devtools.build.lib.remote.http.HttpException: 403 Forbidden
<?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>Anonymous caller does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).</Details></Error>
com.google.devtools.build.lib.remote.http.HttpException: 403 Forbidden
<?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>Anonymous caller does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).</Details></Error>
com.google.devtools.build.lib.remote.http.HttpException: 403 Forbidden
<?xml version='1.0' encoding='UTF-8'?><Error><Code>AccessDenied</Code><Message>Access denied.</Message><Details>Anonymous caller does not have storage.objects.create access to the Google Cloud Storage object. Permission 'storage.objects.create' denied on resource (or it may not exist).</Details></Error>
```

Wouldn't it be better if we leave this out and ask users who have access to put the next two lines in their `.user.bazelrc` ?

(also I don't quite like the rename to `.user.bazelrc` as this make the file harder to stumble over when debugging an issue). Common is to use a visible `user.bazelrc`.

oharboe: I thought this repository was already available as anonymous read only. Would fixing that be satisfactory?

I have no strong opinion on user.bazelrc having a . or not infront, I was just trying to stay consistent. We chose .user-bazelrc in megaboom, ORFS and bazel-orfs. Whatever we pick it should be consistent.

Also, whatever we pick, it should be in .gitignore.

hzeller: If there can be an anonymous access to the cache so that anyone can read from it, then that would be sufficient indeed. I am wondering how bazel knows that this is only readonly here ? Maybe my bazel attempted to push something and thus ran into the error.

Dot style for `user.bazelrc`: also not a strong opinion,  if this is consistent with the other projects, then keep it with dot-prefix (I have observed/used non-dot prefix  in various projects such as [verible](https://github.com/chipsalliance/verible/blob/master/.bazelrc#L52), [xls](https://github.com/google/xls/blob/main/.bazelrc#L104), [bant](https://github.com/hzeller/bant/blob/main/.bazelrc#L46), [re2](https://github.com/google/re2/blob/main/.bazelrc#L20), and others derived from the [best practice section](https://bazel.build/configure/best-practices#bazelrc-file) in the bazel documentation.)

oharboe: @vvbandeira @maliberty Can you set up read only anonymous access to the bucket?

I didn't realize/spot there was a best practice on naming in bazel.

Change so as to stick to [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)?

PR #6934, opened 2025-03-28
Initial switch to bzlmod, moving most dependencies there.

hzeller:
Most dependencies are now pulled from bzlmod, but there are a few remaining that are still pulled from bazel_rules_hdl, fetched in WORKSPACE. Minimal change in that file for now, but next step would be to prune this file and _only_ pull in the dependencies only needed from there (such as lib ABC).

There is one issue with boost.beast in which the provided library does not seem to provide all link targets and fails to link. Deferred to later, for now, use the old `@boost//:beast` (from bazel_rules_hdl) in `src/utl/BUILD`.


hzeller:
I found the issue with `@boost.beast`. I am sending a pull request upstream to BCR, so once that is merged we then can also switch that dependency ( https://github.com/bazelbuild/bazel-central-registry/pull/4174 )

Independent of that, this PR is ready to merge.

oharboe:
@hzeller Do I need OpenROAD to br fully converted to MODULE.bazel before I can make progress with https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3008 ?

PR #6935, opened 2025-03-28
Fix unnecessary const qualified return values. Make some cons refs.

hzeller:
Returning a _value_ 'const' does not actually have any effect, it
just makes the reader wonder why that is (the compiler spits
out a ton of warnings about this).
    
Looking at this it turns out that some expensive types such as
strings, vectors, or maps have been returned by value which is
expensive.
    
Replace these with const reference for simple things (such as
strings), but left containers to be returned by value as it
requires first to analyse if the call sites, while iterating
through it, don't add more components, invalidating said
iterators. Added comment to that effect.


hzeller:
the one test-failure looks like a fluke, but I also don't have access to the CI to fix or re-run.

PR #6939, opened 2025-03-29
clang-tidy: fix a few missing includes

hzeller:
Fix some includes missing, found with `etc/run-clang-tidy-cached.cc`, then adding the relevant header.

First, finding candiates
```
awk '/is directly included.*misc-include-cleaner/ {print $6}' OpenROAD-bazel_clang-tidy.out | sort | uniq -c | sort -n
```

Then, say we want to fix every file that needs the header for `std::string` add this semi-automatically. We still need to know what header that is, then finding the list of files that need that header from the `clang-tidy` output and using a [little tool to insert headers](https://github.com/hzeller/dev-tools/blob/main/insert-header.cc), folloed by a `clang-format`:

```
../dev-tools/insert-header.cc '<string>' $(awk -F: '/no header providing "std::string".*misc-include-cleaner/ {print $1}' OpenROAD-bazel_clang-tidy.out | sort | uniq)
clang-format -i $(awk -F: '/no header providing "std::string".*misc-include-cleaner/ {print $1}' OpenROAD-bazel_clang-tidy.out | sort | uniq)
```

Four symbols, four commits: for `std::string`, `std::string_view`, `std::pair` and `size_t`.


maliberty (approved):

maliberty:
I approved as this is conceptually fine but you do have some clang-format to fix.  You also need to fix the templates for odb not the generated files.  clang-tidy will be impossible for this PR as it has a max size limit on the action.

hzeller:
`clang-format` I have already done on each of the files.

For now, I have re-run the generator and did not fix the includes in the generator. Looks like the script is written in Python, that would burn my eyes out :)

maliberty:
You don't actually need to read the python, just updating the Jinja templates would do it.  I can take a look later.

hzeller:
The set of needed includes are different in some of these situations, so I suspect the script needs to keep track of if ever a `std::string` or `std::pair` is used and only then emit the relevant header.

maliberty:
More than a few 😉

PR #6945, opened 2025-03-29
Move `//:opendb_def` -> `//src/odb/src/def:def`

hzeller:
Move the BUILD for this target more closely to the sources.

While at it, run `buildifier` on all BUILD files
to get a canonical formatting (`buildifier` is
for BUILD files what `clang-format` is for c++ files).


src/odb/src/def/BUILD
```diff
@@ -0,0 +1,119 @@
+# -*- Python -*-
```
maliberty, line 1: Please put at least the minimal license header

maliberty (approved):

PR #6948, opened 2025-03-30
odb: Adds basic support for non-rectangular floorplans

QuantamHD:
This PR adds support for polygonal floorplans. This PR lacks support for pin placement and specifying a polygon through the initialize floorplan command.

The current intended use case is that the user will provide a completed floorplan def that is polygonal with an existing pin placement.

See example floorplan
image


src/odb/src/db/dbBlock.cpp
```diff
@@ -2089,7 +2098,65 @@ void dbBlock::setDieArea(const Rect& new_area)
...
+        obs->setIsVirtual(true);
+      }
+    }
+  }
+}
+
 Rect dbBlock::getDieArea()
+{
+  _dbBlock* block = (_dbBlock*) this;
+  return block->_die_area.getEnclosingRect();
```
maliberty, line 2156: I think this should become `getDieBBox()` and `getDieAreaPolygon()` should be `getDieArea()`.

src/odb/test/cpp/TestPolygonalFloorplan.cc
```diff
@@ -0,0 +1,63 @@
+// Copyright 2023 Google LLC
```
maliberty, line 1: 2025

src/odb/test/cpp/TestPolygonalFloorplan.cc
```diff
@@ -0,0 +1,63 @@
...
+#include "nangate45_test_fixture.h"
+#include "odb/db.h"
+#include "odb/defin.h"
+#include "utl/Logger.h"
+
+namespace odb {
+
+using ::testing::HasSubstr;
+
+TEST_F(Nangate45TestFixture, AbstractLefWriterMapsTieOffToSignal)
```
maliberty, line 24: This is a strange test name - copy-paste error?

src/odb/test/cpp/nangate45_test_fixture.h
```diff
@@ -0,0 +1,39 @@
...
+
+  void SetUp() override
+  {
+    db_ = OdbUniquePtr<odb::dbDatabase>(odb::dbDatabase::create(),
+                                        &odb::dbDatabase::destroy);
+    odb::lefin lef_reader(
+        db_.get(), &logger_, /*ignore_non_routing_layers=*/false);
+    lib_ = OdbUniquePtr<odb::dbLib>(
+        lef_reader.createTechAndLib(
+            "sky130", "sky130", "Nangate45/Nangate45.lef"),
```
maliberty, line 29: Why is sky130 mixed with ng45?

src/odb/test/cpp/TestPolygonalFloorplan.cc
```diff
@@ -0,0 +1,63 @@
...
+  // There's essentially 1 rectangle cut out of it.
+  dbSet<dbBlockage> blockages = block->getBlockages();
+  std::vector<dbBlockage*> virtual_blockages;
+  for (dbBlockage* blockage : blockages) {
+    if (!blockage->isVirtual()) {
+      continue;
+    }
+    virtual_blockages.push_back(blockage);
+  }
+  EXPECT_EQ(virtual_blockages.size(), 1);
```
maliberty, line 48: It would be good to check its dimensions.  Same below

src/ifp/test/init_floorplan_polygon.tcl
```diff
@@ -0,0 +1,10 @@
+# init_floorplan called after a voltage domain is specified
```
maliberty, line 1: Wrong description.

Have you verified the routing tracks are correctly inside the polygon boundary?

src/gui/src/renderThread.cpp
```diff
@@ -813,6 +814,10 @@ void RenderThread::drawBlockages(QPainter* painter,
     if (restart_) {
       break;
     }
+    if (blockage->isVirtual()
+        && !viewer_->options_->areVirtualObstructionsAndBlockagesVisible()) {
```
maliberty, line 818: Hoist this out of the loop into a const variable.  Likewise drawObstructions

maliberty (changes requested):

src/odb/include/odb/geom.h
```diff
@@ -337,15 +337,28 @@ class Polygon
   Polygon(const Oct& oct);
 
   std::vector<Point> getPoints() const;
+
+  const std::vector<Point>& getPointsConstReference() const;
+  std::vector<Point>& getPointsMutable();
```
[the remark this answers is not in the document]
QuantamHD: Boost geom requires these methods in order to be interoperable. See the mapping in geom_boost.h

src/odb/include/odb/geom.h
```diff
@@ -358,6 +371,7 @@ class Polygon
 
  private:
   std::vector<Point> points_;
+  std::vector<std::vector<Point>> interior_points_;  // always empty
```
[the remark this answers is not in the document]
QuantamHD: Boost geom requires refs for exterior and interior polys. We need something to reference.

src/gui/src/displayControls.cpp
```diff
@@ -527,6 +527,10 @@ DisplayControls::DisplayControls(QWidget* parent)
   makeLeafItem(
       misc_.manufacturing_grid, "Manufacturing grid", misc, Qt::Unchecked);
   makeLeafItem(misc_.gcell_grid, "GCell grid", misc, Qt::Unchecked);
+  makeLeafItem(misc_.virtual_obstructions_and_blockages,
+               "Virtual Obstructions & Blockages (Debug)",
+               misc,
+               Qt::Unchecked);
```
gadfort, line 533: lets not add a debugging view to the main viewer, this should be a separate renderer

src/gui/src/displayControls.h
```diff
@@ -397,6 +398,7 @@ class DisplayControls : public QDockWidget,
     ModelRow module;
     ModelRow manufacturing_grid;
     ModelRow gcell_grid;
+    ModelRow virtual_obstructions_and_blockages;
```
gadfort, line 401: remove

src/odb/include/odb/db.h
```diff
@@ -4103,6 +4113,21 @@ class dbObstruction : public dbObject
...
+  /// by that difference are then decomposed into rectangles which
+  /// create obstructions with the virtual annotation.
+  ///
+  bool isVirtual();
+
+  ///
+  /// Sets this obstruction as virtual.
+  ///
+  void setIsVirtual(bool is_virtual);
+
```
gadfort, line 4130: this feels really hacky to me. If the issue is that the various tools dont support this, then we should create issues and assign them out to ensure the tools can handle this properly.

src/odb/include/odb/db.h
```diff
@@ -4159,6 +4184,21 @@ class dbBlockage : public dbObject
...
+  /// by that difference are then decomposed into rectangles which
+  /// create blockages with the virtual annotation.
+  ///
+  bool isVirtual();
+
+  ///
+  /// Sets this blockage as virtual.
+  ///
+  void setIsVirtual(bool is_virtual);
+
```
gadfort, line 4201: same

src/odb/include/odb/geom_boost.h
```diff
@@ -280,48 +280,42 @@ struct tag<odb::Polygon>
...
-  using type = std::vector<std::vector<odb::Point>>;
+  using type = std::vector<std::vector<odb::Point>>&;
 };
 
 template <>
 struct exterior_ring<odb::Polygon>
 {
-  static std::vector<odb::Point> get(odb::Polygon& p) { return p.getPoints(); }
-  static const std::vector<odb::Point> get(const odb::Polygon& p)
+  static auto& get(odb::Polygon& p) { return p.getPointsMutable(); }
```
gadfort, line 307: what is making us have to support this? Since it wasn't needed before

src/odb/src/db/dbBlock.cpp
```diff
@@ -2089,7 +2098,65 @@ void dbBlock::setDieArea(const Rect& new_area)
...
+  for (odb::Polygon& blockage_area : results) {
+    std::vector<Rect> blockages;
+    decompose_polygon(blockage_area.getPoints(), blockages);
+    for (odb::Rect& blockage_rect : blockages) {
+      dbBlockage* db_blockage = dbBlockage::create(this,
+                                                   blockage_rect.xMin(),
+                                                   blockage_rect.yMin(),
+                                                   blockage_rect.xMax(),
+                                                   blockage_rect.yMax());
+      db_blockage->setIsVirtual(true);
```
gadfort, line 2133: I don't think this is the correct solution to deal with tool support. ifp, pdn, grt, drt seems like these would be the only ones that are sensitive to this and we should just invest the time to fix them without this.

gadfort (changes requested):

src/odb/src/db/dbBlock.cpp
QuantamHD: I think I disagree. I think for all the tools you listed this is essentially what they have to do anyways.

Each of their tooling support relies on deriving blockage information from odb. We could duplicate that functionality in each of them with a helper API, but this feels like this solves the problem for most tools without them having to know about it.

It doesn't fully solve the problem for ifp and ppl. Which will need to be made directly aware long term.

src/odb/include/odb/db.h
QuantamHD: I feel like this is a reasonably good solution that covers most tools requirements for awareness here both now and going forward.

Some tools will need to be more aware, but for most this seems sufficient. The only time virtual comes up as necessary is in the GUI and in the lef and def writers. Where we do take on some kludge.

But it feels like we more often deal with tools that will not care about than virtual annotation than those who do like LEF/DEF/GUI. All of which have rather minor additional if conditions.

src/odb/include/odb/geom_boost.h
QuantamHD: `boost::geometry::difference(bounding_rect, block->_die_area, results);` Specifically has a template that says these methods must be references and const references in order to support swap operations. 

Returning these by value I guess is technically incorrect https://stackoverflow.com/questions/67204323/how-to-adapt-a-custom-polygon-type-in-boost-geometry

src/odb/include/odb/geom.h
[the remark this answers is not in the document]
QuantamHD: boost::geometry::difference(bounding_rect, block->_die_area, results); Specifically has a template that says these methods must be references and const references in order to support swap operations.

Returning these by value I guess is technically incorrect https://stackoverflow.com/questions/67204323/how-to-adapt-a-custom-polygon-type-in-boost-geometry

src/odb/src/db/dbBlock.cpp
QuantamHD: After talking with Peter below I think you're right. I'll change it.

src/odb/include/odb/geom.h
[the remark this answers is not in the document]
maliberty: This is not a correct solution and we can't guarantee it will be empty in general.  The real problem is that Polygon is not a boost polygon (from [concept_polygon](https://www.boost.org/doc/libs/1_87_0/libs/geometry/doc/html/geometry/reference/concepts/concept_polygon.html))
```
A polygon of Boost.Geometry is a polygon with or without holes. (A polygon without holes is a helper geometry within Boost.Geometry, and referred to as a ring.)
```
It is a ring.  I think we should remove the attempt to make it seem like one and you should use the boost classes instead for computing the blockages.

src/odb/include/odb/geom.h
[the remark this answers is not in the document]
maliberty: remove this attempt to fit into the boost API

src/odb/include/odb/db.h
maliberty: @gadfort I suspect most tools will wind up doing something similar internally as they naturally want a rectangle to work in.  The 'virtual' name isn't great though it is hard to come up with a nicer one.

src/ifp/test/init_floorplan_polygon.tcl
QuantamHD: image

They are not cut. Does this matter if there are routing obstructions? Or do we want to cut them similar to the standard cell rows?

maliberty: I suppose it won't matter as those areas are 'virtually obstructed'.

maliberty:
I am not sure it is worth having a 'virtual' flag on the blockage/obstructions.  You could simply create them as normal and in def-out just skip writing any that are outside the die outline.  @gadfort does that seem better to you?

gadfort:
I would love to not have the virtual flag, but there are a couple of issues I see:
1. the GUI should also honor this behavior as I don't think it should be visible in the GUI.
2. These blockages really cannot be removed, so they need away to stop users from destroying them (this is missing at the moment @QuantamHD)

QuantamHD:
@gadfort @maliberty I might be missing something, but it looks like there's currently no way to delete blockages or obstructions. Is that correct?

gadfort:
`odb::dbObstruction_destroy` can be called from tcl (same for Blockage)

maliberty:
> I would love to not have the virtual flag, but there are a couple of issues I see:
> 
> 1. the GUI should also honor this behavior as I don't think it should be visible in the GUI.
> 2. These blockages really cannot be removed, so they need away to stop users from destroying them (this is missing at the moment @QuantamHD)

In that case I think it makes sense to flag them in odb as various parts of the system need this and it needs to be reasonably fast for the GUI.  Perhaps instead of isVirtual we could use isExteriorToDieArea with an explanatory comment?

QuantamHD:
> `odb::dbObstruction_destroy` can be called from tcl (same for Blockage)

Hmm I'm not seeing the destroy method for blockages

```c++
///
/// A dbBlockage is the element that represents a placement blockage in a block.
///
class dbBlockage : public dbObject
{
 public:
  ///
  /// Get the bbox of this blockage.
  ///
  dbBox* getBBox();

  ///
  /// Get the instance associated with this blockage.
  /// Returns nullptr of no instance was associated with this blockage
  ///
  dbInst* getInstance();

  ///
  /// Declare this blockage to have been pushed into this block.
  ///
  void setPushedDown();

  ///
  /// Returns true if this blockage was pushed into this block.
  ///
  bool isPushedDown();

  ///
  /// Declare this blockage is soft.
  ///
  void setSoft();

  ///
  /// Returns true if this blockage is soft.
  ///
  bool isSoft();

  ///
  /// Set the max placement density percentage in [0,100]
  ///
  void setMaxDensity(float max_density);

  ///
  /// Returns the max placement density percentage
  ///
  float getMaxDensity();

  ///
  /// Get the block this blockage belongs too.
  ///
  dbBlock* getBlock();

  ///
  /// Create a placement blockage.
  ///
  static dbBlockage* create(dbBlock* block,
                            int x1,
                            int y1,
                            int x2,
                            int y2,
                            dbInst* inst = nullptr);

  ///
  /// Translate a database-id back to a pointer.
  ///
  static dbBlockage* getBlockage(dbBlock* block, uint oid);
};
```

QuantamHD:
> > I would love to not have the virtual flag, but there are a couple of issues I see:
> > 
> > 1. the GUI should also honor this behavior as I don't think it should be visible in the GUI.
> > 2. These blockages really cannot be removed, so they need away to stop users from destroying them (this is missing at the moment @QuantamHD)
> 
> In that case I think it makes sense to flag them in odb as various parts of the system need this and it needs to be reasonably fast for the GUI. Perhaps instead of isVirtual we could use isExteriorToDieArea with an explanatory comment?

I feel like what's actually special about these blockages is that they are system created, and should be ignored to everything outside the system. I think the exterior designation isn't quite right.

gadfort:
> > `odb::dbObstruction_destroy` can be called from tcl (same for Blockage)
> 
> Hmm I'm not seeing the destroy method for blockages
> 
> ```c++
> ///////////////////////////////////////////////////////////////////////////////
> ///
> /// A dbBlockage is the element that represents a placement blockage in a block.
> ///
> ///////////////////////////////////////////////////////////////////////////////
> class dbBlockage : public dbObject
> {
>  public:
>   ///
>   /// Get the bbox of this blockage.
>   ///
>   dbBox* getBBox();
> 
>   ///
>   /// Get the instance associated with this blockage.
>   /// Returns nullptr of no instance was associated with this blockage
>   ///
>   dbInst* getInstance();
> 
>   ///
>   /// Declare this blockage to have been pushed into this block.
>   ///
>   void setPushedDown();
> 
>   ///
>   /// Returns true if this blockage was pushed into this block.
>   ///
>   bool isPushedDown();
> 
>   ///
>   /// Declare this blockage is soft.
>   ///
>   void setSoft();
> 
>   ///
>   /// Returns true if this blockage is soft.
>   ///
>   bool isSoft();
> 
>   ///
>   /// Set the max placement density percentage in [0,100]
>   ///
>   void setMaxDensity(float max_density);
> 
>   ///
>   /// Returns the max placement density percentage
>   ///
>   float getMaxDensity();
> 
>   ///
>   /// Get the block this blockage belongs too.
>   ///
>   dbBlock* getBlock();
> 
>   ///
>   /// Create a placement blockage.
>   ///
>   static dbBlockage* create(dbBlock* block,
>                             int x1,
>                             int y1,
>                             int x2,
>                             int y2,
>                             dbInst* inst = nullptr);
> 
>   ///
>   /// Translate a database-id back to a pointer.
>   ///
>   static dbBlockage* getBlockage(dbBlock* block, uint oid);
> };
> ```

Maybe blockage doesn't have it (although it probably should).

src/odb/include/odb/geom.h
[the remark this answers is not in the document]
QuantamHD: boost::difference will always potentially result in internal rings though. Even if you just use rings you can cut a hole in another polygon creating an internal ring.

Imagine a `square - circle` where the circle is inside the square. Even if we only use boost's classes all we're doing is ignoring the internal rings which is the same thing we're doing here.

We could move the difference into odb::Polygon and make sure that the internal rings are empty after difference. But the Polygon mapping is used elsewhere, and is currently incorrectly implemented with value types. I don't see the benefit of doing what you suggest. 

Could you expand more on why you think it would be better to convert polygons in an adhoc way? It seems potentially more error prone than other options.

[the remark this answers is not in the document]
maliberty: I don't think you need to use boost geometry at all as boost polygon already has the operation you need (operator-).

src/odb/src/db/dbBlock.cpp
QuantamHD: Do you mind if I send this in a follow up PR. It modified about 100 files, and I think would be better as a stand alone PR.

src/odb/include/odb/geom.h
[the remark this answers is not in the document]
QuantamHD: Why prefer boost polygon to boost geometry? Both implement the difference algorithm.

The polygon mapping for boost geometry already existed before my PR, it was just implemented incorrectly. Should we leave it broken?

What is the value in constructing a boost object from odb::Polygon just to then convert it back. It seems like we should leverage the mapping.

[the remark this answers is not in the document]
maliberty: My suggestion is to remove the broken boost geometry mapping.  I'm suggesting using odb::Polygon's mapping to boost polygon where it is a correct mapping.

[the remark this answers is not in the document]
QuantamHD: I'm fine doing this, but there is no mapping to boost.polygon either. 

image

I'll make a second suggestion what if we create subclass of Polygon called MutablePolygon which contains interior rings and has these mutable ref methods. Then cast back to Polygon on return. That way we don't in general allow mutable access to polygon

QuantamHD:
What about is isSystemCreated()

QuantamHD:
The problem I have is that isExteriorToDieArea represents its position not its purpose. It feels like these methods typically denote a label rather than an object state. I prefer the negative space, because it defines what its purpose is.

QuantamHD:
image

Like take for example this floorplan. Some of the blockages are conceptually "inside".

maliberty:
I would not say they are inside as they touch the left side.  If they were truly inside it couldn't be expressed in DEF.

gadfort:
> isExteriorToDieArea vs isNegativeSpaceBlockage : I prefer the former as more precise
> 
> isExteriorToDieArea vs isSystemCreated : neutral as they emphasize different things. I'd like to hear @gadfort thoughts

I am partial to isSystemCreated or something like it, as this is exactly what it is (is isSystemReserved) isExteriorToDieArea seems like a parameter that should just be computed and doesn't indicate that this is not something a user should be modifying.

maliberty:
I'm ok with isSystemCreated or isSystemReserved

QuantamHD:
I'll go with isSystemReserved. I like that best

QuantamHD:
@gadfort @maliberty This should be ready for a re-review.

src/odb/src/db/dbObstruction.cpp
```diff
@@ -392,6 +410,15 @@ void dbObstruction::destroy(dbObstruction* obstruction)
...
   _dbBlock* block = (_dbBlock*) obs->getOwner();
+
+  if (obstruction->isSystemReserved()) {
+    utl::Logger* logger = block->getLogger();
+    logger->critical(
+        utl::ODB,
+        1111,
+        "You cannot delete a system created obstruction (isSystemReserved).");
+  }
+
```
gadfort, line 421: I would be okay with this being a silent error. The usecase I would worry about is one where a user gets all the obstructions and iterates over them to destroy them, that would just throw this critical error, which shouldn't be needed. Same for blockage

gadfort (commented):
looks good to me, except for the critical errors

src/odb/src/db/dbObstruction.cpp
maliberty: I agree it should be an error but why silent?

src/odb/src/db/dbBlock.cpp
```diff
@@ -1048,6 +1057,28 @@ dbIStream& operator>>(dbIStream& stream, _dbBlock& block)
...
+  dbSet<dbBlockage> blockages(this, _blockage_tbl);
+  dbSet<dbObstruction> obstructions(this, _obstruction_tbl);
+
+  for (dbBlockage* blockage : blockages) {
+    if (!blockage->isSystemReserved()) {
+      continue;
+    }
+    blockage->setIsSystemReserved(false);
+    dbBlockage::destroy(blockage);
+  }
```
maliberty, line 1071: It is not safe to iterate blockages while destroying them.  The usual idiom is handled by dbNet as:
```
  static dbSet<dbNet>::iterator destroy(dbSet<dbNet>::iterator& itr);
```
You need to add a similar methods to dbBlockage and dbObstruction.

QuantamHD: Okay yeah. I wasn't sure if this was a thing or not. Good to know. I can just append them to a temp vector then destroy them.

src/odb/src/db/dbObstruction.cpp
QuantamHD: This follows the idiom of dbInst DONT_TOUCH instances. It feels like we're protecting the integrity of the app. We could make it warning, and just not delete them if you want it to be silent.

gadfort: Because these get returned to the users, it would be very easy to call destroy on them and they don't want their tool to come crashing down, an error would be fine (I suggested silent because then users dont have to consider this type of blockage when deleting all the blockages they created, not realizing that these are in the list too), but critical is too restrictive

maliberty: Not a warning, an error.  Critical is too severe and does not match what we do with dont-touch.

QuantamHD: A silent error is just a warning. So we need to decide what happens if a user tries delete a system blockage. Should it throw and stop mid way or keep going, ignoring the deletion request?

maliberty: I don't agree it should be silent.   It should just be a regular error (which is what dont-touch does).

gadfort: Error is fine with me

gadfort (approved):
Lgtm

maliberty (approved):

PR #6949, opened 2025-03-30
Add a simple CI to run the bazel build.

hzeller:
Since it is experimental, I wrapped it in a `|| /bin/true` so that even if it fails, the CI is fine.

### DIFF
--- a/.github/workflows/github-actions-bazel-build.yml
+++ b/.github/workflows/github-actions-bazel-build.yml
@@ -1,0 +1,51 @@
+name: bazel-build
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  Build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Create Cache Timestamp
+        id: cache_timestamp
+        uses: nanzm/get-time-action@v2.0
+        with:
+          format: 'YYYY-MM-DD-HH-mm-ss'
+
+      - name: Mount bazel cache
+        uses: actions/cache@v4
+        with:
+          path: "~/.cache/bazel"
+          key: bazelcache_${{ steps.cache_timestamp.outputs.time }}
+          restore-keys: bazelcache_
+
+      - name: Build
+        run: |
+          # There are no cc_test()s yet, so just build for now.
+          # For now, just run, but don't fail the whole CI, as bazel build is
+          # experimental at this stage.
+          bazel build --keep_going --show_timestamps --curses=no ... || /bin/true
+
+      - name: Smoke test
+        run: |
+          bazel-bin/openroad -help
+
+      - name: Trim Bazel Cache
+        run: |
+          echo "Trim download cache"
+          du -sh $(bazel info repository_cache)
+          rm -r $(bazel info repository_cache)
+
+          echo "Other remaining things that are big"
+          du -h -t 200M ~/.cache/bazel/*/ | sort -hr
--- a/bazel/build_helper.bzl
+++ b/bazel/build_helper.bzl
@@ -141,7 +141,7 @@ OPENROAD_COPTS = [
 ]
 
 OPENROAD_DEFINES = [
-    "OPENROAD_GIT_DESCRIBE=\\\"bazel_rules_hdl\\\"",
+    "OPENROAD_GIT_DESCRIBE=\\\"bazel-build\\\"",
     "BUILD_TYPE=\\\"release\\\"",
     "GPU=false",
     "BUILD_PYTHON=false",


hzeller:
So looks like the bazel compilation finishes (and does not need any set-up w.r.t external dependencies), however caching results does not seem to be possible with fastbuild option. Will try an opt-build tomorrow and see if the resulting cache and binary is smaller so that the cache can proceed.

Das the OpenROAD project maybe have access to larger github runners ? Or are you usually doing things in Jenkins @maliberty  ?

maliberty:
We use Jenkins for building/testing the software.  Once we have a web cache setup we can enable it in Jenkins.

hzeller:
If in Jenkins you have a way to store something after a run and restore before a next run, you can use even just the local on-disk cache of bazel to only have to deal with incremental things, no webcache needed (but would of course be nice)

maliberty:
Since we are going to need the web cache anyways I think I'd like to focus on that.  I don't want everyone to have to pay to build all the dependencies.

hzeller:
Until jenkins is running, this is a good intermediate.

It will run the bazel build, and call `bazel-bin/openroad -help` to make sure the binary does something.

Bazel cache is kept between runs, so only incremental changes should be build.

maliberty:
@vvbandeira any concerns?

.github/workflows/github-actions-bazel-build.yml
```diff
@@ -0,0 +1,51 @@
...
+          path: "~/.cache/bazel"
+          key: bazelcache_${{ steps.cache_timestamp.outputs.time }}
+          restore-keys: bazelcache_
+
+      - name: Build
+        run: |
+          # There are no cc_test()s yet, so just build for now.
+          # For now, just run, but don't fail the whole CI, as bazel build is
+          # experimental at this stage.
+          bazel build --keep_going --show_timestamps --curses=no ... || /bin/true
```
maliberty, line 38: It would be better to use bazelisk than rely on the native version of bazel.  Why `--keep_going` ?

.github/workflows/github-actions-bazel-build.yml
```diff
@@ -0,0 +1,51 @@
...
+        id: cache_timestamp
+        uses: nanzm/get-time-action@v2.0
+        with:
+          format: 'YYYY-MM-DD-HH-mm-ss'
+
+      - name: Mount bazel cache
+        uses: actions/cache@v4
+        with:
+          path: "~/.cache/bazel"
+          key: bazelcache_${{ steps.cache_timestamp.outputs.time }}
```
maliberty, line 30: If the key contains the timestamp how does it get hits?  Wouldn't the key always differ?

.github/workflows/github-actions-bazel-build.yml
hzeller: Since we have a `.bazelversion` bazel will bail out anyway if the version doesn't match. So we'll have the correct version of baze.
(I also don't like bazelisk, as it essentially downloads a new version of bazel if it needs to. I want to be in control of that).

`--keep_going` so that if there is an error (which we might expect while working on the bazel build), we don't stop at the first issue but see if there are other things lurking.

.github/workflows/github-actions-bazel-build.yml
hzeller: The look-up looks for the prefix `bazelcache_` (see the `restore-keys` part), and then sorts by filename. So with this set-up we automatically get the latest cache that has been written.
The filename is _written_ with the timestamp.

.github/workflows/github-actions-bazel-build.yml
maliberty: Its ok for now though in the CI I'd rather not have to manage the bazel version.  That seems more like a local concern for an expert user.

hzeller: We do not have to manage the bazel version. It comes with the github CI, they have all the versions available.

hzeller:
Once tests are running, the bazel invocation should probably be changed to add `-c opt` to make tests pass as @oharboe observed that a `-c fastbuild` without optimization makes them too slow.

.github/workflows/github-actions-bazel-build.yml
```diff
@@ -0,0 +1,51 @@
...
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  Build:
+    runs-on: ubuntu-24.04
```
vvbandeira, line 13: Can we use latest instead?

.github/workflows/github-actions-bazel-build.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Create Cache Timestamp
+        id: cache_timestamp
+        uses: nanzm/get-time-action@v2.0
```
vvbandeira, line 22: What is the benefit of using this action instead of just running the date command?

.github/workflows/github-actions-bazel-build.yml
hzeller: Sure, done.

.github/workflows/github-actions-bazel-build.yml
hzeller: This is just the code-snippet I have written many years ago and refine for bazel caches and used in many github CI's since, so I know it works. There are probably other ways of doing that, feel free to experiment.

hzeller:
Once you merge this starting point, we can iterate over it and incorporate the tests.

.github/workflows/github-actions-bazel-build.yml
vvbandeira: Ok, it should be fine to merge for now, but I would like to reduce the number of dependencies on external actions. It will make porting to Jenkins easier and reduce attack vectors.

vvbandeira (approved):

PR #6951, opened 2025-03-30
[DO NOT REVIEW] bazel: hook up one openroad test

oharboe:
@hzeller @maliberty I know it is too early, but I had a quick stab at adding a test. I think this PR is useful to leave in draft as DO NOT REVIEW, it shows some problems that have to be resolved as Bazel support for OpenROAD progresses. Eventually some form of it should be merged to make it possible to run `bazel test ...` to run the regression tests.

Note that this PR also proposes a solution to fix the [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control) that stops tests from running in parallel without bazel.

The test fails because openroad won't work unless PWD is the runfiles _main folder:

```
$ bazel build :openroad
$ bazel-bin/openroad
application-specific initialization failed: unknown encoding "ascii"
```

Expected was:

```
$ bazel run :openroad
INFO: Analyzed target //:openroad (194 packages loaded, 16307 targets configured).
INFO: Found 1 target...
Target //:openroad up-to-date:
  bazel-bin/openroad
INFO: Elapsed time: 3.024s, Critical Path: 0.06s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/openroad
OpenROAD HDL-HEAD bazel_rules_hdl
Features included (+) or not (-): -GPU -GUI -Python
% 
$
```

```
$ bazel test //test:place_inst
[deleted]
INFO: Found 1 test target...
Target //test:place_inst up-to-date:
  bazel-bin/test/place_inst
INFO: Elapsed time: 366.250s, Critical Path: 33.09s
INFO: 792 processes: 4 internal, 788 linux-sandbox.
INFO: Build completed, 1 test FAILED, 792 total actions
//test:place_inst                                                        FAILED in 0.0s
  .../.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/execroot/_main/bazel-out/k8-fastbuild/testlogs/test/place_inst/test.log

Executed 1 out of 1 test: 1 fails locally.
```

```
$ cat .../.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/execroot/_main/bazel-out/k8-fastbuild/testlogs/test/place_inst/test.log
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //test:place_inst
-----------------------------------------------------------------------------
Directory: .../.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/sandbox/linux-sandbox/815/execroot/_main/bazel-out/k8-fastbuild/bin/test/place_inst.runfiles/_main
Directory: .../.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/sandbox/linux-sandbox/815/execroot/_main/bazel-out/k8-fastbuild/bin/test/place_inst.runfiles/_main/test
Command:   .././openroad  -no_splash -no_init -exit  place_inst.tcl > results/place_inst-tcl.log
Exitcode:  0
Diff:      .../.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/sandbox/linux-sandbox/815/execroot/_main/bazel-out/k8-fastbuild/bin/test/place_inst.runfiles/_main/test/results/place_inst-tcl.diff
application-specific initialization failed: unknown encoding "ascii"
```

### DIFF
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -1,16 +1,16 @@
 [sh_test(
-    name = "{test}".format(test=test),
+    name = "{test}".format(test = test),
     srcs = ["bazel_test.sh"],
     args = [],
     data = [
         "//:openroad",
-    ] + glob(["*.sh", "*.ok"]),
+    ] + glob(["**/*"]),
     env = {
         "TEST_NAME_BAZEL": test,
-        "OPENROAD_EXE": "../$(location //:openroad)",
+        "OPENROAD_EXE": "$(location //:openroad)",
         "TEST_EXT": "tcl",
-        "TEST_CHECK_LOG": "True",
-        "TEST_CHECK_PASSFAIL": "True",
-
     },
-) for test in ["gcd_sky130hs", "place_inst"]]
+) for test in [file.split(".")[0] for file in glob(
+    ["*.tcl"],
+    exclude = ["regression.tcl"],
+)]]
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -14,6 +14,9 @@
         "OPENROAD_EXE": "$(location //:openroad)",
         "TEST_EXT": "tcl",
     },
+    # No point in running these tests with unless they fail,
+    # they would be too slow.
+    tags = ["optonly"],
 ) for test in [file.split(".")[0] for file in glob(
     ["*.tcl"],
     exclude = ["regression.tcl"],
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -1,0 +1,16 @@
+[sh_test(
+    name = "{test}".format(test=test),
+    srcs = ["bazel_test.sh"],
+    args = [],
+    data = [
+        "//:openroad",
+    ] + glob(["*.sh", "*.ok"]),
+    env = {
+        "TEST_NAME_BAZEL": test,
+        "OPENROAD_EXE": "../$(location //:openroad)",
+        "TEST_EXT": "tcl",
+        "TEST_CHECK_LOG": "True",
+        "TEST_CHECK_PASSFAIL": "True",
+
+    },
+) for test in ["gcd_sky130hs", "place_inst"]]
--- a/test/BUILD.bazel
+++ b/test/BUILD.bazel
@@ -1,5 +1,9 @@
 [sh_test(
     name = "{test}".format(test = test),
+    # top showed me 50-400mByte of usage, so "enormous"...
+    #
+    # https://bazel.build/reference/be/common-definitions#test.size
+    size = "enormous",
     srcs = ["bazel_test.sh"],
     args = [],
     data = [
--- a/test/bazel_test.sh
+++ b/test/bazel_test.sh
@@ -4,9 +4,6 @@ set -e
 
 echo "Directory: ${PWD}"
 
-DIR=$(dirname "$0")
-cd $DIR
-
 # TEST_NAME is used in cmake and bazel...
 export TEST_NAME=$TEST_NAME_BAZEL
-./regression_test.sh
+test/regression_test.sh
--- a/test/bazel_test.sh
+++ b/test/bazel_test.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo "Directory: ${PWD}"
+
+DIR=$(dirname "$0")
+cd $DIR
+
+# TEST_NAME is used in cmake and bazel...
+export TEST_NAME=$TEST_NAME_BAZEL
+./regression_test.sh
--- a/test/regression_test.sh
+++ b/test/regression_test.sh
@@ -20,7 +20,7 @@ $OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit <(cat <<EOF
 cd "$PREFIX"
 source "$TEST_NAME.$TEST_EXT"
 EOF
-) 2>&1 | tee $LOG_FILE
+) 2>&1 | tee $(realpath $LOG_FILE)
 
 echo "Exitcode:  $?"
 
--- a/test/regression_test.sh
+++ b/test/regression_test.sh
@@ -2,25 +2,31 @@
 
 set -e
 
-mkdir -p results
+# Make this work with a different pwd
+PREFIX=$(realpath $(realpath --relative-to=$(pwd) $(dirname $0)))
 
-LOG_FILE=results/$TEST_NAME-$TEST_EXT.log
+mkdir -p $PREFIX/results
+
+LOG_FILE=$PREFIX/results/$TEST_NAME-$TEST_EXT.log
 
 ORD_ARGS=""
 if [ "$TEST_TYPE" == "python" ]; then
     ORD_ARGS="-python"
 fi
 
-echo "Directory: ${PWD}"
-echo "Command:   $OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit  $TEST_NAME.$TEST_EXT > $LOG_FILE"
+echo "Command:   $OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit  $PREFIX/$TEST_NAME.$TEST_EXT > $LOG_FILE"
 
-$OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit $TEST_NAME.$TEST_EXT &> $LOG_FILE
+$OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit <(cat <<EOF
+cd "$PREFIX"
+source "$TEST_NAME.$TEST_EXT"
+EOF
+) 2>&1 | tee $LOG_FILE
 
 echo "Exitcode:  $?"
 
 if [ "$TEST_CHECK_LOG" == "True" ]; then
-    echo "Diff:      ${PWD}/results/$TEST_NAME-$TEST_EXT.diff"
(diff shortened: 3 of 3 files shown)


maliberty:
I've been thinking of doing a test just to see how hard it is and to have some confidence in the binary.  I was going to start with utl but this is fine too.  Let's see if we can get it working.

oharboe:
> I've been thinking of doing a test just to see how hard it is and to have some confidence in the binary. I was going to start with utl but this is fine too. Let's see if we can get it working.

:+1: 

I think it boils down to fixing this, the rest is easy:

```
$ bazel build :openroad
$ bazel-bin/openroad
application-specific initialization failed: unknown encoding "ascii"
```

maliberty:
Rather than directly invoke the binary you have to do `bazel run //:openroad` currently.  It doesn't find tcl otherwise.  I'd like it to be more self-contained...

oharboe:
> Rather than directly invoke the binary you have to do `bazel run //:openroad` currently. It doesn't find tcl otherwise. I'd like it to be more self-contained...

That doesn't work. The test is not the use-case, it just illustrates the problem.

The problem is that we want to use openroad as binary when CWD  != the runfiles._main folder, otherwise we've got lots of use-cases that can't work.

hzeller:
`bazel run` is indeed changing directories and that is not necessarily desired.

I suggest to `realpath` the binary so that you get an absolute path of where the `openroad` binary ends up, then you can use that absolute path in whatever test context.

oharboe:
> `bazel run` is indeed changing directories and that is not necessarily desired.
> 
> I suggest to `realpath` the binary so that you get an absolute path of where the `openroad` binary ends up, then you can use that absolute path in whatever test context.

That will find openroad, but that is not the problem.

The problem is that openroad doesnt find files it expects to find as  bazel built openroad uses pwd and not the path of openroad to find the tcl files, unlike the classic openroad built binary.

QuantamHD:
Bazel exports a number of env variables for tests.

https://bazel.build/reference/test-encyclopedia

One of which is an absolute path to the runfiles. $TEST_SRCDIR

You'll want to export TCL_LIBRARY=TEST_SRCDIR/external/tcl_tk/library.

Double check the actual path to the library folder.

oharboe:
@QuantamHD Thanks!

This does not cover the ORFS use case, which executes during build, so the problem persists.

QuantamHD:
You'll probably need to create a custom rule and then use the runfiles directly python value, and export it as an env variable

maliberty:
@QuantamHD how does this work inside of G?  Presumably you use the same bazel_rules_hdl tcl setup.  It appears none of the data deps get exported to bazel-*

maliberty:
```
find bazel* -name '*tcl*'
bazel/tcl_encode.bzl
bazel/tcl_wrap_cc.bzl
```

QuantamHD:
Let me take a look

oharboe:
After a coffee and a snooze in the spring sun it came to me how this works in Bazel.

Bazel relies on relative paths to runfiles, so don't change pwd.

image

hzeller:
If you see issues with timeouts, there are some specifiers to annotate tests
 https://bazel.build/reference/be/common-definitions#test.size

If this is only an issue if not compiled with `-c opt`: Sometimes, it only really makes sense to run tests when compiled in opt mode because otherwise they would be too slow. I _think_ there is a way to do that by adding `tags = ["optonly"]` on that test.

hzeller:
>  I _think_ there is a way to do that by adding `tags = ["optonly"]` on that test.

I just tried that, and it does not make a difference: tests are run independent of `-c opt`, `-c fastbuild` or `-c dbg`. I suspect there could be a way using configurations. @QuantamHD knows a lot more, maybe he has some ideai.

test/BUILD.bazel
```diff
@@ -14,6 +14,9 @@
         "OPENROAD_EXE": "$(location //:openroad)",
         "TEST_EXT": "tcl",
     },
+    # No point in running these tests with unless they fail,
+    # they would be too slow.
+    tags = ["optonly"],
```
hzeller, line 19: Let's see if it works, I think my previous suggestion is something I misremembered.

oharboe:
@maliberty @hzeller I probably need some help on this PR...

```
$ tail -n 20 /home/oyvind/.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/execroot/_main/bazel-out/k8-fastbuild/testlogs/test/aes_nangate45/test.log
[deleted]
[INFO DRT-0036] via9 guide region query size = 0.
[INFO DRT-0036] metal10 guide region query size = 0.
[INFO DRT-0179] Init gr pin query.
-- Test timed out at 2025-03-30 23:13:02 UTC --
```


```
INFO: Build completed, 8 tests FAILED, 25 total actions
//test:commands_without_load                                             PASSED in 0.4s
//test:error1                                                            PASSED in 0.2s
//test:flow                                                              PASSED in 0.2s
//test:flow_helpers                                                      PASSED in 0.2s
//test:flow_metrics                                                      PASSED in 0.2s
//test:gcd_nangate45                                                     PASSED in 264.0s
//test:gcd_sky130hd                                                      PASSED in 787.5s
//test:gcd_sky130hd_fast_slow                                            PASSED in 803.7s
//test:gcd_sky130hs                                                      PASSED in 820.4s
//test:get_core_die_areas                                                PASSED in 1.2s
//test:helpers                                                           PASSED in 0.2s
//test:place_inst                                                        PASSED in 1.2s
//test:regression_tests                                                  PASSED in 0.2s
//test:regression_vars                                                   PASSED in 0.2s
//test:upf_aes                                                           PASSED in 530.1s
//test:upf_test                                                          PASSED in 2.9s
//test:aes_nangate45                                                    TIMEOUT in 3600.1s
  /test/aes_nangate45/test.log
//test:aes_sky130hd                                                     TIMEOUT in 3600.2s
  /test/aes_sky130hd/test.log
//test:aes_sky130hs                                                     TIMEOUT in 3600.1s
  /test/aes_sky130hs/test.log
//test:ibex_sky130hd                                                    TIMEOUT in 3600.1s
  /test/ibex_sky130hd/test.log
//test:ibex_sky130hs                                                    TIMEOUT in 3600.1s
  /test/ibex_sky130hs/test.log
//test:jpeg_sky130hd                                                    TIMEOUT in 3600.2s
  /test/jpeg_sky130hd/test.log
//test:jpeg_sky130hs                                                    TIMEOUT in 3600.2s
  /test/jpeg_sky130hs/test.log
//test:tinyRocket_nangate45                                             TIMEOUT in 3600.1s
  /test/tinyRocket_nangate45/test.log

Executed 24 out of 24 tests: 16 tests pass and 8 fail locally.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

hzeller:
How long would you expect these tests to take ?
If they are slow, I hightly recommend to run them with  `bazel test - c opt ...` (the default 'fastbuild' is essentially `-O0` or something).
(after that, we need to figure out how to make it work that they are not run when non-opt is selected)

test/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+[sh_test(
+    name = "{test}".format(test = test),
```
hzeller, line 2: While the formatting is neat, wouldn't it be much simpler to just say `name = test` ?

Though probably it would be good to have some prefix or suffix, e.g. a test could be  `tcl_foo` (then of course some sort of string manipulation needed, something like `"tcl_" + test` or `"tcl_{}.format(test)`")

test/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
...
+    # top showed me 50-400mByte of usage, so "enormous"...
+    #
+    # https://bazel.build/reference/be/common-definitions#test.size
+    size = "enormous",
+    srcs = ["bazel_test.sh"],
+    args = [],
+    data = [
+        "//:openroad",
+    ] + glob(["**/*"]),
```
hzeller, line 11: that is a big glob :) I guess that is for testing now, but of course being more specific would be good.

oharboe: I'll do a pass of squash and refine this PR before it is ready for review.

Trying now `bazel test -c opt aes_nangate45`

oharboe: @maliberty I've started at my share of detailed routing logs, I have not seen it stuck where it is stuck below. Is that indicative that I'm not running an optimized openroad build?

```
[INFO DRT-0036] via8 guide region query size = 0.
[INFO DRT-0036] metal9 guide region query size = 12.
[INFO DRT-0036] via9 guide region query size = 0.
[INFO DRT-0036] metal10 guide region query size = 0.
[INFO DRT-0179] Init gr pin query.
[no more output for minutes]
```

oharboe: It completed, timed out in 3600 seconds before. so `bazel test -c opt ...` is at least 6x faster.

```
INFO: Build completed successfully, 4497 total actions
//test:aes_nangate45                                                     PASSED in 561.8s

Executed 1 out of 1 test: 1 test passes.
```

oharboe:
closing review to preserve comment history.

oharboe:
proposed baseline to merge for further collaboration and refinement https://github.com/The-OpenROAD-Project/OpenROAD/pull/6959

PR #6958, opened 2025-03-31
DPL DPO merge Node

osamahammad21:
- Use DbuX and DbuY in Pin class.
- Moved Pin, Edge objects to dpl/objects.h
- Made Objects.h public for DPO to use.
- Eliminated Architecture::Region class and used dpl::Group instead.
- Merged GridNode, Node and Cell classes into the final Node class.


src/dpl/include/dpl/PlacementDRC.h
```diff
@@ -0,0 +1,59 @@
...
+  EdgeSpacingEntry(const int spc_in,
+                   const bool is_exact_in,
+                   const bool except_abutted_in)
+      : spc(spc_in), is_exact(is_exact_in), except_abutted(except_abutted_in)
+  {
+  }
+  bool operator<(const EdgeSpacingEntry& rhs) const { return spc < rhs.spc; }
+  int spc;
+  bool is_exact;
+  bool except_abutted;
```
maliberty, line 28: can these be const?

src/dpl/include/dpl/PlacementDRC.h
```diff
@@ -0,0 +1,59 @@
...
+  int getEdgeTypeIdx(const std::string& edge_type) const;
+  bool hasCellEdgeSpacingTable() const;
+  int getMaxSpacing(int edge_type_idx) const;
+
+ private:
+  // Member variables
+  Grid* grid_{nullptr};  // Pointer to the grid for placement
+  std::unordered_map<std::string, int> edge_types_indices_;
+  std::vector<std::vector<EdgeSpacingEntry>>
+      edge_spacing_table_;  // Edge spacing rules
```
maliberty, line 52: It is helpful to add what the indices are (I believe it is two edge types).

src/dpl/include/dpl/Objects.h
```diff
@@ -0,0 +1,266 @@
...
+  unsigned int edge_type_idx_{0};
+  Rect bbox_;
+};
+
+class Master
+{
+ public:
+  bool isMultiRow() const;
+  const std::vector<MasterEdge>& getEdges();
+  Rect getBBox();
```
maliberty, line 62: const methods

src/dpl/include/dpl/Objects.h
```diff
@@ -0,0 +1,266 @@
...
+  Master* master_{nullptr};
+  Group* group_{nullptr};
+  const Rect* region_{nullptr};  // group rect
+  // // Regions.
+  int group_id_{0};
+  // Pins.
+  std::vector<Pin*> pins_;
+};
+
+class Group
```
maliberty, line 180: Is it possible for any fields to set by a ctor rather than setters?  Similarly with the following classes.

src/dpl/include/dpl/Objects.h
```diff
@@ -0,0 +1,266 @@
...
+class Node
+{
+ public:
+  enum Type
+  {
+    UNKNOWN,
+    CELL,
+    TERMINAL,
+    MACROCELL,
+    FILLER
```
maliberty, line 86: I suppose this is needed by dpo but it is strange to have these lumped into a single class as most of these are not moveable.  Something for another PR

maliberty (changes requested):

src/dpl/include/dpl/Objects.h
osamahammad21: Possible but I think it is better handled in a separate refactor PR.

src/dpl/include/dpl/PlacementDRC.h
osamahammad21: I don't think, because of the way I initialize `edge_spacing_table_` in `PlacementDRC::makeCellEdgeSpacingTable`

src/dpl/include/dpl/Objects.h
osamahammad21: Agreed.

osamahammad21:
@maliberty very weird failure in psm regression tests

maliberty (approved):

maliberty:
I'll leave this to you to merge - not sure if you are doing any further testing

PR #6959, opened 2025-03-31
bazel: hook up one openroad test, baseline for collaboration

oharboe:
I propose to merge this as a baseline for further collaboration and refinement.

Note that the tests must be run in optimized mode or they will time out after an hour instead of completing in ca. 10 minutes.

    cd test
    bazel test -c opt ...

Future refinements:

- add python tests
- use some heuristic to know which pdk files to glob
- reduce glob to exclude test .py/.tcl files other than the current test
- there are lots more of these tests. To find folders that have such tests, search for src/**/test/regression. All of these tests should be added.


oharboe:
@maliberty It looks like these tests is gnawing on some bone I haven't seen them gnaw on in ORFS....

Thoughts?

```
std::__1::vector<boost::polygon::polygon_90_with_holes_data<int>, std::__1::allocator<boost::polygon::polygon_90_with_holes_data<int>>>::insert(std::__1::__wrap_iter<boost::polygon::polygon_90_with_holes_data<int> const*>, boost::polygon::polygon_90_with_holes_data<int> const&) (@std::__1::vector<boost::polygon::polygon_90_with_holes_data<int>, std::__1::allocator<boost::polygon::polygon_90_with_holes_data<int>>>::insert(std::__1::__wrap_iter<boost::polygon::polygon_90_with_holes_data<int> const*>, boost::polygon::polygon_90_with_holes_data<int> const&):136)
unsigned int boost::polygon::get_polygons<std::__1::vector<boost::polygon::polygon_90_with_holes_data<int>, std::__1::allocator<boost::polygon::polygon_90_with_holes_data<int>>>, std::__1::__wrap_iter<std::__1::pair<int, std::__1::pair<int, int>>*>, boost::polygon::polygon_90_with_holes_concept>(std::__1::vector<boost::polygon::polygon_90_with_holes_data<int>, std::__1::allocator<boost::polygon::polygon_90_with_holes_data<int>>>&, std::__1::__wrap_iter<std::__1::pair<int, std::__1::pair<int, int>>*>, std::__1::__wrap_iter<std::__1::pair<int, std::__1::pair<int, int>>*>, boost::polygon::orientation_2d, bool, boost::polygon::polygon_90_with_holes_concept, unsigned long) (@unsigned int boost::polygon::get_polygons<std::__1::vector<boost::polygon::polygon_90_with_holes_data<int>, std::__1::allocator<boost::polygon::polygon_90_with_holes_data<int>>>, std::__1::__wrap_iter<std::__1::pair<int, std::__1::pair<int, int>>*>, boost::polygon::polygon_90_with_holes_concept>(std::__1::vector<boost::polygon::polygon_90_with_holes_data<int>, std::__1::allocator<boost::polygon::polygon_90_with_holes_data<int>>>&, std::__1::__wrap_iter<std::__1::pair<int, std::__1::pair<int, int>>*>, std::__1::__wrap_iter<std::__1::pair<int, std::__1::pair<int, int>>*>, boost::polygon::orientation_2d, bool, boost::polygon::polygon_90_with_holes_concept, unsigned long):147)
drt::FlexGCWorker::Impl::initNet_pins_polygonCorners_helper(drt::gcNet*, drt::gcPin*) (@drt::FlexGCWorker::Impl::initNet_pins_polygonCorners_helper(drt::gcNet*, drt::gcPin*):415)
drt::FlexGCWorker::Impl::initNet(drt::gcNet*) (@drt::FlexGCWorker::Impl::initNet(drt::gcNet*):44)
drt::FlexGCWorker::Impl::updateGCWorker() (@drt::FlexGCWorker::Impl::updateGCWorker():407)
drt::FlexDRWorker::initMazeCost_connFig() (@drt::FlexDRWorker::initMazeCost_connFig():48)
drt::FlexDRWorker::init(drt::frDesign const*) (@drt::FlexDRWorker::init(drt::frDesign const*):61)
drt::FlexDRWorker::main(drt::frDesign*) (@drt::FlexDRWorker::main(drt::frDesign*):435)
drt::FlexDR::processWorkersBatch(std::__1::vector<std::__1::unique_ptr<drt::FlexDRWorker, std::__1::default_delete<drt::FlexDRWorker>>, std::__1::allocator<std::__1::unique_ptr<drt::FlexDRWorker, std::__1::default_delete<drt::FlexDRWorker>>>>&, drt::FlexDR::IterationProgress&) (@drt::FlexDR::processWorkersBatch(std::__1::vector<std::__1::unique_ptr<drt::FlexDRWorker, std::__1::default_delete<drt::FlexDRWorker>>, std::__1::allocator<std::__1::unique_ptr<drt::FlexDRWorker, std::__1::default_delete<drt::FlexDRWorker>>>>&, drt::FlexDR::IterationProgress&):41)
drt::FlexDR::optimizationFlow(drt::FlexDR::SearchRepairArgs const&, drt::FlexDR::IterationProgress&) (@drt::FlexDR::optimizationFlow(drt::FlexDR::SearchRepairArgs const&, drt::FlexDR::IterationProgress&):413)
drt::FlexDR::searchRepair(drt::FlexDR::SearchRepairArgs const&) (@drt::FlexDR::searchRepair(drt::FlexDR::SearchRepairArgs const&):188)
drt::FlexDR::main() (@drt::FlexDR::main():179)
drt::TritonRoute::dr() (@drt::TritonRoute::dr():74)
drt::TritonRoute::main() (@drt::TritonRoute::main():457)
detailed_route_cmd(char const*, char const*, char const*, char const*, char const*, bool, int, char const*, char const*, int, double, char const*, char const*, int, bool, bool, bool, int, bool, char const*, int) (@detailed_route_cmd(char const*, char const*, char const*, char const*, char const*, bool, int, char const*, char const*, int, double, char const*, char const*, int, bool, bool, bool, int, bool, char const*, int):393)
_wrap_detailed_route_cmd(void*, Tcl_Interp*, int, Tcl_Obj* const*) (@_wrap_detailed_route_cmd(void*, Tcl_Interp*, int, Tcl_Obj* const*):402)
Tcl_EvalObjv (@Tcl_EvalObjv:42)
TclEvalEx (@TclEvalEx:593)
Tcl_Eval (@Tcl_Eval:15)
sta::sourceTclFile(char const*, bool, bool, Tcl_Interp*) (@sta::sourceTclFile(char const*, bool, bool, Tcl_Interp*):32)
ord::tclInit(Tcl_Interp*) (@ord::tclInit(Tcl_Interp*):416)
Tcl_MainEx (@Tcl_MainEx:203)
main (@main:101)
__libc_start_call_main (@__libc_start_call_main:29)
__libc_start_main_impl (@__libc_start_main@@GLIBC_2.34:44)
_start (@_start:15)
```

oharboe:
@hzeller FYI. I wanted to have a simple example of how to run these tests as a starting point for further collaboration. I'm thinking these tests can be run locally until all the tests in all the folders that contain src/**/test/regression are hooked up. I see there's some work in flight to hook up bazel to CI https://github.com/The-OpenROAD-Project/OpenROAD/pull/6949.

test/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+    # https://bazel.build/reference/be/common-definitions#test.size
+    size = "enormous",
+    srcs = ["bazel_test.sh"],
+    args = [],
+    data = [
+        "//:openroad",
+    ]
+    # TODO Refine blob later.
+    #
+    # This is a very broad blob, but most of the data in size
```
hzeller, line 14: s/blob/glob/

test/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+    # comes from .lib files and we can't know which pdk files
+    # to include for a test given the name only.
+     + glob(["**/*"]),
+    env = {
+        "TEST_NAME_BAZEL": test,
+        "OPENROAD_EXE": "$(location //:openroad)",
+        "TEST_EXT": "tcl",
+    },
+    # FIXME bazel test ... should run these tests in optimized mode only
```
hzeller, line 23: @QuantamHD do you know if there is a way to only enable these tests on `bazel test ...` when `-c opt` is enabled ?

QuantamHD: No there's not. We shouldn't try to do that either. That's really a CI related decision

oharboe: Maybe https://github.com/The-OpenROAD-Project/OpenROAD/pull/6963 ?

oharboe:
@maliberty CI passed last time, I only changed a comment.

test/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
+[sh_test(
```
hzeller, line 1: I'd call this file `BUILD` not `BUILD.bazel`.

oharboe:
@maliberty Anything else I can do?

maliberty:
How do you get the tests to run?  `bazelisk test -c opt //:openroad` runs no tests `ERROR: No test targets were found, yet testing was requested`

maliberty:
nvm, I missed `cd test`

maliberty:
The tests look to all fail with something like:
```
tee: /home/mliberty_openroad_tools/w4/tools/OpenROAD/test/results/gcd_nangate45-tcl.log: Read-only file system
```

hzeller:
FYI it is uncommon to first change into the directory, just use the relative names.

So in the toplevel just say `bazel test //test/...` or simply `bazel test ...` which will find all the tests.

Yeah it is not possible to clobber the bazel output directory, but if you want to write somewhere, I suggest to write relative to `$TEST_TMPDIR`. Alternatively this could be an output of this rule, but don't know exactly if `sh_test()` has such an option.

maliberty:
The results directory is for the test outputs so they can be inspected in case of failure.

Note `bazelisk test -c opt ...` fails with 
```
ERROR: /home/mliberty_openroad_tools/w4/tools/OpenROAD/build/_deps/googletest-src/googletest/test/BUILD.bazel:529:8: errors encountered resolving select() keys for //build/_deps/googletest-src/googletest/test:googletest-json-output-unittest
```
while `bazelisk test -c opt //test/...` runs.

maliberty:
I have no desire for my external dependencies to run their tests.

hzeller:
sounds like you have a random directory in your path that starts with `build/`  :)
Leftover from a previous cmake run ?

maliberty:
Yes I need to check changes against both bazel and cmake.  Is there a way to exclude it?

hzeller:
Don't know. My preferred way of dealing with superfluous cmake temp build directories is the `rm -rf` operation :)
Maybe crate the build output outside the main repo.

test/regression_test.sh
```diff
@@ -4,23 +4,50 @@ set -e
...
+fi
 
-$OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit $TEST_NAME.$TEST_EXT &> $LOG_FILE
+# Ideally, bazel would show continious output in the console when it is running
+# a single command, but it doesn't.
+#
+# It is useful in bazel to ps -Af | grep tee to find the absolute path of
+# the log file and tail -f it.
+cd ${BAZEL_SAVED_PWD:-$SAVED_PWD}
+$OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit <(echo "$script_content") 2>&1 | tee $LOG_FILE > /dev/null
```
hzeller, line 43: bazel should already store the logfile from the stdout, no need to create an additional output.

If you want to see things while they are happening, you can `bazel test -c opt --test_output=streamed ...`

test/regression_test.sh
```diff
@@ -4,23 +4,50 @@ set -e
 
 mkdir -p results
 
-LOG_FILE=results/$TEST_NAME-$TEST_EXT.log
+LOG_FILE=$(realpath results/$TEST_NAME-$TEST_EXT.log)
```
hzeller, line 7: Not sure if you can write into the bazel-build directory, that should be read-only. You can write to `${TEST_TMPDIR}` if you want, but for the logging reasons, you can just use bazels logging it anyway does for the test.

test/regression_test.sh
maliberty: stdout is not the only output

hzeller: ah, at least this script only addresses stdout/stderr.
Anyway if they are writing temporary stuff, they should use `${TEST_TMPDIR}`

test/regression_test.sh
```diff
@@ -4,23 +4,50 @@ set -e
...
+$OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit <(echo "$script_content") 2>&1 | tee $LOG_FILE > /dev/null
+cd $SAVED_PWD
 
 echo "Exitcode:  $?"
 
 if [ "$TEST_CHECK_LOG" == "True" ]; then
-    echo "Diff:      ${PWD}/results/$TEST_NAME-$TEST_EXT.diff"
-    diff $LOG_FILE $TEST_NAME.ok > results/$TEST_NAME-$TEST_EXT.diff
+    echo "Diff:      results/$TEST_NAME-$TEST_EXT.diff"
+    diff $LOG_FILE $TEST_NAME.ok > results/$TEST_NAME-$TEST_EXT.diff || (head -n 10 $LOG_FILE && exit 1)
```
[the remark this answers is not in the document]
maliberty: The point is that if a test fails you want to be able to see those files so they aren't purely temporary.  That's why we write the to results/ so you can inspect them as needed.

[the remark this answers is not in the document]
hzeller: You can't write to the bazel directory though. Only stuff that bazel knows will end up there is allowed (declared outputs of rules), other filesystem access is sandboxed.

So either you write files to some `${TMPDIR}` (one of the few directories you can write to outside the sandbox and can still can manually inspect them later), or declare them as an output to the test so that they end up in `bazel-out/` somewhere (I suspect that would require modifying `sh_test` or writing a specific rule for it, as it looks like `sh_test` does not have some configuration of additional outputs according to the [documentation](https://bazel.build/reference/be/shell#sh_test)

Not only allows that reproducability, but it also allows to cache the results. Say you run against a distributed build cluster, and want to inspect the outputs of your tests, only if bazel knows that a file is written it can transfer it back for you to inspect.

oharboe:
To test updated version and view output, run:

```
$ bazel test -c opt ... --test_output=streamed
[deleted]
ORD-0062
[ERROR ORD-0063] origin and location specified for test2, only one is supported.
ORD-0063
[ERROR ORD-0063] Instance test2 not in the design, -cell must be specified to create a new instance.
ORD-0063
No differences found.
Exitcode:  0
INFO: Found 5 test targets...
INFO: Elapsed time: 38.450s, Critical Path: 37.08s
INFO: 21 processes: 11 internal, 10 linux-sandbox.
INFO: Build completed successfully, 21 total actions
//test:error1                                                   (cached) PASSED in 0.1s
//test:get_core_die_areas                                       (cached) PASSED in 0.2s
//test:place_inst                                               (cached) PASSED in 0.2s
//test:upf_aes                                                  (cached) PASSED in 37.5s
//test:upf_test                                                 (cached) PASSED in 0.4s

Executed 0 out of 5 tests: 5 tests pass.
```

test/regression_test.sh
oharboe: sh_test() is a simple thing thing: run a test and a pass/fail. It is not designed to keep artifacts, like genrule().

However, with OpenROAD and especially long running flows, it can be useful to cache output of a stage, even if that stage failed. An example is global routing: it is useful to examine global routing output, even if it failed.

The solution to this mismatch in use-cases and how Bazel works is straightforward enough: add a genrule() that runs the flow and returns non-zero exit code if the test ran to completion. Ran to completion means that it either passed or failed, but that all artifacts should be retained. On top of this rule, add a sh_test() that returns non-zero exit code if the flow/test executed in the genrule() failed.

This situation is not unique to Bazel or OpenROAD/EDA flows. I had the exact same situation in geoscience: simulations results should be kept if it ran to completion, whether or not they converged. There should be a separate, post simulation reporting stage, to evaluate the simulation and where "success" and "failure" is a possible conclusion.

The first question I had was "why doesn't Bazel cache failed builds and tests"? I quickly discoverd that this would fix one problem and create another: it would be tricky to re-run the test locally as you would only get the cached result. Flaky tests are still a problem an accidental success can populate the cache with a success, in which case it is trickier to re-run tests locally. If I need to iterate on a test locally, I sometimes modify it to always fail as early as possible, so that it will re-run always. Bazel does have a concept of flaky tests, but it wasn't an exact match for my use-cases and I didn't study it.

test/regression_test.sh
oharboe: fixed. Backed this out and use stdout only for this logging.

test/regression_test.sh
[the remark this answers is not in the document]
oharboe: fixed

oharboe:
@hzeller @maliberty I want to keep this PR as simple as possible and to address as few concerns as possible so as to serve as a starting point for refinements and discussions. I did discover that CMakeLists.txt explicitly list the tests, which I've ameded BUILD to follow.

oharboe:
@maliberty pr-head worked....

What's going on here?

https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-6959-merge/6/pipeline-graph/

```
10:29:18  The following tests FAILED:
10:29:18  	785 - grt.congestion1.py (Failed)
10:29:18  Errors while running CTest
```

I tried reproducing locally without luck:

```
$ cd src/grt/test
$ ./regression -j 48
```

oharboe:
@maliberty merged with latest master, which will trigger two more builds. I suspect that there's some intermittency to this problem...

oharboe:
@maliberty The best explanation I have is that https://github.com/The-OpenROAD-Project/OpenROAD/pull/6959#issuecomment-2772376392 is an intermittent problem. Perhaps it already exists on master and you are familiar with it? If it is introduced by this PR, I don't see where the race condition would come from.

maliberty:
I had to add `timeout = "moderate",` to avoid a timeout on upf_aes

test/regression_test.sh
hzeller: If you need a test that has a defined set of outputs which can't be covered by `sh_test`, you could also write a bzl rule that better reflects the test scenario, with different outputs etc  (@QuantamHD knows a lot about rules, so might be able to give you pointers to starting points how to get that going). But: something for later.

To run tests without caching, use  `--nocache_test_results`

Bazel does have a concept of flaky tests, see https://bazel.build/reference/be/common-definitions#test.flaky

maliberty:
I can't push the change to your branch so please make the change

oharboe:
> I had to add `timeout = "moderate",` to avoid a timeout on upf_aes

Fixed.

maliberty:
It passes now but you get a bunch of warnings that some tests don't need the longer timeout.  I guess it really wants you to set it per test.  A future thing to clean up.

maliberty (approved):

PR #6963, opened 2025-03-31
bazel: default to release build

oharboe:
CI and tests would take too long with debug builds, default to release and enable debug explicitly if there's something to investigate

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -32,6 +32,12 @@ build:asan --copt -g
 build:asan --copt -fno-omit-frame-pointer
 build:asan --linkopt -fsanitize=address
 
+# Mostly users and build systems want the release build
+build:release -c opt
+build --config=release
+run --config=release
+test --config=release
+
 # TODO: document
 build --incompatible_strict_action_env
 


oharboe:
@QuantamHD @hzeller Thoughts? OpenROAD is too slow in CI and local builds when built in debug mode for most integration tests.

I don't know bazel quite well enough to know if some build:debug lines are needed too.

hzeller:
It is uncommon to fix the default build to a particular configuration, so by default still things would default to fastbuild; it provides faster turn-around for quick incremental compilation updates.

Typically people put `build -c opt` into their `user.bazelrc` if they want that in their development flow.

For the Continuous Integration, we also can just explicitly choose `-c opt` and are good, so no need to bake that into the main `.bazelrc`.

Having said that, having a `release` config probably might make sense at some point, in particular when it extends to beyond `-c opt` but also includes FDO and possibly other optimization techniques. But I only would add such a configuration once we have more than `-c opt`.

TL;DR:
  * having a `release` config is good, but I'd only establish it once we have more than simply `-c opt`, but possibly other flags we'd like to have
  * fixing the default build to `--config=release` or `-c opt` is probably a less popular idea.

oharboe:
> It is uncommon to fix the default build to a particular configuration, so by default still things would default to fastbuild; it provides faster turn-around for quick incremental compilation updates.
> 
> Typically people put `build -c opt` into their `user.bazelrc` if they want that in their development flow.

My concern is the OpenROAD and ORFS use case and remote caching: build times for ORFS flows and tests are very long, so when working locally, I don't want to start a build from scratch when I change from the default build.

So what I'm trying to do with the default build settings is that they should be fast(minutes, not hours) and also that the remote cache should generally be fully populated by CI on master and for pull requests that have been built.

maliberty:
How much is the build time difference?

maliberty:
I meant the build time of OR.  How much of a hit is it to change the default.  In cmake we have the equivalent of opt & dbg but not fastbuild and its been tolerable.

oharboe:
> I meant the build time of OR. How much of a hit is it to change the default. In cmake we have the equivalent of opt & dbg but not fastbuild and its been tolerable.

I can't say I have timed it carefully enough to have noticed a different in build times for OR itself between configurations. Most of the waiting I do are on tests + flows.

I think CI could build the different OR configurations and populate the cache, but I'm thinking it only makes sense to populate the test/ORFS flow builds based on OR release builds.

hzeller:
You can put multiple configurations in the cache, so there should not be much impact of what people choose.

On my older laptop it took me about ~1.5h for fastbuild and longer for opt when building from scratch after a `bazel clean --expunge`.

oharboe:
> You can put multiple configurations in the cache, so there should not be much impact of what people choose.

My concern is that the cache for all the downstream ORFS flows and tests. That will amount to many, many times more than the cache for OpenROAD binary.

maliberty:
> My concern is that the cache for all the downstream ORFS flows and tests. That will amount to many, many times more than the cache for OpenROAD binary.

That's unrelated to how OR itself is built.  We can limit ORFS to just opt for OR.

hzeller:
You can use openroad as a 'tool' in ORFS flows, which will have different compilation flags. A tool is typically compiled with some sort of release flags, such as opt which can be different to the local compilation rules you use (think of `protoc` the protocol compiler: even if you build your project with fastbuild or dbg, you want the protoc, which is also built as part of your build to be always optimized).

So I guess the question for your ORFS flows would be solved if you use openroad as tool, then it always use the release version. The details need to be read up somewhere in the bazel documentaion, but I imagine @QuantamHD can help figuring these out.

oharboe:
> > My concern is that the cache for all the downstream ORFS flows and tests. That will amount to many, many times more than the cache for OpenROAD binary.
> 
> That's unrelated to how OR itself is built.  We can limit ORFS to just opt for OR.

It affects cached artifacts for OR tests within OR too.

QuantamHD:
What exactly are you hoping to cache? The OR build times or the flow tests in ORFS?

QuantamHD:
When you use a target as a tool in another build rule or test like in ORFS. You are required to mark it with a cfg tag. This is typically "exec" which means that it's meant to run on the current host not the target environment. As part of cfg="exec" openroad will get build with -c opt by default.

```
example_library = rule(
    implementation = _example_library_impl,
    attrs = {
        ...
        "_compiler": attr.label(
            default = Label("//tools:example_compiler"),
            allow_single_file = True,
            executable = True,
            cfg = "exec",
        ),
    },
)
```

oharboe:
> What exactly are you hoping to cache? The OR build times or the flow tests in ORFS?

Today I cache OR (through docker images), flow steps, synthesis, etc. Sometimes this is nice because I can modify and test only a step. Ideally the dependencies take care of everything and I only build from the step I changed in the flow, sometimes I manually specify exactly which step I want to test locally based on cached artifacts. I also use cached artifacts to pin versions(such as .lib and .lef for a mock sram). The latter has been a big success on one of the projects I'm working on, combined with paralell synthesis(in hiearchical mode, I can easily do synthesis on each kept/non-flattened module in parallel), synthesis went from 20 minutes to 2 minutes.

I'm also hoping to get distributed builds set up. For now I'm logging into a server from a laptop to do the build for a PR, then I pull re-run that build on a laptop to get the artfacts.

So lots of nice use-cases for caching in bazel :-)

oharboe:
> When you use a target as a tool in another build rule or test like in ORFS. You are required to mark it with a cfg tag. This is typically "exec" which means that it's meant to run on the current host not the target environment. As part of cfg="exec" openroad will get built with -c opt by default.
> 
> ```
> example_library = rule(
>     implementation = _example_library_impl,
>     attrs = {
>         ...
>         "_compiler": attr.label(
>             default = Label("//tools:example_compiler"),
>             allow_single_file = True,
>             executable = True,
>             cfg = "exec",
>         ),
>     },
> )
> ```
> 
> https://bazel.build/extending/rules

Thanks!

@nekronos @hovind FYI

oharboe:
@QuantamHD bazel-orfs is already using `cfg = "exec",`, so all we're waiting for there is for OpenROAD to be fully switched to MODULE.bazel before I can use OpenROAD as a tool from ORFS directly.

https://github.com/The-OpenROAD-Project/bazel-orfs/blob/48cc8783f285d2a274b1613004dcbea21195f49c/openroad.bzl#L390C1-L394C3

However, this issue makes "bazel test ..." fast. Running "bazel test ..." on OpenROAD (it fails today since we're not fully switched to MODULE.bazel and  lots of 3rd party tests that fail are run) would be excruciatingly slow on the integration tests and almost certainly not what the user ever intended.

Unless all OpenROAD tests are intended to run in minutes and all longer running tests belong in ORFS?

Did I just happen to stumble upon much slower tests when I converted test/recursion to Bazel?

test/recursion, even when run in parallel takes hours(unknown how long, it times out) with OpenROAD debug.

oharboe:
The policy here is what is tricky, implementation is trivial.

Co closing the PR for now. We can easily re-open it if it just so happens that this ends up being the exact solution we want.

QuantamHD:
> @QuantamHD bazel-orfs is already using `cfg = "exec",`, so all we're waiting for there is for OpenROAD to be fully switched to MODULE.bazel before I can use OpenROAD as a tool from ORFS directly.
> 
> https://github.com/The-OpenROAD-Project/bazel-orfs/blob/48cc8783f285d2a274b1613004dcbea21195f49c/openroad.bzl#L390C1-L394C3
> 
> However, this issue makes "bazel test ..." fast. Running "bazel test ..." on OpenROAD (it fails today since we're not fully switched to MODULE.bazel and lots of 3rd party tests that fail are run) would be excruciatingly slow on the integration tests and almost certainly not what the user ever intended.
> 
> Unless all OpenROAD tests are intended to run in minutes and all longer running tests belong in ORFS?
> 
> Did I just happen to stumble upon much slower tests when I converted test/regression to Bazel?
> 
> test/regression, even when run in parallel takes hours(unknown how long, it times out) with OpenROAD debug.
> 
> In fact... test/recursion can't ever work with OpenROAD debug builds, so `bazel test ...` with debug build is _always_ wrong for test/regression. There is hard upper timeout of 3600 seconds.

You can define --test_timeout and set any bound you want. Debug performance is pretty slow, and you wouldn't typically do //... in debug mode, but that doesn't mean it's wrong.

Inside Google there's a system called blueprint that tells the CI which targets to run and in which configurations. This is conceptionally similar to GitHub's yml and Jenkins config files.

I wouldn't get hung up on trying to prevent users from doing silly things like running tests in debug mode.

PR #6965, opened 2025-04-01
Upf: choose isolation

LucasYuki:
The smallest isolation cell was selected without considering whether inverters were needed, so in some cases, the smallest isolation plus the required inverters had a bigger area than the isolation cell that didn't require any inverters.

Changes:
- `find_smallest_isolation` now chooses the smallest area of isolation cell plus the required inverters.
- `isolation_select` test added.
- `use_interface_cell` tcl/upf command:
  - `lib_cells` argument now supports lists;
  - `interface_implementation_name` argument has been added to be UPF 2.1 compliant.
- Additional isolation cells options added to the tests:
  - `test/upf_test`;
  - `test/aes`.

### DIFF
--- a/src/upf/README.md
+++ b/src/upf/README.md
@@ -134,9 +134,10 @@ This command sets the interface cell.
 
 ```tcl 
 use_interface_cell
-    [-domain domain]
-    [-strategy strategy]
-    [-lib_cells lib_cells]
+    -domain domain
+    -strategy strategy
+    -lib_cells lib_cells
+    [interface_implementation_name]
 ```
 
 #### Options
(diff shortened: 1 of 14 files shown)


maliberty (changes requested):

src/upf/README.md
```diff
@@ -134,9 +134,10 @@ This command sets the interface cell.
 
 ```tcl 
 use_interface_cell
-    [-domain domain]
-    [-strategy strategy]
-    [-lib_cells lib_cells]
+    -domain domain
+    -strategy strategy
+    -lib_cells lib_cells
+    [interface_implementation_name]
```
maliberty, line 140: UPF gives this as mandatory

maliberty: image

LucasYuki: Since we don't use this argument and it was added just now, I made them optional.
Do you think it is better to make it mandatory?

maliberty: I think it better to follow the standard even if we don't actually need it.

maliberty (approved):

PR #6967, opened 2025-04-01
mpl: always use best valid result

AcKoucher:
### Context
When running Secure-CI for #6649, some results were considerably different than master which doesn't seem to make sense as there's no functionality changes.

### Cause of the Differences
Inside SA, we have a mechanism to store the best valid result. This was made in order to ensure MPL converges if a valid result was generated at any point during annealing. However, this best stored result is only actually used if the final result is invalid. _We should use the best valid result even if the final iteration result is a valid one, because we might have generated a good result during the early iterations that ended up being discarded._

I.e., what's happening is that the change in the random number generator affected the path SA takes and some good results from early iterations that we were luckily choosing are now being discarded.

### Changes
Use the best valid result stored even if the final result is valid.
With these changes, the results should improve compared to master and the results between base/test for #6649 should get more consistent.


QuantamHD:
Noice!

maliberty:
Are you running a secure CI for this?  Hopefully we just see improvements...

AcKoucher:
Re-running Secure-CI.

AcKoucher:
uW is failing due to high congestion in GRT post-drt. I'm taking a look.

AcKoucher:
Re-ran CI with a lower routing layer adjustment for uW (0.15 -> 0.10) and it passed. However, ng5/swerv_wrapper is failing in detail placement. I'll investigate.

AcKoucher:
`ngt45/swerv_wrapper`  is faling during GRT due high congestion. The only difference with master is that the order of the small RAMs changed inside their array. I'm taking a look.

AcKoucher:
[#3190](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3190) for decreasing routing layer adjustment on `ngt45/swerv_wrapper`. Running final Secure-CI to double-check everything.

AcKoucher:
@eder-matheus Secure-CI passed.

eder-matheus (approved):

PR #6968, opened 2025-04-01
drt: update inst

bnmfw:
Supports #5867.

### updateInst
updateInst inputs an inst and updates all its PA data. It tries to disrupt these structures as least as possible. The function can be called both for an inst that was created and needs to have PA data created, as well as when an instance is moved and needs to have PA data updated for its new position. This function does not update rows of instances.

Two APIs were created:
- `updateInst(frInst* inst)` simply expects an existing inst and updates it.
- `updateInst(odb::dbDatabase* db, odb::dbInst* db_inst)` expects a db_inst and creates a frInst if one does not exist. It also return the respective frInst* regardless if it was created by the function or not.


bnmfw:
clang-tidy is really confused

PR #6969, opened 2025-04-02
ord: exit with error with invalid commandline arguments

gadfort:


maliberty (approved):

oharboe:
👍I just ran into this today. In Bazel -python is not supported yet, this caused strange followon errors.

PR #6973, opened 2025-04-02
bazel: wire up first gtests

hzeller:


src/odb/test/cpp/sky130_test_fixture.h
```diff
@@ -26,7 +26,9 @@ class Sky130TestFixutre : public ::testing::Test
         db_.get(), &logger_, /*ignore_non_routing_layers=*/false);
     lib_ = OdbUniquePtr<odb::dbLib>(
         lef_reader.createTechAndLib(
-            "sky130", "sky130", "data/sky130hd/sky130_fd_sc_hd.tlef"),
+            "sky130",
+            "sky130",
+            "src/odb/test/data/sky130hd/sky130_fd_sc_hd.tlef"),
```
QuantamHD, line 31: This is going to break the ctest version of the test. We should probably introduce some helper function that checks if the env is bazel or ctest and modify the path.

hzeller: yeah, I suspect so. Since I can't run the cmake build (that is like soo last week), I'll let the CI figure out what is happening :)

hzeller: Fixed.

src/odb/test/cpp/sky130_test_fixture.h
```diff
@@ -25,8 +30,10 @@ class Sky130TestFixutre : public ::testing::Test
     odb::lefin lef_reader(
         db_.get(), &logger_, /*ignore_non_routing_layers=*/false);
     lib_ = OdbUniquePtr<odb::dbLib>(
-        lef_reader.createTechAndLib(
-            "sky130", "sky130", "data/sky130hd/sky130_fd_sc_hd.tlef"),
+        lef_reader.createTechAndLib(  //
```
maliberty, line 33: Is there a purpose to the `//` ?

hzeller: yes, it forces `clang-format` to format the way I want it to. Without that, it would put the macro in one line and the string literal into another line.
Adding EOL comments like this is a good way to 'help' clang-format do the right thing.

maliberty: Perhaps
```
    dbLib* lib = lef_reader.createTechAndLib(
        "sky130", "sky130", DATA_PREFIX "data/sky130hd/sky130_fd_sc_hd.tlef");
    lib_ = OdbUniquePtr<odb::dbLib>(lib, &odb::dbLib::destroy);
```
avoid the wrapping issue and is a bit clearer imho.

hzeller: well, but that is not what clang-format would do. So we have to force its hand a bit.

maliberty: That is what it would do as I formatted it before sending it.

QuantamHD (approved):

maliberty (approved):

PR #6975, opened 2025-04-03
bazel: add src/grt/test regression tests

oharboe:
Another step in collaboration and discussion.

To test all tests, run:

    bazel test //test:* //src/grt/test:* //src/gpl/test:* -c opt

Later, when WORKSPACE has all been moved to MODULE.bazel, simply:

    bazel test ... -c opt


oharboe:
@hzeller Comments welcome! I'm studying the regression test use-case here a bit more and come up with an articulation of how things can be specialized per module.

oharboe:
Added gpl: copy & paste of a small file + copy in list from CMakeLists.txt.

Is this all there's to it?

Looking in CMakeLists.txt, there's some gtest files, but that's being covered in https://github.com/The-OpenROAD-Project/OpenROAD/pull/6975 presumably.

If this is satisfactory, where to go from here?

- OpenROAD in Bazel needs to support Python before all the functionality in CMakeLists.txt is covered
- Afterwards retire CmakeLists.txt or keep both alive? If both are to be kept alive, do we update Bazel to read in or_integration_tests() using a repository rule for a DRY solution?

I think it makes sense to settle on a pattern and plan for the three modules we have now before proceding with converting more.

```
$ find . | grep regression\$
./sta/test/regression
./pdn/test/regression
./gui/test/regression
./utl/test/regression
./rsz/test/regression
./ppl/test/regression
./rmp/test/regression
./ant/test/regression
./par/test/regression
./pad/test/regression
./stt/test/regression
./psm/test/regression
./gpl/test/regression
./ifp/test/regression
./cts/test/regression
./grt/test/regression
./dpl/test/regression
./dpo/test/regression
./tap/test/regression
./dst/test/regression
./rcx/test/regression
./fin/test/regression
./dft/test/regression
./dbSta/test/regression
./upf/test/regression
./mpl/test/regression
./drt/test/regression
./odb/test/regression
```

src/gpl/test/BUILD
```diff
@@ -0,0 +1,60 @@
...
+                ["**/*"],
+                exclude = [
+                    test + "." + ext
+                    for test in TESTS
+                    for ext in [
+                        "tcl",
+                        "py",
+                    ]
+                ],
+            ),
```
QuantamHD, line 57: Instead of the glob you might want to put them into a filegroup and references them in the data dep.

oharboe: This is used once, so I would still use a glob, it would simply add a level of indirection. Is it a matter of syntactic sugar or is it in some way different?

hzeller: It would give a name to this set of files (as we might need it in other places as well, and it also allows to then compose things from smaller parts, such as pdk data and tcl scripts)

src/grt/test/BUILD
```diff
@@ -0,0 +1,101 @@
...
+                    test + "." + ext
+                    for test in TESTS
+                    for ext in [
+                        "tcl",
+                        "py",
+                    ]
+                ],
+            ),
+    )
+    for name, kwargs in {test: {} for test in TESTS}.items()
```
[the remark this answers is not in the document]
oharboe: fixed

src/gpl/test/BUILD
oharboe: fixed

src/grt/test/BUILD
```diff
@@ -0,0 +1,102 @@
...
+                "tcl",
+                "py",
+            ]
+        ],
+    ),
+)
+
+[regression_test(
+    name = name,
+    data = ["test_resources"],
```
hzeller, line 101: to refer to a local target, you'd typically prefix it with a colon `:test_resources`.
(you already did it in the other BUILD file)

test/BUILD
```diff
@@ -1,49 +1,61 @@
...
+    srcs = glob(["**/*"]),
+    visibility = ["//visibility:public"],
+)
+
+[
+    regression_test(
+        name = name,
+        tags = kwargs.get("tags", []),
+    )
+    for name, kwargs in (
```
hzeller, line 27: this does look somewhat complicated to me.

why not have something like

```python
   tags = ["manual"] if test in ["timing_api", "...and soo on"] else []
   size = "enormous" if test in ["aes_nangate45"] else "normal"
```

That would make it much more immediately readable without having to understand three merged hash-maps generated by list-comprehension.

src/grt/test/BUILD
oharboe: fixed

oharboe: Bazel linting doesn't catch this. I wonder why. When would you not want the ":" for local files or targets?

src/grt/test/BUILD
```diff
@@ -0,0 +1,102 @@
...
+                "py",
+            ]
+        ],
+    ),
+)
+
+[regression_test(
+    name = name,
+    data = [":test_resources"],
+) for name in TESTS]
```
hzeller, line 102: to avoid confusion about the `name = name` I'd probably use a different run-variable, e.g. `for test_name in TESTS`

test/BUILD
hzeller: Only add flexibility when needed. Readability is the key. Apply the KISS principle.

Part of using bazel is to have not the incomprehensible mess that cmake files require users to endure; so we're not in a race to match that :)

src/grt/test/BUILD
hzeller: Don't know if the linting looks closely at it, but it would be nice (I have built some canonicalization in my `bant` tool, but it also only looks at the surface)

oharboe:
@maliberty @vvbandeira Flaky test on master?

grt.congestion1.py

maliberty:
I believe this is the one @eder-matheus is looking at.

test/BUILD
oharboe: fixed

src/grt/test/BUILD
oharboe: fixed

oharboe:
@maliberty You're up. Fixed feedback from @hzeller

src/grt/test/BUILD
hzeller: you still use `for name in TESTS` not `for test_name in TESTS`

test/BUILD
```diff
@@ -1,49 +1,68 @@
...
+    "jpeg_sky130hs.tcl",
+    "aes_nangate45.tcl",
+    "gcd_nangate45.tcl",
+    "tinyRocket_nangate45.tcl",
+]
+
+[
+    regression_test(
+        name = test_name,
+        tags = [] if test_name in COMPULSORY_TESTS else ["manual"],
```
hzeller, line 65: did you mean to add `size =` ... .with a conditional value depending on `test_name in BIG_TESTS` ?

test/regression.bzl
```diff
@@ -0,0 +1,46 @@
+"""Instantiate a regression test based on .py or .tcl
+files using resources in //test:regression_resources"""
+
+def _pop(kwargs, key, default):
+    """BUILD does not support kwargs, use None as a "kwargs at home" workaround"""
```
hzeller, line 5: bazel actually can go beyond kwargs. There are ways to declare what parameters are expected in a rule. I defer to @QuantamHD who knows all about this and can point you to the right documentation.

test/BUILD
oharboe: fixed

test/regression.bzl
oharboe: This is a macro, not a rule, though.

hzeller: Yeah, I don't have any experience with whatever the stuff is called *.bzl files, I try to stay away from stuff that looks like Python :)

oharboe: There are two levels:

- macros(just a few def functions). This is pretty straightforward and what I do here.
- then there are rules(repository, normal rules) and module extensions. This second bit is powerful, fascinating and introduces many new concepts.

I don't think a rule would make the code more readable or improve the use case here. Maybe later if the use case becomes more sophisticated we could use something more than macros.

hzeller: yeah, sounds macro is the right approach.

hzeller (approved):
I think this looks like a good start getting these tests going. Thanks for working on it!

hzeller:
yay [Executed 116 out of 116 tests: 116 tests pass.](https://github.com/The-OpenROAD-Project/OpenROAD/actions/runs/14269809950/job/40000204715?pr=6975#step:5:3683)

test/BUILD
```diff
@@ -1,49 +1,69 @@
...
+    "timing_api_3",
+    "timing_api_4",
+    "two_designs",
+]
+
+# TODO There are tests in this folder that were not enabled in
+# CMakeLists.txt, presumably because they are long running.
+# These should be added with tags = ["manual"], i.e not
+# run with 'bazel test ...', only when invoked explicitly
+# and the "size" set to "enormous".
```
maliberty, line 45: This is incorrect.  When @QuantamHD converted to using cmake/ctest he didn't convert the flow tests.  They are run by [Jenkins ](https://github.com/The-OpenROAD-Project/OpenROAD/blob/1bd109420d604fdcc8834c7f6379d2128836dd34/Jenkinsfile#L29) and are executed for every PR.

maliberty: `make test` will not run them though which is unfortunate

oharboe: I've tried to correct the comments.

What else is there to do on this PR?

@QuantamHD @hzeller What are your thoughts on local developer workflow expectations vs. CI?

My thinking is that the developer would expect the following to run reasonable set of local tests:

    bazel test -c opt ...

In CI, I don't think it is a big burden to require the test script to craft bazel test statements that picks out tests of a desired range in time and size that are a fit for the available testing resources and testing scope.

hzeller: I think the PR is ready to get submitted, we always can iterate later.

My development expectation would be: `bazel test -c opt ...` should run most test unless they take more than a few minutes in opt mode in which case they should be marked manual (hopefully at some point most *.cc files have a corresponding *_test.cc file with local unit tests to have fast local tests that don't require expensive end2end situations most of the time)

CI should run all the tests (maybe we could even have a configuration for running all the tests, that inclues manual ones as well). And yes, if the CI needs to have a slightly longer command line to accommodate these, that is ok.

oharboe: Today all tests are not run on PR, there for instance some confidential tests. These are are not run on all PRs.

In Bazel it is easy to arrange for things such that stable bazel oneliners implement a policy of which tests to run where in CI when and that these queries will discover the right tests to run without having to update the CI scripts.

maliberty: The fact that they don't run in ctest is not a feature.  Prior to ctest they were run locally as well by the regression script.  It got broken during the move to ctest and nobody has fixed it yet.  I'd like to avoid the same mistake in bazel.  I can merge this as is but don't want to fall into the same later=never trap.

oharboe: Thanks!

I'm confused as to which tests are not available via Bazel.

I have tried to replicate what I conceived as the policy in CMakeLists.txt and is also congruent with expectations from Bazel users(see comments above): fast local tests for developers + longer tests can be run locally if invoked explicitly and from CI.

maliberty (approved):

PR #6982, opened 2025-04-03
included libomp-15-dev to clang-tidy action

sombraSoft:
We're seeing linting issues like in this [PR](https://github.com/The-OpenROAD-Project/OpenROAD/pull/6916)
`warning: 'omp.h' file not found [clang-diagnostic-error]`
The underlying problem was that the correct version of libomp was not installed as a dependency.

image

### DIFF
--- a/.github/workflows/github-actions-clang-tidy.yml
+++ b/.github/workflows/github-actions-clang-tidy.yml
@@ -16,7 +16,7 @@ jobs:
         id: review
         with:
           build_dir: './build'
-          cmake_command: cmake . -B build
+          cmake_command: cmake . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=on
           config_file: '.clang-tidy'
           exclude: "*/codeGenerator/templates/*"
           split_workflow: true
--- a/.github/workflows/github-actions-clang-tidy.yml
+++ b/.github/workflows/github-actions-clang-tidy.yml
@@ -20,7 +20,7 @@ jobs:
           config_file: '.clang-tidy'
           exclude: "*/codeGenerator/templates/*"
           split_workflow: true
-          apt_packages: libomp-15-dev,libomp-14-dev
+          apt_packages: libomp-15-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/github-actions-clang-tidy.yml
```diff
@@ -16,10 +16,11 @@ jobs:
         id: review
         with:
           build_dir: './build'
-          cmake_command: cmake . -B build
+          cmake_command: cmake . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=on
```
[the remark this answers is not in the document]
sombraSoft: great!
I have removed it from this PR

maliberty (approved):

.github/workflows/github-actions-clang-tidy.yml
```diff
@@ -20,6 +20,7 @@ jobs:
           config_file: '.clang-tidy'
           exclude: "*/codeGenerator/templates/*"
           split_workflow: true
+          apt_packages: libomp-15-dev
```
vvbandeira, line 23: The Dockerfile in the action installs clang 14 and 15. Let's err in the side of caution and install both for omp as well?

sombraSoft: you got it!, just added it.

sombraSoft: we're now getting a
`
The following packages have unmet dependencies:
   libomp-14-dev : Conflicts: libomp-x.y-dev
`

and that's because development packages (-dev) install header files (like omp.h) into standard system locations (e.g., /usr/include). If you tried to install both versions, they would attempt to place their potentially different omp.h files (and other headers) in the same locations, leading to an inconsistent and broken state. apt prevents this conflict by refusing to install both.

Because of that I will have to remove `libomp-14-dev`

maliberty: This is only for the purpose of clang-tidy finding the header as we are not compiling any code here.  I don't think 14 vs 15 will matter.

PR #6987, opened 2025-04-04
bazel: run tests in the CI

hzeller:
We have now the first tests wired up, let's run them. Switch to `-c opt` mode to not have them time out
(unfortunately, that also means a longer initial build time, but things will be cached)

### DIFF
--- a/.github/workflows/github-actions-bazel-build.yml
+++ b/.github/workflows/github-actions-bazel-build.yml
@@ -32,10 +32,9 @@ jobs:
 
       - name: Build
         run: |
-          # There are no cc_test()s yet, so just build for now.
           # For now, just run, but don't fail the whole CI, as bazel build is
           # experimental at this stage.
-          bazel build --keep_going --show_timestamps --curses=no ... || /bin/true
+          bazel test -c opt --keep_going --show_timestamps --test_output=errors --curses=no --force_pic ... || /bin/true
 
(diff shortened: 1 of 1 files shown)


maliberty:
The action already takes ~41m to run.  There is no caching as the runners are ephemeral and we don't have a web cache yet.  Increasing the build time sounds concerning as pr-merge/head are ~1h.  Let's see how this PR goes.

hzeller:
This first time will probably take a while, I suspect up to 90min.
The bazel cache is stored in github actions, so as long as they run regularly, only incremental bulids happen later, even without of any web cache.
I've seen they only take a few minutes otherwise.

Anyway, how is the set-up of a web cache going ?

maliberty:
Last time we did something like this it took a while to get permission to make it public (Google owns the project and had to approve).   @vvbandeira is looking into it.

How does the caching in GH actions work?

maliberty:
It is >40m so far in this build.  If there were a cache I would expect it to be faster.  We have run this action many times so it doesn't appear to be working.

hzeller:
this is the first time it runs with `-c opt`, so there is no cache yet.

maliberty (approved):

maliberty:
Build time goes up to 51m but when I re-triggered the bazel build action it was a perfect cache hit.

PR #6988, opened 2025-04-04
openroad: add -cd option

oharboe:
This makes for less convoluted test code and use-cases for Bazel

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -229,15 +229,14 @@ static void handler(int sig)
   raise(sig);
 }
 
-namespace {
 // Done launching, show splash and switch to working folder
-bool launched(int& argc, char* argv[])
+static bool completeLaunch(int& argc, char* argv[])
 {
   bool no_splash = findCmdLineFlag(argc, argv, "-no_splash");
   if (!no_splash) {
     showSplash();
   }
-  const auto workingFolder = findCmdLineKey(argc, argv, "-cd");
+  const char* workingFolder = findCmdLineKey(argc, argv, "-cd");
   if (workingFolder) {
     // switch pwd to the working folder. This is useful when
     // OpenROAD must be launched from one folder and the files
@@ -252,7 +251,6 @@ bool launched(int& argc, char* argv[])
   }
   return no_splash;
 }
-}  // namespace
 
 int main(int argc, char* argv[])
 {
@@ -311,7 +309,7 @@ int main(int argc, char* argv[])
     ord::OpenRoad::setOpenRoad(the_tech_and_design.design->getOpenRoad());
     ord::initOpenRoad(interp, log_filename, metrics_filename);
 
-    launched(cmd_argc, cmd_argv);
+    completeLaunch(cmd_argc, cmd_argv);
 
     utl::Logger* logger = ord::OpenRoad::openRoad()->getLogger();
     if (findCmdLineFlag(cmd_argc, cmd_argv, "-gui")) {
@@ -479,7 +477,7 @@ static int tclAppInit(int& argc,
 
     ord::initOpenRoad(interp, log_filename, metrics_filename);
 
-    bool no_splash = launched(argc, argv);
+    bool no_splash = completeLaunch(argc, argv);
 
     const char* threads = findCmdLineKey(argc, argv, "-threads");
     if (threads) {
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -229,6 +229,31 @@ static void handler(int sig)
   raise(sig);
 }
 
+namespace {
+// Done launching, show splash and switch to working folder
+bool launched(int& argc, char* argv[])
+{
+  bool no_splash = findCmdLineFlag(argc, argv, "-no_splash");
+  if (!no_splash) {
+    showSplash();
+  }
+  const auto workingFolder = findCmdLineKey(argc, argv, "-cd");
+  if (workingFolder) {
+    // switch pwd to the working folder. This is useful when
+    // OpenROAD must be launched from one folder and the files
+    // that it needs to run scripts is in a different folder.
+    //
+    // Bazel is a use-case.
+    if (chdir(workingFolder) != 0) {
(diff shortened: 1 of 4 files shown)


oharboe:
@hzeller @QuantamHD FYI

oharboe:
> @oharboe couldn't the tests just do the change directory if needed? it seems odd to me to add this to the openroad commandline

If you look at the code that it replaces, I'm sure you can symphethize with me that it is less of an eyesore + this is also useful [elsewhere in bazel ](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3043/files#diff-8a50ba67f11474fad173b3b2ad0631d87c012f1a9341a95c81d22ff84648f745), or Buck2 for that matter(I'm sure).

I don't see the benefit of conditional code since `cd $(pwd)` is a safe fast no-op. I'd rather avoid the iffy code.

gadfort:
@oharboe I understand that part, but do other projects that rely on `bazel` need new arguments just to be able to do testing? I can't image this isn't a usecase they have already solved for testing in `bazel`. I would be happier with this if you can point to why bazel cannot handle the `cd` part (or another project that needed to do this).

oharboe:
@gadfort I suspect this can be avoided and later retired as we didn't end up needing it when patching the openroad binary from Docker ORFS and using it in Bazel. Using an OpenROAD binary from the ORFS docker image was another stopgap which we'll phase out when enough issues have been sorted out in Bazel OpenROAD support(including adding Qt support).

src/Main.cc
```diff
@@ -229,6 +229,31 @@ static void handler(int sig)
...
 
+namespace {
+// Done launching, show splash and switch to working folder
+bool launched(int& argc, char* argv[])
+{
+  bool no_splash = findCmdLineFlag(argc, argv, "-no_splash");
+  if (!no_splash) {
+    showSplash();
+  }
+  const auto workingFolder = findCmdLineKey(argc, argv, "-cd");
```
hzeller, line 240: Really only use `auto` if the type is complicated (e.g. some templated construct of sorts), otherwise
Avoid `auto` if writing out the type is not a burden. In _particular_ as the reader needs to know the type to understand that it is compatible with the `chdir()` function below.

So make it `const char* workingFolder = ...`

src/Main.cc
```diff
@@ -285,9 +310,8 @@ int main(int argc, char* argv[])
         = std::make_unique<ord::Design>(the_tech_and_design.tech.get());
     ord::OpenRoad::setOpenRoad(the_tech_and_design.design->getOpenRoad());
     ord::initOpenRoad(interp, log_filename, metrics_filename);
-    if (!findCmdLineFlag(cmd_argc, cmd_argv, "-no_splash")) {
-      showSplash();
-    }
+
+    launched(cmd_argc, cmd_argv);
```
hzeller, line 314: `launched()` does not sounds like verb a function typically would be named.

test/regression_test.sh
```diff
@@ -16,29 +16,12 @@ fi
...
-    )
-fi
 
 echo "Directory: ${PWD}"
 echo "Command:   $OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit  $TEST_NAME.$TEST_EXT > $LOG_FILE"
 
 cd ${BAZEL_SAVED_PWD:-$SAVED_PWD}
-$OPENROAD_EXE $ORD_ARGS -no_splash -no_init -exit <(echo "$script_content") &> $LOG_FILE
+$OPENROAD_EXE $ORD_ARGS -cd $SAVED_PWD -no_splash -no_init -exit $TEST_NAME.$TEST_EXT &> $LOG_FILE
 cd $SAVED_PWD
```
hzeller, line 25: why change to $SAVED_PWD here ? Looks like a leftover from before.

hzeller:
LGTM.
I think this is an option that improves quality-of-life independent of if you use bazel testing or not.
There are often situation where you have scripts and other inputs that might have relative references.

src/Main.cc
```diff
@@ -229,6 +229,31 @@ static void handler(int sig)
   raise(sig);
 }
 
+namespace {
+// Done launching, show splash and switch to working folder
+bool launched(int& argc, char* argv[])
```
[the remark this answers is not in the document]
oharboe: I thought the anonymous namespace was the modern C++ version of static.

src/Main.cc
oharboe: no opinion, I wish clang-format-14 would just do something consistent.

test/regression_test.sh
hzeller: if there is no bug now, don't do it as it might _mask_ a bug in waiting if there is. From the perspective of the shell script, you never changed the directory, only inside the process of openroad.

src/Main.cc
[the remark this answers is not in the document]
hzeller: Sure you _can_ use anonymous namespaces for functions as well, but don't do it simply because it is more modern.
The main concern is readability in code, and anonymous namespaces strictly reduce that for functions as you don't see immediately looking at the signature if it is of local linkage.

So my recommendation is generally: use anonymous namespaces for types (as they can't be 'staticified)', use `static` as always in C/C++ for local functions.

test/regression_test.sh
oharboe: Strike that: the cd $SAVED_PWD is necessary for the code below to work. The changing folder to launch openroad is an openroad launching concern that should not be narrowly scoped by a subshell to minimize surprises.

src/Main.cc
oharboe: fixed

src/Main.cc
oharboe: fixed

src/Main.cc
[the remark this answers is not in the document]
oharboe: fixed

QuantamHD:
This is a better solution to the problem https://github.com/The-OpenROAD-Project/OpenROAD/pull/6993

PR #6995, opened 2025-04-04
drt: find adjacent instance cluster

bnmfw:
Supports #5867.
Branched from #6916.

Essentially implements the APIs from #6522 into PA. Implements the same `getAdjacentInstance()` and `getAdjacentInstancesCluster()` APIs which will be used in incremental PA.


src/drt/src/pa/FlexPA.cpp
```diff
@@ -178,6 +196,69 @@ void FlexPA::prep()
...
+  int cnt = 0;
+
+  omp_set_num_threads(router_cfg_->MAX_THREADS);
+  ThreadException exception;
+#pragma omp parallel for schedule(dynamic)
+  for (frInst* unique_inst : unique) {
+    try {
+      // only do for core and block cells
+      // TODO the above comment says "block cells" but that's not what the code
+      // does?
```
rovinski, line 217: I put this in a few years ago. Have we figured it out yet?

maliberty: I think the comment is just wrong and it is only std cells.   For blocks I don't think the code for instance and rows would work well.

src/drt/src/pa/FlexPA_row_pattern.cpp
```diff
@@ -63,6 +63,60 @@ std::vector<std::vector<frInst*>> FlexPA::computeInstRows()
...
+
+  std::reverse(adj_inst_cluster.begin(), adj_inst_cluster.end());
+  adj_inst_cluster.push_back(inst);
+
+  frInst* right_inst = getAdjacentInstance(inst, right);
+  while (right_inst != nullptr) {
+    adj_inst_cluster.push_back(right_inst);
+    // the left instance can be ignored, since it was added in the line above
+    right_inst = getAdjacentInstance(right_inst, right);
+  }
```
osamahammad21, line 115: The multiple calls to getAdjacentInstance does not seem the most efficient I think. Each call searches for the inst which is `log(N)`. Worst case O would be `N.log(N`). Alternatively, we could grab the iterator for the passed instance once and move left and right untill we get to the end of the pattern from both ends. This would reduce T to `log(N) + N` and O to `N`.

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_row_pattern.cpp
bnmfw: Yep, you are absolutely right. I transferred the logic without thinking too much about how the logic would behave on the different structure, my bad.

src/drt/src/pa/FlexPA_row_pattern.cpp
```diff
@@ -63,6 +63,60 @@ std::vector<std::vector<frInst*>> FlexPA::computeInstRows()
...
+    left_inst = inst_1;
+    right_inst = inst_2;
+  } else {
+    left_inst = inst_2;
+    right_inst = inst_1;
+  }
+
+  if (left_inst->getBoundaryBBox().xMax() != right_inst->getOrigin().getX()) {
+    return false;
+  }
```
osamahammad21, line 82: Could we rely on the bbox instead of the origin in this code? There is a reverted PR that uses the instance's origin correctly which would make this incorrect (origin may be the center of the bbox not lower left). I am planning to rework this PR soon.

src/drt/src/pa/FlexPA_row_pattern.cpp
```diff
@@ -63,6 +63,60 @@ std::vector<std::vector<frInst*>> FlexPA::computeInstRows()
...
+
+  return true;
+}
+
+std::vector<frInst*> FlexPA::getAdjacentInstancesCluster(frInst* inst) const
+{
+  auto inst_it = insts_set_.find(inst);
```
osamahammad21, line 89: const

src/drt/src/pa/FlexPA_row_pattern.cpp
```diff
@@ -63,6 +63,60 @@ std::vector<std::vector<frInst*>> FlexPA::computeInstRows()
...
+  std::vector<frInst*> adj_inst_cluster;
+
+  adj_inst_cluster.push_back(inst);
+
+  if (inst_it != insts_set_.begin()) {
+    auto target_inst_it = inst_it;
```
osamahammad21, line 95:
```suggestion
    auto curr_inst_it = inst_it;
```
Just a suggestion

osamahammad21 (changes requested):

Issue #6318, opened 2024-12-06
cts: regression: crash after commit d7c5a8ae4b9cba8318906490d13fd83ea62aefaf


titan73:
### Describe the bug

After the commit:

```
commit d7c5a8ae4b9cba8318906490d13fd83ea62aefaf => BAD
Merge: c00d74874 928902cae
Author: Matt Liberty <mliberty@precisioninno.com>
Date:   Thu Nov 21 12:59:23 2024 -0800

    Merge pull request #6134 from arthurjolo/cts_enhance_clkgater_handling
    
    Cts enhance clkgater handling
```

CTS causes a crash with no backtrace:

```
[WARNING CTS-0002] No paths found for pin I_digital_top/alwayson_top_i/i2c_top_i/u_i2c_bus_monitor/stop_o_reg/CP.
[WARNING CTS-0002] No paths found for pin I_digital_top/alwayson_top_i/i2c_top_i/u_i2c_bus_monitor/start_o_reg/CP.
[WARNING CTS-0002] No paths found for pin I_digital_top/nvm_control_top_0/u_nvm_mmu_softip_top/u_nvm_ctrl/c_nvm_wrapper/s_trim_addr_regx0x/CP.
[WARNING CTS-0002] No paths found for pin I_digital_top/nvm_control_top_0/u_nvm_mmu_softip_top/u_nvm_ctrl/c_nvm_wrapper/s_trim_addr_regx1x/CP.
[WARNING CTS-0002] No paths found for pin I_digital_top/nvm_control_top_0/u_nvm_mmu_softip_top/u_nvm_ctrl/c_nvm_wrapper/s_trim_addr_regx2x/CP.
Segmentation fault (core dumped)
```

Full log before commit: [ok.zip](https://github.com/user-attachments/files/18041167/ok.zip)

Full log after commit: [bad.zip](https://github.com/user-attachments/files/18041170/bad.zip)

### Expected Behavior

No crash

### Environment

```shell
Commit-id provided
```

### To Reproduce

Not been able to make a testcase yet

titan73:
I get this width/height with 0 value twice after commit:
```
[INFO CTS-0025]     Width:  0.0000.
[INFO CTS-0026]     Height: 0.0000.
 Level 1
    Direction: Vertical
    Sinks per sub-region: 1
    Sub-region size: 0.0000 X 0.0000
[INFO CTS-0034]     Segment length (rounded): 1.
```

[...]

PR #6996, opened 2025-04-04
Cts fix recursive arrival time

arthurjolo:
Fixes [#6318](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6318)

During the balancing of the Register tree and macros tree CTS uses a recursive function that goes through the clock gaters to propagate until the registers pins and get the arrival times to correctly balance the trees. This recursive function was wrongly propagating the clock with it reaches a register data pin, thinking it is clock gater. With this the recursive function can end in loops and crash.

This PR adds a function to CTS that checks if the recursive function should keep propagating or if it should stop. We now only propagate if the recursive function reaches:

- Clock gater clock pin.
- Buffer/inverter.
- Latch Data pin (an improvised clock gater).
- Combinational cell.

Any ohter cells and pins the recursive function does not propagate, avoiding leaking into data paths.


src/cts/src/TritonCTS.cpp
```diff
@@ -2277,6 +2278,30 @@ void TritonCTS::computeSinkArrivalRecur(odb::dbNet* topClokcNet,
   }
 }
 
+bool TritonCTS::propagateClock(odb::dbITerm* input)
+{
+  odb::dbInst* inst = input->getInst();
+  sta::Cell* masterCell = network_->dbToSta(inst->getMaster());
+  sta::LibertyCell* libertyCell = network_->libertyCell(masterCell);
```
precisionmoon, line 2285: Need some bullet proofing here.   Some cells may not have LibertyCell.  Need to return false for those.

precisionmoon (approved):

maliberty:
@arthurjolo does this need any metrics update or is it ready to merge?

PR #7014, opened 2025-04-06
bazel: Remove `timeout` which is not used in regression test.

hzeller:
Remove the timeout that is not used in the regression test,
as the `"medium"` fallback prevented the test size (which is passed
as argument) to have an effect on the timeout.

### DIFF
--- a/test/regression.bzl
+++ b/test/regression.bzl
@@ -1,3 +1,6 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025-2025, The OpenROAD Authors
+
 """Instantiate a regression test based on .py or .tcl
 files using resources in //test:regression_resources"""
 
@@ -12,11 +15,17 @@ def _pop(kwargs, key, default):
 def regression_test(
         name,
         **kwargs):
-    test_files = native.glob([name + "." + ext for ext in [
-        # TODO once Python is supported, add the .py files to the
-        # "py",
-        "tcl",
-    ]])
+    # TODO: we should _not_ have the magic to figure out if tcl or py exists
+    # in here but rather in the BUILD file and just pass the resulting
+    # name = "foo-tcl", test_file = "foo.tcl" to this regression test macro.
+    test_files = native.glob(
+        [name + "." + ext for ext in [
+            # TODO once Python is supported, add the .py files to the
+            # "py",
+            "tcl",
+        ]],
+        allow_empty = True,  # Allow to be empty; see also TODO above.
+    )
     for test_file in test_files:
         ext = test_file.split(".")[-1]
         native.sh_test(
@@ -27,7 +36,6 @@ def regression_test(
             #
             # https://bazel.build/reference/be/common-definitions#test.size
             size = _pop(kwargs, "size", "small"),
-            timeout = _pop(kwargs, "timeout", "moderate"),
             srcs = ["//test:bazel_test.sh"],
             args = [],
             data = [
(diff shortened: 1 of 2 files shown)


hzeller:
CC @oharboe

oharboe (approved):

test/regression.bzl
```diff
@@ -1,3 +1,4 @@
+# -*- Python -*-
```
maliberty, line 1: Do we need to put this in every bzl file?  Why not just setup emacs to recognize the extension?

test/regression.bzl
```diff
@@ -27,14 +34,14 @@ def regression_test(
...
             size = _pop(kwargs, "size", "small"),
-            timeout = _pop(kwargs, "timeout", "moderate"),
             srcs = ["//test:bazel_test.sh"],
             args = [],
             data = [
                 "//:openroad",
                 "//test:regression_resources",
                 "//test:regression_test.sh",
             ] + test_files + _pop(kwargs, "data", []),
+            env_inherit = ["PATH"],  # To find system binaries such as `getconf`
```
maliberty, line 44: Does this imply that we are giving up sandboxing in the tests?

test/regression.bzl
hzeller: I should probably do that :)

test/regression.bzl
hzeller: Well, they never could be sandboxed to begin with as right now, the tcl scripts call unexpected programs  such as `getconf` (typically, on a regular system it might just work, but on my  stricter system (NixOS), it became apparent).

If this is something we need more often, maybe we should swig a tcl function that returns the value of `std::thread::hardware_concurrency`

test/regression.bzl
maliberty: I'm using
```
(defun auto-mode (expr mode)
  (setq auto-mode-alist (cons (cons expr mode)
			      auto-mode-alist)))

(auto-mode "\\.bazel$" 'python-mode)
(auto-mode "\\.bzl$" 'python-mode)
(auto-mode "BUILD$" 'python-mode)
(auto-mode "WORKSPACE$" 'python-mode)
```

test/regression.bzl
maliberty: If the only need is to get the number of cpus we could easily add that.  Are there any other programs?

hzeller: I don't know as this was the first it failed on.

test/regression.bzl
hzeller: Thanks, now added to my `~/.emacs`

test/regression.bzl
oharboe: Can/should we define nix environment to define what system installed tools bazel builds can/should rely on and hook it up in CI?

At what point do caching of builds/tests break?

test/regression.bzl
hzeller: Removed the `-*- Python -*-` and added a SPDX license identifier.

hzeller:
This should be ready to merge. #7021 captures the things to do to avoid this workaround.

hzeller:
I created a separate #7026 to replace all the calls to `getconf`, once that is merged, I can update this one (and update the descripption; it still contains the `allow_empty`)

hzeller:
Alright: here removed the sand-box breaking `PATH` again and adapted the CL description to capture the remaining changes.

What was to be addressed by `PATH` is done in #7026

maliberty (approved):

PR #2021, opened 2022-07-05
dpl use placement locked to distinguish spacer cell used as tapcells

jjcherry56:
Signed-off-by: James Cherry <cherry@parallaxsw.com>

PR #7026, opened 2025-04-07
Replace calls to `getconf` with cpu_count

hzeller:
Issues: #2021

### DIFF
--- a/src/drt/test/aes_nangate45.tcl
+++ b/src/drt/test/aes_nangate45.tcl
@@ -3,7 +3,7 @@ read_lef Nangate45/Nangate45_tech.lef
 read_lef Nangate45/Nangate45_stdcell.lef
 read_def aes_nangate45_preroute.def
 read_guides aes_nangate45.route_guide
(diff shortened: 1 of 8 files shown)


src/drt/test/aes_nangate45.tcl
```diff
@@ -3,7 +3,7 @@ read_lef Nangate45/Nangate45_tech.lef
 read_lef Nangate45/Nangate45_stdcell.lef
 read_def aes_nangate45_preroute.def
 read_guides aes_nangate45.route_guide
-set_thread_count [expr [exec getconf _NPROCESSORS_ONLN] / 4]
+set_thread_count [expr [cpu_count] / 4]
```
hzeller, line 6: Q: is it actually well-defined what happens if `set_thread_count` is called with zero ? Because a machine with 2 cores divided by 4 would result in zero.

maliberty: You get a warning and 1 thread

maliberty (approved):

PR #7029, opened 2025-04-08
bazel: don't allow empty glob() patterns.

hzeller:
Patterns that don't result in anything are certainly accidental, so have bazel double-check that we don't use such patterns. With bazel 8, this will be the default, so this prepares us for that as well.

Required to update a bunch of boost minor packaged versions as earlier ones had that exact problem which are now reported.


hzeller:
This also makes OR compatible with bazel 8, so it would compile with bazel 7 and bazel 8 (though the lock-file will be slightly different).

maliberty (approved):

PR #7033, opened 2025-04-08
bazel: add dbSta tests

oharboe:


src/dbSta/test/BUILD
```diff
@@ -0,0 +1,156 @@
...
+            [
+                test_name + ".*",
+            ],
+        ) + {
+            "report_cell_usage_file": [
+                "report_cell_usage.def",
+            ],
+            "write_verilog6": [
+                "write_verilog5.def",
+            ],
```
maliberty, line 142: This looks odd

maliberty (changes requested):

src/dbSta/test/BUILD
```diff
@@ -0,0 +1,156 @@
...
+    "write_verilog6",
+    "write_verilog7",
+    "write_verilog8",
+]
+
+ALL_TESTS = COMPULSORY_TESTS
+
+filegroup(
+    name = "regression_resources",
+    # Dependencies could be specified more narrowly per test case,
```
hzeller, line 58: Yeah, I think for now, we should just get all the tests in, but make sure once we're done, we revisit the filegroup assembly.

There might be some things even in the project structure that we can combine ? Looks like in many cases these are currently a bunch of symbolic links inside the project, but what if we had all test resources in `//test` as a filegroup or set of filegroups, define _there_ and re-use that. Not only improves that readability, but also bazel will be able to cache that one filegroup instead of having to build many filegroups that have essentially the same content but through symbolic links.

We still can leave the symbolic links in place to simplify the cmake build while it is still there, but ideally the bazel build will work without the need for the symbolic links as we use the original place and then refer to them by the convenient `//test:foobar` filegroup name.

Can you file an issue where we record all these 'post cranking out tests to bazel' ideas, so that we have a systematic way of addressing them afterwards ?

hzeller (approved):
Looks good, but see the request to file an issue where we can collect follow-up ideas.

src/dbSta/test/BUILD
hzeller: Filed #7042 for the later work on file groups so that we don't have to rely on symlinks.

src/dbSta/test/BUILD
oharboe: It is a pattern that I have adopted to handle the cases where the test uses files outside of the `test_name + ".*"` globl.

maliberty (approved):

PR #7037, opened 2025-04-08
bazel: test add dpo

oharboe:


oharboe:
@maliberty Since Python support is so limited anyway and this is a flaky test that we'll eventually move to Bazel, perhaps just disable it?

    792 - grt.congestion1.py (Failed)

maliberty (approved):

PR #7038, opened 2025-04-08
bazel: fin regression tests

oharboe:


src/fin/test/BUILD
```diff
@@ -0,0 +1,58 @@
...
+
+# From CMakeLists.txt or_integration_tests(TESTS
+COMPULSORY_TESTS = [
+    "gcd_fill",
+]
+
+ALL_TESTS = COMPULSORY_TESTS
+
+filegroup(
+    name = "regression_resources",
```
hzeller, line 11: Once most the tests are in, we can probably coalesce many of these and have groups defined somewhere in `//test` like `sky130_std_cells` or something so that we can refer to them in all the tests that need these and only have to define them once.

Not now, right now it is faster moving to have the local file-groups, just as a trajectory for "after".

hzeller (approved):
Thanks for cranking out wiring up all these tests to bazel!

src/fin/test/BUILD
oharboe: One concern at the time.... Get it under test first :-)

maliberty (approved):

PR #7039, opened 2025-04-08
bazel: gui regression tests

oharboe:


src/gui/test/BUILD
```diff
@@ -0,0 +1,29 @@
...
+# Copyright (c) 2022-2025, The OpenROAD Authors
+load("//test:regression.bzl", "regression_test")
+
+# From CMakeLists.txt or_integration_tests(TESTS
+ALL_TESTS = [
+    "supported",
+]
+
+[
+    filegroup(
```
maliberty, line 11: What are the [] around filegroup and regression_test for?

oharboe: it is a list comprehension.

This pattern is from the other foldes with multiple tests. I suppose more than one test will be added here in the future so this should turn out useful.

maliberty (approved):

PR #7045, opened 2025-04-08
pdn: retire cmake regression tests, already done in Bazel

oharboe:
Is there anything else that we need to do in Bazel before we take a page from Cortés book and retire cmake regression tests after they are added to Bazel?

We don't want two sets of build files to maintain for any longer than necessary.

@hzeller @QuantamHD Thoughts?


maliberty:
I am not ready to retire the cmake infrastructure.  For a start we are not running it across all platforms that cmake is using.

maliberty:
What is the "Cortés book" reference?

oharboe:
> I am not ready to retire the cmake infrastructure. For a start we are not running it across all platforms that cmake is using.

Any other things worth mentioning that needs to be in place?

Python support in OpenROAD?
qt?

oharboe:
> What is the "Cortés book" reference?

Refers to Hernán Cortés, the Spanish conquistador who, upon landing in Mexico, burned his ships (or scuttled them, historically speaking) to eliminate any thought of retreat, forcing his men to fully commit to the conquest ahead.

Issue #7019, opened 2025-04-07
Convert `[INFO PPL-0070] Pin xxxx placed at (677.53um, 0.04um).` to debug info


oharboe:
### Description

Log spam:

```
[INFO PPL-0070] Pin io_foo placed at (677.53um, 0.04um).
```

### Suggested Solution

Make this a debug message and let the user inspect pin positions in the GUI, or move it out of logs and into a report.

oharboe:
> How about we print it only if the computed location differs from the user given one?  That seems worth knowing.

I didnt know that could happen. The position is specified by float, so what defines different? No such thing as two equal floats....

oharboe:
> We snap it to the manufacturing grid.  Internally the float in micron is converted to an integer in DBU.  Also -force_to_die_boundary will move the pin if needed.

So if something is moved will the log say why?

maliberty:
Not presently but it could.

PR #7049, opened 2025-04-08
ppl: less verbose output from place_pin command

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/7019


eder-matheus:
@oharboe Could you comment if this address your concerns about this message?

oharboe (commented):
Looks nice!

Though... I think it is a shame to hide this behind -verbose, nobody will see it....

eder-matheus:
> Looks nice!
> 
> Though... I think it is a shame to hide this behind -verbose, nobody will see it....

I'm fine with removing the -verbose. I just added it to prevent unwanted logs. I understood that it was your main concern.

oharboe:
> > Looks nice!
> > Though... I think it is a shame to hide this behind -verbose, nobody will see it....
> 
> I'm fine with removing the -verbose. I just added it to prevent unwanted logs. I understood that it was your main concern.

I'm of two minds about the matter :-) But the code is pointless if nobody sees it, might as well be debug message then.

src/ppl/src/IOPlacer.cpp
```diff
@@ -2647,12 +2647,21 @@ void IOPlacer::placePin(odb::dbBTerm* bterm,
...
 
-  logger_->info(PPL,
-                70,
-                "Pin {} placed at ({:.2f}um, {:.2f}um).",
-                bterm->getName(),
-                getBlock()->dbuToMicrons(pos.x()),
-                getBlock()->dbuToMicrons(pos.y()));
+  odb::Point original_pos(x, y);
+  if (pos != original_pos) {
+    debugPrint(logger_,
```
maliberty, line 2652: I don't think we need to move this to a debugPrint now that it is inside the if statement.

src/ppl/src/IOPlacer.cpp
```diff
@@ -2647,12 +2647,21 @@ void IOPlacer::placePin(odb::dbBTerm* bterm,
...
   excludeInterval(interval);
 
-  logger_->info(PPL,
-                70,
-                "Pin {} placed at ({:.2f}um, {:.2f}um).",
-                bterm->getName(),
-                getBlock()->dbuToMicrons(pos.x()),
-                getBlock()->dbuToMicrons(pos.y()));
+  odb::Point original_pos(x, y);
+  if (pos != original_pos) {
```
maliberty, line 2651: I think you can just write ` if (pos != {x, y}) {`

maliberty (changes requested):

src/ppl/src/IOPlacer.cpp
oharboe: This is enormously much less verbose :+1: 

I would normally only look at pin positins in the GUI.

I can't offer much in the way of judgement on this residual logging and how useful it is, I wouldn't expect to ever see it.

The only time I can think of to give pins an exact position is when I am saving the pin placement to a .tcl file and reusing it(locking pin placement to speed up builds or possibly combining pin placement results). Let's say I have an array with heterogenous elements and routing by abutment. I would then need to use pin placements from one type of element on another type of element. If pins were moved at all, there would be something wrong with my flow because the pins were originally automatically placed.

maliberty (approved):

PR #7050, opened 2025-04-08
ppl: use microns instead of DBUs when reporting

gadfort:
I noticed that the ppl messages are frequently reporting warnings and errors with dbus, this changes the ones I found to microns


eder-matheus (approved):
LGTM! I just don't get the DEF files that seems to be changed...

maliberty (approved):

PR #7059, opened 2025-04-09
testing: write up a chapter on testing policy

oharboe:
A bit wordy and I think it can be trimmed down significantly once it has been circulated and agreed upon.

The goal is to avoid a thousand papercuts in PRs from unaligned expectations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,10 +356,37 @@ and their descriptions are available [here](docs/contrib/Logger.md#openroad-tool
 OpenROAD uses Git for version control and contributions. 
 Get familiarised with a quickstart tutorial to contribution [here](docs/contrib/GitGuide.md).
 
-
 ## Understanding Warning and Error Messages
 Seeing OpenROAD warnings or errors you do not understand? We have compiled a table of all messages
 and you may potentially find your answer [here](https://openroad.readthedocs.io/en/latest/user/MessagesFinal.html).
+
+## Supported platforms and testing strategy
+
+The priority of testing is to test the algorithms. OpenROAD uses automated tests and CI to facilitate fast and safe progress in development.
+
+The platforms used in CI are whatever abundant and inexpensive testing resources that are available, currently this is x86 Google Cloud hardware.
+
+The value of a test ranges from extremely valuable(small, fast, easy to maintain, deterministic and frequent true positives) to a liability(slow running, hard to maintain, frequent false positives, unstable).
+
+Re-running all tests of algorithms varying initial conditions such as number of CPUs, CPU architecture, platform type, etc. increases the running time of tests exponentially, but adds virtually no testing value.
+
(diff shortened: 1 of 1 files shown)


oharboe:
@hzeller @QuantamHD Thoughts?

hzeller:
I'd probably boil it down to small tests (that typically run < 10s), medium tests (everything typically not more than 60seconds) and integration tests (huge, long timeout, more memory).

As for testing goals: there should be tons of unit tests to make sure to have high development veolocity testing things from the ground up.

Right now there are lots of end2end tests (I'd classify the Tcl tests as such), but not too many classic unittests, written in c++, testing the API and functionality of a paritcular object or set of functions (`rg gtest.h | wc -l`).

So a good aspiration is that for each *.cpp, there should be a unit test that tests all the API and behavior or that API.  In every project I am involved in, there is typically a `foo.h`, `foo.cpp` and `foo_test.cpp`, and these tests are typically small. So it is possible to work on one test, have compile/test turnaround times of about 1 second, and then can be confident that everything building on top of that just works.

Writing this down helps getting new folks involved, as it is simple to point them to the documentation and requesting a `foo_test.cpp` if they haven't added it already.

hzeller:
I don't disagree: there should be long running integration tests.

The C++ tests are in addition as they are easy to write, run fast, and make sure that it is easy to check correctness of the assumptions.

oharboe:
> I don't disagree: there should be long running integration tests.
> 
> The C++ tests are in addition as they are easy to write, run fast, and make sure that it is easy to check correctness of the assumptions.

Banging my drum: my prediction is that if C++ unit-tests improve quality of life for developers, we'll see more of them. If not, I expect to see mor of what we already have... Small fast OpenROAD .tcl tests + a bunch of ORFS integration tests.

QuantamHD:
My general tact is that knowing you have to write a c++ unit test means that you have you have to write your code differently to make it testable in the first place. Which leads to more reusable code.

I do require that c++ unit tests are generally included in each PR that's done for us for that reason. I'd like to see a lot more library code that can be called from around the system rather than one off algorithms that can only be called from TCL.

QuantamHD:
Agreed, but both of us are customers as well as contributors. We both have the power to set the requirements $$$

oharboe:
> Agreed, that it's not going to be universal until it's the easiest path, which is likely never, but I will continue to make it a requirement for Google related stuff

Excellent, I won't expend any goodwill on that policy then :-)

oharboe:
There are some claims about what is valuable to test and necessary to test that I haven't backed up with data.

Some of these claims better be true, or there's a problem that there's no clear fix for, so might as well assume it and cross fingers.

For instance, that Python and .tcl bindings themselves are stable and don't need much more beyond testing on par with GUI smoke testing better be true or I'm out of ideas on what could be done about it; I don't see a path to support two language bindings in that case.

oharboe:
out of sight, out of mind for now. Lots of other things going on. Triggered some good discussions, but we may not necessarily need this officially written up as a policy and if it isn't maintained, then a written outdated and wrong policy does more harm than good.

PR #7065, opened 2025-04-09
bazel: ppl test

oharboe:


src/ppl/test/BUILD
```diff
@@ -0,0 +1,200 @@
...
+        name = test_name + "_resources",
+        data = [":regression_resources"] + glob(
+            [
+                test_name + ".*",
+            ],
+        ) + {
+            "gcd": [],
+            "blocked_region": [],
+            "large_groups1": [],
+            "group_pins3": [],
```
maliberty, line 167: What is the purpose of the empty [] ?

oharboe: don't want the default list, just an empty list.

oharboe: I think this is terse iffy code. Don't know how to make it read better...

maliberty (approved):

PR #7079, opened 2025-04-09
ppl: add tie break for sections with same HPWL cost

eder-matheus:


src/ppl/src/IOPlacer.cpp
```diff
@@ -1486,15 +1492,24 @@ bool IOPlacer::assignPinToSection(IOPin& io_pin,
...
+    std::vector<int> larger_cost(sections.size());
+
     for (int i = 0; i < sections.size(); i++) {
       dst[i] = netlist_->computeIONetHPWL(idx, sections[i].pos)
                + getMirroredPinCost(io_pin, sections[i].pos);
+
+      if (has_mirrored_pin) {
+        larger_cost[i]
+            = std::max(netlist_->computeIONetHPWL(idx, sections[i].pos),
+                       getMirroredPinCost(io_pin, sections[i].pos));
```
maliberty, line 1509: You just computed both of these values above.  Let's avoid doing the work twice.

src/ppl/src/IOPlacer.cpp
```diff
@@ -1338,19 +1338,25 @@ int IOPlacer::assignGroupToSection(const std::vector<int>& io_group,
...
+        bool has_mirrored_pin = pin.getBTerm()->hasMirroredBTerm();
         int pin_hpwl = net->computeIONetHPWL(pin_idx, sections[i].pos);
         if (pin_hpwl == std::numeric_limits<int>::max()) {
           dst[i] = pin_hpwl;
           break;
         }
         dst[i] += pin_hpwl + getMirroredPinCost(pin, sections[i].pos);
+        if (has_mirrored_pin) {
+          larger_cost[i]
+              += std::max(pin_hpwl, getMirroredPinCost(pin, sections[i].pos));
```
maliberty, line 1354: Are we accumulating here (+=)?  Don't compute getMirroredPinCost twice.

maliberty (changes requested):

src/ppl/src/IOPlacer.cpp
eder-matheus: Yes, we accumulate because this is for pin groups. We accumulate the total distance of the pin groups, so I decided to accumulate the largest cost between the mirrored pins to have a similar logic.

maliberty (approved):
We have few designs with mirrored pins so I'll merge this and keep an eye on mock-array, etc.

PR #7081, opened 2025-04-09
meta: Adds whereami path expansion

oharboe:


MODULE.bazel
```diff
@@ -65,3 +65,26 @@ use_repo(llvm, "llvm_toolchain")
...
+    build_file_content = """
+cc_library(
+    name = "whereami",
+    srcs = [
+        "whereami.c", "whereami.h",
+    ],
+    hdrs = [
+        "whereami.h",
+    ],
+    copts= ["-I./external/_main~_repo_rules~whereami/"],
```
QuantamHD, line 82: Don't do this. Rather do includes = [ "." ].

oharboe: Didn't work:

`` `

oharboe:
> This feels like overkill. Do we know you can't just use std::filesystem::read_symlink on argv[0] all platforms?

```
$ git diff src/Main.cc
diff --git a/src/Main.cc b/src/Main.cc
index a8d1bd4cbd..637cfaa16c 100644
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -215,9 +215,10 @@ int main(int argc, char* argv[])
 
 #ifdef BAZEL_CURRENT_REPOSITORY
   using rules_cc::cc::runfiles::Runfiles;
+  auto argvpath = std::filesystem::read_symlink(argv[0]);
   std::string error;
   std::unique_ptr<Runfiles> runfiles(
-      Runfiles::Create(argv[0], BAZEL_CURRENT_REPOSITORY, &error));
+      Runfiles::Create(argvpath, BAZEL_CURRENT_REPOSITORY, &error));
   if (!runfiles) {
     std::cerr << error << std::endl;
     return 1;
oyvind@gale:~/OpenROAD-flow-scripts/tools/OpenROAD$ which openroad
/home/oyvind/OpenROAD-flow-scripts/tools/OpenROAD/bazel-bin/openroad
oyvind@gale:~/OpenROAD-flow-scripts/tools/OpenROAD$ openroad
libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in read_symlink: No such file or directory ["openroad"]
Aborted (core dumped)
```

maliberty:
How about https://live.boost.org/doc/libs/1_87_0/doc/html/doxygen/shared_library_reference/runtime__symbol__info_8hpp_1a5895fc6b0a1b22247a2179f64fbde75e.html

oharboe:
> How about https://live.boost.org/doc/libs/1_87_0/doc/html/doxygen/shared_library_reference/runtime__symbol__info_8hpp_1a5895fc6b0a1b22247a2179f64fbde75e.html

:face_exhaling:

oharboe:
Let us see if it compiles, it had to start from the top https://github.com/The-OpenROAD-Project/OpenROAD/pull/7083

MODULE.bazel
oharboe: I don't particularly want to patch up the library.

oharboe:
@maliberty Useful illustration of use of http_archive(), sure to come in handy at some point.

oharboe:
This works and is reasonably clean w.r.t. number of files touched, but not a favourite. Closing, out of sight, out of mind for now.

https://github.com/The-OpenROAD-Project/OpenROAD/pull/7099 is the front runner in the race to implement this fix.

PR #7082, opened 2025-04-09
drt: incremental PA

### LINKED ISSUE #5867: DRT: incremental Pin Access with odb callbacks

bnmfw:
Fixes #5867.

ISPD and Secure CI safe.

Pin Access is now Alive! Once Pin Access has a initial batch run done all its structures are kept alive by TritonRoute. DesignCallBack has its callbacks modified to allow for the update of PA data of any instance called.
During premove all PA data is erased, during postmove all PA data is calculated, if the moved instance is a new unique the unique data will also be calculated. In any case its row of instanced will have their access patterns solved again.
There is still a missing callback for creating a new inst.

#### `DesignCallBack`
All callbacks now deal with PA data.
Also transformed a nest level into a guard just because.

#### PA is alive
All instances of PA on TritonRoute now are member variables. PA is fully alive within DRT.

#### DRT/GRT Decouple
PA has already updates the instances data as instances move, therefore the coupling GRT had to DRT is not necessary, as GRT does not need to manually ask DRT to update PA data.

#### `areAllSkipInstTerms`
Some minor logic that was repeated and I though was better expressed as a function.


bnmfw:
I actually really not like how unique insts are handled. Ideally you would have just a singleton map [master][orient][offset] that could map any inst to any unique insts, which would be a new object. The unique insts would hold all the access point and all access pattern data very simply. Moving an inst would simply mean to move it to another unique inst entry, if that unique inst did not have their acc points and patterns computed you'd compute them than. There would literally be 0 cost to moving an inst if the unique inst it fell under already had its data processed, no managing 4 different maps.

eder-matheus (approved):
Looks good to me regarding GRT and the db callbacks. I'll let Osama give his final comments/requests on the specific DRT changes.

src/drt/src/pa/FlexPA_row_pattern.cpp
```diff
@@ -145,7 +145,6 @@ void FlexPA::prepPatternInstRows(std::vector<std::vector<frInst*>> inst_rows)
             result;
         std::unique_ptr<PinAccessJobDescription> uDesc
             = std::make_unique<PinAccessJobDescription>();
-        uDesc->setPath(path);
```
osamahammad21: why is this removed?

src/drt/src/pa/FlexPA.cpp
```diff
@@ -285,6 +306,16 @@ bool FlexPA::isSkipInstTerm(frInstTerm* in)
   return skip_unique_inst_term_.at({inst_class, in->getTerm()});
 }
 
+bool FlexPA::areAllSkipInstTerms(frInst* inst)
```
osamahammad21:
```suggestion
bool FlexPA::isSkipInst(frInst* inst)
```

src/drt/src/pa/FlexPA_row_pattern.cpp
bnmfw: No idea. Must have been some missclick.

src/drt/src/pa/FlexPA.cpp
bnmfw: Good one

src/drt/src/pa/FlexPA_row_pattern.cpp
```diff
@@ -145,9 +145,9 @@ void FlexPA::prepPatternInstRows(std::vector<std::vector<frInst*>> inst_rows)
...
         uDesc->setType(PinAccessJobDescription::INST_ROWS);
         msg.setJobDescription(std::move(uDesc));
+        uDesc->setPath(path);
```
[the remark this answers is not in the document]
osamahammad21: watch out, you're writing to a null unique_ptr. move this line up before setType as it was.

osamahammad21 (changes requested):

PR #7012, opened 2025-04-06
bazel: make it possible to add openroad to PATH

oharboe:
fixes #7006

PR #7083, opened 2025-04-09
bazel: openroad in path

oharboe:
Seems to work


oharboe:
@hzeller Any tips? Had to relax `common --incompatible_disallow_empty_glob` from 36f0c156ce7
```
$ bazel build :openroad -c opt
ERROR: Traceback (most recent call last):
	File "/home/oyvind/.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/external/boost.filesystem~/BUILD.bazel", line 11, column 16, in <toplevel>
		hdrs = glob([
Error in glob: glob pattern 'include/**/*.h' didn't match anything, but allow_empty is set to False (the default value of allow_empty can be set with --incompatible_disallow_empty_glob).
```

QuantamHD:
To answer your other question @oharboe. "Why doesn't google provide the full solution", because in general we don't benefit from it.

Google has it's own way of doing this, and it's not provided by bazel internally. The bazel team is small, and is essentially trying to replicate the infrastructure of Google and the solutions they create aren't used by Google so it's a low priority.

QuantamHD (approved):

.bazelrc
```diff
@@ -7,7 +7,7 @@ common --check_direct_dependencies=off
 
 # bazel 7 is somewhat forgiving for glob patterns that don't
 # match, but bazel 8 will be strict. So start now.
-common --incompatible_disallow_empty_glob
+#common --incompatible_disallow_empty_glob
```
hzeller, line 10: ... don't do this.

MODULE.bazel
```diff
@@ -40,6 +40,7 @@ bazel_dep(name = "boost.property_tree", version = BOOST_VERSION + ".bcr.2")
 bazel_dep(name = "boost.spirit", version = BOOST_VERSION)
 bazel_dep(name = "boost.stacktrace", version = BOOST_VERSION)
 bazel_dep(name = "boost.thread", version = BOOST_VERSION)
+bazel_dep(name = "boost.filesystem", version = BOOST_VERSION)
```
hzeller, line 43: .. add ".bcr.1" to BOOST_VERSION instead. I think that version has the glob fixed.

```
bazel_dep(name = "boost.filesystem", version = BOOST_VERSION + ".bcr.1")
```

hzeller (commented):
I am not necessarily happy having yet another boost dependency.

The PR description does not really say what you attempt to achieve, can you elaborate ?

.bazelrc
oharboe: didn't want to, but it was getting in the way of my immediate objectives. I figured you might have a solution to this...

MODULE.bazel
oharboe: will try...

oharboe:
@maliberty I won't say it was worth it :sweat_smile: but this solution is nice and good to go.

Now there is a better answer to my question to @QuantamHD on why Bazel doesn't do this: it is a solved problem by boost.

oharboe:
@hzeller What is going on here?

Works locally.

```
2025-04-10T08:50:44.8569585Z (08:50:44) [8,921 / 9,491] Compiling src/drt/src/TritonRoute.cpp; 23s processwrapper-sandbox ... (5 actions, 4 running)
2025-04-10T08:50:47.3902363Z (08:50:47) ERROR: /home/runner/work/OpenROAD/OpenROAD/BUILD.bazel:52:10: Compiling src/Main.cc failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target //:openroad) external/toolchains_llvm~~llvm~llvm_toolchain/bin/cc_wrapper.sh -U_FORTIFY_SOURCE '--target=x86_64-unknown-linux-gnu' -U_FORTIFY_SOURCE -fstack-protector -fno-omit-frame-pointer -fcolor-diagnostics ... (remaining 1384 arguments skipped)
2025-04-10T08:50:47.3906236Z 
2025-04-10T08:50:47.3907730Z Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
2025-04-10T08:50:47.3919994Z src/Main.cc:32:10: fatal error: 'boost/dll/runtime_symbol_info.hpp' file not found
2025-04-10T08:50:47.3925661Z    32 | #include <boost/dll/runtime_symbol_info.hpp>
2025-04-10T08:50:47.3930298Z       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2025-04-10T08:50:47.3933307Z 1 error generated.
```

oharboe:
@hzeller Does it need boost.dll and boost.system? Unsure about dependencies here and why it works locally and not on the server.

Tried, but didn't work:

```
bazel_dep(name = "boost.system", version = BOOST_VERSION + ".bcr.1")
bazel_dep(name = "boost.dll", version = BOOST_VERSION + ".bcr.1")
```

oharboe:
@hzeller @QuantamHD If we can't make this work, then we can use https://github.com/The-OpenROAD-Project/OpenROAD/pull/7081 as a stopgap and revivie this when boost works.

oharboe:
> I am not necessarily happy having yet another boost dependency.
> 
> The PR description does not really say what you attempt to achieve, can you elaborate ?

https://github.com/The-OpenROAD-Project/OpenROAD/pull/7012

oharboe:
@hzeller boost.dll only exist in bcr in boost 1.87. Desperately trying to add boost.system 1.83.0 to see if that can fix the compile problem.

hzeller:
Maybe, we should just use a local function which saves us adding two additional boost dependencies and headaches.

```c++
// Returns path to running binary if possible, otherwise nullopt.
static std::optional<std::string> getProgramLocation() {
#if defined(_WIN32)
  char result[MAX_PATH + 1] = {'\0'};
  auto path_len = GetModuleFileNameA(NULL, result, MAX_PATH);
#elif defined(__APPLE__)
  char result[MAXPATHLEN + 1] = {'\0'};
  uint32_t path_len = MAXPATHLEN;
  if (_NSGetExecutablePath(result, &path_len) != 0) {
    path_len = readlink("/proc/self/exe", result, MAXPATHLEN);
  }
#else
  char result[PATH_MAX + 1] = {'\0'};
  ssize_t path_len = readlink("/proc/self/exe", result, PATH_MAX);
#endif
  if (path_len > 0) return result;
  return std::nullopt;
}
```

(lifted from [Surelog](https://github.com/chipsalliance/Surelog/blob/master/src/Common/FileSystem.cpp#L62-L77)  where we did something like that before).

hzeller:
back in the day ... `PATH_MAX` was enough for everybody :)

Anyway, I think these days `PATH_MAX` probably just is a hint per each element in the filename you're right. OTOH, `PATH_MAX` is typically 4096 or more while I expect the full path, even with bazel hashes in the middle be not much more than 150 characters. So probably not something to worry right now.

oharboe:
https://github.com/The-OpenROAD-Project/OpenROAD/pull/7099

oharboe:
This approach is not a frontrunner in the race to implement this functionality, out of sight, out of mind for now:

```
2025-04-11T15:02:08.0065603Z bazel-out/k8-opt/bin/src/stt/src/flt/etc/POWV9.cpp:3:21: warning: string literal of length 3839112 exceeds maximum length 65536 that C++ compilers are required to support [-Woverlength-strings]
2025-04-11T15:02:08.0067078Z 1 warning generated.
2025-04-11T15:02:08.0069460Z (15:02:07) [8,692 / 9,493] Compiling openroad_swig.cc; 10s processwrapper-sandbox ... (4 actions, 3 running)
2025-04-11T15:02:09.0604704Z (15:02:09) ERROR: /home/runner/work/OpenROAD/OpenROAD/BUILD.bazel:52:10: Compiling src/Main.cc failed: (Exit 1): cc_wrapper.sh failed: error executing CppCompile command (from target //:openroad) external/toolchains_llvm~~llvm~llvm_toolchain/bin/cc_wrapper.sh -U_FORTIFY_SOURCE '--target=x86_64-unknown-linux-gnu' -U_FORTIFY_SOURCE -fstack-protector -fno-omit-frame-pointer -fcolor-diagnostics ... (remaining 1384 arguments skipped)
2025-04-11T15:02:09.0609118Z 
2025-04-11T15:02:09.0609857Z Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
2025-04-11T15:02:09.0618369Z src/Main.cc:32:10: fatal error: 'boost/dll/runtime_symbol_info.hpp' file not found
2025-04-11T15:02:09.0619372Z    32 | #include <boost/dll/runtime_symbol_info.hpp>
2025-04-11T15:02:09.0621095Z       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2025-04-11T15:02:09.0621719Z 1 error generated.
```

PR #7085, opened 2025-04-09
bazel: configure timeout times in .bazelrc

hzeller:
... so that it is not necessary to remember the flag everyt time one runs the flow tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,6 +5,9 @@ common --enable_workspace  # explicitly currently still needing this
 # versions found in dependency graph.
 common --check_direct_dependencies=off
 
+# Test timeouts for various levels.
+test --test_timeout=300,1800,1800,9600
+
 # bazel 7 is somewhat forgiving for glob patterns that don't
 # match, but bazel 8 will be strict. So start now.
 common --incompatible_disallow_empty_glob
--- a/.github/workflows/github-actions-bazel-build.yml
+++ b/.github/workflows/github-actions-bazel-build.yml
@@ -32,8 +32,6 @@ jobs:
 
       - name: Build
         run: |
-          # Timeout is not per test, but for the entire CI job. Running
-          # times are tracked in ORFS tests over time.
           bazel test \
            -c opt \
            --keep_going \
@@ -41,7 +39,6 @@ jobs:
            --test_output=errors \
            --curses=no \
            --force_pic \
-           --test_timeout=3600,3600,3600,3600 \
            ...
 
       - name: Smoke test


hzeller:
CC: @oharboe

.bazelrc
```diff
@@ -5,6 +5,9 @@ common --enable_workspace  # explicitly currently still needing this
 # versions found in dependency graph.
 common --check_direct_dependencies=off
 
+# Test timeouts for various levels.
+test --test_timeout=300,1800,1800,9600
```
maliberty, line 9: This does not match the removed `--test_timeout=3600,3600,3600,3600 `

hzeller: correct. Let's say it is refined.

maliberty: I thought the goal was the avoid timeouts without having to classify the test cases now.  This seems to open the door the previous problems.

hzeller: we have essentially a bimodal set of tests. The ones that are small, and the other ones. The small ones typically run in < 1min (and mostly < 1 sec actually), so I gave them 5min.

The big flow tests are large and actually can take longer than 1h, so I gave them a good chunk.

The other test kinds are not used, so I gave them something in the middle.

maliberty (approved):

oharboe:
@QuantamHD Ref my [writeup on thoughts on testing policy and strategy](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7059), .tcl tests are C++ unit-tests in lieu. They should take seconds, not minutes, hours and days. It is the job of ORFS and bazel-orfs(megaboom is on the order of days an occasional post-merge testing) to do integration testing that takes minutes, hours and days. I don't know what itch the hour long .tcl tests in OpenROAD scratches that ORFS can't or shouldn't, or if those test simply predate ORFS in some way.

To have good coverage in C++ unit-tests where no test takes more than seconds in OpenROAD is a pretty lofty goal and may not be expedient, so a few unit-tests (.tcl or C++) that takes minutes in OpenROAD is a pretty good state of affairs.

If that is the direction we're moving in, I don't see an issue with this pull request.

If we get lots of tests in the minutes ++ range, we'll have to revisit this.

hzeller:
The long ones are the `//test:flow_tests` - one of them (jpeg something something) takes about 1.5h on my (probably too small) personal machine - this is why I choose a somewhat large value for the huge tests.

The small tcl tests are observed typically less than a second and already behave speed-wise like c++ unit tests;  there are a few that are currently in the 70-90ish seconds; so that is what the 300s are addressing.

Anyway, I moved the numbers from the CI to the `.bazelrc`   to make sure to not have to remember setting the timeout value when running manual tests on the development machine (and it will automatically cover the CI as well). But I also didn't want to reserve 1h for small tests, so reduced that to an easily fitting observed value, and expanded the long tests from 1h as I've observed longer times there.

Over time, the tests and categories will probably evolve.

oharboe:
@hzeller Yep... //test:flow_tests are long running and marked as manual. They don't offer any testing value for bazel build as such. I think @maliberty said they are enabled in CI, but I think their value is to test the algorithms of OpenROAD in some way.

hzeller:
yep, but if I run them manually, and they time out after 1h I get reminded that I should've set the timeout ... that is why it is good to have that taken care of the `.bazelrc` :)

oharboe:
> yep, but if I run them manually, and they time out after 1h I get reminded that I should've set the timeout ... that is why it is good to have that taken care of the `.bazelrc` :)

I wonder if these tests will simply be retired in favor of ORFS... First I need to learn more about what they are testing to have an opinion.

hzeller:
It looks like they are making some medium designs, like Ibex risc-V design to kindof to end-to-end test .So having a few of these in OpenRoad like in `//test:flow_tests` is probably good, but other, bigger etc tests should be maintained outside as you suggest.

oharboe:
> It looks like they are making some medium designs, like Ibex risc-V design to kindof to end-to-end test .So having a few of these in OpenRoad like in `//test:flow_tests` is probably good, but other, bigger etc tests should be maintained outside as you suggest.

I'm pushing for doing bazel testing of OpenROAD from within ORFS. If these flow tests test nothing beyond ORFS(they may test scratch some itch ORFS can't, I don't know), then I would argue that they are a liability rather than a benefit. https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3055

PR #7087, opened 2025-04-10
Don't use `using std::something` in header files.

hzeller:
Shortcuts like `using std::vector` pollute the namespace for everyone including the header. Even if they are used within a namespace, now a `foo` namespace suddenly has a `foo::vector`.

This first fix works on fixing the `std::` using statements, but there are a bunch more that need to be addressed.

### DIFF
--- a/src/dpl/src/CheckPlacement.cpp
+++ b/src/dpl/src/CheckPlacement.cpp
@@ -21,17 +21,18 @@ using utl::DPL;
 
 using utl::format_as;
 
-void Opendp::checkPlacement(const bool verbose, const string& report_file_name)
+void Opendp::checkPlacement(const bool verbose,
+                            const std::string& report_file_name)
 {
   importDb();
 
-  vector<Node*> placed_failures;
-  vector<Node*> in_rows_failures;
-  vector<Node*> overlap_failures;
-  vector<Node*> one_site_gap_failures;
-  vector<Node*> site_align_failures;
-  vector<Node*> region_placement_failures;
-  vector<Node*> edge_spacing_failures;
+  std::vector<Node*> placed_failures;
+  std::vector<Node*> in_rows_failures;
+  std::vector<Node*> overlap_failures;
+  std::vector<Node*> one_site_gap_failures;
+  std::vector<Node*> site_align_failures;
+  std::vector<Node*> region_placement_failures;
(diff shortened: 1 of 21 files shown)


src/dpl/src/CheckPlacement.cpp
```diff
@@ -20,17 +20,18 @@ using utl::DPL;
...
-  vector<Cell*> site_align_failures;
-  vector<Cell*> region_placement_failures;
-  vector<Cell*> edge_spacing_failures;
+  std::vector<Cell*> placed_failures;
+  std::vector<Cell*> in_rows_failures;
+  std::vector<Cell*> overlap_failures;
+  std::vector<Cell*> one_site_gap_failures;
+  std::vector<Cell*> site_align_failures;
+  std::vector<Cell*> region_placement_failures;
+  std::vector<Cell*> edge_spacing_failures;
```
maliberty, line 34: From https://google.github.io/styleguide/cppguide.html#Aliases

> However, local convenience aliases are fine in function definitions, private sections of classes, explicitly marked internal namespaces, and in .cc files:
> 
> // In a .cc file
> using ::foo::Bar;
> 

So there is no need to make this transformation in the cpp files.  Just move the using statement to the relevant cpp(s).

maliberty:
conflicts to resolve

src/dpl/src/CheckPlacement.cpp
hzeller: sure, one _could_ do that inside a function or compilation unit, but it is very uncommon for `std::` ... in particular since we also have a sta::Vector with an uppercase letter, so that would be doubly confusing.
For `std::` we should spell it out always.

maliberty: Fine but I wouldn't say that is true for other namespaces in future PRs.

Vector is a bad idea - inheritance from a class with a non-virtual dtor just asks for trouble.

hzeller: Other namespaces, yes that is more common to have `using` inside compilation units or functions.

Yeah, `sta::Vector` (and `sta::Map`) is ... yeah indeed, I fully agree.

hzeller:
Fixed conflicts.

maliberty (approved):

PR #7093, opened 2025-04-10
grt: fix test results name to avoid conflicts

eder-matheus:


eder-matheus:
@osamahammad21 FYI, this should fix the random failures in congestion1.py you detected in some PRs.

maliberty (approved):

oharboe:
> @oharboe this should fix one random CI failure - two processes may write the same file by mistake.

😌


Classic. Bazel has one isolated environment per process it launches, which can help move code from single to multiprocess.

PR #7094, opened 2025-04-10
mpl: fix crash when LEF PIN specifies USE CLOCK

mikesinouye:


mikesinouye:
Hi @maliberty, could you take a look when you get a chance?

maliberty:
Why does this crash?  I don't understand what you are fixing.

maliberty:
> Presumably the initial assumption was that non-SIGNAL pins are not needed for consideration of the graph for determining good macro placement, but this is not the case.

ANALOG signal type is also unhandled and subject to the problem.  I think the code should handle the case where a vertex is not found gracefully.

Including clocks is ok but not likely to be very helpful.  We filter out high fanout nets and only very small clock domains will have any effect.  Have you observed it to make a difference?

mikesinouye:
> Including clocks is ok but not likely to be very helpful. We filter out high fanout nets and only very small clock domains will have any effect. Have you observed it to make a difference?

In this IP, the limited number of macros have their own clock which would explain this behavior. The goal here was to prevent the crash, so including clocks is helpful in that regard. I have not run trials to determine any QoR impact by handing this case + including/excluding the low clock domain connectivity.

ANALOG can be added. RESET likely makes sense as well.

maliberty (approved):
@AcKoucher FYI

PR #7097, opened 2025-04-11
Provide id-sorting map/set typedefs for frBlockObjects.

hzeller:
Instead of manually having to define these containers with the comparator function, pre-define such containers with the comparator already configured.

Give them a name that makes it easy to see at a glance what they doe.

This improves readability and allows to play with various container implementations later.

### DIFF
--- a/src/drt/src/db/grObj/grNet.h
+++ b/src/drt/src/db/grObj/grNet.h
@@ -49,13 +49,11 @@ class grNet : public grBlockObject
   {
     return pinGCellNodePairs;
   }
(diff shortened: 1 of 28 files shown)


hzeller:
The `frOrderedIdMap` and `frOrderedIdSet` are just placeholders; I am open for better names for these types.

hzeller:
Note, there is no functional change, this is just adding the typedef and using everywhere.

maliberty (approved):
@osamahammad21 @bnmfw FYI

Issue #7006, opened 2025-04-05
Make it possible to add bazel built OpenROAD to path


oharboe:
### Description

Use cases include make issue/standalone testscases.

There's a problem with adding the folder of the openroad built by Bazel to the path, which is not an uncommon use-case.

```
$ bazel build :openroad -c opt
$ export PATH=$(pwd)/bazel-out/k8-opt/bin:$PATH
$ openroad
ERROR: external/rules_cc~/cc/runfiles/runfiles.cc(123): cannot find runfiles (argv0="openroad")
```

If you launch it passing in the entire path, $(command -v openroad), it works.

```
$ command -v openroad
.../tools/OpenROAD/bazel-out/k8-opt/bin/openroad
$ .../tools/OpenROAD/bazel-out/k8-opt/bin/openroad
OpenROAD HDL-HEAD bazel-build
Features included (+) or not (-): -GPU -GUI -Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
$
```

hzeller:
This hints at another problem: we probably can't simply copy the binary somewhere ('make install' like) as it will now have trouble to deal with the runfiles. We have to figure out exactly what these runfiles comprise and make it possible to have an install script that copies the binary into another place and it work there.

@QuantamHD do you know what currently is in there ? Is it just the tclsh system files ?

hzeller:
```
find bazel-bin/openroad.runfiles
```
yep, looks like mostly tcl stuff

oharboe:
> ```
> find bazel-bin/openroad.runfiles
> ```
> yep, looks like mostly tcl stuff

The big blockers to using bazel built openroad outside of bazel is to add python and qt support, AFAIK.

hzeller:
I think Python might be straightforward, and I think @QuantamHD has some ideas for Qt.

PR #7099, opened 2025-04-11
bazel: make OpenROAD in path work

oharboe:
Without adding any dependencies

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -201,6 +201,32 @@ static void handler(int sig)
   raise(sig);
 }
 
+#ifdef BAZEL_CURRENT_REPOSITORY
+
+// Avoid adding any dependencies like boost.filesystem
+//
+// Returns path to running binary if possible, otherwise nullopt.
+static std::optional<std::string> getProgramLocation()
+{
+#if defined(_WIN32)
+  char result[MAX_PATH + 1] = {'\0'};
+  auto path_len = GetModuleFileNameA(NULL, result, MAX_PATH);
+#elif defined(__APPLE__)
+  char result[MAXPATHLEN + 1] = {'\0'};
+  uint32_t path_len = MAXPATHLEN;
+  if (_NSGetExecutablePath(result, &path_len) != 0) {
+    path_len = readlink("/proc/self/exe", result, MAXPATHLEN);
+  }
+#else
+  char result[PATH_MAX + 1] = {'\0'};
+  ssize_t path_len = readlink("/proc/self/exe", result, PATH_MAX);
+#endif
+  if (path_len > 0)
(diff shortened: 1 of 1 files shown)


src/Main.cc
```diff
@@ -216,8 +242,8 @@ int main(int argc, char* argv[])
 #ifdef BAZEL_CURRENT_REPOSITORY
   using rules_cc::cc::runfiles::Runfiles;
   std::string error;
-  std::unique_ptr<Runfiles> runfiles(
-      Runfiles::Create(argv[0], BAZEL_CURRENT_REPOSITORY, &error));
+  std::unique_ptr<Runfiles> runfiles(Runfiles::Create(
+      getProgramLocation().string(), BAZEL_CURRENT_REPOSITORY, &error));
```
hzeller, line 246: does this comile ? getProgramLocation() returns an optional.

I'd expect something along the lines of
```c++
 auto maybe_location = getProgramLocation();
 Runfiles::Create(maybe_location.has_value() ? *maybe_location : std::string(), ...);
```

hzeller: alternatively, if optinal is not really used, just modify the `getProgramLocation()` to return a string and make it an empty string in case it can't be determined.

oharboe: Fixed compile error.

Rather than add iffy code "log and continue", just throw an exception if the location can't be gotten: early failure.

hzeller: sounds good.

oharboe:
> LGTM
> I still don't know what issue this solves, as I guess you worked with @QuantamHD on the original problem, so maybe he also wants to have a look.

https://github.com/The-OpenROAD-Project/OpenROAD/issues/7006

hzeller:
ah, the second commit contained the reference to the fixed issue, thanks!

oharboe:
> ah, the second commit contained the reference to the fixed issue, thanks!

missed it first. this repo prefers PRs not to be rewritten to keep complete review history.

maliberty (approved):

maliberty:
Good enough for now.  I don't think we have truly solved the install issue as I expect this will only work if you compiled the program locally.

oharboe:
> Good enough for now.  I don't think we have truly solved the install issue as I expect this will only work if you compiled the program locally.

Thanks! My main concern with bazel is CI and developer use cases,  I have not been thinking about deployment to users.

My understanding is that on Linux one has to recompile for all versions of all distributions(ubuntu, red hat, nix, guix, etc.), which sucks compared to windows. Docker papers over this. Static linking can fix this for some apps at the cost of giant executables, which might not be a problem, and could even be faster that many small executables.

PR #7101, opened 2025-04-11
drt: better requirement structure

bnmfw:
Supports #7150.

Separates the access point requirements on their own functions (`EnoughSparsePoints()` and `EnoughPointsFarFromEdge()`). Adds the `pa_requirements_met` structure to keep track of which requirements have already been met, and changes warning 88 to print the exact requirement not met. I want to formalize  the far from edge requirement so it is configurable. I feel it is not necessary in every case and should not be applied everywhere. The introduction of new soft requirements on PA makes it search for more access points, which costs time and might not always be ideal.


maliberty:
@bnmfw would you point to the pr/change with that your change that you suspect is the issue.  I'd like @gadfort to confirm that is the problem before starting on solutions.

gadfort:
@maliberty I'll check this weekend to see if I can get some concrete numbers to confirm if things slowed down. Like I mentioned it feels like it has, but I didn't have anything specific.

bnmfw:
Here are the recent PRs on PA that could have negatively impacted performance:
#6796 (Merged on March 25): Changes the structure that holds the acc_patterns from a `vector` to a `map`.
#6889 (Merged on April 2): Introduces the PA far from edge soft requirement (The one I'm talking about in this PR).
#6916 (Merged on April 10): Changes the way the instance rows are stored from a vector (what is then ordered) to a set.

gadfort:
@maliberty @bnmfw some quick checks:
now: b45988bcd0d413cc435fe717521f0fca74ac9128 (4586s in prepPatternInstRows, 4584s getEdgeCost)
check commit: a672e5caa944a826471e6628258b860525dd9397 (1063s in prepPatternInstRows, 1027s getEdgeCost)

So that is a 3500s slowdown mostly in `getEdgeCost`.
So I think I'm sure there is a slowdown now :-).

These results are with black_parrot on freepdk45 so ORFS should be able to be used to check.

gadfort:
@bnmfw doing a little checking, this PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6889 is the cause of the slowdown

bnmfw:
Yeah, I thought that might be the culprit. We will adress this in this PR, probabilly make the requirement optional

maliberty:
@bnmfw is there a way to make it smarter rather than optional?  I'd like a bit more understanding of the problem.

bnmfw:
@maliberty Sure! `genPinAccess()` will call `genPinAccessCostBounded()` iteratively until the access points generated meet all soft requirements (`EnoughAccessPoints()` is responsible for verifying it requirements were met). Each iteration will seek access points of different costs. 

If all cost combinations were tried and still the access points did not meet the requirements, warning 88 will be raised (this is relatively new) saying that PA exhaustively tried to generate APs and did not meet the requirements. It will go on with those access points, this is not an error.

Generating access points is expensive, mostly because of the calls to the DRC engine to verify if a new AP candidate is valid. Introducing new soft requirements, like in #6889, will always increase runtime. In the best case the previous solution met the requirement, in the worst new APs have to be created. This was create to try and solve issue #6097, as it was one of the proposed solutions on the thread, require at least 1 access point to be far from the inst edge.

What seems to be happening is that most designs did not meet this requirement previously, so a lot of new access points in most designs are being created to accommodate for this requirement, which seems to not be needed in most cases.

About applying it smartly, the design on issue #6097 is a bit non standard as pointed out on the thread:
```
In the picture it looks as if your stdcell sites are not a multiple of the routing pitch.
This is very unusual and leads to none of the stdcell pins being "on track".
```
Maybe we could check for that and only apply the new soft requirement in that case.

bnmfw:
@maliberty Ready

maliberty:
secure CI?

bnmfw:
> secure CI?

Not necessary, this just changes error messages and creates the requirement structure, but does not actually change how APs are created

src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -1367,19 +1393,34 @@ int FlexPA::genPinAccess(T* pin, frInstTerm* inst_term)
...
+          += "\n\tAt least "
+             + (isStdCell(inst_term->getInst())
+                    ? std::to_string(router_cfg_->MINNUMACCESSPOINT_STDCELLPIN)
+                    : std::to_string(
+                          router_cfg_->MINNUMACCESSPOINT_MACROCELLPIN))
+             + " sparse access points";
+    }
+    if (!reqs.sparse_points) {
+      unmet_requirements += "\n\tAt least " + std::to_string(1)
+                            + " access point farm from inst edge";
```
maliberty, line 1415: farm -> far

maliberty (approved):

src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -1166,27 +1158,58 @@ bool FlexPA::EnoughAccessPoints(
...
+bool FlexPA::EnoughPointsFarFromEdge(
+    std::vector<std::unique_ptr<frAccessPoint>>& aps,
+    frInstTerm* inst_term)
+{
+  const int far_from_edge_requirement = 1;
   Rect cell_box = inst_term->getInst()->getBBox();
+  int total_far_from_edge = 0;
   for (auto& ap : aps) {
     const int colision_dist
         = design_->getTech()->getLayer(ap->getLayerNum())->getWidth() * 2;
```
osamahammad21, line 1180: @bnmfw Could you clarify the rationale behind multiplying by 2 here? The current collision_dist calculation seems somewhat conservative, which might result in unnecessarily restrictive pin access

osamahammad21 (approved):

bnmfw:
Took me a while to understand why tinyRocket was having a slight metric failure. The problem was that once we had enough access points the guard stopped checking on furthers checks on the same pin. Although this seems intuitive, this is not how the current master works. Because of the rough approximation of sparse point calculation is actually possible that adding new access points will make the check fail, this is why the guard was removed.

PR #7105, opened 2025-04-11
drt: Use Bottom/Top routing layer in FlexDRWorker::initTrackCoords

smunaut:
When trying to find a matching layer in the other routing direction, the search is limited to layers selected for routing rather than all tech layers.

In addition, the code now also checks the layer is in the actual routing direction that's expected.

Finally because that code logic was duplicated a few times, it gets isolated in a helper function to make it easier to follow and track what's going on.

### DIFF
--- a/src/drt/src/dr/FlexDR_init.cpp
+++ b/src/drt/src/dr/FlexDR_init.cpp
@@ -1562,6 +1562,43 @@ void FlexDRWorker::initNets(const frDesign* design)
   }
 }
 
+frLayerNum FlexDRWorker::initTrackCoords_getNonPref(frLayerNum lNum)
+{
+  const auto lDir = getTech()->getLayer(lNum)->getDir();
+  auto lDir2 = dbTechLayerDir::NONE;
+
+  switch (lDir) {
+    case dbTechLayerDir::HORIZONTAL:
+      lDir2 = dbTechLayerDir::VERTICAL;
+      break;
+    case dbTechLayerDir::VERTICAL:
+      lDir2 = dbTechLayerDir::HORIZONTAL;
+      break;
+    case dbTechLayerDir::NONE:
+      logger_->error(DRT, 49, "initTrackCoords invalid routing direction");
+      return -1;
+  }
+
+  frLayerNum lTop
+      = std::min(router_cfg_->TOP_ROUTING_LAYER, getTech()->getTopLayerNum());
+  frLayerNum lBot = std::max(router_cfg_->BOTTOM_ROUTING_LAYER,
+                             getTech()->getBottomLayerNum());
+
+  if ((lNum + 2 <= lTop)
+      && (getTech()->getLayer(lNum + 2)->getDir() == lDir2)) {
+    return lNum + 2;
+  }
+
+  if ((lNum - 2 >= lBot)
+      && (getTech()->getLayer(lNum - 2)->getDir() == lDir2)) {
+    return lNum - 2;
+  }
+
+  logger_->error(DRT, 64, "initTrackCoords cannot add non-pref track");
+
+  return -1;
+}
+
 void FlexDRWorker::initTrackCoords_route(drNet* net,
                                          frLayerCoordTrackPatternMap& xMap,
                                          frLayerCoordTrackPatternMap& yMap)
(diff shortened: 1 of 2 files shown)


maliberty:
Ready for review?  Still marked as draft.

smunaut:
@maliberty Yeah, I had trouble running unit tests locally so I wanted to see the result of the action here before taking out of draft.

src/drt/src/dr/FlexDR_init.cpp
```diff
@@ -1579,68 +1608,43 @@ void FlexDRWorker::initTrackCoords_route(drNet* net,
       auto obj = static_cast<drPathSeg*>(uConnFig);
       const auto [bp, ep] = obj->getPoints();
       const auto lNum = obj->getLayerNum();
+      const auto lNum2 = initTrackCoords_getNonPref(lNum);
+      if (lNum2 < 0) {
+        std::cout << "Error: initTrackCoords cannot add non-pref track"
+                  << std::endl;
```
maliberty, line 1614: I see you just moved this but following the boy scout rule it would be good to change this to use the logger and error() rather than std::cout.  You could move the check to initTrackCoords_getNonPref as there is nothing the caller is adding to the check.

src/drt/src/dr/FlexDR_init.cpp
```diff
@@ -1579,68 +1608,43 @@ void FlexDRWorker::initTrackCoords_route(drNet* net,
...
-          } else if (lNum - 2 >= getTech()->getBottomLayerNum()) {
-            xMap[lNum - 2][bp.x()] = nullptr;
-          } else {
-            std::cout << "Error: initTrackCoords cannot add non-pref track"
-                      << std::endl;
-          }
-          // add bp, ep
           yMap[lNum][bp.y()] = nullptr;
           yMap[lNum][ep.y()] = nullptr;
+          if (lNum2 >= 0) {
```
maliberty, line 1622: If this were an error above then then this check is unnecessary.  Similarly elsewhere

maliberty (dismissed):

src/drt/src/dr/FlexDR_init.cpp
```diff
@@ -1562,6 +1562,35 @@ void FlexDRWorker::initNets(const frDesign* design)
...
+  auto lDir2 = dbTechLayerDir::NONE;
+
+  switch (lDir) {
+    case dbTechLayerDir::HORIZONTAL:
+      lDir2 = dbTechLayerDir::VERTICAL;
+      break;
+    case dbTechLayerDir::VERTICAL:
+      lDir2 = dbTechLayerDir::HORIZONTAL;
+      break;
+    default:
```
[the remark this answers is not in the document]
smunaut: :ok_hand:

src/drt/src/dr/FlexDR_init.cpp
smunaut: Using `error` would make it fatal and stop the process right ?
ATM (and in the previous code), although the message says "Error", it would still continue.
( The previous code was wrong because it would use 0 as lNum2 which is invalid for insertion in the map, here I'm not inserting anything at all for the non-pref direction if it wasn't found ).

But if stopping is the desired outcome, I can use `error` and then remove the extraneous checks.

osamahammad21: I say we make it a logger error and check if any of the current test cases break. The only case where this would trigger an error is if we are limited by routing on a single layer which is pretty unusual.

src/drt/src/dr/FlexDR_init.cpp
```diff
@@ -1562,6 +1562,35 @@ void FlexDRWorker::initNets(const frDesign* design)
...
+      return -1;
+  }
+
+  if ((lNum + 2 <= router_cfg_->TOP_ROUTING_LAYER)
+      && (getTech()->getLayer(lNum + 2)->getDir() == lDir2)) {
+    return lNum + 2;
+  }
```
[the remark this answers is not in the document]
smunaut: Ok, I had indeed overlooked that, I will check against both getTech()->getTopLayerNum() and TOP_ROUTING_LAYER
Same for BOT_ROUTING_LAYER, it could be 0 which is less than getBotLayerNum() ( which would be 2 ).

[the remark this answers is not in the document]
osamahammad21: actually BOTTOM_ROUTING_LAYER is ok because the default value is always valid. The first two layers are a master slice layer followed by a via layer. So the bottom routing layer should not be below layer idx 2.

src/drt/src/dr/FlexDR_init.cpp
smunaut: Fixed, tests seem to pass.

src/drt/src/dr/FlexDR_init.cpp
smunaut: Fixed

src/drt/src/dr/FlexDR_init.cpp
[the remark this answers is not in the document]
smunaut: Added tests to bound by tech layer as well.

src/drt/src/dr/FlexDR_init.cpp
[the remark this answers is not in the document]
smunaut: Done.

smunaut:
Ping on this ? Anything else that needs changing ?

maliberty:
@osamahammad21 please re-review

smunaut:
Ping ?

osamahammad21:
@smunaut So sorry for the late reply. I am running a secure CI. Once it passes I'll go ahead and merge

osamahammad21 (approved):

smunaut:
@maliberty I think you need to re-approve :sweat_smile:

PR #7108, opened 2025-04-12
ppl: move top layer grid definition to odb

eder-matheus:
Changes in this PR

1. Add dbBTermTopLayerGrid structure to dbBlock, in order to store the top layer grid used to place bumps.
2. Implement checks to avoid creating a top layer grid out of the die area. Also, resize the region if it would allow pins out of the die area.
3. Move the Tcl command to define the top layer grid to odb.tcl.
4. Update PPL code to use the data from ODB instead of having its own copy of the data.


src/odb/src/db/dbBlock.cpp
```diff
@@ -1600,6 +1630,94 @@ void dbBlock::addBTermGroup(const std::vector<dbBTerm*>& bterms, bool order)
...
+{
+  _dbBlock* block = (_dbBlock*) this;
+  _dbBTermTopLayerGrid& top_layer_grid = block->_bterm_top_layer_grid;
+
+  utl::Logger* logger = block->getImpl()->getLogger();
+
+  const int half_width = pin_width / 2;
+  const int half_height = pin_height / 2;
+  const odb::Rect& die_area = getDieArea();
+  bool region_changed = false;
```
maliberty, line 1649: I think the db should not be modifying its input and should just error on invalid input.  The user interface can choose to modify the input on the way to the db if it wishes.  

That said I think we are being overly helpful and I don't know that shifting the pin locations is the right thing to do.  Often bump locations are quite fixed by the package design.   @gadfort opinion?

maliberty (changes requested):

src/odb/src/db/dbBlock.cpp
gadfort: I think the out of die area concern should be addressed in ppl, it seems strange to adjust the area on behalf of the user (and strangely couples ppl concerns into odb). ppl can just choose not to allocate slots to pins that would exceed the die area.

src/odb/include/odb/db.h
```diff
@@ -892,6 +892,23 @@ class dbBlock : public dbObject
     std::vector<dbBTerm*> bterms;
     bool order = false;
   };
+
+  struct dbBTermTopLayerGrid
+  {
+    int layer_id = -1;
+    int x_step = -1;
+    int y_step = -1;
+    Rect region;
```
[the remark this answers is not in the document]
gadfort: Personally, I think we might as well change to polygons

src/odb/include/odb/db.h
```diff
@@ -892,6 +892,23 @@ class dbBlock : public dbObject
     std::vector<dbBTerm*> bterms;
     bool order = false;
   };
+
+  struct dbBTermTopLayerGrid
+  {
+    int layer_id = -1;
+    int x_step = -1;
+    int y_step = -1;
```
gadfort, line 900: std::optional?

src/odb/include/odb/db.h
```diff
@@ -892,6 +892,23 @@ class dbBlock : public dbObject
     std::vector<dbBTerm*> bterms;
     bool order = false;
   };
+
+  struct dbBTermTopLayerGrid
+  {
+    int layer_id = -1;
```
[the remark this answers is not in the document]
gadfort: I would prefer to return dbTechLayer* as well

src/odb/include/odb/db.h
```diff
@@ -892,6 +892,23 @@ class dbBlock : public dbObject
...
+    int x_step = -1;
+    int y_step = -1;
+    Rect region;
+    int pin_width = -1;
+    int pin_height = -1;
+    int keepout = -1;
```
gadfort, line 904: std::optional?

src/odb/include/odb/db.h
```diff
@@ -892,6 +892,23 @@ class dbBlock : public dbObject
...
+
+    int llx() const { return region.xMin(); }
+    int lly() const { return region.yMin(); }
+    int urx() const { return region.xMax(); }
+    int ury() const { return region.yMax(); }
```
gadfort, line 909: this seems redundant

src/odb/src/db/dbBlock.cpp
eder-matheus: @gadfort Would you expect PPL to throw a warn message about not using the slots that would exceed the die area? Or should the tool just be silent about it?

gadfort: Silent would be okay. It feels like if you define a region you would expect the pins to be placed inside the region, not straddling the boundary of the region

src/odb/include/odb/db.h
```diff
@@ -971,6 +983,27 @@ class dbBlock : public dbObject
   ///
   void addBTermGroup(const std::vector<dbBTerm*>& bterms, bool order);
 
+  ///
+  /// Define the top layer grid for pin placement.
+  ///
+  void setBTermTopLayerGrid(dbTechLayer* layer,
```
maliberty, line 989: Why doesn't this take a `const dbBTermTopLayerGrid&`?  The fields are optional in the struct but mandatory in this API.

src/odb/include/odb/db.h
```diff
@@ -892,6 +892,18 @@ class dbBlock : public dbObject
     std::vector<dbBTerm*> bterms;
     bool order = false;
   };
+
+  struct dbBTermTopLayerGrid
+  {
+    dbTechLayer* layer = nullptr;
+    std::optional<int> x_step;
+    std::optional<int> y_step;
```
maliberty, line 900: What does it mean when the optional fields are not specified?  What is the `keepout`?  Please add some descriptive comments.

src/odb/src/db/dbBlock.cpp
```diff
@@ -1604,6 +1634,76 @@ void dbBlock::addBTermGroup(const std::vector<dbBTerm*>& bterms, bool order)
...
+{
+  _dbBlock* block = (_dbBlock*) this;
+
+  dbBlock::dbBTermTopLayerGrid top_layer_grid;
+
+  odb::dbTech* tech = getDb()->getTech();
+  dbTechLayer* layer = nullptr;
+  const auto& top_grid = block->_bterm_top_layer_grid;
+
+  if (top_grid.layer.isValid()) {
```
maliberty, line 1675: When would it not be valid?  `top_layer_grid.layer = layer->getId();` is going to fail if layer is null (which should be checked and error'ed).

src/odb/include/odb/db.h
eder-matheus: I'm using this function in the odb.tcl to define the top layer grid. I understand that if I set the parameter as a `const dbBTermTopLayerGrid&`, it would require some swig work so Tcl can understand it.

src/odb/include/odb/db.h
eder-matheus: I believe the `std::optional` change I've made yesterday doesn't really makes sense. All the files are mandatory when creating the top layer grid.

I'll add the comments for each field and remove the std::optional.

src/odb/src/db/dbBlock.cpp
eder-matheus: Insided PPL we are calling the `getBTermTopLayerGrid` function even when the top layer grid was not defined. Then it checks if the layer is `nullptr` to detect if it was defined or not. I'll change this logic on PPL to avoid calling this function when we don't have the top layer grid defined.

src/odb/include/odb/db.h
maliberty: You could pass the values from tcl to the .i discretely and assemble the structure in the .i file if needed.

src/odb/src/db/dbBlock.cpp
```diff
@@ -1634,65 +1634,48 @@ void dbBlock::addBTermGroup(const std::vector<dbBTerm*>& bterms, bool order)
...
-  top_layer_grid.pin_width = pin_width;
-  top_layer_grid.pin_height = pin_height;
-  top_layer_grid.keepout = keepout;
+  top_grid.layer = top_layer_grid.layer->getId();
+  top_grid.x_step = top_layer_grid.x_step;
+  top_grid.y_step = top_layer_grid.y_step;
+  top_grid.region = top_layer_grid.region;
+  top_grid.pin_width = top_layer_grid.pin_width;
+  top_grid.pin_height = top_layer_grid.pin_height;
+  top_grid.keepout = top_layer_grid.keepout;
```
maliberty, line 1657: Please check for legal values.

maliberty (approved):

PR #7113, opened 2025-04-12
bazel: Add config with debug flags useful for creating a profile.

hzeller:
To create a executables that keep enough debug information to retain useful symbols for use in e.g. running Linux `perf`

```
bazel build -c opt --config=profile //:openroad
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -41,6 +41,12 @@ build:asan --copt -g
 build:asan --copt -fno-omit-frame-pointer
 build:asan --linkopt -fsanitize=address
 
+# Flags with enough debug symbols to get useful outputs with Linux `perf`
+build:profile --strip=never
+build:profile --copt -g
+build:profile --copt -gmlt
+build:profile --copt -fno-omit-frame-pointer
+
 # TODO: document
 build --incompatible_strict_action_env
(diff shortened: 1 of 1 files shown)


hzeller:
Example (assuming Linux `perf`, and the `pprof` and `perf_to_profile` tools are installed:

```
(cd test ; time perf record -o /tmp/or-test.perf -- $(bazel run -c opt --config=profile --run_under="echo" //:openroad) -no_splash -no_init -exit ./upf_aes.tcl)
```

Then run `pprof` on that profile output:

```
pprof -http=localhost:8080 /tmp/or-test.perf
```

Despite the configuration options, the flame graph is still not very useful: There is a huge chunk of memcpy, which indicates that there is some sub-optimal use of containers, however it is hard to address as the containers itself don't show up:
flame-graph-or

I suspect we also need to be able to load the corresponding debug symbols for libc++ in order to have something useful and find where most of that stuff is originating from. @QuantamHD , you have some experience with running `perf` - are there other flags I should be adding ?

maliberty (approved):
I think it is time to start some Bazel documentation.  The write up in this PR is good but will not be found by most people

PR #7124, opened 2025-04-13
fixes The-OpenROAD-Project/OpenROAD#7054

osamahammad21:


src/drt/src/db/tech/frLayer.h
```diff
@@ -123,7 +123,10 @@ class frLayer
     return secondaryViaDefs_.at(idx);
   }
   bool hasVia2ViaMinStepViol() { return hasMinStepViol_; }
-  std::set<const frViaDef*> getViaDefs() const { return viaDefs_; }
+  std::set<const frViaDef*, frViaDefComp> getViaDefs() const
```
maliberty, line 126: Use frOrderedIdSet rather than std::set.  Same in drNet::fNetTerms_, MarkerId::srcs, FlexGCWorker::Impl::targetObjs_, etc.  Please look for all uses of std::set, std::unordered_set, std::map, and std::unordered_map.

osamahammad21: - frViaDef is not frBlockObject, so frOrderedIdSet cannot be used
- I thought of using it with fNetTerms but it won’t be handled correctly. This is because the term id is unique per instance not globally l. So there could exist multiple terms in the same set with the same id which would mess things up. As for srcs, it would be much worse because it would be comparing ids of completely different frBlockObject types. I will handle this but in a separate PR.

maliberty: I missed that it isn't a common base class for tech/*.  Ok to defer but in general we need to be careful with std::set/unordered_set as we've had multiple cases of non-determinism due to pointer ordering and iteration.

maliberty (approved):

PR #7127, opened 2025-04-14
gui: fix inf loop on huge numbers

gadfort:
The casting to and from floats and int causes large numbers to overflow and the while loops just get stuck when presented with a huge difference in min and max values.


maliberty (approved):
This looks fine but I don't understand the case where we have a value so large as to cause an overflow.  Are we including INF?

gadfort:
@maliberty I was editing a liberty file (scary) and I messed something up causing the setup timing for a cell to jump to a huge number (not INF), but because it was greater than MAX_INT, it caused the code to overflow and for the while loops it was game over.

PR #7130, opened 2025-04-14
rsz: honor dont touch in rebuffer for :load

gadfort:
Fixes: disconnects on load pins which are marked dont touch

### DIFF
--- a/src/rsz/src/Rebuffer.cc
+++ b/src/rsz/src/Rebuffer.cc
@@ -937,11 +937,23 @@ int RepairSetup::rebufferTopDown(const BufferedNetPtr& choice,
     }
 
     case BufferedNetType::load: {
+      const Pin* load_pin = choice->loadPin();
+
+      if (resizer_->dontTouch(load_pin)) {
+        debugPrint(logger_,
+                   RSZ,
+                   "rebuffer",
+                   3,
+                   "{:{}s}connect load: skipped on {} due to dont touch",
+                   "",
+                   level,
+                   sdc_network_->pathName(load_pin));
+        return 0;
+      }
+
       odb::dbNet* db_net = nullptr;
       odb::dbModNet* db_modnet = nullptr;
       db_network_->staToDb(net, db_net, db_modnet);
-
-      const Pin* load_pin = choice->loadPin();
 
       // only access at dbnet level
       Net* load_net = network_->net(load_pin);
--- a/src/rsz/src/RepairSetup.cc
+++ b/src/rsz/src/RepairSetup.cc
@@ -543,9 +543,10 @@ bool RepairSetup::repairPath(PathRef& path,
                RSZ,
                "repair_setup",
                3,
-               "Path slack: {}, repairs: {}",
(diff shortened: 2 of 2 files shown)


maliberty:
I think this is ok but would like @precisionmoon or @povik to confirm.

gadfort:
@maliberty It's very hard to follow that chunk of code, so it's entirely possible that the fix should have been elsewhere

povik (approved):
Looks fine (we should account for the don't touch load in evaluation of the buffering choice, but we can make this change together with other rebuffering changes which are incoming)

povik:
> @maliberty It's very hard to follow that chunk of code, so it's entirely possible that the fix should have been elsewhere

@gadfort This is fine in so far as making sure we don't disconnect the don't touch load, but it can lead to suboptimal buffering results when there's mixture of don't touch, non-don't touch loads

gadfort:
@povik if we've marked something dont_touch it's fine to get less than optimal results (I need to do this until we have proper hierarchy support), and the tool should move on and try something else to resolve it.

precisionmoon:
The fix looks fine.   Dont-load on load pin is like a dont-touch on the relevant net.   @gadfort, can you also add a regression test so we don't break this later?

povik:
> Dont-load on load pin is like a dont-touch on the relevant net.

@precisionmoon I think we can have a don't-touch on the load pin but not on the net connected to the load pin so that other loads on the same net are free to be buffered. E.g. this gain buffering test checks for treatment of exactly that: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/test/gain_buffering1.tcl#L20

povik:
> https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/test/gain_buffering1.tcl#L20

To be precise In the test the don't touch is on the instance, and by extension on the load pin, but not on the net

precisionmoon:
> > Dont-load on load pin is like a dont-touch on the relevant net.
> 
> @precisionmoon I think we can have a don't-touch on the load pin but not on the net connected to the load pin so that other loads on the same net are free to be buffered. E.g. this gain buffering test checks for treatment of exactly that: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/test/gain_buffering1.tcl#L20

You bring up a good point.   set_dont_touch applies to nets or instances only, not pins.    Commercial tools have a separate set_dont_touch_network command that accepts pins or ports and applies dont-touch to the transitive fanout logic.

In gain_buffering1.tcl, there is a dont-touch on an instance.   We should be able to add a buffer at the output net of the instance.

As for the dont-touch on a load pin, how should we define the dont-touch behavior?
1) If this comes from the dont-touch of its cell, the load pin net should be bufferable.
2) If this comes from the connected net, then the net should not be.

gadfort:
@precisionmoon I cannot find a way to replicate the issue for a testcase without juts sharing the whole design unfortunately, so at this time a testcase is not possible

maliberty (approved):

Issue #6352, opened 2024-12-13
cts fails to build clock tree when clock is also an output port


gadfort:
### Describe the bug

In this design the clock is getting forwarded to an output port, ie. `assign mclk = dco_clk`
This causes openroad to put a buffer between the two ports, which appears to cause CTS to skip clock tree generation all together.

### Expected Behavior

Build a clock tree on `dco_clk`

### Environment

```shell
OpenROAD v2.0-17598-ga008522d8 
Features included (+) or not (-): +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
```

### To Reproduce

https://drive.google.com/file/d/1GijuZkVXgaxFMtEAyJejZCT3I8CQEH9k/view?usp=sharing

```
tar xvf sc_issue_openmsp430_job0_cts.clock_tree_synthesis0_20241213-143445.tar.gz
cd sc_issue_openmsp430_job0_cts.clock_tree_synthesis0_20241213-143445
./run.sh
```

### Relevant log output

```shell
[INFO CTS-0007] Net "dco_clk" found for clock "clk".
[WARNING CTS-0105] Net "dco_clk" already has clock buffer _12043_. Skipping...
[WARNING CTS-0105] Net "smclk" already has clock buffer _12043_. Skipping...
[WARNING CTS-0105] Net "mclk" already has clock buffer _12036_. Skipping...
[WARNING CTS-0105] Net "aclk" already has clock buffer _12029_. Skipping...
[WARNING CTS-0083] No clock nets have been found.
[INFO CTS-0008] TritonCTS found 0 clock nets.
[WARNING CTS-0082] No valid clock nets in the design.
[INFO RSZ-0058] Using max wire length 232um.
```

precisionmoon:
As a temporary workaround, can you manually remove the feedthrough buffers before CTS?

gadfort:
@precisionmoon as a temporary bypass yes, I think that should be doable.

precisionmoon:
@gadfort, thanks for the testcase.  We'll review it to see how we can eliminate the workaround.

gadfort:
@maliberty @precisionmoon any update on this?

maliberty:
@arthurjolo please update

arthurjolo:
[...]

PR #7134, opened 2025-04-14
CTS fix wrong clock net skiping

arthurjolo:
Fixes [#6352](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6352)

CTS skips building the clock tree for clock nets that already have a clock tree. The way CTS checks if the clock tree is already generated is by looking at the clock nets and checking if there is a clock buffer in it and if the master of the buffer is the same as the masters used by CTS.

This check can cause a misunderstanding, if there is a buffer in the clock tree that is not from a clock tree but has the same master, and it will skip creating a clock tree when it shouldn't.

This PR fixes this by also checking the source type of the buffer, and only skips if the source of the buffer is `TIMING`.

### DIFF
--- a/src/cts/src/TritonCTS.cpp
+++ b/src/cts/src/TritonCTS.cpp
@@ -1159,7 +1159,8 @@ bool TritonCTS::separateMacroRegSinks(
   for (odb::dbITerm* iterm : net->getITerms()) {
     odb::dbInst* inst = iterm->getInst();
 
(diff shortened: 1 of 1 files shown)


maliberty (approved):
@gadfort please confirm this works for you

gadfort:
@arthurjolo @maliberty this appears to be working in the design I have. Thanks

PR #7137, opened 2025-04-14
drt: flexGCworker end deletion

bnmfw:
I'm debugging a PA issue, and while looking through flexGC found this.
Is there a reason for this to exist, should there be some joining or deletion?
Given that the function has been untouched for some years I'm guessing it is just dead code.


bangqixu:
My guess is that there was some print-out message to gather how long it takes for GC to completes. Probably for benchmarking purposes. There must be better solutions, but we did not know at that time.

maliberty (approved):

maliberty:
You need to update the cmake file
image

Issue #3985, opened 2023-09-13
Add support for Python scripting in GUI


mithro:
### Description

Currently the OpenROAD GUI has a "Scripting" window which accepts TCL commands. TCL isn't a commonly known language and not that user friendly, while Python has become one of the most popular languages on the planet and is well praised for its user friendliness.

### Suggested Solution

It would be **_really_** nice if you could select the "Scripting" window to be either the TCL or Python language.

maliberty:
There is a start to this in https://github.com/The-OpenROAD-Project/OpenROAD/pull/2762 but I don't think @gadfort has had any time for it recently.

gadfort:
@maliberty yeah, that fell through the cracks.
Basically, there are a couple of issues:
- the logging doesn't go through the logger, so that doesn't work right
- exit and hide needs to be reworked to handle the GUI properly

Issue #1759, opened 2022-04-06
Meta: Top 3 Painpoints the Toolchain


QuantamHD:
What are the top three pain points of OpenROAD for chip design?

@antonblanchard 
@arlpetergadfort 
@maliberty
@proppy

maliberty:
@donn

donn:
Question is maybe better suited for our chip designers: @kareefardi 

Me personally, I have no complaints, just would like a Python interface, which is a monumental effort.

arlpetergadfort:
@QuantamHD 
1. Parasitic extraction, requiring a different tool to jumpstart the PDK bring up
2. Yosys/ABC synthesis, generates a netlist that requires a lot of additional work to meet constraints.

I'm sure there are more, but these are the two that stand out the most to me at the moment.

kareefardi:
I am not sure what is the state of any of these today but mainly my struggles are:

- automatic macro placement
- signal routing with multiple macros
- (multiple domain) power routing of multiple macros

msaligane:
- System Verilog support in Yosys. The generated netlist doesn't work (probably I am doing something wrong but it isn't an easy flow)
- Power planning and routing
- Timing results are really hard to understand (No hockey stick trend yet)

maliberty:
@msaligane would you elaborate on your last point.

mithro:
@hzeller / @kgugala -- See @msaligane's comment about SystemVerilog above.

maliberty:
Lack of timing model generator for hierarchical flows.  LEF abstract generator improvements would then follow.

antonblanchard:
- +1 for "Lack of timing model generator for hierarchical flows". Even a basic (and pessimistic) liberty file generator would be very useful.
- +1 for "ABC/synthesis". We don't pass FFs, manually instantiated standard cells or other black boxed things to ABC, so are missing a lot of optimisations including the ability to do register retiming. I've looked into this but tripped up on what looks like ABC issues.

msaligane:
> @msaligane would you elaborate on your last point.


@maliberty This is what I meant by the hokey stick (or knee). Below is the expected timing on a RISC-V core when I push the clock:
![image](https://user-images.githubusercontent.com/48068702/162547909-0b9f1cb8-86ca-4cb4-80df-04aa3e6daea4.png)

This is a reference from [Ketkar, et al ICCAD02] with area instead of power but the two correlates:
[...]

PR #7141, opened 2025-04-15
bazel: fix build, reduce build dependencies.

hzeller:
Simple c++ to avoid dependency-hungry Python.

Context: #7120

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -5,16 +5,16 @@ load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load("@rules_hdl//dependency_support/com_github_westes_flex:flex.bzl", "genlex")
 load("@rules_hdl//dependency_support/org_gnu_bison:bison.bzl", "genyacc")
 load(
-    "//:bazel/build_helper.bzl",
+    "//bazel:build_helper.bzl",
     "OPENROAD_BINARY_SRCS_WITHOUT_MAIN",
     "OPENROAD_LIBRARY_HDRS_INCLUDE",
     "OPENROAD_LIBRARY_INCLUDES",
     "OPENROAD_LIBRARY_SRCS_EXCLUDE",
     "OPENROAD_LIBRARY_SRCS_INCLUDE",
 )
-load("//:bazel/tcl_encode_sta.bzl", "tcl_encode_sta")
-load("//:bazel/tcl_encode_or.bzl", "tcl_encode")
-load("//:bazel/tcl_wrap_cc.bzl", "tcl_wrap_cc")
+load("//bazel:tcl_encode_or.bzl", "tcl_encode")
+load("//bazel:tcl_encode_sta.bzl", "tcl_encode_sta")
+load("//bazel:tcl_wrap_cc.bzl", "tcl_wrap_cc")
 
 package(
     features = [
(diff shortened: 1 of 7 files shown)


maliberty:
I don't share your " One personal goal in my life is to keep the amount of Python people of to suffer through to a minimum :)".   I actually feel that way about TCL but have no issue with Python.

As stated in #7120 "BUILD_PYTHON=false is not a long term solution and OR will require python itself."  So we are saving no dependency by doing this.  I do have to validate your implementation of b64 encoding which is a negative.

At a minimum you would also need to update the cmake build as I don't want to have two implementation of this.  If fixing py_binary isn't hard I think it would be better.

hzeller:
Ok, if you want to work on unbreaking the build for the Python binary, making a `py_binary()` is similar to a `cc_binary` give it a name and the srcs

```Python
py_binary(
    name = "file_to_string",
    srcs = ["//:etc/file_to_string.py"],
)
```

if you want to wire it up in the `//bazel` package but refer to the location in `etc/`, you might need to add `"etc/file_to_string.py"` in the `exports_files` in the toplevel bazel.

@QuantamHD can give guidance how to wire that up in that rule.

oharboe:
Some pointers on how to use python from bazel, which in my experience works quite nicely.

No need to rely on system installed python or add README.md instructions on dependencies. Run `bazel bulid/run` as the case may be and it works.


https://github.com/The-OpenROAD-Project/bazel-orfs/blob/89d264ebe055dac4dc4a85fedf854f156569bf7d/MODULE.bazel#L59

https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/requirements.in

https://github.com/The-OpenROAD-Project/bazel-orfs/blob/89d264ebe055dac4dc4a85fedf854f156569bf7d/BUILD#L395

https://github.com/The-OpenROAD-Project/bazel-orfs/blob/89d264ebe055dac4dc4a85fedf854f156569bf7d/ppa.bzl#L41

maliberty:
@oharboe do you want to fix it given your experience?

oharboe:
> @oharboe do you want to fix it given your experience?

:crossed_fingers:  https://github.com/The-OpenROAD-Project/OpenROAD/pull/7167

hzeller:
I simply don't like Python, as it is a language with whitespaces having a meaning, interpreted, has a huge runtime, etc.
So yeah, I prefer C++ of course as a forever language: write it once, will compile and work forever.

In general, Python is unproblematic in Bazel, so some people that feel less pain seeing Python use it :)

maliberty:
You seem to be at odds with requests from Google to turn openroad into a pip module and not a standalone application at all.  https://github.com/The-OpenROAD-Project/OpenROAD/issues/1424 https://github.com/The-OpenROAD-Project/OpenROAD/issues/3985

Request for a python interface: https://github.com/The-OpenROAD-Project/OpenROAD/issues/1759 from Donn.

Every ML group I've talked to has asked for a Python API and not a tcl or c++ one.

hzeller:
Yeah, I am just a personal spare-time contributor, independent from any place I might work at.

maliberty:
Replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/7167

PR #7144, opened 2025-04-15
DPO: remove dead code "router.h"

osamahammad21:


src/dpo/src/detailed_manager.cxx
```diff
@@ -192,50 +186,6 @@ void DetailedMgr::findBlockages(const bool includeRouteBlockages)
     }
   }
 
-  if (includeRouteBlockages && rt_ != nullptr) {
-    // Turn M1 and M2 routing blockages into placement blockages.  The idea
```
maliberty, line 196: Similar to the pdn stripe issue that we need to address elsewhere.

maliberty (approved):

Issue #3046, opened 2023-03-20
Add "Show in..." feature to OpenROAD timing paths


oharboe:
### Description

Select a data path row, then right click and get a menu to "Show in ..."

![image](https://user-images.githubusercontent.com/2798822/226295495-bf15596b-c2e1-41f1-8a9e-9141fc9f5596.png)

It would be helpful to go from timing paths elements to other parts of the design(backannotation).

Some ideas for features:

- [ ] Show original source code Verilog/RTL code line corresponding to row
- [ ] Show Yosys schematics for original Verilog/RTL as well as current .odb file
- [ ] Display cone of logic or path full path of logic
- [ ] Corresponding .sdc file

### Suggested Solution

Implement in GUI...

gadfort:
Logic cone might be relatively easy since that is already supported in the inspector. I think source line annotation would require additional support from OpenSTA to provide that information.

rovinski:
Showing the original RTL from the timing path seems like a doozy which would probably require a lot of hacking in yosys. OpenROAD starts from a netlist which has no idea what RTL went into it. It would require yosys to annotate the source lines into the generated netlist, and then OpenSTA would have to parse those annotations along with the netlist and then expose them through an API to the GUI.

It sounds like a wonderful feature, but I think that an easier stopgap is to look at the source and destination register names (assuming they are unmangled) and infer the source line from there.

maliberty:
I think the OR work to support this is tractable if we can get yosys/abc to give us the line numbers.

QuantamHD:
Yeah I have a mostly finished PR in OpenSTA which reads the src line annotations that yosys adds to the verilog. Our flow at Google usually keeps around a good amount of the src line annotations. I'm not sure why the ORFS doesn't. @maliberty has asked me to look into it, but I just haven't had the time. At minimum the DFF's should have line numbers.

maliberty:
The fact that we use -noattr on write_verilog in flow/scripts/synth.tcl suppresses the annotations.  If I turn that off I see line numbers for the ff and some wires though not for the combinational logic.

rovinski:
I would be pleasantly surprised if yosys had source line annotation of combinational logic, considering it has to hand off between yosys and ABC.

QuantamHD:
There's an abc pass that's supposed to help with that, but as of yet I haven't been super successful using it.

oharboe:
[...]

PR #7148, opened 2025-04-15
mpl: create abstractions for intervals and tilings

### LINKED ISSUE #7092: MPL should have structs for shape curves and tilings
### Description

Requested [here](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7027#discussion_r2036420240).

### Suggested Solution

Have objects for those abstractions.

### Additional Context

_No response_

AcKoucher:
Resolve #7092 

Previously, width/height intervals and tilings were `std::pair<float, float>`. Now, we have proper objects for them:
 - `Interval` is a simple struct to describe the bounds of a certain dimension.
 - `Tiling` is a small class that represents the coarse shape of clusters with macros.


AcKoucher:
I'll run a Secure-CI if the approach here if fine.

src/mpl/src/shapes.h
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2021-2025, The OpenROAD Authors
```
maliberty, line 2: start 2025

src/mpl/src/shapes.h
```diff
@@ -0,0 +1,38 @@
...
+namespace mpl {
+
+// Bounds of a certain cluster's dimension.
+// Used for either width or height.
+struct Curve
+{
+  Curve(float min, float max) : min(min), max(max) {}
+
+  float min{0.0f};
+  float max{0.0f};
```
maliberty, line 15: Could these be const?

src/mpl/src/util.h
```diff
@@ -33,10 +35,17 @@ struct PenaltyData
...
+                             const Curve& width_curve_b)
+{
+  return width_curve_a.min < width_curve_b.min;
+}
+
+// Utility to help sorting tilings.
+inline bool isAreaSmaller(const Tiling& tiling_a, const Tiling& tiling_b)
 {
-  return p1.first < p2.first;
+  return tiling_a.area() < tiling_b.area();
```
maliberty, line 48: It would be good to have a secondary criteria to handle equal areas (eg width)

src/mpl/src/util.h
```diff
@@ -33,10 +35,17 @@ struct PenaltyData
   float normalization_factor{0.0f};
 };
 
-inline bool isFirstSmaller(const std::pair<float, float>& p1,
-                           const std::pair<float, float>& p2)
+// Utility to help sorting width curves.
+inline bool isMinimumSmaller(const Curve& width_curve_a,
```
maliberty, line 39: This is a confusing name.  isWidthSmaller ?

src/mpl/src/shapes.cpp
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2021-2025, The OpenROAD Authors
```
maliberty, line 2: start date

src/mpl/src/shapes.h
AcKoucher: No, because there are situations in which in these intervals are updated.

src/mpl/src/shapes.h
```diff
@@ -0,0 +1,38 @@
...
+  float height() const { return height_; }
+  float area() const { return width_ * height_; }
+  float aspectRatio() const { return height_ / width_; }
+
+  bool operator==(const Tiling& tiling) const;
+  bool operator<(const Tiling& tiling) const;
+
+ private:
+  float width_{0.0f};
+  float height_{0.0f};
```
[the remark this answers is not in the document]
AcKoucher: We use `std::sort` to sort tilings during shaping. If we make these members `const` we won't be able to do that.

maliberty:
go ahead with ci

AcKoucher:
Found the problem. Re-running Secure-CI.

AcKoucher:
Secure-CI is fine. All results are identical except for one private design which shows a minor discrepancy beginning at the RSZ iterations during CTS (it looks the same problem reported in [#3046](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3046).

maliberty (approved):

PR #7149, opened 2025-04-15
gpl: simplify divergence detection logic

gudeh:
Instead of using both an error counter per region and a boolean flag to check for divergences, use only the counter and check if it's greater than zero where needed.

Remove unnecessary variables. Since NesterovPlace reports the errors, there's no need for a separate error message in NesterovBase.

Ensure that the divergence snapshot is stored after checking for divergence.

### DIFF
--- a/src/gpl/src/nesterovPlace.cpp
+++ b/src/gpl/src/nesterovPlace.cpp
@@ -102,7 +102,7 @@ void NesterovPlace::updatePrevGradient(const std::shared_ptr<NesterovBase>& nb)
   // Wirelength / density gradient calculation
   if (std::isnan(wireLengthGradSum_) || std::isinf(wireLengthGradSum_)
       || std::isnan(densityGradSum_) || std::isinf(densityGradSum_)) {
-    isDiverged_ = true;
+    num_region_diverged_ = 1;
     divergeMsg_ = "RePlAce diverged at wire/density gradient Sum.";
     divergeCode_ = 306;
   }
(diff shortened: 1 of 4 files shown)


maliberty:
@mikesinouye please try this on your case

src/gpl/src/nesterovPlace.cpp
```diff
@@ -768,6 +751,22 @@ void NesterovPlace::updateDb()
...
+void NesterovPlace::checkInvalidValues(float wireLengthGradSum,
+                                       float densityGradSum)
+{
+  if (std::isnan(wireLengthGradSum) || std::isnan(densityGradSum)
+      || std::isinf(wireLengthGradSum) || std::isinf(densityGradSum)) {
+    divergeMsg_
+        = "RePlAce diverged at wire/density gradient Sum. An internal value is "
+          "NaN or Inf.";
+    divergeCode_ = 306;
+    num_region_diverged_ = 1;
```
maliberty, line 765: Does setting it to one here conflict with counting other regions?

mikesinouye:
Thanks for working on this refactor @gudeh! I will try this out.

src/gpl/src/nesterovPlace.cpp
gudeh: I don't think so. I reviewed the logic and [the variable is refreshed ](https://github.com/gudeh/OpenROAD/blob/43aaf424c7d64872c7b502455854e701b7543691/src/gpl/src/nesterovPlace.cpp#L321C8-L321C32) whenever we check if a divergence happens.

mikesinouye:
This PR fixes the issue I raised in #7147. Thanks for the fix, I will close that PR.

PR #7167, opened 2025-04-17
bazel: idiomatic Python usage example

oharboe:
@hzeller Thoughts?


hzeller:
I'll leave this to @QuantamHD to have a look

etc/BUILD
```diff
@@ -0,0 +1,9 @@
+py_binary(
+    name = "file_to_string",
+    srcs = [
+        "file_to_string.py",
+    ],
+    main = "file_to_string.py",
+    visibility = ["//visibility:public"],
+    # deps = ["@openroad-pip//matplotlib"],
```
hzeller, line 8: this looks like a leftover comment.

MODULE.bazel
```diff
@@ -77,3 +77,24 @@ use_repo(llvm, "llvm_toolchain")
 
 # FYI: Comment out on NixOS where you'd like to use the local clang toolchain.
 register_toolchains("@llvm_toolchain//:all")
+
+bazel_dep(
+    name = "rules_python",
+    version = "1.2.0",
+    # reduce namespace collisions by prefixing with openroad
```
hzeller, line 84: what are the namespace collisions you observe ?

etc/BUILD
oharboe: It is. I thought a pip install something would be required.

MODULE.bazel
oharboe: I am not sure this is enough, but if multiple MODULEs are involved they should be able to use different python pip venvs side by side.

hzeller: I suspect this will not be an issue. I'd say for now do not do this, and once a problem of that nature arises we can tackle it then.

etc/BUILD
hzeller: file_to_string doesn't really have a `mathplotlib` dependency. Looking at the Python file, I see `base64`, `sys`, `os` and `argsparse`.
There is also a superfluous `json` dependency in the Python file, but that looks like a leftover that can be removed.

oharboe: none of these need be pip installed

QuantamHD (approved):

QuantamHD:
LGTM. I do think that Henner's PR that uses C++ instead of python is more ideal, but if we want to keep python around for this I'm fine.

maliberty:
What makes it more ideal?

MODULE.bazel
oharboe: I disagree. The problem that I ran into was with bazel-orfs using "pip" + a repository using bazel-orfs also using "pip". If one copies and pastes examples on how to use Python from the net, one is likely to use the "pip" identified and I think it is better to use a "openroad-pip" identified to try to be a good citizen.

oharboe: I didn't keep the details of this namespace collisions. Someone who knows all there is to know about bazel could probably explain :-)

Here is what I got out of x.ai, but I have no idea if this is completley accurate, but it matches *my* hallucinations about how Bazel works and describes the problem I'm trying to avoid and future proof against:

Avoiding Repository Name Conflicts in Multi-Module Projects:

In Bazel’s bzlmod system (used with MODULE.bazel), external dependencies are registered as repositories. Each repository must have a unique name to avoid collisions across the dependency graph. When multiple modules (or projects) depend on rules_python, using the default repository name (rules_python) can lead to conflicts if different modules require different versions of rules_python or configure it differently (e.g., with different Python versions).
By prefixing the repository name with openroad_ (e.g., openroad_rules_python), you create a unique namespace for your project’s instance of rules_python. This ensures that your module’s configuration does not clash with other modules that might also depend on rules_python but use a different version or configuration.
Similarly, openroad-pip as the hub_name for the pip.parse extension ensures that the external repositories generated for Python dependencies (from requirements_lock_3_13.txt) are uniquely namespaced, preventing conflicts with other pip.parse invocations that might use the default pip hub name.
Example Scenario:

Suppose your project depends on two modules, module_a and module_b, both of which use rules_python and pip.parse. If both modules use the default rules_python and pip names, Bazel may fail to resolve the dependency graph if the versions or configurations differ (e.g., module_a uses Python 3.11, and module_b uses Python 3.13). By using openroad_rules_python and openroad-pip, your project’s configuration is isolated, avoiding such conflicts.

oharboe:
> LGTM. I do think that Henner's PR that uses C++ instead of python is more ideal, but if we want to keep python around for this I'm fine.

Lots of Python is used:

```
$ find . -regex .*\.py\$ | grep -v test | wc -l
42
```

Some examples that aren't used yet, AFAIK, but should be used in Bazel build: generate and check documentation for correctness.

```
./docs/getMessages.py
./docs/src/scripts/manpage.py
./etc/find_messages.py
```

oharboe:
@maliberty No further changes planned on my part

maliberty (approved):

hzeller:
`encode_tcl()` is still not working #7226

PR #7174, opened 2025-04-17
Cts use different parameters to cluster macros

arthurjolo:
Currently the macro tree uses the same parameters to cluster the its sinks (cluster size, cluster diameter and minimum number of sinks to enable clustering), however due to the size of the macros CTS almost never finds a clustering solution. In this PR new parameters for clustering the macro tree sinks are used, allowing the macros to be clustered.

This PR uses default values for clustering parameters:

    cluster size = 4
    cluster diameter = 2 * largest macro
    minimum number of macros = 10

Eg. ng45 swerv_wrapper:

Not clustering macros
image

Clustering macros
image (5)


src/cts/test/balance_levels.ok
```diff
@@ -13,7 +13,7 @@
 [INFO CTS-0201] 0 blockages from hard placement blockages and placed macros will be used.
 [INFO CTS-0027] Generating H-Tree topology for net clk.
 [INFO CTS-0028]  Total number of sinks: 1.
-[INFO CTS-0029]  Sinks will be clustered in groups of up to 5 and with maximum cluster diameter of 60.0 um.
+[INFO CTS-0029]  Sinks will be clustered in groups of up to 4 and with maximum cluster diameter of 50.0 um.
```
precisionmoon, line 16: Why override a user-specified value without any warning?

src/cts/src/CtsOptions.h
```diff
@@ -192,6 +192,24 @@ class CtsOptions : public odb::dbBlockCallBackObj
   {
     sinkClusteringLevels_ = levels;
   }
+
```
precisionmoon, line 195: We need Tcl interface to exercise these new APIs.

src/cts/src/TreeBuilder.cpp
```diff
@@ -26,16 +26,28 @@ void TreeBuilder::mergeBlockages()
...
   for (odb::dbInst* inst : block->getInsts()) {
     if (inst->getMaster()->getType().isBlock()
         && inst->getPlacementStatus().isPlaced()) {
+      macros_max_dx = std::max(macros_max_dx, inst->getBBox()->getDX());
+      macros_max_dy = std::max(macros_max_dy, inst->getBBox()->getDY());
       blockage_polygons += inst->getBBox()->getBox();
     }
   }
 
+  // Set macros clustering diameter as 2 * macros highest dimention.
```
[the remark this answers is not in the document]
arthurjolo: We do not override the user specified macro cluster size and macro cluster diameter. For the cluster size it is defaulted as 4 and for the diameter we set it as 2 * macro size only if the diameter was not yet defined.

I am going to add a -macro_clustering_size and -macro_clustering_max_diameter so it is more direct for the user to specify these values.

src/cts/test/balance_levels.ok
arthurjolo: In my point of view we are not overriding the user specified value. The user specified these values for the register sinks, and for the macros none was specified so we used the default one. Should we have a warning that the values will not be same for macro and register tree?

precisionmoon: We can print separate messages for macro sinks vs. register sinks.   For example,
Register sinks will be clustered in groups of up to 5 and with maximum cluster diameter of 60.0.
Macro sinks will be clustered in groups of up to 4 and with maximum cluster diameter of 50.0.

arthurjolo: Yes that makes sense.

arthurjolo: Just a question, on the macro trees sinks we also include clock gaters. Should the message say "Macro and clock gater sinks"?

precisionmoon (approved):

Issue #6991, opened 2025-04-04
[ERROR DRT-0073] No access point for ces_0_0/io_ins_left[10].


oharboe:
### Describe the bug

![Image](https://github.com/user-attachments/assets/7409b7d3-cf64-4c63-a00f-c685ad696c62)

![Image](https://github.com/user-attachments/assets/e19e1703-397a-4589-99ed-3388982279a8)

### Expected Behavior

Should work

### Environment

```shell
asf
```

### To Reproduce

See above

### Relevant log output

```shell

```

maliberty:
Are they on the same layer?  The colors differ.

oharboe:
> Are they on the same layer? The colors differ.

I added one to a selection set, they are mirrored. I think they work everywhere else, or at least there are only two errors listed.

maliberty:
@bnmfw there is no need to do PA on abutted pins as there is nothing to route.   dbNet::isConnectedByAbutment can be used

maliberty:
grt calls pin access to get a better sense which gcell it should route to.

bnmfw:
I'm currently working on this task. The issue here is a bit deeper than only abutment. I'm currently working on a branch where any pin already routed by abutment will skip PA. However the error still persists.
Error 73 is a bit misleading, I think there is a case for changing it. The PA problem is not only with `ces_1_0/io_ins_left[10]` but actually any instance on its unique inst family, such as `ces_1_7`. The latter inst is not routed by abutment and PA can't find a valid access point for that pin as well (it should). I'm currently trying to understand why that is the case.

maliberty:
That constrait does prevent off-grid planar access so the PA problem is real.  It looks like this is an mpl placement issue (@joaomai @AcKoucher )

oharboe:
Interesting.

Could the error message be improved?

It was evidently incredibly tough to figure out that this is an mpl issue...

AcKoucher:
@oharboe #6738 should ensure that the alignment of the signal pins is correct.



> Could the error message be improved?

@joaomai It would be interesting to add a warning when we failed to align all the pins.
Perhaps something like "Couldn't align pins <pin_1> <pin_2> ... <pin_n> with the  track-grid."

maliberty:
If the lef58RightWayOnGridOnlyConstraint_ exists for the layer on which you can't align the pins then I think we should make it an error, otherwise a warning is fine.

[...]

PR #7178, opened 2025-04-18
drt: PA skips abutment

bnmfw:
Helps #6991.
Pins that are already connected by abutment qualify for pin access skip.
Also introduces `forceInstAsClassHead` which is a function made for purely debugging purposes on PA


src/drt/src/pa/FlexPA_unique.cpp
```diff
@@ -278,6 +278,24 @@ bool UniqueInsts::hasUnique(frInst* inst) const
...
+  }
+  for (frInst* inst : temporary_set) {
+    deleteInst(inst);
+  }
+  // target inst is now the only inst in the set, therefore its head
+  for (frInst* inst : temporary_set) {
+    addInst(inst);
+  }
+}
+
```
osamahammad21, line 298: I hate having unused debug code. DRT is already full of it and at some point, no one knows what it does. Could you either actually use it or keep it locally?
Also, better to have in a separate PR.

osamahammad21 (approved):
The connectedByAbutment changes are ok but the debug code needs taking care of.

Issue #7153, opened 2025-04-15
DRT: PA try_all_vias wrong order


bnmfw:
### Description

When generating access points different cost combinations can be tried, we go from best to worst. Also, when considering an access point, we first the first via on a vector, if those do not suffice we try every via.
In master, within a cost combination we the first via within a cost combination if that combination does yield any valid access point then we try every via. This ordering is wrong. We should first try the first via with all cost combinations, then try other vias.

bnmfw:
My comment on #7150 explains part of the motivation for this issue.
I have two main thoughts about this problem:

### Dummy Via for DRC checking
Do we have a dummy via for DRC checking? validating access for a via can be exhausting since we try every specific via. I think we should take the via with the lesser area or make a dummy via that is smaller in every dimension that all others. If the via cannot be used due to some DRC than the access point cannot be accessed through a via, we should not keep trying ad nauseam.

### Every Via should get their own access coords.
#7180 adds a base for this. Common access points (generated by Centered/Ongrid/HalfGrid/NearbyGrid coords) should try all vias. Enclosed access points should only try the via used to create it. I think there is an argument to run through vias on a layer, create custom access points and try and mount on them. This was my end goal with this issue

PR #7180, opened 2025-04-19
drt: genViaEnclosedCoord

bnmfw:
Supports #7153.

Isolates the piece of code that receives a Via, a Rectangle and a direction and inserts the PA coords onto the input map.
This is a very cheap and almost inconsequential refactor but I think it can be useful later down the line when trying to generate new access points for vias besides the first 2


maliberty (approved):

maliberty:
Needs a tidy but otherwise looks fine

PR #7183, opened 2025-04-21
bazel: ignore src/sta/build and debug folders

oharboe:
vscode with CMake plugins will create these dirs if you open src/sta and this will confuse Bazel.

Same as f3ec3697be2e3462bb684531f791838b49dc54c3 but for src/sta

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,2 +1,4 @@
 build/
 debug/
+src/sta/build/
+src/sta/debug/


oharboe:
@maliberty This must be an unrelated CI snag or master bug

maliberty (approved):

oharboe:
> I guess you are building sta manually.  The OR/ORFS build wouldn't create these by default.

These folders are the result of a needy visual code cmake plugin... Open the project and vscode creates these folders and spams me with a thousand dialogs.

.bazelignore is nice... I looked for a way to use .gitignore in bazel but didnt find anything.


I have uninstalled cmake vscode plugins. My instructions to this plugin are: go home, stay there and await my instructions.

I find that vscode is a bit of a disaster with large heterogenous projects: lots of needy plugins that fail to hit the nail on the head w.r.t. configurations and vscode scales poorly. Lots of spamming by needy plugins that try to be part of your family. This is made worse if you switch between computers.

It is still the least bad editor for my heterogenous use cases, but there are colleagues who ditched vscode.

maliberty:
vscode's security model is pretty terrible - see the series https://blog.extensiontotal.com/the-story-of-extensiontotal-how-we-hacked-the-vscode-marketplace-5c6e66a0e9d7

oharboe:
What is curious here is that a test runs before all .cpp files are built. Merits further investigation at some point if it doesn't go away.


image

PR #7190, opened 2025-04-22
grt: Only insert required number of diodes in repairAntennas

TheMightyDuckOfDoom:
DRAFT!

This change aims to rework the repair_antenna logic and the number of inserted Antenna-Diodes to a minimum.

Approach:
- [x] flow: Add number of diodes to metric limits
- [x] ant: Make number of diodes in violation structure absolut
- [x] grt/repairAntennas: Only insert up to a total number of diodes. Distribute them evenly among gates. Test results should be the same
- [ ] ant/checkGates: Rework antenna diode logic

Testing:
We should see the same number and severity of Antenna Violations initially, but if violations are fixed with a lower number of Diodes.

### DIFF
--- a/src/ant/include/ant/AntennaChecker.hh
+++ b/src/ant/include/ant/AntennaChecker.hh
@@ -103,7 +103,7 @@ struct Violation
 {
   int routing_level;
   std::vector<odb::dbITerm*> gates;
-  int diode_count_per_gate;
+  int diode_count;
   double excess_ratio;
 };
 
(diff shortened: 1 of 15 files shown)


maliberty:
Are you iterating in a mode where you need to have CI results?  Every time you push you are starting another CI run that take a lot of compute.

TheMightyDuckOfDoom:
Hi @maliberty ,

It was not my intention to hoard CI runtime andI will be more mindful with pushing changes.

Is there a way to disable CI runs for this PR (for the moment)?
I can also close it and open it again once it is ready.

Best,
Tobias

maliberty:
Closing it is the only method.

Issue #7150, opened 2025-04-15
DRT: Progress Oriented Pin Access


bnmfw:
### Description

Currently, Pin Access generates access points until it meets all soft requirements (or generates all possible and raises warning 88). The problem is that in this, it generates access points that do not further the progress to complete any of the soft requirements. PA should only consider the points that might make this progress. Currently the generation of these redundant access point is slowing down some designs

### Suggested Solution

Implement a function `accessPointMakesProgress()` which returns true if the inclusion of the access point on the set of already generated would further the completion of some requirement. If false, don't even check for DRCs, go onto to next.

bnmfw:
#7706 was created with some of the code needed to solve this issue. There is, however an earlier problem that I was not able to solve. To implement the proposed solution, once the access points are created, a single loop through all access points would need to be created. This refactor is **almost** possible. `filterMultipleAPAccesses` impedes it.

#### Deep Search and the non optimal way we treat vias in PA
Pin Access likes the first two vias a lot. The rest, not so much. When generating access coords only the first two vias get their custom enclosed coordinates. Not the rest. Every via tries to use the first two coordinates.

During via validation, something similar happens. For every access point that was created, we first look if the first two vias can be used on them. This would be the "shallow search". If **None** were successful in that group we go into deep search. Instead of trying only the first two vias we will go through every via. We early stop for some access point if at some point we have already two vias valid for it.

Refactoring this into a single loop is challenging. I've tried implementing a deep search through every access point separately. It did not work a while back but #7541 might have fixed it, so maybe we can try this again. I also ran a branch called drt_no_deep_search through secure CI in which I actually only do deep search from the get go (yeah bad name). It failed two designs, one by adding an antenna and other by adding an actual DRC error. 

Deep search impedes the unifying of a single access point loop. It probably is possible to implement progress oriented pin access without the single loop, but I did not try it.

PR #7196, opened 2025-04-22
drt: minor refactors

bnmfw:
Supports #7150 and #7153.

I'm currently working on a solution for the aforementioned issues on a separate branch. I ended up doing a lot minor changes and don't want pollute those with these minor changes, so I compiled them all here.

#### Better frAccess API
Just added a bunch of calls to the frAccess class to get Horizontal, Vertical an Planar accesses more directly.

#### frLayer isRoutable and isHorizontal
Direct `isRoutable` call to clean up it in some other parts of the code. `isHorizontal` and i`sVertical` already exist, I just used them when appropriate.

#### UpperLayerOnlyAllowOnGridAccess
Condensed a bunch a logic that essentially says if the layer above the input layer will only allow onGrid access. Improves readability on PA.

#### `createMultipleAccessPoints` is responsible for determining if planar and via access is allowed.
This information used to be plumbed through 2 functions even though it can be determine at the `createMultipleAccesspPoints` function, so I just moved it there. (Spoiler for the next PR: `createMultipleAccessPoints` will be moved out of `genAPsFromPinShapes`)

#### `is_macro_cell_pin && use_center_line` abreviation
`use_center_line` by definition is only true if `is_macro_cell_pin` is true, so checking just for `use_center_line` is enough.

#### Layer Shape iteration
Iterate over pin shapes with with the way and in crescent order, instead of using the iterator directly and going backwards (the change in direction has no impact)

#### `updatePinStats` and `pin_access_idx`
`updatePinStats()` can now receive an IO pin. `pin_access_idx` can be calculated for IO pins as well. Now the end of `genPinAccessCostBounded` does not require any ifs.


src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -434,74 +447,66 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
     }
   }
 
-  if (is_macro_cell_pin && use_center_line && is_layer1_horz) {
-    lower_type = frAccessPointEnum::OnGrid;
+  if (use_center_line && is_layer1_horz) {
+    for (auto& [layer1_coord, cost] : layer1_coords) {
+      layer1_coords[layer1_coord] = frAccessPointEnum::OnGrid;
+    }
```
osamahammad21, line 453: @bnmfw how is this equivalent?
This won't produce the same behavior for this check:
https://github.com/bnmfw/OpenROAD_Bernardo_Sandoval/blob/c0f5e592edb87dd99f80d17e14be4f74fe338b60/src/drt/src/pa/FlexPA_acc_point.cpp#L337
Also this for loop is duplicate to the else block of the previous if condition.

src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -514,33 +519,28 @@ void FlexPA::genAPsFromLayerShapes(
...
                             inst_term,
-                            *it,
+                            layer_shapes,
                             layer_num,
-                            allow_via,
                             lower_type,
                             upper_type);
     }
-    layer_num--;
+    layer_num++;
```
osamahammad21, line 543: I think the idea is to prefer upper layer access. I am not sure but I think the ordering of the Access points matters elsewhere in DRT. Are you strongly confident that this won't haunt us in the future?

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -434,74 +447,66 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
...
-                             allow_planar,
-                             allow_via,
-                             is_layer1_horz,
                              x_coords,
                              y_coords,
                              lower_type,
                              upper_type);
 }
 
+bool FlexPA::UpperLayerOnlyAllowOnGridAccess(const frLayerNum layer_num,
```
bnmfw, line 467:
```suggestion
bool FlexPA::isUpperLayerOnGridOnly(const frLayerNum layer_num,
```

src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -434,74 +447,66 @@ void FlexPA::genAPsFromRect(std::vector<std::unique_ptr<frAccessPoint>>& aps,
...
-  const auto upper_layer
-      = (layer_num + 2 <= getDesign()->getTech()->getTopLayerNum())
-            ? getDesign()->getTech()->getLayer(layer_num + 2)
-            : nullptr;
-  if (!is_macro_cell_pin && upper_layer
-      && upper_layer->getLef58RightWayOnGridOnlyConstraint()
+  // IO term is treated as the MacroCellPin as the top block
+  bool is_macro_cell_pin = inst_term ? isMacroCell(inst_term->getInst()) : true;
+
+  if (UpperLayerOnlyAllowOnGridAccess(layer_num, is_macro_cell_pin)
```
bnmfw, line 498:
```suggestion
  if (isUpperLayerOnGridOnly(layer_num, is_macro_cell_pin)
```

src/drt/src/pa/FlexPA_acc_point.cpp
bnmfw: The ordering matters? That was something I was going to actively change (make these APs exist in a set). We will go through them all no matter what, so the only thing changing here is the order. If it does matter then yeah, it could be a problem.

bnmfw:
(Message about the commit above)

#### NearbyGrid Path seg elimination
The code almost folded onto itself as I was cleaning this part.
The first bunch of ifs and elses code is just a **clamp** operation. `end` is just `ftp` but if it is not withing the rectangle we snap it to it.
This made me question what `e` is. It essentially is `ftp`, but if `end` differs from `ftp` then it is just `end`. This is just a convoluted way of saying `e = end` always.
When changing every instance of `e` to `end` we get the `if (end != end)` contradiction, which is always false, so I just deleted it.

Issue #6717, opened 2025-02-14
mpl: the macro arrays in ngt45/swerv_wrapper look odd


AcKoucher:
### Description

The macros arrays are being grouped differently than not so long ago and it looks like it could be causing timing and WL degradation.

### Suggested Solution

Improve the evaluation of similar connection signature between clusters.

PR #7198, opened 2025-04-22
mpl: hard macro partitioning

joaomai:
This PR was motivated by issue #6717 and aims to move the responsibility of partitioning hard macros from PAR to MPL. This hopefully makes MPL results more stable and predictable by relying less on PAR results.

The implementation assigns hard macros to PAR created clusters based on connectivity with said clusters.


src/mpl/src/clusterEngine.cpp
```diff
@@ -1547,6 +1547,60 @@ void ClusteringEngine::breakLargeFlatCluster(Cluster* parent)
...
+    }
+  }
+
+  std::map<odb::dbInst*, int> macro_to_iterms_count;
+  for (odb::dbNet* net : non_supply_nets) {
+    for (odb::dbITerm* iterm : net->getITerms()) {
+      odb::dbInst* inst = iterm->getInst();
+      odb::dbMaster* master = inst->getMaster();
+
+      if (master->isPad() || master->isCover() || master->isEndCap()) {
```
AcKoucher, line 1591: This should check for `isIgnoredInst`

AcKoucher (changes requested):
With this approach couldn't we end up putting all the macros in the same partition?
I think we need to respect the current level's max threshold.

PR #7200, opened 2025-04-23
Grt: Add the placement blockage geometries to avoid insert diodes within those regions

luis201420:
Add the hard placement blockage geometries along with the fixed instances in `r_tree`.
This to prevent diode placement within those regions.


luis201420:
> @luis201420 Did you run a CI for this change?

Yes, I ran it and all public designs passed without errors.

PR #7205, opened 2025-04-23
test: est_rc4.tcl was using src folder for output files

oharboe:


maliberty (approved):
`-spef_file results/est_rc4.spef` doesn't look like it was using the src dir.

oharboe:
> `-spef_file results/est_rc4.spef` doesn't look like it was using the src dir.

It was in a bazel context. Our solution in the past has been to use the make_result_file idiom which works in both bazel and cmake case.

I didn't hunt for residual use-cases, I just fixed one problematic case.

PR #7206, opened 2025-04-23
Orfs test case - basis for further collaboration and discussion

oharboe:
These tests are disabled for now because there's CI infrastructure updates that are needed and also various dominos in OpenROAD that have to be toppled(performance problems such as omp not working in bazel and various other things).

However, merging now will set direction and simplify collaboration and discussion.

Use case is to update fast local tests for OpenROAD to include some important ORFS integration tests.

1. Modify OpenROAD source code as usual
2. Run `bazel test -c opt ...`, this will build openroad and run all tests in a few minutes.

Tests that are too long running or not yet ready to be included in this fast local test case, have the `tags = ["manual"]` set such that they have to be invoked explicitly. The ORFS tests currently go one step further and add the test/orfs/ folder to .bazelignore to work around CI infrastructure limitations.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -2,3 +2,5 @@ build/
 debug/
 src/sta/build/
 src/sta/debug/
+# delete line below when CI can handle it and it runs in minutes
+test/orfs/
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -98,3 +98,25 @@ pip.parse(
     requirements_lock = "//bazel:requirements_lock_3_13.txt",
 )
 use_repo(pip, "openroad-pip")
+
+bazel_dep(name = "bazel-orfs")
+
+# To bump version, run: bazelisk run @bazel-orfs//:bump
+git_override(
+    module_name = "bazel-orfs",
+    commit = "afa5638555b82222d3e2f20aa29c7c7c05ad87c5",
+    remote = "https://github.com/The-OpenROAD-Project/bazel-orfs.git",
+)
+
+orfs = use_extension("@bazel-orfs//:extension.bzl", "orfs_repositories")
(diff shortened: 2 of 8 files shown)


oharboe:
> Is the direction to copy over orfs test cases? I thought the goal was to run them from orfs.

I want to have a few key unique ORFS test cases in OpenROAD that run in minutes and gives you a pretty good indication that a PR to ORFS is good to go and not just a waste of build resources and futile 2 hour wait.

maliberty:
That's what the existing flow tests try to do.  Is the difference here which flow scripts are used?

oharboe:
> That's what the existing flow tests try to do. Is the difference here which flow scripts are used?

My intent is to add a few key bazel-orfs tests to the OpenROAD so that the feedback loop is closed for the developer on OpenROAD in minutes locally with minimum cognitive load. Testing locally with ORFS is also possible, but I expect that to be more of a "test my PR" use-case realistically.

oharboe:
Now with rules-base.json testing:

```

oharboe:
I'm thinking the use-case for bazel-orfs tests in OpenROAD is fast turnaround local testing+CI of key bazel-orfs/ORFS use-cases.

The main use-case for ORFS bazel-orfs testing is local bazel-orfs+ORFS testing and no CI testing to PRs to bazel-orfs that have to go along with PRs in scripts in ORFS.

So I expect number of ORFS bazel-orfs tests to be reduced and ORFS to be mainly makefile based testing as before.

Until and unless the preference shifts to bazel-orfs over makefile based articulation in ORFS testcases for local and CI testing.

maliberty (approved):

PR #7214, opened 2025-04-23
gui: background color support

### LINKED ISSUE #7191: Add an option to change the GUI background color
### Description

I stumbled across #6214 while looking for the exact same feature. I would like to set the GUI background to white before saving an image. Perhaps it would even be possible to set the background to transparent when saving as png?

It would be great if this could be implemented.

### Suggested Solution

_No response_

### Additional Context

_No response_

gadfort:
Closes #7191 

Adds:
- control of background color to the display controls (as Misc/Background)
- add color option to `gui::set_display_controls` to support common colors in openroad or as `#rrbbggaa`


maliberty (approved):
Please fix tidy and then you can merge

PR #7216, opened 2025-04-24
mpl: remove shrink & restart mechanism

AcKoucher:
I believe that the rationale of this mechanism was attempting faster convergence for large designs with high utilization. However, apparently, most of the convergence problems that MPL had was due to bugs associated with the lack of a mechanism to store the valid results generated along the simulated annealing steps - which we now have.

_I tested MPL using a very large private design with 80% utilization and it properly converges._

Also, currently, as use batches of annealing (10) to check for convergence and lowest cost, we may choose a convergent annealing run that had it's std cell area shrunk over a convergent annealing run with its actual chosen utilization (we decrease the utilization as runs go on). MPL wise this is not a problem, but, we'll be generating a macro placement for a very different placement density that GPL will use for global placement.

In investigating whether or not we should actually be comparing costs. That's why this is a draft.


AcKoucher:
Secure-CI passed.

eder-matheus (approved):

PR #7217, opened 2025-04-24
ppl: assign pins to sections with more slots available in case of tie

eder-matheus:


src/ppl/src/HungarianMatching.cpp
```diff
@@ -397,7 +397,7 @@ std::vector<uint8_t> HungarianMatching::getTieBreakRank(
 {
   std::vector<uint8_t> rank(costs.size());
   uint8_t ranking = 1;
-  for (int i : sortIndexes(costs, costs)) {
+  for (int i : sortIndexes(costs, costs, costs)) {
```
maliberty, line 400: It seems odd to have tie breaking code that won't see any difference in the tie breaking values.

eder-matheus: In this specific case, I want only the regular sorting, as we don't have any tie break criteria here. Perhaps I should add an overload to this function for clarity?

maliberty: I think that would help

maliberty (approved):
I'll leave this in the approved state as the change is minor and I'll be out.

PR #7218, opened 2025-04-24
gpl: allow instance removal

gudeh:
This PR introduces no functional changes until RSZ starts triggering destruction callbacks.

It enables the removal of instances during GPL execution by responding to such callbacks. RSZ is expected to trigger destruction callbacks in the future; it already does so for insertions.

Destructors were already defined in the code but had not been properly implemented.

Both the destructors and the previously existing creators for the callbacks assume that the NesterovBase::fixPointers() method will be called afterward.

This implementation takes into account RSZ calls during timing-driven, non-virtual GPL executions. All instance creations and destructions are handled through callbacks, which dynamically modify the vector storages for GCell, GPin, and GNet objects in both NesterovBaseCommon and NesterovBase. Since these modifications can invalidate pointers, fixPointers() is invoked after each timing-driven iteration by RSZ to refresh them.

To maintain consistency, we preserve the ordering of parallel vectors in NesterovBase, following the layout of nb_gcells_. For deletions, we use the swap-and-pop method: the element to be removed is swapped with the last one in the vector, and references are updated to point to the moved (previously last) element before it is removed.

There are three main challenges involved in supporting instance destruction during execution:
1. Maintaining pointer validity for elements referenced in vectors.
2. Preserving the order among parallel vectors.
3. Managing the nb_gcells_ vector in NesterovBase, which holds pointers to GCell objects stored across both NesterovBase (filler cells) and NesterovBaseCommon (instance cells).


gudeh:
I couldn't rebase to edit the missing signoff on the first commit (a revert) because src/dpo was deleted and git is getting confused during rebase. I will create a new branch and new PR.

Issue #6908, opened 2025-03-25
Global place with routability drops solution if it fails to perform routability


gadfort:
### Describe the bug

When running global place with routability it eventually fails with:
```
[INFO GPL-0087] FinalRC: 1.0801135
[INFO GPL-0079] MinRC (1.079737) violation occurred, total count: 3.
[INFO GPL-0045] InflatedAreaDelta:      411.943 um^2 (+0.25%)
[INFO GPL-0046] TargetDensity:            0.189
Revert Routability Procedure. Target density higher than max, or minRC max violations.
[INFO GPL-0080] minRcViolatedCnt: 3
[INFO GPL-0047] SavedMinRC: 1.0797
[INFO GPL-0048] SavedTargetDensity: 0.1875
[INFO GPL-0089] Routability: revert back to snapshot
```
and continues to run global place but without any of the interim solutions for rouability eventually resulting in a design that cannot be routed

### Expected Behavior

Either:
1. continues to attempt routabilty until it works
2. preserve the last solution even if its not perfect since it's probably better than nothing

### Environment

```shell
OpenROAD v2.0-19622-g9e747a8bc 
Features included (+) or not (-): +GPU +GUI +Python
```

### To Reproduce

[sc_issue_darkriscv_job0_ihp130_sg13g2_stdcell_place.global0_20250325-134902.tar.gz](https://github.com/user-attachments/files/19453869/sc_issue_darkriscv_job0_ihp130_sg13g2_stdcell_place.global0_20250325-134902.tar.gz)

```
tar xvf sc_issue_darkriscv_job0_ihp130_sg13g2_stdcell_place.global0_20250325-134902.tar.gz
cd sc_issue_darkriscv_job0_ihp130_sg13g2_stdcell_place
./run.sh
```

### Relevant log output

```shell
[NesterovSolve] Iter:  860 overflow: 0.329 HPWL: 1094242171
[INFO GPL-0075] Routability iteration: 8
[INFO GPL-0036] TileBBox: (    0    0 ) ( 7200 7200 ) DBU
[INFO GPL-0038] TileCnt:     180  180
[INFO GPL-0040] NumTiles: 32400
[INFO GPL-0081] TotalRouteOverflow: 63.8768
[INFO GPL-0082] OverflowTileCnt: 4306
[INFO GPL-0083] 0.5%RC: 1.0888
[INFO GPL-0084] 1.0%RC: 1.0714
[...]

PR #7223, opened 2025-04-25
gpl: allow instance removal

gudeh:
This PR introduces no functional changes until RSZ starts triggering destruction callbacks.

It enables the removal of instances during GPL execution by responding to such callbacks. RSZ is expected to trigger destruction callbacks in the future; it already does so for insertions.

Destructors were already defined in the code but had not been properly implemented.

Both the destructors and the previously existing creators for the callbacks assume that the NesterovBase::fixPointers() method will be called afterward.

This implementation takes into account RSZ calls during timing-driven, non-virtual GPL executions. All instance creations and destructions are handled through callbacks, which dynamically modify the vector storages for GCell, GPin, and GNet objects in both NesterovBaseCommon and NesterovBase. Since these modifications can invalidate pointers, fixPointers() is invoked after each timing-driven iteration by RSZ to refresh them.

To maintain consistency, we preserve the ordering of parallel vectors in NesterovBase, following the layout of nb_gcells_. For deletions, we use the swap-and-pop method: the element to be removed is swapped with the last one in the vector, and references are updated to point to the moved (previously last) element before it is removed.

There are three main challenges involved in supporting instance destruction during execution:

1. Maintaining pointer validity for elements referenced in vectors.
2. Preserving the order among parallel vectors.
3. Managing the nb_gcells_ vector in NesterovBase, which holds pointers to GCell objects stored across both NesterovBase (filler cells) and NesterovBaseCommon (instance cells). A GCellHandle was previously implemented for that.


gudeh:
I ran a test-CI and everything passed: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/test-gpl-instances-removal/5/pipeline

gudeh:
I understand there is no rush to merge this since rsz is under work. But it makes it easier for @povik to test rsz behavior with gpl.

src/gpl/src/nesterovBase.cpp
```diff
@@ -2829,15 +2868,17 @@ bool NesterovBase::revertToSnapshot()
 
 void NesterovBaseCommon::moveGCell(odb::dbInst* db_inst)
 {
-  GCell* gcell = getGCellByIndex(db_inst_map_.find(db_inst)->second);
+  GCell* gcell
+      = getGCellByIndex(db_inst_to_nbc_index_map_.find(db_inst)->second);
```
eder-matheus, line 2872: I believe you should check for the return value of `find` to ensure the key exists in the map. If it's guaranteed to exists, you could use the `[]` operator.

src/gpl/src/nesterovBase.cpp
```diff
@@ -2829,15 +2868,17 @@ bool NesterovBase::revertToSnapshot()
...
   odb::dbBox* bbox = db_inst->getBBox();
   gcell->setAllLocations(
       bbox->xMin(), bbox->yMin(), bbox->xMax(), bbox->yMax());
 }
 
 void NesterovBaseCommon::resizeGCell(odb::dbInst* db_inst)
 {
-  GCell* gcell = getGCellByIndex(db_inst_map_.find(db_inst)->second);
+  GCell* gcell
+      = getGCellByIndex(db_inst_to_nbc_index_map_.find(db_inst)->second);
```
eder-matheus, line 2881: Same comment here about using `find` vs `[]`.

src/gpl/src/nesterovBase.cpp
```diff
@@ -3003,69 +3053,136 @@ size_t NesterovBaseCommon::createCbkGCell(odb::dbInst* db_inst)
...
+    if (gcell_index != last_index) {
+      if (!nb_gcells_[gcell_index]->isFiller()) {
+        odb::dbInst* swapped_inst
+            = nb_gcells_[gcell_index]->insts()[0]->dbInst();
+        db_inst_to_nb_index_map_.erase(swapped_inst);
+        db_inst_to_nb_index_map_[swapped_inst] = gcell_index;
+      }
+    } else {
+      db_inst_to_nb_index_map_.erase(db_inst);
+    }
```
eder-matheus, line 3108: This part of the code is a bit confuse. Could you add some comments on the code for clarity?

I'm not following what are the `gcell_index` and the `last_index` and what they are doing in the `db_inst_to_nb_index_map_` map.

src/gpl/src/nesterovBase.cpp
```diff
@@ -3003,69 +3053,136 @@ size_t NesterovBaseCommon::createCbkGCell(odb::dbInst* db_inst)
...
-  size_t last_index = gCellStor_.size() - 1;
-  if (index_remove > last_index) {
-    return;
+  auto it = db_inst_to_nbc_index_map_.find(db_inst);
+  if (it == db_inst_to_nbc_index_map_.end()) {
+    log_->error(GPL,
+                307,
+                "db_inst not found in db_inst_to_NBC_index_map_ when trying to "
+                "destroy GCell on NBC");
+    return {};
```
eder-matheus, line 3155: You don't need this return here. The `Logger::error` should be enough to stop the execution of this function.

src/gpl/src/nesterovBase.cpp
```diff
@@ -3003,69 +3053,136 @@ size_t NesterovBaseCommon::createCbkGCell(odb::dbInst* db_inst)
...
   }
+
+  size_t index_remove = it->second;
+  size_t last_index = gCellStor_.size() - 1;
+
+  db_inst_to_nbc_index_map_.erase(it);
+
+  std::pair<odb::dbInst*, size_t> replacement{};
```
eder-matheus, line 3163: Are the `{}` necessary in the variable declaration?

src/gpl/src/nesterovBase.cpp
```diff
@@ -3003,69 +3053,136 @@ size_t NesterovBaseCommon::createCbkGCell(odb::dbInst* db_inst)
...
-  gCellStor_.pop_back();
 
-  // TODO: update deltaArea_ after cell destruction.
+  gCellStor_.pop_back();
+  return replacement;
 }
 
 void NesterovBase::destroyFillerGCell(size_t index_remove)
 {
+  log_->report("Warning, deleting filler gcell!");
```
eder-matheus, line 3180: Move to a `Logger::warn` or debug print.

src/gpl/src/nesterovBase.cpp
```diff
@@ -3003,69 +3053,136 @@ size_t NesterovBaseCommon::createCbkGCell(odb::dbInst* db_inst)
...
   size_t last_index = fillerStor_.size() - 1;
   if (index_remove > last_index) {
+    log_->report("error: index {} out of bounds for fillerStor_ (max:{})",
```
eder-matheus, line 3183: `Logger::error`?

eder-matheus (changes requested):

src/gpl/src/nesterovBase.cpp
gudeh: I still need to implement this to try and use it whenever we change the area of the design (we currently increase density instead of removing fillers to compensate for more space). This function is expected to never be called, currently.

eder-matheus: I understand that it will not be used now, but I'll like to avoid leave pending code to be fixed.

gudeh: I only commented it out for now, I am already implementing the removal of fillers. I think it makes more sense for us to do another PR for the removal of fillers.

eder-matheus (approved):

gudeh:
The discussion of removing fillers was already implemented and seems to be working, some preliminary results are shown here: https://github.com/The-OpenROAD-Project/OpenROAD/issues/6908#issuecomment-2844723932

src/gpl/src/nesterovBase.cpp
```diff
@@ -3126,15 +3327,38 @@ void NesterovBase::swapAndPopParallelVectors(size_t remove_index,
...
+    return;
+  }
+
+  out << "gCellStor_.size(): " << gCellStor_.size() << "\n";
+  out.close();  // reuse printToFile which appends
+
+  for (size_t i = 0; i < gCellStor_.size(); ++i) {
+    std::ofstream out_idx(filename, std::ios::app);
+    out_idx << fmt::format("idx:{}\n", i);
+    out_idx.close();
```
maliberty, line 3360: Why are you opening and closing this file on every gcell?

src/gpl/src/nesterovBase.cpp
```diff
@@ -3003,102 +3072,226 @@ size_t NesterovBaseCommon::createCbkGCell(odb::dbInst* db_inst)
...
                               size_t last_index)
 {
-  if (last_index != vec.size() - 1) {
+  if (vec.empty()) {
+    debugPrint(
+        log_, GPL, "callbacks", 1, "Warn Attempted to pop from empty vector.");
+    return;
+  }
+
+  if (remove_index >= vec.size()) {
```
maliberty, line 3287: Is this debug code?  Do we expect these things to happen?

src/gpl/src/nesterovBase.cpp
```diff
@@ -3003,102 +3072,226 @@ size_t NesterovBaseCommon::createCbkGCell(odb::dbInst* db_inst)
...
+void NesterovBaseCommon::destroyCbkGNet(odb::dbNet* db_net)
+{
+  debugPrint(log_, GPL, "callbacks", 2, "NesterovBaseCommon::destroyGNet");
+  auto db_it = db_net_to_index_map_.find(db_net);
+  if (db_it == db_net_to_index_map_.end()) {
+    log_->error(GPL,
+                308,
+                "db_net not found in db_net_to_NBC_index_map_ for net: {}",
+                db_net->getName());
+    return;
```
maliberty, line 3217: no need to return after error as it throws an exception (same elsewhere)

src/gpl/src/nesterovBase.cpp
gudeh: This is an unexpected condition, just to be safe.

maliberty: Would the results be correct if this happened?

src/gpl/src/nesterovBase.cpp
```diff
@@ -3097,17 +3116,13 @@ void NesterovBase::destroyCbkGCell(odb::dbInst* db_inst)
...
         odb::dbInst* swapped_inst
             = nb_gcells_[gcell_index]->insts()[0]->dbInst();
         db_inst_to_nb_index_map_.erase(swapped_inst);
         db_inst_to_nb_index_map_[swapped_inst] = gcell_index;
       }
-    } else {
-      db_inst_to_nb_index_map_.erase(db_inst);
-    }
 
-    if (!is_filler) {
```
povik: @gudeh Was this a mistake? The condition for the code below changed from `!is_filler` to `gcell_index != last_index`

gudeh: This part was fixed in PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/7362

PR #7231, opened 2025-04-28
Cts add user clock buffer prefix for list inferring

arthurjolo:
Add the option for user to define the clock buffer name that CTS will look for when inferring the buffer list (e.g. "CLKBUF"). If no cells are found using the user's name it will proceed with the normal inferring procedure.  This helps users which PDKs don't follow the default clock buffer name defined in CTS.

### DIFF
--- a/src/cts/README.md
+++ b/src/cts/README.md
@@ -44,6 +44,7 @@ Perform clock tree synthesis.
 clock_tree_synthesis 
     [-wire_unit wire_unit]
     [-buf_list <list_of_buffers>]
+    [-buf_list_prefix prefix]\
     [-root_buf root_buf]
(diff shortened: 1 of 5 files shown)


gadfort:
Why not use `[get_lib_cells CLKBUF*]` as the input to the buf_list?

arthurjolo:
> Why not use `[get_lib_cells CLKBUF*]` as the input to the buf_list?

The main reason to add this command is that when CTS does the buffer list inference it sometimes will fail to identify the clock specific buffers. And if you use -buf_list CTS can be slowed down if the list is too big. So this command would help guide the buffer inference and it would take advantage of the buffer list trimming. But i believe this will not be added to production, we need to improve how the buffer inference works.

PR #7232, opened 2025-04-28
tap: improve error message for autoselection of cells with more than one option

eder-matheus:


src/tap/test/endcap_autoselection_error1.ok
```diff
@@ -0,0 +1,10 @@
+[INFO ODB-0227] LEF file: Nangate45/Nangate45_tech.lef, created 22 layers, 27 vias
+[INFO ODB-0227] LEF file: Nangate45/Nangate45_stdcell.lef, created 135 library cells
+[INFO ODB-0227] LEF file: Nangate45_data/endcaps.lef, created 13 library cells
+[INFO ODB-0227] LEF file: Nangate45_data/endcaps_symmetric.lef, created 5 library cells
+[INFO ODB-0227] LEF file: Nangate45/fakeram45_64x7.lef, created 1 library cells
+[INFO ODB-0128] Design: gcd
+[INFO ODB-0131]     Created 5 components and 160 component-terminals.
+[INFO ODB-0303] The initial 479 rows (2081802 sites) were cut with 5 shapes for a total of 699 rows (2014702 sites).
+[ERROR TAP-0104] Unable to find a single master for ENDCAP RIGHTEDGE. Choose one between the following cells: ENDCAP_X4_RIGHTEDGE_Y ENDCAP_X4_RIGHTEDGE.
```
eder-matheus, line 9: @jeffng-or This is an example of how the new error message would look. Do you have any comments/suggestions/requests regarding this message?

jeffng-or: How about "Found multiple masters for ENDCAP RIGHTEDGE. Use -whatever_the_option_is to specify one of the following cells: ..."?

That way, the message more accurately describes the situation and we are directing the user on how to fix it. The only issue might be that the check is in a central spot and we might not know which of the two commands was used (e.g. tapcell or place_endcaps)

eder-matheus: I found that detecting which of the two commands was used is not the only issue. Which option inside each command is hard too, because the option names don't match between `tapcell` and `place_endcaps`.

Making them match is a project itself, since it would require changes in both OR and ORFS.

src/tap/test/endcap_autoselection_error2.ok
```diff
@@ -0,0 +1,9 @@
+[INFO ODB-0227] LEF file: Nangate45/Nangate45_tech.lef, created 22 layers, 27 vias
+[INFO ODB-0227] LEF file: Nangate45/Nangate45_stdcell.lef, created 135 library cells
+[INFO ODB-0227] LEF file: Nangate45_data/endcaps.lef, created 13 library cells
+[INFO ODB-0227] LEF file: Nangate45_data/endcaps_symmetric.lef, created 5 library cells
+[INFO ODB-0227] LEF file: Nangate45/fakeram45_64x7.lef, created 1 library cells
+[INFO ODB-0128] Design: gcd
+[INFO ODB-0131]     Created 5 components and 160 component-terminals.
+[ERROR TAP-0104] Found multiple masters for ENDCAP RIGHTEDGE. Use -right_edge/-endcap to specify one of the following cells:  ENDCAP_X4_RIGHTEDGE_Y ENDCAP_X4_RIGHTEDGE
```
eder-matheus, line 8: @jeffng-or I was able to implement a solution to add the command and the option in the message.

precisionmoon (approved):

PR #7233, opened 2025-04-28
gitignore: ignore bazel & Scala stuff

oharboe:
make bazel-* ignore pattern narrower to clean up local goop more regularly. bazel-* folders can accrue in local `make issue` folders.

gitignore should:

- include all files that should exist locally, like IDE files, but should not be committed
- not include local goop that should be cleaned up regularly

git status should list:

- changes to be committed new or exisiting files
- files that should be deleted locally to keep the source folder nice and short and make sure that use-cases like IDE and "grep -r" remain nice and fast.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,4 +54,12 @@ debian/openroad
 obj-x86_64-linux-gnu/
 
 user.bazelrc
-bazel-*
+
+bazel-bin
+bazel-out
+bazel-testlogs
+bazel-OpenROAD
+
+projectview.bazelproject
+.bsp/
+.bazelbsp/


oharboe:
@vvbandeira @maliberty unrelated CI error

oharboe:
@vvbandeira @maliberty How are we doing w.r.t. getting some more reasonable build resources online for Bazel?

A reasonable machine with caching intact between builds should build bazel in minutes, not hours.

oharboe:
@vvbandeira @maliberty Just FYI w.r.t. expectations for Bazel caching there are three important parts to bazel caching:

1. building all dependencies. This is almost never done, except when MODULE.bazel changes and then only for the changed bits of MODULE.bazel
2. if we modify some part of OpenROAD that causes a full rebuild of OpenROAD that can take a while
3. if OpenROAD binary changes(this excludes git commit ID as there is no git commit ID or build date in OpenROAD), then tests have to be re-run.

Mostly tests in OpenROAD are based on the openroad binary so have to be rerun whenever the openroad binary changes, however, some unit-tests may be linked only with the relevant changes and might have narrower dependencies.

maliberty (approved):

PR #7235, opened 2025-04-29
bazel: fix python toolchain usage

oharboe:


maliberty (approved):

maliberty:
It sounds fine but fails in the CI.

hzeller (approved):
LGTM
Looks like the build-failure is due to eigen lib could not be downloaded temporarily. So restarting the failed CI should fix it.

PR #7236, opened 2025-04-29
gpl: debug generate images

gudeh:
This is a new feature for gpl debug mode. It does not affect functionality.

If the `-generate_images` is used in the debug mode command (for example including `global_placement_debug -generate_images` in `flow/scripts/global_placement.tcl`, before the normal global_placement command), it enables the creation of informative images. Images are saved in ORFS' reports path from env var.

It will create a gif animation with the placement procedure:
placement

And also a set of images showing the placement state at the end of every routability and timing driven iterations, including heatmaps for routability mode:
image


src/gpl/src/graphics.cpp
```diff
@@ -508,4 +508,57 @@ bool Graphics::guiActive()
...
+  std::string cmd = fmt::format(
+      "convert {} -resize 50% -colors 64 -strip -quality 85 "
+      "-gravity SouthEast -pointsize 20 -fill {} "
+      "-annotate +5+5 '{}' PNG8:{}",
+      input_path,
+      fill_color,
+      label,
+      output_path);
+
+  int ret = std::system(cmd);
```
gadfort, line 534: not a huge fan of calls to `system`, seems like the GUI could be augmented to support both the image and gif generation. Separate from this PR, but might be a nice thing to support anyways.

maliberty: Perhaps https://github.com/charlietangora/gif-h - single header library that we could put in third-party

gudeh:
I intend to add this TCL command to the timing driven images, which selects rsz buffers: `select -type Inst -filter {Description=Timing Repair Buffer} `

gudeh:
Here is an example of what gets generated: https://drive.google.com/file/d/1E80n-clONg2bjSmRfNUb-AQ-WQxPbVr4/view?usp=sharing

gadfort:
> The CI reported some errors on the non-GUI build, how should I fix it?

You'll need to add the missing definitions to `src/gui/src/stub.cpp`

src/gpl/src/graphics.cpp
```diff
@@ -534,4 +534,53 @@ bool Graphics::guiActive()
...
+                             const std::string& label,
+                             const std::string& label_name,
+                             int image_width_px)
+{
+  int label_x = bbox.xMin() + 300;
+  int label_y = bbox.yMin() + 300;
+
+  gui::Painter::Color color = gui::Painter::stringToColor("yellow", logger_);
+  gui::Painter::Anchor anchor
+      = gui::Painter::stringToAnchor("bottom left", logger_);
```
gadfort, line 548: you could just reference then as `gui::Painter::yellow` and `gui::Painter::BOTTOM_LEFT` that avoid the string lookup which is only needed in the tcl interface

maliberty (approved):

Issue #5975, opened 2024-10-17
drt; no access point issue


titan73:
### Describe the bug

No access point and routing fails.

### Expected Behavior

Find access points for 2 cell pins and route.

### Environment

```shell
v2.0-16472-gc7283c089
```

### To Reproduce

Test case with 1 buffer and 2 ports:
[no_access_point.zip](https://github.com/user-attachments/files/17412706/no_access_point.zip)

### Relevant log output

```shell
[INFO DRT-0167] List of default vias:
  Layer V1
    default via: M2_M1_VH_via
[INFO DRT-0162] Library cell analysis.
[INFO DRT-0163] Instance analysis.
[INFO DRT-0164] Number of unique instances = 1.
[INFO DRT-0168] Init region query.
[INFO DRT-0024]   Complete FR_MASTERSLICE.
[INFO DRT-0024]   Complete Fr_VIA.
[INFO DRT-0024]   Complete M1.
[INFO DRT-0024]   Complete V1.
[INFO DRT-0024]   Complete M2.
[INFO DRT-0033] FR_MASTERSLICE shape region query size = 0.
[INFO DRT-0033] FR_VIA shape region query size = 0.
[INFO DRT-0033] M1 shape region query size = 15.
[INFO DRT-0033] V1 shape region query size = 0.
[INFO DRT-0033] M2 shape region query size = 2.
[INFO DRT-0165] Start pin access.
[ERROR DRT-0073] No access point for instance/A.
Error: test.tcl, 21 DRT-0073
```

maliberty:
@bnmfw you might start with the debug graphics to see what is being tried: Add `detailed_route_debug -pa -pa_markers -pin instance:A` and run with the GUI.

titan73:
@bnmfw Any news on this issue?

jeffng-or:
FYI. Also seeing this with MegaBoom when M3/M4 routing layer adjustment is set to 0.3.
```
[INFO DRT-0165] Start pin access.
[INFO DRT-0076]   Complete 1000 pins.
[INFO DRT-0076]   Complete 2000 pins.
[ERROR DRT-0073] No access point for frontend/bpd/banked_predictors_0/btb/btb_0_ext/R0_addr[0].
[ERROR DRT-0073] No access point for frontend/bpd/banked_predictors_0/btb/btb_1_ext/R0_addr[0].
[...]

PR #7239, opened 2025-04-29
odb/lef: ignore bad MINIMUMCUT

VoltXd:
Also reallocate memory in case of rule deletion.
Prints a warning when a line is ignored, using code 2504. Is this code OK to use ? How can I know which one I can use, is there a rule (beside that 2500 is for lef warning) ?


AcKoucher:
@VoltXd Regarding the message ID, I think you can rely on `lef.msg` file to see the available IDs.

VoltXd:
Is it expected that "**Read the Docs build**" fails on `npm install -g @mermaid-js/mermaid-cli` ? Should I do anything to correct that ?

maliberty:
> /OpenROAD/src/odb/src/lef/lef/lefiLayer.cpp:1979:3: error: 'uint32_t' was not declared in this scope

Needs `#include <cstdint>`

maliberty:
You can ignore the mermaid error

gadfort:
@maliberty what is wrong with the rule? I though min cut was specified as `>= WIDTH` so the fact that the min cut rule specifies something less than the min width still seems correct to me. @VoltXd what is the issue with keeping the rule in the database?

maliberty:
What makes this a "bad" mincut?  The rule is "Specifies the number of cuts a via must have when it is on a wide wire or pin whose width is greater than _width_".  So having a rule less than min width isn't wrong as it applies to larger values.

VoltXd:
The thing is Innovus ignores these "bad" rules (cf. https://github.com/The-OpenROAD-Project/OpenROAD/issues/5975#issuecomment-2830807641) and as @titan73 told me, it induces many "no access point" issues in openroad. I will have to talk about it with him to get more information.

gadfort:
@VoltXd it seems to me that either the LEF is wrong and should be corrected, or that the PDK requires two cuts. Just ignoring a rule seems dangerous.

PR #7241, opened 2025-04-30
utl: add Histogram helper class

gadfort:
Adds:
- `utl::Histogram` to allow for histograms to to generated throughout openroad.
- Class handles generating the text report for the histogram (matching the behavior of `report_timing_histogram` and `report_logic_depth_histogram`

Changes:
- removes custom class in `dbSta` handling histograms in favor of `utl::Histogram`
- change GUI charts to use histogram as well

@oharboe could you check that the GUI histogram works as you expect? I spot checked and didn't notice anything, but I don't use it that often, so you might have better insight.


oharboe:
Does this mean that there will be a more explicit way to "warm up" the histogram?

open.tcl populates the timing reports to minimize click+wait minutes+click+wait minutes and make the use case => wait for load, then click & immediate response + click.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/9c1f98ec2951e0b6aa908a251474329e84d03b11/flow/scripts/open.tcl#L67

oharboe:
Tried it, I saw no differences in behavior. A nice improvement was the option to view hold and setup slacks separately in addition to timing groups.

Tried on megaboom design(large).

image

src/utl/include/utl/histogram.h
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2023-2025, The OpenROAD Authors
```
maliberty, line 2: start in 2025

src/utl/src/histogram.cpp
```diff
@@ -0,0 +1,143 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2023-2025, The OpenROAD Authors
```
maliberty, line 2: 2025

src/utl/include/utl/histogram.h
```diff
@@ -0,0 +1,84 @@
...
+#include <string>
+#include <vector>
+
+namespace utl {
+
+class Logger;
+
+template <typename DataType>
+class Histogram
+{
```
maliberty, line 17: As you are using a cpp with explicit instantiation it would be good to comment here that this can only be used with DataType = int | float.

maliberty (changes requested):
Would you add the logic depth histogram to the GUI (in this PR or another)?

maliberty (approved):

gadfort:
Looks like the CI is failing with:
```
No space left on device
	java.base/sun.nio.ch.FileDispatcherImpl.write0(Native Method)
	java.base/sun.nio.ch.FileDispatcherImpl.write(FileDispatcherImpl.java:62)
	java.base/sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:[1](https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-7241-merge/6/pipeline-console/?start-byte=0&selected-node=39#log-1)32)
	java.base/sun.nio.ch.IOUtil.write(IOUtil.java:97)
	java.base/sun.nio.ch.IOUtil.write(IOUtil.java:67)
```

maliberty:
@sombraSoft see the disk space issue above

maliberty (approved):

PR #7243, opened 2025-04-30
[DO NOT REVIEW] bazel: use exec config for openroad in tests to avoid rebuilds

oharboe:
Use target configuration for tests so that `bazel build -c :opt openroad` and `bazel test -c opt ...` uses same OpenROAD build.

This saves time and is idiomatic in that OpenROAD is a tool used to build things in bazel, it is intended to be used as the build process, it is not an application that is to be created and has nothing to do with the bazel build stages afterwards.

However, a file like `src/rsz/src/Resizer.cc` is built twice, because ...(???)

Before, multiple configurations, i.e. openroad has to be built multiple times:

```
bazelisk cquery -c opt --output=graph 'allpaths(//test/..., //:openroad)' | dot -Tsvg -o file.svg
```

To view differences of configurations, run:

```
diff -u <(bazelisk config 0b24ccd)  <(bazelisk config 206f41d) | bat
```

image

After, one configuration, openroad has to be built once:

image

### DIFF
--- a/src/ant/test/BUILD
+++ b/src/ant/test/BUILD
@@ -44,7 +44,7 @@ filegroup(
 [
     filegroup(
         name = test_name + "_resources",
-        data = [":regression_resources"] + glob(
+        srcs = [":regression_resources"] + glob(
             [
                 test_name + ".*",
             ],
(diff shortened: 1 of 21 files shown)


oharboe:
```
diff -u <(bazelisk config 0b24ccd)  <(bazelisk config 206f41d) | bat
```

PR #7245, opened 2025-04-30
Fix SEGFAULT when two nets have the same name

Faholan:
When loading a .def file in which two nets have the same name, OpenROAD segfaults due to dereferencing a null pointer. I thus add an assert to check for this specific case.

For example, the following file, loaded through the `read_def` command, segfaults.

```def
VERSION 5.8 ;
DIVIDERCHAR "/" ;
BUSBITCHARS "[]" ;
DESIGN segfault ;
NETS 2 ;
    - a
        + ROUTED
        met2
            ( 0 0 ) ( 0 0 ) ;
    - a
        + ROUTED
        met2
            ( 0 0 ) ( 0 0 ) ;
END NETS
END DESIGN
```

### DIFF
--- a/src/odb/src/defin/definNet.cpp
+++ b/src/odb/src/defin/definNet.cpp
@@ -331,6 +331,7 @@ void definNet::wire(dbWireType type)
         _wire = dbWire::create(_cur_net);
       }
     }
+    assert(_wire != nullptr);
     _wire_encoder.begin(_wire);
   }
 


maliberty:
How does adding an assert fix this?

gadfort:
seems like this should be an error for the def reader so we can get the line number information in the DEF.

Faholan:
> How does adding an assert fix this?

The function `dbWire::create` returns `nullptr` if the net passed already has a wire attached (which happens when two nets share the same name, as the net is then retrieved instead of created). The assert checks for this case with minimal overhead

maliberty:
Exiting with an assert is not the right solution.  I'll take this as a bug report rather than a fix.

maliberty:
Thanks for pointing out the issue and looking into it.  https://github.com/The-OpenROAD-Project/OpenROAD/pull/7247 is more what I have in mind for a fix.

PR #7255, opened 2025-05-01
CTS during buffer inference prioritize buffers that have input ports with LEF USE as "CLOCK"

arthurjolo:
During CTS buffer inference prioritize buffers that have input ports with LEF USE as "CLOCK" alongside with looking for buffers that have "is_clock_cell: true". If none are found proceed with searching buffers by name.


jeffng-or:
can we add a regression test that has a fake library that has both clkbuf and buf cells to ensure that we choose the clkbuf?

jeffng-or:
Another option is to pull the buffer inference code into its own class and build a unit test around it. That would require more work, but would be useful for testing out the scenarios directly.

maliberty:
I don't see the prioritization as both conditions just add to selected_buffers.  Also no unit test per Jeff's request.

arthurjolo:
> I don't see the prioritization as both conditions just add to selected_buffers. Also no unit test per Jeff's request.

Here we are prioritization selecting the buffer that have any atribute that indicates if they are clock buffers, that is if they have the liberty "is_clock_cell : true" or if they have an input pin that has a LEF USE as CLOCK. So I didn't define a priority between them because in my opinion both are clock buffer and should be used for clock tree.

jeffng-or:
> It sounds like you changed the selection criteria not that a priority was imposed. Is that correct?

Based on what I saw in the code, if it finds a buffer that matches the clock conditions Arthur mentioned above, it will stop searching for other buffers. So, effectively, there is a priority. 

I had the same question since the private PDK had both CLKBUF and BUF cells. Only the CLKBUF ones were selected by inference and the BUF cells were ignored.

jeffng-or:
> I'm just looking at this diff and all I see is more buffers added to selected_buffers. Where does it stop searching?

every successive "if" checks to see if selected_buffers is empty before entering the conditional. So, since the CLKBUFs were added to selected_buffers, all of the subsequent places where additional buffers can be added are skipped.

maliberty:
Has this been through a secure CI?

I see the prioritization isn't between these two sources but between selected and non-selected cases elsewhere.

precisionmoon (approved):
The priority is given to "is_clock_cell" liberty attribute first.  The second is LEF input port.  This is OK.

arthurjolo:
Your right Matt, this PR doesn't add any post buffer selection priority. I understand that the from test case from Jeff we don't select the correct buffers, so this PR makes that we try to only select the buffers that have an attribute indicating they are clock.

arthurjolo:
> The priority is given to "is_clock_cell" liberty attribute first. The second is LEF input port. This is OK.

I will change so we prioritize this.

precisionmoon:
> > The priority is given to "is_clock_cell" liberty attribute first. The second is LEF input port. This is OK.
> 
> I will change so we prioritize this.

Liberty attribute should be higher priority than LEF.

arthurjolo:
I stated a secure-CI, but this will probably not affect any design as none of our current PDKs have a cell if a pin that has LEF USE "CLOCK".

src/cts/test/check_buffer_inference1.ok
jeffng-or, line 1: Should this be an info() message vs report()?

src/cts/test/check_buffer_inference3.tcl
jeffng-or, line 1: Maybe add a small comment at the beginning of the file to doc what this test is testing (e.g. choosing clk buffers with based on CLKBUF naming match). Same thing goes for the other two Tcl files.

src/cts/test/check_buffer_inference1.ok
arthurjolo: It is as a report because it is only done if the debug mode is on. I didn't add that and didn't really give it much attention. I will change it to a `debugprint`

arthurjolo:
CI is passing, results are identical. The bazel build is failing with `No space left on device`.

PR #7256, opened 2025-05-02
Fixed -library documentation and added a buffer inference debug state…

jeffng-or:
…ment

Fixed -library command line arg doc
Added debug statement when we remove an inferred buffer due to being within 10% of its neighbor's max cap

### DIFF
--- a/src/cts/README.md
+++ b/src/cts/README.md
@@ -66,6 +66,7 @@ clock_tree_synthesis
     [-dont_use_dummy_load]
     [-sink_buffer_max_cap_derate derate_value]
     [-delay_buffer_derate derate_value]
+    [-library liberty_library_name]
 ```
 
 #### Options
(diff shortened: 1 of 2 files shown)


src/cts/README.md
```diff
@@ -66,6 +66,7 @@ clock_tree_synthesis
     [-dont_use_dummy_load]
     [-sink_buffer_max_cap_derate derate_value]
     [-delay_buffer_derate derate_value]
+    [-library liberty_library_name]
```
maliberty, line 69: @luarss shouldn't the doc checker have flagged this missing arg earlier?

maliberty (approved):

src/cts/README.md
precisionmoon: Sorry, this initially went in as a hidden option.

jeffng-or: Interesting, since the -library option was described in the table below. It was just missing from the usage section. We're OK with it being a public option, right?

PR #7267, opened 2025-05-03
bazel: break out more libraries into their respective directories.

hzeller:
Added
  * src/ant/BUILD
  * src/cts/BUILD
  * src/dpl/BUILD
  * src/dbSta/BUILD
  * src/grt/BUILD
  * src/rsz/BUILD
  * src/stt/BUILD


src/ant/BUILD
```diff
@@ -0,0 +1,85 @@
...
+
+load("//bazel:tcl_encode_or.bzl", "tcl_encode")
+load("//bazel:tcl_wrap_cc.bzl", "tcl_wrap_cc")
+
+package(
+    default_visibility = ["//:__subpackages__"],
+    features = ["layering_check"],
+)
+
+# Broken out due to cyclic dependency between grt and ant
```
maliberty, line 12: @eder-matheus @luis201420 I thought this was resolved?

oharboe:
🥳

Silly question: I was spelunking in Bazel docs and asking ChatGPT, but I cant figure out how Bazel handles the dynamic dependency for .cc files on headers. Make uses .d files, how does this work in Bazel?

Any pointers to read up on?

src/cts/src/TritonCTS.i
```diff
@@ -5,9 +5,9 @@
 #include <cstdint>
 
 #include "cts/TritonCTS.h"
-#include "CtsOptions.h"
-#include "TechChar.h"
-#include "CtsGraphics.h"
+#include "src/cts/src/CtsOptions.h"
```
oharboe, line 8: What's going on here? Why should the path be included?

This is included from `bazel-bin/src/cts/triton_cts_swig.cc`, so shouldn't `#include "CtsOptions.h"` work?

Not a suggestion, just trying to learn....

hzeller: Don't remember; it didn't work, possibly  because the include path was not complete ? In the end, the absolute path is the better choice anyway, so didn't think much about it.

oharboe:
I mean simply that if you have foo.c/h and bar.c/h, where foo.c includes foo.h and same for bar, in a cc_library()

What stops everything from being recompiled when either foo.h or bar.h changes?

src/cts/src/TritonCTS.i
hzeller: ah, and looks like the cmake build doesn't like it, so I need to undo this change and work around it in bazel.
So, staying draft for now until that is fixed.

oharboe: Yep...

```
$ bazel build -c opt src/cts:triton_cts_swig
INFO: Invocation ID: 70621226-8435-4182-a789-86760a671184
INFO: Analyzed target //src/cts:triton_cts_swig (0 packages loaded, 4825 targets configured).
INFO: Found 1 target...
Target //src/cts:triton_cts_swig up-to-date:
  bazel-bin/src/cts/triton_cts_swig.cc
INFO: Elapsed time: 2.095s, Critical Path: 1.65s
INFO: 2 processes: 1 internal, 1 processwrapper-sandbox.
INFO: Build completed successfully, 2 total actions
$ git diff FETCH_HEAD
diff --git a/src/cts/src/TritonCTS.i b/src/cts/src/TritonCTS.i
index 662f257c8..ef6642565 100644
--- a/src/cts/src/TritonCTS.i
+++ b/src/cts/src/TritonCTS.i
@@ -5,9 +5,9 @@
 #include <cstdint>
 
 #include "cts/TritonCTS.h"
-#include "src/cts/src/CtsOptions.h"
-#include "src/cts/src/TechChar.h"
-#include "src/cts/src/CtsGraphics.h"
+#include "CtsOptions.h"
+#include "TechChar.h"
+#include "CtsGraphics.h"
 #include "ord/OpenRoad.hh"
 
 namespace ord {
```

QuantamHD:
@oharboe Due to legacy reasons you are not required to sandbox your header files in the same ways. Internally at google there's a header scanning which is the equivalent of the .d 

It's disabled by default in bazel.
 
image

QuantamHD:
@oharboe See also https://github.com/bazelbuild/bazel/issues/8319

maliberty:
That option suggests a saga....

QuantamHD:
99.99999999% of things will work, and it's enabled inside of Google. What doesn't work, and 1 million people will complain about is this exact code segment

```
#include __MY_DOPE_MACRO_INCLUDE_FILE__
```

The bazel team doesn't want to deal with it, so they've named it that.

maliberty:
Where `__MY_DOPE_MACRO_INCLUDE_FILE__` is a macro itself?  Icky.

maliberty:
I would be fine with saying that is not allowed in the OR code base.  It trickier to say it doesn't happen in any of our dependencies.

QuantamHD:
> Where `__MY_DOPE_MACRO_INCLUDE_FILE__` is a macro itself? Icky.

Correct.

oharboe:
> @oharboe Due to legacy reasons you are not required to sandbox your header files in the same ways. Internally at google there's a header scanning which is the equivalent of the .d
> 
> It's disabled by default in bazel.

Thanks!

I've been trying to find out why Bazel builds a lot more than I expect in OpenROAD and when I tried to understand how Bazel was doing the equivalent of .d files, I couldn't find anything.

Link for reference https://bazel.build/reference/command-line-reference#flag--experimental_unsupported_and_brittle_include_scanning

oharboe:
https://github.com/The-OpenROAD-Project/OpenROAD/pull/7269

hzeller:
There is one test from one of the installers that failed. Sounds like a fluke.
Ready to merge.

src/grt/BUILD
```diff
@@ -0,0 +1,154 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025, The OpenROAD Authors
+
+load("//bazel:tcl_encode_or.bzl", "tcl_encode")
+load("//bazel:tcl_wrap_cc.bzl", "tcl_wrap_cc")
+
+package(
+    default_visibility = ["//:__subpackages__"],
```
maliberty, line 8: How is this different than `//visibility:public` ?  It seems to apply to everything.

maliberty:
I'll merge this and my questions and can discussed/addressed subsequently.

src/stt/BUILD
```diff
@@ -0,0 +1,90 @@
...
+    ],
+    outs = ["src/flt/etc/POWV9.cpp"],
+    cmd = "$(execpath //etc:file_to_string) --inputs $(location src/flt/etc/POWV9.dat) --output \"$@\" --varname powv9 --namespace stt::flt",
+    tools = [
+        "//etc:file_to_string",
+    ],
+)
+
+cc_library(
+    name = "stt",
```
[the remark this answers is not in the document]
hzeller: Here, I just transferred things as-is from the toplevel BUILD.bazel, but yeah things should be broken up more for testing.

src/cts/test/BUILD
```diff
@@ -158,3 +158,15 @@ filegroup(
     )
     for test_name in ALL_TESTS
 ]
+
+cc_test(
+    name = "cts_unittest",
+    srcs = ["cts_unittest.cc"],
+    linkstatic = True,  # TODO: remove once deps define all symbols
```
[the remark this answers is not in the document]
hzeller: If you  set it to `False`, there is some issue that emerges (usually cc_tests() are dynamically linked to allow for faster incremental tests, but last time I checked there were issues in the current dependencies).

src/grt/BUILD
hzeller: With public, everyone using openroad as a dependency can see also this target, while this makes sure it is limited to openroad deps.

src/stt/BUILD
```diff
@@ -0,0 +1,90 @@
...
+    ],
+    hdrs = [
+        "include/stt/MakeSteinerTreeBuilder.h",
+        "include/stt/SteinerTreeBuilder.h",
+        "include/stt/flute.h",
+        "include/stt/pd.h",
+    ],
+    includes = [
+        "include",
+        "include/stt",
```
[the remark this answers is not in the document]
hzeller: Don't remember, but looks like there is at least one header included where that is needed.

PR #6504, opened 2025-01-11
clang-tidy: add `<vector>` where needed.

hzeller:
Automatically created by fixing all files mentioned in the clang-tidy output:

```
../dev-tools/insert-header "<vector>" $(awk -F: '/no header providing.*std::vector.*misc-include-cleaner/ {print $1}' OpenRoad_clang-tidy.out | sort | uniq)
clang-format -i $(awk -F: '/no header providing.*std::vector.*misc-include-cleaner/ {print $1}' OpenRoad_clang-tidy.out | sort | uniq)
```

PR #7269, opened 2025-05-04
bazel: enable C++ include file scanning

oharboe:
Flip the switch and see what happens...

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -50,6 +50,11 @@ build:profile --copt -fno-omit-frame-pointer
 # TODO: document
 build --incompatible_strict_action_env
 
+# Faster builds by pruning the dependency graph. OpenROAD
+# forbids the `#include FOO` pattern in the code, which
+# is the case that will never be supported with this feature.
+build --experimental_unsupported_and_brittle_include_scanning
+
 # TODO: setup a new cache for OpenROAD
 # Read only remote cache, updated by OpenROAD team members and CI.
 #build --remote_cache=https://storage.googleapis.com/megaboom-bazel-artifacts
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -155,34 +155,6 @@ cc_library(
     deps = OPENROAD_LIBRARY_DEPS,
 )
 
-cc_library(
-    name = "openroad_lib",
-    srcs = OPENROAD_BINARY_SRCS_WITHOUT_MAIN + glob(
-        include = OPENROAD_LIBRARY_SRCS_INCLUDE,
-        allow_empty = True,
-        exclude = OPENROAD_LIBRARY_SRCS_EXCLUDE,
-    ) + [
-        "src/OpenRoad.cc",
-        "src/stt/src/flt/etc/POST9.cpp",
-        "src/stt/src/flt/etc/POWV9.cpp",
-    ],
-    hdrs = glob(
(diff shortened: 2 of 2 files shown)


hzeller:
I think I wouldn't do this until we consistently use the project-root relative paths for headers (right now finding headers relies on a myriad of `-I` paths to work).
Also we should make sure that we actually include the headers we use (last time I checked, there were still many thousands `misc-include-cleaner` warnings).

oharboe:
> I think I wouldn't do this until we consistently use the project-root relative paths for headers (right now finding headers relies on a myriad of `-I` paths to work). Also we should make sure that we actually include the headers we use (last time I checked, there were still many thousands `misc-include-cleaner` warnings).

Are there any downsides to enabling this early?

The OpenROAD bazel build won't be in production(replace cmake) until it has Qt + readline support(minimally), so there's time before any downsides affects anyone but developers of bazel builds. I'm thinking enabling this early gives us time to learn.

hzeller:
I think the option assumes a somewhat strict defined project which this is not, so we don't know if the heuristics that assume that make the right decisions; that is why this flag is default off.

To prepare for a better future (which also works with cmake and makes cmake builds better as well) is

  * Fix `misc-include-cleaner` warnings. Run the `cmake` build to create the compilation db, then run `etc/run-clang-tidy-cached.cc --checks="-*,misc-include-cleaner"` to find all the things that need addressing. Have a look at some of my earier PRs #7119 #6939 #6504 #6423 (in some of them I describe what scripts I used to make it somewhat easy).
  There are two kinds of `misc-include-cleaner` warnings: 
   1) the ones that complain about headers not included that provide a particular symbol (they only accidentally work due to other headers happen to include that). `"warning: no header providing "somethingsomething" is directly included"` (majority of the warnings and most important to fix)
   2) warnings about includes that are not providing symbols needed  and can be removed `"included header geometry.hpp is not used directly"` (< 20% of warnings).

We first need to fix (1) so that files are fully defined, then we can do the satisfying removal of (2) (We can't do (2) first, as these header might contribute to the missing symbols accidentally being included).

  * Use includes relative to project root. .Say we defined `src` being the project root, we should have includes like `#include "db/include/odb/db.h"`. Even if it looks somewhat verbose right now, but can be fixed easily later once the project layout is improved without the confusing nested set of src/ and include/ directories (so I imagine a `db/odb/db.h` in a fixed layout structure at some point).
     * Ideally rename the toplevel `src/` to `openroad/` and use that as toplevel, which now has a nice name. Then we can have `#include "openroad/db/include/odb/db.h"`, and things would start to look like how things would be expected by bazel.

Once all that is done, we're more closely to a well-defined bazel project, so then can play with enabling that flag.

oharboe:
@hovind Do we have something to create compile_commands.json other than cmake? I thought this file was created by bazel in our case to support vscode use-case?

hzeller:
bazel does not create compile commands unfortunately, but I have written a tool [`bant`](https://github.com/hzeller/bant) that can do that.

With the current state of the project it will not work well yet though as a bunch of information such as the large include search path that is hidden away in the `build_helper.bzl` that `bant` won't read. So once we have a clean project, we can use bant to create a compilation DB (and use it for other interesting insights)

oharboe:
Tried it out: https://github.com/The-OpenROAD-Project/OpenROAD/pull/7270

PR #7270, opened 2025-05-04
includes: prune some unused includes

oharboe:
Tried these instructions https://github.com/The-OpenROAD-Project/OpenROAD/pull/7269#issuecomment-2849049020

There are many 1358 "warning: included header ... is not used directly" warnings, fixed a few at the compilation unit level.

Fixing those in include headers is trickier as it requires adding include files in some other file where they are used.

Google coding standard, which I guess we're following here, is to disallow forward declarations like "struct Foo;". Rather one should include onlyh what is used in a .h/.cpp file.

### DIFF
--- a/src/dpl/src/Opendp.cpp
+++ b/src/dpl/src/Opendp.cpp
@@ -3,13 +3,8 @@
 
 #include "dpl/Opendp.h"
 
-#include <boost/property_tree/json_parser.hpp>
-#include <boost/property_tree/ptree.hpp>
 #include <cfloat>
 #include <cmath>
-#include <iostream>
-#include <limits>
-#include <map>
 #include <memory>
 #include <string>
 #include <utility>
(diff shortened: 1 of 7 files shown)


hzeller:
this is why we should first fix the missing headers (per instructions), then it is easy to remove the non-needed ones.

hzeller:
Yes, we also should get rid of forward declarations as much as possible.

hzeller (approved):
LGTM
(but first I'd focus on adding missing headers, as that is harder and don't accidentally break things)

oharboe:
this is fine, far as it goes, and ready to be merged

maliberty:
It is a large effort to get to IWYU-style in OR.

maliberty (approved):

oharboe:
> It is a large effort to get to IWYU-style in OR.

Yes, I just tried out the workflow and fixed a few cases to educate myself.

PR #7272, opened 2025-05-05
gpl: fix compile error, missing include

oharboe:


oharboe:
@hzeller FYI

@maliberty I suppose the regular flow must be fixed first(merge this PR), then bazel can be fixed at a more leisurly pace?

hzeller:
The bazel build might be fixed with #7276

Issue #7013, opened 2025-04-06
Bogus(?) warnings when running open.tcl on global_route result


oharboe:
### Describe the bug

untar and run [global-route-warning.tar.gz](https://github.com/user-attachments/files/19621177/global-route-warning.tar.gz)

```
OpenROAD v2.0-20272-g655150e07 
Features included (+) or not (-): +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
read_db results/asap7/gcd/base/5_1_grt.odb
GUI_TIMING=1 reading timing, takes a little while for large designs...
read_sdc /home/oyvind/OpenROAD-flow-scripts/flow/bar/results/asap7/gcd/base/5_1_grt.sdc
estimate_parasitics -global_routing
[WARNING GRT-0026] Missing route to pin _562_/CON in net _004_.
[WARNING GRT-0026] Missing route to pin _569_/CON in net _017_.
[WARNING GRT-0026] Missing route to pin _560_/CON in net _028_.
[WARNING GRT-0026] Missing route to pin _487_/Y in net _066_.
[WARNING GRT-0026] Missing route to pin _526_/Y in net _075_.
[WARNING GRT-0026] Missing route to pin _299_/Y in net _086_.
[WARNING GRT-0026] Missing route to pin _325_/Y in net _109_.
[WARNING GRT-0026] Missing route to pin _328_/Y in net _112_.
[WARNING GRT-0026] Missing route to pin _340_/Y in net _122_.
[WARNING GRT-0026] Missing route to pin _420_/B in net _126_.
[WARNING GRT-0026] Missing route to pin _346_/Y in net _127_.
[WARNING GRT-0026] Missing route to pin _439_/B in net _130_.
[WARNING GRT-0026] Missing route to pin _368_/Y in net _133_.
[WARNING GRT-0026] Missing route to pin _378_/Y in net _142_.
[WARNING GRT-0026] Missing route to pin _380_/Y in net _144_.
[deleted]
Populating timing paths...OK
openroad> 
```

Don't know what I'm looking at here, but this is net _004_ highlighed.

![Image](https://github.com/user-attachments/assets/c8212eeb-75c8-4729-a18a-6c8fac58585a)

### Expected Behavior

No warnings

### Environment

```shell
OpenROAD v2.0-20272-g655150e07
```

### To Reproduce

See above.

Or:

```
[...]

PR #7280, opened 2025-05-05
odb: clear iterm aps when instance is changed

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD/issues/7013

Secure-CI started.


src/odb/src/db/dbITerm.cpp
```diff
@@ -813,6 +813,14 @@ std::vector<dbAccessPoint*> dbITerm::getPrefAccessPoints() const
   return aps;
 }
 
+void dbITerm::clearPrefAccessPoints()
+{
+  _dbITerm* iterm = (_dbITerm*) this;
+  // Clear aps_ map instead of destroying dbAccessPoint object to prevent
+  // destroying APs of other iterms.
+  iterm->aps_.clear();
```
maliberty, line 821: Are the APs shared somehow?

eder-matheus: From my tests, it seems so. When I destroyed the APs of the instances that were moved or had their masters swapped, a bunch of other instances that weren't changed at all lost their APs.

If I understand it correctly, the APs are associated to the dbMPin, under dbMTerm. So all the dbITerms that share the same dbMTerm would lost their APs if we delete it from one of them.

maliberty: In that case we should eventually reference count the APs so they can be cleaned up on the last inst.  That and undo can wait for a future PR.

eder-matheus:
@maliberty Thanks for approving. I'll merge once secure-ci is done.

PR #7281, opened 2025-05-05
odb: Marking table as final

QuantamHD:
Provides a 13% improvment on 2 million standard cell block `report_checks` call.


maliberty (approved):

QuantamHD:
The dbTable implementation is surprisingly heavy. 
image

PR #7282, opened 2025-05-05
Do not skip the destructor in extMeasure

Faholan:
Since 4c290dc, the extMeasure destructor is skipped, leading to memory leaks notably of the _rc array. Why was such behaviour introduced in the first place, and shouldn't it be reverted?

### DIFF
--- a/src/rcx/src/extRCmodel.cpp
+++ b/src/rcx/src/extRCmodel.cpp
@@ -2363,7 +2363,6 @@ void extMeasure::allocOUpool()
 
 extMeasure::~extMeasure()
 {
-  return;
   for (auto& ii : _rc) {
     delete ii;
   }


Faholan:
Additionally, I have identified quite a few other memory leaks and am working on fixes. Should I introduce them as separate pull requests, or as a bundle in one PR? I am inclined to do them separately but I don't want to spam you with too many PRs at the same time

maliberty:
Small PRs are easy to review so separate is fine.  OpenRCX is has lots of room for improvement :)

maliberty (approved):

maliberty:
> Why was such behaviour introduced in the first place, and shouldn't it be reverted?

It was part of a massive PR that can't be reverted and was mostly impossible to review well.  Thanks for noticing this.

PR #7283, opened 2025-05-05
Fix memory leak in the extDistRCTable destructor

Faholan:
empty the table before dropping it


maliberty (approved):

oharboe:
Cleanups! :relieved:

Faholan:
I have investigated further, this fix is actually incorrect, the leak happens in another place and requires a separate fix, I will do another PR for it

PR #7284, opened 2025-05-05
Clean the model table in extmain

Faholan:


maliberty (approved):
Depending on how big a task you want to bite off you could look at moving away from such manual memory management to a more modern c++ style.

Faholan:
I would've like to implement this in the destructor of Ath__array1D, but I checked and it is often used with scalar types that do not require freeing, and I am unsure of whether I am capable of doing this

maliberty:
> I would've like to implement this in the destructor of Ath__array1D, but I checked and it is often used with scalar types that do not require freeing, and I am unsure of whether I am capable of doing this

No worries, I'm glad to have it fixed in any case.

PR #7286, opened 2025-05-05
Correctly free allocated memory in pldata array

Faholan:


src/rcx/src/gs.cpp
```diff
@@ -64,6 +64,12 @@ gs::~gs()
 void gs::freeMem()
 {
   if (init_ & ALLOCATED) {
+    for (int i = 0; i < pldata_.size(); i++) {
+      pixmap* pm = pldata_[i].plane;
+      if (pm != nullptr) {
+        free(pm);
+      }
```
maliberty, line 71: Freeing a nullptr is a no-op so no need for the check.

Faholan: Oh I wasn't sure of that, thank you

maliberty: https://en.cppreference.com/w/c/memory/free  

cppreference.com is helpful for language questions.

QuantamHD:
I think you need to remove your submodule change

src/rcx/src/gs.cpp
```diff
@@ -145,6 +148,10 @@ void gs::setSize(const int plane,
     exit(-1);
   }
 
+  if (plc.plane != nullptr) {
+    free(plc.plane);
+  }
```
maliberty, line 153: No need to check for null before free (again)

Faholan: Indeed

src/rcx/src/gs.cpp
```diff
@@ -64,6 +64,9 @@ gs::~gs()
 void gs::freeMem()
 {
   if (init_ & ALLOCATED) {
+    for (auto& pm : pldata_) {
+      free(pm.plane);
+    }
```
maliberty, line 69: To be safe it would be good to set place to a nullptr, eg

```
  struct plconfig
  {
...
    pixmap* plane{nullptr};
  };
```

maliberty (approved):

PR #7314, opened 2025-05-07
Setup fixes

oharboe:
Test in CI...


oharboe:
@maliberty bazel builds still fail in CI, unrelated

PR #7320, opened 2025-05-07
Cleanup the search field

Faholan:


src/rcx/src/extFlow.cpp
```diff
@@ -318,6 +318,7 @@ uint extMain::initSearchForNets(int* X1,
   }
   const uint layerCnt = n + 1;
 
+  delete _search;
   _search = new GridTable(&maxRect, 2, layerCnt, pitchTable, X1, Y1);
```
maliberty, line 322: If you are comfortable, this is a case where std::unique_ptr and make_unique would be a more elegant solution.  If not I can merge it as is.

src/rcx/src/extmain.cpp
```diff
@@ -109,6 +109,7 @@ extMain::extMain()
 }
 extMain::~extMain()
 {
+  _search.reset();
```
maliberty, line 112: not needed, the dtor will clean up the memory

Faholan: Will the destructor run automatically if I don't delete the `_search`, or should I add a `delete _search`? The documentation isn't really clear on what will be the behaviour

maliberty: Class members are always destroyed when the class is destroyed.  See https://en.cppreference.com/w/cpp/language/destructor:
```
...the compiler calls the destructors for all non-static non-variant data members of the class, in reverse order of declaration,...
```

maliberty (approved):

PR #7327, opened 2025-05-08
Fix leak of rcDistTable

Faholan:
Cache the size of the arrays to ensure that every allocated element is deleted


maliberty:
Please give some description of these changes.  It isn't obvious why you need to cache these sizes.

Faholan:
The issue is that currently the code references the length of the arrays in the destructor, however it is not necessarily equal to their size, which would be the right metric.

Notably, one of the constructors takes as input a `maxWidthCnt` and not an exact count, so it needs to be cached in order for the cleanup to be accurate.

In that constructor the `_widthTable` is initialized empty with the right size.
`_widthTable = new Ath__array1D<int>(maxWidthCnt);`.

Similarly, for the diagonals, the `_diagWidthTable` and `_diagDistTable` are have their elements initialized empty.

Caching the correct sizes allows the destructor to clean up correctly

maliberty (approved):

PR #7335, opened 2025-05-09
rcx: fix msan violation in destruction

mikesinouye:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/6b031011d4fc38320530de283fe709046c9239e7/src/rcx/src/extmain.cpp#L185 can be called in destruction before `extMain::initContextArray()`, in which the values of `_ccContextPlanes` and `_dgContextPlanes` can be checked before initialization. MemorySanitizer will fail on this behavior.

### DIFF
--- a/src/rcx/include/rcx/extRCap.h
+++ b/src/rcx/include/rcx/extRCap.h
@@ -2734,7 +2734,7 @@ class extMain
 
   Ath__array1D<SEQ*>*** _dgContextArray = nullptr;
   uint _dgContextDepth;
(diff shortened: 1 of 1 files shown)


Faholan:
In #7329 I remove the check of `_ccContextPlanes` and `_dgContextPlanes`, as memory would be leaked if they were set to 0, which in turn fixes this uninitialized read.

mikesinouye:
Ah I should have checked for the open PRs first, glad it's already properly fixed. I'll close this.

PR #7336, opened 2025-05-09
Keep track of whether measuretable should be deleted

Faholan:


src/rcx/src/extmeasure_res.cpp
```diff
@@ -292,6 +292,10 @@ extDistRC* extDistRCTable::getComputeRC_res(uint dist1, uint dist2)
     }
   }
   if (found) {
+    if (!_measureInR) {
```
maliberty, line 295: Why is _measureInR neede?

Faholan: This variable tracks whether the current measure table is stored in `_measureTableR` or not. This is to avoid risking a use after free, or a double free. If it is, then it shouldn't be deleted. But if it isn't, it should be deleted to avoid leaking.

Assignments of the type `_measureTable = _measureTableR[index_dist];` are the issue

maliberty (approved):

PR #7339, opened 2025-05-10
Fix a leak in PDN

Faholan:
Replace the shared pointers upper_ and lower_ with weak pointers, to avoid a reference cycle that ends up leaking

### DIFF
--- a/src/pdn/src/PdnGen.cc
+++ b/src/pdn/src/PdnGen.cc
@@ -167,8 +167,8 @@ void PdnGen::updateVias()
     grid->getVias(all_vias);
 
     for (const auto& via : all_vias) {
-      via->getLowerShape()->lock()->addVia(via);
-      via->getUpperShape()->lock()->addVia(via);
+      via->getLowerShape().lock()->addVia(via);
+      via->getUpperShape().lock()->addVia(via);
     }
   }
 }
--- a/src/pdn/src/PdnGen.cc
+++ b/src/pdn/src/PdnGen.cc
@@ -167,8 +167,8 @@ void PdnGen::updateVias()
     grid->getVias(all_vias);
 
     for (const auto& via : all_vias) {
-      via->getLowerShape()->addVia(via);
-      via->getUpperShape()->addVia(via);
+      via->getLowerShape()->lock()->addVia(via);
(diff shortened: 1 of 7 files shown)


maliberty:
You have many test failures so something is wrong.  @gadfort you can hold off until this is passing

fivemserverhostings:
Impressive project structure! The architecture really demonstrates solid design principles.

myeyestore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

eyestorefivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

eyesaider:
This is a well-structured codebase. I appreciate the clean separation of concerns.

godofcodeeyes:
The performance optimizations here are quite clever. Nice engineering work!

eyesgodcode:
Excellent contribution to open source! The community will benefit greatly from this.

gadfort:
@Faholan what leak is this resolving?

fivemeyestore:
Excellent contribution to open source! The community will benefit greatly from this.

Faholan:
> @Faholan what leak is this resolving?

There is a reference cycle between Shape and Via.

It's very explicit in this line:

`via->getLowerShape()->addVia(via);`

This cannot be automatically collected as the via indirectly hols a `shared_ptr` to the upper and lower shapes, which reference the via in turn

gadfort:
@Faholan I see, yes that is an issue. At the moment because the shapes are shared there is no need to ensure they exist in the vias, but they can be deleted (modified, etc) while the via is still there (this gets resolved later, but it's possible this is that you are running into in this fix).
There are a lot of via operations, so hopefully he added overhead of `.lock()` and checking for a valid pointer doesn't cause too much of a runtime penalty.

gadfort:
@Faholan is there a tool that indicated the leak? if so, I can take a stab at it

Faholan:
Yes the leak is indicated by adress sanitizer. I have attached the output it gives after execution.
ASAN can be enabled by passing the `-DASAN=ON` flag when building OpenRoad
[indirect_leakslog](https://github.com/user-attachments/files/20196580/leak_all.log)

PR #7344, opened 2025-05-12
gui: add labels

gadfort:
Adds:
- user defined labels in GUI.

This should allow us to avoid calls to system in https://github.com/The-OpenROAD-Project/OpenROAD/pull/7236 for adding a label to the image.

@maliberty I'm not totally happy with having to expose `QFont` via the `printer`, but maybe it's okay since that's already exposed via the options.h. If you have a better suggestion let me know.


src/gui/README.md
```diff
@@ -644,6 +644,54 @@ To remove all the rulers:
...
+
+Returns: name of the newly created label.
+
+```tcl
+gui::add_label
+    x y
+    text
+    [color]
+    [size]
+    [name]
```
maliberty, line 659: It would be nicer to use switches than optional positional arguments.

src/gui/README.md
```diff
@@ -644,6 +644,54 @@ To remove all the rulers:
...
+
+#### Options
+
+| Switch Name | Description |
+| ---- | ---- |
+| `x, y,` | point of the label in microns. |
+| `text` | text for the label. |
+| `color` | color to use for the label. |
+| `size` | size of the label. |
+| `name` | name of the label. |
```
maliberty, line 670: What are the defaults?

src/gui/src/gui.cpp
```diff
@@ -186,6 +187,87 @@ static void resetConversions()
       = [](const std::string& value, bool*) { return 0; };
 }
 
+std::map<std::string, Painter::Color> Painter::colors()
```
maliberty, line 190: Why aren't Painter methods in painter.cpp?  (same below)

src/gui/README.md
```diff
@@ -644,6 +644,54 @@ To remove all the rulers:
...
+    [color]
+    [size]
+    [name]
+```
+
+#### Options
+
+| Switch Name | Description |
+| ---- | ---- |
+| `x, y,` | point of the label in microns. |
```
maliberty, line 666: The center point of the label (I guessed it would be the lower-left).  It might be nice to have an alignment option.

maliberty:
Clicking the trash icon in the inspector doesn't delete the label from the layout

src/gui/src/renderThread.cpp
```diff
@@ -431,6 +442,38 @@ void RenderThread::drawRulers(Painter& painter, const Rulers& rulers)
...
 }
 
+void RenderThread::drawLabels(Painter& painter, const Labels& labels)
+{
+  if (!viewer_->options_->areLabelsVisible()) {
+    return;
+  }
+
+  for (auto& label : labels) {
+    painter.saveState();
```
maliberty, line 452: Save and restore can be outside the loop

src/gui/src/mainWindow.cpp
```diff
@@ -1128,6 +1133,55 @@ void MainWindow::addHighlighted(const SelectionSet& highlights,
...
+                                text,
+                                anchor.value_or(Painter::Anchor::CENTER),
+                                color.value_or(gui::Painter::white),
+                                size,
+                                std::move(name));
+  std::string new_name = new_label->getName();
+
+  // check if ruler name is unique
+  for (const auto& ruler : rulers_) {
+    if (new_name == ruler->getName()) {
```
maliberty, line 1155: copy-paste error?  Should be checking labels

src/gui/src/label.cpp
```diff
@@ -0,0 +1,242 @@
...
+             return LabelDescriptor::editPoint(
+                 value, dbu_per_uu_, label->getPt(), true);
+           })},
+          {"y", makeEditor([label, dbu_per_uu_](const std::any& value) {
+             return LabelDescriptor::editPoint(
+                 value, dbu_per_uu_, label->getPt(), false);
+           })}};
+}
+
+bool LabelDescriptor::editPoint(std::any value,
```
maliberty, line 162: I can't edit the x/y values and get them to stick.  When I hit enter the just return to their previous values.

maliberty (changes requested):

maliberty:
I want to keep QFont out of the header if possible.  One idea is to have a simpler `PainterFont` that just holds the family name and size.  Internally we can convert that to a QFont when rendering.  It would make it possible for other parts of the code to use fonts as well.  We would have to down convert the QFontDialog result.

maliberty:
Do you want to persist the label font in settings?

src/gui/src/label.cpp
gadfort: This seems to be true for the rulers as well

src/gui/README.md
gadfort: I'll add a proper TCL cmd

gadfort:
> Do you want to persist the label font in settings?

I figured we preserve the other fonts so no reason to not store it, but if you feel different, I can switch it to be non-persistent

maliberty:
> I figured we preserve the other fonts so no reason to not store it, but if you feel different, I can switch it to be non-persistent

I was thinking it wasn't persisted - did I miss it in the PR?

src/gui/src/displayControls.cpp
```diff
@@ -704,6 +711,7 @@ void DisplayControls::readSettings(QSettings* settings)
   settings->beginGroup("font");
   getFont(pin_markers_font_, "pin_markers");
   getFont(ruler_font_, "ruler");
+  getFont(label_font_, "label");
```
gadfort, line 714: @maliberty  label font stored here

maliberty (approved):

Issue #7305, opened 2025-05-07
Remove debug output(?) [WARNING DRT-0088] Exhaustive access point generation for io_outs_down_mult.final_adder.U$194/B (AND2x2_ASAP7_75t_R) is unsatisfactory.


oharboe:
### Description

Is this debug output?

untar and run [detail-route-warnings.tar.gz](https://github.com/user-attachments/files/20082635/detail-route-warnings.tar.gz)

Standalone version of:

```
make DESIGN_CONFIG=designs/asap7/mock-array/config.mk route
```

```
OpenROAD v2.0-21262-g5df41f35ec 
[deleted]
[INFO DRT-0165] Start pin access.
[WARNING DRT-0088] Exhaustive access point generation for io_outs_down_mult.final_adder.U$194/B (AND2x2_ASAP7_75t_R) is unsatisfactory.
[WARNING DRT-0088] Exhaustive access point generation for io_outs_down_mult.final_adder.U$198/B (AND2x2_ASAP7_75t_R) is unsatisfactory.
```

### Suggested Solution

Fix warning or if there is nothing the user can or should do with this, make it debug output. 

If the user should do something, give eh user a clue as to what to do. Is this a DRC error? Is it obvious to anyone in skilled in the art?

PR #7351, opened 2025-05-12
drt: temporary deletion of far from edge soft requirement

bnmfw:
Helps #7305.

Takes the far from edge soft requirement out of PA while a better requirement structure is not implemented. This does not remove warning 88 completely, it will show for not enough sparse points but this is much more rare then far from edge.


bnmfw:
Thought the problem were just with logs, have to change some defok files.

maliberty (approved):

PR #7353, opened 2025-05-12
Fix leak in hasMultipleOutputs

Faholan:


src/rsz/src/Resizer.cc
```diff
@@ -2313,10 +2313,12 @@ bool Resizer::hasMultipleOutputs(const Instance* inst)
     if (network_->direction(pin)->isAnyOutput() && network_->net(pin)) {
       output_count++;
       if (output_count > 1) {
+        delete pin_iter;
```
maliberty, line 2316: This is better fixed with a unique_ptr
```
std::unique_ptr<InstancePinIterator> pin_iter(network_->pinIterator(inst));
```
Then it will be cleaned up regardless of subsequent logic.  See Resizer::findOverdrivenNets for example.

maliberty:
Needs a clang-format

PR #7354, opened 2025-05-12
Fix leak in Spef

Faholan:


src/rcx/src/extSpef.cpp
```diff
@@ -87,6 +87,13 @@ extSpef::~extSpef()
   delete _nodeCoordParser;
   free(_msgBuf1);
   free(_msgBuf2);
+
+  for (uint ii = 0; ii < 1024; ii++) {
```
maliberty, line 91: It would be better to use `_hcnrc->getSize()` than `1024` in case it ever gets resized.

maliberty (approved):

Issue #7349, opened 2025-05-12
`bazelisk build -c dbg :openroad --keep_going` fails to link


oharboe:
### Description

```
$ bazelisk build -c dbg :openroad --keep_going
INFO: Analyzed target //:openroad (0 packages loaded, 0 targets configured).
ERROR: /home/oyvind/OpenROAD-flow-scripts/tools/OpenROAD/BUILD.bazel:115:10: Linking openroad failed: (Exit 1): cc_wrapper.sh failed: error executing CppLink command (from target //:openroad) external/toolchains_llvm~~llvm~llvm_toolchain/bin/cc_wrapper.sh @bazel-out/k8-dbg/bin/openroad-2.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
ld.lld: error: undefined symbol: std::__glibcxx_assert_fail(char const*, int, char const*, char const*)
>>> referenced by atomic_base.h:498 (/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:498)
>>>               bazel-out/k8-dbg/bin/external/edu_berkeley_abc/_objs/abc-lib/utilPth.pic.o:(std::atomic<bool>::load(std::memory_order) const)
>>> referenced by atomic_base.h:499 (/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:499)
>>>               bazel-out/k8-dbg/bin/external/edu_berkeley_abc/_objs/abc-lib/utilPth.pic.o:(std::atomic<bool>::load(std::memory_order) const)
>>> referenced by atomic_base.h:473 (/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:473)
>>>               bazel-out/k8-dbg/bin/external/edu_berkeley_abc/_objs/abc-lib/utilPth.pic.o:(std::atomic<bool>::store(bool, std::memory_order))
>>> referenced 2 more times
clang: error: linker command failed with exit code 1 (use -v to see invocation)
INFO: Found 1 target...
Target //:openroad failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 6.029s, Critical Path: 5.76s
INFO: 2 processes: 2 internal.
ERROR: Build did NOT complete successfully
```

oharboe:
[...]

PR #7358, opened 2025-05-13
bazel: stopgap way to create debug build

oharboe:
This produces an openroad build now:
    
    bazelisk build --cxxopt=-stdlib=libstdc++ --linkopt=-lstdc++ -c dbg :openroad

Whereas this [still fails](https://github.com/The-OpenROAD-Project/OpenROAD/issues/7349):

    bazelisk build -c dbg :openroad

### DIFF
--- a/src/odb/src/def/BUILD
+++ b/src/odb/src/def/BUILD
@@ -54,7 +54,7 @@ cc_library(
     ],
     hdrs = DEF_HEADERS,
     # Ignore warnings in generated code
-    copts = ["-Wno-unused-but-set-variable"],
+    copts = ["-Wno-unused-but-set-variable", "-Wno-modules-import-nested-redundant"],
     includes = ["def"],
 )
 


oharboe:
@QuantamHD Any ideas why this works?

src/odb/src/def/BUILD
```diff
@@ -54,7 +54,7 @@ cc_library(
     ],
     hdrs = DEF_HEADERS,
     # Ignore warnings in generated code
-    copts = ["-Wno-unused-but-set-variable"],
+    copts = ["-Wno-unused-but-set-variable", "-Wno-modules-import-nested-redundant"],
```
maliberty, line 57: what is this warning?

maliberty (approved):

src/odb/src/def/BUILD
oharboe: Unclear. Something about bcr bison hardcoding YYDEBUG to 1 and this causing problems with include if stdio.h and C++ modules due to some namespace scoping or something...

PR #7362, opened 2025-05-13
gpl: callbacks should not touch filler gcells

gudeh:
The logic was previously left this way so that the same function could potentially be used for filler removal also. However, in the implementation of filler GCell removal (currently under development and not yet merged), I chose to create a new function.

In summary, when callbacks are triggered for destroying instances, it is not expected to also trigger the destruction of filler GCells, which are internal to GPL only.

### DIFF
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -3107,8 +3107,15 @@ void NesterovBase::destroyCbkGCell(odb::dbInst* db_inst)
     size_t gcell_index = db_it->second;
 
     GCellHandle& handle = nb_gcells_[gcell_index];
-    // TODO delete fillers
-    bool is_filler = handle->isFiller();
+
+    if(handle->isFiller()) {
+      debugPrint(log_,
+                GPL,
+                "callbacks",
+                1,
+                "error: trying to destroy filler gcell during callback!");
+      return;
+    }
 
     if (gcell_index != last_index) {
       std::swap(nb_gcells_[gcell_index], nb_gcells_[last_index]);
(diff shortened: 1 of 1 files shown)


maliberty:
Needs a clang-format

gudeh:
@povik FYI

gudeh:
I wasn't able to test this because the testing case is with @povik branch which is currently with conflicts when merging with master.

src/gpl/src/nesterovBase.cpp
```diff
@@ -3118,13 +3125,9 @@ void NesterovBase::destroyCbkGCell(odb::dbInst* db_inst)
...
-        odb::dbInst* swapped_inst
-            = nb_gcells_[gcell_index]->insts()[0]->dbInst();
-        db_inst_to_nb_index_map_.erase(swapped_inst);
-        db_inst_to_nb_index_map_[swapped_inst] = gcell_index;
-      }
-
+      odb::dbInst* swapped_inst = nb_gcells_[gcell_index]->insts()[0]->dbInst();
+      db_inst_to_nb_index_map_.erase(swapped_inst);
+      db_inst_to_nb_index_map_[swapped_inst] = gcell_index;
       std::pair<odb::dbInst*, size_t> replacer = nbc_->destroyCbkGCell(db_inst);
```
povik, line 3131: @gudeh Is it correct that this call is conditioned on `gcell_index != last_index`?

gudeh: Ah, well noticed, this condition is blocking the call on `nbc_->destroyCbkGCell(db_inst)`. I fixed the closing bracket position.

I believe it is correct now. Unfortunately I could not find a way to properly test this. Do you have a suggestion on how I could do so?

gudeh:
What happens here is that we replace the element we wish to destroy with the last one. After the pop back we need to update the element position with the previously last element data. This is only required when the element to be removed is not the last one.

gudeh:
Furthermore, I noticed another problem which should not trigger a seg fault but would provide incorrect results. There is a parallel vector which belongs to RouteBase which also needs to be managed when an instance is removed. I am thinking of moving such vector to NesterovBase instead of RouteBase, this way we maintain all parallel vectors in NesterovBase.

maliberty:
Please advise when this is ready to merge

povik:
I have merged this fix into my working branch and it addressed all known crashes. Since this code path isn't exercised in unpatched openroad (there are no instance removals from timing-driven iterations) I believe it's safe to merge. @gudeh are you fine with merging?

gudeh:
I understand it is good to merge since it at least fixes a crash. The other modification I mentioned can go on a another PR

PR #7365, opened 2025-05-13
Fix leak in RepairDesign

Faholan:


maliberty (approved):

src/rsz/src/RepairDesign.cc
```diff
@@ -55,14 +55,18 @@ RepairDesign::RepairDesign(Resizer* resizer) : resizer_(resizer)
...
+  delete pre_checks_;
+}
 
 void RepairDesign::init()
 {
   logger_ = resizer_->logger_;
   dbStaState::init(resizer_->sta_);
   db_network_ = resizer_->db_network_;
   dbu_ = resizer_->dbu_;
+  delete pre_checks_;
```
QuantamHD, line 69: Could we make this unique_ptr since assigning to a unique_ptr will auto free the currently store object. That way we can avoid specifying an explicit destructor for this class.

maliberty (approved):

PR #7367, opened 2025-05-13
Cleanup dbSta fields

Faholan:


maliberty:
It would be an error to call these functions more than once.  If you want to check that the pointers are null and error if they aren't that is ok.

Faholan:
Ah alright. Knowing, that, I have located the actual issue that caused the `makeComponents` function to be called twice instead of once

Faholan:
The `initVars` function already calls `makeComponents`, so the extra call should be removed

src/dbSta/src/dbSta.cc
```diff
@@ -231,7 +231,6 @@ void dbSta::unregisterStaState(dbStaState* state)
 std::unique_ptr<dbSta> dbSta::makeBlockSta(odb::dbBlock* block)
 {
   auto clone = std::make_unique<dbSta>();
-  clone->makeComponents();
```
maliberty, line 234: Where is the other call?

Faholan: It's the line right after, `clone->initVars()` :

```c++
void dbSta::initVars(Tcl_Interp* tcl_interp,
                     odb::dbDatabase* db,
                     utl::Logger* logger)
{
  db_ = db;
  db->addObserver(this);
  logger_ = logger;
  makeComponents();
  if (tcl_interp) {
    setTclInterp(tcl_interp);
  }
  db_report_->setLogger(logger);
  db_network_->init(db, logger);
  db_cbk_ = std::make_unique<dbStaCbk>(this, logger);
  buffer_use_analyser_ = std::make_unique<BufferUseAnalyser>();
}
```

maliberty (approved):

PR #7369, opened 2025-05-13
Correctly delete leaking fields

Faholan:


maliberty (approved):

maliberty:
I'm very glad to have the cleanups but in almost all cases where you are writing `delete` it would be better to upgrade the code to `unique_ptr`.   Most cases require no attention to deletion. 
 Such manual memory management is error prone as you are demonstrating.

PR #7370, opened 2025-05-14
odb: move to flat_map in iterm to save 24 bytes

QuantamHD:


src/odb/src/db/dbITerm.h
```diff
@@ -55,7 +56,7 @@ class _dbITerm : public _dbObject
   dbId<_dbITerm> _next_modnet_iterm;
   dbId<_dbITerm> _prev_modnet_iterm;
   uint32_t _sta_vertex_id;  // not saved
-  std::map<dbId<_dbMPin>, dbId<_dbAccessPoint>> aps_;
+  boost::container::flat_map<dbId<_dbMPin>, dbId<_dbAccessPoint>> aps_;
```
maliberty, line 59: Requires update the schema id and providing backwards compatibility in reading old odb files.

QuantamHD: I'm not clear that this does require that. The schema for maps is writing out the size then writing the pairs of values. It's the same format here. There should be no change, and old odbs should read normally

maliberty: Nice!

maliberty:
Looks like it just needs a tidy

maliberty (approved):

PR #7376, opened 2025-05-14
bazel: garbage collect cache

oharboe:
try it out...

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -67,4 +67,8 @@ build --incompatible_strict_action_env
 #build --remote_cache=https://storage.googleapis.com/megaboom-bazel-artifacts
 #build --remote_cache_compression=true
 
+# Tune these to fit our needs.
+build --experimental_disk_cache_gc_max_size=100G
+build --experimental_disk_cache_gc_max_age=7d
+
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty Seems to do no harm, reenable CI and try it out?

.bazelrc
```diff
@@ -67,4 +67,8 @@ build --incompatible_strict_action_env
 #build --remote_cache=https://storage.googleapis.com/megaboom-bazel-artifacts
 #build --remote_cache_compression=true
 
+# Tune these to fit our needs.
+build --experimental_disk_cache_gc_max_size=100G
```
maliberty, line 71: "By default, public shared Github runners come with around 25-29 GB of free disk space to be consumed by your build job." so this threshold is too high.  ORFS itself is going to need a bunch of that space.

oharboe: So then we might as well leave it at 100gbyte and wait for a more suitable runner?

oharboe:
@vvbandeira @maliberty We are using this setting in-house so we are gaining some experience with it.

Closing for now, easy to configure as CI build resources are set up for bazel.

Issue #5214, opened 2024-06-08
CI for no differences in .ok file


rovinski:
### Description

As a simple sanity check, there could be a CI test that any modified `.ok` files do not have differences in them. This will ensure that they don't get missed on review.

### Suggested Solution

A simple script which greps for "Differences found at line" should suffice.

maliberty:
@vvbandeira could this be a GH action?

vvbandeira:
> @vvbandeira could this be a GH action?

I think Jenkins should be fine. It would be a simple `grep "Differences found at line" **/*.ok` or `for f in **/*.ok; do ...` plus some minor parsing/substitutions just to print the names of the tests in a nice format.

rovinski:
Resolved by #7389

PR #7389, opened 2025-05-15
Fix ok files and add workflow to catch issues

gadfort:
I found a couple of "Differences found ..." I fixed them and added a workflow to try and prevent these again.


maliberty (approved):

.github/workflows/github-actions-are-differences-found.yml
```diff
@@ -0,0 +1,25 @@
...
+  format:
+    runs-on: ${{ vars.USE_SELF_HOSTED == 'true' && 'self-hosted' || 'ubuntu-latest' }}
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Check ok files
+        run: |
+          find -name "*.ok" | xargs grep -n "Differences found " > diffs
```
maliberty, line 21: Find need a directory argument eg `find . -name ....`

rovinski:
I think finally closes #5214 🚀

PR #7391, opened 2025-05-16
Delete LUT

Faholan:
Correctly delete LUT.

A better solution I'd prefer would imply using shared_ptr, however I'm unsure whether they work correctly with struct arrays, and in `readLUTfiles` `malloc` is used instead of `new[]`, which is why I went with manually tracking the allocations

### DIFF
--- a/src/stt/src/flt/flute.cpp
+++ b/src/stt/src/flt/flute.cpp
@@ -74,7 +74,6 @@ static void readLUTfiles(LUT_TYPE LUT, NUMSOLN_TYPE numsoln)
 {
   unsigned char charnum[256], line[32], *linep, c;
   FILE *fpwv, *fprt;
-  struct csoln* p;
   int d, i, j, k, kk, ns, nn;
 
   for (i = 0; i <= 255; i++) {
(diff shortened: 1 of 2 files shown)


maliberty:
You can use a shared_ptr with an array of objects.  The objects being structs is fine.  From [std::shared_ptr](http://en.cppreference.com/w/cpp/memory/shared_ptr): std::shared_ptr<float[]> sp3 = std::make_shared<float[]>(64);

malloc can be replaced by new as needed (with matching deallocation changes).

Where would you use the shared_ptr?

maliberty:
Better still would be  `boost::multi_array<std::shared_ptr<struct csoln[]>, 2>` so the whole thing wouldn't need manual memory management

Faholan:
I'm gonna look into implementing a solution based on that

maliberty:
ready for review?

Faholan:
Yes, everything should be working fine now

src/stt/src/flt/flute.cpp
```diff
@@ -114,12 +114,12 @@ static void readLUTfiles(LUT_TYPE LUT, NUMSOLN_TYPE numsoln)
...
         numsoln[d][k] = numsoln[d][kk];
-        LUT[d][k] = LUT[d][kk];
+        (*LUT)[d][k] = (*LUT)[d][kk];
       } else {
         fgetc(fpwv);  // '\n'
         numsoln[d][k] = ns;
-        p = (struct csoln*) malloc(ns * sizeof(struct csoln));
-        LUT[d][k] = p;
+        p = new struct csoln[ns];
+        (*LUT)[d][k] = std::shared_ptr<struct csoln[]>(p);
```
maliberty, line 122: Better to use std::make_shared

Faholan: If I remember correctly I tried exactly that, but had issues making `std::make_shared` behave as I wanted, which is why I decided to use that solution, since in any case I need access to the raw pointer for the rest of the code to work

maliberty: I pushed a change that I think is an improvement

maliberty (approved):

PR #7398, opened 2025-05-18
ord: add gzip reading and writing to ord::writeDb

gadfort:
Adds:
- gzip reading and writing to odb, in a similar interface as opensta commands like `write_sdc`.

This allows users to select if compression should be used avoiding the runtime penalty during writing if its not requested.

The change to the scope code in `pushScope` is due to the gzip writing not supporting `pos()`.


stefanottili:
Couple of thoughts in order of ease of implementation:
1. Please give gzip compression level 1 a try.
For gds/oas/def one gains ~2x read/write speed with only 10% larger .gz files.
2. https://github.com/cloudflare/zlib contains several zlib speed improvements
3. Why a switch instead of using the .gz extension as a hint ?
write_db [-gzip] filename  instead of write_db <name>.odb.gz ?
4. To make the write fast, have a look at pigz.
It's multithreaded writing scales nicely with # of cores.
5. In order to allow multithreaded reading odb would need to gain an offset_table (like oasis).

oharboe:
compressing .odb files have already been discussed and discarded https://github.com/The-OpenROAD-Project/OpenROAD/pull/4261

What is different this time?

oharboe:
Adding a specific compression format to OpenROAD is a point solution that OpenROAD will have to support forever.

I think this is, and measurements in #4261 support this, a concern for the build systems to handle, such as Bazel or SiliconCompiler.

gadfort:
@oharboe the difference between this and #4261 is that that PR always used compression, while this makes it a choice, so we dont have to worry about the runtime issues raised in that PR (if the build system wants to support this then great, but it's not being forced upon anyone).

oharboe:
> @oharboe the difference between this and #4261 is that that PR always used compression, while this makes it a choice, so we dont have to worry about the runtime issues raised in that PR (if the build system wants to support this then great, but it's not being forced upon anyone).

If this feature must be added, I want to be clear that my advice is not to — I believe it's misplaced in OpenROAD. This functionality belongs in build systems, not in EDA tools.

If we're forced to include it, then I would suggest relying on the file extension to distinguish compressed files, and simply do nothing for uncompressed ones. This way, the code path remains dormant for those not using compression, minimizing its footprint and impact.

Also, the fact that we're already heading into a debate over which compression format to use only reinforces my point: this is a distraction. OpenROAD should steer well clear of this and let build systems handle it, where it naturally belongs.

gadfort:
@stefanottili I did not mess with the default compression (it's hard to gauge the impacts on speed vs. filesize without a good diversity of files), I agree that triggering on the extension would be a good idea, the `-gzip` is to mirror the opensta commands which use that flag, but if just triggering on the extension would be simpler.

@oharboe I'm not sure why you are framing this as "forced to include this", gzipping files is common and being able to read them without extra steps is nice to have. Again, this is optional to use, so should not impact anything if you don't want to use it.

oharboe:
@gadfort fixed wording.

maliberty:
I agree that using the extension is preferrable to -gzip.  This needs to be documented.  Unit test(s) are needed.

As we support gzip elsewhere I am ok with adding this an option.  I don't want to pull in more dependencies though  - let's just take what Boost offers.   The cost of owning this code is pretty low in this form.

gadfort:
@maliberty I switched writing to be based on the extension and added a test. I wasn't sure how to test this, so I just write the DB twice and compare their md5sums (not a huge fan of this, but I didn't know how else to test this).

oharboe (changes requested):
Modify it to only decompress if there is a .gz extension to leave current codepath unchanged.

gadfort:
> Modify it to only decompress if there is a .gz extension to leave current codepath unchanged.

done

oharboe (approved):
No effect on existing use cases 👍

maliberty (approved):

stefanottili:
This broke the linking openroad on MacOS M1.
/opt/homebrew/lib/libboost_iostreams.a is missing from the openroad link command, adding it makes it work.

stefanottili:
adding `-L/opt/homebrew/opt/zstd/lib` to the link command makes `-lzstd` find it, but now I get
`ld: library 'icudata' not found`.
https://github.com/Homebrew/homebrew-core/issues/67427 shows that icu4c/lib has to be added to the link path
-L/opt/homebrew/opt/zstd/lib -L/opt/homebrew/opt/icu4c/lib

PR #7403, opened 2025-05-19
utl: Fix include for stdlib exit function in decode

QuantamHD:


maliberty (approved):

QuantamHD:
Google internally must have stricter rules for includes. It fails with exit not defined when I build it within google3.

PR #7407, opened 2025-05-21
upf: add missing port in upf tests

### LINKED ISSUE #7363: [bug] Error in some UPF files.
### Description

Hello!

I am using a custom UPF interpreter to check correctness of source UPF specifications.

A port with the name `power_down_aes_2` is used [here](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/test/upf/mpd_aes.upf#L23), [here](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/upf/test/write.upfok#L36), [here](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/upf/test/data/mpd_top/mpd_top.upf#L23) and [here](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/upf/test/data/mpd_top/mpd_top_combined.upf#L23) but it is never created in any of those files. According to the standard for UPF I believe that is a mistake. As far as I can see the power switch created using this port is not used, but, for example, Questa 2024.3 after analyzing the design with a mock .sv file results in this error.

```
** Error (suppressible): /home/runner/upfsvtest.upf(16): UPF: (vopt-9655) Control port: 'power_down_aes_2' doesn't exist in scope: 'topt'.
```

Is this a bug or is it intended?

### Suggested Solution

Append `create_logic_port power_down_aes_2 -direction in`

### Additional Context

_No response_

LucasYuki:
Closes #7363.

Add missing port in upf tests.

### DIFF
--- a/src/upf/test/data/mpd_top/mpd_top.upf
+++ b/src/upf/test/data/mpd_top/mpd_top.upf
@@ -8,6 +8,7 @@ create_power_domain PD_AES_2 \
   -elements {u_aes_2} \
 
 create_logic_port power_down_aes_1 -direction in
+create_logic_port power_down_aes_2 -direction in
 
 create_power_switch psw_aes_1 \
   -domain PD_AES_1 \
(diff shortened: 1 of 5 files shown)


maliberty (approved):

maliberty:
Is there some warning or error that should have been generated?

LucasYuki:
We currently don't check for UPF errors in general, but I think it should generate a warning or error.

PR #7410, opened 2025-05-21
Downsize move

mguthaus:
Added a new move to downsize off-critical gates to improve setup slack. It respects timing by trying to not make the off-ciritical fanouts into critical nets. In general, this reduces area in many cases, but can have no impact on some other cases.


maliberty:
Ready for review?

mguthaus:
I think so

maliberty:
There are still some CI failures to resolve.

src/rsz/src/SizeDownMove.cc
```diff
@@ -0,0 +1,326 @@
...
+using sta::Slack;
+using sta::Slew;
+using sta::Vertex;
+using sta::VertexOutEdgeIterator;
+
+bool SizeDownMove::doMove(const Path* drvr_path,
+                          int drvr_index,
+                          Slack drvr_slack,
+                          PathExpanded* expanded,
+                          float setup_slack_margin)
```
maliberty, line 41: Make args const when possible.   (Note only do so in the definition not the declaration for by-value types).  Same in sizeup

src/rsz/src/SizeDownMove.cc
```diff
@@ -0,0 +1,326 @@
...
+  Pin* drvr_pin = drvr_path->pin(this);
+  Vertex* drvr_vertex = drvr_path->vertex(sta_);
+  const Path* load_path = expanded->path(drvr_index + 1);
+  Vertex* load_vertex = load_path->vertex(sta_);
+  Pin* load_pin = load_vertex->pin();
+
+  const int fanout = this->fanout(drvr_vertex);
+  if (fanout < 2) {
+    return false;
+  }
```
maliberty, line 52: Why not?

src/rsz/src/SizeDownMove.cc
```diff
@@ -0,0 +1,326 @@
...
+  const bool tristate_drvr = resizer_->isTristateDriver(drvr_pin);
+  if (tristate_drvr) {
+    return false;
+  }
+  const Net* net = db_network_->dbToSta(db_network_->flatNet(drvr_pin));
+  if (resizer_->dontTouch(net)) {
+    return false;
+  }
+  dbNet* db_net = db_network_->staToDb(net);
+  if (db_net->isConnectedByAbutment()) {
```
maliberty, line 62: Similar check in sizeup?

src/rsz/src/SizeDownMove.cc
```diff
@@ -0,0 +1,326 @@
...
+  Pin* load_pin = load_vertex->pin();
+
+  const int fanout = this->fanout(drvr_vertex);
+  if (fanout < 2) {
+    return false;
+  }
+  const bool tristate_drvr = resizer_->isTristateDriver(drvr_pin);
+  if (tristate_drvr) {
+    return false;
+  }
```
maliberty, line 56: Why can't a tristate be downsized?  Afaik only buffer insertion after a tristate is a problem

src/rsz/src/SizeDownMove.cc
```diff
@@ -0,0 +1,326 @@
...
+  const Net* net = db_network_->dbToSta(db_network_->flatNet(drvr_pin));
+  if (resizer_->dontTouch(net)) {
+    return false;
+  }
```
maliberty, line 60: dbInst::swapMaster doesn't change the net so this shouldn't matter.  Only a dont-touch on the instance should matter.

maliberty (changes requested):

src/rsz/README.md
```diff
@@ -355,9 +356,10 @@ repair_timing
 | `-setup_margin` | Add additional setup slack margin. |
 | `-hold_margin` | Add additional hold slack margin. |
 | `-allow_setup_violations` | While repairing hold violations, buffers are not inserted that will cause setup violations unless `-allow_setup_violations` is specified. |
-| `-sequence` | Specify a particular order of setup timing optimizations. The default is "unbuffer,buffer,swap,size,clone,split". Ignores skip flags when used. |
+| `-sequence` | Specify a particular order of setup timing optimizations. The default is "unbuffer,buffer,swap,sizeup,clone,split". Obeys skip flags also. |
```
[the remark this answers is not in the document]
mguthaus: Do we want to not have the default with sizedown?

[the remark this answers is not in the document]
maliberty: I was pointing out the the description of the default doesn't seem to match what actually happens.  I'm fine to have it but we the docs to match.

src/rsz/src/SizeDownMove.cc
mguthaus: Good point. I'll check how often this happens, but it may require some different logic since this path is critical and has negative slack. We are only considering paths that have positive slack right now.

src/rsz/src/SizeDownMove.cc
mguthaus: Yes, this was used elsewhere for almost all the optimizations.

src/rsz/src/SizeDownMove.cc
mguthaus: This interface will be changed in my next PR. The current one was copied from the moves prior to my refactor. I'd like to defer this until my next PR.

src/rsz/src/SizeDownMove.cc
maliberty: It should really only apply to optimizations that want to insert a buffer (not possible) but it won't hurt here except for some wasted runtime.

maliberty (approved):

maliberty:
Not currently active due to increased ERCs.

src/rsz/test/repair_setup2_multi.ok
```diff
@@ -14,15 +14,15 @@ Iteration |   Area    | Resized | Buffers | Nets repaired | Remaining
...
+[INFO RSZ-0100] Repair move sequence: UnbufferMove SizeUpMove SwapPinsMove BufferMove CloneMove SplitLoadMove 
 [INFO RSZ-0094] Found 2 endpoints with setup violations.
 [INFO RSZ-0099] Repairing 2 out of 2 (100.00%) violating endpoints...
    Iter   | Removed | Resized | Inserted | Cloned |  Pin  |   Area   |    WNS   |   TNS      |  Viol  | Worst
           | Buffers |  Gates  | Buffers  |  Gates | Swaps |          |          |            | Endpts | Endpt
 --------------------------------------------------------------------------------------------------------------
         0 |       0 |       0 |        0 |      0 |     0 |    +0.0% |   -0.118 |       -0.2 |      2 | y2
         5 |       0 |       0 |        0 |      0 |     0 |    +0.0% |   -0.118 |       -0.2 |      2 | y2
-       10 |       0 |       3 |        0 |      0 |     1 |   +53.7% |   -0.118 |       -0.2 |      2 | y2
+       10 |       0 |       0 |        0 |      0 |     0 |   +53.7% |   -0.118 |       -0.2 |      2 | y2
```
rovinski, line 25: This seems suspicious, no? 0 changes but +53% area?

src/rsz/test/repair_setup2.ok
```diff
@@ -14,15 +14,15 @@ Iteration |   Area    | Resized | Buffers | Nets repaired | Remaining
...
+[INFO RSZ-0100] Repair move sequence: UnbufferMove SizeUpMove SwapPinsMove BufferMove CloneMove SplitLoadMove 
 [INFO RSZ-0094] Found 2 endpoints with setup violations.
 [INFO RSZ-0099] Repairing 2 out of 2 (100.00%) violating endpoints...
    Iter   | Removed | Resized | Inserted | Cloned |  Pin  |   Area   |    WNS   |   TNS      |  Viol  | Worst
           | Buffers |  Gates  | Buffers  |  Gates | Swaps |          |          |            | Endpts | Endpt
 --------------------------------------------------------------------------------------------------------------
         0 |       0 |       0 |        0 |      0 |     0 |    +0.0% |   -0.118 |       -0.2 |      2 | y2
         5 |       0 |       0 |        0 |      0 |     0 |    +0.0% |   -0.118 |       -0.2 |      2 | y2
-       10 |       0 |       3 |        0 |      0 |     1 |   +53.7% |   -0.118 |       -0.2 |      2 | y2
+       10 |       0 |       0 |        0 |      0 |     0 |   +53.7% |   -0.118 |       -0.2 |      2 | y2
```
rovinski, line 25: Same

src/rsz/test/repair_setup2_multi.ok
mguthaus: There is an existing bug in the area report, but this may be another. Right now, the area includes moves that aren't "accepted" but pending whereas the other stats are only the accepted moves.

mguthaus: Earlier implementations would present data that could be reversed later. I agree it should be consistent though. This isn't limited to my PR but can happen in general.

mguthaus: FYI
https://github.com/The-OpenROAD-Project/OpenROAD/pull/7533

PR #7411, opened 2025-05-21
Free allocated SWIG input

Faholan:


maliberty:
I prefer to remove this mess rather than add to it (see https://github.com/The-OpenROAD-Project/OpenROAD/pull/7416)

Issue #7375, opened 2025-05-14
Segfault in `rcx::extDistWidthRCTable::~extDistWidthRCTable()`


gadfort:
### Describe the bug

Segfault in parasitic extraction.

### Expected Behavior

No segfault

### Environment

```shell
Check https://github.com/The-OpenROAD-Project/OpenROAD/compare/5932228c68f463ad701f2deacdb38c5c0ee120a8...f9cfe3ce45b77a2bb7ef9b34c8909c9770da9b5c for changes
```

### To Reproduce

https://drive.google.com/file/d/1-5pdFo-5iyK3HcJEnPg_vBrlZO6ju6X_/view?usp=sharing

```
tar xvf sc_issue_ethmac_job0_asap7_asap7sc7p5t_rvt_write.views0_20250514-011923.tar.gz
cd sc_issue_ethmac_job0_asap7_asap7sc7p5t_rvt_write.views0_20250514-011923
./run.sh
```

### Relevant log output

```shell
| ERROR    | job0_asap7_asap7s... | write.views          | 0 | Signal 11 received
| ERROR    | job0_asap7_asap7s... | write.views          | 0 | Stack trace:
| ERROR    | job0_asap7_asap7s... | write.views          | 0 |  0# 0x000055F29DBAE802 in /sc_tools/bin/openroad
| ERROR    | job0_asap7_asap7s... | write.views          | 0 |  1# 0x00007F9090A1D520 in /lib/x86_64-linux-gnu/libc.so.6
| ERROR    | job0_asap7_asap7s... | write.views          | 0 |  2# rcx::extDistWidthRCTable::~extDistWidthRCTable() in /sc_tools/bin/openroad
| ERROR    | job0_asap7_asap7s... | write.views          | 0 |  3# rcx::extMetRCTable::~extMetRCTable() in /sc_tools/bin/openroad
| ERROR    | job0_asap7_asap7s... | write.views          | 0 |  4# rcx::extRCModel::~extRCModel() in /sc_tools/bin/openroad
| ERROR    | job0_asap7_asap7s... | write.views          | 0 |  5# rcx::extMain::makeBlockRCsegs(char const*, unsigned int, unsigned int, double, bool, double, int, char const*) in /sc_tools/bin/openroad
| ERROR    | job0_asap7_asap7s... | write.views          | 0 |  6# 0x000055F29E3366D6 in /sc_tools/bin/openroad
[...]

PR #7414, opened 2025-05-22
Fix leak in model table

Faholan:
This is #7355 once more, fixing #7375 that was caused by one of the `extDistWidthRCTable` constructors returning early and not initializing all the fields.

This also fixes `extMetRCTable::allocateInitialTables` that in practice must always be called with `_layerCnt` as argument, so it is unnecessary.


src/rcx/src/extRCmodel.cpp
```diff
@@ -760,7 +760,7 @@ extDistWidthRCTable::extDistWidthRCTable(bool over,
   _firstDiagDist = nullptr;
   _lastDiagDist = nullptr;
 
-  for (jj = 0; jj < 16; jj++) {
+  for (jj = 0; jj < 32; jj++) {
```
maliberty, line 763: Why is 32 more correct than 16?  What do either represent?

This value is to be used in multiple places so it should be a named constexpr static class member.

src/rcx/src/extRCmodel.cpp
```diff
@@ -876,12 +876,29 @@ extDistWidthRCTable::extDistWidthRCTable(bool over,
...
+
   // dkf 09202024 skip width map table when  not knowing number of widths is not
   // know in advance after reading rules of different width width mapping should
   // be re-done before writting rules
   bool skip_width_map_table = widthTable == nullptr;
 
   if (!skip_width_map_table && widthTable->getCnt() == 0) {
+    _rcDistTable = nullptr;
+    _widthTable = nullptr;
+    _rcDistTable = nullptr;
```
maliberty, line 901: Duplicate `_rcDistTable = nullptr;`

src/rcx/src/netRC.cpp
```diff
@@ -2017,7 +2015,9 @@ void extMain::makeBlockRCsegs(const char* netNames,
     }
 
   */
-  _modelTable->resetCnt(0);
+  while (_modelTable->notEmpty()) {
```
maliberty, line 2018: Do you know that all instances of extRCModel are in _modelTable?

src/rcx/src/extRCmodel.cpp
```diff
@@ -876,12 +876,29 @@ extDistWidthRCTable::extDistWidthRCTable(bool over,
   _layerCnt = layerCnt;
   _met = met;
 
+  _firstDiagWidth = nullptr;
```
hzeller, line 879: I think it would be a lot less error prone, more readable and concise if all the empty defaults would be using default member initialization (anything that is initialized to `nullptr` here is probably falling into that category).

Members that do get initialized from parameters can also get much more readable by using  the RAII-syntax (`... : _over(over), _layerCnt(layerCnt)...`).

The duplication of code between all these constructors could probably also be reduced by using delegating constructors.

src/rcx/src/netRC.cpp
Faholan: There is an instance in `Ext::rc_estimate` but I added a delete for it.

The remaining problematic instance is allocated in `Ext::init_rcx_model`, and set to `extMain::_currentModel`. The issue is that this variable can also be set to an element of `_modelTable`. I can add tracking through a boolean variable to correctly free this instance and avoid risking a double free

maliberty:
Ready for re-review?

Faholan:
Yes, I've fixed very comment, they were very useful thank you all

PR #7420, opened 2025-05-23
mpl: improve debug messages

AcKoucher:
1. Create debug table for thresholds. Example:
```
Current cluster: root
  Macros: 28
  Std Cells: 91483

    Level 1  |  Min  |  Max
-----------------------------
  Std Cells  |   900 |   5500
     Macros  |     1 |      5
```
   2. Group clustering debug functions in a section;
   3. Improve existent messages.


AcKoucher:
@joaomai FYI

PR #7433, opened 2025-05-26
Bazel llvm 20 1 0

oharboe:
Compiles... Bump?


oharboe:
> @oharboe what does updating the compile toolchain provide?

I was eliminating it as a cause for some issues and it works, so it is a choice we have.

Do you see any downsides?

gadfort:
@oharboe I was just wondering, I don't see any obvious downside, other than the churn of compilers (a fairly hefty download). In the bazel builds it seems I'm needing to update the llvm (I know I don't compile often in bazel at the moment, so this might just be a perception issue) every time I need to compile, while I don't recall the last time I updated gcc on the cmake build and updating that doesn't seem to need updates this often.

maliberty (approved):

oharboe:
@maliberty merge?

PR #7441, opened 2025-05-27
bazel: add openmp, same version as LLVM

oharboe:


oharboe:
@QuantamHD @hzeller Any ideas?

PR #7442, opened 2025-05-27
Bazel openmp fix

oharboe:
Upgrading OpenMP to match LLVM to fix linking errors so that `copts` can contain `-fopenmp`. Note that we're not adding `-fopenmp` to `linkopts` because we're linking with OpenMP that we built, not the system provided OpenMP(hermetic builds).

A bit of a hack and slash job based upon what I found in https://github.com/hdl/bazel_rules_hdl/tree/main/dependency_support/org_llvm_openmp

With this PR, I get multithreading with OpenMP

First build everything up to:

```
bazelisk build -c opt test/orfs/mock-array:MockArray_grt
```

Now run detailed routing:

```
bazelisk build -c opt test/orfs/mock-array:MockArray_route
```

And see multithreading:

image


maliberty:
@QuantamHD does it make more sense for these changes to be in the upstream?

oharboe:
> @QuantamHD does it make more sense for these changes to be in the upstream?

You mean bcr?

PR #7450, opened 2025-05-27
mpl: add lower limit for pin access depth

### LINKED ISSUE #7084: MPL pin access blockages' depth should have a lower limit
### Description

In #4682 we introduced a macro dominance factor to decrease the depth generated for pin access blockages inside MPL. One of the reasons was that for some cases we ended up generating very - unnecessarily - huge blockages. However, with the current approach some designs e.g., ngt45/bp_fe get an overly thin depth.

### Suggested Solution

Add a lower depth limit.

### Additional Context

_No response_

AcKoucher:
Resolve #7084 

Example of the table in new debug messages:
```
  Pin Access Depth (μm)  |  Min  |  Max
-----------------------------------------
             Horizontal  | 32.00 | 160.00
               Vertical  | 28.00 | 140.00
```


maliberty:
The code looks fine; just awaiting the CI

maliberty (approved):

PR #7461, opened 2025-05-29
drt: frVia takes origin during construction

bnmfw:
Just quickly did this because it was bothering me a little. Pet peeve PR.

Full Safe

`frVia` can now take an origin point during construction. We keep using `setOrigin()` when it clearly wants to be in construction. The part of the code on `FlexPA_acc_pattern.cpp` touched by this is very verbose for essentially just creating 3 `frVias`. 

I actually think `frVia` should have a constructor that just takes and access point as input but that would require each access point to know the transform from its unique instance which is not the case currently.


osamahammad21 (approved):
Nice cleanup

PR #7465, opened 2025-05-30
mpl: add wirelength metrics estimation based on nets

AcKoucher:


AcKoucher:
@maliberty @eder-matheus @osamahammad21 I was thinking, would it make sense for `report_wire_length` to have the behavior of this draft (w.r.t. to the wire length computation itself) when we don't have routing data in ODB?

eder-matheus (approved):

PR #7466, opened 2025-05-30
gpl filler gcell removal instead of density modification for routability mode

gudeh:
This PR introduces a significant change to how filler cells are handled within the global placement (GPL) routability mode. Previously, when area increased due to routability mode inflations, we compensated by increasing density. This approach was counter-intuitive and hindered the routability mode's ability to spread instances effectively. Now, filler cells are removed to directly compensate for inflated area. The previous density modification approach is only utilized when no more fillers can be removed.

A special situation arises when routing congestion doesn't reduce for three consecutive routability iterations. In such cases, instead of reverting to the conventional snapshot (0.6 overflow), we revert to the inflation values that yielded the minimum observed routing congestion (i.e., the best solution found). When this occurs, previously removed fillers are restored using NesterovBase::restoreRemovedFillers().

Currently, this filler removal modification is exclusive to the routability mode. It has not been extended to the timing-driven mode due to consistent divergences observed on three Nangate45 designs when attempting to remove fillers after non-virtual timing-driven iterations.

This PR also includes the following modifications:

    -Some renaming of functions (including 'get' prefix on getter functions). 
    -New reports providing delta area information after modifications made by both routability and timing-driven modes.
    -New functions have been added to write a CSV file containing all GCell information, including all parallel vector values and internal GCell attributes.

EDIT: Here are some examples, master on top and new version on bottom images:
sky130hd/aes
image

nangate45/swerv
image

All public designs images are available in [this link](https://drive.google.com/drive/folders/1ddmyymOzd7DyzH3JutCDBuLzZx6VamTD?usp=sharing).


gudeh:
@osamahammad21 FYI

src/gpl/test/simple01-td.ok
```diff
@@ -137,19 +137,20 @@ Iteration |   Area    | Resized | Buffers | Nets repaired | Remaining
...
+[INFO GPL-1001] Finished with Overflow: 0.098474
 [INFO GPL-1002] Placed Cell Area              488.1100
 [INFO GPL-1003] Available Free Area           953.8760
-[INFO GPL-1004] Minimum Feasible Density        0.5117 (cell_area / free_area)
+[INFO GPL-1004] Minimum Feasible Density        0.5900 (cell_area / free_area)
 [INFO GPL-1006]   Suggested Target Densities:
 [INFO GPL-1007]     - For 90% usage of free space: 0.5686
 [INFO GPL-1008]     - For 80% usage of free space: 0.6396
 [INFO GPL-1009]     - For 50% usage of free space: 1.0234
+[INFO GPL-0083] Original area um^2: 0.00, new area: 488.11, change: inf%, New area due to routability inflation and/or timing-driven otimizations.
```
gadfort, line 154: how come there is an "inf" change?

gudeh: Thanks @gadfort. I fixed this, I was saving the original area in the wrong moment.

gudeh:
Last Friday I ran a secure-CI and noticed only metrics failing. I triggered a new run with a master merge.

src/gpl/src/nesterovBase.cpp
```diff
@@ -1494,12 +1494,17 @@ void NesterovBaseCommon::revertGCellSizeToMinRc()
...
+GCell* NesterovBaseCommon::getGCellByIndex(size_t idx)
 {
-  if (i >= gCellStor_.size()) {
+  if (idx >= gCellStor_.size()) {
+    log_->error(GPL,
+                315,
+                "getGCellByIndex out of bounds: idx = {}, size = {}",
+                idx,
+                gCellStor_.size());
     return nullptr;
```
eder-matheus: You don't need the return after the error.

src/gpl/src/nesterovPlace.cpp
```diff
@@ -830,6 +832,25 @@ int NesterovPlace::doNesterovPlace(int start_iter)
...
+
+      if (!is_routability_need_) {
+        auto block = pbc_->db()->getChip()->getBlock();
+        int64_t end_routability_area = 0;
+        for (auto& nb : nbVec_) {
+          end_routability_area += nb->nesterovInstsArea();
+        }
+        double percent_diff
+            = 100.0 * (end_routability_area - original_area) / original_area;
+        log_->info(
```
maliberty: This should be a debugPrint as I don't think most users will know what it means (or care).

src/gpl/src/nesterovPlace.cpp
```diff
@@ -845,6 +866,21 @@ int NesterovPlace::doNesterovPlace(int start_iter)
       break;
     }
   }
+  auto block = pbc_->db()->getChip()->getBlock();
+  int64_t new_area = 0;
+  for (auto& nb : nbVec_) {
+    new_area += nb->nesterovInstsArea();
+  }
+  double percent_diff = 100.0 * (new_area - original_area) / original_area;
+  log_->info(GPL,
```
maliberty, line 892: This should be a debugPrint as I don't think most users will know what it means (or care).

src/gpl/src/nesterovBase.cpp
```diff
@@ -3208,20 +3208,250 @@ std::pair<odb::dbInst*, size_t> NesterovBaseCommon::destroyCbkGCell(
...
+
+  int64_t filler_area_before_removal = totalFillerArea_;
+  size_t num_filler_before_removal = fillerStor_.size();
+  int64_t availableFillerArea = single_filler_area * fillerStor_.size();
+  int64_t originalInflationArea = inflation_area;
+
+  for (int i = nb_gcells_.size() - 1;
+       i >= 0 && removed_count < max_fllers_to_remove;
+       --i) {
+    if (nb_gcells_[i]->isFiller()) {
```
maliberty: I don't know how the fillers in nb_gcells_ are distributed over the layout.  Is this effectly random removal or are we potentially removing the fillers from a specific region (which would not be great).

gudeh: They are randomly placed on the initialization. And I am also removing them randomly (looping backwards in nb_gcells_, in the initialization, fillers are at the end of such vector).

I thought of trying to search for fillers close to any instances, or even fillers close to instances that were modified. Although this would cost some runtime.

src/gpl/src/nesterovPlace.cpp
gudeh: You think so Matt? We already show the changes in area on internal iterations (both routability and timing driven).

We are changing the gpl behavior to be more dynamic and to be modified during execution, and this log message summarizes the modifications made. 

With it, is also easier to have a grasp if problems that eventually happened in gpl would potentially be because of a large change in area.

Sorry to insist, but last call is yours.

maliberty: A change in area in TD is due to actual changes to the design (buffering/sizing).  The change in area in RD is an internal artifact of inflation.  No design area is being consumed so I think the message will be confusing.  Am I misunderstanding something?

gudeh: You are totally right! Maybe I should report the change in area separately at the end.

gudeh:
In the later commits, I added a limitation on the amount of filler cells that can be removed. I experimented with allowing up to 10%, 25%, 33%, and 50% removal based on the initial filler area. Allowing up to 10% removal seemed to yield slightly better results, with a greater reduction in initial violations observed at the start of DRT. Results are available [here](https://docs.google.com/spreadsheets/d/1MeCrsdF7k11NZbnIHXJlwz1K8cVo97UP6VPh2RHe6HA/edit?gid=0#gid=0) showing the difference from master on each run.

gudeh:
@maliberty secure-CI is passing for public designs, I just triggered the updateRules in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3220

maliberty (approved):

PR #7469, opened 2025-05-30
Free allocated Polygon90Set

Faholan:


Faholan:
The reason why it was failing is that actually `newfree` is apparently used to immediately free objects allocated by a function, instead of specifying a destructor for later use as I was thinking was the case

PR #7470, opened 2025-05-31
pdn: add marker category for repair channels

gadfort:
Changes:
- removes markers for recheck vias
- adds markers for repair channels

### DIFF
--- a/src/pdn/src/connect.cpp
+++ b/src/pdn/src/connect.cpp
@@ -925,22 +925,22 @@ void Connect::recordFailedVias() const
     std::string reason_str;
     switch (reason) {
       case failedViaReason::OBSTRUCTED:
-        reason_str = "Obstructed";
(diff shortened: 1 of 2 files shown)


src/pdn/src/connect.cpp
```diff
@@ -925,22 +925,22 @@ void Connect::recordFailedVias() const
     std::string reason_str;
     switch (reason) {
       case failedViaReason::OBSTRUCTED:
-        reason_str = "Obstructed";
+        reason_str = "Via - Obstructed";
```
gadfort, line 928: @maliberty the vias could also be in their own category under the tool, what do you think?

maliberty: I think that makes sense

maliberty:
Perhaps it should be PDN Via to avoid confusion with signal vias?

PR #7472, opened 2025-06-02
mpl: consider data flow connections during macro placement

### LINKED ISSUE #7435: Study QoR considering data flow when placing macros in macro clusters
### Description

Since the original implementation we never considered the connections from the data flow data structure when placing the macros of a macro cluster (i.e., macro placement stage of HierRTLMP). It looks like we could benefit from those indirect connections.

### Suggested Solution

Include data flow connections when placing macros.

### Additional Context

_No response_

AcKoucher:
Resolve #7435 

### Context
Mock-array had poor macro placement results with the new [floorplan rebuffering](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3101). Some investigation showed that the nets between macros and IOs were not being considered, because the data flow connections are not taken into account during the macro placement step inside MPL. The changes here are not enough to make MPL able to handle mock-array with the new rebuffering.

_Also did some renaming to clarify what is going on._


AcKoucher:
Running Secure-CI.

maliberty (approved):
The code changes are simple, its just about the CI

AcKoucher:
Secure-CI passed.

PR #7474, opened 2025-06-02
Ant: Save the number of diodes reported for a gate in another layer.

luis201420:
ANT reports the number of diodes required for each gate to repair an antenna violation in each layer for repair antennas, but in some cases, this number may be more than necessary.

These cases occur when a gate has violations in different layers. In this case, the number of diodes reported for the gate in one layer does not consider whether a number of diodes have already been reported in another layer, which could cover or reduce the number of diodes required.

This modification solves this problem by saving the number of diodes reported for a gate, avoiding adding unnecessary diodes. This modifies the results in designs that have a considerable number of diodes added such as:
- sky130hd/microwatt: **3.10%** reduced diodes.
- sky130hd/jpeg: **7.31%** Reduced diodes.
- sky130hd/chameleon: **26.05%** reduced diodes.
- sky130hd/aes: **8.46%** reduced diodes.
- sky130hs/jpeg: **4.47%** educed diodes.


eder-matheus:
@luis201420 Please add the reduction value for the designs you mentioned. A PR updating the diode metric for theses designs is desirable.

eder-matheus (approved):

PR #7477, opened 2025-06-02
ppl/odb: enure constraint regions are inside die area

eder-matheus:
Fix https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3192

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -1007,6 +1007,12 @@ class dbBlock : public dbObject
   Polygon getBTermTopLayerGridRegion();
 
   ///
+  /// Ensure that the constraint region defined in the given edge is completely
+  /// inside the die area.
+  ///
+  void ensureConstraintRegion(const Direction2D& edge, int& begin, int& end);
+
+  ///
   /// Find the rectangle corresponding to the constraint region in a specific
   /// edge of the die area.
   ///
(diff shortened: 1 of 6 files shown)


eder-matheus:
@AcKoucher FYI.

src/odb/include/odb/db.h
```diff
@@ -1006,6 +1006,12 @@ class dbBlock : public dbObject
   ///
   Polygon getBTermTopLayerGridRegion();
 
+  ///
+  /// Ensure that the constraint region defined in the given edge is completely
+  /// inside the die area.
+  ///
+  void ensureConstraintRegion(const Direction2D& edge, int& begin, int& end);
```
maliberty: I don't think this needs to be a public API.

maliberty (changes requested):

src/odb/include/odb/db.h
```diff
@@ -1689,6 +1689,12 @@ class dbBlock : public dbObject
 
  private:
   void ComputeBBox();
+
+  ///
+  /// Ensure that the constraint region defined in the given edge is completely
+  /// inside the die area.
+  ///
+  void ensureConstraintRegion(const Direction2D& edge, int& begin, int& end);
```
maliberty: I should have been more explicit.  I mean it should be a method of _dbBlock and not in db.h.

eder-matheus: Oh, my bad! Now it should be correct.

PR #7484, opened 2025-06-03
gui: CTS Viewer show/hide subtree

### LINKED ISSUE #7388: In clock tree view add choice for macro / std cell display
### Description

In the clock tree view add an option to show only macros, only std cells, or both (current)

### Suggested Solution

_No response_

### Additional Context

_No response_

LucasYuki:
Closes #7388.

Adds the feature to hide/show subtrees in CTS Viewer by right-clicking on a buffer.


arthurjolo:
I tested this PR with some designs that are hard to look at the macros trees and it was really helpful. On a user side I think this PR is Ok.

maliberty:
When you hide a subtree it doesn't seem to adjust the drawing of the tree on the layout, eg
image

maliberty:
If you hide a subtree, eg
image

and then hide a higher node, eg
image

and the unhide the higher node you get the full tree
image

it appears to unhide all the lower nodes as well.  Is this intended?

PR #7490, opened 2025-06-04
DRT: support higher via access layer num correctly and minor fixes

osamahammad21:
- Add VIA access layer range in the pin access command
- Apply the correct via cost considering the via access layer range not only the via layer.
- Correctly evaluate via guides in guides processing
- Show the path segment connecting to the via in pin access debug mode.

### DIFF
--- a/src/drt/src/TritonRoute.cpp
+++ b/src/drt/src/TritonRoute.cpp
@@ -1028,6 +1028,9 @@ int TritonRoute::main()
 
 void TritonRoute::pinAccess(const std::vector<odb::dbInst*>& target_insts)
 {
+  if (router_cfg_->DBPROCESSNODE == "GF14_13M_3Mx_2Cx_4Kx_2Hx_2Gx_LB") {
+    router_cfg_->USENONPREFTRACKS = false;
+  }
   if (distributed_) {
     asio::post(*dist_pool_, [this]() {
       sendDesignDist();
--- a/src/drt/src/TritonRoute.tcl
+++ b/src/drt/src/TritonRoute.tcl
@@ -259,6 +259,8 @@ sta::define_cmd_args "pin_access" {
     [-db_process_node name]
     [-bottom_routing_layer layer]
     [-top_routing_layer layer]
+    [-via_in_pin_bottom_layer layer]
+    [-via_in_pin_top_layer layer]
     [-min_access_points count]
     [-verbose level]
     [-distributed]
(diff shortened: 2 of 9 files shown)


src/drt/src/TritonRoute.tcl
```diff
@@ -259,6 +259,8 @@ sta::define_cmd_args "pin_access" {
     [-db_process_node name]
     [-bottom_routing_layer layer]
     [-top_routing_layer layer]
+    [-via_in_pin_bottom_layer layer]
+    [-via_in_pin_top_layer layer]
```
maliberty, line 263: Add to the docs

src/drt/src/TritonRoute.cpp
```diff
@@ -1028,6 +1028,9 @@ int TritonRoute::main()
 
 void TritonRoute::pinAccess(const std::vector<odb::dbInst*>& target_insts)
 {
+  if (router_cfg_->DBPROCESSNODE == "GF14_13M_3Mx_2Cx_4Kx_2Hx_2Gx_LB") {
```
maliberty, line 1040: We really need to kill this off...

osamahammad21: I agree.

src/drt/README.md
```diff
@@ -151,6 +151,8 @@ pin_access
     [-db_process_node name]
     [-bottom_routing_layer layer]
     [-top_routing_layer layer]
+    [-via_in_pin_bottom_layer layer]
+    [-via_in_pin_top_layer layer]
```
maliberty, line 155: You need to add descriptions to the options section below.

src/drt/src/pa/FlexPA_acc_point.cpp
```diff
@@ -1061,10 +1061,12 @@ void FlexPA::filterMultipleAPAccesses(
...
                     inst_term);
     if (is_std_cell_pin) {
-      has_access |= ((layer_num == router_cfg_->VIA_ACCESS_LAYERNUM
-                      && ap->hasAccess(frDirEnum::U))
-                     || (layer_num != router_cfg_->VIA_ACCESS_LAYERNUM
-                         && ap->hasAccess()));
+      const bool isViaAccessLayer
+          = layer_num == router_cfg_->VIA_ACCESS_LAYERNUM
+            || (layer_num >= router_cfg_->VIAINPIN_BOTTOMLAYERNUM
+                && layer_num <= router_cfg_->VIAINPIN_TOPLAYERNUM);
```
maliberty, line 1067: This looks to repeat isViaAccessLayerNum and similarly in FlexPA::genPinAccessCostBounded.  Can this be a single shared function?

PR #7502, opened 2025-06-04
pdn: add test for M1->M3 followpins with split cuts

gadfort:
Adds:
- test for using split cuts on M2 for follow pins in M1 and M3 in asap7


src/pdn/test/asap7_M1_M3_followpins.tcl
```diff
@@ -0,0 +1,23 @@
...
+add_global_connection -net VDD -pin_pattern {^VDD$} -power
+add_global_connection -net VSS -pin_pattern {^VSS$} -ground
+
+set_voltage_domain -power VDD -ground VSS
+
+define_pdn_grid -name "Core"
+add_pdn_stripe -followpins -layer M1 -width 0.072
+add_pdn_stripe -followpins -layer M3 -width 0.090
+
+add_pdn_connect -layers {M1 M3} -split_cuts {M2 1.0}
```
maliberty, line 17: @jeffng-or we can use this for your PDK work

PR #7506, opened 2025-06-04
gpl: reduce counting of new cells after callback removal

gudeh:
This has no functional change, since we do not trigger callback removals yet.


eder-matheus (approved):

gudeh:
@povik FYI, let me know if we still see gpl warning after this change.

PR #7509, opened 2025-06-04
mpl: model blockages as macro clusters constrained by fences

AcKoucher:
Replaces #6502

### Context
Currently, in cluster placement, blockages are modeled as regions that should not be occupied by clusters with macros - penalty based on overlap area. With this approach, the only way to reduce the overlap penalty is to either fill the blockage region with std cell clusters or leave it empty (the latter is very challenging because of the sequence pair representation).

### Idea Here
Model blockages as macro clusters constrained to fences to prevent any cluster from occupying the blockage region.

### Also Needed
Increase fence penalty so that if effectively places the blockages where they belong.

_Just for the record, this is a very aggressive idea. I'm studying it as a possible, but not certain enhancement._

### DIFF
--- a/src/mpl/src/graphics.cpp
+++ b/src/mpl/src/graphics.cpp
@@ -390,7 +390,7 @@ void Graphics::drawObjects(gui::Painter& painter)
 
   int i = 0;
   for (const auto& macro : soft_macros_) {
-    if (isSkippable(macro)) {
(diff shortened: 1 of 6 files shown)


AcKoucher:
Running Secure-CI to evaluate impact.

AcKoucher:
QoR regarding the benefits of these changes is dubious. I'm closing this for now.

PR #7517, opened 2025-06-05
Fix non-static class member arrival_paths_ is not initialized in constructor warning.

mguthaus:
Fix non-static class member arrival_paths_ is not initialized in constructor warning.

### DIFF
--- a/src/rsz/src/BufferMove.hh
+++ b/src/rsz/src/BufferMove.hh
@@ -24,8 +24,6 @@ using sta::Vertex;
 class BufferMove : public BaseMove
 {
  public:
-  using BaseMove::BaseMove;
-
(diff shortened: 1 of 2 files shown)


src/rsz/src/BufferMove.hh
```diff
@@ -26,6 +26,8 @@ class BufferMove : public BaseMove
  public:
   using BaseMove::BaseMove;
 
+  BufferMove(Resizer* resizer);
```
maliberty: I think you need to remove `using BaseMove::BaseMove`

mguthaus: Sure, it would just shadow it, but it is clearer by removing it.

maliberty (approved):

PR #7522, opened 2025-06-05
Bazel CI in Jenkins

vvbandeira:


maliberty:
@oharboe there is still some issue with the bazel-orfs test which uses docker inside the docker that Jenkins started.  We can merge this with the knowledge that this particular test will fail for now or wait for Vitor to figure out a solution.  We could also just run the src/test units for now.

oharboe:
I think we need caching before we putit in production anyway, so might as well iterate on the PR

maliberty:
Without caching the runtime is no more than the regular build & test CI so putting it in parallel won't hurt.  That is without the bazel-orfs which is currently failing but I don't expect to add too much time.

oharboe:
I see.

vvbandeira:
@maliberty @oharboe, with ORFS tests disabled in Bazel I expect a clean CI run. We can investigate and fix the issues with ORFS tests in a later PR.

PR #7530, opened 2025-06-06
gpl: allow load fastroute heatmap in gui

gudeh:
New TCL parameter for gpl to allow load fastroute heatmap in gui

### DIFF
--- a/src/gpl/README.md
+++ b/src/gpl/README.md
@@ -98,6 +98,7 @@ global_placement
     [-timing_driven_nets_percentage]
     [-keep_resize_below_overflow]
     [-disable_revert_if_diverge]
+    [-load_fast_route]
 ```
 
 #### Options
@@ -121,6 +122,7 @@ global_placement
 | `-pad_right` | Set right padding in terms of number of sites. The default value is 0, and the allowed values are integers `[1, MAX_INT]` |
 | `-skip_io` | Flag to ignore the IO ports when computing wirelength during placement. The default value is False, allowed values are boolean. |
 | `-disable_revert_if_diverge` | Flag to make gpl store the placement state along iterations, if a divergence is detected, gpl reverts to the snapshot state. The default value is disabled. |
+| `-load_fast_route` | Loads FastRoute GRT allowing for Heatmap Routing Congestion to be available after global placement. |
 
 #### Routability-Driven Arguments
 
--- a/src/gpl/README.md
+++ b/src/gpl/README.md
@@ -122,7 +122,7 @@ global_placement
 | `-pad_right` | Set right padding in terms of number of sites. The default value is 0, and the allowed values are integers `[1, MAX_INT]` |
 | `-skip_io` | Flag to ignore the IO ports when computing wirelength during placement. The default value is False, allowed values are boolean. |
 | `-disable_revert_if_diverge` | Flag to make gpl store the placement state along iterations, if a divergence is detected, gpl reverts to the snapshot state. The default value is disabled. |
-| `-load_fast_route` | Loads FastRoute GRT allowing for Heatmap Routing Congestion to be available after global placement. |
+| `-load_fast_route` | Flag for loading FastRoute GRT allowing for Heatmap Routing Congestion to be available after global placement. |
 
 #### Routability-Driven Arguments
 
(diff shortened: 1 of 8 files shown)


gudeh:
If we are willing to accept the runtime cost to load grt after gpl finishes, we can leave it as default

src/gpl/src/replace.cpp
```diff
@@ -371,6 +371,15 @@ int Replace::doNesterovPlace(int threads, int start_iter)
              1,
              "NP->doNesterovPlace() runtime: {} seconds ",
              elapsed.count());
+
+  log_->report("are we going to loading fast route?");
```
eder-matheus: Remove this comment. I believe it was debug comment that went to the PR.

src/gpl/src/replace.cpp
```diff
@@ -371,6 +371,15 @@ int Replace::doNesterovPlace(int threads, int start_iter)
...
+  if (load_fast_route_) {
+    log_->report("loading fast route!");
```
eder-matheus: Ditto.

src/gpl/src/routeBase.cpp
```diff
@@ -314,6 +314,14 @@ void RouteBase::getGrtResult()
   updateGrtRoute();
 }
 
+void RouteBase::loadGrt()
```
eder-matheus, line 317: Can you use this function in `replace.cpp`. It seems we have duplicated code and this function is never used.

src/gpl/README.md
```diff
@@ -121,6 +122,7 @@ global_placement
 | `-pad_right` | Set right padding in terms of number of sites. The default value is 0, and the allowed values are integers `[1, MAX_INT]` |
 | `-skip_io` | Flag to ignore the IO ports when computing wirelength during placement. The default value is False, allowed values are boolean. |
 | `-disable_revert_if_diverge` | Flag to make gpl store the placement state along iterations, if a divergence is detected, gpl reverts to the snapshot state. The default value is disabled. |
+| `-load_fast_route` | Flag for loading FastRoute GRT allowing for Heatmap Routing Congestion to be available after global placement. |
```
eder-matheus: Maybe rephrase to: `Flag to run global routing after global placement, enabling the Routing Congestion Heatmap.`

I'm also not sure if `load_fast_route` is the best flag name. Do you intend to have this flag available for final users or is this a developer functionality?

eder-matheus (changes requested):

src/gpl/src/replace.cpp
gudeh: Whops!

src/gpl/src/routeBase.cpp
gudeh: I tried using this internally for gpl debug automatic generating fatroute images. When I use it, fastroute works fine, but RUDY heatmap stops being updated. I left it here for revisiting the issue

src/gpl/README.md
gudeh: I understand users might be interested to have the more precise congestion heatmap from fastroute with the cost of runtime to load grt.

eder-matheus: I would avoid using "fast_route" in the documentation and in the flag name. FastRoute is internal to GRT. I suggest using "load_global_route" or something similar instead.

gudeh: what about enable_routing_congestion?

eder-matheus: That's ok. Please update the description too.

src/gpl/src/replace.cpp
```diff
@@ -490,6 +497,11 @@ void Replace::setDisableRevertIfDiverge(bool mode)
   disableRevertIfDiverge_ = mode;
 }
 
+void Replace::setLoadFastRoute(bool mode)
```
eder-matheus: Just a nit: make the names consistent between the Swig interface and the C++ implementation. `setEnableRoutingCongestion`.

src/gpl/README.md
```diff
@@ -98,6 +98,7 @@ global_placement
     [-timing_driven_nets_percentage]
     [-keep_resize_below_overflow]
     [-disable_revert_if_diverge]
+    [-enable_routing_congestion_fast_route]
```
eder-matheus: Remove `_fast_route` from the end of the flag name.

eder-matheus (approved):
I'll approve since the requested changes are small. Please address them when you can.

PR #7531, opened 2025-06-06
repair_timing upsizeCell swappable sort BugFix?

spongebobaa16:
When call command "repair_timing", in upsize move, the swappable cells will be sorted.
However, the sorting order seems to be strange..?

return std::tie(drive2, intrinsic1, capacitance1) < std::tie(drive1, intrinsic2, capacitance2);

Is there any specific heuristic or reason to sort in this way?

### DIFF
--- a/src/rsz/src/BaseMove.cc
+++ b/src/rsz/src/BaseMove.cc
@@ -632,8 +632,8 @@ LibertyCell* BaseMove::upsizeCell(LibertyPort* in_port,
            const ArcDelay intrinsic2 = port2->intrinsicDelay(this);
            const float capacitance1 = port1->capacitance();
            const float capacitance2 = port2->capacitance();
-           return std::tie(drive2, intrinsic1, capacitance1)
-                  < std::tie(drive1, intrinsic2, capacitance2);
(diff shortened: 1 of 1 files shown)


povik:
Expanded sorting code from before #7311:
```
    sort(swappable_cells,
         [=](const LibertyCell* cell1, const LibertyCell* cell2) {
           LibertyPort* port1
               = cell1->findLibertyPort(drvr_port_name)->cornerPort(lib_ap);
           LibertyPort* port2
               = cell2->findLibertyPort(drvr_port_name)->cornerPort(lib_ap);
           const float drive1 = port1->driveResistance();
           const float drive2 = port2->driveResistance();
           const ArcDelay intrinsic1 = port1->intrinsicDelay(this);
           const ArcDelay intrinsic2 = port2->intrinsicDelay(this);
           return drive1 > drive2
                  || ((drive1 == drive2 && intrinsic1 < intrinsic2)
                      || (intrinsic1 == intrinsic2
                          && port1->capacitance() < port2->capacitance()));
         });
```
We try weak drivers first. Your change might make us upsize to the strongest driver right away.

maliberty:
It sounds like the code is correct as is.

PR #7532, opened 2025-06-06
Cts estimate wire delay for balancing macros and register latencies

arthurjolo:
Currently during CTS macros and register tree latencies balancing we use the ideal clock mode, so the clock arrival times computed don't take into account wire delay, slew, etc. This can cause a missmach between the results CTS thinks it has achieved and the reality. e.g.: asap7/mock-array.

Also the macro macros and register tree latencies balancing happens before reapir_clock_nets and thus the results it achieved are changes and can lead to designs beeing to much delayed. So this PR allong with PR [#7523](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7523) address this issue.

This PR solves these issues by:

1. After CTS builds the trees and before balancing macros and register tree latencies, set all clocks to propagated mode and estimate the wire parasitics.
3. Run reapir_clock_nets.
4. estimate the wire parasitics after reapir_clock_nets.
5. Balance macros and register tree latencies,

This way cts can achieve a much precise result in balancing the macros and regeter latencies.

asap7/mock-array:
Result CTS sees:
Captura de tela de 2025-05-26 13-52-18

Real CTS results:
Captura de tela de 2025-05-26 13-53-18


oharboe:
Wow! Very visual difference!

maliberty:
Conflicts, CI failures.  Does this also need an ORFS metrics PR?

maliberty:
The coding changes look fine.  Is the "Real CTS results" with or without this change?  It would be nice to include whichever isn't shown for comparison.

arthurjolo:
> The coding changes look fine. Is the "Real CTS results" with or without this change? It would be nice to include whichever isn't shown for comparison.

The real CTS results is without my changes. The CTS results with my chages is:
mock_array_test_branch_clock

arthurjolo:
> Conflicts, CI failures. Does this also need an ORFS metrics PR?

It needs to use the PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/7523 option in ORFS. I am discussing with Cho on how to better use it.

PR #7533, opened 2025-06-06
Update PrintProgress to print pending moves which are included in area.

mguthaus:
This fixes a discrepency where the stats on the moves are only the committed moves (from the move classes), but the area is from the ECO module which includes the pending moves. This will report pending moves and area, so the progress could go backwards, but the alternative is to add area tracking to the moves which seems overkill.


maliberty:
I think the committed moves is nicer but at least this is consistent.

PR #7539, opened 2025-06-08
rsz: Add slew violation check to size_down, re-enable size_down.

mguthaus:
Don't merge yet, doing PR to get QoR results.

### DIFF
--- a/src/rsz/src/SizeDownMove.cc
+++ b/src/rsz/src/SizeDownMove.cc
@@ -18,6 +18,7 @@
 #include "sta/Graph.hh"
 #include "sta/GraphDelayCalc.hh"
 #include "sta/Liberty.hh"
+#include "sta/MinMax.hh"
 #include "sta/NetworkClass.hh"
 #include "sta/Path.hh"
 #include "sta/PathExpanded.hh"


QuantamHD:
@mguthaus You need to amend d3d33f85e1a1cfd2db3153be03fff0e2abc94880 with a `git rebase HEAD~2 -s` then force push to make the DCO pass.

mguthaus:
@povik Can you take a look? I think I've got general improvement in QoR in most cases.

mguthaus:
Or can @maliberty take a look?

povik:
@mguthaus I've started a run to evaluate QoR on ORFS. I will try to review the code

povik:
@mguthaus QoR comparison with a clean baseline (no other changes should be interfering): https://dashboard.precisioninno.com/compare?sourceAType=Commit&sourceBType=Commit&sourceBName=1171cb1d34e470c433f8eb5baef6b71213e6a920&sourceBID=1&sourceAName=4a04de24f4b308603ad92c9206633347015530cd&sourceAID=5753

I don't see any notable degradation.

mguthaus:
Cool comparison tool.

It seems there is consistent finish__power__total and finish__design__core__area improvement in most designs (none got worse). This is expected since this is using downsizing to get improvement.  

finish__timing__setup__ws and finish__timing__setup__tns results are a bit mixed but with more improvement than not. (Maybe I'm being subjective...). I think this also makes sense because other moves would get a similar result (albeit using more power).

Also, there is no degredatation in finish__runtime__total which was my main concern on larger designs.

src/rsz/src/SizeDownMove.cc
```diff
@@ -172,171 +168,297 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
            LibertyPort* port1
-               = cell1->findLibertyPort(fanout_port_name)->cornerPort(lib_ap);
+               = cell1->findLibertyPort(load_port_name)->cornerPort(lib_ap);
            const LibertyPort* port2
-               = cell2->findLibertyPort(fanout_port_name)->cornerPort(lib_ap);
+               = cell2->findLibertyPort(load_port_name)->cornerPort(lib_ap);
 
            const float cap1 = port1->capacitance();
            const float cap2 = port2->capacitance();
            const ArcDelay intrinsic1 = port1->intrinsicDelay(this);
```
povik: I think this is calling intrinsicDelay on an input port and that will always come out to zero

src/rsz/src/SizeDownMove.cc
```diff
@@ -172,171 +168,297 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
+             "size_down fanout {} swaps={}",
+             network_->pathName(load_pin),
+             swappable_names.c_str());
+
+  const float load_input_cap = load_port->cornerPort(lib_ap)->capacitance();
+
+  // Get fanouts based on Liberty since STA arcs are not present in DFFs
+  // Could have more than one fanout (e.g. Q and QN of a flop)
+  const Instance* fanout_inst = network_->instance(load_pin);
+  auto output_pins = getFanouts(fanout_inst);
```
povik: It looks like this helper returns output pins not fanouts (I know it's an existing helper)

src/rsz/src/SizeDownMove.cc
```diff
@@ -172,171 +168,297 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
+                 delayAsString(output_delays[i], sta_, 3),
+                 delayAsString(slack_margin, sta_, 3),
+                 delayAsString(delay_change, sta_, 3),
+                 delayAsString(slack_margin, sta_, 3));
+
+      // First case is positive slack and delay change doesn't get worse than
+      // that slack. Second case is negative slack and delay is improved, but
+      // doesn't necessarily fix the slack violation entirely.
+      if ((slack_margin > 0 && delay_change > slack_margin)
+          || (slack_margin < 0 && delay_change > 0)) {
```
povik: We don't check paths through the gate which don't go through `load_pin`; also for flops this check will be comparing delay in driving `Q` with slack on `D`.

mguthaus: > We don't check paths through the gate which don't go through load_pin;

Can you elaborate what you mean here?

povik: The slack considered in the check is evaluated at `load_pin`, also `drvr_delta_delay` only applies for paths going through that pin

mguthaus: I think I get what you mean. We don't have to check other paths that aren't being sized down, because they will always get faster. We only consider smaller input capacitance gates, so the capacitance load of the driver is always decreased and all other fanout gates will always get faster. 

The real evaluation is whether the fanout gate doesn't get *too slow*. We figure this out by checking how much faster the driver gets and how much slower the fanout gate gets.

You are correct about the flop slack issue.

src/rsz/src/SizeDownMove.cc
mguthaus: Renamed.

src/rsz/src/SizeDownMove.cc
povik: > We don't have to check other paths that aren't being sized down,

Oops, I'm not being clear. I mean let's say the gate we are downsizing is a NAND, we only check the slack on pin A but there could be a critical path going through pin B which we don't check for. One way to include paths going through B would be to build the check off the slack on the output pin of the NAND gate, but then it's not precise to include `drvr_delta_delay` for paths going through B.

mguthaus: Got it. Yes, you are correct.

mguthaus:
@povik Can you take a lok?

src/rsz/src/BaseMove.cc
```diff
@@ -700,6 +704,114 @@ vector<const Pin*> BaseMove::getFanouts(const Instance* inst)
...
+}
+
+float BaseMove::computeElmoreSlewFactor(const Pin* output_pin,
+                                        LibertyPort* output_port,
+                                        float output_load_cap)
+{
+  float elmore_slew_factor = 0.0;
+
+  // Get the vertex for the output pin
+  Vertex* output_vertex = graph_->pinLoadVertex(output_pin);
```
povik: pinDrvrVertex to handle bidirectional drivers. I'm not sure this is critical as for unidirectional this is the same vertex and bidirectional drivers might be ruled out for this move

src/rsz/src/BaseMove.cc
```diff
@@ -700,6 +704,114 @@ vector<const Pin*> BaseMove::getFanouts(const Instance* inst)
...
+    // Do this once to populate the set
+    if (buffer_sizes_.empty()) {
+      for (LibertyCell* buffer : resizer_->buffer_fast_sizes_) {
+        buffer_sizes_.push_back(buffer);
+      }
+    }
+    if (resizer_->buffer_fast_sizes_.count(base) == 0) {
+      return LibertyCellSeq();
+    }
+    return buffer_sizes_;
```
povik: There might be a latent bug where we don't clear `buffer_sizes_` when the user does `set_dont_use`. We're creating a copy for the return value anyway so I'm not sure anything is bought by caching `buffer_sizes_`

src/rsz/src/SizeDownMove.cc
```diff
@@ -68,19 +79,11 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
-                 delayAsString(fanout_slack, sta_, 3),
-                 delayAsString(drvr_slack, sta_, 3))
-          // If we already have a move on the load, don't try to size down
-          if (!hasMoves(load_inst))
-      {
+      Pin* fanout_pin = fanout_vertex->pin();
+      Instance* fanout_inst = network_->instance(fanout_pin);
+      // If we already have a move on the fanout gate, don't try to size down
+      // again
+      if (!hasMoves(fanout_inst)) {
```
povik, line 97: I know we have avoided combining moves on instances due to journaling limitations. These limitations are now gone. Do you think we should continue these checks?

src/rsz/src/SizeDownMove.cc
povik: @mguthaus I think this hasn't been addressed.

src/rsz/src/SizeDownMove.cc
```diff
@@ -172,168 +171,271 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
-    // We might want to use the most critical one instead?
-    for (int i = 0; i < fanouts.size(); i++) {
-      const float cur_fanout_load_cap
-          = graph_delay_calc_->loadCap(fanouts[i], dcalc_ap);
-      if (cur_fanout_load_cap > fanout_load_cap) {
-        fanout_load_cap = cur_fanout_load_cap;
-        max_fanout_id = i;
-      }
+  string swappable_names;
+  for (LibertyCell* swappable : swappable_cells) {
```
povik: Nit: ideally we would gate `swappable_names` creation on `logger->debugCheck(RSZ, "size_down", 3)`. It's fine to leave it as is

src/rsz/src/SizeDownMove.cc
```diff
@@ -172,168 +171,271 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
-        continue;
+
+    // Max capacitance and slew checking
+    bool skip_cell = false;
+    for (int i = 0; i < output_pins.size(); i++) {
+      LibertyPort* output_port
+          = swappable->findLibertyPort(output_port_names[i]);
+
+      // Check for max capacitance violation
+      if (checkMaxCapViolation(output_pins[i], output_port, output_caps[i])) {
```
povik, line 338: I think this check suffers from corner confusion. Maybe leave a comment so we can eventually fix it together with other cases throughout the resizer

src/rsz/src/SizeDownMove.cc
```diff
@@ -172,168 +171,271 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
+         output_index++) {
+      LibertyPort* output_port
+          = swappable->findLibertyPort(output_port_names[output_index]);
+      float new_load_delay = resizer_->gateDelay(
+          output_port, output_caps[output_index], dcalc_ap);
+      float delay_change = 0.0;
+
+      if (load_cell->hasSequentials()) {
+        // Sequential elements: output timing determined by clock, not inputs
+        delay_change = drvr_delta_delay - output_delays[output_index];
```
povik: I think this should be `new_load_delay` not `drvr_delta_delay`

src/rsz/src/SizeDownMove.cc
```diff
@@ -172,168 +171,271 @@ bool SizeDownMove::doMove(const Path* drvr_path,
...
+               delayAsString(output_delays[0], sta_, 3),
+               delayAsString(slack_margin, sta_, 3),
+               delayAsString(worst_delay_change, sta_, 3),
+               delayAsString(slack_margin, sta_, 3));
+
+    // First case is positive slack and delay change doesn't get worse than
+    // that slack. Second case is negative slack and delay is improved, but
+    // doesn't necessarily fix the slack violation entirely.
+    if ((slack_margin > 0 && worst_delay_change > slack_margin)
+        || (slack_margin < 0 && worst_delay_change > 0)) {
```
povik: In the case of flops `slack_margin` will be referring to the input side and `worst_delay_change` to the output side. We should fetch a different slack or ignore sequentials in the move

src/rsz/src/BaseMove.hh
```diff
@@ -224,12 +224,33 @@ class BaseMove : public sta::dbStaState
...
+  static constexpr int size_down_max_fanout_ = 10;
   static constexpr int rebuffer_max_fanout_ = 20;
   static constexpr int split_load_min_fanout_ = 8;
   static constexpr int buffer_removal_max_fanout_ = 10;
   static constexpr float rebuffer_relaxation_factor_ = 0.03;
 
-  vector<const Pin*> getFanouts(const Instance* inst);
+  // Experiment flag: when true, only consider the direct input pin for delay
+  // change when false (default), analyze all input pins for worst delay change
+  static constexpr bool use_direct_pin_only_for_delay_change_ = true;
```
povik: Looks to be left over, it's not used

src/rsz/src/BaseMove.cc
mguthaus: This is intentionally trying to use the load pin, because it will include any wire effects on slew. This assumes that the wire is roughly the same and that is why it is computing this "slew factor". If you were to just look at the slew at the driver pin, it won't include that and will miss slew violations.

src/rsz/src/SizeDownMove.cc
mguthaus: Probably not. I suggest we remove when we do it for all moves though. What do you think?

src/rsz/src/BaseMove.cc
povik: > This is intentionally trying to use the load pin, because it will include any wire effects on slew.

That's not what this will do. For non-bidirectional drivers `graph_->pinLoadVertex(output_pin);` will return a vertex corresponding to the driver pin.

`pinLoadVertex` vs `pinDrvrVertex` only matters for bidirectional driver pins, in which case one returns a vertex representing the pin as a load, and the other as a driver.

Screenshot 2025-08-15 at 21 12 52

src/rsz/src/SizeDownMove.cc
povik: That sounds good

src/rsz/src/BaseMove.cc
mguthaus: Got it! Thanks. That makes more sense -- especially since there would be more than one load vertex fanning out from an output pin.

src/rsz/src/SizeDownMove.cc
mguthaus: Good call. I will do this.

mguthaus:
@povik Can you take a look again? I believe I fixed all prior issues.

Results in setup slack are mixed tending to improved slack. But this is because the one path at a time repair setup needs to be improved and is in a local minima. It gets consistent power improvement with similar slack in all other cases.

src/rsz/src/RepairSetup.cc
```diff
@@ -161,6 +161,11 @@ bool RepairSetup::repairSetup(const float setup_slack_margin,
     // TODO: Add size_down_move to the sequence if we want to allow
     // Always  have sizing
     move_sequence.push_back(resizer_->size_up_move_.get());
+    move_sequence.push_back(resizer_->size_down_move_.get());
```
povik: Should this be here? The comment below says it's disabled by default for now

mguthaus: I had enabled it by default. Help and defaults should be updated. I can remove this comment.

mguthaus: Updated.

povik: Thanks

maliberty:
I see merge conflicts to resolve.   Is this otherwise ready for testing?

src/rsz/src/SizeDownMove.cc
```diff
@@ -9,6 +9,10 @@
 #include <tuple>
 
 #include "BaseMove.hh"
+#include "ord/Timing.h"
```
maliberty: rsz should not depend on ord.   That API is for external clients to use.  Internally we should use sta APIs directly.

mguthaus: Fixed. It was unused.

mguthaus:
> I see merge conflicts to resolve. Is this otherwise ready for testing?

I rebased and pushed, but it fails clang-tidy due to the CI issues, I believe?

maliberty:
Clang-tidy is clean but these tests need investigation:
17:17:35  The following tests FAILED:
17:17:35  	339 - rsz.repair_setup9_hier.tcl (Failed)               IntegrationTest tcl rsz log_compare
17:17:35  	422 - rsz.repair_setup_vt_swap.tcl (Failed)             IntegrationTest tcl rsz log_compare
17:17:35  	423 - rsz.repair_setup_vt_swap2.tcl (Failed)            IntegrationTest tcl rsz log_compare
17:17:35  	424 - rsz.repair_fanout7_skip_pin_swap.tcl (Failed)     IntegrationTest tcl rsz log_compare

maliberty:
I updated the metrics to help get this passing.  If it does I'll start a full CI.

src/rsz/src/RepairSetup.cc
mguthaus: I went ahead and disabled it just so I can get the update code PRed.

mguthaus:
@maliberty I disabled this by default again as I want to get the working move into the code in place of the broken one.

src/rsz/src/BaseMove.cc
```diff
@@ -699,21 +701,198 @@ bool BaseMove::replaceCell(Instance* inst, const LibertyCell* replacement)
   }
   return false;
 }
+Slack BaseMove::getWorstInputSlack(Instance* inst)
```
maliberty, line 704: Have you considered bi-directional pins?

mguthaus: Good observation. Fixing this as I try to figure out any other issues.

PR #7540, opened 2025-06-08
gpl: dummies ignore fixed instances in other regions

LucasYuki:
Creates dummy instances on top of fixed instances from other regions.

PR #7405 removed the creation of dummy instances on top of fixed instances, but the placerBase can only see instances that belong to its region, so dummy instances still need to be created on top of fixed instances from other regions.


gudeh:
with this change we do not set as FixedInst the site that holds a fixed inst from another region, correct? This implies a different behavior with dummies? Dummies are the sites tagged as Empty, right?

PR #7541, opened 2025-06-09
drt: simplification of genPinAccessCostBound return

bnmfw:
Supports #7150 and #7153.

Every CI Safe

Simplifies `genPinAccessCostBounded` calls so it does not allow early stops from `genPinAccess`.
This makes so `updatePinStats` is only called once and the moving of access points to the pins only happens in a single place.
I need this so I can have a single place where I can get all access points from a term right before they are transferred to the pin.

Also creates `isStdCellTerm`, `isMacroCellTerm` and `isIOCellTerm` just to clean up some code in some places and make things less confusing with io terms.

### viaMaxExt
The calculation of the via max ext has been separated in order to be calculated somewhere else in the code.

### sortViaDefs
One thing that was stopping a more dynamic insertion of via_defs iteratively and was messing with some DEFs on other PRs was the fact that the via defs have to be sorted by max_ext. I created  `sortViaDefs` to do this sorting later. This takes out the requirement of validating all via defs for an access point in one single batch. This will help with other PRs failing CIs


src/drt/src/pa/FlexPA.h
```diff
@@ -553,6 +556,24 @@ class FlexPA
...
       const std::vector<gtl::polygon_90_data<frCoord>>& layer_polys);
 
+  /**
+   * @brief Calculates the maximum extension a via has overhang on pin shapes
+   *
+   * @param inst_term the term representing the pin
+   * @param ap the access point the via is centered on
+   * @param polyset the polygon set representing pin shapes
+   * @param via_def the via def
+   * @param early_stop some use cases just want to check if max_ext > 0. In
```
osamahammad21: This feels like an unnecessary optimization — it doesn't reduce complexity, and the runtime without the early stop appears negligible. Would you be open to removing it to improve code readability?

osamahammad21 (changes requested):

src/drt/src/pa/FlexPA.h
bnmfw: Sure. I'm sometimes afraid of `boost` operations and wanted to avoid the intersections when possible, but I do agree this might have been overkill

PR #7544, opened 2025-06-09
ord: add API to report the total hpwl

### LINKED ISSUE #7467: `report_wire_length` could have a net-based wire length estimation
### Description

This should be very useful to QoR study across different tools in OR.

### Suggested Solution

Move the current implementation to ODB and add the new logic. Also create a function to avoid always having to write the instantiation of the WireLengthEvaluator.

### Additional Context

Suggested in [#7465](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7465#issuecomment-2922763993).

AcKoucher:
Simpler approach to resolve #7467.

### DIFF
--- a/src/OpenRoad.i
+++ b/src/OpenRoad.i
@@ -584,12 +584,20 @@ std::string get_docs_path()
   return ord->getDocsPath();
 }
 
+void report_each_net_hpwl()
+{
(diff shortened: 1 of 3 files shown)


AcKoucher:
@povik FYI.

Issue #6560, opened 2025-01-20
Refactor repeated code in each rule parser


osamahammad21:
This same splitting occurs in each parser.  It would be nice to factor it out once.  It can happen in a separate PR.

_Originally posted by @maliberty in https://github.com/The-OpenROAD-Project/OpenROAD/pull/6559#discussion_r1922823630_

PR #7565, opened 2025-06-13
odb: refactoring of repeated code in each parser

jfgava:
Fixes [#6560](https://github.com/The-OpenROAD-Project/OpenROAD/issues/6560)
 
## Problem
Several parser classes (`MetalWidthViaMapParser`, `lefTechLayerSpacingEolParser`, `KeepOutZoneParser`, etc.) contained identical logic for:
- Splitting semicolon-delimited rule strings
- Trimming whitespace from individual rules
- Skipping empty rules
- Iterating through each rule for processing

## Solution
Created a new `processRules` template function in `parserUtils.h` that:
- Encapsulates the common string splitting and preprocessing logic
- Uses a template parameter to accept any callable (lambda, function, functor) for rule processing
- Maintains the exact same behavior as the original implementations

### DIFF
--- a/src/odb/src/lefin/lefMacroPropParser.h
+++ b/src/odb/src/lefin/lefMacroPropParser.h
@@ -23,7 +23,7 @@ class lefMacroEdgeTypeParser
   {
   }
   void parse(const std::string&);
-  bool parseSubRule(std::string);
+  bool parseSubRule(const std::string&);
   void setRange(boost::fusion::vector<double, double>& params);
 
  private:
--- a/src/odb/src/lefin/lefMacroPropParser.h
+++ b/src/odb/src/lefin/lefMacroPropParser.h
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <boost/fusion/container/vector.hpp>
 #include <string>
 #include <string_view>
 
--- a/src/odb/src/lefin/lefTechLayerSpacingEolParser.cpp
+++ b/src/odb/src/lefin/lefTechLayerSpacingEolParser.cpp
@@ -363,9 +363,7 @@ void wrongDirSpaceParser(double value,
   sc->setWrongDirSpace(l->dbdist(value));
 }
 
-template <typename Iterator>
-bool parse(Iterator first,
-           Iterator last,
+bool parse(std::string s,
            odb::dbTechLayer* layer,
            odb::lefinReader* l)
 {
@@ -454,7 +452,9 @@ bool parse(Iterator first,
               >> double_[boost::bind(&wrongDirSpaceParser, _1, sc, l)])
          >> (withinRule | toconcavecornerrule | tonotchlengthrule)
          >> -lit(";"));
-
+  
+  auto first = s.begin();
+  auto last = s.end();
   bool valid
       = qi::phrase_parse(first, last, spacingRule, space) && first == last;
 
@@ -471,7 +471,7 @@ void lefTechLayerSpacingEolParser::parse(const std::string& s,
                                          dbTechLayer* layer,
                                          odb::lefinReader* l)
 {
-  processRules(s, [layer, l](std::string& rule) {
+  processRules(s, [layer, l](const std::string& rule) {
     if (rule.find("ENDOFLINE") == std::string::npos) {
       l->warning(254,
                  "unsupported LEF58_SPACING property for layer {} :\"{}\"",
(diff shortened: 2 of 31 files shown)


src/odb/src/lefin/lefTechLayerSpacingEolParser.cpp
```diff
@@ -470,19 +471,13 @@ void lefTechLayerSpacingEolParser::parse(const std::string& s,
...
                                          odb::lefinReader* l)
 {
-  std::vector<std::string> rules;
-  boost::split(rules, s, boost::is_any_of(";"));
-  for (auto& rule : rules) {
-    boost::algorithm::trim(rule);
-    if (rule.empty()) {
-      continue;
-    }
+  processRules(s, [layer, l](std::string& rule) {
```
maliberty: Could rule be a const std::string& ?  (Same in all the other parsers)

jfgava: Only in lefTechLayerSpacingEolParser.cpp rule could be const.

The other parsers have this part 
rule += " ; ";

maliberty: Could that be moved into your helper?  It looks like the parsers expect a terminating ;  (I guess lefTechLayerSpacingEolParser is an oversight)

jfgava: Yes, it can be moved.

The terminating semicolon is optional (-lit(";")) in the lefTechLayerSpacingEolParser.cpp, and required for the other parsers (lit(";")).

The additional semicolon should not affect the lefTechLayerSpacingEolParser behavior and is good for consistency.
I'll make these changes

maliberty: Having made the ';' change can we now use const std::string& ?

jfgava: I tried here and got some compilation errors related to the lefTechLayerSpacingEolParser.cpp 
The other parsers are ok with the const std::string&.
I will investigate

jfgava: Done
Now using the const std::string&

Instead of passing iterators as arguments, I have created the iterators inside the parse method, the same way as the other parser files.

src/odb/src/lefin/lefTechLayerSpacingEolParser.cpp
```diff
@@ -362,11 +363,7 @@ void wrongDirSpaceParser(double value,
   sc->setWrongDirSpace(l->dbdist(value));
 }
 
-template <typename Iterator>
-bool parse(Iterator first,
-           Iterator last,
-           odb::dbTechLayer* layer,
-           odb::lefinReader* l)
+bool parse(std::string s, odb::dbTechLayer* layer, odb::lefinReader* l)
```
maliberty:
```suggestion
bool parse(const std::string& s, odb::dbTechLayer* layer, odb::lefinReader* l)
```

jfgava: I am having some issues when trying to change it to const.

https://github.com/The-OpenROAD-Project/OpenROAD/blob/cbe702ebacd2e598267cba1a26bb25be8292c4d1/src/odb/src/lefin/lefTechLayerSpacingEolParser.cpp#L454-L457

The problem is that the parse function from Boost library does not accept a const iterator.
I can create a string copy here before the iterators, or continue using it without const.

What do you suggest?

maliberty: I believe the problem is that many rules are declared
```
qi::rule<std::string::iterator, space_type>
```
and will now need to be
```
qi::rule<std::string::const_iterator, space_type>
```

jfgava: Yes. You are right.
I thought I had already tried that...

I'm checking the other parsers. Many parser files could be refactored with const.

src/odb/src/lefin/lefMacroPropParser.h
```diff
@@ -23,7 +23,7 @@ class lefMacroEdgeTypeParser
   {
   }
   void parse(const std::string&);
-  bool parseSubRule(std::string);
+  bool parseSubRule(const std::string&);
   void setRange(boost::fusion::vector<double, double>& params);
```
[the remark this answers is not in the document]
jfgava: I think I can add the #include <boost/fusion/container/vector.hpp> to solve this

Issue #7043, opened 2025-04-08
Break circular dependency between GRT and ANT


eder-matheus:
### Description

GRT code includes ANT to evaluate the violations from route guides during repair_antennas.

ANT code includes GRT to build the wires from the route guides.

We should not have this circular dependency between the tools.

### Suggested Solution

Do not include GRT in ANT. The functions that ANT uses from GRT can be moved inside ANT, then only GRT will need to include ANT.

PR #7572, opened 2025-06-13
ant: remove circular dependency with grt

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/7043

**Problem**

GRT need to include ANT in repair_antennas to get the nets with violation. Also, ANT was including GRT to build the net wires from the global routing guides. This created a circular dependency between the tools.

**Solution**

* Move functions from GRT to ODB/ANT in order to make it visible to ANT, removing the necessity of including GRT
* Created the WireBuilder class in ANT, which is responsible for creating the net wires from the route guides stored in ODB
  * Auxiliar functions were created to properly translate the guides to wires, specially to detect which pins connects to which guides.

Secure-ci is running with the modifications.


eder-matheus:
@maliberty I believe this is finally ready for review. I've fixed the last CI issues, now everything seems to be working as expected (i.e., we're replicating the results from the previous version of antenna checker).

FYI, I plan add multithreading to the wire creation, since it is a bit slow compared to the old version. But I decided to do it in a separate PR as it already has a lot to review.

maliberty:
I'll take a look.   Let me know if you need help with the Bazel failure.  I expect you need to adjust the dependencies.

src/odb/src/db/dbGuide.cpp
```diff
@@ -254,6 +264,26 @@ void dbGuide::setIsJumper(bool jumper)
...
+    is_connected_to_term = guide->is_connect_to_term_;
+  }
+  return is_connected_to_term;
+}
+
+void dbGuide::setIsConnectedToTerm(bool is_connected)
+{
+  _dbGuide* guide = (_dbGuide*) this;
+  _dbDatabase* db = guide->getDatabase();
+  if (db->isSchema(db_schema_guide_connected_to_term)) {
```
maliberty: Why is a schema check need here and isConnectedToTerm?

src/ant/src/AntennaChecker.cc
```diff
@@ -1062,6 +1061,46 @@ Violations AntennaChecker::getAntennaViolations(odb::dbNet* net,
...
 
+bool AntennaChecker::designIsPlaced()
+{
+  for (odb::dbBTerm* bterm : block_->getBTerms()) {
+    if (bterm->getFirstPinPlacementStatus() == odb::dbPlacementStatus::NONE) {
+      return false;
+    }
+  }
+
+  for (odb::dbNet* net : block_->getNets()) {
```
maliberty: Why not getInsts()?

src/ant/src/WireBuilder.hh
```diff
@@ -0,0 +1,113 @@
...
+#include "odb/dbWireGraph.h"
+#include "utl/Logger.h"
+
+namespace ant {
+
+struct GuidePoint
+{
+  odb::Point pos;
+  odb::dbTechLayer* layer;
+  friend bool operator<(const GuidePoint& pt1, const GuidePoint& pt2);
```
maliberty: Why does a struct need a friend?

src/ant/src/WireBuilder.hh
```diff
@@ -0,0 +1,113 @@
...
+  GuidePoint pt2;
+  bool isVia() const { return pt1.pos == pt2.pos; }
+  bool operator==(const GuideSegment& segment) const;
+};
+
+struct GuidePtPins
+{
+  std::vector<odb::dbBTerm*> bterms;
+  std::vector<odb::dbITerm*> iterms;
+  bool connected;
```
maliberty: Give a default value for safety

eder-matheus:
> I'll take a look. Let me know if you need help with the Bazel failure. I expect you need to adjust the dependencies.

@maliberty I'm struggle to make bazel work. My last commit is another attempt on it. Unfortunately I can't build it with bazel locally, so I need to wait for the CI.

PR #7575, opened 2025-06-14
bazel-orfs: bump

oharboe:
bazel-orfs now uses PYTHON_EXE to inject Bazel venv dependency, so the few python dependencies that ORFS has, are not requird in the system Python when running bazel-orfs. Nice for CI with bare bones environments.


oharboe:
@maliberty pr-head CI problem, pr-merge worked, so merge?

PR #7582, opened 2025-06-15
utl: move gzip handling to StreamHandler

gadfort:
This should make the gzip handling uniform for the places where the stream handler is used and voids needing to copy the code around.

Changes:
- make cdl use ostream for writing
- move db and def gzip writing to use utl streamhandler' gzip.


src/OpenRoad.cc
```diff
@@ -456,8 +457,7 @@ void OpenRoad::readDb(const char* filename, bool hierarchy)
   stream.open(filename, std::ios::binary);
 
   try {
-    const std::string name(filename);
-    if (name.length() >= 3 && name.compare(name.length() - 3, 3, ".gz") == 0) {
+    if (boost::ends_with(std::string(filename), ".gz")) {
       boost::iostreams::filtering_streambuf<boost::iostreams::input> inbuf;
       inbuf.push(boost::iostreams::gzip_decompressor());
       inbuf.push(stream);
```
maliberty, line 463: Do you think we should do something similar to this for gz read?

maliberty (approved):

src/OpenRoad.cc
gadfort: yeah, we could, just need a class to wrap `ifstream`

src/utl/test/cpp/TestCFileUtils.cpp
```diff
@@ -161,6 +163,29 @@ TEST(Utl, stream_handler_write_and_read)
...
+    StreamHandler sh(filename);
+    std::ostream& os = sh.getStream();
+    os.write(kTestData.c_str(), kTestData.size());
+  }
+
+  std::ifstream is(filename, std::ios_base::binary);
+  boost::iostreams::filtering_streambuf<boost::iostreams::input> inbuf;
+  inbuf.push(boost::iostreams::gzip_decompressor());
+  inbuf.push(is);
+  std::istream zstd_uncompressed(&inbuf);
```
QuantamHD, line 181: Why zstd here? Looks like gzip.

gadfort: I'll fix the name in #7585 that was a copy paste mistake

PR #7591, opened 2025-06-16
bazel: OpenROAD/sta tests runs with cfg=exec, which is opt by default

oharboe:


maliberty:
That's more variation than I expect in runtime.  I'll re-run it to see if it is consistent or not.

oharboe:
> That's more variation than I expect in runtime. I'll re-run it to see if it is consistent or not.

30 minutes this time.

maliberty:
I wonder if it was counting time queued?

oharboe:
> I wonder if it was counting time queued?

Hmmm...

Local use case to build and test is now simply:

    bazelisk test ...

PR #7599, opened 2025-06-16
bazel: update READMEs

oharboe:


docs/user/Bazel.md
```diff
@@ -2,9 +2,9 @@
 
 The main use-case for Bazel is to make modifications to OpenROAD and run local tests before creating a PR.
 
-After [installing Baselisk](https://bazel.build/install/bazelisk), the command line below will discover all tests starting in the current working directory and build all dependencies, including openroad:
+After [installing Baselisk](https://bazel.build/install/bazelisk), the command line below will discover all tests starting in the current working directory and build all dependencies, including openroad and opensta:
 
-    bazelisk test -c opt ...
+    bazelisk test ...
```
maliberty, line 7: It would be good to describe how to get the prior results (faster compile with less opt) or a debug build too.

oharboe: Unsure what you mean "prior results", you mean cached?

What is a "faster compile with less opt"?

I'm hesitant to document how bazel works. Especially with ChatGPT being better and better at explaining it. We're going to end up with out of date and worse advice than ChatGPT very quickly...

I think it makes sense to add some "constraints", some text for some important use-cases, to the documentation that can be fed into ChatGPT for elaboration.

Can we merge fix and then discuss the broader issue of what to document w.r.t. bazel?

PR #7609, opened 2025-06-18
bazel: profiling use case documentation

oharboe:


docs/user/Bazel.md
```diff
@@ -38,6 +38,43 @@ Run ORFS flow and use debugger as usual for ORFS:
...
+At this point, use your favorite preformance tool, such as the Linx Perf tool:
+
+    perf top
+
+Perhaps attach gdb and use ctrl-c from the command line? Use gdb with an IDE, emacs, or vim?
+
+    $ gdb bazel-bin/openroad
+    [deleted]
+    (gdb) attach 578603
+    Attaching to program: /home/oyvind/.cache/bazel/_bazel_oyvind/896cc02f64446168f604c13ad7b60f8b/execroot/_main/bazel-out/k8-fastbuild/bin/openroad, process 578603
```
gadfort, line 62: what happens if the running tests executable is not the same as the one linked in bazel-bin/openroad?

oharboe: Unsure what you mean. bazel-bin points to the most recent bazelisk build.,

maliberty (approved):

docs/user/Bazel.md
gadfort: I just meant that in the example those could actually be two different binaries (just depends if they were recompiled while something was running, ie. changing the most recent build).

oharboe: not bazel or openroad specific: dont do that. :-)

PR #7610, opened 2025-06-18
gpl: use deafult unscaled overflow 

gudeh:
Use default unscaled overflow for checking if routability or timing-driven should run.
The scaled version should be left for solving overflow with macros (normal distribution scaling).

This solves a segmentation fault occurring with @joaomai in https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-mpl-boundary-penalty/4/pipeline/ where we would end up calling the timing-driven iteration during the routability iterations.


eder-matheus (approved):

gudeh:
I ran a secure-CI in: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-use-unscaled-overflow/1/pipeline

PR #7617, opened 2025-06-19
mpl: mixed leaf split by replacement only

### LINKED ISSUE #7020: MPL's mixed leaf splitting should only work by replacement
### Description

HierRTLMP relies on the RTL hierarchy to create a hierarchical tree of clusters used to perform the macro placement. The multilevel autoclustering process (which creates the cluster tree) maps the RTL hierarchy by converting modules to clusters and then recursively breaks/merges clusters based on thresholds of the depth level. At the end of this process, we'll have a tree of only mixed clusters whose leaves will be split into std cell and macro clusters. 

This split can be done in two ways:
- Replacement: The mixed leaf will be replaced by one std cell cluster and N macro clusters (i.e., the mixed leaf will be literally dissolved). _The criteria used for this split type is macro dominance (99% dominance), that is, the amount of macro area in the mixed leaf._
- Sub-tree addition:  The mixed leaf will become a subsequent hierarchical level composed by one std cell cluster and N macro clusters.

For either of these techniques we also add a virtual connection between the split clusters in order to guide the annealer to keep them close to each other.

The rationale of this implementation with two types of split was based on a test case in which a very small cluster of std cells ended up far from its virtually connected macros in the final cluster placement result. However for all the cases we have in our CI, the sub-tree addition leads to the creation of levels that are clearly unneeded due to the size of the designs. _I.e., designs with few instances and small area end up with an unnecessary amount of clustering hierarchy._

One example of this can be seen in the root-Level annealing of `ngt45/bp_be_top`:

Image
  
The disposition of the mixed cluster might generate an unrecoverable small area for placing the 8 RAMs array and that might lead to congestion problems:

### Suggested Solution

Remove the sub-tree adding technique. QoR should tell us if this is actually a good thing.

### Additional Context

_No response_

joaomai:
Closes #7020.

There are clusters made of standard cells and macros as leaves in the cluster tree built by autoclustering in MPL that need to be broken into macro clusters and standard cell clusters. Currently, there are two ways of doing this:
- Append a new sub-tree made of a standard cell cluster and a macro cluster where the mixed cluster was on the hierarchy (sub-tree)
- Directly appending the standard cell cluster and macro cluster to the mixed leaf parent (replacement).

This PR updates the mixed leaf splitting function to always break a mixed leaf using the replacement method.

### DIFF
--- a/src/mpl/src/clusterEngine.cpp
+++ b/src/mpl/src/clusterEngine.cpp
@@ -1829,7 +1829,6 @@ void ClusteringEngine::breakMixedLeaves(
 void ClusteringEngine::breakMixedLeaf(Cluster* mixed_leaf)
 {
   Cluster* parent = mixed_leaf->getParent();
-  const float macro_dominated_cluster_ratio = 0.01;
 
   mapMacroInCluster2HardMacro(mixed_leaf);
 


AcKoucher:
Replaced by #7862.

PR #7619, opened 2025-06-19
mpl: honor macro thresholds 

joaomai:
During clustering, there were places that only the thresholds for standard cells were checked. 
This PR updates MPL to include the thresholds for macros in the appropriate places.

The change breaks asap7/mock-array, which will now require its thresholds to me manually selected instead of calculated by the tool. Future improvements may be done to make peculiar designs such as mock-array to work with the automatic calculations for thresholds.

Should be merged alongside ORFS [#3304](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3304)


oharboe:
@maliberty @joaomai Just FYI, I'm working on improving and documenting the use-case to create `make issue` reports for failing bazel-orfs test cases in OpenROAD.

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3256

oharboe:
What is the user-experience here?

Will there an ORFS variable that the user has to tweak to some magic value?

oharboe:
@joaomai @maliberty How to create a `make issue` file for a bazel-orfs target(ref. MockArray_floorplan failure in this PR):

https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/docs/user/Bazel.md#creating-an-orfs-issue-with-bazel-orfs-targets-using-_deps-targes

AcKoucher:
@oharboe The rationale here is that the clustering engine inside MPL has a bug where we don't respect the thresholds that tell us "this level should have at least N macros and at max M macros" as we're clustering.

Briefly explaining: MPL creates an internal hierarchy of clusters (groups of std cells or macros) and places them hierarchically so that SA can deal with smaller problems, one at the time, instead of a huge one.

That said,

> What is the user-experience here?
> Will there an ORFS variable that the user has to tweak to some magic value?

Currently, mock-array doesn't require user settings, because of this bug. So, answering your question, with this fix, it will require manually set thresholds (there will be a PR for that) until we have a more elaborate solution.

oharboe:
Interesting. Look forward to seeing the PR. Is this an AutoTuner paramater candidate?

To test mock-array from the OpenROAD repository, run:

    bazelisk test test/orfs/mock-array:MockArray_test

test/orfs/mock-array/BUILD will need your modification

@maliberty FYI

AcKoucher:
> Is this an AutoTuner paramater candidate?

It is, but I don't think that mock-array would benefit from auto-tuning these parameters specifically as there's not much to vary after all. I may be wrong.

The idea is that MPL should be smart enough to come up with these thresholds based on the design data.

src/mpl/src/clusterEngine.cpp
```diff
@@ -1490,7 +1492,8 @@ void ClusteringEngine::breakLargeFlatCluster(Cluster* parent)
 {
   // Check if the cluster is a large flat cluster
   if (!parent->getDbModules().empty()
-      || parent->getLeafStdCells().size() < max_std_cell_) {
+      || (parent->getLeafStdCells().size() <= max_std_cell_
+          && parent->getLeafMacros().size() <= max_macro_)) {
```
AcKoucher: As this method is a special case for cluster breaking, I think it would be best to remove this check and have something like
```
if (isLargeFlatCluster(cluster)) {
  breakLargeFlatCluster(cluster);
}
```
amongst the `breakLargeFlatCluster()` calls.

oharboe:
@maliberty @vvbandeira @joaomai @AcKoucher FYI w.r.t. updating bazel-orfs test cases inside OpenROAD.

[Changes needed](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3304):

```
$ git diff
diff --git a/test/orfs/mock-array/BUILD b/test/orfs/mock-array/BUILD
index 192e333459..b4d5b0e58c 100644
--- a/test/orfs/mock-array/BUILD
+++ b/test/orfs/mock-array/BUILD
@@ -130,6 +130,10 @@ orfs_flow(
         ),
         "MACRO_PLACE_HALO": "0 2.16",
         "RTLMP_BOUNDARY_WT": "0",
+"RTLMP_MAX_INST" :"250",
+"RTLMP_MIN_INST" :"50",
+"RTLMP_MAX_MACRO" :"64",
+"RTLMP_MIN_MACRO" :"8",
         "PDN_TCL": "$(PLATFORM_DIR)/openRoad/pdn/BLOCKS_grid_strategy.tcl",
         "MACRO_ROWS_HALO_X": "0.5",
         "MACRO_ROWS_HALO_Y": "0.5",
```

To [update the rules-base.json](https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/test/orfs#updating-rules_json-files) for mock-array:

    bazelisk run test/orfs/mock-array:MockArray_update

Now tests will pass:

    bazelisk test test/orfs/mock-array:MockArray_test

src/mpl/src/clusterEngine.h
```diff
@@ -202,6 +202,7 @@ class ClusteringEngine
   void createCluster(odb::dbModule* module, Cluster* parent);
   void createCluster(Cluster* parent);
   void updateSubTree(Cluster* parent);
+  bool isLargeFlatCluster(Cluster* cluster);
```
AcKoucher: This method should be const as it doesn't alter members.

oharboe:
@joaomai And you'll need `bazelisk run test/orfs/mock-array:MockArray_update` to update rules-base.json:

https://github.com/The-OpenROAD-Project/OpenROAD/pull/7619#issuecomment-3055703833

AcKoucher (commented):
The code looks ok to me.
Please, address @oharboe 's comments.

EDIT: It looks like this is actually good to go. We shouldn't see metrics changes as the results will be the same as we currently have. The check failure looks like an IT issue - I set to rerun.

maliberty (approved):

PR #7622, opened 2025-06-19
[DO NOT REVIEW] docs: bazel-orfs use-case

oharboe:


oharboe:
Hmm... @hovind This use-case is broken for openroad where the openroad binary does not come from the docker image, but from //:openroad. We keep patching up the _deps code in bazel-orfs.... I'm going to have a go at making `--sandbox_debug` do the job. If I can make `--sandbox_debug` do the job, then :crossed_fingers: we can retire _deps targets....

PR #7571, opened 2025-06-13
gui: add descriptor for dbScanInst

AcKoucher:
Requires #7642

Beginning of the work for #7740.

Example (combining the changes with the required PR):

PR #7642, opened 2025-06-23
dft: ensure we create Enable pin when creating dbScanInsts

AcKoucher:
### Context

As I was testing #7571 (using _scan_architect_no_mix_sky130.tcl_), I got a weird exception throw when trying to access the Scan Enable of a dbScanInst. Some investigation showed that we are forgetting to set the Scan Enable when we create the dbScanInst's.

### Changes
1. Add interface to retrieve the ScanEnable of `dft::ScanInst` and its implementation in both the current one bit cell type that we have and the scan mock test cell;
2. Ensure that we set the Scan Enable of `dbScanInst`s.

### DIFF
--- a/src/dft/src/Dft.cpp
+++ b/src/dft/src/Dft.cpp
@@ -111,6 +111,9 @@ void Dft::executeDftPlan()
       odb::dbInst* db_inst = db_block->findInst(inst_name.c_str());
       odb::dbScanInst* db_scaninst = db_scanlist->add(db_inst);
       db_scaninst->setBits(scan_cell->getBits());
+      ScanLoad scan_enable = scan_cell->getScanEnable();
+      std::visit([&](auto&& pin) { db_scaninst->setScanEnable(pin); },
+                 scan_enable.getValue());
       auto scan_in_term = scan_cell->getScanIn().getValue();
       auto scan_out_term = scan_cell->getScanOut().getValue();
       db_scaninst->setAccessPins(
--- a/src/dft/src/Dft.cpp
+++ b/src/dft/src/Dft.cpp
@@ -106,18 +106,26 @@ void Dft::executeDftPlan()
     db_part->setName(kDefaultPartition);
     odb::dbScanList* db_scanlist = odb::dbScanList::create(db_part);
 
+    std::optional<ScanDriver> sc_enable_driver = chain->getScanEnable();
+
     for (const auto& scan_cell : chain->getScanCells()) {
       std::string inst_name(scan_cell->getName());
       odb::dbInst* db_inst = db_block->findInst(inst_name.c_str());
       odb::dbScanInst* db_scaninst = db_scanlist->add(db_inst);
       db_scaninst->setBits(scan_cell->getBits());
+      if (sc_enable_driver.has_value()) {
+        std::visit(
+            [&](auto&& sc_enable_term) {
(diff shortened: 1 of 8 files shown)


AcKoucher:
@fgaray I can't request your review. Could you please take a look?

src/dft/src/Dft.cpp
```diff
@@ -106,18 +106,26 @@ void Dft::executeDftPlan()
...
       std::string inst_name(scan_cell->getName());
       odb::dbInst* db_inst = db_block->findInst(inst_name.c_str());
       odb::dbScanInst* db_scaninst = db_scanlist->add(db_inst);
       db_scaninst->setBits(scan_cell->getBits());
+      if (sc_enable_driver.has_value()) {
+        std::visit(
+            [&](auto&& sc_enable_term) {
+              db_scaninst->setScanEnable(sc_enable_term);
+            },
+            sc_enable_driver.value().getValue());
```
fgaray: You already know that sc_enable_drive has a value, you can just use the "->" operator to access getValue

See: https://en.cppreference.com/w/cpp/utility/optional/operator%252A.html

src/dft/src/Dft.cpp
```diff
@@ -106,18 +106,26 @@ void Dft::executeDftPlan()
...
+
     for (const auto& scan_cell : chain->getScanCells()) {
       std::string inst_name(scan_cell->getName());
       odb::dbInst* db_inst = db_block->findInst(inst_name.c_str());
       odb::dbScanInst* db_scaninst = db_scanlist->add(db_inst);
       db_scaninst->setBits(scan_cell->getBits());
+      if (sc_enable_driver.has_value()) {
+        std::visit(
+            [&](auto&& sc_enable_term) {
+              db_scaninst->setScanEnable(sc_enable_term);
```
fgaray: This is setting the scan enable of the cell to be the scan enable of the scan chain.

Those are different pins. The scan enable of the scan chain drives all the scan enable loads of the insts.

I think the correct code is something like

```
ScanLoad scan_enable = scan_cell->getScanEnable();
//etc
```

If you see a diagram:

image

SE in this case is the top level scan enable driver that is connected to each inst's scan enable load.

src/dft/src/Dft.cpp
```diff
@@ -106,18 +106,26 @@ void Dft::executeDftPlan()
...
+            },
+            sc_enable_driver.value().getValue());
+      }
       auto scan_in_term = scan_cell->getScanIn().getValue();
       auto scan_out_term = scan_cell->getScanOut().getValue();
       db_scaninst->setAccessPins(
           {.scan_in = scan_in_term, .scan_out = scan_out_term});
     }
 
-    std::optional<ScanDriver> sc_enable_driver = chain->getScanEnable();
```
fgaray: This is correct. This is the top level driver.

src/dft/src/Dft.cpp
AcKoucher: Thanks for pointing that out. The connectivity confused me. I'm fixing this.

AcKoucher:
@fgaray I removed the dbScanInst code as I understand it is no longer necessary (and perhaps incorrect) after addressing your comments. Could you take another look?

fgaray:
LGTM, thanks!

@maliberty

maliberty (approved):

Issue #7581, opened 2025-06-15
Dynamic node fails with routing congestion


gadfort:
### Describe the bug

Between https://github.com/The-OpenROAD-Project/OpenROAD/compare/4b729ea474e9bd0a3513579b41a8c96fe20dc75e...6e29df0746191b48963a35bf45422d76e2c9922c something changed that is causing a routing congestion issue on a previously working design.
It appears that the change that broke this was after 588bc37d48f2a5ffef6f2b0d93f24e1810a13038

### Expected Behavior

No routing congestion issue

### Environment

```shell
OpenROAD v2.0-22186-g6e29df0746 
Features included (+) or not (-): +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
```

### To Reproduce

[sc_issue_dynamic_node_job0_skywater130_sky130hd_route.global0_20250615-011417.tar.gz](https://github.com/user-attachments/files/20745947/sc_issue_dynamic_node_job0_skywater130_sky130hd_route.global0_20250615-011417.tar.gz)

```
tar xvf sc_issue_dynamic_node_job0_skywater130_sky130hd_route.global0_20250615-011417.tar.gz
cd sc_issue_dynamic_node_job0_skywater130_sky130hd_route.global0_20250615-011417
./run.sh
```

### Relevant log output

```shell
[INFO GRT-0101] Running extra iterations to remove overflow.
[INFO GRT-0103] Extra Run for hard benchmark.
[WARNING GRT-0230] Congestion iterations cannot increase overflow, reached the maximum number of times the total overflow can be increased.
[INFO GRT-0197] Via related to pin nodes: 52735
[INFO GRT-0198] Via related Steiner nodes: 2330
[INFO GRT-0199] Via filling finished.
[INFO GRT-0111] Final number of vias: 95509
[INFO GRT-0112] Final usage 3D: 393857

[INFO GRT-0096] Final congestion report:
Layer         Resource        Demand        Usage (%)    Max H / Max V / Total Overflow
---------------------------------------------------------------------------------------
li1                  0             0            0.00%             0 /  0 /  0
met1             30211         25809           85.43%             4 /  1 / 799
met2             38127         36714           96.29%             0 / 10 / 3532
met3             32346         27842           86.08%             4 /  1 / 826
met4             14910         12881           86.39%             0 /  5 / 1358
[...]

PR #7644, opened 2025-06-24
gpl: fix insts start position

gudeh:
Both `-skip_io` and `-skip_initial_place` options set the CG initial placement procedure to zero max iterations.

However, using the zero-iteration condition to decide whether to use existing database instance positions causes issues when `-skip_initial_place` is used. In this case, we end up using the final placement from a previous run, which is not desirable.

The expected behavior when skipping CG initial placement (e.g., via `-skip_initial_place`) is to place instances at the center of the domain, not to reuse existing placement. For example, in stage 3-3 in ORFS, it might lead to using the final placement from stage 3-2 in ORFS.

I believe it is better to explicitly check for `-skip_io` when deciding to use database positions.

### DIFF
--- a/src/gpl/src/initialPlace.cpp
+++ b/src/gpl/src/initialPlace.cpp
@@ -137,6 +137,8 @@ void InitialPlace::placeInstsCenter()
                               domain_y_max - (domain_y_max - domain_y_min) / 2);
       ++count_region_center;
     } else if (pbc_->skipIoMode() && db_inst->isPlaced()) {
+      // It is helpful to pick up the placement from mpl if available,
+      // particularly when you are going to run skip_io.
       const auto bbox = db_inst->getBBox()->getBox();
       inst->setCenterLocation(bbox.xCenter(), bbox.yCenter());
       ++count_db_location;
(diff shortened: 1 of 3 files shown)


gudeh:
I suggested @gadfort to use `-skip_initial_place` gpl parameter as a potential workaround to the issue reported in https://github.com/The-OpenROAD-Project/OpenROAD/issues/7581. Although I noticed another issue which this PR should be able to fix.

maliberty:
This will need a secure CI and unit test fixes.

gudeh:
I ran a secure-CI at [secure-gpl-fix-insts-start-position](https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-fix-insts-start-position/1/pipeline). So far public designs are all green.

I do not expect any functional change with this PR, only to enable the usage of `-skip_initial_place` to use the core center location if it is used.

gudeh:
All green on secure-CI, including private designs.

maliberty (approved):

PR #7650, opened 2025-06-24
grt: append nodes and edges from removed nets when merging nets

eder-matheus:
**Problem**
When merging the routing of two nets after buffer removal, the FastRouteCore data regarding edges and nodes of the deleted net was just being deleted, but the congestion information of them wasn't updated. This caused multiple congestion issues during post-GRT incremental repair, since the congestion data of these edges was never updated.

**Solution**
Instead of just deleting the edges and nodes of the deleted nets, append them to the edges and nodes of the preserved net (i.e., the net that will "receive" the routing of the deleted net). This way, the congestion data can be properly updated in future iterations of the global routing tool.


maliberty (approved):

eder-matheus:
@maliberty FYI, I'll hold the merge of this PR until I have a complete secure-ci. It didn't went through due to the sta issue.

PR #7663, opened 2025-06-26
gpl: fix code scan defects

gudeh:
This PR addresses all high (1) and medium (6) severity defects recently reported in the GPL code. Most of the fixes are straightforward, such as adding nullptr checks. For one of the defects, I chose to remove an unused function that had a clear issue (a pointer was set to nullptr and then dereferenced afterwards).

There are still 4 low-severity defects left unaddressed. They follow the pattern below:
```
CID 1522333: (#1 of 1): Calling risky function (DC.WEAK_CRYPTO)
dont_call: rand should not be used for security-related applications, because linear congruential algorithms are too easy to break. 
```

I’m unsure whether we want to replace `rand()` with a more robust RNG like `std::mt19937`. While the defect message warns about cryptographic weakness, I believe the current usages of rand() are not related to any security-sensitive functionality.


gudeh:
I separated the defects with individual commits for each. 

I also triggered some secure-CI runs in [secure-gpl-fix-code-scan-defects](https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-fix-code-scan-defects/5/pipeline)

gudeh:
The jenkins CI is reporting a failing test only for Ubuntu 24: 
image

Failing is happening with incremental02.tcl

image

I do not get the error on my gcp machine. Are we able to access the logs from the runs in jenkins? I could not find it.

src/gpl/src/solver.cpp
```diff
@@ -20,16 +20,20 @@ ResidualError cpuSparseSolve(int maxSolverIter,
...
-  error.x = solver.error();
+  if (solver.info() == Eigen::Success) {
+    residual_error.x = solver.error();
+  }
 
   solver.compute(placeInstForceMatrixY);
   instLocVecY = solver.solveWithGuess(fixedInstForceVecY, instLocVecY);
-  error.y = solver.error();
-  return error;
+  if (solver.info() == Eigen::Success) {
```
maliberty: If eigen fails we will return uninit values.  I think it should be an error in that case.

maliberty:
The tests now fail on all platforms.  Hopefully you can reproduce it.

src/gpl/src/solver.cpp
```diff
@@ -20,16 +20,28 @@ ResidualError cpuSparseSolve(int maxSolverIter,
...
+    residual_error.y = solver.error();
+  } else {
+    logger->error(utl::GPL, 155, "Eigen solver failed on Y at iter {}.", iter);
+    residual_error.y = std::numeric_limits<float>::quiet_NaN();
```
rovinski, line 42: `error` throws so this line would have to come first to have an effect

gudeh:
Well, it seems the GPL test at `src/gpl/test/incremental02.tcl `was failing with `solver.info() != Eigen::Success`, although this was previously ignored.

I assume the CG initial placement continues execution even after receiving potentially garbage values from Eigen, resulting in a poor initial placement for Nesterov.

Usually, the conjugate gradient residual value decreases with each iteration, as observed in nangate45/aes:
image

Although, in incremental02 test the solver error is masked, the residual values fluctuate instead of converging:
image


Using the GUI, I confirmed that the initial placer appears to move the unlocked instances randomly. Despite this, Nesterov still completes the placement, even with the initial placement behaving erratically.

I believe incremental GPL is more susceptible to such failures due to the high number of locked instances.

Maybe we should not throw an error in such case and just make nesterov to initiate with instances on the middle of the core.

gudeh:
I implemented the solution I suggested. I also had a quick peak on the behavior with gui debug mode and nesterov placement seems more smooth with this modification for incremental02 test.

The previous secure-CI I triggered were reporting an issue with at least one private design. I am triggering a new run to investigate.

maliberty:
That sounds like an ok solution

gudeh:
We got 4 tests in `OpenROAD/test/` which are currently failing with `solver.info() != Eigen::Success`. Here are their behaviors on master. With the new update we would only put all instances in the middle of the core.


`aes_nangate45.tcl`, which converges during CG initial place even with Eigen fail. It just moves a smaller cluster upwards, I highlighted its connections on the image:
```
[InitialPlace]  Iter: 1 conjugate gradient residual: 0.00046178 HPWL: 837067900
[InitialPlace]  Iter: 2 conjugate gradient residual: 0.00002337 HPWL: 656049444
[InitialPlace]  Iter: 3 conjugate gradient residual: 0.00002931 HPWL: 655172079
[InitialPlace]  Iter: 4 conjugate gradient residual: 0.00002877 HPWL: 655639162
[InitialPlace]  Iter: 5 conjugate gradient residual: 0.00002162 HPWL: 655347033
[InitialPlace]  Iter: 6 conjugate gradient residual: 0.00002532 HPWL: 655720247
[InitialPlace]  Iter: 7 conjugate gradient residual: 0.00000796 HPWL: 655430199
```
image


And this is `aes_sky130hs.tcl`, does not converge and stops after the default 20 max CG initial iterations:
```
[InitialPlace]  Iter: 1 conjugate gradient residual: 0.00065200 HPWL: 876409820
[InitialPlace]  Iter: 2 conjugate gradient residual: 0.00033705 HPWL: 693254925
[InitialPlace]  Iter: 3 conjugate gradient residual: 0.00032558 HPWL: 691409079
[InitialPlace]  Iter: 4 conjugate gradient residual: 0.00015019 HPWL: 691959689
[InitialPlace]  Iter: 5 conjugate gradient residual: 0.00019565 HPWL: 691368400
[InitialPlace]  Iter: 6 conjugate gradient residual: 0.00014839 HPWL: 691799089
[InitialPlace]  Iter: 7 conjugate gradient residual: 0.00011467 HPWL: 691308190
[InitialPlace]  Iter: 8 conjugate gradient residual: 0.00009899 HPWL: 691663041
[InitialPlace]  Iter: 9 conjugate gradient residual: 0.00012311 HPWL: 691119937
[InitialPlace]  Iter: 10 conjugate gradient residual: 0.00012633 HPWL: 691544832
[InitialPlace]  Iter: 11 conjugate gradient residual: 0.00049594 HPWL: 691090897
[InitialPlace]  Iter: 12 conjugate gradient residual: 0.00014961 HPWL: 691556273
[InitialPlace]  Iter: 13 conjugate gradient residual: 0.00013032 HPWL: 691078631
[InitialPlace]  Iter: 14 conjugate gradient residual: 0.00010355 HPWL: 691439530
[InitialPlace]  Iter: 15 conjugate gradient residual: 0.00011363 HPWL: 690996493
[InitialPlace]  Iter: 16 conjugate gradient residual: 0.00010711 HPWL: 691350676
[InitialPlace]  Iter: 17 conjugate gradient residual: 0.00012241 HPWL: 690937561
[InitialPlace]  Iter: 18 conjugate gradient residual: 0.00011744 HPWL: 691292109
[InitialPlace]  Iter: 19 conjugate gradient residual: 0.00015518 HPWL: 690878721
[InitialPlace]  Iter: 20 conjugate gradient residual: 0.04246838 HPWL: 691232620
```

image


With `ibex_sky130hs` we also do not converge and we get this behavior from CG initial place which sometimes put instances right on top of external pins:
image


With the new change we just put all instances by the center of the core.

gudeh:
I updated the tests for `OpenROAD/tests`, but the secure-CI is triggering fatal failures:
image

We also had 5 metrics updates on public designs. And we have 51 designs failing with `solver.info() != Eigen::Success`

maliberty:
Would it work better to keep the last passing result rather than going to all instances in the center?

gudeh:
I realized we have other options for [info() return object](https://eigen.tuxfamily.org/dox/group__enums.html#ga85fad7b87587764e5cf6b513a9e0ee5e). If we do `if (solver.info() == Eigen::NoConvergence || solver.info() == Eigen::Success)` we should cover actual errors and not change the functional behavior.

This way we can focus on the code scan issues on this PR, and explore further improvements for CG initial place on another PR.

gudeh:
The latest code version is showing the same behavior as master, with the addition of covering actual errors such as `NumericalIssue` and `InvalidInput`. 

The secure-CI reported an error with private PDK,  and the OR CI reported an error with bazel. I asked @vvbandeira for help with those.

gudeh:
All green on secure-CI (1 private design finishing)

PR #7666, opened 2025-06-27
Add `@boost.{stacktrace,icl,graph}` dependency where needed.

hzeller:
The bazel build didn't compile without these anymore.

Why did it compile before ? I suspect the system-install of boost was leaking through; the headers in OR are using angle brackets for includes (implying that the headers are on the system), which is why it was probably not discovered.


src/pdn/BUILD
```diff
@@ -60,6 +60,7 @@ cc_library(
         "//src/utl",
         "@boost.geometry",
         "@boost.polygon",
+        "@boost.stacktrace",
```
gadfort, line 63: This dependency comes from the swig file, should it be expressed tcl_wrap_cc instead? (This is true for all tools)

hzeller: Ideally yes, but [`tcl_wrap_cc()`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/bazel/tcl_wrap_cc.bzl) is only generating the swigged code it seems, not providing a `cc_library()` (which would be cool).

Probably something to refactor later (it then could then also include the `@tk_tcl//:tcl` dependency)

PR #7671, opened 2025-06-27
drt: getLayer, getTech and Layer function calls.

bnmfw:
This is just some simple code cleaning. I got sick of seeing this verbose nothing structure all overt drt:
`getDesign()->getTech()->getLayer(laye_num)`
Most of the tools already were able to just get the Tech directly. I made a call for them to get the Layer directly also.
I also applied the `isRouting()`, `isHorizontal()` and `isVertical()` calls of the layers when appropriate.
That is it, just all over drt.
This essentially turns things like:
`getDesign()->getTech()->getLayer(layer_num)->getDir() == dbTechLayerDir::HORIZONTAL`
into the much cleaner:
`getLayer(layer_num)->isHorizontal()`


src/drt/src/dr/FlexDR_end.cpp
```diff
@@ -74,7 +74,7 @@ void FlexDRWorker::endRemoveNets_pathSeg(
         // change boundary style to ext if orig pathSeg crosses boundary
         if (end.y() > routeBox.yMin()) {
           frSegStyle style = pathSeg->getStyle();
-          style.setEndStyle(frEndStyle(frcExtendEndStyle), style.getEndExt() /*getTech()->getLayer(pathSeg->getLayerNum())->getWidth() / 2*/);
+          style.setEndStyle(frEndStyle(frcExtendEndStyle), style.getEndExt() /*getLayer(pathSeg->getLayerNum())->getWidth() / 2*/);
```
osamahammad21: very weird inline comment. Should be handled in a separate PR.

bnmfw: Oh yeah, I also thought this was weird. I did not case by case when changing these, I used some replace all to change the many instances. I guess I can change comments to be left as they were

PR #7673, opened 2025-06-28
gui: fix color change to vias and allow site to be selected

gadfort:
Changes:
- while looking into the via visibility issue, I noticed that the color of the via is actually not updated correctly when the display controls changes the color. This ensures that is correct.
- when selecting the site from the rows options, the site is now selected in the inspector.

### DIFF
--- a/src/gui/src/displayControls.cpp
+++ b/src/gui/src/displayControls.cpp
@@ -1021,16 +1021,13 @@ void DisplayControls::displayItemSelected(const QItemSelection& selection)
     if (!user_data.isValid()) {
       continue;
     }
-    auto* tech_layer = user_data.value<dbTechLayer*>();
-    if (tech_layer != nullptr) {
+
+    if (auto* tech_layer = user_data.value<dbTechLayer*>()) {
       emit selected(Gui::get()->makeSelected(tech_layer));
+    } else if (auto* site = user_data.value<odb::dbSite*>()) {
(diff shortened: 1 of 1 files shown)


maliberty:
I pushed a small change to make the code more readable.  I don't understand the via color issue you are fixing but I see it still doesn't work entirely correctly.  For example if I have:

image

and I changed the via color I see it works:
image

but if I then change the M1 color it also changes V1:
image

gadfort:
@maliberty the issue was that when the metal color changes we changed the color in the display controls, but the actual color used to draw was not updated. This just ensures that the color displayed is the one used to draw.

maliberty:
I tried in master and I don't see a problem:
image

maliberty:
Do you want to address the issue I mention above or leave it for another PR?

gadfort:
@maliberty if you change the color of M2 to blue, then the via color (displayed will be blue, the color used to draw will remain red), changing the via color by itself is not affected. The only thing the added code it doing is ensure the color in the display control is actually the color used to draw the vias. (whether or not that is a good idea is a issue issue).

image

Notice the V2 color:
image

maliberty (approved):

maliberty:
This is fine but I do think we should make the via color more independent of the enclosure layer(s)

gadfort:
> This is fine but I do think we should make the via color more independent of the enclosure layer(s)

I agree, I have some ideas I want to try.

PR #7676, opened 2025-06-28
gui: change placement, pin, and power density heatmaps to use core ar…

gadfort:
…ea instead of die area for map bounds

Changes:
- heatmaps that only need to cover the core area uses the the core area to build the map.
- add a check for nullptr in `addToMap` (this should not happen right now, but once a Polygon core area can be used some of the map sites will be blank.

Notes:
- when the Polygon floorplanning is available and the db has a `getCorePolygon()`, it should be relatively easy to update this to handle that correctly.


maliberty (approved):

maliberty:
@eder-matheus see the note on Polygon floorplanning

PR #7680, opened 2025-06-29
gui: separate routing and cut colors

gadfort:
Changes:
 - when routing layer color changes, the cut layers do not change with them (this also makes it easier for users to specify the color of a layer they want without having to undo the cut layer change as well).
 - change default cut colors

Initial cut colors were chosen by calling `color.lighter()` or `color.darker()` to generate a color that is "related" to the initial metal color, but with some contrast to the metal.

If someone wants to suggest a different color for the specific layer, feel free to do so. My goal here was to get a starting point that I thought was sufficient to begin the discussion.

Examples m1-m10:
m1
m2
m3
m4
m5
m6
m7
m8
m9
m10

### DIFF
--- a/src/gui/src/displayControls.cpp
+++ b/src/gui/src/displayControls.cpp
@@ -1033,50 +1033,43 @@ void DisplayControls::displayItemSelected(const QItemSelection& selection)
   }
 }
 
-std::tuple<QColor*, Qt::BrushStyle*, bool> DisplayControls::lookupColor(
+std::pair<QColor*, Qt::BrushStyle*> DisplayControls::lookupColor(
     QStandardItem* item,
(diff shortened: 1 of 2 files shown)


maliberty:
It looks fine to me.  The only downside is that when you exceed the builtin list the via and routing layers' colors will be uncorrelated random values.

maliberty:
Compile errors to fix

gadfort:
> It looks fine to me. The only downside is that when you exceed the builtin list the via and routing layers' colors will be uncorrelated random values.

It turns out that was true before too, after the end of the list, the via colors were auto generated.

maliberty:
Approved and should merge once you cleanup the build issue

Issue #7556, opened 2025-06-11
GRT uses layers below the bottom routing layer in repair antennas


osamahammad21:
### Describe the bug

I have the following case where GRT uses a long M1 guide for a net. This happens in the first iteration of  repair antenna flow:

![Image](https://github.com/user-attachments/assets/c762aa0d-e362-47e0-93d0-b61484da710d)

This is causing failures in a [ci](https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-Private/detail/TR-via-access-layer/2/pipeline/530) run for my branch which introduces errors when such a guide is present. The error is:
[ERROR DRT-0155] Guide in net _01412_ uses layer li1 (2) that is outside the allowed routing range [met1 (4), met5 (12)] with via access on [li1 (2)].

### Expected Behavior

GRT to follow the signal layer constraints post drt.

### Environment

```shell
[WARNING] Your current OpenROAD version is outdated.
It is recommened to pull the latest changes.
If problem persists, file a github issue with the re-producible test case.
kernel: Linux 6.8.0-60-generic
os: Ubuntu 22.04.5 LTS (Jammy Jellyfish)
cmake version 3.24.2
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-22112-g5ffc612b38
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
[...]

PR #7685, opened 2025-06-30
grt: check layer range when adding resources from deleted net wires

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/7556.

Secure-ci started, as it will affect resources values and can lead to congestion in some designs.


eder-matheus:
@maliberty The bazel build is failing, but I can't see the reason as my changes don't touch anything related to it, and ORFS is also green.

maliberty:
Bazel seems unrelated

PR #7691, opened 2025-07-01
ci: split bazel build and test

vvbandeira:


vvbandeira:
> splitting build and test line this is not bazel idiomatic: only bazel test ... should be invoked
> 
> this slows down things _significantly_ for two reasons
> 
> 1. bazel test builds only what is needed for testing, importantly cfg=exec for openroad/sta tests. bazel build will build openroad cfg=target, so this change builds openroad/sta twice
> 2. tests are not run  in parallel with other build activities, which means verilator has to be built before src/... tests start to run
> 
> can you explain what the motivation is for splitting?

The goal is to separate the "build" and "test" steps so we can clearly identify which step is failing. We've observed some build errors in the CI that do not prevent the unit tests from running -- likely due to the use of the `--keep_going` flag. Not only did the tests run, but they all passed. When this happens, it can be confusing to understand what is failing and why if both build and test happen in the same CI stage.

FYI @sombraSoft (he is going to pick up this if we move forward with the split)

oharboe:
@vvbandeira I see. There should be some failing test if something fails to build, or we are missing a test. Can you give me a pointer, I'll have a look at the log?

I don't think we should merge this change, we should fix the underlying problem that makes this desirable.

A big part of the value of Bazel is that the user should be able to reproduce any problem without installing anything and by a) checking out a commit b) running `bazelisk test <failing target>`

Also, I'm thinking we should remove "--keep_going", because we want early failure in CI(fast turnaround times). The user should re-run locally the failing tests to get details.

@maliberty Regarding policy: I think we should try to create the simplest, fastest and most robust CI setup for bazel possible that needs no maintenance or further development. No fancy artifacts, logs, web GUI representations.  `bazelisk test ...` finds and runs all tests for all time, the rest is done locally.

vvbandeira:
@oharboe here's an example:
https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-7663-head/8/pipeline-overview/log?nodeId=266
Removing the `--keep_going` looks like a good idea if you are also ok.

oharboe:
@vvbandeira @maliberty `--keep_going` can be useful locally, but I think it solves one problem (get a list of all failing tests, not just whichever is first) and creates two others: the logs get harder to read, containing multiple errors and it increases turnaround times in failing PRs.

oharboe:
> @oharboe here's an example: https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-7663-head/8/pipeline-overview/log?nodeId=266 Removing the `--keep_going` looks like a good idea if you are also ok.



@maliberty Reached out to me about this error. I'm hoping it is gone in master after we upgraded .bazelversion to 8.3.1

@vvbandeira Do you have an example of this failing after .bazelversion 8.3.1 upgrade?

```
2025-06-30T17:42:44.292Z] Repository rule coursier_fetch defined at:
[2025-06-30T17:42:44.292Z]   /root/.cache/bazel/_bazel_root/4ec1920a10a623c8097845d1a0ddf570/external/rules_jvm_external+/private/rules/coursier.bzl:1434:33: in <toplevel>
[2025-06-30T17:42:44.292Z] (17:42:44) ERROR: /root/.cache/bazel/_bazel_root/4ec1920a10a623c8097845d1a0ddf570/external/rules_jvm_external+/private/rules/coursier.bzl:928:13: An error occurred during the fetch of repository 'rules_jvm_external++maven+openroad_maven':
[2025-06-30T17:42:44.292Z]    Traceback (most recent call last):
[2025-06-30T17:42:44.292Z] 	File "/root/.cache/bazel/_bazel_root/4ec1920a10a623c8097845d1a0ddf570/external/rules_jvm_external+/private/rules/coursier.bzl", line 1039, column 38, in _coursier_fetch_impl
[2025-06-30T17:42:44.292Z] 		dep_tree = make_coursier_dep_tree(
[2025-06-30T17:42:44.292Z] 	File "/root/.cache/bazel/_bazel_root/4ec1920a10a623c8097845d1a0ddf570/external/rules_jvm_external+/private/rules/coursier.bzl", line 928, column 13, in make_coursier_dep_tree
[2025-06-30T17:42:44.292Z] 		fail("Error while fetching artifact with coursier: " + exec_result.stderr)
[2025-06-30T17:42:44.293Z] Error in fail: Error while fetching artifact with coursier: Resolution error: Error downloading org.chipsalliance:chisel_2.13:7.0.0-RC1+22-2cb90ea2-
```

oharboe:
@vvbandeira Interesting: the reason you saw build errors and no failing tests is because the build errors caused test enumeration to fail. This is another example of `--keep_going` introducing problems: if the build stopped at the first error, then the confusion of all tests passing would not be introduced.

This goes to my thinking to keep the CI as simple as possible and re-run tests locally for more detail. 

The more complicated the CI system is, the more we'll be debugging complexities of the CI system, which is hard because we have to debug a setup on a server, which introduces more complexities because we need debug features and the cycle continues.

PR #7692, opened 2025-07-01
gpl: automatic adjust density instead of throwing error

gudeh:
Using uniform density if provided target density value is below the feasible value.

I tested with gcd and aes and works as intended. I triggered a secure-CI in:
[secure-gpl-automatic-adjust-density](https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/activity?branch=secure-gpl-automatic-adjust-density)

Current PR code version changes the default behavior, messaging a warning to the user that the density was changed.

This feature was requested by @jeffng-or, and was also mentioned in discussion https://github.com/The-OpenROAD-Project/OpenROAD/discussions/4257.

### DIFF
--- a/src/gpl/test/error01.ok
+++ b/src/gpl/test/error01.ok
@@ -20,7 +20,50 @@
 [INFO GPL-0019] Utilization:                    59.732 %
 [INFO GPL-0020] Standard cells area:           569.772 um^2
 [INFO GPL-0021] Large instances area:            0.000 um^2
-[ERROR GPL-0302] Consider increasing the target density or re-floorplanning with a larger core area.
-Given target density: 0.00
-Suggested target density: 0.60 (uniform density)
-GPL-0302
+[WARNING GPL-0302] Target density 0.0010 is too high for the available whitespace.
+Automatically adjusting to uniform density 0.6000.
+[INFO GPL-0023] Placement target density:       0.6000
+[INFO GPL-0024] Movable insts average area:      1.938 um^2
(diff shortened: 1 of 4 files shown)


gudeh:
There was a modification in a test which used to fail with incorrect density. I just removed the part to try and catch the error, maybe it requires the final defok file for comparison now?

maliberty:
ready for review?

gudeh:
> ready for review?

yes, the change is quite simple. I realized we do not use the target density before nesterovBase constructor, so we should be free to change it before its actual usage afterwards. Not requiring a restart as we first imagined.

src/gpl/test/error01.ok
```diff
@@ -20,7 +20,50 @@
 [INFO GPL-0019] Utilization:                    59.732 %
 [INFO GPL-0020] Standard cells area:           569.772 um^2
 [INFO GPL-0021] Large instances area:            0.000 um^2
-[ERROR GPL-0302] Consider increasing the target density or re-floorplanning with a larger core area.
-Given target density: 0.00
-Suggested target density: 0.60 (uniform density)
-GPL-0302
+[WARNING GPL-0302] Target density 0.0010 is too high for the available whitespace.
+Automatically adjusting to uniform density 0.6000.
```
rovinski, line 24: This is confusing to me? 0.0010 is too high so it is _increased_ to 0.6000?

maliberty: Good point - it should be too low

PR #7697, opened 2025-07-02
gpl: fix debug print segmentation fault and levels for callbacks

gudeh:
Avoid a segmentation fault by not attempting to access the name of a destructed db inst.

### DIFF
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -3017,13 +3017,12 @@ void NesterovBase::updateGCellState(float wlCoeffX, float wlCoeffY)
       // analogous to updatePrevGradient()
       updateSinglePrevGradient(gcells_index, wlCoeffX, wlCoeffY);
     } else {
-      debugPrint(log_,
-                 GPL,
-                 "callbacks",
-                 1,
-                 "warning: updateGCellState, db_inst not found in "
-                 "db_inst_index_map_ for instance: {}",
-                 db_inst->getName());
+      debugPrint(
+          log_,
+          GPL,
+          "callbacks",
+          1,
+          "warning: updateGCellState, db_inst not found in db_inst_index_map_");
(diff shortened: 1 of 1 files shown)


src/gpl/src/nesterovBase.cpp
```diff
@@ -3017,13 +3017,12 @@ void NesterovBase::updateGCellState(float wlCoeffX, float wlCoeffY)
       // analogous to updatePrevGradient()
       updateSinglePrevGradient(gcells_index, wlCoeffX, wlCoeffY);
     } else {
-      debugPrint(log_,
-                 GPL,
-                 "callbacks",
-                 1,
-                 "warning: updateGCellState, db_inst not found in "
-                 "db_inst_index_map_ for instance: {}",
-                 db_inst->getName());
```
maliberty, line 3026: If db_inst has been destroyed then why is it still in new_instances ?

gudeh: I understand this happens when rsz creates and destroys an instance on the same iteration. 

[The only place we use new_instances is protected by this condition](https://github.com/The-OpenROAD-Project/OpenROAD/blob/2c6e10d3fdc0a93eec2cf761b2f6359ae3992f66/src/gpl/src/nesterovBase.cpp#L2951-L2955), so it should not be a problem to have the destructed db inst in new_instances. If we want to be safe I can implement its removal.

maliberty: I think it is better to remove it.  Any new code would need similar protection and we could then safely print the instance name.

gudeh:
I realized this section of the code was actually missing the handling of bTerms after the callbacks are processed. There should be a new PR with such changes including the modifications from this PR.

maliberty:
close?

gudeh:
left it as a draft for now so I do not forget your request here.

PR #7709, opened 2025-07-03
mpl: reintroduce ability to handle fixed pins

### LINKED ISSUE #7675: MPL does not consider locations of fixed pins when placing macros
### Description

MPL no longer considers the locations of FIRM/FIXED pins when placing macros. This leads to poor placement of macros when the pins have been manually placed.

@maliberty FYI, here is a testcase for the MPL issue I mentioned.

### Suggested Solution

Consider the location of FIXED and FIRM pins since those will not be moved by PPL.

### Additional Context

Attached is a copy of the io_constraints9.tcl testcase from MPL (renamed to io_constraints11) which fixes the placement of the two pins in the testcase instead of using the pin constraints. In this case I would expect a solution similar to io_constraints9, but instead it places the macro against the left edge of the die instead the of the right edge.

[io_constraints11.zip](https://github.com/user-attachments/files/20961584/io_constraints11.zip)

AcKoucher:
Resolve #7675

## Overview
- This PR adds support for placed IOs in MPL. The idea is that, to avoid dealing with a huge number of connections inside annealing, we divide each edge of the die area into regions and bundle the IOs of that region into an IO bundle. Objects connected with those pins will be pushed towards the center of the bundle.
- These changes allow the user to run automated placement with both unplaced (either constrained or unconstrained) and fixed pins.

### IO Bundle Clusters
We divide the die area into a N number of bundles - based on a const value that we can easily change - and create a Cluster object for each IO Bundle. We then associate each fixed pin to its respective IO bundle based on its location. Empty IO bundles are deleted.

### Pin Access
We create pin access blockages to prevent objects from occupying the space close the IO bundle. We take into account the span of all the existent IO bundles, the std cell area in the design and the IO density in each bundle.

_Note: I adapted the pin access blockages generation for constraint regions to consider only the unplaced ios when computing the IO density per constraint region. This is more concise, because the base depth is computed based on the span of the constraint regions so we shouldn't be including unconstrained and fixed IOs._

### Orientation Improvement
When computing HPWL for a certain orientation and iterating bterms of a net, if the bterm is fixed, we use its actual location.


AcKoucher:
@gadfort Would you have some specific test case apart from the one in the issue that you'd like me to test these changes on?

AcKoucher:
Running Secure-CI.

gadfort:
@AcKoucher not one I can share, but I should be able to run this PR on it and let you know if it works. The plan is to generate the testcase (along with others) for PII.

maliberty:
The code looks ok but I'll wait for CI and Peter to merge.

PR #7724, opened 2025-07-05
bazel: test -c opt, because that is what is being deployed

oharboe:
Testing non -c opt builds is a bonus, but minimally we should test what is deployed.


oharboe:
@maliberty Actually, openroad and opensta is cfg=exec, so -c opt/dbg has no effect on bazelisk test.

So I will let this lie until I learn more bazel.

PR #7728, opened 2025-07-06
Fix a couple of dependency and linking issues

hzeller:


maliberty:
Once #7730 merges, please merge master to this PR so we can see bazel run

src/mpl/BUILD
```diff
@@ -43,6 +43,7 @@ cc_library(
         "//src/sta:opensta_lib",
         "//src/utl",
         "@boost.random",
+        "@cudd",
```
maliberty: Why is this necessary?  mpl has no direct use of cudd

hzeller: ah, it was not. It was due to an include of `"util.h"` which is a very generically named header without the use of project-relative paths. And it is also provided by cudd. Anyway, my scripts picked up on that.

We really should establish project-relative paths soon.

Fixed.

maliberty: I don't see it as fixed.  I see the same in rcx as well.

hzeller: done.

maliberty (approved):

oharboe:
@hzeller I think a merge with master will fix the bazelisk snag.

maliberty:
Bazelisk is fixed but it now fails on build errors, eg `: error: undefined symbol: grt::GlobalRouter::getPinGridPositions(odb::dbNet*)`

src/drt/BUILD
```diff
@@ -211,6 +211,7 @@ cc_library(
         "//src/dst",
         "//src/odb",
         "//src/stt",
+        "//src/stt:ui",
```
maliberty: why is this needed?

src/cts/test/BUILD
```diff
@@ -166,8 +166,8 @@ cc_test(
     linkstatic = True,  # TODO: remove once deps define all symbols
     deps = [
         "//src/cts",
-        "//src/grt",
         "//src/utl",
+        "//src/grt",  # work around cyclic dependncy between rsz and grt
```
maliberty, line 170: FYI @eder-matheus  is working fixing this cycle.

src/drt/BUILD
hzeller: There is an ambiguity as `SteinerTreeBuilder.h` is provided by both, `stt` and `stt:ui`. So not really needed here.
Removed.

I'll file an issue with ambiguous mult-mentions of header files.

PR #7592, opened 2025-06-16
jenkins: Bazel early failure, run in parallel with docker builds

oharboe:

Issue #7568, opened 2025-06-13
Speed up Bazel CI so that common errors are spotted in minutes and seconds, not hours


oharboe:
### Description

**Call to arms:** minutes and seconds, not hours to get CI errors. For much larger projects than OpenROAD we're seing CI pass/fail for simple changes in 60 seconds to minutes.

- [ ] Do a [shallow clone](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7763) for OpenROAD and submodules, brings checkout down to 1 minute instead of 1m40s
- [ ] The docker image for Bazel should hardly ever change. Built it and tag it and us it. If it needs updating, do that manually out of bands. This avoids running `docker build -f docker/Dockerfile.bazel -t openroad/bazel-ci` for every bazel build, cuts ca. 3 minutes before tests start.
- [ ] Download .bazelversion in docker image. `COPY .bazelversion .` and `RUN bazelisk info` should do it.
- [ ] Do a survey of CI errors for OpenROAD and file feature request to extend bazel testing to include coverage of common problems so they can be spotted in minutes

Simplify docker image for BAzel:

- use podman, no need for docker. bazel-orfs looks for podman or docker and will use either.
- clang is not needed, bazelisk downloads this
- python is not needed, bazelisk will download the right version and pip install modules within Bazel
- time is not needed, bazelisk will get this via bazel-orfs/ORFS.

When all that is in place, use the Docker idiom to reduce layer size:

```
RUN apt-get update && \
    apt-get install -y --no-install-recommends \
        curl \
        ca-certificates && \
    rm -rf /var/lib/apt/lists/*
```

**First cut done, further refinement and bugfixes in followup issues:**

- [x] Start running bazel build immediately for OpenROAD PRs, don't wait for installer/docker images testing(screenshot below). I had a quick stab at it https://github.com/The-OpenROAD-Project/OpenROAD/pull/7592
- [x] Add caching of bazel
- [x] Use a [minimal docker image](https://github.com/The-OpenROAD-Project/OpenROAD/issues/7665#issuecomment-3012188491) for bazel so bazel startup is faster.

![Image](https://github.com/user-attachments/assets/b6ca1ba7-d57a-4441-b828-aba464de08f5)

### Suggested Solution

Reorder activities in Jenkins so that common errors are spotted in minutes, not hours.

oharboe:
@vvbandeira @maliberty One down? (caching)

vvbandeira:
[...]

PR #7732, opened 2025-07-06
ci: run Bazel earlier

vvbandeira:
Approx. 10 min TAT for Bazel: ~2-3 minutes for setup and ~7-8 minutes to run tests w/cache enabled.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -260,12 +260,14 @@ def bazelTest = {
                 stage('Bazel Build') {
                     withCredentials([file(credentialsId: 'bazel-cache-sa', variable: 'GCS_SA_KEY')]) {
                         timeout(time: 120, unit: 'MINUTES') {
-                            def cmd = 'bazelisk test --config=ci --show_timestamps --test_output=errors --curses=no --force_pic'
-                                if (env.BRANCH_NAME != 'master') {
-                                    cmd += ' --remote_upload_local_results=false'
-                                }
-                            sh label: 'Bazel Build', script: cmd + ' --google_credentials=$GCS_SA_KEY ...'
+                            def cmd = 'bazelisk test --config=ci --show_timestamps --test_output=errors --curses=no --force_pic';
+                            if (env.BRANCH_NAME != 'master') {
+                                cmd += ' --remote_upload_local_results=false';
+                            }
+                            sh label: 'Bazel Build', script: cmd + ' --google_credentials=$GCS_SA_KEY ...';
                         }
+                        sh label: 'Archive logs', script:  'tar zcvf bazel-logs.tgz bazel-testlogs/.';
+                        archiveArtifacts artifacts: 'bazel-logs.tgz';
                     }
                 }
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -162,31 +162,6 @@ def getParallelTests(String image) {
             }
         },
 
-        'Build with Bazel': {
-            node {
-                withDockerContainer(args: '-u root -v /var/run/docker.sock:/var/run/docker.sock', image: image) {
-                    stage('Setup Bazel Build') {
-                        echo "Build with Bazel";
-                        sh label: 'Configure git', script: "git config --system --add safe.directory '*'";
-                        checkout scm;
-                    }
-                    stage('Bazel Build') {
-                        withCredentials([file(credentialsId: 'bazel-cache-sa', variable: 'GCS_SA_KEY')]) {
-                            timeout(time: 120, unit: 'MINUTES') {
-                                def cmd = 'bazelisk test --config=ci --show_timestamps --test_output=errors --curses=no --force_pic'
-
-                                if (env.BRANCH_NAME != 'master') {
-                                    cmd += ' --remote_upload_local_results=false'
-                                }
-
-                                sh label: 'Bazel Build', script: cmd + ' --google_credentials=$GCS_SA_KEY ...'
-                            }
-                        }
-                    }
-                }
-            }
-        },
-
         'Check message IDs': {
             dir('src') {
                 sh label: 'Find duplicated message IDs', script: '../etc/find_messages.py > messages.txt';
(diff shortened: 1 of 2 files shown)


vvbandeira:
@oharboe
Not ready yet, but Bazel would run in parallel with Docker with this PR. Was this what you had in mind with #7592?

Jenkinsfile
```diff
@@ -274,21 +249,33 @@ def getParallelTests(String image) {
...
+                    withCredentials([file(credentialsId: 'bazel-cache-sa', variable: 'GCS_SA_KEY')]) {
+                        timeout(time: 120, unit: 'MINUTES') {
+                            def cmd = 'bazelisk test --config=ci --show_timestamps --test_output=errors --curses=no --force_pic';
+                            if (env.BRANCH_NAME != 'master') {
+                                cmd += ' --remote_upload_local_results=false';
+                            }
+                            sh label: 'Bazel Build', script: cmd + ' --google_credentials=$GCS_SA_KEY ...';
+                        }
+                        sh label: 'Archive logs', script:  'tar zcvf bazel-logs.tgz bazel-testlogs/.';
+                        archiveArtifacts artifacts: 'bazel-logs.tgz';
```
vvbandeira: @oharboe 
Does this address https://github.com/The-OpenROAD-Project/OpenROAD/pull/7722#issuecomment-3042332591, or do we need to retain more files/logs from elsewhere?

oharboe: No. What I need are the ORFS logs, not bazel logs. Come to think of it, those logs are in the bazel cache, so I should be able to just build with caches and get the logs that way. Checking...

Just remove the above line, something else will be needed. Let's do that in a separate PR, if needed.

Jenkinsfile
```diff
@@ -274,21 +249,33 @@ def getParallelTests(String image) {
...
-    properties([
-        buildDiscarder(logRotator(
-            daysToKeepStr:         daysToKeep,
-            artifactDaysToKeepStr: daysToKeep,
+def bazelTest = {
+    stage ('Build with Bazel') {
+        node {
+            stage('Setup Bazel Build') {
+                checkout scm;
+                sh label: 'Setup Docker Image', script: 'docker build -f docker/Dockerfile.bazel -t openroad/bazel-ci .';
```
oharboe, line 257: Don't build docker image here, use prebuilt image. This will cut down the time it takes to start a docker build with several minutes. Added as a task to #7568

docker/Dockerfile.bazel
```diff
@@ -0,0 +1,31 @@
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get -y update \
+    && apt-get -y install curl git
+
+RUN curl -Lo bazelisk https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-amd64 \
+    && chmod +x bazelisk \
+    && mv bazelisk /usr/local/bin/bazelisk
```
oharboe, line 10: Refinement for later: download .bazelversion here. #7568

docker/Dockerfile.bazel
```diff
@@ -0,0 +1,31 @@
...
+    && chmod a+r /etc/apt/keyrings/docker.asc \
+    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/ubuntu \
+        $(. /etc/os-release && echo "${VERSION_CODENAME}") stable" | \
+        tee /etc/apt/sources.list.d/docker.list > /dev/null
+
+RUN apt-get -y update \
+    && apt-get -y install --no-install-recommends \
+      build-essential \
+      clang \
+      containerd.io \
```
oharboe, line 24: We should be building within seconds and minutes for bazel, so small things will matter...

Refinement for later: simplify docker image. #7568

Jenkinsfile
vvbandeira: In our CI, this takes less than 1 min. The last three builds were 57s, 47s, and 44s. We can use prebuilt, but I don't think it is a high priority. The breakdown of the runtime is ~1 min to provision the VM, ~2 min to checkout the code (this can be improved with shallow clone, I will give it a try as is low effort), ~1 min to build the image and 7+min to run the `bazelisk test [...]`

docker/Dockerfile.bazel
vvbandeira: How often do you see this being updated? If it is somewhat often, I would say to wait until we have the prebuilt image set up to automatically update; otherwise, it will require manual steps to update, which is not ideal.

docker/Dockerfile.bazel
vvbandeira: All the current packages are required to run the tests. I iterated over adding one package at a time and checking the error from `bazelisk test [...]`. Unless we can move the dependencies into Bazel itself, these are the minimal requirements as far as I can tell.

Jenkinsfile
oharboe: Came up with a better solution https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3294

vvbandeira: Where would the files be located in the Bazel directory? We can save them and make it easier to find in the CI rather than re-running from cache.

Jenkinsfile
oharboe: I guess we'll whittle away at the things that move the needle and then eventually this sort of thing would come to the top of the list... That these quicker things start to matter in terms of time to failed CI and successful Bazel build is a good problem to have.

docker/Dockerfile.bazel
oharboe: Low priority, listed more as an observation here.

Jenkinsfile
oharboe: re-running from cache is the idiomatic way to download built files in bazel, we should not implement an alternative to this, so just remove this line for now.

Let us do https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3294 and then revisit alternatives if that isn't good enough.

PR #7748, opened 2025-07-08
Don't stop at first compile issue: keep going

hzeller:


oharboe (changes requested):
I don't think this is a good idea, see commit 043cc8bafbeacc24bf78313e194c1486d6af8794  when we removed --keep_going recently.

- it makes error messages more complicated
- it takes longer for a CI error to be returned
- More detailed analysis(archiving, etc.) should happen locally instead of making CI more complicated.

etc.

gadfort:
@oharboe I find that continuing can be helpful so we can see all the errors that occurred, stopping at the first one is only useful as a signal that there is an issue, but doesn't help determine to extent of the issue. Personally, I'm in favor of this change.

oharboe:
> @oharboe I find that continuing can be helpful so we can see all the errors that occurred, stopping at the first one is only useful as a signal that there is an issue, but doesn't help determine to extent of the issue. Personally, I'm in favor of this change.

I was too, I added it initially, but then when I worked with @vvbandeira and @maliberty I was convinced otherwise, because the error messages are *terrible* with `--keep_going` and the turnaround times are the full build time of OpenROAD(not hours as for ORFS) instead of minutes on a failure.

Adding `--keep_going` fixes one simple problem and introduces more complicated ones.

It is, however, a very useful local tool.

hzeller:
the CI should keep going, so that it is possible to fix things. Otherwise you have a turnaround time of 45 minutes for each fix.

I don't know what terrible error messages you got (except the regular terribleness of c++ error messages), but it is _way_ more worthwhile to `--keep_going` than spending hours and days roundtrips to satisfy a CI.

oharboe:
> the CI should keep going, so that it is possible to fix things. Otherwise you have a turnaround time of 45 minutes for each fix.
> 
> I don't know what terrible error messages you got (except the regular terribleness of c++ error messages), but it is _way_ more worthwhile to `--keep_going` than spending hours and days roundtrips to satisfy a CI.

I think both approaches have advantages and disadvantages.

hzeller:
I don't see a disadvantage to always run `--keep_going` , only disadvantages not doing that (because it takes forever to narrow down on an issue that is only visible in the CI).

oharboe:
@vvbandeira Could we split this in two in jenkins, first build without `--keep_going`, ignore errors, then build with `--keep_going` where errors stop the build. The first failed build is immediately available in Jeknins with maximum readbility, the second is available later with all failures listed.

Then both use cases are covered.

hzeller:
oh, is it not possible to see the ongoing build in Jenkins ?

oharboe:
> oh, is it not possible to see the ongoing build in Jenkins ?

It is, but the problem is that the first error drowns in an ongoing build. By splitting in two, we add a little bit of complexity to Jenkins, but I think this covers both use-cases reasonably well.

vvbandeira:
> I don't know what terrible error messages you got (except the regular terribleness of c++ error messages), but it is _way_ more worthwhile to `--keep_going` than spending hours and days roundtrips to satisfy a CI.

The confusing message we got was due to a test failing to build, maybe it was also not set to run, because the last message in the logs could be read as "all tests passed" but "build has failed" type of situation. We had to dig at the very top of the logs to understand the reason for the failure, which was confusing to other devs as well.

If we could separate the build of the tests and running them it would be great.

maliberty:
Doubling the number of tests (with and without keep-going) just raises the likelihood of other errors creeping in (e.g. networking) and doesn't reduce the test time for the PR.  I am not in favor of that suggestion.

maliberty:
> If we could separate the build of the tests and running them it would be great.

We could just do `bazelisk build`  and then `bazelisk test`

maliberty:
I generally favor keep_going as well as it is more useful to developers to see the full extent of a problem.

oharboe:
> > If we could separate the build of the tests and running them it would be great.
> 
> We could just do `bazelisk build`  and then `bazelisk test`

No, tests use cfg=exec so that builds openroad twice and also does not allow for overlap in test/build graph

oharboe:
> I generally favor keep_going as well as it is more useful to developers to see the full extent of a problem.

It is false choice, we can have both https://github.com/The-OpenROAD-Project/OpenROAD/pull/7751

gadfort:
> > > If we could separate the build of the tests and running them it would be great.
> > 
> > 
> > We could just do `bazelisk build`  and then `bazelisk test`
> 
> No, tests use cfg=exec so that builds openroad twice and also does not allow for overlap in test/build graph

I agree with @maliberty a build followed by a test seems completely reasonable, if we cannot test what we built directly that seems odd to me.

oharboe:
@gadfort I am still learning cfg=exec/target bazel idoms, but the long and the short of it is that  exec is host and we are running tests on host. bazel build builds cfg=target, so master is doing the correct idiomatic thing today.

oharboe:
Reminder to self, add example of how to visualize dependencies to this PR

https://github.com/The-OpenROAD-Project/OpenROAD/discussions/7295

oharboe:
@hzeller Close?

This is what you have today on master `--keep_going` at failing log when you click on `bazelisk test ...`:

image

hzeller:
Yeah, if we have a keep_going now, then this can be closed.

PR #7751, opened 2025-07-08
ci: add try/catch to consider keep_going

vvbandeira:
Possible replacement for #7748
FYI @oharboe @hzeller @gadfort

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -264,7 +264,13 @@ def bazelTest = {
                             if (env.BRANCH_NAME != 'master') {
                                 cmd += ' --remote_upload_local_results=false';
                             }
-                            sh label: 'Bazel Build', script: cmd + ' --google_credentials=$GCS_SA_KEY ...';
+                            cmd += ' --google_credentials=$GCS_SA_KEY';
+                            try {
+                                sh label: 'Bazel Build', script: cmd + ' ...';
+                            } catch (e) {
+                                currentBuild.result = 'FAILURE';
+                                sh label: 'Bazel Build (keep_going)', script: cmd + ' --keep_going ...';
+                            }
                         }
                     }
                 }


Jenkinsfile
```diff
@@ -264,7 +264,13 @@ def bazelTest = {
                             if (env.BRANCH_NAME != 'master') {
                                 cmd += ' --remote_upload_local_results=false';
                             }
-                            sh label: 'Bazel Build', script: cmd + ' --google_credentials=$GCS_SA_KEY ...';
+                            cmd += ' --google_credentials=$GCS_SA_KEY';
+                            try {
+                                sh label: 'Bazel Build', script: cmd + ' ...';
+                            } catch (e) {
+                                currentBuild.result = 'FAILURE';
```
oharboe, line 271: Nice!

Just some thoughts, I'm not suggesting making any changes:

I think `currentBuild.result = 'FAILURE';` is an interesting touch. Consider network outage. We really only want to try again if there were build and not transient errors as continuing makes transient errors less obvious. 

Does the exception contain any useful information? The catch is very broad, can or should the catch be narrowed to catch more specific errors only?

oharboe (approved):

Jenkinsfile
vvbandeira: I will look further, if possible, I will narrow down the handling on another PR.

oharboe: Yes: it is a refinement in future PR.

Probably best to let it lie until there's something that happens(like network errors) that gives us data on how there could be an improvement.

The alternative is to write more complicated code now without really knowing if and how it will be used and how often. Keep it simple and solid for now.

oharboe:
As I understand it there will be:

- "Bazel Build"
- "Bazel Build (keep going)"

Instead of just 1 as today. What I have observed is that it is master that decides what Jenkins should do, not the PR, so we'll see the effects of this change *after* it is merged with master.

image

QuantamHD:
Just as an FYI most people on bazel use a remote UI to collect the build and test stats. 

https://www.buildbuddy.io/ is a common one.

oharboe:
@vvbandeira Some outage?

```
15:20:53  (15:20:51) WARNING: Download from https://ftp.gnu.org/gnu/ncurses/ncurses-6.2.tar.gz failed: class java.io.IOException Connect timed out
```

vvbandeira:
> @vvbandeira Some outage?
> 
> ```
> 15:20:53  (15:20:51) WARNING: Download from https://ftp.gnu.org/gnu/ncurses/ncurses-6.2.tar.gz failed: class java.io.IOException Connect timed out
> ```

Yes, I noticed that earlier while trying to download the Bison source code for another task. Looks like ftp.gnu had an outage.

vvbandeira:
> As I understand it there will be:
> 
> * "Bazel Build"
> * "Bazel Build (keep going)"
> 
> Instead of just 1 as today. What I have observed is that it is master that decides what Jenkins should do, not the PR, so we'll see the effects of this change _after_ it is merged with master.

Point of clarification: on the left you have stages, i.e., `Bazel Build`, on the right you have steps, i.e., `Build Bazel` and `Build Bazel (keep_going)`. The try/catch will add an additional step, not an additional stage.

image

Changes to Jenkinsfile are run on the PR if the GH user has enough privilege to do so, which avoids bad actors injecting code on our CI.

oharboe:
@vvbandeira Looked at this after merge, noticed that the math is off:

2+7 minutes is 9, not 14.

Weird. I'm not suggesting anything be done, just surprising.

image

maliberty:
Perhaps it include waiting time for an machine

oharboe:
> Perhaps it include waiting time for an machine

Neat. That would be useful information to have, if it is the case.

vvbandeira:
@oharboe 
This might be useful to you: https://jenkins.openroad.tools/plugin/pipeline-timeline/index.html?build_url=https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-7751-head/2/

PR #7763, opened 2025-07-09
jenkins: shallow checking for Bazel

oharboe:
Testing speed


oharboe:
@vvbandeira Looks like 1 minute vs 1m40 or thereabouts. If we can avoid building the docker image every time, we can start bazel build in 1 minute. With caches we can have errors to the user in 2 minutes or so, happy case. :-)

vvbandeira:
> @vvbandeira Looks like 1 minute vs 1m40 or thereabouts. If we can avoid building the docker image every time, we can start bazel build in 1 minute. With caches we can have errors to the user in 2 minutes or so, happy case. :-)

Sounds great! I will try to schedule some time later in the week.

oharboe:
@vvbandeira I'll leave you to consider this and make the switch

PR #7772, opened 2025-07-10
gpl: extend max iter cap by routability internal iterations

gudeh:
Increase the maximum number of nesterov iterations by the number of iterations performed during routability mode (i.e., the section between 0.6 and 0.3 overflow, where snapshot saving and revert triggering occur). This acts as a global stopping criteria for Nesterov.

This change addresses the behavior observed in [issue #7581](https://github.com/The-OpenROAD-Project/OpenROAD/issues/7581), where a design exhibited very small but consistent reductions in congestion during routability. As a result, it performed an unusually high number of routability reverts (~30), eventually hitting the existing 5000-iteration limit.

This PR also introduces a new routability stopping criteria: a fixed cap of 50 routability reverts.

Log changes:

- Remove unnecessary + 1 when printing iterations.
- Report overflow and HPWL at the final iteration.
- Compute and display the final weighted routing congestion.
- Add log messages when each of the three main GPL modes begins: incremental, CG initial place, and nesterov.


gudeh:
CI failed only on bazel with: `There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.`

maliberty:
The real issue is mock-array
```
[ERROR] finish__timing__setup__ws fail test: -351.636 >= -351.15
```
The "too big" is more minor.

gudeh:
I ran the bazel test locally with my branch, and also ran the same test with master version, and I see no difference among logs for gpl:
image

gudeh:
> The real issue is mock-array
> 
> ```
> [ERROR] finish__timing__setup__ws fail test: -351.636 >= -351.15
> ```
> 
> The "too big" is more minor.

I realized why this is happening. I made a small logic change to detect whether routability mode is active: instead of relying solely on the overflow values, I added a check for whether the snapshot has already been saved. As a result, TD is now allowed to run even before the snapshot is saved and overflow is low. This change only affects designs that start with a low overflow (e.g., below 0.60).

gudeh:
secure-CI is green in https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3355

maliberty (approved):

PR #7773, opened 2025-07-10
Clock gating

kbieganski:
Adds `cgt`, a module that performs automatic clock gating based on [this paper](https://dl.acm.org/doi/10.1145/1391469.1391637).
It uses the logic extractor from `rmp`, so that part was extracted into `lext`, a common module for both `rmp` and `cgt`.
ABC is used for verifying gating conditions.

#### Before

```
Group                  Internal  Switching    Leakage      Total
                          Power      Power      Power      Power (Watts)
----------------------------------------------------------------
Sequential             2.00e-02   9.81e-04   1.18e-08   2.10e-02  32.9%
Combinational          1.19e-02   1.28e-02   2.99e-08   2.47e-02  38.7%
Clock                  1.18e-02   6.35e-03   1.83e-09   1.82e-02  28.4%
Macro                  0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
Pad                    0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
----------------------------------------------------------------
Total                  4.37e-02   2.01e-02   4.35e-08   6.38e-02 100.0%
                          68.5%      31.5%       0.0% 
```

#### After

```
Group                  Internal  Switching    Leakage      Total
                          Power      Power      Power      Power (Watts)
----------------------------------------------------------------
Sequential             5.46e-03   9.85e-04   1.18e-08   6.45e-03  13.7%
Combinational          1.19e-02   1.28e-02   3.00e-08   2.47e-02  52.5%
Clock                  1.17e-02   4.25e-03   1.87e-09   1.59e-02  33.8%
Macro                  0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
Pad                    0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
----------------------------------------------------------------
Total                  2.91e-02   1.80e-02   4.36e-08   4.71e-02 100.0%
                          61.7%      38.3%       0.0%
```

#### Before

```
Group                  Internal  Switching    Leakage      Total
                          Power      Power      Power      Power (Watts)
----------------------------------------------------------------
Sequential             2.14e+00   1.94e-01   2.65e-06   2.33e+00  15.7%
Combinational          7.34e+00   5.20e+00   3.75e-05   1.25e+01  84.3%
Clock                  0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
Macro                  0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
Pad                    0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
----------------------------------------------------------------
Total                  9.48e+00   5.39e+00   4.01e-05   1.49e+01 100.0%
                          63.7%      36.3%       0.0% 
```

#### After

```
Group                  Internal  Switching    Leakage      Total
                          Power      Power      Power      Power (Watts)
----------------------------------------------------------------
Sequential             9.28e-01   1.94e-01   2.65e-06   1.12e+00   8.2%
Combinational          7.33e+00   5.19e+00   3.76e-05   1.25e+01  91.7%
Clock                  5.44e-03   3.73e-03   9.18e-09   9.17e-03   0.1%
Macro                  0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
Pad                    0.00e+00   0.00e+00   0.00e+00   0.00e+00   0.0%
----------------------------------------------------------------
Total                  8.26e+00   5.39e+00   4.02e-05   1.37e+01 100.0%
                          60.5%      39.5%       0.0% 
```


maliberty:
I'll hold off on review until you resolve the various failures

maliberty:
Note that you will need to make this work with Bazel as well as cmake

kbieganski:
Sure, no problem. Shouldn't the CI run on our fork as well? ([only the security check does](https://github.com/antmicro/OpenROAD/commit/a6bbd53e0074535894ddd56749129f55aeb41190))
Also, what is that check about? The error is not clear to me - are digits forbidden in filenames?

maliberty:
I don't think we allow Jenkins to be used on forks as that would allow arbitrary workloads to be started (eg someone will inevitably replace OR with a bit coin miner).

kbieganski:
Right, but I mean the format checks etc, that's just GitHub Actions, right?

maliberty:
I expect those would run in a fork

src/utl/include/utl/Logger.h
```diff
@@ -49,6 +50,7 @@ class Progress;
   X(GRT)                \
   X(GUI)                \
   X(IFP)                \
+  X(LEXT)               \
```
maliberty: Module names should be three letters as the logger formatting assumes that.  I find the name logic extraction not very descriptive so let's discuss before you spend time to change it.

QuantamHD:
@vvbandeira @maliberty Can you take a look at the scan code ci step, looks like they need a whitelist?

src/cgt/include/cgt/RandomBits.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <cstdint>
+#include <random>
+
+namespace cgt {
+
+class RandomBits
+{
+ private:
+  std::minstd_rand generator{std::random_device{}()};
```
QuantamHD: All randomness must be deterministic in OpenROAD. For a given set of input files with the same 256 hash, it should always produce the same output files with the same sha256 hash.

src/cgt/src/.gdb_history
QuantamHD, line 1: Please delete this file.

maliberty:
I don't think we should worry about the whitelist until we settle the question of module names.

src/cgt/include/cgt/RandomBits.h
kbieganski: Yes, accounted for [here](https://github.com/The-OpenROAD-Project/OpenROAD/blob/f04a99febafefb4f7860df5e0c220417ca7340e8/src/cgt/src/ClockGating.cpp#L384).

kbieganski:
What name would you propose for the logic extractor? It extracts/exports logic to ABC. How about just `abc`, and make it contain all helpers for interacting with ABC?

maliberty:
I don't think of it as extracting logic but more of a local cutout or window.  Extraction is a very overloaded word as well and often refers to computing a derived result (e.g. rc extraction, connectivity extraction, device parameter extraction, etc).  I expect the idea of a small piece of logic to be useful in many context so abc would not be a good name.  I think rsz will want the same for example.

I like 'Logic Cutout' but am open to other ideas.

rovinski:
"logic partition"?

Also, I am curious about how this compares to the clock gating done at the synthesis level. I think Yosys supports this now, and it is generally more suitable to be done at the synthesis stage. While 15% power savings are nothing to scoff at, I usually see 30-60% savings from commercial tools.

maliberty:
Partition sounds a bit like min-cut paritioning of par/TritonPart.

QuantamHD:
> "logic partition"?
> 
> Also, I am curious about how this compares to the clock gating done at the synthesis level. I think Yosys supports this now, and it is generally more suitable to be done at the synthesis stage. While 15% power savings are nothing to scoff at, I usually see 30-60% savings from commercial tools.

This is a similar technique as what's used in other tools. The one in Yosys just goes off trivial enable signals which doesn't actually save much power since those signals have very low toggle rates in practice. 

I asked Yosys to implement the former, and it didn't really deliver the results I was expecting.

QuantamHD:
@maliberty I think your name suggestions need to be a 3 letter monicor not a general thing about Logic Extraction. The class is already called LogicCut.

I would propose syn, and it would encapsulate synthesis related functionality. Or throw it in utl, or get rid of the 3 letter name requirement.

QuantamHD:
@kbieganski The decision has been made by Matt and I. Please rename it `cut`

src/OpenRoad.cc
```diff
@@ -240,6 +242,7 @@ void OpenRoad::init(Tcl_Interp* tcl_interp,
   initOpenRCX(extractor_, db_, logger_, getVersion(), tcl_interp);
   initICeWall(icewall_, db_, logger_, tcl_interp);
   initRestructure(restructure_, logger_, sta_, db_, resizer_, tcl_interp);
+  cgt::initClockGating(this);
```
maliberty: Don't pass the entire object, just pass the actual dependencies (see all other modules).

src/cgt/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Simply run the following script:
+
+```shell
+./test/regression
+```
+
+## Limitations
+
+Clock gating is not supported for designs that contain cells not supported by ABC, such as adders.
```
maliberty: Meaning that adders won't be considered for gating or that the presence of an adder mean nothing will be gated at all?

src/cgt/README.md
```diff
@@ -0,0 +1,65 @@
+# Clock gating
+
+This module provides automatic insertion of clock gates for reducing power usage.
```
maliberty, line 3: A high level description of the method would be helpful.

src/cgt/include/cgt/ClockGating.h
```diff
@@ -0,0 +1,197 @@
...
+
+}  // namespace lext
+
+namespace cgt {
+
+using utl::Logger;
+
+enum class GroupInstances
+{
+  No,
```
maliberty: What is 'No'?  Is that no grouping?  If so I would use None.  

Using https://google.github.io/styleguide/cppguide.html#Enumerator_Names (leading k).

src/cgt/include/cgt/ClockGating.h
```diff
@@ -0,0 +1,197 @@
...
+  // conditions into the network.
+  void insertClockGate(const std::vector<sta::Instance*>& instances,
+                       const std::vector<sta::Net*>& gate_cond_nets,
+                       bool clk_enable);
+
+  void dump(const char* name);
+  void dumpAbc(const char* name, abc::Abc_Ntk_t* network);
+
+  using path = std::filesystem::path;
+  path getNetworkGraphvizDumpPath(const char* name)
```
maliberty: This method and the ones that follow should be out-of-line.

src/lext/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+
+messages(
+TARGET lext
+)
+
+target_include_directories(lext
+  PUBLIC
+    ./include
+  PRIVATE
+    ./include
```
maliberty: No need for the duplicate PRIVATE section

src/lext/README.md
```diff
@@ -0,0 +1 @@
+# Logic extractor
```
maliberty: Better documentation

src/lext/include/lext/abc_library_factory.h
```diff
@@ -1,5 +1,8 @@
-// SPDX-License-Identifier: BSD-3-Clause
-// Copyright (c) 2024-2025, The OpenROAD Authors
+// Copyright 2024 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
```
maliberty: Please do not change the copyright header (here and elsewhere)

src/lext/include/lext/abc_library_factory.h
```diff
@@ -38,6 +48,7 @@ class AbcLibrary
   void InitializeConstGates();
 
   utl::UniquePtrWithDeleter<abc::SC_Lib> abc_library_;
+  abc::Mio_Library_t* mio_library_;  // TODO: handle lifetime of this lib
```
maliberty: Are you planning to fix this TODO?

src/lext/include/lext/blif.h
```diff
@@ -0,0 +1,98 @@
+/////////////////////////////////////////////////////////////////////////////
```
maliberty: Use the standard copyright notice

src/lext/include/lext/delay_optimization_strategy.h
maliberty, line 1: This feel like it belongs in rmp not lext (cut).  Why was it moved over?

src/lext/include/lext/logic_optimization_strategy.h
maliberty, line 1: in rmp?

src/lext/include/lext/resynthesis_strategy.h
maliberty, line 1: rmp?

src/lext/include/lext/unique_name.h
maliberty, line 1: to utl?  Nothing specific to lext or rmp

src/lext/include/lext/unique_name.h
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#include <string>
+
+namespace lext {
+class UniqueName
+{
+ public:
+  std::string GetUniqueName(const std::string& prefix = "lext_")
```
maliberty: It seems odd that prefix wouldn't be a ctor argument.  Would you really want to change prefixes per call?

src/lext/src/delay_optimization_strategy.cpp
```diff
@@ -85,6 +88,19 @@ utl::UniquePtrWithDeleter<abc::Abc_Ntk_t> DelayOptimizationStrategy::Optimize(
...
+	// TODO(jbylicki): Where did this come from?!
+
+  // auto library = static_cast<abc::Mio_Library_t*>(ntk->pManFunc);
+  // // Install library for NtkMap
+  // abc::Abc_FrameSetLibGen(library);
+  //
+  // // Set up libraries for buffering
+  // AbcLibraryFactory library_factory(logger);
+  // library_factory.AddDbSta(sta_);
+  // AbcLibrary abc_sc_library = library_factory.Build();
```
maliberty: rm unless you are fixing the TODO

src/rmp/README.md
maliberty, line 1: Why is there no readme anymore for this module?

src/cgt/include/cgt/RandomBits.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include <cstdint>
+#include <random>
+
+namespace cgt {
+
+class RandomBits
+{
+ private:
```
maliberty: The private section should come after the public https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/cgt/include/cgt/RandomBits.h
maliberty: Why not construct it with a fixed seed or just use the default_seed?

src/cgt/include/cgt/RandomBits.h
```diff
@@ -0,0 +1,40 @@
...
+  bool get()
+  {
+    if (bits_remaining == 0) {
+      bit_buffer = generator();
+      bits_remaining = sizeof(bit_buffer) * 8;
+    }
+
+    bool bit = bit_buffer & 1;
+    bit_buffer >>= 1;
+    bits_remaining--;
```
maliberty, line 23: The low order bits of an LCG are not that random.  It would be better to use mt19937

maliberty:
It feels like too much of rmp got pulled into lext.  I stopped commenting on it per file but I feel lext should not need an abc dependency and should be generic infrastructure.  Rmp is the right place for applying abc to a logic cut.

src/cgt/src/cgt.tcl
```diff
@@ -0,0 +1,45 @@
...
+                                    }
+
+proc clock_gating { args } {
+  sta::parse_key_args "clock_gating" args \
+    keys {-instances -gate_cond_nets -min_instances -max_cover -group_instances -dump_dir}
+
+  set instances {}
+  set gate_cond_nets {}
+  set min_instances 10
+  set max_cover 100
```
maliberty: It is better to set the defaults in c++ that way they can be common between tcl & python

src/cgt/src/cgt.i
```diff
@@ -0,0 +1,63 @@
...
+  int list_len;
+  Tcl_Obj **list_elems;
+  Tcl_Interp* interp = getOpenRoad()->tclInterp();
+  if (Tcl_ListObjGetElements(interp, instances, &list_len, &list_elems) != TCL_OK) {
+      return;
+  }
+  for (int i = 0; i < list_len; i++) {
+      const char *elem = Tcl_GetString(list_elems[i]);
+      cgt->addInstance(elem);
+  }
```
maliberty: Why such low level handling?  It would be easier to take an STL container as an arg and let swig handle this.

src/cgt/test/aes_nangate45.tcl
maliberty, line 1: Please make python versions of the cgt tests

QuantamHD:
> It feels like too much of rmp got pulled into lext. I stopped commenting on it per file but I feel lext should not need an abc dependency and should be generic infrastructure. Rmp is the right place for applying abc to a logic cut.

Logic cuts build mapped abc networks. So it will by default need an abc dep. That's part of the design of logic cut. It can be built from STA and emit networks in other formats, and take those other formats and insert them back into odb/sta

src/cgt/src/NetworkBuilder.cpp
```diff
@@ -0,0 +1,130 @@
...
+             qualifier,
+             network_->name(pin),
+             network_->name(inst_),
+             port->name(),
+             network_->name(net));
+  network_->connectPin(pin, net);
+  return *this;
+}
+
+void NetworkBuilder::init(Logger* logger, sta::dbNetwork* network)
```
maliberty: This code should consider the dont-use list.

src/cgt/src/NetworkBuilder.cpp
```diff
@@ -0,0 +1,130 @@
...
+  network_ = network;
+
+  // Find Liberty cells
+  sta::LibertyLibraryIterator* lib_iter = network_->libertyLibraryIterator();
+  while (lib_iter->hasNext()) {
+    sta::LibertyLibrary* lib = lib_iter->next();
+    sta::LibertyCellIterator cell_iter(lib);
+    while (cell_iter.hasNext()) {
+      auto cell = cell_iter.next();
+      if (!clkgate_cell_ && cell->isClockGateLatchPosedge()) {
```
maliberty: Is isClockGateLatchNegedge not supported?

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+    auto child = child_iter->next();
+    dumpGraphviz(network, child, out);
+  }
+  out << "}\n";
+}
+
+// Times the given function `f` and logs the time taken. Aggregates the time in
+// microseconds.
+template <typename F>
+static auto timed(Logger* logger, size_t& total_us, const char* label, F f)
```
maliberty: Why not use DebugScopedTimer ?

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+  }
+  for (auto seq : cell->sequentials()) {
+    if (seq->isRegister() && seq->output()) {
+      sta::LibertyCellPortIterator port_iter(cell);
+      while (port_iter.hasNext()) {
+        auto port = port_iter.next();
+        if (port->direction()->isAnyOutput()) {
+          assert(port->function()->op() == sta::FuncExpr::op_port);
+          assert(port->function()->port() == seq->output());
+          return network->findPin(inst, port);
```
maliberty, line 154: A reg can have Q and QN outputs (and scan out).  You need to actually check that the output port is the one in the function.

I'm not certain if you only have QN if it will appear in output or outputInv.  You should verify this works in that case.

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+        }
+      }
+    }
+  }
+  return nullptr;
+}
+
+// Returns the function of a register cell, if it exists.
+static sta::FuncExpr* getRegFunction(sta::dbNetwork* network,
```
maliberty: The name is not very descriptive.  This returns the data function.

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+  }
+  std::filesystem::create_directory(*dump_dir_);
+}
+
+// Gathers networks downstream of the given instances (until it encounters
+// registers, and no more than max_count).
+static std::vector<sta::Net*> downstreamNets(
+    sta::dbSta* open_sta,
+    const std::vector<sta::Instance*>& instances,
+    size_t max_count)
```
maliberty: const

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+  }
+  return nets;
+}
+
+// Creates a string that combines names of the given gating condition.
+static std::string combinedGatingCondNames(
+    sta::dbNetwork* network,
+    std::vector<sta::Net*>::const_iterator begin,
+    std::vector<sta::Net*>::const_iterator end,
+    bool clk_enable)
```
maliberty: const

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+    result += network->name(*it);
+  }
+  return result;
+}
+
+// Creates a string that combines names of the given instances.
+static std::string combinedInstanceNames(
+    sta::dbNetwork* network,
+    std::vector<sta::Instance*>::const_iterator begin,
+    std::vector<sta::Instance*>::const_iterator end)
```
maliberty: const

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+             max_cover_);
+  debugPrint(logger_,
+             CGT,
+             "clock_gating",
+             1,
+             "Group registers: {}",
+             group_instances_ == GroupInstances::No ? "No"
+             : group_instances_ == GroupInstances::NetPrefix
+                 ? "Net prefix"
+                 : "Instance prefix");
```
maliberty: avoid nested ternary operators - just make a function for this.

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+  for (const auto& inst_name : instances_) {
+    auto inst = network->findInstance(inst_name.c_str());
+    if (inst) {
+      auto clk_pin = getClockPin(network, inst);
+      if (!clk_pin) {
+        logger_->error(CGT,
+                       1,
+                       "Clock pin not found for instance '{}'",
+                       network->name(inst));
+        continue;
```
maliberty: the continue is unreachable

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+    if (group_instances_ == GroupInstances::NetPrefix) {
+      auto out_pin = getRegOutPin(network, inst);
+      prefix = network->name(network->net(out_pin));
+    }
+    if (group_instances_ != GroupInstances::No) {
+      auto last_dot_pos = prefix.find_last_of('.');
+      if (last_dot_pos == std::string::npos) {
+        last_dot_pos = 0;
+      }
+      auto bracket_pos = prefix.find('[', last_dot_pos);
```
maliberty: A name may have multiple [ - do you want the first?  (eg a[5]/b[10])

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1479 @@
...
+                     std::vector<std::pair<ClockNet*, sta::Instance*>>>
+      inst_groups;
+  for (auto [clk_net, inst] : found_insts) {
+    std::string prefix = network->name(inst);
+    if (group_instances_ == GroupInstances::NetPrefix) {
+      auto out_pin = getRegOutPin(network, inst);
+      prefix = network->name(network->net(out_pin));
+    }
+    if (group_instances_ != GroupInstances::No) {
+      auto last_dot_pos = prefix.find_last_of('.');
```
maliberty: What is special about the '.'?  The option description `instance name prefix` doesn't make this clear.

src/cgt/README.md
gadfort: That could be nearly all designs

QuantamHD: I think they mean FA and HA cells will not be included in ABC. They are treated as black boxes so they can't be leveraged for gating conditions.

maliberty: If so the wording needs improving.

src/lext/include/lext/unique_name.h
kbieganski: It's in `rmp` on master, I don't want to change this code other than moving out the logic cut stuff

src/lext/include/lext/unique_name.h
kbieganski: Again, it's in `rmp` now, I'd rather not touch it

src/cgt/README.md
kbieganski: Thought the wording is OK based on [this](https://github.com/The-OpenROAD-Project/OpenROAD/blob/fe1622a4e97dddc4c0d53c38069aaed8e2301725/src/rmp/src/abc_library_factory.h#L30) but I'll expand it.

src/lext/include/lext/abc_library_factory.h
kbieganski: Yes

src/cgt/src/cgt.i
kbieganski: Wasn't aware Swig can handle it, do I just put an `std::vector` here? I'll try it

src/cgt/src/NetworkBuilder.cpp
kbieganski: It's not, all the ORFS PDKs only have posedge (unless I missed something), so I wasn't able to test negedge.

src/cgt/src/ClockGating.cpp
kbieganski: First from the last dot

src/cgt/src/ClockGating.cpp
kbieganski: It's a hierarchical separator in Verilog. I can drop this feature altogether if you find it too weird

src/cgt/src/cgt.i
maliberty: Yes but you need to add `%include <std_vector.i>` - see other .i files

src/cgt/src/ClockGating.cpp
maliberty: That doesn't mean it will be the hierarchy separator in odb.  I'm not really clear what you are trying to do here with the names.

kbieganski:
Did a rebase to avoid a messy history/git blame, sorry about that.
Stil working on a few review notes (the unresolved ones).

src/cgt/src/ClockGating.cpp
kbieganski: It doesn't let me aggregate the times.

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1428 @@
...
+
+  auto network = sta_->getDbNetwork();
+  network_builder_.init(logger_, network);
+
+  sta_->ensureGraph();
+  sta_->ensureLevelized();
+  sta_->searchPreamble();
+
+  // Initialize ABC
+  abc::Abc_Start();
```
QuantamHD: This probably needs to be refactored into some neutral location on startup, or you need to check to ensure that it has not already been started. RMP also does this.

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1428 @@
...
+  };
+  if constexpr (std::is_void_v<decltype(f())>) {
+    f();
+    end_timer();
+  } else {
+    auto result = f();
+    end_timer();
+    return result;
+  }
+}
```
QuantamHD: I think this is a little messy. Could we change this to a timer object instead that Would have the same-ish params but had an API that looked more like this.
```c++
TimerDebugLogger timer(logger, label, utl::CGT, "clock_gating", 2);
timer.Start();
// Do Stuff
timer.Stop();
```

That way we don't have weird indentation and capture groups for the lambda.

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1428 @@
...
+    Abc_NtkForEachObjReverse(abc_network.get(), obj, idx)
+    {
+      if (Abc_ObjIsCo(obj)) {
+        abc::Abc_NtkDeleteObj(obj);
+      }
+    }
+  }
+
+  abc::Vec_Ptr_t* xor_states = abc::Vec_PtrAlloc(1);
+  abc::Vec_Ptr_t* fanins = abc::Vec_PtrAlloc(2);
```
QuantamHD: All ABC pointer allocs should be managed by utl::UniquePtrWtihDeleter<...> or with a std::unique_ptr where the free function is delete/free

src/cgt/src/ClockGating.cpp
gadfort: There is a `DebugScopedTimer` https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/utl/include/utl/timer.h  This is what I use in PSM and PDN

QuantamHD: Probably should use that. Noice!

src/cgt/test/.gitignore
```diff
@@ -0,0 +1 @@
+abc.history
```
QuantamHD: This should probably go in the root .gitignore

src/cgt/src/ClockGating.cpp
kbieganski: `DebugScopedTimer` doesn't aggregate measurements, and I think it's useful to have the total time for each step.
I'll add it as an option to `DebugScopedTimer`.

src/cgt/src/NetworkBuilder.cpp
gadfort: You could create a fake cell in that has a negedge for your unit testing to be able to test this.

kbieganski:
@maliberty Can you take a look at the security check?

src/cgt/src/ClockGating.cpp
kbieganski: You mean I should check it as a condition, not assert it (second assert)?

Do `output` and `outputInv` correspond to the two pins in `ff()` in Liberty? Like `ff (IQN,IQNN)`? If so, I don't think I need to care about `outputInv`.

maliberty: You may not care but both will match isAnyOutput and the assert would fail on the outputInv

QuantamHD:
@kbieganski It looks like this test is failing in the CI. Can you take a look?

`19:15:22  	756 - cgt.ibex_sky130hd.tcl (Failed)                    IntegrationTest tcl cgt log_compare`

kbieganski:
That's a confusing one. After I merged in master, it's a different failing test:

```
16:38:55  	754 - cgt.aes_nangate45.py (Failed)                     IntegrationTest python cgt log_compare
```

And the [diff](https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-7773-head/14/artifact/src/cgt/test/results.tgz) is:

```diff
43,45c43
< Differences found at line 28809.
<  XNOR2_X1 _17627_ (.A(_08534_),
<  XNOR2_X1 _17627_ (.A(_08534_),
---
> No differences found.
```

(keep in mind this is the log diff, the actual diff is that the line in the output Verilog changed... or did it? :exploding_head:)

QuantamHD:
It might be that there's non-determinism in the CGT algorithm.

kbieganski:
But what's the actual difference? The two lines are the same. (The two Verilog lines from the Python diff function)

QuantamHD:
@maliberty Can you review this? It looks ready to me

maliberty:
Please add the new modules to HeaderFilterRegex in .clang-tidy

src/cut/test/sky130/sky130_fd_sc_hd__ss_n40C_1v40.lib
maliberty, line 1: Why not just link the sky130hd directory instead of file by file - like asap7?

src/cgt/include/cgt/MakeClockGating.h
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2025, The OpenROAD Authors
+
+#pragma once
+
+namespace ord {
+class OpenRoad;
+}
```
maliberty: Unused - rm

src/cgt/src/ClockGating.cpp
```diff
@@ -0,0 +1,1427 @@
...
+          1,
+          "Not enough instances ({}) for clock gate: {}",
+          instances.size(),
+          combinedInstanceNames(network, instances.begin(), instances.end()));
+    }
+  }
+
+  dump("post");
+
+  debugPrint(logger_, CGT, "clock_gating", 1, "Clock gating completed.");
```
maliberty: Just wrap this in a single logger->debugCheck rather than repeatedly checking the same condition.  It will also be more compact.  logger->debug can be used inside.

maliberty (changes requested):

maliberty:
What process have you used to verify the correctness of this code?  I haven't tried to verify you have correctly implemented the paper.

src/cut/test/sky130/sky130_fd_sc_hd__ss_n40C_1v40.lib
kbieganski: I think it was this way in `rmp` or something, so idk. I'll link the directory

kbieganski:
I ran post-synthesis simulation on `aes` and Ibex (simple system example).

maliberty:
Please resolve the merge conflicts

kbieganski:
I'm in the process of doing that. Latest changes are a bit tricky to integrate.

maliberty:
Please ping when this is ready for review

kbieganski:
@maliberty It's ready, the CI failures seem to be some network issues.
Also, updated the description, as after some tweaks the Ibex case now gives 35% improvement.

maliberty:
cgt.ibex_sky130hd.tcl failed in pr-merge.  Please merge master and resolve.

kbieganski:
The CI is green now.

maliberty:
@kbieganski since merging this we find it fails non-deterministically in our ci for the aes_nangate45.tcl test.  I'm looking into it but I hope you can help as well as the author.  If the cause isn't found soon I will revert or disable this change.

kbieganski:
I'll help. The diff from the artifacts is:

```diff
30,32c30
< Differences found at line 53020.
<     .ZN(_02445_));
<     .ZN(_0
---
> No differences found. 
```

So it's as if the file was cut off, but the one in the artifacts is correct.

Are the tests run in parallel? They write to the same file, so perhaps one test overwrites the other's result?

gadfort:
@kbieganski yes the tests can run in parallel, you might want to update the tests to write the unique files based on the test name to avoid this

kbieganski:
@gadfort @maliberty https://github.com/The-OpenROAD-Project/OpenROAD/pull/8281

maliberty:
You should use make_result_file - see any other unit test for examples

maliberty:
I did run with asan while trying to figure it out and there are a bunch of leaks from within abc.  I presume they are somehow related to cgt's usage:

```
=================================================================
==1616090==ERROR: LeakSanitizer: detected memory leaks

Indirect leak of 22800 byte(s) in 285 object(s) allocated from:
    #0 0x784fd68fd340 in calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x61c9aadd0b89 in abc::Mio_LibraryReadPin(char**, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bab89) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 11280 byte(s) in 94 object(s) allocated from:
    #0 0x784fd68fd340 in calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x61c9aadd1d3e in abc::Mio_LibraryRead(char*, char*, char*, int, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bbd3e) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 4096 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aaeaef7c in abc::Mem_FlexEntryFetch(abc::Mem_Flex_t_*, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x5298f7c) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 2304 byte(s) in 36 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcda59 in abc::Mio_ParseFormulaOper(int*, int, abc::Vec_Ptr_t_*, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b7a59) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 2240 byte(s) in 35 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcd750 in abc::Mio_ParseFormulaOper(int*, int, abc::Vec_Ptr_t_*, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b7750) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 2208 byte(s) in 92 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aae895e8 in abc::st__insert(abc::st__table*, char const*, char*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x52735e8) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 1227 byte(s) in 94 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd1de3 in abc::Mio_LibraryRead(char*, char*, char*, int, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bbde3) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 960 byte(s) in 15 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcf46f in abc::Mio_ParseFormula(char*, char**, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b946f) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 795 byte(s) in 94 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd1d6e in abc::Mio_LibraryRead(char*, char*, char*, int, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bbd6e) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 794 byte(s) in 285 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd0bb2 in abc::Mio_LibraryReadPin(char**, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51babb2) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 752 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd0f7f in abc::Mio_LibrarySortGates(abc::Mio_LibraryStruct_t_*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51baf7f) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 752 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd0fc2 in abc::Mio_LibrarySortGates(abc::Mio_LibraryStruct_t_*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bafc2) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 576 byte(s) in 36 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcda07 in abc::Mio_ParseFormulaOper(int*, int, abc::Vec_Ptr_t_*, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b7a07) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 560 byte(s) in 35 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcd572 in abc::Mio_ParseFormulaOper(int*, int, abc::Vec_Ptr_t_*, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b7572) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 512 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aaeaee4b in abc::Mem_FlexStart() (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x5298e4b) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 384 byte(s) in 6 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcd356 in abc::Mio_ParseFormulaOper(int*, int, abc::Vec_Ptr_t_*, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b7356) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 265 byte(s) in 94 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd0a7a in abc::chomp(char*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51baa7a) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd340 in calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x61c9aadd1aed in abc::Mio_LibraryRead(char*, char*, char*, int, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bbaed) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 240 byte(s) in 15 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcf462 in abc::Mio_ParseFormula(char*, char**, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b9462) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 184 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aae88f39 in abc::rehash(abc::st__table*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x5272f39) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 100 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd1b33 in abc::Mio_LibraryRead(char*, char*, char*, int, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bbb33) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 96 byte(s) in 6 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadcd30b in abc::Mio_ParseFormulaOper(int*, int, abc::Vec_Ptr_t_*, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51b730b) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9ab4ad3fe in abc::Mio_GateParseFormula(abc::Mio_GateStruct_t_*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x58973fe) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9ab4ad0be in abc::Mio_GateParseFormula(abc::Mio_GateStruct_t_*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x58970be) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd340 in calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x61c9aaeaee33 in abc::Mem_FlexStart() (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x5298e33) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aae8918d in abc::st__init_table(int (*)(char const*, char const*), int (*)(char const*, int)) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x527318d) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 30 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd1f62 in abc::Mio_LibraryRead(char*, char*, char*, int, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bbf62) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 16 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9ab4ad0b1 in abc::Mio_GateParseFormula(abc::Mio_GateStruct_t_*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x58970b1) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 16 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9ab4ad3f1 in abc::Mio_GateParseFormula(abc::Mio_GateStruct_t_*) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x58973f1) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

Indirect leak of 16 byte(s) in 1 object(s) allocated from:
    #0 0x784fd68fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x61c9aadd1b1f in abc::Mio_LibraryRead(char*, char*, char*, int, int) (/workspace/w7/tools/install/OpenROAD/bin/openroad+0x51bbb1f) (BuildId: c2e87312e5212cb87ed4ab50a77728fe0865f25b)

SUMMARY: AddressSanitizer: 53691 byte(s) leaked in 1237 allocation(s).
```

kbieganski:
I'll take a look at those

Issue #7655, opened 2025-06-25
Can you exclude nets from clock tree synthesis?


maliberty:
### Discussed in https://github.com/The-OpenROAD-Project/OpenROAD/discussions/7653

<div type='discussions-op-text'>

<sup>Originally posted by **alexroach-rl** June 25, 2025</sup>
Hi Guys,

I was wondering if it was possible to exclude a net from clock tree synthesis even if it is combinatorially connected to the root clk? An example would be if the clock was used to generate a pulse signal for a memory read or something like that. In this case you would not want it be part of the H-tree and have delay cells inserted to balance skew, etc.

As a related question, is it possibly to keep a net in the clock tree, but exclude it from skew balancing? I have been using "create_generated_clk -combinational" for this purpose, but it may not be the best approach.</div>

arthurjolo:
'As a related question, is it possibly to keep a net in the clock tree, but exclude it from skew balancing? I have been using "create_generated_clk -combinational" for this purpose, but it may not be the best approach.'
We don't have support for doing it for a specific net, but you can disable skew balancing by adding `-no_insertion_delay `to the CTS arguments.

arthurjolo:
I created a PR to add the option to skip clock tree synthesis for a set of nets. Another alternative would be to specify the root clock nets for CTS by hand with `-clk_nets` and leave out the clocks you don't want to generate. This option, however, would not work if you want to leave out only a subnet of the clock (eg.: the net of gated region of the clock), but with the PR this is possible.

PR #7785, opened 2025-07-10
Cts: add option skip nets

arthurjolo:
Fixes [#7655](https://github.com/The-OpenROAD-Project/OpenROAD/issues/7655)

Add the user option to skip clock tree synthesis for a set of nets: `skip_nets`.

The option receives a string containing the names of the nets that CTS should skip.

  1. If a specified net is the root net for the clock, it will skip the clock tree synthesis for all the sub-nets as well.
  2. If a specified net is a sub-net, it will only skip this net.

### DIFF
--- a/src/cts/README.md
+++ b/src/cts/README.md
@@ -69,6 +69,7 @@ clock_tree_synthesis
     [-library liberty_library_name]
     [-repair_clock_nets]
     [-no_insertion_delay]
+    [-skip_nets <list_of_clk_nets_to_skip>]
 ```
 
 #### Options
@@ -99,6 +100,7 @@ clock_tree_synthesis
 | `-library` | This option specifies the name of Liberty library from which clock buffers will be selected, such as the LVT or uLVT library.  It is assumed that the library has already been loaded using the read_liberty command.  If this option is not specified, clock buffers will be chosen from the currently loaded libraries, which may not include LVT or uLVT cells. |
 | `-repair_clock_nets` | This option includes fixing long wires inside CTS prior to latency adjustment with delay buffers. This can lead to a more balanced clock tree.  The default is not to perform clock net repair. |
 | `-no_insertion_delay` | Ignore sink insertion delay in clock tree construction and balancing. |
+| `-skip_nets` | String containing the names of the clock nets to be skipped. If the net is a root clock net, the whole clock is skipped, otherwise only the subnet specified is skipped. |
 
 #### Instance Name Prefixes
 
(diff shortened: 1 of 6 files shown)


maliberty:
Needs a unit test

src/cts/src/TritonCTS.tcl
```diff
@@ -165,6 +166,14 @@ proc clock_tree_synthesis { args } {
     }
   }
 
+  if { [info exists keys(-skip_nets)] } {
+    set clk_nets $keys(-skip_nets)
+    set fail [cts::set_skip_clock_nets $clk_nets]
+    if { $fail } {
+      utl::error CTS 57 "Error when finding -skip_nets in DB."
```
maliberty, line 173: Let's error in c++ that way python would be the same.

src/cts/README.md
```diff
@@ -99,6 +100,7 @@ clock_tree_synthesis
 | `-library` | This option specifies the name of Liberty library from which clock buffers will be selected, such as the LVT or uLVT library.  It is assumed that the library has already been loaded using the read_liberty command.  If this option is not specified, clock buffers will be chosen from the currently loaded libraries, which may not include LVT or uLVT cells. |
 | `-repair_clock_nets` | This option includes fixing long wires inside CTS prior to latency adjustment with delay buffers. This can lead to a more balanced clock tree.  The default is not to perform clock net repair. |
 | `-no_insertion_delay` | Ignore sink insertion delay in clock tree construction and balancing. |
+| `-skip_nets` | String containing the names of the clock nets to be skipped. If the net is a root clock net, the whole clock is skipped, otherwise only the subnet specified is skipped. |
```
maliberty, line 103: It would be better to take a list of names rather than a string you have to parse.  get_nets could then be used by the user.

maliberty (changes requested):

precisionmoon:
It'd be good to introduce separate commands to configure CTS.   I suggest that we start with these to close the gap with commercial tools.

- set_cts_config -exclude_pins "pins which should not be treated as clock sinks"
- set_cts_config -no_balance_pins "pins which should be treated as clock sinks but ignored for skew balancing"
- set_cts_config -propagate_pins "FF/CP pins that are clock dividers so we should continue propagating clocks until we find real clock sinks" (** can be implemented later**)

report_cts_config

reset_cts_config -exclude_pins  / -no_balance_pins / -propagate_pins

It'd be good for clock_tree_synthesis to print report_cts_config at the beginning.

arthurjolo:
Closing this PR as it is beeing continued by PR [#8502](https://github.com/The-OpenROAD-Project/OpenROAD/pull/8502).

PR #7796, opened 2025-07-12
Fix segfault when parsing a corrupt def file

Faholan:
DEF error numbers are handled by the defError function with an offset of 5000.

However, the defyyerror function uses the `defMsgCnt` attribute, which is initialized at 0, causing an out-of-bounds access.

According to the `def.y` file, yyerror error codes start at 5500, so the fix is imply to initialize the `defMsgCnt` variable to this value

### DIFF
--- a/src/odb/src/def/def/defrData.hpp
+++ b/src/odb/src/def/def/defrData.hpp
@@ -128,7 +128,7 @@ class defrData
   int componentWarnings{0};
   int constraintWarnings{0};
   int cover_is_keyword{0};
-  int defMsgCnt{0};
+  int defMsgCnt{5500};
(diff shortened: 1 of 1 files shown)


maliberty (approved):
This will fix the invalid access.  Using defMsgCnt as a msgNum seems wrong from the start.  Probably it should just be a fixed value.

maliberty:
Please fix the DCO (see the details link)

Faholan:
I don't know whether using a fixed value is a better idea. However, `defMsgCnt` is only used as `msgNum` and not referenced anywhere else in the code. And sorry for the DCO, I always forget to signoff

Issue #7745, opened 2025-07-08
[ERROR GPL-0305] RePlAce diverged during gradient descent calculation


oharboe:
### Describe the bug

untar and run [global_place_skip_io_tag_array_64x184_asap7_mocked_2025-07-08_09-04.tar.gz](https://github.com/user-attachments/files/21116463/global_place_skip_io_tag_array_64x184_asap7_mocked_2025-07-08_09-04.tar.gz)

```
     2210 |   0.3337 |  0.000000e+00 |   +0.00% |  1.42e+33 |      
     2220 |   0.3337 |  0.000000e+00 |   +0.00% |  2.09e+33 |      
     2230 |   0.3337 |  0.000000e+00 |   +0.00% |  3.07e+33 |      
     2240 |   0.3338 |  0.000000e+00 |   +0.00% |  4.53e+33 |      
     2250 |   0.3337 |  0.000000e+00 |   +0.00% |  6.67e+33 |      
[ERROR GPL-0305] RePlAce diverged during gradient descent calculation, resulting in an invalid step length (Inf or NaN). This is often caused by numerical instability or high placement density. Consider reducing placement density to potentially resolve the issue.
Error: global_place_skip_io.tcl, 12 GPL-0305
openroad> 
```

### Expected Behavior

No error, used to work.

### Environment

```shell
OpenROAD v2.0-22423-gd59854ee88
```

### To Reproduce

See above

### Relevant log output

```shell

```

oharboe:
Tried a bisection, a090ebaa03b3171cc7282e49d651c8280d9372b2 is the most recent that supports the .odb format in the test-case and it fails too, so nothing to bisect on with this test-case.

```
$ git log --pretty=format:'%h' -- OpenROAD | while read commit; do  git show $commit OpenROAD | grep "Subproject commit" | awk '{print $3}'; done | head -n 10
c79c8317e05ecbf45eba5096a74cb421ab97b7ae
a090ebaa03b3171cc7282e49d651c8280d9372b2
...
```

gudeh:
Hi @oharboe, was this design working previously or is it a new one? Have you notice the constant 0 HPWL?

looking at it we seem to not have any net. I right clicked and highlighted all nets of all instances and no net showed up:

<img width="293" height="532" alt="Image" src="https://github.com/user-attachments/assets/46b9a0c7-ba9b-4e1a-bd74-5bcc1f917a21" />

oharboe:
[...]

PR #7798, opened 2025-07-12
gpl: avoid error at 0 hpwl

gudeh:
Change requested in https://github.com/The-OpenROAD-Project/OpenROAD/issues/7745


src/gpl/test/nograd01.defok
```diff
@@ -406,12 +406,12 @@ TRACKS Y 270 DO 444 STEP 270 LAYER M2 ;
...
-    - _2_ BUFx2_ASAP7_75t_R + PLACED ( 59898 59698 ) N ;
-    - _3_ BUFx2_ASAP7_75t_R + PLACED ( 59907 59880 ) N ;
-    - _4_ BUFx2_ASAP7_75t_R + PLACED ( 59421 60683 ) N ;
-    - _5_ BUFx2_ASAP7_75t_R + PLACED ( 60760 60083 ) N ;
+    - _0_ BUFx2_ASAP7_75t_R ;
+    - _1_ BUFx2_ASAP7_75t_R ;
+    - _2_ BUFx2_ASAP7_75t_R ;
+    - _3_ BUFx2_ASAP7_75t_R ;
+    - _4_ BUFx2_ASAP7_75t_R ;
+    - _5_ BUFx2_ASAP7_75t_R ;
```
rovinski, line 414: Doesn't this mean that the cells are now unplaced? Does the detailed placer handle this?

maliberty: Given that a circuit with no nets can't actually do anything this seems an unrealistic test (a bunch of floating buffers).  I think the motivation was an empty block with only pins.

rovinski: But wasn't the whole point of this PR to have the flow not crash? Wouldn't this just kick the problem over to DPL which will then crash due to no placed instances?

maliberty: To crash and to error are different things.  I don't know if dpl would crash or not.  We could refine this to error if there are any instances and no nets, and just skip if there are no instances.

maliberty: In that case we should place the instances as otherwise there may be downstream problems.  What produces the crash currently?

gudeh: I shared some more things in the issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/7745, and answered your question.

maliberty:
After some discussion I think this is better fixed another way so that we do place these cells.

gudeh:
replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/7814

PR #7799, opened 2025-07-12
IP checker utility

MarufHossain-tech:
6 files are added.  the code in pinchecker.py
Screenshot 2025-07-13 011908

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,366 +1,59 @@
-# OpenROAD
+# OpenROAD IP Checker
 
-[![Build Status](https://jenkins.openroad.tools/buildStatus/icon?job=OpenROAD-Public%2Fmaster)](https://jenkins.openroad.tools/job/OpenROAD-Public/job/master/)
-[![Coverity Scan Status](https://scan.coverity.com/projects/the-openroad-project-openroad/badge.svg)](https://scan.coverity.com/projects/the-openroad-project-openroad)
-[![Documentation Status](https://readthedocs.org/projects/openroad/badge/?version=latest)](https://openroad.readthedocs.io/en/latest/?badge=latest)
-[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/5370/badge)](https://bestpractices.coreinfrastructure.org/en/projects/5370)
+Utility for checking supported IP blocks in the OpenROAD project.
 
-## About OpenROAD
+## Features
+- Check if an IP is supported
+- List all supported IPs
+- JSON output option
+- Colorized terminal output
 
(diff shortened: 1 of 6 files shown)


MarufHossain-tech:
I want your supervision. I uploaded the code in ip_checker.py

maliberty:
I don't think this is the right path.  You'll want to build something in OpenROAD itself otherwise you would have to develop something equivalent to odb.  You could write the checker in tcl or python that uses the OR APIs to implement the checks.

MarufHossain-tech:
Thanks for the guidance! To ensure I’m aligned with OpenROAD’s design philosophy, I’d appreciate clarification on a few points:
1. Should the IP checker be implemented as,
A new Tcl command (e.g., check_supported_ip {ip_name})? A Python utility using odbpy (e.g., querying dbTechLib)? Or extend an existing tool like report_lib?
2. Should we validate IPs against Loaded LEF/DEF libraries (db.getLibs())? or A predefined PDK configuration file? or Other OpenROAD database structures?
3. Are there preferred test cases (e.g., gcd flow) or CI hooks to validate this feature?

I’ll adapt my implementation based on your input and submit a revised PR. For transparency, here’s a rough sketch of the proposed Tcl/Python approach. Let me know if this direction makes sense:
\\TCL example: \\

proc is_supported_ip {ip_name} {
    set libs [get_libs]
    return [expr {[lsearch $libs $ip_name] >= 0}]
}

\\Python example: \\

from odb import dbDatabase
def is_supported(ip_name):
    db = dbDatabase.create()
    db.readLef("tech.lef")  # Assume pre-loaded
    return ip_name in [lib.getName() for lib in db.getLibs()]

Looking forward to your thoughts!

maliberty:
If you are comfortable with c++ you can implement the checker in c++ and then add interface in tcl & python both.  If not you could implement it in just one of those languages.  I think the former is preferable.

OR already has commands to load data so I would suggest the checker assume it is already loaded into odb as a separate step.

MarufHossain-tech:
@maliberty Thanks for clarifying! I’ll:
1. Implement the core checker in C++ (using odb).  
2. Add Tcl/Python interfaces.  
3. Remove the standalone files.  

Should I place this under `src/odb/` or create a new module?  
Also, is there an existing test design (e.g., `gcd`) to validate against?

maliberty:
The checker may need access to timing data so it might be better to put it in dbSta (which depends on odb but not visa-versa).   You could test gcd though it doesn't have any errors afaik.  You will probably need to construct some failing tests yourself.  See the tests in dbSta/test for examples.

PR #7808, opened 2025-07-14
C++ pin_checker & tcl interface , uploaded in dbSta 

MarufHossain-tech:


MarufHossain-tech:
as per your instruction I made the core checker in c++ & interface in tcl. added them on dbSta

MarufHossain-tech:
image
seems like there is a cross on a commit. I'm having trouble finding our what is the issue. can you help me identifying this?

maliberty:
I retriggered the build as it seems like it might be an intermittent issue.

src/dbSta/pin_checker.cc.cpp
```diff
@@ -0,0 +1,74 @@
...
+    
+    result.valid = result.errors.empty();
+    return result;
+}
+
+double LefChecker::getManufacturingGrid() const {
+    return db_->getTech() ? db_->getTech()->getManufacturingGrid() : 0.001;
+}
+
+bool LefChecker::isOnGrid(double value, double grid) const {
```
maliberty: Layout is done in integer DBU and you should avoid floating point.  It solves a multitude of problems.

src/dbSta/CMakeLists2.txt
maliberty, line 1: Why is a new cmakefile file needed?  It would be better to just add to the existing one as this is not connected to anything.

src/dbSta/pin_checkerTcl_interface.i
```diff
@@ -0,0 +1,22 @@
+# Define the Tcl command
+namespace eval ord {
+    proc check_lef {lib_name} {
```
maliberty: tcl code should go in dbSta.tcl

src/dbSta/pin_checkerTcl_interface.i
maliberty, line 1: No need for a new .i file, just add to dbSta.i

MarufHossain-tech:
I have update based on your requirements. let me know if further change required

MarufHossain-tech:
I have update based on your requirements. let me know if further change required

PR #7817, opened 2025-07-15
Avoid a stack use after scope

Faholan:
std::minmax returns references, so the arguments need to have a valid allocation


Faholan:
On an array, std::minmax actually returns the min and max by value and not by reference, so this avoids introducing extra variables

PR #7823, opened 2025-07-16
Cts: create new balance latencies method

arthurjolo:
Needs [#7821](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7821)

This PR creates a new clock latencies balancer. Currently CTS only balance latencies between macros and registers trees, but when we have multiple sub-trees (eg.: multiple clock gaters) the sub-trees are not balanced between them eg.:

unity test gated_clock3 with master
gated_clock3_master

This PR proposes a new balance latencies method, that takes into account all sub-trees connected to a same clock source, the LatenciesBalancer. This class builds a graph that connects all sub-trees of the clock and travels throw it to insert delay buffers where its needed. The delay buffers are inserted as high up in the tree as possible, and try to share the mod amount of buffers as possible as well. A more detailed explanation of the algorithm can be found [here](https://docs.google.com/presentation/d/1G5pWCrPHad6GCvFkKXFHwl0x5MI2cGiPcu-HwUBJMsU/edit?slide=id.g37046e681e6_0_109#slide=id.g37046e681e6_0_109).

This changes help improve skew specially for designs with multiple clock gaters, where the problem mention tends to appear, eg.: 

unity test gated_clock3 with this PR
gated_clock3_LatenciesBalancer


src/cts/src/CMakeLists.txt
maliberty, line 1: You'll need to add the new file to BUILD as well for Bazel

src/cts/src/LatenciesBalancer.cpp
```diff
@@ -0,0 +1,512 @@
...
+    }
+    debugPrint(
+        logger_, CTS, "insertion delay", 3, "  {}", sinkInput->getName());
+    sinkInput->disconnect();
+    int sinkX, sinkY;
+    sinkInput->getAvgXY(&sinkX, &sinkY);
+    maxSinkX = std::max(maxSinkX, sinkX);
+    maxSinkY = std::max(maxSinkY, sinkY);
+    minSinkX = std::min(minSinkX, sinkX);
+    minSinkY = std::min(minSinkY, sinkY);
```
maliberty: It would be simpler to use a Rect and mergeInit() / merge() and xCenter/yCenter below.

src/cts/src/LatenciesBalancer.cpp
```diff
@@ -0,0 +1,512 @@
...
+  if (inputPort) {
+    return inputPort->isRegClk();
+  }
+
+  return false;
+}
+
+void LatenciesBalancer::showGraph()
+{
+  logger_->report("Graph builded:");
```
maliberty: builded ->built

maliberty (changes requested):

src/cts/src/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ add_library(cts_lib
     TritonCTS.cpp
     Clustering.cpp
     LevelBalancer.cpp
+    LatenciesBalancer.cpp
```
precisionmoon: Can we rename the file LatencyBalancer.cpp?

precisionmoon: And rename LatenciesBalancer.h to LatencyBalancer.h?

arthurjolo: Ok, Do you want me to rename the class as well or just the files?

precisionmoon: Let's rename the classes too.

precisionmoon (commented):
Do we also want to update the CTS summary to include new delay buffers?
For example,

[INFO CTS-0098] Clock net "clk"
[INFO CTS-0099]  Sinks 225
[INFO CTS-0100]  Leaf buffers 103
**[INFO CTS-XXXX] Delay buffers 25**
[INFO CTS-0101]  Average sink wire length 9313.40 um
[INFO CTS-0102]  Path depth 16 - 17
[INFO CTS-0207]  Leaf load cells 4
[INFO CTS-0098] Clock net "clk_regs"
[INFO CTS-0099]  Sinks 2254
[INFO CTS-0100]  Leaf buffers 227
**[INFO CTS-XXXX] Delay buffers 10**
[INFO CTS-0101]  Average sink wire length 4121.94 um
[INFO CTS-0102]  Path depth 17 - 17
[INFO CTS-0207]  Leaf load cells 4
Total number of Clock Roots: 2.
Total number of Buffers Inserted: 556.
Total number of Clock Subnets: 556.
Total number of Sinks: 2475.

arthurjolo:
We can't do it for each clock net as you are suggesting, because LatencyBalancer is not running for each clock net, it is running for all clock nets of one clock together. We can report how many were inserted by each clock and the total amount.

precisionmoon:
> We can't do it for each clock net as you are suggesting, because LatencyBalancer is not running for each clock net, it is running for all clock nets of one clock together. We can report how many were inserted by each clock and the total amount.

Yes, this makes sense.

src/cts/src/LatencyBalancer.cpp
```diff
@@ -0,0 +1,569 @@
...
+namespace cts {
+
+using utl::CTS;
+
+int LatencyBalancer::run()
+{
+  logger_->info(
+             CTS,
+             33,
+             "Balancing latancy for clock {}",
```
precisionmoon: A small typo:  latancy => latency

maliberty:
Is there a metrics update PR?

src/cts/test/CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ or_integration_tests(
     dummy_load
     find_clock
     find_clock_pad
+    gated_clock1
+    gated_clock2
+    gated_clock3
+    gated_clock4
+    gated_clock5
```
maliberty, line 33: Add to the BUILD file.

src/cts/src/TritonCTS.cpp
```diff
@@ -2340,23 +2341,31 @@ void TritonCTS::balanceMacroRegisterLatencies()
...
+      LatencyBalancer balancer = LatencyBalancer(builder,
+                                                 options_,
+                                                 logger_,
+                                                 db_,
+                                                 network_,
+                                                 openSta_,
+                                                 techChar_->getLengthUnit(),
+                                                 capPerDBU);
+      totalDelayBuff += balancer.run();
+      parasitics_guard.update();
```
maliberty: Move the guard to just inside the if-statement and the end of the scope will handle the update.  The goal of the guard is to avoid needing to manually call update and to be exception safe.

src/cts/src/LatencyBalancer.cpp
```diff
@@ -0,0 +1,616 @@
...
+  std::string newNetName
+      = "delaynet_" + std::to_string(delayBufIndex_) + "_" + clockName;
+  odb::dbNet* newNet = odb::dbNet::create(block, newNetName.c_str());
+  newNet->setSigType(odb::dbSigType::CLOCK);
+
+  // create a new delay buffer
+  std::string newBufName
+      = "delaybuf_" + std::to_string(delayBufIndex_++) + "_" + clockName;
+  odb::dbMaster* master = db_->findMaster(options_->getRootBuffer().c_str());
+  odb::dbInst* newBuf = odb::dbInst::create(block, master, newBufName.c_str());
```
maliberty: @jhkim-pii any concerns about hierarchy?

arthurjolo: This code should be the same as what we already have in CTS for inserting delay buffers (`TritonCTS::insertDelayBuffer`), I see that this code has changed, probably to deal with hierarchy. So probably what we have on this PR should be changed as well.

src/cts/src/LatencyBalancer.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+  for (odb::dbITerm* sinkInput : sinksInput) {
+    sinkInput->connect(drivingNet);
+  }
+  return getFirstInput(returnBuffer);
+}
+
+// Create a new delay buffer and connect output pin of driver to input pin of
+// new buffer. Output pin of new buffer will be connected later.
+odb::dbInst* LatencyBalancer::createDelayBuffer(odb::dbNet* driverNet,
```
arthurjolo, line 512: @jhkim-pii can you give your review on this function to insure hierarchy is fine.

I tried to maintain the idea you used on TritonCTS::insertDelayBuffer, so the delay buffers are on the same module as their drivers, this way we avoid creating any extra ports on the modules for the worst cases.

jhkim-pii: It looks good. As you mentioned, this makes a flat connection b/w the driver and the new buffer. So it does not need any hierarchy handling.

src/cts/src/LatencyBalancer.cpp
```diff
@@ -0,0 +1,635 @@
...
+  GraphNode* node = &graph_[nodeId];
+
+  // Compute number of buffer needed for leaf node
+  if (node->childrenIds.empty()) {
+    return;
+  }
+
+  // If it is not a leaf node compute the amount of buffers needed for its
+  // children
+  std::vector<odb::dbITerm*> sinksInput;
```
jhkim-pii, line 394: In a hierarchical flow, `sinksInput` can be located at different hierarchies, right?
Then, a hierarchy handling is required when the sink pin is disconnected and reconnected to the output pin of the new buffer because the hierarchy of the driver and the sink instance are different.

A simple solution is to use `dbNetwork::hierarchicalConnect(dbITerm* source_pin, dbITerm* dest_pin, const char* connection_name)` for the connection b/w the buffer output and the sink. `hierarchicalConnect()` automatically handles the hierarchy. 
But its drawback is that it does not reuse the existing hierarchical ports and punches many new hierarchical ports.

jhkim-pii: @arthurjolo Could you please add a test case with hierarchical flow? I think we can find issues from the hierarchical test case.

arthurjolo: There is a testcase `hier_insertion_delay` but maybe it is too simple.

arthurjolo: Is it necessary to do anything else at the disconnect? If i disconnect the pins and create a new hierarchical port than the previous connection port is dangling right?

jhkim-pii: Yes. `hierarchicalConnect()` will create a new port and make the previous dangling. It is the drawback of the current `hierarchicalConnect()`. 
I will enhance `hierarchicalConnect()` to fully reuse the existing nets and ports to minimize the port punching.

arthurjolo: I modified testcase `gated_clock4` to have a simple hierarchy and added `hierarchicalConnect()` for the delay buffers and its sinks. On the verilog output of `gated_clock4` it looks like there the dangling ports are deleted but @jhkim-pii if you could take a look to confirm it is ok.

jhkim-pii: Yes. It is the expected behavior. `hierarchicalConnect()` removes unused ports and pins in the end. I forgot to mention it yesterday.

```
void dbEditHierarchy::hierarchicalConnect(dbITerm* source_pin,
                                          dbITerm* dest_pin,
                                          const char* connection_name)
...                                         
    // 3.6. During the addition of new ports and new wiring we may
    // leave orphaned pins, clean them up.
    std::set<dbModInst*> cleaned_up;
    for (auto module_to_clean_up : source_parent_tree) {
      dbModInst* mi = module_to_clean_up->getModInst();
      if (mi) {
        dlogHierConnCleaningUpSrc(mi);
        mi->removeUnusedPortsAndPins();
        cleaned_up.insert(mi);
      }
    }
    for (auto module_to_clean_up : dest_parent_tree) {
      dbModInst* mi = module_to_clean_up->getModInst();
      if (mi) {
        if (cleaned_up.find(mi) == cleaned_up.end()) {
          dlogHierConnCleaningUpDst(mi);
          mi->removeUnusedPortsAndPins();
          cleaned_up.insert(mi);
        }
      }
    }
  }

  dlogHierConnDone();    
}
```

arthurjolo:
Started a new CI to update the metrics. I was having an error in DPL after grt for one design, but it looks like it was fixed.

maliberty:
If the metrics are fine we can merge

arthurjolo:
Metrics are fine, there is one PR to update some small failures. However, I have one timeout in the private CI.

Issue #7740, opened 2025-07-07
Missing descriptors for dbDft objects


AcKoucher:
### Suggested Solution

Add descriptors to the following objects so we can inspect them from the GUI:

 - [x] dbScanChain
 - [x] dbScanPartition
 - [x] dbScanList
 - [x] dbScanInst

### Additional Work

 - [x] Add ability to inspect dbScanInst through dbInst.

PR #7824, opened 2025-07-16
odb: move ownership of _dbScanInst to _dbBlock

AcKoucher:
Requires #7825.

## Context and Changes
As I was working on #7740, I realized that we need some way of linking the instance to its scan version in order to allow fast inspection through the descriptor of the `dbInst`.

One solution would  be to hash instances to their scan version. However, with the current ownership hierarchy that is not possible. This PR is to allow the cited solution.

With this, `_dbScanInst`s in a `_dbScanList` will work as `_dbIterm`s in a `_dbNet`.

## Important Considerations
Now that we insert the `ScanInst`s in a `ScanList` in the opposite order, we need to manually reverse the order when reading a DEF file. Same idiom used with `dbIterm`s in a `dbNet`.


AcKoucher:
@fgaray Could you take a look?

src/dft/test/scandef_sky130.defok
fgaray, line 1: Why is the order of the scandef changing?

fgaray:
> @fgaray Could you take a look?

I think you have more experience on odb than me. As long as it is functional equivalent to the original code it LGTM. One small comment as seems like the order of some flops is changing? Why is that?

src/odb/include/odb/db.h
```diff
@@ -8440,6 +8440,8 @@ class dbScanInst : public dbObject
 
   dbInst* getInst() const;
 
+  void insertInScanList(dbScanList* scan_list);
```
maliberty: This should be clear about whether you are adding to the front or back of the list.  I guess that may relate to the order difference.

src/odb/src/codeGenerator/schema/scan/dbScanList.json
```diff
@@ -9,6 +9,25 @@
...
+    "dbScanInstItr.h"],
+  "fields": [
+    {
+      "name": "_unused",
+      "type": "uint",
+      "flags": [
+        "private"
+      ],
+      "default": 0,
+      "comment": "// As the elements of a free dbTable are 12 bytes long, we need this\n// additional member in order to make _dbScanList big enough to allow\n// safe casting between table members."
```
maliberty: I would expect clang-format to wrap the comment for you (we format all generated files).

src/odb/src/db/dbScanInst.cpp
```diff
@@ -59,6 +65,12 @@ dbIStream& operator>>(dbIStream& stream, _dbScanInst& obj)
   stream >> obj.inst_;
   stream >> obj.scan_clock_;
   stream >> obj.clock_edge_;
+  if (obj.getDatabase()->isSchema(db_schema_block_owns_scan_insts)) {
+    stream >> obj._next_list_scan_inst;
+  }
+  if (obj.getDatabase()->isSchema(db_schema_block_owns_scan_insts)) {
+    stream >> obj._prev_list_scan_inst;
+  }
```
maliberty, line 73: These can be combined into one if.

src/odb/src/db/dbScanInst.cpp
```diff
@@ -70,6 +82,12 @@ dbOStream& operator<<(dbOStream& stream, const _dbScanInst& obj)
   stream << obj.inst_;
   stream << obj.scan_clock_;
   stream << obj.clock_edge_;
+  if (obj.getDatabase()->isSchema(db_schema_block_owns_scan_insts)) {
+    stream << obj._next_list_scan_inst;
+  }
+  if (obj.getDatabase()->isSchema(db_schema_block_owns_scan_insts)) {
+    stream << obj._prev_list_scan_inst;
+  }
```
maliberty: Likewise

src/odb/src/db/dbScanInst.cpp
```diff
@@ -221,19 +231,33 @@ dbScanInst::AccessPins dbScanInst::getAccessPins() const
...
 }
 
+void dbScanInst::insertInScanList(dbScanList* scan_list_)
+{
+  _dbScanInst* scan_inst = (_dbScanInst*) this;
+  _dbScanList* scan_list = (_dbScanList*) scan_list_;
+  _dbBlock* block = (_dbBlock*) scan_inst->getOwner();
+
+  if (scan_list->_scan_insts != 0) {
+    _dbScanInst* tail = block->_scan_inst_tbl->getPtr(scan_list->_scan_insts);
```
maliberty: This is the head of the list not the tail

src/odb/src/codeGenerator/schema/scan/dbScanList.json
```diff
@@ -9,6 +9,25 @@
...
+    },
+    {
+      "name": "_scan_insts",
```
maliberty: It would be clearer to call this _first_inst

src/odb/src/db/dbScanList.cpp
```diff
@@ -31,43 +34,27 @@ bool _dbScanList::operator<(const _dbScanList& rhs) const
...
-  scan_insts_ = new dbTable<_dbScanInst>(
-      db, this, (GetObjTbl_t) &_dbScanList::getObjectTable, dbScanInstObj);
+  _unused = 0;
 }
 
 dbIStream& operator>>(dbIStream& stream, _dbScanList& obj)
 {
-  stream >> *obj.scan_insts_;
+  stream >> obj._unused;
+  stream >> obj._scan_insts;
```
maliberty: Needs a schema check

src/odb/src/db/dbScanInst.cpp
gadfort: this is writing, so we care what the incoming schema was? seems like this is a no-op or not needed.

src/odb/src/db/dbScanInst.cpp
AcKoucher: These came from the code generator. Is there a way to enforce it to keep it a single condition?

src/dft/test/scandef_sky130.defok
AcKoucher: The new logic to add a `_dbScanInst` to a `_dbScanList` works as a `push_front()` (same behavior as inserting an iterm in a net). Previously, we were using something like `dbScanInst::create(_scan_list, inst);` which works as a `push_back()`.

@maliberty Please, correct me if I'm wrong.

src/odb/src/db/dbScanInst.cpp
AcKoucher: FYI this also comes from the generator.

src/dft/test/scandef_sky130.defok
maliberty: We could reverse the list once it is created.  That idiom is used elsewhere.

src/odb/src/db/dbScanInst.cpp
maliberty: It would make the generator more complex so let's just live with it.

src/odb/src/db/dbScanInst.cpp
maliberty: I agree we don't need to write in an old schema.  I'll update the code generator in a separate PR.

maliberty: #7870

maliberty: @AcKoucher you can merge master to this branch and re-run the code generator to resolve this.

src/dft/test/scandef_sky130.defok
AcKoucher: Added commits to ensure the current order is preserved when reading from a `.def`.

PR #7825, opened 2025-07-16
dft: keep track of just .defok for scandef_sky130 test

AcKoucher:
As I was updating scandef_sky130 test result for #7824, I realized that its input `.def` file is tracked by git. That feels unnecessary and makes us need to update two files when updating the test.

This PR creates the symbolic link for `save_defok` and changes the test to have a somewhat more friendly `.defok` update process.

Also removing an unused file.


AcKoucher:
@fgaray FYI.

AcKoucher:
The PR check failures seem to be IT related.

PR #7842, opened 2025-07-18
bazel-orfs: bump

oharboe:
Simplified use-case for bazel run foo_deps, WORK_HOME is in generated config.mk


oharboe:
@vvbandeira @maliberty same unrelated CI failure

Issue #7833, opened 2025-07-17
Usability of route guides


maliberty:
### Description

Two improvements that would make the route guides more usable:

1) The inspector's 'route guide' button functions as an on/off toggle but it doesn't give any visual feedback as to whether the guides are on or off for the inspected net.

2) When we have detailed routes we draw the net highlight along the detailed routes but when we have global routes we just use the same flightlines as we do before grt.  We should do similar highlighting with the guides.

@gadfort would you have any time for these?

maliberty:
@precisionmoon FYI

gadfort:
@maliberty I can take a look this weekend. The first one should be fairly easy, the second one I think would require a little more effort,  it should be doable if I understand correctly

gadfort:
@maliberty I was poking around at using the guides, at first I tried to draw the guide outlines, but that was crazy busy and nearly useless. So right now I'm just drawing the center line of the guide and then flywires from the box center to the iterms. For some nets this doesn't look too odd, but for short nets it looks a little funny. What are your thoughts?

Clock net: 

<img width="778" height="1081" alt="Image" src="https://github.com/user-attachments/assets/3a883d71-a703-4569-b71e-b76e5fb8eb0f" />

Short net:

<img width="1145" height="848" alt="Image" src="https://github.com/user-attachments/assets/225e78e1-10be-489b-adfb-7c908ec6e131" />

Net0:

<img width="1593" height="889" alt="Image" src="https://github.com/user-attachments/assets/c9bb2124-424c-4e21-be04-ac6f49296092" />

maliberty:
I think it would be good to lightly draw the gcell boundaries.  I think it would make it clearer that we are going gcell center to pin.  Btw, that's not too far from what parasitic estimation does.

@eder-matheus any opinion?

maliberty:
@precisionmoon FYI

gadfort:
@maliberty to make the guides a lot of less busy, I added a merging function to the polygon in odb.

Clock net:

<img width="1314" height="1508" alt="Image" src="https://github.com/user-attachments/assets/4a12f434-30d0-49a1-b112-569af48a33c7" />

Short net:

<img width="1345" height="511" alt="Image" src="https://github.com/user-attachments/assets/33e90d69-4935-4208-a15e-42a8e05a4dd4" />

Bterm:

[...]

PR #7846, opened 2025-07-19
gui: update actions in inspector and change button for route guides based on visibility

gadfort:
Part 1 of #7833

Before turning on: 
image

After turning on:
image

### DIFF
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -1590,14 +1590,19 @@ Descriptor::Actions DbNetDescriptor::getActions(std::any object) const
                        }});
   }
   if (!net->getGuides().empty()) {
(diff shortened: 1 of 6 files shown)


maliberty:
If you show a route guide, right click and select clear route guides, the inspector still shows hide.

gadfort:
@maliberty try again, I added net tracks to this as well.

maliberty:
I see it works for route guides but not for net tracks.  Clearing the tracks still leaves it on hide.

gadfort:
@maliberty it seems that net tracks were just doing their own thing, I don't know how to view them / load them, so I don't have a good way to check if this works.

maliberty:
I mention it since you said you worked on it.  Its a debug tool so it is enabled with `detailed_route_debug -write_net_tracks` before running detailed routing.  It draws the results of track assignment.  Its not crucial.

maliberty (approved):

PR #7849, opened 2025-07-20
Ip utility checker 1

MarufHossain-tech:


MarufHossain-tech:
@maliberty Can you give  me feedback on what should i do next? I've made the changes you asked for

maliberty:
This doesn't actually do anything so I think a PR is premature.  If you want to check that a pin is on-grid then you need to iterate over pins and check their shapes.

PR #7851, opened 2025-07-20
DRT: skip non active and special bterms warnings

osamahammad21:


osamahammad21:
@maliberty No CI change is expected

maliberty (approved):

PR #7852, opened 2025-07-20
gui: use guides for net highlighting when available

### LINKED ISSUE #7833: Usability of route guides

gadfort:
Closes #7833

Adds:
- odb::Polygon merge functions to merge sets of shapes into a set of polygons
- gui::Painter add drawLine to take in a odb::Line.

Changes:
- when detailed routing information is available it will be used for highlighting, when guides are available it will be used, and finally flywires will be used if neither is available (in that order)

Notes:
- This code will draw the outline of the guide boxes and use the centerline of the box to represent the routing with a "flywire" connecting the center line to the iterm or bpins.
- The center line is drawn with two pixels (or 4 seemed like good numbers, 4 is a more clear, but I left it 2 for now), 1 pixel was nearly impossible to see what was going on.
- The "fatness" of the routing in the examples stem from the routing box outlines and don't seem to be impacted by the centerline with that much.

Examples:
Clock routing before for cva6:
cva6 core_clock_before
After for cva6:
cva6 core_clock

Clock routing before for ethmac:
ethmac wb_clk_i
After for ethmac:
ethmac wb_clk_i

Timepath drawing:
image

Selecting a single net:
image

### DIFF
--- a/src/gui/include/gui/gui.h
+++ b/src/gui/include/gui/gui.h
@@ -216,6 +216,7 @@ class Painter
 
   // Draw a line with coordinates in DBU with the current pen
   virtual void drawLine(const odb::Point& p1, const odb::Point& p2) = 0;
+  void drawLine(const odb::Line& line) { drawLine(line.pt0(), line.pt1()); }
 
   // Draw a circle with coordinates in DBU with the current pen
   virtual void drawCircle(int x, int y, int r) = 0;
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -1430,10 +1430,13 @@ void DbNetDescriptor::highlight(std::any object, Painter& painter) const
           }
         }
 
+        std::vector<odb::Rect> guide_rects;
+        guide_rects.reserve(guides.size());
         for (const auto* guide : guides) {
           const auto& box = guide->getBox();
           const auto center = box.center();
           const int width_half = box.minDXDY() / 2;
+          guide_rects.push_back(box);
           odb::Point p0, p1;
           switch (box.getDir()) {
             case 0: {
@@ -1494,6 +1497,11 @@ void DbNetDescriptor::highlight(std::any object, Painter& painter) const
         }
 
         painter.restoreState();
+
+        // draw outlines of guides
+        for (const odb::Polygon& outline : odb::Polygon::merge(guide_rects)) {
+          painter.drawPolygon(outline);
+        }
       }
     }
   }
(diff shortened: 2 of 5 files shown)


maliberty:
I ran into this case where one pin (lower left) appears with two different flight lines

image

I suspect it has something to do with spanning a gcell boundary.

maliberty:
Do you need me to upload the odb?

maliberty:
[gcd.zip](https://github.com/user-attachments/files/21335912/gcd.zip)

maliberty:
Also `_192_` is similar

gadfort:
@maliberty I thought that might happen, it's not clear to me that the correct solution is for the GUI. The database just has a flag that it is connected, but doesn't say to what, so in the code I make a guess that anything that overlaps the guide is connected.
If we only draw one flywire (per terminal, right now it's one flywire per terminal per guide), then the last guide might just look like it's going nowhere.
If we want the GUI to correctly display the connectivity expressed by the guides, then the database would need more information.

maliberty:
If I open the final db with drt results and I turn on 'flywires only' then I see the pre-grt rendering.  Do we need a way to see the grt style or should that control give the grt style?

maliberty:
A minor refinement might be to skip computing/drawing the gcell boundaries when zoomed far out where they don't really show, eg
image

maliberty:
Btw, in your ethmac picture - is that a buffer tree?  I would expect CTS to have broken up that net.

gadfort:
> Btw, in your ethmac picture - is that a buffer tree? I would expect CTS to have broken up that net.

it has one more level, it's just that the macros and stdcells are split at the top, so it ends up looking super lopsided.
image

gadfort:
> If I open the final db with drt results and I turn on 'flywires only' then I see the pre-grt rendering. Do we need a way to see the grt style or should that control give the grt style?

We would probably need some better control to specify the routing data source then, the flywires only is exactly that just the flywires, but if you want to bypass the routing data and see the guide data that would need something else (might be best as a separate issue and PR once this is in).

maliberty:
Which net from the clock tree are you showing?

maliberty:
@osamahammad21 @eder-matheus how is the guide hand-off handling this case?

maliberty (approved):
We can refine it later

PR #7853, opened 2025-07-20
pad: add check for empty call to place_io_terminals

gadfort:
Changes:
- checks if the arguments are empty for `place_io_terminals`

### DIFF
--- a/src/pad/src/pad.tcl
+++ b/src/pad/src/pad.tcl
@@ -349,6 +349,10 @@ proc place_io_terminals { args } {
     keys {} \
     flags {-allow_non_top_layer}
 
+  if { [llength $args] == 0 } {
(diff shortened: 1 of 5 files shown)


src/pad/src/pad.tcl
```diff
@@ -349,6 +349,10 @@ proc place_io_terminals { args } {
     keys {} \
     flags {-allow_non_top_layer}
 
+  if { [llength $args] == 0 } {
+    utl::error PAD 8 "place_io_terminals requires atleast one terminal"
```
maliberty:
```suggestion
    utl::error PAD 8 "place_io_terminals requires at least one terminal"
```

maliberty (approved):

PR #7856, opened 2025-07-21
Select buckets in End point Slack Histogram by x-coordinate

### LINKED ISSUE #7789: It is hard to hit small buckets in End point Slack Histogram
### Description

It is very hard to hit the buckets with few paths in them:

Image

### Suggested Solution

Allow clicking on the entire bucket column, regardless of the height of the bucket.

### Additional Context

_No response_

LucasYuki:
Closes #7789.

Adds clickable and hoverable transparent BarSets stacked on top of the existing ones. This makes it easier to interact with small buckets.

### DIFF
--- a/src/gui/src/chartsWidget.cpp
+++ b/src/gui/src/chartsWidget.cpp
@@ -712,16 +712,22 @@ void HistogramView::setVisualConfig()
     return;
   }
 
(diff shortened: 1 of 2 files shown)


maliberty:
@oharboe would you give this a try?

oharboe (approved):
:relieved:

PR #7859, opened 2025-07-21
gui: expose clock depth to swig interface

gadfort:
Adds:
- clock tree depth to the swig interface, this proved useful when working on the guide routing drawing.

Notes:
- there is no tcl interface to this functionality at this time, if we want to add that we can do it in a separate PR.

### DIFF
--- a/src/gui/src/clockWidget.cpp
+++ b/src/gui/src/clockWidget.cpp
@@ -152,6 +152,10 @@ void ClockTreeRenderer::resetTree()
 
 void ClockTreeRenderer::setMaxColorDepth(int depth)
 {
+  if (max_depth_ < 1) {
+    return;
+  }
+
   max_depth_ = depth;
   redraw();
 }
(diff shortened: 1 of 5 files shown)


src/gui/src/clockWidget.cpp
```diff
@@ -152,6 +152,10 @@ void ClockTreeRenderer::resetTree()
 
 void ClockTreeRenderer::setMaxColorDepth(int depth)
 {
+  if (max_depth_ < 1) {
+    return;
+  }
+
   max_depth_ = depth;
```
maliberty: When would max_depth_ be less than 1?  If that did occur it would be unchangeable.  Did you mean `if (depth < 1)` ?

gadfort: oops, yes, thanks for catching that

maliberty (approved):

PR #7860, opened 2025-07-21
ppl: remove random pin placement functionality

eder-matheus:
Since it is not used in ORFS anymore, the `-random` related code is basically dead code. Removing it to have better readability.


eder-matheus:
@gadfort @donn @QuantamHD Do you guys still using the random pin placement in your flows?

gadfort:
> @gadfort @donn @QuantamHD Do you guys still using the random pin placement in your flows?

@eder-matheus it's still enabled for me. If it's not a lot of effort, it might be nice to still have the two flags (I only use -random) but return immediately with a warning that this is deprecated and does nothing now? Just to provide some transition time.

eder-matheus:
> @eder-matheus it's still enabled for me. If it's not a lot of effort, it might be nice to still have the two flags (I only use -random) but return immediately with a warning that this is deprecated and does nothing now? Just to provide some transition time.

@gadfort  Sure, that should be easy to do. But I believe if your flow may not continue as expected with this. If ppl -random returns doing nothing, it's possible that errors in the following stages will happen, such as routability-driven GPL that requires the pins to be already placed.

QuantamHD:
Doesn't look like I use random

donn:
We allow people to use random but I don't mind getting rid of it

src/ppl/src/IOPlacer.tcl
```diff
@@ -228,146 +228,144 @@ proc place_pins { args } {
 
   sta::check_argc_eq0 "place_pins" $args
 
-  set regions $list(-exclude)
-  set pin_groups $list(-group_pins)
+  if { [info exists flags(-random)] || [info exists keys(-random_seed)] } {
+    utl::warn PPL 113 "-random and -random_seed are deprecated."
+  } else {
```
maliberty: It appears we skip everything if these are seen.  Deprecated usually means still available but not recommended.  It seems like this is more of an error as the code no longer exists.

gadfort: @maliberty an error would be just as interruptive as the full removal of the arguments. I think updating the warning to indicate that nothing is going to happen would be good.

maliberty: @gadfort no pins will be placed so the downstream flow will not have much success.  Would it be better to just do a non-random placement?

gadfort: In the SC flow, this gets called, but because MPL doesn't use it and we call GPL -skip_io before pin placement it didn't matter. I would just prefer this become a no-op so there is some time to transition flows to not include this.

maliberty: ok then let's say "-random and -random_seed are obsolete.  Skipping placement."

PR #7871, opened 2025-07-22
rsz: Caching drive resistance value.

QuantamHD:
Calculating drive resistance takes about 2.8 hours of the currently 30 hour runtime of our block. It's very expensive to calculate so we should cache it.

image

### DIFF
--- a/src/rsz/src/Rebuffer.cc
+++ b/src/rsz/src/Rebuffer.cc
@@ -1394,7 +1394,7 @@ void Rebuffer::init()
     buffer_sizes_.push_back(BufferSize{
         cell,
         FixedDelay(out->intrinsicDelay(sta_)),
-        /*intrinsic_delay=*/0.0f,
+        /*margined_max_cap=*/0.0f,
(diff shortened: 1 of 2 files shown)


maliberty:
@QuantamHD I think that is a good suggestion from @povik

QuantamHD:
Done. Thanks for the suggestion

maliberty:
@povik approve?

src/rsz/src/Rebuffer.cc
```diff
@@ -1394,7 +1394,8 @@ void Rebuffer::init()
     buffer_sizes_.push_back(BufferSize{
         cell,
         FixedDelay(out->intrinsicDelay(sta_)),
-        0.0f,
+        /*intrinsic_delay=*/0.0f,
```
povik: I believe this comment is wrong (this is margined_max_cap not intrinsic_delay)

QuantamHD: Done.

PR #7878, opened 2025-07-23
odb: hash instance to its scan version

AcKoucher:


maliberty:
What happens on dbInst::destroy if there is an associated dbScanInst?

src/odb/src/db/dbBlock.cpp
```diff
@@ -1024,6 +1025,9 @@ dbIStream& operator>>(dbIStream& stream, _dbBlock& block)
   if (db->isSchema(db_schema_bterm_top_layer_grid)) {
     stream >> block._bterm_top_layer_grid;
   }
+  if (db->isSchema(db_schema_block_owns_scan_insts)) {
```
maliberty: Where sis the change to dbDatabase.h to define this?

AcKoucher: This is the schema introduced in #7824 which was just merged. Should I add new one just for the changes here?

maliberty: Any db generated in the window between these PRs would become unloadable so let's make a new rev.

AcKoucher:
> What happens on dbInst::destroy if there is an associated dbScanInst?

I added an error for when that occurs.

PR #7887, opened 2025-07-24
mpl: make conversion methods usable across MPL namespace

AcKoucher:


src/mpl/src/util.h
```diff
@@ -201,4 +201,111 @@ inline double computeDistToNearestRegion(
   return std::sqrt(smallest_distance);
 }
 
+// Here we redefine the Rect class
+// odb::Rect use database unit
+// Rect class use float type for Micron unit
```
rovinski, line 206: Maybe it would be a good idea to make `odb::Rect` templated instead? I think there are other places that use `float` as well. e.g.

```
template <typename T = int>
odb::Rect(
  // ...
  T lx = 0;
  T ly = 0;
  T ux = 0;
  T uy = 0;
```
Then you could do `odb::Rect<float>` and none of the other `int` instances would have to change.

maliberty: I think it is better to not use float.  It is a separate project to remove this class.  I don't want to promote such misusage.

PR #7893, opened 2025-07-25
Support for Rectilinear floorplans in ifp

gs-chaitanya:


src/ifp/src/InitFloorplan.cc
```diff
@@ -60,6 +60,11 @@ using odb::uint;
 
 using upf::eval_upf;
 
+namespace {
+// for the polygoin points vector buffer
+static std::vector<odb::Point> die_polygon_buf_;
+}  // namespace
```
gudeh: We had some issue when setting this as an attribute of `InitFloorplan`, right? Should we eventually do it?

src/ifp/src/InitFloorplan.cc
```diff
@@ -172,6 +182,108 @@ void InitFloorplan::makeDie(const odb::Rect& die)
...
+  for (auto site : additional_sites) {
+    sites_by_name[site->getName()] = site;
+  }
+  addUsedSites(sites_by_name);
+
+  // Remove all existing rows
+  auto rows = block_->getRows();
+  for (auto row_itr = rows.begin(); row_itr != rows.end();) {
+    row_itr = dbRow::destroy(row_itr);
+  }
```
gudeh, line 269: Would it be too hard to check for Rect or Polygon before creating the rows so we don't have to destroy anything?

third-party/abc
eder-matheus, line 1: @gs-chaitanya You probably changed this submodule by accident. Please check the correct version of abc submodule in the OpenROAD master branch and fix it accordingly.

src/ifp/include/ifp/InitFloorplan.hh
```diff
@@ -114,6 +125,7 @@ class InitFloorplan
   odb::dbSite* findSite(const char* site_name);
 
  private:
+  // std::vector<odb::Point> die_polygon_buf_;
```
eder-matheus: If it will not be used, you can remove the commented code.

src/ifp/src/InitFloorplan.cc
```diff
@@ -60,6 +60,11 @@ using odb::uint;
 
 using upf::eval_upf;
 
+namespace {
+// for the polygoin points vector buffer
```
eder-matheus: Typo: polygoin --> polygon?

src/ifp/src/InitFloorplan.cc
```diff
@@ -60,6 +60,11 @@ using odb::uint;
...
+static std::vector<odb::Point> die_polygon_buf_;
```
eder-matheus: Do you still need it? It seems unused, and global vars are not a good thing to be added. Please remove it.

src/ifp/src/InitFloorplan.cc
```diff
@@ -163,6 +168,11 @@ void InitFloorplan::makeDieUtilization(double utilization,
   makeDie({die_lx, die_ly, die_ux, die_uy});
 }
 
+void InitFloorplan::clearPolygonData()
+{
+  die_polygon_buf_.clear();
+}
```
eder-matheus: Same for this function, it doesn't seem that you need it.

src/ifp/src/InitFloorplan.cc
```diff
@@ -952,4 +1064,233 @@ void InitFloorplan::makeTracksNonUniform(odb::dbTechLayer* layer,
...
+      if (site->getHeight() % base_site->getHeight() != 0) {
+        logger_->error(
+            IFP,
+            1001,
+            "Site {} height {}um is not a multiple of site {} height {}um.",
+            site->getName(),
+            block_->dbuToMicrons(site->getHeight()),
+            base_site->getName(),
+            block_->dbuToMicrons(base_site->getHeight()));
+        continue;
```
eder-matheus: You don't need the `continue;` after the Logger::error.

eder-matheus (changes requested):

src/ifp/include/ifp/InitFloorplan.hh
```diff
@@ -97,6 +99,13 @@ class InitFloorplan
                 RowParity row_parity = RowParity::NONE,
                 const std::set<odb::dbSite*>& flipped_sites = {});
 
+  // Create rows for a polygon core area using true polygon-aware generation
+  void makePolygonRows(const std::vector<odb::Point>& core_polygon,
```
gadfort: any reason to not use `odb::Polygon`?

src/ifp/src/InitFloorplan.cc
```diff
@@ -952,4 +1054,232 @@ void InitFloorplan::makeTracksNonUniform(odb::dbTechLayer* layer,
...
+  // Get the bounding box for the polygon (reuse existing polygon if available)
+  odb::Polygon core_poly(core_polygon);
+  odb::Rect core_bbox = core_poly.getEnclosingRect();
+
+  if (base_site->hasRowPattern()) {
+    logger_->error(
+        IFP,
+        1000,
+        "Hybrid rows not yet supported with polygon-aware generation.");
+    return;
```
gadfort: return is not needed, the error will throw an exception

src/ifp/src/InitFloorplan.tcl
```diff
@@ -6,16 +6,22 @@ sta::define_cmd_args "initialize_floorplan" {[-utilization util]\
 					       [-core_space space | {bottom top left right}]\
 					       [-die_area {lx ly ux uy}]\
 					       [-core_area {lx ly ux uy}]\
+                  #  we only plan to use die_polygon and core_polygon combination for maunually specifying the polygon die/core area 
+					       [-die_polygon {x1 y1 x2 y2 ...}]\
+					       [-core_polygon {x1 y1 x2 y2 ...}]\
```
gadfort: would it be possible to use the `die_area` and `core_area`? It seems that anything with more than 2 points would be considered a polygon and two points is just a rectangle

src/ifp/test/init_floorplan_polygon2.ok
```diff
@@ -0,0 +1,4 @@
+[INFO ODB-0227] LEF file: Nangate45/Nangate45.lef, created 22 layers, 27 vias, 135 library cells
+[INFO IFP-0005] Added 6 die polygon vertices to the list.
+[ERROR IFP-0085] no -core_polygon specified for polygon floorplan.
```
gadfort: might be nice to support margin (in the future).

src/ifp/src/InitFloorplan.cc
eder-matheus: I think this was removed, since it's not used anymore.

src/ifp/include/ifp/InitFloorplan.hh
gs-chaitanya: resolved

src/ifp/test/init_floorplan_polygon2.ok
gs-chaitanya: discussed with @eder-matheus , will make sure to include this in the next PR

src/ifp/src/InitFloorplan.tcl
```diff
@@ -352,16 +357,16 @@ proc make_die_helper { key_array } {
 proc make_polygon_die_helper { key_array } {
   array set keys $key_array
 
-  if { ![info exists keys(-die_polygon)] } {
+  if { ![info exists keys(-die_area)] } {
     utl::error IFP 75 "no -die_polygon specified for polygon floorplan."
```
eder-matheus: The -die_polygon option doesn't exists anymore. Please update the error message.

eder-matheus (approved):
The code looks good overall. Just address my last comment before merging it.

PR #7901, opened 2025-07-27
Bazel bump

oharboe:


oharboe:
@maliberty Only modified bazel files, bazel passed. If there is a failure, then that must be a CI snag. We don't lint & error bazel in CI yet.

oharboe:
@maliberty Anything else to do here? This is just a preemptive measure prompted intermittent instability in mirrors where I discovered we could move from a nightly snapshot to RC3. I'll upgrade again when a release is available, there were no changes in generated Verilog for the nightly snapshot and RC3 and I don't expect any actual changes for the release, this is just about using more stable sources for dependencies.

oharboe:
> Do we actually need to be on an RC or would 6.7.0 suffice?

We are using 7 and 6 is nearly in the mirror so I am loath to give 6 much thought.

PR #7904, opened 2025-07-27
pdn: add error checking to connect rules to prevent rules with no actual target

gadfort:
Adds a `-check_only` to run the setup check, but not build the power grid.


maliberty (approved):

oharboe:
@gadfort @maliberty Caught a problem in our PDN, working as intended I'm sure. :relieved:

Issue #7897, opened 2025-07-26
rcx: Crash when generating patterns for RC rules


titan73:
### Describe the bug

The command bench_wires crashes:
```
read_lef technology.lef
bench_wires -len 100 -all
```

### Expected Behavior

No crash

### Environment

```shell
OpenROAD v2.0-23106-g0e46398a83
```

### To Reproduce

Test case file crash_bench_wires.zip uploaded to PII sftp server.

### Relevant log output

```shell
OpenROAD v2.0-23106-g0e46398a83 
Features included (+) or not (-): +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
read_lef
[INFO ODB-0227] LEF file: technology_fixed.lef, created 13 layers, 45 vias
bench_wires
[INFO RCX-0055] Finished 180 bench measurements for pattern MET_OVER_MET
[WARNING ODB-0406] Cannot create net R6_M2oM0_W200W200_S00000S00210_3, duplicate net
Signal 11 received
Stack trace:
 0# 0x00005CE42D8E3732 in ./openroad
 1# 0x000070BDF5042520 in /lib/x86_64-linux-gnu/libc.so.6
 2# odb::dbNet::getBTerms() in ./openroad
 3# odb::dbNet::get1stBTerm() in ./openroad
 4# rcx::extMeasure::createNetSingleWire(char*, unsigned int, unsigned int, unsigned int, int) in ./openroad
 5# rcx::extRCModel::writeBenchWires_DB_res(rcx::extMeasure*) in ./openroad
 6# rcx::extRCModel::benchDB_WS(rcx::extMainOptions*, rcx::extMeasure*) in ./openroad
 7# rcx::extRCModel::linesOverBench(rcx::extMainOptions*) in ./openroad
 8# rcx::extMain::benchWires(rcx::extMainOptions*) in ./openroad
 9# rcx::Ext::bench_wires(BenchWiresOptions const&) in ./openroad
10# 0x00005CE42E099BDE in ./openroad
11# TclNRRunCallbacks in /lib/x86_64-linux-gnu/libtcl8.6.so
12# 0x000070BDF9C54028 in /lib/x86_64-linux-gnu/libtcl8.6.so
13# Tcl_EvalEx in /lib/x86_64-linux-gnu/libtcl8.6.so
[...]

PR #7905, opened 2025-07-27
rcx: Fix double comparison.

titan73:
For https://github.com/The-OpenROAD-Project/OpenROAD/issues/7897.


src/rcx/src/extBenchDB.cpp
```diff
@@ -496,7 +497,7 @@ uint extRCModel::benchDB_WS(extMainOptions* opt, extMeasure* measure)
         double s = pitch * ii;
         double s1 = s - minWidth;
 
-        if (spacing != s1) {
+        if (spacing - s1 > DBL_EPSILON) {
```
maliberty:
```suggestion
        if (std::fabs(spacing - s1) > DBL_EPSILON) {
```

titan73: Yes indeed there was regression in tests. Fixed. Thanks.

maliberty (approved):

PR #7906, opened 2025-07-27
pdn: add support for staggering split cuts

gadfort:
Adds:
- support to stagger splitcuts, currently `GROUND` will be the offset via layer, but if needed in the future we could make that user controllable.


maliberty (approved):
Many thanks!  I've verified it on the test case I sent you.

PR #7907, opened 2025-07-28
rcx: write_rules and ext_patterns.tcl fixes

titan73:


src/rcx/src/OpenRCX.tcl
```diff
@@ -399,7 +399,7 @@ proc write_rules { args } {
   if { [info exists keys(-name)] } {
     set name $keys(-name)
   }
-  rcx::write_rules $filename $dir $name
+  rcx::write_rules $dir/$filename $name
```
maliberty, line 402: `set dir "./"` doesn't need the / as you are adding it here

titan73: Ah indeed. Fixed. Thanks

gadfort: What if user provides a value without the trailing /?

maliberty: A / is being added so wouldn't that be ok?

gadfort: Ahh, I didn't see the fix.

titan73:
The failures seem related to bazel build.

Issue #7858, opened 2025-07-21
Make set_routing_layers command control both grt and drt


titan73:
### Describe the bug

As discussed in https://github.com/The-OpenROAD-Project/OpenROAD/issues/7684.

That would mean the `detailed_route` options `-bottom_routing_layer` and `-top_routing_layer` would be removed.

### Expected Behavior

As requested above.

### Environment

```shell
Latest OR
```

### To Reproduce

Whatever test case.

### Relevant log output

```shell

```

oharboe:
@maliberty Just FYI: Vestiges removed https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3348

PR #7909, opened 2025-07-28
drt: use odb layer range instead of own options to define layers

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/7858.

I've decided to remove the outdated options, but I can revert it and make them deprecated if it's preferable.


maliberty:
@osamahammad21 please handle the review

osamahammad21 (approved):
Looks fine to me. I prefer having them deprecated with an error so that the user knows this is not the way we set the routing layer range anymore and update their script. Just temporarily until all flows are updated. This change should be in a separate commit that could be reverted in the future.

While I was looking at the set_routing_layers command under odb I found that it calls a command in grt. Is there an actual dependency on GRT or could we just make it pure odb?
https://github.com/The-OpenROAD-Project/OpenROAD/blob/a5a209483269c4eac9fd18158fe3892f96756647/src/odb/src/swig/tcl/odb.tcl#L584-L588

eder-matheus:
> While I was looking at the set_routing_layers command under odb I found that it calls a command in grt. Is there an actual dependency on GRT or could we just make it pure odb?

I don't think we have a real dependency with GRT. I can update the command in another PR to make it pure ODB.

osamahammad21:
> > While I was looking at the set_routing_layers command under odb I found that it calls a command in grt. Is there an actual dependency on GRT or could we just make it pure odb?
> 
> I don't think we have a real dependency with GRT. I can update the command in another PR to make it pure ODB.

Great, then we will handle this in a separate PR. I will change my review to "Request changes" until the deprecation is added.

oharboe:
https://github.com/The-OpenROAD-Project/OpenROAD/pull/7921

src/drt/src/TritonRoute.tcl
```diff
@@ -141,17 +139,17 @@ proc detailed_route { args } {
     if { [info exists keys(-remote_host)] } {
       set rhost $keys(-remote_host)
     } else {
-      utl::error DRT 506 "-remote_host is required for distributed routing."
+      utl::warn DRT 506 "-remote_host is required for distributed routing."
```
osamahammad21: Why?

eder-matheus: Bazel build is breaking because of this error, since ORFS still uses these options. I'll turn them back into a error after updating ORFS (https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3362)

oharboe: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/docs/user/Bazel-breaking-changes.md

PR #7913, opened 2025-07-28
gui: add descriptor for dbScanList

AcKoucher:
For #7740.

### DIFF
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -4685,16 +4685,14 @@ bool DbScanListDescriptor::getBBox(std::any object, odb::Rect& bbox) const
   auto scan_list = getObject(object);
   bbox.mergeInit();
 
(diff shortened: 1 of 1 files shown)


src/gui/src/dbDescriptors.cpp
```diff
@@ -4665,6 +4665,85 @@ Descriptor::Property DbScanInstDescriptor::getScanPinProperty(
...
+  bool has_placed_insts = false;
+  for (odb::dbScanInst* scan_inst : scan_list->getScanInsts()) {
+    odb::dbInst* inst = scan_inst->getInst();
+    if (inst->getPlacementStatus().isPlaced()) {
+      has_placed_insts = true;
+      bbox.merge(inst->getBBox()->getBox());
+    }
+  }
+
+  return has_placed_insts;
```
maliberty: No need for `has_placed_insts` just use `!bbox.isInverted()`

PR #7921, opened 2025-07-30
bazel: procedure to handle OpenROAD breaking changes to ORFS

oharboe:


oharboe:
@maliberty OK?

maliberty:
@eder-matheus is this helpful?  I think it might be easier to just make it a warning, update the flow, and then make it an error.

eder-matheus:
> @eder-matheus is this helpful? I think it might be easier to just make it a warning, update the flow, and then make it an error.

Yes, I'll follow your suggestion.

PR #7926, opened 2025-07-30
odb: code generate dbProperty

osamahammad21:


maliberty:
Please sort out "Check That ODB Files Are Generated" before I review.

src/odb/include/odb/dbObject.h
```diff
@@ -90,6 +90,7 @@ enum dbObjectType
   dbPolygonObj,
   dbPowerDomainObj,
   dbPowerSwitchObj,
+  dbPropertyObj,
```
maliberty, line 93: Reordering the values here will change their numeric value.  Usages like _dbAccessPoint::vias_ will need backward compatibility code for a schema rev.

osamahammad21: Nice catch. I am wondering if the actual problem is storing the dbObjectType. Do you know of any other classes that actually store the dbObjectType?

osamahammad21: I am picking up on a few other classes like dbMarker. I can't think of an easy way to handle this to be backward compatible. Also I think this an already underlying issue with code generation. Each time we add a new class, we change the order of the dbObjectTypes starting from the code generated objects and including the lib and tech objects.

osamahammad21:
odb-gen-objtype needs to be merged first.

maliberty (approved):

PR #7927, opened 2025-07-30
GPU-Accelerated Global Routing in TritonRoute-WXL

ZhiangWang033:
This PR is about the GPU-accelerated global routing (GGR-TR) in OpenROAD.   The goal is to replace FastRoute with GGR-TR.  The code is not fully cleaned yet.  I will keep on working on cleaning the code.   It will be done in the next 2-3 weeks.


maliberty:
Do you have an eta for when the remaining work will be done?

PR #7946, opened 2025-08-04
dft: fix null pointer access

donn:
Regression introduced in
c950307c1eeffebde769060e68e54340bca3b96e -- physical-only cells without lib views would cause a null pointer access.


gadfort:
@donn is there a way to add a test to avoid this regression again?

donn:
@gadfort Done.

maliberty:
Please merge head of master to your branch as that should resolve the bazel build error.

maliberty (approved):

maliberty:
nvm, the change is so trivial I'll just merge it and assume it will pass

PR #7948, opened 2025-08-04
dft: save clock domain info in the database

### LINKED ISSUE #7947: dft: scan instance clocks and edges not saved into database
### Description

https://github.com/The-OpenROAD-Project/OpenROAD/blob/3a358b51a46ef79291a771de85221755190a4fbb/src/dft/src/Dft.cpp#L109

The code does not store the clocks and edges for the scan instances, rendering them unretrievable outside of the `dft` module.

### Suggested Solution

They should be implemented.

### Additional Context

Just keeping track of this- might do it myself

donn:
Easier than I expected.

- `dft::Dft::executeDftPlan()`: amend data saved to database to include basic clock domain information - the clock name and the edge, specifically

---

Resolves #7947


src/dft/src/Dft.cpp
```diff
@@ -118,6 +118,24 @@ void Dft::executeDftPlan()
...
+        case ClockEdge::Rising:
+          edge = odb::dbScanInst::ClockEdge::Rising;
+          break;
+        case ClockEdge::Falling:
+          edge = odb::dbScanInst::ClockEdge::Falling;
+          break;
+        default:
+          // TODO: std::unreachable() once we reach c++23
+          // For now, intentionally leave the behavior undefined
+          break;
```
maliberty: Don't add a default case that way the compiler will complain if any value is uncovered.

donn: noted

maliberty:
@AcKoucher please give it a first pass

AcKoucher (approved):
The code LGTM.

@donn just one doubt - w.r.t. std::unreachable how do you envision its usage for this case - as we're switching between `ClockEdge`s shouldn't we be safe?

maliberty:
I wouldn't add a default case regardless of c++ version.  By not having a default the compiler will warn/error if the cases don't cover all the enum values.  Once you have a default you lose that.  

Therefore I would remove the comment but otherwise ok.

donn:
@maliberty done

maliberty (approved):

donn:
@maliberty I think this is a one-off build failure?

PR #7956, opened 2025-08-05
Grt: Fix the resources updates post-DRT

luis201420:
The changes made are:
- Release resources used by nets added to dirty nets during incremental GRT congestion iterations.
- When dirty net congestion cannot be resolved in incremental GRT, find and add to dirty nets the nets that affect congestion areas based on their wires.
- When rerouting a net in GRT, release the correct amount of resources that its wires reduced when loading capacities to FastRouter.


src/grt/include/grt/GlobalRouter.h
```diff
@@ -324,7 +325,11 @@ class GlobalRouter
                                 float reduction_percentage);
   void applyObstructionAdjustment(const odb::Rect& obstruction,
                                   odb::dbTechLayer* tech_layer,
-                                  bool is_macro = false);
+                                  bool is_macro = false,
+                                  bool has_release = false);
```
eder-matheus: `has_release` is a bad name for the flag. Maybe just `release` works better.

src/grt/include/grt/GlobalRouter.h
```diff
@@ -295,6 +295,7 @@ class GlobalRouter
   Net* addNet(odb::dbNet* db_net);
   void removeNet(odb::dbNet* db_net);
 
+  void findNetsWithWiresOnCongestion(std::set<odb::dbNet*>& congestion_nets);
```
eder-matheus: Maybe a shorter name? `getCongestionNets` seems good enough, and you could add a short comment on the function implementation mentioning the wires from detailed routing.

src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -123,18 +123,29 @@ class FastRouteCore
...
+      bool has_release = false);
   void addHorizontalAdjustments(
       const odb::Point& first_tile,
       const odb::Point& last_tile,
       const int layer,
       const interval<int>::type& first_tile_reduce_interval,
-      const interval<int>::type& last_tile_reduce_interval);
+      const interval<int>::type& last_tile_reduce_interval,
+      const std::vector<int>& track_space,
+      bool has_release = false);
```
eder-matheus: Same comment about flag name. Just `release` is better, IMO.

eder-matheus (changes requested):

PR #7960, opened 2025-08-05
ODB: Hash dbObjectType and use that hash in serialization and deserialization.

osamahammad21:


src/odb/src/codeGenerator/gen.py
```diff
@@ -311,6 +312,20 @@ def generate(schema, env, includeDir, srcDir, keep_empty):
...
                     klass["h_includes"].append("odb/db.h")
                 break
+        # Add hash to class
+        hash_dict = {}
+        if "hash" not in klass:
+            hash_value = fnv1a_32(klass["name"])
+        else:
+            hash_value = int(klass["hash"], 16)
+
+        if hash_value in hash_dict:
```
maliberty, line 322: You just did ` hash_dict = {}` so how could this be true?  Will you check against the manual values (range)?

osamahammad21: Nice catch. Should be defined before the for loop of the classes.

maliberty (approved):
Approved conditional on the loop fix

Issue #7959, opened 2025-08-05
Display MPL cluster results


mguthaus:
### Description

Right now, there is no way to determine the quality of the soft-macro placement during MPL. The cells are pre-placed in clusters for GPL, but then before and after GPL there is no way to highlight the cell clusters. 

Doing this with a script to highlight each instance, even using some smart wildcards, results in a TCL script that is too long and cannot finish do to the GUI refresh rate using commands like this:

`select -name "i_frontend/i_fetch_fifo.i_fifo_v3.usage_o\\\[*" -type Inst -highlight 0`

### Suggested Solution

There are a few possible solutions:
1) Allow highlighting in batch from a file with a list of names/regexes.
2) Add properties to the ODB output (but probably make it optional since most people won't use it)

### Additional Context

The goal is to make plots like this from the Hier-RTLMP paper:

<img width="167" height="155" alt="Image" src="https://github.com/user-attachments/assets/1b8105fb-3db5-40c8-8eeb-534c5feb74b4" />

AcKoucher:
@maliberty Do we want this feature on both GUI and WEB?

maliberty:
Web suffices for me.

AcKoucher:
@mguthaus I removed the GUI label. If you want it on the GUI as well, please, just add it back.

mguthaus:
I think any way to view it would be useful. It doesn't have to be on the GUI. I'll have to check out the web interface.

PR #7968, opened 2025-08-06
mpl: add option to store clustering data as dbGroups

AcKoucher:
For #7959.

Next step is to allow browsing groups using the Hierarchy Browse widget in the GUI.

### DIFF
--- a/src/mpl/test/BUILD
+++ b/src/mpl/test/BUILD
@@ -120,6 +120,7 @@ filegroup(
                 "io_constraints6": ["testcases/io_constraints6.def"],
                 "io_constraints7": ["testcases/io_constraints6.def"],
                 "io_constraints8": ["testcases/io_constraints6.def"],
+                "keep_clustering_data": ["testcases/io_constraints6.def"],
             }.get(test_name, []),
         ),
     )


maliberty:
@eder-matheus please review as I'm getting backed up

eder-matheus (commented):
The code looks good, but I would like to see how creating the groups affects the flow after MPL. It's possible that we need to change some other tools and prevent them of looking at this groups created by MPL, since they are just for visualization and debugging.

maliberty:
If necessary we could a distinguishing value to dbGroupType

AcKoucher:
I added a new value to dbGroupType and I'm working to make the placer understand that it should ignore it.

AcKoucher:
I updated this branch locally with master now that both #7992 and #7986 we're merged and I still get different placement results when running the flow. It looks like DPL is also affected by the new group, so apparently there's more work needed to allow the changes here.

maliberty:
Yes we support groups in dpl/dpo too

AcKoucher:
@maliberty @eder-matheus I ran a secure-CI enforcing the new flag and I got the same results as running without it, so this should be good to merge.

eder-matheus:
@AcKoucher I believe your unit test is failing in the bazel build due to a missing dependency on the input files. I had something similar in a recent PR, see how I fixed it here: https://github.com/The-OpenROAD-Project/OpenROAD/pull/8554/files#diff-85df7b235e804a7f2ad2d172187227f957651f66e912f40a0f41fc521c86ba53R332-R337

AcKoucher:
Thanks @eder-matheus.

There were two issues. The first one was the actual test result, the other was a missing bazel resource as you pointed out. I missed that there are multiple tests using the same resource and that's being addressed later in the file:

https://github.com/The-OpenROAD-Project/OpenROAD/blob/9a0e3dfa90189bf51eac4ba6ca243f478653678f/src/mpl/test/BUILD#L112-L126

The last commit should handle it.

AcKoucher:
@maliberty The check failure looks unrelated to the changes here:

```
docker: Error response from daemon: failed to create task for container:
Unavailable: error reading from server: EOF: unavailable.
[...]
```

maliberty:
@mguthaus FYI

PR #7972, opened 2025-08-07
CTS: Add inverters in register tree

arthurjolo:
Currently in CTS, inverters are part of the macro tree. Results, however results are beter when they are in the register tree.

eg.: sky130hd/uw:
master:
master_uw_cts_jtag_tck

inverters in register tree:
my_uw_cts_jtag_tck

### DIFF
--- a/src/cts/src/TritonCTS.cpp
+++ b/src/cts/src/TritonCTS.cpp
@@ -1239,8 +1239,12 @@ bool TritonCTS::separateMacroRegSinks(
 
     if (iterm->isInputSignal() && inst->isPlaced()) {
       odb::dbMTerm* mterm = iterm->getMTerm();
+      sta::Cell* masterCell = network_->dbToSta(mterm->getMaster());
+      sta::LibertyCell* libertyCell = network_->libertyCell(masterCell);
       // Treat clock gaters like macro sink
-      if (hasInsertionDelay(inst, mterm) || !isSink(iterm) || inst->isBlock()) {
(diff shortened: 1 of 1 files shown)


maliberty:
You might want to classify the inverter by what it drives (macro vs flop)

arthurjolo:
Yes, that seems to be the best way. I will work on the changes.

rovinski:
Maybe tangential, but it seems like the first buffer has 1+ ns of delay on it? Doesn't that imply that the slew is pretty bad in either case?

arthurjolo:
> Maybe tangential, but it seems like the first buffer has 1+ ns of delay on it? Doesn't that imply that the slew is pretty bad in either case?

Are you referring to the triangle in red? The red triangle represents the input pin and the green lines represent wire delay, so the 1ns delay is wire delay to get to the first buffers.

arthurjolo:
FYI, I created a new PR with @maliberty comment (https://github.com/The-OpenROAD-Project/OpenROAD/pull/7995). So I wll be closing this one.

rovinski:
Yes, 1 ns from the input pin to first buffer is generally far too long. Either the input drive is not properly constrained, or CTS should be adding buffers closer to the input pin.

arthurjolo:
Oh ok, sorry I understand buffer delay as the internal delay. We do have a fix for long wires that break these wires from the input pin to the first buffers. I can check why it isn't doing anything for this case.

PR #7974, opened 2025-08-07
gui: fix the charts view hold with path_group filter

LucasYuki:
The Charts View was always displaying the setup slack histogram when filtering by path_group.

Changes:
- Sends the correct argument to findPathGroup
  - sta::MinMax::max() when displaying setup slack
  - sta::MinMax::min() when displaying hold slack


AcKoucher (approved):
Nice catch!

maliberty (approved):

PR #7976, opened 2025-08-08
pdn: dont add last strap if it is not inside grid region

gadfort:
Changes:
- does not add the last strap if it does not fit inside the grid area, before if the strap was more than 50% over the area it would be accepted.


gadfort:
@maliberty this should resolve the issue you mentioned, but it might be worth doing a larger check to ensure this doesn't cause any other strange behavior.

maliberty:
secure CI started

src/pdn/src/straps.cpp
```diff
@@ -293,8 +293,8 @@ void Straps::makeStraps(int x_start,
         // no portion of the strap is inside the limit
         return;
       }
-      if (group_pos > pos_end) {
-        // strap center is outside of alotted area
+      if (strap_end > pos_end) {
+        // strap end is outside of alotted area
```
maliberty, line 293: Is the prior check
```
if (strap_start >= pos_end) {
        // no portion of the strap is inside the limit
        return;
      }
```
now redundant?  I assume this check would always fail in that case

maliberty:
Would it make sense to ensure the strap is not covering any signal pins that could be access with a via?  In my case this would suffice but I can imagine another case where the strap is just inside but still covering the pins.

gadfort:
> Would it make sense to ensure the strap is not covering any signal pins that could be access with a via? In my case this would suffice but I can imagine another case where the strap is just inside but still covering the pins.

I can think of atleast one technology where that would lead to missing power straps over the access circuits. As a future improvement I think it could be added as an option to the macro grid

src/pdn/src/straps.cpp
gadfort: you are correct

maliberty:
I see some unable to repair channels on a private PDK so I'll have to investigate a bit...

gadfort:
> I see some unable to repair channels on a private PDK so I'll have to investigate a bit...

Since this only impacts the last strap, it would likely be in regions with little room next to a macro where before the strap was placed by was partially outside the core area.

PR #7980, opened 2025-08-08
odb: Fixing lef parsing error message

QuantamHD:
Fix` ERROR: Attempt to call configuration function 'lefrUnsetReadFunction' in LEF parser before lefrInit() call in session-based mode`. from lef parser when parsing multiple LEF files.

This was a fix to an interal binary that uses the OpenROAD C++ API.

### DIFF
--- a/src/odb/src/lef/lef/lefrReader.cpp
+++ b/src/odb/src/lef/lef/lefrReader.cpp
@@ -1579,6 +1579,7 @@ void lefrUnsetReadFunction()
 {
   LEF_INIT;
   lefSettings->ReadFunction = nullptr;
+  init_call_func = nullptr;
 }
 
 // Set the maximum number of warnings


maliberty:
How does clearing init_call_func fix that?

QuantamHD:
@haojcx Would you mind adding some additional context?

haojcx:
`ERROR: Attempt to call configuration function '%s' in LEF parser before `lefrInit()` call in session-based mode.\n"` was logged from `lefrInitSession` if `init_call_func != nullptr`.

This happens when the multiple LEF files are given. After the first file was read, `init_call_func` remains none empty even after `lefGZipClose()` and `lefrClear()`.

Functionally, I don't see this affecting anything except the error message.

maliberty:
Are you trying to read more than one LEF in one session or are you creating a new session?  If the later it would make more sense to put this in lefrClear which is the session terminator.

PR #7986, opened 2025-08-10
gpl: include the consideration of debug group,

gudeh:
when this group is present, instances will be placed together with the set of instances that belong to no group, associated with no region.

I tested with an example provided by @AcKoucher at branch `mpl-keep-clustering`, and it seems to work as intended:
placement(3)(1)

without the changes in the current PR gpl was setting all instances participating in the debug group as not movable.

I triggered a secure-CI at: [secure-gpl-consider-debug-group](https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-consider-debug-group/1/pipeline)

### DIFF
--- a/src/gpl/src/placerBase.cpp
+++ b/src/gpl/src/placerBase.cpp
@@ -1015,7 +1015,9 @@ void PlacerBase::init()
       continue;
     }
 
-    if (inst->dbInst() && inst->dbInst()->getGroup() != group_) {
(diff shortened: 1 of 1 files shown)


eder-matheus:
@maliberty The only failure in this PR is the bazel download that happened in other PRs. This should be safe to merge.

PR #7992, opened 2025-08-11
odb: add dbGroupType for visual debug

AcKoucher:
So that we can merge #7986 in order to allow the changes in #7968.


AcKoucher:
The build is failing due to new warnings generated when SWIG parsing. I'm taking a look.

PR #7995, opened 2025-08-11
Cts classify inverters

arthurjolo:
Currently in CTS, inverters are always part of the macro tree, even if the inverter is driving a register.

This PR makes CTS classify the inverters by what they drive. If the inverter drives a macro put it in the macro tree if it drives a register put it in the register tree.

Also a new unity test is added to verify the correct behavior (`inverters`):

image

This can improve QoR for designs with inverters that are currently classified in the wrong tree.
eg.: sky130hd/uw jtag_tck:
master:
master_uw_cts_jtag_tck

this PR:

cts_jtag_tck

### DIFF
--- a/src/cts/src/TritonCTS.cpp
+++ b/src/cts/src/TritonCTS.cpp
@@ -1245,7 +1245,7 @@ bool TritonCTS::separateMacroRegSinks(
       bool nonSinkMacro = !isSink(iterm);
       sta::Cell* masterCell = network_->dbToSta(mterm->getMaster());
       sta::LibertyCell* libertyCell = network_->libertyCell(masterCell);
-      if (libertyCell->isInverter()) {
+      if (libertyCell && libertyCell->isInverter()) {
         odb::dbITerm* invertedTerm
             = inst->getFirstOutput()->getNet()->get1stSignalInput(false);
         nonSinkMacro &= invertedTerm->getInst()->isBlock();


maliberty:
secure ci?

arthurjolo:
Secure CI is fine for all designs except asap7/cva6, where I get a timeout (which is expected due to the problems with nightly). But there is one metric change needed I am going to create a PR for it.

src/cts/src/TritonCTS.cpp
```diff
@@ -1238,9 +1238,20 @@ bool TritonCTS::separateMacroRegSinks(
...
-      // Treat clock gaters like macro sink
-      if (hasInsertionDelay(inst, mterm) || !isSink(iterm) || inst->isBlock()) {
+
+      bool nonSinkMacro = !isSink(iterm);
+      sta::Cell* masterCell = network_->dbToSta(mterm->getMaster());
+      sta::LibertyCell* libertyCell = network_->libertyCell(masterCell);
+      if (libertyCell && libertyCell->isInverter()) {
+        odb::dbITerm* invertedTerm
+            = inst->getFirstOutput()->getNet()->get1stSignalInput(false);
+        nonSinkMacro &= invertedTerm->getInst()->isBlock();
```
precisionmoon, line 1251: What happens when there are multiple inverters before a sink?   Do we need to look beyond just one inverter?

maliberty: It would be unusual to see multiple inverters in the pre-cts netlist as synthesis would tend to remove pairs of inverters.

arthurjolo: This is a tricky case, but it would end up on the register tree, as the inverter is driving another inverter (that is not a macro)

Issue #2904, opened 2023-02-15
Add respect ordering constraint to set_io_pin_constraint


oharboe:
### Description

Add option to set_pin_constraint so that it is possible to constrain to an edge as well as deciding the ordering of the pins.

### Suggested Solution

Add ```-order``` option that orders the pins in increasing x/z ordering as the names were specified to the ```-pin_names``` argument

```
set_io_pin_constraint -order
```

https://github.com/The-OpenROAD-Project/OpenROAD/blob/42cea5fc6fc5244bbaf93d591fa7c3ab162cbc59/src/ppl/src/IOPlacer.tcl#L121

eder-matheus:
@oharboe Do you want the pins adjacent to each other? Or can we have other pins between two pins of the list?

oharboe:
> @oharboe Do you want the pins adjacent to each other? Or can we have other pins between two pins of the list?

I thought there was a "group" constraint that could express this?

I can easily imagine someone wanting control over this, such as when having a bus and other unconstrained pins. In this case thet would want the bus pins not intermingled with other pins.

eder-matheus:
@oharboe Yes, we have the -group option where you can define a set of pins to be placed together, following the ordering of the names. In this case, they will be placed following the min distance between each pin. I'm updating this option to have the pins ordered in ascending x/y position.

You can see an example of how to use the `-group_pins` option here: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/ppl/test/group_pins6.tcl.
 
You can constrain the whole group to an edge by only constraining one pin of the group, so your script doesn't need to be as verbose as this example.

oharboe:
> @oharboe Yes, we have the -group option where you can define a set of pins to be placed together, following the ordering of the names. In this case, they will be placed following the min distance between each pin. I'm updating this option to have the pins ordered in ascending x/y position.
> 
> You can see an example of how to use the `-group_pins` option here: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/ppl/test/group_pins6.tcl.
> 
> You can constrain the whole group to an edge by only constraining one pin of the group, so your script doesn't need to be as verbose as this example.

@maliberty Makes sense to me. The only concern I have is that this breaks the API, something that as OpenROAD becomes more mature should never do without careful consideration. E.g. if this has an.impact on QoR, good or bad, an upgrade can become less trivial.

[...]

PR #7997, opened 2025-08-11
Fix issue with `openroad_deps_prefixes.txt` permissions & add case for AlmaLinux 9

kcaisley:
Hi there, thanks for this awesome project. Just managed to build from source. :)

This PR addresses two issues found when trying to build OpenROAD via the OpenROAD flow scripts repo on AlmaLinux 9. (binary-compatible with RHEL9 & RockyLinux 9). AlmaLinux 9 gets lots of use at universities here in Germany, and is used now by scientific institutions like CERN, so I guess it's worth adding this.

## Fix 1
The first issues, was basically that a case for AlmaLinux when checking the `os-release` variable didn't exist in the `etc/DependencyInstaller.sh`. So I simply added a match:

```bash
# etc/DependencyInstaller.sh
elif [[ "${os}" == "AlmaLinux" ]]; then
    rhelVersion=$(rpm -q --queryformat '%{VERSION}' almalinux-release | cut -d. -f1)
```

## Fix 2
The second change I'm a bit more unsure of, but essentially the docs recommend the higher-level flow scripts repo `setup.sh` script be run with `sudo`, but this causes the `openroad_deps_prefixes.txt` to be owned by `root`. During setup, within the main `OpenROAD` submodule this causes the error:

```bash
line 954: $HOME/OpenROAD-flow-scripts/tools/OpenROAD/etc/openroad_deps_prefixes.txt: Permission denied
```

I tested the flow script `setup.sh` build on AlmaLinux 9, Ubuntu 24.04, and Ubuntu 20.04 and this issue appears to exist regardless of platform.

Would hopefully resolves several related issues:
[#2904](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2904) - Permission denied error on Ubuntu 24 during install
[#3137](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3137) - Permission denied accessing openroad_deps_prefixes.txt
[#2505](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2505) - Permission denied error when running setup.sh (was inside Docker container)

My fix is to essentially `chmod` the file permissions if the user is running as `root`:

```bash
# etc/DependencyInstaller.sh

 1020    if [[ ! -z ${saveDepsPrefixes} ]]; then
 1021        mkdir -p "$(dirname $saveDepsPrefixes)"
 1022        echo "$CMAKE_PACKAGE_ROOT_ARGS" > $saveDepsPrefixes
 1023 +      # Fix permissions if running as root to allow user access
 1024 +      if [[ $(id -u) == 0 && ! -z "$SUDO_USER" ]]; then
 1025 +          chown "$SUDO_USER:$(id -gn "$SUDO_USER")" "$saveDepsPrefixes" 2>/dev/null || true
 1026 +      fi
 1027    fi

```
If this is a bit too much of a "hack", I'm open to feedback and can modify my pull request. Just hoping to get the build working in upstream so others in my group can more easily install OpenROAD. :+1:

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -1021,7 +1021,7 @@ if [[ ! -z ${saveDepsPrefixes} ]]; then
     mkdir -p "$(dirname $saveDepsPrefixes)"
     echo "$CMAKE_PACKAGE_ROOT_ARGS" > $saveDepsPrefixes
     # Fix permissions if running as root to allow user access
-    if [[ $(id -u) == 0 && ! -z "$SUDO_USER" ]]; then
+    if [[ $(id -u) == 0 && ! -z "${SUDO_USER+x}" ]]; then
         chown "$SUDO_USER:$(id -gn "$SUDO_USER")" "$saveDepsPrefixes" 2>/dev/null || true
     fi
 fi
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -933,11 +933,13 @@ case "${os}" in
             _installOrTools "ubuntu" "${ubuntuVersion}" "amd64"
         fi
         ;;
-    "Red Hat Enterprise Linux" | "Rocky Linux")
+    "Red Hat Enterprise Linux" | "Rocky Linux" | "AlmaLinux")
     if [[ "${os}" == "Red Hat Enterprise Linux" ]]; then
         rhelVersion=$(rpm -q --queryformat '%{VERSION}' redhat-release | cut -d. -f1)
     elif  [[ "${os}" == "Rocky Linux" ]]; then
         rhelVersion=$(rpm -q --queryformat '%{VERSION}' rocky-release | cut -d. -f1)
+    elif [[ "${os}" == "AlmaLinux" ]]; then
+        rhelVersion=$(rpm -q --queryformat '%{VERSION}' almalinux-release | cut -d. -f1)
     fi
         if [[ "${rhelVersion}" != "8" ]] && [[ "${rhelVersion}" != "9" ]]; then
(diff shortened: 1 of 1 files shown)


maliberty:
Thanks for the PR. 

- DCO needs to be satisfied by signing your commits with `git commit -s`.  You can fix existing commits are described in the details link
- This is still a draft so I'm expecting it is not ready for a review

etc/DependencyInstaller.sh
```diff
@@ -1018,4 +1020,8 @@ esac
 if [[ ! -z ${saveDepsPrefixes} ]]; then
     mkdir -p "$(dirname $saveDepsPrefixes)"
     echo "$CMAKE_PACKAGE_ROOT_ARGS" > $saveDepsPrefixes
+    # Fix permissions if running as root to allow user access
+    if [[ $(id -u) == 0 && ! -z "$SUDO_USER" ]]; then
```
vvbandeira: All variables must be set to avoid undefined behavior. The following should make the CI pass the Docker build stage.
```suggestion
    if [[ $(id -u) == 0 && ! -z "${SUDO_USER+x}" ]]; then
```

kcaisley: Okay, thanks for letting me know. I think I understand that this would potentially be an issue in a case where SUDO_USER is set, but to an empty string.

kcaisley:
I've added a partner pull request in the ORFS repo to support installs on AlmaLinux 9 (actually EL8 / EL9 in general, I suppose).

The-OpenROAD-Project/OpenROAD-flow-scripts#3408

etc/DependencyInstaller.sh
kcaisley: Ah, and as I just realized: also when the variable is completely unset.

PR #7999, opened 2025-08-11
cts: multiple strategies to apply NDR to clock nets

jfgava:
# Add flexible NDR application strategies for clock tree levels in TritonCTS

## Summary

This PR introduces multiple Non-Default Rule (NDR) application strategies for clock nets in TritonCTS, providing fine-grained control over which clock tree levels receive NDR treatment. Previously, NDR was applied uniformly to all non-leaf clock nets. This enhancement allows users to target specific levels, ranges, or portions of the clock tree hierarchy.

### New Functions Added

- **`extractClockTreeLevelFromNetName()`**: Extracts clock tree level from net names using regex pattern matching (e.g., "clknet_2_clk_i" → level 2)
- **`getAllClockTreeLevels()`**: Discovers all unique clock tree levels present in the design
- **`applyNDRToClockLevels()`**: Applies NDR to a specified list of clock tree levels
- **`applyNDRToClockLevelRange()`**: Applies NDR to a contiguous range of levels (min to max)
- **`applyNDRToFirstHalfLevels()`**: Applies NDR to the first half of clock tree levels (typically higher levels in the hierarchy)

### Modified Functions

- **`writeClockNDRsToDb()`**: Updated to demonstrate multiple NDR application strategies with commented options

## NDR Application Strategies

The implementation provides four distinct strategies:

1. **Specific Levels**: Target individual levels (e.g., only level 0)
2. **Level Range**: Apply to a continuous range (e.g., levels 0-3)
3. **First Half**: Apply to the upper half of the clock tree hierarchy
4. **All Non-Leaf**: Original behavior - apply to all non-leaf clock nets

## Technical Details

- Uses regex pattern `_([0-9]+)_` to extract level numbers from clock net names
- Maintains backward compatibility with existing NDR behavior
- Includes comprehensive debug logging for NDR application tracking
- Handles edge cases where no level pattern is found (defaults to level 0)

## Future Work

- Integrate with command-line options for strategy selection
- Create new regression test cases
- Add user-configurable parameters to select NDR strategy at runtime


src/cts/src/TritonCTS.cpp
```diff
@@ -1538,6 +1538,120 @@ void TritonCTS::writeClockNetsToDb(TreeBuilder* builder,
...
+int TritonCTS::applyNDRToClockLevels(odb::dbBlock* block_,
+                                     odb::dbTechNonDefaultRule* clockNDR,
+                                     const std::set<odb::dbNet*>& clkLeafNets,
+                                     const std::vector<int>& targetLevels)
+{
+  int ndrAppliedNets = 0;
+
+  // clang-format off
+  debugPrint(logger_, CTS, "clustering", 1, "Applying NDR to clock tree levels: ");
+  // clang-format on
```
maliberty: No need for the format on/off as it is just a single line.

src/cts/src/TritonCTS.cpp
```diff
@@ -1538,6 +1538,120 @@ void TritonCTS::writeClockNetsToDb(TreeBuilder* builder,
...
+int TritonCTS::applyNDRToFirstHalfLevels(
+    odb::dbBlock* block_,
+    odb::dbTechNonDefaultRule* clockNDR,
+    const std::set<odb::dbNet*>& clkLeafNets)
+{
+  // Get all unique levels in the design
+  std::vector<int> allLevels = getAllClockTreeLevels(block_, clkLeafNets);
+
+  // Calculate first half (rounding up if odd number of levels)
+  size_t halfCount = (allLevels.size() + 1) / 2;
```
maliberty: const

src/cts/src/TritonCTS.cpp
```diff
@@ -1538,6 +1538,120 @@ void TritonCTS::writeClockNetsToDb(TreeBuilder* builder,
...
+  for (int level : targetLevels) {
+    debugPrint(logger_, CTS, "clustering", 1, "{} ", level);
+  }
+
+  // Single pass: check clock nets and apply NDR if level matches
+  for (odb::dbNet* net : block_->getNets()) {
+    if (net->getSigType() == odb::dbSigType::CLOCK
+        && (clkLeafNets.find(net) == clkLeafNets.end())) {
+      std::string netName = net->getConstName();
+      int level = extractClockTreeLevelFromNetName(netName);
```
maliberty: const

maliberty:
@arthurjolo is there a place to apply NDRs that doesn't need the regex and can know the level more structurally?

arthurjolo:
> @arthurjolo is there a place to apply NDRs that doesn't need the regex and can know the level more structurally?

So far, there is not any information of the clock net level. I think there is an easy way to add this information to CTS clock nets structures, this way we can avoid using regex.

jfgava:
These last modifications fixes #8026

jfgava:
ndr-applied

@maliberty I modified the code to not use regex. It is working fine.
Do you think it's ok to include a TCL command line option to select different strategies to apply NDR to clock nets?
I did it to test here, it's already working. Something like:
clock_tree_synthesis -apply_ndr -ndr_strategy level0 ...
The options could be "level0", "half", "full" (default). Or maybe numbers? 0, 1, 2. And explain what it means in the README

maliberty:
I think it is ok to make a user level options.  Please use names and not numbers.

precisionmoon:
> I think it is ok to make a user level options. Please use names and not numbers.

Perhaps we can combine -apply_ndr and -ndr_strategy options into one option:  -apply_ndr none | root_only | half | full
The default can be root_only or whichever option provides the best QoR out of the box (OOTB).

jfgava:
> > I think it is ok to make a user level options. Please use names and not numbers.
> 
> Perhaps we can combine -apply_ndr and -ndr_strategy options into one option: -apply_ndr none | root_only | half | full The default can be root_only or whichever option provides the best QoR out of the box (OOTB).

Sounds good. Thanks Cho

src/cts/README.md
```diff
@@ -92,7 +92,7 @@ clock_tree_synthesis
 | `-num_static_layers` | Set the number of static layers. The default value is `0`, and the allowed values are integers `[0, MAX_INT]`. |
 | `-sink_clustering_buffer` | Set the sink clustering buffer(s) to be used. |
 | `-obstruction_aware` | Enables obstruction-aware buffering such that clock buffers are not placed on top of blockages or hard macros. This option may reduce legalizer displacement, leading to better latency, skew or timing QoR.  The default value is `False`, and the allowed values are bool. |
-| `-apply_ndr` | Applies 2X spacing non-default rule to all clock nets except leaf-level nets. The default value is `False`. |
+| `-apply_ndr` | Applies 2X spacing non-default rule to clock nets except leaf-level nets following some strategy. There are four strategy options: `none, root_only, half, full`. The default value is `root_only`. |
```
maliberty: I assume the previous default would have been none as this was a boolean switch.  Are you confident we should change the default (secure CI is happy)?

src/cts/src/CtsOptions.cpp
```diff
@@ -5,6 +5,31 @@
 
 namespace cts {
 
+void CtsOptions::setApplyNDR(const std::string& strategy)
```
maliberty: It would be better to have the c++ code use an enum and have the string conversion handled in the swig layer.  See for example dft.i.

maliberty (changes requested):

src/cts/README.md
jfgava: Yes, I think I need to be clearer in the description.
The default is actually "none" if the apply_ndr arg is not used. If it is used with an incorrect option, then it uses root_only as the default.

maliberty: If the arg is used then a value should be mandatory.

src/cts/src/CtsOptions.cpp
jfgava: Done. I didn't know we could handle this inside the swig layer. Thanks for pointing out

src/cts/README.md
jfgava: Yes. Now it is working as expected. The value is mandatory.
And if the -apply_ndr is not used, "none" is the default.

maliberty (approved):

PR #8002, opened 2025-08-12
ODB: codeGenerate dbDatabase and dbChip

osamahammad21:


src/odb/src/db/dbDatabase.cpp
```diff
@@ -192,125 +137,231 @@ _dbDatabase::_dbDatabase(_dbDatabase* /* unused: db */, int id)
...
-  stream << db._schema_minor;
-  stream << db._master_id;
-  stream << db._chip;
-  stream << *db._tech_tbl;
-  stream << *db._lib_tbl;
-  stream << *db._chip_tbl;
-  stream << *db._gds_lib_tbl;
-  stream << NamedTable("prop_tbl", db._prop_tbl);
-  stream << *db._name_cache;
-  stream << *db._gds_lib_tbl;
```
osamahammad21, line 223: @maliberty  I just noticed that we out stream the _gds_lib_tbl twice. is that intentional?

maliberty: No

src/odb/src/codeGenerator/schema/chip/dbChip.json
```diff
@@ -0,0 +1,159 @@
...
+      "name": "thickness_",
+      "type": "int",
+      "default": "0",
+      "schema": "db_schema_chip_extended"
+    },
+    {
+      "name": "shrink_",
+      "type": "float",
+      "default": "0.0",
+      "comment": "// range ]0, 1]",
```
rovinski, line 62: `[0, 1]` ?

maliberty: Some people use this as an alternative to `(0, 1]` which is how I took it (a shrink of 0 would be useless).

maliberty: See https://en.wikipedia.org/wiki/Interval_(mathematics) "Notations for intervals"

rovinski: Interesting, I've never seen that before. I think `(0, 1]` is much more common.

maliberty: I'm ok to accept a PR changing it

rovinski: Sure #8018

PR #8003, opened 2025-08-12
ODB: database prop

osamahammad21:
- dbDatabase <-> dbProperty relation codeGeneration
- better plural name in codeGenerator


osamahammad21:
after https://github.com/The-OpenROAD-Project/OpenROAD/pull/8002

osamahammad21:
after https://github.com/The-OpenROAD-Project/OpenROAD/pull/8011

src/odb/src/codeGenerator/helper.py
```diff
@@ -99,6 +99,14 @@ def is_bit_fields(field, structs):
             return True
     return False
 
+def get_plural_name(name):
```
maliberty, line 103: I remember this same issue in another db from a past life.  English is weird enough that there is a library just for this purpose `inflect`.  This will probably hold us for a while.

maliberty (approved):

src/odb/src/codeGenerator/helper.py
osamahammad21: Yes I came across that library but I didn't want to add a new python library to the build.

PR #8009, opened 2025-08-12
gui: pass object not Selected object to getProperties

gadfort:
Fixes the `std::bad_cast` which was attempting to convert a Selected to a dbInst

FYI: @gudeh


maliberty (approved):
Thanks!

PR #8011, opened 2025-08-13
ODB: chip table hash and add a test case

osamahammad21:


src/odb/include/odb/db.h
```diff
@@ -6921,7 +6921,7 @@ class dbChip : public dbObject
     HIER
   };
 
-  std::string getName() const;
+  const char* getName() const;
```
QuantamHD, line 6938: Why are we downgrading this from string to const char?

osamahammad21: the dbHashTable is built using const char. I need to change the name variable from string to const char so that it could be used in that table.

QuantamHD: Can we use string view in the hash table instead?

osamahammad21: Yeah sure. I'll handle that in a separate PR.

src/odb/test/cpp/TestChips.cpp
maliberty, line 1: Add to BUILD

maliberty (changes requested):

src/odb/include/odb/db.h
maliberty: I don't think swig likes string_view but if it is entirely internal that should be ok.

src/odb/test/cpp/TestChips.cpp
osamahammad21: For some reason, the BUILD has only three tests. They seem to be specially handled in the CMakeLists.txt and the BUILD. In the CMakeLists, they are grouped under OdbGTests. Is it okay if I added the TestChips to the BUILD?

src/odb/include/odb/db.h
osamahammad21: I don't think the hashtable is exposed. It is only accessed in the find* APIs so it should be easy to manage internally. I too don't like the c strings.

src/odb/test/cpp/TestChips.cpp
maliberty: Looks like we need to add the others to BUILD but the new test is enough for this PR

maliberty (approved):

maliberty:
You need to fix the bazel error before merging

osamahammad21:
> You need to fix the bazel error before merging

I think I understand the reason behind not adding the other tests to the BUILD. Bazel doesn't work with the boost tests. That's the reason behind the failures in the CI of this PR. Since TestChips.cpp uses boost test, not gtest, could we handle adding it to BUILD in a separate PR that is focused on including all cpp tests under bazel?

PR #8016, opened 2025-08-13
ODB: add dbChipInst

osamahammad21:
after https://github.com/The-OpenROAD-Project/OpenROAD/pull/8003 and https://github.com/The-OpenROAD-Project/OpenROAD/pull/8011


osamahammad21:
@maliberty FYI

src/odb/include/odb/geom.h
```diff
@@ -57,13 +57,46 @@ class Point
...
 
- private:
+ protected:
   int x_ = 0;
   int y_ = 0;
 };
 
 std::ostream& operator<<(std::ostream& os, const Point& pIn);
 
+class Point3D : public Point
```
maliberty: I'm no sure this is a good idea.  Do the rotate* or *Distance methods make sense on a 3D point?

src/odb/src/db/dbChipInst.cpp
```diff
@@ -0,0 +1,221 @@
...
+  obj->orient_ = orient;
+}
+
+dbOrientType dbChipInst::getOrient() const
+{
+  _dbChipInst* obj = (_dbChipInst*) this;
+  return obj->orient_;
+}
+
+dbTransform dbChipInst::getTransform() const
```
maliberty, line 141: The current dbTransform doesn't handle the extra orientations in Z

src/odb/include/odb/geom.h
osamahammad21: I think the rotate methods can still make sense if we assume they operate around a fixed z-axis.

As for the distance methods, I agree they don’t extend to 3D points. We could handle this either by overloading them specifically for Point3D, or by flipping the hierarchy by making Point3D the base class and letting Point inherit from it with z = 0. We could also make Point3D separate, but I fear that this may cause us issues in the future.

src/odb/src/db/dbChipInst.cpp
osamahammad21: Not yet. Should be handled in a separate PR.

src/odb/include/odb/geom.h
osamahammad21: I don’t think flipping the hierarchy makes sense. We could just give Point an extra z coordinate instead and use that directly for 2d and 3d points.

maliberty: My concerns is that a Point3D isn't a Point2D which inheritance usually implies.  There is the possibility of slicing as well.  What are we gaining by inheritance?  You could use composition or just make them independent classes.

osamahammad21: I think making them independent works as well. I was just considering cases where using a 3D point as 2D might be convenient if the z coordinate doesn’t matter. However, this could become misleading for developers later, similar to what happened in DPL with GridX and DbuX

maliberty (approved):

osamahammad21:
@maliberty @vvbandeira @sombraSoft The CI failure seems to be unrelated. Could we set it to pass and merge the PR?

PR #8025, opened 2025-08-14
est: fix coverity issue

eder-matheus:
Coverity complains about the possibility of having a `throw std::runtime_error` inside the `IncrementalParasiticsGuard` destructor. It happens because the destructor calls the `EstimateParasitics::updateParasitics` function, which can fall into a case where a Logger::error would be generated.

In practice, is very unlikely that it would happen, but Coverity complains about it anyway. My approach to fix it is to add a `IncrementalParasiticsGuard::end` function, which needs to be explicitly called at the end of the escope of which use of this class.

### DIFF
--- a/src/cts/src/TritonCTS.cpp
+++ b/src/cts/src/TritonCTS.cpp
@@ -2223,6 +2223,8 @@ void TritonCTS::balanceMacroRegisterLatencies()
       }
     }
   }
+
+  parasitics_guard.end();
 }
 
 float TritonCTS::getVertexClkArrival(sta::Vertex* sinkVertex,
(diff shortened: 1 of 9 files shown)


maliberty:
In what case would an exception be thrown?  This weakens the RAII aspect of the guard as end won't be called if an exception is propagated out.

eder-matheus:
> In what case would an exception be thrown? This weakens the RAII aspect of the guard as end won't be called if an exception is propagated out.

One of the cases that Coverity is complaining is not possible, since it requires a specific flag to be false, which will never be when using the incremental guard. The other case is similar, and I believe it's just Coverity being overly pessimistic. I can mark them as false positives and close this PR.

maliberty:
Let's do that

PR #8028, opened 2025-08-14
mpl: adjust regression test to new IO Pads handling approach

AcKoucher:
IO Pad clusters don't depend on the data flow to exist since #6894.

### DIFF
--- a/src/mpl/test/io_pads1.ok
+++ b/src/mpl/test/io_pads1.ok
@@ -19,4 +19,5 @@ Die Area: (0.00, 0.00) (290.00, 125.00),  Floorplan Area: (0.00, 0.00) (149.91, 
 	Floorplan Utilization: 0.21
 	Manufacturing Grid: 10
 
+[WARNING MPL-0014] No Liberty data found for std cells. Continuing without dataflow.
 No differences found.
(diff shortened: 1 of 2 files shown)


src/mpl/test/io_pads1.ok
```diff
@@ -19,4 +19,5 @@ Die Area: (0.00, 0.00) (290.00, 125.00),  Floorplan Area: (0.00, 0.00) (149.91,
 	Floorplan Utilization: 0.21
 	Manufacturing Grid: 10
 
+[WARNING MPL-0014] No Liberty data found for std cells. Continuing without dataflow.
```
eder-matheus, line 22: Even with this warning message, the results are the same. Why running without dataflow didn't affect the final result?

maliberty: @AcKoucher was just telling me earlier that he found a major bug in data flow so possibly that's the cause?

AcKoucher: That's actually not the cause. The thing is that this test case literally does not depend on the dataflow. Before #6894, however, we would create the Cluster objects for Pads **from** the dataflow structure, as this is not done anymore, we don't need to include the libs in this test (and as there's no liberty data, there's no dataflow).

I.e., the result shouldn't change.

eder-matheus: Thanks for explaining!

AcKoucher: Np!

PR #8029, opened 2025-08-14
mpl: ensure valid virtual connections when breaking mixed leaves

AcKoucher:
Bug from the original implementation - the virtual connections literally don't exist nowadays.
I expect QoR to improve as now the placer will be biased to place together clusters from a mixed leaf split.

Creating as a draft as it will need to be rebased after #8028.


AcKoucher:
Running Secure-CI.

AcKoucher:
Secure-CI passed and there are many winners w.r.t. QoR in both public and private.

eder-matheus (approved):

PR #8039, opened 2025-08-15
Improve initial placement solver convergence and exit criteria

calewis:
This commit introduces several improvements to the initial placement solver to enhance convergence, runtime, and the reliability of the exit condition.
- **Refactor Method Name:** Renamed the primary method to `doPlace` to decouple it from a specific implementation. This will avoid a name mismatch if swapping in other solvers (e.g., `MINRES`) in the future.
- **Improve Solver Convergence:** Switched the `BiCGSTAB` solver to use its default diagonal preconditioner. This offers better convergence with negligible performance overhead.
- **Refine Exit Condition:** The solver now terminates based on the stabilization of HPWL. This replaces the previous ad-hoc method of using the solver residual with a fixed minimum of 5 iterations.
- **Reduce Time in Solver:** The maximum number of solver iterations has been reduced. This is justified by the faster convergence from the new preconditioner and because the previous high iteration count was often adding noise to the solution rather than improving it.

### DIFF
--- a/src/gpl/include/gpl/Replace.h
+++ b/src/gpl/include/gpl/Replace.h
@@ -65,6 +65,8 @@ class Replace
 
   // Initial Place param settings
   void setInitialPlaceMaxIter(int iter);
+  void setInitialResidualThreshold(double threshold);
+  void setInitialHpwlPctChangeThreshold(double threshold);
   void setInitialPlaceMinDiffLength(int length);
   void setInitialPlaceMaxSolverIter(int iter);
   void setInitialPlaceMaxFanout(int fanout);
@@ -142,6 +144,8 @@ class Replace
   std::unique_ptr<NesterovPlace> np_;
 
   int initialPlaceMaxIter_ = 20;
+  double initialResidualThreshold_ = 1.0e-5;
+  double initialPlaceHpwlPctChangeThreshold_ = 1.0;
   int initialPlaceMinDiffLength_ = 1500;
   int initialPlaceMaxSolverIter_ = 100;
   int initialPlaceMaxFanout_ = 200;
--- a/src/gpl/src/solver.h
+++ b/src/gpl/src/solver.h
@@ -3,11 +3,9 @@
 
 #pragma once
 
-#include <Eigen/IterativeLinearSolvers>
+#include <Eigen/Core>
 #include <Eigen/SparseCore>
-#include <memory>
 
-#include "graphics.h"
 #include "odb/db.h"
 #include "placerBase.h"
 #include "utl/Logger.h"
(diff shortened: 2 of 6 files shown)


calewis:
I'll fix the diff failures in the tests next week.

maliberty:
DCO is required (use `git commit -s` to sign your commits).  The details link gives steps to fix it.

src/gpl/src/solver.h
```diff
@@ -24,20 +22,13 @@ struct ResidualError
...
-                             int iter,
-                             SMatrix& placeInstForceMatrixX,
-                             Eigen::VectorXf& fixedInstForceVecX,
-                             Eigen::VectorXf& instLocVecX,
-                             SMatrix& placeInstForceMatrixY,
-                             Eigen::VectorXf& fixedInstForceVecY,
-                             Eigen::VectorXf& instLocVecY,
-                             utl::Logger* logger,
-                             int threads);
+ResidualError cpuSparseSolve(
```
maliberty: clang-format your changes.

src/gpl/include/gpl/Replace.h
```diff
@@ -65,6 +65,8 @@ class Replace
 
   // Initial Place param settings
   void setInitialPlaceMaxIter(int iter);
+  void setInitialResidualThreshold(double threshold);
+  void setInitialHpwlPctChangeThreshold(double threshold);
```
maliberty: These are not exposed to the user through tcl or python.  What is their use model?

maliberty:
I think it reasonble to move on to the unit tests.  Once we see good results there we can run all designs.

Note we have a few large tests that are not part of the check-in test but can be run with ./src/gpl/test/regression-large.  It would be good to test/update them as well.

src/gpl/include/gpl/Replace.h
calewis: I was just trying to mirror the other existing functions.  I can make them available to TCL/python.

maliberty: My guess is this is too low level a detail for users to care about or know how to set effectively.  I suggest we can pick reasonable defaults until something demonstrates the need for user control.

src/gpl/test/large01.ok
```diff
@@ -26,26 +26,26 @@
...
-[InitialPlace]  Iter: 12 conjugate gradient residual: 0.00041858 HPWL: 2501437898
-[InitialPlace]  Iter: 13 conjugate gradient residual: 0.00050194 HPWL: 2483385672
-[InitialPlace]  Iter: 14 conjugate gradient residual: 0.00012450 HPWL: 2481000217
-[InitialPlace]  Iter: 15 conjugate gradient residual: 0.00016931 HPWL: 2467165861
-[InitialPlace]  Iter: 16 conjugate gradient residual: 0.00010803 HPWL: 2468526865
-[InitialPlace]  Iter: 17 conjugate gradient residual: 0.00017008 HPWL: 2465897237
-[InitialPlace]  Iter: 18 conjugate gradient residual: 0.00010513 HPWL: 2468180915
-[InitialPlace]  Iter: 19 conjugate gradient residual: 0.00017248 HPWL: 2465550210
-[InitialPlace]  Iter: 20 conjugate gradient residual: 0.00010280 HPWL: 2467735216
+[InitialPlace]  Iter: 1 residual: 0.03180663 HPWL: 4066226359, abs percent change: 75.19178500424849
```
maliberty, line 29: Too many digits.   I think this would be better recast as a table like we do in nestrov place with similar formatting precision.

```
Iteration | Residual |     HPWL (um) |  HPWL(%) 
---------------------------------------------------
....
```

PR #8046, opened 2025-08-18
odb: dbChipRegion, dbChipRegionInst and dbChipConn

osamahammad21:


src/odb/src/codeGenerator/schema/chip/dbChip.json
```diff
@@ -148,6 +148,12 @@
       "type": "dbId<_dbChipInst>",
       "flags": ["private"],
       "schema":"db_schema_chip_inst"
+    },
+    {
+      "name": "conns_",
```
maliberty, line 153: Is the owning chip a 'hier' type?  Is there always a root chip?

osamahammad21: yes and yes

osamahammad21: I could move the Conn table into chip if that's the follow up question. I do think it makes more sense to do so.

maliberty: It isn't clear that 3dblox requires a top chip - is that something we will need to invent and avoid writing back out?

I don't know how many conns per chip there will be.  The tables allocate in pages so there is the potential for waste if we push it down (if there are many it is less likely to matter).

osamahammad21: The design, identified by the design name, is going to be the top chip of type HIER. It won’t receive any special treatment during write-back if you think of it this way:
- The user specifies which chip to write, usually the top chip.
- That chip corresponds to the design, identified by the design name, in the 3dblox files.

Since 3dblox allows encapsulating the entire design as a hierarchical chiplet to be instantiated in a larger design, it’s natural for the design itself to be represented as a chiplet.

maliberty: It allows it but does not require it.  If there was no hier chiplet then we would have created it and the write out would have something the user didn't input.

Issue #8036, opened 2025-08-15
Placement check failed after antenna repair and filler placement


povik:
### Describe the bug

See the log below taken from the OR flow test aes_sky130hs.

### Expected Behavior

Test completes, possibly errors out on metrics since there are resizer changes

### Environment

```shell
.
```

### To Reproduce

This is on commit https://github.com/The-OpenROAD-Project/OpenROAD/pull/8024/commits/8a8e760c69cadb0d1ffbebb9189420cfc3564add. CI artifact can be found at https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-8024-head/2/artifact/aes_sky130hs.tgz.

### Relevant log output

```shell
################################################################
# Repair antennas post-DRT
set repair_antennas_iters 0
utl::set_metrics_stage "drt__repair_antennas__pre_repair__{}"
while { [check_antennas] && $repair_antennas_iters < 5 } {
  utl::set_metrics_stage "drt__repair_antennas__iter_${repair_antennas_iters}__{}"
  repair_antennas
  detailed_route -output_drc [make_result_file "${design}_${platform}_ant_fix_drc.rpt"] \
    -output_maze [make_result_file "${design}_${platform}_ant_fix_maze.log"] \
    -verbose 0
  incr repair_antennas_iters
}
[INFO ANT-0002] Found 55 net violations.
[INFO ANT-0001] Found 55 pin violations.
[INFO GRT-0020] Min routing layer: met1
[INFO GRT-0021] Max routing layer: met5
[INFO GRT-0022] Global adjustment: 0%
[INFO GRT-0023] Grid origin: (0, 0)
[INFO GRT-0088] Layer li1     Track-Pitch = 0.4800  line-2-Via Pitch: 0.3400
[INFO GRT-0088] Layer met1    Track-Pitch = 0.3700  line-2-Via Pitch: 0.3400
[INFO GRT-0088] Layer met2    Track-Pitch = 0.4800  line-2-Via Pitch: 0.3500
[INFO GRT-0088] Layer met3    Track-Pitch = 0.7400  line-2-Via Pitch: 0.6150
[INFO GRT-0088] Layer met4    Track-Pitch = 0.9600  line-2-Via Pitch: 1.0400
[INFO GRT-0088] Layer met5    Track-Pitch = 3.3300  line-2-Via Pitch: 3.1100
[INFO GRT-0003] Macros: 0
[INFO GRT-0004] Blockages: 1549
[...]

PR #8053, opened 2025-08-18
dpl: avoid padding violation on cells adjacent to the core edge

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/8036


src/dpl/test/mirror2.ok
```diff
@@ -8,12 +8,12 @@
...
-average displacement        1.4 u
-max displacement            2.9 u
-original HPWL               0.9 u
+total displacement          2.7 u
+average displacement        1.3 u
+max displacement            2.7 u
+original HPWL               0.7 u
 legalized HPWL              2.1 u
-delta HPWL                  136 %
+delta HPWL                  199 %
```
maliberty, line 16: It seems odd that HPWL went up - particularly in a mirror test that shouldn't care about this change.

maliberty: nvm, I see it is a delta between two quantities that both reduced.

src/dpl/src/PlacementDRC.cpp
```diff
@@ -302,7 +302,9 @@ bool PlacementDRC::checkPadding(const Node* cell,
     for (GridY grid_y{y}; grid_y < cell_y_end; grid_y++) {
       const Pixel* pixel = grid_->gridPixel(grid_x, grid_y);
       if (pixel == nullptr) {  // at the core edge
-        if (!cell->isFixed()) {
+        const bool left_edge = x == 0;
+        const bool right_edge = cell_x_end == grid_->getRowSiteCount();
+        if (!cell->isFixed() && !left_edge && !right_edge) {
           // Padding violation with the core edge.
           return false;
         }
```
osamahammad21, line 310: When would `pixel` ever be null without being at the core edge? I remember spending time on this case when I first introduced the new padding handling in DPL, and I didn’t want to alter the existing results. That’s why I added this check to return false if there isn’t enough padding between the cell and the core edge. If that’s no longer necessary, then just change all of this to:
```
if (pixel == nullptr)
{
  continue;
}
```

osamahammad21 (commented):
@eder-matheus sorry for the late review.

src/dpl/src/PlacementDRC.cpp
osamahammad21: Also, if the required behavior is to ignore padding violations to the core edge altogether then this misses the following case:
- CellA: (1, y_begin) (x_end, y_end)
- CellA Padding is 2+ sites

With the current implementation, this would produce a padding violation with the core edge because CellA is not at the left core edge exactly.

eder-matheus: I could imagine a scenario where an instance has 1-site distance fro the core edge, and the padding is 2. If I understand correctly, the `pixel` would still be `nullptr`, and a potential error would occur, where we'd have a 1-site gap between the cell and the core edge. That's why I decide to ensure the instance is adjacent to the edge.

While writing it, I realized I should have added a test for this scenario, at least to confirm that this is what happens. I'll try it later.

PR #8054, opened 2025-08-18
ODB: dbChipBump, dbChipBumpInst and dbChipNet

osamahammad21:


stefanottili:
Could you please elaborate what these dbChipXX data structures are needed for ?
Why would the hierarchical def "designs with placed pins and nets" model not be sufficient ?

maliberty:
They are for multi-chip systems (think heterogenous 3DIC).

stefanottili:
﻿> They are for multi-chip systems (think heterogenous 3DIC).

well, I guessed right, but that didn’t quite answer my question 

>> Why would the hierarchical def "designs with placed pins and nets" model not be sufficient ?

Even a 3d chip is „just“ a design with components, pins, nets, (thru) vias.
An interposed is just a piece of silicon after all.
Thru via are also just via, no ?

The missing piece might be the missing z of the component placement.

maliberty:
Chip can also be flipped in Z so DEF transforms are insufficient.  A lot of additional information is also needed (eg thickness) that doesn't make sense to pay for everywhere.

It is certainly possible to build a system entirely from uniform netlist representation and we did have that discussion.  However forcing every part of the system to be aware of that possibility felt like more overhead than have a set of distinct objects for what likely to be separate algorithms.  Placement across dies is quite different than within a die for example (3d vs 2d planar).

A format like 3dblox also has more abstract representations than you see in DEF (think DSO/prototyping).

maliberty (approved):

PR #8060, opened 2025-08-19
Add clock filtering to the Charts View

### LINKED ISSUE #8057: gui: add clock filter in Charts view
### Description

In the Charts view, we display all of the paths regardless of clock in the histogram. Enhance the filtering to allow the user to specify which clock(s) they want to include in the histogram. Then, use the clock selection to tie into the click action so that the Timing Report that comes up is filtered by the clock selection. So, if they are viewing the histogram for a specific clock and click on one of the bars, only those paths are displayed in the Timing Report.

### Suggested Solution

_No response_

### Additional Context

_No response_

LucasYuki:
Closes #8057

Changes:
- Add clock filtering to the Charts View
- Now, if a mode was previously selected and then the Select Mode is selected again, the charts hide the other buttons
- Unused attribute of ChartsWidget removed (prev_filter_index_)
- Unconstrained pins count moved, so the mode selection button doesn't change position

### DIFF
--- a/src/gui/src/chartsWidget.cpp
+++ b/src/gui/src/chartsWidget.cpp
@@ -182,26 +182,30 @@ ChartsWidget::ChartsWidget(QWidget* parent)
   setObjectName("charts_widget");  // for settings
   chart_tabs_->setTabBarAutoHide(true);
 
-  QHBoxLayout* controls_layout = new QHBoxLayout;
-  controls_layout->addWidget(mode_menu_);
-  controls_layout->addWidget(label_);
+  QHBoxLayout* controls_layout_top = new QHBoxLayout;
+  controls_layout_top->addWidget(mode_menu_);
   setModeMenu();
-  controls_layout->addWidget(path_group_menu_);
+  controls_layout_top->addWidget(path_group_menu_);
   path_group_menu_->hide();
-  controls_layout->addWidget(clock_menu_);
+  controls_layout_top->addWidget(clock_menu_);
   clock_menu_->hide();
-  controls_layout->addWidget(refresh_filters_button_);
+  controls_layout_top->addWidget(refresh_filters_button_);
   refresh_filters_button_->hide();
-  controls_layout->insertStretch(2);
+  controls_layout_top->insertStretch(1);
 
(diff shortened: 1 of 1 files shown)


maliberty:
> Now, if a mode was previously selected and then the Select Mode is selected again, the charts hide the other buttons

I don't see the value of the "Select Mode" choice.  It seems simpler to just have setup & hold with setup being the default.

maliberty:
The minimum width of the chart is now rather wide due to all the text across the top (ethmac)
image

I think it would help to move the unconstrained pin text to the bottom
image

LucasYuki:
> I don't see the value of the "Select Mode" choice. It seems simpler to just have setup & hold with setup being the default.

I think it is better to load the charts view only if it's requested.
With the setup or hold as default, if the charts view is open in the GUI, the chart is going to load automatically even if the user is not going to use, increasing the GUI startup time.

maliberty:
The "Timing Report" requires you to click update to populate it so the same could be applied here.

maliberty:
Note that ORFS does 
```
    log_cmd gui::select_chart "Endpoint Slack"
    log_cmd gui::update_timing_report
```
automatically on the gui_* targets with GUI_TIMING=1 so we are already paying the cost

LucasYuki:
> I don't see the value of the "Select Mode" choice. It seems simpler to just have setup & hold with setup being the default.

I think it is better to load the charts view only if it's requested.
With the setup or hold as default, if the charts view is open in the GUI, the chart will be loaded automatically even if the user does not use it, increasing the GUI startup time.

LucasYuki:
Looking into the Charts View code, I notice that the charts were implemented as tabs, but there is only one tab now (everything is inside the "Slack" tab), so I want to remove this part of the code to reduce the size of the frame.

Can I make another PR to change the layout and keep this PR with only the clock filtering?

In another PR, I can remove the tabs, remove the "Select Mode" choice, and change the update button behavior so it only updates the charts if the button is clicked.

maliberty:
We can defer work to another PR.  My thought on tabs was that we may have other charts over time and tabs would then be useful.

maliberty (approved):

PR #8067, opened 2025-08-19
gpl: fix hpwl print

gudeh:
We are currently printing the progression of the HPWL during GPL execution in DBU values. This is incorrect, the values should be shown in microns (µm).


maliberty (approved):

rovinski:
I just always assumed this was on purpose :)

Issue #6033, opened 2024-10-28
Support partial automatic macro placement


oharboe:
### Description

untar  [partial-automatic-macro-placement.tar.gz](https://github.com/user-attachments/files/17539427/partial-automatic-macro-placement.tar.gz)

### Suggested Solution

Support partial automatic macro placement

### Additional Context

If support is added, I'll update https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/2516

AcKoucher:
@oharboe I had a discussion with @maliberty  and, before actually digging into the implementation of partial placement, our idea is to allow the already existing mpl2 guidance mechanism to be used. We support it internally, but there's no .tcl API to actually used it.

Basically we'd let the user specify a certain region in which the macro must be placed in order to guide the automated placement. I think that can help solving the issue.

Having a mechanism capable of incorporating the data of a placed macro into the floor-plan packing during SA would be challenging to implement. It'd be a bigger project.

oharboe:
> @oharboe I had a discussion with @maliberty  and, before actually digging into the implementation of partial placement, our idea is to allow the already existing mpl2 guidance mechanism to be used. We support it internally, but there's no .tcl API to actually used it.
> 
> Basically we'd let the user specify a certain region in which the macro must be placed in order to guide the automated placement. I think that can help solving the issue.
> 
> Having a mechanism capable of incorporating the data of a placed macro into the floor-plan packing during SA would be challenging to implement. It'd be a bigger project.

Specifying regions sounds like a better feature. For megaboom, I could specify that integer register file with pins on horizontal edge is placed near the center top of the design and that the floating point register file with pins on vertical edge is placed near middle of the left edge, for instance.

However, how small can these regions be?  A single location? What should the user expect if he chooses a region that is effectively a single location? Just curious if this is supported and how it is handled.

AcKoucher:
> However, how small can these regions be? A single location? What should the user expect if he chooses a region that is effectively a single location? Just curious if this is supported and how it is handled.

We'll try to align the center of the macro with the center of the specified region. With regards to how small it could be, I don't see a reason to have a lower limit. However, I think that using a single location would be less practical - and a bit odd - than just specifying a rectangle that covers a somewhat interesting region for the macro. That's just my opinion though.

Note: I'm interpreting a single location as a region made of a single point:
( x1 , y1 ), ( x2 , y2 ) 
where x1 = x2 and y1 = y2.

[...]

PR #8076, opened 2025-08-20
mpl: remove unneeded reset of macros' positions in SA

AcKoucher:
I'm not sure about the rationale behind this reset, but it doesn't seem to make sense as the new positions don't rely on resetting the old ones.

This change will be useful in the implementation for #6033.


AcKoucher:
Running Secure-CI.

AcKoucher:
Secure-CI is fine.

PR #8077, opened 2025-08-20
Use -corner for inner corner cell if -edge_corner not specified

jeffng-or:


jeffng-or:
> I think it would be wise to add a test case to this

Just pushed up the addition of boundary_macros_separate2.tcl, which is the same as boundary_macros_separate.tcl + removal of -edge_corner argument. The end result is the same.

src/tap/test/CMakeLists.txt
maliberty, line 1: Please add to BUILD as well

jeffng-or: done!

jeffng-or:
@maliberty, it looks like the bazel tests failed on my new test. Do I need to do anything special/different to enable it?

maliberty:
Looks like you are using boundary_macros_separate.defok from boundary_macros_separate2.  Usually we only allow tests to access their own files.  I just made a symlink to resolve that.

PR #8083, opened 2025-08-21
Gui: Add clock filter in the Timing Report

### LINKED ISSUE #8058: gui: add clock filter in the Timing Report
### Description

In the Timing Report, enable clock selection in case the user isn't using the Charts as an entry point into the Timing Report

### Suggested Solution

_No response_

### Additional Context

_No response_

LucasYuki:
Closes #8058

Changes:
- Add clock filter in the Timing Report
- Add new class (DropdownCheckboxes) so it's possible to select multiple items.


src/gui/src/staGui.cpp
```diff
@@ -1321,4 +1342,12 @@ std::vector<std::set<const sta::Pin*>> TimingControlsDialog::getThruPins() const
   }
   return pins;
 }
-}  // namespace gui
+
+void TimingControlsDialog::getClocks(sta::ClockSet* clock_set) const
+{
+  for (const auto& clk : clock_box_->selectedItems()) {
+    clock_set->insert(clks_.at(clk));
+  }
```
AcKoucher: The name `clks_` for the map make this operation a bit confusing. `name_to_clk_` or `qstring_to_clk_` is clearer.
Also, here you're iterating clock names, not clock pointers, so `clk_name` instead of `clk` makes it clearer as well.

src/gui/src/staGuiInterface.cpp
```diff
@@ -1026,14 +1026,15 @@ std::unique_ptr<TimingPathNode> STAGuiInterface::getTimingNode(
 
 TimingPathList STAGuiInterface::getTimingPaths(const sta::Pin* thru) const
 {
-  return getTimingPaths({}, {{thru}}, {}, "" /* path group name */);
+  return getTimingPaths({}, {{thru}}, {}, "" /* path group name */, nullptr);
```
AcKoucher: Please, add a comment like the one for the path group name.

src/gui/src/qComboCheckBoxes.h
```diff
@@ -0,0 +1,57 @@
...
+class Clock;
+}  // namespace sta
+
+namespace gui {
+class qComboCheckBoxes : public QComboBox
+{
+  Q_OBJECT
+ public:
+  qComboCheckBoxes(const char* name,
+                   const char* all_text = "All",
```
AcKoucher: Use `std::string` instead of `const char*` for all the string manipulation unless the latter is actually required.

src/gui/src/staGui.cpp
```diff
@@ -1106,6 +1111,7 @@ TimingControlsDialog::TimingControlsDialog(QWidget* parent)
       layout_(new QFormLayout),
       path_count_spin_box_(new QSpinBox(this)),
       corner_box_(new QComboBox(this)),
+      clock_box_(new qComboCheckBoxes("Select Clocks", "All Clocks", this)),
```
AcKoucher: I think we don't need the "Select Clocks" option.

LucasYuki: I put the "Select Clocks" as an inactive item because I'm not changing the selected item (since the user can choose none or multiple items), so it always displays this text when the QComboBox is closed.

If the "Select Clocks" is removed, what do you suggest showing when the QComboBox is closed?

AcKoucher: I don't think that being able to choose none would be very useful. My idea was to have a similar behavior to the Corner selection. Does that make sense?

AcKoucher (approved):

PR #8091, opened 2025-08-21
Resynthesis: Fix handling of memories and properly honor set_dont_touch and set_dont_use attributes

TheMightyDuckOfDoom:
This fixes the following issues:

- Memories: IsCombinational didn't check for memories, but memories can not be used for resynthesis
- set_dont_use: Resynthesis did not check if cells are marked as dont-use
- set_dont_touch: Resynthesis did not check if instances/pins/nets are marked as dont-toch

Each feature has its own testcase.

### DIFF
--- a/src/rmp/src/CMakeLists.txt
+++ b/src/rmp/src/CMakeLists.txt
@@ -66,6 +66,7 @@ target_link_libraries(rmp_abc_library
     OpenSTA
     dbSta_lib
     utl_lib
+    rsz_lib
     ${ABC_LIBRARY}
 )
 
--- a/src/rmp/src/zero_slack_strategy.cpp
+++ b/src/rmp/src/zero_slack_strategy.cpp
@@ -24,15 +24,24 @@
 
 namespace rmp {
 
-std::vector<sta::Vertex*> GetNegativeEndpoints(sta::dbSta* sta)
+std::vector<sta::Vertex*> GetNegativeEndpoints(sta::dbSta* sta,
+                                               rsz::Resizer* resizer)
 {
   std::vector<sta::Vertex*> result;
 
   sta::dbNetwork* network = sta->getDbNetwork();
   for (sta::Vertex* vertex : *sta->endpoints()) {
-    sta::PortDirection* direction = network->direction(vertex->pin());
+    sta::Pin* pin = vertex->pin();
+    sta::PortDirection* direction = network->direction(pin);
     if (!direction->isInput()) {
       continue;
+    }
+
+    if (resizer != nullptr) {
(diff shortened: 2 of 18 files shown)


maliberty:
Thanks for the contribution and the unit tests.  There are CI issues that need to be addressed along with format/lint in order to be mergeable.

src/rmp/src/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ target_link_libraries(rmp_abc_library
     OpenSTA
     dbSta_lib
     utl_lib
+    rsz
```
QuantamHD: You should depend on rsz_lib instead

src/rmp/src/zero_slack_strategy.cpp
```diff
@@ -24,17 +24,26 @@
...
   for (sta::Vertex* vertex : *sta->endpoints()) {
-    sta::PortDirection* direction = network->direction(vertex->pin());
+    sta::Pin* pin = vertex->pin();
+    sta::PortDirection* direction = network->direction(pin);
     if (!direction->isInput()) {
       continue;
     }
 
+    if (resizer != nullptr) {
+      if (resizer->dontTouch(pin) || resizer->dontTouch(network->net(pin))
```
QuantamHD, line 41: @maliberty should this functionality be moved to odb instead? Seems like purely database related functionality

QuantamHD (commented):
Thanks for the contribution! Do you have any additional plans or are you just fixing some bugs?

TheMightyDuckOfDoom:
@QuantamHD I tried to use resynthesis with [https://github.com/pulp-platform/croc](https://github.com/pulp-platform/croc) and ran into these issues.

It gives nice QoR improvements, so I want to upstream the fixes.

If the pull-request passes review and CI, then it would be ready for merging.

src/rmp/src/CMakeLists.txt
TheMightyDuckOfDoom: Done!

src/rmp/src/zero_slack_strategy.cpp
maliberty: As it operating in terms of sta objects it would make more sense to move to dbSta.

maliberty: I think this is a separate concern so we can merge this and address afterwards.  @TheMightyDuckOfDoom are you interested in doing this change?

TheMightyDuckOfDoom: I'll create an issue for further discussion, as the Resizer keeps track of set_dont_touch and set_dont_use attributes sperately from OpenSTA.

Issue #8090, opened 2025-08-21
GRT: Global routes not connected to pin access point location


osamahammad21:
### Describe the bug

In asap7/aes, pin _24562_/B is not covered by the global route guides

<img width="1656" height="1073" alt="Image" src="https://github.com/user-attachments/assets/5f31356c-2189-46a6-8ce3-d1d571b58aaf" />
Showing just M1 guides:

<img width="1656" height="1073" alt="Image" src="https://github.com/user-attachments/assets/5c4e8838-419f-45c3-b82f-11bec7b2941f" />

### Expected Behavior

GRT uses the correct access point 3d location in routing.(considering the layer of the access point as well)

### Environment

```shell
NA
```

### To Reproduce

https://drive.google.com/file/d/1W1MVNXMT9DWHmkTfFKY5a2zZ941Qgsm4/view?usp=sharing
./run-me.....
the scripts/global_route.tcl is edited to show gui after the first run of global_route to avoid incremental global route changes to PA.

### Relevant log output

```shell

```

PR #8092, opened 2025-08-22
grt: fix pin connection layer

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/8090

Before this fix, GRT was always using the upper layer of a pin (BTerm or ITerm) as its connection layer, without considering if the access point used as the position on grid is on this specific layer.

This PR modifies how it work by looking at the layer of the access point choosed to compute the position on grid, and setting it as the connection layer. This way, GRT will always produce guides that covers the pin AP.


eder-matheus:
Started a secure-ci for precaution, but this should be safe.

@osamahammad21 FYI.

maliberty:
Please give a brief description of this fix

eder-matheus:
> Please give a brief description of this fix

I've updated the description.

maliberty (approved):

PR #8097, opened 2025-08-22
gpl: reduce timing driven iterations from 5 to 2

gudeh:
This way we do one timing driven iteration before and another after routability mode.

We get this behavior:
| Overflow | Event                     |
|----------|----------------------------|
| 64       | TD iteration, non-virtual |
| 60       | Save routability snapshot |
| 30       | Revert routability        |
| 20       | TD iteration, non-virtual |
| 10       | Finish nesterov |

Experiments showed reductions in gpl runtime compared to master.
image

### DIFF
--- a/src/gpl/src/replace.tcl
+++ b/src/gpl/src/replace.tcl
@@ -101,7 +101,7 @@ proc global_placement { args } {
 
     if { [info exists keys(-timing_driven_net_reweight_overflow)] } {
       set overflow_list $keys(-timing_driven_net_reweight_overflow)
-    } else {    
+    } else {
       set overflow_list [list 64 20]
     }
 


gudeh:
Stable results for wirenlength, final TNS and final max frequency:
image

gudeh:
Secure-CI is currently passing for all public designs (with few metrics updates), although there is a fatal happening in a private design: `[INFO DPL-0034] Detailed placement failed on the following 1 instances:`

src/gpl/src/replace.tcl
```diff
@@ -102,7 +102,8 @@ proc global_placement { args } {
     if { [info exists keys(-timing_driven_net_reweight_overflow)] } {
       set overflow_list $keys(-timing_driven_net_reweight_overflow)
     } else {
-      set overflow_list [list 79 64 29 21 15]
+      #previous values: [list 79 64 29 21 15]
```
eder-matheus: Remove this commented code. If we need this information in the future, it's better to get it from git history.

PR #8099, opened 2025-08-22
mpl: support fixed macros in floorplan packing

AcKoucher:
This approach currently does NOT work.
I'm opening this draft mainly for discussion.


AcKoucher:
@mikesinouye Would you have some test case for the changes here?

@oharboe @gadfort @QuantamHD I think this may be of your interest.

mikesinouye:
Thank you for all of your work to support this! I have a couple of internal test cases for this. I have cherry picked this locally and will report back with the results. They're both larger than what's in the public CI so it will likely be tomorrow.

AcKoucher:
@mikesinouye Nice!

AcKoucher:
Running Secure-CI.

mikesinouye:
This seems to work as expected with our internal test cases. Thanks again!

AcKoucher:
I got a problem in a private CI due to the way I was handling some special placed cells. The last commit fixes it.
Re-running Secure-CI.
This should be ready for review.

PR #8118, opened 2025-08-26
mpl: group soft annealing weights in a struct

AcKoucher:
No changes in functionality.

The idea is to make the code less bug prone for adding/removing penalties.

### DIFF
--- a/src/mpl/src/hier_rtlmp.h
+++ b/src/mpl/src/hier_rtlmp.h
@@ -269,8 +269,8 @@ class HierRTLMP
   float notch_v_th_ = 10.0;
   float notch_h_th_ = 10.0;
 
-  // For cluster and macro placement.
-  SACoreWeights placement_core_weights_;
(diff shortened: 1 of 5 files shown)


src/mpl/src/hier_rtlmp.h
```diff
@@ -279,11 +279,6 @@ class HierRTLMP
                                             0.0f /* guidance */,
                                             0.0f /* fence */};
 
-  // Soft-Especific Weights
-  float boundary_weight_ = 5.0;
-  float notch_weight_ = 1.0;  // Used inside Core, but only for Soft.
-  float macro_blockage_weight_ = 1.0;
```
eder-matheus, line 285: Are these values being set in the new structure via Tcl?

AcKoucher: These values were never used.

The values that we actually use are still being set via Tcl.

PR #8124, opened 2025-08-26
ram: initial complete version

braydenlouie:
- Added hierarchical decoder
  - Tested on x2, x4, and x8 sizes
- Added input buffers
- Created dbRows
- Added fill cells through Tcl 
- Started work on adding bPins
- Verified through RTL simulation that behavioral logic is correct


oharboe:
what is the behavioral model of these SRAMs.?

add a cmd to write out the behavioral model? (faster simulation, equivalence checks)


How are properties modified: ports, synchronius read, common/shared read/write clock, are there any aspect ratio choices, pin placment options?

oharboe:
Given that there are an enormous number of possible condifigurations and that openroad changes all the time, a one-off equivalence check is not enough: it there needs to be a way for users to add equivalence checks to  a flow or when they generate the srams.

rovinski:
This is an undergraduate project. "Complete" in the title is only meant that it is an actual working RAM with buffers and decoders, rather than just the memory array.

> what is the behavioral model of these SRAMs.?

Same as https://github.com/AUCOHL/DFFRAM

> add a cmd to write out the behavioral model?

We can add that to the TODO list

> How are properties modified: ports

Not sure what this means. The ports are generated based on the generator options

> synchronius read

The address port combinationally determines the output data, if that's what you're asking.

> common/shared read/write clock

Read and write are on the same clock. We don't have plans to add separate clocks, but it could be put on the TODO list as a very low priority item. Timing is checked by STA so defining the constraints will not be straightforward.

> are there any aspect ratio choices

No, there is no mux support yet.

> pin placment options?

`D` and `Q` will be at the top edge, `addr` will be on the right edge (in the default orientation). We have plans to allow other configurations, but that is medium-to-low priority. The main configuration we would look to support is all pins on the same edge.

>  a one-off equivalence check is not enough

It was checked for all word sizes which are supported in this commit (x2, x4, x8). Formal equivalence checking is low priority.

maliberty:
Note that this is a DFF ram built out of standard cells so you could also simulate it at gate level.

oharboe:
> Note that this is a DFF ram built out of standard cells so you could also simulate it at gate level.

That would make sense for power simulations, maybe, but it would be awfully slow compared to a behavioral model. 

Pre-synthesis RTL for a pipelined multiplier vs netlist simulation after retiming comes to mind. There is a difference in simulation speed of orders of magnitude, not considering how long Verilator takes to build netlist.

Also, the behavioral model is used for eqy(or similar) to validate that the behavioral model matches the netlist.

rovinski:
I agree that a behavioral model is useful and I will put it on the TODO list. To give you an idea of scope, we currently have about 30 items on the TODO list for @braydenlouie to work on for as long as he is interested in the project.

oharboe:
Lots of niggling little details, or it would be done already...

src/ram/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+## IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+## ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+## LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+## CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+## SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+## INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+## CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+## ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+## POSSIBILITY OF SUCH DAMAGE.
+###########################################################################
```
rovinski, line 32: All copyright headers should be updated to the SPDX format, as seen in other files.

src/ram/test/make_8x8.defok
rovinski, line 1: This file likely has to be updated using `make ok`

src/ram/include/ram/ram.h
```diff
@@ -0,0 +1,135 @@
...
+  sta::dbNetwork* network_;
+  Logger* logger_;
+
+  odb::dbMaster* storage_cell_;
+  odb::dbMaster* tristate_cell_;
+  odb::dbMaster* inv_cell_;
+  odb::dbMaster* and2_cell_;
+  odb::dbMaster* clock_gate_cell_;
+  odb::dbMaster* buffer_cell_;
+  odb::dbMaster* filler_cell_;
```
rovinski: `filler_cell_` doesn't seem to be used. Even when it is used, it should probably be a `std::vector<odb::dbMaster*>`. Remove for now.

src/ram/src/ram.cpp
```diff
@@ -0,0 +1,609 @@
...
+  if (!chip) {
+    chip = odb::dbChip::create(db_);
+  }
+
+  block_ = chip->getBlock();
+  if (!block_) {
+    block_ = odb::dbBlock::create(chip, ram_name.c_str());
+  }
+
+  // 9 tracks for 8 bits per word plus
```
rovinski: "track" has a specific meaning which I'm not sure applies here. Perhaps you mean columns?

src/ram/src/ram.cpp
```diff
@@ -0,0 +1,609 @@
...
+  auto clock = makeBTerm("clk");
+
+  vector<dbBTerm*> write_enable(bytes_per_word, nullptr);
+  for (int byte = 0; byte < bytes_per_word; ++byte) {
+    auto in_name = fmt::format("we[{}]", byte);
+    write_enable[byte] = makeBTerm(in_name);
+  }
+
+  // input bterms
+  int numImputs = std::log2(word_count);
```
rovinski: `std::log2` returns a float, which is then truncated when assigning to an `int`. e.g. `std::log2(5)` will return ~2.3, which gets truncated to 2. You need to round up the number of bits required.
```suggestion
  int numImputs = std::ceil(std::log2(word_count));
```

src/ram/src/ram.cpp
```diff
@@ -0,0 +1,609 @@
...
+}
+
+dbBTerm* RamGen::makeOutputBTerm(const std::string& name)
+{
+  auto net = dbNet::create(block_, name.c_str());
+  auto bTerm = dbBTerm::create(net, name.c_str());
+
+  bTerm->setIoType(odb::dbIoType::OUTPUT);
+  return bTerm;
+}
```
rovinski: It probably makes more sense to have just one `makeBTerm` function and then pass in the IoType as an argument. Less code that way.

src/ram/test/make_8x8.tcl
```diff
@@ -15,7 +15,8 @@ generate_ram_netlist \
 
 ord::design_created
 
+filler_placement {sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 sky130_fd_sc_hd__fill_4}
+
 set def_file [make_result_file make_8x8.def]
 write_def $def_file
-diff_files make_8x8.defok $def_file
```
rovinski: Do not remove the `diff_files` check

src/ram/test/make_8x8.tcl
```diff
@@ -15,7 +15,8 @@ generate_ram_netlist \
 
 ord::design_created
 
+filler_placement {sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 sky130_fd_sc_hd__fill_4}
```
rovinski: Make sure to include all filler sizes
```suggestion
filler_placement {sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 sky130_fd_sc_hd__fill_4 sky130_fd_sc_hd__fill_8}
```

src/ram/src/ram.cpp
```diff
@@ -480,27 +444,27 @@ void RamGen::generate(const int bytes_per_word,
...
-      D[bit] = makeBTerm(fmt::format("D[{}]", bit + col * 8));
+      D[bit] = makeBTerm(fmt::format("D[{}]", bit + col * 8), dbIoType::INPUT);
       D_nets[bit] = makeNet(fmt::format("D_nets[{}]", bit + col * 8), "net");
     }
 
     vector<array<dbBTerm*, 8>> Q;
     // if readports == 1, only have Q outputs
     if (read_ports == 1) {
-      array<dbBTerm*, 8> d;
+      array<dbBTerm*, 8> q;
```
rovinski: Where does the static 8 come from? Is it because we assume only byte-level masking? This is fine for now, but I don't think there's a reason why we couldn't support arbitrary levels of masking (or at least something reasonable like powers of 2).

For example, it may be desirable to have word-level masking instead of byte-level masking if byte-level masking isn't needed.

src/ram/src/ram.cpp
```diff
@@ -480,27 +444,27 @@ void RamGen::generate(const int bytes_per_word,
     array<dbBTerm*, 8> D;     // array for b-term for external inputs
```
rovinski: I would use a more descriptive name, like `d_bterms`.
@maliberty what is the convention for capitalization in snake case? Should it be `d_bterms` or can it be `D_bterms`?

src/ram/src/ram.cpp
rovinski: Similarly, I would call this `q_bterms`

src/ram/src/ram.cpp
```diff
@@ -480,27 +444,27 @@ void RamGen::generate(const int bytes_per_word,
...
     array<dbNet*, 8> D_nets;  // net for buffers
     for (int bit = 0; bit < 8; ++bit) {
-      D[bit] = makeBTerm(fmt::format("D[{}]", bit + col * 8));
+      D[bit] = makeBTerm(fmt::format("D[{}]", bit + col * 8), dbIoType::INPUT);
       D_nets[bit] = makeNet(fmt::format("D_nets[{}]", bit + col * 8), "net");
     }
 
     vector<array<dbBTerm*, 8>> Q;
```
rovinski, line 444: The name on this is tough because it represents Q bterms of multiple ports. Maybe instead name it `read_port_bterms`?

src/ram/src/ram.cpp
```diff
@@ -603,7 +563,12 @@ void RamGen::generate(const int bytes_per_word,
                   sites_width);
   }
 
+  int max_y_coord = ram_grid.getHeight() * (word_count + 1);
+  int max_x_coord = ram_grid.getRowWidth();
```
rovinski, line 563: Pedantic: move these declarations below `ram_grid.placeGrid()` so that they are closer to where they are used.

Also, mark them as `const`.

rovinski (changes requested):

rovinski:
@maliberty very close to ready to review here

src/ram/src/ram.cpp
braydenlouie: At the moment, the static 8 is assuming byte-level masking because I've only been testing one-word DFFRAMs, but this can be changed moving forward.

rovinski:
@maliberty @vvbandeira ready for review

rovinski (approved):

src/ram/CMakeLists.txt
rovinski: @braydenlouie this copyright still needs to be adjusted

src/ram/src/MakeRam.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+  return new ram::RamGen;
+}
+
+void deleteRamGen(ram::RamGen* ram_gen)
+{
+  delete ram_gen;
+}
+
+void initRamGen(OpenRoad* openroad)
```
rovinski: This function needs to be adjusted so that it matches the declaration after the merge. You can look at any other tool to see how to use the new style of initialization.

This is what's currently causing the build error.

maliberty:
Please clean up the various action failures before review.

vvbandeira:
@braydenlouie CI is failing due to compiler warnings; we consider them as errors in CI builds.

rovinski:
@braydenlouie thanks - after you address the clang-tidy comments, we can request approval again. You can install `clang-tidy` locally to speed things up

src/ram/include/ram/ram.h
```diff
@@ -0,0 +1,104 @@
...
+////////////////////////////////////////////////////////////////
+class Cell;
+class Layout;
+class Grid;
+
+class RamGen
+{
+ public:
+  RamGen(sta::dbNetwork* network, odb::dbDatabase* db, Logger* logger);
+  ~RamGen();
```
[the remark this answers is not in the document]
braydenlouie: This destructor definition follows the convention for the rest of the repository, and I'm unsure if I should make this change.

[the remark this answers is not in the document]
rovinski: I think it's probably ok to ignore this one. The rest seem like they can be addressed.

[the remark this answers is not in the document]
rovinski: Actually, I take that back. I would just go ahead and follow the comment and declare the default in the header rather than the cpp

rovinski:
@braydenlouie next steps are to address the recent round of clang-tidy and then rerun clang-format. Also, follow the changes recommended by Tclint on the .tcl file.
The failures on pr-head and pr-merge are unrelated to your changes so they should be good to go.

Let's try to wrap this up ASAP so that you can move forward with other changes.

rovinski:
Still need to fix formatting on src/openroad.cc and the tcl file

maliberty:
> Still need to fix formatting on src/openroad.cc and the tcl file

Just did it to help

maliberty (approved):

oharboe:
How about an A/B comparison of synthesized RAM from behavioral Verilog to this implementation?

oharboe:
@rovinski Such as https://github.com/The-OpenROAD-Project/RegFileStudy

hzeller:
Is there a `ram-py.i` file missing as building with bazel results in the following issue:
```
bazel build src/ram/...
ERROR: ~/OpenROAD/src/ram/BUILD:68:15: Action src/ram/swig-py.cc failed: missing input file '//src/ram:src/ram-py.i'
ERROR: ~/OpenROAD/src/ram/BUILD:68:15: Action src/ram/swig-py.cc failed: 1 input file(s) do not exist
Use --verbose_failures to see the command lines of failed build steps.
ERROR: ~/OpenROAD/src/ram/BUILD:68:15 Action src/ram/swig-py.cc failed: 1 input file(s) do not exist
```

rovinski:
@oharboe several steps left to make this usable. First is power grid, pin placement, routing, and abstracting. Technically we also need to do tapcell insertion, but that won't affect PPA too much.

You are welcome to test drive it yourself and run some experiments once it's ready.

oharboe:
I'm mainly interested in the time/area compared to flip flop.

I thought the idea of this RAM implementation was to improve PPA in some way. Stands to reason that you'd want to measure that early to validate the advantages of this approach.

Currently I'm mainly interested in register files(port dominated, so implemented via flip flops) and I'd need behavior verilog for simulation before this work becomes relevant for RAMs.

rovinski:
> I thought the idea of this RAM implementation was to improve PPA in some way. Stands to reason that you'd want to measure that early to validate the advantages of this approach.

We are 100% confident that structured placement and routing is better than synthesized. It is well established and also the experiments have already been run a while ago, see: https://github.com/AUCOHL/DFFRAM?tab=readme-ov-file#comparisons

oharboe:
> > I thought the idea of this RAM implementation was to improve PPA in some way. Stands to reason that you'd want to measure that early to validate the advantages of this approach.
> 
> We are 100% confident that structured placement and routing is better than synthesized.

Well... it depends on the context and the quality of results you get from the tools in a context. The link below mentions "register files". My definition of register files is storage dominated by ports and access rather than storage bits.

A paper written on this topic would be much more gripping if it has data on when it is worth using a dedicated RAM generator compared to synthesis with a tool like OpenROAD (lower bound for quality of results).

Especially considering the downsides to RAM w.r.t not having an accurate behavioral model for simulation and the need for glue logic to match RAMs to use-case (In Chipyard they have the bartools to generate this glue logic).

> It is well established and also the experiments have already been run a while ago, see: https://github.com/AUCOHL/DFFRAM?tab=readme-ov-file#comparisons

Interesting. Though a RAM never exists in isolation, it is used in some context. What matters to me is to see what use-cases that make the effort of using this new RAM feature worth it considering that OpenROAD is known to reach 80%+ placement densities and that, especially as number of ports grow, ports and a more holistic flow can get better results than isolating separate parts of the system and joining at the end. 

As an example with a CPU, a register file(a 2R1W was classified as such in the link above) never exists in isolation, it comes with a bypass network(or you can't do back to back ALU operations). As number of ports grow, ports and bypass network dwarf the RAM/flip flops.

Even for storage dominated use-cases, there's the matter of speed required clock period. I already know that Yosys will go off the rails(take a very, very long time) to do synthesis for a *large* RAM, but I'm thinking it would be possible to do synthesis for a RAM block and get a netlist, then use retiming on the glue logic + yosys retiming to combine many such RAMs into larger RAMs. I've done similar things and this should get me a large RAM in OpenROAD through floorplan in minutes and I find that global placement is easily able to place this sort of thing.

rovinski:
> My definition of register files is storage dominated by ports and access rather than storage bits.

Electrically speaking, a register file is when the storage medium is through latch-based devices (i.e. latch or flip-flop). SRAMs are when the storage medium is SRAM bit cells (i.e. 6T SRAM), and DRAM is stored through dynamic capacitive devices (i.e. 1T1C cells). There are generally access limitations implied as well, for example register files can perform same-cycle accesses, but DRAMs must have pre-charge and readout phases, etc.

Architecturally speaking, a register file is anything that implements the architected set of registers. Anything more specific is a microarchitectural decision.

Although a high number of read/write ports is typical for an architected register file, it is not really categorized as part of the taxonomy. It is simply a matter of how many ports does the RAM have.

> when it is worth using a dedicated RAM generator compared to synthesis with a tool like OpenROAD

The answer is always, as long as you have the right RAM generator. A bespoke RAM generator is always going to beat out a generic synthesis/placement/routing tool. RAM generators can generate non-CMOS logic which is not possible with synthesized netlists. Even non-SRAM latch-based RFs will use dynamic logic, sense amplifiers, etc. to outperform any CMOS configuration.

As to whether it outperforms *this* RAM generator is purely a function of how much time is spent developing it. We only plan to incorporate logic which uses std cells. Technically it is already non-synthesizable because it uses a multi-driver bus for outputs, which is generally not synthesizable by tools.

>  OpenROAD is known to reach 80%+ placement densities

OpenROAD does a lot of timing optimization and therefore uses more logic area to get those results. Anecdotally, I recall some experiments some years ago where we compared synthesized RFs vs generated RFs. The synthesized version used about 40% more logic area (ignoring utilization) just to struggle and fail to hit the same timing as the generated RAM.

> As number of ports grow, ports and bypass network dwarf the RAM/flip flops.

This happens much more quickly in synthesized logic than it does with generated/bespoke logic. Synthesized logic will not produce multi-driver buses but instead use extremely wide muxes. This has rapidly diminishing returns.

> it comes with a bypass network(or you can't do back to back ALU operations).

Well, forwarded values typically exist outside the register file for in-order cores and they exist in the reorder buffer for out-of-order cores for that reason. RAM ports are simply fundamentally expensive, so many architectures workaround the need for large numbers of accesses within the same cycle by storing them in explicit structures. And yes, high-performance ROBs are hard to physically implement - I don't have any advice on that.

> Even for storage dominated use-cases

RAM generators will do better than this, 100% guaranteed. If it doesn't, then it's a problem with the RAM generator.

rovinski:
If we implement enough of the planned features for this RAM generator, it will give you better PPA than any synthesized RAM, almost guaranteed (only exception is if library std cells suck). We are focusing on basic and common use cases right now, which are 1rw RAMs and 1r1w RAMs. With good fortune we can explore higher-port RAMs, but it makes no sense to explore the complicated cases before we've polished the simple ones.

oharboe:
@rovinski Thank you for this excellent account. Very insightful...

For out of order CPUs, the port count goes beyond 3R2W... I asked ChatGPT to create a scaling table... :-)

Metric | Synthesized RF | Generated RAM
-- | -- | --
Area | O(n²) | O(n)
Delay | O(n) | O(1)–O(n)
Energy | O(n²) | O(n)
Typical crossover point | ~1R1W → 2R1W starts to explode | Scales up to 3R2W gracefully

oharboe:
@rovinski Can this be made to work on asap7?

Probably a silly question... I did look for some documentation (.md files), but didn't see any.

rovinski:
> @rovinski Can this be made to work on asap7?

I believe so, yes. The generator only requires these cells to work: 1) a flip-flop, 2) an AND gate, 3) a buffer and inverter, 4) a tristate driver, 5) a clock gate. I believe ASAP7 has all of those.

> Probably a silly question... I did look for some documentation (.md files), but didn't see any.

When we have something that's ready for an end-user, we'll try to add some.

> I asked ChatGPT to create a scaling table...

Not sure where it's getting those complexities, but I wouldn't put too much stake in that, especially for the synthesized ones.

maliberty:
For the open PDKs we don't have open RAM generators (openram hasn't seen wide adoption and some academic PDKs don't even have the building blocks).  This beats the alternative which is general PnR not a RAM generator.

PR #8128, opened 2025-08-27
Use `""` instead of `<>` for vendored boost dependency headers.

hzeller:
Boost is a vendored resources (in bazel but also in the way it is installed with etc/DependencyInstaller.sh).

Thus, use the quote instead of angle-bracket includes for them.


QuantamHD:
@hzeller will this work when we import into google3

hzeller:
@QuantamHD : don't know yet, but I'll have a look once it is in here, maybe it requires some copybara updates.

Edit: confirmed it will work.

src/cts/BUILD
```diff
@@ -24,7 +24,11 @@ cc_library(
         "src/Util.h",
     ],
     visibility = ["//visibility:private"],
-    deps = ["//src/sta:opensta_lib"],
+    deps = [
+        "//src/sta:opensta_lib",
+        "@boost.container_hash",
+        "@boost.unordered",
+    ],
```
maliberty, line 31: I'm curious why this change would require changes to the deps.  Were we getting these by some non-hermetic means before?

hzeller: Simple: In the course of fixing the headers I discovered that these deps were missing

maliberty: Wouldn't the code fail to compile without them then?

hzeller: maybe there was boost installed on the system, so it was accidentally compiling.

Because the include was with angle-brackets and thus looked as system header, the blaze would also not necessarily do a layering check.

maliberty: Can we avoid having it look at anything outside the sandbox?  This seems like a hole in the hermetic nature.

PR #8132, opened 2025-08-27
mpl: consider root cluster location in coarse shaping

AcKoucher:
Needed for partial automated macro placement support.

The idea is that the fixed macros will also have to be considered at the root-level annealing we performing coarse shaping. (the root-level annealing prevents moving on to cluster placement with a set of macros that we can't fit in the core area).


eder-matheus (approved):
The changes look very simple. Are there any expected changes in results?

AcKoucher:
I don't expect changes in the results.

Issue #8111, opened 2025-08-25
Automatic clustering can't be enabled despite message saying so


mole99:
### Describe the bug

During CTS I get the following message:

```
[INFO CTS-0204] A clustering solution was found from clustering size of 8 and clustering diameter of 50.
[INFO CTS-0205] Better solution may be possible if either -sink_clustering_size, -sink_clustering_max_diameter, or both options are omitted to enable automatic clustering.
```

Well, I did not pass either `-sink_clustering_size` nor `-sink_clustering_max_diameter`, so what is going on here?

In order for this message to be printed either `clusterSizeSet` or `maxDiameterSet` must be true: https://github.com/The-OpenROAD-Project/OpenROAD/blob/6a23a4b44f9e8b7a8ca3d700bab0e9e87b2dcc57/src/cts/src/HTreeBuilder.cpp#L142

They are set a few lines above by `options_->isMacroMaxDiameterSet()` and `options_->isMacroSinkClusteringSizeSet()` or `options_->isMaxDiameterSet()` and `options_->isSinkClusteringSizeSet()` respectively depending on the TreeType.
(Btw. the datatype of `clusterSizeSet` should be `bool` and not `unsigned`: https://github.com/The-OpenROAD-Project/OpenROAD/blob/6a23a4b44f9e8b7a8ca3d700bab0e9e87b2dcc57/src/cts/src/HTreeBuilder.cpp#L36).

It seems that for a TreeType of `TreeType::MacroTree` the automatic solution is never possible, since `macroSinkClustersSizeSet_` is `true` by default: https://github.com/The-OpenROAD-Project/OpenROAD/blob/6a23a4b44f9e8b7a8ca3d700bab0e9e87b2dcc57/src/cts/src/CtsOptions.h#L319
I think this should be `false`?

Still, why do I get this message for a sink tree?

`isSinkClusteringSizeSet()` is true in my case despite not passing `-sink_clustering_size`, and the only other place that sets it is: https://github.com/The-OpenROAD-Project/OpenROAD/blob/6a23a4b44f9e8b7a8ca3d700bab0e9e87b2dcc57/src/cts/src/TritonCTS.cpp#L214

[...]

PR #8133, opened 2025-08-27
Cts: improve automatic clustering for designs with max fanout 

arthurjolo:
Fixes [#8111](https://github.com/The-OpenROAD-Project/OpenROAD/issues/8111)

For designs with a max fanout set, if the limit is smaller than the default cluster size (20), CTS will set the cluster size to the max fanout limit and disable automatic clustering.

The comparison with the default value of 20 is incorrect, because, for automatic clustering we use three possible values (10, 20, 30). So, if a design has a max fanout of 25 and automatic clustering is enabled, CTS can still choose a cluster size of 30. Furthermore, in the same case, cluster sizes of 10 and 20 are still valid options but are never tried.

This PR fixes the issue mentioned above and improves automatic clustering for designs with a max fanout:

- When a cluster size is set and there is a max fanout limit, if the set value is greater than the max fanout limit, we set the size to the max fanout limit.

- When automatic clustering is enabled and there is a max fanout limit, we remove values greater than the max fanout limit and add the limit to the possible values, e.g.:
    - `cluster sizes = {10, 20 , 30}`
    - `max fanout = 25`
    - `limited cluster sizes = {10, 20, 25}`.

### DIFF
--- a/src/cts/src/CtsOptions.cpp
+++ b/src/cts/src/CtsOptions.cpp
@@ -35,10 +35,6 @@ void CtsOptions::limitSinkClusteringSizes(unsigned limit)
       sinkClusteringSizes_.begin(), sinkClusteringSizes_.end(), limit);
   sinkClusteringSizes_.erase(lowerBound, sinkClusteringSizes_.end());
   sinkClusteringSizes_.push_back(limit);
-
-  for(auto i : sinkClusteringSizes_) {
-    logger_->report("i: {}", i);
-  }
 }
 
 void CtsOptions::recordBuffer(odb::dbMaster* master, MasterType type)
--- a/src/cts/test/check_max_fanout3.ok
+++ b/src/cts/test/check_max_fanout3.ok
@@ -4,7 +4,6 @@
 [INFO ODB-0130]     Created 1 pins.
 [INFO ODB-0131]     Created 300 components and 2100 component-terminals.
 [INFO ODB-0133]     Created 1 nets and 300 connections.
-i: 8
 [INFO CTS-0050] Root buffer is sg13g2_buf_4.
 [INFO CTS-0051] Sink buffer is sg13g2_buf_4.
 [INFO CTS-0052] The following clock buffers will be used for CTS:


arthurjolo:
It seems we don´t have a unity test for cases where we have a max fanout and automatic clustering is enabled, I am creating one.

arthurjolo:
CI shows no changes.

src/cts/src/CtsOptions.cpp
```diff
@@ -25,6 +25,18 @@ void CtsOptions::inDbInstCreate(odb::dbInst* inst, odb::dbRegion* region)
...
 
+void CtsOptions::limitSinkClusteringSizes(unsigned limit)
+{
+  if (sinkClustersSizeSet_) {
+    setSinkClusteringSize(std::min(limit, sinkClustersSize_));
+    return;
+  }
+  auto lowerBound = std::lower_bound(
+      sinkClusteringSizes_.begin(), sinkClusteringSizes_.end(), limit);
+  sinkClusteringSizes_.assign(sinkClusteringSizes_.begin(), lowerBound);
```
maliberty: I don't think this is safe to do.  sinkClusteringSizes_.erase(lowerBound, sinkClusteringSizes_.end()); should work.  Are you intending to remove the lower_bound element and replace it with limit?

arthurjolo: Yes, the lower bound should not be a part of the vector.  From my tests this has worked fine, the vector is not corrupted, but I can see it being dangerous.

src/cts/src/CtsOptions.cpp
```diff
@@ -25,6 +25,22 @@ void CtsOptions::inDbInstCreate(odb::dbInst* inst, odb::dbRegion* region)
...
+    return;
+  }
+  auto lowerBound = std::lower_bound(
+      sinkClusteringSizes_.begin(), sinkClusteringSizes_.end(), limit);
+  sinkClusteringSizes_.erase(lowerBound, sinkClusteringSizes_.end());
+  sinkClusteringSizes_.push_back(limit);
+
+  for(auto i : sinkClusteringSizes_) {
+    logger_->report("i: {}", i);
+  }
```
maliberty: debug code?

PR #8140, opened 2025-08-28
ODB: Allow multiple chip def reads

osamahammad21:
This PR allows specifying the chip to populate with the read_def command. Needed for multiple chips in the same database.


include/ord/OpenRoad.hh
```diff
@@ -186,6 +187,7 @@ class OpenRoad
 
   void readDef(const char* filename,
                odb::dbTech* tech,
+               odb::dbChip* chip,
```
QuantamHD, line 194: Could you add a default value for chip to be nullptr. That way existing users of this API don't need to update

osamahammad21: I intentionally left `chip` non-default to ensure that the caller explicitly specifies which chip should be populated. The goal is to make the behavior unambiguous—readDef should focus solely on populating the provided chip rather than creating or interpreting which chip to target.

maliberty: @QuantamHD 👍 👎 ?

maliberty:
Summarizing a conversation with @QuantamHD I think he would like to see us not need both of these args: 
```
               odb::dbTech* tech,
               odb::dbChip* chip,
```
I think the best way to fix that is to move the dbTech to be associated with the dbChip rather than the dbBlock.  It feels more semantically accurate as you can't have multiple techs on a single chip.

include/ord/OpenRoad.hh
maliberty: I believe this is resolved as we now only require chip and not tech.

PR #8158, opened 2025-08-29
Don't use stdio.h remove(), but std::filesystem::remove() instead.

hzeller:
Follow-up change to #8154 : use modern `std::filesystem::remove()` instead of old-school `remove()` from stdio.h

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -14,6 +14,7 @@
 #include <iostream>
 #include <memory>
 #include <string>
+#include <system_error>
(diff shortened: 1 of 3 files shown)


hzeller:
this is in the other places that used the `remove()` function and thus also would need to use the `stdio.h` header.

maliberty:
conflict to resolve

hzeller:
conflict resolved, ready to merge.

PR #8162, opened 2025-08-30
boost: include bind.hpp where needed; Use single-redirect geometry he…

hzeller:
…ader.

Include header needed to use boost::bind where missing

Instead of `boost/geometry.hpp`, include `boost/geometry/geometry.hpp`. The reason is that boost/geometry.hpp is a header that just includes the boost/geometry/geometry.hpp header, which in turn is a header that includes a bunch of the actual implementation headers.

The misc-include-cleaner complains about symbols it doesn't see, but that is because it is picky that it actualy wants the direct header that defines a thing. With redirecting headers, there is a way to add a `IWHY pragma: export` to tell that it forwards includes (and I'll send a pull request upstream
to boost to make it happen), but that only works over one level. So we need to include the header that actually has all the forwarding includes (`"boost/geometry/geometry.hpp"`), not the header that forwards to the forwarding header (`"boost/geometry.hpp"`).  (similar for stacktrace).

Until I have sent a fix boost-upstream to add the IWYU pragmas, that won't do anything yet, but we're prepared.


hzeller:
Fixing about 130 misc-include-cleaner warnings.

maliberty:
I can imagine that one level will not always be sufficient.

PR #8164, opened 2025-08-30
IWYU: Include utl/Logger.h where it is used.

hzeller:


src/dft/src/stitch/BUILD
```diff
@@ -26,6 +26,7 @@ cc_library(
         "//src/dft/src/config",
         "//src/dft/src/utils",
         "//src/odb",
+        "//src/utl",
```
maliberty, line 29: Earlier we have system headers being non-hermetic but this seems even more surprising.  How would the code have compiled without being able to see utl?

maliberty (approved):

src/dft/src/stitch/BUILD
hzeller: the header was imported through indirect inclusions, linking a library that happend to bring the visibility in; the final library did not have the correct dependency. It happens to work because the final openroad binary will link everything together.

Once you IWYU, the build-system can make that an error. This is why it is so important to do and fix all the misc-include-cleaner issues.

So it is all as expected, just the project needs a little more work.

PR #8174, opened 2025-08-31
buildifier: run it to reduce PR noise later

oharboe:
Add buildifier to CI?


maliberty (approved):

maliberty:
> Add buildifier to CI?

I'll open a task though it will be a lower priority.

PR #8177, opened 2025-08-31
Move MakeExample to exa:ui as it is needed for the tcl-swig

hzeller:


hzeller:
I don't know exactly how the swig and tcl stuff works, but looks like the `exa_tcl_inits` variable needs to be visible. Anyway, did the same thing as in the other tcl build rules where the `Make*.cc` is part of the ui-library.

hzeller:
This got messed up in my previous PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/8163

maliberty (approved):

maliberty:
src/exa/src/example.tcl gets turned into an encoded string that is compiled into OR.  That way we don't have to ship additional files.  At startup we decode such strings and evaluate them in the TCL interpreter.

maliberty:
The encoded string is exa_tcl_inits

PR #8178, opened 2025-08-31
power: hierarchical .odb testing

oharboe:
mock-array/Element now has submodules, which is useful for power testing.

More testing to come, as hierarhical .odb gets better, this is just the prerequisites on top of which I can add more tests.

The design is bigger because more kept modules means less efficient synthesis.

### DIFF
--- a/test/orfs/mock-array/BUILD
+++ b/test/orfs/mock-array/BUILD
@@ -19,17 +19,11 @@ POWER_STAGE_STEM = {
 MOCK_ARRAY_TABLE = [
     8,
     8,
-    20,
-    20,
-    20,
-    22,
+    30,
+    30,
+    30,
+    32,
 ]
-
-# Element'd data width
-MOCK_ARRAY_DATAWIDTH = 64
-
-# Must be zero for routing by abutment
-MACRO_BLOCKAGE_HALO = 0
 
 MOCK_ARRAY_SCALE = 45
 
@@ -76,10 +70,6 @@ ce_margin_y = placement_grid_y * 0.5
 array_spacing_x = margin_x * 2
 
 array_spacing_y = margin_y * 2
-
-array_offset_x = array_spacing_x + margin_x
-
-array_offset_y = array_spacing_y + margin_y
 
 # top level core and die size
 core_width = (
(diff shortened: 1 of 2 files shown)


maliberty:
This pushes bazel tests to be longer than cmake.  Can you reduce this further and still test what you need?  OR should not have ORFS level tests

oharboe:
> This pushes bazel tests to be longer than cmake. Can you reduce this further and still test what you need? OR should not have ORFS level tests

I don't know what "ORFS level tests" mean, but the design goal of the test/orfs/... tests is to run within minutes, not hours.

To test power, we need a .vcd file from a netlist + estimated parasitics.

The problem is actually that verilator is taking long for some reason(30 minutes). After verilator has run, the actual compilation of the .cpp files and generation of .vcd is snappy.

maliberty:
I mean the runtime is far too long:
image

maliberty:
Versus master
image

oharboe:
@maliberty fixed here https://github.com/The-OpenROAD-Project/OpenROAD/pull/8181. I suggest merging letting https://github.com/The-OpenROAD-Project/OpenROAD/pull/8181 merge first, so that we have some data on the books w.r.t. build speeds and then I'll merge this PR w/master and we should be good.

oharboe:
@maliberty :crossed_fingers: build times should now be short for bazel...

oharboe:
@maliberty Mostly waiting for a build server, the actual tests are 18m, which is not bad for the testing value.

Power testing requires the flexibility of Bazel to support all the tools required for this intricate flow and OpenROAD is the only place where we've got Bazel plugged into CI. It is "integration testing" rather than "unit-testing" of power, but I don't know of a simpler, faster, simpler, way to test power w/.vcd activations.

image

maliberty:
>  I don't know of a simpler, faster, simpler, way to test power w/.vcd activations.

18m is ok though you can just use a smaller design to make it faster.

oharboe:
> > I don't know of a simpler, faster, simpler, way to test power w/.vcd activations.
> 
> 18m is ok though you can just use a smaller design to make it faster.

Possibly, but the next step is to update CI to record the critical path in the bazel build so that we can see if there's any point in spending more time. I'm unsure we'll get the same coverage with a smaller design. mock-array has been VERY versatile w.r.t. testing... it doesn't only test power.

PR #8181, opened 2025-08-31
power: faster verilator to generate .vcd file

oharboe:


oharboe:
@maliberty With this change, power testing, based on estimated parasitics of cts w/verilator .vcd simulation, finishes before detailed routing of mock-array.

maliberty:
How does --flatten change when it happens in the flow?

oharboe:
> How does --flatten change when it happens in the flow?

It doesnt change anything in the .vcd: it just inlines C++ modules for  behavioral models of flip flops, buffers, etc. which are pathologically slow to generate for some reason. I dont particularly want or need to know why verilator is slow here, I am happy with the pinata result and power.tcl checks that all pins are annotated by the .vcs file, so we are good...

maliberty (approved):

PR #8182, opened 2025-08-31
Support for Rectilinear Dies in PPL

gs-chaitanya:
This PR aims to add support for pin placement for rectilinear dies. Rectilinear die edges are treated as odb::Line objects and, therefore, new functions were made as replacements for old functions in the flow, which used the Edge enum to deal with edges.
A new function for calculating pin orientations was developed which uses Ray Tracing to properly determine which side of the edge face the polygon. (New PR after fixing merge issues)


test/final.def
eder-matheus, line 1: It seems this file is not used anywhere. Please remove it.

src/ppl/src/IOPlacer.tcl
```diff
@@ -216,18 +216,27 @@ sta::define_cmd_args "place_pins" {[-hor_layers h_layers]\
                                   [-exclude region]\
                                   [-group_pins pin_list]\
                                   [-annealing] \
-                                  [-write_pin_placement file_name]
+                                  [-write_pin_placement file_name]\
+                                  [is_rectilinear_die]
```
eder-matheus: The tool should be able to identify if the design has a polygonal die area automatically. There shouldn't be a new option in the tool.

src/ppl/src/Core.h
```diff
@@ -74,6 +77,7 @@ class Core
   std::map<int, int> min_width_x_;
   std::map<int, int> min_width_y_;
   int database_unit_;
+  std::vector<odb::Line> polygon_edges_;  // stores edges of the polygon
```
eder-matheus: A more descritive name would be `die_edges_` or `die_area_edges_`. It would also be consistent with the function `getDieAreaEdges` that you've created.

src/sta
eder-matheus, line 1: There shouldn't be changes in the sta submodule. Please look at the current commit of sta in the OpenROAD master branch and update the sta commit in your branch accordingly.

src/ppl/src/IOPlacer.cpp
```diff
@@ -442,6 +446,34 @@ bool IOPlacer::checkBlocked(Edge edge, const odb::Point& pos, int layer)
...
+        || blocked_interval.getLayer() == layer) {
+      if (coord > blocked_interval.getBegin()
+          && coord < blocked_interval.getEnd()) {
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
```
eder-matheus: The differences between this function and `checkBlocked` are very small. I would like to see them merged into a single one.

You could take both `Edge` and `odb::Line` as arguments. If you're handling an Edge, the odb::Line could be an empty line. If you're handling a line, the `Edge` can be `Edge::invalid`

src/ppl/src/IOPlacer.cpp
```diff
@@ -2128,6 +2600,101 @@ void IOPlacer::runHungarianMatching()
   clear();
 }
 
+void IOPlacer::runHungarianMatchingPolygon()
```
eder-matheus: I don't think you need an specific run function for polygon shapes. The only difference I see here is the call of the `assignPinsToSectionsPolygon(constrained_pins_cnt);` function (correct me if I'm wrong, please). You could just update the original `runHungarianMatching` function with a check if the die is polygonal or not, and call the appropriate function according to the die shape (rectangle or polygon).

src/ppl/src/IOPlacer.i
```diff
@@ -159,6 +159,12 @@ run_hungarian_matching()
   getIOPlacer()->runHungarianMatching();
 }
 
+void
+run_hungarian_matching_polygon()
+{
+  getIOPlacer()->runHungarianMatchingPolygon();
+}
```
eder-matheus: You won't need this function once you automatically detect the die shape.

src/ppl/src/IOPlacer.cpp
```diff
@@ -819,6 +898,109 @@ std::vector<Point> IOPlacer::findLayerSlots(const int layer, const Edge edge)
   return slots;
 }
 
+std::vector<Point> IOPlacer::findLayerSlotsPolygon(const int layer,
```
eder-matheus: My comments here are similar to the comment in the `checkBlockedPolygon` function. The differences between the `findLayerSlotsPolygon` and `findLayerSlots` are small and don't justify having two different functions, specially when they are very long functions.

Please merge them into a single one, using the same logic suggested previously.

src/ppl/src/IOPlacer.cpp
```diff
@@ -933,6 +1174,40 @@ void IOPlacer::findSections(int begin,
...
+    }
+    n_sec.begin_slot = begin;
+    n_sec.end_slot = end_slot;
+    n_sec.used_slots = 0;
+    n_sec.edge = Edge::polygonEdge;
+
+    sections.push_back(n_sec);
+    begin = ++end_slot;
+  }
+}
```
eder-matheus: Notice how the `odb::Line poly_edge` is not even used inside the function. You should have only the `findSections` function, and pass a `bool` determine if the section will be for an `Edge::polygonEdge` or for a regular Edge.

eder-matheus: Now that I notice, this function is not even used in your code. This can just be removed.

src/ppl/src/IOPlacer.cpp
eder-matheus: You'll need to update `runAnnealing()` too.

src/ppl/src/IOPlacer.cpp
```diff
@@ -709,7 +741,54 @@ void IOPlacer::findSlots(const std::set<int>& layers, Edge edge)
...
       bool blocked = checkBlocked(edge, pos, layer);
-      slots_.push_back({blocked, false, pos, layer, edge});
+      // For regular die boundary edges, use default-constructed line
+      odb::Line empty_line;  // Default constructor creates empty line
+      slots_.push_back({blocked, false, pos, layer, edge, empty_line});
+    }
+  }
+}
+
+void IOPlacer::findSlotsPolygon(const std::set<int>& layers, odb::Line line)
```
eder-matheus: This function should be merged with `findSlots` too. The differences between them are pretty minor, and could follow the same logic you'll apply for the other functions that will be merged.

src/ppl/src/IOPlacer.cpp
```diff
@@ -899,6 +1081,65 @@ void IOPlacer::defineSlots()
...
+void IOPlacer::defineSlotsPolygon()
+{
+  for (auto line : core_->getDieAreaEdges()) {
+    bool is_vertical_pin = (line.pt0().getY() == line.pt1().getY());
+    if (is_vertical_pin) {
+      findSlotsPolygon(ver_layers_, line);
+    } else {
+      findSlotsPolygon(hor_layers_, line);
+    }
+  }
```
eder-matheus: It seems that only this block of code is different when comparing wiwth `defineSlots`. Merge both functions too.

src/ppl/src/IOPlacer.cpp
```diff
@@ -1293,6 +1635,59 @@ bool IOPlacer::assignPinsToSections(int assigned_pins_count)
   return false;
 }
 
+bool IOPlacer::assignPinsToSectionsPolygon(int assigned_pins_count)
+{
+  Netlist* net = netlist_.get();
+  std::vector<Section>& sections = sections_;
+
+  createSectionsPolygon();
```
eder-matheus: This call is the only thing different from `assignPinsToSections`. Merge them into a single function.

src/ppl/src/IOPlacer.cpp
```diff
@@ -1417,6 +1812,83 @@ void IOPlacer::updateOrientation(IOPin& pin)
   }
 }
 
+bool IOPlacer::isPointInsidePolygon(odb::Point point,
```
eder-matheus, line 1567: Add a brief comment describing how the ray casting algorithm works.

eder-matheus (changes requested):

src/ppl/src/Core.h
gs-chaitanya: refactored

src/ppl/src/IOPlacer.cpp
gs-chaitanya: merged them for now, but we need to discuss a better way to handle branching the flow for normal die vs polygon

src/ppl/src/IOPlacer.i
```diff
@@ -159,6 +159,12 @@ run_hungarian_matching()
   getIOPlacer()->runHungarianMatching();
 }
 
+// void
+// run_hungarian_matching_polygon()
+// {
+//   getIOPlacer()->runHungarianMatchingPolygon();
+// }
```
eder-matheus: Remove commented code that will not be used anymore.

src/ppl/include/ppl/IOPlacer.h
```diff
@@ -202,14 +215,18 @@ class IOPlacer
...
   void updatePinArea(IOPin& pin);
   void movePinToTrack(odb::Point& pos,
                       int layer,
                       int width,
                       int height,
                       const Rect& die_boundary);
   Interval getIntervalFromPin(IOPin& io_pin, const Rect& die_boundary);
-  bool checkBlocked(Edge edge, const odb::Point& pos, int layer);
+  bool checkBlocked(Edge edge, odb::Line, const odb::Point& pos, int layer);
+  // bool checkBlockedPolygon(odb::Line& line, const odb::Point& pos, int
```
eder-matheus: Remove commented code.

src/ppl/include/ppl/IOPlacer.h
```diff
@@ -89,6 +90,7 @@ class IOPlacer
   void clear();
   void clearConstraints();
   void runHungarianMatching();
+  // void runHungarianMatchingPolygon();
```
eder-matheus: Remove commented code.

src/ppl/src/IOPlacer.cpp
```diff
@@ -703,30 +719,66 @@ void IOPlacer::findSlots(const std::set<int>& layers, Edge edge)
...
+      }
     }
   }
 }
 
-std::vector<Point> IOPlacer::findLayerSlots(const int layer, const Edge edge)
+std::vector<Point> IOPlacer::findLayerSlots(const int layer,
+                                            const Edge edge,
+                                            odb::Line line,
+                                            bool isDiePolygon)
```
eder-matheus: `isDiePolygon` --> `is_die_polygon`

gs-chaitanya: corrected this in all the functions

QuantamHD:
@gs-chaitanya Congratulations! Very nice work!

gs-chaitanya:
Thanks @QuantamHD 
Thank you, @eder-matheus and @gudeh, for your guidance!

PR #8190, opened 2025-09-01
grt: introduce CUGR-2 as an alternative solver for global routing

eder-matheus:
CUGR-2 original code can be found here: https://github.com/cuhk-eda/cu-gr-2

Original license is added inside the cugr directory and was preserved untouched.

This is the first step on integrating the new solver. Initial tests shows better wirelength results in the final routing, but a few designs ended with DRCs in the private CI. Also, support for incremental repairs are necessary.


src/grt/src/cugr/README.md
```diff
@@ -0,0 +1,75 @@
...
+Please read our paper for more technical details:
+
+* Jinwei Liu, Evangeline F. Y. Young,
+EDGE: Efficient DAG-Based Global Routing Engine,
+ACM/IEEE Design Automation Conference (DAC), San Francisco, CA, USA, July 9-13, 2023.
+
+(CUGR 2.0 supports ICCAD'19 benchmarks ([v2](http://iccad-contest.org/2019/Problem_C/iccad19_benchmarks_v2.tar.gz), [hidden](http://iccad-contest.org/2019/Problem_C/iccad19_hidden_benchmarks.tar.gz)).
+This version of code is consistent with the DAC paper.)
+
+## 1. How to Build
```
maliberty: No need to keep the "How to Build" section

src/grt/src/cugr/README.md
```diff
@@ -0,0 +1,75 @@
...
+CUGR 2.0 is a detailed routability-driven global router and its solution quality is solely determined by the final detailed routing results.
+
+Please read our paper for more technical details:
+
+* Jinwei Liu, Evangeline F. Y. Young,
+EDGE: Efficient DAG-Based Global Routing Engine,
+ACM/IEEE Design Automation Conference (DAC), San Francisco, CA, USA, July 9-13, 2023.
+
+(CUGR 2.0 supports ICCAD'19 benchmarks ([v2](http://iccad-contest.org/2019/Problem_C/iccad19_benchmarks_v2.tar.gz), [hidden](http://iccad-contest.org/2019/Problem_C/iccad19_hidden_benchmarks.tar.gz)).
+This version of code is consistent with the DAC paper.)
```
maliberty: rm as "This version of code is consistent with the DAC paper" is unlikely to remain true and the benchmarks are not important in the OR context.

src/grt/src/cugr/README.md
```diff
@@ -0,0 +1,75 @@
...
+**Step 2:** Go to the project root and build by
+```bash
+$ cd cu-gr-2
+$ scripts/build.py -o release
+```
+
+Note that this will generate two folders under the root, `build` and `run` (`build` contains intermediate files for build/compilation, while `run` contains binaries and auxiliary files).
+More details are in [`scripts/build.py`](scripts/build.py).
+
+### 1.1. Dependencies
```
maliberty: rm dependencies section - stale in OR.

src/grt/src/GlobalRouter.tcl
```diff
@@ -149,15 +149,16 @@ sta::define_cmd_args "global_route" {[-guide_file out_file] \
                                   [-allow_congestion] \
                                   [-verbose] \
                                   [-start_incremental] \
-                                  [-end_incremental]
+                                  [-end_incremental] \
+                                  [-use_cugr]
```
maliberty, line 153: This will need to be documented but lets hold off until we have more confidence.

eder-matheus: I think the `Run Docs Test` will fail in the jenkins run if I don't add it. Perhaps I can add it with a disclaimer that it's not ready for production yet?

maliberty: Ok.

maliberty (approved):
we can deal with clang-tidy later

maliberty:
@hzeller FYI there may be new IWYU changes needed for this newly imported code.  Its still experimental so not a high priority; just want you to be aware.

PR #8192, opened 2025-09-01
test: add option for pre-placed macros in flow.tcl

AcKoucher:


test/flow.tcl
```diff
@@ -56,6 +56,10 @@ source $tracks_file
 # remove buffers inserted by synthesis
 remove_buffers
 
+if { $pre_placed_macros_file != "" } {
```
eder-matheus, line 59: The regression tests are failing because it can't find this variable. I believe you'll have to add it into the `.vars` file of each tech with the default value of `""` to prevent the error.

AcKoucher:
@eder-matheus I just realized that I have to add an empty file string to the existing tests in order for them not to fail with these changes. I'll ping once it's done.

eder-matheus (approved):

PR #8195, opened 2025-09-01
IWYU: Include needed odb headers in files uses classes from them.

hzeller:
Fixing about 1500 misc-include-cleaner clang-tidy warnings.


hzeller:
Some of the forward declarations are probably obsolete now, but I didn't specifically target that. And if a forward declaration is there and might be sufficient, clang-tidy will also not complain about the missing header. First I want to focus on adding all headers, then go from there.

roughly
  1) to add all headers needed in places that use a symbol, but rely on accidentally having the symbol visible due to other includes (current phase; 8000 fixes to go)
  2) Remove forward declares and see if that makes a difference or requires more includes added.
  3) Now that all the necessary headers are there, we can remove headers that are included but are not needed (maybe they were needed in the past). That alone is about 1500 of the warnings, but this can't be done yet, because some of these might be needed for the 'accidental' dependencies they provide.

Then, do all the fixes from above in the generated files which I am currently not touching.

PR #8200, opened 2025-09-02
drt: use relative paths to the respective implementation headers.

hzeller:
This makes it easier to follow where things are, as well as reduces the proliferation of needed `-I` include search paths.


maliberty:
Is there no corresponding cmake change that would enforce the same?

hzeller:
I don't know cmake enough to know.

PR #8202, opened 2025-09-02
Fix line-endings in src/grt/src/cugr

hzeller:
Looks like these files picked up CRLF line endings at some point src/grt/src/cugr/

Looks like .clang-format was not actually instructed to generate canonical line ending.

Changed that and ran it over the src/grt files.


hzeller:
@eder-matheus Looks like you've been working on these files, so you should have a look so that this PR is not modifying things you're working on.

eder-matheus:
> @eder-matheus Looks like you've been working on these files, so you should have a look so that this PR is not modifying things you're working on.

Thanks for the heads up, and thanks for catching this bug in clang-format. It will not affect any work I'm doing now, so it should be safe to go.

PR #8211, opened 2025-09-03
More use of `""` instead of `<>` for vendored boost includes.

hzeller:
Like #8128, but thse went under the radar due to unusual file suffices.

### DIFF
--- a/src/ant/src/Polygon.hh
+++ b/src/ant/src/Polygon.hh
@@ -3,11 +3,10 @@
 
 #pragma once
 
-#include <boost/functional/hash.hpp>
-#include <boost/polygon/polygon.hpp>
-
 #include "PinType.hh"
 #include "ant/AntennaChecker.hh"
+#include "boost/functional/hash.hpp"
+#include "boost/polygon/polygon.hpp"
 
 namespace ant {
 
--- a/src/dpl/src/optimization/detailed_global.cxx
+++ b/src/dpl/src/optimization/detailed_global.cxx
@@ -4,7 +4,6 @@
 #include "detailed_global.h"
 
 #include <algorithm>
-#include <boost/tokenizer.hpp>
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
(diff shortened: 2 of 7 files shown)


hzeller:
Thre are too many different c++ suffices in use :)

Excluding sta/, these are the ones:

```
     17 .cxx
     55 .hh
     61 .cc
    714 .h
    733 .cpp
```

Maybe the `.hh`, `.cc` and `.cxx` 'outliers' should be renamed to `*.h` and `*.cpp` respectively at some point to fit the majority and reduce confusion.

(personally, I would of course prefer `*.cc` to `*.cpp` because it does not sound like the c-preprocessor binary that is called `cpp`, but looks like that ship has sailed)

maliberty:
| Thre are too many different c++ suffices in use :)

Totally agree.  Also camel case vs snake case vs kebab case.  Fwiw I prefer kebab for files as it requires no use of the shift key.  I'm not sure how well git handles such renames in merges (ie existing PRs/branches).  We may need to coordinate such changes.

hzeller:
Yes, kebab-case is the most beautiful way of naming files (here is when I did that for verible https://github.com/chipsalliance/verible/commit/6acd488d9ca3b35bf95586ff40a872ae7cb991d9 (Only using underscore for deliniating tests. For every `foo-bar.cc` there is a `foo-bar_test.cc`)).

Since the filename is different afterwards, it requires to use `git log --follow` to see the full history of a file.

I had written a script once that asks the github API to figure out what files are currently open in pull requests to avoid touching in a cleanup, and then have multiple waves of doing things on files currently not in any open PRs. Still it might be useful to coordinate for everyone who is currently working on something but has not opened a PR yet.

maliberty:
I'm curious why you used _test.cc over -test.cc as both seem equally clear and the later avoids an exception.

hzeller:
It is mostly because then it reads as two words, as underscore is a 'weaker' binding, and we don't want to use spaces in filennames of course.

So it reads a bit lik`foo-bar-baz test` (a test that tests `foo-bar-baz`) but is written`foo-bar-baz_test`.

Mostly personal preference of course, but that is the reasoning.

PR #8212, opened 2025-09-03
Fix some missing includes in new cugr/ sources

hzeller:
Found with misc-include-cleaner (fixed using
scripts from https://github.com/hzeller/dev-tools )


hzeller:
@eder-matheus FYI so that it does not interfere with what you might be doing.

PR #8214, opened 2025-09-03
ODB: 3dBlox parsers

osamahammad21:


src/odb/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ add_subdirectory(src/lefout)
 add_subdirectory(src/def)
 add_subdirectory(src/zutil)
 add_subdirectory(src/cdl)
+add_subdirectory(src/3dblox)
```
maliberty, line 19: You'll need to setup bazel too and probably add yaml-cpp as a bazel_dep in MODULE.bazel (it is available in BCR)

src/odb/include/odb/3dblox.h
```diff
@@ -0,0 +1,45 @@
...
+{
+ public:
+  ThreeDBlox(utl::Logger* logger, odb::dbDatabase* db);
+  ~ThreeDBlox() = default;
+  void readDbv(const std::string& dbv_file);
+  void readDbx(const std::string& dbx_file);
+
+ private:
+  utl::Logger* logger_ = nullptr;
+  odb::dbDatabase* db_ = nullptr;
```
maliberty: Methods before data members https://google.github.io/styleguide/cppguide.html#Declaration_Order

src/odb/include/odb/3dblox.h
```diff
@@ -0,0 +1,45 @@
...
+class dbDatabase;
+class ChipletDef;
+class ChipletRegion;
+class dbChip;
+class ChipletInst;
+class Connection;
+class DesignDef;
+class dbChipRegionInst;
+class dbChipInst;
+class ThreeDBlox
```
maliberty, line 24: A blank link separating the definition from the declarations would be helpful

src/odb/include/odb/db.h
```diff
@@ -7238,6 +7238,10 @@ class dbChipRegionInst : public dbObject
 class dbDatabase : public dbObject
 {
  public:
+  void setDbuPerMicron(uint dbu_per_micron);
```
maliberty, line 7241: Does this replaced the methods at lower levels?

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -0,0 +1,307 @@
...
+                     dbv_file);
+    }
+  }
+  for (const auto& [_, chiplet] : data.chiplet_defs) {
+    createChiplet(chiplet);
+  }
+}
+
+std::string ThreeDBlox::resolveIncludePath(const std::string& include_path,
+                                           const std::string& current_file_path)
```
maliberty, line 59: Better to use std::filesystem::path over strings

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -0,0 +1,307 @@
...
+{
+  DbxParser parser(logger_);
+  DbxData data = parser.parseFile(dbx_file);
+  for (const auto& include : data.header.includes) {
+    std::string resolved_path = resolveIncludePath(include, dbx_file);
+    if (include.find(".3dbv") != std::string::npos) {
+      readDbv(resolved_path);
+    } else if (include.find(".3dbx") != std::string::npos) {
+      readDbx(resolved_path);
+    }
```
maliberty, line 81: Warn on unknown type?

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -0,0 +1,307 @@
...
+    return dbChip::ChipType::IP;
+  }
+  if (type == "substrate") {
+    return dbChip::ChipType::SUBSTRATE;
+  }
+  if (type == "hier") {
+    return dbChip::ChipType::HIER;
+  }
+  logger->error(
+      utl::ODB, 527, "3DBV Parser Error: Invalid chip type: {}", type);
```
maliberty, line 110: You can throw from here rather than require a logger arg.  This is non-standard compared to other types in dbTypes.h/cpp.  Likewise elsewhere

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -0,0 +1,307 @@
...
+  dbTechLayer* layer = nullptr;
+  if (!region.layer.empty()) {
+    // TODO: add layer
+  }
+  dbChipRegion* chip_region = dbChipRegion::create(
+      chip, region.name, getChipRegionSide(region.side, logger_), layer);
+  Rect box;
+  bool init = true;
+  for (const auto& coord : region.coords) {
+    if (init) {
```
maliberty: Use Rect::mergeInit()

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -0,0 +1,307 @@
...
+                   519,
+                   "3DBX Parser Error: Chiplet instance reference {} not found "
+                   "for chip inst {}",
+                   chip_inst.reference,
+                   chip_inst.name);
+  }
+  dbChipInst* inst = dbChipInst::create(db_->getChip(), chip, chip_inst.name);
+  inst->setLoc(Point3D(chip_inst.loc.x * db_->getDbuPerMicron(),
+                       chip_inst.loc.y * db_->getDbuPerMicron(),
+                       chip_inst.z * db_->getDbuPerMicron()));
```
maliberty, line 195: Does this behave like DEF where you have to place the corner after orientation is done?

maliberty (changes requested):
Not a full review but some feedback

src/odb/src/3dblox/baseParser.cpp
```diff
@@ -0,0 +1,128 @@
...
+void BaseParser::parseIncludes(std::vector<std::string>& includes,
+                               const std::string& content)
+{
+  includes.clear();
+
+  std::istringstream stream(content);
+  std::string line;
+
+  while (std::getline(stream, line)) {
+    if (line.find("#!include") == 0) {
```
maliberty: I don't see #!include in the spec but I do see #!define

src/odb/include/odb/db.h
osamahammad21: Not really. tech does still have dbu_per_micron of its own

src/odb/src/3dblox/3dblox.cpp
osamahammad21: Yes, but that shouldn't matter now since I don't actually place it anywhere or create a bounding box for it yet.

src/odb/src/3dblox/3dblox.cpp
osamahammad21: Not yet. there are other includes that I don't support yet that are not necessarily wrong to issue an error for.

src/odb/src/3dblox/3dblox.cpp
osamahammad21: What's wrong with using the logger in here?

maliberty (approved):

Issue #8194, opened 2025-09-01
EST/GRT: add missing via resistance on parasitics estimation


eder-matheus:
### Description

When estimating parasitics from route guides, the via resistances from Metal1-Metal2 are not being added in the net parasitics. This is probably caused by the min routing layer set as Metal2.

### Suggested Solution

Independently of the min routing layer, the parasitics estimated from the global routing should include the vias down to the pin layer.

PR #8215, opened 2025-09-03
grt: add resistance for vias below min routing layer

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/8194

Pending secure-ci.


eder-matheus:
@jeffng-or FYI, this should fix the missing via you saw in your experiments.

maliberty (approved):

PR #8221, opened 2025-09-03
grt/odb: add db callback for access point update

eder-matheus:
This new ODB callback will be necessary for incremental pin access.

Calling pin access in each iteration of incremental GRT might be prohibitive regarding runtime. This PR introduces a new callback that will track instance terms that had their access points modified by PA, and add their nets to the list of nets to reroute. This way, we can run incremental iterations and only update the access points at the end of it, leaving GRT to reroute only the nets with modified APs.


eder-matheus:
@osamahammad21 FYI. I've tested this approach with ORFS by disabling pin access during DRT and making a second call to pin access in the end of GRT. This way we "mimic" what the incremental flow would do, and everything looks to work as expected.

PR #8222, opened 2025-09-03
Allow for MBFF tray identification on flops that have scan, preset, and clear pins.

hongted:


src/gpl/test/BUILD
```diff
@@ -59,3 +59,44 @@ filegroup(
...
+    "-Wno-sign-compare",
+    "-Wno-unused-parameter",
+    "-Wno-tautological-constant-out-of-range-compare",
+    "-Wno-ctad-maybe-unsupported",
+    "-Wno-nullability-extension",
+    "-fopenmp",
+]
+
+cc_test(
+    name = "mbff_test",
```
maliberty, line 157: It is problematic to have bazel only tests.

src/gpl/src/mbff.cpp
```diff
@@ -357,11 +356,41 @@ bool MBFF::IsClearPin(dbITerm* iterm)
   if (lib_port == nullptr) {
     return false;
   }
+
+  // Check the lib cell if the port is a clear.
```
maliberty: You have moved the comment to the wrong block of code.

src/gpl/test/BUILD
hongted: Adding the test to cmake is in progress.  Two prerequisite PRs will come first:

 https://github.com/The-OpenROAD-Project/OpenROAD/pull/8250
 https://github.com/The-OpenROAD-Project/OpenROAD/pull/8252

maliberty:
some conflicts to resolve

hongted:
> some conflicts to resolve

Resolved.  Rest will be fixed once I rebase onto the parent PR.

src/gpl/test/mbff_test.cpp
```diff
@@ -0,0 +1,137 @@
...
+  rsz::Resizer* rsz = ord->getResizer();
+
+  // Read tech.
+  tech.readLiberty("src/gpl/test/library/test/test0.lib");
+  tech.readLef("src/gpl/test/library/test/test0.lef");
+  sta->postReadLef(db->getTech(), db->findLib("test0"));
+
+  // Create empty chip and design.
+  odb::dbChip* chip = odb::dbChip::create(db, tech.getTech());
+  ord->readDef("src/gpl/test/library/test/test0.def",
```
maliberty: I don't think this makes sense as a c++ unit test.  You are just doing read_liberty/read_lef/read_def and introducing a whole program dependency into this test.

hongted: Agreed that this isn't really a unit test, but rather a C++ integration test which I think is still valuable to have.  

Ideally, we'd build some test functionality so that simple lefs and defs could be created in-sitsu without needing to create lef/def files and them read them in.  We'd also modularize the openroad class so that programs could utilize openroad as a library without bringing in all subcomponents.

This test was written to minimize the need for such changes and simply does the following:
1. Instantiate open road / open sta databases by reading liberty/lef/def and creating an empty chip/design
2. Instantiate mutiple mbffs flops 
3. Check that the mbff module is able to correctly identify those flops.

Let me know if you have any other preference on how to test this functionality.

maliberty: There is no advantage to an integration test in c++ over tcl.  The disadvantage is that it involves a large link step (equivalent to openroad itself).  I don't wish to encourage this direction.  I also find that using def files tends to encourage people to drop in overly large/unspecific integration tests in lieu of building proper tests.

I agree we should make utilities to make proper c++ unit testing easier.  You can see src/drt/test/fixture.* as an example for DRC checking tests.  I would rather make such changes than "minimize the need for such changes".

If you want this test in the short term then I would follow the established pattern for integration tests while we work on the better path.

hongted: The advantage of C++ vs tcl is the ability to test at a level lower then that provided by the public tcl api.  I see tcl as useful to testing end-to-end, while C++ enables testing at all levels.  As it stands, there is no way to perform the equivalent test of this PR in tcl without adding and exporting additional tcl functions.

For C++, we can certainly follow what drt does.  The issue there would be that we'd basically be proliferating stripped down versions of [class OpenRoad](https://github.com/The-OpenROAD-Project/OpenROAD/blob/6dadf2db4d0eab32fb53bb5168ba4f9cb6a27f2c/include/ord/OpenRoad.hh#L132).  

I think the natural consequence of that would be we'd eventually end up factoring said versions out and ending up with an equivalent class that can bring in any set of OpenRoad modules via dependency injection to solve that large link step.  If that's a direction you are ok with moving to, we can discuss what that'll look like.

maliberty: I think the main challenge is that it is painful to setup the tool individual in c++ ([example](https://github.com/The-OpenROAD-Project/OpenROAD/blob/024ae4397eb3c67b748d935c52a17e861e4fc92c/src/rsz/test/cpp/TestBufferRemoval.cc#L220)).  The Design object hides a lot of that for you.  

I'm playing with trying to simplify the process so that it would be easier to do this by avoiding the two phase setup (ctor + init -> ctor).  Then you would just put this simplified setup in a test fixture and use it for a suite of tests without needing Design to do it for you and all the dependencies that implies.  I'll need a few days to try.

hongted: Migrated to the test fixture.

src/OpenRoad.cc
```diff
@@ -3,6 +3,8 @@
 
 #include "ord/OpenRoad.hh"
 
+#include <tcl.h>
```
maliberty: #include "tcl.h"  per @hzeller

maliberty:
Merged master so I could see the changes here without the noise of the gui changes that are already merged.

src/gpl/test/mbff_test.cpp
```diff
@@ -0,0 +1,149 @@
...
+    return uut->IsValidTray(tray);
+  }
+};
+
+namespace {
+
+class MBFFTestFixture : public tst::Fixture
+{
+ protected:
+  void SetUp() override
```
maliberty, line 40: My thought is that we would create one fixture per tool that would handle all the dependency setup so that all the tests could share it.  Since there is just one test for gpl now this is ok but we should plan to split it out in the future.

PR #8224, opened 2025-09-04
gpl: support regions for routability mode

### LINKED ISSUE #6774: gpl: support regions for routability mode.
### Description

In the GPL code, NesterovBase consists of a class representing each region. We have the nbVec_ vector, which holds an object for each region. The first object in this vector (nbVec_[0]) contains all instances that do not belong to any specific region. For example, if a Verilog design is described without any defined regions, as is usually the case, GPL will hold all instances in nbVec_[0].

Currently, in routeBase.cpp, whenever we reference a NesterovBase object, we only do so for the object without regions (nbVec_[0]), while all other objects are ignored. For example:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/6c60401f98c6ec75633b3aa892456818e55fdeb2/src/gpl/src/routeBase.cpp#L763

I have not yet tested routability mode on a design with regions, but there is no way it would function correctly. However, I believe it will not necessarily throw an error.

### Suggested Solution

- Test what currently happens on a design with regions (e.g., upf_aes in OpenROAD/test/).
- Most importantly, replace every instance of nbVec_[0] in routeBase.cpp with a loop iterating over nbVec_ objects, and move forward from there.

### Additional Context

_No response_

LucasYuki:
Closes #6774.


gudeh:
it would be nice to see the placement progression with debug mode. You can generate a gif with:
`global_placement_debug -initial -generate_images -start_iter 5000`

test/upf_aes.tcl
```diff
@@ -23,7 +23,8 @@ make_tracks
 place_pins \
   -hor_layers met3 \
   -ver_layers met2
-global_placement -skip_initial_place -density uniform -routability_driven -routability_target_rc_metric 1.00
+global_placement -skip_initial_place -density uniform \
+  -routability_driven -routability_target_rc_metric 1.00
```
gudeh: why use 1.0 and not the 1.01 default? From previous experiments 1.0 we "aim for perfection" and spend unnecessary time inflating more.

LucasYuki: I used as base the test gpl/simple02-rd.tcl, in this test the value used is 1.0. 

I can change to the default value.

src/gpl/src/nesterovBase.h
```diff
@@ -1133,6 +1133,8 @@ class NesterovBase
 
   std::shared_ptr<PlacerBase> getPb() const { return pb_; }
 
+  odb::dbGroup* group() const { return pb_->group(); }
```
gudeh: getGroup()

gudeh:
I believe we should run a secure-CI for this change

src/gpl/src/routeBase.cpp
```diff
@@ -250,14 +251,49 @@ RouteBase::RouteBase(RouteBaseVars rbVars,
...
+                 57,
+                 "Target density at minimum routing congestion: {:.4f} ({})",
+                 minRcTargetDensity_[j],
+                 nbVec_[j]->group() ? "" : nbVec_[j]->group()->getName());
+    } else {
+      log_->info(GPL,
+                 58,
+                 "Target density at minimum routing congestion: {:.4f}",
+                 minRcTargetDensity_[j]);
+    }
```
gudeh: We don't need two different messages here, we can do something similar to this one: https://github.com/The-OpenROAD-Project/OpenROAD/blob/94ab17621aa8150675eee8703607482eab396dac/src/gpl/src/nesterovBase.cpp#L1684-L1687

src/gpl/src/routeBase.cpp
```diff
@@ -575,212 +613,216 @@ std::pair<bool, bool> RouteBase::routability(
...
+    prev_total_gcells_area[i]
+        = nbVec_[i]->getNesterovInstsArea() + nbVec_[i]->getTotalFillerArea();
+    prev_expected_gcells_area[i]
+        = inflatedAreaDelta_[i] + prev_total_gcells_area[i];
+
+    nbVec_[i]->cutFillerCells(inflatedAreaDelta_[i]);
+
+    // max density detection
+    if (nbVec_[i]->getTargetDensity() > rbVars_.maxDensity) {
+      string group_text;
```
gudeh: I believe we don't need this group_text variable

gudeh:
@LucasYuki noticed an important detail. When we revert to the minimal observed congestion we do so AFTER applying new inflations (but new inflations won't be applied since we have a `revertToMinCongestion()`). This should not make a difference, since we are going to update the inflations again with the `revertToMinCongestion()`. 

When he attempted to do the `revertToMinCongestion()` before applying new inflations, results changed, which is unexpected.

We are not including this changes in this PR, but this should be addressed! We might have something wrong with the `revertToMinCongestion()`, and/or the removal and revert of fillers might have an issue.

src/gpl/src/nesterovBase.h
gudeh: we may also change the placerbase version of group() to getGroup() also.

LucasYuki:
@gudeh FYI, I fixed the issue that you mentioned here.

The filler restoration method was placing the fillers back in the wrong order, I fixed this and included back the revertToMinCongestion() before applying new inflations.

gudeh:
Great! If you wish to apply this changes on this PR we should have a new run on the secure-CI with this also, we expect no changes on designs without regions.

LucasYuki:
> The filler restoration method was placing the fillers back in the wrong order, I fixed this and included back the revertToMinCongestion() before applying new inflations.

I checked again, and this didn't really fix the issue.
So I reverted these changes and the order of the revert checks.

This PR now only adds support for the routability mode without addressing those issues.
I ran secure-CI and this PR didn't change the results of any designs without regions.

test/upf_aes.ok
```diff
@@ -1,35 +1,36 @@
 [INFO ODB-0227] LEF file: sky130hd/sky130hd.tlef, created 13 layers, 25 vias
 [INFO ODB-0227] LEF file: sky130hd/sky130_fd_sc_hd_merged.lef, created 437 library cells
-[WARNING IFP-0028] Core area lower left (2.000, 2.000) snapped to (2.300, 2.720).
-[INFO IFP-0001] Added 497 rows of 2944 site unithd.
+[WARNING IFP-0028] Core area lower left (30.000, 30.000) snapped to (30.360, 32.640).
```
gudeh, line 3: Why do we have differences here?

LucasYuki: Because I also changed the initialize_floorplan arguments in `test/upf_aes.tcl`

maliberty:
@gudeh approve?

gudeh:
@LucasYuki sent me this gif with the new behavior for UPF. He also mentioned secure-CI showed no changes in circuits without regions, so everything looking good!

output

maliberty (approved):

PR #8228, opened 2025-09-04
GRT: Suggest adjustment value when having congestion problems

### LINKED ISSUE #6638: GRT: suggest adjustment value after congestion error
Related to https://github.com/The-OpenROAD-Project/OpenROAD/issues/6404

luis201420:
Calculate an ideal adjustment value when a design cannot be routed due to congestion.

This update will calculate whether the user-defined adjustment can be reduced to address congested areas.

Fixes #6638


src/grt/src/fastroute/include/FastRoute.h
```diff
@@ -151,6 +151,12 @@ class FastRouteCore
       const interval<int>::type& last_tile_reduce_interval,
       const std::vector<int>& track_space,
       bool release = false);
+  void saveResourcesBeforeAdjustments();
+  bool getSuggestAdjustment(int& suggested_adjustment);
```
eder-matheus: Maybe `computeSuggestedAdjustment` is a better name.

src/grt/src/fastroute/include/Graph2D.h
```diff
@@ -78,6 +78,8 @@ class Graph2D
                                bool stop_decreasing,
                                int& max_adj);
   void str_accu(int rnd);
+  void saveResources(int x, int y, bool is_horizontal);
+  bool getSuggestAdjustment(int x, int y, bool is_horizontal, int& adjustment);
```
eder-matheus: Ditto.

eder-matheus (changes requested):
There are GRT unit tests that require updates in the .ok files.

eder-matheus (approved):

PR #8230, opened 2025-09-04
odb: name mapping for chip objects

osamahammad21:


src/odb/src/db/dbChipInst.h
```diff
@@ -35,6 +36,7 @@ class _dbChipInst : public _dbObject
   dbId<_dbChip> parent_chip_;
   dbId<_dbChipInst> chipinst_next_;
   dbId<_dbChipRegionInst> chip_region_insts_;
+  std::unordered_map<std::string, dbId<_dbChipRegionInst>> region_insts_map_;
```
maliberty: Could the key be the region on the master chip rather than a string?  Look up the region on the master by name and then map it to the dbChipRegionInst

src/odb/src/db/dbChip.cpp
```diff
@@ -223,6 +226,10 @@ dbIStream& operator>>(dbIStream& stream, _dbChip& obj)
   if (obj.getDatabase()->isSchema(db_schema_chip_hash_table)) {
     stream >> obj._next_entry;
   }
+  auto chip = (dbChip*) &obj;
+  for (const auto& chip_region : chip->getChipRegions()) {
```
maliberty, line 230: Doesn't chipinsts_map_ need to be populated too?

osamahammad21: it is in the dbDatabase.cpp

src/odb/src/db/dbChipInst.h
osamahammad21: Yeah that makes more sense.

maliberty (approved):

PR #8234, opened 2025-09-04
update yosys version to 0.57

eder-matheus:
Should be merged after https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3471


maliberty (approved):
@sombraSoft do you need to do anything to update dev gcp machines?

PR #8245, opened 2025-09-05
dpl: repaint after rejected move at tryswap

gudeh:
This fixes a case not handled by the journal when undoing a move, which can leave the pixel grid inconsistent.

During a swap, we call addToMoveList() first for ndi, then for ndj. We commit ndi’s move (both the cell attribute and the pixel grid), which can overwrite pixels originally belonging to ndj. If ndj’s move is later rejected, those overwritten ndj pixels are not repainted, leaving the grid incorrect.

It may be possible to address this automatically with journal (it would need to be aware of overridden pixels), but this simple fix addresses issue https://github.com/The-OpenROAD-Project-private/OpenROAD-flow-scripts/issues/1203.

I would like to understand why it is a no-op, [as can be seen at the dashboard](https://dashboard.precisioninno.com/compare?sourceAType=Branch&sourceBType=Branch&sourceAName=nightly+%28nightly%29&sourceAID=5970&sourceBName=secure-dpl-repaint-after-rejected-move&sourceBID=1). Perhaps the issue only triggers under certain padding configuration?


maliberty (approved):

osamahammad21:
@gudeh Very nice catch. I think we should have a journal SwapAction to handle this cleanly.

PR #8247, opened 2025-09-05
test: MockArray flat and hierarchical flow

oharboe:


jhkim-pii:
Thank you for this PR!
Having the both flat and hierarchical flow tests will be definitely helpful.

jhkim-pii:
By the way, there is no such variant for "Element". Is this ok? 

```
orfs_flow(
    name = "Element",
        ...
        # We want to report power per module using hierarhical .odb
        "OPENROAD_HIERARCHICAL": "1",
```

oharboe:
> By the way, there is no such variant for "Element". Is this ok?
> 
> ```
> orfs_flow(
>     name = "Element",
>         ...
>         # We want to report power per module using hierarhical .odb
>         "OPENROAD_HIERARCHICAL": "1",
> ```

I think it is fine to test hier/no hier with a Element that is build with .hier as the flat array only sees the .lib/.lef files...?

I can easily update to build the Element flat as well, but I didn't think it added much in the way of testing value.

oharboe:
@jhkim-pii FYI to update rules.json https://github.com/The-OpenROAD-Project/OpenROAD/tree/master/test/orfs#updating-orfs-and-bazel-orfs

test/orfs/mock-array/BUILD
```diff
@@ -103,7 +103,12 @@ filegroup(
     visibility = [":__subpackages__"],
 )
 
-orfs_flow(
+VARIANTS = [
```
maliberty, line 106: Will these run serially or concurrently?

oharboe: parallel by default

maliberty (approved):

PR #8248, opened 2025-09-05
DPL: Remove the use of unordered_set in searchNearestSite

calewis:
By carefully picking which direction nodes can expand we can avoid needing to keep track of already visited nodes.  This avoids the overhead of keeping track of where we've already visited.

Also a minor optimization of not adding the center to the PQ, if the center is a match we can exit early and avoid PQ construction. 

## The expansion strategy goes as follows:

The center naturally expands in all 4 directions. 

Nodes in the same column as the center expand in 3 directs
- Up for nodes above the row of the center
- Down for nodes below the row of the center
- Left and right

Assume we are on the y-axis above the center, then we know the only way we could have reached this point is from below so we don't need to add our neighbor below us to the queue, opposite goes for the y-axis nodes below the center, they don't need to add the node directly above them.

Finally we get to off center off y-axis nodes, expanding only to the left or right.

Because we are expanding on a structured grid instead of an arbitrary graph we can avoid keeping track of visited nodes saving us the overhead of using the `unordered_set`.

### DIFF
--- a/src/dpl/src/Place.cpp
+++ b/src/dpl/src/Place.cpp
@@ -744,17 +744,29 @@ PixelPt Opendp::searchNearestSite(const Node* cell,
       return manhattan_distance == other.manhattan_distance;
     }
   };
+
+  // Check the first candidate to avoid unnecessary priority queue creation.
+  GridPt center{x, y};
+  if (canBePlaced(cell, center.x, center.y)) {
+    return PixelPt(
+        grid_->gridPixel(center.x, center.y), center.x, center.y);
+  }
+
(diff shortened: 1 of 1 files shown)


maliberty (approved):

maliberty:
test failures need addressing

maliberty:
I thought the goal here is to do the same search as would have done previously with better efficiency.  From the test failures it is apparent that we are not.  Is it possible to make this equivalent and avoid the need to do a big validation cycle?

maliberty:
(Also DCO needs fixing)

PR #8249, opened 2025-09-05
mpl: use single thread when running with graphics

AcKoucher:
Small partial revert of #8216 .

Graphics require single thread to run properly. Not running single thread causes problems when attempting to `pause()` the GUI - it won't pause.


AcKoucher:
@joaomai FYI.

eder-matheus (approved):

PR #8252, opened 2025-09-05
Added Abseil into DependencyInstaller.sh

hongted:


maliberty:
@vvbandeira I'll leave this to you to merge and update infra as needed.

vvbandeira:
@hongted,
I see you mentioned this PR and the bin/library separation as the two blockers for the MBFF change on GPL.
We plan to drop Ubuntu 20 and Debian 11 soon, so if this PR becomes the only blocker, ping me again.
To make CI pass right now, you will likely need to change the incantation to compile on Ubuntu 20 and Debian 11.

hongted:
> @hongted, I see you mentioned this PR and the bin/library separation as the two blockers for the MBFF change on GPL. We plan to drop Ubuntu 20 and Debian 11 soon, so if this PR becomes the only blocker, ping me again. To make CI pass right now, you will likely need to change the incantation to compile on Ubuntu 20 and Debian 11.

Incantation is now updated.

vvbandeira (approved):

PR #8265, opened 2025-09-08
Spec file generation flow for ArtNet

shypark98:
I have created PR.

Please let me know any comments.

Best,
Seonghyeon


src/par/test/openroad
maliberty, line 1: rm

src/par/test/run_artnetgen.tcl
```diff
@@ -0,0 +1,39 @@
...
+set spec_dir .
+set verilog_dir .
+
+if { ![file exists $spec_dir] } {
+    file mkdir $spec_dir
+}
+
+read_liberty asap7/asap7sc7p5t_AO_LVT_TT_nldm_201020.lib
+read_liberty asap7/asap7sc7p5t_AO_RVT_TT_nldm_201020.lib
+read_liberty asap7/asap7sc7p5t_AO_SLVT_TT_nldm_201020.lib
```
maliberty: Does this test actually need multiple Vts?  Could you use the existing .lib without adding any new ones?

src/par/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ add_library(par_lib
   src/KWayFMRefine.cpp
   src/KWayPMRefine.cpp
   src/PriorityQueue.cpp
+  src/ArtNetSpec.cpp
```
maliberty: You need to add this to the BUILD file as well for bazel.

src/par/test/run_artnetgen.tcl
maliberty, line 1: This test is not connected to the test runner for cmake or bazel

maliberty (changes requested):

src/par/test/run_artnetgen.tcl
shypark98: I added new testcase. It does not need multiple Vts.

src/par/CMakeLists.txt
shypark98: I added ArtNetSpec.cpp to BUILD file

src/par/test/run_artnetgen.tcl
shypark98: Was the issue resolved by adding ArtNetSpec.cpp to the BUILD file? Or is there anything else I need to do?

rovinski:
Side note: would recommend either a rebase or merge by squash so that the hefty asap7 libs that were added and then deleted do not get committed to the repo

PR #8273, opened 2025-09-09
test/orfs/mock-array: test IO_CONSTRAINTS and MACRO_PLACEMENT_TCL

oharboe:


oharboe:
@maliberty This test write IO/macro placement + use of these.

oharboe:
@maliberty unrelated CI failure on pr-merge

image

maliberty:
Why do these need to be part of mock-array and not an ordinary unit test that would run much faster.  What additional coverage does this provide?

oharboe:
> Why do these need to be part of mock-array and not an ordinary unit test that would run much faster.  What additional coverage does this provide?

Easy to articulate this way and it tests reading as well as writing of these files for routing by abutment and IO pins on multiple layers.

maliberty:
| Easy to articulate this way and it tests reading as well as writing of these files for routing by abutment and IO pins on multiple layers.

All those things can be unit tested.  This adds a lot of runtime overhead.  I view flow tests as testing the interaction across a full flow not testing the functionality of a single module.

oharboe:
Tests also need to be cheap to articulate. 

Do you already have this coverage in unit tests?

I think runtime concerns/cost of developing tests are best addressed by having a post merge test set.

things that fail frequently are worth the effort to develop unit tests for.

oharboe:
I see that the bazel tests and build finish sooner than other CI work by at least 10 minutes, so there is no visible slowdown now.

oharboe:
@maliberty Merge?  As long as this is well under the other running times in the CI, can we merge this to cover more use-cases?

maliberty:
Its not just about performance.  This isn't a flow test.  Using large designs to test small bits of functionality is not a good methodology.

oharboe:
> Its not just about performance. This isn't a flow test. Using large designs to test small bits of functionality is not a good methodology.

Ideally yes, but do you have these tests today?

Integration tests are a result of accepting that fast unittests for everything and having a mix of fast unittests + integration is a false choice.

maliberty:
If we lack tests then they should be added as unit tests.   We do have tests for write_pin_placement (in odb) and write_pin_placement (in ppl)

PR #8277, opened 2025-09-09
mpl util.h: use less ambiguous name.

hzeller:
The "util.h" has a very generic name and it is included not relative to project root, but relative to current files, so relies on the include search path.

Make this a more unique name and use mpl-util.h instead.

While at it, fix another "util.h" ambiguity in the rcx header directory. Here, we can dismabiguate by adding the "rcx/" prefix as it should be.


hzeller:
the CI failure looks unrelated.

maliberty:
I wonder if it wouldn't be better to press on towards the conversion to bazel and the use of full paths.  There are many other such cases and we would have to undo all this at the end.  

```
 find src/*/[is]* -type f|sed -e 's@.*/@@'|sort|uniq -d|wc -l
30
```

hzeller:
I think there is still no need to undo these names, as they are better even if looked at it alone.

This particular one was was standing out as `@cudd` unfortunately also exports a `"util.h" without a needed include path, so it was ambiguous with that.

w.r.t the bazel conversion: I think what is left is mostly mentioned in the WORKSPACE file and possibly @oharboe probably also has some overview what to do. I am only having spare-time time to contribute unfortunately ...

oharboe:
Emptying WORKSPACE is the big missing feature for my use case.

I mostly care about improved dev flow, testing.

bazel-orfs uses the docker image.

maliberty:
I don't view WORKSPACE as gating though it is good to have.  @QuantamHD is working on python support which I do view as gating since we have many python tests that I don't wish to lose.

Assuming we used full paths, why would mpl-util.h be be better as the full path would already contain mpl.  It would just be redundant.

hzeller:
(I am not opposed to renaming the header later. But for now, having non-ambiguous names is probably a good idea.)

BTW, we're not bound to a full bazel conversion _now_ to use full names, they would just look slightly annoying right now `src/foo/include/foo/flub.h`. Or maybe leave out the toplevel `src` for a `foo/include/foo/flub.h`

It would be be possible now and probably requrires a little bit of cmake tweaking (having it have a `-Isrc`).

Long term, we'd probably want to rename src/ to openroad/ and flatten the include structure a bit, so that we have
`openroad/foo/flub.h`. The visibility of the implementation and public headers would be fully dealt with by bazel.

oharboe:
I care about emptying WORKSPACE so that I can add OR as a bazel_dep() from other projects

maliberty:
> I care about emptying WORKSPACE so that I can add OR as a bazel_dep() from other projects

I said it isn't gating not that I don't care.

maliberty:
-Isrc would be a bad idea in cmake as it is not hermetic and you now allow arbitrary inclusion without any expressed dependency.

maliberty:
I want to know what your plan is going forward.  Are you planning to rename 30+ headers or just util.h?

oharboe:
I was just clarifying how it is useful to me and for what use case, nothing else implied.

emptying WORKSPACE is not a high priority item for me.

docker + bazel-orfs works very well and to get rid of docker is a much bigger task, it includes all binaries, up to and including klayout.

locally, bazel works *great* with OR dev flows for me.

hzeller:
Just util.h

maliberty (approved):

PR #8283, opened 2025-09-09
abc: Enable usage of rmp library for custom implementations.

mikesinouye:
`abcInit()` needs to be called in rmp implementations. However, abcInit() is defined in Main.cc, so it is not possible to link the openroad_lib and :rmp libraries that exclude Main.cc in a different binary with a custom rmp implementation.

This PR moves the abc start and stop logic to OpenRoad.cc. abc already carries statics so the abc initialized static is not a new limitation for binaries that use multiple OpenRoad objects.

The PR also exposes some public rmp headers needed for 3rd party implementations.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -89,7 +89,6 @@ static const char* log_filename = nullptr;
 static const char* metrics_filename = nullptr;
 static bool no_settings = false;
 static bool minimize = false;
-static bool abc_initialized = false;
 
 static const char* init_filename = ".openroad";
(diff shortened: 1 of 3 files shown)


maliberty:
Addressed in #8275

maliberty:
If there are more changes needed after that PR then please reopen after merging.

mikesinouye:
Ah, I had made this yesterday afternoon and missed that. Thanks.

PR #8285, opened 2025-09-09
Updated DependencyInstaller.sh env.sh script to support ZSH.

hongted:


oharboe (commented):
This idiom is endemic, what about other cases?

maliberty:
The script has `#!/usr/bin/env bash` - are you substituting zsh for bash?

vvbandeira:
@oharboe, indeed, there are multiple places where we assume `bash` as the default shell, but I am not particularly against accepting PRs that expand the support to other shells -- we have done so in the past, in ORFS, I think.

Users just need to be clear that we support and test `bash`, and other shells would be unofficially supported.

@maliberty, the changes relate to the `env.sh` file that the `DependencyInstaller.sh` generates. Users can source `env.sh` to get the correct `PATH`/`LD_LIBRARY_PATH`, etc that are needed to compile OR correctly.

hongted:
> @oharboe, indeed, there are multiple places where we assume `bash` as the default shell, but I am not particularly against accepting PRs that expand the support to other shells -- we have done so in the past, in ORFS, I think.
> 
> Users just need to be clear that we support and test `bash`, and other shells would be unofficially supported.
> 
> @maliberty, the changes relate to the `env.sh` file that the `DependencyInstaller.sh` generates. Users can source `env.sh` to get the correct `PATH`/`LD_LIBRARY_PATH`, etc that are needed to compile OR correctly.

Yes, the env.sh script is source'd rather than run has has a dependence on the shell the user is running under.  This PR is to extend it so that it can either be sourced in bash or zsh.  Other scripts that are run (with a shebang line) should work independent of the shell the user is running it under.

If there are other similar scripts that are source'd and have a dependence on the user shell , please point them out to me and I'd be happy to test and update if necessary.

etc/DependencyInstaller.sh
```diff
@@ -289,7 +289,12 @@ _installCommonDev() {
     if [[ ! -z ${PREFIX} ]]; then
         # Emit an environment setup script
         cat > ${PREFIX}/env.sh <<EOF
-depRoot="\$(dirname \$(readlink -f "\${BASH_SOURCE[0]}"))"
+if [ -n "\$ZSH_VERSION" ]; then
+  depRoot="\$(dirname \$(readlink -f "\${(%):-%x}"))"
+else
```
oharboe, line 294: Maybe this could work and be even more shell independent?

dash, bash, zsh...

    depRoot="$(dirname -- "$(readlink -f "${BASH_SOURCE:-${(%):-%x:-$0}}")")"

hongted: It could, but I don't have access to dash to see how the third branch would work.

```
depRoot="$(dirname -- "$(readlink -f $0)")"
```

oharboe:
```
$ pwd
/home/oyvind
$ cat foo.sh
depRoot="$(dirname -- "$(readlink -f $0)")"
echo $depRoot
$ dash foo.sh
/home/oyvind
$ bash foo.sh
/home/oyvind
```

vvbandeira: `$0` is only POSIX-compliant across shells if the script is executed, not sourced.

Using the example above and just adding another echo to show `$0`:
```
$  cat env.sh
depRoot="$(dirname -- "$(readlink -f $0)")"
echo $depRoot
echo $0
$  realpath env.sh
/home/vvbandeira/etc/env.sh
```

If we source from the same folder that the script is in, it "works" but is actually a false positive.
```
$  . ./env.sh
/home/vvbandeira/etc
bash
```

If we try to source from any other folder, we get the wrong path:
```
$  pwd
/home/vvbandeira
$  . ./etc/env.sh
/home/vvbandeira
bash
```

This happens because `$SHELL $FILE` spawns a new child process, thus setting ⁠`$0` to `$FILE`.However, when you source with `. $FILE` the value of⁠`$0` remains the original caller (your shell or parent script). Unfortunately, there is no POSIX-compliant way across shells to get the correct path to the file that is being sourced, and each shell does it differently.

For example, using `BASH_SOURCE` or whatever shell-specific idiom:
```
$  cat env.sh
depRoot="$(dirname -- "$(readlink -f ${BASH_SOURCE[0]})")"
echo $depRoot
echo $0
$  realpath env.sh
/home/vvbandeira/etc/env.sh
```

Still works if we are in the same folder.
```
$  pwd
/home/vvbandeira/etc
$  . ./env.sh
/home/vvbandeira/etc
bash
```

But now it also works from any folder in the system.
```
$  pwd
/home/vvbandeira
$  . ./etc/env.sh
/home/vvbandeira/etc
bash
```

PR #8287, opened 2025-09-10
bazel: creature comforts

oharboe:
Needs https://github.com/The-OpenROAD-Project/OpenROAD/pull/8288 merged first

Minimize surprises in bazel usage.

This will build all dependencies, such as binaries and ORFS build targets and run all tests and is normally what the user should do:

    bazelisk test ...

Most users should run a faster set of tests localized to the module of interest and then go directly to creating a PR to run a full set of tests, possibly iterating later on specific failures locally if the PR fails.

This is the fastest option as it builds only the exec version of the build and runs only tests that are likely to fail based upon module changed:

    bazelisk test src/gpl/...

This will build all binaries in the target configuration(tests use the exec configuration), which is not normally something a developer would want if they want to run tests afterwards. However, bazelisk build is a natural choice when working specifically on build of the code and not testing changes:

    bazelisk build ...


oharboe:
@maliberty Ready for review as such, but needs https://github.com/The-OpenROAD-Project/OpenROAD/pull/8288

oharboe:
@maliberty Ready for review and local testing

maliberty:
Thanks

PR #8293, opened 2025-09-10
ANT: avoid checking the positioning of special nets

luis201420:
This modification will avoid checking the special nets.

Fixes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3476

It also prevents returning ant violations when the number of diodes required is 0.

### DIFF
--- a/src/ant/src/AntennaChecker.cc
+++ b/src/ant/src/AntennaChecker.cc
@@ -1095,15 +1095,28 @@ Violations AntennaChecker::getAntennaViolations(odb::dbNet* net,
 
 bool AntennaChecker::designIsPlaced()
 {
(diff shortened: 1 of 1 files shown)


src/ant/src/AntennaChecker.cc
```diff
@@ -1095,15 +1095,28 @@ Violations AntennaChecker::getAntennaViolations(odb::dbNet* net,
 
 bool AntennaChecker::designIsPlaced()
 {
+  if (db_->getChip() == nullptr) {
```
eder-matheus: This check seems unnecessary, since the function will only be called after initializing `block_`. Please remove

PR #8308, opened 2025-09-11
Refactor(Binary): Specify binary path

### LINKED ISSUE #8303: The target location of `openroad`
### Description

This is a nit-picky proposal.
So the current `openroad` after `./etc/Build.sh`(before `make install`) is located at `build/src` while `build/bin` is empty which is a bit awkward.

### Suggested Solution

So I suggest just add 
```
set_target_properties(openroad PROPERTIES
  RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
)
```
if you don't mind.

### Additional Context

_No response_

wjrforcyber:
Closes #8303 .

### DIFF
--- a/docker/Dockerfile.binary
+++ b/docker/Dockerfile.binary
@@ -12,6 +12,6 @@ FROM $copyImage AS copyfrom
 
 FROM $fromImage
 
-COPY --from=copyfrom /OpenROAD/build/src/openroad /usr/bin/.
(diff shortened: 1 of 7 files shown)


maliberty:
You'll need to update the regression script to match
```
08:54:16  Run regression test on openroad/ubuntu20.04-builder-gcc:17eb1b
08:54:16  bash: ./build/src/openroad: No such file or directory
```

maliberty (approved):

PR #8310, opened 2025-09-11
CTS: create a new command for CTS configuration

### LINKED ISSUE #8159: CTS: reduce clock_tree_synthesis command arguments
### Description

CTS command `clock_tree_synthesis` has 27 arguments, most of them are configuration parameters, .

Separating the configuration parameters into a different command could help with legibility and an easier understanding of the configuration options.

### Suggested Solution

Have a separate command from `clock_tree_synthesis` to configure CTS parameters. Like suggested in [PR #7708](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7785#issuecomment-3091454558)

### Additional Context

_No response_

luis201420:
Fixes #8159 

Creating new `set_cts_config` command which will work as CTS configuration command.

The CTS key parameters are added to the new command:
```
-apply_ndr
-branching_point_buffers_distance
-buf_list
-clustering_exponent
-clustering_unbalance_ratio
-delay_buffer_derate
-distance_between_buffers
-library
-macro_clustering_max_diameter
-macro_clustering_size
-num_static_layers
-root_buf
-sink_buffer_max_cap_derate
-sink_clustering_levels
-sink_clustering_max_diameter
-sink_clustering_size
-tree_buf
-wire_unit
```
The command can be used to add settings on a single line:
```
set_cts_config -library LIBRARY -wire_unit UNIT
```
Or it can also be used to add configurations on multiple separate lines:
```
set_cts_config -library LIBRARY
set_cts_config -wire_unit UNIT
```


maliberty:
@arthurjolo please do a first pass review

oharboe:
I think this is the wrong approach.

A generic tcl/python side idiom should be found for all commands to deal with long list of arguments, idiomatic for EDA stages.

https://github.com/The-OpenROAD-Project/OpenROAD/issues/8159#issuecomment-3283732619

arthurjolo:
> I think this is the wrong approach.
> 
> A generic tcl/python side idiom should be found for all commands to deal with long list of arguments, idiomatic for EDA stages.
> 
> [#8159 (comment)](https://github.com/The-OpenROAD-Project/OpenROAD/issues/8159#issuecomment-3283732619)

I see this PR being very similar to your suggestion in [#8159](https://github.com/The-OpenROAD-Project/OpenROAD/issues/8159#issuecomment-3283732619), configuration is done via tcl/python and it is, in my opinion, more explicit and readable with the separate command.

Maybe I am misinterpreting your comment but it seems this PR is inline with what you suggested, all parameters are set in TCL but instead of having a variable passed to `clock_tree_synthesis` we use a configuration command.

src/cts/src/TritonCTS.tcl
```diff
@@ -36,6 +36,109 @@ proc configure_cts_characterization { args } {
...
+                                       [-sink_clustering_levels levels] \
+                                       [-sink_clustering_max_diameter] \
+                                       [-sink_clustering_size] \
+                                       [-tree_buf buf] \
+	                               [-wire_unit unit]
+}
+proc set_cts_config { args } {
+
+  sta::parse_key_args "set_cts_config" args \
+    keys {-wire_unit -sink_clustering_size \
```
arthurjolo: I think `-buf_list` and `-root_buf `should be in this new command.

precisionmoon:
> > I think this is the wrong approach.
> > A generic tcl/python side idiom should be found for all commands to deal with long list of arguments, idiomatic for EDA stages.
> > [#8159 (comment)](https://github.com/The-OpenROAD-Project/OpenROAD/issues/8159#issuecomment-3283732619)
> 
> I see this PR being very similar to your suggestion in [#8159](https://github.com/The-OpenROAD-Project/OpenROAD/issues/8159#issuecomment-3283732619), configuration is done via tcl/python and it is, in my opinion, more explicit and readable with the separate command.
> 
> Maybe I am misinterpreting your comment but it seems this PR is inline with what you suggested, all parameters are set in TCL but instead of having a variable passed to `clock_tree_synthesis` we use a configuration command.

Having set_cts_config also makes it applicable to multiple commands like clock_tree_synthesis and repair_clock_nets.  For example, if clock buffers are specified with set_cts_config, the same buffer list should be used for both CTS and repair_clock_nets.  This is similar to how set_opt_config applies to repair_timing and repair_design.    We need 3 new commands:

1) set_cts_config
2) reset_cts_config
3) report_cts_config

luis201420:
Two new commands added: `report_cts_config` and `reset_cts_config`

The `reset_cts_config` can be used without parameters to reset all configurations, or you can specify which configuration you want to reset using the same flags as `set_cts_config`.

arthurjolo (approved):

src/cts/src/TritonCTS.tcl
arthurjolo: Oh, I think it got the old comment, but this was resolved.

maliberty:
I think it is better to discuss the general approach to arguments in #2745 where there is already a lot of context.   This is consistent with similar commands elsewhere so I don't want to hold up this PR.  If we decide to take a different approach then many places will need updating (including this).

maliberty (approved):

precisionmoon:
Here are comments from Claude code.  We must fix the delay buffer selection bug.

The direction is good. Replacing (int)((worseDelay_ - arrival) / bufferDelay_) with a real DP over a heterogeneous buffer set is the right call, and propagating dlyNeeded up the tree instead of the old nBuffInsert + maxArrival pair is a genuine simplification. The zero-delay buffer filter is well-reasoned, and TestDelayBufList asserts against values read back from Liberty rather than pinning magic strings.

But the last commit broke the headline feature and left two tests that must fail. 9ea1092bc "cts: use only clk delay buffers" removed the population of the DLY-family candidate lists but left everything downstream. TechChar::isDlyCell() — the function that recognizes DEL/DLY/dlygate — is defined and never called. footprintDly and nameDly are declared, never written, and their two else if branches are dead.

The consequence is mechanical. src/cts/test/ihp-sg13g2 symlinks to the IHP library, whose delay cells carry cell_footprint : "DLY1". The surviving check is footprint.find("clkdly"), which "DLY1" doesn't match, and the cell names contain neither CLKDLY nor CLKDEL. So DelayBufListDlyCellTest.AppendsInferredDelayCells and LibraryOptionRestrictsInference both fail — and the test's own comment says exactly what was deleted: "tagged with a DLY* cell_footprint." Net effect: "identifies dedicated delay cells via footprint or naming" now works only for libraries with a literal lowercase clkdly footprint. Related, isClkDlyCell is case-sensitive while the name matching uses containsIgnoreCase — so even an uppercase CLKDLY footprint misses.

Three of the review comments marked as addressed are still open: no null checks in getDrvrResistance/getinternalDelay/computeBufferDelay (now a public API, so the caller-side invariant that protects them is no longer local); (drvrRes - prevDrvrRes) / drvrRes still divides by a possibly-zero resistance; resizer_->dontUse() still unguarded while cts_unittest.cc constructs TechChar with a null resizer.

Two logic bugs I'd want fixed regardless. The convergence loop while (nBufs < prevNBufs) exits on a non-decrease — and at that moment dpResult holds the non-improving solve, so it returns the longer chain and throws away the better previous one. And computeWireLumpedDelay declares sta::LibertyPort *input, *output; uninitialized, calls bufferPorts, then tests if (input) — bufferPorts doesn't write its out-params for a non-buffer cell, so that guard reads indeterminate stack.

The test gap is the thing I'd push hardest on. Eight goldens churned and buffer counts move both directions (gated_clock3 12→15, array_max_wl 1→0, lvt_lib 0→1), but nothing asserts residual latency mismatch — the only number that says whether this is better. solveDP has no unit test at all, and that's where the risk lives. skip_nets also now prints "Not possible to insert buffers" eight times where it previously balanced; that may be correct restraint, but it should be in the PR description with numbers rather than buried in golden churn.

The file has everything with file:line, plus a "what's good" section and a suggested merge order — B1 first, since nothing else is evaluable until CI is green.

arthurjolo:
Hi @precisionmoon looks like you meant to send this comment on PR: https://github.com/The-OpenROAD-Project/OpenROAD/pull/11314.

PR #8314, opened 2025-09-12
Enable output log checking with bazel test.

hongted:
Currently bazel test does not perform the same checks as ctest.  This PR updates regressions under bazel to write and diff output logs to checked-in golden references.

Currently WIP.

### DIFF
--- a/src/dst/test/BUILD
+++ b/src/dst/test/BUILD
@@ -21,6 +21,9 @@ cc_test(
     includes = [
         "../src",
     ],
+
+    # TODO(https://github.com/The-OpenROAD-Project/OpenROAD/issues/8599): re-enable once issue is resolved.
+    tags = ["manual"],
     deps = [
         ":test_helpers",
         "//src/dst",
@@ -43,6 +46,8 @@ cc_test(
     includes = [
         "../src",
     ],
+    # TODO(https://github.com/The-OpenROAD-Project/OpenROAD/issues/8599): re-enable once issue is resolved.
+    tags = ["manual"],
     deps = [
         ":test_helpers",
         "//src/dst",
@@ -66,6 +71,9 @@ cc_test(
     includes = [
         "../src",
     ],
+
+    # TODO(https://github.com/The-OpenROAD-Project/OpenROAD/issues/8599): re-enable once issue is resolved.
+    tags = ["manual"],
     deps = [
         ":test_helpers",
         "//src/dst",


maliberty:
I just noticed this earlier today when looking at @hzeller 's PR.  Thanks for addressing it.

hzeller:
What is the status here ? I think it would be great to have (as I typically use bazel to compile and test and almost never cmake, so it is important to me that the bazel tests are actually testing and comparing fully everything).

maliberty:
Its a draft so I haven't looked at it.

hongted:
Current status is still WIP.  A couple of changes are needed to get output log checking to pass as some tests have a very tight dependence on the exact log output so changes in test run location and compilers make a difference.

Changes todo
1. Sanitize output logs to reduce test dependence on trivial changes
2. Migrate to GCC from LLVM to confirm that test differences are due to a compiler difference

A wider discussion would be if there is a better way to check the behavior of the tests.

src/rsz/test/repair_tie11_hier.ok
```diff
@@ -40,5 +40,9 @@ Hierarchical pins
...
 0.0 0.0
 0.0 0.1
-No differences found.
-No differences found.
+Differences found at line 7.
+ wire net2;
+ wire net;
+Differences found at line 9.
+    - alu0_1 TIEHIx1_ASAP7_75t_R + SOURCE TIMING + PLACED ( 0 0 ) N ;
+    - alu0_2 TIEHIx1_ASAP7_75t_R + SOURCE TIMING + PLACED ( 0 0 ) N ;
```
maliberty: There should be no diffs in the .ok files

MODULE.bazel
```diff
@@ -111,19 +110,19 @@ bazel_dep(name = "googletest", version = "1.16.0", dev_dependency = True)
 bazel_dep(name = "glpk", version = "5.0.bcr.3", repo_name = "org_gnu_glpk")
 
 ## Fix the compiler version and avoid any local compiler
-bazel_dep(name = "toolchains_llvm", version = "1.5.0")
+#bazel_dep(name = "toolchains_llvm", version = "1.5.0")
```
maliberty: Why is this commented (similar below)?

hongted: This is WIP and intended to enable testing with system-level gcc instead of llvm to minimize differences due  to the compiler difference.

src/rsz/test/repair_tie11_hier.ok
hongted: Understood, this is WIP as there are differences between bazel test and cmake test that need to be minimized and or waived.

MODULE.bazel
hongted: Reverted and marked tests as manual.

hongted:
This is finally ready for review.

src/dbSta/BUILD
```diff
@@ -143,10 +143,9 @@ cc_library(
 
 tcl_encode(
     name = "tcl",
-    srcs = [
+    srcs = ["//src/sta:tcl_scripts"] + [
         "src/dbReadVerilog.tcl",
         "src/dbSta.tcl",
-        "//src/sta:tcl_scripts",
     ],
```
maliberty, line 149: Why split this into two arrays?

maliberty:
Please open an issue will all the tests that need to be re-enabled so we can track the burn down.

src/dbSta/BUILD
hongted: That's because src/dbSta.tcl re-defines sta_warn and sta_error so it needs to be loaded after src/sta:tcl_scripts.

https://github.com/The-OpenROAD-Project/OpenROAD/blob/c62be3f9526a0f925c4c996ad4d99193c8f820d3/src/dbSta/src/dbSta.tcl#L92-L99

Splitting it into two arrays is there so that buildifier doesn't reorder them when auto-formatting the file.

maliberty: An explanatory comment would be good as that is subtle.

hongted:
https://github.com/The-OpenROAD-Project/OpenROAD/issues/8610 filed to track follow-up.

src/dbSta/BUILD
hongted: Done in https://github.com/The-OpenROAD-Project/OpenROAD/pull/8706

PR #8319, opened 2025-09-12
bazel: tests should have _test suffix

oharboe:
https://bazel.build/versions/7.3.0/build/style-guide#target-naming


hzeller:
LGTM

PR #8327, opened 2025-09-13
dbTypes: Define trivial constructors directly in header.

hzeller:
Make it easier to read for humans and enable the compiler to see that these are trivially copyable types.


maliberty (approved):

hzeller:
I did another push (had do undo number literal readability improvements as apparently some swig-tooling could not handle it in the CI). So maybe another auto-merge stamp is needed.

src/odb/include/odb/dbTypes.h
```diff
@@ -207,22 +208,22 @@ class dbGroupType
   /// The explicit orientation must be a string of one of the
   ///  following values: "PHYSICAL_CLUSTER", "VOLTAGE_DOMAIN".
   ///
-  dbGroupType(const char* type);
+  dbGroupType(const char* orient);
```
gadfort: How come this changed from type to orient?

hzeller: I came across it while editing. The name in the *.cpp file for this is "orient", so adpted them to be the same.
(clang tidy warning about inconsistent names in declaration and implementation)

gadfort: The cpp file is probably incorrect, the other constructors for this class all use type.

maliberty: Agreed - orient is wrong.

hzeller: Alright, renamed in the compilation unit instead to `type`.

PR #8336, opened 2025-09-15
python: Add pypi package infrastructure to bazel

QuantamHD:
There is now infrastructure to build python wheels that we could upload to pypi. This rule can be built with 

```
bazelisk build -c opt --features=thin_lto //openroad:openroad_wheel
```
 
This is a precursor to removing the -python option from the OpenROAD binary, and adopting a more standard pypi distribution.
    
This will require CI changes to build the wheel on manylinux Docker images.
    
### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -131,8 +131,7 @@ OPENROAD_DEFINES = [
     "OPENROAD_GIT_DESCRIBE=\\\"bazel-build\\\"",
     "BUILD_TYPE=\\\"$(COMPILATION_MODE)\\\"",
     "GPU=false",
-    "BUILD_PYTHON=true",
-    "ENABLE_PYTHON3",
+    "BUILD_PYTHON=false",
     "ABC_NAMESPACE=abc",
     "TCLRL_VERSION_STR=",
 ]
@@ -141,15 +140,6 @@ cc_binary(
     name = "openroad",
     srcs = [
         "src/Main.cc",
-        "src/OpenRoad.cc",
-        ":openroad_swig",
-        ":openroad_swig-py",
-        ":openroad_py",
-        ":openroad_tcl",
-        ":rmp_swig",
-        ":rmp_swig-py",
-        ":rmp_python",
-        ":rmp_tcl",
         "//bazel:runfiles",
     ],
     copts = OPENROAD_COPTS,
@@ -167,12 +157,7 @@ cc_binary(
         "@boost.stacktrace",
         "@rules_cc//cc/runfiles",
         "@tk_tcl//:tcl",
-        "@openroad_rules_python//python/cc:current_py_cc_headers",
-        "@openroad_rules_python//python/cc:current_py_cc_libs",
     ],
-    data = [
-        "@python_3_13//:py3_runtime",
-    ]
 )
 
 GUI_BUILD_FLAGS = select(
@@ -209,18 +194,8 @@ cc_library(
         "src/Tech.cc",
         "src/Timing.cc",
         ":openroad_swig",
-        ":openroad_swig-py",
-        ":openroad_py",
         ":openroad_tcl",
-        ":rmp_swig",
-        ":rmp_swig-py",
-        ":rmp_python",
-        ":rmp_tcl",
-    ] + glob([
-        "src/rmp/src/*.cpp",
-        "src/rmp/src/*.h",
-    ]),
-    hdrs = glob(["src/rmp/include/rmp/*.h"]),
+    ],
     copts = OPENROAD_COPTS,
     defines = OPENROAD_DEFINES + GUI_BUILD_FLAGS,
     features = ["-use_header_modules"],
@@ -294,6 +269,55 @@ tcl_wrap_cc(
     ],
 )
 
+# This target compiles the SWIG C++ wrapper into a shared library (.so)
+# that Python can load dynamically.
+cc_binary(
+    name = "_openroadpy.so",
+    srcs = [":openroad_swig-py"],
+    linkshared = True,
+    deps = [
+        ":openroad_lib",  # Depends on the core odb C++ library
+        ":ord",
+        "//src/odb",
+        "//src/ifp",
+        "//src/utl",
+        "//src/gpl",
+        "//src/ant",
+        "//src/cts",
+        "//src/dpl",
+        "//src/drt",
+        "//src/exa",
+        "//src/fin",
+        "//src/grt",
+        "//src/par",
+        "//src/pdn",
+        "//src/ppl",
+        "//src/psm",
+        "//src/rcx",
+        "//src/stt",
+        "//src/gui",
+        "//src/tap",
+        "@boost.stacktrace",
+        "@openroad_rules_python//python/cc:current_py_cc_headers",
+    ],
+)
+
+# This packages the SWIG-generated Python wrapper (odb.py) and the
+# compiled C++ extension (_odb.so) together.
+py_library(
+    name = "ord_py",
+    srcs = [":openroad_swig-py"],  # Use the .py output from the swig-py rule
+    # The data attribute makes the .so file available at runtime.
+    data = [":_openroadpy.so"],
+    deps = [
+        "//src/utl:utl_py",
+        "//src/odb:odb_py",
+    ],
+    # This allows imports relative to the workspace root.
+    imports = ["."],
+    visibility = ["//visibility:public",]
+)
+
 python_wrap_cc(
     name = "openroad_swig-py",
     srcs = [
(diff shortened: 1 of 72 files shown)


maliberty:
@vvbandeira please review/comment re:
`This will require CI changes to build the wheel on manylinux Docker images.`

maliberty:
Is --features=thin_lto  required?

QuantamHD:
It's not required, but it helps with wheel size a lot. By default pypi caps each wheel at 100MB, and we're 345MB without LTO+Opt and 45MB with LTO+Opt. I can try to see what it's like with just opt.

donn:
Hi Ethan, could I ask you to run this CI on GitHub so arm64 and macOS wheels are built as well?

Here's the runner info:

https://github.com/YosysHQ/yosys/blob/09742e27f797cba49dacca1cdbebd546d9d782c3/.github/workflows/wheels.yml#L14

maliberty:
We do not support mac as we have no hardware on which to debug issues.  Our CI runs on Jenkins/GCP not GHA so I don't see that request as possible.

donn:
Then I would kindly ask you to leave the embedded interpreter as-is. Or at least a setup.py/pyproject.toml to independently build the requisite wheels outside of Bazel.

maliberty:
If you (@donn) can do all the bring up and support to get it going we will take the changes.  Mac remains a best-effort platform and I can't guarantee resources to work on it.

We will not gate the move to Bazel on mac support and may remove -python when Linux platforms are working.

donn:
Is the long-term plan to move all building to Bazel?

QuantamHD:
Yes, we're going to drop cmake as soon as we're able. Python being the last major hurtle

gadfort:
> It's not required, but it helps with wheel size a lot. By default pypi caps each wheel at 100MB, and we're 345MB without LTO+Opt and 45MB with LTO+Opt. I can try to see what it's like with just opt.

they also cap projects at 10GB (you can request more), so that would get used up fairly quickly. @QuantamHD at what interval would you be expecting it be uploaded?

QuantamHD:
Google is a PSF sponsor so I think we could get that bumped pretty easily. I'm imagining weekly or perhaps once every 2 weeks. Even at the 10GB limit we're looking at about 4 years of weekly releases or 8 years of twice a month releases.

maliberty:
I'm surprised LTO shrinks the size that much.  How does it impact build time?

src/Exception-py.i
```diff
@@ -29,11 +49,21 @@
...
       }
     }
+    #endif
 
     PyErr_SetString(PyExc_RuntimeError, excp.what());
     SWIG_fail;
   }
   catch (...) {
+    #ifdef BAZEL
+    Logger* logger = Logger::defaultLogger();
```
maliberty, line 59: Why just for Bazel?

BUILD.bazel
```diff
@@ -69,7 +70,9 @@ OPENROAD_LIBRARY_DEPS = [
     "//src/exa",
     "//src/exa:ui",
     "//src/fin",
+    # "//src/fin:ui",
```
maliberty: Are these comments for some future purpose?

BUILD.bazel
```diff
@@ -262,6 +269,92 @@ tcl_wrap_cc(
     ],
 )
 
+# This target compiles the SWIG C++ wrapper into a shared library (.so)
+# that Python can load dynamically.
+cc_binary(
```
maliberty, line 269: I would have thought this would be a cc_library rule to produce a .so.

src/cts/BUILD
```diff
@@ -47,8 +48,10 @@ cc_library(
     ],
     hdrs = [
         "include/cts/TritonCTS.h",
+        "src/CtsOptions.h",
+        "src/TechChar.h",
     ],
-    includes = ["include"],
+    includes = ["include", "src"],
```
maliberty, line 54: Why are we exposing files from src externally?

src/gpl/test/ar02.py
```diff
@@ -2,6 +2,9 @@
 import helpers
 import gpl_aux
 
+bazel_working_dir = "/_main/src/gpl/test/"
+helpers.if_bazel_change_working_dir_to(bazel_working_dir)
```
maliberty, line 6: What is this for?

src/odb/test/test_module.py
```diff
@@ -11,16 +11,13 @@
...
-        self.db, self.lib = helper.createSimpleDB()
+        self.db, self.lib, self.design, self.ord_tech = helper.createSimpleDB()
         self.block = helper.create1LevelBlock(self.db, self.lib, self.db.getChip())
         self.master_mod = odb.dbModule_create(self.block, "master_mod")
         self.parent_mod = odb.dbModule_create(self.block, "parent_mod")
         self.i1 = odb.dbModInst_create(self.parent_mod, self.master_mod, "i1")
         self.inst1 = self.block.findInst("inst")
 
-    def tearDown(self):
-        self.db.destroy(self.db)
```
maliberty, line 22: Is the reason we no longer need this is that the tests are run in separate processes?

BUILD.bazel
QuantamHD: No a cc_library does not trigger a link step. There's a dedicated shared library rule, but I'm less familiar with it. But proper so's are binaries in bazel's world.

src/Exception-py.i
QuantamHD: I was trying not to change the default behavior of the -python embedded interp. Once we removed Cmake and -python the non bazel stuff would get deleted.

src/cts/BUILD
QuantamHD: Because both CtsOptions and TechChar are exposed to the swig bindings, which defacto makes them public hdrs.

src/odb/test/test_module.py
QuantamHD: Because I changed all the odb tests to get their DB from Design which owns the database, and once Design goes out of scope it cleans up the db automatically.

src/gpl/test/ar02.py
QuantamHD: Bazel's default working directory is the .runfiles folder of the binary in which you are currently executing. GPL tests are written in a way that assumes their CWD is the test folder of GPL. This line of code checks if you're being run in bazel, and if so changes the working directory to gpl/test within the runfiles folder.

QuantamHD:
> I'm surprised LTO shrinks the size that much. How does it impact build time?

Maybe 30 seconds?

BUILD.bazel
QuantamHD: Removed.

openroad/BUILD
```diff
@@ -27,11 +27,14 @@ py_package(
 
 py_wheel(
     name = "openroad_wheel",
+    abi = "abi3",
     distribution = "openroad",
     homepage = "https://github.com/The-OpenROAD-Project/OpenROAD",
-    python_tag = "py3",
+    python_requires = ">=3.10",
+    python_tag = "cp310",
```
gadfort, line 34: does this imply that only python3.10 will be supported?
It would be nice if the wheel could be built for 3.10, 3.11, 3.12, 3.13, and 3.14, otherwise this will not be particularly portable.
With the `openroad -python` we controlled the python version, but that is not necessarily true with something published on pypi.

QuantamHD: This wheel is built with ABI3.10 which means it will work with any python greater than or equal to 3.10.

ABI3 specifies a subset of the CPython API that is ABI compatible so this tag tells pip that this wheel is capable of being loaded by any CPython greater than or equal to 3.10

oharboe:
🥳

openroad/BUILD
vvbandeira: @QuantamHD 
Did you try with older versions, or was 3.10 what you had available?
Would it be worth trying with older versions to increase support?

gadfort: @vvbandeira python3.9 is EOL in a month or two, so not sure how much value there is to support that.

vvbandeira: We still have users on legacy OS like RHEL8 (which ships with older versions of Python), so there may be requests for older versions, even after EOL. Hopefully, that will not happen, but I thought it was worth asking.

oharboe: Python is a marginal use case in its current form compared to the new world of pypi....

Perhaps dropping Python from older RHELs is acceptable while still building with Bazel?

vvbandeira: @oharboe
Yes, that should be an acceptable alternative.
I also don't think that support for older Python versions is something that should block this PR or the adoption of Bazel.

QuantamHD:
@maliberty This is good to go as is. I think there's a bit more work in setting up the CI, but I would like to do that in a separate PR.

PR #8339, opened 2025-09-15
Add description for write_cdl command.

titan73:
Also fix write_db description.


rovinski:
@titan73 the command doc has `-masters` but the example has `-master`. I think `-masters` is correct.

titan73:
@rovinski Thanks for spotting this. Sent https://github.com/The-OpenROAD-Project/OpenROAD/pull/8361.

PR #8340, opened 2025-09-15
est: fix typo on function call

eder-matheus:
I've found this small bug when investigating a related issue. I'm not sure how it was working even with the missing `s`.


maliberty:
I suspect the tclreadline is autocompleting it.

PR #8349, opened 2025-09-16
Fix header ambiguously exported by two targets

hzeller:
`rmp/Restructure.h` was exported by two targets.

Found with:

```
bant lib-headers | awk '{print $1}' | sort | uniq -cd
```

While at it, canonicalize all targets:

```
. <(bant canonicalize ...)
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -142,7 +142,7 @@ cc_binary(
     deps = [
         ":openroad_lib",
         ":openroad_version",
-        "//:ord",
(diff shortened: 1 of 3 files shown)


oharboe:
Silly question: what is "bant"?

hzeller:
> Silly question: what is "bant"?

My little swiss-army knife for dealing with bazel projects http://bant.build/

maliberty (approved):

Issue #7699, opened 2025-07-02
Expand multi-clock functionality in OR GUI


jeffng-or:
### Description

Based on feedback from a customer, we should enhance/expand the OR GUI functionality dealing with multiple clocks. The CTS viewer currently has tabs to view each clock tree, which is good.

Here are some quick thoughts on how to enhance the GUI:

- In the Charts view, we display all of the paths regardless of clock in the histogram. Enhance the filtering to allow the user to specify which clock(s) they want to include in the histogram. Then, use the clock selection to tie into the click action so that the Timing Report that comes up is filtered by the clock selection. So, if they are viewing the histogram for a specific clock and click on one of the bars, only those paths are displayed in the Timing Report
- In the Timing Report, enable clock selection in case the user isn't using the Charts as an entry point into the Timing Report

jeffng-or:
we can use asap7 ethmac as the test case since it has three clocks

jeffng-or:
Another note, Tom and I looked at the output from "report_checks -group_path_count 5" and it looks like there are separate path groups for each of the clocks. So, it looks like the information is there, but possibly not hooked up to the Charts Path Group selector.

LucasYuki:
I made an initial implementation where the clock selection is made in the same dropdown menu as the path.
Do you think this is ok, or would it be better if the clock selection were independent of the path group selection?

<img width="771" height="586" alt="Image" src="https://github.com/user-attachments/assets/b8c68024-9881-45b4-a7fd-07fb37f7937a" />

maliberty:
These are separate options (in principle you could choose a path group & a clock)

jeffng-or:
@LucasYuki , are you still working on this one? If so, please remove the Stale label.

PR #8356, opened 2025-09-17
gui: add path group filter to the Timing report 

LucasYuki:
Part of #7699.

Changes:
- Add path group filter to the Timing report.
- Makes the timing report settings window resizable.
  - At lower resolutions, such as 720p, the timing report settings were bigger than the screen, making it difficult to access some settings.


maliberty:
Why is it necessary to adjust the size hints?  Isn't the scrollbar enough for Qt to ensure the rest?

LucasYuki:
Without the size hints, on a 1080p display, the timing report settings window doesn't open in full size.

image

PR #8362, opened 2025-09-18
meta: Working towards a working MacOS build

QuantamHD:
This fixes some, but not all of the issues. Some of which are located in rules_hdl BUILD files. They will either need to be moved to bzlmod's or fixed upstream, but these are good changes regardless.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -138,11 +138,13 @@ cc_binary(
     name = "openroad",
     srcs = [
         "src/Main.cc",
-        "//bazel:runfiles",
     ],
     copts = OPENROAD_COPTS,
     features = ["-use_header_modules"],
-    malloc = "@tcmalloc//tcmalloc",
+    malloc = select({
+        "@platforms//os:linux": "@tcmalloc//tcmalloc",
+        "@platforms//os:macos": "@bazel_tools//tools/cpp:malloc",
+    }),
     visibility = ["//visibility:public"],
     deps = [
         ":openroad_lib",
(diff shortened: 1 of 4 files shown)


BUILD.bazel
```diff
@@ -138,11 +138,13 @@ cc_binary(
     name = "openroad",
     srcs = [
         "src/Main.cc",
-        "//bazel:runfiles",
     ],
     copts = OPENROAD_COPTS,
     features = ["-use_header_modules"],
-    malloc = "@tcmalloc//tcmalloc",
+    malloc = select({
+        "@platforms//os:linux": "@tcmalloc//tcmalloc",
```
maliberty, line 145: Not related to this PR but does this mean that the cpp unit tests are not using tcmalloc and only OR is?

QuantamHD: Correct. I think you can set a global config in the bazelrc, but to be honest I don't know how

maliberty: It may not matter often but in some edge case it could.  A problem for someone who has the time to figure it out.

QuantamHD:
```
--custom_malloc=<a build target label> default: see description
Specifies a custom malloc implementation. This setting overrides malloc attributes in build rules.

Tags: changes_inputs, affects_outputs
```

I guess there's a flag. Imo leave it alone.

QuantamHD:
@maliberty This is good to go. Just the clang-tidy missing the runfiles include.

Issue #8169, opened 2025-08-30
More flexibility for `-connect_to_pads`


mole99:
### Description

I'm currently working on a padring for gf180mcu. The issue is that the power/ground pads have their core connections only on Metal2.

It seems that `add_pdn_stripe` with `-connect_to_pads` uses the specified layers to connect to the pads. This has two issues:

1. Since Metal2 is vertical, this limits me in placing the power/ground pads only in the north or south IO rows.
2. The vertical part of the core ring must be Metal2, which could be more resistive in some stack ups.

### Suggested Solution

Would it be possible to make the connection to the pads more flexible?

For example, I could imagine a dedicated `connect_to_pads` command that would allow me to pass a vertical and horizontal layer to override the default directions from the Tech LEF. I could then use Metal4/Metal5 for the core ring, and Metal2 to connect to the power/ground pads all around.

### Additional Context

<img width="1703" height="1285" alt="Image" src="https://github.com/user-attachments/assets/c76ea3af-c739-4106-a714-fdb4d8f41730" />

rovinski:
To be honest, this unfortunately is bad pad design. The pads should have pins on multiple layers so that the grid can always be connected in the preferred direction. Generally, a pad will just have internal vias to via up or down to one or more consecutive layers (such as M3 in your example) to avoid a layer change outside of the cell.

That being said, the solution here is to connect to the horizontal pads in M2. The outer ring can be connected directly without changing layers, the inner ring will have to via up to M3 and then via down to connect to the ring.

I'm not sure if PDN does that, so I'll leave someone else to comment on that.

maliberty:
@gadfort suggestions?

mole99:
@rovinski I agree, this is bad pad design. However, these are the cells GF supplied. There may be other PDKs out there with power/ground pads with pins on a single layer.

I know that the power/ground pads were edited before they were open-sourced: GF didn't own the 3.3V stdcell library, thus they only released a 5V stdcell library and connected DVDD with VDD as well as DVSS with VSS in the power/ground pads for only a single power domain. Though, I don't believe this changed anything about the pin layers.

No, pdngen currently only works according to the directions in the tech lef. Like I said, I would use Metal4/Metal5 for the core ring, and Metal2 to connect to the power/ground pads.

gadfort:
@mole99 it may be possible to add some kind of functionality to address this, but the vias would need some careful consideration. Without a test case, there is nothing that can be done though.

[...]

PR #8367, opened 2025-09-18
pdn: fix pad edge connections

gadfort:
Fixes:
- power pads with core facing pins on only non-preferred direction layers were getting rejected from the ring connections.

Addresses: #8169


src/pdn/test/gf180/gf180mcu_6LM_1TM_9K_7t_tech.lef
maliberty, line 1: It would be better to treat this like the other PDKs and put the setup in test/gf180 and symlink it here.  That way any future use of this PDK can be shared.

Issue #8343, opened 2025-09-15
MPL should evaluate clusters' connections based on ratio


AcKoucher:
### Description

The original implementation used a fixed value which is settable by the user.

### Suggested Solution

Use a ratio based on on the total weight of connections.
We shouldn't allow the user to set that as well.

PR #8370, opened 2025-09-18
mpl: update connections after merging clusters

AcKoucher:
Found this bug when working on #8343.

The only reason nothing blows up nowadays is because, when we use the clusters' connection maps, we don't retrieve the `Cluster*`. Trying to do that with `std::map::at()` results in an invalid map access without the changes here.

Also some subtle renaming and code improvements.


AcKoucher:
Running Secure-CI.

AcKoucher:
Interestingly, results don't change and Secure-CI is green.

eder-matheus (approved):

PR #8399, opened 2025-09-22
openroad_lib_private is not used anywhere. Remove.

hzeller:


hzeller:
Looks like a left-over from early days of the build file.

maliberty:
I thought it was for Google internal use.  @mikesinouye @QuantamHD do you need this target?

hzeller:
I asked @mikesinouye and he said they decoupled the toplevel BUILD file, so does not need that target anymore.

mikesinouye:
I had to desync the top level BUILD file between Github and our internal copy due to the number of required changes. So this can be removed in OSS.

maliberty (approved):

PR #8375, opened 2025-09-19
mpl: evaluate connection between clusters based on connections' ratio 

AcKoucher:
Resolve #8343.

Requires https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3504 in order to not break the CI!

The methods that evaluate connections between clusters and search for closely connected clusters were moved out of the `Cluster` class as they should be done by the `ClusteringEngine` itself.

A strong connection is now based on:

```
            connection weight A -> B
------------------------------------------------- >= 0.1 (10%)
weight of A connections + weight of B connections
```

PR #8403, opened 2025-09-22
mpl: weigh nets with IO pins as regular nets

AcKoucher:
Currently, connections involving IO clusters are weighted much strongly than others. This makes the approach in #8375 unfeasible, because, when evaluating connections between clusters by looking at the connections' ratio, connections that don't involve IO clusters may be neglected as they're much weaker.

An example of the problem would be how MPL fails to group the 8x8 array macros when using the connections' ratio approach without the changes here.

Screenshot from 2025-09-22 12-04-11

_(Obs: In the image, some coarse shaping iteration.)_


AcKoucher:
Running Secure-CI.

AcKoucher:
Re-running Secure-CI with updates to fix the mock-array failure cited in #8411.

AcKoucher:
A recently merged bug fix (#8424) had reasonable impact in my previous Secure-CI.
I'm running Secure-CI again to avoid metrics discrepancies.

AcKoucher:
Secure-CI showed a huge timing degradation for asap7/riscv32i which happens because the different weight makes the clustering result change for worse. However, this doesn't happen if we combine the changes here with the ones in #8375 which should be the subsequent PR to this one. So I'm closing this and will merge everything in the same PR.

PR #8404, opened 2025-09-23
pdn: add check for macro pin access before running power grid generation

### LINKED ISSUE #8373: [ERROR PDN-0233] Failed to generate full power grid.
### Describe the bug

[pdn_top_asap7_base_2025-09-19_04-03.tar.gz](https://github.com/user-attachments/files/22417801/pdn_top_asap7_base_2025-09-19_04-03.tar.gz)

```
INFO PDN-0001] Inserting grid: top
[INFO PDN-0001] Inserting grid: CORE_macro_grid_1 - u_sdq_17x64
[WARNING PDN-0232] CORE_macro_grid_1 - u_sdq_17x64 does not contain any shapes or vias.
[ERROR PDN-0233] Failed to generate full power grid.
Error: pdn.tcl, 6 PDN-0233
```

### Expected Behavior

No error

### Environment

```shell
OpenROAD 24Q3-8871-g96512a37e9
```

### To Reproduce

See above

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

gadfort:
Closes #8373

Adds:
- checkSetup for instance/macro grids to check if pins will be accessible from above

### DIFF
--- a/src/pdn/test/BUILD
+++ b/src/pdn/test/BUILD
@@ -273,6 +273,8 @@ filegroup(
                 "ihp_ethmac/RM_IHPSG13_1P_256x48_c2_bm_bist.lef",
                 "ihp_ethmac/floorplan.def",
             ],
+            "asap7_failed_macro_covered": ["asap7_failed_macro_grid.def"],
+            "asap7_macro_notcovered": ["asap7_failed_macro_grid.def"],
         }.get(test_name, []),
     )
     for test_name in ALL_TESTS


src/pdn/src/grid.cpp
```diff
@@ -1666,6 +1666,66 @@ bool InstanceGrid::isValid() const
...
+void InstanceGrid::checkSetup() const
+{
+  Grid::checkSetup();
+
+  // check blockages above pins
+  std::set<odb::dbBox*> obs;
+  odb::dbMaster* master = inst_->getMaster();
+  for (auto* box : master->getObstructions()) {
+    obs.insert(box);
+  }
```
maliberty: Why copy out the obstructions to a set just to later iterate them in `for (auto* master_obs : obs) {`?

src/pdn/src/grid.cpp
```diff
@@ -1666,6 +1666,66 @@ bool InstanceGrid::isValid() const
...
+
+  const auto nets = getNets(startsWithPower());
+  for (auto* iterm : inst_->getITerms()) {
+    if (std::find(nets.begin(), nets.end(), iterm->getNet()) == nets.end()) {
+      continue;
+    }
+    odb::dbTechLayer* top = nullptr;
+    std::map<odb::dbTechLayer*, std::set<odb::Rect>> boxes;
```
maliberty: Why use a map?  Just keep those in the top layer.  Whenever top changes just empty the set.

oharboe:
BUILD.bazel needs an update to include the new resource files for the tests

src/pdn/src/grid.cpp
gadfort: when I started I had thought of something else, but it was not needed, removed.

src/pdn/src/grid.cpp
gadfort: when I started I had thought of something else, but it was not needed, removed.

gadfort:
@oharboe is there a way to suppress the output of the bazel orfs log into a file in the CI? it's insanely large and masks the other error messages.

@maliberty the file bazel is complaining about is used in other tests but not listed in the BUILD file, why was this working before?

oharboe:
> @oharboe is there a way to suppress the output of the bazel orfs log into a file in the CI? it's insanely large and masks the other error messages.

What output, other than ORFS output, are you referring to?

We've worked on reducing the verbosity of ORFS...

oharboe:
@gadfort Regarding noise, don't look at "bazel keep going", but at 

image

From this, you quickly see that you should run the following locally:

    bazelisk test src/pdn/...

oharboe:
Searching for "error:", I find:

```
[2025-09-23T12:22:45.181Z] [ERROR ORD-0003] asap7_failed_macro_grid.def does not exist.
[2025-09-23T12:22:45.181Z] Error: asap7_macro_notcovered.tcl, 7 ORD-0003
```

maliberty:
We allow `<test_name>.*` by default but you have a shared def file between three tests that doesn't match that pattern.  I added an explicit callout for that.

gadfort:
@maliberty thanks.

@oharboe I was just wondering if the ORFS run could be piped to a log file instead of to the stdout for the CI since it adds a lot of text which makes the bazel output hard to parse. Another option would be to run orfs as a separate action so it's output is selfcontained.

oharboe:
We want a simple stable CI interface and run it all in parallel and sometimes we do want to see that output.

    bazelisk test ...

Tests everything and runs it all in parallel

oharboe:
So the current state of affairs is the best attempt at supporting all use cases with no further CI work.

maliberty:
I usually look at the pipeline view (eg https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-8404-head/2/pipeline-overview/).  That lets you see output by pipeline step.  From there the 
image button lets you see the output from just that step in plain text if needed.

PR #8406, opened 2025-09-23
orfs: get a bump on the books

oharboe:
if something happens later on, it is nice to have a last known good version on the books.


oharboe:
@maliberty non-bazel pr-merge errors. Good to go.

PR #8407, opened 2025-09-23
Document cyclic dependency.

hzeller:
To be fixed later, but this way it is already documented. (what needs to be done is probably break out the libraries in the BUILD file better).

### DIFF
--- a/src/dbSta/BUILD
+++ b/src/dbSta/BUILD
@@ -9,6 +9,13 @@ package(
     features = ["layering_check"],
 )
 
+# TODO: there is a cyclic dependency between
+# some of these. dbSta depends on dbNetwork and
+# dbSdcNetwork, but these would need to include
+# dbSta.hh ...
+# So this probably needs to be broken into more
+# libraries.
+
 cc_library(
     name = "dbSta",
     srcs = [
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -68,6 +68,11 @@ Recommended conclusion: use map for concrete cells. They are invariant.
 #include "sta/Search.hh"
 #include "utl/Logger.h"
 
+// Circular dependency! We need this header, but that
+// is provided by :dbSta which in turn depends
+// :dbNetwork ...
+// #include "db_sta/dbSta.hh"
+
 namespace sta {
 
 using utl::ORD;
--- a/src/dbSta/src/dbSdcNetwork.cc
+++ b/src/dbSta/src/dbSdcNetwork.cc
@@ -13,6 +13,11 @@
 #include "sta/ParseBus.hh"
 #include "sta/PatternMatch.hh"
 
+// Circular dependency! We would need this for
+// e.g. sta::Cell, but that requires to depend on :dbSta
+// which in turn depends on :dbStaNetwork ...
+// #include "db_sta/dbSta.hh"
+
 namespace sta {
 
 static std::string escapeDividers(const char* token, const Network* network);


maliberty:
It would be helpful to state the actual symbols involved in the circular dependency.

hzeller:
I mentioned one in one of the files (`sta::Cell`), but there were a few more that all required `dbSta.hh` (currently off computer, but can have a look later)

maliberty:
You can get that from sta/include/sta/NetworkClass.hh

hzeller:
Ok, looked
both need [`sta::Cell`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/dbSta/include/db_sta/dbSta.hh#L40) and [`sta::Instance`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/dbSta/include/db_sta/dbSta.hh#L50)
 
NetwokClass.hh only has a forward declare, not a definition.

maliberty:
There is no definition - it is just used for type puning.

hzeller:
yes, but even opaque types should be defined (like what you do in dbSta.hh; and the comment even gives a reason why this is needed for the RTTI constructs).

maliberty:
As it states there

// talking to the OpenSTA author about implementing these
// upstream instead.

is the right fix.  Putting it in dbSta was a hack and causes the cycles you see.  @QuantamHD what came of those discussions?  We can put it in our fork if necessary.

hzeller:
This might be one of the cases where a Macro can improve readability because it describes the intent.
I am thinking of something like this:

```c++
#define MAKE_OPAQUE_TYPE(T)                     \
class T                                         \
{                                               \
 public:                                        \
  T() = delete;                                 \
}

MAKE_OPAQUE_TYPE(Cell);
MAKE_OPAQUE_TYPE(Instance);
// ...
#undef  MAKE_OPAQUE_TYPE
```

Since it will not only improve the use of the types in RTTI context, but also strictly improve the readability, it might be easier to upstream such change to `sta/NetworkClass.hh`

hzeller:
Converting this to draft, as other includes should be used; but leaving it open so that we don't forget to move the opaque type stuff into sta.

maliberty:
I think the easiest way to break the cycle is to just move these defintions into their own header and make it a more base level dep than dbSta which is more high level.

hzeller:
w.r.t. https://github.com/The-OpenROAD-Project/OpenROAD/pull/8407#issuecomment-3327007666

The change would be something like

```diff
--- a/include/sta/NetworkClass.hh
+++ b/include/sta/NetworkClass.hh
@@ -34,19 +34,29 @@
 
 namespace sta {
 
+#define MAKE_OPAQUE_TYPE(T)                     \
+class T                                         \
+{                                               \
+ public:                                        \
+  T() = delete;                                 \
+}
+
+MAKE_OPAQUE_TYPE(Cell);
+MAKE_OPAQUE_TYPE(Library);
+MAKE_OPAQUE_TYPE(Port);
+MAKE_OPAQUE_TYPE(Instance);
+MAKE_OPAQUE_TYPE(Pin);
+MAKE_OPAQUE_TYPE(Term);
+MAKE_OPAQUE_TYPE(Net);
+MAKE_OPAQUE_TYPE(ViewType);
+
+#undef  MAKE_OPAQUE_TYPE
+
 class Network;
 class NetworkEdit;
 class NetworkReader;
-class Library;
-class Cell;
-class Port;
 class PortDirection;
-class Instance;
-class Pin;
-class Term;
-class Net;
 class ConstantPinIterator;
-class ViewType;
 class LibertyLibrary;
 
 typedef Iterator<Library*> LibraryIterator;
```

should I file a PR against https://github.com/The-OpenROAD-Project/OpenSta or try to parallaxsw ?

maliberty:
I have no idea if you can get it into parallax but let's see.  I would include a description of the motivating issue with RTTI.

PR #8411, opened 2025-09-23
mpl: ensure number of perturbations increase for large arrays

AcKoucher:
I got a failure in the Secure-CI for #8403 on mock-array in which I couldn't find the correct positions of the macros inside the 8x8 array. The problem is that large arrays require more perturbations per step in order to converge, and, currently, there's a bug causing the number of perturbations per step to remain the same.

I.e., we're never entering the condition to increase the number of perturbations due to the way we're setting that same value.


AcKoucher:
Running Test-CI.

AcKoucher:
Test-CI is fine.

PR #8414, opened 2025-09-23
Or tools upgrade

hongted:


hzeller:
I've also pinged https://github.com/google/or-tools/issues/4551#issuecomment-3327916051 so that we can get the same newer version to be used in our [MODULE.bazel](https://github.com/The-OpenROAD-Project/OpenROAD/blob/143ce0188d76d72266f8bbbee44e60454933ef17/MODULE.bazel#L89)

vvbandeira:
Can we close this PR? or-tools have been updated to 9.14

PR #8424, opened 2025-09-24
mpl: fix dataflow connections between clusters with macros

AcKoucher:
Connections made from paths that link macro pins to other macros are being created in such a way that only one of the clusters actually possesses that connection.


AcKoucher:
Running Secure-CI.

eder-matheus (approved):

PR #8434, opened 2025-09-25
GPL: use regions instead of power domains

osamahammad21:


gudeh:
I think this makes a lot of sense, I don't understand why we used the power domain previously, and I assume it worked because examples we tested so far had a region directly associated with a power domain.

Also, we have `OpenROAD/test/upf_test.tcl` and `OpenROAD/test/upf_aes.tcl` seem to be passing with the change.

gudeh (changes requested):
I noticed we are missing a change at initialPlace.cpp: https://github.com/The-OpenROAD-Project/OpenROAD/blob/024ae4397eb3c67b748d935c52a17e861e4fc92c/src/gpl/src/initialPlace.cpp#L134

osamahammad21:
@maliberty @gudeh I think this is ready for merge. Could you please confirm?

maliberty:
lgtm - @gudeh ?

gudeh:
I noticed a few naming occurrences of power domain which can be renamed to region. I created a complementary PR at https://github.com/The-OpenROAD-Project/OpenROAD/pull/8475, is no-op.

PR #8437, opened 2025-09-25
rmp: Simulated annealing

kbieganski:
New resynthesis strategy for `rmp`.
It searches for an ABC script that improves timing via simulated annealing.
The script is a series of operations on ABC's internal AIG data structure.
A neighboring solution is a script with one operation added, removed, or two operations swapped.
The optimization function is defined as the worst slack.


maliberty:
Please share results showing the effectiveness of this change.

kbieganski:
Some results are in the tests: [here](https://github.com/The-OpenROAD-Project/OpenROAD/pull/8437/files#diff-7bb67967582e5d158d8d459d191f7188aedf33b426fe50cb76ccbfc8171745b9), [here](https://github.com/The-OpenROAD-Project/OpenROAD/pull/8437/files#diff-b22512611df848c19c31970c601875701b267b35c4dbdf03d441ca354bae584f), and [here](https://github.com/The-OpenROAD-Project/OpenROAD/pull/8437/files#diff-0667fd36277460b24bf1fa0ab3194e6b0f67f1e57fc65b10e3b8c7446ff9893e).

src/rmp/README.md
```diff
@@ -66,6 +66,34 @@ resynth
...
+    [-iters iters]
+    [-initial_ops initial_ops]
+```
+
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-corner` | Process corner to use. |
+| `-slack_threshold` | Specifies a (setup) timing slack value below which timing paths need to be analyzed for restructuring. The default value is `0`, and the allowed values are floats `[0, MAX_FLOAT]`. |
```
povik: Why no negative thresholds?

src/rmp/README.md
```diff
@@ -66,6 +66,34 @@ resynth
...
+| `-seed` | Seed to use for randomness in simulated annealing. |
+| `-temp` | Initial temperature for simulated annealing. The default is the required of the worst slack path. |
```
povik: Is this the required arrival time on the endpoint for the worst slack path?

src/rmp/src/utils.cpp
```diff
@@ -0,0 +1,72 @@
...
+                                       rsz::Resizer* resizer,
+                                       sta::Slack slack_threshold)
+{
+  std::vector<sta::Vertex*> result;
+
+  sta::dbNetwork* network = sta->getDbNetwork();
+  for (sta::Vertex* vertex : *sta->endpoints()) {
+    sta::Pin* pin = vertex->pin();
+    sta::PortDirection* direction = network->direction(vertex->pin());
+    if (!direction->isInput()) {
```
povik, line 38: What's the reason for this check? I think we disqualify top module outputs here.

povik: I see this is code moved from elsewhere but still I wonder about the purpose of the check

src/rmp/src/annealing_strategy.cpp
```diff
@@ -0,0 +1,650 @@
...
+
+      [&](auto& gia) {
+        // &dc2
+        debugPrint(logger, RMP, "annealing", 1, "Starting heavy rewriting");
+        replaceGia(gia, Gia_ManCompress2(gia, true, false));
+      },
+
+      [&](auto& gia) {
+        // &b
+        debugPrint(logger, RMP, "annealing", 1, "Starting syn2");
```
povik: Nit: mismatched print

src/rmp/src/annealing_strategy.cpp
```diff
@@ -0,0 +1,650 @@
...
+        replaceGia(gia,
+                   Gia_ManAreaBalance(gia, false, ABC_INFINITY, false, false));
+      },
+
+      [&](auto& gia) {
+        // &b -d
+        debugPrint(logger, RMP, "annealing", 1, "Starting syn2");
```
povik: Here too

src/rmp/src/annealing_strategy.cpp
```diff
@@ -0,0 +1,650 @@
...
+          replaceGia(gia, Gia_ManRehash(gia, false));
+        }
+        abc::If_Par_t pars = {};
+        Gia_ManSetIfParsDefault(&pars);
+        pars.fDelayOpt = true;
+        pars.nLutSize = 6;
+        pars.fTruth = true;
+        pars.fCutMin = true;
+        pars.fExpRed = false;
+        debugPrint(logger, RMP, "annealing", 1, "Starting tech mapping (&if)");
```
povik: I think the important part is the SOP balancing enabled with `-g` and the command isn't used for actual technology mapping here. It doesn't matter much given this is a debug print

src/rmp/README.md
povik: Based on the code it is. I think it would be good to say it's the required time on the endpoint in the description.

src/rmp/src/annealing_strategy.cpp
kbieganski: Thanks, good catch

src/rmp/README.md
maliberty: It think users will find required time to be a strange way to express their intent.

povik: @maliberty They're specifying temperature but the default value for it is taken from the required time at the end of the worst path.

src/rmp/README.md
kbieganski: Fixed, interestingly I copied it over from the `restructure` doc above.

src/rmp/src/utils.cpp
kbieganski: @QuantamHD Do you know?

QuantamHD: I believe this was to defend against endpoints that are unconstrained. It's probably overly simplistic.

src/rmp/README.md
```diff
@@ -66,6 +66,36 @@ resynth
...
+#### Options
+
+| Switch Name | Description |
+| ----- | ----- |
+| `-corner` | Process corner to use. |
+| `-slack_threshold` | Specifies a (setup) timing slack value below which timing paths need to be analyzed for restructuring. The default value is `0`. |
+| `-seed` | Seed to use for randomness in simulated annealing. |
+| `-temp` | Initial temperature for simulated annealing. The default is the required arrival time on the worst slack endpoint. |
+| `-iters` | Number of iterations to run simulated annealing for. |
+| `-revert_after` | After the given number of iterations that worsen slack, revert to best fonud solution. |
```
povik:
```suggestion
| `-revert_after` | After the given number of iterations that worsen slack, revert to best found solution. |
```

src/rmp/src/utils.cpp
povik: It's not the right check

maliberty:
Please merge master which should fix the test failure

src/rmp/src/annealing_strategy.cpp
```diff
@@ -0,0 +1,681 @@
...
+
+  cut::AbcLibraryFactory factory(logger);
+  factory.AddDbSta(sta);
+  factory.AddResizer(resizer);
+  factory.SetCorner(corner_);
+  cut::AbcLibrary abc_library = factory.Build();
+
+  // GIA ops as lambdas
+  // All the magic numbers are defaults from abc/src/base/abci/abc.c
+  // Or from the ORFS qbc_speed script
```
maliberty:
```suggestion
  // Or from the ORFS abc_speed script
```

src/rmp/src/annealing_strategy.cpp
```diff
@@ -0,0 +1,681 @@
...
+        = *temperature_ * (static_cast<float>(iterations_ - i) / iterations_);
+
+    if (revert_after_ && worse_iters >= *revert_after_) {
+      logger->info(RMP, 57, "Reverting to the best found solution");
+      ops = best_ops;
+      worst_slack = best_worst_slack;
+      worse_iters = 0;
+    }
+
+    if (i > 0 && (i + 1) % 10 == 0) {
```
maliberty: Redundant - if i == 0 then (i + 1) % 10 != 0 so no need to check i > 0

maliberty (changes requested):

src/rmp/src/annealing_strategy.cpp
RRozak: Done

src/rmp/src/annealing_strategy.cpp
RRozak: Done

maliberty (approved):

maliberty:
I didn't want to hold this PR up further but I think it would be good to add and eqy step to prove equivalence.

RRozak:
Could you tell how it should be added? I wanted to add similar commands to tests we added in this PR and put the annealing between them: https://github.com/The-OpenROAD-Project/OpenROAD/blob/75f345819cc1823b1c2c24ceab713588d75e2c9e/src/rsz/test/eqy_repair_setup2.tcl#L14-L15
Unfortunately it hangs. I also tried to run `eqy_repair_setup2.tcl` with `EQUIVALENCE_CHECK` env variable set to 1 and it didn't complete in 17h.

maliberty:
I have no expertise with those tools and find them obscure.  I suggest asking on their GH for help.

PR #8446, opened 2025-09-26
pdn: add check in macro grid setup

gadfort:
Closes: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3511

Adds:
- warning for macros cells that cannot be found
- ignore non-fixed instances


src/pdn/src/pdn.tcl
```diff
@@ -1024,6 +1024,9 @@ proc define_pdn_grid_macro { args } {
         utl::error PDN 1030 "Unable to find instance: $inst_pattern"
       }
       foreach inst $sub_insts {
+        if { ![$inst isFixed] } {
+          continue
```
maliberty, line 1030: A warning would be helpful

gadfort: good point

maliberty (approved):

PR #8452, opened 2025-09-27
Remove using declaration that spilled a symbol into different namespace.

hzeller:
First step in fixing `using` declarations in header files. In this case, the `odb::dbTransform` was using-declared in a different namespace in a header file which then spilled this symbol into `drt::` namespace. Unless in very special circumstances this is of course never desirable.

(using declaration was in src/drt/src/db/infra/frBox.h)

This is the first step in cleaning up `using` declarations that spill symbols globally into different namespaces.

### DIFF
--- a/src/drt/src/db/drObj/drRef.h
+++ b/src/drt/src/db/drObj/drRef.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include "db/drObj/drFig.h"
+#include "odb/dbTransform.h"
 #include "odb/dbTypes.h"
 
 namespace drt {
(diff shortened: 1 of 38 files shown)


hzeller:
The fixit-journey ahead will take me a while it seems: 

```
rg "using ...::" -g "*.h*" | wc -l
```

maliberty:
https://google.github.io/styleguide/cppguide.html#Aliases talks about avoiding public aliases.  None of this is in the public api of drt (ie src/drt/include/...).  What rule are you driving towards?

hzeller:
if it is in a header, it is essentially public from the readers' perspective and super-confusing.

I stumbled upon it as clang tidy complains about no header included for `drt::dbTransform` which doesn't exist. And then I stumbled upon that alias.

If it is in the header in or outside a namespace, consider it public.
Aliases might be ok if you have a namespace `foo::private` or `foo::internal`. Or it is ok inside the namespace of a struct or class or inside a function body.

hzeller:
Essentially applying "However, local convenience aliases are fine in function definitions, private sections of classes, explicitly-marked internal namespaces, and in .cc files"

maliberty (approved):

maliberty:
"and in .cc files" suggests you could make the .h use the odb:: prefix while the .cc has `using odb::dbTransform;`

PR #8454, opened 2025-09-28
Avoid leaking odb::Rect into different namespaces.

hzeller:
Fix some uses of `using odb::Rect` in header files.


maliberty (approved):
Is there any enforcement method once you have finished with this.  Without it I expect it will rot quickly.

hzeller:
I think once things are done, we should have some `grep` for some pattern like this in the CI.
Maybe even possible now if we only apply the grep to the diff ?

maliberty:
Gemini suggests such a grep is possible in a GH action for just the diffs:
```
git diff "origin/${{ github.base_ref }}"..."origin/${{ github.head_ref }}" | grep "your_pattern"
```

PR #8458, opened 2025-09-28
Remove some `using namespace` uses.

hzeller:


maliberty (approved):

maliberty:
Interestingly https://clang.llvm.org/extra/clang-tidy/checks/google/build-using-namespace.html is more relaxed than your rule (just whole namespace imports are disallowed).

hzeller:
Yes, full namespace imports is what I am reducing here. `using namespace` is _never_ to be used, not even in *.cc files.

`using ParticularType` is allowed in *.cc files, so it is the more relaxed rule.

PR #8476, opened 2025-09-30
mpl: avoid redundant connections

AcKoucher:
Required for the approach in #8375.

These changes are actually a bug fix.

I didn't understand why the regression tests' results were changing as the redundant connections we're getting rid of here should be inoffensive. The thing is that these redundant connections are inoffensive at cluster placement, because when we rebuild the graph at the end of the clustering process, these nets are gone. However, at _macro placement_, these connections go inside the annealer and mess up the normalization, because we use the sum of the weights of all the nets which takes into account these redundant nets.


AcKoucher:
Running Secure-CI.

eder-matheus (approved):

AcKoucher:
Secure-CI is fine.

maliberty (approved):

Issue #8477, opened 2025-09-30
gpl: bug in the maximum number of filler cells that can be removed during routability


LucasYuki:
### Describe the bug

The current code should limit the maximum number of filler cells removed to 10%, but there is a bug that allows up to 100% of the filler cells to be removed.

### Expected Behavior

The maximum number of filler cells removed should be enforced, or the limit should be removed.

### Environment

```shell
Git commit: 2976317355a8370ad56605ce95cbe7a9afdf18e4
kernel: Linux 6.14.0-1011-gcp
os: Ubuntu 24.04.3 LTS (Noble Numbat)
cmake version 3.31.6
-- The CXX compiler identification is GNU 13.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-24950-g2976317355
-- System name: Linux
-- Compiler: GNU 13.3.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 17
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 13.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-array-bounds - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull
-- Performing Test C_COMPILER_SUPPORTS__-Wno-nonnull - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wno-nonnull
[...]

PR #8480, opened 2025-09-30
Gpl: remove filler removal cap

LucasYuki:
Issue #8477.

Removes the 10% limit of fillers cells removed during routability.


LucasYuki:
Summary of metrics across all design tests.

image

maliberty:
I'm confused as the issue says
```
The current code should limit the maximum number of filler cells removed to 10%, but there is a bug that allows up to 100% of the filler cells to be removed.
```
but this is `Removes the 10% limit of fillers`.  There is no such limit per the issue.

gudeh:
As mentioned on a previous meeting we noticed it could be beneficial to remove the max cap filler removal, and allow for all fillers to be potentially removed.

Initially we wanted to remove at most 10% of the fillers. Although, there was a bug which we were accidentally removing all fillers, but even still this showed as a good outcome on some private designs, since after fixing the bug and removing only 10% of fillers showed congestion problems during GRT on private designs. By removing the max cap we see no divergences during GPL (on contrary to what I observed previously where multiple public designs were diverging consistently).

So we understand it would be better to instead of fixing the cap, just remove it. Since we do not see divergences and private designs benefit from removing the cap.

maliberty:
Needs a small metric update
11:59:52  + ./test/regression jpeg_sky130hd
12:21:05  jpeg_sky130hd (tcl) *FAIL* GRT::ANT::errors   1 >   0

gudeh:
All green @maliberty

maliberty (approved):

PR #8489, opened 2025-09-30
ODB: read liberty from 3dblox files

osamahammad21:


src/OpenRoad.cc
```diff
@@ -485,13 +485,13 @@ void OpenRoad::writeCdl(const char* out_filename,
 
 void OpenRoad::read3Dbv(const std::string& filename)
 {
-  odb::ThreeDBlox parser(logger_, db_);
+  odb::ThreeDBlox parser(logger_, db_, sta::Sta::sta());
```
maliberty: You should use the sta_ object.  Same below

maliberty:
Need to fix in bzl `src/odb/src/3dblox/3dblox.cpp:21:10: error: module //src/odb:odb does not depend on a module exporting 'sta/Sta.hh'`

Issue #8183, opened 2025-09-01
Update OpenROAD bazel CI to show critical path w.r.t. running time


oharboe:
### Description

bazel-orfs does this:

https://github.com/The-OpenROAD-Project/bazel-orfs/blob/a94bc4e8e7973750cf818a220fcb2c1e23042469/.github/scripts/smoketests.sh#L11-L23

### Suggested Solution

Show critical path w.r.t. running times

maliberty:
ORFS is a linear flow so that isn't much new information.  What are you hoping to get from this?

oharboe:
> ORFS is a linear flow so that isn't much new information.  What are you hoping to get from this?

We have many parallel activities in ORFS testing. This will tell me which target takes the most time and what it is that takes how much time.

oharboe:
@vvbandeira @maliberty I'm happy to tweak test cases and chip away at the critical path, but I need this information from CI to do so.

oharboe:
@maliberty @vvbandeira Could we prioritize this? Seems like a trivial change...

We need to know exactly which step is taking how much time to optimize running times and to interpret running times in bazel correctly.

Since we have a cache, things that are fast could be inconclusive and things that appear slow could simply be a cache miss and incorrect expectations compared to a full build.

PR #8492, opened 2025-10-01
[DO NOT REVIEW] Mock array scaled down

oharboe:
@maliberty FYI, waiting for #8183 to evaluate.


oharboe:
@jhkim-pii FYI, waiting for #8183 before proceeding with this one...

PR #8496, opened 2025-10-01
Mock array cleanup

oharboe:
No change, make future PRs easier to review.


test/orfs/mock-array/rules-base.json
maliberty, line 1: The description says "no change" so why are the metrics updated?

oharboe: Because I haven't updated them in a while and I wanted to have a separate rules .json file for each variant

oharboe: I'll have some more followup work on this, so it's just a checkpoint to make subsequent PRs easier to review.

oharboe:
@maliberty Thanks!

It will be much easier to inspect these sort of changes when we have GUI in //:openroad...

maliberty:
Try `bazelisk build --//:platform=gui //:openroad`

oharboe:
> Try `bazelisk build --//:platform=gui //:openroad`

Doesn't cover my use-case.

I need a //:openroad GUI target and an //:openroad-cli headless target...

https://github.com/The-OpenROAD-Project/OpenROAD/blob/29b70acdb142efd054e31971b4a168b4ad90316e/MODULE.bazel#L168

maliberty:
I don't follow - you can build either way.  You can also run the gui build without the -gui at startup.

oharboe:
> I don't follow - you can build either way. You can also run the gui build without the -gui at startup.

This is my use-case, run this command to look at something, such as place:

    bazelisk run test/orfs/mock-array:MockArray_place /tmp/place gui_place

maliberty:
Why wouldn't that work if you compiled with the gui enabled?

oharboe:
> Why wouldn't that work if you compiled with the gui enabled?

I don't build openroad, I run tests, which builds OpenROAD for the exec target, then I investigate intermediate build results.

maliberty:
So the tests can build openroad with that option?

oharboe:
> So the tests can build openroad with that option?

Don't know, haven't investigated, not appealing for my use-case/flow:

1. make changes in a free moment
2. run tests before heading off to do something else
3. investigate result, possibly view in GUI
4. create PR

For "options" to work, I'd have to plan ahead in step 2 and remember where I was when getting to 2 and 3... My context is usually flushed when I return to investigating and possibly creating a PR.

maliberty:
This seems no different the regular ORFS.  You build OR with the gui and decide at runtime as to whether you want to pass -gui or not.

oharboe:
> This seems no different the regular ORFS. You build OR with the gui and decide at runtime as to whether you want to pass -gui or not.

Which is not what I do. The command line below will build all dependencies, runs the flow that produces whatever I want to look at, sets up a tmp folder and then launch the GUI. I don't pick a different binary for the GUI.

    bazelisk run test/orfs/mock-array:MockArray_place /tmp/place gui_place

I coud forget all about the differences between GUI and non-GUI builds if I have a //:openroad that produces a GUI and a //:openroad-cli that produces headless. I'm thinking building with a GUI by default [minimizes surprises](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) for the user, hence I think that a //:openroad-cli target makes sense.

maliberty:
In the end that just translates into `openroad -gui ...` or `openroad ....` and the same binary can serve both purposes.  I think making the default platform gui makes sense.

oharboe:
> In the end that just translates into `openroad -gui ...` or `openroad ....` and the same binary can serve both purposes. I think making the default platform gui makes sense.

GUI:

    bazelisk run test/orfs/mock-array:MockArray_place /tmp/place gui_place

CLI (which use today sometimes):

    bazelisk run test/orfs/mock-array:MockArray_place /tmp/place open_place

PR #8502, opened 2025-10-01
Cts: Add option to skip nets

### LINKED ISSUE #7655: Can you exclude nets from clock tree synthesis?

luis201420:
Fixes #7655 

Continue work from PR #7785 

Add the user option to skip clock tree synthesis for a set of nets: `-skip_nets`. 

The option receives a list containing the names of the nets that CTS should skip.

- If a specified net is the root net for the clock, it will skip the clock tree synthesis for all the sub-nets as well.
- If a specified net is a sub-net, it will only skip this net.

The option was added to the following  three CTS commands:

- `set_cts_config -skip_nets <list_nets>` to define the list of nets to be excluded from CTS. (The user can use the `get_nets` command to get the list of nets or can pass a list of net names `{"net1" "net2"}`).
- `reset_cts_config -skip_nets` to reset the list of skipped nets from CTS.
-  `report_cts_config` to report the cts configuration including the list of of skipped nets.


maliberty:
@arthurjolo please do a first pass review

arthurjolo:
Code wise the PR looks good there is just one point that I think deserves a discussion. This is the clock nets pre CTS for the unity test added by this PR, in red are the nets we are skipping and in grey the nets we build a clock tree.
Captura de tela de 2025-10-01 14-03-34

The net `gclk1` is skipped and the net `gclk2` is driven by `gclk1`, the way PR [#7785](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7785) and this one we are building a clock tree for `gclk2`, should we still build a clock tree for this net, given that its root has no tree. When I did PR  [#7785](https://github.com/The-OpenROAD-Project/OpenROAD/pull/7785), I didn't considered cases like this, but after seen this unity testI think we should skip it as well.

maliberty (approved):

PR #8504, opened 2025-10-01
Include vendored `"omp.h"` with quotes not angle brackets.

hzeller:


maliberty (approved):
I guess by "vendored" you mean that we have it locally for the bazel build.

hzeller:
Yes, "vendored" is typically used in the sense that a dependency is shipped with the build.

maliberty:
It is not vendored in cmake but I guess that is less important.

hzeller:
Don't know what is going on in the CI. Since that part is using some docker stuff, it is not easy to run locally.

maliberty:
There is a failure on master so its not this PR's issue.  You can see that in Jenkins `bazelisk test //test/orfs/mock-array:MockArray_flat_test` failed

PR #8507, opened 2025-10-01
mock-array: less noisy .v files generated

oharboe:
Should have no effect. Only thing is that yosys is sensitive to initial conditions, so even adding whitespace can result in slightly different synthesis results...


maliberty (approved):

oharboe:
@maliberty Manual merge? Unerlated non-bazel CI errors only


image

maliberty:
retriggered the CI - we had a Jenkins glitch after the dependabot update

PR #8510, opened 2025-10-01
dbCompare is an implementation artifact of db.h. Make an inc.

hzeller:
To avoid accidentally creating circular dependencies that confuses tools and users, make it very clear that this is an implementation detail.


maliberty:
I like this change.  If you merge master it should clear up the CI failures.

hzeller:
Looks like `//test/orfs/mock-array:MockArray*` tests are still not happy.
Is this something you work on in #8519 @oharboe  ?

oharboe:
> Looks like `//test/orfs/mock-array:MockArray*` tests are still not happy. Is this something you work on in #8519 @oharboe ?

Yes, fixed in #8519

hzeller:
This looks like some CI glitch to me.

PR #8511, opened 2025-10-01
gui: guard getProperties failure

gadfort:
Fixes:
- crash in GUI when getProperties throws an error (in this case sta reporting that libraries are missing), this causes the GUI to hang and then crash.
- This ensures the GUI reports the error to the user and fails gracefully.


gadfort:
@maliberty I generally agree, but Qt doesn't really seem to have anything obvious that would allow us to catch issues in the eventloop (that I'm aware of) and once we catch outside of that it (like in gui.cpp) the GUI would crash anyways.

maliberty:
Gemini suggests
```
The standard approach is to subclass QApplication and override its notify() method. This method is the central point through which all events are dispatched in a Qt application, making it the perfect place to install a global exception handler for the main GUI thread.
```
which sounds reasonable.
```
bool SafeApplication::notify(QObject *receiver, QEvent *event) {
    try {
        // The base class implementation does the actual event dispatching.
        // All the work is done here.
        return QApplication::notify(receiver, event);
    } catch (const std::exception &e) {
        // Catch standard exceptions
        qCritical() << "Caught C++ exception:" << e.what();
        
        // You can show a message box to the user
        QMessageBox::critical(nullptr,
                              "Application Error",
                              QString("An unexpected error occurred: %1").arg(e.what()));
                              
        // Returning true indicates the event has been handled. In this case,
        // we've "handled" it by catching the exception, so we prevent
        // further processing that might rely on a corrupt state.
        return true; 
    } catch (...) {
        // Catch any other unknown exceptions (e.g., non-standard)
        qCritical() << "Caught an unknown exception.";
        QMessageBox::critical(nullptr,
                              "Application Error",
                              "An unknown and unexpected error occurred.");
        return true;
    }

    // This line should not be reached if an exception is thrown,
    // but it's good practice for compilers.
    return false;
}
```

gadfort:
Once the other ones merge I can take a stab at it and see if that works.

PR #8514, opened 2025-10-01
Use `inline constexpr` uniformly for constants in headers.

hzeller:
Adressing `clang-diagnostic-unused-const-variable` clang-tidy warning.

### DIFF
--- a/src/dst/src/LoadBalancer.h
+++ b/src/dst/src/LoadBalancer.h
@@ -24,8 +24,9 @@ namespace dst {
 
 namespace ip = asio::ip;
 
-const int kWorkersDiscoveryPeriod = 15;  // time in seconds between retrying to
-                                         // find new workers on the network
+// time in seconds between retrying to find new workers on the network
+inline constexpr int kWorkersDiscoveryPeriod = 15;
+
 class Distributed;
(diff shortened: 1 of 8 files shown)


src/dst/src/LoadBalancer.h
```diff
@@ -24,8 +24,9 @@ namespace dst {
 
 namespace ip = asio::ip;
 
-const int kWorkersDiscoveryPeriod = 15;  // time in seconds between retrying to
-                                         // find new workers on the network
+inline constexpr int kWorkersDiscoveryPeriod
+    = 15;  // time in seconds between retrying to
+           // find new workers on the network
```
maliberty:
```suggestion
// time in seconds between retrying to find new workers on the network
inline constexpr int kWorkersDiscoveryPeriod = 15;  
```

hzeller: Done.

hzeller: (one sec)

maliberty:
It didn't matter much here but in general it is helpful not to force push after a code review as it breaks the 'changes since last review' feature.

maliberty (approved):

PR #8524, opened 2025-10-02
dpl: remove pad and block masters from filler placement before placing

gadfort:
Changes:
- filter the list of filler masters before using them in the event that the user provided "extra" fillers which cannot be placed by `filler_placement`. In this case PAD cells and BLOCK cells.

Why:
- it turns out it is easy to glob more than you expect and later steps will fail if PADs or BLOCKs have been used. Blocks would be detected by the check_placement, but PAD cells do not trigger an issue.

There should be no functional changes to current setups (unless they have been silently adding more than they should).


gadfort:
@maliberty I don't believe the contributor guide has any information on the "buildifier", might be nice to add something about where to get it and how to run it.

src/dpl/src/FillerPlacement.cpp
```diff
@@ -49,6 +49,52 @@ Opendp::MasterByImplant Opendp::splitByImplant(
...
+                                        filtered_masters.end(),
+                                        [](dbMaster* master) -> bool {
+                                          if (master->isPad()) {
+                                            return true;
+                                          }
+                                          if (master->isBlock()) {
+                                            return true;
+                                          }
+                                          return false;
+                                        }),
```
maliberty:
```suggestion
                                          return master->isPad() || master->isBlock();
                                        }),
```

PR #8527, opened 2025-10-02
pad: correct use of r -> l to fix pad filler sorting

gadfort:
Fixes:
- minor typo in pad filler sorting that rendered it useless.
- Updated a test to provide fillers out of preferred placement order to ensure sorting is done.


gadfort:
@maliberty  bazel test failure seems unrelated

maliberty (approved):

PR #8533, opened 2025-10-03
README: Bazel docs pointer

oharboe:
I'm thinking front and center and a call to arms...

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,14 @@
 [![Documentation Status](https://readthedocs.org/projects/openroad/badge/?version=latest)](https://openroad.readthedocs.io/en/latest/?badge=latest)
 [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/5370/badge)](https://bestpractices.coreinfrastructure.org/en/projects/5370)
 
+## Building OpenROAD
+
+OpenROAD has grown beyond what CMake/Make can reasonably handle. With so many operating systems, use cases, and dependencies, we needed a build system designed for complexity. That’s why we chose Bazel - it gives us reproducibility, scalability, and flexibility to support everyone from researchers to industry users. Bazel can take some effort to learn, but we encourage you to dive in and explore it for your own workflows. Every contribution helps, and your experience with Bazel will make the whole community stronger.
+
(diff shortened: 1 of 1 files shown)


gadfort:
That blurb feels out of place at the top of the main readme. Maybe this belongs in the contributor guide or something? Or is this not intended for merging?

oharboe:
> That blurb feels out of place at the top of the main readme. Maybe this belongs in the contributor guide or something? Or is this not intended for merging?

I'm wondering what to do: we have bazel docs in various places close to the use-case, but apparently it is not being found/read.

I don't see any point in adding more bazel references in a place that isn't being read :-)

oharboe:
Maybe increasing Bazel doc readership is futile as long as the old flows are available?

gadfort:
@oharboe certainly doesn't help that bazel isn't mentioned on the build docs:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/docs/user/Build.md
Maybe merging Build.md and https://github.com/The-OpenROAD-Project/OpenROAD/blob/bb2b36f26048b7a8f5f83c7c49f7ce81c63b7fac/docs/user/Bazel.md would help in the transition to bazel.
At a minimum the build docs should reference Bazel.md otherwise there is literally no direct reference to bazel in the docs.

oharboe:
> @oharboe certainly doesn't help that bazel isn't mentioned on the build docs: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/docs/user/Build.md Maybe merging Build.md and https://github.com/The-OpenROAD-Project/OpenROAD/blob/bb2b36f26048b7a8f5f83c7c49f7ce81c63b7fac/docs/user/Bazel.md would help in the transition to bazel. At a minimum the build docs should reference Bazel.md otherwise there is literally no direct reference to bazel in the docs.

Actually, if you start with ORFS (more likely?) and not OpenROAD, then Bazel appears prominently...

https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/docs/user/BuildLocally.md


Perhaps we should have a single source of truth and not spread build information across two repositories...

PR #8538, opened 2025-10-03
Enable layering_check in test BUILD files.

hzeller:
There were a bunch of BUILD files that didn't have layering check enabled yet. This change enables it, and fixes the reported missing dependencies.

Unfortunately, not all tests are clean and import only visible headers, so disable layering check for these for a later fix.

### DIFF
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -1,6 +1,8 @@
 load("@openroad_rules_python//python:pip.bzl", "compile_pip_requirements")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 
+package(features = ["layering_check"])
+
 compile_pip_requirements(
     name = "requirements",
     src = "requirements.in",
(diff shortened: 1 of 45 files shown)


maliberty:
Is this something we can enable globally (eg .bazelrc)?

maliberty:
Looks to be a real issue

[2025-10-03T13:57:30.611Z] test/orfs/mock-array/simulate.cpp:7:10: error: module //test/orfs/mock-array:simulator_cts does not depend on a module exporting 'verilated_vcd_c.h'
[2025-10-03T13:57:30.611Z]     7 | #include "verilated_vcd_c.h"

hzeller:
It is possible to add `build --features=layering_check` so that this setting is the default.

For the `test/orfs/`, I've now disabled that `layering_check` there as I can't test it locally (requires some sort of docker config).

PR #8542, opened 2025-10-03
etc: try different mirrors for bison

eder-matheus:


gadfort:
@maliberty is this mergable? The current build on SC is failing since bison cannot be downloaded.

maliberty (approved):

PR #8574, opened 2025-10-07
test/orfs/mock-array: test IO_CONSTRAINTS and MACRO_PLACEMENT_TCL

oharboe:
There's no other live test that does this and since 8x8 takes substantially longer, this won't increase the max running time and will reduce the total workload for CI:

    bazelisk run test/orfs/mock-array:MockArray_4x4_flat_floorplan /tmp/floorplan print-IO_CONSTRAINTS

Outputs:

    IO_CONSTRAINTS=...

Then:

    bazelisk run test/orfs/mock-array:MockArray_4x4_flat_place_deps /tmp/place do-3_1_place_gp_skip_io

Outputs:

    All pins are placed. Skipping global placement without IOs

Macro placement similarly outputs.

    [INFO MPL-0017] No unfixed macros.

### DIFF
--- a/test/orfs/mock-array/BUILD
+++ b/test/orfs/mock-array/BUILD
@@ -1,24 +1,10 @@
+load("@bazel-orfs//:openroad.bzl", "orfs_run")
 load("@bazel-orfs//toolchains/scala:chisel.bzl", "chisel_binary")
 load(":mock-array.bzl", "config", "element", "mock_array", "verilog")
 
 package(
     features = ["-layering_check"],  # TODO: enable
 )
-
-# mock_array(
-#     name = "base",
-#     verilog = ":4x4",
-# )
-
-# mock_array(
-#     name = "flat",
-#     verilog = ":4x4",
-# )
-
-# mock_array(
-#     name = "8x8",
-#     verilog = ":8x8",
-# )
 
 chisel_binary(
     name = "generate_verilog",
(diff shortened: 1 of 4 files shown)


maliberty:
What is being tested?  I don't see any check on the output.

oharboe:
> What is being tested?  I don't see any check on the output.

It generates warning, not an error.

We talked about rules.json counting warnings.

That would catch it ....

oharboe:
@maliberty https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/1905 would close the loop

maliberty:
So this is testing that the pins and macros are all placed?  I still don't get what this is a test of.

oharboe:
> So this is testing that the pins and macros are all placed?  I still don't get what this is a test of.


This is testing the artifact pinning mechanics for macros and pins.

writing out pins, reading. Writing out macro placement reading.

It is testing OpenROAD *and* ORFS for this flow.

oharboe:
ping?

maliberty:
Sorry, lost track of this one.  If we had warning checking this would at most test that no warning is produced (and currently would only check no error is produced).  Is that right?  As we will remove mock array from ORFS, this would mean we are only using 4x4 from now on?

oharboe:
>  If we had warning checking this would at most test that no warning is produced (and currently would only check no error is produced).  Is that right? 

No. I do recall warnings added removed being caught, but also  ORFS flow issues(no macro layout and skip non io gp) + it provides significant fuzzing (it changes as openroad makes tiny tweaks to initial conditions). I dont have a way to tell what errors this have caught locally vs in CI on PRs.

I think this will create significant testing value and a net *decrease* in CI workload as we run 4x4 in flat and -hier.

> As we will remove mock array from ORFS, this would mean we are only using 4x4 from now on?


No, OR mock array tests 4x4 as well as 8x8. These test cases are now MUCH faster than they used to be in ORFS: I have whittled down running times where it didnt sacrifice testing value materially.

PR #8581, opened 2025-10-08
DPL: ignore pdn patch wires

osamahammad21:


maliberty (approved):
@gadfort please try this on your case

maliberty:
This is a stop gap solution

gadfort:
@maliberty yes, this is passing. thanks

PR #8584, opened 2025-10-08
Use style-guide recommended names in enum class DataType

hzeller:
The name `NO_DATA` conflicts with a system macro.


hzeller:
Looks like clang-tidy failed independently as it didn't find the correct clang-tidy version ?

hzeller:
I see that was already addressed in #8583

Issue #8558, opened 2025-10-06
ifp: Gap between power domains is hardcoded to 6 row heights and should be configurable


titan73:
### Describe the bug

The gap between power domains should be configurable. Currently it is hardcoded to 6 row heights which is too big.

```
  // The unit for power_domain_y_space is the site height. The real space is
  // power_domain_y_space * site_dy
  const int power_domain_y_space = 6;
```

<img width="551" height="414" alt="Image" src="https://github.com/user-attachments/assets/2c54df97-5bfe-478d-831d-637c2e65d1b8" />

Morever, this shouldn't be necessary to have a multiple of row height as this is a constraints from the DRM. 

This should be configurable with either:
  - a global command to set the value
  - a option to `initialize_floorplan` and `make_rows`

I have a preference to the first solution

This shouldn't be to hard. I can do PR.

Also note:
The site is 1.53 x 4.51.
6 row height is 27.06
In the pic, in X, the gap is ceil(27.06 / 1.53) * 1.53 = 27.64 which is correct.
In the pic in Y, the gap is 7 row heights (31.57) which is incorrect.

### Expected Behavior

The gap should be configurable.

### Environment

```shell
Latest OR
```

### To Reproduce

I can make a test case if needed.

### Relevant log output

```shell

```

titan73:
The gap on the right is also wrong. It's 26.01. It is < 27.05 so the min distance is not respected. It should be 27.54. 

<img width="1156" height="653" alt="Image" src="https://github.com/user-attachments/assets/25d8b821-eb2a-42fb-92bf-1d3ca43c692c" />

maliberty:
`I have a preference to the first solution`, why?  I'd rather avoid having more globals so I tend to prefer the option.

titan73:
> `I have a preference to the first solution`, why? I'd rather avoid having more globals so I tend to prefer the option.

I'm used to Innovus. Power domains can have different gaps so an option to `initialize_floorplan` and `make_rows` would make it global.

If we want an option, we could pass it to `set_domain_area` or `create_voltage_domain`. That said, I'm not sure different gaps can be supported now. So, in the meantime, maybe an option to `initialize_floorplan` and `make_rows` would be enough. What do you think?

maliberty:
You are the only one to ever ask about it so if you can live with a single gap that is fine with me.

PR #8595, opened 2025-10-09
ifp: Fix y gap between power domains

titan73:
1st bug described at https://github.com/The-OpenROAD-Project/OpenROAD/issues/8558#issue-3487395214.

2nd is described un the following pic:

single_double_row

The tests init_floorplan8.tcl and init_floorplan8.py cover the changes well.


titan73:
How can I fix the new no-diffs-in-ok-files check?

gadfort:
@titan73 you will need to run the ./save_defok to update the def and then update the okfile with ./save_ok

maliberty:
Note you will need to rerun the test before the save_ok to get rid of the diff.

titan73:
@gadfort @maliberty Thanks. Fixed.
I pushed a new commit in my branch but the PR as not been updated. Is there a a way to force the sync?

titan73:
Ah the sync just happened.

titan73:
pr-head fail to unrelated test remove_buffers1.

titan73:
Actually the tests init_floorplan8.tcl and init_floorplan8.py cover only the changes of the 1st fix.
I've just pushed init_floorplan_dbl_row.tcl and init_floorplan_dbl_row.py which cover changes of both fixes.
Basically it's the same test as init_floorplan8 but with double height sites.

PR #8604, opened 2025-10-10
Fix some missing includes.

hzeller:


hzeller:
the CI issue looks like an independent glitch.

PR #8606, opened 2025-10-10
only_use

shypark98:


shypark98:
Please modify this part for compatibility with the ArtNet code.

maliberty:
What does ONLY_USE  mean?

shypark98:
We use ONLY_USE to specify the cells that should be used for netlist generation — that is, only the cells listed in the spec will be used among those read from the LEF.

maliberty:
It seems wrong to limit yourself to only cells that are currently instantiated.  At a minimum you would want to include other sizes, no?

shypark98:
The goal of ArtNet is to generate an artificial netlist using the given list of cells, while leaving sizing to the downstream stages. In addition, the updated parts do not affect the existing code behavior (ArtNetSpec.cpp) — only the format of the final output file is changed.

maliberty:
par.write_artnet is failing and needs updating.

PR #8607, opened 2025-10-10
Fix missing `defi*.hpp` includes.

hzeller:
.. and a few Qt include.


stefanottili:
Just out of curiosity, how do you come up with these "missing" headers ?
The compile succeeded without them til now, so why are they needed ?

hzeller:
The symbols are accidentally visible because a header they include might include the header that is needed.
But the goal is that we don't rely on these accidental leakage of symbols.

I am finding these by looking at the output of `clang-tidy`, this is the `misc-include-cleaner` warning.

maliberty:
How many misc-include-cleaner remain?

hzeller:
IIRC, about 5000 left (I started with > 13000). About 1500 or so of these are due to includes that are included, but not needed (but I can only remove them at the very end after all the needed ones are there - currently the not needed ones might help the accidental availability of symbols). 

A lot of these will be fixed once I have landed some IWYU pragmas in boost (similar to the OpenSTA patch that is merged [there](https://github.com/parallaxsw/OpenSTA/blob/master/include/sta/Delay.hh#L29) after it lands in here). Other than that, some collection of symbols with some grep/awk/sed/rg means.

The generated db sources are source of a few hundred. Maybe I should make a new sample PR for these over the weekend.

PR #8612, opened 2025-10-11
pad: check corner cell placement for overlaps with covers

gadfort:
Expansion on https://github.com/The-OpenROAD-Project/OpenROAD/pull/8608

Avoids placing corners then they will overlap with fixed bumps that are placed there.


maliberty:
bzl needs fixing

PR #8614, opened 2025-10-11
Don't pollute other namespace with `odb::dbOrientType`

hzeller:
Fix `using odb::dbOrientType;` in header files in a different namepace, which pollutes that namespace with that alias.

### DIFF
--- a/src/dpl/src/optimization/detailed_orient.cxx
+++ b/src/dpl/src/optimization/detailed_orient.cxx
@@ -154,6 +154,8 @@ int DetailedOrient::orientCells(int& changed)
 ////////////////////////////////////////////////////////////////////////////////
 bool DetailedOrient::orientMultiHeightCellForRow(Node* ndi, int row)
 {
+  using odb::dbOrientType;
+
   // Takes a multi height cell and fixes its orientation so
   // that it is correct/agrees with the power stripes.
   // Return true is orientation is okay, otherwise false to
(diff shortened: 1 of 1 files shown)


hzeller:
the CI seemed to have glitched to unrelated reason.

src/dpl/src/infrastructure/Objects.cpp
```diff
@@ -372,10 +372,12 @@ bool Node::adjustCurrOrient(const dbOrientType& newOri)
...
-  if (curOri == dbOrientType::R90 || curOri == dbOrientType::MXR90
-      || curOri == dbOrientType::R270 || curOri == dbOrientType::MYR90) {
-    if (newOri == dbOrientType::R0 || newOri == dbOrientType::MY
-        || newOri == dbOrientType::MX || newOri == dbOrientType::R180) {
+  if (curOri == odb::dbOrientType::R90 || curOri == odb::dbOrientType::MXR90
+      || curOri == odb::dbOrientType::R270
+      || curOri == odb::dbOrientType::MYR90) {
+    if (newOri == odb::dbOrientType::R0 || newOri == odb::dbOrientType::MY
+        || newOri == odb::dbOrientType::MX
+        || newOri == odb::dbOrientType::R180) {
```
maliberty: I don't find this improves readability.  I think it is fine to remove using-decls from the header but as the guide allows it in .cc files, I would just move it there rather than change all uses in the those files.

hzeller: Yeah, this is an automated change using `sed`, so it does not take aestetics into account :)
Anyway, fixed this by adding a local using declaration.

PR #8617, opened 2025-10-11
pdn: error on complete pin coverage and warn on partial coverage

gadfort:
Closes https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3571

Changes:
- only generate error when pin is completely covered by an obstruction, if the pin is only partially covered emit a warning, since it may be possible to build the grid, with the obstruction, but if the grid fails to build there will be a warning to indicate a possible issue.

Note:
- I didn't want to put in an arbitrary percentage of overlap to fail on, if we find this will be needed, additional test cases will be needed to help determine that fail criteria.

### DIFF
--- a/src/pdn/src/grid.cpp
+++ b/src/pdn/src/grid.cpp
@@ -1727,7 +1727,7 @@ void InstanceGrid::checkSetup() const
         int64_t total_overlap = 0;
         std::string layer_txt;
         for (const auto* layer : layers) {
-          if (layer_txt.size() > 0) {
+          if (!layer_txt.empty()) {
             layer_txt += ", ";
           }
           layer_txt += layer->getName();


src/pdn/src/grid.cpp
```diff
@@ -1707,18 +1710,41 @@ void InstanceGrid::checkSetup() const
...
+                overlap_area += overlap.area();
+                pin_area += pin.area();
+              }
             }
           }
         }
       }
+
+      if (pin_area > 0 && overlap_area != 0) {
+        const float pct = 100 * static_cast<float>(overlap_area) / pin_area;
```
maliberty: Could you overlap with more than one box and the sum is full overlap?  It seems to only warn in that case

gadfort: good point, its possible for a pin to be covered by multiple boxes so the counting here is wrong, I'll fix that.

PR #8618, opened 2025-10-11
utl: ensure utl::metric correctly handles large integers

gadfort:
Ran into this issue running a "large" chip through and the site count exceeded the limit for int (likely an indicator of additional issues down the line).

### DIFF
--- a/src/utl/src/LoggerCommon.h
+++ b/src/utl/src/LoggerCommon.h
@@ -16,7 +16,7 @@ void critical(utl::ToolId tool, int id, const char* msg);
 void open_metrics(const char* metrics_filename);
 void close_metrics(const char* metrics_filename);
 void metric(const char* metric, const char* value);
-void metric_integer(const char* metric, const int64_t value);
+void metric_integer(const char* metric, int64_t value);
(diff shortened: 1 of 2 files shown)


src/utl/src/LoggerCommon.h
```diff
@@ -15,7 +16,7 @@ void critical(utl::ToolId tool, int id, const char* msg);
 void open_metrics(const char* metrics_filename);
 void close_metrics(const char* metrics_filename);
 void metric(const char* metric, const char* value);
-void metric_integer(const char* metric, const int value);
+void metric_integer(const char* metric, int64_t value);
```
rovinski, line 19: Dropping `const`?

hzeller: using const for integers in function prototypes is confusing, so this looks like an improvement.

maliberty: const is needed in the definition not the declaration

rovinski: Oh I missed the clang tidy

PR #8623, opened 2025-10-12
Don't use `using odb::dbBlock;` in header files to not pollute namespace

hzeller:


hzeller:
Also, fix `src/rcx/include/rcx/extRulesPattern.h` along the way that had a bunch of pre-existing `clang-diagnostic-error`s.

PR #8624, opened 2025-10-12
Fix typo in instance variable.

hzeller:
The variable in question does not exist, but there is an instance variable with underscore...


maliberty:
it is dead code which is why there is no compiler error

PR #8629, opened 2025-10-13
Fix missing comma in `.clang-tidy` config.

hzeller:
`clang-tidy --dump-config` shows that it had trouble around that line and the next and thus was not properly processing these.


hzeller:
This goes all the way back to 03bf3da00799e82169a11ed90461c3ed63ea9b36 (fyi @QuantamHD )

hzeller:
CI failure looks like independent glitch.

PR #8635, opened 2025-10-13
grt: use all APs for non-core cells

eder-matheus:
Necessary for DRT incremental pin access.

For non-core cells, DRT doesn't assign a preferred access point for the dbITerms. Therefore, GRT must consider all APs stored in ODB for these dbITerms, to ensure that at least one of their APs will be covered by the guides.

### DIFF
--- a/src/grt/src/Pin.cpp
+++ b/src/grt/src/Pin.cpp
@@ -192,7 +192,8 @@ bool Pin::isCorePin() const
   }
 
   odb::dbInst* inst = iterm->getInst();
-  return inst->isCore();;
+  return inst->isCore();
+  ;
 }
 
 }  // namespace grt


src/grt/src/Pin.cpp
```diff
@@ -185,4 +185,22 @@ odb::Point Pin::getPositionNearInstEdge(const odb::Rect& pin_box,
...
+  if (connected_to_pad_or_macro_) {
+    return false;
+  }
+
+  odb::dbInst* inst = iterm->getInst();
+  if (inst->getMaster()->isCore()) {
+    return true;
+  }
+
+  return false;
```
osamahammad21:
```suggestion
  return inst->isCore();
```

src/grt/src/Pin.cpp
```diff
@@ -185,4 +185,22 @@ odb::Point Pin::getPositionNearInstEdge(const odb::Rect& pin_box,
...
 
+bool Pin::isCorePin() const
+{
+  if (is_port_) {
+    return false;
+  }
+
+  if (connected_to_pad_or_macro_) {
+    return false;
+  }
```
osamahammad21: Isn't this covered by the following lines?

eder-matheus: I changed the checks for a simpler one.

PR #8636, opened 2025-10-13
odb: dont add each polygon to set, add them in bulk

gadfort:
Turns out the merging is insanely slow, this ensures boost performs a bulk add to the polygon set avoiding long recomputing periods.

Before on design with 21481 rows:
* Merge: 200.67s

After on same design:
* Merge: 0.07s


maliberty:
What operation did this show up in?

gadfort:
> What operation did this show up in?

After init floorplan when it calls the computeCoreArea function. Had it been 1 to 10 seconds I probably wouldn't have noticed, but 200 is a noticeable delay (measured with vtune)

PR #8643, opened 2025-10-14
gui: fix "through" filter in the timing widget

LucasYuki:
The "through" filter in the timing widget was only working when there was also a "from" selection.

Changes:
- TimingControlsDialog::getClocks can return nullptr if "all" is selected (small optimization)

This PR just fixes the bug only when all clocks are selected. 
I created the Issue #8671 for filtering "through" and clock at the same time.

### DIFF
--- a/src/gui/src/dropdownCheckboxes.h
+++ b/src/gui/src/dropdownCheckboxes.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <qchar.h>
+#include <qnamespace.h>
 #include <qobjectdefs.h>
 
 #include <QComboBox>


oharboe:
🥳

maliberty:
ready for review?

src/gui/src/staGui.h
```diff
@@ -410,7 +410,7 @@ class TimingControlsDialog : public QDialog
   std::set<const sta::Pin*> getFromPins() const { return from_->getPins(); }
   std::vector<std::set<const sta::Pin*>> getThruPins() const;
   std::set<const sta::Pin*> getToPins() const { return to_->getPins(); }
-  const sta::ClockSet* getClocks();
+  const sta::ClockSet* getClocks(bool return_null_if_all = true);
```
maliberty: What is the purpose of return_null_if_all?  I don't see anyone passing an argument so it seems to always be true.

src/gui/src/staGuiInterface.cpp
```diff
@@ -1119,6 +1120,7 @@ TimingPathList STAGuiInterface::getTimingPaths(
   }
 
   sta::Search* search = sta_->search();
+  search->clear();
```
maliberty: why is clear needed?

LucasYuki: Testing the timing widget, I notice that the last search can affect the current search, so I tried to clean the search and the "through" filter started to work again.

To reproduce the bug on Master (design asap7/ethmac):

- Filter "through" \_058181\_/A
  - No results found
image3

- Clear the "through" filter.
- Filter "from" wishbone.ShiftEndedSync\_c1$\_DFF\_PP0\_/QN
  - 2 results
image1

- Clear the "from" filter.
- Filter through \_058181\_/A
  - 1 result (the same result as filter "from" and "through" combined)
image4

maliberty: If -through gave none then combining any filter with it would be no results not one.  Clearing search is runtime expensive so I would like to understand the root cause rather than use this approach if possible.

maliberty: I tried to reproduce your result but `_058181_/A` doesn't match anything in my version.  Where can I find the version you used?

LucasYuki: sorry, here is the report of this design so you can reproduce https://drive.google.com/file/d/12fTm2kBktrYOxQ4MHbtX4DPc1i0X3f4f/view?usp=sharing

LucasYuki: I removed this line for this PR, since the other modification already solves the most common use of this feature, and created the Issue #8671 for filtering "through" and clock at the same time.

maliberty (approved):

PR #8651, opened 2025-10-14
Fix various namespace pollutions by `using odb::...` in header files

hzeller:


hzeller:
I've added a function local using-clause to improve readability, so auto-merge got disabled and needs to be re-enabled.

PR #8663, opened 2025-10-16
gui: Switch to inspector view only when double clicking on violation in drc viewer.

titan73:
Currently clicking on a violation in the drc viewer zooms to the violation in the renderer and switch to the inspector.
When quickly browsing several violations, we always have to switch back to the drc viewer for the next violation which is annoying.

This PR makes the changes:
- single click : Only zooms to the violation
- double click : Zooms to the violation and switch to the inspector


titan73:
I can add an option to enable the new behavior if needed. Let me know.

maliberty:
This makes sense to me.  @gadfort any concern?

gadfort:
@maliberty no, makes sense me as well

PR #8665, opened 2025-10-16
mpl: re-work utilization handling

### LINKED ISSUE #8732: MPL should use the utilization set by the user correctly
### Description

Currently there are two controls that handle how utilization varies across runs:
 - target_util which handles the inflation of Mixed clusters and whose initial value corresponds to the actual utilization set by the user;
 - target_dead_space which handles the inflation of Std Cell clusters and whose initial value comes from the tests made for the paper results (we begin at 5% which doesn't make any sense for our usage);

As we currently don't have multilevel physical hierarchies in our current test suite, we only the latter which means that literally all of our designs with macros are having their macros placed using a unreasonable value of utilization.

### Suggested Solution

Use the utilization correctly and have a single control for the utilization sweep.

### Additional Context

_No response_

### LINKED ISSUE #9073: MPL shouldn't expand tiny std cell clusters
### Description

Small regression from #8950 - we're expanding tiny std cell clusters when filling dead space after SA.

### Suggested Solution

Include this in the project to adapt internal utilization handling.

### Additional Context

_No response_

AcKoucher:
Resolve #8732,
Resolve #9073.

## Context

When we're placing clusters, we generate multiple annealing runs in order to attempt convergence. For each new run, we decrease the amount of dead space i.e., increase the utilization by 5% so that it becomes easier to fit the rectangles in the outline.

Currently there are two controls that handle how utilization varies across runs:
 - _target_util_ which handles the inflation of Mixed clusters;
 - _target_dead_space_ which handles the inflation of Std Cell clusters;

The rationale is cited in a comment in the code:
```
// During our experiements, we found that keeping
// the same utilization of standard-cell clusters and mixed cluster will make
// SA very difficult to find a feasible solution.  With different utilization,
// SA can more easily find the solution.
```
 
However, after quite some testing, it looks like the convergence problems MPL had were mainly due to different problems such as the shaping engine or the annealing engine not working due to bugs.

## Goals

- Retire _target_dead_space_ and use _target_utilization_ as the unique control for the inflation of both Mixed and Std Cell clusters;
- Actually aim for the target utilization set by the user;

## Changes

1. We now have a fixed amount of runs (10) for which the utilization increases exponentially all the way up to `1.0` (we want closer values to the target);
2. Choose the final result i.e., the "best" run based on utilization rather than cost. We now show an info message with the utilization that was used in cases in which we couldn't converge with the utilization specified by the user.
3. Fine shaping method refactoring*;
4. Simplifications and adaptations to tests now that the target utilization is respected;
5. Adaptations to pin access depth computations. These were made in order to: make the io density impact more noticeable (changes to how we compute `io_density_factor` - see methods for computing blockage depth) and avoid generating unreasonably deep blockages (lower max depth).

*The fine shaping function was doing many things at the same time. So I broke it into 3 separate pieces, each one of them takes care of a single thing:

- `setMacroClustersShapes`;
- `validUtilization` is to check if the current utilization value generates a total soft area that surpasses the available area for inflation.
- `applyUtilization` is the function that actually inflates the objects - I also fixed a bug that emerged from the float -> dbu work in which the tiny std cell clusters where not being evaluated as they should (now, for tiny std cell clusters `width == height == area == 1`. Clusters that were supposed to be considered tiny were not and ended up getting expanded by `fillDeadSpace` @joaomai FYI). _Now we also evaluate if a std cell cluster should be considered tiny only based on the amount of std cells in the design. Some specific test cases such as asap7/mock-array were affected by this change and so I had to add a specific pass for mock-array's corner case._

Also, there was a mechanism that verified if the tilings generated for macro/mixed clusters were valid i.e., fit in the outline. That was only needed, because the original code was written using floating-point types and used a convergence tolerance variable to deal with bad approximations. This mechanism was removed.

### DIFF
--- a/src/mpl/test/BUILD
+++ b/src/mpl/test/BUILD
@@ -135,17 +135,17 @@ filegroup(
                     "testcases/macro_only.lib",
                 ],
                 "io_constraints3": [
-                    "testcases/io_constraints1.def",
+                    "testcases/io_constraints3.def",
                     "testcases/macro_only.lef",
                     "testcases/macro_only.lib",
                 ],
                 "io_constraints4": [
-                    "testcases/io_constraints1.def",
+                    "testcases/io_constraints4.def",
                     "testcases/macro_only.lef",
                     "testcases/macro_only.lib",
                 ],
                 "io_constraints5": [
-                    "testcases/io_constraints1.def",
+                    "testcases/io_constraints4.def",
                     "testcases/macro_only.lef",
                     "testcases/macro_only.lib",
                 ],


AcKoucher:
The last push should fix the BAZEL failures.

maliberty (approved):

PR #8667, opened 2025-10-16
DRT: correct use of drcBox_ inside FlexGCWorker

osamahammad21:
In the case, of TritonRoute::getDRCMarkers, drWorker is null for the gc worker. This removes the incorrect use of the member drWorker_.


osamahammad21:
@gadfort FYI

PR #8670, opened 2025-10-16
CUT: wrap mio_library_ with unique_ptr

calewis:
mio_library_ was not being deleted AFAICT.


maliberty (approved):

calewis:
@maliberty when merging upstream master I had signing issues so I basically squashed and repushed.  Sorry that you might want to look at the changes again, but at least it's a short PR, the main thing since you approved was to make new copies of the library and suppress the stdout for abc.

PR #8673, opened 2025-10-17
ifp: Check -gap value is positive.

titan73:


src/ifp/src/InitFloorplan.tcl
```diff
@@ -113,6 +113,9 @@ proc make_rows_helper { key_array } {
   # Get gap space
   set gap -1
   if { [info exists keys(-gap)] } {
+    if { $keys(-gap) <= 0 } {
```
gadfort: probably a good idea to add a test for these two error conditions

titan73: Done thanks

titan73:
tcl and py tests differ now. Is there a way to have different ok files for py and test?
Should I create a new test specific for testing tcl errors?

maliberty:
Why do the results differ?  There is no mechanism as that is not a normal outcome.

titan73:
> Why do the results differ? There is no mechanism as that is not a normal outcome.

I added the error checking in tcl for -gap value which are not in py test.
```
[ERROR IFP-0061] -gap must be positive
[ERROR IFP-0061] -gap must be positive
[ERROR IFP-0065] -gap must be positive
[ERROR IFP-0065] -gap must be positive
```

maliberty:
Then you should add the same to python or move the check to c++

titan73:
> Then you should add the same to python or move the check to c++

It's not possible to have the check in python as the methods are directly mapped.
Sent a version with the check in C++. I use the min int32_t value for the default (6 min site height) and use the method for the check to have the same message id as python and tcl does not use the same code paths.

PR #8692, opened 2025-10-20
ram: additions to example tcl script and changes to ok files

braydenlouie:
- Added pdngen, global and detailed routing to example tcl script
- Added filler_instances to tcl script
- Revised defok and added lefok files

### DIFF
--- a/src/ram/test/make_8x8.tcl
+++ b/src/ram/test/make_8x8.tcl
@@ -1,5 +1,7 @@
 source "helpers.tcl"
 
+set_thread_count [expr [cpu_count] / 4]
+
 read_liberty sky130hd/sky130_fd_sc_hd__tt_025C_1v80.lib
 
 read_lef sky130hd/sky130hd.tlef
@@ -37,7 +39,7 @@ filler_placement {sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 \
 	sky130_fd_sc_hd__fill_4 sky130_fd_sc_hd__fill_8}
 
 global_route
-detailed_route
+detailed_route -verbose 0
 
 set lef_file [make_result_file make_8x8.lef]
 write_abstract_lef $lef_file
--- a/src/ram/test/make_8x8.ok
+++ b/src/ram/test/make_8x8.ok
@@ -6,4 +6,67 @@
 [INFO RAM-0016] Selected and2 cell sky130_fd_sc_hd__and2_0
 [INFO RAM-0016] Selected clock gate cell sky130_fd_sc_hd__dlclkp_1
 [INFO RAM-0016] Selected buffer cell sky130_fd_sc_hd__buf_1
+[INFO PDN-0001] Inserting grid: ram_grid
+Found 0 macro blocks.
+Using 2 tracks default min distance between IO pins.
+[INFO PPL-0067] Restrict pins [ D[0] D[1] D[2] D[3] D[4] ... ] to region 0.00u-120.52u at the TOP edge.
+[INFO PPL-0001] Number of available slots 280
+[INFO PPL-0002] Number of I/O             29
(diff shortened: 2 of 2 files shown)


src/ram/test/make_8x8.tcl
```diff
@@ -15,13 +15,33 @@ generate_ram_netlist \
...
+
+global_route
+detailed_route
+
+set lef_file [make_result_file make_8x8.lef]
+write_abstract_lef $lef_file
+diff_files make_8x8.lefok $lef_file
 
 set def_file [make_result_file make_8x8.def]
 write_def $def_file
```
rovinski, line 49: Just noticed that there's a missing
```
diff_files make_8x8.defok $def_file
```
here

src/sta
rovinski, line 1: No need to update sta in this PR

rovinski:
@maliberty can we get `make_8x8.lefok` on the whitelist? I assume we want a check for it.

maliberty:
When I run the 8x8 test I get
```
[ERROR PDN-0185] Insufficient width (0.00 um) to add straps on layer met2 in grid "ram_grid" with total strap width 23.0 um and offset 0.0 um.
```

maliberty:
> @maliberty can we get `make_8x8.lefok` on the whitelist? I assume we want a check for it.

done

src/ram/test/make_8x8.tcl
braydenlouie: I have the `diff_files` on line 48 on the make_8x8.tcl file. Does this still need to be fixed?

rovinski: Oh nevermind. It for some reason didn't show up in my file view even though I tried to extend to the end of the file.

src/ram/test/make_8x8.tcl
```diff
@@ -15,13 +15,33 @@ generate_ram_netlist \
...
-# filler_placement {sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 \
-# 	sky130_fd_sc_hd__fill_4 sky130_fd_sc_hd__fill_8}
+add_global_connection -net VDD -pin_pattern {^VPWR$} -power
+add_global_connection -net VSS -pin_pattern {^VGND$} -ground
+global_connect
+set_voltage_domain -power VDD -ground VSS
+define_pdn_grid -name ram_grid -voltage_domains {CORE}
+add_pdn_stripe -grid ram_grid -layer met1 -followpins -width 0.48
+add_pdn_stripe -grid ram_grid -layer met2 -width 0.48 -pitch 45
+add_pdn_stripe -grid ram_grid -layer met3 -width 0.48 -pitch 20
```
rovinski: @braydenlouie the `-pitch` values that you used don't seem to work. I think they are not the same as the ones you demo'd with in our last meeting. Can you reset them to values that work? Can you also double check that the widths are correct?

braydenlouie: I double checked the width values and they are the same that I demo'd in our last meeting. I switched the pitch value for the vertical layers to 45 when I was demo'ing it to you because I was running into issues with the `pin_placer` putting pins to close to the `pdn_stripes` without causing collisions in the routing. It should be changed back to the original values with a fix now.

maliberty: You needed to set the core area.  I've pushed a fix for that.

maliberty: Updated the unit test as well.

rovinski:
```
The following tests FAILED:
  	234 - dbSta.report_json1.tcl (Failed)                   IntegrationTest tcl dbSta log_compare
Errors while running CTest
```
@braydenlouie try merging in the master branch to this branch?

rovinski:
Although pr-merge is failing on the same thing too, so this seems to be an unrelated failure?

PR #8693, opened 2025-10-21
gpl: introduce region hard constraint

gudeh:
The main change in this PR is how we address regions in gpl. 

- Before:
We previously used dummies as a soft constraint for regions, allowing instances to be placed outside their region. This approach proved imperfect because the global placement could finish with instances outside their assigned regions. Also, for each region we used to have a bin grid based on the complete core area, including bins outside of the region scope.

- After:
Now we do not use dummies for regions anymore (we still use dummies for blockages and unusable sites). We still use a placerBase object and a nesterovBase object for each region, although their bin grid dimensions are defined by the region size. This introduces a hard constraint for regions in gpl, ensuring that instances never go outside their regions.

The top-level (placerBase object without any group or region) region space is defined as the core space.

This PR also includes multiple small changes.

- Functional changes:
  - Fillers starting position is now based on the region instead of the whole die. (In the previous version, they could be not only outside the region, but even outside the core and within the die. They would get relocated afterwards, but that is no longer required, as they now respect the region boundaries.)
  - Restrict initial placement positions of instances to be within the region area.
  - Prevent the max phi adjustment from being applied multiple times (previously we reduced the max value for each region), and ensure it never goes below 1.01. If the max phi goes below 1, the density penalty is reduced, which undesirably increases the importance of HPWL instead of density. Now each region has its own max phi, relocated from npVars to nbVars.
  - Introduce a new test case called region01 using two regions, based on issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/5197

- Graphics changes:
  - New GUI chart for density penalty values, protected by debugCheck().
 
image

  - Use a color palette for regions. Each region now uses the same color for fillers, bin forces, and dummies (the latter with a lighter tone).
  - Include the drawing of gradient-based vectors when an instance is selected: blue for density gradient, red for HPWL, and black for overall. The vectors correspond to the printed gradient values when the instance is selected.
image  
- Fix a bug when selecting an instance. We previously used nbVec_[0] (top-level) for any instance. Now we identify and use the proper nbVec_ that the instance belongs to when selected.

TODO: we still have some references to nbVec_[0] when drawing the gpl debug mode heatmaps (density and overflow).

### DIFF
--- a/src/gpl/src/graphicsImpl.cpp
+++ b/src/gpl/src/graphicsImpl.cpp
@@ -3,8 +3,6 @@
 
 #include "graphicsImpl.h"
 
-#include <spdlog/fmt/bundled/format.h>
-
 #include <algorithm>
 #include <cmath>
 #include <cstddef>


gudeh:
I triggered a secure-CI at: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-new-regions/1/pipeline

gudeh:
I was updating the tests with UPF at tools/OpenROAD/test, and noticed a potentially undesired behavior. We have the top-level (no region) fillers in dark magenta, fillers from the top region in blue and the fillers of bottom region in red. Notice how both fillers and normal gcells from the top-level are placed within the regions, we even get a strong density and undeteced density on the top region because of this.

output


We should not allow instances without regions to enter a region, correct @maliberty ?

I did not notice this before because other test cases do not have instances on the top-level.

maliberty:
We should consider the top like its own region and all cells in top (ie not in any region) should be there and not in another region.

gudeh:
For the top-level, I believe we should keep using the previous approach with dummies. The top-level should treat regions the same way it treats blockages.

This is because, in the new approach, the bin grid is defined by the region itself, and if we would do that for the top-level we would have to make holes in the top-level bin grid.

gudeh:
We’re getting a good result on the upf_aes test case! I couldn’t find any top-level instances placed inside a region. Also, the bin grid ensures that no instances are placed outside their assigned regions.

I made some changes to the colors of instances.

output

gudeh:
I see OR failures with "Build on RHE8L" which seem unrelated.

I just updated the metrics on the ORFS PR and expect everything to be green as soon as it finishes.

src/gpl/src/placerBase.cpp
```diff
@@ -990,6 +994,25 @@ PlacerBase::~PlacerBase()
...
   die_ = pbCommon_->getDie();
+  if (group_ != nullptr) {
+    auto boundaries = group_->getRegion()->getBoundaries();
+
+    if (!boundaries.empty()) {
+      region_bbox_.mergeInit();
+      for (auto boundary : boundaries) {
+        region_bbox_.merge(boundary->getBox());
+      }
+      region_area_ = region_bbox_.area();
```
maliberty, line 1005: This is a bit conservative as the boundaries might form a non-rectangular area.

PR #8702, opened 2025-10-22
bazel-orfs: bump

oharboe:


oharboe:
@maliberty non bazel CI failures. Merge?

PR #8705, opened 2025-10-22
Utility to temporarily suppress stdout

calewis:
Move the stdout suppresser from annealing_strategy.cpp to its own library.

I'd like to use this in #8670 to limit output from ABC.  I don't have a way to test on windows.


src/utl/src/SuppressStdout.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <cstdio>
+
+#include "utl/Logger.h"
+
+namespace utl {
+
+#ifdef _WIN32
```
maliberty: We don't support a WIN32 build so I would just remove this as it doesn't work anyways.

calewis: Done

maliberty:
needs a clang-format

PR #8706, opened 2025-10-22
Updated gitignore to ignore bazel-*. added come comments to src/dbSta/BUILD

hongted:


.gitignore
```diff
@@ -61,6 +61,7 @@ bazel-bin
 bazel-out
 bazel-testlogs
 bazel-OpenROAD
+bazel-*
```
maliberty, line 60: You can remove 
bazel-bin
bazel-out
bazel-testlogs
bazel-OpenROAD

hongted: Done

PR #8710, opened 2025-10-22
gpl: Factor out graphics use in gpl

hongted:


src/gpl/include/gpl/Replace.h
```diff
@@ -46,13 +48,21 @@ using Clusters = std::vector<Cluster>;
...
           rsz::Resizer* resizer,
           grt::GlobalRouter* router,
           utl::Logger* logger);
+
   ~Replace();
 
+  // Use the following class as a template for graphics interface.
+  //
+  // Note: no ownership is transfered as the object will create a new
+  // graphics boject of the same class.
```
maliberty:
```suggestion
  // graphics object of the same class.
```

src/gpl/src/AbstractGraphics.cpp
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2020-2025, The OpenROAD Authors
+
+#include "gpl/AbstractGraphics.h"
+
+namespace gpl {
+
+AbstractGraphics::~AbstractGraphics() = default;
```
maliberty, line 8: Why not put this in the header?

hongted: That's to define a key function for the class, without it dynamic_cast may not work across shared library boundaries, so I have a habit of making sure that there is one for every class. 

https://developer.android.com/ndk/guides/common-problems#rttiexceptions_not_working_across_library_boundaries

https://lld.llvm.org/missingkeyfunction.html

src/gpl/include/gpl/Replace.h
hongted: Done

src/gpl/src/AbstractGraphics.cpp
maliberty: There are likely many interfaces that don't follow this rule.  Since we don't use shared libraries it has never mattered but it could in the future so it is a good idea.

maliberty:
Please don't force push after code review has started as it breaks the "changes since last review" feature in GH (the last reviewed commit is elimiated).

src/gpl/include/gpl/AbstractGraphics.h
```diff
@@ -0,0 +1,138 @@
...
+class AbstractGraphics
+{
+ public:
+  using LineSeg = std::pair<odb::Point, odb::Point>;
+  using LineSegs = std::vector<LineSeg>;
+
+  virtual ~AbstractGraphics();
+
+  // Create a new object of the same class.
+  virtual std::unique_ptr<AbstractGraphics> MakeNew(utl::Logger* logger) const
```
maliberty, line 34: Why are we making copies of the graphics object rather than share one?

hongted: That's to match the existing behavior where each object has its own Graphics object.  The graphics object does have internal state, but I've not done the analysis to see if every object can share one.

https://github.com/The-OpenROAD-Project/OpenROAD/blob/41754504d3cdf2016bca9d9e78f3a3dd5217f9ca/src/gpl/src/nesterovPlace.cpp#L56-L63

https://github.com/The-OpenROAD-Project/OpenROAD/blob/41754504d3cdf2016bca9d9e78f3a3dd5217f9ca/src/gpl/src/mbff.cpp#L2498

https://github.com/The-OpenROAD-Project/OpenROAD/blob/41754504d3cdf2016bca9d9e78f3a3dd5217f9ca/src/gpl/src/initialPlace.cpp#L52

https://github.com/The-OpenROAD-Project/OpenROAD/blob/41754504d3cdf2016bca9d9e78f3a3dd5217f9ca/src/gpl/src/graphics.h#L135-L143

hongted:
> Please don't force push after code review has started as it breaks the "changes since last review" feature in GH (the last reviewed commit is elimiated).

Understood, I'll do so in the future.

Also, I don't know if it's the same as "changes since last review",  you can see the updates after a force-push via the "compare" link in this discussion.  It does include changes due to the rebase and not just the commit though.

gudeh:
I realized a small oversight on this change. We now only plot the GUI charts if in debug mode.

PR #8713, opened 2025-10-23
Test orfs warnings rules json baseline

oharboe:
@jhkim-pii FYI


test/orfs/gcd/rules-base.json
```diff
@@ -1,4 +1,74 @@
 {
+    "cts__flow__warnings__count:RSZ-62": {
+        "value": 1,
+        "compare": "<=",
+        "level": "warning"
+    },
+    "cts__flow__warnings__count:STA-1212": {
+        "value": 1001,
+        "compare": "<=",
```
oharboe, line 9: @jhkim-pii @maliberty Shouldn't this be ==?

If the number of warnings change, that could be good or bad...

maliberty: I think that is too strict.

oharboe: Could be... either it would save or create work... I think running with what we have for a while and learning and observing how this work in real life is good next step...

oharboe:
@maliberty Non bazel related CI errors 
image

PR #8720, opened 2025-10-23
gui: Updated singleton pattern for gui::Gui::get().

hongted:
The stub verson of gui::Gui::get() no longer returns nullptr.


hongted:
@maliberty -- This is needed to address code under which `gui::Get()` is called by code and assumed to be non-null.  While true when compiled under gui, cli mode compilation uses the stub implementation.

https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/gui/src/stub.cpp#L28
https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/gui/src/stub.cpp#L56-L59

Example place where I've assumed this:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/gpl/src/graphicsImpl.cpp#L30

maliberty:
This seems wrong - Gui::enabled() should prevent anyone from ever using the Gui object.

hongted:
Gui::enabled has a different function -- it returns true if there is a main window.  It isn't consistent, but it appears that certain Gui functions are written with that understanding.

That said, it is odd to me that a singleton would be able to return nullptr.  This PR fixes that with minimal cost.

maliberty:
If someone is using the gui without checking if it is enabled that is a bug.  This would just mask that.

hongted:
To confirm -- users are expected to call Gui::enabled() and ensure it is true before retrieving the Gui::Gui::get() singleton and any other method from said singleton object except for Gui::enabled()?

If that is what you want, then ok.  I just don't think that assumption is well known.  Even in Gui code, the assumption is that there is always a Gui object, and during a single program run, the Gui::enabled() can switch between true and false depending on the state of the program.

https://github.com/The-OpenROAD-Project/OpenROAD/blob/7cbd77e24833bca2a745fb5c85024728c3ff362c/src/gui/src/gui.cpp#L237-L241

https://github.com/The-OpenROAD-Project/OpenROAD/blob/7cbd77e24833bca2a745fb5c85024728c3ff362c/src/gui/src/gui.cpp#L776-L808

hongted:
This is also hard to debug, if you don't want the stub's Gui::gui() to return an actual object, would you be ok with having it assert and fail?

maliberty:
If you only have a stub it makes little difference.  If you have a real gui and it isn't enabled then you are invoking useless work.  @gadfort can you think of any case where you would want to do a gui operation when the gui is not enabled?

gadfort:
@maliberty no, returning the nullptr is expected when the GUI is not active

maliberty:
@hongted an assert in the stub is fine (or a critical error).

src/gui/src/gui.cpp
```diff
@@ -208,13 +209,12 @@ static void resetConversions()
...
 }
 
-Gui* Gui::singleton_ = nullptr;
-
 Gui* Gui::get()
 {
-  if (singleton_ == nullptr) {
-    singleton_ = new Gui();
-  }
+  static Gui* singleton_ = new Gui();
```
maliberty: No need for _ on a local variable.

src/gui/src/gui.cpp
```diff
@@ -208,13 +209,12 @@ static void resetConversions()
...
+
+  CHECK(enabled())
+      << "Internal Error: gui::Gui::get() called when enabled() is false";
```
maliberty: I think an error would suffice here and the check should be before allocating the object.

hongted: I'm going to move this to another PR as the fallout is a bit larger.

hongted: Draft PR created for reference -- https://github.com/The-OpenROAD-Project/OpenROAD/pull/8747 -- in order to document fallout.

src/gpl/src/graphicsImpl.cpp
```diff
@@ -27,7 +27,9 @@ namespace gpl {
 GraphicsImpl::GraphicsImpl(utl::Logger* logger)
     : HeatMapDataSource(logger, "gpl", "gpl"), logger_(logger), mode_(Mbff)
 {
-  gui::Gui::get()->registerRenderer(this);
+  if (gui::Gui::get() != nullptr) {
+    gui::Gui::get()->registerRenderer(this);
+  }
```
maliberty: This shouldn't be needed as you shouldn't create a GraphicsImpl if the gui isn't enabled.  An unregistered renderer won't do anything.

hongted: What would you like the code to read then?

hongted: Refer to https://github.com/The-OpenROAD-Project/OpenROAD/issues/8748 where GraphicsImpl functions are called when the gui isn't enabled.

hongted: Feel free to make any edits you feel is appropriate.  We can talk about this 1-1, if needed.

maliberty (approved):

PR #8729, opened 2025-10-24
mpl: fix convergence problems with large macro arrays

### LINKED ISSUE #8694: MPL should deal better with large arrays of macros with same signature
### Description

With proper macro grouping, arrays with >40 may fail in macro placement, because the annealer can't find the correct arrangement.

Example in ngt45/ariane133 with the changes in #8375:

Image

### Suggested Solution

Use similar techniques to mock-array.

### Additional Context

_No response_

AcKoucher:
Resolve #8694.

Changes here:
 - Create flag for macro arrays of any type (we only had a flag for arrays of interconnected macros such as mock-array, however, macro arrays should be handled regardless of the reason used to cluster the macros together);
 - Use only swap moves for macro arrays that have no empty space;
 - Only allow valid sa states for macro arrays with empty space. This should allow us to explore a bit more the possibilities for an array that doesn't fulfill its tiling e.g., an array of 15 macros in a 4x4 tiling.


AcKoucher:
Running Secure-CI.

maliberty:
does this need secure-ci/metrics?

AcKoucher:
@maliberty It does. I'm investigating a huge run time increase in a private design.

AcKoucher:
Recent GPL changes changed the metrics substantially. I'm rerunning Secure-CI to prevent conflicts.

AcKoucher:
This should be paired with [#3627](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3627).

AcKoucher:
@maliberty Would it be possible to merge this?

PR #8731, opened 2025-10-24
Use quote for vendored yaml-cpp dependency.

hzeller:
This will trigger a new `misc-include-cleaner` warning as the yaml-header does not have IWYU annotations. So we should ignore such clang-tidy warning for now.

I am upstreaming the necessary IWYU pragmas in
https://github.com/jbeder/yaml-cpp/pull/1378


hzeller:
all the clang-tidy complaints shall be ignored for this one.

hzeller:
... and the CI failure is unrelated

PR #8733, opened 2025-10-24
tcl: use vendored include and fix missing includes.

hzeller:


hzeller:
Looks like the CI failure is unrelated (some IOException in one of the jenkins runs)
and the clang-tidy: unclear why for that particular header the CI clang-tidy configuration wouldn't find the tcl.h file but everything else.

QuantamHD:
Will we need to patch this inside google3 Henner?

hzeller:
I think not @QuantamHD : internally we already deal with a mix of `"tcl.h"` and `<tcl.h>` like in this code. This just fixes the the places of `<tcl.h>` to be `"tcl.h"`

hzeller:
FYI, it looks like tcl.h itself includes some implementation headers, so I try to upstream the necessary IWYU pragmas

https://core.tcl-lang.org/tcl/tktview/c7dc599e73

Issue #8728, opened 2025-10-24
grt: connection between non-adjacent layers


jfgava:
### Description

This rare grt error started to appear last week.
`[ERROR GRT-0075] Connection between non-adjacent layers in net XXXX.`

It is happening in the grt end incremental post repair timing step.
The net that appears in the error is created during the repair_timing iterations.
There is something wrong when building the via stack for this net. 
It is connecting the first and last metal layers directly

jfgava:
Resolved with #8735

PR #8735, opened 2025-10-24
grt: create via stack when connecting merged nets

eder-matheus:
Fixes the multiple `[ERROR GRT-0075] Connection between non-adjacent layers in net net281114.` errors that happened in the last week.

Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/8728


eder-matheus:
FYI @jeffng-or and @precisionmoon

precisionmoon:
Does this have any QoR impact?

eder-matheus:
> Does this have any QoR impact?

No, it shouldn't impact QoR. It will just prevent the error we're having lately.

precisionmoon (approved):

PR #8739, opened 2025-10-25
.gitignore: make junk visible for clean

oharboe:
Narrow ignore so that junk is visible for clean

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,11 +57,8 @@ obj-x86_64-linux-gnu/
 
 user.bazelrc
 
-# Limit ignore to these files, so we see and clean out other cruft
-/bazel-bin
-/bazel-OpenROAD
-/bazel-out
-/bazel-testlogs
+# Limit ignore to these files, so we see and clean out other cruft that pollutes grep
+/bazel-*
 
 projectview.bazelproject
 .bsp/


oharboe:
@maliberty @vvbandeira Unrelated CI problems with pr-merge:

```
Cannot run program "nohup" (in directory "/tmp/workspace/OpenROAD-Public_PR-8739-merge"): error=2, No such file or directory
	java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1170)
	java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1089)
	Jenkins v2.522//hudson.Proc$LocalProc.(Proc.java:252)
```

maliberty:
You and @hongted need to work this out as he just made the opposite change.

oharboe:
@hongted I want to make junk visible and clean it out regularly. The reason is that junk creates noise when I search OpenROAD folder with grep or from my IDE.

Are you good with being more specific here?

Is there anything in particular that bazel-* ignores that you want to have ignored other than the explicit list that I have in this PR?

hongted:
The directories/symlinks that bazel creates depends o the working directory's basename.  Because of that depending on how you checked out the repo, you'll have different names.

https://github.com/bazelbuild/bazel/blob/f56e92799afb2578b77b60f43ea775752c3eef17/.gitignore#L25-L27

oharboe:
Good point.

Would /bazel-* work for you?

That limits it to the root folder.

hongted:
Yes, that will work and was my intention.

oharboe:
> Yes, that will work and was my intention.

Fixed.

@maliberty good to go.

PR #8740, opened 2025-10-25
.gitignore: make junk visible for clean

oharboe:
I think flow folders were created in OpenROAD in the distant past?


maliberty (approved):

maliberty:
Long ago we checked out orfs in a style similar to what you are doing with bazel-orfs

Issue #8723, opened 2025-10-24
Allow interactive updates to group_path


oharboe:
### Description

To see only paths running through the "foo" module:

1. Run `group_path -name foo -through *foo*`
2. Click "Update" in Endpoint Slack histogram
3. Select from dropdown

No paths are shown in Endpoint Slack histogram(blank), even if `report_checks -through *foo*` returns paths

The use-case is to avoid re-running the flow after adding the path_group to the .sdc just to add a path group. This sort of exploration is exactly what I'd like to do in seconds instead of hours to figure out where to look.

### Suggested Solution

Allow adding path groups interatively so as to use path groups in dropdown from Endpoint Slack histogram

maliberty:
I ran ng45/gcd with
```
export SYNTH_HIERARCHICAL = 1
export SYNTH_MINIMUM_KEEP_SIZE = 1
```
I ran `group_path -name foo -through *ctrl*`, clicked Update and see paths

<img width="578" height="551" alt="Image" src="https://github.com/user-attachments/assets/4579288a-d757-4aa6-bb57-3e670f3c08da" />

Please give a reproducible problem.

oharboe:
To reproduce:

```
$ bazelisk run test/orfs/mock-array:Element_4x4_base_synth /tmp/x open_synth
[deleted]
read_db -hier ./test/orfs/mock-array/results/asap7/Element/4x4_base/1_synth.odb
[WARNING ORD-0012] Hierarchical flow (-hier) is currently in development and may cause multiple issues. Do not use in production environments.
[deleted]
GUI_TIMING=1 reading timing, takes a little while for large designs...
% report_checks 
[deleted]

Startpoint: io_outs_down_mult/mod.a_registered[1]$_DFF_P_
            (rising edge-triggered flip-flop clocked by clock)
Endpoint: io_outs_down_mult/mod.pp_row3_1$_DFF_P_
          (rising edge-triggered flip-flop clocked by clock)
Path Group: reg2reg
Path Type: max

  Delay    Time   Description
---------------------------------------------------------
   0.00    0.00   clock clock (rise edge)
   0.00    0.00   clock network delay (ideal)
   0.00    0.00 ^ io_outs_down_mult/mod.a_registered[1]$_DFF_P_/CLK (DFFHQNx1_ASAP7_75t_R)
[...]

PR #8741, opened 2025-10-25
mock-array: more OpenSTA testing

oharboe:
Passes now after fixes on OpenROAD master :relieved: 

Previously failing test case for which this test was added:

    bazelisk test //test/orfs/mock-array:MockArray_4x4_base_power_cts_test  --experimental_ui_max_stdouterr_bytes=7000000

@jhkim-pii FYI, Megabytes of these warnings. Even if there is nothing wrong, these warnings are much to spammy. Ideally we could tell OpenSTA that this is OK or the warnings can be fixed/removed.

    Warning: bazel-out/k8-fastbuild/bin/test/orfs/mock-array/results/asap7/MockArray/4x4_base/../../Element/4x4_base/4_cts.spef line 587, instance io_outs_up_mult\/mod.b_registered\[2\]$_DFF_P_:D not found.

@maliberty Failing test case for https://github.com/The-OpenROAD-Project/OpenROAD/issues/8723 at the bottom the log when running the command above:

```
Checking if we can find timing paths from REG_10$_DFF_P_/QN to io_lsbs_10
Checking if we can create a group_path from REG_10$_DFF_P_/QN to io_lsbs_10
Warning: test/orfs/mock-array/power.tcl line 1, 'REG_10$_DFF_P_/QN' is not a valid start point.
Error: find_timing_paths -path_group test_group failed
//test/orfs/mock-array:MockArray_4x4_base_power_cts_test        FAILED TO BUILD
```

### DIFF
--- a/test/orfs/mock-array/path_groups.tcl
+++ b/test/orfs/mock-array/path_groups.tcl
@@ -10,7 +10,11 @@ if { [llength [find_timing_paths -from $from -to $to]] == 0 } {
   exit 1
 }
 puts "Checking if we can create a group_path from $from to $to"
-group_path -name test_group -from $from -to $to
+
+# There is no guarantee that the first node in the
+# tcl path structure is an exception start point, so use -through
+# to ensure we capture a valid path for our testing purposes.
+group_path -name test_group -through $from -to $to
 if { [llength [find_timing_paths -path_group test_group]] == 0 } {
   puts "Error: find_timing_paths -path_group test_group failed"
   exit 1


jhkim-pii:
@oharboe OK. Let me look into it.

jhkim-pii:
I found the reason of the massive warnings.
The escaping logic has an issue.

```
In src/dbSta/src/SpefWriter.cc:41,

std::string escapeSpecial(const std::string& name)
{
  std::string result = name;
  size_t pos = 0;
  while ((pos = result.find('$', pos)) != std::string::npos) {
    result.replace(pos, 1, "\\$");
    pos += 2;
  }
  pos = 0;
  while ((pos = result.find('/', pos)) != std::string::npos) {  // ISSUE: Escape all the '/', which is wrong.
                                                                // It should not escape the hierarchy delimiter.
    result.replace(pos, 1, "\\/");
    pos += 2;
  }
  return result;
}
```

The escaping logic should be enhanced correctly. 
e.g., It should be able to handle the following case.

```
hier1/hier2_name_has_\/char/instance_name_has\/char
```

I'll find a solution.

oharboe:
@jhkim-pii Do you have a suggestion for a modification to check that the .spef reading doesn't have any warnings?

https://github.com/The-OpenROAD-Project/OpenROAD/blob/7cbd77e24833bca2a745fb5c85024728c3ff362c/test/orfs/mock-array/load_power.tcl#L18-L23

oharboe:
@maliberty @jhkim-pii Narrow concerns per integration test. This means loading the designs a couple more times, but once the bizillion .spef warnings are fixed, I don't think loading of these small designs will (certainly shouldn't) take much time and smaller single concern integration tests have a much lower cognitive load when working with them

Path group test:

    bazelisk test --profile=analyze.txt //test/orfs/mock-array:MockArray_4x4_base_path_groups_cts_test --test_output=streamed --experimental_ui_max_stdouterr_bytes=10000000

jhkim-pii:
> @jhkim-pii Do you have a suggestion for a modification to check that the .spef reading doesn't have any warnings?
> 
> https://github.com/The-OpenROAD-Project/OpenROAD/blob/7cbd77e24833bca2a745fb5c85024728c3ff362c/test/orfs/mock-array/load_power.tcl#L18-L23

The `read_spef` warnings are generated by STA code, which is different from the typical OpenROAD warning messages. 
To check the OpenSTA warnings, I think you can make an additional step to grep the STA log file.

oharboe:
@jhkim-pii How can we most simply amend the rules.json log?

jhkim-pii:
The read_spef warnings are generated by Open STA run. So it is not related to `rules-*.json` because it is not a part of ORFS run.

oharboe:
> The read_spef warnings are generated by Open STA run. So it is not related to `rules-*.json` because it is not a part of ORFS run.

But the code is there, so "all" that is needed is to inject that information into rules.json files...

Bazel makes this sort of trickestry easy enough to integrate with ORFS.

jhkim-pii:
'OpenROAD-flow-scripts/flow/util/checkMetadata.py' reads `metadata.json` to report warning/errors.
To reuse the same metadata check flow of ORFS, the warning information of OpenSTA run should be fed into `metadata.json`.
It does not look trivial to me. Do you have an easy way to do this?

oharboe:
> 'OpenROAD-flow-scripts/flow/util/checkMetadata.py' reads `metadata.json` to report warning/errors. To reuse the same metadata check flow of ORFS, the warning information of OpenSTA run should be fed into `metadata.json`. It does not look trivial to me. Do you have an easy way to do this?

Let's look at that later, for now this should do the job: https://github.com/The-OpenROAD-Project/OpenROAD/pull/8743

maliberty:
@jhkim-pii we have dbStaReport::warn / vfileWarn which uses the logger for STA messages.  Doesn't that cover capturing sta messages?  What message do you see that isn't captured?

jhkim-pii:
> @jhkim-pii we have dbStaReport::warn / vfileWarn which uses the logger for STA messages. Doesn't that cover capturing sta messages? What message do you see that isn't captured?

Oh, I see. `dbStaReport` is a child class of `Report`. Then ORFS script can capture the STA warnings if OpenROAD binary is used instead of OpenSTA binary.

@oharboe We should use OpenROAD binary instead of OpenSTA binary in order to capture the `read_spef` warnings in `json` output file.

oharboe:
power.tcl only works from sta, it doesnt work woth openroad.

I forget the exact details why, but @maliberty concluded for power reporting I had to use sta standalone

jhkim-pii:
> power.tcl only works from sta, it doesnt work woth openroad.
> 
> I forget the exact details why, but @maliberty concluded for power reporting I had to use sta standalone

Oops. I see.

maliberty:
My recollection is that you wanted to report power on hiearchical modules and OR didn't have hierarchy at the time.  In theory it should be possible in OR with -hier now though nobody has tried it afaik.

oharboe:
> My recollection is that you wanted to report power on hiearchical modules and OR didn't have hierarchy at the time. In theory it should be possible in OR with -hier now though nobody has tried it afaik.

I tried to add one more test case for OpenROAD power reporting with mock-array, but it doesn't work yet:

https://github.com/The-OpenROAD-Project/OpenROAD/pull/8750

oharboe:
@maliberty @jhkim-pii My understanding is that the issues in this PR is being worked on first, so I'll hold of tinkering with the rest of the PRs to extend test/orfs/mock-array power testing.

jhkim-pii:
@oharboe @maliberty I found that the PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6405 caused the wrong escaping of the hierarchy delimiter '/'. So rolling back the PR can resolve the massive SPEF warning issue. 

But PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/6405 was intended to resolve another issue. I don't know what it was. Do you remember?

maliberty:
I don't recall but it was before we had hierarchy so I'm guessing it occurs in the flat case where we would write a name like `a/b` and sta would expect `a\/b`.  You may have to verify any change in both hier & flat modes.

jhkim-pii:
> I don't recall but it was before we had hierarchy so I'm guessing it occurs in the flat case where we would write a name like `a/b` and sta would expect `a\/b`. You may have to verify any change in both hier & flat modes.

Thank you.

I found that `dbNetwork::name(const Instance*)` cannot handle the escaped '/' appropriately in `a\/b` case. 
I think the solution is to fix `dbNetwork::name(const Instance*)` and remove '/' escaping code in `SpefWriter.cc`
I'll implement and test this in both flat and hier flows.

oharboe:
I'm hoping to cover all these use-cases in the test/orfs/mock-array tests...

jhkim-pii:
> I'm hoping to cover all these use-cases in the test/orfs/mock-array tests...

We can test it if `test/orfs/mock-array` verilog file contains the following cases.
1. \inst/name
2. inst\/name

The mock-array verilog is generated by Chisel. How can we do that in Chisel?

oharboe:
> > I'm hoping to cover all these use-cases in the test/orfs/mock-array tests...
> 
> We can test it if `test/orfs/mock-array` verilog file contains the following cases.
> 
> 1. \inst/name
> 2. inst/name
> 
> The mock-array verilog is generated by Chisel. How can we do that in Chisel?

OpenROAD only sees post-synthesis verilog, so whatever Chisel outputs is "gone" by the time OpenROAD sees it...

I think the job of mock-array is to test use-cases, I wonder if some OpenROAD unit-tests (the .tcl scripts + collateral files) is what is needed here in addition to mock-array use-case testing.

jhkim-pii:
@oharboe I'll add new Tcl regressions (in src/cts/test) to test the '/' escaping in both flat and hier flows.

maliberty:
I would think dbSta or est more natural place to test it as CTS has nothing to do with spef generation

oharboe:
@vvbandeira @maliberty unrelated CI error, bazel passed. Should be good to go.

maliberty (approved):

PR #8746, opened 2025-10-27
Fix some multi-chip issues.

Logikable:
dbChip::destroy: the dbDatabase's stored main chip ID should not be cleared when the chip being destroyed is not the main chip.

dbSta::postReadDef: reading a DEF for a chip that isn't the main chip should not update which block STA considers the main block.

### DIFF
--- a/src/dbSta/src/dbSta.cc
+++ b/src/dbSta/src/dbSta.cc
@@ -284,7 +284,8 @@ void dbSta::postReadLef(dbTech* tech, dbLib* library)
 
 void dbSta::postReadDef(dbBlock* block)
 {
-  if (!block->getParent()) {
+  // If this is the top block of the main chip:
+  if (!block->getParent() && block->getChip() == block->getDb()->getChip()) {
     db_network_->readDefAfter(block);
     db_cbk_->addOwner(block);
     db_cbk_->setNetwork(db_network_);
--- a/src/odb/src/db/dbChip.cpp
+++ b/src/odb/src/db/dbChip.cpp
@@ -662,7 +662,6 @@ void dbChip::destroy(dbChip* chip_)
   dbProperty::destroyProperties(chip);
   db->chip_hash_.remove(chip);
   db->chip_tbl_->destroy(chip);
-  db->_chip = 0;
 }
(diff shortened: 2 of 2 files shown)


src/odb/src/db/dbChip.cpp
```diff
@@ -662,8 +662,7 @@ void dbChip::destroy(dbChip* chip_)
   dbProperty::destroyProperties(chip);
   db->chip_hash_.remove(chip);
   db->chip_tbl_->destroy(chip);
-  db->_chip = 0;
```
maliberty, line 665: How does this accomplish "stored main chip ID should not be cleared when the chip being destroyed is not the main chip."?  It seems to be an unconditional change.

src/dbSta/src/dbSta.cc
```diff
@@ -284,7 +284,8 @@ void dbSta::postReadLef(dbTech* tech, dbLib* library)
 
 void dbSta::postReadDef(dbBlock* block)
 {
-  if (!block->getParent()) {
+  // If this is the top block of the main chip:
+  if (!block->getParent() && block->getChip() == block->getDb()->getChip()) {
```
maliberty, line 288: Why should the callback only happen for the top DEF?  Some clients may want to know about each DEF.

src/odb/src/db/dbChip.cpp
Logikable: L659-661 just above already clears the chip ID.

src/dbSta/src/dbSta.cc
Logikable: Currently: dbDatabase callback clients will all know about each DEF, but Sta will update only for the top block of each chip.

After this change: dbDatabase callback clients will still all know about each DEF, but Sta will update only for the top block _of the main chip_.

Sta is using the callback to set the block that owns the Sta instance. The "owner" of Sta should be in sync with who OpenROAD thinks is the main chip + top block. Therefore the owner should only be updated for the main chip.

Does this sound right?

osamahammad21: @maliberty this looks fine for now until accomodate for multi-chip network.

PR #8747, opened 2025-10-27
Added warning to gui::Gui::get() if called without enabled() being true.

hongted:


src/gpl/src/graphicsImpl.cpp
```diff
@@ -21,15 +20,16 @@
...
 
 namespace gpl {
 
 GraphicsImpl::GraphicsImpl(utl::Logger* logger)
     : HeatMapDataSource(logger, "gpl", "gpl"), logger_(logger), mode_(Mbff)
 {
-  gui::Gui::get()->registerRenderer(this);
+  if (gui::Gui::enabled()) {
+    gui::Gui::get()->registerRenderer(this);
+  }
```
gadfort, line 32: wont this always be false? The renderers can be registered without the main_window being open.

src/gui/src/gui.cpp
```diff
@@ -208,15 +238,19 @@ static void resetConversions()
...
 Gui* Gui::get()
 {
-  if (singleton_ == nullptr) {
-    singleton_ = new Gui();
+  if (!enabled()) {
+    ord::OpenRoad::openRoad()->getLogger()->warn(
+        utl::GUI,
+        150,
+        "Internal Error: gui::Gui::Get() called when enabled() is false\n{}",
+        GetSymbolizedStackTraceAsString(10, 0, true));
```
gadfort, line 248: wont this always trigger on the first get? The GUI class should exits before the main_window.

hongted: I think the desire is that the main window should exist before the GUI class per https://github.com/The-OpenROAD-Project/OpenROAD/issues/8748

src/gpl/src/graphicsImpl.cpp
hongted: Not always, Ideally we'd be able to call registerRenderer even without the main window, but gui::Gui::get() can be nullptr so we need a guard.

My hope is by the time it really matters -- here ie. https://github.com/The-OpenROAD-Project/OpenROAD/blob/dd7a99f782781bc8d3517fa3171cedca06e693a5/src/gpl/src/replace.cpp#L249 the main window is open.

Ideally, I'd prefer my original solution in https://github.com/The-OpenROAD-Project/OpenROAD/pull/8720/commits/de23da88abfb3341bd06d25e700296c1e0e426f8

PR #8374, opened 2025-09-19
gpl: improve divergence detction

gudeh:
This PR includes a new check for gpl detecting if a divergence has occurred.

As reported by @joaomai at branch https://github.com/The-OpenROAD-Project-private/OpenROAD/tree/mpl-new-blockages with nangate45/black_parrot there was an undetected divergence happening in gpl. The newly included check detects such divergence, and reverts to a valid solution, also providing a valid legalized placement and routable design.

make gui_place:
image

make gui_final:
image

I triggered a secure-CI at https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-improve-divergence-detection/2/pipeline and should be an no-op on current master version.

I noticed the divergence happened after the last timing-driven iteration, where RSZ removed highly inflated instances.

PR #8754, opened 2025-10-28
gpl: relax new divergence check

gudeh:
We recently added a new check for divergences based on the increase in overflow and HPWL over 10 iterations (PR #8374).

The motivation for this new check is that, when a divergence occurs, instances tend to move erratically, causing an increase in HPWL. This behavior is typically observed across multiple iterations, whereas the previous check only compared consecutive iterations.

The only issue with the new check appears to be that it is too strict (causing a false positive detection). Therefore, this PR relaxes the acceptable differences to 5% for overflow and 30% for HPWL, up from the previous thresholds of 2% and 5%, respectively.

This PR also includes a few minor changes:

- Fixed incorrect area change calculations from `rsz` modifications and added counting of deleted instances.  
- Included a function checkConsistency() to check for unexpected modifications with density related values due to area change.
- Renamed functions for clarity (for example, both `NesterovBase` and `NesterovPlace` had a function named `updateGradients`, which often caused confusion when reading the code).  
- Added generation of the first and last images in debug mode when using the `-generate_images` option.

### DIFF
--- a/src/gpl/test/simple01-rd.ok
+++ b/src/gpl/test/simple01-rd.ok
@@ -65,8 +65,8 @@ Iteration | Overflow |     HPWL (um) |  HPWL(%) |   Penalty | Group
       190 |   0.6798 |  2.191922e+03 |   +1.30% |  8.46e-10 |      
       200 |   0.6576 |  2.228829e+03 |   +1.68% |  1.38e-09 |      
       210 |   0.6213 |  2.261743e+03 |   +1.48% |  2.24e-09 |      
-[INFO GPL-0038] Routability snapshot saved at iter = 216
-   Iter: 217, overflow: 0.598, HPWL: 4564337
+[INFO GPL-0038] Routability snapshot saved at iter = 217
+      216 |   0.5977 |  2.282169e+03 |          |           |      
       220 |   0.5823 |  2.300463e+03 |   +1.71% |  3.66e-09 |      
       230 |   0.5418 |  2.340447e+03 |   +1.74% |  5.96e-09 |      
       240 |   0.4880 |  2.354778e+03 |   +0.61% |  9.70e-09 |      
--- a/src/gpl/test/simple02-rd.ok
+++ b/src/gpl/test/simple02-rd.ok
@@ -65,8 +65,8 @@ Iteration | Overflow |     HPWL (um) |  HPWL(%) |   Penalty | Group
       190 |   0.6798 |  2.191922e+03 |   +1.30% |  8.46e-10 |      
       200 |   0.6576 |  2.228829e+03 |   +1.68% |  1.38e-09 |      
       210 |   0.6213 |  2.261743e+03 |   +1.48% |  2.24e-09 |      
-[INFO GPL-0038] Routability snapshot saved at iter = 216
-   Iter: 217, overflow: 0.598, HPWL: 4564337
+[INFO GPL-0038] Routability snapshot saved at iter = 217
+      216 |   0.5977 |  2.282169e+03 |          |           |      
       220 |   0.5823 |  2.300463e+03 |   +1.71% |  3.66e-09 |      
       230 |   0.5418 |  2.340447e+03 |   +1.74% |  5.96e-09 |      
       240 |   0.4880 |  2.354778e+03 |   +0.61% |  9.70e-09 |      
--- a/src/gpl/test/simple03-rd.ok
+++ b/src/gpl/test/simple03-rd.ok
@@ -64,8 +64,8 @@ Iteration | Overflow |     HPWL (um) |  HPWL(%) |   Penalty | Group
       190 |   0.6798 |  2.191922e+03 |   +1.30% |  8.46e-10 |      
       200 |   0.6576 |  2.228829e+03 |   +1.68% |  1.38e-09 |      
       210 |   0.6213 |  2.261743e+03 |   +1.48% |  2.24e-09 |      
-[INFO GPL-0038] Routability snapshot saved at iter = 216
-   Iter: 217, overflow: 0.598, HPWL: 4564337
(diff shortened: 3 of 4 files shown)


src/gpl/src/nesterovBase.cpp
```diff
@@ -2014,16 +2017,75 @@ void NesterovBase::setTargetDensity(float density)
...
 
+void NesterovBase::checkConsistency()
+{
+  if (!log_->debugCheck(GPL, "checkConsistency", 1)) {
+    return;
+  }
+  auto block = pb_->db()->getChip()->getBlock();
+  const int64_t tolerance = 10000;
+
+  int64_t expected_white_space
```
maliberty: const - same with all variables in this method

src/gpl/src/nesterovBase.cpp
```diff
@@ -2553,12 +2615,6 @@ void NesterovBase::updateSingleGradient(
       = nbc_->getWireLengthGradientWA(gCell, wlCoeffX, wlCoeffY);
   densityGrads[gCellIndex] = getDensityGradient(gCell);
 
-  wireLengthGradSum_ += std::fabs(wireLengthGrads[gCellIndex].x);
-  wireLengthGradSum_ += std::fabs(wireLengthGrads[gCellIndex].y);
-
-  densityGradSum_ += std::fabs(densityGrads[gCellIndex].x);
-  densityGradSum_ += std::fabs(densityGrads[gCellIndex].y);
```
maliberty, line 2560: Why removed?

src/gpl/src/nesterovBase.cpp
```diff
@@ -2985,8 +3042,17 @@ bool NesterovBase::checkDivergence()
     float hpwl_increase = (static_cast<float>(prev_hpwl_ - prev_reported_hpwl_))
                           / static_cast<float>(prev_reported_hpwl_);
 
-    if (overflow_change >= 0.02f && hpwl_increase >= 0.05f) {
+    float overflow_acceptance = 0.05f;
+    float hpwl_acceptance = 0.25f;
```
maliberty: const

src/gpl/src/nesterovPlace.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include "nesterovPlace.h"
 
+#include <spdlog/fmt/bundled/format.h>
```
maliberty: Just use fmt.h - I've added a clang-tidy exception for it.

src/gpl/src/nesterovPlace.cpp
```diff
@@ -426,29 +430,36 @@ void NesterovPlace::runTimingDriven(int iter,
...
 
+    int nbc_total_gcells_delta
+        = nbc_->getNewGcellsCount() - nbc_->getDeletedGcellsCount();
     td_accumulated_delta_area += nbc_->getDeltaArea();
     for (auto& nb : nbVec_) {
       nb_gcells_after_td += nb->getGCells().size();
     }
-    nb_total_gcells_delta = nb_gcells_after_td - nb_gcells_before_td;
-    if (nb_total_gcells_delta != nbc_->getNewGcellsCount()) {
+    int nb_total_gcells_delta = nb_gcells_after_td - nb_gcells_before_td;
```
maliberty: const

src/gpl/test/simple01-rd.ok
```diff
@@ -66,6 +66,7 @@ Iteration | Overflow |     HPWL (um) |  HPWL(%) |   Penalty | Group
       200 |   0.6576 |  2.228829e+03 |   +1.68% |  1.38e-09 |      
       210 |   0.6213 |  2.261743e+03 |   +1.48% |  2.24e-09 |      
 [INFO GPL-0038] Routability snapshot saved at iter = 216
+   Iter: 217, overflow: 0.598, HPWL: 4564337
       220 |   0.5823 |  2.300463e+03 |   +1.71% |  3.66e-09 |
```
maliberty: Why not use the same format as the other iterations?

src/gpl/src/nesterovBase.cpp
gudeh: I realized this has no effect. We do the sum for all gcells at updateNextIter() afterwards. The removed part is the version for a single gcell which is used after callbacks.

src/gpl/src/nesterovPlace.cpp
gudeh: should we have to modify cmake for this? I did it based on other usages:

```
dbSta/src/dbSdcNetwork.cc:12:#include "spdlog/fmt/fmt.h"
grep: gpl/doc/image/adaptec2.inf.gif: binary file matches
gpl/test/fft_test.cc:8:#include "spdlog/fmt/fmt.h"
utl/include/utl/Logger.h:29:#include "spdlog/fmt/fmt.h"
utl/src/Metrics.cpp:11:#include "spdlog/fmt/fmt.h"
rsz/src/BufferedNet.hh:11:#include "spdlog/fmt/fmt.h"
rsz/src/BufferedNet.cc:22:#include "spdlog/fmt/fmt.h"
dft/src/utils/Formatting.hh:7:#include "spdlog/fmt/fmt.h"
odb/test/cpp/CallBack.h:10:#include "spdlog/fmt/fmt.h"
```

maliberty: I modified clang-tidy
https://github.com/The-OpenROAD-Project/OpenROAD/blob/5114b0e8815189fc467ddc349443eaabdb73df3f/.clang-tidy#L66

maliberty: We can do it separately globally

gudeh:
You think it is good to merge @maliberty? I see some errors on the CI but seem unrelated. Secure-CI is green with the single update in the ORFS PR

PR #8758, opened 2025-10-28
Use ABC fork in Bazel

kbieganski:
Currently, the Bazel build uses ABC from [bazel_rules_hdl](https://github.com/hdl/bazel_rules_hdl), whereas the CMake build uses the ABC submodule. The submodule one is older than the Bazel one.

One problem with this is that some `rmp` tests fail when built with Bazel due to producing different output.

Using one ABC for both build systems will fix that and prevent version drift in the future.

PR that adds Bazel to the ABC fork: https://github.com/The-OpenROAD-Project/abc/pull/8

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -104,7 +104,7 @@ OPENROAD_LIBRARY_DEPS = [
     "//src/upf:ui",
     "//src/utl",
     "//src/utl:ui",
-    "@edu_berkeley_abc//:abc-lib",
+    "//third-party/abc:abc-lib",
     ":ord",
 ] + select(
     {
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -25,7 +25,6 @@ http_archive(
 # This essentially reads as a TODO list of what needs to be upstreamed to BCR
 load("@rules_hdl//dependency_support/com_github_quantamhd_lemon:com_github_quantamhd_lemon.bzl", "com_github_quantamhd_lemon")
 load("@rules_hdl//dependency_support/com_github_westes_flex:com_github_westes_flex.bzl", "com_github_westes_flex")
-load("@rules_hdl//dependency_support/edu_berkeley_abc:edu_berkeley_abc.bzl", "edu_berkeley_abc")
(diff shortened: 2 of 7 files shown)


maliberty:
Once we switch to bazel we can drop the abc fork.  I am also fine to update from the upstream it if that is useful.

BUILD.bazel
```diff
@@ -104,7 +104,7 @@ OPENROAD_LIBRARY_DEPS = [
     "//src/upf:ui",
     "//src/utl",
     "//src/utl:ui",
-    "@edu_berkeley_abc//:abc-lib",
+    "//third-party/abc:abc-lib",
```
oharboe, line 107: https://registry.bazel.build/modules/abc is not an option?

kbieganski: I'm okay with any ABC version, the point is to just use one, so I went with the one that's already a submodule in OpenROAD.

oharboe: Yes.

Also, we should move away from using WORKSPACE and use MODULE.bazel when we can...

kbieganski:
> Once we switch to bazel we can drop the abc fork.

Sure, I'm happy with either one, but for now let's keep the two build methods in sync.

WORKSPACE
```diff
@@ -37,8 +36,6 @@ load("@rules_hdl//dependency_support/tk_tcl:tk_tcl.bzl", "tk_tcl")
 # Direct dependencies needed in Openroad
 com_github_quantamhd_lemon()
```
oharboe, line 38: :relieved:

maliberty:
Please merge master into this branch which should help resolve the CI issue.

kbieganski:
The CI failures seem random/unrelated.

vvbandeira:
> The CI failures seem random/unrelated.

The erros in pr-head are due to the PR branch being out of date with master. Since pr-merge is green, this is not a blocker -- no action needed. 

FYI @maliberty

maliberty (approved):

PR #8774, opened 2025-11-03
odb: add 3dbvWriter

rafaelmoresco:
Base 3dbvWriter.

When called directly by CLI, does the writeFile function, writing all chiplets defs in the same level. The function writeChiplet is made to be called by the 3dbxWriter, writing all the dependencies of a top 3dbx chiplet, with a call back to the 3dbxWriter if a hier chiplet is found.


src/odb/src/3dblox/baseWriter.h
```diff
@@ -0,0 +1,44 @@
...
+
+ protected:
+  // Common YAML content writing
+  void writeHeader(YAML::Node& header_node, odb::dbDatabase* db);
+  void writeLef(YAML::Node& external_node,
+                odb::dbDatabase* db,
+                odb::dbChip* chiplet);
+  void writeDef(YAML::Node& external_node,
+                odb::dbDatabase* db,
+                odb::dbChip* chiplet);
```
osamahammad21: Should be moved to dbvWriter as dbxWriter should never handle lef and def.

src/odb/src/3dblox/baseWriter.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+namespace odb {
+
+BaseWriter::BaseWriter(utl::Logger* logger) : logger_(logger)
+{
+}
+
+void BaseWriter::writeHeader(YAML::Node& header_node, odb::dbDatabase* db)
+{
+  header_node["version"] = "2.5";  // TODO: add version to DB
```
osamahammad21: Version 3.
I don't think it should be a variable in the database. We will always write in the most updated version. We will not write in the read version.

src/odb/src/3dblox/baseWriter.cpp
```diff
@@ -0,0 +1,125 @@
...
+      ++cnt;
+    }
+  } else if (db->getTech()) {
+    utl::OutStreamHandler stream_handler(
+        (std::string(chiplet->getName()) + ".lef").c_str());
+    odb::lefout lef_writer(logger_, stream_handler.getStream());
+    lef_writer.writeTech(db->getTech());
+    external_node["APR_tech_file"] = (std::string(chiplet->getName()) + ".lef");
+  }
+}
```
osamahammad21: Couple of comments on this:
- It's better and cleaner to separate the tech lef from the lib when we write, even though we don't enforce this when we read. So we shouldn't use writeTechAndLib. 
- We should infer the tech used by the current chip instead of using `db->getTech()`. we should use `chip->getTech()`. This would not work correctly on a design that has multiple technologies
- We should resolve which libs are being used in the current chip instead of linking it to all libs in the current database. This could be done by going through the insts of the chip's block and collecting the set of libs used.

src/odb/src/3dblox/dbvWriter.cpp
```diff
@@ -0,0 +1,251 @@
...
+    YAML::Node chiplet_node = chiplets_node[dependecy->chip->getName()];
+    writeChipletInternal(chiplet_node, dependecy->chip, db);
+  }
+
+  writeYamlToFile(filename, root);
+}
+
+void DbvWriter::writeChiplet(const std::string& base_filename,
+                             odb::dbDatabase* db,
+                             odb::dbChip* top_chip)
```
osamahammad21: You could get one from the other. Only one is sufficient, either db or top_chip.

src/odb/src/3dblox/dbvWriter.cpp
```diff
@@ -0,0 +1,251 @@
...
+    YAML::Node includes_node = header_node["include"];
+    for (const auto& include : included) {
+      includes_node.push_back(include);
+    }
+  }
+}
+
+void DbvWriter::writeChipletToFile(const std::string& filename,
+                                   odb::dbChip* chiplet,
+                                   odb::dbDatabase* db,
```
osamahammad21: ditto. in that case, chiplet is sufficient.

src/odb/src/3dblox/dbvWriter.cpp
```diff
@@ -0,0 +1,251 @@
...
+  }
+  YAML::Node coords_node = region_node["coords"];
+  writeCoordinates(coords_node, region->getBox(), db);
+}
+
+void DbvWriter::writeExternal(YAML::Node& external_node,
+                              odb::dbChip* chiplet,
+                              odb::dbDatabase* db)
+{
+  BaseWriter::writeLef(external_node, db, chiplet);
```
osamahammad21: only for non-Hier chiplets

src/odb/src/3dblox/baseWriter.h
```diff
@@ -0,0 +1,43 @@
...
+  void logError(const std::string& message);
+  std::string trim(const std::string& str);
+  void writeYamlToFile(const std::string& filename, const YAML::Node& root);
+  template <typename IntType>
+  std::string dbuToMicron(IntType dbu) const;
+
+  // Member variables
+  utl::Logger* logger_ = nullptr;
+  std::string current_dir_path_;
+  uint dbu_per_micron_ = 0;
```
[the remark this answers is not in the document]
osamahammad21: @maliberty is this really needed?

osamahammad21:
@maliberty Could we merge this PR? clang-tidy keeps complaining about using yaml-cpp/yaml.h and wants us to include the needed headers.

maliberty:
@osamahammad21 I believe if you had approved the PR you would have been able to merge it.

PR #8775, opened 2025-11-03
odb: add 3dbxWriter

rafaelmoresco:
Add 3dbxWriter.

Still not connected to main code, first commit is for analysis only. 3dbvWriter is needed for proper working. CLI would call the writeFile function, where the user gives the filename. This function will call the 3dbv writer for the top chiplet, writting all the dependencies in a file. In the case of a hier chiplet, 3dbvWriter will call the writeChiplet function, that has the same internal working but creates its own file name.

### DIFF
--- a/src/odb/src/3dblox/dbxWriter.cpp
+++ b/src/odb/src/3dblox/dbxWriter.cpp
@@ -3,8 +3,6 @@
 
 #include "dbxWriter.h"
 
-#include <yaml-cpp/emitterstyle.h>
-#include <yaml-cpp/node/node.h>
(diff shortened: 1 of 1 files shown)


osamahammad21 (approved):

povik:
@osamahammad21 FYI, CI failing on master after this merge

osamahammad21:
@povik I think the failure is irrelevant to this PR.
`[2025-12-11T13:34:26.274Z] //test/orfs/mock-array:MockArray_8x8_base_cts_openroad_test     FAILED TO BUILD`

PR #8777, opened 2025-11-03
gpl: debug mode colors and charts

gudeh:
This PR includes new charts when set_debug_level is enabled with "debugPlot".

It also changes the coloring in gpl debug mode for modifications made to instances, now rsz changes are more detailed: red for new instances, orange for upsize, and blue for downsize.

### DIFF
--- a/src/gpl/src/nesterovBase.h
+++ b/src/gpl/src/nesterovBase.h
@@ -1192,11 +1192,10 @@ class NesterovBase
   float uniformTargetDensity_ = 0;
 
   // StepLength parameters (also included in the np debugPrint)
+  // alpha
   float stepLength_ = 0;
   float coordiDistance_ = 0;
   float gradDistance_ = 0;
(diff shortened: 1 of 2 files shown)


src/gpl/src/nesterovBase.h
```diff
@@ -1228,7 +1244,7 @@ class NesterovBase
   float densityGradSum_ = 0;
 
   // alpha
-  float stepLength_ = 0;
+  // float stepLength_ = 0;
```
maliberty: rm

maliberty:
Generally fine but I see a bunch of commented out code.  Its better to remove such unless you are imminently going to enable it.

gudeh:
I removed the comments. Here are some examples of the new plots and colors:

image

image

output

maliberty:
What cause the large jump in the density penalty?

gudeh:
It's expected. The density penalty grows exponentially. If we plot in log scale we see a straight line.

PR #8778, opened 2025-11-03
odb: Nested ECO journaling

erendn:
@povik was working on nested ECO journaling that wasn't completed. I finished what he started.

This PR adds the following:
1. Each `odb::dbBlock` holds a `std::vector<dbJournal*>` (ECO history) to keep track of the uncommitted ECOs.
1. `odb::dbDatabase` pushes ECOs to the history.
1. `odb::dbDatabase` merges a committed ECO with the previous ECO from the history so we can still undo the previous ECO.
1. Updates `debugPrint` calls with the new behavior.
1. Adds a few missing `utl::Logger::critical` calls to prevent undefined behavior. (`odb::dbJournal::undo` is incomplete so undo will fail if OpenROAD attempts to undo the unimplemented things.)
1. Fixes a unit test (`rsz/repair_setup_undo1.tcl`) and adds another coverage test (`rsz/repair_setup_undo2.tcl`). (The fixed unit test was supposed to undo an entire `repair_timing -setup` call but didn't. This PR fixes that.)

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -1756,12 +1756,12 @@ class dbNet : public dbObject
 {
  public:
   ///
-  /// Get the net name.
+  /// Get the hierarchical net name (not a base name).
   ///
   std::string getName() const;
 
   ///
-  /// Get the net name.
+  /// Need a version that does not do strdup every time
   ///
   const char* getConstName() const;
 
@@ -2520,6 +2520,8 @@ class dbNet : public dbObject
   /// Dump dbNet info for debugging
   ///
   void dump() const;
+
+  void checkSanityModNetConsistency() const;
 };
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -2532,14 +2534,14 @@ class dbInst : public dbObject
 {
  public:
   ///
-  /// Get the instance name.
+  /// Get the hierarchical instance name (not a base name).
   ///
-  std::string getName();
+  std::string getName() const;
 
   ///
   /// Need a version that does not do strdup every time
   ///
-  const char* getConstName();
+  const char* getConstName() const;
 
   ///
   /// Compare, like !strcmp
@@ -2828,7 +2830,7 @@ class dbInst : public dbObject
   ///
   /// Get the block of this instance.
   ///
-  dbBlock* getBlock();
+  dbBlock* getBlock() const;
 
   ///
   /// Get the Master of this instance.
--- a/src/odb/src/db/dbBTerm.cpp
+++ b/src/odb/src/db/dbBTerm.cpp
@@ -619,7 +619,11 @@ dbBTerm* dbBTerm::create(dbNet* net_, const char* name)
                utl::ODB,
                "DB_ECO",
                1,
-               "ECO: dbBTerm:create");
+               "ECO: create dbBTerm '{}' on dbNet({}, {:p}) '{}'",
+               name,
+               net->getId(),
+               static_cast<void*>(net),
+               name);
     block->_journal->beginAction(dbJournal::CREATE_OBJECT);
     block->_journal->pushParam(dbBTermObj);
     block->_journal->pushParam(net->getId());
@@ -680,11 +684,13 @@ void _dbBTerm::connectModNet(_dbModNet* mod_net, _dbBlock* block)
                utl::ODB,
                "DB_ECO",
                1,
-               "ECO: connect dbBTerm {} at id {} to dbModNet {} at id {}",
-               _name,
-               getId(),
-               mod_net->_name,
-               mod_net->getId());
+               "ECO: connect dbBTerm({} {:p}) '{}' to dbModNet({} {:p}) '{}'",
+               bterm->getId(),
+               static_cast<void*>(bterm),
+               bterm->_name,
+               mod_net->getId(),
+               static_cast<void*>(mod_net),
+               ((dbModNet*) mod_net)->getHierarchicalName());
     block->_journal->beginAction(dbJournal::CONNECT_OBJECT);
     block->_journal->pushParam(dbBTermObj);
     block->_journal->pushParam(bterm->getId());
@@ -715,11 +721,13 @@ void _dbBTerm::connectNet(_dbNet* net, _dbBlock* block)
                utl::ODB,
                "DB_ECO",
                1,
-               "ECO: connect dbBTerm {} at id {} to dbNet {} at id {}",
-               _name,
-               getId(),
-               net->_name,
-               net->getId());
+               "ECO: connect dbBTerm({} {:p}) '{}' to dbNet({} {:p}) '{}'",
+               bterm->getId(),
+               static_cast<void*>(bterm),
+               bterm->_name,
+               net->getId(),
+               static_cast<void*>(net),
+               ((dbNet*) net)->getName());
     block->_journal->beginAction(dbJournal::CONNECT_OBJECT);
     block->_journal->pushParam(dbBTermObj);
     block->_journal->pushParam(bterm->getId());
@@ -781,7 +789,10 @@ void dbBTerm::destroy(dbBTerm* bterm_)
                utl::ODB,
                "DB_ECO",
                1,
-               "ECO: dbBTerm:destroy");
+               "ECO: delete dbBTerm({}, {:p}) '{}'",
+               bterm->getId(),
+               static_cast<void*>(bterm),
+               bterm->_name);
     block->_journal->beginAction(dbJournal::DELETE_OBJECT);
     block->_journal->pushParam(dbBTermObj);
     block->_journal->pushParam(bterm_->getId());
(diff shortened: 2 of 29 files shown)


src/dbSta/test/cpp/TestHconn.cpp
```diff
@@ -945,7 +945,6 @@ TEST_F(TestHconn, ConnectionMade)
   // Journalling test.
   // Undo everything and check initial state preserved
   //
-  odb::dbDatabase::endEco(block_);
```
povik, line 948: @erendn is this right? It looks left over from my initial patch

erendn: I think this is fine. `odb::dbDatabase::undoEco` will undo the active ECO so no need to end it first.

maliberty:
The ODB files are generated check indicates that you have modified some of the generated code that needs to be addressed by either modifying the schema, templates, or adding appropriate comments marking it as user code (I haven't investigated in detail).

erendn:
I think this is ready for a review.

src/odb/include/odb/db.h
```diff
@@ -7416,46 +7416,49 @@ class dbDatabase : public dbObject
...
+  static void commitEco(dbBlock* block);
 
   ///
-  /// Read the eco changes from the specified stream to be applied to the
-  /// specified block.
+  /// Undo the last ECO changes on the specified block.
+  /// Only supports:
+  ///   create and destroy of dbInst and dbNet
+  ///   dbInst::swapMaster
+  ///   connect and disconnect of dbBTerm and dbITerm
```
maliberty: This is an incomplete list and should be fixed or removed.

src/odb/include/odb/db.h
```diff
@@ -7416,46 +7416,49 @@ class dbDatabase : public dbObject
...
-  /// Write the eco netlist changes to the specified stream.
+  /// Returns true if the current ECO is empty
   ///
-  static void writeEco(dbBlock* block, const char* filename);
-  static int checkEco(dbBlock* block);
+  static bool ecoEmpty(dbBlock* block);
 
   ///
-  /// Commit any pending netlist changes.
+  /// Return true if the ECO history is empty.
```
maliberty: What is the "history" of an ECO?  This comment needs more explanation

src/odb/include/odb/db.h
```diff
@@ -7416,46 +7416,49 @@ class dbDatabase : public dbObject
...
   ///
-  /// Undo any pending netlist changes.  Only supports:
-  ///   create and destroy of dbInst and dbNet
-  ///   dbInst::swapMaster
-  ///   connect and disconnect of dbBTerm and dbITerm
+  /// Read the ECO changes from the specified stream to be applied to the
+  /// specified block.
   ///
-  static void undoEco(dbBlock* block);
+  static void readEco(dbBlock* block, const char* filename);
```
maliberty: It take a filename not a stream (same with writeEco)

src/odb/src/db/dbBTerm.cpp
```diff
@@ -680,10 +680,9 @@ void _dbBTerm::connectModNet(_dbModNet* mod_net, _dbBlock* block)
                utl::ODB,
                "DB_ECO",
                1,
-               "ECO: connect Bterm {} to modnet {}",
-               bterm->getId(),
-               mod_net->getId());
-
+               "ECO: connect dbBTerm {} at id {}",
+               _name,
+               getId());
```
maliberty: Why are you removing information from these messages (here and elsewhere)?

povik:
Looking at the CI failure it's strange that this has changed QoR on `MockArray_8x8_flat_test` and `MockArray_8x8_base_test`:

```
[ERROR] detailedroute__timing__hold__tns fail test: -494.643 >= -475
```

As far as I can tell master was passing those tests when last merged into this branch: https://github.com/The-OpenROAD-Project/OpenROAD/commit/01b4063aafec3f070c107baa3d07415d49753064

src/odb/src/db/dbJournalLog.cpp
```diff
@@ -237,6 +239,64 @@ void dbJournalLog::pop(std::string& value)
...
+void dbJournalLog::append(dbJournalLog& other)
+{
+  if (debug_ != other.debug_) {
+    logger_->error(utl::ODB, 429, "Journal debug mode mismatch.");
+  }
+  if (other.size() == 0) {
+    return;
+  }
+  // Scan from the end to find all action indices
+  std::vector<uint> action_indices;
```
povik, line 251: From this code I'm learning when we splice two journals we need to fix up the pointer that's at the end of each item:
```
void dbJournal::endAction()
{
  _start_action = false;
  _log.push((unsigned char) END_ACTION);
  _log.push(_action_idx);  // This value allows log to be scanned backwards.
}
```

povik:
I'll test this on ORFS to see if we get a result change there, too. Otherwise the code looks good to me.

povik:
ORFS results look fine. At the same time I can reproduce the failing mock array test locally.

povik:
The pr-merge job is green. @maliberty from my end I think we can merge this. I've tested it with public ORFS.

src/rsz/test/repair_setup_undo1.ok
```diff
@@ -25,6 +25,6 @@ post repair_timing QoR
 worst slack max -0.36
 tns max -2.054
 post undo repair_timing QoR
-worst slack max -0.36
-tns max -2.054
+worst slack max -1.95
+tns max -6.294
```
maliberty, line 29: Why did this result change?

src/odb/src/db/dbDatabase.cpp
```diff
@@ -707,47 +707,121 @@ void dbDatabase::write(std::ostream& file)
...
+  if (!block->_journal) {
+    block->_journal = block->_journal_history.back();
+    block->_journal_history.pop_back();
+  }
+  if (!block->_journal_history.empty()) {
+    dbJournal* next_journal = block->_journal_history.back();
+    block->_journal_history.pop_back();
+    int old_size = next_journal->size();
+    next_journal->append(block->_journal);
+    block->_journal_history.push_back(next_journal);
```
maliberty: Why is it necessary to pop and push next_journal?  It seems like a no-op.

src/odb/src/db/dbDatabase.cpp
```diff
@@ -707,47 +707,121 @@ void dbDatabase::write(std::ostream& file)
...
+             block->_journal_history.back()->size());
+}
 
-  {
-    delete block->_journal_pending;
-  }
+void dbDatabase::commitEco(dbBlock* block_)
+{
+  _dbBlock* block = (_dbBlock*) block_;
+  // Commit the current journal or the last journal in history
```
maliberty: Previously we called redo on _journal_pending.  What is the corresponding action when there is an empty _journal_history (ie no merge)?

src/odb/src/db/dbBlock.h
```diff
@@ -301,7 +301,7 @@ class _dbBlock : public _dbObject
   void* _extmi;
 
   dbJournal* _journal;
-  dbJournal* _journal_pending;
+  std::vector<dbJournal*> _journal_history;
```
maliberty: I find the word history to be a misnomer.  This is a stack of active, nested journals not a record of past journals.  std::stack might be a better data type as well to make that clear.

src/rsz/test/repair_setup_undo1.ok
povik: Previously this test was broken, i.e. wrapping repair_timing in beginEco/undoEco didn't work because that involved nested eco. Now the undo works properly and we get the pre-repair numbers from lines 8, 9

src/odb/include/odb/db.h
erendn: Removed.

src/odb/include/odb/db.h
erendn: Done.

src/odb/src/db/dbDatabase.cpp
erendn: Done.

src/odb/src/db/dbBlock.h
erendn: I changed it to use std::stack and renamed it to _journal_stack.

src/odb/include/odb/db.h
erendn: Updated.

src/odb/src/db/dbBTerm.cpp
erendn: "ECO" and "UNDO ECO" debug prints didn't match. They print the same information as dbJournal now. I can take a look and add the ones that were removed.

src/odb/src/db/dbDatabase.cpp
erendn: Redo was never actually used because commitEco was never called. If redo is supposed to be used to reverse an "undo", I can add another member variable to hold the last undone ECO. But I don't know if/where it will be used.

src/odb/src/db/dbBTerm.cpp
erendn: Done.

maliberty (approved):

src/odb/src/db/dbDatabase.cpp
maliberty: redo is a use model that we don't use in OR (odb was a contribution from a previous system).   The semantic changes but this seems ok.

maliberty:
Conflicts to resolve...

Issue #7872, opened 2025-07-23
pdn: Power ports removed after calling pdngen


titan73:
### Describe the bug

A def is loading with all the ports (signal and power) but pdgen removes some of then:

Before pdnden:
<img width="1139" height="842" alt="Image" src="https://github.com/user-attachments/assets/d236dc69-60c1-4e45-a7a7-48d907e40555" />

After pdnden:
<img width="1149" height="842" alt="Image" src="https://github.com/user-attachments/assets/2d30db18-ef44-4b0c-99b7-c3af4c1dad6e" />

### Expected Behavior

Power ports not removed

### Environment

```shell
OpenROAD v2.0-22931-gac11984103
```

### To Reproduce

Test case full_test_case_2025-07-23.zip uploaded to PII sftp server.

### Relevant log output

```shell

```

PR #8779, opened 2025-11-03
pdn: Remove non-fixed bpins only when adding pins is requested.

titan73:
Fixes: https://github.com/The-OpenROAD-Project/OpenROAD/issues/7872

If `pdngen` is called with `-dont_add_pins` flag, no pins are created so there is not reason to remove the existing ones even if they are not fixed.

Also, add a message when a bpin is removed to help understand what's done.

### DIFF
--- a/src/pdn/test/bpin_removal.ok
+++ b/src/pdn/test/bpin_removal.ok
@@ -5,4 +5,5 @@
 [INFO ODB-0131]     Created 482 components and 2074 component-terminals.
 [INFO ODB-0133]     Created 385 nets and 1110 connections.
 [INFO PDN-0001] Inserting grid: Core
+[INFO PDN-0242] Removing non-fixed bpins VDD
 No differences found.


gadfort:
What is the hope here? PDN does not magically connect to the BTerms you have, are you expecting them to be left floating?

titan73:
The analog team defines the floorplan shape and bterms with a def they provide to us. The grid is created with `-dont_add_pins` and the bterms are connected manually to the grid.

titan73:
The CI seems to have crash:
```
Failed to rm container '8aaa2f916d385100946fa1434db9d85859d6a2d851eea7612992450d64e17de4'.
```

maliberty:
The CI issue is known and being investigated

gadfort:
@titan73 couldn't you just mark the bpins as fixed? This change would likely just result in later errors in the flow when psm attempts to check connectivity and the pins have not been removed.

titan73:
The straps does not go to the boundary with fixed bpins:
bterms1
The result with placed bpins and my commit:
bterms2

titan73:
> This change would likely just result in later errors in the flow when psm attempts to check connectivity and the pins have not been removed.

Why not removing the pins would make psm fail? Is there a test for that?
Either we:
   - remove current pins and create new ones
   - keep the current pins without creating new ones with pdn and manage to get the grid connected to the current pins
What's the difference? In both case the pins are connected to the grid

gadfort:
psm will fail because the port were not connected to the grid.
if tell PDN to extend the strap to the boundary then ports would be inserted for you automatically (since it looks like you have perfect overlap it shouldn't really change anything for you).
we could look at connecting to fixed pins (non-fixed pins should be ignored in the floorplanning), but that would require time to implement since the current assumption in PDN is the pins will be created by PDN and PDN will ensure connectivity to those.

gadfort:
This is replaced by https://github.com/The-OpenROAD-Project/OpenROAD/pull/8851

PR #8787, opened 2025-11-04
ram: updated ram.tcl and example test file

braydenlouie:
- Created function in ram.tcl to handle pdn generation, global routing, pin placement, and filler scripts
- Changed make_8x8.tcl file to reflect changes in ram.tcl

### DIFF
--- a/src/ram/src/ram.tcl
+++ b/src/ram/src/ram.tcl
@@ -182,5 +182,5 @@ proc generate_ram { args } {
   filler_placement $filler_cells
 
   global_route
-  detailed_route
+  detailed_route -verbose 0
 }
--- a/src/ram/test/make_8x8.tcl
+++ b/src/ram/test/make_8x8.tcl
@@ -7,8 +7,6 @@ read_liberty sky130hd/sky130_fd_sc_hd__tt_025C_1v80.lib
 read_lef sky130hd/sky130hd.tlef
 read_lef sky130hd/sky130_fd_sc_hd_merged.lef
 
-puts $tcl_version
-
 generate_ram \
   -bytes_per_word 1 \
   -word_count 8 \


src/ram/src/ram.tcl
```diff
@@ -48,3 +48,124 @@ proc generate_ram_netlist { args } {
...
+  global_route
+  detailed_route
+
+  set lef_file [make_result_file make_8x8.lef]
+  write_abstract_lef $lef_file
+  diff_files make_8x8.lefok $lef_file
+
+  set def_file [make_result_file make_8x8.def]
+  write_def $def_file
+  diff_files make_8x8.defok $def_file
```
rovinski: This part should not be in ram.tcl, it should be in the test

src/ram/src/ram.tcl
```diff
@@ -48,3 +48,124 @@ proc generate_ram_netlist { args } {
...
+				     [-storage_cell name]
+				     [-tristate_cell name]
+				     [-inv_cell name]
+				     [-read_ports count]
+				     -power_net name
+				     -ground_net name
+				     -routing_layer config
+				     -ver_layer config
+				     -hor_layer config
+				     -filler_cells fillers}
```
rovinski: Does this pass the formatting check? Seems odd.

src/ram/src/ram.tcl
```diff
@@ -48,3 +48,124 @@ proc generate_ram_netlist { args } {
...
+  }
+
+  if { [info exists keys(-filler_cells)] } {
+    set filler_cells $keys(-filler_cells)
+  } else {
+    utl::error RAM 18 "The -filler_cells argument must be specified."
+  }
+
+  add_global_connection -net VDD -pin_pattern $power_net -power
+  add_global_connection -net VSS -pin_pattern $ground_net -ground
```
rovinski, line 157: I think you may want to change `power_net` and `ground_net` to `power_pin` and `ground_pin`. Because that is technically what it is setting (power net is statically set to VDD and ground net is statically set to VSS here).

rovinski (changes requested):

src/ram/src/ram.tcl
```diff
@@ -48,3 +48,124 @@ proc generate_ram_netlist { args } {
...
+proc generate_ram { args } {
+  sta::parse_key_args "generate_ram" args \
+    keys {-bytes_per_word -word_count -storage_cell -tristate_cell -inv_cell -read_ports
+      -power_net -ground_net -routing_layer -ver_layer -hor_layer -filler_cells} flags {}
+
+  generate_ram_netlist \
+    -bytes_per_word $keys(-bytes_per_word) \
+    -word_count $keys(-word_count) \
+    -storage_cell $keys(-storage_cell) \
+    -read_ports $keys(-read_ports)
```
gadfort: shoudn't tristatecell be forwarded as well?

src/ram/src/ram.tcl
```diff
@@ -48,3 +48,124 @@ proc generate_ram_netlist { args } {
...
+				     [-read_ports count]
+				     -power_net name
+				     -ground_net name
+				     -routing_layer config
+				     -ver_layer config
+				     -hor_layer config
+				     -filler_cells fillers}
+
+# user arguments for generate ram netlist arguments
+proc generate_ram { args } {
```
gadfort, line 66: this command might need some safeguards to check if a design is loaded and error out

rovinski:
@braydenlouie can you try merging in the master branch to see if it fixes the CI failures?

src/ram/src/ram.tcl
rovinski: @braydenlouie I think you can add something like

```
set chip [$db getChip]
if { $chip != "NULL" } {
    utl::error RAM ## "A design is already loaded. Cannot generate RAM."
}
```

rovinski:
@braydenlouie you can mark as ready for review

src/ram/src/ram.tcl
braydenlouie: Should the safeguard be added to the `generate_ram_netlist` too or just to the `generate_ram`?

rovinski: Considering `generate_ram_netlist` is the first thing called, you shouldn't need to duplicate any checks that are run inside of that proc.

maliberty (approved):

maliberty:
In a future PR it would be nice to see if we can get more symmetry across rows, eg the routing for *select_b
image

rovinski:
> In a future PR it would be nice to see if we can get more symmetry across rows, eg the routing for *select_b

Yep, @braydenlouie and I talked about that. We're not sure how to encourage the detailed router to do that. We thought about manually creating stright wires, but the problem is that it would be very difficult to handle different cell libraries / technologies as the routing rules and pin locations will change.

@braydenlouie is working on tapcell insertion next, as that's essentially the last step to create a tapeout-ready block. After that is more rigorous testing for all bit/word sizes from 1x1 to 8x8, and then probably looking at higher word sizes (e.g. 16, 32, 64).

Issue #8782, opened 2025-11-03
[ERROR GRT-0076] Net net1217 does not have route guides.


gudeh:
### Description

GRT is failing with sky130hd/riscv
artifact: https://drive.google.com/file/d/1PAr9GR7IpCJ8ZvYnZKA2YKnHNl40NxMg/view?usp=sharing
also available at: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-reduce-inflation/4/pipeline/247

gudeh:
Also happens with sky130hs/riscv (hs instead of hd) in this run: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-reduce-inflation/2/pipeline/247

eder-matheus:
The net with this issue has an instance out of the die area:
<img width="1305" height="1519" alt="Image" src="https://github.com/user-attachments/assets/20ec365a-b3f6-4ef0-8b72-cfb401c0ffb5" />

I think what's happenining is that during the repair_timing iterations, RSZ will insert buffers and run GRT to get parasitics data without legalizing the cells (due to runtime concerns). Only at the end of the repair_timing iteratios we call DPL to legalize the cells and generate a final and valid placement.

I'll try to update this process to detect such cases and legalize to avoid GRT errors.

@maliberty FYI.

maliberty:
Why would rsz ever put a buffer outside the core area?  Normally they fall along the Steiner tree.  @povik @precisionmoon

povik:
Yes, Steiner tree or in the case of "split load" on top of another gate. @eder-matheus do you know the name of the out of bounds instance?

eder-matheus:
> Yes, Steiner tree or in the case of "split load" on top of another gate. [@eder-matheus](https://github.com/eder-matheus) do you know the name of the out of bounds instance?

Yes, the instance is `rebuffer1194`, master `sky130_fd_sc_hd__buf_4`. 

EDIT: I found that the instance changes on repairPath with `SplitLoadMove`. I'm looking at this code.

gudeh:
Hey guys, I changed some parameters in gpl and got a seg fault this time with sky130hs/riscv.

[...]

PR #8793, opened 2025-11-04
odb: use setOrigin on dbJournal::undo_deleteObject for dbInstObj

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/8782

When restoring a deleted instance, dbJournal was using `setLocation`. This function is not the correct for it, as it modifies the input (x, y) based on the master bbox.

Changed it to setOrigin, to ensure that the input (x, y) will be preserved.


maliberty:
Please add a unit test in TestJournal.cpp

PR #8795, opened 2025-11-05
pad: expand place instance checks and add ability to shift cells during automatic placement

gadfort:
Changes:
- checks now check for blockages
- when nets overlap they are allowed
- when using `place_pads` the pads will shift to avoid obstructions.

Tests:
- add tests for blockage avoidance, blockage error, and allowing net overlaps

### DIFF
--- a/src/pad/src/ICeWall.cpp
+++ b/src/pad/src/ICeWall.cpp
@@ -983,8 +983,8 @@ void ICeWall::placePadsBumpAligned(
                               snapToRowSite(row, select_pos),
                               ginst,
                               odb::dbOrientType::R0,
-                              false,
-                              true)
+                              /* allow_overlap */ false,
+                              /* allow_shift */ true)
                 * row->getSpacing();
 
       performPadFlip(row, ginst, iterm_connections);
@@ -1018,8 +1018,8 @@ void ICeWall::placePadsUniform(const std::vector<odb::dbInst*>& insts,
                   snapToRowSite(row, offset),
                   inst,
                   odb::dbOrientType::R0,
-                  false,
-                  true);
+                  /* allow_overlap */ false,
+                  /* allow_shift */ true);
     offset += inst_widths.at(inst);
     offset += target_spacing;
   }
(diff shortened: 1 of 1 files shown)


src/pad/src/ICeWall.cpp
```diff
@@ -1034,7 +982,9 @@ void ICeWall::placePadsBumpAligned(
       offset += placeInstance(row,
                               snapToRowSite(row, select_pos),
                               ginst,
-                              odb::dbOrientType::R0)
+                              odb::dbOrientType::R0,
+                              false,
+                              true)
```
maliberty:
```suggestion
                              /* allow_overlap */ false,
                              /* allow_shift */ true)
```

src/pad/src/ICeWall.cpp
```diff
@@ -1064,7 +1014,12 @@ void ICeWall::placePadsUniform(const std::vector<odb::dbInst*>& insts,
                inst->getName(),
                offset / dbus);
 
-    placeInstance(row, snapToRowSite(row, offset), inst, odb::dbOrientType::R0);
+    placeInstance(row,
+                  snapToRowSite(row, offset),
+                  inst,
+                  odb::dbOrientType::R0,
+                  false,
+                  true);
```
maliberty: same as above

src/pad/src/ICeWall.cpp
```diff
@@ -1202,38 +1158,69 @@ int ICeWall::placeInstance(odb::dbRow* row,
...
+    const auto& [check_inst, check_rect] = *check_obs;
+
+    int obs_index = index;
+    switch (row_edge) {
+      case odb::Direction2D::North:
+        obs_index = snapToRowSite(row, check_rect.xMin());
+        break;
+      case odb::Direction2D::South:
+        obs_index = snapToRowSite(row, check_rect.xMin());
+        break;
```
maliberty, line 1170:
```suggestion
      case odb::Direction2D::North:
      case odb::Direction2D::South:
        obs_index = snapToRowSite(row, check_rect.xMin());
        break;
```

src/pad/src/ICeWall.cpp
```diff
@@ -1202,38 +1158,69 @@ int ICeWall::placeInstance(odb::dbRow* row,
...
+      case odb::Direction2D::West:
+        obs_index = snapToRowSite(row, check_rect.yMin());
+        break;
+      case odb::Direction2D::East:
+        obs_index = snapToRowSite(row, check_rect.yMin());
+        break;
```
maliberty, line 1174:
```suggestion
      case odb::Direction2D::West:
      case odb::Direction2D::East:
        obs_index = snapToRowSite(row, check_rect.yMin());
        break;
```

Issue #8790, opened 2025-11-04
PDR crash on GRT call


eder-matheus:
PDR is crashing with the test case below, but the reason is probably related to GRT not passing the correct inputs.

> This is the run in jenkins: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-less-inflation/4/pipeline/247
> 
> And this is the artifact: https://drive.google.com/file/d/1ZULw73Xmhae5lTQ8IapdeX21KFzPXW6P/view?usp=sharing
> 
> ```
> [2025-11-03T21:11:57.548Z]       610 |      36 |     113 |       22 |      6 |    37 |    +0.7% |   -0.796 |     -149.8 |    730 | aluout[0]
> [2025-11-03T21:11:57.548Z]       620 |      38 |     117 |       23 |      8 |    37 |    +0.7% |   -0.797 |     -149.4 |    730 | aluout[0]
> [2025-11-03T21:11:57.548Z]       624 |      36 |     113 |       22 |      6 |    37 |    +0.7% |   -0.796 |     -150.1 |    729 | aluout[0]
> [2025-11-03T21:11:58.174Z]       630 |      36 |     116 |       22 |      8 |    37 |    +0.7% |   -0.796 |     -152.3 |    729 | aluout[0]
> [2025-11-03T21:11:58.174Z] /usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = int; _Alloc = std::allocator<int>; std::vector<_Tp, _Alloc>::reference = int&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
> [2025-11-03T21:11:58.174Z] Signal 6 received
> [2025-11-03T21:11:58.174Z] Stack trace:
> [2025-11-03T21:11:58.174Z]  0# 0x00005E2675DCD0E2 in /tmp/workspace/rivate_secure-gpl-less-inflation/tools/install/OpenROAD/bin/openroad
> [2025-11-03T21:11:58.174Z]  1# 0x0000710F7EBBE520 in /lib/x86_64-linux-gnu/libc.so.6
[...]

PR #8796, opened 2025-11-05
grt: check if instance is inside die area

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/8790

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -3890,11 +3890,10 @@ void GlobalRouter::makeItermPins(Net* net,
     }
 
     if (!die_area.contains(inst->getBBox()->getBox())) {
-      logger_->error(
-          GRT,
-          280,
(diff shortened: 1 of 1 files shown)


src/grt/src/GlobalRouter.cpp
```diff
@@ -3888,6 +3888,15 @@ void GlobalRouter::makeItermPins(Net* net,
     if (!inst->isPlaced()) {
       logger_->error(GRT, 10, "Instance {} is not placed.", inst->getName());
     }
+
+    if (!die_area.contains(inst->getBBox()->getBox())) {
```
maliberty, line 3892: @gadfort can you see any cases in pad rings where this might be violated?

gadfort: if you mean IO pad rings, then no, they should always be inside the die area (but outside the core area), so this check should be valid.

PR #8800, opened 2025-11-06
gpl: reduce inflation parameters

gudeh:
A design with a tighter core area has shown problems due to too much inflation during gpl routability mode. We would solve congestion on one side by inflating instances, although we generate congestion on the other side of the design, leading to DPL and GRT issues.

For this reason we are reducing the default inflation parameters.


gudeh:
By the way @jfgava made some experiments with this reduction in inflation and mentioned better QoR overall. If I recall correctly, easier to route for GRT.

PR #8804, opened 2025-11-06
ifp: remove old tracks when floorplan is changed

gadfort:
Previous:
- when `initialize_floorplan` is called after tracks are present, they were not removed causing multiple track patterns to appear in the design

Now:
- when `initialize_floorplan` is called the tracks are removed to avoid invalid track patterns from getting left behind.

### DIFF
--- a/src/rsz/test/BUILD
+++ b/src/rsz/test/BUILD
@@ -39,6 +39,7 @@ TESTS = [
     "rebuffer1",
     "rebuffer1_hier",
     "recover_power_verbose",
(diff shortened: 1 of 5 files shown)


maliberty (approved):

maliberty:
idk if the failure of pdn.core_grid_with_single_edge_pins.tcl (Failed) is related or not in pr-head but it failed in both cmake & bazel

gadfort:
@maliberty test should have been fixed now

PR #8810, opened 2025-11-07
ord: forward debug level to STA when STA is requested

### LINKED ISSUE #8749: set_debug_level does not pass along debug level to STA
### Describe the bug

set_debug_level does not set the debug level for STA in openroad.  Instead the user needs to remember to use sta::set_debug.  

### Expected Behavior

User can use set_debug_level to control debug level for all of openroad.  Logging between openroad and opensta should be integrated. 

### Environment

```shell
N/A
```

### To Reproduce

```
$ openroad

set_debug_level "STA" "liberty" 100
read_liberty <LIBERTY_FILE>
# Observe no debug messages

sta::set_debug "liberty" 100
read_liberty <LIBERTY_FILE>
# Observe debug messages
```

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

gadfort:
Closes #8749

@maliberty I couldn't find a place where this is handed in the python swig wrapper, let me know if it's there and I missed it.

### DIFF
--- a/src/OpenRoad.i
+++ b/src/OpenRoad.i
@@ -417,6 +417,9 @@ set_debug_level(const char* tool_name,
   if (id == utl::UKN) {
     logger->error(utl::ORD, 15, "Unknown tool name {}", tool_name);
   }
+  if (id == utl::STA) {
+    getSta()->setDebugLevel(group, level);
+  }
   logger->setDebugLevel(id, group, level);
 }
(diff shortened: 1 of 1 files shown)


maliberty:
I think to make this work in python we would need to be able to add a hook into the Logger when we create it.

PR #8811, opened 2025-11-07
gpl: rename variables for clarity

gudeh:
These changes are no-op.

renamings are related to the scalings gpl uses:
  -target density
  -cell smoothing
  -binormal distribution for macros

This PR removes the "unscaled" suffix from variables, the unscaled variables versions now have no suffix.  The scaling versions relate to the type os scaling performed, for example, scaling performed from binormal distribution will have the suffix "binormal".

### DIFF
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -255,7 +255,7 @@ bool GCell::isLargeInstance() const
   if (!isInstance()) {
     return false;
   }
-  return insts_[0]->isPbLargeInstance();
+  return insts_[0]->isLargeInstance();
 }
 
 bool GCell::isStdInstance() const
@@ -263,7 +263,7 @@ bool GCell::isStdInstance() const
   if (!isInstance()) {
     return false;
   }
-  return !insts_[0]->isPbLargeInstance();
+  return !insts_[0]->isLargeInstance();
 }
 
 void GCell::print(utl::Logger* logger, bool print_only_name = true) const
@@ -3987,7 +3987,7 @@ static int64_t getOverlapAreaBiNormal(const Bin* bin,
     return 0;
   }
 
-  if (inst->isPbLargeInstance()) {
+  if (inst->isLargeInstance()) {
     const float meanX = (inst->cx() - inst->lx()) / (float) dbu_per_micron;
     const float meanY = (inst->cy() - inst->ly()) / (float) dbu_per_micron;
 
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -938,7 +938,7 @@ void BinGrid::updateBinsGCellDensityArea(const std::vector<GCellHandle>& cells)
     const float overflowAreaUnscaled
         = std::max(0.0f,
                    static_cast<float>(bin.getInstPlacedArea())
-                       + static_cast<float>(bin.getNonPlaceAre())
+                       + static_cast<float>(bin.getNonPlaceArea())
                        - density_scaled_bin_area);
     sum_overflow_area_ += overflowAreaUnscaled;
 
@@ -952,7 +952,7 @@ void BinGrid::updateBinsGCellDensityArea(const std::vector<GCellHandle>& cells)
                    block->dbuAreaToMicrons(density_scaled_bin_area));
       log_->report("bin.getInstPlacedArea():{}, bin.getNonPlaceArea():{}",
                    block->dbuAreaToMicrons(bin.getInstPlacedArea()),
-                   block->dbuAreaToMicrons(bin.getNonPlaceAre()));
+                   block->dbuAreaToMicrons(bin.getNonPlaceArea()));
       log_->report(
           "bin.getInstPlacedAreaBiNormal():{}, "
           "bin.getNonPlaceAreaBiNormal():{}",
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -35,11 +35,11 @@ using utl::GPL;
 
 static float calculateBiVariateNormalCDF(biNormalParameters i);
 
-static int64_t getOverlapArea(const Bin* bin,
-                              const Instance* inst,
-                              int dbu_per_micron);
+static int64_t getOverlapAreaBiNormal(const Bin* bin,
+                                      const Instance* inst,
+                                      int dbu_per_micron);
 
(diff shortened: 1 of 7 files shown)


gudeh:
Me, @eder-matheus, @AcKoucher and @LucasYuki had a discussion about this, and how this part of the code was not clear. I hope with this changes things are more clear now.

When we talked about this, we got confused among the different types of scalings performed.

src/gpl/src/placerBase.h
```diff
@@ -109,7 +109,7 @@ class Instance
   int ux_ = 0;
   int uy_ = 0;
   int extId_ = INT_MIN;
-  bool is_macro_ = false;
+  bool is_large_instance_ = false;
```
eder-matheus, line 113: Could you add a comment describing what is a "large instance" in the context of GPL?

src/gpl/src/nesterovBase.h
```diff
@@ -612,54 +615,54 @@ inline int Bin::dy() const
...
 {
-  nonPlaceArea_ += area;
+  non_place_area_binormal_ += area;
 }
 
-inline void Bin::addInstPlacedArea(int64_t area)
-{
-  instPlacedArea_ += area;
-}
+// inline void Bin::addInstPlacedArea(int64_t area)
```
eder-matheus: Is this suppose to still commented? Can it be removed?

eder-matheus (changes requested):

src/gpl/src/placerBase.h
gudeh: Yes! Historically RePlace is able to move and place macros. It considers macros as any instance which has more than 6 rows height. This concept conflicts with how we treat macros in gpl itself, where we check if an instance is fixed (which is the case for macros) as is done [in this part of the code](https://github.com/The-OpenROAD-Project/OpenROAD/blob/d69f08d01e675b65eb5f1300f5fb0969e11399a0/src/gpl/src/placerBase.cpp#L1044-L1053) . 

To make it more clear I renamed this to "large instances", and in practice in OpenROAD flow we never have any large instances instantiated. Either way I think we should keep this concept so we may explore gpl's capability of placing macros in the future, if we wish to.

I just noticed I should also rename macroInstsArea_ to `large_insts_area_`. I will apply the changes.

src/gpl/src/nesterovBase.h
gudeh: This function is never used. I believe we can remove.

gudeh:
I ran a secure-CI with public designs only for sanity check and everything is green, as expected: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/test-gpl-remove-unscaled-suffix/1/pipeline

src/gpl/src/nesterovBase.h
eder-matheus: If that's the case, please remove. Let's avoid dead code

src/gpl/src/nesterovBase.h
```diff
@@ -545,27 +545,30 @@ class Bin
...
-  int64_t getNonPlaceArea() const { return nonPlaceArea_; }
-  int64_t instPlacedArea() const { return instPlacedArea_; }
-  int64_t getNonPlaceAreaUnscaled() const { return nonPlaceAreaUnscaled_; }
-  int64_t getInstPlacedAreaUnscaled() const { return instPlacedAreaUnscaled_; }
+  int64_t getNonPlaceAreaBiNormal() const { return non_place_area_binormal_; }
+  int64_t getInstPlacedAreaBiNormal() const
+  {
+    return inst_placed_area_binormal_;
+  }
+  int64_t getNonPlaceAre() const { return non_place_area_; }
```
AcKoucher: Typo `getNonPlaceArea()`

src/gpl/src/nesterovBase.cpp
```diff
@@ -218,9 +218,9 @@ void GCell::setDensitySize(int dDx, int dDy)
   dUy_ = dCenterY + dDy / 2;
 }
 
-void GCell::setDensityScale(float densityScale)
+void GCell::setDensitySmoothing(float densitySmoothing)
```
AcKoucher, line 221: Is this in fact a scale? If yes perhaps `setDensitySmoothingScale` or `setSmoothingScale` is clearer.

src/gpl/src/nesterovBase.cpp
```diff
@@ -2191,7 +2186,9 @@ float NesterovBase::getTargetDensity() const
   return targetDensity_;
 }
 
-// update densitySize and densityScale in each gCell
+// update densitySize and densitySmoothing in each gCell
+// Density smoothing from eplace paper: "4.5. Local Smoothness Over Discrete
```
AcKoucher: Typo in the comment.

src/gpl/src/placerBase.cpp
```diff
@@ -246,9 +246,10 @@ void Instance::setExtId(int extId)
   extId_ = extId;
 }
 
-bool Instance::isMacro() const
+// ePlace macro definition
+bool Instance::isPbLargeInstance() const
 {
-  return is_macro_;
+  return is_large_instance_;
```
AcKoucher: Why not just `isLargeInstance()`?

AcKoucher (changes requested):

AcKoucher:
W.r.t. to names such as `getInstPlacedAreaBiNormal()`, the `InstPlaced` still feel vague to me. Would it refer to movable cells by GPL i.e., instances that are being placed? If that's the case switching the order perhaps would make it a bit clearer? `getPlacedInstsArea()`?

Just a nit though.

src/gpl/src/nesterovBase.cpp
gudeh: This smoothing scales the instance to be at least the same size of a bin. I understand the term smoothing here already implies a scaling.

gudeh:
I was checking the interaction between rsz reisizings and routability inflations and realized some things. The change in size already considers the smoothing reapplying (which is good). Although there are some things to rename:

TODO: 
- `updateDensityCoordiLayoutInside()` is the function that applies the actual smoothing, rename it to `GCellSmoothing()` or similar.
- Rename `GCell::setDensityLocation()`, `GCell::setDensityScale()`, `GCell::setDensitySize()`,  and `GCell::setDensityCenterLocation()` to have the keyword smoothing.
- Also: `GPin::updateDensityLocation()`, potentially others.

gudeh:
@LucasYuki we talked about the naming of functions. Unfortunately this went stale and now we have conflicts, but I believe we should do this renamings

PR #8822, opened 2025-11-11
grt: add option to skip large fanout nets

eder-matheus:


src/grt/src/GlobalRouter.cpp
```diff
@@ -3744,6 +3744,9 @@ std::vector<Net*> GlobalRouter::findNets(bool init_clock_nets)
   }
   std::vector<Net*> clk_nets;
   for (odb::dbNet* db_net : db_nets) {
+    if (skip_large_fanout_ > 0 && db_net->getTermCount() > skip_large_fanout_) {
+      continue;
+    }
```
eder-matheus: @maliberty should we generate a message (WARN or INFO) about the skipped nets?

src/grt/README.md
```diff
@@ -42,6 +43,7 @@ global_route
 | `-congestion_report_iter_step` | Set the number of iterations to report. The default value is `0`, and the allowed values are integers `[0, MAX_INT]`. |
 | `-grid_origin` | Set the (x, y) origin of the routing grid in DBU. For example, `-grid_origin {1 1}` corresponds to the die (0, 0) + 1 DBU in each x--, y- direction. |
 | `-critical_nets_percentage` | Set the percentage of nets with the worst slack value that are considered timing critical, having preference over other nets during congestion iterations (e.g. `-critical_nets_percentage 30`). The default value is `0`, and the allowed values are integers `[0, MAX_INT]`. |
+| `-skip_large_fanout_nets` | Skips routing for nets with a fanout higher than the specified limit. Nets exceeding this pin count threshold will be ignored by the global router. The default value is `-1`, and the allowed values are integers `[0, MAX_INT]`. |
```
oharboe: A word about use-case would be nice. Debugging? What happens when they are ignored? Left to detailed router to sort out somehow or is the idea to stop the flow at grt to get some sort of estimate?

Nit: contradiction... default -1 and allowed [0, MAX_INT]...

eder-matheus: I've updated the readme with the requested changes. Let me know if something is unclear.

src/grt/src/GlobalRouter.cpp
maliberty: Info seems fine since the user requested it.  I think a warning about a large non-skipped net would be good too (say >1k fanout).

src/grt/src/GlobalRouter.cpp
```diff
@@ -3744,6 +3744,9 @@ std::vector<Net*> GlobalRouter::findNets(bool init_clock_nets)
   }
   std::vector<Net*> clk_nets;
   for (odb::dbNet* db_net : db_nets) {
+    if (skip_large_fanout_ > 0 && db_net->getTermCount() > skip_large_fanout_) {
```
maliberty: It feels like it would be simpler to default to max_int and then you wouldn't need that >0 check.

maliberty:
This arose in the context of someone skipping cts but still wanting grt results without intending to do to drt.  grt bogs down on the clock net which isn't interesting as they are using ideal clocks.

PR #8832, opened 2025-11-12
gpl: new incremental messages

gudeh:
Include a few messages for gpl incremental mode, making it more clear to what it is doing.

An important addition is reporting the amount of instances placed and unplaced, which currently has a high difference in how gpl will behave in incremental mode.


gudeh:
here is a gif showing how our existing regression test `incremental02` behaves. Instances painted in teal are the locked ones.
output

We do a fixed 300 iterations unlocking only the unplaced instances and using 0.2 target overflow. Then we unlock all instances and execute nesterov again with target overflow of 0.1.

gudeh:
FYI @jferreiraOpenRoad

eder-matheus (approved):

PR #8836, opened 2025-11-13
pad: add placer mode to pad placement

gadfort:
Adds:
- new mode `placer` to `place_pads` which implements a general purpose pad placement algorithm.
- test for placement with placer

Changes:
- updates single test to add placer on one edge.

Notes:
- algorithm preserves the order in which pads a requested by the user.
- avoids obstructions in the rows when possible.


src/pad/README.md
```diff
@@ -245,6 +245,12 @@ In `linear` mode, the pads will be place starting from the bottom or left of the
 
 <img src="./doc/image/mode_linear.png" width=450px>
 
+In `placer` mode, the pads will be placed according to the following algorithm:
```
maliberty, line 248: Image?  The others have an example

gadfort: added

PR #8845, opened 2025-11-14
odb: add read_3dblox_bmap command

gadfort:
Adds:
- `read_3dblox_bmap` to read in and place the bmap into the current design.

Notes:
- I made the bpin creation optional, but I'm not sure it should be since it would be needed for LVS later.
- I'm open to name change suggestions to the command if needed.

### DIFF
--- a/src/gui/src/mainWindow.cpp
+++ b/src/gui/src/mainWindow.cpp
@@ -1870,7 +1870,17 @@ void MainWindow::showGlobalConnect()
 
 void MainWindow::openDesign()
 {
-  const QString filefilter = "OpenDB (*.odb *.ODB)";
+  const std::vector<QString> exts{".odb", ".odb.gz", ".db", ".db.gz"};
+
+  QString filefilter = "OpenDB (";
+  for (const auto& ext : exts) {
+    if (filefilter.length() != 8) {
+      filefilter += " ";
+    }
+    filefilter += "*" + ext;
+    filefilter += " *" + ext.toUpper();
+  }
+  filefilter += ")";
   const QString file = QFileDialog::getOpenFileName(
       this, "Open Design", QString(), filefilter);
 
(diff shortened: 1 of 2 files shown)


src/odb/src/3dblox/3dblox.cpp
```diff
@@ -444,4 +446,141 @@ void ThreeDBlox::createConnection(const Connection& connection)
...
+        for (dbMPin* mpin : mterm->getMPins()) {
+          for (dbBox* geom : mpin->getGeometry()) {
+            auto* layer = geom->getTechLayer();
+            if (layer == nullptr) {
+              continue;
+            }
+            if (max_layer == nullptr
+                || max_layer->getRoutingLevel() <= layer->getRoutingLevel()) {
+              max_layer = layer;
+              if (max_layer->getRoutingLevel() < layer->getRoutingLevel()) {
```
maliberty: how can the `if` be true if the two layers are the same?

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -444,4 +446,141 @@ void ThreeDBlox::createConnection(const Connection& connection)
...
+    }
+
+    // create bpins
+    for (dbInst* inst : bumps) {
+      auto masterbpin = bpininfo.find(inst->getMaster());
+      if (masterbpin == bpininfo.end()) {
+        continue;
+      }
+
+      BPinInfo& pin_info = masterbpin->second;
```
maliberty: const

maliberty:
I don't know a use case for leaving off the bpins

src/odb/src/3dblox/3dblox.cpp
gadfort: oops, assignment was in the wrong place

gadfort:
> I don't know a use case for leaving off the bpins

I'll remove the option then

maliberty (approved):

PR #8846, opened 2025-11-14
gui: add .gz and .db to allowed extension for open db

gadfort:

Changes:
- adds `.db` to allowed extension in open/save dialogs in gui. The command to write is `write_db` so it's natural to just use the `.db` file extension.
- adds `.gz` to allowed extensions as well.

### DIFF
--- a/src/gui/src/mainWindow.cpp
+++ b/src/gui/src/mainWindow.cpp
@@ -1914,7 +1914,7 @@ void MainWindow::openDesign()
 
 void MainWindow::saveDesign()
 {
-  const std::vector<QString> exts{".odb", ".odb.gz", ".db", ".db.gz"};
+  const std::vector<QString> exts{".odb", ".odb.gz"};
 
   QString filefilter = "OpenDB (";
   for (const auto& ext : exts) {


stefanottili:
.db has been used as extension for physical verification error files for decades.
And p&r tools load them to visualize errors, so don’t use .db, it’s already taken.

You might rather change read/write_db to _odb.

rovinski:
Do we really want to allow .db to avoid aliasing with a certain other tool?

rovinski:
> .db has been used as extension for physical verification error files for decades. And p&r tools load them to visualize errors, so don’t use .db, it’s already taken.
> 
> You might rather change read/write_db to _odb.

Same thought :)

gadfort:
@stefanottili changing the command name doesn't many much sense since that would impact all kinds of flows and internally the functions are called readDb and writeDb.
This only matters in the GUI, being overly restrictive with the readDB is annoying, I could understand attempting to limit the extensions allowed in saveDb. The GUI is actually the only place the extension is enforced.

maliberty:
There is no requirement on the extension for read/write db.  .db or .odb are convention at most.   ORFS uses .odb.   The file starts with magic bytes so read_db will not be confused by files from another tool.  I agree write_db is the place to be restrictive if anywhere.

maliberty:
.db(.gz) is removed from the write dialog but still allowed in open.

PR #8854, opened 2025-11-17
odb: cache x/y grids in `dbTrackGrid`

gadfort:

Changes:
- caches the results of the x/y grids in `dbTrackGrid`, these rarely change and take a noticeable amount of time to compute on larger designs (from this weekends testcase, about 45 seconds was spent repeatedly calculating this value in PDN (of course that is an inefficiency there, but this solves it for all)).  When the grid is modified the cache is cleared.
- make gui/pad/pdn use returned references to the grid vector.

### DIFF
--- a/src/odb/src/db/dbTrackGrid.cpp
+++ b/src/odb/src/db/dbTrackGrid.cpp
@@ -19,6 +19,7 @@
 #include "odb/db.h"
 #include "odb/dbSet.h"
 #include "odb/dbTypes.h"
+#include "odb/odb.h"
 #include "utl/Logger.h"
 
 namespace odb {
--- a/src/odb/src/db/dbTrackGrid.h
+++ b/src/odb/src/db/dbTrackGrid.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <vector>
+
 #include "dbCore.h"
 #include "dbDatabase.h"
 #include "dbVector.h"


maliberty:
Its not that much memory but you'll still pay the inefficiency once.  If it is faster there may be no need to cache.  There is no reserve call, no need to sort for a single pattern, even multi-pattern could be more efficient than an append and sort.

gadfort:
@maliberty I'm not sure what you mean by faster? The memory is trivial, and the improvements you are suggesting can be implemented, in a different PR.

maliberty:
By faster I mean efficient enough that no cache is needed.  Caching tends to be a source of bugs so I avoid it if I can (for instance this PR missed create / destroy invalidation).

gadfort:
On create, there is nothing to do (the grid vector would be empty, thus load loaded).
On adding pattern, the grid is cleared to ensure it would be created again if requested.
On destroy the object is destroyed along with its members.
I don't think there is any risk here of a invalid cache (if you see something that is missing specifically, then I can address that).

Allocating the memory over and over again is also not particularly efficient. In the profile, insertion still takes real time (at scale).

src/pdn/src/techlayer.cpp
```diff
@@ -47,11 +47,11 @@ void TechLayer::populateGrid(odb::dbBlock* block, odb::dbTechLayerDir dir)
 
   auto* tracks = block->findTrackGrid(layer_);
   if (dir == odb::dbTechLayerDir::HORIZONTAL) {
-    tracks->getGridY(grid_);
+    grid_ = tracks->getGridY();
```
maliberty, line 50: You are making a copy here fwiw

maliberty:
I think this is unnecessary but not worth arguing over.

PR #8855, opened 2025-11-17
mpl: remove unneeded checks and fix condition to store best result

AcKoucher:
We don't need to check if the best result is empty now that we always update it before beginning the annealer steps.
We also should be comparing the cost of the current step with the cost of the best result when evaluating if we actually found a new best result.

Here are the best result cost curves with/without the changes here:
| Base  | Branch |
| ---- |  ------- |
| image | image |

FWIW we should probably stop earlier as we're spending a huge amount of steps after the temperature drops.


AcKoucher:
Running Secure-CI.

maliberty:
The code changes are fine, just waiting on your CI results

AcKoucher:
@maliberty This should be paired with https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3659.

maliberty (approved):

PR #8881, opened 2025-11-20
Cts clone gaters

arthurjolo:
This PR implements 2 new features for CTS: moving and cloning clock gaters.

### **Moving clock gaters**

Some clock gaters can be placed to far from the sinks it connects, creating a long wire to connect the clock gater to the center of its tree increasing the internal delay of the clock gater
eg.: sky130hs/ibex
The highlighted net is the connection between the clock gater and first buffer of its tree, and in the tree viewer we can see the large internal delay of this clock gater.
image

To fix this issue we enable CTS to move the clock gaters to the center of its output net bounding box, reducing the connection between the clock gater and the first buffer of the tree, reducing the internal delay of the clock gater and the overall latency of the gated sub-clock.
eg.: sky130hs/ibex
image

### **Cloning clock gaters**

Still related to the placement, since clock gaters drive a small portion of the sinks they can some time be very spread out throw the design, causing some clusters of cells far from one and the other. This can lead to some big skew between this clusters.
To fix this this PR identifies these clusters and separates them into different trees, by creating a clone of the original clock gater for each new cluster. And in the final step the new trees are balanced by the `LatencyBalancer`.
The clusters are identified using a steiner tree, from the tree we find the long edges and separate the net into the clusters.

- In this PR the "long edge" is defined by a threshold value, that is currently determined as: the size of biggest size of the die / 5.
- This is a provisional value and it should be refined as it does not take into account PDK data, but it is set to this value as we do not have many examples of cases where we need cloning.

eg.: unity test gated_clock1

Original gated clock net:
Captura de tela de 2025-11-09 14-35-50

Steiner tree and long edge separation:
image

New cloned gated clock and original gated clock:
Captura de tela de 2025-11-09 14-55-23


rovinski:
Great to see this PR! Gate cloning is exceptionally important for clock gating to have reasonable performance.

> To fix this issue we enable CTS to move the clock gaters to the center of its output net bounding box, reducing the connection between the clock gater and the first buffer of the tree, reducing the internal delay of the clock gater and the overall latency of the gated sub-clock.

I am curious - do you check the timing on the enable signal as well as the clock signal? Both the clock signal and the enable signal could be the critical path, so it could be that optimizing for the clock signal actually slows the enable signal down and makes the overall path worse.

> the size of biggest size of the die / 5.

Can you elaborate on this? If the die size is very small, would it do cloning because it is greater than 1/5 the die size? Is there a rollback if cloning the clock gate doesn't actually improve timing?

> it does not take into account PDK data

I think RSZ already calculates a "maximum wire length" for buffering. Perhaps this value can be used somehow?

QuantamHD:
Have you tried this with the new cgt module. I'm curious how the two interact

arthurjolo:
CI is showing very good results for ibex on sky130hs and sky130hd, but some degradation on ibex ng45 and gf180. One small metric failure that I will create a PR to fix.

> Great to see this PR! Gate cloning is exceptionally important for clock gating to have reasonable performance.
> 
> > To fix this issue we enable CTS to move the clock gaters to the center of its output net bounding box, reducing the connection between the clock gater and the first buffer of the tree, reducing the internal delay of the clock gater and the overall latency of the gated sub-clock.
> 
> I am curious - do you check the timing on the enable signal as well as the clock signal? Both the clock signal and the enable signal could be the critical path, so it could be that optimizing for the clock signal actually slows the enable signal down and makes the overall path worse.

We only focus on the clock signal. On the entire flow, if I am not mistaken timing paths don't consider the enable signal path during timing analysis.
> 
> > the size of biggest size of the die / 5.
> 
> Can you elaborate on this? If the die size is very small, would it do cloning because it is greater than 1/5 the die size? Is there a rollback if cloning the clock gate doesn't actually improve timing?
> 
> > it does not take into account PDK data
> 
> I think RSZ already calculates a "maximum wire length" for buffering. Perhaps this value can be used somehow?

Indeed this initial value is not ideal and it is preliminary, but it showed to have good results in our testcases. There is no rollback, the cloning is done pre clock tree building, it is hard to determine the improvement or not as we are still going to build the tree.

The value for the resizer is one of the candidates I want to try. But I also wan't to have a way for CTS to determine this value (not only for clock gate cloning but for repair_clock_nets as well). So my idea was to create this initial PR that has good results for our testcases and work on the improvements next.

arthurjolo:
> Have you tried this with the new cgt module. I'm curious how the two interact

No, I don't think we have any examples on the CI that uses ctg. Is there any examples using that use ctg and complete the flow?

rovinski:
> We only focus on the clock signal. On the entire flow, if I am not mistaken timing paths don't consider the enable signal path during timing analysis.

STA definitely considers (or should consider) the enable signal path. While it typically wouldn't be considered part of the clock tree, it does affect the timing of signals that pass through it.

Thought experiment: Imagine an enable source on one side of the chip, and the clock and sink on the other side of the chip. If you move the clock gate close to the sink, now the enable source is very far away and will take longer than the clock.

arthurjolo:
I looked into some designs with clock gaters and found that, as I imagined, we do not consider the enable pin as part of the clock path. But we do have timing constraints on the enable pin to make sure the clock signal is stable. I checked the results when using this branch and no design showed problems on this enable pins.

rovinski:
> I checked the results when using this branch and no design showed problems on this enable pins.

That's fine. But just be forewarned that this may show up in the future. I don't think that the current examples using clock gating are very robust right now.

arthurjolo:
> > I checked the results when using this branch and no design showed problems on this enable pins.
> 
> That's fine. But just be forewarned that this may show up in the future. I don't think that the current examples using clock gating are very robust right now.

Yes, we don't have a robust test for clock gaters. I will create a ticket to keep track of the delay in enable pins from clock gaters. Thanks for pointing this issue!

maliberty:
ok file needs fixing.  Are there any metrics updates?

src/cts/src/TritonCTS.cpp
```diff
@@ -1018,6 +1020,320 @@ std::string TritonCTS::selectBestMaxCapBuffer(
...
+}
+
+void TritonCTS::findLongEdges(
+    stt::Tree& clkSteiner,
+    odb::Point driverPt,
+    std::map<odb::Point, std::vector<odb::dbITerm*>>& point2pin)
+{
+  double maxWlMicrons
+      = resizer_->findMaxWireLength(/* don't issue error */ false) * 1e+6;
+  int threshold = block_->micronsToDbu(maxWlMicrons);
```
maliberty: const

src/cts/src/TritonCTS.cpp
```diff
@@ -1018,6 +1020,320 @@ std::string TritonCTS::selectBestMaxCapBuffer(
...
+  for (int b = 0; b < clkSteiner.branchCount(); b++) {
+    const stt::Branch branch = clkSteiner.branch[b];
+    const stt::Branch* neighbor = &clkSteiner.branch[branch.n];
+    odb::Point branchPt = {branch.x, branch.y};
+
+    odb::Point neighborPt = {neighbor->x, neighbor->y};
+    int64_t dist = odb::Point::manhattanDistance(branchPt, neighborPt);
+    int clusterFrom
+        = iterm2cluster.find(b) == iterm2cluster.end() ? -1 : iterm2cluster[b];
+    int clusterTo = iterm2cluster.find(branch.n) == iterm2cluster.end()
```
maliberty: const

src/cts/src/TritonCTS.cpp
```diff
@@ -1018,6 +1020,320 @@ std::string TritonCTS::selectBestMaxCapBuffer(
...
+                 clone->getName());
+      debugPrint(logger_,
+                 CTS,
+                 "clock gate cloning",
+                 2,
+                 " Original net {}",
+                 cloneNet->getName());
+    } else {
+      // Create the ICG clone
+      // Creat a new input net
```
maliberty: typo Creat

maliberty (changes requested):

arthurjolo:
> ok file needs fixing. Are there any metrics updates?

There are some metrics updates, I am making the PR now.

src/cts/src/TritonCTS.cpp
```diff
@@ -1018,6 +1020,320 @@ std::string TritonCTS::selectBestMaxCapBuffer(
...
+                   2,
+                   "  Connects sink {}",
+                   sink->getName());
+        int sinkX, sinkY;
+        sink->getAvgXY(&sinkX, &sinkY);
+        sinksBbox.merge({sinkX, sinkY});
+        if (disconectNets) {
+          // Connect sinks to new clone instance
+          sink->disconnect();
+          sink->connect(cloneNet);
```
jhkim-pii, line 1319: `hierarchicalConnect()` can make the flat net connection. So you can omit the flat net connection codes.
However, it is ok to include the code. I am mentioning this to inform you about the `hierarchicalConnect()` function.

arthurjolo: Do I still need to disconnect the destination Pin? Or `hierarchicalConnect()` also disconnects it?

maliberty:
Needs an update to
```
06:25:37  + ./test/regression ibex_sky130hd
06:29:37  ibex_sky130hd (tcl) *FAIL* DRT::ANT::errors   1 >   0
```

src/cts/src/TritonCTS.cpp
jhkim-pii: You don't have to disconnect it manually. `dbITerm::connect(...)` API used in `hierarchicalConnect()` automatically disconnects the old connection if exists.

PR #8885, opened 2025-11-21
ram: added tapcell placement and spacing

braydenlouie:
- Added option to insert tapcells with user-defined maximum distance/spacing

### DIFF
--- a/src/ram/src/ram.cpp
+++ b/src/ram/src/ram.cpp
@@ -177,7 +177,7 @@ std::unique_ptr<Layout> RamGen::generateTapColumn(const int word_count,
                                                   const int tapcell_col)
 {
   auto tapcell_layout = std::make_unique<Layout>(odb::vertical);
-  for (int i = 0; i < word_count; ++i) {
+  for (int i = 0; i <= word_count; ++i) {
     auto tapcell_cell = std::make_unique<Cell>();
     makeCellInst(tapcell_cell.get(),
                  "tapcell",
@@ -575,9 +575,6 @@ void RamGen::generate(const int bytes_per_word,
       int nearest_tap
           = (max_tap_dist / ram_grid.getWidth()) * ram_grid.getLayoutWidth(0);
       int tapcell_count = 0;
-      // x distance between transistor and tapcell means 2x distance between
-      // tapcells
-      max_tap_dist *= 2;
       // iterates through each of the columns
       for (int col = 0; col < ram_grid.numLayouts(); ++col) {
         if (nearest_tap + ram_grid.getLayoutWidth(col) >= max_tap_dist) {
@@ -591,7 +588,7 @@ void RamGen::generate(const int bytes_per_word,
         nearest_tap += ram_grid.getLayoutWidth(col);
       }
       // check for last column in the grid
-      if (nearest_tap >= max_tap_dist / 2) {
+      if (nearest_tap >= max_tap_dist) {
         auto tapcell_layout = generateTapColumn(word_count, tapcell_count);
         ram_grid.addLayout(std::move(tapcell_layout));
       }
--- a/src/ram/src/ram.tcl
+++ b/src/ram/src/ram.tcl
@@ -58,7 +58,7 @@ proc generate_ram_netlist { args } {
     }
     set tapcell $keys(-tapcell)
   } else {
-    utl::info RAM 22 "No tapcell is specified.
+    utl::warn RAM 22 "No tapcell is specified.
         The generated layout may not pass Design Rule Checks."
   }
 
--- a/src/ram/test/make_8x8.defok
+++ b/src/ram/test/make_8x8.defok
@@ -3,372 +3,425 @@ DIVIDERCHAR "/" ;
 BUSBITCHARS "[]" ;
 DESIGN RAM8x8 ;
 UNITS DISTANCE MICRONS 100 ;
-DIEAREA ( 0 0 ) ( 12282 2448 ) ;
-ROW RAM_ROW0 unithd 0 0 N DO 267 BY 1 STEP 46 0 ;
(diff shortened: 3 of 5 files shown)


src/ram/src/layout.cpp
```diff
@@ -34,6 +34,10 @@ void Cell::addInst(dbInst* inst)
 
 void Cell::cellInit()
 {
+  if (width_ > 0 || height_ > 0) {
+    width_ = 0;
+    height_ = 0;
+  }
```
rovinski: What is the purpose of this?

src/ram/src/ram.tcl
```diff
@@ -45,8 +47,19 @@ proc generate_ram_netlist { args } {
...
+  set tapcell ""
+  set max_tap_dist 0
+  if { [info exists keys(-tapcell)] } {
+    if { [info exists keys(-max_tap_dist)] } {
+      set max_tap_dist $keys(-max_tap_dist)
+    } else {
+      utl::error RAM 21 "The -max_tap_dist argument must be specified with tapcell."
+    }
+    set tapcell $keys(-tapcell)
+  }
```
rovinski, line 63: Generate a warning if no tapcell is specified. Something like "No tapcell is specified. The generated layout may not pass Design Rule Checks."

src/ram/src/ram.tcl
```diff
@@ -45,8 +47,19 @@ proc generate_ram_netlist { args } {
...
+
   ram::generate_ram_netlist_cmd $bytes_per_word $word_count $storage_cell \
-    $tristate_cell $inv_cell $read_ports
+    $tristate_cell $inv_cell $read_ports $tapcell $max_tap_dist
```
rovinski, line 66: Recommend that the user specifies the tap distance in um and then you convert to dbu before passing to `ram::generate_ram_netlist_cmd`

src/ram/src/ram.cpp
```diff
@@ -546,6 +550,44 @@ void RamGen::generate(const int bytes_per_word,
...
+          ++col;
+          nearest_tap = 0;
+        } 
+        nearest_tap += ram_grid.getLayoutWidth(col);
+      }
+    } 
+  }  
+
+  ram_grid.gridInit();
+
```
rovinski: Can you add some comments?

src/ram/src/ram.cpp
```diff
@@ -382,10 +384,12 @@ void RamGen::generate(const int bytes_per_word,
   storage_cell_ = storage_cell;
   tristate_cell_ = tristate_cell;
   inv_cell_ = inv_cell;
+  tapcell_ = tapcell;
   and2_cell_ = nullptr;
   clock_gate_cell_ = nullptr;
   buffer_cell_ = nullptr;
   findMasters();
+  max_tap_dist *= 10;
```
rovinski: What does this do?

src/ram/src/ram.cpp
```diff
@@ -546,6 +550,44 @@ void RamGen::generate(const int bytes_per_word,
...
+        if (nearest_tap >= max_tap_dist) {
+          auto tapcell_layout = std::make_unique<Layout>(odb::vertical);
+          for (int i = 0; i < word_count; ++i) {
+            auto tapcell_cell = std::make_unique<Cell>();
+            makeCellInst(tapcell_cell.get(),
+                         "tapcell",
+                         fmt::format("cell{}_{}", col, i),
+                         tapcell_,
+                         {});
+            tapcell_layout->addCell(std::move(tapcell_cell));
```
rovinski: Maybe you should make a function like `generateTapColumn()` which generates the tapcell column, then you can insert it.

src/ram/src/layout.cpp
braydenlouie: If the grid gets reinitialized, then the width and height get reset. Otherwise, the spacing between the cells gets doubled

src/ram/src/ram.cpp
braydenlouie: Supposed to be a unit conversion, but will fix before passing it in

rovinski:
@braydenlouie please update the ok files and fix the code formatting. Then you can mark the PR as ready.

src/ram/src/layout.cpp
rovinski: But isn't the logic here "if width or height are non-zero, set them to zero"? Which is the exact same as "always set width and height to 0"?
e.g.
```suggestion
  width_ = 0;
  height_ = 0;
```

src/ram/src/ram.tcl
maliberty: use warn not info

src/ram/src/ram.tcl
maliberty: More than recommended - please make this change to be consistent with other commands (eg tapcell itself).

maliberty:
There is no tapcell in the topmost row:
image

maliberty:
The test uses a tap distance of 15 but that doesn't seem to have been obeyed:
image

maliberty:
Since the rows inside the RAM will not be continuous with those outside endcaps will also be needed.

rovinski:
> There is no tapcell in the topmost row:

@braydenlouie this needs to be addressed - all rows should have tapcells

> The test uses a tap distance of 15 but that doesn't seem to have been obeyed:

That's my fault; `max_tap_dist` is the maximum tapcell to transistor distance. Therefore the tap-to-tap distance is twice that amount. @braydenlouie can you remove the 2x factor on the tap distance?

> Since the rows inside the RAM will not be continuous with those outside endcaps will also be needed.

To be handled in a later PR. Sky130 doesn't have endcaps.

src/ram/src/ram.tcl
rovinski: this was addressed

rovinski (approved):
@maliberty I think this is ready to go

maliberty (approved):

PR #8888, opened 2025-11-24
pad: fix pads stuck under obstructions causing failing placement

gadfort:
Fixes:
- when pads would get under an obstruction they could get pushed into the middle and from there pads that needed to tunnel through the obstruction where prevented.
- This adds a push for pads under obstructions when a tunnel was needed, but not able to be completed.

### DIFF
--- a/src/pad/src/PadPlacer.cpp
+++ b/src/pad/src/PadPlacer.cpp
@@ -1347,13 +1347,15 @@ bool PlacerPadPlacer::padSpreading(
         total_force);
   }
 
-  // Push instances causing tunneling to fail
+  // Push instances causing tunneling to fail, these are instances that are too
+  // far from the instance that needs to tunnel/jump over the obstruction and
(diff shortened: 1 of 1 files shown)


maliberty:
I'm not sure what tunneling means in this context.  Please explain further.

gadfort:
@maliberty an obstruction is anything in that the pads interfere with placement (such as a bump or other fixed instances). When the iterative pad spreading is running eventually a pad will get "stuck" on the edge of the obstruction and because there is no "smooth" way to move around the obstruction, I just chose to call it tunneling when the instance jumps over the obstruction.

maliberty:
It would be good to add a comment to the code to explain that as others may have the same confusion in the future.

gadfort:
> It would be good to add a comment to the code to explain that as others may have the same confusion in the future.

Added some comments.

maliberty (approved):

PR #8892, opened 2025-11-25
Gui show 3d

### LINKED ISSUE #8856: GUI: Chiplets gui on calling gui::show
### Describe the bug

Currently, when I run openroad with the -gui option and I read_3dbx a design, the gui is showing the chiplets correctly. However, when I run openroad without the -gui option, read_3dbx and then call gui::show, the gui is not showing the design and it looks like an empty design.

### Expected Behavior

openroad -gui should perform the same as gui::show.

### Environment

```shell
NA
```

### To Reproduce

.

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

LucasYuki:
Closes #8856

3DBlox chips doesn't have block, so it's necessary to load only the chip in the GUI.


maliberty:
Theoretically you could have one hier chip that instantiates another hier chip.  @osamahammad21 is there any way to detect that?

PR #8899, opened 2025-11-25
odb: Bmap writer

rafaelmoresco:


src/odb/src/3dblox/bmapWriter.cpp
```diff
@@ -0,0 +1,50 @@
...
+      const auto inst_name = bump->getInst()->getName();
+      const auto cell_type = bump->getInst()->getMaster()->getName();
+      auto point = bump->getInst()->getOrigin();
+      line.append(inst_name);
+      line.append(" ");
+      line.append(cell_type);
+      line.append(" ");
+      line.append(std::to_string(point.getX() / u));
+      line.append(std::to_string(point.getY() / u));
+      line.append(" - -\n");
```
osamahammad21: Missing the pin and the net name.

PR #8889, opened 2025-11-24
rmp: eqy test check

b-chmiel:
Adds eqy check for gcd_annealing1 and gcd_annealing2 tests in rmp.

As those tests require loading liberty files for eqy, changed `run_equivalence_test`'s interface to allow selecting both a library directory (for globbing sources) and explicit listing of liberty files.

Follow-up to https://github.com/The-OpenROAD-Project/OpenROAD/pull/8437#issuecomment-3365934747

PR #8908, opened 2025-11-26
rmp: Genetic algorithm for slack tuning

b-chmiel:
New resynthesis strategy for `rmp`.
Alternative approach to https://github.com/The-OpenROAD-Project/OpenROAD/pull/8437 for optimizing ABC timing.
Utilizes a genetic algorithm with the following tunable parameters:
- population size
- mutation probability
- cross probability
- tournament size
- tournament probability

Results from regression tests:

|test|slack|execution time [s]|
|---------|------|--------------|
|[asap7 aes_genetic](https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908/files#diff-a71881a40b67295fcbb984645dd09f94848fc1558a3e91c0e4401428a994009a)|+26.52|164.3|
|[asap7 aes_annealing](https://github.com/The-OpenROAD-Project/OpenROAD/blob/089d41d806190e95da38bac29965b37bcd95fd4c/src/rmp/test/aes_annealing.tcl#L1)|+14.28|52.7|

Marking as draft for a discussion, currently working on some minor implementation restructuring and algorithm parameter tuning.

Depends on https://github.com/The-OpenROAD-Project/OpenROAD/pull/8889.


maliberty:
aes on what pdk?  Are the slacks negative or positive?

b-chmiel:
Tested on `asap7` PDK.
Results (in the regression test suite):
[aes_annealing.ok](https://github.com/The-OpenROAD-Project/OpenROAD/blob/089d41d806190e95da38bac29965b37bcd95fd4c/src/rmp/test/aes_annealing.ok#L161)
[aes_genetic.ok](https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908/files#diff-fd9ae51ed47beba73a0556f391ceff9fd151e679c339c9c640c012791681b9c1R166)

Slacks are positive, the higher the better. I'll adjust the description.

maliberty:
@QuantamHD please assign a Google reviewer

QuantamHD:
@calewis could you review this PR from antmicro?

src/rmp/README.md
```diff
@@ -96,6 +96,11 @@ resynth_annealing
 | `-revert_after` | After the given number of iterations that worsen slack, revert to best found solution. |
 | `-initial_ops` | Size of the initial random solution (number of commands in the script for ABC). |
 
+### Resynth with genetic slack tuning
+
+TODO
```
calewis: Need to finish this TODO

src/rmp/src/genetic_strategy.h
```diff
@@ -0,0 +1,76 @@
...
+#include <random>
+#include <vector>
+
+#include "cut/abc_library_factory.h"
+#include "gia.h"
+#include "slack_tuning_strategy.h"
+#include "sta/Delay.hh"
+#include "utl/unique_name.h"
+
+namespace rsz {
```
calewis: Probably makes sense to put the fwd decls in slack_tuning_strategy.h so you don't have to repeat them for the genetic and annealing strategies.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+                             utl::Logger* logger)
+{
+  struct HashVector final
+  {
+    size_t operator()(const SolutionSlack::Type& sol) const
+    {
+      size_t res = 0;
+      for (const auto& item : sol) {
+        size_t hash = boost::hash<size_t>()(item.id);
+        // TODO set seed
```
calewis: Need to implement. 

Or make the underlying data hashable and use `absl::Hash` as your hash function.  absl automatically hashes common containers with hashable types.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "utl/deleter.h"
+#include "utl/unique_name.h"
+
+namespace rmp {
+using utl::RMP;
+
+static void removeDuplicates(std::vector<SolutionSlack>& population,
+                             utl::Logger* logger)
+{
+  struct HashVector final
```
calewis: Given this is a function scoped struct in a cpp file, `final` is a bit verbose.  It's not hurting anything, but is a bit weird looking since it's only purpose is to stop someone from inheriting from it in the future.   It's a noop from an optimization standpoint.

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+std::string SolutionSlack::toString() const
+{
+  if (solution_.empty()) {
+    return "[]";
+  }
+
+  std::ostringstream resStream;
+  resStream << '[';
+  if (!solution_.empty()) {
```
calewis: We already confirmed this to be the case with the check on line 35

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,198 @@
...
+  } else {
+    resStream << "not computed";
+  }
+  return resStream.str();
+}
+
+bool SolutionSlack::operator<(const SolutionSlack& other) const
+{
+  // Highest slack first
+  return worst_slack_ > other.worst_slack_;
```
calewis: This is less than other if our worst_slack_ is greater than other's worst slack?  It's a bit surprising that the return doesn't use `<`.

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+SolutionSlack::Type SolutionSlack::RandomNeighbor(
+    const SolutionSlack::Type& all_ops,
+    utl::Logger* logger,
+    std::mt19937& random) const
+{
+  SolutionSlack::Type sol = solution_;
+  enum Move
```
calewis: Is there a reason to not use `enum class`?

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,198 @@
...
+  {
+    ADD,
+    REMOVE,
+    SWAP,
+    COUNT
+  };
+  Move move = ADD;
+  if (sol.size() > 1) {
+    move = Move(random() % (COUNT));
```
calewis: Consider using absl::uniform_int_distribution in this function (not std unfortunately), but the % behavior is probably okay for our goals here.

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,198 @@
...
+      sol.erase(sol.begin() + i);
+    } break;
+    case SWAP: {
+      debugPrint(
+          logger, RMP, "annealing", 2, "Swapping adjacent GIA operations");
+      size_t i = random() % (sol.size() - 1);
+      std::swap(sol[i], sol[i + 1]);
+    } break;
+    case COUNT:
+      // unreachable
```
calewis: I would add a `__builtin_unreachable` here so you can optimize out this case.   You get slightly nicer asm https://godbolt.org/z/1ehnhqrcx, given that it actually is unreachable.   Clang and gcc both have __builtin_unreachable so I think it's very unlikely to encounter a situation where it doesn't work.

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,198 @@
...
+    case REMOVE: {
+      debugPrint(logger, RMP, "annealing", 2, "Removing a GIA operation");
+      size_t i = random() % sol.size();
+      sol.erase(sol.begin() + i);
+    } break;
+    case SWAP: {
+      debugPrint(
+          logger, RMP, "annealing", 2, "Swapping adjacent GIA operations");
+      size_t i = random() % (sol.size() - 1);
+      std::swap(sol[i], sol[i + 1]);
```
calewis, line 92: We have to check that sol.size() >= 2 before doing a swap. Otherwise you'll hit `x % 0` which is UB.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+        // TODO set seed
+        boost::hash_combine(hash, res);
+      }
+      return res;
+    }
+  };
+
+  std::unordered_set<SolutionSlack::Type, HashVector> taken;
+  population.erase(
+      std::ranges::begin(std::ranges::remove_if(
```
calewis, line 35: We should be C++20 soon, but we can't use ranges till we move from 17.

src/rmp/src/gia.h
```diff
@@ -0,0 +1,42 @@
...
+class Vertex;
+}  // namespace sta
+
+namespace utl {
+class Logger;
+}  // namespace utl
+
+namespace rmp {
+
+struct GiaOp final
```
calewis, line 25: Can make GiaOp absl hashable via:
```
template <typename H>
    friend H AbslHashValue(H h, ...
```

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+      for (const auto& item : sol) {
+        size_t hash = boost::hash<size_t>()(item.id);
+        // TODO set seed
+        boost::hash_combine(hash, res);
+      }
+      return res;
+    }
+  };
+
+  std::unordered_set<SolutionSlack::Type, HashVector> taken;
```
calewis: If you make GiaOp hashable then you can use absl::flat_hash_set<SolutionSlack::Type> here or you can use the unordered_set with `absl::Hash`.  Check with @maliberty, but I think using abseil is preferable to using boost when abseil provides equivalent functionality.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "rsz/Resizer.hh"
+#include "slack_tuning_strategy.h"
+#include "utl/Logger.h"
+#include "utl/deleter.h"
+#include "utl/unique_name.h"
+
+namespace rmp {
+using utl::RMP;
+
+static void removeDuplicates(std::vector<SolutionSlack>& population,
```
calewis, line 28: If you don't care about order then sort + unique won't require hashable and unless population is really big the extra log N factor shouldn't be an issue.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+  debugPrint(logger,
+             RMP,
+             "population",
+             1,
+             "Generating and evaluating the initial population");
+  std::vector<SolutionSlack> population(pop_size_);
+  for (auto& ind : population) {
+    ind.solution_.reserve(initial_ops_);
+    for (size_t i = 0; i < initial_ops_; i++) {
+      ind.solution_.push_back(all_ops[random_() % all_ops.size()]);
```
calewis: I assume it is okay to get the same op multiple times?

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+      }
+      sol_slack.Evaluate(candidate_vertices,
+                         abc_library,
+                         corner_,
+                         sta,
+                         name_generator,
+                         logger);
+    }
+    // Selection
+    std::ranges::sort(population, std::less{});
```
calewis: Not quite on ranges just yet, you might be able to hold the PR till we bump to 20 though.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+    newPopulation.reserve(pop_size_);
+    for (int j = 0; j < pop_size_; j++) {
+      std::vector<size_t> tournament(tourn_size_);
+      std::generate_n(tournament.begin(), tourn_size_, [&]() {
+        return random_() % population.size();
+      });
+      std::ranges::sort(tournament);
+      tournament.erase(std::ranges::begin(std::ranges::unique(tournament)),
+                       tournament.end());
+      std::bernoulli_distribution bern_dist{tourn_prob_};
```
calewis: Construction is probably cheap, but you can remove this declaration from the loop.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+      for (const auto& candidateId : tournament) {
+        if (bern_dist(random_)) {
+          newPopulation.push_back(population[candidateId]);
+          break;
+        }
+      }
```
calewis, line 145: Is it okay if bern_dist is false for every candidate?

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+                         corner_,
+                         sta,
+                         name_generator,
+                         logger);
+    }
+    // Selection
+    std::ranges::sort(population, std::less{});
+    std::vector<SolutionSlack> newPopulation;
+    newPopulation.reserve(pop_size_);
+    for (int j = 0; j < pop_size_; j++) {
```
calewis: Maybe I'm misunderstanding the goal, but could we do: 
```
decltype(population) tourny;
tourny.reserve(tourn_size_);
std::bernoulli_distribution bern_dist{tourn_prob_};

for (int j = 0; j < pop_size_; j++) {
  tourny.clear();
  std::sample(population.begin(), population.end(), 
                       std::back_inserter(tourny), tourn_size_, random_);
  auto winner = std::find_if(tourny.begin(), tourny.end(), 
                          [&](auto const &_) {return bern_dist(random_);});
  if (winner != tourny.end()){
     newPopulation.push_back(*winner);
  } else { // is this okay? }
}
  ```

File: src/rmp/src/genetic_strategy.cpp [hunk 20]
```diff
@@ -0,0 +1,171 @@
...
+      std::vector<size_t> tournament(tourn_size_);
```
calewis: If I misunderstood above then we can at least do something like this. Extract the tournament vector from the `j` loop.   Make another vector called indices that has indices 0..pop_size_-1.   Then you can do:
```
tournament_candidates.clear();
std::sample(indices.begin(), indices.end(),
                    std::back_inserter(tournament_candidates),
                    tourn_size_,
                    random_);
```
The indices will come out already sorted so you won't need to sort and erase.   The only difference is that tournament will always have size tourn_size_ since std::sample doesn't produce duplicates.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+      std::bernoulli_distribution bern_dist{tourn_prob_};
+      for (const auto& candidateId : tournament) {
+        if (bern_dist(random_)) {
+          newPopulation.push_back(population[candidateId]);
+          break;
+        }
+      }
+    }
+    removeDuplicates(newPopulation, logger);
+    population = newPopulation;
```
calewis: Can move here to avoid a third allocation.

src/rmp/src/genetic_strategy.cpp
calewis: We may need to switch to absl's bernoulli_distribution since the guts of std::bernoulli_distribution are implementation defined.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+    debugPrint(logger, RMP, "genetic", 1, candidate.toString());
+  }
+
+  for (unsigned i = 0; i < iterations_; i++) {
+    logger->info(RMP, 65, "Resynthesis: Iteration {} of genetic algorithm", i);
+    unsigned generation_size = population.size();
+    // Crossover
+    unsigned cross_size = std::max<unsigned>(cross_prob_ * generation_size, 1);
+    for (unsigned j = 0; j < cross_size; j++) {
+      auto rand1 = random_() % generation_size;
```
calewis: Could just std/boost sample 2 parents to avoid computing rand1, rand2, and checking for collisions.

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,171 @@
...
+      population.emplace_back(child_sol_slack);
+    }
+    // Mutations
+    unsigned mut_size = std::max<unsigned>(mut_prob_ * generation_size, 1);
+    for (unsigned j = 0; j < mut_size; j++) {
+      SolutionSlack sol_slack;
+      auto rand = random_() % generation_size;
+      sol_slack.solution_
+          = population[rand].RandomNeighbor(all_ops, logger, random_);
+      population.emplace_back(sol_slack);
```
calewis: Mutations should be in place? Or we want to add them to the end of the population?

src/rmp/src/gia.cpp
```diff
@@ -0,0 +1,426 @@
...
+#include "sta/GraphDelayCalc.hh"
+#include "sta/Search.hh"
+#include "utils.h"
+#include "utl/Logger.h"
+#include "utl/SuppressStdout.h"
+#include "utl/deleter.h"
+#include "utl/unique_name.h"
+
+namespace abc {
+extern Abc_Ntk_t* Abc_NtkFromAigPhase(Aig_Man_t* pMan);
```
calewis, line 46: There isn't a header for these?

src/rmp/src/slack_tuning_strategy.h
```diff
@@ -0,0 +1,97 @@
...
+                               unsigned iterations,
+                               unsigned initial_ops)
+      : corner_(corner),
+        slack_threshold_(slack_threshold),
+        seed_(seed),
+        iterations_(iterations),
+        initial_ops_(initial_ops)
+  {
+    if (seed_) {
+      random_.seed(*seed_);
```
calewis: This is the only place seed_ is used, if we are never going to use it anywhere else we can just remove it from the member variables.

calewis:
> @calewis could you review this PR from antmicro?

Almost done, I just need to go through src/rmp/src/gia.cpp tomorrow.

src/rmp/src/gia.cpp
```diff
@@ -0,0 +1,426 @@
...
+  gia = new_gia;
+}
+
+std::vector<GiaOp> GiaOps(utl::Logger* logger)
+{
+  // GIA ops as lambdas
+  // All the magic numbers are defaults from abc/src/base/abci/abc.c
+  // Or from the ORFS abc_speed script
+  using OpType = std::function<void(abc::Gia_Man_t*&)>;
+  std::vector<OpType> all_ops = {
```
calewis: You could reduce some boiler plate with:
```
auto step = [logger](const char* msg, auto func) {
    return [logger, msg, func](auto& gia) {
        debugPrint(logger, RMP, "gia", 1, msg);
        // func(gia) executes the logic and returns the new pointer
        replaceGia(gia, func(gia));
    };
};

std::vector<OpType> all_ops = {
    // 2. Use the helper for the standard one-liners
    step("Starting rehash", [](auto& g) { 
        return Gia_ManRehash(g, false); 
    }),
```
 
 Just keep the longer ones as is and replace the simple ones with this.

calewis: https://godbolt.org/z/xvza18faz as a simple example.

src/rmp/src/gia.cpp
```diff
@@ -0,0 +1,426 @@
...
+               "gia",
+               1,
+               "Mapped ABC network has {} nodes and {} POs.",
+               abc::Abc_NtkNodeNum(current_network.get()),
+               abc::Abc_NtkPoNum(current_network.get()));
+
+    current_network->pManFunc = library;
+    abc::Gia_Man_t* gia = nullptr;
+
+    {
```
calewis: Is this scope triggering some RAII cleanup? If not why is it here?

src/rmp/src/gia.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+  utl::UniquePtrWithDeleter<abc::Abc_Ntk_t> mapped_abc_network
+      = cut.BuildMappedAbcNetwork(abc_library, network, logger);
+
+  utl::UniquePtrWithDeleter<abc::Abc_Ntk_t> current_network(
+      abc::Abc_NtkToLogic(
+          const_cast<abc::Abc_Ntk_t*>(mapped_abc_network.get())),
+      &abc::Abc_NtkDelete);
+
+  {
```
calewis, line 278: Why this scope, the only thing not in it is: 
```
  cut.InsertMappedAbcNetwork(
      current_network.get(), abc_library, network, name_generator, logger);
```
Does something need to be cleaned up before that call? 

This function is a bit long so I wouldn't be opposed to breaking some of the scopes into helpers to make things more clear.

src/rmp/src/slack_tuning_strategy.cpp
calewis: Check that that's okay with @maliberty though, maybe we don't allow non-standard builtins.

maliberty:
I'll review after your comments are addressed

src/rmp/src/genetic_strategy.cpp
rovinski: C++20 should be the default as of #8886

calewis: Thanks I knew it was imminent, but missed the switch flip.

src/rmp/src/genetic_strategy.cpp
b-chmiel: I decided to go with absl::Hash with unordered_set approach to keep order stable and leave debugging output for each step.

src/rmp/src/slack_tuning_strategy.cpp
b-chmiel: I decided to go with absl::Uniform approach here (and in every other place where this patter occurs) to make it deterministic.

src/rmp/src/slack_tuning_strategy.cpp
b-chmiel: I used `ABSL_UNREACHABLE` from abseil. The preferable approach, `std::unreachable`, will be available since c++23.

Macro is defined here: https://github.com/abseil/abseil-cpp/blob/88c48235c8add3442800a19d5a7ee639dc242afa/absl/base/optimization.h#L207

src/rmp/src/slack_tuning_strategy.cpp
b-chmiel: Added an assert as it is assumed that RandomNeighbor moves (apart from ADD) will be performed on sol.size()>1

src/rmp/src/genetic_strategy.cpp
b-chmiel: Rebased branch and changed to C++ counterparts

src/rmp/src/genetic_strategy.cpp
b-chmiel: https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908#discussion_r2602351553

src/rmp/src/genetic_strategy.cpp
b-chmiel: https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908#discussion_r2602379386

src/rmp/src/gia.cpp
b-chmiel: Unfortunately no: https://github.com/search?q=repo%3AThe-OpenROAD-Project%2Fabc%20Abc_NtkFromAigPhase&type=code

As these methods are used here exclusively, I think making a facade is not needed now

src/rmp/src/slack_tuning_strategy.h
b-chmiel: Right, removed.

src/rmp/src/gia.cpp
b-chmiel: Done, thx for the example

src/rmp/src/gia.cpp
b-chmiel: I restructured gia/aig handling to use unique ptrs. Now it should be more readable with RAII cleanups

src/rmp/src/gia.cpp
b-chmiel: Simplified scopes and introduced unique ptrs, should be cleaner now

src/rmp/src/genetic_strategy.cpp
b-chmiel: In a single tournament round it shouldn't be a problem. I added handling of edge case when tournament does not add any candidate to the new population.

src/rmp/src/genetic_strategy.cpp
b-chmiel: I simplified it by using find_if. As `std::sample` is using std::uniform_int_distribution` I restrained it's usage and implemented explicit randomization with absl::Uniform utility.

src/rmp/src/genetic_strategy.cpp
b-chmiel: As in https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908#discussion_r2578989331

src/rmp/src/genetic_strategy.cpp
b-chmiel: I don't think we need to implement fully deterministic sample just for this case, but I can add it if required

src/rmp/src/genetic_strategy.cpp
b-chmiel: I'll measure efficiency of both solutions.

src/rmp/src/genetic_strategy.cpp
b-chmiel: I do not see any contraindications for that. For instance, the operation [Gia_ManAigSyn2](https://github.com/The-OpenROAD-Project/abc/blob/8adf31774762ba1e37818e25b5ecb5916607c9c6/src/aig/gia/giaScript.c#L69) invokes `Gia_ManAreaBalance` mutliple times with the same parameters from which I presume that there is no idempotence here.

Operations with complex dependencies such as `&nf` or `&dsd` are disabled for now.

b-chmiel:
The only errors I'm getting right now are related to ABC imports. For instance, importing header such as `map/if/if.h` causes include cycle warning (which is there as it includes `map/if/acd/ac_wrapper.h` which includes `if.h`) and redeclaration errors in cmake build.
Not including this header causes [misc-include-cleaner](https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html) warning as those functions are not directly included in `ac_wrapper.h`. I'm also getting warnings from the inside of ABC library too.

@maliberty Should I avoid including this header altogether or ignore all problematical warnings with lint comments and compiler pragmas?

I'm also running a benchmark for https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908#discussion_r2579031594 to decide which mutation approach will be best.

Besides that, I believe this PR is ready for a review.

maliberty:
If you don't include if.h in ac_wrapper.h, what are "those functions" that it complains about?

b-chmiel:
@maliberty The `if.h` header is required to be able to construct `If_Par_t` struct with custom parameters (defined there).
This header throws two warnings:
-  `Wredundant-decls` (for repeated externs)
- `Wunused-but-set-variable` (for inline code in `utilTruth.h`)
I'd silenced those warnings for this header only, this resolved issues.

I'd also removed call to `ABSL_UNREACHABLE` as it requires bumping abseil to at least 2023. Then we may also add it to https://github.com/The-OpenROAD-Project/OpenROAD/blob/b57cfc641359379a86bd13e6a66bccdde1840a7e/src/dft/src/clock_domain/ClockDomain.cpp#L38 and nix environment files would need to be bumped as well in a separate PR.

b-chmiel:
@maliberty All tests pass now without warnings

src/rmp/test/.gitignore
```diff
@@ -0,0 +1,6 @@
+aes_annealing
+aes_annealing.v
+gcd_annealing1
+gcd_annealing1.v
+gcd_annealing2
+gcd_annealing2.v
```
maliberty: Test output should be written to the results directory so no .ignore should be present.

src/rmp/test/gcd_annealing1.ok
maliberty, line 1: Why does this PR affect the annealing results?  I thought it to be independent.

b-chmiel:
@maliberty Changes are ready, I rebased them and removed one GIA operation, as we observed it had no visible effect on slack: https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908/commits/e5b9e6c5b6b0fd65e6ca7c9dbe7a06b662322a63


Currently CI is failing due old docker images in the registry after bumping of absl:
https://github.com/The-OpenROAD-Project/OpenROAD/commit/3d6152aeb786eeaf1edc459ac8ecdff46d7df2f5

error message (note old absl 20240722 instead of 20260107):
```
/usr/bin/ld: warning: libbz2.so.1, needed by /opt/or-tools/lib/libortools.so.9.14.6206, may conflict with libbz2.so.1.0

15:41:47  #9 348.4 /usr/bin/ld: /usr/local/lib/libabsl_hash.a(hash.cc.o): in function `absl::lts_20240722::hash_internal::MixingHashState::CombineLargeContiguousImpl32(unsigned long, unsigned char const*, unsigned long)':

15:41:47  #9 348.4 hash.cc:(.text+0x30): undefined reference to `absl::lts_20240722::hash_internal::CityHash32(char const*, unsigned long)'

15:41:47  #9 348.4 /usr/bin/ld: /usr/local/lib/libabsl_hash.a(hash.cc.o): in function `absl::lts_20240722::hash_internal::MixingHashState::LowLevelHashImpl(unsigned char const*, unsigned long)':

15:41:47  #9 348.4 hash.cc:(.text+0x226): undefined reference to `absl::lts_20240722::hash_internal::LowLevelHashLenGt16(void const*, unsigned long, unsigned long, unsigned long const*)'

15:41:47  #9 348.4 /usr/bin/ld: /usr/local/lib/libabsl_hash.a(hash.cc.o): in function `absl::lts_20240722::hash_internal::MixingHashState::CombineContiguousImpl(unsigned long, unsigned char const*, unsigned long, std::integral_constant<int, 4>)':

15:41:47  #9 348.4 hash.cc:(.text._ZN4absl12lts_2024072213hash_internal15MixingHashState21CombineContiguousImplEmPKhmSt17integral_constantIiLi4EE[_ZN4absl12lts_2024072213hash_internal15MixingHashState21CombineContiguousImplEmPKhmSt17integral_constantIiLi4EE]+0x63): undefined reference to `absl::lts_20240722::hash_internal::CityHash32(char const*, unsigned long)'
```

vvbandeira:
@b-chmiel can you provide a link to the logs where you see these messages?
Looking at the CI what I see is an issue with the newer version?
```
[2026-02-10T10:43:31.435Z] #9 457.8 hash.cc:(.text+0x85f): undefined reference to `absl::lts_20260107::hash_internal::CityHash32(char const*, unsigned long)'
[2026-02-10T10:43:31.933Z] #9 458.2 collect2: error: ld returned 1 exit status
```

maliberty:
It is not a CI issue, you need to add
```
diff --git a/src/rmp/src/CMakeLists.txt b/src/rmp/src/CMakeLists.txt
index 89d7e0f288..d82e8eb509 100644
--- a/src/rmp/src/CMakeLists.txt
+++ b/src/rmp/src/CMakeLists.txt
@@ -65,6 +65,7 @@ target_link_libraries(rmp_lib
     cut
     ${ABC_LIBRARY}
     absl_hash
+    absl_city
 )
```

b-chmiel:
@maliberty @vvbandeira Thanks for help, fixed.

src/rmp/test/.gitignore
maliberty, line 1: This should not be needed as all tests should write to results/.  Please fix any tests writing directly to test/

src/rmp/src/rmp.tcl
```diff
@@ -152,3 +152,54 @@ proc resynth_annealing { args } {
...
+                                            [-tourn_prob tourn_prob]
+                                            [-tourn_size tourn_size]
+                                            [-iters iters]
+                                            [-initial_ops initial_ops]
+                                          }
+
+proc resynth_genetic { args } {
+  sta::parse_key_args "resynth_genetic" args \
+    keys {-corner -iters -seed -pop_size -mut_prob -cross_prob \
+              -tourn_size -tourn_prob -initial_ops -slack_threshold} \
```
maliberty: Please use full words for clarity (pop/mut/cross/tourn/prob/ops)

src/rmp/src/gia.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+      // &synch2
+      [logger](auto& g) {
+        abc::Dch_Pars_t pars = {};
+        Dch_ManSetDefaultParams(&pars);
+        pars.nBTLimit = 100;
+        debugPrint(logger, RMP, "gia", 1, "Starting synch2");
+        replaceGia(g, Gia_ManAigSynch2(g.get(), &pars, 6, 20));
+      }};
+  /* Some ABC functions/commands that could be used, but crash in some
```
maliberty: Is there a reason to keep this?  Is someone planning on enabling this code?

src/rmp/src/slack_tuning_strategy.h
```diff
@@ -0,0 +1,94 @@
...
+
+namespace utl {
+class Logger;
+}  // namespace utl
+
+namespace rmp {
+
+struct SolutionSlack final
+{
+  using Type = std::vector<GiaOp>;
```
maliberty: Type is very generic.  Is there nothing more descriptive?

src/rmp/src/gia.h
```diff
@@ -0,0 +1,51 @@
...
+{
+  using Type = utl::UniquePtrWithDeleter<abc::Gia_Man_t>;
```
maliberty: More descriptive name than Type?

src/rmp/src/slack_tuning_strategy.h
```diff
@@ -0,0 +1,94 @@
...
+  Type solution_;
+  std::optional<sta::Slack> worst_slack_ = std::nullopt;
```
maliberty: Struct data members do not having a trailing _ in their name (https://google.github.io/styleguide/cppguide.html#Variable_Names).

Should this be a class?  It has methods with behavior and fields that seem correlated.

src/rmp/src/slack_tuning_strategy.h
```diff
@@ -0,0 +1,94 @@
...
+
+  sta::Vertex* Evaluate(const std::vector<sta::Vertex*>& candidate_vertices,
+                        cut::AbcLibrary& abc_library,
+                        sta::Corner* corner,
+                        sta::dbSta* sta,
+                        utl::UniqueName& name_generator,
+                        utl::Logger* logger);
+};
+
+// TODO docs
```
maliberty: Any reason to defer?

src/rmp/src/slack_tuning_strategy.h
```diff
@@ -0,0 +1,94 @@
...
+                               unsigned iterations,
+                               unsigned initial_ops)
+      : corner_(corner),
+        slack_threshold_(slack_threshold),
+        iterations_(iterations),
+        initial_ops_(initial_ops)
+  {
+    if (seed) {
+      random_.seed(*seed);
+    }
```
maliberty: In what scenario would we want an unseeded random number generator?

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "sta/MinMax.hh"
+#include "utils.h"
+#include "utl/Logger.h"
+#include "utl/unique_name.h"
+
+namespace rmp {
+
+// The magic numbers are defaults from abc/src/base/abci/abc.c
+constexpr size_t SEARCH_RESIZE_ITERS = 100;
+constexpr size_t FINAL_RESIZE_ITERS = 1000;
```
maliberty: Use https://google.github.io/styleguide/cppguide.html#Constant_Names

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,193 @@
...
+         name_generator,
+         logger);
+
+  odb::dbDatabase::endEco(block);
+
+  auto [worst_slack, worst_vertex] = GetWorstSlack(sta, corner);
+  worst_slack_ = worst_slack;
+
+  odb::dbDatabase::undoEco(block);
+  return worst_vertex;
```
maliberty: How do you know worst_vertex still exists after undo?

src/rmp/test/gcd_annealing2.ok
```diff
@@ -136,34 +138,62 @@ Corner: slow
...
+        -1018.42   data arrival time
 ---------------------------------------------------------
-         -47.70   slack (VIOLATED)
+        -548.20   slack (VIOLATED)
 
 
-wns max -47.70
-tns max -1112.78
+wns max -548.20
+tns max -20829.82
```
maliberty: This result is substantially worse now.  Please explain.

src/rmp/src/CMakeLists.txt
```diff
@@ -48,6 +51,8 @@ target_include_directories(rmp
 target_include_directories(rmp_lib
   PUBLIC
     ../include
+  PRIVATE
+    .
```
maliberty: Unnecessary

src/rmp/src/slack_tuning_strategy.cpp
```diff
@@ -0,0 +1,193 @@
...
+    std::mt19937& random) const
+{
+  SolutionSlack::Type sol = solution_;
+  enum class Move : uint8_t
+  {
+    ADD,
+    REMOVE,
+    SWAP,
+    COUNT
+  };
```
maliberty, line 69: Use https://google.github.io/styleguide/cppguide.html#Enumerator_Names

src/rmp/src/genetic_strategy.cpp
```diff
@@ -0,0 +1,168 @@
...
+      auto winner = std::ranges::find_if(tournament, [&](auto const& _) {
+        return absl::Bernoulli(random_, static_cast<double>(tourn_prob_));
+      });
+      if (winner != tournament.end()) {
+        newPopulation.emplace_back(population[*winner]);
+      }
+    }
+    removeDuplicates(newPopulation, logger);
+
+    if (newPopulation.empty()) {
```
maliberty, line 150: How would you get an empty newPopulation?

src/rmp/test/gcd_annealing1.ok
b-chmiel: Annealing and genetic use the same logic for getting random neighbors which now uses compiler-agnostic `absl::Uniform`. This way got deterministic results for both compilers. 

The annealing strategy results should not be impacted, I'll make random trial of tests for different seeds to confirm this.

src/rmp/src/gia.cpp
b-chmiel: I think it is worth noting here that this is only a subset of ABC transformations and the reason why only those are used. I may remove comment from here and only use this PR thread as a reference or add a comment in code

src/rmp/src/genetic_strategy.cpp
b-chmiel: If there is no winner from the tournament, population is empty, then the first candidate is selected

src/rmp/test/gcd_annealing1.ok
b-chmiel: I'd run 100 tests with different seeds for aes_genetic, aes_annealing, gcd_genetic, gcd_annealing1 and gcd_annealing2. Additionally, I'd tested aes_annealing against master version (commit 809a904c3ff84c945ebd200f565344facf1c4f58) and found no observable regressions. Also, genetic algorithm seems to have less spread in terms of slack leading to more consistent results across runs.

For each test, I'd took the best seeds and used them for regression tests.

## AES

### Genetic

```
Summary statistics for 100 runs:

Metric            Min        Max         Mean        Stdev
--------------------------------------------------
SLACK          -44.62      47.23         0.76        22.11
WNS            -44.62       0.00        -8.82        12.95
TNS           -249.63       0.00       -26.03        52.37
```

### Annealing (after genetic changes)

```
Summary statistics for 100 runs:

Metric            Min        Max         Mean        Stdev
--------------------------------------------------
SLACK         -180.42      45.68       -61.80        43.22
WNS           -180.42       0.00       -63.16        40.72
TNS          -1457.49       0.00      -370.12       340.87
```

### Annealing master

```
Summary statistics for 100 runs:

Metric            Min        Max         Mean        Stdev
--------------------------------------------------
SLACK         -191.97      39.35       -66.19        42.96
WNS           -191.97       0.00       -67.16        41.03
TNS          -1697.43       0.00      -408.20       372.12
```

## GCD

### Genetic

```
Summary statistics for 100 runs:

Metric            Min        Max         Mean        Stdev
--------------------------------------------------
SLACK          -94.46     -37.28       -63.27        13.99
WNS            -94.46     -37.28       -63.27        13.99
TNS          -3269.22    -526.75     -1985.41       596.58
```

### Annealing1 (after genetic changes)

```
Summary statistics for 100 runs:

Metric            Min        Max         Mean        Stdev
--------------------------------------------------
SLACK         -597.07     -37.21      -175.06       140.87
WNS           -597.07     -37.21      -175.06       140.87
TNS         -23332.39    -587.16     -6135.38      5437.95
```

### Annealing2 (after genetic changes)

```
Summary statistics for 100 runs:

Metric            Min        Max         Mean        Stdev
--------------------------------------------------
SLACK         -599.23     -40.48      -187.32       141.47
WNS           -599.23     -40.48      -187.32       141.47
TNS         -23001.21    -703.01     -6578.20      5444.68
```

src/rmp/src/slack_tuning_strategy.cpp
b-chmiel: Fixed, thanks.

src/rmp/test/gcd_annealing2.ok
b-chmiel: See https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908#discussion_r2852088048

src/rmp/src/slack_tuning_strategy.h
b-chmiel: Fixed, thanks.

b-chmiel:
@maliberty Resolved all comments. I'd also run experiments with different seeds for genetic and annealing slack tuning: https://github.com/The-OpenROAD-Project/OpenROAD/pull/8908#discussion_r2852088048

maliberty (approved):
Just needs conflicts resolved

b-chmiel:
@maliberty Done

PR #8914, opened 2025-11-26
rsz: clamp max fanout in pre-place gain based buffering

mikesinouye:
If no max_fanout constraint has been applied via SDC and the PDK has no default, sta will return -inf for the max_fanout constraint. This is implicitly converted to int in the following call to `performGainBuffering()`.

In certain environments, the compiler enforces mandatory UBSAN and this conversion out of bounds will cause runtime failures.

### DIFF
--- a/src/rsz/src/RepairDesign.cc
+++ b/src/rsz/src/RepairDesign.cc
@@ -191,7 +191,7 @@ void RepairDesign::performEarlySizingRound(int& repaired_net_count)
 
       bool repaired_net = false;
       if (max_fanout <= 0) {
-        max_fanout = std::numeric_limits<int>::max();
+        max_fanout = 1e9;
       }
 
       if (performGainBuffering(net, drvr_pin, max_fanout)) {


maliberty:
Where does " sta will return -inf for the max_fanout constraint" happen?

mikesinouye:
> Where does " sta will return -inf for the max_fanout constraint" happen?

https://github.com/The-OpenROAD-Project/OpenSTA/blob/eb8ab089dfc2b4643ecea571f329bf97985224f5/search/CheckFanoutLimits.cc#L96

And is never updated as a limit does not exist

maliberty:
I don't think using 0 for max_fanout is a good idea.  It seems we should identify this case and avoid using the limit.

mikesinouye:
Should we set the limit to int max if the sta limit is less than or equal to zero?

maliberty:
> Should we set the limit to int max if the sta limit is less than or equal to zero?

I think that will work

maliberty (approved):

PR #8917, opened 2025-11-27
grt: implement layer ranges for signal and clock in CUGR

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/8398

### DIFF
--- a/src/grt/src/cugr/src/Design.cpp
+++ b/src/grt/src/cugr/src/Design.cpp
@@ -1,6 +1,7 @@
 #include "Design.h"
 
 #include <iostream>
+#include <set>
 #include <vector>
 
 #include "CUGR.h"
@@ -12,7 +13,6 @@
 #include "odb/dbShape.h"
 #include "odb/dbTypes.h"
 #include "odb/geom.h"
-#include "sta/Liberty.hh"
 #include "utl/Logger.h"
 
 namespace grt {
@@ -23,7 +23,7 @@ Design::Design(odb::dbDatabase* db,
                const Constants& constants,
                const int min_routing_layer,
                const int max_routing_layer,
-               std::set<odb::dbNet*> clock_nets)
+               const std::set<odb::dbNet*>& clock_nets)
     : block_(db->getChip()->getBlock()),
       tech_(db->getTech()),
       logger_(logger),
--- a/src/grt/src/cugr/include/CUGR.h
+++ b/src/grt/src/cugr/include/CUGR.h
@@ -2,6 +2,7 @@
 
 #include <csignal>
 #include <memory>
+#include <set>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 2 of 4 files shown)


src/grt/src/cugr/src/Design.cpp
```diff
@@ -289,6 +302,34 @@ void Design::setUnitCosts()
   }
 }
 
+bool Design::isClkTerm(odb::dbITerm* iterm, sta::dbNetwork* network)
```
maliberty: This and others look like they are copied from GlobalRouter.  Is it possible to avoid copying?

eder-matheus: I updated the code to not rely on these functions. Now, it finds the clock nets on GlobalRouter and pass them to CUGR in a set. It uses the set to determine which nets need to have a specific layer range.

I decided to keep the sta class member into CUGR::Design as we'll eventually use it to add timing-driven to the code. Let me know if that's ok or if I should remove.

src/grt/src/cugr/src/Design.cpp
```diff
@@ -16,15 +19,19 @@ namespace grt {
...
+               sta::dbSta* sta,
                const Constants& constants,
                const int min_routing_layer,
-               const int max_routing_layer)
+               const int max_routing_layer,
+               const std::set<odb::dbNet*>& clock_nets)
     : block_(db->getChip()->getBlock()),
       tech_(db->getTech()),
       logger_(logger),
+      sta_(sta),
```
maliberty, line 30: Where is the used?  I see you are passing both sta and clock_nets which seems redundant unless there is another use for sta_.

src/grt/src/cugr/src/Design.cpp
```diff
@@ -130,7 +137,15 @@ void Design::readNetlist()
       pins.emplace_back(pin_count, db_iterm, pin_shapes);
       pin_count++;
     }
-    nets_.emplace_back(net_index, db_net, pins);
+
+    LayerRange layer_range
+        = {.min_layer = min_routing_layer_, .max_layer = max_routing_layer_};
+    if (clock_nets_.find(db_net) != clock_nets_.end()) {
+      layer_range.min_layer = block_->getMinLayerForClock() - 1;
+      layer_range.max_layer = block_->getMaxLayerForClock() - 1;
```
maliberty, line 145: Why the -1 ?

src/grt/src/cugr/src/Design.cpp
eder-matheus: I decided to keep the sta class member into CUGR::Design as we'll eventually use it to add timing-driven to the code. Should I remove it?

src/grt/src/cugr/src/Design.cpp
eder-matheus: The ODB M1 index is 1, while internally CUGR uses 0 as the M1 index.

src/grt/src/cugr/src/Design.cpp
maliberty: Ok to keep so long as there is a (future) use.

PR #8932, opened 2025-12-02
rcx: fix coordinates option in write_spef

shinkuan:
This PR includes fix for `-coordinates` option in `write_spef` command, for detail informations, refer to the issue mentioned in the https://github.com/The-OpenROAD-Project/OpenROAD/discussions/8931

The original `ext.i` was missing the `opt.N` and that causes `Ext::write_spef` never writes coordinate information to spef.
Modify the `ext.i` to have the same behavior as the Python protocol below:
https://github.com/The-OpenROAD-Project/OpenROAD/blob/3f01e9f673aad5eb56ad76acd598c0bb7263274b/src/rcx/test/rcx_aux.py#L56-L64

Edit: force-pushed to add commit sign-off message.


QuantamHD:
Lgtm. Thank you for the fix.

Would you mind adding a regression test for this option to ensure that the behavior does not regress. Examples should be in src/rcx/test

shinkuan:
Sure, the regression tests have been added. Please check.

maliberty (approved):

maliberty:
Thanks!

Issue #8909, opened 2025-11-26
Notches in mpl result for ng45/ariane133


maliberty:
### Describe the bug

There are notches in NW & SE corners:

<img width="495" height="497" alt="Image" src="https://github.com/user-attachments/assets/ef9680fe-4e54-4b26-98c1-086629097ba3" />

### Expected Behavior

No notches

### Environment

```shell
Head of OR (19294a8c77a6663a307fc5622bbbd82bd428cf1c)
ORFS https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3678
```

### To Reproduce

Run the design

### Relevant log output

```shell

```

PR #8934, opened 2025-12-02
gpl: modify instance area based on averaga instances pin density

gudeh:
The motivation on this change is to make overall design pin density more homogeneous, and help with routing congestion. Experimental results showed consistent improvement in TNS on slang for a private design.

We calculate the average instance pin density (`avg_density` variable) by counting all signal pins in movable instances and dividing by the accumulation of the area of all those instances.

The modification in instance size is controlled by a max cap of 120% (increase) and 95% (decrease) for each dimension, and depends if the instance has a higher or lower pin density than the average.

Two reasons for the max cap: 
  - Decreasing the area of instances with low pin density too much led to issues with DPL. I noticed GPL would position some large instances on regions with not enough area to move them around, leading to larger wires in routing and congestion. Furthermore, I suspect we have limited DPL ability to legalize instances when there is not a lot of empty space, mostly on edges and corners.

  - The desired homogeneity of overall design pin density was only fully achieved by forcing all instances to match the average pin density. Although, this stronger modification led to inconsistencies in routability mode later on, due to too much modification in area. The controlled max cap avoids this issue.

### DIFF
--- a/test/upf_aes.ok
+++ b/test/upf_aes.ok
@@ -29,9 +29,9 @@ Using 2 tracks default min distance between IO pins.
 [INFO GPL-0014] Region name: top-level.
 [INFO GPL-0015] Region area:                878902.938 um^2
 [INFO GPL-0017] Fixed instances area:       595260.902 um^2
-[INFO GPL-0018] Movable instances area:     126782.845 um^2
-[INFO GPL-0019] Utilization:                    44.698 %
-[INFO GPL-0020] Standard cells area:        126782.845 um^2
+[INFO GPL-0018] Movable instances area:     143938.422 um^2
+[INFO GPL-0019] Utilization:                    50.747 %
+[INFO GPL-0020] Standard cells area:        143938.422 um^2
 [INFO GPL-0021] Large instances area:            0.000 um^2
 [INFO GPL-0032] Initializing region: PD_AES_1
 [INFO GPL-0006] Number of instances:             16862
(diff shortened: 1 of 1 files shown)


gudeh:
Here are some plots comparing this branch with master for public designs. These give me the impression we are doing a better routability job, with slight increase in wirelength (due to inflations I believe) and less GRT violations.

We get some reduced amount of GRT violations at the start of DRT, and reduced runtime in DRT (3rd and 4th plots):
Routability_metrics

Slight increase in wirelength and slight stable timing:
WL_timing

gudeh:
Yesterday there was a divergence happening in GPL with nangate45/ariane133, leading to a fatal in GRT. I merged the branch with master with a newer version, there is no more GPL divergence, although GRT is still failing with:

```
[WARNING GRT-0704] Try reduce the layer adjustment from 25% to 9%
[ERROR GRT-0116] Global routing finished with congestion. Check the congestion regions in the DRC Viewer.
```

gudeh:
I suspect the issue is due to macro placement, with my new branch we end up using all the empty space available and placing instances on the top left corner of the design, where the GRT issues happen:
image

On master, we place only a few instances on the top left corner:
image

Here is RUDY during stage 3-3, master versus new branch:
imageimage

And here is the pin density with equalized min and max values, master and branch. This shows how we do achieve our goal of a more balanced pin density!
imageimage





Furthermore, I also noticed GPL is currently diverging on the master version (although we are still able to finish). I suspect it is due to the holes in the corner.

maliberty:
Looks like https://github.com/The-OpenROAD-Project/OpenROAD/issues/8909  @joaomai please help out as this is a change I want to get in (workaround or solution).

AcKoucher:
@maliberty @joaomai @gudeh I think that these notches are also a symptom of another behavior: _we're generating shapes during coarse shaping that allow these macro clusters to have empty spaces_ e.g., the array in the top has 44 macros arranged as a 5x9 array. Perhaps it would make more sense for these macros to be arranged as 4x11.

However, I'm not sure that if we generated only shapes without empty space we might end up stumbling into cases in which we can't converge.

maliberty:
I think it is ok to generate a shape with gaps but we shouldn't then place the macros to leave a notch like this.

gudeh:
Maybe we could make the gap to be towards the middle of the design? Or avoid edges/corners?

gudeh:
Error with large tests again, all of them need updating: `ERROR: File src/gpl/test/large02.defok is big but not whitelisted (hash 34176aa8aa0dac9e2ec35dc0adf21111)`

Also this one I have never seen: `unidiff.errors.UnidiffParseError: Hunk is shorter than expected`

maliberty (approved):

PR #8936, opened 2025-12-02
odb: add callback to BPin when adding or removing boxes so the gui and psm get updated

gadfort:
I noticed that when deleting dbBox's from BPins the GUI was not updating. This corrects this to ensure the search is cleaned up. Along with ensure PSM is reset as well.


maliberty:
Please add a test to src/odb/test/cpp/TestCallBacks.cpp

gadfort:
@maliberty tests added

Issue #8698, opened 2025-10-22
rsz: Resizer is non-deterministic causing tests to be flaky


hongted:
### Describe the bug

Resizer.cc runs are non-deterministic.  Running the same test multiple times results in different results.  

One reason is the use of ordered sets of pointers, which need to be sorted again before any iteration where order matters. 

Example:  
https://github.com/The-OpenROAD-Project/OpenROAD/blob/0a9b58c36689b10d3c0db4ce38391e75acddefe9/src/rsz/include/rsz/Resizer.hh#L741
which causes the iteration order here to change  https://github.com/The-OpenROAD-Project/OpenROAD/blob/0a9b58c36689b10d3c0db4ce38391e75acddefe9/src/rsz/src/Resizer.cc#L748

As a separate note, as these ordered sets can't be relied on to be ordered, changing to unordered_set can be considered. 

### Expected Behavior

Resizer code will give the same results when run multiple times.

### Environment

```shell
N/A
```

### To Reproduce

Run tests multiple times to observe the failure rate:

```
bazelisk test -c opt //src/rsz/test/... --cache_test_results=no --runs_per_test=100
```

### Relevant log output

```shell

```

maliberty:
unordered_set  won't help as it will just hash on the pointer value.  A comparison should be provided to any set of pointers that will iterated.

hongted:
That's the point.  The code isn't relying on the sorted property of sets, so an unordered set is no better and can be considered for a performance/memory fragmentation gain.

Sorting or adding a proper compare function would be potential solutions.

povik:
@hongted Please if you have it at hand post which tests are failing other than the fast buffer reporting tests

hongted:
Below is a list from my last run:

```
//src/rsz/test:pin_swap1-tcl_test                                        FAILED in 5 out of 100 in 1.0s
//src/rsz/test:repair_tie11_hier-tcl_test                                FAILED in 2 out of 100 in 0.7s
//src/rsz/test:repair_tie12_hier-tcl_test                                FAILED in 2 out of 100 in 0.7s
//src/rsz/test:resize_slack2-tcl_test                                    FAILED in 1 out of 100 in 1.2s
```

hongted:
Below command can be used to run those tests

```
[...]

PR #8942, opened 2025-12-02
RSZ: Improve resizer determinism

calewis:
- Replace std::set with std::unordered_set for `buffer_fast_sizes_` to remove unnecessary pointer-address sorting.
- Ensure deterministic sorting at call sites where order matters.
- Fix pessimization in BaseMove.cc: check condition before constructing vector.

Addresses issues pointed out in [#8698](https://github.com/The-OpenROAD-Project/OpenROAD/issues/8698), although it's hard to prove that all of them have been fixed.

```
bazelisk test -c opt //src/rsz/test/... --cache_test_results=no --runs_per_test=100
```
yields no failures on my machine, but I didn't go out of my way to try and force allocation order or insertion order changes.

### DIFF
--- a/src/gui/src/search.cpp
+++ b/src/gui/src/search.cpp
@@ -61,6 +61,16 @@ void Search::inDbBPinCreate(odb::dbBPin* pin)
   clearShapes();
 }
 
+void Search::inDbBPinAddBox(odb::dbBox* box)
+{
+  clearShapes();
+}
+
+void Search::inDbBPinRemoveBox(odb::dbBox* box)
+{
+  clearShapes();
+}
+
 void Search::inDbBPinDestroy(odb::dbBPin* pin)
 {
   clearShapes();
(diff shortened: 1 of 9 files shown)


src/rsz/src/BaseMove.cc
```diff
@@ -982,14 +982,25 @@ float BaseMove::computeElmoreSlewFactor(const Pin* output_pin,
...
-      buffer_sizes.push_back(buffer);
-    }
     if (resizer_->buffer_fast_sizes_.count(base) == 0) {
       return LibertyCellSeq();
     }
+
+    LibertyCellSeq buffer_sizes;
+    for (LibertyCell* buffer : resizer_->buffer_fast_sizes_) {
+      buffer_sizes.push_back(buffer);
+    }
```
QuantamHD: Probably should reserve the size?

PR #8949, opened 2025-12-03
CTS: improve automatic clustering

### LINKED ISSUE #8540: CTS: improve automatic clustering
### Description

CTS has an option for an automatic clustering parameters selection (`-sink_clustering_size`, `-sink_clustering_max_diameter`). This automatic selection uses a set of predetermined values, {10, 20, 30} for clustering size and {50, 100, 200} for clustering max diameter.

The values are arbitrary and have no relation to the design or PDK. As discussed breafilly in [issue #8111 ](https://github.com/The-OpenROAD-Project/OpenROAD/issues/8111#issuecomment-3229468213) the selection for the clustering max diameter could be selected based on PDK rows. 

### Suggested Solution

_No response_

### Additional Context

_No response_

luis201420:
Fixes #8540 

It eliminates the need for predefined lists to find the best combination of diameter and cluster size. Both lists are no longer needed since buffer max cap includes them. The iterations that performed this search are also removed.

The following cases will now be considered as follows:
- If the cluster size or diameter is not defined, then the buffer max cap is used to create the clusters.
- If the cluster size is defined, then that size, along with the buffer max cap, will be used to create the clusters.
- If the cluster diameter is defined, then the diameter, along with the buffer max cap, will be used to create the clusters.
- Finally, in case If both values ​​are defined, then they will be used by disabling the buffer max cap.


luis201420:
I ran a CI and all the public designs passed. I am waiting for the results of the private designs.

src/cts/src/SinkClustering.cpp
```diff
@@ -415,7 +421,17 @@ bool SinkClustering::isLimitExceeded(const unsigned size,
                                      const unsigned sizeLimit)
 {
   if (useMaxCapLimit_) {
```
arthurjolo, line 425: I think this if statement can be a bit different now that we define when to use each of the limits.

suggestion:
```

bool is_limit_exceeded = false
if (useMaxCapLimit_) {
  is_limit_exceeded |= (capCost > options_->getSinkBufferInputCap() * max_cap__factor_);
// size is defined by the user
if (use_max_size_) {
  is_limit_exceeded |= (size >= sizeLimit);
}
// diameter is defined by the user
if (use_max_diameter_) {
  is_limit_exceeded |= (cost > maxInternalDiameter_);
}
return is_limit_exceeded;
```


like this we don't need to separate into 2 different cases useMaxCap and not useMaxCap

arthurjolo (approved):
I think the code changes look fine, it is good to leave the list of cluster sizes and diameter behind. Just one small suggestion on the method for defining if the limit is exceeded.

src/cts/src/CtsOptions.h
arthurjolo, line 208: By the comment of the PR we would only want to disable buffer max cap if both cluster size and diameter were set. But here if we set the diameter we immediately disable the useMaxCap_.

src/cts/src/CtsOptions.h
```diff
@@ -232,15 +228,11 @@ class CtsOptions : public odb::dbBlockCallBackObj
   }
```
arthurjolo, line 220: Same here.

maliberty:
Conflicts to resolve.  Have you tested with private designs as well?

luis201420:
> Conflicts to resolve. Have you tested with private designs as well?

Yes, I ran an CI, and in the case of the private designs, I didn't see any differences in the metrics.

maliberty (approved):

PR #8956, opened 2025-12-04
ODB: fix multiple 3dblox issues

osamahammad21:
- Store dbChipInst::origin instead of loc and calculate loc when needed
- Fix dbChipInst::setLoc and dbChipInst::setOrient to always place the chipInst so that it's lower left lower corner is at loc_
- Extend dbTransform class to handle 3D shapes
- Fix 3DBV Reader to disregard optional values that are not defined.
- Correctly calculate the die area and core area for a block of a die chip without an associated block and accounting for the scribe_line and seal_ring
- Fix the placement of bumps read from the bump map by accounting for the chip offset.
- Create an unfolded view of the chiplets in the 3dblox checker which fixed multiple anomalies noticed when running it

### DIFF
--- a/src/odb/src/db/dbChipInst.cpp
+++ b/src/odb/src/db/dbChipInst.cpp
@@ -144,7 +144,6 @@ void dbChipInst::setOrient(dbOrientType3D orient)
   _dbChipInst* obj = (_dbChipInst*) this;
 
   obj->orient_ = orient;
-  setLoc(getLoc());
 }
 
 void dbChipInst::setLoc(const Point3D& loc)


osamahammad21:
@maliberty Could you please review?

src/odb/src/db/dbChipInst.cpp
```diff
@@ -156,9 +136,34 @@ dbChip* dbChipInst::getParentChip() const
...
-                     Point(obj->loc_.x(), obj->loc_.y()));
+  return dbTransform(obj->orient_, obj->origin_);
+}
+
+void dbChipInst::setOrient(dbOrientType3D orient)
+{
+  _dbChipInst* obj = (_dbChipInst*) this;
+
+  obj->orient_ = orient;
+  setLoc(getLoc());
```
maliberty: This seems to be different from how dbInst::setOrient works (just sets the orientation and not the loc).

osamahammad21: The original intent was to make the order of setting the orientation and the location irrelevant to the final result. So calling setOrient followed by setLoc would yield the same outcome as calling setLoc followed by setOrient. I can avoid this entire gray area by introducing a single function that sets both at once.

maliberty: I think it should be the same as dbInst otherwise it is confusing that different types of instances behave differently.

maliberty (approved):

PR #8957, opened 2025-12-04
dbsta: skip supply nets when iterating over pins

gadfort:
Issue:
- if the power or ground nets are connected to an instance where the LEF has not specified the pins as POWER/GROUND, opensta attempts to build a graph with the nets, which causes the memory to balloon. In the case I was running into, the instances were `COVER BUMP` cells, which just mark their pins as `INOUT`.

Notes:
- @maliberty please double check this, I think I found a method to only remove the real power and ground nets while leaving the tieoff nets (which are also marked as power and ground).

### DIFF
--- a/src/dbSta/include/db_sta/dbNetwork.hh
+++ b/src/dbSta/include/db_sta/dbNetwork.hh
@@ -163,7 +163,7 @@ class dbNetwork : public ConcreteNetwork
                          const dbIoType& io_type) const;
 
   bool isPGSupply(dbITerm* iterm) const;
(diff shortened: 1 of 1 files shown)


maliberty:
I started a secure CI

maliberty:
You are relying on `net->isSpecial` to avoid the tie nets?

gadfort:
> You are relying on `net->isSpecial` to avoid the tie nets?

Yes

maliberty (approved):

PR #8963, opened 2025-12-05
pdn: fix working error to ensure stable ordering in write to db

gadfort:
Replaces #8904

I think I found the source of the bazel / cmake issue.
The sort was using the wrong layer number.

DEF files updated to account for the corrected ordering.


gadfort:
@maliberty there is still one test that wasn't fixed with this that I need to work on.

Issue #8838, opened 2025-11-13
gpl: regions with timing-driven


gudeh:
### Description

Currently GPL does not support timing-driven mode when a design has regions. Issue https://github.com/The-OpenROAD-Project/OpenROAD/issues/8768 has an example where this happens.

I understand we can also reproduce the issue with the gpl regresssion test with regions (region01) or the upf test at OpenROAD/test, by including the timing-driven parameter.

### Suggested Solution

We may have two different approaches:

- An easy approach would be to allow for RSZ instances to move freely. I imagine having the new buffer on the top-level and allowing it to move freely along the core area or regions (although currently the top-level sees regions as filled with dummy instances). [I believe this approach would only require to address this comment](https://github.com/The-OpenROAD-Project/OpenROAD/blob/94ab17621aa8150675eee8703607482eab396dac/src/gpl/src/nesterovPlace.cpp#L1217).
- Or we can modify RSZ to identify where an instance was placed, if it is inside a region, set the new instance to be inside the region. GPL callback processing would also need to be updated to account for the region. We have an ODB callback version with a region defined, [but it has no implementation currently in GPL](https://github.com/The-OpenROAD-Project/OpenROAD/blob/94ab17621aa8150675eee8703607482eab396dac/src/gpl/src/nesterovPlace.cpp#L1285).
  - A potential con about this approach is that the RSZ buffer would be constrained to the region, I wonder if we would like for the buffer to move freely on the design.

PR #8965, opened 2025-12-05
gpl: always create new instances from TD mode in top-level

gudeh:
The existing code was incorrect. I remember just throwing a sketch loop to be refined afterwards regarding the region behavior and it was left incorrect.

I just modified it so we always create new instances from callbacks in the top-level region of gpl. This would be the first step mentioned in issue #8838. The actual desired behavior might be to set the actual region a buffer may belong to.


gudeh:
This change should be no-op for designs without regions, I triggered a secure-CI for public designs at: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/test-gpl-buffer-in-top-level/2/pipeline

PR #8968, opened 2025-12-05
RSZ: Remove non-determinism and platform-specific behavior

calewis:
Fix non-determinism in Resizer that caused flaky tests. This enables pin_swap1 and resize_slacks2 to run by default in Bazel tests.

Note: Some tests remain flaky if `write_verilog` is called without `--sort`. This is deferred to an upstream STA update that will enforce deterministic sorting by default.


calewis:
Looks like clang-format on ranges stuff might change between versions, my file is formatted, by my clang-format.

QuantamHD:
I've run into this before. I just downloaded the clang-format version that's in the CI, and used that one. 18.1.3

calewis:
> I've run into this before. I just downloaded the clang-format version that's in the CI, and used that one. 18.1.3

Was able to use `apt install clang-format-18` and that one matches well enough.

src/dbSta/src/dbNetwork.cc
```diff
@@ -447,6 +447,10 @@ DbInstanceNetIterator::DbInstanceNetIterator(const Instance* instance,
         }
       }
       flat_nets_vec_.assign(flat_nets_set.begin(), flat_nets_set.end());
+      // Sort these nets so their order is determinisitc
+      std::ranges::sort(flat_nets_vec_, [](auto const* net1, auto const* net2) {
+        return net1->getId() < net2->getId();
+      });
```
maliberty: This solves this one problem and leaves us vulnerable to this everywhere else.  I think we should approach std::ranges::less like we did std::less instead.

oharboe:
> I've run into this before. I just downloaded the clang-format version that's in the CI, and used that one. 18.1.3

Add a `bazelisk run //:format` target to insulate ourselves from locally installed tools?

This would autoformat

- python
- tcl
- bazel
- C++

maliberty:
That sounds useful

src/dbSta/src/dbNetwork.cc
calewis: I would prefer to define a functor with the pointers that you can drop in when you want.  Alternatively I could swap to `std::sort`, which is what I used originally, but clang-tidy complained that we need to use ranges now so I switched to ranges. 

I'd prefer not to define a bunch of `std::ranges::less` specializations in this PR if at all possible. 

I'll see if I can use a projection to make it work with ranges without needing to provide a `std::ranges::less` overload, other wise I'll try to modify the clang-tidy runs to avoid telling people to swap to `std::ranges::sort`.

calewis: @maliberty please take a look.

maliberty: I don't think we should tell people  not to use ranges.  I don't see any reason not to do the same as we did for the non-ranges code.  Do you prefer not to do it in _this_ PR or at all?

calewis: I'd prefer not to do it at all, but I won't stop someone else from doing it or from doing it here if you tell me that is the only way you'll accept this pr.  BTW the code now uses regular unspecialized std::ranges::less, by providing a projector.

maliberty: If you think we shouldn't do that then suggest what we should do instead.  Relying on everyone to use some functor without error is not a better solution imho.

calewis: Well I think we should use something like the `ValueWrappedPtr` (name can change) that I pointed out already.  I think that specializing `std::ranges::less` for pointer types is a big violation of the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). 

As an outsider how could I possibly suspect that arbitrary pointer types won't sort by address for `std::ranges::sort(vec_of_ptrs)` when that's how every other container of pointers I've ever seen sorts?  And within OR how can I remember which class hierarchies have specialized `std::less` and which haven't?

If you want to treat class hierarchies of pointers as values, making a value wrapper seems straight forward and not surprising to anyone.

calewis: @maliberty here is a simple example showing how depending on a specialization of std::less can lead to dangerous outcomes.  There are both logic errors and ODR violations here, and given how we use a lot of forward declarations, it wouldn't even be that surprising if this happened.  https://godbolt.org/z/c8n6x1of4

maliberty: I don't follow your example.  There is no definition of std::less at all nor any main that would make it executable.

calewis: That's the entire point!  You don't need to be able to see the specialization of `std::less` to compile and run your code, that's the error/issue.  Having a main has nothing to do with it.  The only thing that matters is that different TUs have different definitions of `std::less<dbNet*>`, which leads to ODR violations. Even if the linker deals with that by letting you have two definitions, you get different sort orders in different TUs.

maliberty: I don't see how your example shows that.  Let me see if I understand your point

If you only forward declare dbNet and use it in a container then you would not see the std::less specialization.   The resulting code would have the std::less<dbNet*> default comparison in that case.   Other code might see the OR specialization and fail ODR.

I think that is theoretically possible though practically unlikely.  @hzeller push to never forward declare would make it impossible.

calewis: You do understand the problem.  This being theoretically possible, is just one reason not to depend on specializations of std::less for pointers.  That said, I'm not trying to remove them in this PR, I just don't want to add those specializations to std::ranges::less, because I think it's dangerous.

calewis: @maliberty I don't think your current std::less specializations are necessarily working as intended before this PR.   While trying to actually generate an ODR violation with the above code I discovered the following:
https://godbolt.org/z/8czo9saqs appears to be doing what you want, except if you run it in Google's env the output becomes:
```
Mi: 127352220737536 -> 4
Mi: 127352220737544 -> 12
Mi: 127352220737552 -> 11
Mi: 127352220737560 -> 10
Mi: 127352220737568 -> 9
Mi: 127352220737576 -> 8
Mi: 127352220737584 -> 7
Mi: 127352220737592 -> 6
Mi: 127352220737600 -> 5
Mi: 127352220739640 -> 13
Mi: 140729647569640 -> 3
Mi: 140729647569648 -> 2
Mi: 140729647569656 -> 1
Mi: 140729647569664 -> 0
```
And just to make sure this isn't some Googlism if you run the same code on your Mac (I think you said you have a mac so it should be easy to reproduce this result):
```
mac > clang++ --std=c++20 test.cpp
mac >  ~/s/small_stuff> ./a.out
Mi: 6155530584 -> 3
Mi: 6155530592 -> 2
Mi: 6155530600 -> 1
Mi: 6155530608 -> 0
Mi: 105553175183424 -> 4
Mi: 105553175183440 -> 5
Mi: 105553175183456 -> 6
Mi: 105553175183472 -> 7
Mi: 105553175183488 -> 8
Mi: 105553175183504 -> 9
Mi: 105553175183520 -> 10
Mi: 105553175183536 -> 11
Mi: 105553175183552 -> 12
Mi: 105553175183568 -> 13
```

This comes down to how `std::sort` is implemented in libc++ and is not something we can change. 

In libc++ `std::sort` without a third argument will use `<` to do the sorting and we can't change that or overload `<` for pointers.  

With this discovery I don't think there is anything reasonable I can do besides provide either a compare functor or a projector for ranges::sort.

hzeller: I do not like the `std::less` specializations on the pointers used in OpenRoad at all, it is a very dangerous construct and is a huge surprise once found out (I remember when I stumbled upon that).
(It is one of the things that technically _is_ possible in c++, but should not be done :))

This relies on the includes being included properly, but if not, there is no indication of failure. For my header cleanups, I am using the compiler to help me find missing includes, but this is a case where not including the specializations will also be a perfectly valid compilation but with unintended semantic consequences; the developer always needs to be on their toes to make sure the dependency is there.

TL;DR: I think we should _not_ have `std::less` specializations in this way, and code should be refactored to not need it (e.g. by having a wrapper object).

So for new code, using an explicit comparison, as @calewis is doing, looks like a fine way of avoiding that confusion and a first step in the right direction.

oharboe: [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

calewis:
@maliberty what do I need to do to finish this up?

maliberty (approved):

maliberty:
The code looks ok but I'm going to start a secure-CI run to ensure there is no QOR impact

Issue #8871, opened 2025-11-19
mpl should have a timing-driven cost component


AcKoucher:
### Description

The investigation in [#3402](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3402) showed that we require some sort of component in the cost function that looks at timing in order to improve correlation.

### Suggested Solution

Use wire load models and create another layer of connectivity between clusters that involves only the critical nets.

PR #8969, opened 2025-12-05
mpl: refactor to avoid future duplicated code

AcKoucher:
For #8871.

There will be another penalty computation based on WL.


AcKoucher:
Running Secure-CI.

PR #8970, opened 2025-12-06
dpl: Boost Geometry IWYU

QuantamHD:
geometry.hpp includes a bunch of other unneeded headers that interact poorly with Google's no exception policy. This includes only the subset needed for OpenDP

### DIFF
--- a/src/dpl/include/dpl/Opendp.h
+++ b/src/dpl/include/dpl/Opendp.h
@@ -15,14 +15,12 @@
 #include <utility>  // pair
 #include <vector>
 
-// clang-format off
-#include "boost/geometry/geometries/point.hpp"
+#include "boost/geometry/core/cs.hpp"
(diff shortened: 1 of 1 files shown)


src/dpl/include/dpl/Opendp.h
```diff
@@ -15,7 +15,13 @@
 #include <utility>  // pair
 #include <vector>
 
-#include "boost/geometry/geometry.hpp"
+// clang-format off
```
maliberty: Turning off format doesn't affect tidy which still warns (and I don't see a reason not to format).

QuantamHD: Turned off tidy on the strategies line which is implicitly needed by rtree with including geometry.h

maliberty (approved):

PR #8979, opened 2025-12-08
mpl: refactor to avoid future duplicated code

AcKoucher:
For #8871.

Another wire length-based penalty will be added and the changes here are to make that easier.

I'm using `double` when computing WL for clusters of unplaced IOs to preserve identical results (using `float` causes some very subtle approximation changes that causes interference). Ideally we should probably have the annealer wire length as `int64_t` so that we get rid of this kind of problem.


AcKoucher:
Secure-CI is fine.

PR #8987, opened 2025-12-09
grt: improve capacity calculation

eder-matheus:
Instead of a pessimistic approximation, GRT now computes the capacity on each GCell based on the actual number of tracks that crosses the GCell rectangle. With that, more accurate capacities are computed, preventing losing usable tracks and helping with congestion.


src/grt/src/GlobalRouter.cpp
```diff
@@ -698,26 +693,111 @@ void GlobalRouter::checkAdjacentLayersDirection(int min_routing_layer,
...
+                                                                    track_init,
+                                                                    track_pitch,
+                                                                    track_count,
+                                                                    horizontal)
+                                             : 0;
+          min_cap = std::min(min_cap, cap);
+          fastroute_->setEdgeCapacity(x - 1, y - 1, x, y - 1, layer, cap);
+        }
+      }
+      min_cap = min_cap == std::numeric_limits<int>::max() ? 0 : min_cap;
```
maliberty, line 727: The only way I could see min_cap == std::numeric_limits<int>::max() is if x_grids==0 && y_grids==0.  Could that happen?

maliberty (approved):

src/grt/src/GlobalRouter.cpp
eder-matheus: It actually happens on single-gcell designs, where x_grids_ == 1 and y_grids_ == 1. I made this for a specific unit test we have in GRT, but it should not be common in production designs.

PR #8996, opened 2025-12-09
grt: Add use of DRT access points to CUGR

### LINKED ISSUE #8824: GRT: extra guides covering IO pin when using CUGR
### Describe the bug

When running `global_route -use_cugr`, the IO pins are covered by route guides, but a guide on a layer above the pin layer is also created.

### Expected Behavior

Guides should be cover all pins, but avoid adding unnecessary extra guides.

### Environment

```shell
N/A
```

### To Reproduce

Under `src/grt/test`, run `openroad clock_route_cugr.tcl`. The single IO pin of the design is on met3, but the guides go up to met4, as shown below.

### Relevant log output

```shell

```

### Screenshots

Image

### Additional Context

_No response_

### LINKED ISSUE #8826: GRT: CUGR should use the access points generated by DRT when available
### Description

CUGR has its own methods to compute the access points of IO pins and instance pins. This can lead to mismatches between the global router and the detailed router.

### Suggested Solution

Update CUGR to make it use the DRT access points that are stored in the ODB. The internal calculation should only be used for pins that don't have any APs stored in ODB.

### Additional Context

FastRoute has a similar approach for this problem. The following code can be used as reference:
- Computing positions on grid: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/grt/src/GlobalRouter.cpp#L995-L1050
- Searching for APs in ODB: https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/grt/src/GlobalRouter.cpp#L949-L993

rafaelmoresco:
CUGR uses the preferred access points generated by DRT when abailable in ODB. When not avaliable, CUGR does the usual calculations to find its own access points.

closes #8826 
closes #8824

### DIFF
--- a/src/grt/src/cugr/src/GridGraph.h
+++ b/src/grt/src/cugr/src/GridGraph.h
@@ -13,6 +13,7 @@
 #include "GRTree.h"
 #include "Layers.h"
 #include "geo.h"
+#include "odb/db.h"
 #include "robin_hood.h"
 
 namespace grt {


src/grt/src/cugr/src/GridGraph.cpp
```diff
@@ -365,6 +367,65 @@ CostT GridGraph::getViaCost(const int layer_index, const PointT loc) const
   return cost;
 }
 
+bool GridGraph::findODBAccessPoints(
+    const GRNet* net,
+    AccessPointSet& selected_access_points) const
+{
+  std::vector<odb::dbAccessPoint*> access_points;
+  int amount_per_x = design_->getDieRegion().hx() / x_size_;
+  int amount_per_y = design_->getDieRegion().hy() / y_size_;
```
eder-matheus: Make them `const int`

src/grt/src/cugr/src/GridGraph.cpp
```diff
@@ -373,61 +434,65 @@ AccessPointSet GridGraph::selectAccessPoints(const GRNet* net) const
...
-      if (point.getLayerIdx() >= constants_.min_routing_layer) {
-        const int direction = getLayerDirection(point.getLayerIdx());
-        accessibility
-            += getEdge(point.getLayerIdx(), point.x(), point.y()).capacity >= 1;
-        if (point[direction] > 0) {
-          auto lower = point;
-          lower[direction] -= 1;
+  if (findODBAccessPoints(net, selected_access_points)) {
+    // Skips calculations if DRT already created APs in ODB
+  } else {
```
eder-matheus: This could be simplified to:
```
// Skips calculations if DRT already created APs in ODB
if (!findODBAccessPoints(net, selected_access_points)) {
...
}
```

eder-matheus:
@rafaelmoresco It seems that some clang-tidy requestes were not addressed. Please take a look at it.

src/grt/src/cugr/src/GridGraph.cpp
```diff
@@ -365,6 +367,82 @@ CostT GridGraph::getViaCost(const int layer_index, const PointT loc) const
...
+                               : point.getY() / amount_per_y;
+          const PointT selected_point = PointT(ap_x, ap_y);
+          const int num_layer
+              = ((layer->getNumber() - 2) > (getNumLayers() - 1))
+                    ? getNumLayers() - 1
+                    : layer->getNumber() - 2;
+          const IntervalT selected_layer = IntervalT(num_layer);
+          const AccessPoint ap_new{.point = selected_point,
+                                   .layers = selected_layer};
+          selected_access_points.emplace(ap_new);
```
eder-matheus: This block of code seems to be duplicated below. Could you refactor it into a new function to prevent copying code?

eder-matheus (changes requested):

PR #9014, opened 2025-12-11
odb: remove leaks of name_ in dbModule and dbChip

mikesinouye:
These may have been inadvertently removed in https://github.com/The-OpenROAD-Project/OpenROAD/commit/1a67f34ba6a04154fc19d1b9c054f069f1c92fac#diff-4d65abd71d0a33724e0a9859253b3d5c3a74e5966fb3f6a2a24020c1b3ef757eL132.


maliberty:
Thanks for point this out.  The problem is in the code generator so I've fixed it in #9015

PR #9018, opened 2025-12-11
gui: default -key to most recent and add additional error checks for array out of bounds

gadfort:
This makes the change allow users to create more than one gif serially without needing to keep track of the `-key` param (compatible with the old behavior).

### DIFF
--- a/src/gui/src/gui.cpp
+++ b/src/gui/src/gui.cpp
@@ -1470,8 +1470,7 @@ void Gui::gifAddFrame(std::optional<int> key,
   if (!key.has_value()) {
     key = gifs_.size() - 1;
   }
-  if (gifs_.empty() || *key < 0 || *key >= gifs_.size()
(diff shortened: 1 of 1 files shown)


src/gui/src/gui.cpp
```diff
@@ -1549,14 +1553,18 @@ void Gui::gifAddFrame(const int key,
...
 }
 
-void Gui::gifEnd(const int key)
+void Gui::gifEnd(std::optional<int> key)
 {
-  if (key >= gifs_.size() || gifs_[key] == nullptr) {
+  if (!key.has_value()) {
+    key = gifs_.size() - 1;
+  }
+  if (gifs_.empty() || *key < 0 || *key >= gifs_.size()
```
maliberty: if gifs_.empty() is redundant to the two checks that follow (ie key < 0 || key >= 0 is always true).  Two places (end & add)

PR #9020, opened 2025-12-11
pdn: add option to allow stripes outside core area and add nets to report

gadfort:
Changes:
- adds option to allow stripes outside the core area + one test.
- adds net information to stripes in report if nets are not the same as the grid.

This is related to the other PR: https://github.com/The-OpenROAD-Project/OpenROAD/pull/9010


src/pdn/src/straps.cpp
```diff
@@ -98,7 +98,12 @@ void Straps::checkLayerSpecifications() const
 bool Straps::checkLayerOffsetSpecification(bool error) const
 {
   const int strap_width = getStrapGroupWidth();
-  const odb::Rect grid_area = getGrid()->getDomainArea();
+  odb::Rect grid_area = getGrid()->getDomainArea();
+  if (allow_out_of_core_) {
+    const odb::Rect die = getGrid()->getBlock()->getDieArea();
```
maliberty, line 103: So the strap will go to the die edge?  Won't that conflict with the pads?

gadfort: It could, in the case where I have pads, I'm also using the -number_of_straps option to control how many I get (avoiding the pads).

maliberty: Would it be better to automate that?  It feel error prone

gadfort: What do you mean automate? The pads themselves will block straps that overlap them (because of the obstructions in them).

gadfort: An alternative would be to require the use of number_of_straps as well

maliberty: If the pads will block the straps then why do you need to specify number_of_straps ?

gadfort: well, on the right side of the chip I don't because it would just be the pads and and then the edge. On the left hand side, I only want straps near the pads and not throughout the core

PR #9021, opened 2025-12-12
gpl: debug changes, new routability plot

gudeh:
Functional changes (no-op):
  -  New TCL parameter for user setting of routability snapshot saving overflow `-routability_snapshot_overflow 0.6.` The settling of the revert threshold was already available for the user via `-routability_check_overflow 0.3`. Now user has total control of begin and end of routability mode.

Debug and graphics changes:
  - Show all GPL plots if gui active. No more need for `set_debug_level GPL debugPlot 1`.
  - New GIF for routability reverts (each image is each time we reach 0.3 overflow, the moment we recalculate rudy):
routability

  - A new plot for routability data, sky130hd/aes: 
image

  - Now you can use `-start_rudy X` and `-rudy_stride Y` to update RUDY starting at X and every Y iteration, for example, using `global_placement_debug -start_rudy 1000 -rudy_stride 1`
image

### DIFF
--- a/src/gpl/src/routeBase.cpp
+++ b/src/gpl/src/routeBase.cpp
@@ -10,7 +10,6 @@
 #include <functional>
 #include <limits>
 #include <memory>
-#include <ranges>
 #include <string>
 #include <utility>
 #include <vector>


gudeh:
There is a gui warning about inf I need to fix.

src/gpl/README.md
```diff
@@ -85,12 +85,14 @@ global_placement
...
     [-routability_use_grt]
     [-routability_target_rc_metric routability_target_rc_metric]
+    [-routability_snapshot_overflow routability_snapshot_overflow]
     [-routability_check_overflow routability_check_overflow]
     [-routability_max_density routability_max_density]
     [-routability_max_inflation_iter routability_max_inflation_iter]    
     [-routability_inflation_ratio_coef routability_inflation_ratio_coef]
     [-routability_max_inflation_ratio routability_max_inflation_ratio]
     [-routability_rc_coefficients routability_rc_coefficients]
+    [-routability_snapshot_overflow routability_snapshot_overflow]
```
maliberty: Why does routability_snapshot_overflow  appear twice?

gudeh:
I still need to make some modifications:
- recalculating rudy every iteration on larger designs made gpl super slow. Don't set this as default, maybe a new debug parameter setting a stride (run rudy every X iterations).
- fix inf warning from gui.
- reset overflowed tiles counting (test error with upf).
- fix new snapshot parameter appearing twice.
- remove unused variable `routabilityMaxInflationIter`.

src/gpl/src/nesterovBase.cpp
```diff
@@ -2576,6 +2577,10 @@ void NesterovBase::initDensity1()
 
   initCoordi_.resize(gCellSize, FloatPoint());
 
+  snapshotCoordi_.resize(gCellSize, FloatPoint());
+  snapshotSLPCoordi_.resize(gCellSize, FloatPoint());
+  snapshotSLPSumGrads_.resize(gCellSize, FloatPoint());
```
maliberty, line 2584: This seems unrelated to debugging.  What is the motivation?

src/gpl/src/nesterovPlace.cpp
```diff
@@ -530,6 +545,13 @@ void NesterovPlace::runTimingDriven(int iter,
         nesterov->checkConsistency();
       }
 
+      // Recalculate RUDY after non-virtual TD
+      // TODO not sure to make this default behavior
+      // if (npVars_.routability_driven_mode) {
+      //   rb_->calculateRudyTiles();
+      //   rb_->updateRudyAverage();
```
maliberty: any reason to keep?

src/gpl/src/routeBase.cpp
```diff
@@ -335,7 +330,7 @@ std::vector<int64_t> RouteBase::inflatedAreaDelta() const
 
 int RouteBase::numCall() const
 {
-  return numCall_;
+  return revert_count_;
```
maliberty: How does revert count relate to numCall?

src/gpl/src/routeBase.cpp
```diff
@@ -819,38 +800,41 @@ std::pair<bool, bool> RouteBase::routability(
...
-        (static_cast<double>(overflowTileCnt) / tg_->tiles().size()) * 100);
+        overflowed_tiles_count_,
+        (static_cast<double>(overflowed_tiles_count_) / tg_->tiles().size())
+            * 100);
   }
 
-  int arraySize = edgeCongArray.size();
-  std::sort(edgeCongArray.rbegin(), edgeCongArray.rend());
+  int arraySize = edge_cong_array.size();
+  std::ranges::sort(std::ranges::reverse_view(edge_cong_array),
```
maliberty: why the reverse_view?

src/gpl/src/nesterovBase.cpp
gudeh: This is just to be safe. 

I made this change when trying the linear routability, and I would get seg faults when trying a routability iteration before 0.6 which is the snapshot saving. 

Either way, I believe we should keep all parallel vectors with the same size, and this guarantees we initiate them as so.

src/gpl/src/routeBase.cpp
gudeh: the previous numCall_ is bad a name IMO, it refers to the routability number of calls, which should be the same as the number of reverts we do in routability mode.

src/gpl/src/nesterovPlace.cpp
gudeh: no, I will remove this.

src/gpl/src/routeBase.cpp
maliberty: Then please rename the method as well

gudeh: oh wow! indeed, I missed that!

PR #9022, opened 2025-12-12
mpl: initialize global fence in cpp to allow changing ORFS defaults

AcKoucher:
This is temporary and will be corrected subsequently by #8950.


AcKoucher:
I ran a Test-CI just to ensure nothing would break and it's fine.

PR #9028, opened 2025-12-12
Some fixes for python libraries bindings when building in bazel

fgaray:


fgaray:
Hi @maliberty, 

I am not sure how to propertly test this change so feel free to reject this PR. This is needed for our internal build system.

Thanks!

maliberty:
why is it bazel specific?

fgaray:
When building OpenROAD as a monolithic python module (which is done in the Bazel build via OpenRoad-py.i), multiple interface files are combined. Since std::vector<int> can only be instantiated once per module (or requires careful handling of names to avoid conflicts), the explicit instantiation in PdnGen-py.i (%template(split_cuts_stuff) std::vector<int>;) causes a collision with the existing vector_int definition provided by TritonRoute or other modules included in the same build.

However, I think I can do this better with some header guards (#ifndef , #define). I updated this PR with the new version.

maliberty (approved):

PR #9031, opened 2025-12-13
gui: add two width spacing table to tech layer descriptor

gadfort:
Adds:
- two width table to techlayer descriptor.

### DIFF
--- a/src/gui/src/dbDescriptors.cpp
+++ b/src/gui/src/dbDescriptors.cpp
@@ -36,7 +36,6 @@
 #include "odb/dbTypes.h"
 #include "odb/dbWireGraph.h"
 #include "odb/geom.h"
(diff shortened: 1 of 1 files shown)


src/gui/src/dbDescriptors.cpp
```diff
@@ -2946,6 +2947,39 @@ Descriptor::Properties DbTechLayerDescriptor::getDBProperties(
     props.push_back({std::move(title), widths});
   }
 
+  if (layer->hasTwoWidthsSpacingRules()) {
+    std::vector<std::vector<odb::uint>> spacing_table;
```
maliberty: unused?

gadfort: yes, that was left over. Removed it.

PR #9032, opened 2025-12-13
gui: add wire/swire/sbox descriptors

gadfort:
Adds:
- Descriptors for SBox, SWire, and Wire.

Note:
- I did not implement the slow highlight for SWire, to mirror the behavior of dbNet. @maliberty if you think it should be there, I can add that.


maliberty:
Can you click on a wire to select it?  
How does it scale for large pdn nets?

gadfort:
> Can you click on a wire to select it? How does it scale for large pdn nets?

Yes, the wire is accessible via the dbNetDescriptor.
There is a cap on the number of sboxes (similar to the cap on iterms), to avoid large power and ground nets from slowing down the inspector.

PR #9038, opened 2025-12-14
DPL Two-Pass with toggle

PrecisEDAnon:
The following content was AI generated.

# Two-Pass Congestion-Aware Global Swap (DPL) — Optional “Extra DPL” Path

## Summary
This PR adds an optional alternative implementation of the `gs` (global swap) phase used by `improve_placement` (DPO / detailed placement optimization). The new path is designed to help **high core-utilization** designs where routability-driven metrics (e.g., DRWL) become limiting.

**Expected outcome (high-util targets):** ~2–3% DRWL reduction near the highest utilization where the design still completes the full flow; outliers up to ~8% were observed.

## Motivation / When It Helps
At high utilization (typically ~85%+ and/or “near max util that still converges”), a pure HPWL-driven global swap can “clump” cells into locally congested regions. This often harms routability and increases DRWL. The new algorithm explicitly trades bounded HPWL for congestion relief.

## What This PR Changes
- **New algorithm:** a *two-pass* congestion-aware global swap (“profile then optimize”).
  - **Pass 1 (profiling):** run HPWL-only optimization to estimate a near-optimal HPWL target.
  - **Rollback:** restore the exact initial placement using Journal undo.
  - **Pass 2 (guarded optimization):** re-run global swap using a congestion-aware objective while enforcing an HPWL budget derived from Pass 1.
- **Runtime selection (no rebuild needed):** legacy vs. new global swap is chosen at runtime via a toggle.
- **Tcl configurability:** key hyperparameters are set via Tcl (`improve_placement -global_swap_args {...}`), not compile-time.

## Usage / Toggle Behavior
This PR is intentionally modular:
- If your flow **does not call** `improve_placement`, nothing changes.
- If `improve_placement` runs but the toggle is **off**, behavior matches the legacy global swap.
- If `improve_placement` runs and the toggle is **on**, the new two-pass global swap runs for the `gs` phase.

### Enable / Disable
Enable per call:
```tcl
improve_placement -enable_extra_dpl 1
```

Or enable via environment (useful for flows/CI harnesses):
```tcl
set ::env(ENABLE_EXTRA_DPL) 1
```

Disable:
```tcl
improve_placement -enable_extra_dpl 0
```

## Hyperparameters (Tcl)
Hyperparameters are set through `-global_swap_args` as key/value pairs. They are **reset each `improve_placement` call**, so flows can A/B test cleanly.

Example:
```tcl
improve_placement \
  -enable_extra_dpl 1 \
  -global_swap_args {
    -tradeoff 0.4
    -profiling_excess 1.10
    -budget_multipliers {1.50 1.25 1.10 1.04}
    -area_weight 0.4
    -pin_weight 0.6
    -congestion_user_weight 35.0
    -sampling_moves 150
    -normalization_interval 1000
  }
```

### Recommended tuning choices (2–4 each)
Tune on any circuit that struggles at high utilization near its “max util that still completes”.

| Knob | Meaning | Suggested values |
|---|---|---|
| `-tradeoff` | Exploration vs. wirelength-optimal moves | `0.2`, `0.4`, `0.6` |
| `-profiling_excess` | HPWL budget slack from Pass 1 | `1.05`, `1.10`, `1.15` |
| `-budget_multipliers` | Budget schedule across stages | `{1.30 1.15 1.07}`, `{1.50 1.25 1.10 1.04}`, `{1.20 1.10 1.05}` |
| `-area_weight` | Area term in utilization map | `0.3`, `0.4`, `0.5` |
| `-pin_weight` | Pin term in utilization map | `0.5`, `0.6`, `0.7` |
| `-congestion_user_weight` | Global scaling for congestion term | `20.0`, `35.0`, `50.0` |
| `-sampling_moves` | Samples for adaptive weighting | `50`, `150`, `300` |
| `-normalization_interval` | Utilization renormalization cadence | `500`, `1000`, `2000` |

## Validation Recommendation
1. Pick a circuit that becomes routing-limited at high utilization.
2. Push utilization to the **highest value that still completes** (or just below).
3. Run baseline vs. new path under identical conditions (seed, constraints, etc.).
4. Sweep a small subset of the tuning grid above and compare DRWL; a consistent ~2–3% DRWL reduction is the expected outcome in this regime.

## CI Note
For regression detection, it is recommended to run CI with the alternate path **enabled** (so the new code path is exercised). After merge, this feature should remain available as an alternate setting with **default = off** to preserve legacy behavior unless explicitly enabled.

## Related Work

Private communication.


PrecisEDAnon:
@maliberty - PR now shows merge CI pass, and the linting should all be fixed as well. The head shows bazel weirdness, but since this resolves post-merge, it should be alright (no changes were made on the bazel end, and the tests pass locally, so I would guess this is a case of uploading the head that resolves post-merge).  Please re-review and confirm if OK for merge.

maliberty:
Linting / formatting is not resolved.  Bazel is likely a known issue so I won't worry about that.
image

PrecisEDAnon:
Fixed on the flagged files, and re-pushed.

PrecisEDAnon:
@maliberty Let me know if we can run the lint checks.

maliberty:
As this is off by default I'll go ahead and merge it.  It needs a bit of metrics work to actually enable in ORFS which will be scheduled later.

PR #9039, opened 2025-12-15
test/orfs: sv-bugpoint instructions

oharboe:
Tried these instructions on gcd and got https://github.com/YosysHQ/eqy/issues/88#issuecomment-3656786488


maliberty (approved):
Nice.  bugpoint seems similar to the original https://github.com/dsw/delta which was for c++

PR #9044, opened 2025-12-15
Cts update techchar debugprints

arthurjolo:
Update debug print messages for tech char to give better descriptions of the steps.

### DIFF
--- a/src/cts/src/TechChar.cpp
+++ b/src/cts/src/TechChar.cpp
@@ -2,7 +2,6 @@
 // Copyright (c) 2019-2025, The OpenROAD Authors
 
 #include "TechChar.h"
-#include <spdlog/spdlog.h>
 
 #include <algorithm>
 #include <bitset>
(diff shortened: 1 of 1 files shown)


src/cts/src/TechChar.cpp
```diff
@@ -1041,6 +1045,11 @@ std::vector<TechChar::SolutionData> TechChar::createPatterns(
     odb::dbBPin* outPortPin = odb::dbBPin::create(outPort);
     // Updates the topology with the output port, old new, possible instances
     // and other attributes.
+    for (int i = 0; i < nodesWithoutBuf; i++) {
+      tmp << "-- ";
+    }
+    tmp << "OUT";
+    debugPrint(logger_, CTS, "tech char", 1, tmp.str());
```
maliberty, line 1052: What are you building up in tmp? It seems to be `--` for nodes without buffers.   Is that what it showing?

maliberty (approved):

src/cts/src/TechChar.cpp
arthurjolo: Yes, I use `--` node that are just wires, and `->` for nodes with buffers.

PR #9049, opened 2025-12-16
Fix ambiguity in inspector.cpp.

hzeller:
`makeItem()` exists for `const QString &` and `const std::any &` which resulted in the compiler being unsure what to call.


hzeller:
FYI @gadfort as you modified this file last.

PR #9051, opened 2025-12-16
gui: draw bbox when highlight is smaller than coarse limit (10px)

gadfort:
Changes:
- when animating the highlight, this changes the drawing to a rendering of the bounding box if the bounding box is less than 10px. Otherwise the original highlighting is called.

Notes:
- I chose 10px since it's the coarse resolution we use elsewhere, but it would be easy to modify.


maliberty:
@precisionmoon this should help make seeing instTerm highlights easier

maliberty (approved):

PR #9057, opened 2025-12-16
DRT: set unidirectional layers in router_cfg_

osamahammad21:


maliberty:
What motivates this change?  More of a description would be helpful

osamahammad21:
> What motivates this change? More of a description would be helpful

ETHZ observed that DRT does not respond correctly to the `set_unidirectional_layer` command. While debugging the flow, I found that when `set_unidirectional_layer` is followed by `pin_access`, the effect of the former is effectively nullified. This happens because the pin access flow begins by calling `clearDesign`, which clears the stored unidirectional layers that live in the tech object. That’s why I moved the unidirectional layer settings into `router_cfg_`.

maliberty:
I wonder what else might be cleared in that call.

PR #9077, opened 2025-12-17
mpl: properly handle fixed macros outside the macro placement area

### LINKED ISSUE #9060: MPL fails to take into account fixed macros without pins
### Description

For this situation, overlap is generated, because MPL ignores those cells.

### Suggested Solution

Take macros without pins into account correctly.

### Additional Context

_No response_

AcKoucher:
Resolve #9060.

Before these changes, we would ignore every macro without pins in order to cover some special cases with fixed macros without pins that are outside of the core. However, now, we'll check if there are fixed macros outside the macro placement area and store those in a set so that they can be properly ignored. With this, we don't need to care about the pins.

I included some new messages that I think will make things a bit clearer as well.

### DIFF
--- a/src/mpl/test/BUILD
+++ b/src/mpl/test/BUILD
@@ -105,11 +105,11 @@ filegroup(
         "testcases/macro_only.lef",
         "testcases/macro_only.lib",
         "testcases/macro_only.v",
+        "testcases/macros_without_pins1.def",
         "testcases/mixed_ios1.def",
         "testcases/no_unfixed_macros.def",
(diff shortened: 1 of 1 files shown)


AcKoucher:
@gadfort Could you please check if the changes here work for you?

AcKoucher:
Running Secure-CI.

gadfort:
@AcKoucher I'm getting:
```[ERROR MPL-0003] There are no valid tilings for mixed cluster: root```

PR #9083, opened 2025-12-18
pad: add DB property to enable/disable RDL routing to specific iterms

gadfort:
Adds:
- additional controls for RDL routing. By default RDL routing will attempt to connect all iterms, but that may not be needed in all cases. This adds the ability to tell the router that some iterms can be ignored.

### DIFF
--- a/src/pad/src/RDLRouter.h
+++ b/src/pad/src/RDLRouter.h
@@ -248,7 +248,7 @@ class RDLRouter
   odb::dbNet* debug_net_{nullptr};
 
   // Consts
-  static constexpr const char* kRouteProperty_ = "RDL_ROUTE";
+  static constexpr const char* kRouteProperty = "RDL_ROUTE";
 };
 
 }  // namespace pad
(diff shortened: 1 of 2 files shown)


stefanottili:
> ... do not need to be RDL routed, this can be done by assigning `RDL_ROUTE` to a terminal.

I would think that `RDL_ROUTE 0` would describe this better, no ?

gadfort:
> > ... do not need to be RDL routed, this can be done by assigning `RDL_ROUTE` to a terminal.
> 
> I would think that `RDL_ROUTE 0` would describe this better, no ?

good point, I'll change it

src/pad/src/RDLRouter.h
```diff
@@ -246,6 +246,9 @@ class RDLRouter
   // Debugging
   RDLGui* gui_;
   odb::dbNet* debug_net_{nullptr};
+
+  // Consts
+  static constexpr const char* kRouteProperty_ = "RDL_ROUTE";
```
maliberty: No need for the trailing _ (https://google.github.io/styleguide/cppguide.html#Constant_Names)

maliberty (changes requested):

maliberty (approved):

PR #9087, opened 2025-12-18
gpl: fix bug related to snapshot saving

gudeh:
Bug inserted at https://github.com/The-OpenROAD-Project/OpenROAD/pull/9021, small oversight.

With the new parameter for user setting of routability snapshot overflow it required a checking if routability was finished, but the added check did not consider when routability is disabled at all.

The issue would only be reproducible  in 3-1 if a divergence happens.


gudeh:
I triggered a secure-CI at: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-gpl-fix-bug/1/pipeline

gudeh:
All green on public designs, 2 private currently executing

maliberty (approved):

PR #9091, opened 2025-12-18
mpl: cover cases in which a fixed macro is partially inside the core

### LINKED ISSUE #9085: MPL should handle fixed macros that are partially inside the macro placement area
### Description

Currently neither shaping/placement can handle it.

### Suggested Solution

Use only the area of the macro that is within the outline.

### Additional Context

_No response_

AcKoucher:
Resolve #9085.

With the changes here, when a fixed macro is partially inside the outline, we'll compute and use their intersection.


AcKoucher:
Running Secure-CI.

AcKoucher:
Re-running Secure-CI due to possible interference from #9077.

AcKoucher:
Secure-CI is fine.

PR #9098, opened 2025-12-19
odb: use move instead of copy

gadfort:
Fixes Coverity 1643823

### DIFF
--- a/src/odb/src/db/dbBlock.cpp
+++ b/src/odb/src/db/dbBlock.cpp
@@ -3473,7 +3473,7 @@ int _dbBlock::globalConnect(const std::vector<dbGlobalConnect*>& connects,
                                }),
                 insts.end());
 
-    inst_map[inst_pattern] = insts;
+    inst_map[inst_pattern] = std::move(insts);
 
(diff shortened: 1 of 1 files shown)


calewis (commented):
Looks good.

src/odb/src/db/dbBlock.cpp
```diff
@@ -3473,7 +3473,7 @@ int _dbBlock::globalConnect(const std::vector<dbGlobalConnect*>& connects,
                                }),
                 insts.end());
```
calewis, line 3474: While here would you consider replacing the complicated erase sequence above with:
```
std::erase_if(insts, [](dbInst* inst) {
  return inst->isDoNotTouch();
});
```
Simplified example: https://godbolt.org/z/cM198sf8W

gadfort: `remove_insts` is used later on.
I think this is out of scope for this PR (which is just to resolve the coverity thing).

calewis: That's fair.

PR #9103, opened 2025-12-19
Cts optimize tech char iterations

arthurjolo:
Tech Char was performing unnecessary operations in 2 cases:

1. For pure wire topologies it was testing all slew and input cap combinations for each of the available buffers, even though there is no buffers in the topology and it doesn't change the results.
2. For the first combination of buffers it was repeating the slew and input cap combinations at the last iteration.

This PR prevents these unnecessary operations and does not impact QoR


maliberty:
needs a clang-format

Issue #8586, opened 2025-10-08
Generate compile.json from the bazel build for clang-tidy


maliberty:
### Description

Needs to replace this being generated by cmake

hzeller:
I can take this

maliberty:
Does clang-tidy need to run in the sandbox from bazel?  I wonder if the compile.json will point to files that are removed when the sandbox is cleaned up.

maliberty:
@QuantamHD suggests https://github.com/hedronvision/bazel-compile-commands-extractor

hzeller:
Whenever I tried, I never had good experience with the hedronvision tool.

I'll built something with [`bant`](http://bant.build) that works pretty well, and is maintained by someone close to the team.
Probably this weekend.

hongted:
https://bazel.build/install/ide#autocomplete-for-source-code has additional options.

Though if this is only for clang-tidy, then something like:
https://github.com/erenon/bazel_clang_tidy

can be used without needing a compile.json file.

maliberty:
It is only for clang_tidy afaik.

hzeller:
I also use the compilation DB for support for `clangd`, this is why I prefer using bant to create a compilation DB.
Also, if we use our clang-tidy script, it has more sophisticated caching than what bazel can provide, so it is also nicer.

I have a week of vacation now, so I might get to it.

mlyoung101:
Sorry to post so close to Christmas, but I have returned to this as I've been reading some PnR papers while traveling and am trying to read the OpenROAD GPL/RePlAce code with clangd jump-to-reference.

First - re @maliberty the above:
> It is only for clang_tidy afaik.

I'm not sure what the dev setup at OpenROAD/Precision Inno is like, but a functional `compile_commands.json` is _very_ useful for clangd as Henner said. There are of course a number of other clang-tools (-tidy, -format, etc) that will benefit from a compile_commands.

---

Second - I did a little sitrep on the Bazel compile commands situation with existing tools. I've only invested about a full day into it, but I haven't gotten anything to work yet.

Hendrovision's tool (https://github.com/hedronvision/bazel-compile-commands-extractor) is non-functional with this project and likely abandoned (see https://github.com/hedronvision/bazel-compile-commands-extractor/issues/232). We run into the bug that this PR fixes, but has not been merged: https://github.com/hedronvision/bazel-compile-commands-extractor/pull/219.

[...]

PR #9122, opened 2025-12-22
Provide a way to create a compilation DB in the bazel project

hzeller:
This uses http://bant.build/ to create a compilation DB.

The change provides two useful scripts:

  * `etc/bazel-make-compilation-db.sh` creates a compilation db. To avoid a huge compile_commands.json, this generates the compile_flags.txt which is a much shorter file applied to all files. This keeps memory usage low for tools such as clang-tidy
  * `etc/run-clang-tidy.sh` a convenience script that creates the compilation DB and then calls `run-clang-tidy-cached.cc`. It uses `clang-tidy` from the llvm-toolchain configured in the MODULE.bazel

Issues #8586

### DIFF
--- a/etc/bazel-make-compilation-db.sh
+++ b/etc/bazel-make-compilation-db.sh
@@ -15,13 +15,16 @@ BANT="$($(dirname "$0")/get-bant-path.sh)"
 # Important to run with --remote_download_outputs=all to make sure generated
 # files are actually visible locally in case a remote cache (that includes
 # --disk_cache) is used ( https://github.com/hzeller/bazel-gen-file-issue )
-BAZEL_OPTS="${BAZEL_OPTS:--c opt --remote_download_outputs=all}"
+BAZEL_REMOTE_MATERIALIZE=--remote_download_outputs=all
+
+# Bazel options
+BAZEL_OPTS="${BAZEL_OPTS:--c opt ${BAZEL_REMOTE_MATERIALIZE}}"
 
 # Tickle some build targets to fetch all dependencies and generate files,
 # so that they can be seen by the users of the compilation db.
 # (Note, we need to limit targets to everything below //src, otherwise
 # it requires docker)
-"${BAZEL}" fetch //src/...
+"${BAZEL}" fetch "${BAZEL_REMOTE_MATERIALIZE}" //src/...
 "${BAZEL}" build -k ${BAZEL_OPTS} \
   @openmp//:omp_header \
   $("${BANT}" list-targets -g "genrule|tcl_encode|tcl_wrap_cc" -g "//src" | awk '{print $3}')


etc/bazel-make-compilation-db.sh
```diff
@@ -0,0 +1,33 @@
...
+"${BAZEL}" build -k ${BAZEL_OPTS} src/...
+
+# Create compilation DB. Command 'compilation-db' would create a huge *.json file,
+# but compile_flags.txt is perfectly sufficient and easier for tools to use as
+# these tools will require much less memory.
+
+"${BANT}" compile-flags -o compile_flags.txt
+
+# The QT headers are not properly picked up; add them manually.
+for f in bazel-out/../../../external/qt-bazel*/qt_source/qtbase*/build/include/Q* ; do echo "-I$f" ; done >> compile_flags.txt
```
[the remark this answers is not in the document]
hzeller: will complicate code more, but the consequence is unproblematic.

etc/bazel-make-compilation-db.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -u
```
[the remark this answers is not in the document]
hzeller: Added -e, but not pipefail as its is not very portable and not really relevant here.

etc/run-clang-tidy.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Invocation without parameters simply uses the .clang-tidy config to run on
+# all *.{cc,h} files. Additional parameters passed to this script are passed
+# to clang-tidy as-is, e.g. if you only want to run a particular check and
+# the auto-fix
+#   run-clang-tidy.sh --checks="-*,modernize-loop-convert" --fix
+
+set -u
```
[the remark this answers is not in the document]
hzeller: Added -e, but not pipefail as its is not very portable and not really relevant here.

etc/run-clang-tidy.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Use either CLANG_TIDY provided by the user as environment variable or use
+# our own from the toolchain we configured in the MODULE.bazel
+export CLANG_TIDY=${CLANG_TIDY:-$(bazel run -c opt --run_under="echo" @llvm_toolchain//:clang-tidy 2>/dev/null)}
+
+# The user should keep the compilation DB fresh, but refresh here
+# if substantial things changed or it is not there in the first place.
+if [ MODULE.bazel -nt compile_flags.txt -o BUILD -nt compile_flags.txt ] ; then
```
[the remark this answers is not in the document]
hzeller: Good catch. fixed.

etc/run-clang-tidy.sh
```diff
@@ -0,0 +1,21 @@
...
+   $(dirname $0)/bazel-make-compilation-db.sh
```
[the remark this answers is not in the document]
hzeller: done.

etc/run-clang-tidy.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+$(dirname $0)/run-clang-tidy-cached.cc "$@"
```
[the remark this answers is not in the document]
hzeller: done.

etc/bazel-make-compilation-db.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Which bazel and bant to use. If unset, defaults are used.
+BAZEL=${BAZEL:-bazel}
+BANT=$($(dirname $0)/get-bant-path.sh)
```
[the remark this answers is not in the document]
hzeller: done.

etc/get-bant-path.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
hzeller: pipefail is not very portable and it is not super relevant here.

etc/bazel-make-compilation-db.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# Which bazel and bant to use. If unset, defaults are used.
+BAZEL=${BAZEL:-bazel}
```
maliberty: We use bazelisk so it would be good to take it from .bazelversion

etc/get-bant-path.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -eu
+
+# Print path to a bant binary. Can be provided by an environment variable
+# or built from our dependency.
+
+BAZEL=${BAZEL:-bazel}
```
maliberty: same

etc/bazel-make-compilation-db.sh
hzeller: Would not recommend.
Bazel will also look for the right bazelversion and start that anyway, and complain if it is not installed, which is a good thing. So no need for baselisk. I think bazelisk is more dangerous as it attempts to download binaries which might not work on the platform in question; it is better to use the bazel binary compiled for the system itself.

hzeller: Ok, using `baselisk` now first, and if that is not installed (like on my machine), fall back to bazel.

etc/get-bant-path.sh
hzeller: Same here; attempt to use baselisk, if not installed, fall back to bazel.

hzeller:
(forgot to make it a separate commit, so it is `--amend`-ed)

etc/bazel-make-compilation-db.sh
```diff
@@ -0,0 +1,40 @@
...
+BANT="$($(dirname "$0")/get-bant-path.sh)"
+
+# Important to run with --remote_download_outputs=all to make sure generated
+# files are actually visible locally in case a remote cache (that includes
+# --disk_cache) is used ( https://github.com/hzeller/bazel-gen-file-issue )
+BAZEL_OPTS="-c opt --remote_download_outputs=all"
+
+# Tickle some build targets to fetch all dependencies and generate files,
+# so that they can be seen by the users of the compilation db.
+# Right now, comile everything (which should not be too taxing with the
```
maliberty:
```suggestion
# Right now, compile everything (which should not be too taxing with the
```

etc/bazel-make-compilation-db.sh
```diff
@@ -0,0 +1,40 @@
...
+if ! command -v "${BAZEL}">/dev/null ; then
+  BAZEL=bazel
+fi
+
+BANT="$($(dirname "$0")/get-bant-path.sh)"
+
+# Important to run with --remote_download_outputs=all to make sure generated
+# files are actually visible locally in case a remote cache (that includes
+# --disk_cache) is used ( https://github.com/hzeller/bazel-gen-file-issue )
+BAZEL_OPTS="-c opt --remote_download_outputs=all"
```
maliberty: It doesn't seem to hit the cache.  I see the CI builds with
bazelisk test --config=ci --show_timestamps --test_output=errors --curses=no --force_pic --profile=build.profile --remote_header=Authorization=Basic **** ...

which I'm guessing isn't a match

maliberty:
compile_flags.txt seems to try to be a universal set of includes rather than the specific includes used by a given translation unit.  I don't see how that will work in general.   For example:
```
find src/ -name graphics.h
src/mpl/src/graphics.h
src/fin/src/graphics.h
src/exa/src/graphics.h
```
build/compile_commands.json is 3.4M so larger but not terrible.   cmake is much faster at this job as it doesn't require a build.  I think you only need to generate the swig wrappers and tcl encoding files.

etc/bazel-make-compilation-db.sh
hzeller: Since I can not run the ci configuration as a regular user (no permissions), I've now kept that default, but made it overridable with an environment variable, so that the CI or personal set-ups can set that to desired values.

etc/bazel-make-compilation-db.sh
hzeller: Done.

hzeller:
> compile_flags.txt seems to try to be a universal set of includes rather than the specific includes used by a given translation unit. I don't see how that will work in general. For example:
> 
> ```
> find src/ -name graphics.h
> src/mpl/src/graphics.h
> src/fin/src/graphics.h
> src/exa/src/graphics.h

Yeah, that is a general problem of the project layout which I have said many times before. Once we can abandon CMake, we should fix the paths to always use paths relative to project root.

> build/compile_commands.json is 3.4M so larger but not terrible. 

For bazel, it would be several hundred megabytes, as it needs to encode all the paths. With that, running clang-tidy on 128 cores will require several hundred gigabytes of RAM. So unfortunately, a compile_commands.json is not an option right now.

> cmake is much faster at this job as it doesn't require a build.

We do that now also but build the relevant generated files. Ideally, `bazel` would provide a compilation-db in the analysis phase, alas, it is not a feature they have. Bant is the best we have.

> I think you only need to generate the swig wrappers and tcl encoding files.

Yep, changed that now, so only the fetch of the dependencies is done plus the handful of generated things (much faster)

mlyoung101:
Thank you for your work Henner, just confirming this works exactly as expected from my side, including from clangd.

The one observation I do have is that clangd does not seem to generate its symbol database without a compile_commands.json, meaning that it has to re-index the project on the fly as you edit; which is a bit slow in some large files, particularly OpenDB-related things.

I expect that clangd resource usage will be high, but more tame than clang-tidy. If Bant supports generating compile_commands without too much extra trouble, it would be interesting to try and see how this behaves.

mlyoung101:
The above ended up being simple enough to do, so I've got some numbers on it. The `compile_commands.json` database is ~900MB in the default generated format with no minification. `clangd` uses peak 12GB of RAM to generate the index, which takes about half an hour. I haven't yet tested `clang-tidy`, the reason being that the database seems invalid; it produces symbol redefinition errors that the `compile_commands.txt` doesn't seem to generate for some reason.

Machine specs are 32 cores, 64GB DDR4 RAM.

hzeller:
There are a lot of paths that are needed for bazel, as each external dependency has its own lengthy path. The paths bant emits are also for possible generated sources, so there might be a superset that is not needed in practice.
Usually, just relative paths (the `../../../externa./`) should be needed, alas `clangd` has a fault which does not follow symbolic links directly, hence the same paths show up twice. In any case, that is the reason why you observe such a gigantic file if you use compile_commands.js, which is not very practical with `clangd` and in particular not `clang-tidy` if you run it on many cores in parallel.
Hence the recommendation (as implemented in this PR) to use `compile_commands.txt`.

Anyway, ready to merge ?

maliberty (approved):

maliberty:
Thanks we can always refine more later.

hzeller:
The clangd works for me well with the compile_flags.txt @mlyoung101 ; after a bit of tweaking in recent submits, the coverage of symbols it finds looks good ( `etc/bazel-make-compilation-db.sh` )

The re-definition errors were some cyclic dependencies I've fixed recently.

If you run `etc/run-clang-tidy.sh` it will run clang-tidy on all your cores, so you do want a small compilation DB; that is why I am using compile_flags.txt, but it also tremendously makes work with `clangd` faster, as it also only has to deal with a much smaller file.

Can you try from head ?

mlyoung101:
Thanks @hzeller, I pulled master and ran the compilation script, and it works exactly as expected, no bogus errors. Very much appreciated for your work here, it's looking fantastic :))

PR #9127, opened 2025-12-22
gui: new tcl command for worst path image

gudeh:
Allows for automatically generating the worst path image at the final report from ORFS.

`gui::show_worst_path`
image


src/gui/src/timingWidget.cpp
```diff
@@ -789,6 +789,27 @@ void TimingWidget::hideEvent(QHideEvent* event)
   toggleRenderer(false);
 }
 
+void TimingWidget::showWorstPath(bool setup)
+{
+  populatePaths();
```
gadfort: this causes the table to be rebuilt, probably only needed if the model is empty

src/gui/src/gui.cpp
```diff
@@ -863,6 +864,13 @@ void Gui::saveHistogramImage(const std::string& filename,
       filename, chart_mode, width, height);
 }
 
+void Gui::showWorstPath(bool setup)
+{
+  if (main_window->getTimingWidget()) {
```
gadfort: check if enabled.
The timing widget will never be nullptr if enabled.

src/gui/src/gui.cpp
```diff
@@ -863,6 +864,13 @@ void Gui::saveHistogramImage(const std::string& filename,
       filename, chart_mode, width, height);
 }
 
+void Gui::showWorstPath(bool setup)
```
gadfort: `showWorstTimingPath`

src/gui/src/timingWidget.cpp
gudeh: how can I check if the model is empty? Perhaps:
```
  if (setup_timing_paths_model_ == nullptr
      || hold_timing_paths_model_ == nullptr) {
    populatePaths();
  }
```

gadfort: I think: https://doc.qt.io/qt-6/qabstractitemmodel.html#rowCount

gadfort:
@gudeh it occurred to me that it might be useful to have a way to hide the timing path from TCL as well. Otherwise this command becomes a one-way thing in the scripts

src/gui/src/gui.i
```diff
@@ -332,6 +332,15 @@ void save_histogram_image(const char* filename, const char* mode, int width_px =
   gui->saveHistogramImage(filename, mode, width_px, height_px);
 }
 
+void show_worst_path(bool setup = true)
```
maliberty, line 335: Add to the docs

gudeh:
> @gudeh it occurred to me that it might be useful to have a way to hide the timing path from TCL as well. Otherwise this command becomes a one-way thing in the scripts

I included a `gui::clear_timing_path`, and tested both commands at `make do-final`, one after the other:
```
gui::show_worst_path
save_image -resolution $resolution $::env(REPORTS_DIR)/final_worst_path.webp
```
image

```
gui::clear_timing_path
save_image -resolution $resolution $::env(REPORTS_DIR)/final_clear_timing_path.webp
```

image

src/gui/README.md
```diff
@@ -945,6 +945,30 @@ Update the paths in the Timing Report widget:
 gui::update_timing_report
 ```
 
+### Show Worst Path
+
+Update the paths in the Timing Report widget and select the path with the worst slack:
+
+```tcl
+gui::show_worst_path
```
gadfort, line 953: @gudeh this does not match the code implemented. Please correct.

PR #9137, opened 2025-12-23
mpl: remove ignore-std-cell-area mechanism

AcKoucher:
Never used and if used generates a result that won't make much sense.
We'll have a different way of handling and informing the user that MPL is struggling to converge with the work being made for #8732.


AcKoucher:
Secure-CI is fine.

eder-matheus (approved):

PR #9145, opened 2025-12-24
CTS techChar: add max wl constraint

arthurjolo:
### Problem
During the HTree building process CTS chooses wire topologies made by techChar. The decision on which topology to use is made by the delay and output slew of the topology, choosing the one with that respects the output slew threshold and has the minimum delay.

This approach ends up causing CTS to always choose pure wire topologies, that are usually faster than buffered ones, only choosing a buffered topology when it violates to the slew threshold. For some designs we use too many pure wire topologies in seqeunce that a buffered topology would be faster but techChar doesn't notices that.

### Fix
To fix this issue, this PR adds a max wirelength threshold that is applied the same way as the slew threshold, preventing CTS from adding long wires in the middle of the tree. The value for this threshold is the `resizer max_wirelength`.

### Disclamer
Although this PR is able to fix this issue, techChar should be able to understand that choosing too many wired topology in sequence is slower that adding a buffered topology without a threshold to prevent this. TechChar needs some more investigation regarding this issue.

### DIFF
--- a/src/cts/test/BUILD
+++ b/src/cts/test/BUILD
@@ -107,6 +107,10 @@ filegroup(
                 "array_tile.lef",
                 "array_tile_ins_delay.lib",
             ],
+            "array_max_wl": [
+                "array_tile.lef",
+                "array_tile.lib",
+            ],
             "array_no_blockages": [
                 "array_tile.lef",
                 "array_tile.lib",
@@ -114,6 +118,7 @@ filegroup(
             "array_repair_clock_nets": [
                 "array_tile.lef",
                 "array_tile.lib",
+                "array_max_wl.defok",
             ],
             "check_buffer_inference1": [
                 "ModNangate45/ModNangate45_typ.lib",


maliberty:
secure-ci?

src/cts/src/HTreeBuilder.cpp
```diff
@@ -1407,23 +1408,44 @@ void HTreeBuilder::computeLevelTopology(const unsigned level,
...
     inputCap = previousLevel.getOutputCap();
     inputSlew = previousLevel.getOutputSlew();
     remainingLength = previousLevel.getRemainingLength();
+    currWl = previousLevel.getCurrWl();
   }
 
   const unsigned SLEW_THRESHOLD = options_->getMaxSlew();
   const unsigned INIT_TOLERANCE = 1;
+
+  int WIRELENGTH_THRESHOLD;
```
maliberty: This is an ordinary variable so use lower case naming.

src/cts/src/HTreeBuilder.cpp
```diff
@@ -1407,23 +1408,44 @@ void HTreeBuilder::computeLevelTopology(const unsigned level,
...
   if (level > 1) {
     const LevelTopology& previousLevel = topologyForEachLevel_[level - 2];
     inputCap = previousLevel.getOutputCap();
     inputSlew = previousLevel.getOutputSlew();
     remainingLength = previousLevel.getRemainingLength();
+    currWl = previousLevel.getCurrWl();
   }
 
   const unsigned SLEW_THRESHOLD = options_->getMaxSlew();
   const unsigned INIT_TOLERANCE = 1;
```
maliberty:
```suggestion
  const unsigned kSlewThreshold = options_->getMaxSlew();
  const unsigned kInitTolerance = 1;
```

src/cts/src/HTreeBuilder.cpp
```diff
@@ -1548,6 +1575,17 @@ unsigned HTreeBuilder::computeMinDelaySegment(const unsigned length,
               return;
             }
 
+            if (seg.isBuffered()
+                && (currWl + seg.getWl2FisrtBuffer() > wirelengthThreshold)) {
```
maliberty: Please fix the typo in getWl2FisrtBuffer -> getWl2FirstBuffer

arthurjolo:
Secure-CI shows no changes.

arthurjolo:
Added a unity test `array_max_wl`

arthurjolo:
No metrics update needed.

maliberty:
The code looks ok but you need to fix the failure in cts.array_max_wl.tcl

PR #9147, opened 2025-12-25
test/orfs/gcd: command line adjustable clock period

oharboe:
First build:

    bazelisk build test/orfs/gcd:gcd_synth

Observe that clock period is picked up from variable and written out by write_sdc after synthesis, hence no further dependency on ABC_CLOCK_PERIOD_IN_PS variable:

    $ grep create_clock $(bazelisk info bazel-bin)/test/orfs/gcd/results/asap7/gcd/base/1_synth.sdc
    create_clock -name core_clock -period 310.0010 [get_ports {clk}]

### DIFF
--- a/test/orfs/gcd/BUILD
+++ b/test/orfs/gcd/BUILD
@@ -12,6 +12,9 @@ orfs_flow(
         "DIE_AREA": "0 0 16.2 16.2",
         "PLACE_DENSITY": "0.35",
         "OPENROAD_HIERARCHICAL": "1",
+        # Demonstrate clock period specified in arguments, useful
+        # for Optuna scripts.
+        "ABC_CLOCK_PERIOD_IN_PS": "310.001",
     },
     sources = {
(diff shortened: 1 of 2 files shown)


oharboe:
@MrAMS FYI

test/orfs/gcd/BUILD
```diff
@@ -12,6 +12,9 @@ orfs_flow(
         "DIE_AREA": "0 0 16.2 16.2",
         "PLACE_DENSITY": "0.35",
         "OPENROAD_HIERARCHICAL": "1",
+        # Demonstrate clock period specified in arguments, useful
+        # for Optuna scripts.
+        "ABC_CLOCK_PERIOD_IN_PS": "310.001",
```
maliberty, line 17: Curious why you went for 310.001 and not 310

oharboe: Same number as it used to be, but textually different, so I was in no doubt it is working and I don't have to redo metadata for tests.

PR #9164, opened 2025-12-29
gui: correct documentation for new worst path image TCL command

gudeh:
Leftover from PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/9127

### DIFF
--- a/src/gui/src/gui.tcl
+++ b/src/gui/src/gui.tcl
@@ -471,8 +471,6 @@ proc add_label { args } {
 }
 
 namespace eval gui {
-sta::define_cmd_args "show_worst_path" {[-setup|-hold]}
-
 proc show_worst_path { args } {
   sta::parse_key_args "show_worst_path" args \
     keys {} flags {-setup -hold}


gadfort:
@gudeh I think it would be better to make the command as documented, these direct swig calls are hard to read and the scripts become very sensitive to any changes to the API

gudeh:
What do you think about this version @gadfort? I made it so the command uses the `gui::` namespace, similar to `gui::set_display_controls`

src/gui/README.md
```diff
@@ -951,14 +951,15 @@ Update the paths in the Timing Report widget and select the path with the worst
 
 ```tcl
 gui::show_worst_path
```
gadfort, line 953: I think the command is now show_worst_path, not inside the gui namespace

gudeh: Have you noticed this part of the code in gui.tcl:

```
namespace eval gui {

sta::define_cmd_args "show_worst_path" {[-setup|-hold]}

proc show_worst_path { args } {
  sta::parse_key_args "show_worst_path" args \
    keys {} flags {-setup -hold}

  sta::check_argc_eq0 "show_worst_path" $args

  set setup 1
  if { [info exists flags(-hold)] } {
    set setup 0
  }

  gui::show_worst_path_internal $setup
}

sta::define_cmd_args "clear_timing_path" {}

proc clear_timing_path { args } {
  sta::parse_key_args "clear_timing_path" args \
    keys {} flags {}

  sta::check_argc_eq0 "clear_timing_path" $args

  gui::clear_timing_path_internal
}

} ;# namespace gui
```

gudeh: the usage is also the same with `gui::show_worst_path`, like in ORFS `save_iamges.tcl`: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/528ad3d8ad11cd703c99fc5c223da17b204d0ffe/flow/scripts/save_images.tcl#L116C1-L116C21

gadfort: then you need to remove `sta::define_cmd_args "show_worst_path" {[-setup|-hold]}` this is not the correct path and adding causes opensta to emit the function into the global namespace.

gudeh: Done! Let me know if you think its ok. I tested with all options (-setup, -hole, or none) and behaves as expected.

PR #9165, opened 2025-12-29
“More Recover Power” engine for repair_timing -recover_power

PrecisEDAnon:
This PR gates behind a toggle a power-saving resizer that cuts, on average, 8.5% power across various designs (range : 2-20%).

The following content was AI generated.

# Optional “More Recover Power” engine for `repair_timing -recover_power`

## Summary
This PR adds an **optional** alternative implementation of `recover_power` in the resizer. The new engine (`RecoverPowerMore`) is designed to recover additional power by **systematically trading slack headroom for power** via controlled cell swaps (size and VT) and (non-clock) buffer removal, while enforcing timing and DRV guardrails.

**Default behavior is unchanged.** The new engine is gated behind a runtime toggle so existing flows match the legacy `recover_power` unless explicitly enabled.

## Motivation / When It Helps
`repair_timing -recover_power` is commonly used to reduce post-route power once timing is reasonably stable. In practice, additional power recovery is often available in non-critical regions (and sometimes within clock networks) but requires:

- Better prioritization of “high power × high slack headroom” opportunities.
- Iteration (recompute timing, spend slack again) rather than a single pass.
- Guardrails to avoid regressions in **setup/hold** and **DRV** (slew/cap/fanout).

The new engine targets these gaps while remaining opt-in.

## What This PR Changes
- **New engine:** `RecoverPowerMore` (C++) as an alternate `recover_power` implementation.
- **Runtime selection (no rebuild needed):** legacy vs. new engine is chosen at runtime via a toggle.
- **Tcl integration:** adds a new flag to `repair_timing`:
  - `repair_timing ... -recover_power <pct> -more_recover_power`
- **Environment fallback:** `MORE_RECOVER_POWER=1` enables the new engine when the Tcl flag is not set.
  - Accepted “true” values: `1`, `true`, `yes`, `on`
  - Accepted “false” values: `0`, `false`, `no`, `off`, empty
  - Unrecognized values warn and default to **false**.

## Usage / Toggle Behavior
This PR is intentionally modular:

- If your flow does **not** call `repair_timing -recover_power`, nothing changes.
- If `-recover_power` is used and the toggle is **off**, behavior matches the legacy engine.
- If `-recover_power` is used and the toggle is **on**, the new `RecoverPowerMore` engine runs.

### Enable / Disable
Enable per call:
```tcl
repair_timing -recover_power 100 -more_recover_power
```

Enable via environment (useful for CI/harness A/B):
```tcl
set ::env(MORE_RECOVER_POWER) 1
repair_timing -recover_power 100
```

Disable:
```tcl
repair_timing -recover_power 100
```

## Technical Details (High Level)
The new engine is a bounded, multi-pass greedy optimizer with rollback:

1. **Baseline capture (guardrails):**
   - Records initial setup/hold WNS.
   - Records initial DRV violation counts (slew/cap/fanout).
2. **Timing floors:**
   - If setup is closed (WNS ≥ 0): enforce `WNS ≥ 0`.
   - If setup is already failing: allow a small, bounded additional WNS degradation budget proportional to clock period and the requested `-recover_power` percent.
   - Enforce hold floor similarly (do not introduce hold failures when closed).
3. **Candidate selection and ranking:**
   - Considers leaf logic standard cells (excluding `dont_touch`) and includes clock-driving instances for swap opportunities.
   - Ranks candidates by **(instance power × available slack headroom)**, prioritizing high-impact, non-critical instances.
4. **Per-instance optimization (bounded effort):**
   - Optionally tries **buffer removal** for non-clock buffers (clock buffers are not removed).
   - Attempts a limited number of swaps:
     - **Downsize / area reduction** first (dynamic power).
     - **Lower leakage VT** swaps next (leakage power).
5. **Acceptance criteria and rollback:**
   - Each tentative move runs under `Resizer` journal, updates parasitics, and re-runs STA requireds.
   - A move is accepted only if:
     - Setup/hold remain above their floors, and
     - Estimated global DRV violation counts do not exceed the recorded baseline.
   - Rejected moves are cleanly rolled back via journal restore.
6. **Progress logging:**
   - Prints a periodic progress table (area delta, swap/remove counts, WNS, total power, DRV counts).
   - `-verbose` emits per-move accept/reject debug prints.

## Validation Recommendation
1. Run a baseline with legacy `repair_timing -recover_power <pct>`.
2. Re-run with `-more_recover_power` enabled under identical conditions.
3. Compare at minimum: total power, WNS (setup/hold), DRV counts, and runtime.
4. Start with smaller `-recover_power` values (e.g. 25/50) if runtime is a concern, then scale up.

## CI Note
Default is **off** to preserve legacy behavior. It is recommended to exercise the new path in some CI configuration (e.g., nightly) by enabling `MORE_RECOVER_POWER=1` so the alternate code path is continuously built and tested.

### DIFF
--- a/src/rsz/src/RecoverPowerMore.cc
+++ b/src/rsz/src/RecoverPowerMore.cc
@@ -5,39 +5,30 @@
 
 #include <algorithm>
 #include <cmath>
+#include <cstdint>
 #include <cstring>
 #include <limits>
 #include <memory>
 #include <optional>
 #include <string>
-#include <tuple>
 #include <unordered_set>
-#include <utility>
 #include <vector>
 
-#include "db_sta/dbNetwork.hh"
 #include "db_sta/dbSta.hh"
+#include "est/EstimateParasitics.h"
+#include "grt/GlobalRouter.h"
 #include "odb/db.h"
 #include "rsz/Resizer.hh"
 #include "sta/Clock.hh"
 #include "sta/Corner.hh"
-#include "sta/DcalcAnalysisPt.hh"
 #include "sta/Delay.hh"
-#include "sta/Fuzzy.hh"
 #include "sta/Graph.hh"
-#include "sta/GraphDelayCalc.hh"
-#include "sta/InputDrive.hh"
 #include "sta/Liberty.hh"
 #include "sta/Network.hh"
 #include "sta/NetworkClass.hh"
-#include "sta/Parasitics.hh"
-#include "sta/PortDirection.hh"
 #include "sta/PowerClass.hh"
 #include "sta/Sdc.hh"
-#include "sta/Search.hh"
-#include "sta/TimingArc.hh"
 #include "sta/Units.hh"
-#include "sta/Vector.hh"
 #include "utl/Logger.h"
 
 namespace rsz {
@@ -224,7 +215,8 @@ bool RecoverPowerMore::recoverPower(const float recover_power_percent,
       sta_->worstSlack(max_, current_wns, current_wns_vertex);
       (void) current_wns_vertex;
 
-      std::vector<CandidateInstance> candidates = collectCandidates(current_wns);
+      std::vector<CandidateInstance> candidates
+          = collectCandidates(current_wns);
       candidates.erase(std::remove_if(candidates.begin(),
                                       candidates.end(),
                                       [&](const CandidateInstance& cand) {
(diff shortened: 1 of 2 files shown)


maliberty:
It is helpful to give more specific data vs
```
This PR gates behind a toggle a power-saving resizer that cuts, on average, 8.5% power across various designs (range : 2-20%).
```
A table of improvements and how they compare to the current recover_power implementation would be helpful.  Including power, wns for setup & hold.

If you evaluate timing on every move I would also be concerned about runtime so it would be good to include that in the comparison.

PR #9177, opened 2025-12-31
Gui goto

### LINKED ISSUE #8700: gui: "Go to position" bug
### Describe the bug

If I load a design in OpenROAD tool, go to "View", and "Go to position" (or use Shift+G), when trying to modify the X or Y values it does not work, I start typing but the cursor is removed from the typing filed while typing. I can't even use backspace to remove the existing values in X and Y.

It works fine if I first modify the "Size", than I can modify the X and Y and it zooms in properly.

Image

On this video I noticed delete or backspace both do not work, I am not able to remove the last number (I am consistently pressing backspace in the video, but the last digit is not erased). If I select the whole text I am able to replace the existing value (but only after changing "Size" value).

https://github.com/user-attachments/assets/4b020d11-2b75-4c75-9075-4dbcd7211925

### Expected Behavior

Be able to freely modify X and Y in "Go to position"

### Environment

```shell
I currently see this, but it happened with the latest version when this issue was created.

./etc/Env.sh 
Error: Access to the public repository 'The-OpenROAD-Project/OpenROAD' is restricted.
Please use the private repositories for your work.
```

### To Reproduce

Load a design, press shift+G.

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

LucasYuki:
Closes #8700

image

Changes:
- The "Go to position" dialog only changes the position when the "Goto" button is pressed.
- Adds the "Update" button to update the values of the current view.
- Changes the LayoutViewer::zoomTo padding calculations to make it easier to calculate the inverse operation.
- Adds the method LayoutViewer::getVisibleDiameter() to get the current view size.

### DIFF
--- a/src/gui/src/mainWindow.cpp
+++ b/src/gui/src/mainWindow.cpp
@@ -45,6 +45,7 @@
 #include "displayControls.h"
 #include "drcWidget.h"
 #include "globalConnectDialog.h"
+#include "gotoDialog.h"
 #include "gui/gui.h"
 #include "gui/heatMap.h"
 #include "helpWidget.h"


maliberty:
I'm holding of reviewing until you make the changes to the update button we discussed.

src/gui/src/gotoDialog.cpp
```diff
@@ -32,32 +33,28 @@ void GotoLocationDialog::updateUnits(int dbu_per_micron, bool use_dbu)
...
 // NOLINTNEXTLINE(readability-non-const-parameter)
-void GotoLocationDialog::updateLocation(QLineEdit* x_edit, QLineEdit* y_edit)
+void GotoLocationDialog::updateLocation()
 {
   auto viewer = viewers_->getCurrent();
-  x_edit->setText(QString::fromStdString(Descriptor::Property::convert_dbu(
+  if (!viewer) {
+    return;
+  }
+  xEdit->setText(QString::fromStdString(Descriptor::Property::convert_dbu(
```
maliberty: Please change the naming in the .ui so that the name here corresponds to the standard convention (x_edit/y_edit)

maliberty:
If I switch to 'Show DBU' it is should use integers (shows microns & dbus for the same values)
image
image

maliberty:
An oddity is that if I enter Size=5 and hit goto I get a slightly different value.  I am not hitting the edges of the area
image
image

LucasYuki:
> An oddity is that if I enter Size=5 and hit goto I get a slightly different value. I am not hitting the edges of the area image image

Unfortunately, there are integer conversions that introduce errors from the requested size (float) to the view (pixels), so the inverse function is a little bit inaccurate.

LucasYuki:
> If I switch to 'Show DBU' it is should use integers (shows microns & dbus for the same values) image image

The GUI already does the conversion automatically.
The method GotoLocationDialog::updateUnits was doing the conversion again, so I removed it.

maliberty:
Please fix the tidy

PR #9179, opened 2025-12-31
odb: fix 3dblox declaration mismatches

ahmed532:
Some structs in src/odb/include/odb/3dblox.h where forward declared as classes. Which errored with "-Wmismatched-tags" compiler flag. Also, found another mismatch in PartitionMgr.h
Note that C++ does not distinguish between class and struct in a forward declaration. Only MSVC can have issues with it.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -125,6 +125,7 @@ OPENROAD_COPTS = [
     "-Wformat-security",
     "-Wno-sign-compare",
     "-Wno-unused-parameter",
+    "-Wmismatched-tags",
     "-fopenmp",
 ]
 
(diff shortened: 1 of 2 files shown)


maliberty:
Apparently it matters on mac as well which doesn't use MSVC.

ahmed532:
> Apparently it matters on mac as well which doesn't use MSVC.

OK I'll add it to compile options for clang then.

ahmed532:
> This started with https://github.com/The-OpenROAD-Project/OpenROAD/issues/9093 which was reported by Stefan who uses a mac.  To my knowledge we have zero MSVC users (nobody has ever reported an issue).   In the report it is a warning but we should clean it up.  In the CI we treat warnings as errors to keep them from creeping in.
I don't think you need to make it always an error as non-devs should be able to build even if we have a mismatch
option(ALLOW_WARNINGS "Flag to allow compilation with compiler warnings: on by default" ON)
```
if (NOT ALLOW_WARNINGS)
  add_compile_options(
    $<$<CXX_COMPILER_ID:GNU>:-Werror>
  )
endif()
```
> in src/CMakeLists.txt
    in the CI `ALLOW_WARNINGS=0`
    
Quote by @maliberty

PR #9181, opened 2025-12-31
GRT: add options for repair antennas

### LINKED ISSUE #9042: `repair_antennas`: additional options
#### `-allow_congestion`

I would appreciate an `-allow_congestion` option for `repair_antennas`.

For ORFS, this is not a problem, since the entire flow runs within a single process of OpenROAD. Therefore, a previous invocation of GRT sets the `-allow_congestion` option, which is later also used for `repair_antennas`.
However, in LibreLane we load the ODB at every step (checkpoint). Consequently, `-allow_congestion` is not set from a previous invocation of GRT within the current process.

#### `-jumper_only` / `-diode_only`

I would also appreciate the option of fixing antenna violations using only one of the two methods.

Some designs only integrate macros at the top-level with standard cell rows disabled.
In this case, antenna violations can only be fixed using jumpers, since it is not possible to place antenna diodes.

(I can open a separate issue for this feature request if you prefer that.)

### Suggested Solution

_No response_

### Additional Context

_No response_

luis201420:
Fixes #9042 
Add the next flags for `repair_antennas` command:
- `allow_congestion`: Allow congestion when rerouting repaired nets.
- `jumper_only`: Use only jumpers to repair nets.
- `diode_only`: Use only diodes to repair nets.


eder-matheus (changes requested):
Please address clang-format and also add some unit tests for the new flags.

Issue #9117, opened 2025-12-21
Incorrect yosys command in ./etc/DependencyInstaller.sh


hariramakrishnan1:
### Describe the bug

./etc/DependencyInstaller.sh
`_install_yosys() {
    local yosys_prefix=${PREFIX:-"/usr/local"}
    local yosys_installed_version="none"
    if _command_exists "yosys"; then
        yosys_installed_version=$(yosys --version | awk '{print $2}')
    fi
   ...
`
Line 144 calls yosys --version when it should be yosys -V. calling --version is incorrect (see yosys -h) and causes setup.sh from Open-Road-flow-scripts to fail. However, this problem is technically contained to just the OpenROAD dependency installer. 

### Expected Behavior

I expected the setup script to run but it failed at the yosys build step. 

### Environment

```shell
Git commit: 86b9236cdd564adea819b0d595cf871bc92c4f9d
kernel: Linux 6.8.0-90-generic
os: Ubuntu 22.04.5 LTS (Jammy Jellyfish)
cmake version 3.22.1
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- OpenROAD version: v2.0-27248-g86b9236cdd
-- System name: Linux
-- Compiler: GNU 11.4.0
-- Build type: RELEASE
-- Install prefix: /usr/local
-- C++ Standard: 20
-- C++ Standard Required: ON
-- C++ Extensions: OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test C_COMPILER_SUPPORTS__-Wall
-- Performing Test C_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall
-- Performing Test CXX_COMPILER_SUPPORTS__-Wall - Success
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds
-- Performing Test C_COMPILER_SUPPORTS__-Wno-array-bounds - Success
[...]

PR #9186, opened 2026-01-03
Changed incorrect yosys command option from --version to -V

hariramakrishnan1:
This pull request addresses issue #9117 (https://github.com/The-OpenROAD-Project/OpenROAD/issues/9117). Unsure whether test cases are necessary since this is rather straightforward.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -141,7 +141,7 @@ _install_yosys() {
     local yosys_prefix=${PREFIX:-"/usr/local"}
     local yosys_installed_version="none"
     if _command_exists "yosys"; then
-        yosys_installed_version=$(yosys --version | awk '{print $2}')
+        yosys_installed_version=$(yosys -V | awk '{print $2}')
     fi
 
     local required_version="${YOSYS_VERSION#v}"


hariramakrishnan1:
Gemini code review incorrect after first line.
```
adam@adamc:~/dev/OpenROAD/etc$ eqy --version
EQY v0.60
adam@adamc:~/dev/OpenROAD/etc$ sby --version
SBY v0.58
adam@adamc:~/dev/OpenROAD/etc$ sby -V
usage: sby [options] [<jobname>.sby [tasknames] | <dirname>]
sby: error: unrecognized arguments: -V
adam@adamc:~/dev/OpenROAD/etc$ eqy -V
usage: eqy [options] <config>.eqy
eqy: error: unrecognized arguments: -V
```

hariramakrishnan1:
Found out that --version isn't included on older versions of Yosys, but was included in this commit: https://github.com/YosysHQ/yosys/commit/b76b02358439cd04ed5cad5cb889e8b43c801ac1. However there is this issue posted that indicates there is or was still some bug going around after that commit https://github.com/TerosTechnology/vscode-terosHDL/issues/770. How I managed to get that version of Yosys on my system in the original issue is unknown, but either way newer versions should be properly accepting --version, however, -V still works. 

We could keep --version as it is standard, or switch to -V so the installer works, regardless of system Yosys version. 
Since the --version issue has still made its way to recent issues, https://github.com/The-OpenROAD-Project/OpenROAD/issues/9117, I would recommend -V only for the installer, since it is meant to be universal, but keep --version anywhere else.

maliberty (approved):

Issue #9160, opened 2025-12-29
3DBlox: Check region connectivity


osamahammad21:
### Description

Add a linter check to validate that connected regions in a 3DBlox design geometrically intersect as intended.
For any region-to-region connection defined in the 3DBlox connectivity description, the corresponding regions must overlap or intersect in physical space.

### Additional Context

You can find the linter code in src/odb/src/3dblox/checker.cpp.
Note how the unfolded view of the chiplets is constructed and use the same idea for regions.

PR #9203, opened 2026-01-06
odb: foundation for 3DBlox connectivity checker (1)

ahmed532:
This PR provides the foundational fixes and utilities required for the 3DBlox connectivity linter.
- **Simplification (`dbChipRegionInst`)**: Refactors `getCuboid()` to delegate to the master region and apply the instance transform, removing duplicated logic.
- **Utility Extraction (`utl/UnionFind`)**: Moves the Union-Find data structure from an anonymous namespace to a reusable utility in `utl/`.
- **Output Fix (`dbxWriter`)**: Ensures correct origin coordinates are written to `.3dbx` files.
- **Recursion Safety**: Added `path.pop_back()` to `unfoldChip` to ensure correct path state during DFS traversal.
- **Hierarchy tracking fix**: `UnfoldedChip` was passed by value which could degrade performance. 

Relates to #9160

### DIFF
--- a/src/utl/include/utl/unionFind.h
+++ b/src/utl/include/utl/unionFind.h
@@ -1,3 +1,6 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2023-2026, The OpenROAD Authors
+
 #pragma once
 #include <cstddef>
 #include <vector>
--- a/src/utl/src/unionFind.cpp
+++ b/src/utl/src/unionFind.cpp
@@ -1,3 +1,6 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2023-2026, The OpenROAD Authors
+
 #include "utl/unionFind.h"
 
 #include <cstddef>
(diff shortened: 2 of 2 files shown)


src/utl/include/utl/unionFind.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
```
osamahammad21, line 4: Please add copyrights

src/utl/src/unionFind.cpp
```diff
@@ -0,0 +1,36 @@
+#include "utl/unionFind.h"
```
osamahammad21, line 4: ditto

src/utl/src/unionFind.cpp
```diff
@@ -0,0 +1,36 @@
...
+  if (parent[x] != x) {
+    parent[x] = find(parent[x]);
+  }
+  return parent[x];
+}
+
+void UnionFind::unite(size_t x, size_t y)
+{
+  size_t px = find(x);
+  size_t py = find(y);
```
osamahammad21:
```suggestion
  const size_t px = find(x);
  const size_t py = find(y);
```

ahmed532:
The dbxWriter in this PR is just a typo fix. 
Yes, UnfoldedChip is not buggy. It is just an improvement like described.

PR #9205, opened 2026-01-06
pad: add some debugging information and use int for positions

gadfort:
Changes:
- I noticed a few places where the code uses floats instead of int like it should. This corrects that
- adds some debugging information.

### DIFF
--- a/src/pad/src/PadPlacer.cpp
+++ b/src/pad/src/PadPlacer.cpp
@@ -1134,9 +1134,9 @@ std::map<odb::dbInst*, int> PlacerPadPlacer::poolAdjacentViolators(
       updated = true;
       // Calculate new value
       const float total_weight = weights[i] + weights[i - 1];
-      const int pooled_value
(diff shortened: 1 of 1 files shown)


src/pad/src/PadPlacer.cpp
```diff
@@ -944,6 +944,25 @@ std::map<odb::dbInst*, int> PlacerPadPlacer::initialPoolMapping() const
     odb::dbInst* inst = insts[i];
     mapping[inst] = position[i];
   }
+
+  if (getLogger()->debugCheck(utl::PAD, "PAVA", 2)) {
```
maliberty, line 948: what is PAVA?

maliberty (approved):

src/pad/src/PadPlacer.cpp
gadfort: Pool Adjacent Violators Algorithm (PAVA)

Issue #9209, opened 2026-01-07
3Dblox: Unfolded view


osamahammad21:
### Description

In the current 3DIC implementation in OpenROAD, ODB stores the folded (hierarchical) view of the 3DBlox design. However, correct visualization and downstream handling require an unfolded (flattened) view that resolves hierarchy transforms and produces geometry/connectivity in a single coordinate space.

Today, src/odb/src/3dblox/checker.* contains an incomplete unfolded representation that effectively unfolds chiplets only. We need to extend this into a well-defined class that represents the unfolded view of all relevant 3DBlox nodes. The current node types are:
- chiplets
- regions
- connections
- bumps

PR #9217, opened 2026-01-08
odb: Refactor 3DBlox checker into UnfoldedModel and ThreeDBloxValidator

ahmed532:
This commit separates the monolithic 3DBlox checker into two distinct components to improve modularity, maintainability, and architectural clarity:

1. UnfoldedModel: Responsible for hierarchical design "unfolding".
    - Recursively traverses the design hierarchy to build a flat, global representation of chips, regions, and bumps.
    - Employs std::deque for internal storage to guarantee pointer stability for hierarchical object references.
    - Maintains an efficient path-based map for fast object resolution.

2. ThreeDBloxValidator: Responsible for verification logic.
    - Ports existing floating chip and overlapping chip detection to the new unfolded data structure.
    - Provides a dedicated entry point for future physical alignment and connectivity rule implementations.

3. Checker Facade:
    - Updated to serve as a clean API facade that orchestrates the UnfoldedModel construction and ThreeDBloxValidator execution.
    - Maintains full backward compatibility with the current Tcl interface.

Technical Improvements:
- Decouples data construction from rule analysis (Single Responsibility Principle).
- Improved memory safety via stable containers for cross-referenced structs.
- Modularized codebase into dedicated header/implementation files.

Verified by successfully passing all 85 existing 'odb' integration tests.

Related to: #9209


src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+  return {};
+}
+
+std::string UnfoldedChip::getName() const
+{
+  std::string name;
+  for (auto* chip_inst : chip_inst_path) {
+    if (!name.empty()) {
+      name += "/";
```
jhkim-pii: "/" <--- Is the fixed delimiter ok? 
In `dbBlock`, `_dbBlock::hier_delimiter_` is configurable. 
I am not sure if the "/" should be the same as `_dbBlock::hier_delimiter_` or not.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -0,0 +1,142 @@
...
+  return getName();
+}
+
+UnfoldedModel::UnfoldedModel(utl::Logger* logger) : logger_(logger)
+{
+}
+
+void UnfoldedModel::build(dbChip* chip)
+{
+  for (auto chip_inst : chip->getChipInsts()) {
```
jhkim-pii: IMO, using `auto` for the trivial usage case hurts code readability. Changing to `dbChipInst*` would be more readable.

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -0,0 +1,117 @@
...
+  int getSurfaceZ() const;
+  bool isFacingUp() const;
+  bool isFacingDown() const;
+};
+
+struct UnfoldedConnection
+{
+  dbChipConn* connection;
+  UnfoldedRegionFull* top_region;
+  UnfoldedRegionFull* bottom_region;
```
jhkim-pii, line 56: Although there are no specific implementation for `top_region` & `bottom_region` yet, they may point entries in `std::vector<UnfoldedRegionFull> regions;`.

If # of regions is not constant and changes dynamically after construction, the vector can be reallocated and `top_region` and `bottom_region` require proper management because of the pointer address change. 

If this scenario can happen, using vector may not be a good choice.

src/odb/src/3dblox/unfoldedModel.cpp
ahmed532: I think you're right. I'm now using hier_delimiter and '/' as a default.

src/odb/src/3dblox/unfoldedModel.cpp
ahmed532: Done.

src/odb/src/3dblox/unfoldedModel.h
ahmed532: I reviewed all other instances of vector and changed some of them to deque for pointer stability. Thanks.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -0,0 +1,152 @@
...
+  char delimiter = '/';
+  if (!chip_inst_path.empty()) {
+    dbBlock* block = chip_inst_path[0]->getParentChip()->getBlock();
+    if (block) {
+      char d = block->getHierarchyDelimiter();
+      if (d != 0) {
+        delimiter = d;
+      }
+    }
+  }
```
jhkim-pii: Let's consider the following cases.
1. There are multiple chips. Should we allow different hierarchical delimiters in multiple chips?
2. If there should be a single hierarchical delimiter, storing it in ODB-level (not in block or chip) is better.

src/odb/src/3dblox/unfoldedModel.cpp
jhkim-pii: Thanks for the update.
If you agree that `auto` for trivial type can hurt readability, it is better to change other `auto` statements as well. : )

src/odb/src/3dblox/unfoldedModel.cpp
ahmed532: 1. It depends on the way we use the path later. For this particular function, it does not matter as long as we're consistent. And, it does not affect anything outside the checker. 
2. We do not need to enforce this.

In general you are right to question this, but I think it is outside the scope of this PR. I can create another issue for it.

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -0,0 +1,117 @@
...
+  bool z_flipped = false;
+  std::deque<UnfoldedRegionFull> regions;
+  std::vector<UnfoldedConnection*> connected_conns;
+};
+
+class UnfoldedModel
+{
+ public:
+  UnfoldedModel(utl::Logger* logger);
+  void build(dbChip* chip);
```
osamahammad21:
```suggestion
  UnfoldedModel(utl::Logger* logger, dbChip* chip);
```
Since you already don't clear the member variables when you call build, then calling build multiple times is gonna lead to a misleading behavior.

src/odb/src/3dblox/unfoldedModel.cpp
osamahammad21: I don't think you should infer the delimiter from the block as it will inevitably lead to misalignment and ambiguity. The delimiter should be `/`, as this is the implicitly defined delimiter in the 3DBlox standard.

PR #9218, opened 2026-01-08
Add multi-cut via support to pin access

Damonan:
### Problem

Pin access is currently hard-coded to using single-cut vias. I'm working with a PDK that has sufficiently large pins to require multi-cut vias. In addition, minimum cut violations are not handled properly.

### MINIMUMCUT Types Targeted:

**Type 1 - WIDTH only**: If the metal rect is >WIDTH, the specified number of cuts is required
**Type 2 - WIDTH and LENGTH+WITHIN**: From LEFDEF, "this rule only checks the thin wire connected to a wide wire, and does not check the wire itself". If the metal rect is >WIDTH and >LENGTH then any other connected segments must use the specified number of cuts WITHIN the specified distance.

### Key Modifications

`src/drt/src/gc/FlexGC_impl.cpp` - Update checkMinimumCut_main to correctly handle WIDTH/LENGTH MinimumCut constraints. Here's the basic algorithm:
1. If the rect is <WIDTH or <LENGTH for the constraint, skip the constraint
2. Search for all vias above/below in a box bloated by the constraint distance (WITHIN)
3. For each via in the resulting list:
           a. If the via is inside the rect of interest and it is a WIDTH constraint, add it as a valid via, and continue
           b. If the via is inside the rect of interest and we're dealing with a LENGTH/WITHIN type constraint, skip it
           b. If the via is outside the rect of interest and we're dealing with LENGTH/WITHIN type, check the other pin rects to make sure the via is connected to the pin
4. Check that the number of valid vias is either 0 (none attempted) or >= the number required by the constraint

`TritonRoute(.h, .cpp, .i, .tcl)` and `global.h` - Add two new router_cfg variables: VIA_MAX_CUT specifies the maximum number of cuts allowed. VIA_CANDIDATE_PER_CUT specifies how many cuts to use per cut type.
`src/drt/src/pa/FlexPA_acc_point.cpp` - Update filterViaAccess(...) to use VIA_MAX_CUT and VIA_CANDIDATE_PER_CUT to properly populate via_defs. genAPEnclosedBoundary(...) is updated similarly.
`src/drt/src/db/obj/frAcces.h` - Update getViaDef(s) functions to first check if hasVia for the requested numCut. If none exist, then search for a numCut that does have valid vias, and return the first one.

### To-do, looking for some feedback

1. Does VIA_MAX_CUT/VIA_CANDIDATE_PER_CUT fit in with the long-term vision? Would you prefer to move towards something more arbitrary. For example, what if the user wants a specific list of vias?
2. Magic numbers are remaining in getViaDef(s) functions. I'm not sure what the best way to get VIA_MAX_CUT into that function is. An input might be too cumbersome. Perhaps it's best to give the frAccessPoint a max-cut variable on creation?

### Test Cases:

Passes DRT regression, but only after modification. The MinimumCut constraint generator in the testbench did NOT assign any numCuts, so the constraint was invalid. Updated the constraint to have numCuts = 2, as it seems to be expected by the test case. This does suggest that MinimumCut test cases may need to be developed for robustness.


Damonan:
@osamahammad21 Hello, can you let me know if there's anything I need to do to get this through the pending checks? It was stuck in this state since Jan, and then went stale. I finally spent some time to update it, but I don't want it to go stale again. Let me know if there's anything I need to do here, thanks!

bnmfw:
This PR makes a change that highlights one of the issues on the current Pin Access structure. 

A shared logic between two functions exists, in which both functions independently know to do the same thing:
- During access point generation we take some vias to use as a basis for enclosed cordinates.
- During access point validation we use those same vias (outside deep_search) to check if an ap is valid.
That is an issue with master that is a simple one to solve. I will open a separate issue and a subsequent PR so your PR is more concise.

I still have to review the GC files, but otherwise most of the content on this PR seems fine.

bnmfw:
@Damonan, recentrly PR #11175 was merged, which creates a function that handles the choice of representative vias on DRT. Now the choice is in a single place instead of two. With this you should be able to merge master into your PR and drop the plumbing of the `viaMaxCut` and `viaCandidatePerCut` variables. This should make your whole PR way smaller and more concise.

osamahammad21 (commented):
@Damonan, Could you please @bnmfw comments

PR #9219, opened 2026-01-08
test/orfs: bump bazel-orfs and update rules .json files

AcKoucher:
For #8665.

I had to update the rules .json of all designs to get rid of the metrics that were removed by https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3750. The tests would fail otherwise.


oharboe:
@AcKoucher you may also want to run `bazelisk run @bazel-orfs//:bump` to update ORFS and bazel-orfs version used.

PR #9227, opened 2026-01-09
dpl: fix one site gap false positive

gudeh:
The checker in this part of the code was incorrectly examining more rows than intended. This appears to have been an oversight while trying to handle multi-row instances.

The issue surfaced with a private PDK that disallows one-site gaps between adjacent instances. In this case, instances were being moved even when no actual modifications were applied to the design. It happened because we were checking the one site gap horizontally in a row that the instance was not present.

The problem was reproducible with gcd during GRT when repair_design made no changes, and also when invoking DPL multiple times in sequence using the disallows one-site gaps.

I added a test that performs multiple consecutive DPL calls in the presence of a one-site gap. The test ensures that the one site gap is fixed on the first pass and no further changes are made in subsequent runs.

Finally, this issue occurs because the one-site gap check is implemented independently in both Place.cpp and CheckPlacement.cpp, without a shared helper function. A potential follow-up improvement would be to factor this logic into a common function used by both.

### DIFF
--- a/src/dpl/src/Place.cpp
+++ b/src/dpl/src/Place.cpp
@@ -73,10 +73,17 @@ void Opendp::detailedPlacement()
   if (!arch_->getRegions().empty()) {
     placeGroups();
   }
+
+  if (debug_observer_) {
+    logger_->report("Pause before detail placement.");
+    debug_observer_->redrawAndPause();
+  }
+
   place();
 
   if (debug_observer_) {
-    debug_observer_->endPlacement();
+    logger_->report("Pause after detail placement.");
+    debug_observer_->redrawAndPause();
   }
 }
 
@@ -957,7 +964,7 @@ bool Opendp::checkPixels(const Node* cell,
       }
     }
   }
-  
+
   const auto orient = grid_->getSiteOrientation(x, y, site).value();
   return drc_engine_->checkDRC(cell, x, y, orient);
 }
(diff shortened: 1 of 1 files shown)


gudeh:
A secure-CI is running at: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-dpl-fix-one-site-gap/1/pipeline

I expect no-op on public designs, but private designs may need metrics update.

gudeh:
@jfgava FYI

maliberty:
tcl lint needs fixing.  @osamahammad21 please review/merge

src/dpl/test/gcd_no_one_site_sanity.tcl
```diff
@@ -0,0 +1,17 @@
...
+detailed_placement
+detailed_placement
+detailed_placement
+detailed_placement
+detailed_placement
+check_placement -verbose
+
+set def_file [make_result_file gcd_no_one_site_sanity.def]
+write_def $def_file
+diff_file gcd_no_one_site_sanity.defok $def_file
```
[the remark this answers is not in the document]
gudeh: the two defs are not identical

```
ools/OpenROAD/src/dpl/test$ diff gcd_no_one_site_sanity.def gcd_no_one_site_sanity.defok 
617c617
<     - _525_ MUX2_X1 + PLACED ( 166130 179200 ) FS ;
---
>     - _525_ MUX2_X1 + PLACED ( 165940 179200 ) FS ;
```

src/dpl/src/Place.cpp
```diff
@@ -928,13 +928,29 @@ bool Opendp::checkPixels(const Node* cell,
...
+          debugPrint(logger_,
+                     DPL,
+                     "one_site_gap",
+                     1,
+                     "One site gap right of {} at ({}, {})",
+                     cell->name(),
+                     x,
+                     y);
           return false;
         }
```
osamahammad21: @gudeh Do you think we should move this outside the `y` for loop? It seems weird we are doing it for every y

gudeh: don't we have to do so for multirow cells?

osamahammad21: we do, but this whole block of code exists in an outer loop that iterates over the y indices of the cell.
Currently it is something like this:
```
for (GridY y1 = y; y1 < y_end; y1++) {  // line 907
  for (GridY y = y1; y < y_end; ++y) {
    ...
  }
}
```
This is repetitive. It's not caused by PR and has always been there, but it's incorrect.

gudeh: Ah, I see! We are checking for one site gap for every row, although we should require to check only once. I will move it out of the loop.

gudeh:
This requires updates on metrics for private designs. I have it running at the moment to fetch the values. A PR for it should follow soon.

PR #9228, opened 2026-01-09
utl: Fix crash for write_db ""

mikesinouye:
`write_db ""` will crash openroad. In practice this happens for a misconfigured flow in which the command `write_db $output` is used when output is "".

### DIFF
--- a/src/utl/src/ScopedTemporaryFile.cpp
+++ b/src/utl/src/ScopedTemporaryFile.cpp
@@ -68,7 +68,7 @@ OutStreamHandler::OutStreamHandler(const char* filename, bool binary)
     : filename_(filename)
 {
   if (filename_.empty()) {
-    std::runtime_error("filename is empty");
+    throw std::runtime_error("filename is empty");
   }
   tmp_filename_ = generate_unused_filename(filename_);
 


src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -67,6 +67,9 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
 OutStreamHandler::OutStreamHandler(const char* filename, bool binary)
     : filename_(filename)
 {
+  if (filename_.empty()) {
+    std::throw_with_nested(std::runtime_error("filename is empty"));
```
maliberty: Why are you using throw_with_nested here?  I thought that is just for in a catch clause.

mikesinouye: Wasn't thinking when I copy/pasted this from below, fixed.

src/utl/src/ScopedTemporaryFile.cpp
```diff
@@ -67,6 +67,9 @@ ScopedTemporaryFile::~ScopedTemporaryFile()
 OutStreamHandler::OutStreamHandler(const char* filename, bool binary)
     : filename_(filename)
 {
+  if (filename_.empty()) {
+    std::runtime_error("filename is empty");
```
maliberty: throw?

maliberty (approved):

PR #9229, opened 2026-01-10
fix replace deprecated swig $function by new $action

hashhsah:
swig 4.4 removed the long deprecated $function, switch to $action instead.

### DIFF
--- a/src/Exception-py.i
+++ b/src/Exception-py.i
@@ -23,7 +23,7 @@
 #endif
 
 %exception {
-  try { $function }
+  try { $action }
   catch (std::bad_alloc &) {
     fprintf(stderr, "Error: out of memory.");
     abort();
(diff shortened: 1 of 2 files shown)


maliberty:
Thanks for contributing.  You need to sign your commits to pass DCO (`git commit -s`).  Please see the DCO link for info on how to amend your commits.

hashhsah:
amended the commit with sign-off.

stefanottili:
Using swig 4.4.1 on MacOS with homebrew, the compile fails because src/sta/tcl/Exception.i needs to be updated too.

maliberty:
> Using swig 4.4.1 on MacOS with homebrew, the compile fails because src/sta/tcl/Exception.i needs to be updated too.

This will be resolved when we can get through the next sta update (blocked on an error currently)

PR #9234, opened 2026-01-11
gui: add screenshot tests to GUI tests to provide basic (very basic) testing

gadfort:
Notes:
- bazel requires `sudo apt install libxcb-cursor0` to work, might want to add that to the docs.
- bazel test is failing unable to write the screenshot, no clue how to debug `bazel test --//:platform=gui src/gui/test/...`

We should probably setup a second bazel CI run that performs the GUI build as well along with tests.


gadfort:
@maliberty also looks like the CI runners needs some additional "stuff" installed to be able to run the test

maliberty:
@QuantamHD I could use some help with Bazel here.  If I build the executable `bazelisk build -c dbg --//:platform=gui //:openroad` and run it myself the test passes.  It only fails from within `bazelisk test -c dbg --sandbox_debug --//:platform=gui //src/gui/test:screenshot-tcl_test`.  Any idea why?

src/gui/src/layoutViewer.cpp
```diff
@@ -2221,9 +2221,11 @@ void LayoutViewer::saveImage(const QString& filepath,
 
   QImage img = createImage(region, width_px, dbu_per_pixel);
 
-  if (!img.save(save_filepath)) {
-    logger_->warn(
-        utl::GUI, 78, "Failed to write image: {}", save_filepath.toStdString());
+  QImageWriter writer(save_filepath);
```
gadfort, line 2251: @maliberty why was this needed? We have this code in several other places

maliberty: QImage::save just return a bool pass/fail.  If you want to get why it failed you have to do it this way.

gadfort: okay

gadfort:
@maliberty this is the only error that needs to be fixed in the CI:
```
16:13:29  [INFO ODB-0227] LEF file: Nangate45/Nangate45_tech.lef, created 22 layers, 27 vias
16:13:29  [INFO ODB-0227] LEF file: Nangate45/Nangate45_stdcell.lef, created 135 library cells
16:13:29  [INFO ODB-0128] Design: gcd
16:13:29  [INFO ODB-0130]     Created 54 pins.
16:13:29  [INFO ODB-0131]     Created 482 components and 2074 component-terminals.
16:13:29  [INFO ODB-0133]     Created 385 nets and 1110 connections.
16:13:29  [WARNING GUI-0076] could not connect to display 
16:13:29  [INFO GUI-0075] Could not load the Qt platform plugin "xcb" in "" even though it was found.
16:13:29  [ERROR GUI-0077] This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
16:13:29  
16:13:29  Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.
16:13:29  
16:13:29  terminate called after throwing an instance of 'std::runtime_error'
16:13:29    what():  GUI-0077
16:13:29  Signal 6 received
16:13:29  Stack trace:
16:13:29   0# 0x000055F74B42E572 in /OpenROAD/build/bin/openroad
16:13:29   1# 0x00007F6651B3D520 in /lib/x86_64-linux-gnu/libc.so.6
16:13:29   2# pthread_kill in /lib/x86_64-linux-gnu/libc.so.6
16:13:29   3# raise in /lib/x86_64-linux-gnu/libc.so.6
16:13:29   4# abort in /lib/x86_64-linux-gnu/libc.so.6
```

PR #9237, opened 2026-01-12
Introduce fast solution for doomed DRT runs

PrecisEDAnon:
This is a PR to cut down on long-tail DRT runs (taking over hours). AI generated content follows.

# PR: `drt` — Optional “Doomed Clips” Scheduling for Long‑Tail Runtime

This PR introduces an **opt‑in** detailed routing mode that targets **long‑tail wall time** caused by a small number of expensive tiles/clips. The mode is enabled only when explicitly requested via `detailed_route -doomed_clips`.

I plan a **follow‑up PR to OpenROAD-flow-scripts (ORFS)** that adds flow‑level toggles, a benchmark harness, and a reproducible “horror story” case to demonstrate and compare baseline vs. `-doomed_clips`. This PR is **OpenROAD‑only**.

---

## Summary

- Adds a new `detailed_route` option: **`-doomed_clips`**
- When enabled, DRT collects per‑clip stats each iteration and **prioritizes the worst clips earlier** in the next iteration.
- Optional diagnostic/tuning knobs are exposed to Tcl (`help detailed_route`) and documented in `src/drt/README.md`.

**Default behavior is unchanged** unless `-doomed_clips` is set.

---

## Motivation

On some designs, detailed routing iterations become dominated by a small number of clips. This produces:

- Very low effective parallelism late in iterations (workers idle while a few clips run long).
- Large iteration‑to‑iteration runtime variance (“long tail”), sometimes driving multi‑hour routing even when most tiles are easy.

The core idea of this PR is simple: **if we can identify which clips were slow/problematic in iteration *k*, we should schedule them earlier in iteration *k+1*** so the expensive work overlaps better with other work, reducing the tail and improving thread utilization.

---

## User‑Facing Changes (Tcl / CLI)

New `detailed_route` flags:

- `-doomed_clips`
- `-doomed_clips_report_n <n>`: report top‑N worst tiles each iteration (0 disables printing)
- `-doomed_clips_min_iter <n>`: start using previous‑iteration stats at/after iteration N (default 1)
- `-doomed_clips_w_runtime <w>`, `-doomed_clips_w_drvs <w>`, `-doomed_clips_w_congestion <w>`: scoring weights
- `-doomed_clips_top_n <n>`: (optional) for top‑N scored tiles, run multiple cost variants and keep the best result (default 0/off)

Example:

```tcl
detailed_route -doomed_clips -doomed_clips_report_n 10
```

---

### 1) Per‑clip stats collection

When `-doomed_clips` is enabled, each clip records:

- runtime (seconds)
- initial marker/DRV count
- best marker/DRV count achieved
- congested flag

These are keyed by the clip’s gcell coordinate and persisted in memory for the next iteration.

### 2) Scoring + scheduling

At the start of an iteration (after `-doomed_clips_min_iter`), clips are scored using a weighted, normalized sum:

- runtime normalized by max runtime seen
- best DRVs normalized by max best DRVs seen
- optional congestion weight if the clip was flagged congested

Workers (and worker batches) are then **stable‑sorted** so higher‑scored (“worse”) clips execute earlier.

### 3) Iteration‑0 seeding (no prior stats)

For iteration 0 there are no previous‑iteration stats. When `-doomed_clips` is enabled, we seed an initial ordering using a lightweight proxy signal (boundary pin count) to avoid starting from an essentially arbitrary order.

### 4) Optional multi‑cost exploration (`-doomed_clips_top_n`)

If enabled and prior stats exist, the top‑N scored tiles can be expanded into multiple cost variants (e.g. 0.5×, 1×, 2× for selected costs). The best result is selected using:

1) lower best DRC markers, then
2) lower runtime

This is intentionally guarded (default off) because it can increase total work.

---

## Implementation Notes

Primary implementation is in DRT:

- Scheduling + scoring + optional multi‑cost: `src/drt/src/dr/FlexDR.cpp`
- Public params/plumbing:
  - `src/drt/include/drt/TritonRoute.h`
  - `src/drt/src/TritonRoute.cpp`
  - `src/drt/src/TritonRoute.i`
  - `src/drt/src/TritonRoute.tcl`
  - `src/drt/src/global.h`
- User docs: `src/drt/README.md`

---

## Backwards Compatibility / “Does this change baseline?”

- All new behavior is guarded by `-doomed_clips`.
- When the flag is not set:
  - no clip scoring/sorting is applied
  - no reporting is emitted
  - no multi‑cost exploration happens

In other words: **default routing behavior is intended to remain identical** unless the new flag is enabled.

---

## Testing

- Build: local CMake build completed successfully.
- Formatting: `clang-format` applied to touched files.
- Static analysis: `clang-tidy` run on the modified DRT sources locally.

---

## Follow‑Up (ORFS)

After this lands, I will open a separate PR to **OpenROAD-flow-scripts** that:

- Adds flow toggles (no raw arg strings required).
- Ships a reproducible long‑tail case + “how‑to” docs.
- Provides a benchmark harness to compare baseline vs. `-doomed_clips` runs from identical GRT checkpoints.

### DIFF
--- a/src/drt/src/dr/FlexDR.cpp
+++ b/src/drt/src/dr/FlexDR.cpp
@@ -804,7 +804,8 @@ void FlexDR::endWorkersBatch(
       increaseClipsize_ = true;
     }
     if (router_cfg_->DOOMED_CLIPS) {
-      const odb::Point gcell_idx = block->getGCellIdx(worker->getRouteBox().ll());
+      const odb::Point gcell_idx
+          = block->getGCellIdx(worker->getRouteBox().ll());
       const uint64_t tile_key
           = (static_cast<uint64_t>(static_cast<uint32_t>(gcell_idx.x())) << 32)
             | static_cast<uint32_t>(gcell_idx.y());
@@ -844,7 +845,8 @@ void FlexDR::endWorkersBatchSelectBest(
       increaseClipsize_ = true;
     }
     if (router_cfg_->DOOMED_CLIPS) {
-      const odb::Point gcell_idx = block->getGCellIdx(worker->getRouteBox().ll());
+      const odb::Point gcell_idx
+          = block->getGCellIdx(worker->getRouteBox().ll());
       const uint64_t tile_key
           = (static_cast<uint64_t>(static_cast<uint32_t>(gcell_idx.x())) << 32)
             | static_cast<uint32_t>(gcell_idx.y());
@@ -1382,8 +1384,8 @@ void FlexDR::optimizationFlow(const SearchRepairArgs& args,
               });
   }
 
-  const bool use_multi_cost = use_prev_stats && !dist_on_
-                              && router_cfg_->DOOMED_CLIPS_TOP_N > 0;
+  const bool use_multi_cost
+      = use_prev_stats && !dist_on_ && router_cfg_->DOOMED_CLIPS_TOP_N > 0;
   std::unordered_set<uint64_t> doomed_tiles;
   std::vector<frUInt4> drc_costs;
   std::vector<frUInt4> marker_costs;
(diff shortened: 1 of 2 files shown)


oharboe:
Is all this code for the concern of failing dommed runs?

Thoughts:

- A lot of code was added. Is this code there *only* to handle doomed runs or does this PR handle other concerns that could be broken off to a separate PR?
- There are no doomed runs under CI, how will this be tested?
- Every user option makes OpenROAD increases the cognitive load and makes OpenROAD less user friendly, can we do without options? Perhaps the C++ code just needs more work and then we can do without options?
- What is the root cause of the doomed runs. Can we fail earlier? In floorplan? Placement? CTS? grt?

rovinski:
Are there any measurements on even synthetic test cases that demonstrate the improvement? Rescheduling the tiles only helps if there are many other tiles to process, which only helps in the earlier iterations. A lot of the long tail behavior you are mentioning are from "stubborn tiles" which already have a separate flow that leverages more parallelism to speed it up.

maliberty:
Given @osamahammad21 recent improvements I think this is moot.

PR #9242, opened 2026-01-12
Mpl iterative implementation

oharboe:


oharboe:
@maliberty Is this helpful?

Just a quick Copilt attempt at making an iterative instead of recurseive implementation after seeing a stack trace that indicated that there was a performance problem in this function.

maliberty:
Did it make a runtime difference?

oharboe:
> Did it make a runtime difference?

I didn't try it, I just was tinkering with copilot and wondering if it could be helpful in rewriting this code that showed up in the early stages of MPL.

maliberty:
Then I don't know if it is helpful or not as I haven't seen the problem nor do I know if this is a solution.

PR #9244, opened 2026-01-12
Introduce tomography plumbing to OpenROAD

PrecisEDAnon:
The following content was AI generated. Note that this only adds the plumbing for soft-assignments to be used, and does not supply the model itself. You have to add it from the ORFS end.

# [GPL] Plumbing: seedable placement + external cluster import

## Summary
This PR adds two small pieces of plumbing to support multi-run “view generation” and external guidance injection (e.g., netlist tomography experiments), without shipping any model:

- `global_placement -random_seed <int>`: makes GPL’s internal RNG seed configurable (previously hardcoded to `42`; default remains `42` so behavior is unchanged unless the flag is used).
- `read_soft_placement_clusters -file <csv>`: reads an external “soft assignment” file and converts it into existing GPL `placement_cluster` constraints.

## Motivation
For A/B experiments that fuse multiple placement “views”, we need:

1) deterministic yet controllable randomization across runs (`-random_seed`), and  
2) a stable, tool-native interface to consume externally generated instance→cluster assignments.

### `global_placement -random_seed`
- New optional argument to `global_placement`.
- When provided, sets the seed used by GPL’s Nesterov engine.
- Default remains `42` (matches prior hardcoded seed).

### `read_soft_placement_clusters`
Creates one or more `placement_cluster` constraints from a CSV file.

- File format: `inst_name,cluster_id[,weight]`
  - Header is allowed (line 1 `inst_name,...` / `instance,...` / `inst,...` is ignored).
  - Blank lines and `#` comments are ignored.
  - `weight` is optional (defaults to `1.0`).
  - If an instance appears multiple times, the highest `weight` row wins.
- Options:
  - `-min_weight` (default `0.0`): drop assignments below this weight.
  - `-min_cluster_size` (default `2`): drop clusters smaller than this.
  - `-max_cluster_size` (default `50`): split clusters larger than this into chunks.
  - `-split_large_clusters`: suppress warning when splitting oversized clusters.
  - `-best_effort`: skip unknown instance names instead of erroring.

## Example
```tcl
read_soft_placement_clusters -file clusters.csv \
  -min_weight 0.5 -min_cluster_size 2 -max_cluster_size 50 -best_effort

global_placement -random_seed 12345 -timing_driven
```

## Backward compatibility
- No behavior change unless `-random_seed` and/or `read_soft_placement_clusters` are used.
- The default seed remains `42` (matching previous behavior).

## Notes
- This change is intentionally “plumbing only”: the producer of `clusters.csv` is external to OpenROAD (any valid model/heuristic can be used).

### DIFF
--- a/src/gpl/src/replace.cpp
+++ b/src/gpl/src/replace.cpp
@@ -22,7 +22,6 @@
 #include "placerBase.h"
 #include "routeBase.h"
 #include "rsz/Resizer.hh"
-#include "sta/StaMain.hh"
 #include "timingBase.h"
 #include "utl/Logger.h"
 #include "utl/validation.h"
@@ -74,7 +73,7 @@ void Replace::checkHasCoreRows()
 {
   if (!dbHasCoreRows(db_)) {
     log_->error(
-      GPL,
+        GPL,
         130,
         "No rows defined in design. Use initialize_floorplan to add rows.");
   }
--- a/src/gpl/src/replace.tcl
+++ b/src/gpl/src/replace.tcl
@@ -268,7 +268,8 @@ proc read_soft_placement_clusters { args } {
 
     set fields [split $line ","]
     if { [llength $fields] < 2 } {
-      utl::warn GPL 1622 "$file:$line_no: expected at least 2 CSV fields (inst_name,cluster_id[,weight]); skipping."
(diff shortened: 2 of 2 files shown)


maliberty:
Low interest without a lot more investigation.

Issue #9089, opened 2025-12-18
OpenROAD GUI Crash When Using the "Ruler" After Loading a 3DBX File


LeeJaKang:
### Describe the bug

After loading a 3DBX file using read_3dbx, OpenROAD terminates when the "Tools/Ruler" is clicked, producing the error log shown below.

"""
Signal 11 received
Stack trace:
 0# 0x00000000011DACA3 in openroad
 1# 0x00007F61955075B0 in /lib64/libc.so.6
 2# odb::dbBlock::getDieArea() in openroad
 3# gui::LayoutViewer::searchNearestEdge(odb::Point const&, bool, bool) in openroad
 4# gui::LayoutViewer::mouseMoveEvent(QMouseEvent*) in openroad
 5# QWidget::event(QEvent*) in /lib64/libQt5Widgets.so.5
 6# QApplicationPrivate::notify_helper(QObject*, QEvent*) in /lib64/libQt5Widgets.so.5
 7# QApplication::notify(QObject*, QEvent*) in /lib64/libQt5Widgets.so.5
 8# gui::SafeApplication::notify(QObject*, QEvent*) in openroad
 9# QCoreApplication::notifyInternal2(QObject*, QEvent*) in /lib64/libQt5Core.so.5
10# QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) in /lib64/libQt5Widgets.so.5
11# 0x00007F61980A55CA in /lib64/libQt5Widgets.so.5
12# 0x00007F61980A8A05 in /lib64/libQt5Widgets.so.5
13# QApplicationPrivate::notify_helper(QObject*, QEvent*) in /lib64/libQt5Widgets.so.5
14# QApplication::notify(QObject*, QEvent*) in /lib64/libQt5Widgets.so.5
15# gui::SafeApplication::notify(QObject*, QEvent*) in openroad
16# QCoreApplication::notifyInternal2(QObject*, QEvent*) in /lib64/libQt5Core.so.5
17# QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) in /lib64/libQt5Gui.so.5
18# QGuiApplicationPrivate::processWindowSystemEvent(QWindowSystemInterfacePrivate::WindowSystemEvent*) in /lib64/libQt5Gui.so.5
[...]

PR #9264, opened 2026-01-15
Gui: 3DBX fix ruler 

LucasYuki:
Closes issue #9089

Changes:
- Adds the method LayoutViewer::getChips().
  - Return all dbChip and dbChipInst.
- Iterate through all blocks instead of using getBlock().
- LayoutViewer::searchNearestViaEdge does the search using the block coordinates:
  - Transform the search variables to the current block coordinate system.
  - After searching the block, turn back the search variables to the global coordinate system.

### DIFF
--- a/src/gui/src/layoutViewer.cpp
+++ b/src/gui/src/layoutViewer.cpp
@@ -191,15 +191,15 @@ void LayoutViewer::setChip(odb::dbChip* chip)
   fit();
 }
 
-std::map<odb::dbChip*, odb::dbChipInst*> LayoutViewer::getChips() const
+std::map<odb::dbChipInst*, odb::dbChip*> LayoutViewer::getChips() const
 {
-  std::map<odb::dbChip*, odb::dbChipInst*> chips;
+  std::map<odb::dbChipInst*, odb::dbChip*> chips;
   if (getChip() == nullptr) {
     return chips;
   }
   std::vector<odb::dbChip*> stack;
 
-  chips[getChip()] = nullptr;
+  chips[nullptr] = getChip();
   stack.push_back(getChip());
   while (!stack.empty()) {
     odb::dbChip* curr_chip = stack.back();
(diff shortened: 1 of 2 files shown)


src/gui/src/layoutViewer.cpp
```diff
@@ -511,10 +548,8 @@ void LayoutViewer::searchNearestViaEdge(
   }
 }
 
-std::pair<LayoutViewer::Edge, bool> LayoutViewer::searchNearestEdge(
-    const odb::Point& pt,
-    bool horizontal,
-    bool vertical)
+std::pair<LayoutViewer::Edge, bool>
+LayoutViewer::searchNearestEdge(odb::Point pt, bool horizontal, bool vertical)
```
maliberty, line 552: Why switch `pt` to by-value?

maliberty:
This PR does not work for me

- The measurements seem to always be in dbu regardless of Options->Show DBU.
- It won't snap to a second die.

I tested with `openroad -gui read_3dbx.tcl` and
```
diff --git a/src/odb/test/data/example.3dbx b/src/odb/test/data/example.3dbx
index ef4a84dfb7..eb0617ad62 100644
--- a/src/odb/test/data/example.3dbx
+++ b/src/odb/test/data/example.3dbx
@@ -30,7 +30,7 @@ Stack:
     z: 0.0
     orient: R0
   soc_inst_duplicate:
-    loc: [100.0, 200.0]
+    loc: [100.0, 1300.0]
     z: 300.0
     orient: MZ
```
and couldn't snap to the upper die at all.

LucasYuki:
@maliberty 
>     * The measurements seem to always be in dbu regardless of Options->Show DBU.
Fixed by getting the dbu per micron from db instead of the block

>     * It won't snap to a second die.
I think the problem was with the top chip, testing with other files I could snap to a second chiplet.

> and couldn't snap to the upper die at all.

It was not searching the top chip if it didn't have a block. I thought we didn't draw the top chip in 3d designs, only the chiplets.

maliberty:
I don't think it is the top chip.

LucasYuki:
Sorry, I didn't apply your modifications in src/odb/test/data/example.3dbx. 

I was using a map<dbChip*, dbChipInst*>, but we can have multiple dbChipInst with the same dbChip as master.
I inverted the map, and now it should work.

maliberty:
It works for me now.  Needs some lint fixing.

LucasYuki:
Looking into the logs, I think it was only a network problem.
Can you put those tests to run again?

src/gui/src/layoutViewer.cpp
LucasYuki: Because now I'm temporally changing the value of pt to be in the dbChipInst coordinate system

PR #9267, opened 2026-01-15
odb: fix mask on special nets in DEF

gadfort:
The MASK flag was getting added incorrectly in defout.
Before:
```
FIXED metal3 10 + SHAPE RING + MASK 2 + ( 5 0 ) ( 5 10 )
```

After:
```
FIXED metal3 10 + SHAPE RING + ( 5 0 ) MASK 2 ( 5 10 )
```

The before syntax would be correct if it were:
```
FIXED metal3 10 + SHAPE RING + MASK 2 + RECT ( 5 0 ) ( 5 10 )
```

### DIFF
--- a/src/odb/test/data/design58.def
+++ b/src/odb/test/data/design58.def
@@ -60,6 +60,7 @@ SPECIALNETS 2 ;
    + FIXED + SHAPE RING + MASK 2 + RECT metal3 ( 0 0 ) ( 10 10 )
    + ROUTED metal1 2000 ( 10 0 ) MASK 3 ( 10 20 ) M2_M1_via
      NEW metal2 1000 ( 10 10 ) ( 20 10 ) MASK 1 ( 20 20 ) MASK 031 M2_M1_via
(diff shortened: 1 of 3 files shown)


stefanottili:
I get syntax errors from my own def parser in line 57 and 64.
Removing the + before the first coordinates fixes that.
```
src/odb/test/write_def58.defok
 57       + FIXED metal3 10 + SHAPE RING + ( 5 0 ) MASK 2 ( 5 10 )
...
 64       NEW metal2 1000 + SHAPE DRCFILL + ( 20 10 ) MASK 1 ( 20 20 ) ;
```

gadfort:
> I get syntax errors from my own def parsers in line 57 and 64. Removing the + before the first coordinates fixes that.
> 
> ```
> src/odb/test/write_def58.defok
>  57       + FIXED metal3 10 + SHAPE RING + ( 5 0 ) MASK 2 ( 5 10 )
> ...
>  64       NEW metal2 1000 + SHAPE DRCFILL + ( 20 10 ) MASK 1 ( 20 20 ) ;
> ```

You are correct, I'll fix it now Thanks for checking

PR #9279, opened 2026-01-17
odb: use new/delete consistently instead of mixing in malloc/free

oharboe:
malloc had to be amended with a wafer thin wrapper to throw an exception, new/delete does this out of the box.

Motivated by valgrind errors


oharboe:
@maliberty Looks like malloc/free new/delete usage is used a bit interchangably and subtly throughout the code, so endemic... Verilator protests, but I get the sense that the runtime libraries are forgiving on this point. Letting the dog lie.

PR #9280, opened 2026-01-17
gui: add separate classes for rendering legends

gadfort:
Changes:
- adds ability to create custom legends.

There is an empty parent class at the moment. I can remove it, but I think it would be nice to create a separate legend render pass, so we can "organize" them so they dont' all overlap. But this just gets the ball rolling.

Timing path:
image


src/gui/include/gui/gui.h
```diff
@@ -651,6 +651,44 @@ class SpectrumGenerator
...
+class Legend
+{
+ public:
+  Legend() = default;
+  virtual ~Legend() = default;
+
+  virtual void draw(Painter& painter) const = 0;
+};
+
+class LinearLegend : public Legend
```
maliberty, line 671: A descriptive comment would be good for LinearLegend & DiscreteLegend

src/gui/include/gui/gui.h
```diff
@@ -651,6 +651,44 @@ class SpectrumGenerator
  private:
   static const unsigned char kSpectrum[256][3];
   double scale_;
+  static constexpr int kLegendColorIncrement = 2;
+};
+
+class Legend
+{
+ public:
+  Legend() = default;
```
maliberty: move the ctor to protected (the class is abstract but if you are going to define it then the visibility should be correct)

gadfort:
@maliberty it might make sense to pin the black formatter to a version and let the github actions update the version automatically to avoid this random set of failures.

maliberty:
Agree about black but fixed for the moment (looks to be a yearly update to stable).

Issue #3536, opened 2023-06-26
ispd19_test6: global_route increases runtime from 50min to 2h45 compared to using route_guides


stefanottili:
### Describe the bug

This is a follow up to https://github.com/The-OpenROAD-Project/OpenROAD/issues/1888.

I've been running ispd19_test6 without read_guides and saw the runtime go from 50min to 2h45.
Matt suggested to run without M1, but that reduced the runtime only from 2h45 to 2h15.

global_route seems to severely overestimate the routing resources available for the std cell area.

M1/M2 power pins, M1 vertical, M2 horizontal, even at the low std cell utilization, there is really no M1/M2 routing resource available, only room for pin access.
Credits to the detailed router to be able to resolve this, but at the expense of 2.5x times the runtime compared to the route guides (from a different tool).
 
It should also be noted that in May 2022 with route guides, the drt finished in 30min.
See https://github.com/The-OpenROAD-Project/OpenROAD/issues/3535

### Expected Behavior

global_route should block all of M1 and nearly all of M2 for anything but pin access.

[log_groute_M1M9_Jun2023.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/11874874/log_groute_M1M9_Jun2023.gz)
[log_groute_M2M9_Jun2023.gz](https://github.com/The-OpenROAD-Project/OpenROAD/files/11874875/log_groute_M2M9_Jun2023.gz)

### Environment

```shell
Jun 2023 OR build --local on M1 16GB MacOS
```

### To Reproduce

read_lef ispd19_test6/ispd19_test6.input.lef.gz
read_def ispd19_test6/ispd19_test6.input.def.gz
set_thread_count 8
#set_routing_layers -signal Metal2-Metal9 -clock Metal2-Metal9
#read_guides ispd19_test6/ispd19_test6.guide 
global_route -verbose
detailed_route -output_drc 5_route_drc.rpt -verbose 1

maliberty:
Can you try 
```
set_global_routing_layer_adjustment Metal2-Metal9 0.5
```
to see if it helps?

stefanottili:
That did help, 2h50 -> 0h46 !!
[INFO DRT-0267] cpu time = 05:19:51, elapsed time = 00:46:20, memory = 6077.78 (MB), peak = 7793.19 (MB)

[...]

PR #9292, opened 2026-01-19
grt: add option to show runtime at the end of global_route execution

eder-matheus:
Partially covers the comments made here: https://github.com/The-OpenROAD-Project/OpenROAD/issues/3536#issuecomment-3765917690

This PR adds the ability to show the runtime of the global_route command. It is only enabled by using the flag -show_runtime. This is off by default for reproducibility purposes (e.g., unit tests that relies on the tool log).


maliberty:
As I commented there ORFS is already using log_cmd to report the time.  Having an off by default option doesn't seem that useful.  @stefanottili is this what you are asking for and it is still relevant?

stefanottili:
a) not everybody is using ORFS, sometimes running a P&R flow in "one script, without getting in/out of the tool" is a good thing.
b) considering that drt is very verbose with it's runtime, why not grd ?
What's different between drt and grt with regards to unit tests that rely on log files ?

eder-matheus:
> b) considering that drt is very verbose with it's runtime, why not grd ? What's different between drt and grt with regards to unit tests that rely on log files ?

GRT is less verbose than DRT, so the `-verbose` guards everything that is logged by the tool. Information such as capacity/usage per layer, final congestion, etc. are used in the unit tests to ensure the tool is not changing its behavior. Adding runtime to this flag would break our unit tests.

maliberty:
@eder-matheus we suppress runtime related messages in test/helpers.tcl for unit testing to avoid output differences (they occur elsewhere like drt).

eder-matheus:
@maliberty I believe this is ready for review/merge.

maliberty (approved):

Issue #9291, opened 2026-01-19
Strengthen 3dblox checker. Encompasses many checks.


ahmed532:
### Description

The 3DBlox checker performs a comprehensive set of validations across three main categories to ensure design integrity and correct connectivity:
1. Physical Alignment Checks
These checks verify the spatial placement and contact between components.
  * Bump Inside Region: Confirms that every bump's global coordinates (X, Y) are contained within the physical footprint of its designated parent region.
  * Non-Intersecting Connections (Rule 2/3): Ensures that all regions declared as connected via Connectivity actually exhibit physical contact or overlap in the 3D design space.
  * INTERNAL_EXT Usage: Validates that any region marked as INTERNAL_EXT is explicitly utilized in at least one Connectivity declaration. 
2.  Logical Alignment & Connectivity
The checker performs two levels of logical verification to bridge the gap between physical bumps and the Verilog netlist:
  * Master-level Alignment (Rule 2 - ODB-560)
Scope: Intra-chiplet (Master definition).
Check: Verifies that the logical_net specified for a bump in the bmap corresponds to a valid Port in the chiplet's Verilog module.
Rationale: This ensures the chiplet's physical interface (bumps) aligns with its logical interface (internal ports). It uses intra-chiplet names because the master definition is reusable across different assembly contexts.
  * Assembly-level Connectivity (Rule 4 - Open Net)
Scope: Inter-chiplet (Full assembly).
Check: Performs a Union-Find analysis to ensure all bumps assigned to the same Inter-chiplet Net in the top-level Verilog/DB are physically connected.
Rationale: This verifies that the "cross-chiplet" signals are physically supported by the bonded bumps and internal routing.
3. Structural & Design Constraints
  * Region Type Validation: Ensures that only appropriate chip types INTERPOSER utilize INTERNAL or INTERNAL_EXT regions. (Log ODB-463)

Related to: #9204 #9209 #9172

PR #9293, opened 2026-01-19
odb: implement UnfoldedModel for 3DBlox geometric flattening

ahmed532:
- Implement UnfoldedModel class to flatten 3DBlox hierarchical instances into a global geometric representation.
- Add Transform3D helper class to unify 2D/3D transformations and Z-mirroring.
- Implement orientation-aware Z-calculation for FRONT/BACK/INTERNAL regions.
- Support layer-based Z-offsets using dbTech information.
- Provide pointer-stable containers (std::deque) for efficient geometric analysis.

Related to: #9291

### DIFF
--- a/src/odb/src/db/dbScanChain.cpp
+++ b/src/odb/src/db/dbScanChain.cpp
@@ -141,13 +141,13 @@ std::variant<dbBTerm*, dbITerm*> _dbScanChain::getPin(
       ->getPin();
 }
 
-void _dbScanChain::setPin(dbId<dbScanPin> _dbScanChain::* field, dbBTerm* pin)
+void _dbScanChain::setPin(dbId<dbScanPin> _dbScanChain::*field, dbBTerm* pin)
 {
   dbDft* dft = (dbDft*) getOwner();
   this->*field = dbScanPin::create(dft, pin);
 }
 
-void _dbScanChain::setPin(dbId<dbScanPin> _dbScanChain::* field, dbITerm* pin)
+void _dbScanChain::setPin(dbId<dbScanPin> _dbScanChain::*field, dbITerm* pin)
 {
   dbDft* dft = (dbDft*) getOwner();
   this->*field = dbScanPin::create(dft, pin);
--- a/src/odb/src/db/dbScanChain.h
+++ b/src/odb/src/db/dbScanChain.h
@@ -33,8 +33,8 @@ class _dbScanChain : public _dbObject
   void collectMemInfo(MemInfo& info);
   // User Code Begin Methods
   std::variant<dbBTerm*, dbITerm*> getPin(const dbId<dbScanPin>& scan_pin_id);
-  void setPin(dbId<dbScanPin> _dbScanChain::* field, dbBTerm* pin);
-  void setPin(dbId<dbScanPin> _dbScanChain::* field, dbITerm* pin);
+  void setPin(dbId<dbScanPin> _dbScanChain::*field, dbBTerm* pin);
+  void setPin(dbId<dbScanPin> _dbScanChain::*field, dbITerm* pin);
   // User Code End Methods
 
   std::string name_;


ahmed532:
4 tests failed because of a new warning was added and .ok files were not updated correctly. I'll update them in the next commit.

src/odb/src/db/dbTransform.cpp
```diff
@@ -79,11 +79,13 @@ static const dbOrientType::Value orientMul[8][8] = {{dbOrientType::R0,
...
+// dbOrientType when needed.
 dbOStream& operator<<(dbOStream& stream, const dbTransform& t)
 {
-  stream << (int) t.orient_;
-  stream << t.offset_;
-  stream << (int) t.orient_;
+  stream << static_cast<int>(t.orient_);
   stream << t.offset_;
   return stream;
 }
```
osamahammad21, line 87: nice catch. Although I don't think we ever serialize dbTransform because we don't ever store it. I have been thinking of removing the serialization part altogether.

src/odb/src/db/dbTransform.cpp
```diff
@@ -225,20 +227,28 @@ void dbTransform::apply(Point3D& p) const
...
+  Point p[4];
+  p[0] = r.ll();
+  p[1] = r.ul();
+  p[2] = r.ur();
+  p[3] = r.lr();
+  r.mergeInit();
+  for (int i = 0; i < 4; i++) {
+    apply(p[i]);
+    r.merge(p[i]);
+  }
```
osamahammad21: why?

src/odb/src/db/dbTransform.cpp
```diff
@@ -225,20 +227,28 @@ void dbTransform::apply(Point3D& p) const
...
-  apply(uur);
-  c.init(lll.x(), lll.y(), lll.z(), uur.x(), uur.y(), uur.z());
+  std::vector<Point3D> points = c.getPoints();
+  Cuboid result;
+  result.mergeInit();
+  for (Point3D& p : points) {
+    apply(p);
+    result.merge(p);
+  }
+  c = result;
```
osamahammad21: why?

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -43,49 +45,66 @@ struct UnfoldedBump
 struct UnfoldedRegionFull : public UnfoldedRegion
```
osamahammad21: Is there a good reason for this hierarchy? It's a bit confusing why not to merge both already.

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -107,10 +128,15 @@ class UnfoldedModel
...
   utl::Logger* logger_;
+
+  // Using std::deque for pointer stability: elements maintain their addresses
+  // even when the container grows, which is critical since UnfoldedRegion,
+  // UnfoldedBump, etc. hold raw pointers to their parent structures.
   std::deque<UnfoldedChip> unfolded_chips_;
   std::deque<UnfoldedConnection> unfolded_connections_;
   std::deque<UnfoldedNet> unfolded_nets_;
   std::map<std::string, UnfoldedChip*> chip_path_map_;
+  std::unordered_map<dbChipBumpInst*, UnfoldedBump*> bump_inst_map_;
```
osamahammad21: How exactly does this work? There is no one to one mapping of a dbChipBumpInst to UnfoldedBump. If there was, we wouldn't need the unfolded model

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +402,318 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+
+        local_z = std::max(0, z_from_chip_bottom);
+
+      } else if (original_side == dbChipRegion::Side::BACK) {
+        // Back-side BEOL: stack starts at bottom of chip (Z=0)
+        auto [total_tech_thick, layer_z]
+            = getLayerZ(master_chip->getTech(), layer);
+        // Stack aligned to bottom. Layer Z is simply layer_z
+        local_z = layer_z;
+      }
```
osamahammad21: is this really needed?
If it is, then it needs to consider in the folded chiplet object. Also, I think that the layer thickness should be considered in the region's cuboid thickness.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +402,318 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+                         odb::Point(inst->getLoc().x(), inst->getLoc().y()));
+      t.apply(footprint);
     }
+    uf_region.cuboid.set_xlo(footprint.xMin());
+    uf_region.cuboid.set_xhi(footprint.xMax());
+    uf_region.cuboid.set_ylo(footprint.yMin());
+    uf_region.cuboid.set_yhi(footprint.yMax());
+    // Update Z of global cuboid to match surface_pt
+    uf_region.cuboid.set_zlo(surface_pt.z());
+    uf_region.cuboid.set_zhi(surface_pt.z());
```
osamahammad21: duplicate of 513-514

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +402,318 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+    // Global coordinates for region FOOTPRINT (XY)
+    for (auto inst : path | std::views::reverse) {
+      odb::dbTransform t(inst->getOrient().getOrientType2D(),
+                         odb::Point(inst->getLoc().x(), inst->getLoc().y()));
+      t.apply(footprint);
     }
+    uf_region.cuboid.set_xlo(footprint.xMin());
+    uf_region.cuboid.set_xhi(footprint.xMax());
+    uf_region.cuboid.set_ylo(footprint.yMin());
+    uf_region.cuboid.set_yhi(footprint.yMax());
```
osamahammad21: Why not:

```suggestion
    uf_region.cuboid = region_inst->getChipRegion()->getCuboid();
    // Global coordinates for region FOOTPRINT (XY)
    for (auto inst : path | std::views::reverse) {
      
      inst->getTransform().apply(uf_region.cuboid);
    }    
```

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +402,318 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+    auto original_side = region_inst->getChipRegion()->getSide();
+
+    // Default matching dbChipRegion::getCuboid() defaults
+    if (original_side == dbChipRegion::Side::FRONT) {
+      local_z = master_thick;
+    } else if (original_side == dbChipRegion::Side::BACK) {
+      local_z = 0;
+    } else {
+      local_z = master_thick / 2;
+    }
```
osamahammad21: This looks to be considered in `Cuboid dbChipRegion::getCuboid() const`. I think you should consider using that instead of redoing this.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,53 +3,394 @@
...
+odb::dbChipRegion::Side computeEffectiveSide(
+    odb::dbChipRegion::Side original,
+    const std::vector<odb::dbChipInst*>& path)
+{
+  bool z_flipped = false;
+  for (auto inst : path) {
+    if (inst->getOrient().isMirrorZ()) {
+      z_flipped = !z_flipped;
+    }
+  }
```
osamahammad21: Duplicate of 450-455

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,53 +3,394 @@
...
+    }
+    if (layer == target_layer) {
+      reached_target = true;
+    }
+  }
+  return {total_thickness, layer_z};
+}
+
+// Helper class to unify 3D transformations for Points and Cuboids
+class Transform3D
```
osamahammad21: What's wrong with using dbTransform directly? it handles 3d transformations correctly

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +402,318 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+    uf_region.cuboid.set_zlo(surface_pt.z());
+    uf_region.cuboid.set_zhi(surface_pt.z());
+
+    odb::Rect footprint = region_inst->getChipRegion()->getBox();
+    // Global coordinates for region FOOTPRINT (XY)
+    for (auto inst : path | std::views::reverse) {
+      odb::dbTransform t(inst->getOrient().getOrientType2D(),
+                         odb::Point(inst->getLoc().x(), inst->getLoc().y()));
+      t.apply(footprint);
     }
```
osamahammad21: I see that we are doing this quite a lot. How about having a concatenated transform for each unfolded chiplet
```
    odb::dbTransform result;
    for (auto inst : path | std::views::reverse) {
      result.concat(inst->getTransform, result);
    }
    unfolded_chip.transform = result;
```
We could then use that transform directly.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +402,318 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+
+    // Z is the region's connecting surface
+    Point3D global_pos(global_xy.x(), global_xy.y(), uf_region.getSurfaceZ());
+    uf_bump.global_position = global_pos;
+
+    // Extract logical net name from property
+    dbProperty* net_prop = dbProperty::find(bump, "logical_net");
+    if (net_prop && net_prop->getType() == dbProperty::STRING_PROP) {
+      uf_bump.logical_net_name = ((dbStringProperty*) net_prop)->getValue();
+    }
```
osamahammad21: not sure if that's useful here. The logical net is intra-chiplet net which is out of scope in this implementation.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -422,44 +456,51 @@ void ThreeDBlox::createBump(const BumpMapEntry& entry,
...
   if (entry.net_name != "-") {
     auto net = block->findNet(entry.net_name.c_str());
-    if (net == nullptr) {
-      logger_->error(utl::ODB,
-                     534,
-                     "3DBV Parser Error: Bump net {} not found",
-                     entry.net_name);
+    if (net != nullptr) {
+      bump->setNet(net);
+      inst->getITerms().begin()->connect(net);
```
osamahammad21: Is it wise to be silent in that case? Should we create the net instead same way we create the dbInst?

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -422,44 +456,51 @@ void ThreeDBlox::createBump(const BumpMapEntry& entry,
...
-      logger_->error(utl::ODB,
-                     533,
-                     "3DBV Parser Error: Bump port {} not found",
-                     entry.port_name);
-    }
-    bump->setBTerm(bterm);
-    if (bump->getNet()) {
-      bterm->connect(bump->getNet());
+    if (bterm != nullptr) {
+      bump->setBTerm(bterm);
```
osamahammad21, line 529: ditto

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +402,318 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+
+  auto bumps = region->getChipBumps();
+
+  // Pre-map bump instances to their definitions for O(1) lookup
+  std::unordered_map<dbChipBump*, dbChipBumpInst*> bump_to_inst;
+  for (auto* bi : uf_region.region_inst->getChipBumpInsts()) {
+    bump_to_inst[bi->getChipBump()] = bi;
+  }
+
+  for (auto* bump : bumps) {
```
osamahammad21:
```suggestion
  for (auto* bump_inst : uf_region.region_inst->getChipBumpInsts()) {
    dbChipBump* bump = bump_inst->getChipBump();
```

src/odb/src/3dblox/unfoldedModel.h
ahmed532: Agreed. I have merged UnfoldedRegionFull into UnfoldedRegion to simplify the structure. All members (like bumps) and methods are now part of the single UnfoldedRegion struct.

src/odb/src/db/dbTransform.cpp
ahmed532: Even with Manhattan geometry (90° rotations), transforming just the min/max points breaks the bounding box invariants. For example, a 90° rotation transforms the 'Max' point `(10, 5)` into `(-5, 10)`. If we simply assigned this to the new max, we'd end up with `x_min=0` and `x_max=-5`, which is an invalid box. Iterating all corners and using `merge()` relies on the standard API to automatically sort coordinates and ensure a valid bounding box, avoiding complex `switch` logic to handle coordinate swapping for every orientation. I have, however, updated the loop syntax to a range-based `for` loop as requested by the linter.

The Example: 90-Degree Rotation

   1. Original Box: Min (0,0) to Max (10, 5).
       * x_min=0, x_max=10
       * y_min=0, y_max=5

   2. Rotation: 90 degrees counter-clockwise ($x \to -y, y \to x$).

   3. Transforming only Min/Max:
       * Transform Min(0,0) $\to$ (0,0)
       * Transform Max(10,5) $\to$ (-5, 10)
       * Result:
           * New x_min = 0
           * New x_max = -5
       * ERROR: x_min is greater than x_max. The box is mathematically invalid (negative width).

src/odb/src/3dblox/unfoldedModel.h
ahmed532: It's true that across the entire hierarchy there isn't a simple 1-to-1 mapping. However, `bump_inst_map_` maps a specific *instance* of a bump (`dbChipBumpInst`) back to its corresponding *unfolded representation* (`UnfoldedBump`) within the current unfolded model context. This is crucial for the checker implementation—specifically for the connectivity and alignment checks—which need to look up the global geometry (the `UnfoldedBump`) starting from a `dbChipBumpInst` found in a net. I verified this against the `odb-3dblox-verification-suite` branch and confirmed that this map is essential for the full checker logic, so I have retained it.

src/odb/src/3dblox/unfoldedModel.cpp
ahmed532: The `getTransform` helper serves a specific, narrow purpose: it bridges the `dbChipInst` API (which separates 2D orientation, 2D location, and special Z-mirroring logic) into a standard `dbTransform` object. While `dbTransform` handles 3D, constructing it correctly from a `dbChipInst` requires handling the Z-offset logic (especially for mirrored chips) consistently. I've kept this as a small static helper to encapsulate that specific construction logic, but the rest of the code now uses standard `dbTransform` operations.

src/odb/src/3dblox/unfoldedModel.cpp
ahmed532: That is an excellent optimization. I have implemented it by adding a `dbTransform transform` member to the `UnfoldedChip` struct. Now, in `buildUnfoldedChip`, I calculate the full global transform (accumulating from root to leaf) exactly once and cache it. This cached transform is then reused everywhere—for calculating region cuboids and for unfolding bumps—which makes the code much cleaner and more efficient.

src/odb/src/3dblox/unfoldedModel.cpp
ahmed532: You are completely right. The previous logic was manually recalculating things that the database already knows. I have refactored `buildUnfoldedChip` to rely directly on `region_inst->getChipRegion()->getCuboid()`. By taking this authoritative local geometry and applying the cached global `transform` (implemented per your previous suggestion), we get the correct global coordinates automatically. This eliminated the redundant manual Z-stack calculations and the duplicated code you flagged.

src/odb/src/3dblox/unfoldedModel.cpp
ahmed532: I understand. I've left the logical net extraction in place for now mainly because it is used by the `checkLogicalAlignment` function in the full verification suite (which checks if the logical net in the bump map matches the Verilog). While not strictly needed for physical connectivity, it is valuable for the broader validation flow.

src/odb/src/3dblox/3dblox.cpp
ahmed532: That is a much better approach. I have updated `createBump` to be proactive: if the `dbNet` specified in the 3DBlox definition is missing in the block, the parser now automatically creates it and logs an `INFO` message (ID 534) to inform the user, instead of failing or remaining silent.

src/odb/src/db/dbTransform.cpp
osamahammad21: that's already handled in Rect::init implementation. It assumes the given x and y ranges are unsorted and handles it correctly.

ahmed532: OK, I will revert the change.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -78,6 +80,8 @@ void ThreeDBlox::readDbv(const std::string& dbv_file)
 
 void ThreeDBlox::readDbx(const std::string& dbx_file)
 {
+  std::string full_path = std::filesystem::absolute(dbx_file).string();
```
osamahammad21, line 84: not sure why absolute. not a big deal tho.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -193,6 +197,10 @@ void ThreeDBlox::calculateSize(dbChip* chip)
 {
   Cuboid cuboid;
   cuboid.mergeInit();
+  if (chip->getWidth() > 0 && chip->getHeight() > 0) {
+    cuboid.merge(Cuboid(
+        0, 0, 0, chip->getWidth(), chip->getHeight(), chip->getThickness()));
+  }
```
osamahammad21: Did you find any case where this is actually useful? We call calculateSize on the top chiplet which is the defined by the design name in dbx file. The design doesn't specify a width nor height, so it should never be useful. If it is then there is something wrong.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -335,8 +357,7 @@ void ThreeDBlox::createChiplet(const ChipletDef& chiplet)
 
   chip->setOffset(Point(chiplet.offset.x * db_->getDbuPerMicron(),
                         chiplet.offset.y * db_->getDbuPerMicron()));
-  if (chip->getChipType() != dbChip::ChipType::HIER
-      && chip->getBlock() == nullptr) {
+  if (chip->getBlock() == nullptr) {
```
osamahammad21: you should never create a block for a HIER chiplet. please revert

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -373,7 +394,19 @@ void ThreeDBlox::createRegion(const ChipletRegion& region, dbChip* chip)
...
+    }
+    if (!layer) {
+      logger_->warn(utl::ODB,
+                    563,
+                    "3DBV Parser: Layer {} not found in tech.",
+                    region.layer);
+    }
+  } else {
+    logger_->warn(
+        utl::ODB, 206, "Region {} has no layer specified.", region.name);
```
osamahammad21: I would remove that warning for now. There is no enforcement to specify the layer for a region currently.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -422,44 +456,81 @@ void ThreeDBlox::createBump(const BumpMapEntry& entry,
...
   auto bump = dbChipBump::create(chip_region, inst);
+
+  if (entry.port_name != "-" && !entry.port_name.empty()) {
+    odb::dbStringProperty::create(
+        bump, "logical_port", entry.port_name.c_str());
+  }
+  if (entry.net_name != "-" && !entry.net_name.empty()) {
+    odb::dbStringProperty::create(bump, "logical_net", entry.net_name.c_str());
+  }
+
```
osamahammad21: Are these needed now? Your last change makes sure that a dbChipBump holds all the related port and net information directly.

osamahammad21: Also do you know when would the entry.port_name or entry.net_name ever occur? I don't think we should allow this in the parser.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -422,44 +456,81 @@ void ThreeDBlox::createBump(const BumpMapEntry& entry,
...
-                     534,
-                     "3DBV Parser Error: Bump net {} not found",
-                     entry.net_name);
+      net = dbNet::create(block, entry.net_name.c_str());
+      logger_->info(utl::ODB,
+                    534,
+                    "Creating missing net {} for bump {}",
+                    entry.net_name,
+                    entry.bump_inst_name);
     }
```
osamahammad21: This could mess up the log very quickly. Could we make it a debugPrint instead?

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -422,44 +456,81 @@ void ThreeDBlox::createBump(const BumpMapEntry& entry,
...
-                     533,
-                     "3DBV Parser Error: Bump port {} not found",
-                     entry.port_name);
+      if (net != nullptr) {
+        bterm = dbBTerm::create(net, entry.port_name.c_str());
+        logger_->info(utl::ODB,
+                      533,
+                      "Creating missing port {} for bump {}",
+                      entry.port_name,
+                      entry.bump_inst_name);
```
osamahammad21: ditto

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -422,44 +456,81 @@ void ThreeDBlox::createBump(const BumpMapEntry& entry,
...
 }
 
 dbChip* ThreeDBlox::createDesignTopChiplet(const DesignDef& design)
 {
-  dbChip* chip
-      = dbChip::create(db_, nullptr, design.name, dbChip::ChipType::HIER);
+  dbChip* chip = db_->findChip(design.name.c_str());
+  if (chip == nullptr) {
+    chip = dbChip::create(db_, nullptr, design.name, dbChip::ChipType::HIER);
+  }
```
osamahammad21: Could a chip with the same name as the design name ever exist? I think we should catch that and mark it as an error.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -492,8 +563,17 @@ void ThreeDBlox::createChipInst(const ChipletInst& chip_inst)
...
                        chip_inst.z * db_->getDbuPerMicron()));
+  if (!chip_inst.external.verilog_file.empty()) {
+    if (odb::dbProperty::find(chip, "verilog_file") == nullptr) {
+      std::string verilog_file = chip_inst.external.verilog_file;
+      if (std::filesystem::path(verilog_file).is_relative()) {
+        verilog_file = std::filesystem::absolute(verilog_file).string();
+      }
+      odb::dbStringProperty::create(chip, "verilog_file", verilog_file.c_str());
+    }
+  }
```
osamahammad21: I know that the standard states that the verilog file exists under the chiplet instantiation part, but it actually should be under the chiplet defintion. There should never be 2 chipInsts of the same chiplet definition with different verilog files. Please move to the chip creation.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+  unfolded_chip.transform = total_transform;
+  total_transform.apply(unfolded_chip.cuboid);
+
+  bool z_flipped = false;
+  for (auto inst : path) {
+    if (inst->getOrient().isMirrorZ()) {
+      z_flipped = !z_flipped;
     }
+  }
+  unfolded_chip.z_flipped = z_flipped;
```
osamahammad21: why is this useful? You could get that information from the total_transform now

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,53 +3,351 @@
...
+      key += delimiter;
+    }
+    key += chip_inst->getName();
+  }
+  return key;
+}
+
+odb::dbChipRegion::Side computeEffectiveSide(
+    odb::dbChipRegion::Side original,
+    const std::vector<odb::dbChipInst*>& path)
```
osamahammad21: I fear that using odb::dbChipRegion::Side for this function is going to be very confusing very quickly. The front and back sides are relative to the chiplet. They should not be mapped or used to infer up and down. We could have a separate enum for that.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,53 +3,351 @@
...
+    case odb::dbChipRegion::Side::BACK:
+      return odb::dbChipRegion::Side::FRONT;
+    default:
+      return original;
+  }
+}
+
+odb::Cuboid computeConnectionCuboid(const odb::UnfoldedRegion& top,
+                                    const odb::UnfoldedRegion& bottom,
+                                    int thickness)
```
osamahammad21: unused?!

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+      unfolded_chip.cuboid.merge(sub_local_cuboid);
     }
   } else {
-    // Original logic: Leaf chip cuboid calculation
-    unfolded_chip.cuboid = chip_inst->getMasterChip()->getCuboid();
-    for (dbChipInst* inst :
-         unfolded_chip.chip_inst_path | std::views::reverse) {
-      inst->getTransform().apply(unfolded_chip.cuboid);
+    unfolded_chip.cuboid = master_chip->getCuboid();
+  }
```
osamahammad21: Why do you store the hierarchical chiplet in the unfolded model now? This doesn't look correct.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
   }
   unfoldConnections(chip);
   unfoldNets(chip);
 }
 
-void UnfoldedModel::unfoldChip(dbChipInst* chip_inst,
-                               UnfoldedChip& unfolded_chip)
+UnfoldedChip* UnfoldedModel::buildUnfoldedChip(dbChipInst* chip_inst,
+                                               std::vector<dbChipInst*>& path,
+                                               Cuboid& local_cuboid)
```
osamahammad21: local_cuboid?!

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,53 +3,351 @@
...
+odb::dbTransform getTransform(odb::dbChipInst* inst)
+{
+  int z_offset = inst->getLoc().z();
+  if (inst->getOrient().isMirrorZ()) {
+    z_offset += inst->getMasterChip()->getThickness();
+  }
+  return odb::dbTransform(
+      inst->getOrient(),
+      odb::Point3D(inst->getLoc().x(), inst->getLoc().y(), z_offset));
+}
```
osamahammad21: I don't know where should this be more useful than the total transform calculated from concatenating the transforms

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -22,70 +24,86 @@ class dbChipConn;
...
-  UnfoldedChip* parent_chip = nullptr;
-};
+struct UnfoldedRegion;
 
 struct UnfoldedBump
 {
-  dbChipBumpInst* bump_inst;
-  UnfoldedRegion* parent_region;
+  dbChipBumpInst* bump_inst;      // non-owning, managed by dbChip
+  UnfoldedRegion* parent_region;  // non-owning, points to UnfoldedChip::regions
```
osamahammad21: set to nullptr

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -22,70 +24,86 @@ class dbChipConn;
...
   Point3D global_position;
   std::string logical_net_name;
   std::string port_name;
 };
 
-struct UnfoldedRegionFull : public UnfoldedRegion
+struct UnfoldedRegion
 {
+  dbChipRegionInst* region_inst;  // non-owning, managed by dbChip
```
osamahammad21: ditto

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -22,70 +24,86 @@ class dbChipConn;
...
 struct UnfoldedConnection
 {
-  dbChipConn* connection;
-  UnfoldedRegionFull* top_region;
-  UnfoldedRegionFull* bottom_region;
+  dbChipConn* connection;  // non-owning, managed by dbChip
+  UnfoldedRegion*
+      top_region;  // non-owning, may be null for virtual connections
+  UnfoldedRegion*
+      bottom_region;  // non-owning, may be null for virtual connections
```
osamahammad21: ditto

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -22,70 +24,86 @@ class dbChipConn;
...
+  dbBTerm* bterm = nullptr;  // non-owning, managed by dbBlock
 
   bool isValid() const;
 };
 
 struct UnfoldedNet
 {
-  dbChipNet* chip_net;
-  std::vector<UnfoldedBump*> connected_bumps;
+  dbChipNet* chip_net;  // non-owning, managed by dbChip
```
osamahammad21: ditto

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+  // GLOBAL cuboid for the UnfoldedChip object
+  // Calculate total transform for the path
+  dbTransform total_transform;  // Identity
+  for (auto inst : path | std::views::reverse) {
+    dbTransform t = getTransform(inst);
+    // total = t * total
+    total_transform.concat(t, total_transform);
+  }
+  unfolded_chip.transform = total_transform;
+  total_transform.apply(unfolded_chip.cuboid);
```
osamahammad21:
```suggestion
  for (auto inst : path | std::views::reverse) {
    unfolded_chip.transform.concat(inst->getTransform);
  }
  unfolded_chip.transform.apply(unfolded_chip.cuboid);
```

Again, I don't think there should be something wrong with inst->getTransform in terms of transformation across the z coordinates. if there is, then we should fix it.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+      z_flipped = !z_flipped;
     }
+  }
+  unfolded_chip.z_flipped = z_flipped;
+
+  // Process Regions
+  for (auto* region_inst : chip_inst->getRegions()) {
+    UnfoldedRegion uf_region;
+    uf_region.region_inst = region_inst;
+    uf_region.parent_chip = nullptr;  // Set later
```
osamahammad21: not needed

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
 
-  unfolded_chip.chip_inst_path.pop_back();
+  unfolded_chips_.push_back(unfolded_chip);
+  UnfoldedChip* stable_chip_ptr = &unfolded_chips_.back();
+  for (auto& region : stable_chip_ptr->regions) {
+    region.parent_chip = stable_chip_ptr;
+    stable_chip_ptr->region_map[region.region_inst] = &region;
+    for (auto& bump : region.bumps) {
+      bump.parent_region = &region;
+      bump_inst_map_[bump.bump_inst] = &bump;
```
osamahammad21: I still think that that bump_inst_map_ should be handled the same way region_map is handled. It should be unique per region, not globally.

src/odb/src/3dblox/unfoldedModel.cpp
osamahammad21: Comment still not addressed.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+      continue;
+    }
+    Point local_xy = bump_inst->getLocation();
+
+    Point global_xy_pt = local_xy;
+    transform.apply(global_xy_pt);
+
+    Point3D global_pos(
+        global_xy_pt.x(), global_xy_pt.y(), uf_region.getSurfaceZ());
+    uf_bump.global_position = global_pos;
```
osamahammad21:
```suggestion
    Point location = bump_inst->getLocation();

    transform.apply(location);

    uf_bump.global_position = Point3D(location, uf_region.getSurfaceZ()) ;
```

src/odb/src/3dblox/unfoldedModel.cpp
osamahammad21: Ok. But just note that the net defined in the top verilog of the design should be different from that net. Meaning that the net defined in the top design verilog is the inter-chiplet net, while the net in the bump map is the intra-chiplet net inside that chip.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+    if (!top_region && !bottom_region) {
+      continue;  // Skip totally virtual/unresolved
+    }
+
+    UnfoldedConnection uf_conn;
+    uf_conn.connection = conn;
+    uf_conn.top_region = top_region;
+    uf_conn.bottom_region = bottom_region;
+
+    uf_conn.is_bterm_connection = false;
```
osamahammad21: Could you explain more on this variable? it's always false.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+          *top_region, *bottom_region, conn->getThickness());
+    }
+
+    // Mark used for INTERNAL_EXT check
+    if (top_region && top_region->isInternalExt()) {
+      top_region->isUsed = true;
+    }
+    if (bottom_region && bottom_region->isInternalExt()) {
+      bottom_region->isUsed = true;
+    }
```
osamahammad21: Could you explain more on the is_used variable? it is set but never used.

osamahammad21: Also, I don't think it could/should ever happen that a connection involves an internal_ext region

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -94,23 +112,29 @@ class UnfoldedModel
...
-  void unfoldChip(dbChipInst* chip_inst, UnfoldedChip& unfolded_chip);
-  void unfoldBumps(UnfoldedRegionFull& uf_region,
-                   const std::vector<dbChipInst*>& path);
+  UnfoldedChip* buildUnfoldedChip(dbChipInst* chip_inst,
+                                  std::vector<dbChipInst*>& path,
+                                  Cuboid& local_cuboid);
+  void unfoldBumps(UnfoldedRegion& uf_region, const dbTransform& transform);
   void unfoldConnections(dbChip* chip);
   void unfoldConnectionsRecursive(dbChip* chip,
                                   const std::vector<dbChipInst*>& parent_path);
```
osamahammad21: Could we merge?

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -61,81 +359,251 @@ UnfoldedModel::UnfoldedModel(utl::Logger* logger, dbChip* chip)
...
+
+    for (uint32_t i = 0; i < net->getNumBumpInsts(); i++) {
+      std::vector<dbChipInst*> path;
+      dbChipBumpInst* bump_inst = net->getBumpInst(i, path);
+
+      auto it = bump_inst_map_.find(bump_inst);
+      if (it != bump_inst_map_.end()) {
+        uf_net.connected_bumps.push_back(it->second);
+      }
+    }
```
osamahammad21, line 250: Something is wrong here. path is not used which doesn't make sense.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,53 +3,351 @@
...
+{
+  return effective_side == dbChipRegion::Side::INTERNAL;
+}
+
+bool UnfoldedRegion::isInternalExt() const
+{
+  return effective_side == dbChipRegion::Side::INTERNAL_EXT;
 }
 
 bool UnfoldedConnection::isValid() const
```
osamahammad21: I think this should be moved to the checker. We should keep the unfoldedModel for data storage only and push the logic to the checker.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,53 +3,351 @@
...
+      return false;
+    }
+  }
+  return true;
 }
 
 std::vector<UnfoldedBump*> UnfoldedNet::getDisconnectedBumps(
+    utl::Logger* logger,
+    const std::deque<UnfoldedConnection>& connections,
     int bump_pitch_tolerance) const
```
osamahammad21: ditto

src/odb/src/3dblox/3dblox.cpp
ahmed532: I initially removed calculateSize, but it caused regressions in the hierarchical connectivity tests (odb.check_3dblox_connectivity_hier and odb.check_3dblox_connectivity_recursive_nested).

The issue is that in hierarchical flows, readDbx is called for sub-designs (e.g., a MidChip.3dbx included in a top design). While the top design might not specify dimensions, the sub-chiplets need their dbChip dimensions calculated from their internal instances so that:
   1. Hierarchical Placement: The parent design can correctly transform and place instances of these chiplets.
   2. User Scripts/Tests: Scripts that rely on the database (like the failing tests) use $master getWidth/getHeight to calculate displacements. Without this call, sub-chiplets like MidChip return 0, so the test fails to move instances and never triggers the expected connectivity errors.
   3. Database Integrity: It ensures that all dbChip objects in the database have a valid bounding box based on their content, even if not explicitly defined in the .dbx.

osamahammad21: I think you misunderstood my comment. I am not saying we should get rid of calculateSize, I am saying that we should get rid of the check for the width and height. When we call readDbx for internal chiplets, we still set those chiplets to be the top chiplets temporarily. They shouldn't have any width or height, because they are the top design at this point.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -351,32 +372,41 @@ void ThreeDBlox::createChiplet(const ChipletDef& chiplet)
...
 {
   dbTechLayer* layer = nullptr;
   if (!region.layer.empty()) {
-    // TODO: add layer
+    dbTech* tech = chip->getTech();
+    if (tech) {
+      layer = tech->findLayer(region.layer.c_str());
+    }
+  }
+  dbTechLayer* layer_to_pass = (chip->getBlock() != nullptr) ? layer : nullptr;
```
osamahammad21: why do you need the block to exist for setting the region's layer?

src/odb/src/3dblox/3dblox.cpp
osamahammad21: not addressed yet

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -489,11 +554,34 @@ void ThreeDBlox::createChipInst(const ChipletInst& chip_inst)
...
+  // Z should be "low-point" based (starts at loc.z) to avoid overlap in
+  // stacking.
+  Point3D adjustment(master_cuboid.lll().x(),
+                     master_cuboid.lll().y(),
+                     master_cuboid.lll().z());
+  Point3D adjusted_loc(
+      chip_inst.loc.x * db_->getDbuPerMicron() + adjustment.x(),
+      chip_inst.loc.y * db_->getDbuPerMicron() + adjustment.y(),
+      chip_inst.z * db_->getDbuPerMicron());
+  inst->setLoc(adjusted_loc);
```
osamahammad21: why? location should be store as it is given by the 3dbx. It represents the lower left low corner of the instance.

src/odb/include/odb/db.h
```diff
@@ -7369,6 +7369,8 @@ class dbChipInst : public dbObject
 
   void setOrient(dbOrientType3D orient);
 
+  void setOrigin(const Point3D& origin);
```
osamahammad21: unneeded

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -317,6 +339,9 @@ void ThreeDBlox::createChiplet(const ChipletDef& chiplet)
     chip = dbChip::create(
         db_, tech, chiplet.name, getChipType(chiplet.type, logger_));
   }
+  if (tech != nullptr) {
+    odb::dbStringProperty::create(chip, "3dblox_tech", tech->getName().c_str());
+  }
```
osamahammad21: unneeded

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -398,10 +423,21 @@ void ThreeDBlox::createRegion(const ChipletRegion& region, dbChip* chip)
...
+  dbTechLayer* layer_to_pass = (chip->getBlock() != nullptr) ? layer : nullptr;
+  dbChipRegion* chip_region
+      = dbChipRegion::create(chip,
+                             region.name,
+                             getChipRegionSide(region.side, logger_),
+                             layer_to_pass);
+  if (layer != nullptr && layer_to_pass == nullptr) {
+    odb::dbStringProperty::create(
+        chip_region, "3dblox_layer", layer->getName().c_str());
```
osamahammad21: unneeded

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -514,11 +584,21 @@ void ThreeDBlox::createChipInst(const ChipletInst& chip_inst)
...
+      static_cast<int>(chip_inst.z * db_->getDbuPerMicron()),
+  });
+
+  if (!chip_inst.external.verilog_file.empty()) {
+    if (odb::dbProperty::find(chip, "verilog_file") == nullptr) {
+      std::string verilog_file = chip_inst.external.verilog_file;
+      verilog_file = std::filesystem::absolute(verilog_file).string();
+      odb::dbStringProperty::create(chip, "verilog_file", verilog_file.c_str());
+    }
+  }
```
osamahammad21: remove.

src/odb/src/3dblox/dbxWriter.cpp
```diff
@@ -78,8 +80,9 @@ void DbxWriter::writeStackInstance(YAML::Node& stack_instance_node,
                                    odb::dbChipInst* inst)
 {
   const double u = inst->getDb()->getDbuPerMicron();
-  const double loc_x = inst->getLoc().x() / u;
-  const double loc_y = inst->getLoc().y() / u;
+  const Point3D origin = inst->getTransform().getOffset3D();
+  const double loc_x = origin.x() / u;
+  const double loc_y = origin.y() / u;
```
osamahammad21: unneeded.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
+  uf_chip.name = getFullPathName(path);
+  uf_chip.chip_inst_path = path;
+  uf_chip.transform = total;
+
+  if (master->getChipType() == dbChip::ChipType::HIER) {
+    uf_chip.cuboid.mergeInit();
+    for (auto* sub : master->getChipInsts()) {
+      Cuboid sub_local;
+      buildUnfoldedChip(sub, path, total, sub_local);
+      uf_chip.cuboid.merge(sub_local);
```
osamahammad21: We should not consider the hierarchical chiplet in the unfolded model.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
+  unfoldConnections(chip, {});
+  unfoldNets(chip, {});
 }
 
-void UnfoldedModel::unfoldChip(dbChipInst* chip_inst,
-                               UnfoldedChip& unfolded_chip)
+UnfoldedChip* UnfoldedModel::buildUnfoldedChip(dbChipInst* inst,
+                                               std::vector<dbChipInst*>& path,
+                                               const dbTransform& parent_xform,
+                                               Cuboid& local)
```
osamahammad21: unneeded

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
   }
-  unfoldConnections(chip);
-  unfoldNets(chip);
+  unfoldConnections(chip, {});
+  unfoldNets(chip, {});
 }
 
-void UnfoldedModel::unfoldChip(dbChipInst* chip_inst,
-                               UnfoldedChip& unfolded_chip)
+UnfoldedChip* UnfoldedModel::buildUnfoldedChip(dbChipInst* inst,
```
osamahammad21, line 81: inst could be inferred from the path

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
+    }
+  }
+  chip_path_map_[chip.chip_inst_path] = &chip;
 }
 
-UnfoldedChip* UnfoldedModel::findUnfoldedChip(
-    const std::vector<dbChipInst*>& path)
+void UnfoldedModel::unfoldRegions(UnfoldedChip& uf_chip,
+                                  dbChipInst* inst,
+                                  const dbTransform& transform)
```
osamahammad21: unneeded.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
+      bump_inst_map_[{bump.bump_inst, chip.chip_inst_path}] = &bump;
+    }
+  }
+  chip_path_map_[chip.chip_inst_path] = &chip;
 }
 
-UnfoldedChip* UnfoldedModel::findUnfoldedChip(
-    const std::vector<dbChipInst*>& path)
+void UnfoldedModel::unfoldRegions(UnfoldedChip& uf_chip,
+                                  dbChipInst* inst,
```
osamahammad21: unneeded.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
-  return nullptr;
+  for (auto* region_inst : inst->getRegions()) {
+    auto side = region_inst->getChipRegion()->getSide();
+    if (transform.isMirrorZ()) {
+      side = mirrorSide(side);
+    }
+
+    UnfoldedRegion uf_region;
+    uf_region.region_inst = region_inst;
+    uf_region.effective_side = side;
```
osamahammad21, line 166: use a different enum for representing top/bottom.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
-void UnfoldedModel::unfoldBumps(UnfoldedRegionFull& uf_region,
-                                const std::vector<dbChipInst*>& path)
+void UnfoldedModel::registerUnfoldedChip(UnfoldedChip& chip)
 {
+  for (auto& region : chip.regions) {
+    region.parent_chip = &chip;
+    chip.region_map[region.region_inst] = &region;
+    for (auto& bump : region.bumps) {
+      bump.parent_region = &region;
+      bump_inst_map_[{bump.bump_inst, chip.chip_inst_path}] = &bump;
```
osamahammad21, line 128: to be changed for better efficiency in future PRs.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,252 @@
...
 {
+  for (auto* conn : chip->getChipConns()) {
+    UnfoldedRegion* top = findUnfoldedRegion(
+        findUnfoldedChip(concatPath(parent_path, conn->getTopRegionPath())),
+        conn->getTopRegion());
+    UnfoldedRegion* bot = findUnfoldedRegion(
+        findUnfoldedChip(concatPath(parent_path, conn->getBottomRegionPath())),
+        conn->getBottomRegion());
+
+    if (top || bot) {
```
osamahammad21, line 221: unlikely to trigger else.

src/odb/src/3dblox/unfoldedModel.h
```diff
@@ -22,63 +25,62 @@ class dbChipConn;
...
-  std::vector<UnfoldedBump*> getDisconnectedBumps(
-      int bump_pitch_tolerance) const;
 };
 
 struct UnfoldedChip
 {
-  std::string getName() const;
-  std::string getPathKey() const;
+  std::string name;
+  bool isParentOf(const UnfoldedChip* other) const;
```
osamahammad21: unused

src/odb/src/db/dbChipInst.cpp
```diff
@@ -148,6 +148,12 @@ void dbChipInst::setOrient(dbOrientType3D orient)
   obj->orient_ = orient;
 }
 
+void dbChipInst::setOrigin(const Point3D& origin)
+{
+  _dbChipInst* obj = (_dbChipInst*) this;
+  obj->origin_ = origin;
+}
+
```
osamahammad21: remove.

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -141,16 +144,71 @@ dbSet<dbChipBump> dbChipRegion::getChipBumps() const
...
+    dbTech* tech = layer->getTech();
+    int total_tech_thick = 0;
+    int layer_z = 0;  // Top of layer relative to stack bottom
+    bool reached_target = false;
+    uint32_t target_layer_thick = 0;
+
+    if (tech) {
+      for (dbTechLayer* l : tech->getLayers()) {
+        auto type = l->getType();
+        if (type == dbTechLayerType::ROUTING || type == dbTechLayerType::CUT) {
```
osamahammad21: unneeded check.

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -141,16 +144,71 @@ dbSet<dbChipBump> dbChipRegion::getChipBumps() const
...
+            total_tech_thick += thick;
+            if (!reached_target) {
+              layer_z += thick;
+            }
+          }
+        }
+        if (l == layer) {
+          reached_target = true;
+          layer->getThickness(target_layer_thick);
+        }
```
osamahammad21: rewrite/cleanup

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -141,16 +144,71 @@ dbSet<dbChipBump> dbChipRegion::getChipBumps() const
...
   } else {
-    z = getChip()->getThickness() / 2;
+    // Default logic if no layer
+    const auto side = getSide();
+    if (side == dbChipRegion::Side::FRONT) {
+      z_top = z_bot = chip_thickness;
+    } else if (side == dbChipRegion::Side::BACK) {
+      z_top = z_bot = 0;
+    } else {
+      z_top = z_bot = chip_thickness / 2;
```
osamahammad21: Suggestion: we could make z_bot = 0 and z_top = chip_thickness to establish connection with the front and back regions.

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -169,13 +227,33 @@ dbTechLayer* dbChipRegion::getLayer() const
 {
   _dbChipRegion* obj = (_dbChipRegion*) this;
   dbChip* chip = (dbChip*) obj->getOwner();
-  if (chip->getBlock() == nullptr || chip->getBlock()->getTech() == nullptr
-      || obj->layer_ == 0) {
+
+  dbTech* tech = nullptr;
+  if (chip->getBlock() != nullptr) {
+    tech = chip->getBlock()->getTech();
+  }
```
osamahammad21: unneeded.

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -169,13 +227,33 @@ dbTechLayer* dbChipRegion::getLayer() const
...
+  if (tech == nullptr) {
     return nullptr;
   }
 
-  _dbTech* tech = (_dbTech*) chip->getBlock()->getTech();
-  return (dbTechLayer*) tech->layer_tbl_->getPtr(obj->layer_);
+  dbTechLayer* layer = nullptr;
+  if (obj->layer_ != 0) {
+    _dbTech* _tech = (_dbTech*) tech;
+    layer = (dbTechLayer*) _tech->layer_tbl_->getPtr(obj->layer_);
```
osamahammad21: Suggestion: use dbTechLayer::getTechLayer

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -169,13 +227,33 @@ dbTechLayer* dbChipRegion::getLayer() const
...
+  }
+
+  if (layer == nullptr) {
+    if (auto prop
+        = odb::dbStringProperty::find((dbChipRegion*) this, "3dblox_layer")) {
+      layer = tech->findLayer(prop->getValue().c_str());
+    }
+  }
```
osamahammad21: unneeded

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -447,44 +475,75 @@ void ThreeDBlox::createBump(const BumpMapEntry& entry,
...
+      }
     }
   }
 }
 
 dbChip* ThreeDBlox::createDesignTopChiplet(const DesignDef& design)
 {
-  dbChip* chip
-      = dbChip::create(db_, nullptr, design.name, dbChip::ChipType::HIER);
+  dbChip* chip = db_->findChip(design.name.c_str());
```
osamahammad21: Why find?

src/odb/src/3dblox/unfoldedModel.cpp
osamahammad21: not resolved

src/odb/src/3dblox/unfoldedModel.cpp
osamahammad21: not resolved

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,283 @@
...
+    Cuboid merged_cuboid;
+    merged_cuboid.mergeInit();
+    for (auto* sub : master->getChipInsts()) {
+      Cuboid sub_local;
+      buildUnfoldedChip(sub, path, total, sub_local);
+      merged_cuboid.merge(sub_local);
     }
-    name += chip_inst->getName();
+    local = merged_cuboid;
+    inst_xform.apply(local);
```
osamahammad21: Again the merged_cuboid/local should be removed. They are not needed.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,283 @@
...
+  // Transform cuboid to global space
+  uf_chip.transform.apply(uf_chip.cuboid);
+  unfoldRegions(uf_chip, inst);
+
+  unfolded_chips_.push_back(std::move(uf_chip));
+  UnfoldedChip* created_chip = &unfolded_chips_.back();
+  registerUnfoldedChip(*created_chip);
+
+  unfoldConnections(master, path);
+  unfoldNets(master, path);
```
osamahammad21: not expected for non-hierarchical chiplets. please remove.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,283 @@
...
-        unfolded_chip.cuboid.zMax() / chip_inst->getDb()->getDbuPerMicron());
-
-    unfolded_chips_.push_back(unfolded_chip);
-    chip_path_map_[unfolded_chip.getPathKey()] = &unfolded_chips_.back();
-  }
-
-  unfolded_chip.chip_inst_path.pop_back();
+    if (uf_chip.transform.isMirrorZ()) {
+      side = mirrorSide(side);
+    }
```
osamahammad21, line 162: flawed logic. Use dbChipRegion::Side with isMirrorZ to decide the UnfoldedRegionSide.

src/odb/src/3dblox/unfoldedModel.cpp
```diff
@@ -3,139 +3,283 @@
...
 {
+  const int z = uf_region.getSurfaceZ();
+  for (auto* bump_inst : uf_region.region_inst->getChipBumpInsts()) {
+    dbChipBump* bump = bump_inst->getChipBump();
+    if (auto* inst = bump->getInst()) {
+      Point global_xy = inst->getLocation();
+      transform.apply(global_xy);
+      uf_region.bumps.push_back(
+          {.bump_inst = bump_inst,
+           .parent_region = &uf_region,
```
osamahammad21: broken link

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -141,16 +144,54 @@ dbSet<dbChipBump> dbChipRegion::getChipBumps() const
...
+    } else if (side == Side::BACK) {
+      z_top = z_bot = 0;
+    } else {
+      z_bot = 0;
+      z_top = chip_thickness;
+    }
   }
-  return Cuboid(box.xMin(), box.yMin(), z, box.xMax(), box.yMax(), z);
+
+  return Cuboid(box.xMin(), box.yMin(), z_bot, box.xMax(), box.yMax(), z_top);
```
osamahammad21: move to dbChipRegion::create and use the calculated layer stack z coordinates directly without falling back to the defined region side.

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -169,13 +165,26 @@ dbTechLayer* dbChipRegion::getLayer() const
...
-  if (chip->getBlock() == nullptr || chip->getBlock()->getTech() == nullptr
-      || obj->layer_ == 0) {
+
+  dbTech* tech = nullptr;
+  if (chip->getBlock() != nullptr) {
+    tech = chip->getBlock()->getTech();
+  }
+  if (tech == nullptr) {
+    tech = chip->getTech();
+  }
```
osamahammad21: still not handled
```suggestion
  dbTech* tech = chip->getTech();
```

src/odb/src/db/dbChipRegion.cpp
```diff
@@ -227,6 +236,31 @@ dbChipRegion* dbChipRegion::create(dbChip* chip,
...
+        }
+      }
+    }
+  } else {
+    z_bot = 0;
+    z_top = chip_thickness;
+  }
+
+  chip_region->z_min_ = z_bot;
+  chip_region->z_max_ = z_top;
```
osamahammad21: no z_min_ nor z_max_ defined for dbChipRegion.

osamahammad21 (changes requested):

src/odb/src/db/dbChipRegion.h
```diff
@@ -36,6 +36,8 @@ class _dbChipRegion : public _dbObject
   uint8_t side_;
   dbId<_dbTechLayer> layer_;
   Rect box_;
+  int z_min_;
+  int z_max_;
```
osamahammad21, line 40: You added those manually, not with the code generator. Running the code generator will erase this. `[Check That ODB Files Are Generated / Are-Odb-Files-Generated (pull_request)]` github action will fail this.

osamahammad21 (changes requested):

Issue #7957, opened 2025-08-05
ODB: multiple dbGDSLibObj types in dbObjectType enum


osamahammad21:
### Describe the bug

https://github.com/The-OpenROAD-Project/OpenROAD/blob/59ef9a94573c0a6799d32e2dfc9fc7a93feeb30f/src/odb/include/odb/dbObject.h#L38
and
https://github.com/The-OpenROAD-Project/OpenROAD/blob/59ef9a94573c0a6799d32e2dfc9fc7a93feeb30f/src/odb/include/odb/dbObject.h#L123

### Expected Behavior

Only one entry should be there. Removing any of them will mess up the enum order and compromise the integrity of the enum deserialization from older odb files.

### Environment

```shell
NA
```

### To Reproduce

NA

### Relevant log output

```shell

```

osamahammad21:
@maliberty FYI

PR #9296, opened 2026-01-20
odb: fix duplicate dbGDSLibObj enum entry

ahmed532:
There were two redundant enum entries for GDS Library objects in `dbObjectType`: `dbGdsLibObj` and `dbGDSLibObj`.
This duplication caused ambiguity and risked compromising the integrity of the database format.

This commit resolves the duplication by:
1. Deprecating `dbGDSLibObj` and renaming it to `dbObsoleteGdsLibObj_DO_NOT_USE`. The entry is retained in the enum to preserve the integer values of subsequent members, ensuring binary compatibility with existing tools that rely on enum order.
2. Redirecting the legacy hash (`0x53`) to the canonical `dbGdsLibObj` type during deserialization. This allows older ODB files containing the obsolete type to be read correctly.
3. Removing the obsolete type from the serialization table so that all GDS Library objects are now written with the canonical hash (`0x2`).
4. Adding `TestObjectType` to verify hash redirection and ensure the obsolete type is correctly handled.

Closes: #7957

### DIFF
--- a/src/odb/include/odb/dbObject.h
+++ b/src/odb/include/odb/dbObject.h
@@ -129,7 +129,6 @@ enum dbObjectType
 
   // Lib Objects
   dbLibObj,
-  dbObsoleteGdsLibObj_DO_NOT_USE,
   dbSiteObj,
   dbMasterObj,
   dbMPinObj,
--- a/src/odb/src/db/dbObject.cpp
+++ b/src/odb/src/db/dbObject.cpp
@@ -126,7 +126,6 @@ static const char* name_tbl[] = {"dbGDSLib",
 
                                  // Lib Objects
                                  "dbLib",
-                                 "dbGdsLib_obsolete",
                                  "dbSite",
                                  "dbMaster",
                                  "dbMPin",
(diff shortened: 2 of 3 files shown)


maliberty:
Please expand on `binary compatibility with existing tools that rely on enum order.`.  The only binaries we ship are the whole application not libraries.  It seems remapping it in serialization would be sufficient to remove it altogether.

ahmed532:
I meant the libraries like odb, drt, ord, and all others (inside openroad). 
In any case we cannot remove it because that will change all the subsequent enum values (the old value - 1) and then all old odb files will fail to deserialize. Because when you read a value from db file it (the odb file to be deserialized correctly) will expect the mapping of the OpenRoad binary that created it, and if we change all enum values after the duplicate we'll need special handling for all of them.

maliberty:
Libraries inside OR will be recompiled and work fine with a removal.  Isn't the purpose of object_type_to_hash and hash_to_object_type to allow the enum to shift and keep serialization working.  I'm afraid I am still not getting the issue.

ahmed532:
Understood. I was operating under the assumption that we needed to support linking against libodb independently. Since we only ship the full application and don't guarantee ABI stability for external tools, preserving the enum integer values is indeed unnecessary.
​You are correct that the serialization mapping (object_type_to_hash) is sufficient to handle legacy files. I will remove the enum entry entirely and update the mapping tables to redirect the old hash to the canonical type.

PR #9307, opened 2026-01-21
gpl: fix area reports

gudeh:
This PR is no-op.

Changes in this PR:

- Includes new area reports: one matching `report_cell_usage` for total design area, one for movable instances area, and one for the area due to instance pin density adjustment (from PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/8934).
- Includes a new TCL option to disable the recently added instance pin density area adjustment (from PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/8934).
- Fixes the final routability area inflation, which was incorrect because it did not account for the area changes due to timing-driven (TD) modifications.


gudeh:
I triggered a secure-CI to be sure it is no-op.

gudeh:
the final report for inflation is still incorrect

PR #9317, opened 2026-01-22
run-clang-tidy.sh: pick up the correct compilation db

hzeller:
Make run-clang-tidy.sh more robust against not picking up the build/compile_commands.json compilation DB (if there happens to be a leftover old cmake build).

  * set `COMPILE_JSON` default to `comile_flags.txt` if invoked from run-clang-tidy.sh (but still overridable).
  * run-clang-tidy-cached.cc: if it falls back to compile_flags.txt (if the provided `COMPILE_JSON` or the fallback `build/compile_commands.json` is not found), make sure to use that name.
  * while at it: include the compilation db used in the base-hash of the clang tidy cache directory. That way it is quick to switch these without overriding the clang-tidy results for a particular compdb.

### DIFF
--- a/etc/run-clang-tidy-cached.cc
+++ b/etc/run-clang-tidy-cached.cc
@@ -272,9 +272,12 @@ class ContentAddressedStore
 class ClangTidyRunner
 {
  public:
-  ClangTidyRunner(const std::string& cache_prefix, int argc, char** argv)
+  ClangTidyRunner(const std::string& cache_prefix,
+                  std::string_view compilation_db,
+                  int argc,
+                  char** argv)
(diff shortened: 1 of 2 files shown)


maliberty:
What is the recommended way to run clang-tidy now?

hzeller:
The script `etc/run-clang-tidy.sh` combines creating the and running `run-clang-tidy-cached.cc`  so this is the simplest.

You still can invoke `etc/run-clang-tidy-cached.cc` directly, then it uses whatever last state of the compilation DB (same as before).

But to answer the question: I recommend to run `etc/run-clang-tidy.sh`

PR #9321, opened 2026-01-22
Do not clang-tidy code that is not used in the project.

hzeller:
The `lef/` and `def/` directories contain some code that is actually not used. Skip these for clang-tidy, they generate superfluous `clang-diagnostic-error`s.


hzeller:
@maliberty do I see that correctly that these are nowhere referenced ?

PR #9328, opened 2026-01-23
mpl: remove data flow driven mechanism

AcKoucher:
Testing showed that:
1. Timing does not improve with our test cases (actually, it generally gets worse) when using data flow;
2. Building data flow structure is expensive and can have a huge impact in run time for large designs.

Also, I'm working on a TD approach that should help with our timing results - see #9249.


AcKoucher:
Running Secure-CI.

AcKoucher:
Re-running Secure-CI now that #9282 is in.

PR #9338, opened 2026-01-24
odb: add antenna gate diff parsing

gadfort:
Adds:
- antenna gate plus parsing.

Notes:
This mirrors the code that is already present in lefiLayer but gated by the parse65nm function.


maliberty:
`//src/odb/test:write_lef_and_def-tcl_test ` failed in bazel

gadfort:
@maliberty it required an update to the lefok.

PR #9351, opened 2026-01-26
Add bazel infra for doc unit tests

luarss:
- add central python scripts for manpage/readme validations
- modify regression.bzl for .py detection, enable python regression testing
- add temporary skips for broken `openroad -python tests`. This is out of scope for the doc unit tests PRs.
- export `TEST_TYPE` env var

This serves as the foundation for next PR which is to symlink to central scripts (rather than each module having actual .py scripts)

### DIFF
--- a/test/regression.bzl
+++ b/test/regression.bzl
@@ -142,8 +142,15 @@ def regression_test(
     data = _pop(kwargs, "data", [])
     size = _pop(kwargs, "size", "small")
     test_type = _pop(kwargs, "test_type", "")
+    tags = _pop(kwargs, "tags", [])
     for test_file in test_files:
         ext = test_file.split(".")[-1]
+
+        # Python tests that need openroad -python are disabled because
+        # the Bazel build doesn't include Python embedding support.
+        # Tests with test_type="standalone_python" can still run.
+        is_openroad_python_test = ext == "py" and test_type != "standalone_python"
+        test_tags = tags + (["manual"] if is_openroad_python_test else [])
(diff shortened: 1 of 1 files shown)


BUILD.bazel
```diff
@@ -150,7 +150,8 @@ OPENROAD_DEFINES = [
     "OPENROAD_GIT_DESCRIBE=\\\"bazel-build\\\"",
     "BUILD_TYPE=\\\"$(COMPILATION_MODE)\\\"",
     "GPU=false",
-    "BUILD_PYTHON=false",
+    "BUILD_PYTHON=true",
+    "ENABLE_PYTHON3",
```
maliberty: This produces 
```
src/Main.cc:28:10: error: module //:openroad does not depend on a module exporting 'Python.h'
   28 | #include "Python.h"
      |          ^
```
for me.

docs/src/scripts/man_tcl_check.py
```diff
@@ -0,0 +1,82 @@
...
+    if "odb" in tool_dir:
+        tool_dir = "./src/odb"
+    if not os.path.exists(os.path.join(tool_dir, "README.md")):
+        continue
+    if re.search(f".*{'|'.join(e for e in exclude)}.*", path):
+        continue
+
+    # special handling for pad, since it has 3 Tcls.
+    if "ICeWall" in path or "PdnGen" in path:
+        continue
```
maliberty: No longer true

PR #9353, opened 2026-01-26
gpl: modify default initial position of instances to use ODB

gudeh:
This PR modifies the default behavior of the initial position of instances in GPL.

Now if an instance has an existing odb position, does not have a region and `-force_center_initial_place` is not used, GPL will initiate with the position from the odb. Otherwise it will initiate the instance on the core center or region center.

Previously GPL would only initiate instances from the positions given by the odb if `skip_io` option was used, which is non intuitive.

Notice that the initial position of instances is independent of the initial placement run, which uses the conjugate gradient methodology, and is followed by the main nesterov run. 

We can think of GPL run in three steps:

1. Initial position of instances.
2. Run initial placement with conjugate gradient method.
3. Run the main nesterov run.

Where step 1. has three options that can be set by:
  - Having an exisiting odb position set for the instance when GPL is run (default)
  - Using the new TCL parameter: `-force_center_initial_place`.
  - Having a region associated with the instance.
And step 2. can be skipped with `-skip_initial_place`.

Now the user has more ability to control GPL behavior depending on his desires. Now he can use `-skip_initial_place`, plus modifying the position of instances as given by the previous run in the flow (for example, a previous GPL run itself), and have the main nesterov run (step 3) to initiate as he wishes.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -53,9 +53,9 @@ PCRE_VERSION="10.42"
 PCRE_CHECKSUM="37d2f77cfd411a3ddf1c64e1d72e43f7"
 SWIG_VERSION="4.3.0"
 SWIG_CHECKSUM="9f74c7f402aa28d9f75e67d1990ee6fb"
-BOOST_VERSION_BIG="1.86"
+BOOST_VERSION_BIG="1.89"
 BOOST_VERSION_SMALL="${BOOST_VERSION_BIG}.0"
-BOOST_CHECKSUM="ac857d73bb754b718a039830b07b9624"
+BOOST_CHECKSUM="187b577ce9f485314fcf17bcba2fb542"
 EIGEN_VERSION="3.4"
 CUDD_VERSION="3.0.0"
 LEMON_VERSION="1.3.1"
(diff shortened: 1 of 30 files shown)


gudeh:
Thanks to @dbekatli for a detailed exploration of GPL behavior and exposing this limitation!

maliberty:
Various tests fail under bazel.

What was the detailed explanation?

gudeh:
If we wish to maintain the previous behavior we have to use the new TCL parameter `-force_center_initial_place`. Otherwise we are now using the initial position from the ODB (as given from the input test file).

In other words, tests failed because we used the initial position from ODB instead of starting positions all on center which was the previous default.

gudeh:
Ah! I just noticed I should have updated tests from other tools. I will update the tests. I also made some small changes on log messages.

gudeh:
@maliberty I expect tests to be green after they finish this time. There is also an ORFS PR that should follow: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3838#event-22293860701

gudeh:
I see this on the error for bazel: `There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.`

maliberty:
That is more informational, the real problem is
```
[ERROR] Failed to open file: bazel-out/k8-opt/bin/test/orfs/ram_8x7/logs/asap7/ram_8x7/base/5_1_grt.log
```

PR #9375, opened 2026-01-28
odb: identify terminals that can be used with wildcard during defout

gadfort:
Fixes:
- the current implementation doesn't ensure the def and odb have the same power connections, this means the def has the wrong connections.
-  this fixes it by allowing defout to identify the wildcards itself and not rely on a  flag that can only be applied to a single net and not specific terminals.

Updates the defoks that are impacted by this.

### DIFF
--- a/src/odb/src/defout/defout_impl.h
+++ b/src/odb/src/defout/defout_impl.h
@@ -7,6 +7,7 @@
 #include <list>
 #include <map>
 #include <ostream>
+#include <set>
 #include <string>
 #include <unordered_map>
 


src/dpl/test/fillers1.defok
```diff
@@ -53,7 +53,7 @@ PINS 2 ;
         + FIXED ( 0 60 ) N ;
 END PINS
 SPECIALNETS 2 ;
-    - VDD ( * VDD ) + USE POWER
+    - VDD ( _277_ VDD ) + USE POWER
```
maliberty, line 56: I found myself wondering about this and learned that the problem is the fill cells are not connected to vdd while the instance is.  In many tests I think we should setup/run global_connect rather than update the .ok as that better reflect how the tool would really be used.

Otherwise the code is fine.

src/odb/src/defout/defout_impl.cpp
```diff
@@ -1365,7 +1376,14 @@ void DefOut::Impl::writeSNet(dbNet* net)
     dbInst* inst = iterm->getInst();
     dbMTerm* mterm = iterm->getMTerm();
     char* mtname = mterm->getName(inst, &ttname[0]);
-    if (net->isWildConnected()) {
+    bool iswildcard = false;
+    const auto snet_term_map_itr = snet_term_map.find(mterm->getName());
+    if (snet_term_map_itr != snet_term_map.end()
+        && snet_term_map_itr->second.size() == 1) {
```
maliberty: It is possible for the key not to be in the map?

gadfort: I suppose it cannot since this is iterating over the same iterms

src/dpl/test/fillers1.defok
gadfort: yeah, I think that would need to be a followup thing though.

maliberty: Is it less work to update all the def ok versus update the scripts?

gadfort: well, updating the defoks I can do with a simple script to fix all failing defs, updating the scripts requires a little more effort (and since the defs are updated, now it requires more effort to undo)

PR #9376, opened 2026-01-28
ci: change clang branch

vvbandeira:


vvbandeira:
@sombraSoft @maliberty 
FYI in case we need to update boost we need to change here, this is a side effect of the boost vs or-tools conflict.

PR #9379, opened 2026-01-28
mpl: warn total amount of unaligned pins instead of each pin name

AcKoucher:
This is to make the output more useful and less verbose.

The changes here also fix a bug w.r.t. the position used to evaluate whether or not unaligned right way on grid pins exist.


AcKoucher:
@gadfort FYI.

eder-matheus:
@AcKoucher Please check this failure: `1416 - TestSnapper.SingleLayerWithUnalignablePin (Failed)`

AcKoucher:
@eder-matheus fixed.

PR #9389, opened 2026-01-29
GPL: add support to TD with multiple regions

LucasYuki:
Part of #8838.

Changes:
- Fixes cell creation inside GPL.
  - Insert the new cell in the right `dbRegion` (`inDbInstCreate`).
- Fixes cell destruction inside GPL.
  - Each region has a `NesterovBase` object, but the cells are stored in the `NesterovBaseCommon`
  - When a cell is deleted in `NesterovBaseCommon`, we replace the deleted cell with the last cell in the list. After this, we need to update the index of this cell in all `NesterovBases`.
- Fixes a small bug in DPL:
  - The `dummy_cell_` was not set as fixed, causing a segmentation fault in some cases.   
- Updates the test/upf_aes.tcl to use TD.
- Removes the `inDbInstCreate(dbInst* inst, dbRegion* region)` callback.
  - Since it's possible to get the dbRegion from the dbInst, only the `inDbInstCreate(dbInst* inst)` is necessary.

This PR doesn't assign regions to buffers inserted by the RSZ. 
Currently, all buffers are placed in the top region, but support for inserting instances in a dbRegion was added.

### DIFF
--- a/src/gpl/src/nesterovPlace.cpp
+++ b/src/gpl/src/nesterovPlace.cpp
@@ -1281,7 +1281,7 @@ void NesterovPlace::createCbkGCell(odb::dbInst* db_inst)
 void NesterovPlace::destroyCbkGCell(odb::dbInst* db_inst)
 {
   if (db_inst == nullptr) {
-    log_->warn(GPL, 327, "Trying to destroy odb::dbInst* nullptr");
+    log_->warn(GPL, 328, "Trying to destroy odb::dbInst* nullptr");
     return;
   }
 
@@ -1296,7 +1296,7 @@ void NesterovPlace::destroyCbkGCell(odb::dbInst* db_inst)
       }
       if (!updated) {
         log_->error(GPL,
-                    326,
+                    329,
                     "NesterovPlace destroyCbkGCell failed to update db_inst {}",
                     replaced.first->getName());
       }


gudeh:
is this no-op for current designs without regions?

gudeh:
we should have secure-CI for this branch PR

src/gpl/src/nesterovPlace.cpp
```diff
@@ -1318,12 +1365,13 @@ void nesterovDbCbk::inDbPostMoveInst(odb::dbInst* db_inst)
 
 void nesterovDbCbk::inDbInstCreate(odb::dbInst* db_inst)
 {
-  nesterov_place_->createCbkGCell(db_inst);
+  nesterov_place_->createCbkGCell(db_inst, nullptr);
```
gudeh: I think we should "mimic" the ODB setup and have different method overloads, one for the creation ignoring regions (top-level), and another with regions.

LucasYuki: I removed the `inDbInstCreate(dbInst* inst, dbRegion* region)` callback, since it's possible to get the dbRegion from the dbInst.

LucasYuki:
@gudeh, I ran a secure-CI and it didn't change any current designs without regions.

src/dpl/src/Place.cpp
```diff
@@ -738,6 +738,9 @@ bool Opendp::shiftMove(Node* target_cell)
     }
   }
 
+  deepIterativePause("pause after legalGridPt() inside shiftMove(), cell "
+                     + target_cell->name());
+
   deepIterativePause("pause after legalGridPt() inside shiftMove(), cell "
                      + target_cell->name());
```
maliberty, line 745: Why is this duplicated?

test/sky130hd/sky130hd.rc
```diff
@@ -12,4 +12,4 @@ set_layer_rc -via mcon -resistance 9.249146E-3
 set_layer_rc -via via -resistance 4.5E-3
 set_layer_rc -via via2 -resistance 3.368786E-3
 set_layer_rc -via via3 -resistance 0.376635E-3
-set_layer_rc -via via4 -resistance 0.00580E-3
+set_layer_rc -via via4 -resistance 0.00580E-3
```
maliberty, line 15: No reason to remove the last newline.  Do you have an editor setting issue?

test/upf_aes.tcl
```diff
@@ -26,9 +30,9 @@ set_routing_layers -signal li1-met5
 place_pins \
   -hor_layers met3 \
   -ver_layers met2
-global_placement -skip_initial_place -density uniform -routability_driven
+global_placement -skip_initial_place -density uniform -routability_driven -timing_driven
 
-detailed_placement
+detailed_placement -max_displacement 1000
```
maliberty, line 35: Why is such a large limit needed?

src/dpl/src/Place.cpp
LucasYuki: I probably messed up the rebase, I'm removing the duplicate.

test/sky130hd/sky130hd.rc
LucasYuki: I edited this file to add the set_wire_rc command like the `OpenROAD-flow-scripts/flow/platforms/sky130hd/setRC.tcl` file, but I notice that this information was available in `OpenROAD/test/sky130hd/sky130hd.vars`, so I moved this command to `test/upf_aes.tcl` using the value stored in the vars file.
I'm adding the newline back.

test/upf_aes.tcl
LucasYuki: Since all buffers are placed in the top region, some buffers need to be moved up to 680.5 u.
1000 was the first value that worked in my tests, but I can decrease this value to 650.

PR #9397, opened 2026-01-30
dpl: eeq initial patch

bnmfw:
Ignore this for now, this is the most draftiests draft I have ever made, I'm opening this in this state only to have the PR to link to.

### The EEQ Feature
Some PDKs use the LEF58 EEQ Feature. Essentially multiple equivalent LEFs of the same cells in the same sizing are provided. Each one has a different alignment with the tracks of metal:
Outlook-vfm5bwxw
 This essentially ensures that no matter where the instance is placed, if there are default you have a LEF that has all pins aligned in tracks, so every access is guaranteed to be on grid.
### This Patch
This PR contains a VERY simple solution for the EEQ problem used for a specific advanced PDK. In this case misalignment can only happen at the M1 tracks and it is possible to determine the correct version only by name. Please do not pay mind to the name of the functions.

I intend on formalizing this PR later to bring it up to standard. Currently the solution is janky.

### DIFF
--- a/src/dpl/src/Place.cpp
+++ b/src/dpl/src/Place.cpp
@@ -58,6 +58,63 @@ std::string Opendp::printBgBox(
                      queryBox.max_corner().y());
 }
 
+bool Opendp::M1Missaligned(Node* cell) 
+{
+  auto inst = cell->getDbInst();
+  std::string master_name = inst->getMaster()->getName();
+  bool isEEQ = master_name.size() > 4 && master_name.substr(master_name.size() - 4) == "_EEQ";
+  int site_witdh = 480;
+
+  const bool alignment_swapped = (inst->getOrient() == dbOrientType::R180 || inst->getOrient() == dbOrientType::MY);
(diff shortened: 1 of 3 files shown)


src/dpl/src/Place.cpp
```diff
@@ -58,6 +58,63 @@ std::string Opendp::printBgBox(
                      queryBox.max_corner().y());
 }
 
+bool Opendp::M1Missaligned(Node* cell)
```
gudeh, line 61: Better naming: isM1Missaligned()

src/dpl/src/Place.cpp
```diff
@@ -83,6 +140,7 @@ void Opendp::detailedPlacement()
   }
 
   place();
+  fixParity();
```
gudeh, line 143: I wonder if this changes lead to failures, for example the ones checked with `checkDRC()`. You can check with check_placement TCL command.

If they do happen, you could change the call of your function to happen together with mapMove() and shiftMove() since they already provide means of checking for failures during the move of the cells.

PR #9401, opened 2026-01-30
grt: fixes and enhancements for TD-CUGR

eder-matheus:
In this PR:
- Set the slack of non-critical nets as the max float value, making it irrelevant for sorting.
- Set initial slacks before running CUGR flow based on placement parasitics.
- Calculate partial slacks on later stages of CUGR flow.
- Sort the nets using slack as a new parameter, where the nets with most negative slack (critical) are routed first.
- Ignore local nets when getting the routes, fixing the issue with parasitics estimation.
- Use stable_sort instead of sort (this might help with Bazel build failures).
- Add voting system when defining access points for terms using ODB APs.

### DIFF
--- a/src/grt/src/cugr/src/CUGR.cpp
+++ b/src/grt/src/cugr/src/CUGR.cpp
@@ -12,6 +12,7 @@
 #include <memory>
 #include <set>
 #include <sstream>
+#include <tuple>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


eder-matheus:
@rafaelmoresco FYI, this PR fixes the issue with estimate parasitics you're having.

maliberty (approved):

Issue #4872, opened 2024-03-29
Create an IP Checker Utility


rovinski:
### Description

_Originally posted by @rovinski in https://github.com/The-OpenROAD-Project/OpenROAD/issues/4871#issuecomment-2026137319_

Currently, it is very possible to import an IP block which is poorly designed and will cause downstream issues in OpenROAD, which becomes difficult to root-cause. A checker utility which checks against a list of common problems can go a long way in identifying problems early.

### Suggested Solution

List of common problems to check (far from an exhaustive list):
- LEF
  - [ ] Macro dimensions are a multiple of the manufacturing grid
  - [ ] Pins are aligned to the manufacturing grid
  - [ ] Each signal pin must obey one of the following:
    - Minimum width, and each pin lies on the routing grid (with an offset is okay)
    - Greater than minimum size, has an access point on the routing grid, and connections from a default width wire OR an NDR rule does not cause a DRC violation.
    - Pins are wide enough such that regardless of offset, a connection from a default width wire (OR NDR rule) does not cause a DRC violation.
  - [ ] Power ring or grid straps are accessible (no obstructions in the way)
  - [ ] Signal pins are accessible (no obstructions in the way)
  - [ ] LEF file does not have an excessive number of polygons which will slow down OR
  - [ ] Ports have antenna info
  - [ ] LEFs in finfet processes have a fin grid property
- LIB
  - All quantities (rise, fall, delay, power, cap, etc.) are sane
    - [ ] No instantaneous transitions for reasonable input values
    - [ ] No excessively long transitions for reasonable input values
    - [ ] No excessively large capacitances
  - [ ] No missing properties
  - [ ] Characterization corner(s) match(es) the characterization corner(s) of the design
- LIB+LEF
  - [ ] Every LEF digital pin has lib timing info
  - [ ] LEF power pins defined match lib power pins
  - [ ] LEF and LIB area match

QuantamHD:
Huge +1 from me.

maliberty:
Thanks for writing this up.  A new comer doesn't have to implement every check in order to contribute.  Ticking any box is helpful.

rovinski:
Indeed, I just wanted to list everything I could think of.

razmikarshakyann:
Hi everyone,
I am currently working on this subtask: Each signal pin must obey one of the following:
- Minimum width, and each pin lies on the routing grid (with an offset is okay)
[...]

PR #9405, opened 2026-01-31
added an Ip Checker utility

Mohamed-Khairy-SWE:
this utility currently do some LEF file checks, i just made a pull request to see if i'm on the right track

releated to #4872

### DIFF
--- a/src/dbSta/test/check_ip_test.tcl
+++ b/src/dbSta/test/check_ip_test.tcl
@@ -27,7 +27,7 @@ proc test_check_fails { test_name master_name { extra_args "" } } {
   puts "Testing master: $master_name"
   set cmd "check_ip -master $master_name $extra_args"
   puts "Command: $cmd"
-  if { [catch { eval $cmd } err] } {
+  if { [catch { eval { $cmd } } err] } {
     puts "EXPECTED FAILURE: $err"
     return 1
   } else {
@@ -42,7 +42,7 @@ proc test_check_passes { test_name master_name { extra_args "" } } {
   puts "Testing master: $master_name"
   set cmd "check_ip -master $master_name $extra_args"
   puts "Command: $cmd"
-  if { [catch { eval $cmd } err] } {
+  if { [catch { eval { $cmd } } err] } {
     puts "UNEXPECTED FAILURE: $err"
     return 0
   } else {
@@ -80,7 +80,8 @@ test_check_fails "LEF-CHK-003d: Multi-pattern still incompatible (FAIL)" "lef003
 test_check_fails "LEF-CHK-004-005a: Pin fully blocked (M1+M2)" "lef004_005a_fully_blocked"
 
 # Test 4-5b: LEF-CHK-004-005 - Pin blocked on M1 but accessible from M2 above (PASS)
-test_check_passes "LEF-CHK-004-005b: Blocked on M1 but from M2 (PASS)" "lef004_005b_multi_shape_pass"
+test_check_passes \
+  "LEF-CHK-004-005b: Blocked on M1 but from M2 (PASS)" "lef004_005b_multi_shape_pass"
 
 # Test 4-5c: LEF-CHK-004-005 - Power pin blocked on M1+M2 (FAIL)
 test_check_fails "LEF-CHK-004-005c: Power pin blocked (M1+M2)" "lef004_005c_power_blocked"
@@ -90,10 +91,12 @@ test_check_fails "LEF-CHK-004-005c: Power pin blocked (M1+M2)" "lef004_005c_powe
 test_check_passes "LEF-CHK-004-005d: Edge pin accessible from outside (PASS)" "lef004_005d_edge_pin"
 
 # Test 6a: LEF-CHK-006 - Excessive polygon count (below threshold = FAIL)
-test_check_fails "LEF-CHK-006: Excessive polygon count (threshold 5)" "lef006_polygon_count" "-max_polygons 5"
+test_check_fails \
+  "LEF-CHK-006: Excessive polygon count (threshold 5)" "lef006_polygon_count" "-max_polygons 5"
 
 # Test 6b: LEF-CHK-006 - Polygon count at threshold (should PASS)
-test_check_passes "LEF-CHK-006: Polygon count at threshold (threshold 6)" "lef006_polygon_count" "-max_polygons 6"
+test_check_passes \
+  "LEF-CHK-006: Polygon count at threshold (threshold 6)" "lef006_polygon_count" "-max_polygons 6"
 
 # Test 7: LEF-CHK-007 - Missing antenna info
 test_check_fails "LEF-CHK-007: Missing antenna model" "lef007_no_antenna"


src/utl/include/utl/IpChecker.h
```diff
@@ -0,0 +1,127 @@
...
+{
+  int max_polygons = 10000;  // Polygon count warning threshold
+  bool verbose = false;      // Include informational messages in output
+};
+
+// IP Checker utility for validating LEF macros
+class IpChecker
+{
+ public:
+  IpChecker(odb::dbDatabase* db, sta::dbSta* sta, Logger* logger);
```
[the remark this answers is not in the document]
mohamed-khairy-swe: i will use this when i start working on the LIB checks

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+{
+  odb::dbTech* tech = db_->getTech();
+  if (!tech || !tech->hasManufacturingGrid()) {
+    return;
+  }
+
+  int mfg_grid = tech->getManufacturingGrid();
+  if (mfg_grid <= 0) {
+    return;
+  }
```
rovinski: I don't think that's possible

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+  results_.clear();
+  track_grid_warning_emitted_ = false;
+}
+
+void IpChecker::addWarning(const std::string& code,
+                           const std::string& message,
+                           const std::string& object)
+{
+  results_.emplace_back(code, message, object);
+}
```
rovinski: I would not use this structure. I would simply use `utl::Logger::warning()` to print the message immediately and increment a counter for number of warnings. Then, use `utl::Logger::error()` at the end if the counter > 0. This is the same way it is done in the LEF/DEF parser.

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+          aligned = false;
+          coords += "yMax=" + std::to_string(yMax) + " ";
+        }
+
+        if (!aligned) {
+          addWarning("LEF-002",
+                     "Pin geometry not aligned to manufacturing grid ("
+                         + std::to_string(mfg_grid) + "): " + coords,
+                     master_name + "/" + mterm->getName());
+        }
```
rovinski: I would simplify this by just reporting a warning on the first edge that isn't aligned. Message can be something like "Pin {master_name}/{pin_name} is not aligned to the manufacturing grid ({manufacturing grid in um})"

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+          }
+
+          // Check for overlap
+          if (pin_box->xMin() < obs->xMax() && pin_box->xMax() > obs->xMin()
+              && pin_box->yMin() < obs->yMax()
+              && pin_box->yMax() > obs->yMin()) {
+            addWarning("LEF-004",
+                       "Signal pin obstructed on layer " + obs_layer->getName(),
+                       master_name + "/" + mterm->getName());
+            break;  // One warning per pin is enough
```
rovinski: Two issues here:

1. You should use odb geometry functions such as `odb::intersect()` or `odb::overlap` to perform polygon operations
2. It is actually okay if an obstruction overlaps a pin shape, as long as at least one edge of the pin is accessible (north, south, east, west, up, down). The main issue is if no edge of the pin is accessible such that there is no viable way to connect to it.

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+      }
+    }
+  }
+
+  if (polygon_count > config_.max_polygons) {
+    addWarning("LEF-006",
+               "Excessive polygon count: " + std::to_string(polygon_count)
+                   + " (threshold: " + std::to_string(config_.max_polygons)
+                   + ")",
+               master_name);
```
rovinski: More descriptive is helpful here. "Master {master_name} has more than 10000 polygons, which will significantly slow down processes."

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+        uint32_t min_width = layer->getWidth();
+        if (min_width <= 0) {
+          continue;
+        }
+
+        uint32_t width = box->getDX();
+        uint32_t height = box->getDY();
+
+        // The smaller dimension is the "width" of the shape
+        uint32_t shape_width = std::min(width, height);
```
rovinski: `odb::Rect::minDXDY()`

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+  checkManufacturingGridAlignment(master);       // LEF-001
+  checkPinManufacturingGridAlignment(master);    // LEF-002
+  checkPinRoutingGridAlignment(master);          // LEF-003
+  checkSignalPinAccessibility(master);           // LEF-004
+  checkPowerPinAccessibility(master);            // LEF-005
+  checkPolygonCount(master);                     // LEF-006
+  checkAntennaInfo(master);                      // LEF-007
+  checkFinFetProperty(master);                   // LEF-008
+  checkPinGeometryPresence(master);              // LEF-009
+  checkPinMinDimensions(master);                 // LEF-010
```
rovinski: I would not use `LEF` as it is already in use by the LEF parser. Perhaps `CHK` for checker. I am personally not fond of IPC because it makes me think of "Inter-Process Communication" (or Instructions Per Clock)

src/utl/src/Logger.i
rovinski, line 1: I don't think Logger.i needs to be modified at all

src/utl/test/check_ip_tests.tcl
rovinski, line 1: I would encourage using existing LEF test files to perform checks. If they don't satisfy what you need, It would be better to make the .lef files themselves rather than a script that generates them.

.gitignore
```diff
@@ -63,3 +63,7 @@ projectview.bazelproject
 .bsp/
 .bazelbsp/
 /buildifier
+
+check_ip_test.lef
+check_ip_test.v
+
```
rovinski: This should not be in the top-level .gitignore

rovinski (commented):
Initial comments

src/utl/src/Logger.i
```diff
@@ -101,6 +105,36 @@ void startPrometheusEndpoint(uint16_t port)
   logger->startPrometheusEndpoint(port);
 }
 
-} // namespace
+}  // namespace utl
+
+bool check_ip_cmd(const char* master_name,
```
maliberty: The checker doesn't belong in utl as it is a low level library with few dependencies.  The checker will need more dependencies.  dbSta might be a more natural home to deal with timing and odb.

mohamed-khairy-swe:
That was a valuable code review, thank you very much

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+            break;
+          }
+        }
+
+        // Check Y alignment
+        for (int track_y : grid_y) {
+          if (pin_center_y == track_y) {
+            y_aligned = true;
+            break;
+          }
```
gadfort: I'm not sure this check work properly.
The routing tracks and master pins are not really comparable. Imagine a macro where the pins are all placed at 0.25 tracks over (expecting that they would align when the macro is placed).
You probably need to check that pin pitches are a multiple of the routing track pitches.

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+        odb::dbTechLayer* pin_layer = pin_box->getTechLayer();
+        if (!pin_layer) {
+          continue;
+        }
+
+        for (odb::dbBox* obs : obstructions) {
+          odb::dbTechLayer* obs_layer = obs->getTechLayer();
+          if (obs_layer != pin_layer) {
+            continue;
+          }
```
gadfort: You should check that the obs is above the pins.

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
```
gadfort: same, check that the obs is above the pins

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+  }
+
+  // Heuristic: check if any layer has "fin" in property names
+  bool has_fin_layer = false;
+
+  for (odb::dbTechLayer* layer : tech->getLayers()) {
+    // Check for fin-related layer names or properties
+    std::string layer_name = layer->getName();
+    if (layer_name.find("fin") != std::string::npos
+        || layer_name.find("FIN") != std::string::npos) {
```
gadfort: What is the purpose of this? name matching is very brittle.

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+    logger_->warn(IPC, 1, "{}", msg);
+  }
+
+  // Summary
+  logger_->info(IPC, 2, "IP Check complete: {} warnings", warnings);
+
+  // Error out if any warnings were found
+  if (warnings > 0) {
+    logger_->error(IPC, 3, "IP Check failed with {} warnings", warnings);
+  }
```
gadfort: It might be nice to have warnings that would cause an error (pins blocked for example), but track alignment and shape count only impact runtime and thus would not really be an error (at least to me)

src/utl/src/IpChecker.cpp
```diff
@@ -0,0 +1,587 @@
...
+    file << "WARNING [" << result.code << "]";
+    if (!result.object.empty()) {
+      file << " " << result.object;
+    }
+    file << ": " << result.message << "\n";
+  }
+
+  file << "\nSummary: " << getWarningCount() << " warnings\n";
+
+  file.close();
```
gadfort: report this to the logger (from there we can use `tee` to pipe the report into a file if needed.)

mohamed-khairy-swe:
@maliberty @rovinski @gadfort 
I made changes

mohamed-khairy-swe:
@maliberty @rovinski @gadfort 
Should I make a new PR for the changes?

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,546 @@
...
+    return;
+  }
+
+  int mfg_grid = tech->getManufacturingGrid();
+  uint32_t width = master->getWidth();
+  uint32_t height = master->getHeight();
+  std::string master_name = master->getName();
+
+  // Convert to microns for display (assuming DBU = nm)
+  double mfg_grid_um = mfg_grid / 1000.0;
```
maliberty: Use dbTech::getDbUnitsPerMicron() - it isn't safe to assume 1000.  Same throughout

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,546 @@
...
+            || rect.xMax() % mfg_grid != 0 || rect.yMax() % mfg_grid != 0) {
+          logger_->warn(
+              utl::CHK,
+              20,
+              "Pin {}/{} not aligned to manufacturing grid ({:.3f} um)",
+              master_name,
+              mterm->getName(),
+              mfg_grid_um);
+          warning_count_++;
+          return;  // One warning per pin is enough
```
maliberty: This is one warning per master not per pin.

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,546 @@
...
+        odb::dbTrackGrid* track_grid = block->findTrackGrid(layer);
+        if (!track_grid) {
+          continue;
+        }
+
+        // Get track pitches (step) from track grid patterns
+        int origin_x = 0, line_count_x = 0, x_pitch = 0;
+        int origin_y = 0, line_count_y = 0, y_pitch = 0;
+        if (track_grid->getNumGridPatternsX() > 0) {
+          track_grid->getGridPatternX(0, origin_x, line_count_x, x_pitch);
```
maliberty: You are assuming a single pattern per layer which is not always true (eg M2 in asap7).

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,546 @@
...
+
+        // Check if pin dimensions are compatible with track pitch
+        bool is_horizontal
+            = (layer->getDirection() == odb::dbTechLayerDir::HORIZONTAL);
+
+        bool compatible = false;
+        if (is_horizontal) {
+          // Horizontal layer: pin height should be compatible with y_pitch
+          compatible
+              = (pin_height % y_pitch == 0) || (y_pitch % pin_height == 0);
```
maliberty: I don't think this is right.  The pin may contain multiple shapes and its bbox doesn't have to be a multiple of the track pattern.  What is the goal of this check?

maliberty (changes requested):

maliberty:
Have you run this checker on all the public PDKs?  I think that would be a good form of testing.

src/dbSta/src/IpChecker.cc
mohamed-khairy-swe: Well, the first time i implemented the check where it compared pin center positions to track grid positions wasn't right because @gadfort pointed out that macros might have pins intentionally offset (like 0.25 tracks over), expecting alignment when the macro is placed. He suggested checking that pin pitches are multiples of routing track pitches instead.

So While implementing that, I was thinking about how to verify pin-to-track compatibility. My initial approach was to check if individual pin dimensions (width/height) are multiples of track pitch

mohamed-khairy-swe:
@maliberty 
I ran the checker on (Nangate45.lef, sky130_fd_sc_hd.tlef, sky130_fd_sc_hd_merged.lef ,asap7_tech_1x.lef, asap7sc7p5t_28_L_1x_220121a.lef)

Nangate45: 533 warnings (all the warnings was just missing antenna model)

asap7: 2228 warnings (mostly missing antenna model, some missing pin geometry)

sky130hd: 3114 warnings (mostly missing antenna model, some missing pin geometry)

src/dbSta/src/IpChecker.cc
mohamed-khairy-swe: @maliberty 
what is the right way to implement LEF-CHK-003: Pin routing grid alignment?

rovinski: From the original issue:

* Each signal pin must obey one of the following:
  * Minimum width, and each pin lies on the routing grid (with an offset is okay)
  * Greater than minimum size, has an access point on the routing grid, and connections from a default width wire OR an NDR rule does not cause a DRC violation.
  * Pins are wide enough such that regardless of offset, a connection from a default width wire (OR NDR rule) does not cause a DRC violation.

The easiest check is 1, 2/3 are harder. I would say always heir on the side of assuming the IP is okay unless it definitely isn't (for example, not aligned to the tracks is okay, assuming that there is some offset that makes it aligned. If there is no offset that can align it, then that's not okay)

So basically the check should be (pseudocode)
```
if (pin is minimum width) {
  find the distances between each pin
  find the largest common multiple (LCM) of these distances
  if LCM is not a multiple of the track pitch, then warning because there is no offset that could align it
} else {
  // to be implemented later. Don't issue warning yet.
}
```

For example, let's say there are 3 minimum-sized pins. One has an offset of 15um, two is 20um, and three is 30um. The LCM is 5um. If the track pitch is 1um, then this is fine because 1 is a factor of 5. If the track pitch were 2um, this is not fine because it is not a factor of 5. Therefore there is no offset that could be applied to the block in order to align it to a 2um track pitch.

Do note that this, for the most part, only applies to blocks and not to standard cells. Standard cells have more complicated access patterns that are more challenging to analyze. But I think that standard cells are beyond the scope of #4872, it is mostly concerned with block IP.

rovinski:
I think that checking PDKs is okay, but many of the checks related to geometry mainly apply to block designs. Standard cells are more complicated and outside the scope of #4872. Those checks would have to much more heavily rely on using the Pin Access engine which seems complicated.

Block IPs are generated by users and are more prone to problems, whereas standard cells are generally made by library designers with more care.

maliberty:
In that case we should have some check to determine this is a not a standard cell and that simplified assumptions are true (eg pins are simple rectangles on track).

rovinski:
Should be as simple as checking class == block

src/dbSta/src/IpChecker.cc
mohamed-khairy-swe: @rovinski you said largest common multiple (LCM) but I think you actually meant Greatest Common Divisor (GCD) right😅?
Because say pins are at positions 15, 20, 30. Distances are 5 and 10. The GCD of those is 5. That's basically the spacing grid that all the pins naturally sit on. If the track pitch divides into that grid (GCD % pitch == 0), then we can always find some offset to align everything.

so the pseudocode i'm thinking of:
```
for each routing layer:
  collect center positions of all minimum-width signal pins on this layer
  if less than 2 pins: skip (nothing to compare)
  
  sort the positions
  compute GCD of consecutive distances between pins
  
  for each track pattern on this layer:
    if GCD % track_pitch == 0: it's fine, some offset exists
  
  if no pattern works: warn
```
Is that right?

rovinski: > I think you actually meant Greatest Common Divisor (GCD)

Sure, same thing.

> ```
>   for each track pattern on this layer:
>     if GCD % track_pitch == 0: it's fine, some offset exists

I think the only tricky thing here is that the offsets between the track patterns do matter. For example if you have track patterns 1 and 2, with both of them having a pitch of 10um, but then the offsets for the patterns are 0um and 5um (respectively), then the effective pitch among those patterns is 5um.

mohamed-khairy-swe: @rovinski  oh i know , how about we compute the effective pitch as: `GCD(all pattern pitches, all origin differences)
`
so for your example that'd be` GCD(10, 10, |5-0|) = 5`. Then we just check `pin_distance_GCD % effective_pitch == 0` as before.

rovinski: I would recommend just writing some additional tests. It would be easier to demonstrate that way.

mohamed-khairy-swe:
@rovinski @maliberty 
Done

src/dbSta/test/check_ip_test.tcl
```diff
@@ -0,0 +1,115 @@
...
+
+# Helper to run a test that expects failure
+proc test_check_fails { test_name master_name { extra_args "" } } {
+  puts "\n--- Test: $test_name ---"
+  puts "Testing master: $master_name"
+  if { $extra_args ne "" } {
+    set cmd "check_ip -master $master_name $extra_args"
+  } else {
+    set cmd "check_ip -master $master_name"
+  }
```
rovinski: this isn't necessary. If `extra_args` is an empty string, it will be parsed the exact same as above

rovinski: Also, if you intend to supply multiple extra args, you should use argument expansion, e.g. `${*}extra_args`

src/dbSta/test/check_ip_test.lef
```diff
@@ -0,0 +1,389 @@
...
+    PORT
+      LAYER M1 ;
+      RECT 1.000 1.000 1.200 1.200 ;
+    END
+  END SIG
+  OBS
+    LAYER M1 ;
+    RECT 0.900 0.900 1.300 1.300 ;
+  END
+END lef004_signal_blocked
```
rovinski: This example is not correct. It could technically be accessed above with M2 and dropping a via down. You should also have an example where the pin is accessible from underneath (for example, pin is on M2 but M1 is open).

You should also consider the case where there are multiple shapes on the pin - one is blocked but another is available.

src/dbSta/test/check_ip_test.lef
```diff
@@ -0,0 +1,389 @@
...
+    PORT
+      LAYER M1 ;
+      RECT 1.000 1.000 1.200 1.200 ;
+    END
+  END VDD
+  OBS
+    LAYER M1 ;
+    RECT 0.900 0.900 1.300 1.300 ;
+  END
+END lef005_power_blocked
```
rovinski: Is this checking anything different from 004?

src/dbSta/test/check_ip_test.tcl
```diff
@@ -0,0 +1,115 @@
...
+test_check_fails "LEF-CHK-003d: Multi-pattern still incompatible (FAIL)" "lef003d_multi_fail"
+
+# Test 4: LEF-CHK-004 - Signal pin not accessible
+test_check_fails "LEF-CHK-004: Signal pin blocked by obstruction" "lef004_signal_blocked"
+
+# Test 5: LEF-CHK-005 - Power pin not accessible
+test_check_fails "LEF-CHK-005: Power pin blocked by obstruction" "lef005_power_blocked"
+
+# Test 6: LEF-CHK-006 - Excessive polygon count (use low threshold)
+test_check_fails "LEF-CHK-006: Excessive polygon count" "lef006_polygon_count" "-max_polygons 5"
```
rovinski: While you're at it, you should test that `-max_polygons 6` passes.

src/dbSta/test/check_ip_test.lef
```diff
@@ -0,0 +1,389 @@
...
+      LAYER M1 ;
+      RECT 0.000 1.000 0.200 1.200 ;
+    END
+  END A
+END lef007_no_antenna
+
+# =============================================================================
+# LEF-CHK-008: FinFET technology detection (info only, use -verbose)
+# Uses fin_drawing layer defined above
+# =============================================================================
```
rovinski, line 398: It is unclear how this check works.

src/dbSta/test/check_ip_test.lef
```diff
@@ -0,0 +1,389 @@
...
+  PIN A
+    DIRECTION INPUT ;
+    USE SIGNAL ;
+    ANTENNAMODEL OXIDE1 ;
+    PORT
+      LAYER M1 ;
+      RECT 0.000 1.000 0.050 1.050 ;
+    END
+  END A
+END lef010_small_pin
```
rovinski: This check should be more specific. It should check that the min width is specifically violating in the width direction (perpendicular to the preferred routing direction in that layer). It can still be violating even if the length > 0.1. It can also be _not_ violating even if the length < 0.1 (in that case, you should defer to the min area check)

src/dbSta/test/check_ip_test.lef
```diff
@@ -0,0 +1,389 @@
...
+
```
rovinski, line 469: Please also add a test case for each pin shape's min area.

src/dbSta/test/check_ip_test.lef
mohamed-khairy-swe: @rovinski I was thinking signal pins and power pins might need different accessibility logic down the road — like maybe power pins have different routing requirements or something. But yeah it's literally the same exact code with the dbSigType filter flipped 😅

So I'm going to merge them into a single `checkPinAccessibility()` that just loops over all pins regardless of type. The warning message should say "Signal" or "Power" based on the sig type so you can still tell them apart in the output.

mohamed-khairy-swe:
on it 🫡

mohamed-khairy-swe:
@rovinski 
wait i forgot the test cases that actually triggers the min area check

mohamed-khairy-swe:
@rovinski 
Done

src/dbSta/test/check_ip_test.lef
rovinski: I don't think the port type matters here for pin accessibility. As long as the pin is accessible.

src/dbSta/test/check_ip_test.lef
```diff
@@ -247,9 +248,10 @@ MACRO lef003d_multi_fail
 END lef003d_multi_fail
 
 # =============================================================================
-# LEF-CHK-004: Signal pin has NO accessible edge (blocked by obstruction)
+# LEF-CHK-004-005a: Pin fully blocked - obstruction covers pin on same layer
+# AND above layer (M2) so it can't be accessed from above either.
```
rovinski: Macro does not match description. There is no M2.

src/dbSta/test/check_ip_test.lef
```diff
@@ -266,12 +268,37 @@ MACRO lef004_signal_blocked
...
     RECT 0.900 0.900 1.300 1.300 ;
   END
-END lef004_signal_blocked
+END lef004_005a_fully_blocked
 
 # =============================================================================
-# LEF-CHK-005: Power pin has NO accessible edge (blocked by obstruction)
+# LEF-CHK-004-005b: Pin has multiple shapes - one blocked, one accessible (PASS)
+# First shape is blocked by obstruction but second shape is open.
+# The check should pass because at least one shape is accessible.
```
rovinski: Also no M2 here

rovinski (commented):
Please address the clang-tidy comments

mohamed-khairy-swe:
@rovinski 
Done

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,681 @@
...
+        // For wider pins, routing can connect regardless of offset
+        int pin_dim = is_horizontal ? rect.dy() : rect.dx();
+        if (min_width > 0 && static_cast<uint32_t>(pin_dim) > min_width) {
+          continue;  // Wider than minimum, skip for now
+        }
+
+        // Pin center along the routing direction
+        int center = is_horizontal
+                         ? (rect.yMin() + rect.yMax()) / 2
+                         : (rect.xMin() + rect.xMax()) / 2;
```
rovinski: Use `Rect::yCenter()` and `Rect::xCenter()`

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,681 @@
...
+    }
+  }
+
+  // For each layer, compute GCD of distances and check against track pitch
+  for (auto& [layer, centers] : layer_pin_centers) {
+    if (centers.size() < 2) {
+      continue;  // Need at least 2 pins to compute distances
+    }
+
+    // Sort and compute distances between consecutive pin centers
```
rovinski: Why is sorting necessary?

Also if you do need to sort, you should use `std::ranges::sort()`

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,681 @@
...
+  std::string master_name = master->getName();
+
+  // Collect minimum-width signal pin centers grouped by layer
+  // key: layer, value: list of pin center positions along routing direction
+  std::map<odb::dbTechLayer*, std::vector<int>> layer_pin_centers;
+
+  for (odb::dbMTerm* mterm : master->getMTerms()) {
+    odb::dbSigType sig_type = mterm->getSigType();
+    if (sig_type == odb::dbSigType::POWER
+        || sig_type == odb::dbSigType::GROUND) {
```
rovinski: use `dbSigType::isSupply()`

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,681 @@
...
+          continue;
+        }
+
+        odb::Rect rect = box->getBox();
+        uint32_t min_width = layer->getWidth();
+        bool is_horizontal =
+            (layer->getDirection() == odb::dbTechLayerDir::HORIZONTAL);
+
+        // Only check minimum-width pins
+        // For wider pins, routing can connect regardless of offset
```
rovinski: I would just clarify this comment to "might" be able to connect. Checking whether it actually can is a much more complicated check, so it doesn't need to be in an initial implementation.

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,681 @@
...
+          warning_count_++;
+          return;  // One warning per master is enough
+        }
+      }
+    }
+  }
+}
+
+// LEF-CHK-003: Pin routing grid alignment
+void IpChecker::checkPinRoutingGridAlignment(odb::dbMaster* master)
```
rovinski, line 200: This function seems to be checking that all pins within one layer are aligned with the tracks. I think that is fine for an initial version. An ideal version would check that pins across multiple layers can be aligned to the tracks.

Consider the following block which has a pin on M1 (blue) and M2 (red)
Image

Individually, they can both easily be shifted to align with the track pattern on the left. But together, there is no way to shift the block up or down such that both M1 and M2 are aligned to the track pattern.

Something to consider in the future after this PR.

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,681 @@
...
+  }
+
+  odb::dbBlock* block = chip->getBlock();
+  if (!block) {
+    return;
+  }
+
+  auto track_grids = block->getTrackGrids();
+  if (track_grids.empty()) {
+    return;
```
rovinski, line 219: If the check returns early, I would log an `info` that the check is being skipped and the reason why, e.g. "Skipping pin routing grid alignment check due to no tracks present in the design."

Otherwise, a skipped check looks the same as a passing check.

src/dbSta/src/IpChecker.cc
```diff
@@ -0,0 +1,681 @@
...
+    if (obs_rect.intersects(south_edge)) {
+      south_blocked = true;
+    }
+    if (obs_rect.intersects(east_edge)) {
+      east_blocked = true;
+    }
+    if (obs_rect.intersects(west_edge)) {
+      west_blocked = true;
+    }
+  }
```
rovinski, line 429: This strategy won't work. It seems like you are creating a 1 dbu wide rectangle on each edge and then using `Rect::intersects()` to check for intersection. However, intersects returns true if edges are touching.

A pin shape overrides any obstructions, and is in fact quite common in blocks.

Consider this case, where a pin in M1 (dark blue) is on an edge but also completely covered by an M1 obstruction (light blue). It is still valid to connect to this pin from the west side, but I think your check will fail.

Image

mohamed-khairy-swe:
@rovinski 
Do you have resources that can help me have a deep understanding of everything, because AI is not very reliable

rovinski:
I think the Gemini bot does a decent job at giving code feedback in this repo.

For fundamental concepts and understanding, I'm not sure if there's a great direct resource. Asking questions here is probably best.

I think this PR is nearly there. The main concern left that I see is changing the obstruction strategy. This is something that even AI might not tell you because I think it isn't defined in the LEF/DEF spec, but is rather a conventional practice that tools abide by.

src/dbSta/src/IpChecker.cc
mohamed-khairy-swe: because pins are collected by iterating over mterms not in spatial order. We need consecutive pin positions to compute the distances between them, The GCD of distances between consecutive sorted positions gives us the minimal spacing pattern which is what we check against the track pitch. Without sorting, centers[i] - centers[i-1] wouldn't give meaningful distances.

src/dbSta/src/IpChecker.cc
mohamed-khairy-swe: @rovinski 
you'er right I wrote a passing test and it failed so I switched from` intersects()` to `overlaps()` so obstructions merely touching pin edges don't count as blocking.

mohamed-khairy-swe:
@rovinski 
Done, sorry for the delay but i was busy with college stuff

mohamed-khairy-swe:
@rovinski 
Are you busy?

src/dbSta/src/IpChecker.cc
```diff
@@ -371,11 +374,28 @@ bool IpChecker::hasAccessibleEdge(odb::dbMaster* master,
...
 
   // Check 1: Same-layer edge access
-  // If any edge of the pin is not blocked, routing can reach it
+  // If any edge of the pin is not blocked, routing can reach it.
+  // A pin edge on the macro boundary is always accessible from outside.
   if (same_layer_obs.empty()) {
     return true;  // No same-layer obstructions, all edges accessible
   }
 
+  // Get macro bounding box (origin is always 0,0 in master coords)
```
rovinski, line 386: The origin is actually unfortunately not always (0,0). It almost always is by convention, but the LEF/DEF spec doesn't require it to be. There is a function to get the origin.

src/dbSta/src/IpChecker.cc
```diff
@@ -371,11 +374,28 @@ bool IpChecker::hasAccessibleEdge(odb::dbMaster* master,
...
+  int macro_width = static_cast<int>(master->getWidth());
+  int macro_height = static_cast<int>(master->getHeight());
+
+  // An edge touching the macro boundary is accessible from outside
+  bool north_on_boundary = (pin_rect.yMax() >= macro_height);
+  bool south_on_boundary = (pin_rect.yMin() <= 0);
+  bool east_on_boundary = (pin_rect.xMax() >= macro_width);
+  bool west_on_boundary = (pin_rect.xMin() <= 0);
```
rovinski, line 394: The macro boundary doesn't deal with pins. It's only the obstructions that should be checked against.

Although it is VERY unconventional, I think it is technically legal for an obstruction to be outside of the macro boundaries.

rovinski:
I think this PR might be good enough as it is for an initial implementation. It's not perfect but it can serve as a good start for another PR to fixup the remaining items. Maybe ping @maliberty for review.

mohamed-khairy-swe:
@rovinski 
Should I work on the things that you pointed out or leave it for another PR?

rovinski:
If you can fix them up that would be great. I just also don't want this PR to drag on forever.

mohamed-khairy-swe:
@rovinski 
Yeah this PR has over 90 comments, so if there isn't anything mandatory I would prefer to make it in another PR

maliberty:
I agree it is time to merge this PR and continue refinements in another.

maliberty (approved):

mohamed-khairy-swe:
@maliberty 
I think you forgot to run the github actions

rovinski:
@Mohamed-Khairy-SWE thank you very much for the contribution to OpenROAD! If you would like to continue the refinements to improve the checker in another PR, I'd be happy to continue iterating there.

mohamed-khairy-swe:
@rovinski @maliberty @gadfort 
It was my pleasure to work with great engineers like you ♥️, I hope one day I can reach your level 

Oh I forgot to mention that I'm currently working on my proposal for a gsoc project related to CIRCT (the Verilog compiler) So once I finish my proposal and submit it, I can comeback to openROAD and continue refinement to the IP checker,

I actually was hoping openROAD would propose a project this year but it's all good 😊

On last question:
If I kept contributing to openROAD and take on bigger tasks, Will that make me a maintainer at openROAD?

maliberty:
Thanks for contributing.  There will be some OR related GSoC projects this year (as in the past).  Please keep an eye out.

You would need to be an active contributor for quite a while to become a maintainer.  You need to demonstrate the skill and also the commitment over time to the project.

PR #9410, opened 2026-02-02
Ifp log messages

gudeh:
This PR is no-op. And just adds and changes some log messages

- New log messages showing core area, instances area, number of instances and final utilization.
- Also new messages for stating which of the 4 types we have for initiating the floorplan.
- Rearrange error messages IDs to start at 300.


src/gpl/test/region01.ok
```diff
@@ -3,9 +3,18 @@
 The NOWIREEXTENSIONATPIN statement will be ignored. See file sky130hd/sky130_fd_sc_hd.lef at line 2.
 
 [INFO ODB-0227] LEF file: sky130hd/sky130_fd_sc_hd.lef, created 437 library cells
+[INFO IFP-0025] Defining die area using utilization: 5.00% and aspect ratio: 1.
+[INFO IFP-0029] Defining die area: (0, 0) to (377475, 377475) DBU
+[INFO IFP-0031] Defining core area: (2000, 2000) to (375475, 375475) DBU
```
eder-matheus: Showing these die and core areas in DBU are a bit confuse, since it comes before the snapping of the core area. Also, DBUs aren't that useful for the final user, IMO. I believe this could be turned into debug messages.

src/cts/test/array.ok
```diff
@@ -1,8 +1,16 @@
 [INFO ODB-0227] LEF file: Nangate45/Nangate45_tech.lef, created 22 layers, 27 vias
 [INFO ODB-0227] LEF file: Nangate45/Nangate45_stdcell.lef, created 135 library cells
 [INFO ODB-0227] LEF file: array_tile.lef, created 1 library cells
+[INFO IFP-0029] Defining die area: (0, 0) to (9880000, 9880000) DBU
+[INFO IFP-0031] Defining core area: (0, 0) to (9880000, 9880000) DBU
 [INFO IFP-0001] Added 3528 rows of 26000 site FreePDK45_38x28_10R_NP_162NW_34O.
 [INFO ODB-0303] The initial 3528 rows (91728000 sites) were cut with 225 shapes for a total of 52263 rows (14219856 sites).
+[INFO IFP-0012] Die BBox:  (  0.000  0.000 ) ( 4940.000 4940.000 ) um
+[INFO IFP-0013] Core BBox: (  0.000  0.000 ) ( 4940.000 4939.200 ) um
```
maliberty: Messages 29/31 seem to duplicate 12/13 (also agree with Eder about DBUs)

src/cts/test/array.ok
```diff
@@ -1,8 +1,16 @@
...
+[INFO IFP-0016] Core area:                     24399648.000 um^2
+[INFO IFP-0017] Instances area:                20260174.500 um^2
+[INFO IFP-0020] Result utilization:                   0.830
```
maliberty: "Result utilization" - what is it a result of?  I think "Utilization" is enough.

src/cts/test/array.ok
```diff
@@ -1,8 +1,16 @@
...
 [INFO ODB-0227] LEF file: Nangate45/Nangate45_stdcell.lef, created 135 library cells
 [INFO ODB-0227] LEF file: array_tile.lef, created 1 library cells
+[INFO IFP-0029] Defining die area: (0, 0) to (9880000, 9880000) DBU
+[INFO IFP-0031] Defining core area: (0, 0) to (9880000, 9880000) DBU
 [INFO IFP-0001] Added 3528 rows of 26000 site FreePDK45_38x28_10R_NP_162NW_34O.
 [INFO ODB-0303] The initial 3528 rows (91728000 sites) were cut with 225 shapes for a total of 52263 rows (14219856 sites).
+[INFO IFP-0012] Die BBox:  (  0.000  0.000 ) ( 4940.000 4940.000 ) um
+[INFO IFP-0013] Core BBox: (  0.000  0.000 ) ( 4940.000 4939.200 ) um
+[INFO IFP-0016] Core area:                     24399648.000 um^2
+[INFO IFP-0017] Instances area:                20260174.500 um^2
```
maliberty: `Total instance area` is clearer (it would be `Instance area` otherwise)

src/ifp/src/InitFloorplan.cc
```diff
@@ -716,38 +748,27 @@ void InitFloorplan::makeUniformRows(odb::dbSite* base_site,
...
           base_site->getName(),
           block_->dbuToMicrons(base_site->getHeight()));
     }
-    total_rows += make_rows(site);
+    make_rows(site);
   }
   block_->setCoreArea(block_->computeCoreArea());
-
-  if (total_rows == 0) {
-    logger_->error(IFP, 65, "No rows created in the core area.");
```
maliberty, line 749: Why is this removed?

maliberty (changes requested):

maliberty:
What is the purpose of `Rearrange error messages IDs to start at 300.` ?

src/cts/test/array.ok
gudeh: I think it is important because if we use the utilization mode to initiate floorplan the IFP may perform adjustments which modify the utilization a bit, not necessarily a perfect fit to the value given by the user. In fact I added one message for the received utilization and another for the resulting one.

src/cts/test/array.ok
gudeh: I had the impression the input values in config.mk were given in DBU, but I was mistaken. So I agree, no need for the DBU message.

gudeh:
> What is the purpose of `Rearrange error messages IDs to start at 300.` ?

I understand we don't have an official pattern, but at least for GPL I try to keep errors around 300, initialization messages starting from 1, final messages starting from 1000, etc. So it is only to keep them better organized.

src/ifp/src/InitFloorplan.cc
gudeh: removed by accident.

src/cts/test/array.ok
gudeh: For example, have a look at `src/grt/test/no_tracks.ok‎`

maliberty: How about "Adjusted utilization" or "Effective utilization"?

maliberty:
If we had started with your error numbering scheme I would be less concerned.  Renumbering existing messages can break any use of suppress_message, any log scraping scripts, web searches, or llm training.  I'd rather not change existing message numbers.

PR #9419, opened 2026-02-04
Refresh absl-cpp and tcmalloc

hzeller:


gadfort:
@maliberty due to the recent issues with this script, can we make sure this works correctly before merging?

hzeller:
Probably a good idea as I am doing the changes in the dependency script blind. I only use the `bazel` build.

hzeller:
the CI failures look unrealated.
I can't test the dependency script on my machine (I am using nix), so I'd say yolo it, then we can fix it later if there is an issue.

maliberty:
@gadfort I tested with your script from the previous issue and it built ok.

The pr-merge failure is the known flaky pdn test and is unrelated

PR #9421, opened 2026-02-04
odb: impl 3dblox checker checkFloatingChips

ahmed532:
**Connection-Aware Integrity**: Integrated dbChipConn and dbChipRegionInst data. The checker now recognizes that chips connected via valid 3DBlox connections are part of the same assembly, even if their bounding volumes do not physically intersect.
**Improved Diagnostics**: Floating sets are now sorted by size, identifying the largest group as the primary system.

**Tests**:
  * Validate basic floating chip detection using a manually constructed database (no parser dependencies).
  * Test the mathematical limit of connectivity versus connection thickness (bloat).

Related to: #9291

### DIFF
--- a/src/odb/src/3dblox/checker.cpp
+++ b/src/odb/src/3dblox/checker.cpp
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <ranges>
 #include <string>
 #include <unordered_map>
 #include <utility>
@@ -26,12 +27,14 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
 void Checker::check(dbChip* chip)
 {
   UnfoldedModel model(logger_, chip);
+  auto* top_cat = dbMarkerCategory::createOrReplace(chip, "3DBlox");
 
-  checkFloatingChips(chip, model);
-  checkOverlappingChips(chip, model);
+  checkFloatingChips(top_cat, model);
+  checkOverlappingChips(top_cat, model);
 }
 
-void Checker::checkFloatingChips(dbChip* chip, const UnfoldedModel& model)
+void Checker::checkFloatingChips(dbMarkerCategory* top_cat,
+                                 const UnfoldedModel& model)
 {
   const auto& chips = model.getChips();
   // Add one more node for "ground" (external world: package, PCB, ...)
@@ -68,19 +71,31 @@ void Checker::checkFloatingChips(dbChip* chip, const UnfoldedModel& model)
 
   std::vector<std::vector<const UnfoldedChip*>> groups(chips.size() + 1);
   for (size_t i = 0; i < chips.size(); ++i) {
-    groups[uf.find((int) i)].push_back(&chips[i]);
+    groups[uf.find(i)].push_back(&chips[i]);
   }
+  auto ground_leader = uf.find(ground_node);
+  const bool ground_empty = groups[ground_leader].empty();
+  groups.erase(groups.begin() + ground_leader);
+
   std::erase_if(groups, [](const auto& g) { return g.empty(); });
 
   std::ranges::sort(
-      groups, [](const auto& a, const auto& b) { return a.size() > b.size(); });
+      groups, [](const auto& a, const auto& b) { return a.size() < b.size(); });
 
-  if (groups.size() > 1) {
-    auto* top_cat = dbMarkerCategory::createOrGet(chip, "3DBlox");
-    auto* conn_cat = dbMarkerCategory::createOrGet(top_cat, "Connectivity");
-    auto* cat = dbMarkerCategory::createOrReplace(conn_cat, "Floating chips");
+  if (ground_empty) {
+    logger_->warn(
+        utl::ODB,
+        206,
+        "No ground group found. Erasing biggest group from floating chips.");
+    if (!groups.empty()) {
+      groups.pop_back();
+    }
+  }
+
+  if (!groups.empty()) {
+    auto* cat = dbMarkerCategory::createOrReplace(top_cat, "Floating chips");
     logger_->warn(utl::ODB, 151, "Found {} floating chip sets", groups.size());
-    for (const auto& group : groups) {
+    for (const auto& group : groups | std::views::reverse) {
       auto* marker = dbMarker::create(cat);
       for (auto* chip : group) {
         marker->addShape(Rect(chip->cuboid.xMin(),
@@ -94,7 +109,8 @@ void Checker::checkFloatingChips(dbChip* chip, const UnfoldedModel& model)
   }
 }
 
-void Checker::checkOverlappingChips(dbChip* chip, const UnfoldedModel& model)
+void Checker::checkOverlappingChips(dbMarkerCategory* top_cat,
+                                    const UnfoldedModel& model)
 {
   const auto& chips = model.getChips();
   std::vector<std::pair<const UnfoldedChip*, const UnfoldedChip*>> overlaps;
@@ -110,10 +126,7 @@ void Checker::checkOverlappingChips(dbChip* chip, const UnfoldedModel& model)
   }
 
   if (!overlaps.empty()) {
-    auto* top_cat = dbMarkerCategory::createOrGet(chip, "3DBlox");
-    auto* conn_cat = dbMarkerCategory::createOrGet(top_cat, "Connectivity");
-    auto* cat
-        = dbMarkerCategory::createOrReplace(conn_cat, "Overlapping chips");
+    auto* cat = dbMarkerCategory::createOrReplace(top_cat, "Overlapping chips");
     logger_->warn(
         utl::ODB, 156, "Found {} overlapping chips", (int) overlaps.size());
 
@@ -140,16 +153,18 @@ void Checker::checkOverlappingChips(dbChip* chip, const UnfoldedModel& model)
   }
 }
 
-void Checker::checkConnectionRegions(dbChip* chip, const UnfoldedModel& model)
+void Checker::checkConnectionRegions(dbMarkerCategory* top_cat,
+                                     const UnfoldedModel& model)
 {
 }
 
-void Checker::checkBumpPhysicalAlignment(dbChip* chip,
+void Checker::checkBumpPhysicalAlignment(dbMarkerCategory* top_cat,
                                          const UnfoldedModel& model)
 {
 }
 
-void Checker::checkNetConnectivity(dbChip* chip, const UnfoldedModel& model)
+void Checker::checkNetConnectivity(dbMarkerCategory* top_cat,
+                                   const UnfoldedModel& model)
 {
 }
 
@@ -193,10 +208,11 @@ bool Checker::isValid(const UnfoldedConnection& conn) const
     return false;
   }
   if (conn.top_region->isInternalExt() || conn.bottom_region->isInternalExt()) {
-    return std::max(conn.top_region->cuboid.zMin(),
-                    conn.bottom_region->cuboid.zMin())
-           <= std::min(conn.top_region->cuboid.zMax(),
-                       conn.bottom_region->cuboid.zMax());
+    return conn.top_region->parent_chip == conn.bottom_region->parent_chip
+           && std::max(conn.top_region->cuboid.zMin(),
+                       conn.bottom_region->cuboid.zMin())
+                  <= std::min(conn.top_region->cuboid.zMax(),
+                              conn.bottom_region->cuboid.zMax());
   }
 
   auto surfaces = getMatingSurfaces(conn);
--- a/src/odb/src/3dblox/checker.h
+++ b/src/odb/src/3dblox/checker.h
@@ -26,11 +26,16 @@ class Checker
     int top_z;
     int bot_z;
   };
-  void checkFloatingChips(dbChip* chip, const UnfoldedModel& model);
-  void checkOverlappingChips(dbChip* chip, const UnfoldedModel& model);
-  void checkConnectionRegions(dbChip* chip, const UnfoldedModel& model);
-  void checkBumpPhysicalAlignment(dbChip* chip, const UnfoldedModel& model);
-  void checkNetConnectivity(dbChip* chip, const UnfoldedModel& model);
+  void checkFloatingChips(dbMarkerCategory* top_cat,
+                          const UnfoldedModel& model);
+  void checkOverlappingChips(dbMarkerCategory* top_cat,
+                             const UnfoldedModel& model);
+  void checkConnectionRegions(dbMarkerCategory* top_cat,
+                              const UnfoldedModel& model);
+  void checkBumpPhysicalAlignment(dbMarkerCategory* top_cat,
+                                  const UnfoldedModel& model);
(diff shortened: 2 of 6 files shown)


src/odb/src/3dblox/checker.cpp
```diff
@@ -20,138 +23,143 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
 {
 }
 
-void Checker::check(odb::dbChip* chip)
+void Checker::check(dbChip* chip)
 {
   UnfoldedModel model(logger_, chip);
+  auto* top_cat = dbMarkerCategory::createOrReplace(chip, "3DBlox");
+  auto* conn_cat = dbMarkerCategory::createOrReplace(top_cat, "Connectivity");
```
osamahammad21: create only when there are markers

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,138 +23,143 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
-}
+  if (!conn.top_region || !conn.bottom_region) {
+    return true;
+  }
+  if (!conn.top_region->cuboid.xyIntersects(conn.bottom_region->cuboid)) {
+    return false;
+  }
+  if (conn.top_region->isInternalExt() || conn.bottom_region->isInternalExt()) {
+    return true;
+  }
```
osamahammad21: What if they are from different chiplets?

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,138 +23,143 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
 
-void Checker::checkNetConnectivity(const UnfoldedModel& model,
-                                   dbChip* chip,
-                                   dbMarkerCategory* category)
-{
-}
+  if ((conn.top_region->isInternal() || conn.bottom_region->isInternal())
```
osamahammad21: I don't think that's allowed

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,138 +23,143 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
-        utl::ODB, 156, "Found {} overlapping chips", (int) overlaps.size());
-
-    for (const auto& [inst1, inst2] : overlaps) {
-      odb::dbMarker* marker = odb::dbMarker::create(overlapping_chips_category);
-
-      auto cuboid1 = inst1->cuboid;
-      auto cuboid2 = inst2->cuboid;
-      auto intersection = cuboid1.intersect(cuboid2);
+  auto up = [](auto* r) { return r->isTop() || r->isInternal(); };
+  auto down = [](auto* r) { return r->isBottom() || r->isInternal(); };
```
osamahammad21: Again, according to the 3dblox standard, internal regions shouldn't be defined in the connections. INTERNAL_EXT could

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,138 +23,143 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
+      marker->setComment("Isolated chip set starting with "
+                         + groups[i][0]->name);
     }
   }
 }
 
-void Checker::checkOverlappingChips(const UnfoldedModel& model,
-                                    odb::dbMarkerCategory* category)
+Checker::ContactSurfaces Checker::getContactSurfaces(
+    const UnfoldedConnection& conn) const
```
osamahammad21: The function implementation is a unclear to me. Could you document?

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,138 +23,143 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
+void Checker::check(dbChip* chip)
 {
   UnfoldedModel model(logger_, chip);
+  auto* top_cat = dbMarkerCategory::createOrReplace(chip, "3DBlox");
+  auto* conn_cat = dbMarkerCategory::createOrReplace(top_cat, "Connectivity");
 
-  odb::dbMarkerCategory* category
-      = odb::dbMarkerCategory::createOrReplace(chip, "3DBlox");
-  checkFloatingChips(model, category);
-  checkOverlappingChips(model, category);
```
osamahammad21, line 30: Why did you remove the overlap check?

ahmed532: I wanted to simplify the tests and reduce the changes since the new overlap check is different from the current implementation. I'll add it again (improved) in the next PR.

maliberty:
Agree, it would be a more direct to test in c++ unit tests.

src/odb/src/3dblox/checker.cpp
osamahammad21: Please keep it as is and change it in the next PR.

src/odb/src/3dblox/checker.cpp
ahmed532: Done

src/odb/src/3dblox/checker.cpp
ahmed532: Good catch. did the z assert to make sure the connection is valid.

src/odb/src/3dblox/checker.cpp
ahmed532: Removed.

src/odb/src/3dblox/checker.cpp
ahmed532: Removed.

src/odb/src/3dblox/checker.cpp
ahmed532: I simplified and renamed the function for clarity.

src/odb/src/3dblox/checker.cpp
ahmed532: Restored.

ahmed532:
> Agree, it would be a more direct to test in c++ unit tests.

Ported the tcl tests to C++

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,73 +23,78 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
-                              inst->cuboid.xMax(),
-                              inst->cuboid.yMax()));
-        marker->addSource(inst->chip_inst_path.back());
+  std::erase_if(groups, [](const auto& g) { return g.empty(); });
+
+  std::ranges::sort(
+      groups, [](const auto& a, const auto& b) { return a.size() > b.size(); });
+
+  if (groups.size() > 1) {
+    auto* top_cat = dbMarkerCategory::createOrGet(chip, "3DBlox");
```
osamahammad21: Create the 3DBlox category at the check() function whether there are violations or not. However, the sub-categories should be created if and only if there are violations

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,73 +23,78 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
+    auto* conn_cat = dbMarkerCategory::createOrGet(top_cat, "Connectivity");
+    auto* cat = dbMarkerCategory::createOrReplace(conn_cat, "Floating chips");
```
osamahammad21: Merge into 1 sub-category. I don't see a need for that layered categorization of the violations.

src/odb/src/3dblox/checker.cpp
```diff
@@ -102,13 +110,15 @@ void Checker::checkOverlappingChips(const UnfoldedModel& model,
   }
 
   if (!overlaps.empty()) {
-    odb::dbMarkerCategory* overlapping_chips_category
-        = odb::dbMarkerCategory::createOrReplace(category, "Overlapping chips");
+    auto* top_cat = dbMarkerCategory::createOrGet(chip, "3DBlox");
```
osamahammad21: ditto

src/odb/src/3dblox/checker.cpp
```diff
@@ -102,13 +110,15 @@ void Checker::checkOverlappingChips(const UnfoldedModel& model,
...
+    auto* conn_cat = dbMarkerCategory::createOrGet(top_cat, "Connectivity");
+    auto* cat
+        = dbMarkerCategory::createOrReplace(conn_cat, "Overlapping chips");
```
osamahammad21: merge

src/odb/src/3dblox/checker.cpp
```diff
@@ -154,4 +160,53 @@ bool Checker::isOverlapFullyInConnections(const UnfoldedChip* chip1,
...
+  auto* bot = r1_down_r2_up ? r2 : r1;
+  return {
+      .valid = true, .top_z = top->getSurfaceZ(), .bot_z = bot->getSurfaceZ()};
+}
+
+bool Checker::isValid(const UnfoldedConnection& conn) const
+{
+  if (!conn.top_region || !conn.bottom_region) {
+    return true;
+  }
```
osamahammad21, line 206: Assume the following example:

Image
Conn1 would pass as valid although there it's impossible for r1 to be connected to the virtual plane.

src/odb/src/3dblox/checker.cpp
```diff
@@ -154,4 +160,53 @@ bool Checker::isOverlapFullyInConnections(const UnfoldedChip* chip1,
...
+  if (!conn.top_region->cuboid.xyIntersects(conn.bottom_region->cuboid)) {
+    return false;
+  }
+  if (conn.top_region->isInternalExt() || conn.bottom_region->isInternalExt()) {
+    return std::max(conn.top_region->cuboid.zMin(),
+                    conn.bottom_region->cuboid.zMin())
+           <= std::min(conn.top_region->cuboid.zMax(),
+                       conn.bottom_region->cuboid.zMax());
+  }
```
osamahammad21, line 216: There are many corner cases that would bypass this check. Assume they are from different parallel (on the z axis) chiplelts. We should check that they belong to the same chiplet and that they are touching (if the region layer is defined).

src/odb/src/3dblox/checker.h
```diff
@@ -17,26 +17,25 @@ class Checker
...
-
+  struct MatingSurfaces
+  {
+    bool valid;
+    int top_z;
+    int bot_z;
+  };
+  void checkFloatingChips(dbChip* chip, const UnfoldedModel& model);
+  void checkOverlappingChips(dbChip* chip, const UnfoldedModel& model);
+  void checkConnectionRegions(dbChip* chip, const UnfoldedModel& model);
```
osamahammad21: Isn't that covered by the Floatingchips check or is it gonna be a different check?

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,73 +23,78 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
+
+  std::ranges::sort(
+      groups, [](const auto& a, const auto& b) { return a.size() > b.size(); });
+
+  if (groups.size() > 1) {
+    auto* top_cat = dbMarkerCategory::createOrGet(chip, "3DBlox");
+    auto* conn_cat = dbMarkerCategory::createOrGet(top_cat, "Connectivity");
+    auto* cat = dbMarkerCategory::createOrReplace(conn_cat, "Floating chips");
+    logger_->warn(utl::ODB, 151, "Found {} floating chip sets", groups.size());
+    for (const auto& group : groups) {
```
osamahammad21: skip creating a violation for the first/largest group.

src/odb/test/cpp/Test3DBloxChecker.cpp
```diff
@@ -36,189 +34,206 @@ class CheckerFixture : public tst::Fixture
...
-  ASSERT_NE(category, nullptr);
-
-  auto floating_category = category->findMarkerCategory("Floating chips");
-  ASSERT_NE(floating_category, nullptr);
-
-  auto markers = floating_category->getMarkers();
-  EXPECT_EQ(markers.size(), 1);
+  check();
+  auto markers = getFloatingMarkers();
+  EXPECT_EQ(markers.size(), 2);
```
osamahammad21: should be 1 if previous comment is addressed

osamahammad21 (changes requested):

src/odb/src/3dblox/checker.cpp
osamahammad21: Tho that could be left for a separate check.

src/odb/src/3dblox/checker.cpp
ahmed532: OK.

src/odb/src/3dblox/checker.cpp
ahmed532: You mean under 3dblox we have Floating chips directly?

src/odb/src/3dblox/checker.cpp
ahmed532: I made it a separate check yes.

src/odb/src/3dblox/checker.cpp
ahmed532: regions belong to the same chiplet is sound. But, I'm not sure about touching. the standard says the interanl_ext regions can connect via long vias (pages 17 and 34). Also, check the picture on page 17.

src/odb/src/3dblox/checker.h
ahmed532: It will be a different check.

src/odb/src/3dblox/checker.cpp
ahmed532: I'm not sure, since interpreting this is hard. What if the smaller group is connected to ~ or what if they are equal. I think marking all is safer. Then the designer can decide which one is actually in fault.

src/odb/src/3dblox/checker.cpp
osamahammad21: yes

src/odb/src/3dblox/checker.cpp
osamahammad21: Actually you have a point, we could mark the group that is connected to the virtual plane as the base group and not mark it as a violation. However, I am not sure that the standard enforces defining such connection to the virtual plane. In case none exist, then we choose the largest group to be the base.

src/odb/src/3dblox/checker.cpp
osamahammad21: BTW, this is how detailed routing DRCs are handled and I want them both to follow the same strategy.

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,73 +24,93 @@ Checker::Checker(utl::Logger* logger) : logger_(logger)
...
+  std::erase_if(groups, [](const auto& g) { return g.empty(); });
+
+  std::ranges::sort(
+      groups, [](const auto& a, const auto& b) { return a.size() < b.size(); });
+
+  if (ground_empty) {
+    logger_->warn(
+        utl::ODB,
+        206,
+        "No ground group found. Erasing biggest group from floating chips.");
```
osamahammad21, line 89: ambiguous message. You may just warn that a connection to the virtual plane doesn't exist.

ahmed532: OK will fix it
"connection to the virtual plane doesn't exist"?

PR #9424, opened 2026-02-04
gui: allow focusing the nets of paths from the timing report

AcKoucher:
Also create a new check box to show the guides of the nets that are focused.

Example:

Path highlighted by the Timing Report selection (we don't know the metal layers used for the segments):
image

New menu for focusing the nets of the path:
image

Check "Nets" and uncheck "Timing Path":
image

Misc->Focused nets guides:
image

### DIFF
--- a/src/gui/src/renderThread.h
+++ b/src/gui/src/renderThread.h
@@ -12,6 +12,7 @@
 #include <cstdint>
 #include <map>
 #include <mutex>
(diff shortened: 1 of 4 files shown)


maliberty:
The gemini suggestions look reasonable

gadfort (approved):
LGTM

PR #9434, opened 2026-02-06
Fix: Multi-port read and multi-byte word bugs in RAM generator

tnguy19:
Fixed two bugs affecting multi-port RAMs and multi-byte words:

**1.Multi-port decoder bug, only port 0 output correct data:** 
- Decoder output only connected to selects[0] (port 0) 
- Added buffer with fanout to connect decoder signal to all read port

**2.Multi-byte bit indexing bug, only byte 0 of all words and read ports output correct data:** 
- Bit index calculated as byte_idx * 9 instead of byte_idx * 8 
- This cause error making higher bytes unreadable

**Files with changes:**
- ram.cpp: makeDecoder() and makeCellByte()
- .ok files : Updated golden files

**The new code tested and passed with these configs:**
- RAM 8x8 with 2,3,4 read ports
- RAM 8x16 with 2,3,4 read ports
- RAM 16x32 with 2, 3 read ports

@rovinski

### DIFF
--- a/src/ram/test/make_8x8.ok
+++ b/src/ram/test/make_8x8.ok
@@ -7,8 +7,6 @@
 [INFO RAM-0016] Selected clock gate cell sky130_fd_sc_hd__dlclkp_1
 [INFO RAM-0016] Selected buffer cell sky130_fd_sc_hd__buf_1
 [INFO PDN-0001] Inserting grid: ram_grid
-Found 0 macro blocks.
-Using 2 tracks default min distance between IO pins.
 [INFO PPL-0067] Restrict pins [ D[0] D[1] D[2] D[3] D[4] ... ] to region 0.00u-127.88u at the TOP edge.
 [INFO PPL-0001] Number of available slots 294
 [INFO PPL-0002] Number of I/O             29


src/ram/src/ram.cpp
```diff
@@ -131,7 +131,7 @@ void RamGen::makeCellByte(Grid& ram_grid,
   auto gclock_net = makeNet(prefix, "gclock");
   auto we0_net = makeNet(prefix, "we0");
 
-  int bit_idx = byte_idx * 9;
+  int bit_idx = byte_idx * 8;
```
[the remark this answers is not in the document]
tnguy19: Working on this , will update code with changes soon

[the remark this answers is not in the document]
tnguy19: Added suggested fix from gemini in 3751bb8

src/ram/src/ram.cpp
```diff
@@ -131,10 +131,16 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
+  const int physical_col_base
+      = byte_idx * 9;  // For placement taking into acount select bit of each
+                       // byte: 0, 9, 18, 27...
   for (int local_bit = 0; local_bit < 8; ++local_bit) {
-    const int global_bit_idx = bit_idx + local_bit;
-    auto name = fmt::format("{}.bit{}", prefix, global_bit_idx);
+    const int global_logical_bit_idx
+        = logical_bit_base + local_bit;                          // For naming
+    const int physical_col_idx = physical_col_base + local_bit;  // For
+                                                                 // placement
```
rovinski: Please do above-line comments rather than in-line comments unless the code+comment can all fit on a single line.

src/ram/src/ram.cpp
```diff
@@ -131,10 +131,16 @@ void RamGen::makeCellByte(Grid& ram_grid,
   auto gclock_net = makeNet(prefix, "gclock");
   auto we0_net = makeNet(prefix, "we0");
 
-  int bit_idx = byte_idx * 9;
+  const int logical_bit_base = byte_idx * 8;  // For naming bits: 0, 8, 16,...
```
rovinski: You, me, and @braydenlouie need to discuss a code refactoring, because this logical bit / physical bit distinction doesn't make sense from a software architecture perspective

src/ram/src/ram.cpp
```diff
@@ -210,24 +216,27 @@ std::unique_ptr<Cell> RamGen::makeDecoder(
...
     auto input_net = makeNet(prefix, fmt::format("layer_in{}", i));
     // sets up first AND gate, closest to byte's select + write enable gate
     if (i == 0 && i == layers - 1) {
+      makeCellInst(word_cell.get(),
+                   prefix,
+                   fmt::format("and_layer{}", i),
+                   and2_cell_,
+                   {{"A", addr_nets[i]},
+                    {"B", addr_nets[i + 1]},
+                    {"X", decoder_out_net}});
```
rovinski, line 255: This doesn't need to be fixed in this PR, but this is brittle because it can't be expected that the input and output ports have the names A/B/X. A more comprehensive solution would be to iterate over the input ports and then get the names from the pointer. @braydenlouie

rovinski (changes requested):
Minor comments

rovinski:
@tnguy19 can you post a screenshot of the layout before and after? I'm curious why the overall dimensions changed

Please also fix the merge conflict, and then mark the PR as ready to review.

tnguy19:
RAM8x8 before PR:
 **Without nets:**
image

**With nets:** 
image

RAM8x8 after the PR:
**Without nets:**
image

**With nets:**
image

The difference in area is probably from the addition of buffers for the decoders for each read port. So for a 2 read port RAM each word decoder has 2 buffer cells (one for each read port) so word 7 have decoder_7.buf_port0, decoder_7.buf_port1 that connect the signal to each read ports. So we have in total 8 words x 2 buffers = 16 buffer cells which cause the area to increase

rovinski:
Oh right, the example is for 2 read ports so of course it should change.

Looks like you need to fix up the .ok one more time then you can mark as ready for review

tnguy19:
I've been working on fixing the .ok files but I'm running into persistent issues that are causing the script to not generate the correct output. My `make_8x8.ok` file always produce these two output lines that I think was removed from the latest update:
```
Found 0 macro blocks.
```

I have tried pulling the updated code from master, re run the installer script and the setup.sh and rebuild using the --local flag but I couldn't get the ok files to not output those two lines to match the CI script.
Do you know of a fix or a way to setup Openroad flow script to ensure the local build matches this CI version?

rovinski:
The CI is just running using exactly what is on this fork/branch.

1. Make sure you have synced your fork's master to the upstream master
2. Make sure you have merged the master into the branch
3. Make sure your local copy is updated with the remote
4. Run the build making sure that you specify your fork and your branch with the flags, not just `--local`

rovinski:
@maliberty ready for review

@vvbandeira problem with the CI infrastructure

rovinski:
@tnguy19 please mark PR as ready for review

maliberty:
There is an unrelated issue on master that is being fixed.

PR #9441, opened 2026-02-08
dpl: debug deep iterative and new paintings

gudeh:
This PR is no-op.

It includes the following changes to debug mode:

- New option to paint occupied pixels during dpl debug mode: `-paint_pixels`
- New option to debug mode: `-deep_iterative`
  - The shiftMove() function is called when the main mapMove() function fails to find a suitable position for a target cell. The shiftMove() itself internally calls mapMove() multiple times to place cells surrounding the target cell. The deep_iterative option will stop every time an internal cell is positioned with the internal mapMove() call.
- Always paint unplaced cells in magenta.
- A few summarizing final messages for successful/failed mapMove() and shiftMove() calls

This PR also includes a new mode for DPL: `-incremental`, which initiates all cells in a legal position and without collisions as placed, instead of the current default which initiates all cells as unplaced (this option has functional implications, but is not used by default currently).


src/dpl/src/Opendp.tcl
```diff
@@ -4,11 +4,12 @@
 sta::define_cmd_args "detailed_placement" { \
                            [-max_displacement disp|{disp_x disp_y}] \
                            [-disallow_one_site_gaps] \
+                           [-incremental] \
```
maliberty, line 7: Add to docs

src/dpl/src/graphics/Graphics.cpp
```diff
@@ -86,7 +88,28 @@ void Graphics::drawObjects(gui::Painter& painter)
...
 
   for (const auto& cell : dp_->network_->getNodes()) {
-    if (!cell->isPlaced() || !cell->getDbInst()) {
+    if (!cell->getDbInst()) {
+      continue;
+    }
+
+    if (!cell->isPlaced()) {
+      auto color = gui::Painter::kDarkMagenta;
+      // color.a = 50;
```
maliberty: rm unused?

src/dpl/src/Opendp.cpp
```diff
@@ -278,6 +291,72 @@ void Opendp::findOverlapInRtree(const bgBox& queryBox,
...
 }
 
+void Opendp::setInitialGridCells()
+{
+  std::unordered_set<Node*> conflicted;
+  const DbuX site_width = grid_->getSiteWidth();
+
+  for (auto& node : network_->getNodes()) {
+    if (node->getType() == Node::CELL && !node->isFixed() && node->isPlaced()) {
+      if ((node->getLeft() % site_width != 0) || !checkInRows(*node)) {
```
maliberty: The row might not start on a multiple of site_width so this may be incorrect.

src/dpl/src/Opendp.cpp
```diff
@@ -278,6 +291,72 @@ void Opendp::findOverlapInRtree(const bgBox& queryBox,
...
+        continue;
+      }
+
+      bool node_conflicted = false;
+      grid_->visitCellPixels(
+          *node, false, [&](Pixel* pixel, [[maybe_unused]] bool padded) {
+            if (pixel->cell != nullptr && pixel->cell != node.get()) {
+              node_conflicted = true;
+              if (!pixel->cell->isFixed()) {
+                conflicted.insert(pixel->cell);
```
maliberty, line 386: Is it not a conflict to be on a fixed object?

src/dpl/src/Opendp.tcl
gudeh: of course! I intend to investigate in the future to include this option to CTS and GRT calls to DPL. I hope it will provide less displacement.

src/dpl/src/Opendp.cpp
gudeh: what if we do the following, the snapping to the grid pixels has to be same as the cell position.
```
      const GridX x = grid_->gridX(node.get());
      const GridY y = grid_->gridSnapDownY(node.get());
      if (node->getLeft() != gridToDbu(x, site_width)
          || node->getBottom() != grid_->gridYToDbu(y)
```

src/dpl/src/Opendp.cpp
gudeh: This ensures we do not attempt to unplace fixed cells. In other words, when `!pixel->cell->isFixed()` is true (the cell is movable), we mark both cells as conflicted. But if the cell already in the pixel is fixed, the condition is false, and only the current node is marked for unplacement.

PR #9452, opened 2026-02-10
rsz: design area include tap

gudeh:
We are currently not accounting for end caps and tap cells when reporting the area and utilization for the user.

The modified rsz function is used for the command used multiple times in ORFS: `report_design_area`. 

The `report_design_area` is called by `report_metrics`, and has the following format:

image

I triggered a secure-CI and this seems to be no-op.


maliberty:
The change is trivial but will need a secure CI check

gudeh:
@maliberty this actually did not require metrics update on secure-CI. It seems we use another method to calculate areas for secure-CI metrics.

maliberty:
Is this consistent with the method used by metrics?  Can we consolidate the two?

gudeh:
I noticed the area does **not** match for some private designs, but does for public ones.

Here is a small summary for potential future reference:

- `report_metrics` is costly and reports multiple things, it also calls `report_design_area`. 
- `report_design_area` is calculated with RSZ with the function I modified in this PR: `Resizer::computeDesignArea()`, I understand it is not costly in runtime, and only iterates over `block_->getInsts()` once.
- OR submodule flow tests in `OpenROAD/test/` uses `Resizer::computeDesignArea()` for reporting metrics.
  - This is why we required test updates in this PR.
- ORFS metrics on the other hand uses another command with a similar name: `report_design_area_metrics`. It iterates over instances and checks the area directly on a TCL script in `OpenROAD/src/Metrics.tcl` at [these lines](https://github.com/The-OpenROAD-Project/OpenROAD/blob/2543b4ef55b3ef2788a1e49764db3a5063ae3c07/src/Metrics.tcl#L188C18-L301) with the TCL loop: `foreach inst [$block getInsts] {`.

Finally, I checked a few designs and the areas reported by the two commands `report_design_area` and `report_design_area_metrics` match for pubic designs. 

For example, asap7/aes:
OR metric: `Design area 1998 um^2 48% utilization.`
ORFS metric: `[INFO] finish__design__instance__area pass test: 1997.66 <= 2269`

This can be easily check by the end of secure-CI report, both are a few lines apart:
image
image

maliberty:
@gudeh can we replace the tcl one with the OR one?  It is better to have a single method for this.

gudeh:
> @gudeh can we replace the tcl one with the OR one? It is better to have a single method for this.

The C++ version (used in OR test) named `report_design_area` exclusively calculates the instances area.

While the TCL version (used in ORFS metrics) named `report_design_area_metrics` is a lot more complex, calculating the area for all of these:
```
  utl::metric_int "design__io" $num_ios
  utl::metric_float "design__die__area" $die_area
  utl::metric_float "design__core__area" $core_area
  utl::metric_int "design__instance__count" $num_insts
  utl::metric_float "design__instance__area" $total_active_area
  utl::metric_int "design__instance__count__stdcell" $num_stdcells
  utl::metric_float "design__instance__area__stdcell" $stdcell_area
  utl::metric_int "design__instance__count__macros" $num_macros
  utl::metric_float "design__instance__area__macros" $macro_area
  utl::metric_int "design__instance__count__padcells" $num_padcells
  utl::metric_float "design__instance__area__padcells" $padcell_area
  utl::metric_int "design__instance__count__cover" $num_cover
  utl::metric_float "design__instance__area__cover" $cover_area
  utl::metric_float "design__instance__utilization" $core_util
  utl::metric_float "design__instance__utilization__stdcell" $stdcell_util
```

Do you think we should have only one version somehow? They do not provide the exact same result (TCL provides more things).

Also, I suspect the area mismatch in private designs has to do with identifying fillers during CTS (I mentioned this briefly last week).

PR #9461, opened 2026-02-12
3Dblox: Do not use layer stack in extracting the cuboid of the region

osamahammad21:
I’ve just noticed that most tech LEFs typically don’t specify layer thickness, which makes relying on them to derive a chip region’s thickness and Z-offset obscure. While the removed logic is valid when thickness values are present, it would rarely hold in practice, since it would usually end up assuming zero thickness for all layers.


maliberty:
Even if specified it wouldn't necessarily tell you the thickness of the chip (eg there is no substrate thickness).

osamahammad21:
> Even if specified it wouldn't necessarily tell you the thickness of the chip (eg there is no substrate thickness).

Ahh. then there is really no way to know. Noted.

Issue #8483, opened 2025-09-30
Evaluate QoR impact of removing push-to-edges post-process in MPL


AcKoucher:
### Description

This post-process was created, because we weren't seeing too much peripheral pushing with the annealer on its own. However, after many bug fixes the annealing results changed a lot and it's not clear that we're getting benefits from this extra step.

PR #9465, opened 2026-02-12
mpl: treat virtual pin access blockages as soft constraints

### LINKED ISSUE #9226: Different types of blockages should be treated differently in MPL
### Description

Currently pin access blockages are treated as hard constraints, i.e., area of pin access is inaccessible for std cell clusters, which generates a clear discrepancy between MPL and GPL behavior, as GPL will definitely be able to use an empty area to place std cells.

### Suggested Solution

Virtual pin access blockages generated by MPL should work better as soft constraints i.e., clusters of std cells should be allowed to occupy the area of pin access blockage.

### Additional Context

See #9214.

### LINKED ISSUE #9985: RTLMP gives error MPL-0040 with very easy floorplan problem
### Describe the bug

`rtl_macro_placer` fails with `MPL-0040` (annealing engine failed to find a valid solution) on a design with 70 small FakeRAM SRAM macros that occupy only 1.5% of the core area. The failure occurs at the root cluster level, suggesting the partitioning fails before any sub-clustering is attempted.

The design is a PULP Platform snitch_cluster (9 RISC-V cores with SRAM banks) targeting ASAP7 with hierarchical synthesis. The macros are very small relative to the die — the largest is 99x35 um in a 1500x1500 die.

### Design characteristics

| Parameter | Value |
|-----------|-------|
| Platform | ASAP7 |
| Die area | 1500 x 1500 um |
| Std cell instances | 3,268,160 |
| Macros | 70 |
| Macro area | 33,330 um² (1.5% of core) |
| Macro area with halos | 47,263 um² (2.1% of core) |
| Design utilization | 0.18 |
| Floorplan utilization | 0.17 |
| Hierarchical synthesis | yes |

### Macro types

The 70 macros are 4 FakeRAM types, all very small relative to the 2.22M um² core:

| Macro | Size (um) | Area (um²) | % of core |
|-------|-----------|-----------|-----------|
| fakeram7_128x48 | 9.3 x 16.4 | 153 | 0.007% |
| fakeram7_256x64 | 12.4 x 31.1 | 387 | 0.017% |
| fakeram7_256x256 | 49.8 x 41.5 | 2,064 | 0.093% |
| fakeram7_64x512 | 99.5 x 34.6 | 3,440 | 0.155% |

### What was tried

All of the following produce the same `MPL-0040` error:

| Parameter | Values tried |
|-----------|-------------|
| PLACE_DENSITY | 0.35, 0.40 |
| Die area | auto-sized, explicit 1500x1500 |
| max_num_level | 2 (default), 5 |
| target_util | 0.35, 0.40 |
| halo | 2, 5 |

### rtl_macro_placer command

```tcl
rtl_macro_placer \
    -max_num_level 5 \
    -halo_width 2 \
    -halo_height 2 \
    -target_util 0.35
```

### Environment

- OpenROAD 26Q1-1929-g7cf7b720fd
- Docker image: openroad/orfs:26Q1-656-g97416b2e4

### Expected Behavior

It should easily find a floorplan since it is loosely constrainted.

### Environment

```shell
This is run in Docker image: openroad/orfs:26Q1-656-g97416b2e4
```

### To Reproduce

Download the standalone test case from: [mpl0040-repro.tar.gz](https://drive.google.com/file/d/186WBvaIyGSFqj4ET4jloD3vtMwnGru4e/view?usp=drivesdk)]

```bash
tar xzf mpl0040-repro.tar.gz
cd mpl0040-repro
docker run --rm -v $(pwd):/work -w /work openroad/orfs:26Q1-656-g97416b2e4 bash run-me.sh
```

Sorry it is so big! I couldn’t create a smaller test case.

The test case includes a pre-floorplanned ODB and a minimal TCL script that calls `rtl_macro_placer`. Runtime: ~3 minutes to load, then fails.

### Relevant log output

```shell
Die Area: (0.00, 0.00) (1500.00, 1500.00),  Floorplan Area: (5.02, 5.13) (1494.99, 1494.99)
	Number of std cell instances: 3268160
	Area of std cell instances: 375384.17
	Number of macros: 70
	Area of macros: 33329.88
	Default halo (L, B, R, T): (2.00, 2.00, 2.00, 2.00)
	Area of macros with halos: 47263.46
	Area of std cell instances + Area of macros: 408714.05
	Floorplan area: 2219843.72
	Design Utilization: 0.18
	Floorplan Utilization: 0.17
	Manufacturing Grid: 1

[ERROR MPL-0040] Annealing engine failed to find a valid solution.
Cluster Id: 0
 Cluster Name: root
```

### Screenshots

_No response_

### Additional Context

- Similar failures reported in #5655, #5884, #5996, #6055 (MegaBoom on ASAP7).
- Unlike MegaBoom, these macros are extremely small (1.5% of core) — the annealing engine may be struggling because the macros are negligible relative to the floorplan.
- There is currently no way to skip `rtl_macro_placer` when it fails, since commit e308334c removed the `else` branch that made `MACRO_PLACEMENT_TCL` an alternative to RTLMP. A way to disable RTLMP when it cannot handle a design would be helpful.

AcKoucher:
See #9214 for the rationale.

Resolve #9226.
Resolve #9985.

I thought it was time to update the naming convention w.r.t. the blockages (the original name `macro_blockages` was very misleading and I think made understanding what's going on much harder). Now the annealer deals with:
 - Soft Blockages: Virtual pin access blockages generated for MPL;
 - Hard Blockages: Fixed macros and placement blockages;

I'm removing two tests (`boundary_push2` and `boundary_push3`) whose purpose was to check if the overlap with io blockages was properly avoided when applying the Pusher post-process. However, with the changes here, these tests started working worked in a hackish way due to how the original implementation treated "tiny" std cell clusters (we'd benefit from the fact that these clusters have no area and the centralization of the macros would happen regardless of the position of that cluster).

As we'll eventually try to get rid of this post process (see #8483) I don't think this coverage is that important. If desired we can come up with a different test suite (maybe in Cpp as the architecture allows it).

### DIFF
--- a/src/mpl/src/SACoreSoftMacro.cpp
+++ b/src/mpl/src/SACoreSoftMacro.cpp
@@ -996,13 +996,10 @@ void SACoreSoftMacro::printResults() const
           boundary_weight_,
           boundary_penalty_,
           norm_boundary_penalty_});
-  report({"Soft Blockage",
-          soft_blockage_weight_,
-          soft_blockage_penalty_,
-  report({.name="Soft Blockage",
-          .weight=soft_blockage_weight_,
-          .value=soft_blockage_penalty_,
-          .normalization_factor=norm_soft_blockage_penalty_});
+  report({.name = "Soft Blockage",
+          .weight = soft_blockage_weight_,
+          .value = soft_blockage_penalty_,
+          .normalization_factor = norm_soft_blockage_penalty_});
   report({"Notch", notch_weight_, notch_penalty_, norm_notch_penalty_});
   reportTotalCost();
   if (logger_->debugCheck(MPL, "hierarchical_macro_placement", 2)) {


AcKoucher:
@joaomai FYI.

AcKoucher:
Running Secure-CI.

src/mpl/src/mpl.tcl
```diff
@@ -151,8 +151,8 @@ proc rtl_macro_placer { args } {
   if { [info exists keys(-notch_weight)] } {
     set notch_weight $keys(-notch_weight)
   }
-  if { [info exists keys(-macro_blockage_weight)] } {
-    set macro_blockage_weight $keys(-macro_blockage_weight)
+  if { [info exists keys(-soft_blockage_weight)] } {
+    set soft_blockage_weight $keys(-soft_blockage_weight)
```
maliberty, line 168: It would be good to also accept the old name for compatibility.   It doesn't need to be documented.

PR #9469, opened 2026-02-14
odb: refactor 3dblox checkOverlappingChips

ahmed532:
## refactors the 3DBlox checkOverlappingChips
Significantly improve performance.

### Changes:
 * Logic Fix: Relaxed Checker::isValid to allow INTERNAL_EXT connections between different parent chips, provided their Z-ranges intersect. 
 * Use a line sweep algorithm for overlap checking for better performance.
 * Testing: Added comprehensive test cases to Test3DBloxChecker covering:
   * Complex multi-chip overlap scenarios.

### Validation
Passed all tests in Test3DBloxChecker.
Verified that existing functionality for floating chips and simple overlaps remains correct.

### Related to: #9291

### DIFF
--- a/src/odb/src/3dblox/checker.cpp
+++ b/src/odb/src/3dblox/checker.cpp
@@ -21,6 +21,58 @@
 
 namespace odb {
 
+namespace {
+
+MatingSurfaces getMatingSurfaces(const UnfoldedConnection& conn)
+{
+  auto* r1 = conn.top_region;
+  auto* r2 = conn.bottom_region;
+  if (!r1 || !r2) {
+    return {.valid = false, .top_z = 0, .bot_z = 0};
+  }
+
+  // r1 faces down (Bottom side) and r2 faces up (Top side) -> r1 is above r2
+  bool r1_down_r2_up = r1->isBottom() && r2->isTop();
+  // r1 faces up (Top side) and r2 faces down (Bottom side) -> r2 is above r1
+  bool r1_up_r2_down = r1->isTop() && r2->isBottom();
+
+  if (r1_down_r2_up == r1_up_r2_down) {
+    return {.valid = false, .top_z = 0, .bot_z = 0};
+  }
+
+  auto* top = r1_down_r2_up ? r1 : r2;
+  auto* bot = r1_down_r2_up ? r2 : r1;
+  return {
+      .valid = true, .top_z = top->getSurfaceZ(), .bot_z = bot->getSurfaceZ()};
+}
+
+bool isValid(const UnfoldedConnection& conn)
+{
+  if (!conn.top_region || !conn.bottom_region) {
+    return true;
+  }
+  if (!conn.top_region->cuboid.xyIntersects(conn.bottom_region->cuboid)) {
+    return false;
+  }
+  if (conn.top_region->isInternalExt() || conn.bottom_region->isInternalExt()) {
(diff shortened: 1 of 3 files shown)


src/odb/src/3dblox/checker.cpp
```diff
@@ -168,10 +183,47 @@ void Checker::checkNetConnectivity(dbMarkerCategory* top_cat,
...
+                     r->cuboid.xMax(),
+                     r->cuboid.yMax())
+                    .contains(overlap_rect);
+    };
+    if (valid(r1) || valid(r2)) {
+      return true;
+    }
+  }
   return false;
 }
```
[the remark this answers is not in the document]
osamahammad21: please define externally in a namespace

src/odb/src/3dblox/checker.cpp
```diff
@@ -168,10 +183,47 @@ void Checker::checkNetConnectivity(dbMarkerCategory* top_cat,
...
+      return r->isInternalExt()
+             && Rect(r->cuboid.xMin(),
+                     r->cuboid.yMin(),
+                     r->cuboid.xMax(),
+                     r->cuboid.yMax())
+                    .contains(overlap_rect);
+    };
+    if (valid(r1) || valid(r2)) {
+      return true;
+    }
```
osamahammad21: why would this be valid?

osamahammad21 (changes requested):
I feel that using connections to verify overlap violations is an overkill. It is simply a violation if 2 chiplets overlap at any condition.

ahmed532:
The spec allows embedded chiplets connected via internal_ext. That means one chiplet completely inside another -- an overlap. But that is not a violation if the internal_ext region covers the overlap. You can find examples in the tests.

src/odb/src/3dblox/checker.cpp
ahmed532: The spec allows it like in the diagram on page 18 cases 1 and 3. We define the internal_ext region to span the whole chip z (for now) and this check ensure that it's x and y span the whole overlap (embedded chip).

src/odb/src/3dblox/checker.cpp
```diff
@@ -20,6 +21,58 @@
...
+    return true;
+  }
+  if (!conn.top_region->cuboid.xyIntersects(conn.bottom_region->cuboid)) {
+    return false;
+  }
+  if (conn.top_region->isInternalExt() || conn.bottom_region->isInternalExt()) {
+    return std::max(conn.top_region->cuboid.zMin(),
+                    conn.bottom_region->cuboid.zMin())
+           <= std::min(conn.top_region->cuboid.zMax(),
+                       conn.bottom_region->cuboid.zMax());
```
osamahammad21, line 61: Didn't we agree on removing this and checking if they are of the same chiplet?

ahmed532: Which part exactly should be removed? I think the check that z positions are correct is useful. 
If we enforce that both regions are in the same chiplet, then internal_ext would have the exact same semantics as internal. I know we discussed this together with Matt, but I'm not sure this extra check would be useful. Given that overlap checking would catch any embedded chiplets. 
But anyway, I leave it up to you. Enforce it?

osamahammad21: we shouldn't allow the idea of "embedded chiplets" altogether. There should never be a chip inside another chip in the unfolded model.

PR #9473, opened 2026-02-16
Update version of qt prebuilt to latest.

hzeller:
The qt library was using xcb headers, but they were not declared in the dependencies
(fixed in
https://github.com/The-OpenROAD-Project/qt_bazel_prebuilts/pull/6 )

Also, update harfbuzz and glib dependencies that I've fixed upstream and work better with with Nix (glib) and don't have missing symbols (harfbuzz).


hzeller:
To preempt the question: no, this will not make it possible to use `<QOpenGLWidget>` which recently got added in OpenROAD. So gui-build is still broken.

This just fixes general dependencies.

PR #9474, opened 2026-02-16
Use abc from BCR instead of third-party/

hzeller:
BCR version is a fairly recent implementation from the actively maintained Yosys branch. It comes with its own BUILD file that we don't have to maintain here.

Putting third-party/abc into bazelignore so that we don't process that part anymore.

With that, we now can also get rid of the very outdated `readline` and `ncurses` dependencies in `WORKSPACE`, as they now all come (much newer) from BCR.
(both of them I updated recently upstream in BCR)


hzeller:
steps toward further reducing `WORKSPACE` dependencies.
I talked with @QuantamHD and he said it is ok to use the Yosys version.

Also CC:  @oharboe @mikesinouye

oharboe:
😌

PR #9479, opened 2026-02-16
gui: Replace OpenGL widget with software rendered 3dViewPort

QuantamHD:
Replacing the OpenGL component for 3d chiplets with a software rendered one instead.

image

### DIFF
--- a/src/gui/src/chiplet3DWidget.cpp
+++ b/src/gui/src/chiplet3DWidget.cpp
@@ -10,6 +10,7 @@
 #include <array>
 #include <cmath>
 #include <cstddef>
+#include <cstdint>
 
 #include "odb/db.h"
 #include "odb/dbTransform.h"


src/gui/src/chiplet3DWidget.cpp
```diff
@@ -116,42 +107,29 @@ void Chiplet3DWidget::buildGeometries()
...
-    indices_lines_.push_back(base + 3);
-    indices_lines_.push_back(base + 7);
+    // Add line indices for a cube (12 lines)
+    const int lines[24] = {0, 1, 1, 2, 2, 3, 3, 0,   // Bottom face
+                           4, 5, 5, 6, 6, 7, 7, 4,   // Top face
+                           0, 4, 1, 5, 2, 6, 3, 7};  // Connecting pillars
+
+    for (int i : lines) {
+      indices_lines_.push_back(base + i);
+    }
```
maliberty, line 119: You could use std::array and insert at once rather than looping w/ one at a time push

src/gui/src/chiplet3DWidget.h
```diff
@@ -4,52 +4,57 @@
...
 
  protected:
-  void initializeGL() override;
-  void paintGL() override;
-
-  void mousePressEvent(QMouseEvent* e) override;
-  void mouseReleaseEvent(QMouseEvent* e) override;
-  void mouseMoveEvent(QMouseEvent* e) override;
-  void wheelEvent(QWheelEvent* e) override;
+  void paintEvent(QPaintEvent* event) override;  // Replaces paintGL
```
maliberty: No need to mention paintGL

src/gui/src/chiplet3DWidget.cpp
```diff
@@ -167,44 +145,123 @@ void Chiplet3DWidget::paintGL()
...
+  // 2. Clip against Near Plane
+  // In OpenGL view space, camera looks down -Z.
+  // Near plane is at z = -nearVal (e.g. -10.0). Points with z > -nearVal are
+  // behind the near plane. However, QMatrix4x4::perspective sets up a standard
+  // frustum. We can extract the near plane distance from our setup, but a small
+  // fixed epsilon usually works for preventing divide-by-zero or "behind head"
+  // artifacts.
+
+  // Simple clipping: Clip against Z = -0.1 (very close to camera)
+  const float CLIP_Z = -0.1f;
```
maliberty: kClipZ

src/gui/src/chiplet3DWidget.cpp
```diff
@@ -167,44 +145,123 @@ void Chiplet3DWidget::paintGL()
...
+
+  bool p1_visible = p1_view.z() < CLIP_Z;
+  bool p2_visible = p2_view.z() < CLIP_Z;
```
maliberty: const here and various vars below

src/gui/src/chiplet3DWidget.cpp
```diff
@@ -167,44 +145,123 @@ void Chiplet3DWidget::paintGL()
...
+
+    if (!p1_visible) {
+      p1_view = intersection;
+    } else {
+      p2_view = intersection;
+    }
+  }
+
+  QVector3D p1_ndc = projection.map(p1_view);
+  QVector3D p2_ndc = projection.map(p2_view);
```
maliberty: What is ndc?

maliberty (changes requested):

maliberty:
This is to resolve a concern about servers that have no OpenGL available.

src/gui/src/chiplet3DWidget.cpp
QuantamHD: You still need to add base to each element, and that requires looping one way or another. This also only occurs if geometry is changing so it's not really a performance issue IMO.

I did try it your way, but adding base to each element in the std::array was way messier and adding a for loop to modify the array afterwards just felt confusing.

PR #9501, opened 2026-02-19
Use std::span, reducing one boost dependency

hzeller:


QuantamHD:
Why absl::Span instead of std::span? This code predates the move to c++20

hzeller:
Ah, right, with c++20 we can use regular span
(I just wanted to match what we did at Google).

maliberty (approved):

PR #9506, opened 2026-02-20
Tell compilation db where to find the Python.h includes.

hzeller:
There was one diagnostic error when running clang-tidy that was due to not finding Python.h. Add it to the `-I` paths.

### DIFF
--- a/etc/bazel-make-compilation-db.sh
+++ b/etc/bazel-make-compilation-db.sh
@@ -49,6 +49,16 @@ done >> compile_flags.txt
 # Qt include files check for this
 echo '-fPIC' >> compile_flags.txt
 
+# Python include bindings.
+for f in bazel-out/../../../external/*/include/python3.*/Python.h; do
+  if [ -f "${f}" ]; then
+    PY_INC="$(dirname "${f}")"
+    echo "-I${PY_INC}"
+    echo "-I$(realpath "${PY_INC}")"  # work around clangd bug
+    break
+  fi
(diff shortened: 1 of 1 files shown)


etc/bazel-make-compilation-db.sh
```diff
@@ -49,6 +49,12 @@ done >> compile_flags.txt
 # Qt include files check for this
 echo '-fPIC' >> compile_flags.txt
 
+# Python include bindings.
+PYTHON_INC=$(dirname bazel-out/../../../external/*/include/python3.*/Python.h | head -1)
+if [ ! -z "${PYTHON_INC}" ]; then
+  echo "-I${PYTHON_INC}" >> compile_flags.txt
+fi
```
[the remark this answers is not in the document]
hzeller: done.

etc/bazel-make-compilation-db.sh
```diff
@@ -49,6 +49,16 @@ done >> compile_flags.txt
 # Qt include files check for this
 echo '-fPIC' >> compile_flags.txt
 
+# Python include bindings.
+for f in bazel-out/../../../external/*/include/python3.*/Python.h; do
```
maliberty, line 53: `bazel-out/..` is a no-op, what is the intention in including it?

hzeller: `bazel-out/` is a symlink that points to the bazel internal directories.

PR #9513, opened 2026-02-20
Add elapsed time reporting for long-running commands

### LINKED ISSUE #5003: Long running OR commands should report their runtime
### Description

Every “long running” command should report how long it took to run it.
The goal would be to be able to easily compare log files from different runs.
E.g.
vimdiff dir[12]/log
or
grep runtime dir2/log > dir2/runtime
diff dir[12]/runtime

### Suggested Solution

Currently there seems to be a way to get ORFS to report runtime and memory usage.
But this relies on running the openroad executable for each and every step.
And it's not "built into" openroad, it relies on external scripting.
It should be possible to have one OR tcl script to run "the whole flow" and still get this information in the log.
```
Log                            Elapsed seconds Peak Memory/KB
1_1_yosys                                  243         971412
2_1_floorplan                               27         510688
2_2_floorplan_io                             4         363812
2_3_floorplan_tdms                           4         362360
2_4_floorplan_macro                          4         367364
2_5_floorplan_tapcell                        4         325804
2_6_floorplan_pdn                            6         381020
3_1_place_gp_skip_io                        38         461672
3_2_place_iop                                4         372996
3_3_place_gp                               613        1048224
3_4_place_resized                           94         602152
3_5_place_dp                                78         657092
4_1_cts                                    152         747132
5_1_grt                                   6137        1298604
5_2_fillcell                                 5         494576
5_3_route                                  543       11470100
6_1_merge                                   15         815860
6_report                                   198        2751824
Total                                     8169       11470100
```

### Additional Context

_No response_

Bbn08:
Add INFO-level elapsed time logging to 9 long-running OpenROAD commands using utl::Timer:

- DPL: Opendp::detailedPlacement (msg 500)
- CTS: TritonCTS::runTritonCts (msg 500)
- GRT: GlobalRouter::globalRoute (msg 501)
- DRT: TritonRoute::main (msg 501)
- GPL: Replace::doPlace (msg 500)
- MPL: MacroPlacer::place (msg 500)
- RSZ: Resizer::repairDesign (msg 500)
- PDN: PdnGen::buildGrids (msg 500)
- IFP: InitFloorplan::initFloorplan (msg 500)

Each function logs 'Elapsed time: X.XXs' at INFO level upon completion, providing users with runtime visibility without requiring verbose mode.

Closes #5003


src/grt/src/GlobalRouter.cpp
```diff
@@ -59,6 +59,7 @@
 #include "utl/CallBackHandler.h"
 #include "utl/Logger.h"
 #include "utl/algorithms.h"
+#include "utl/timer.h"
```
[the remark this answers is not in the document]
Bbn08: Resolved — refactored `globalRoute` to use `utl::Timer` instead of manual `std::chrono`. The include is now actively used: `utl::Timer timer;` at function entry, and both the existing verbose log (msg 303) and the new elapsed time log (msg 501) use `timer.elapsed()`.

src/grt/src/GlobalRouter.cpp
```diff
@@ -389,16 +389,15 @@ void GlobalRouter::globalRoute(bool save_guides)
...
-    int min = (runtime.count() % 3600) / 60;
-    int sec = runtime.count() % 60;
+    const int elapsed = static_cast<int>(timer.elapsed());
+    const int hour = elapsed / 3600;
+    const int min = (elapsed % 3600) / 60;
+    const int sec = elapsed % 60;
     logger_->info(
         GRT, 303, "Global routing runtime = {:02}:{:02}:{:02}", hour, min, sec);
   }
+  logger_->info(GRT, 501, "Runtime: {:.2f}s", timer.elapsed());
```
maliberty: We are already reporting runtime here under -verbose, why add it again? 

Note that existing h:m:s is easier to read the just a total number of seconds.  It might be nice to do that generally.  You could make a method on Timer to return such a string.

maliberty (changes requested):

maliberty:
The PDN change seems unrelated and should be its own PR

src/pdn/src/PdnGen.cc
```diff
@@ -60,6 +61,7 @@ void PdnGen::resetShapes()
 
 void PdnGen::buildGrids(bool trim)
 {
+  utl::Timer timer;
```
gadfort: this only captures one "phase" of the build, it seems like what you want is to measure the time from the start of buildGrids through writeDb (which can also make a lot of time)

src/grt/src/GlobalRouter.cpp
Bbn08: Removed, the existing verbose h:m:s line is sufficient. Will follow up with a Timer::toString() method in a separate cleanup if desired.

src/pdn/src/PdnGen.cc
Bbn08: Fixed in the separate PDN pr (shall be pushing shortly) added `PdnGen::run()` that wraps `buildGrids + writeToDb` under a single timer, so the reported runtime covers both phases.

maliberty:
Please merge master which should fix the pr-merge error

PR #9515, opened 2026-02-21
odb: impl 3dblox checker checkConnectionRegions

ahmed532:
Check all regions in all connections are well defined. 

Related to: #9291

### DIFF
--- a/src/odb/src/3dblox/checker.cpp
+++ b/src/odb/src/3dblox/checker.cpp
@@ -224,10 +224,14 @@ void Checker::checkOverlappingChips(dbMarkerCategory* top_cat,
 void Checker::checkInternalExtUsage(dbMarkerCategory* top_cat,
                                     const UnfoldedModel& model)
 {
-  auto* cat = dbMarkerCategory::createOrReplace(top_cat, "Unused internal_ext");
+  dbMarkerCategory* cat = nullptr;
   for (const auto& chip : model.getChips()) {
     for (const auto& region : chip.regions) {
       if (region.isInternalExt() && !region.isUsed) {
+        if (!cat) {
+          cat = dbMarkerCategory::createOrReplace(top_cat,
+                                                  "Unused internal_ext");
+        }
         logger_->warn(utl::ODB,
                       464,
                       "Region {} is internal_ext but unused",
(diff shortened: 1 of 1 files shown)


src/odb/src/3dblox/checker.cpp
```diff
@@ -233,6 +249,49 @@ void Checker::checkInternalExtUsage(dbMarkerCategory* top_cat,
...
+    return fmt::format("{}/{} (faces {})",
+                       r->parent_chip->name,
+                       r->region_inst->getChipRegion()->getName(),
+                       sideToString(r->effective_side));
+  };
+  int count = 0;
+  for (const auto& conn : model.getConnections()) {
+    if ((!conn.top_region || !conn.bottom_region)
+        && conn.connection->getTopRegion()
+        && conn.connection->getBottomRegion()) {
```
osamahammad21: How?

src/odb/src/3dblox/checker.cpp
```diff
@@ -233,6 +249,49 @@ void Checker::checkInternalExtUsage(dbMarkerCategory* top_cat,
...
+                                      describe(conn.bottom_region, marker));
+        marker->setComment(msg);
+        logger_->warn(utl::ODB, 207, msg);
+        count++;
+      }
+    }
+  }
+  if (count > 0) {
+    logger_->warn(
+        utl::ODB, 273, "Found {} non-intersecting connections", count);
```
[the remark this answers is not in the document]
osamahammad21: Good suggestion

src/odb/test/data/missing_regions.3dbv
```diff
@@ -0,0 +1,23 @@
...
+      front_reg:
+        side: front
+    external:
+      APR_tech_file: [NG45_PATH/*_tech.lef]
+    
+  HierChiplet:
+    type: hier
+    regions:
+      hier_reg:
+        side: front
```
osamahammad21: This is a good example of a parser error that should be marked earlier as hierarchical chiplets should be purely hierarchical and never have regions defined.

src/odb/src/3dblox/checker.cpp
ahmed532: I added it because of the parser error. I think we should fix the error in a separate PR. But, should we keep the check as defensive programming? Or is it confusing?

osamahammad21: No I think it is confusing.

ahmed532: OK will remove it from this PR now.

src/odb/src/3dblox/checker.cpp
```diff
@@ -233,6 +249,35 @@ void Checker::checkInternalExtUsage(dbMarkerCategory* top_cat,
 void Checker::checkConnectionRegions(dbMarkerCategory* top_cat,
                                      const UnfoldedModel& model)
 {
+  auto* cat = dbMarkerCategory::createOrReplace(top_cat, "Connection regions");
```
osamahammad21: Create only if there is a marker

PR #9521, opened 2026-02-23
Pdn secondary ground

Bbn08:
As per your request i have moved the PDN to its own pr @maliberty.

Added `-secondary_ground` parameter to `set_voltage_domain` for designs with secondary ground rails (e.g. VSSA).

Also fixes `VoltageDomain::getNets()` which was placing secondary nets after ground instead of between primary power and ground. Correct order is:

primary power → secondary power(s) → secondary ground(s) → primary ground
Two regression tests added; `CMakeLists` registration pending `.defok` generation.

### DIFF
--- a/src/pdn/src/PdnGen.cc
+++ b/src/pdn/src/PdnGen.cc
@@ -29,6 +29,7 @@
 #include "straps.h"
 #include "techlayer.h"
 #include "utl/Logger.h"
+#include "utl/timer.h"
 #include "via.h"
 #include "via_repair.h"
 
@@ -56,6 +57,14 @@ void PdnGen::resetShapes()
     grid->resetShapes();
   }
   updateRenderer();
+}
+
+void PdnGen::run(bool trim, bool add_pins, const std::string& report_file)
+{
+  utl::Timer timer;
+  buildGrids(trim);
+  writeToDb(add_pins, report_file);
+  logger_->info(utl::PDN, 500, "Runtime: {:.2f}s", timer.elapsed());
 }
 
 void PdnGen::buildGrids(bool trim)
--- a/src/pdn/src/PdnGen.cc
+++ b/src/pdn/src/PdnGen.cc
@@ -254,7 +254,7 @@ VoltageDomain* PdnGen::getCoreDomain() const
 void PdnGen::ensureCoreDomain()
 {
   if (core_domain_ == nullptr) {
-    setCoreDomain(nullptr, nullptr, nullptr, {});
+    setCoreDomain(nullptr, nullptr, nullptr, {}, {});
   }
 }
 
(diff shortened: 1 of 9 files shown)


src/pdn/src/domain.cpp
```diff
@@ -210,12 +224,20 @@ void VoltageDomain::report() const
...
+
+  if (!secondary_ground_.empty()) {
     std::string nets;
-    for (auto* net : secondary_) {
+    for (auto* net : secondary_ground_) {
       nets += net->getName() + " ";
     }
-    logger_->report("  Secondary nets: {}", nets);
+    logger_->report("  Secondary ground nets: {}", nets);
   }
```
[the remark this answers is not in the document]
Bbn08: Refactored into a `reportSecondaryNets() `free function in an anonymous namespace.

src/pdn/src/pdn.tcl
```diff
@@ -115,14 +116,29 @@ proc set_voltage_domain { args } {
...
+  if { [info exists keys(-secondary_ground)] } {
+    foreach snet $keys(-secondary_ground) {
+      set db_net [[ord::get_db_block] findNet $snet]
+      if { $db_net == "NULL" } {
+        utl::error PDN 1007 "Unable to find secondary ground net: $snet"
+      } else {
+        lappend secondary_ground $db_net
       }
     }
   }
```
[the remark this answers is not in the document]
Bbn08: Extracted into a `pdn::collect_secondary_nets `helper proc shared by both secondary power and ground collection

src/pdn/src/domain.cpp
```diff
@@ -18,19 +18,37 @@
...
+  if (nets.empty()) {
+    return;
+  }
+  std::string names;
+  for (auto* net : nets) {
+    names += net->getName() + " ";
+  }
+  logger->report("  Secondary {} nets: {}", label, names);
+}
+}  // namespace
```
gadfort, line 35: I think you can just make a loop in the reporting function to avoid needing to add this function.

src/pdn/src/pdn.tcl
```diff
@@ -1,6 +1,21 @@
...
 # Copyright (c) 2022-2025, The OpenROAD Authors
 
+# Resolve a list of net name strings to dbNet objects.
+# Reports PDN error $err_id if any name is not found.
+proc pdn::collect_secondary_nets { net_names err_id type_label } {
+  set nets {}
+  foreach snet $net_names {
+    set db_net [[ord::get_db_block] findNet $snet]
+    if { $db_net == "NULL" } {
+      utl::error PDN $err_id "Unable to find secondary $type_label net: $snet"
```
gadfort, line 11: I would avoid the use of a variable for the err_id since that will not be detectable by the id parser.

src/pdn/src/PdnGen.cc
```diff
@@ -58,6 +59,14 @@ void PdnGen::resetShapes()
   updateRenderer();
 }
 
+void PdnGen::run(bool trim, bool add_pins, const std::string& report_file)
+{
+  utl::Timer timer;
+  buildGrids(trim);
+  writeToDb(add_pins, report_file);
+  logger_->info(utl::PDN, 500, "Runtime: {:.2f}s", timer.elapsed());
```
gadfort, line 67: I like this function, but I think we need to resolve the formatting question before adding the timer code here.

src/pdn/src/domain.cpp
```diff
@@ -75,21 +95,31 @@ std::vector<odb::dbNet*> VoltageDomain::getNets(bool start_with_power) const
...
+    nets.insert(
+        nets.end(), secondary_ground_.begin(), secondary_ground_.end());
+    nets.insert(nets.end(), secondary_power_.begin(), secondary_power_.end());
     if (switched_power_ != nullptr) {
       nets.push_back(switched_power_);
     }
+    nets.push_back(power_);
   }
 
-  nets.insert(nets.end(), secondary_.begin(), secondary_.end());
```
gadfort, line 91: I think it will be okay, but I'm a little worried about reordering of the nets for folks who are currently expecting the current order.

gadfort (changes requested):

PR #9522, opened 2026-02-23
Support Kali GNU/Linux and generic Debian derivatives in DependencyInstaller

### LINKED ISSUE #3910: Add support for Kali GNU/Linux (and generic Debian derivatives)
## Summary

Building OpenROAD-flow-scripts on Kali GNU/Linux fails at multiple stages due to OS detection not recognizing Kali (or generic Debian derivatives). Since Kali is Debian-based, it should be routed through the existing Debian/Ubuntu installation paths.

## Issues Encountered

1. **OS detection failure** — Both `etc/DependencyInstaller.sh` and `tools/OpenROAD/etc/DependencyInstaller.sh` only match `"Ubuntu"`, `"Debian GNU/Linux rodete"`, and `"Debian GNU/Linux"` exactly. Kali reports its OS as `"Kali GNU/Linux"`, causing an "unsupported system" error.

2. **`libtcl` virtual package** — On Kali (and some Debian versions), `libtcl` is a virtual package with no install candidate. The script needs to fall back to `libtcl8.6`.

3. **Hardcoded `libpython3.8`** — The Debian install path defaults to `libpython3.8`, which doesn't exist on newer Debian-based distros (Kali has Python 3.13+). The version should be auto-detected.

4. **or-tools download URL** — Kali's `VERSION_ID` is `2025.4`, which doesn't correspond to any or-tools release. Non-standard Debian version IDs need to be normalized to a compatible Debian release (e.g., `12`).

5. **Abseil compiled without `-fPIC`** — On GCC 15 (Kali/Debian Testing), PIE is the default. Abseil static libraries need `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` to link into PIE executables.

## Environment

- **OS**: Kali GNU/Linux 2025.4 (Rolling, based on Debian Testing)
- **GCC**: 15.2.0
- **glibc**: 2.42
- **Python**: 3.13

## Expected Behavior

`setup.sh` and `build_openroad.sh` should complete successfully on Kali and other Debian derivatives.

Subhankar-hub:
## Summary

- Add `"Kali GNU/Linux"` and wildcard `"Debian GNU/Linux"*` to OS case match so Debian derivatives are recognized
- Fall back to `libtcl8.6` when the virtual `libtcl` package has no install candidate
- Auto-detect system Python version instead of hardcoding `3.8` for Debian
- Normalize non-standard Debian version IDs (e.g., Kali's `2025.4`) to Debian `12` for the or-tools download URL
- Add `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` to Abseil cmake config for GCC 15+ where PIE is the default

## Test plan

- [x] Tested `setup.sh` on Kali GNU/Linux 2025.4 (GCC 15.2, Python 3.13, glibc 2.42)
- [x] Verified OpenROAD builds and runs (`openroad -version` succeeds)
- [ ] Verify no regressions on Ubuntu 22.04/24.04
- [ ] Verify no regressions on Debian 12

Fixes The-OpenROAD-Project/OpenROAD-flow-scripts#3910

Made with [Cursor](https://cursor.com)

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -655,7 +655,7 @@ _install_abseil() {
             _execute "Extracting Abseil..." tar xf "abseil-cpp-${ABSL_VERSION}.tar.gz"
             cd "abseil-cpp-${ABSL_VERSION}"
             local cmake_bin=${PREFIX:-/usr/local}/bin/cmake
-            _execute "Configuring Abseil..." "${cmake_bin}" -DCMAKE_INSTALL_PREFIX="${absl_prefix_install}" -DCMAKE_CXX_STANDARD=17 -B build .
+            _execute "Configuring Abseil..." "${cmake_bin}" -DCMAKE_INSTALL_PREFIX="${absl_prefix_install}" -DCMAKE_CXX_STANDARD=17 -DCMAKE_POSITION_INDEPENDENT_CODE=ON -B build .
             _execute "Building and installing Abseil..." "${cmake_bin}" --build build --target install
         )
         absl_prefix_found="${absl_prefix_install}"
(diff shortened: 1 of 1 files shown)


maliberty:
Note that commits need to be signed (`git commit -s`) to pass DCO.  If you click the link it has info on correcting existing commits.

PR #9528, opened 2026-02-23
Fix some uses of `using namespace`

hzeller:


povik:
What's the motivation? I understood the motivation to get rid of `using` from headers but I'm not sure why `using namespace` would be undesirable in these .cc files. My experience is it's an additional burden to keep those `using` sections updated during rsz development. It uses a lot of sta types.

povik:
I see #9525, let's discuss there

maliberty:
This exacerbates the situation @povik is trying to avoid (discussed in #9525).  I think we should jump to FQN directly.

hzeller:
I am doing this manually, so I first need to break out the using clauses before I can fix the inline implementation. Maybe there is a tool for that, but first steps first.

I can keep working on this until we have the inline situation, but I need to break this into several commits to stay sane :)

hzeller:
Alright, done.
Replacing the names with the FQN is something I would've avoided in the first step so that it is easy to do step by step. But, anyway, it's done now.

@maliberty @povik

src/rsz/src/RepairSetup.cc
```diff
@@ -51,7 +51,22 @@
...
+using sta::InstancePinIterator;
+using sta::Path;
+using sta::PathExpanded;
+using sta::Scene;
+using sta::Slack;
+using sta::TimingArc;
+using sta::Vertex;
+using sta::VertexInEdgeIterator;
+using sta::VertexOutEdgeIterator;
+using sta::VertexSet;
```
maliberty: Could you do these too?

maliberty:
Thanks for doing this manual work.  I hope it will be acceptable to all.

src/rsz/src/RepairSetup.cc
hzeller: Done.

hzeller:
The CI error seems to be some docker glitch

PR #9535, opened 2026-02-24
Use bison and flex starlark locally, not from bazel_rules_hdl.

hzeller:
We already use `rules_flex` and `rules_bison` from bcr, but were using `WORKSPACE` to pull in our starlark files. Instead, use them locally to reduce the parts needed from WORKSPACE.


hzeller:
Relevant STA to merge so that we can have it in a clean submodule https://github.com/The-OpenROAD-Project/OpenSTA/pull/296

Issue #9392, opened 2026-01-29
ram: feature tracker


rovinski:
### Description

This is a public tracker for feature development for the ram tool.

- [ ] Add proper support for RAMs with 1 or more r, w, and rw ports
  - The current model only effectively supports 1rw RAMs. Multi-decoder support needs to be added to support anything with more than 1 addr port.
- [ ] Add placement sanity checks to the regression tests (`check_placement`, etc.)
- [ ] Add support for endcap insertion
  - Endcaps need to be inserted around the perimeter of the RAM bounding box after fill
- [ ] Physical verification
  - [ ] Check that timing, LVS, and DRC are clean via external tools
- [ ] Sweep word and bit sizes 1-64 (all values) to verify support for arbitrary array sizes (no placement or routing errors, LVS clean, DRC clean, passes STA)
- [ ] Add automatic detection of standard cell power/ground names
  - Power/ground pin names are currently manually specified by the user. The generator should be able to iterate through each standard cell, detect the power/ground names, and then globally connect them. The generator should also error on any cell that has more than one power/ground.
- [ ] Enable optional user-provided name of power rails
  - The current implementation hardcodes power to `VDD` and ground to `VSS`. Some users may want to change this. The default values should be `VDD` and `VSS` but users should be able to override.
- [ ] Add support for AND3/AND4 cells in the decoder
  - Only AND2 is supported now. Larger arrays can be more efficient with wider ANDs
- [ ] Add support for automatic master selection
  - Find and identify FFs, ICGs, buffers, inverters, tristates, AND2/3/4s automatically if not user specified
- [ ] Automatic detection of the followpin metal layer and width
  - The current implementation requires manual specification of the followpin layer and width. This should be automatically extractable from the library. The user should not need to specify this.
- [x] Add support for arbitrary bit mask granularity
  - Currently there is no mask support. Masking support should be added on an arbitrary basis (1 bit, 2 bits, 3 bits, etc.) as long as the word size is divisible by the mask size (cannot make a 2-bit mask for a 7-bit word)
- [ ] Add cell anchoring
  - Allow instances within cells to be “anchored” at a certain point, such as for input buffers being aligned with the pins which they drive
- [ ] Enable users to specify the input buffer to use, or to even disable completely
  - The current code always inserts a default size buffer. Allow users to manually specify which buffer to use
- [ ] Enable users to specify an output buffer to use, or to disable completely
[...]

PR #9539, opened 2026-02-24
Add behavioral verilog output

tnguy19:
- add argument `-write_behavioral_verilog [filename]` to `generate_ram` allowing the user to generate a behavioral verilog file for RAM 
- update `storage_cell` in `MakeCellInst` (line 122) to allow use of CLK and GATE pin for flip flop and latch cells

@rovinski

### DIFF
--- a/src/ram/src/ram.cpp
+++ b/src/ram/src/ram.cpp
@@ -597,6 +597,14 @@ void RamGen::generate(const int bytes_per_word,
   const std::string ram_name
       = fmt::format("RAM{}x{}", word_count, bits_per_word);
 
+  // error checking for read ports != 1 for current version of RamGen, edit
+  // later for future changes
+  if (read_ports != 1) {
+    logger_->error(
+        RAM, 25, "The ram generator currently only supports 1 read port.");
+    return;
+  }
+
   logger_->info(RAM, 3, "Generating {}", ram_name);
 
   storage_cell_ = storage_cell;
--- a/src/ram/test/make_8x8_behavioral.vok
+++ b/src/ram/test/make_8x8_behavioral.vok
@@ -1,18 +1,14 @@
 module RAM8x8 (
   clk,
   D,
-  Q0,
-  Q1,
+  Q,
   addr_rw,
-  addr_r1,
   we
 );
   input clk;
   input [7:0] D;
-  output reg [7:0] Q0;
-  output reg [7:0] Q1;
+  output reg [7:0] Q;
   input [2:0] addr_rw;
-  input [2:0] addr_r1;
   input [0:0] we;
 
   // memory array declaration
@@ -30,11 +26,7 @@ module RAM8x8 (
 
   // read logic
   always @(*) begin
-    Q0 = mem[addr_rw];
-  end
-
-  always @(*) begin
-    Q1 = mem[addr_r1];
+    Q = mem[addr_rw];
   end
 
 endmodule
--- a/src/ram/test/make_8x8.defok
+++ b/src/ram/test/make_8x8.defok
@@ -3,360 +3,270 @@ DIVIDERCHAR "/" ;
 BUSBITCHARS "[]" ;
 DESIGN RAM8x8 ;
 UNITS DISTANCE MICRONS 100 ;
-DIEAREA ( 0 0 ) ( 14260 2448 ) ;
-ROW RAM_ROW0 unithd 0 0 N DO 310 BY 1 STEP 46 0 ;
-ROW RAM_ROW1 unithd 0 272 FS DO 310 BY 1 STEP 46 0 ;
-ROW RAM_ROW2 unithd 0 544 N DO 310 BY 1 STEP 46 0 ;
-ROW RAM_ROW3 unithd 0 816 FS DO 310 BY 1 STEP 46 0 ;
-ROW RAM_ROW4 unithd 0 1088 N DO 310 BY 1 STEP 46 0 ;
-ROW RAM_ROW5 unithd 0 1360 FS DO 310 BY 1 STEP 46 0 ;
-ROW RAM_ROW6 unithd 0 1632 N DO 310 BY 1 STEP 46 0 ;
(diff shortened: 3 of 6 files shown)


src/ram/src/ram.cpp
```diff
@@ -832,6 +833,97 @@ void RamGen::generate(const int bytes_per_word,
...
+void RamGen::writeBehavioralVerilog(const std::string& filename,
+                                    const int bytes_per_word,
+                                    const int word_count,
+                                    const int read_ports){
+    if (filename.empty()) {
+      return;
+    }
+
+    const int word_size_bit = bytes_per_word * 8; 
+    const int address_width = std::ceil(std::log2(word_count));
```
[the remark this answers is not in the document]
tnguy19: updated in af4dd47

src/ram/src/ram.cpp
```diff
@@ -832,6 +836,105 @@ void RamGen::generate(const int bytes_per_word,
...
+                                         read_port_logic);
+
+  std::ofstream vf(filename);
+  if (!vf.is_open()) {
+    logger_->error(RAM, 23, "Unable to open file {}", filename);
+    return;
+  }
+
+  vf << verilog_code;
+  vf.close();
```
rovinski, line 964: @maliberty isn't there a utility function for writing files? Should that be used here?

src/ram/test/make_8x8_behavioral.tcl
rovinski, line 1: Does this need to be a separate test? Can't you just add the flag to the make8x8 test?

rovinski:
test still requires fixing for CI

src/ram/test/make_8x8_behavioral.vok
rovinski, line 1: A few notes:

* No indentation for `module`
* Please name the module the exact same as the block. This will help developers seamlessly swap between the gate-level and behavioral versions.
* Indent ports (D, Q0, etc.)
* Please use 2-space indenting
* Power and ground ports should not be included
* Consistent indentation and spacing
** e.g. `for (i=0; i< 1; i = i +1 )` --> `for (i = 0; i < 1; i = i + 1)`
** e.g. no empty line after `// read logic`

rovinski:
CI is still failing though, please make sure everything is up to date.

src/ram/src/ram.cpp
```diff
@@ -832,6 +836,109 @@ void RamGen::generate(const int bytes_per_word,
...
+                                         bytes_per_word - 1,
+                                         word_size_bit - 1,
+                                         word_count - 1,
+                                         bytes_per_word,
+                                         read_port_logic);
+
+  std::ofstream vf(filename);
+  if (!vf.is_open()) {
+    logger_->error(RAM, 23, "Unable to open file {}", filename);
+    return;
```
maliberty: rm - return is unreachable as error will throw

src/ram/test/make_8x8_behavioral.vok
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  // read logic
+  always @(*) begin
+    Q0 = mem[addr];
+  end
+
+  always @(*) begin
+    Q1 = mem[addr];
+  end
```
maliberty, line 30: Is this correct?  You have two identical read ports using a single address?

tnguy19: Yes, because the structural verilog code for the ram that the behavioral code generator is based on uses the same address bus shared with all the read ports. So all read ports currently read from the same address

maliberty: So a correct description of the hardware but not normal meaning to real ports afaik.  @rovinski is this what you intend?

rovinski: @tnguy yeah, that's not right. If that's what the structural netlist does, then it's wrong.

There should be an address port per read port. Fix up the behavioral model for now and we'll have to modify the netlist to be correct later.

tnguy19: @rovinski  I have added the update in 190a188 please let me know if theres any changes needed in the new behavioral code

rovinski: Hmmm, I think this raises an issue on the lack of clarity for the tool interface. It is true that there is one write port and two read ports. But it is unclear whether this means that it should be 3 ports total (2 read 1 write) or 2 ports (1 read-write 1 read).

Right now, we only implicitly support 1 write port. I think it should be that there is always 1 read-write port and then additional read ports. In other words, the first addr port is used for both reading and writing, and then any additional ports are only for reading.

If you make this change, then we can mark it done for now and leave an issue to come back to it when more write ports are supported.

tnguy19: I have added the updated code in 80c67ad, ready for review

maliberty: @rovinski you are wanting to merge this with code the writes a behavioral model that doesn't match the hardware generated?  That seems confusing/dangerous.

rovinski: I am inclined to for now with the goal that the hardware is fixed to match. It didn't occur to me until now that essentially only 1rw memories actually work. Everything else is broken because there is no routine to implement multiple decoders.

Perhaps the ok test should be changed to be 1rw rather than having 2 broken read ports.

maliberty: Then let's put in an error if the user selects a broken configuration.  I don't want to give misleading results.

rovinski:
@tnguy19 based on conversation, can you please do the following before merge:

1) Change the ram test and ok files to be based on `-read_ports 1`.
2) Add a check for read_ports != 1, and if so, error with "The ram generator currently only supports 1 read port."

tnguy19:
@maliberty I have added the updates in a526f03 code is ready for review

rovinski:
@maliberty CI looks green now, I think this is ready to merge. I added fixing muli-read/write ports as a top priority item in #9392

Issue #9363, opened 2026-01-27
RSZ: enhance Rebuffer::fullyRebuffer to understand global routing parasitics


eder-matheus:
### Description

If Rebuffer::fullyRebuffer runs with grt parasitics, it fails with `[CRITICAL RSZ-1003] unhandled BufferedNet type` error.

This function should be able to understand the grt parasitics and perform its moves based on them.
This is specially useful for the integration between GRT and GPL during timing-driven placement.

krrish175-byte:
Hey @eder-matheus can I work on this issue?

eder-matheus:
> Hey [@eder-matheus](https://github.com/eder-matheus) can I work on this issue?

Hi @krrish175-byte! Sure, feel free to work on it and contribute with a PR!

krrish175-byte:
@eder-matheus, raised a pr, please check. Thanks!

eder-matheus:
Hey @krrish175-byte, thanks for the fix! I see that your PR https://github.com/The-OpenROAD-Project/OpenROAD/pull/9548 was already merged, so I'm closing it.

Thanks again!

PR #9548, opened 2026-02-25
Add missing via type handling in fullyRebuffer

krrish175-byte:
This PR addresses issue #9363 where `Rebuffer::fullyRebuffer` crashes with `[CRITICAL RSZ-1003] unhandled BufferedNet type` when running with GRT parasitics.

### DIFF
--- a/src/rsz/src/Rebuffer.cc
+++ b/src/rsz/src/Rebuffer.cc
@@ -888,6 +888,7 @@ BufferedNetPtr Rebuffer::recoverArea(const BufferedNetPtr& root,
         switch (node->type()) {
           case BnetType::wire:
           case BnetType::buffer:
+          case BnetType::via:
             recurse(node->ref(), arrival + node->delay());
             break;
           case BnetType::junction:


src/rsz/src/Rebuffer.cc
```diff
@@ -905,10 +907,11 @@ BufferedNetPtr Rebuffer::recoverArea(const BufferedNetPtr& root,
       [&](auto& recurse, int level, const BnetPtr& node, int upstream_wl)
           -> BufferedNetSeq {
         switch (node->type()) {
+          case BnetType::via:
```
povik: Today via should never appear here because recoverArea() gets called on a tree from a prior round of recoverArea(), or one that was returned from bufferForTiming(). Neither should contain vias.

I think if vias were there, we would want to handle them transparently, just skip over them, which the code change doesn't do. It's easiest if we keep them unhandled for now.

src/rsz/src/Rebuffer.cc
```diff
@@ -1099,6 +1101,17 @@ void Rebuffer::annotateTiming(const BnetPtr& tree)
...
+            int ret = recurse(bnet->ref());
+            BnetPtr p = bnet->ref();
+            double via_res
+                = bnet->viaResistance(corner_, resizer_, estimate_parasitics_);
+            FixedDelay via_delay = FixedDelay(via_res * p->cap(), resizer_);
+            bnet->setDelay(via_delay);
+            bnet->setSlack(p->slack() - via_delay);
+            bnet->setSlackTransition(p->slackTransition());
+            return ret;
+          }
```
povik, line 1113: This looks correct.

test/regression_test.sh
```diff
@@ -23,7 +23,7 @@ case "$TEST_TYPE" in
 esac
 
 echo "Command: $CMD"
-$CMD |& tee $LOG_FILE
+$CMD 2>&1 | tee $LOG_FILE
```
povik, line 26: I'm not sure what this does but gemini says it's a compatibility improvement

povik (approved):
Looks good as far as Rebuffer.cc; I defer to @maliberty on regression_test.sh

test/regression_test.sh
maliberty: They are equivalent and |& is more modern.  Perhaps it matters on some ancient version of bash but unless you actually have a problem I would revert it as nobody has ever complained.

krrish175-byte: |& works fine on CI and modern Bash. The change was made because Bash 3.2 (macOS default) does not support |&, which caused all tests to fail locally on macOS. Happy to revert if local macOS testing isn't a priority, just wanted to flag it since it bit me during development.

maliberty: I'm ok to leave it if helps mac

maliberty:
I see ` /OpenROAD/src/rsz/src/Rebuffer.cc:888:9: error: enumeration value ‘via’ not handled in switch [-Werror=switch]` needs fixing

Issue #9222, opened 2026-01-09
Bogus error when viewing macro placement `[ERROR GRT-0042] Pin clock does not have geometries in a valid routing layer.`


oharboe:
### Description

When I run `make gui_2_2_floorplan_macro`, I get a bogus error.

```
OpenROAD 24Q3-12098-gb2a2369ddb 
Features included (+) or not (-): +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
source external/bazel-orfs++orfs_repositories+docker_orfs/OpenROAD-flow-scripts/flow/platforms/asap7/liberty_suppressions.tcl
read_db -hier .../2_2_floorplan_macro.odb
[WARNING ORD-0012] Hierarchical flow (-hier) is currently in development and may cause multiple issues. Do not use in production environments.
[WARNING GRT-0300] Timing is not available, setting critical nets percentage to 0.
[deleted]
[ERROR GRT-0042] Pin clock does not have geometries in a valid routing layer.
[WARNING GUI-0066] Heat map "Estimated Congestion (RUDY)" has not been populated with data.
```

### Suggested Solution

Remove bogus errors due to later stages not being run while viewing earlier stages so as to reduce cognitive load and not to drown helpful actionable errors and warnings.

maliberty:
It looks like you have the rudy heatmap turned on.  Are you suggesting with automatically disable that?

oharboe:
> It looks like you have the rudy heatmap turned on.  Are you suggesting with automatically disable that?

The GUI config file is shared between all my openroad use-cases, I must have recently looked at the heatmap for another design.

I dont want to get into what scope config files should be saved in, but I certainly find bogus error messages when I switch between designs surprising and noisy.

PR #9553, opened 2026-02-26
grt: do not error out when pins are not placed yet during initialization for Rudy

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/9222.


maliberty:
Wasn't this fixed in #9544?

eder-matheus:
> Wasn't this fixed in #9544?

I thought that, but no. This PR extends that fix to the moment where we create the Pins for dbBTerms. That's when the error reported on https://github.com/The-OpenROAD-Project/OpenROAD/issues/9222 happens.

PR #9557, opened 2026-02-26
Fix segmentation fault when selecting instance in debug mode

### LINKED ISSUE #9017: gpl: debug select inst bug
### Description

We get a segmentation fault when trying to select an instance with debug mode. It started happening after a fix considering regions and selecting an instance.

### Suggested Solution

It is related to searching the instance in NB and NBC vectors. I also wonder if we cannot only use the NBC vector for debug mode drawings (although NB vector is required for gpl fillers).

### Additional Context

_No response_

krrish175-byte:
### Description
Fixes a segmentation fault that occurs when trying to select an instance in GPL debug mode. This regression was introduced after a previous fix that added region-aware instance selection logic.

Fixes #9017


gudeh:
Hi @krrish175-byte, thanks for your contribution! Were you able to reproduce the issue? I don't remember what/when the issue is triggered exactly. I tried but couldn't reproduce the issue with master version on my end.

krrish175-byte:
Hi @gudeh! Yes, I was able to reproduce the issue on the master branch.

The segmentation fault occurs in GPL's debug mode if we try to select an instance that is not part of any NesterovBase (for example, an unplaced filler cell outside defined regions).

How to reproduce: If you create a simple test with regions and place a dummy filler cell outside of the region, selecting that filler with -inst in debug mode will trigger the crash immediately:

```tcl
# Setup design with region, then:
set master [$lib findMaster "sky130_fd_sc_hd__fill_1"]
set dummy_inst [odb::dbInst_create $block $master "my_dummy_filler"]
$dummy_inst setPlacementStatus "PLACED"
$dummy_inst setOrigin 5000 5000

global_placement_debug -pause 1 -update 1 -inst my_dummy_filler
global_placement -density 0.7
```
In graphicsImpl.cpp, the debug initialization GraphicsImpl::debugForNesterovPlace() looks for the instance in nbVec_ to get nb_selected_index_. If it isn't found in any NesterovBase, nb_selected_index_ remains kInvalidIndex (which is std::numeric_limits<size_t>::max()).

Later, in GraphicsImpl::drawNesterov() and GraphicsImpl::reportSelected(), the code accesses nbVec_[nb_index] assuming it's valid, which leads to a segmentation fault:

```cpp
    size_t nb_index = 0;
    if (nb_selected_index_ != kInvalidIndex) {
      nb_index = nb_selected_index_;
    } else {
      logger_->warn(
          utl::GPL, 317, "Selected instance not found in any NesterovBase");
    }
    FloatPoint densityGrad = nbVec_[nb_index]->getDensityGradient(gcell); // <--- SEGFAULT HERE!
```

When nb_selected_index_ is invalid, the code logs a warning but proceeds to use nb_index = 0 by default. And when it calls nbVec_[0]->getDensityGradient(gcell) on an unmapped gcell, it crashes.

The proposed fix (which I've pushed to fix/gpl-debug-select-inst) safely bails out of rendering if the index is invalid instead of defaulting to 0. Here is the screenshot:

Screenshot 2026-03-03 at 12 49 09 PM

src/gpl/src/graphicsImpl.cpp
```diff
@@ -27,6 +27,30 @@
 
 namespace gpl {
 
+namespace {
```
gudeh: Why include this namespace? getSelectedNesterovBaseIndex() should be a private member function of GraphicsImpl. No need for all these parameters.

src/gpl/src/graphicsImpl.cpp
```diff
@@ -524,13 +547,12 @@ void GraphicsImpl::reportSelected()
...
-    } else {
-      logger_->warn(
-          utl::GPL, 318, "Selected instance not found in any NesterovBase");
+    auto opt_nb_index = getSelectedNesterovBaseIndex(
+        nb_selected_index_, nbVec_, logger_, 318);
+    if (!opt_nb_index) {
+      return;
     }
+    size_t nb_index = *opt_nb_index;
     FloatPoint densityGrad = nbVec_[nb_index]->getDensityGradient(gcell);
```
gudeh, line 575: I understand the code is only missing a proper protection from an invalid `nb_index` (I don't see it happening on my tests, but would be nice to have). I mean, when nb_selected_index_ == kInvalidIndex. Maybe we don't even need the nb_index variable.

gudeh (changes requested):
I believe the segmentation issue mentioned in https://github.com/The-OpenROAD-Project/OpenROAD/issues/9017 was solved by commit https://github.com/The-OpenROAD-Project/OpenROAD/commit/ff6eed246a23e19bec1f844f26e6e8f1fd2c8b3f. I wasn't able to reproduce the segmentation fault in any way, even with the dummy insertion mentioned by @krrish175-byte.

The code is only missing some polishing after being solved with the mentioned commit: 

We find the selected instance in two different places in two different ways, would be nice to use the same methodology, maybe a private member function. 

We can also make sure we don't use the nb_index when it's invalid.

krrish175-byte:
@gudeh, PTAL

krrish175-byte:
@gudeh , any updates regarding this?

gudeh:
LGTM. I didn't test this yet though

src/gpl/src/graphicsImpl.cpp
```diff
@@ -103,6 +118,7 @@ void GraphicsImpl::debugForNesterovPlace(
     for (const auto& nb : nbVec_) {
       for (size_t idx = 0; idx < nb->getGCells().size(); ++idx) {
         GCellHandle cell_handle = nb->getGCells()[idx];
+        if (cell_handle->contains(inst)) {
```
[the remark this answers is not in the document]
gudeh: I renamed this to debg_inst I believe

gudeh:
I misspelled on my previous message in this discussion: **debug_inst**

You also missed your last commit not including your signoff, that's the reason for the DCO error.

PR #9569, opened 2026-02-28
ci: Adds MacOS Bazel Builder

QuantamHD:


.github/workflows/github-actions-macos-bazel.yml
```diff
@@ -0,0 +1,50 @@
...
+  push:
+    branches:
+      - master # Updated to strictly track master
+      - macos-bazel # for local testing
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  Mac-Build:
+    if: github.repository_owner != 'The-OpenROAD-Project' && github.repository_owner != 'The-OpenROAD-Project-private' && github.repository_owner != 'The-OpenROAD-Project-staging'
```
maliberty: What is the goal of excluding these?

QuantamHD: This mirrors the other macos action

image

maliberty:
Introduced in https://github.com/The-OpenROAD-Project/OpenROAD/commit/e3ee598ffad7e152b471534ecf046032122aa342 was for back when we had a self-hosted mac build.  We may again someday but this currently means the action never runs.

QuantamHD:
Fixed

QuantamHD:
As you point out we should probably populate a bazel macos cache for the CI. Right now we're limited by the 10GB limit action cache.

MODULE.bazel
```diff
@@ -92,8 +92,8 @@ git_override(
 bazel_dep(name = "qt-bazel")
 git_override(
     module_name = "qt-bazel",
-    commit = "2b450c5d0b84e0a48e0e7797486e62d24f65c78b",
-    remote = "https://github.com/The-OpenROAD-Project/qt_bazel_prebuilts",
+    commit = "c42fdd2cb15d284f8ec78f13485a1181d722ea85",
+    remote = "https://github.com/feuerste/qt_bazel_prebuilts",
```
maliberty: Why change to this fork?

QuantamHD: That was an accident. I'll revert that. Thanks for the catch

QuantamHD:
@maliberty the ci failed pulling a docker container should be good to go

.github/workflows/github-actions-macos-bazel.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Build OpenROAD
+        run: |
+          bazelisk build \
+            --disk_cache=~/.cache/bazel-disk-cache \
+            --experimental_disk_cache_gc_max_size=5G \
+            --//:platform=gui //:openroad
+
+      # 2. SAVE: Only executes if this is a push/merge directly to the master branch
+      - name: Save Bazel Disk Cache
+        if: github.ref == 'refs/heads/master' && github.event_name == 'push'
```
maliberty, line 44: Do we need to check if this is being run in the The-OpenROAD-Project org and not in a fork?

QuantamHD: I don't think so the upstream cache can only be run by the upstream action 

https://github.com/orgs/community/discussions/25242

maliberty (approved):

maliberty:
As a point of policy this will not be merge blocking and mac remains best effort.  It is nearly impossible to debug build problems without an actual mac.

PR #9574, opened 2026-02-28
Regression testing / Reproducability: Fix MacOS and Linux differences for rand() and qsort()

fredowski:
As already discussed in #9480 the regression yields different results on MacOS and on Linux. Today around 60 tests fail on MacOS when running a part of the regression in bazel with

`bazelisk test src/...`

One major reason are different implementations of rand() and qsort() between glibc and the apple functions. qsort() gives different results for equal elements. This PR addresses this problem by

- Change the rand function in gpl to mt19937 which yields the same pseudorandom sequences on apple and linux
- Update the new "golden" reference files due to this change
- Provide a copy of rand() and qsort() for apple which is copied from glibc and link this to the abc code. This avoids to change abc. This update does not need an update of the "golden" reference files as they are produces on linux anyway.

I tested this on MacOS and debian 12 x86_64. On MacOS the failed tests are reduced to 5. On linux x86_64 there are no regression failures.

I think these changes have no impact on actual function but only on regression and reproducability. You can see the small differences in the updated golden files.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -175,7 +175,6 @@ cc_binary(
         ":opt_notification",
         ":ord",
         "//bazel:runfiles",
-        "//bazel/glibc_for_apple:portable_rand_qsort",
         "//src/cut",
         "//src/gui",
         "//src/sta:opensta_lib",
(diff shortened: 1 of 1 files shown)


QuantamHD:
Instead of pulling LGPL code, could we just stub these function with the c++ version std::mt1997 etc. or std::stable_sort

fredowski:
Yes this exploded in clang-tidy, google code hints and friends. I will remove the last part with rand() and qsort() and just do a new pr.

PR #9579, opened 2026-03-01
bazel-orfs: bump

oharboe:


oharboe:
@hzeller @maliberty This bumps bazel-orfs and removes the dependency on local Python having PyYAML installed.

hzeller (approved):

PR #9580, opened 2026-03-01
Fix: replace_arith_modules ignores -slack_threshold due to key mismatch

suhr25:
### **SUMMARY**

This PR fixes a bug in `replace_arith_modules` where the `-slack_threshold` option was being ignored due to a mismatched key name. The issue is limited to `src/rsz/src/Resizer.tcl` and only affects Tcl argument parsing, with no impact on the C++ layer.

---

### **FIX / CHANGES**

* **Root cause:**
  The option was renamed from `-slack_margin` to `-slack_threshold`, but the later checks still used the old name. As a result, the user-provided value was never picked up and always defaulted to `0.0`.

* **Approach:**
  Updated the key name in both the guard and parsing call so it matches what `parse_key_args` produces.

#### Fix

```tcl
- if { [info exists keys(-slack_margin)] } {
-   set slack_margin [rsz::parse_time_margin_arg "-slack_margin" keys]
+ if { [info exists keys(-slack_threshold)] } {
+   set slack_margin [rsz::parse_time_margin_arg "-slack_threshold" keys]
```

* No changes to core logic—this just fixes the Tcl → C++ argument wiring.

---

### **VERIFICATION**

1. Run a normal flow up to timing estimation.
2. Call:

```tcl
replace_arith_modules -path_count 100 -slack_threshold 0.5 -target setup
```

3. **After the fix:**

* The log now shows `-slack_margin 0.5` instead of `0.0`
* The provided threshold is actually used during replacement


suhr25:
Hi @maliberty,
Please take a look at this PR when you get time.
Thanks!

maliberty:
Apparently these need updating
20:32:20  	456 - rsz.replace_arith_modules1.tcl (Failed)           IntegrationTest tcl rsz log_compare
20:32:20  	457 - rsz.replace_arith_modules2.tcl (Failed)           IntegrationTest tcl rsz log_compare
20:32:20  	458 - rsz.replace_arith_modules3.tcl (Failed)           IntegrationTest tcl rsz log_compare

suhr25:
Hi @maliberty,
I have made changes according to your comment.
Please take a look
Thanks!

PR #9585, opened 2026-03-01
odb: impl 3dblox checker checkLogicalConnectivity

ahmed532:
Now reading missing nets from verilog in 3dblox in `void ThreeDBlox::createChiplet(const ChipletDef& chiplet)`


src/odb/include/odb/3dblox.h
```diff
@@ -66,5 +66,6 @@ class ThreeDBlox
   std::unordered_set<odb::dbTech*> written_techs_;
   std::unordered_set<odb::dbLib*> written_libs_;
   std::unordered_set<std::string> read_files_;
+  std::unordered_set<std::string> read_verilog_files_;
```
osamahammad21: I don't think you should keep track of the read verilog files. It doesn't make any sense that 2 3dbx files reference the same verilog file, unlike lef and lib which can be referenced/shared by different chiplets.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -337,6 +341,67 @@ void ThreeDBlox::createChiplet(const ChipletDef& chiplet)
       odb::dbStringProperty::create(
           chip, "verilog_file", chiplet.external.verilog_file.c_str());
     }
+    if (sta_ != nullptr) {
+      std::error_code ec;
+      auto path = std::filesystem::weakly_canonical(
```
osamahammad21: why?

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -337,6 +341,67 @@ void ThreeDBlox::createChiplet(const ChipletDef& chiplet)
...
+      }
+
+      if (!cell) {
+        logger_->warn(utl::ODB,
+                      555,
+                      "Verilog module {} not found in loaded libraries.",
+                      chiplet.name);
+      }
+
+      if (cell) {
```
osamahammad21:
```suggestion
      else {
```

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -337,6 +341,67 @@ void ThreeDBlox::createChiplet(const ChipletDef& chiplet)
...
+        std::unordered_set<std::string> existing_nets;
+        for (auto* net : chip->getChipNets()) {
+          existing_nets.insert(net->getName());
+        }
```
osamahammad21: How would it happen that any dbChipNet exist before reading the verilog?

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -337,6 +341,67 @@ void ThreeDBlox::createChiplet(const ChipletDef& chiplet)
...
+        std::unique_ptr<sta::CellPortBitIterator> port_iter(
+            network->portBitIterator(cell));
```
osamahammad21: You should use netIterator not portBitIterator. dbChipNet should essentially reference the nets while the ports should reference the bumps. look at OpenRoad::linkDesign for reference

src/odb/src/3dblox/checker.cpp
```diff
@@ -288,4 +316,158 @@ void Checker::checkNetConnectivity(dbMarkerCategory* top_cat,
...
+
+    if (loaded_files.insert(file).second) {
+      debugPrint(logger_,
+                 utl::ODB,
+                 "3dblox",
+                 552,
+                 "Reading Verilog file {} for design {}",
+                 path.filename().string(),
+                 master->getName());
+      if (!sta->readVerilog(file.c_str())) {
```
osamahammad21: why are you reading the verilog at the checker?
The checker should only check for correctness while the parser at 3dblox should read the verilog and populate the data.

osamahammad21 (changes requested):

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -78,6 +82,111 @@ void ThreeDBlox::readDbv(const std::string& dbv_file)
...
+    }
+
+    for (auto* region : master->getChipRegions()) {
+      for (auto* bump : region->getChipBumps()) {
+        if (auto* bterm = bump->getBTerm()) {
+          master_bump_map[master][bterm->getName()] = bump;
+        }
+      }
+    }
+  }
```
osamahammad21, line 134: This is done because there is no mapping from the bterm to the owner dbChipBump object. A future PR to add this missing mapping will be implemented and that shall be removed then.

src/odb/src/3dblox/3dblox.cpp
```diff
@@ -78,6 +82,111 @@ void ThreeDBlox::readDbv(const std::string& dbv_file)
...
+  sta::ConcreteNetwork temp_network;
+  temp_network.copyState(sta_);
+  sta::VerilogReader verilog_reader(&temp_network);
+
+  if (!verilog_reader.read(verilog_file.c_str())) {
+    return;
+  }
+
+  sta::Instance* top_inst
+      = verilog_reader.linkNetwork(data.design.name.c_str(), true, false);
```
osamahammad21, line 107: @maliberty Does this look fine for reading the top verilog? The idea is that we want to extract connections/nets that connect ports across chiplets in the top design without contaminating the state of sta

maliberty: It looks safe afaik as you are using a temporary network.

PR #9591, opened 2026-03-02
bazel: use stamp=1 to avoid global --stamp cache pollution

oharboe:
The global --stamp flag changes the build configuration for every target, preventing cache sharing between stamped and unstamped builds. Switch the OpenRoadVersion genrule to stamp = 1 (unconditional) so it always embeds the real git version without needing --stamp on the command line.  This lets all .o files share a single cache regardless of whether version stamping is desired.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,9 +1,9 @@
 common --enable_bzlmod
 build --workspace_status_command=tools/workspace_status.sh
 
-# Release builds embed real git version info; dev builds use cache-safe placeholders.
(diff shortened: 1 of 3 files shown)


oharboe:
didn't work, it generates stamp without --config=release:

```
$ bazelisk run //:openroad
[deleted]
INFO: Running command line: bazel-bin/openroad
OpenROAD 26Q1-1497-ga53fbc42e2 
```

PR #9592, opened 2026-03-02
etc: add whittle.py a deltaDebug.py without Python in OpenROAD dependency

oharboe:
An alternative implementation of deltaDebug.py with a
standalone Python script (whittle.py) that delegates ODB manipulation
to small TCL scripts via `openroad -exit`. This works with any OpenROAD
build, even without Python compiled in.

deltaDebug.py is kept as-is during the transition period, but the
plan is to delete it once we're happy with whittle.py

Run with stock python3 — no extra packages needed beyond the standard
library.

Quick test using bazel-orfs gcd targets:

    bazelisk run //test/orfs/gcd:gcd_place_deps -- do-place ISSUE_TAG=test global_place_issue print-PWD
    mkdir whittle && cd whittle
    tar --strip-components=1 -xzf ../tmp/test/orfs/gcd/gcd_place_deps/_main/test/orfs/gcd/global_place_test.tar.gz
    sed -i 's/openroad -no_init/openroad -exit -no_init/g' \
            run-me-gcd-asap7-base.sh
    export PATH=$(pwd)/../../install/OpenROAD/bin:$PATH
    python3 ../etc/whittle.py \
            --persistence 2 --use_stdout \
            --error_string "      100 | " \
            --timeout 120 \
            --base_db_path \
                ../tmp/test/orfs/gcd/gcd_place_deps/_main/test/orfs/gcd/results/asap7/gcd/base/2_floorplan.odb \
            --step ./run-me-gcd-asap7-base.sh

After a minute or two:

```
Step 19, Nets level debugging, Insts 0, Nets 1, cut elements 1, timeout 2 minutes
[C]
Error Code found:       100 |
Removed 212 unused masters.
___________________________________
Resultant file is ../tmp/test/orfs/gcd/gcd_place_deps/_main/test/orfs/gcd/results/asap7/gcd/base/deltaDebug_base_result_2_floorpl>
Delta Debugging Done!
```


oharboe:
@maliberty FYI

oharboe:
@maliberty please retrigger pr-merge so I can see if the naming problem is gone on master

etc/whittle_cut.tcl
```diff
@@ -0,0 +1,59 @@
...
+#   WHITTLE_DEF_OUTPUT - (optional) path to write DEF output
+#
+# Prints to stdout:
+#   INSTS <count>
+#   NETS <count>
+
+set db [odb::dbDatabase_create]
+odb::read_db $db $env(WHITTLE_DB_INPUT)
+
+set block [[$db getChip] getBlock]
```
maliberty: Why not read_db and ord::get_db_block ?

etc/whittle_cut.tcl
```diff
@@ -0,0 +1,59 @@
...
+set end $env(WHITTLE_CUT_END)
+
+if { $cut_level eq "insts" } {
+  set elms [$block getInsts]
+} else {
+  set elms [$block getNets]
+}
+
+for { set i $start } { $i < $end } { incr i } {
+  set elm [lindex $elms $i]
```
maliberty: Is this correct?  You are deleting elements from the same set you are indexing into.

etc/whittle_cut.tcl
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+odb::write_db $db $env(WHITTLE_DB_OUTPUT)
+
+if { [info exists env(WHITTLE_DEF_OUTPUT)] } {
+  odb::write_def $block $env(WHITTLE_DEF_OUTPUT)
+}
+
+set block [[$db getChip] getBlock]
```
maliberty: redundant

etc/whittle_cleanup.tcl
```diff
@@ -0,0 +1,24 @@
...
+#   WHITTLE_DB_OUTPUT  - path to write the cleaned ODB file
+#   WHITTLE_DEF_OUTPUT - (optional) path to write DEF output
+#
+# Prints to stdout:
+#   REMOVED <count>
+
+set db [odb::dbDatabase_create]
+odb::read_db $db $env(WHITTLE_DB_INPUT)
+
+set removed [$db removeUnusedMasters]
```
maliberty: Similar to the comment in cut.tcl (with ord::get_db)

oharboe:
> I'll try running it later. Have you used it on a real design? It will be slightly slower due to reload the db in each cut step.

No, I haven't been able to use deltaDebug.py in ages due to the dependency inversion problem and having switched to bazel.

oharboe:
@maliberty addressed concerns and added one more integration test.

maliberty:
I mean have you tested this version on a real design.  Presumably that is now possible.

oharboe:
> I mean have you tested this version on a real design. Presumably that is now possible.

No.

oharboe:
> I'll try running it later. Have you used it on a real design? It will be slightly slower due to reload the db in each cut step.

I think that will be drowned out by running times of the script for any real use-case.

oharboe:
> I'll try running it later. Have you used it on a real design? It will be slightly slower due to reload the db in each cut step.

Thank you! You must have quite a lot of suitable use-cases to reduce reproduction time :grin:

oharboe:
@maliberty I'll merge w/master when you have tested this. OK?

maliberty (approved):

PR #9593, opened 2026-03-02
dpl: ignore unplaced bterms

gadfort:
Changes:
- Ignore unplaced bterms.

Notes:
- I would expect this to be a no-op for most designs.

### DIFF
--- a/src/dpl/src/dbToOpendp.cpp
+++ b/src/dpl/src/dbToOpendp.cpp
@@ -143,6 +143,7 @@ void Opendp::createNetwork()
       continue;
     }
     if (!bterm->getFirstPinPlacementStatus().isPlaced()) {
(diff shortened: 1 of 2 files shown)


src/dpl/src/dbToOpendp.cpp
```diff
@@ -142,6 +142,10 @@ void Opendp::createNetwork()
     if (!net || net->getSigType().isSupply()) {
       continue;
     }
+    if (!bterm->getFirstPinPlacementStatus().isPlaced()) {
+      // skip unplaced terminals
```
maliberty, line 147: Do you think we should have a warning here?

gadfort: We can, what should the say? `BTerm {} has is unplaced.`?

maliberty: Sure (drop `has`)

PR #9598, opened 2026-03-02
Add support for multicore in global routing, especially focusing on speeding up congestion cleanup iterations

ebrevdo:
## Summary

Make snapshot-batched FastRoute the default semantic mode for `global_route` by setting the default `-snapshot_batched_width` to `16`. Users can still force the legacy non-batched path with `-snapshot_batched_width 0`, while execution width continues to follow `set_thread_count`.

The public GRT surface is updated so the new default is visible and testable through Tcl, SWIG, and `GlobalRouter`, instead of being hidden behind internal-only knobs. The same range also wires OpenROAD thread-count changes into GRT so router-side reporting and post-route accounting use the requested thread count.

  - `src/grt/src/GlobalRouter.tcl` adds and defaults `-snapshot_batched_width`
  - `src/grt/README.md` documents the new default and the `0` escape hatch
  - `src/grt/src/GlobalRouter.i` exposes `get_snapshot_batched_width()` and `get_snapshot_batch_count()`
  - `src/grt/include/grt/GlobalRouter.h` and `src/grt/src/GlobalRouter.cpp` carry the new width/thread plumbing
  - `src/OpenRoad.cc` now forwards `setThreadCount()` into `GlobalRouter::setNumThreads()`

## Router Changes

The FastRoute core now batches contiguous nets against frozen planar congestion snapshots, routes each batch on worker copies, and then commits the results back in deterministic order. The implementation keeps bounded serial cleanup for late overflow-reduction iterations and skips snapshot batching in cases that are unlikely to benefit, such as debug flows, tiny net sets, or non-soft NDR nets.

The same range also adds better internal timing and accounting around the router phases, and parallelizes some read-only reporting work so multi-threaded runs scale past the maze kernel itself. That includes wirelength, per-layer resource totals, and congestion summaries.

  - `src/grt/src/fastroute/src/maze.cpp` splits `FastRouteCore::mazeRouteMSMD()` from `mazeRouteMSMDSequential()` and adds snapshot-wave execution
  - `src/grt/src/fastroute/src/FastRoute.cpp` adds snapshot worker construction, batch heuristics, cleanup gates, and per-phase metrics
  - `src/grt/src/fastroute/include/Graph2D.h` and `src/grt/src/fastroute/src/graph2d.cpp` add `Graph2D::copyRoutingStateFrom()`
  - `src/grt/src/GlobalRouter.cpp` parallelizes `computeWirelength()` and propagates snapshot width into `configFastRoute()`
  - `src/grt/src/fastroute/src/FastRoute.cpp` parallelizes `getOriginalResources()` and `computeCongestionInformation()`

## Tests And Compatibility

The regression coverage is retargeted to a real congestion-heavy stress case instead of relying on earlier goldens that no longer guaranteed batching. The new smoke, incremental-state, bus-route, and thread-report tests all exercise the current width-16 default, and the shared helper now asserts that snapshot batching actually executed.


maliberty:
Please note you need to sign your commits to pass DCO (ie `git commit -s`).  See the DCO link for info on how to fix existing commits.

src/grt/README.md
```diff
@@ -46,6 +47,7 @@ global_route
 | `-critical_nets_percentage` | Set the percentage of nets with the worst slack value that are considered timing critical, having preference over other nets during congestion iterations (e.g. `-critical_nets_percentage 30`). The default value is `0`, and the allowed values are integers `[0, MAX_INT]`. |
 | `-skip_large_fanout_nets` | Skips routing for nets with a fanout higher than the specified limit. Nets above this pin count threshold are ignored by the global router and will not have routing guides, meaning they will also be skipped during detailed routing. This option is useful in debugging or estimation flows where high-fanout nets (such as pre-CTS clock nets) can be ignored. The default value is 0, indicating no fanout limit. The default value is `MAX_INT`. The allowed values are integers `[0, MAX_INT]`. |
 | `-allow_congestion` | Allow global routing results to be generated with remaining congestion. The default is false. |
+| `-multicore` | Enable the optional multicore Track B routing path. |
```
maliberty: What does "Track B routing path" mean?

PR #9602, opened 2026-03-03
bazel-orfs: bump

oharboe:
pick up rules_python fix


oharboe:
@hzeller :crossed_fingers: this will fix the rules_json issues.

maliberty (approved):

PR #9603, opened 2026-03-03
Use `@rules_python` directly, don't rename.

hzeller:


hzeller:
Yay, CI passed :tada:

PR #9608, opened 2026-03-03
All uses of `WORKSPACE` are gone; stop requesting that bazel feature.

hzeller:


hzeller:
Let's see if the CI stumbles upon a sub-dependency that still needs it...

hzeller:
The CI's fail for unrelated reasons
  * one fails in bazel, but at the `pads_black_parrot_flipchip_connect_overpads-tcl_test` that is known to be flaky
  * the other in some ubuntu build (which I suspect is cmake, so not bazel)

So I suggest to re-run these CI's

maliberty:
I'm satisfied there is no real issue here.

PR #9609, opened 2026-03-03
bazel: remove unused scalatest dep, document rules_scala use

oharboe:
After a quick investigation on what could be cleaned up:

Remove scala_deps.scalatest() as it is not used in any BUILD file. Add comment documenting that rules_scala is used by chisel_binary from @bazel-orfs in test/orfs/mock-array.

@hzeller FYI, a small step in the right direction.


MODULE.bazel
```diff
@@ -15,6 +15,8 @@ bazel_dep(name = "rules_cc", version = "0.1.2")
 bazel_dep(name = "rules_flex", version = "0.3.1")
 bazel_dep(name = "rules_jvm_external", version = "6.4")
 bazel_dep(name = "rules_python", version = "1.8.5")
+
+# Used by chisel_binary from @bazel-orfs in test/orfs/mock-array
 bazel_dep(name = "rules_scala")
```
hzeller, line 20: wouldn't it be enough for `@bazel-orfs` to depend on `rules_scala`, or am I misreading this ?

oharboe: bazel-orfs provides macros that use rules_scala, so they are instantiated in the openroad context

PR #9615, opened 2026-03-03
A newer `yaml-cpp` does not need our local patch anymore.

hzeller:
While at it, update `rules_cc` as `yaml-cpp` was implicitly bumping it.


maliberty (approved):

hzeller:
github is currently hosed, a lot of URLs return 500. The CI might need to get restarted.

PR #9616, opened 2026-03-03
bazel: move openroad/ to python/openroad/ to fix output path conflict

oharboe:
The //:openroad cc_binary output path (bazel-out/k8-opt/bin/openroad) was a prefix of //openroad:openroadpy py_library output path (bazel-out/k8-opt/bin/openroad/__pycache__/...), causing Bazel to reject building both targets simultaneously.

Rename the package to //python/openroad to eliminate the collision while preserving the Python import path (import openroad) via imports = [".."].


oharboe:
@maliberty CI error fix? Testing...

https://jenkins.openroad.tools/job/OpenROAD-Public/job/PR-9592-head/11/stages/?selected-node=393

```
17:53:02  (17:53:00) ERROR: One of the output paths 'bazel-out/k8-opt/bin/openroad/__pycache__/odb.cpython-313.pyc' (belonging to //openroad:openroadpy) and 'bazel-out/k8-opt/bin/openroad' (belonging to //:openroad) is a prefix of the other. These actions cannot be simultaneously present; please rename one of the output files or build just one of them

```

oharboe:
@vvbandeira @maliberty clang tidy CI network outage

oharboe:
@QuantamHD Any idea about the CI error above and if Claude is onto something here?

hzeller:
I think I have occasionally seen the issue here as well, and this looks like the right solution.

oharboe:
@maliberty @vvbandeira  looks like more CI trouble w checkout

QuantamHD:
@oharboe Looks like you found the right solution, the error is basically saying you have a folder and a binary file named the same thing. You can't have both rename one.

maliberty:
I've seen this with `//...` but I usually just build `//:openroad`

maliberty:
GH is a bit janky today, not really in our control

oharboe:
> GH is a bit janky today, not really in our control

👍I see that you are on top of this and that it looks like it could help more than me to have this merged.

PR #9618, opened 2026-03-04
fix: add missing module and code args to utl::warn in detailed_route

suhr25:
## SUMMARY

This PR fixes a broken utl::warn call in TritonRoute.tcl that caused detailed_route to crash instead of showing a warning.
Now, when -droute_end_iter is set above 64, it prints a warning and continues as intended.

---

## FIX

**Before (invalid call):**

```tcl
utl::warn "-droute_end_iter cannot be greater than 64. Setting -droute_end_iter to 64."
```

**After (correct form):**

```tcl
utl::warn DRT 518 "-droute_end_iter cannot be greater than 64. Setting -droute_end_iter to 64."
```

Adds missing tool tag and warning ID. No logic change.

---

## VERIFICATION

Run:

```
detailed_route -droute_end_iter 80
```

Expected: warning is printed, value is clamped to 64, routing continues.

### DIFF
--- a/src/drt/src/TritonRoute.tcl
+++ b/src/drt/src/TritonRoute.tcl
@@ -42,6 +42,7 @@ proc detailed_route { args } {
            -single_step_dr -save_guide_updates}
   sta::check_argc_eq0 "detailed_route" $args
 
+  set MAX_DROUTE_END_ITER 64
   set enable_via_gen [expr ![info exists flags(-disable_via_gen)]]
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @maliberty,
This PR correctly fixes a crash that occurred when using -droute_end_iter with a value greater than 64
Please take a look at this PR when you get time.
Thanks!

PR #9620, opened 2026-03-04
Use same `rules_scala` version bazel-orfs

hzeller:
That also allows us to not have a git override.


oharboe:
If the computer says yes, it is good to go.

oharboe (approved):
Nice. As long as the computer says yes, we're good to go.

oharboe:
There's a [bazel bcr chisel module](https://registry.bazel.build/modules/rules_chisel), but it too leaves the actual scala configuration a user concern

PR #9621, opened 2026-03-04
Add missing `break` in switch/case

hzeller:
The case label `sta::FuncExpr::Op::zero` in `ClockGating.cpp` looks like a genuine oversight and this will possibly modify behavior. Move all `break;` in this section into the curly brace for easier readability.

The remaining `break;` changes are fixing benign situations (right before an empty default label).


hzeller:
FYI @kbieganski (looks like you edited the code in `ClockGating.cpp` last)

kbieganski:
Thanks for catching it

PR #9623, opened 2026-03-04
rcx: fix diff_spef flags always evaluating to false

suhr25:
## SUMMARY

This PR fixes a bug in **`diff_spef`** where comparison flags (`-r_res`, `-r_cap`, `-r_cc_cap`, `-r_conn`) were silently ignored.
The issue was caused by incorrect flag checks in **`src/rcx/src/OpenRCX.tcl`**, so user-provided options were never detected.

---

## FIX 
Update the checks to reference the correct flags.

```tcl
# Before
set res [info exists flags(-over)]

# After
set res [info exists flags(-r_res)]
```

The same correction is applied for `-r_cap`, `-r_cc_cap`, and `-r_conn`.

---

## VERIFICATION

Run:

```
diff_spef -file design.spef -r_res
```

**Before:** all flags were `0`, so the option had no effect.
**After:** the selected flag is correctly detected and passed to `rcx::diff_spef`.


suhr25:
Hi @maliberty,
This pull request effectively resolves a critical bug in the diff_spef.
Thanks!

maliberty (approved):

maliberty:
Thanks.  How did you happen to come across this?

suhr25:
Hi,
While reviewing the diff_spef implementation in OpenRCX.tcl, I noticed the flags (-r_res, -r_cap, etc.) were checked against the wrong key (-over).
I tested it and confirmed the options were ignored, so I submitted the fix.
Thanks for the merge!

PR #9628, opened 2026-03-04
dpl: single main loop and new reports

gudeh:
This PR is no-op.

Changes include:
 - It renames some important functions. Such as mapMove() renamed to diamondMove(), and shiftMove() to ripUpAndReplace().

 - Currently we perform the main loop twice, once checking if instances are multi-height and another time to check if instances are single height. This is not necessary, we just need to properly order them at `sorted_cells`. I modified `CellPlaceOrderLess` so we now sort cells first based on their height and use a single main loop, the other sorting criteria are kept the same (area, center distance, than name). 

 - This PR also include new reports:
   - Exposing the effective max displacement:
`[INFO DPL-0005] Max displacement: 500 multiples of cell width, 100 multiples of row height.`
   - Total types of movements performed:
```
Movements Summary
---------------------------------------
Total cells:                     555
Diamond Move Success:            555 (100.00%)
Diamond Move Failure:              0
Rip-up and replace Success:        0 (  0.00% of diamond failures)
Rip-up and replace Failure:        0
Total Placement Failures:          0
---------------------------------------
```

### DIFF
--- a/src/cts/test/hier_insertion_delay.ok
+++ b/src/cts/test/hier_insertion_delay.ok
@@ -53,14 +53,92 @@ Using 2 tracks default min distance between IO pins.
 [INFO DPL-0005] Max displacement: +/- 500 sites horizontally, +/- 100 rows vertically.
 Movements Summary
 ---------------------------------------
-Total cells:                     279
-Diamond Move Success:            278 ( 99.64%)
-Diamond Move Failure:              1
+Total cells:                     278
+Diamond Move Success:            278 (100.00%)
+Diamond Move Failure:              0
 Rip-up and replace Success:        0 (  0.00% of diamond failures)
-Rip-up and replace Failure:        1
-Total Placement Failures:          1
+Rip-up and replace Failure:        0
+Total Placement Failures:          0
 ---------------------------------------
-[INFO DPL-0034] Detailed placement failed on the following 1 instances:
-[INFO DPL-0035]  inst_0
-[ERROR DPL-0036] Detailed placement failed.
-Error: hier_insertion_delay.tcl, 19 DPL-0036
+Placement Analysis
+---------------------------------
+total displacement       9387.3 u
+average displacement       33.6 u
+max displacement           51.8 u
+original HPWL             605.0 u
+legalized HPWL            641.4 u
+delta HPWL                    6 %
+
+[INFO CTS-0050] Root buffer is CLKBUF_X3.
+[INFO CTS-0051] Sink buffer is CLKBUF_X3.
+[INFO CTS-0052] The following clock buffers will be used for CTS:
+                    CLKBUF_X3
+[INFO CTS-0049] Characterization buffer is CLKBUF_X3.
+[INFO CTS-0007] Net "clk" found for clock "core".
+[INFO CTS-0011]  Clock net "clk" for macros has 1 sinks.
+[INFO CTS-0011]  Clock net "clk_regs" for registers has 278 sinks.
+[INFO CTS-0008] TritonCTS found 2 clock nets.
+[INFO CTS-0097] Characterization used 1 buffer(s) types.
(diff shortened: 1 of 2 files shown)


gudeh:
I was considering including a displacement report specifically for buffers, but I wasn’t able to integrate STA into the DPL context. I’m wondering whether it would be acceptable to add an STA pointer to DPL.

gudeh:
I made sure the changes are no-op with the secure-CI: https://jenkins.openroad.tools/blue/organizations/jenkins/OpenROAD-flow-scripts-Private/detail/secure-dpl-multiple-improvements/2/pipeline

src/dpl/src/Opendp.cpp
```diff
@@ -141,6 +141,13 @@ void Opendp::detailedPlacement(const int max_displacement_x,
     max_displacement_y_ = max_displacement_y;
   }
 
+  logger_->info(DPL,
+                5,
+                "Max displacement: {} multiples of cell width, {} multiples of "
```
osamahammad21:
```suggestion
                "Max displacement: {} multiples of site width, {} multiples of "
```

osamahammad21 (commented):
LGTM except that small requested change in the log message.

src/dpl/src/Opendp.cpp
gudeh: Ah! Good catch. This message refers to `diamondMove()` (old `mapMove()`). I changed the message a bit, hope it is more clear.

I got confused with the `ripUpAndReplace()` (old `shiftMove()`) displacement which does use multiples of cell width. BTW, the `ripUpAndReplace()` is currently a fixed value and not modifiable by the user. In other words, we have different displacement ranges for  `diamondMove()` and `ripUpAndReplace()`.

eder-matheus:
@gudeh You have a bunch of unit tests failing. Probably related to the message change.

src/cts/test/hier_insertion_delay.ok
```diff
@@ -50,85 +50,17 @@ Using 2 tracks default min distance between IO pins.
...
+Total cells:                     279
+Diamond Move Success:            278 ( 99.64%)
+Diamond Move Failure:              1
+Rip-up and replace Success:        0 (  0.00% of diamond failures)
+Rip-up and replace Failure:        1
+Total Placement Failures:          1
+---------------------------------------
+[INFO DPL-0034] Detailed placement failed on the following 1 instances:
+[INFO DPL-0035]  inst_0
+[ERROR DPL-0036] Detailed placement failed.
```
eder-matheus: I don't think this failure was expected.

gudeh: Indeed! An initial visual inspection shows no differences after DPL finishes in both cases (branch and master):

Image

Image

The failed instance inst_0 is that purple macro on the origin. I will investigate further.

gudeh: I got the bottom of it. The master version has a variable which I removed: `have_multi_row_cells_`, which was used on one of the two main loops. Notice when setting it we check for `isCore()`:
```
    if (inst->getMaster()->isCore()
        && network_->getMaster(inst->getMaster())->isMultiRow()) {
      have_multi_row_cells_ = true;
    }
```
But we do not check for` isCore()` with the function `isMultiRow()`:
```
bool Opendp::isMultiRow(const Node* cell) const
{
  return network_->getMaster(cell->getDbInst()->getMaster())->isMultiRow();
}
```

I believe the test contains a non-fixed macro, which triggers the issue. Maybe `Opendp::isMultiRow()` should include the isCore(), or we should check for isCore() when instantiating instances, what do you think @osamahammad21 ?

gudeh: The latest push should keep the same previous behavior.

Issue #9629, opened 2026-03-04
Feature Request: Publish OpenROAD as a Bazel Central Registry (BCR) Module


oharboe:
## Use Case

Projects like [`bazel-orfs`](https://github.com/The-OpenROAD-Project/bazel-orfs) and other downstream Bazel-based EDA flows need to depend on OpenROAD. Today, this requires a `git_override` pointing at a specific commit:

```starlark
bazel_dep(name = "openroad")
git_override(
    module_name = "openroad",
    commit = "...",
    remote = "https://github.com/The-OpenROAD-Project/OpenROAD.git",
)
```

This has several drawbacks:

- **No version resolution** — `git_override` bypasses bzlmod's version selection, so diamond dependency conflicts are invisible until they break at build time.
- **Slow fetches** — every clean build clones the full OpenROAD repo instead of downloading a compact archive.
- **No reproducibility guarantees** — commit SHAs are opaque; there is no integrity check, no mirroring, and no BCR CI validation.
- **Friction for new users** — downstream projects cannot simply write `bazel_dep(name = "openroad", version = "2025.1.0")` and have it work.

Publishing OpenROAD to the BCR would allow downstream flows to depend on it like any other module:

```starlark
bazel_dep(name = "openroad", version = "2025.1.0")
```

This enables proper version resolution, cached archive downloads, integrity verification, and a standard onboarding experience for the growing ecosystem of Bazel-based EDA tooling built on OpenROAD.

---

## Current State

- **Module name**: `openroad` (defined in `MODULE.bazel`, no version set)
- **Bazel version**: 8.5.0, fully on bzlmod (no WORKSPACE file)
- **88 BUILD files** across `src/`, `third-party/`, `bazel/`, `test/`, etc.
- **Dependencies**: ~60+ `bazel_dep` entries; most already in BCR
- **Blockers**: 2 dependencies use `git_override` (not in BCR):
  - `qt-bazel` (The-OpenROAD-Project/qt_bazel_prebuilts)
  - `bazel-orfs` (The-OpenROAD-Project/bazel-orfs)
- **Resolved**: `rules_scala` now uses BCR version `7.1.5`
- **Non-BCR fetches**: 2 `http_archive` calls (`circt` firtool, `openmp`)

---

### Phase 1: Resolve Dependency Blockers

All `bazel_dep` dependencies **must** already exist in the BCR before OpenROAD can be published.

#### 1.1 Classify dependencies as required vs dev-only

[...]

PR #9630, opened 2026-03-04
Mark bazel-orfs, rules_scala, rules_jvm_external, circt as dev_dependency

oharboe:
@hzeller FYI

These dependencies are only used for OpenROAD's own test infrastructure (ORFS integration tests, Chisel mock-array, firtool binary) and are not needed by downstream consumers of the openroad module.

Whittles away at #9629

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -179,15 +179,9 @@ maven.install(
 )
 use_repo(maven, "maven")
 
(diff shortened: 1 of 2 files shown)


hzeller:
Nice!

oharboe:
fixed a gaffe

PR #9633, opened 2026-03-04
Don't use `&container[0]i` but `container.data()` instead.

hzeller:
Readability improvement. Done automatically by clang-tidy:

```
etc/run-clang-tidy.sh --checks=-*,readability-container-data-pointer --fix
```

### DIFF
--- a/src/ant/src/AntennaChecker.cc
+++ b/src/ant/src/AntennaChecker.cc
@@ -318,7 +318,7 @@ void AntennaChecker::Impl::saveGates(odb::dbNet* db_net,
     dsu_parent[i] = i;
   }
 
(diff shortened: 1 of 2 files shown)


src/ant/src/AntennaChecker.cc
```diff
@@ -318,7 +318,7 @@ void AntennaChecker::Impl::saveGates(odb::dbNet* db_net,
     dsu_parent[i] = i;
   }
 
-  boost::disjoint_sets<int*, int*> dsu(&dsu_size[0], &dsu_parent[0]);
+  boost::disjoint_sets<int*, int*> dsu(dsu_size.data(), dsu_parent.data());
```
maliberty, line 321: The change is ok but the original code looks odd.  @luis201420 @eder-matheus is this right?

maliberty (approved):

PR #9639, opened 2026-03-05
odb: UnfoldedModel as a persistent object in dbDatabase

osamahammad21:


src/odb/src/db/dbDatabase.cpp
```diff
@@ -685,6 +691,15 @@ dbChip* dbDatabase::getChip()
...
 }
 
+UnfoldedModel* dbDatabase::getUnfoldedModel()
+{
+  _dbDatabase* db = (_dbDatabase*) this;
+  if (db->unfolded_model_ == nullptr) {
+    db->unfolded_model_ = new UnfoldedModel(db->logger_, getChip());
+  }
+  return db->unfolded_model_;
+}
```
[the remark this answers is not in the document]
osamahammad21: expected

osamahammad21:
@ahmed532 @jferreiraOpenRoad FYI

src/odb/src/db/dbDatabase.cpp
```diff
@@ -685,6 +692,19 @@ dbChip* dbDatabase::getChip()
   return (dbChip*) db->chip_tbl_->getPtr(db->chip_);
 }
 
+void dbDatabase::constructUnfoldedModel()
+{
+  _dbDatabase* db = (_dbDatabase*) this;
+  absl::MutexLock lock(unfolded_model_mutex);
```
maliberty: In general odb doesn't support MT access.  Do we need it here?  

If so we need to guard against the model already being created by the previous locker.  Also getUnfoldedModel will need a lock.

src/odb/src/db/dbDatabase.cpp
```diff
@@ -685,6 +692,19 @@ dbChip* dbDatabase::getChip()
   return (dbChip*) db->chip_tbl_->getPtr(db->chip_);
 }
 
+void dbDatabase::constructUnfoldedModel()
```
maliberty, line 694: Should we have auto creation in get rather than explicit construction?  Do we want get to return a nullptr?

src/odb/src/db/dbDatabase.h
```diff
@@ -335,6 +336,7 @@ class _dbDatabase : public _dbObject
 
   utl::Logger* logger_;
   std::set<dbDatabaseObserver*> observers_;
+  UnfoldedModel* unfolded_model_;
```
maliberty: mention that it is non-persistent

src/odb/src/db/dbDatabase.cpp
osamahammad21: I think this API is temporary and won't stick long in DRT when we make the unfoldedmodel dynamic and responsive to the database changes. I just added the mutex lock to be consistent with the table creation, but I will remove it now since we don't expect any user of the unfoldedModel to be mutli-threaded.

src/odb/src/db/dbDatabase.cpp
osamahammad21: We could, which I did earlier, but I think this introduces many issues for the current use cases. 

- If we create in the getter, do we create once or each time? If we create once, then any changes to the database after retrieving the unfoldedModel will never be reflected.
- The cpp test cases change the database objects directly and test that change on the checker, which needs a way of updating the unfoldedModel every time since it's static.

I would prefer to keep it that way temporarily until we make it dynamic.

maliberty: As it stand calling it twice leaks the previous unfolding.  At a minimum it needs to manage memory if you want to stick with this temporarily

I think get should create if needed or return otherwise.  Updating the unfolding dynamically should be a separate, automatic process inside the database.

osamahammad21: Yes, the memory leak stands and will be fixed by destroying the unfoldedModel before reconstructing it.
However, the dynamic update is planned but not included in this PR because it requires some work to listen for database objects callbacks. So is it ok to merge this PR and handle the automatic updates in future PRs or is it preferred to handle that in this PR?

maliberty:
Its fine for dynamic to be a future PR.  I was just pointing to where I think we will end up and create won't be necessary at that point.

PR #9641, opened 2026-03-05
WIP: GRT: Add detour metric tracking for routing stages (#8941)

Divinesoumyadip:
Hey @eder-matheus  , opening this as PR to get the initial architecture up. I've wired up the full Tcl -> C++ pipeline based on your feedback.

**What's working so far:**

-     Added the report_net_global_routing_detour -nets <net_names> Tcl command (wildcard support is built-in via get_nets).

-     Wired the SWIG bindings and C++ headers directly to GlobalRouter.

-      Implemented the core metric functions in FastRouteCore (computeTotalWirelength, computeNetSttWirelength, computeNetFinalWirelength).

-      Added the stt_wirelengths_ map to track STT wirelength per net.

**Still left to do:**

-     Hook up the internal FastRoute netID mapping to the new GlobalRouter::reportNetGlobalRoutingDetour function.

-     Clean up and format the final logger output.

Let me know if this structural direction looks good to you before I finalize the internal variable mapping!

PR #9645, opened 2026-03-05
grt: add incremental routing support to CUGR

sparsh-karna:
## Summary
Adds incremental global routing support to the CUGR backend, 
mirroring the existing FastRoute incremental API.

## Changes
- Added `addDirtyNet()`, `startIncremental()`, `endIncremental()` 
  public methods to CUGR
- Added private `rerouteNets()` helper to avoid code duplication
- Incremental mode rips up dirty nets, reroutes them through all 
  three routing stages, and propagates overflow to neighboring nets
- Sort dirty nets by slack (then HPWL) before rip-up phase in 
  rerouteNets(), ensuring critical nets reclaim routing resources 
  first, mirroring FastRoute's incremental behavior

## Testing
- All existing grt regression tests pass
- Builds cleanly on Linux (GCC 13) and but on macOS (AppleClang 17)

### DIFF
--- a/docs/contrib/DeveloperGuide.md
+++ b/docs/contrib/DeveloperGuide.md
@@ -198,9 +198,9 @@ Instructions for building are available [here](../user/Build.md).
 
 ## Example of Adding a Tool to OpenROAD
 
-The directory `OpenRoad/src/example` 
+The directory [src/exa](../main/src/exa/README.md)
 illustrates a tool named "Example" that uses the file structure described above.
- It defines a command to run the tool as illustrated below:
+It defines a command to run the tool as illustrated below:
 
 ```tcl
 > example_instance -name test


Divinesoumyadip:
The bot makes a good point regarding the O(N+M) STL optimization for the secondary nets, but looking at the higher-level architecture, I have a concern regarding the grid state synchronization.

I’ve been deep in the grt core recently wrapping up the GridGraph architectural refactor (PR #9642) and detour metric tracking (PR #9641), and I'm very interested in how your incremental API interacts with the underlying grid state during the rip-up phase.

When rerouteNets() rips up the dirty nets, how are you ensuring that the 3D edge capacities and history costs within the GridGraph are synchronously decremented before the CUGR engine triggers the reroute? If the grid state isn't explicitly cleared from the previous routing iteration, we risk hitting "phantom congestion" where the incremental pass avoids tracks that are actually free.

We should probably align your startIncremental() and net invalidation hooks with the new GridGraph coordinate search logic I'm finalizing to ensure accurate capacity tracking during high-frequency timing loops.

sparsh-karna:
> ## Code Review
> This pull request introduces incremental routing capabilities to the CUGR global router, which is a significant feature enhancement. The implementation mirrors the existing FastRoute API by adding `addDirtyNet`, `startIncremental`, and `endIncremental` methods. A new helper function `rerouteNets` is introduced to avoid code duplication, which is good practice. The logic correctly handles rerouting of dirty nets and also considers secondary overflows by rerouting neighboring affected nets. The code is well-structured. I have one suggestion to improve the implementation by using a standard library algorithm which can make the code more concise and performant.
> 
> **Note:** Security Review did not run due to the size of the PR.

Good catch — applied in the latest commit.

eder-matheus:
@sparsh-karna DCO is failing in the PR. You need to sign-off your commits by adding the -s flag to the git commit call: git commit -s -m "<your commit message>".

Check the DCO link in the checks for advise on how to fix DCO.

I'll review the code in details soon.

sparsh-karna:
> @sparsh-karna DCO is failing in the PR. You need to sign-off your commits by adding the -s flag to the git commit call: git commit -s -m "".
> 
> Check the DCO link in the checks for advise on how to fix DCO.
> 
> I'll review the code in details soon.

got it, will keep that in mind for future

src/grt/src/cugr/src/CUGR.cpp
```diff
@@ -612,4 +615,59 @@ void CUGR::getBTermsAccessPoints(
...
 }
 
+void CUGR::addDirtyNet(odb::dbNet* net)
+{
+  auto it = db_net_map_.find(net);
+  if (it != db_net_map_.end()) {
+    dirty_net_indices_.insert(it->second->getIndex());
+  } else {
+    logger_->warn(
+        GRT, 600, "Net {} not found in CUGR net map.", net->getConstName());
```
eder-matheus: During incremental routing, new nets can be created by buffer insertion, and they'll not be inside CUGR map yet. We need to ensure that the new nets are created in these cases.

This warning would make sense if the process of adding the new nets to CUGR structures is performed before the addDirtyNet calls.

src/grt/src/cugr/src/CUGR.cpp
```diff
@@ -612,4 +615,59 @@ void CUGR::getBTermsAccessPoints(
...
+  }
+}
+
+void CUGR::startIncremental()
+{
+  incremental_mode_ = true;
+  dirty_net_indices_.clear();
+}
```
eder-matheus: I believe a simpler approach would be to mimic how incremental works for FastRouteCore. Instead of managing the start/end incremental, FastRouteCore looks at the list of nets that will be routed on the run. All the management regarding detecting the dirty nets is performed by the GlobalRouter class.

My suggestions would be:
- Instead of a dirty nets list inside CUGR, we'd have a list of nets to route (similar to FastRouteCore)
- Prior to incremental route, all the nets would be routed, so this list would have all the net IDs
- On incremental calls, GlobalRouter will be responsible to tell CUGR what are the dirty nets

APIs to include new individual nets, and to update the net pins will also be necessary.

PR #9646, opened 2026-03-05
Removed log spam from write_verilog

hemannt003:
What changed

Updated write_verilog in:
src/sta/verilog/Verilog.tcl
src/dbSta/src/dbReadVerilog.tcl
Removed the warning emission for -sort ("The -sort flag is ignored."). 

Result
write_verilog still accepts -sort for compatibility. It now ignores -sort silently, so repeated calls won’t flood logs.

### DIFF
--- a/src/dbSta/src/dbReadVerilog.tcl
+++ b/src/dbSta/src/dbReadVerilog.tcl
@@ -34,11 +34,12 @@ sta::define_cmd_args "write_verilog" {[-sort] [-include_pwr_gnd]\
 proc write_verilog { args } {
   sta::parse_key_args "write_verilog" args keys {-remove_cells} \
     flags {-sort -include_pwr_gnd}
-  if { [info exists flags(-sort)] } {
-    if { ![info exists ::_write_verilog_sort_warned] } {
-      utl::warn STA 2065 "The -sort flag is ignored."
-      set ::_write_verilog_sort_warned 1
-    }
+  if { [info exists ::write_verilog_sort_warned] == 0 } {
+    set ::write_verilog_sort_warned 0
+  }
+  if { [info exists flags(-sort)] && !$::write_verilog_sort_warned } {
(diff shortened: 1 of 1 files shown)


maliberty:
I'm unclear about the motivation here.  How often are you calling write_verilog?  No other command implements deprecation this way so it is irregular.

Hemanntnagar:
@maliberty Thanks for pointing this out.

The motivation behind this change was to mark the old behavior of `write_verilog` as deprecated so that users are aware of the newer recommended approach. I noticed that the command is still used in some scripts, so I thought adding a deprecation notice could help guide users toward the updated usage.

However, I understand your concern about the implementation being inconsistent with other commands. I’ll take another look at how deprecation is handled elsewhere in the project and adjust this change to follow the existing pattern.

Please let me know if there is a preferred way deprecations should be implemented for commands like this.

maliberty:
I'm not sure what is wrong with what is there.  As the comment says
```
# Copied from sta/verilog/Verilog.tcl because we don't want sta::read_verilog
# that is in the same file.
```
so the same exists in sta.

Hemanntnagar:
You're right that completely removing the warning could hide the fact that the `-sort` flag is deprecated. The original intent was to reduce the repeated log spam when `write_verilog` is called multiple times.

I've updated the implementation so that the warning is **emitted only once per session instead of being removed entirely**. This keeps users informed about the deprecated `-sort` flag while preventing repeated log messages.

The change has been applied consistently in:

* `src/sta/verilog/Verilog.tcl`
* `src/dbSta/src/dbReadVerilog.tcl`

Please let me know if this approach aligns with the expected behavior.

maliberty:
I'm still unclear why we need to limit it.  write_verilog is probably only going to be used once in a session or at most a handfull of times.  Not other deprecation method implements such a limit so I'm not sure why just this one should.  In short I'm unclear what the problem is that this is solving.

PR #9650, opened 2026-03-05
gpl: skip incremental no unlocked instances are identified

gudeh:
GPL incremental is currently being run at the end of stage 3-4 (resize.tcl). If it identifies no unplaced instances and the overflow is above 10% it will restart the whole placement, undoing all TD and RD work from stage 3-3 (global_place.tcl). This happens at least with sky130hd/aes.

To avoid this I opted to just skip GPL incremental when no unplaced instances are identified.


maliberty:
Does this have an QOR impact (ie secure-CI needed)?

gudeh:
yes, a few designs may fall into this situation. There is an open PR for the metrics: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/3958

I also put it in the queue.

PR #9653, opened 2026-03-05
ifp: fix insert_tiecells null check using wrong variable

suhr25:
## SUMMARY

This PR fixes a bug in `insert_tiecells` where the null check after `findMTerm` was checking `$master` instead of `$mterm`.
Because `$master` had already been validated earlier, the check never triggered and a `NULL` pin could be passed to `insert_tiecells_cmd`.

---

## FIX

Update the null check to validate the correct variable.

**Before**

```
if { $master == "NULL" }
```

**After**

```
if { $mterm == "NULL" }
```

This ensures an invalid pin name is caught before calling `insert_tiecells_cmd`.

---

## VERIFICATION

Run:

```
insert_tiecells TIEHI_X1/WRONG_PIN
```

**Before:** the check is skipped and a NULL `mterm` may reach the C++ command.
**After:** OpenROAD correctly reports "Unable to find master pin".


eder-matheus:
@suhr25 DCO is failing in the PR. You need to sign-off your commits by adding the -s flag to the git commit call: git commit -s -m "<your commit message>".

Check the DCO link in the checks for advise on how to fix DCO.

suhr25:
I will fix it.
Thanks!

suhr25:
Hi @eder-matheus @maliberty ,
I have made changes.
Thanks

suhr25:
Hi @eder-matheus @maliberty ,
Kindly following up on this PR.
Thanks!

eder-matheus:
@suhr25 Thanks for the bug fix, it was a good catch!

Issue #9635, opened 2026-03-05
bazel: After install, openroad still reads file openroad.repo_mapping from build location


fredowski:
### Describe the bug

I moved the openroad exe and the openroad runtime directory to another location to have the applicaton standalone. After deleting the build environment, starting the application results in a crash for the gui. For the cli, the tcl.init is not found - indicating a wrong access to the runtime.

I traced this with strace and found that still the openroad.repo_mapping file is accessed in the build directory.

When I delete the MANIFEST file in the runtime, then it works. 

I opened an issue in rules_cc: https://github.com/bazelbuild/rules_cc/issues/620

### Expected Behavior

No remaining files in the build directory needed.

### Environment

```shell
debian 13 aarch64
```

### To Reproduce

- Build openroad with bazel
- Copy the exe and the runtime out of the bazel-bin directory to /tmp/install with `cp -r -L openroad.runtime /tmp/install/`and `cp openroad /tmp/install/`
- run `bazelisk clean` or rename ~/.cache 
- run `/tmp/install/openroad -gui` => crash

### Relevant log output

```shell

```

QuantamHD:
This might be somewhat intended behavior. If you want a relocatable distribution I believe you're supposed to use the rules_pkg package rules to generate a tar https://bazelbuild.github.io/rules_pkg/latest.html#pkg_tar

fredowski:
But we have this //:install ?

maliberty:
My guess is that it is not correct then.  It was probably vibe coded (see [Co-authored-by: gemini-code-assist[bot]](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9374/changes/4bee233fa133bc799bcd79aab11e390d6d51b264))

QuantamHD:
In particular you need to make sure to pass `include_runfiles = True`. The rules can also generate rpms and debs. The other reason it's important to use these rules is that the things in bazel-bin dir are often symlinked multiple times, and when you just naively copy the folder you're often breaking the symlinks and or not actually including the real contents of the files.

oharboe:
> My guess is that it is not correct then. It was probably vibe coded (see [Co-authored-by: gemini-code-assist[bot]](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9374/changes/4bee233fa133bc799bcd79aab11e390d6d51b264))

Yes.

Note that this has been enormously useful in day-to-day development since DependencyInstaller/CMake no longer works for me.

[...]

PR #9656, opened 2026-03-05
bazel: use pkg_tar for the install procedure

fredowski:
I updated the install procedure to use the pkg_tar module. The install works as before, but now you
can also install optionally to another location.

Closes: #9635

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -4,6 +4,7 @@
 load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
 load("@rules_python//python:defs.bzl", "py_library")
 load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
 load("//bazel:notification.bzl", "notification_rule")
@@ -411,5 +412,14 @@ sh_binary(
     name = "install",
     srcs = ["//bazel:install.sh"],
     args = [],
-    data = [":openroad"],
+    data = [":tarfile"],
+)
+
+pkg_tar(
+    name = "tarfile",
+    srcs = [
+        ":openroad",
+    ],
+    include_runfiles = True,
+    package_file_name = "openroad.tar",
 )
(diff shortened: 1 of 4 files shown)


QuantamHD:
Do you need to remove the repo mapping file?

fredowski:
> Do you need to remove the repo mapping file?

It will work also if I do not delete that file. But the relevant repo_mapping file is in `openroad.runfiles/_repo_mapping`.  Without that, the gui crashes.

oharboe:
The approach looks good — using `pkg_tar` with `include_runfiles = True` is the right way to make a relocatable install per Bazel conventions.

One request: the custom install path usage instructions are currently buried in comments inside `bazel/install.sh`:

```bash
# Run this from bazel build directory with
# bazelisk run --//:platform=gui //:install

# If you want to install to another location, e.g. /tmp, then do
# bazelisk run --//:platform=gui //:install -- /tmp
```

These should go into `docs/user/Build.md` instead, where users will actually find them. The Bazel install section there (lines 19-28) already documents the basic install command but doesn't mention the custom path option. Something like:

```markdown
## Build and install with Bazel

Build OpenROAD with GUI support and install into ../install/OpenROAD/bin

    bazelisk run --//:platform=gui //:install

To install to a custom location:

    bazelisk run --//:platform=gui //:install -- /tmp/myinstall

To embed the real git version string, add `--config=release`:

    bazelisk run --config=release --//:platform=gui //:install

See [Bazel](Bazel.md) for more details on testing, profiling and build configurations.
```

Then remove the usage comments from `install.sh` — shell scripts shouldn't be the documentation surface for user-facing features.

fredowski:
Good idea! I adapted the description in `docs/user/Build.md` and removed the comments in the install.sh script.

BUILD.bazel
```diff
@@ -411,5 +412,23 @@ sh_binary(
...
+)
+
+# Produce a tarfile including the binary and the runtime files
+# The layout is
+# ./openroad
+# ./openroad.runfiles/...
+# This can be untarred and run standalone
+# This includes
+# ./openroad.runfiles/_main/openroad
+# which is again the binary.
```
oharboe: remove these comments, these days you can ask e.g. Claude to explain what is going on and it will read the README.md and these comments are only going to be out of date....

oharboe:
@fredowski FYI, this project prefers not force pushing because then the review history is preserved. This makes the commit history very messy, but it is more information. Perhaps in this day and age of AI, that is information that can be used to train an automatic reviewer :smile:

oharboe:
@fredowski and perhaps reading a messy git history automatically with AI to answer questions for hoomans will pay off...

BUILD.bazel
fredowski: Thanks for your review! I removed the comments.

fredowski:
O.k. - yes the history vanishes with the squashing...

maliberty (approved):

fredowski:
Thanks everybody for reviewing the code and spending your time on this topic!

oharboe:
:relieved:

PR #9680, opened 2026-03-07
ram: add nangate45 cross-platform regression test

### LINKED ISSUE #9468: ram: Cross platform testing - nangate45
The ram module has only been tested with sky130hd. Verify that it works with nangate45 as well.

There should be an additional ram test added that verifies that nangate45 works, similar to the test for sky130. Preferably with a different, odd size (e.g. 7x7)

DeepanshuDadhich:
add test make_7x7_nangate45 exercising generate_ram with Nangate45 standard cells. Uses an intentionally odd 7-word x 1-byte (56-bit) configuration to stress the decoder logic on a non-power-of-2 count.

also fix a latent bug in makeCellBit() where the clock-pin name probe only checked for CLK (sky130hd) or GATE (latch cells), silently missing CK which is the Nangate45 DFF_X1 clock pin name. The probe chain is extended to CLK -> CK -> GATE.

Resolves #9468.


src/ram/src/ram.cpp
```diff
@@ -116,11 +116,21 @@ std::unique_ptr<Cell> RamGen::makeCellBit(const std::string& prefix,
...
+  // Resolve the clock-pin name for the storage cell.
+  // sky130hd DFFs use "CLK", Nangate45 DFFs use "CK", and some latch-type
+  // storage cells use "GATE". Check in priority order so the correct pin
+  // is always found regardless of the technology node.
+  const char* clk_pin = "GATE";
+  if (storage_cell_->findMTerm("CLK")) {
+    clk_pin = "CLK";
+  } else if (storage_cell_->findMTerm("CK")) {
+    clk_pin = "CK";
+  }
```
QuantamHD: Instead of setting a fixed pin name you should use the OpenSTA APIs to query for the clock pin. Which is populated by the liberty data.

You can turn the cell name into a LibertyCell then iterate all of its ports then check which pin is the clock pin.

https://github.com/parallaxsw/OpenSTA/blob/master/include/sta/Liberty.hh#L832

DeepanshuDadhich: got it...thanks for the suggestion..i look into it...

DeepanshuDadhich:
Thanks for the helpful suggestions!

I’ve updated the implementation based on the feedback. Instead of relying on hard-coded clock pin names, the code now queries the Liberty data through the OpenSTA API to determine the clock pin for the storage cell.

This should make the detection logic more robust and technology-independent, without assuming specific pin names like `CLK`, `CK`, or `GATE`.

I also reran the regression tests, including the Nangate45 test added in this PR, and verified that the RAM generation flow still works as expected.

Please let me know if there’s anything else I should adjust or improve.

src/ram/test/make_7x7_nangate45.tcl
```diff
@@ -0,0 +1,42 @@
...
+  -filler_cells {FILLCELL_X1 FILLCELL_X2 FILLCELL_X4 FILLCELL_X8} \
+  -write_behavioral_verilog $behavioral_file
+
+set lef_file [make_result_file make_7x7_nangate45.lef]
+write_abstract_lef $lef_file
+diff_files make_7x7_nangate45.lefok $lef_file
+
+set def_file [make_result_file make_7x7_nangate45.def]
+write_def $def_file
+diff_files make_7x7_nangate45.defok $def_file
```
maliberty, line 40: I see no defok

src/ram/test/make_7x7_nangate45.tcl
```diff
@@ -0,0 +1,42 @@
...
+generate_ram \
+  -bytes_per_word 1 \
+  -word_count 7 \
+  -read_ports 1 \
+  -storage_cell DFF_X1 \
+  -power_pin VDD \
+  -ground_pin VSS \
+  -routing_layer {metal1 0.07} \
+  -ver_layer {metal2 0.07 0.42} \
+  -hor_layer {metal3 0.07 0.42} \
```
rovinski, line 30: How were these values determined?

src/ram/test/make_7x7_nangate45.tcl
```diff
@@ -0,0 +1,42 @@
...
+  -filler_cells {FILLCELL_X1 FILLCELL_X2 FILLCELL_X4 FILLCELL_X8} \
+  -write_behavioral_verilog $behavioral_file
```
rovinski, line 32: I don't think we need to test the behavioral flag. This really isn't significantly different from the 8x8 test.

src/ram/test/make_7x7_nangate45.tcl
rovinski: nor lefok

src/ram/test/make_7x7_nangate45_behavioral.vok
rovinski, line 1: I don't think .vok is needed for this test

rovinski (commented):
Please include layout screenshots to verify that the output looks correct

rovinski:
@DeepanshuDadhich I would kindly ask that you not take issues which are already assigned to others as mentioned in #9468. There are many many other issues not yet being worked on in #9392 which you could request assignment on.

One of the main reasons is that @braydenlouie and I discussed this and we decided to pause this while waiting on https://github.com/The-OpenROAD-Project/OpenROAD/pull/9690 which will probably affect this PR significantly.

PR #9684, opened 2026-03-07
Fix GUI crash rendering when Acess Points or Gcell grid is enabled 

### LINKED ISSUE #8927: GUI: Chiplets gui crashing
### Describe the bug

After reading a 3dbx design, the GUI crashes if Access Points or GCell grid are enabled.

### Expected Behavior

Shouldn't crash

### Environment

```shell
NA
```

### To Reproduce

.

### Relevant log output

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

alokkumardalei-wq:
## What does this PR resolves?
Fixes #8927

Resolves GUI crash that occurred when enabling Access Points or GCell grid display options after reading a 3DBX chiplet design.

## What have caused this issue?
The rendering functions `drawTracks`, `drawManufacturingGrid`, and `drawGCellGrid` used `viewer_->getBlock()`, which returns the top-level HIER chip's block (nullptr in 3DBX designs) instead of the per-chiplet block being rendered. This caused null pointer dereference crashes.

Additionally, `drawAccessPoints` and `drawChip` lacked nullptr guards for `block->getTech()` before iterating technology layers.

## Changes made to resolve this
- **renderThread.h / renderThread.cpp**: 
  - Added `odb::dbBlock* block` parameter to `drawTracks`, `drawManufacturingGrid`, and `drawGCellGrid`
  - Replaced all `viewer_->getBlock()` calls with the passed-in `block` parameter
  - Added nullptr guards for `block` and `block->getTech()` before dereferencing
  - Wrapped layer iteration in `drawChip` with `if (tech != nullptr)` check

### DIFF
--- a/src/gui/src/renderThread.cpp
+++ b/src/gui/src/renderThread.cpp
@@ -314,6 +314,7 @@ bool RenderThread::instanceBelowMinSize(dbInst* inst)
 
 void RenderThread::drawTracks(dbTechLayer* layer,
                               QPainter* painter,
+                              odb::dbBlock* block,
                               const Rect& bounds)
 {
   if (!viewer_->options_->arePrefTracksVisible()
@@ -321,12 +322,16 @@ void RenderThread::drawTracks(dbTechLayer* layer,
     return;
   }
 
-  dbTrackGrid* grid = viewer_->getBlock()->findTrackGrid(layer);
+  if (block == nullptr) {
+    return;
+  }
+
+  dbTrackGrid* grid = block->findTrackGrid(layer);
   if (!grid) {
     return;
   }
 
-  Rect block_bounds = viewer_->getBlock()->getDieArea();
+  Rect block_bounds = block->getDieArea();
   if (!block_bounds.intersects(bounds)) {
     return;
   }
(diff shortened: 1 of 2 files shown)


maliberty:
Did you reproduce the crash and master and verify that this fixes it?

alokkumardalei-wq:
Hi @maliberty, I verified the fix at the code level  but not at gui level— I
 traced the crash through renderThread.cpp
 and confirmed that block->getTech() can r
eturn nullptr for chiplet container blocks in 3
DBX designs, causing the naked dereference crash.

Also you can check the code level logic implemented in this commit 
[1191ec8](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9684/commits/1191ec878c51e0b066419ffa257966be65d3bbca) and please let me know  if it is correct  or not or there are any other suitable and alternative  approach for this.

Unfortunately, I don't have access to an 
actual 3DBX chiplet design dataset to trigger t
he crash at runtime, and @LucasYuki didn't include one in #8927. 
I should have asked for it earlier — apologies for not doing so.

If you could share a chiplet dataset (or point me to one), I would be happy to do a full GUI-level runtime test to confirm both the crash on master and the fix on this branch. Alternatively, let me know if there's another testing approach you prefer.

Thank you so much for time and looking forward for any kind of feedback further.

maliberty (approved):

PR #9685, opened 2026-03-07
bazel: disable ABC warnings by default via build flag

MeloveGupta:
During Bazel builds of OpenROAD, the vendored ABC sources emit several compiler warnings (e.g., -Wmissing-field-initializers, -Wimplicit-int-float-conversion). Because these originate from third-party code, they tend to drown out warnings produced by OpenROAD itself, making actionable warnings harder to identify.

This change suppresses warnings from the vendored ABC sources by default while still allowing developers to enable them when needed.

Implementation

Introduced a global Bazel build setting
//bazel/build_settings:abc_enable_warnings (default: false).

When the flag is disabled (default), -w is appended to ABC compilation options to silence warnings from the vendored sources.

When the flag is enabled, the suppression is removed and ABC warnings are emitted normally.

Added a convenience Bazel config in .bazelrc:

--config=abc_warnings

which enables the flag for developers who want to inspect ABC warnings.

The build setting is defined in bazel/build_settings rather than third-party/abc because the latter is excluded from direct Bazel package resolution via --deleted_packages.

No modifications were made to the ABC source code.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -64,6 +64,9 @@ build --copt "-Wno-dangling"  --host_copt "-Wno-dangling"
 build --per_file_copt=.*external/.*@-w
 build --host_per_file_copt=.*external/.*@-w
 
+# Opt-in config to surface warnings from the vendored ABC library.
+build:abc_warnings --//bazel/build_settings:abc_enable_warnings=true
+
 ##### Platform-specific configs (auto-selected) #############
 # --enable_platform_specific_config makes bazel automatically apply
 # build:macos on macOS, build:linux on Linux, etc.
--- a/bazel/build_settings/BUILD
+++ b/bazel/build_settings/BUILD
@@ -1,0 +1,7 @@
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+
+bool_flag(
+    name = "abc_enable_warnings",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)


maliberty:
I seem to be missing something - how does this flag control the -w?

MeloveGupta:
You're right, the current flag definition doesn't actually affect the -w suppression yet. My intention was to introduce the build setting first and then wire it into the ABC compilation flags so that -w is removed when abc_enable_warnings is true.

maliberty:
Can you do that in a single PR?  I don't see the purpose in having a dangling flag merged.

MeloveGupta:
Wired the global abc_enable_warnings flag into the ABC build: added abc_warnings_enabled + ABC_WARNING_COPTS (default -w) and appended it to all ABC targets (abc, demo, abc-lib), so ABC warnings are suppressed by default and re-enabled with --config=abc_warnings.
Kept .bazelrc opt-in config unchanged.
Submodule commit: 7a4eed534 (signed-off); main commit: d32f1ceaaf (signed-off) updates the submodule pointer.

MeloveGupta:
Added global build flag abc_enable_warnings in bazel/build_settings/BUILD (default false).
Wired that flag into all ABC Bazel targets in third-party/abc/BUILD via abc_warnings_enabled + ABC_WARNING_COPTS so -w is the default and is removed when the flag is true.
Added opt‑in config in .bazelrc: --config=abc_warnings sets the flag to show ABC warnings.
Submodule third-party/abc updated to commit 8272cc186; main repo commit a93fdc102a reflects the new flag file, bazelrc config, and submodule pointer (both commits are Signed‑off).
This was implemented in the latest patch...

maliberty:
I think you need to PR your abc changes as I can't see them from this PR

PR #9686, opened 2026-03-07
fix: typo in resynth_genetic argument -mutation_probability

suhr25:
## SUMMARY

Fixes a typo in the `resynth_genetic` argument definition in `src/rmp/src/rmp.tcl`. The option was documented as
 `-mututation_probability` while the parser expects `-mutation_probability`, causing the user value to be ignored.

---

## FIX

Correct the typo in `define_cmd_args` so it matches the parser.

**Before**

```tcl
[-mututation_probability mututation_probability]
```

**After**

```tcl
[-mutation_probability mutation_probability]
```

---

## VERIFICATION

Run:

```tcl
resynth_genetic -liberty_file my.lib -mutation_probability 0.3
```

Expected: the argument is recognized and the mutation probability value is applied.


src/rmp/src/rmp.tcl
```diff
@@ -158,7 +158,7 @@ sta::define_cmd_args "resynth_genetic" {
                                             [-slack_threshold slack_threshold]
                                             [-seed seed]
                                             [-population_size population_size]
-                                            [-mututation_probability mututation_probability]
+                                            [-mutation_probability mutation_probability]
```
eder-matheus, line 161: Nice fix! Could you also update the README.md file with the correct option names?

suhr25: Thanks for the suggestion! I've updated README.md to use the correct option names.

suhr25:
Hi @eder-matheus,
Thanks for the review.
I have made changes accordingly.
Please take a look.

eder-matheus:
@suhr25 Check the DCO error.

suhr25:
Hi @eder-matheus,
All checks are passing now.
Kindly take a look.
Thanks!

PR #9688, opened 2026-03-07
fix(gui): guard Gui::get() calls with enabled() checks

MeloveGupta:
This PR fixes multiple contract violations where gui::Gui::get() was invoked without verifying that the GUI subsystem was enabled. According to the GUI API contract, Gui::get() must only be called when gui::Gui::enabled() returns true. Violating this condition triggers runtime warnings such as:

[WARNING GUI-0150] Internal Error: gui::Gui::get() called when enabled() is false

These warnings commonly occurred during headless runs or in code paths where GUI components were referenced unconditionally.

Changes

Added gui::Gui::enabled() guards around calls to gui::Gui::get() across multiple modules.

Updated renderer registration, redraw operations, and GUI status updates to ensure they only execute when the GUI is active.

Prevented early GUI initialization paths from calling Gui::get() before the GUI subsystem is fully initialized.

Adjusted several graphics-related classes to avoid triggering warnings during construction or destruction in headless mode.

Ensured that non-GUI flows (e.g., routing, placement, PDN analysis) no longer implicitly depend on GUI availability.

Affected Areas

The fixes primarily impact:

GUI heatmap lifecycle and renderer management

Tool-specific graphics integrations (e.g., routing, placement, PDN visualization)

GUI initialization and layout setup

Renderer and widget destructors that previously triggered warnings in headless execution


MeloveGupta:
The latest commit [gui] Fix gui::Gui::get() called before enabled() violations (hash 960b8c5f37) includes:

HeatMap: pending registry to defer registration until GUI is enabled, destructor cleans pending list, and pending flush after MainWindow creation.
GUI: flush pending heatmaps in startGui after window setup.
DisplayControls: guard layer tooltips, renderer registration, state save/selection emits on Gui::enabled().
ClockWidget: guard descriptor lookups, renderer registration/hide events on Gui::enabled().
MainWindow: guard descriptor unregister and heatmap block assignment when GUI isn’t enabled.
Files touched:

src/gui/include/gui/heatMap.h
src/gui/src/heatMap.cpp
src/gui/src/gui.cpp
src/gui/src/displayControls.cpp
src/gui/src/clockWidget.cpp
src/gui/src/mainWindow.cpp

PR #9690, opened 2026-03-08
ram: simplification of word slice creation

braydenlouie:
Changes in RAM creation order
- Switched to building out each word slice out (row by row)
- For simplifying all the indices used in the current implementation for readability

### DIFF
--- a/src/ram/src/ram.cpp
+++ b/src/ram/src/ram.cpp
@@ -3,7 +3,6 @@
 
 #include "ram/ram.h"
 
-#include <algorithm>
 #include <array>
 #include <cmath>
 #include <fstream>


braydenlouie:
@rovinski Here are the code simplification changes. At the moment the ok files are still not lining up, even when I tested it with the cmake build so still need to look into that. The differences are related to two of the pins and I believe some routing. I don't know if that's an issue with the changes I made or if they are a product of any changes to the other tools. The rest of the source files should be fixed barring linting and formatting changes.

src/ram/src/layout.cpp
```diff
@@ -237,6 +237,19 @@ int Grid::numLayouts() const
...
 
+void Grid::setNumLayouts(int tracks)
+{
+  if (tracks > layouts_.size()) {
+    for (int size = layouts_.size(); size < tracks; ++size) {
+      if (orientation_ == odb::horizontal) {
+        layouts_.push_back(std::make_unique<Layout>(odb::vertical));
+      } else {
+        layouts_.push_back(std::make_unique<Layout>(odb::horizontal));
+      }
```
rovinski: Does this work?
```suggestion
      layouts_.push_back(std::make_unique<Layout>(orientation_.turn_90()));
```

src/ram/include/ram/layout.h
```diff
@@ -0,0 +1,117 @@
...
+  int cell_width_{0};
+  std::vector<std::unique_ptr<Cell>> cells_;
+};
+
+class Grid
+{
+ public:
+  Grid(odb::Orientation2D orientation);
+
+  Grid(odb::Orientation2D orientation, int tracks);
```
rovinski: Can we eliminate all use of the term "tracks"? A "track" has a specific meaning which is associated with metal tracks in global/detailed routing.

Should this be "rows" instead?

src/ram/include/ram/ram.h
rovinski: Are these declarations necessary? They should already exist by including `layout.h`.

src/ram/src/ram.cpp
rovinski, line 56: There is a mismatch here between the argument names here and in the header file (some args have underscores some don't). Please standardize on one.

src/ram/src/ram.cpp
```diff
@@ -137,16 +138,18 @@ std::unique_ptr<Cell> RamGen::makeCellBit(const std::string& prefix,
   return bit_cell;
 }
 
-void RamGen::makeCellByte(Grid& ram_grid,
-                          const int byte_idx,
-                          const std::string& prefix,
+void RamGen::makeCellByte(const int byte_idx,
```
rovinski: There should no longer be a concept of `byte`. All `byte`s should now be `slice`s.
Remember how we discussed that we want each slice to be able to contain an arbitrary number of bits.

e.g. `makeCellByte` --> `makeSlice` (no reason to indicate "Cell" here)
`byte_idx` --> `slice_idx`
etc.

src/ram/src/ram.cpp
```diff
@@ -775,53 +793,53 @@ void RamGen::generate(const int bytes_per_word,
...
-  ram_grid.gridInit();
+  ram_grid_.gridInit();
 
   auto db_libs = db_->getLibs().begin();
   auto db_sites = *(db_libs->getSites().begin());
   auto sites_width = db_sites->getWidth();
 
-  int num_sites = ram_grid.getRowWidth() / db_sites->getWidth();
+  int num_sites = ram_grid_.getRowWidth() / db_sites->getWidth();
   for (int i = 0; i <= word_count; ++i) {  // extra for the layer of buffers
```
rovinski: Please make this comment on a separate line and make it more explanatory. e.g.

`// One extra row is added at the top of the array for placing input buffers`

rovinski: Actually, better yet, add the following

```
// One extra row is added at the top of the array for placing input buffers
const int rows = word_count + 1;

for (int i = 0; i < rows; ++i) {
  //...
```

src/ram/src/ram.cpp
```diff
@@ -837,10 +855,10 @@ void RamGen::generate(const int bytes_per_word,
                   sites_width);
   }
 
-  ram_grid.placeGrid();
+  ram_grid_.placeGrid();
 
-  int max_y_coord = ram_grid.getHeight() * (word_count + 1);
-  int max_x_coord = ram_grid.getRowWidth();
+  int max_y_coord = ram_grid_.getHeight() * (word_count + 1);
```
rovinski: After the previous suggestion:
```suggestion
  int max_y_coord = ram_grid_.getHeight() * rows;
```

This makes it self-explanatory where the +1 comes from.

src/ram/include/ram/layout.h
rovinski, line 1: I'm confused about why the header file is copied here? Also how is this working - the old one is not deleted?

src/ram/src/ram.cpp
```diff
@@ -155,32 +158,19 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
+      outs[read_port] = data_output[read_port][local_bit]->getNet();
     }
-    ram_grid.addCell(makeCellBit(name,
-                                 read_ports,
-                                 gclock_net,
-                                 select_b_nets,
-                                 data_input[local_bit],
-                                 outs),
-                     physical_col_idx);
+    ram_grid_.addCell(makeCellBit(name,
```
rovinski: Any reason to name this `makeCellBit` and not just `makeBit`?

src/ram/include/ram/layout.h
braydenlouie: It's moved over here now that I have the grid as a member of the RamGen class, that way we're not constantly passing the ram_grid variable every single time, since we're only working with one grid anyway. The ram header file needs access to this file. Also, I'm considering renaming this to grid since the class that's being used in the cpp file is the grid. The old one should've been deleted.  I must've missed it, but it's been fixed.

src/ram/src/ram.cpp
braydenlouie: All of the make methods should be just converted to `make` as opposed to `makeCell`. It was just the naming scheme that I had when I was setting up the grid class initially.

src/ram/src/layout.cpp
```diff
@@ -149,9 +149,7 @@ Grid::Grid(odb::Orientation2D orientation, int tracks)
     : orientation_(orientation)
 {
   for (int i = 0; i < tracks; ++i) {
-    if (orientation_ == odb::horizontal) {
-      layouts_.push_back(std::make_unique<Layout>(odb::vertical));
-    }
+    layouts_.push_back(std::make_unique<Layout>(orientation_.turn_90()));
```
rovinski, line 152: Is this the desired behavior? Note that before it wasn't pushing a value if `orientation_ == odb::vertical`. Now it does.

src/ram/src/ram.cpp
```diff
@@ -105,48 +106,50 @@ dbBTerm* RamGen::makeBTerm(const std::string& name, dbIoType io_type)
...
-               {{storage_cell_->findMTerm("CLK") ? "CLK" : "GATE", clock},
-                {"D", data_input},
-                {"Q", storage_net}});
+  makeInst(bit_cell.get(),
+           prefix,
+           "bit",
+           storage_cell_,
+           {{storage_cell_->findMTerm("CLK") ? "CLK" : "GATE", clock},
+            {"D", data_input},
+            {"Q", storage_net}});
```
rovinski, line 125: Let's make a note to fix the hard-coded pin names in the next PR.

src/ram/src/ram.cpp
```diff
@@ -105,48 +106,50 @@ dbBTerm* RamGen::makeBTerm(const std::string& name, dbIoType io_type)
...
-void RamGen::makeCellByte(Grid& ram_grid,
-                          const int byte_idx,
-                          const std::string& prefix,
-                          const int read_ports,
-                          dbNet* clock,
-                          dbNet* write_enable,
-                          const vector<dbNet*>& selects,
-                          const array<dbNet*, 8>& data_input,
-                          const vector<array<dbBTerm*, 8>>& data_output)
+void RamGen::makeSlice(const int byte_idx,
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -105,48 +106,50 @@ dbBTerm* RamGen::makeBTerm(const std::string& name, dbIoType io_type)
...
+                       const int row_idx,
+                       const int read_ports,
+                       dbNet* clock,
+                       dbNet* write_enable,
+                       const vector<dbNet*>& selects,
+                       const vector<dbNet*>& data_input,
+                       const vector<vector<dbBTerm*>>& data_output)
 {
+  const int bits_per_byte = 8;
+  const int start_bit_idx = byte_idx * bits_per_byte;
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -155,61 +158,84 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
-    const int global_logical_bit_idx = logical_bit_base + local_bit;
-
-    // For placement
-    const int physical_col_idx = physical_col_base + local_bit;
-
-    auto name = fmt::format("{}.bit{}", prefix, global_logical_bit_idx);
-    vector<dbNet*> outs;
-    outs.reserve(read_ports);
+  for (int local_bit = 0; local_bit < bits_per_byte; ++local_bit) {
+    auto name = fmt::format("{}.bit{}", prefix, start_bit_idx + local_bit);
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -155,61 +158,84 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
-                                 data_input[local_bit],
-                                 outs),
-                     physical_col_idx);
+    ram_grid_.addCell(makeBit(name,
+                              read_ports,
+                              gclock_net,
+                              select_b_nets,
+                              data_input[local_bit],
+                              outs),
+                      start_bit_idx + local_bit + byte_idx);
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -155,61 +158,84 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
+    makeInst(sel_cell.get(),
+             prefix,
+             fmt::format("select_inv_{}", i),
+             inv_cell_,
+             {{"A", selects[i]}, {"Y", select_b_nets[i]}});
   }
 
-  ram_grid.addCell(std::move(sel_cell), (byte_idx * 9) + 8);
+  ram_grid_.addCell(std::move(sel_cell),
+                    start_bit_idx + bits_per_byte + byte_idx);
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -155,61 +158,84 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
+}
+
+void RamGen::makeWord(const int bytes_per_word,
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -155,61 +158,84 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
+{
+  const int bits_per_byte = 8;
+
+  for (int byte = 0; byte < bytes_per_word; ++byte) {
+    int start_idx = byte * bits_per_byte;
+
+    vector<dbNet*> byte_inputs(data_input.begin() + start_idx,
+                               data_input.begin() + start_idx + bits_per_byte);
+    vector<vector<dbBTerm*>> byte_outputs(read_ports,
+                                          vector<dbBTerm*>(bits_per_byte));
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -155,61 +158,84 @@ void RamGen::makeCellByte(Grid& ram_grid,
...
+    }
+
+    makeSlice(byte,
+              row_idx,
+              read_ports,
+              clock,
+              write_enable[byte]->getNet(),
+              selects,
+              byte_inputs,
+              byte_outputs);
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -593,7 +617,8 @@ void RamGen::generate(const int bytes_per_word,
                       dbMaster* tapcell,
                       int max_tap_dist)
 {
-  const int bits_per_word = bytes_per_word * 8;
+  const int bits_per_byte = 8;
+  const int bits_per_word = bytes_per_word * bits_per_byte;
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -692,73 +717,64 @@ void RamGen::generate(const int bytes_per_word,
...
-                   cell_name,
-                   read_ports,
-                   clock->getNet(),
-                   write_enable[col]->getNet(),
-                   word_decoder_nets[row],
-                   D_nets,
-                   q_outputs_);
-    }
+  for (int row = 0; row < word_count; ++row) {
+    makeWord(bytes_per_word,
```
rovinski: No byte

src/ram/src/ram.cpp
```diff
@@ -692,73 +717,64 @@ void RamGen::generate(const int bytes_per_word,
...
+             read_ports,
+             clock->getNet(),
+             write_enable,
+             word_decoder_nets[row],
+             D_nets,
+             q_outputs_);
+  }
 
-    for (int bit = 0; bit < 8; ++bit) {
+  for (int byte = 0; byte < bytes_per_word; ++byte) {
```
rovinski: No byte

src/ram/src/layout.cpp
braydenlouie: Yes, I think I never added a value for `orientation == odb::vertical` because I wasn't sure how accommodating this grid class should be. The current implementation stores the layouts as columns but keeping this could give us flexibility if we need to change to storing the layouts as rows (or if in the future someone has an additional use for this class)

src/ram/src/ram.cpp
braydenlouie: This was a change made in relation to the switch from using sky130's `dlxtp` to `dfxtp`for generating a behavioral model. I can change it so that it's just hard-coded to `CLK` for now until the next PR if that's preferable.

src/ram/src/ram.cpp
```diff
@@ -655,19 +653,19 @@ void RamGen::generate(const int bytes_per_word,
   // 9 columns for 8 bits per word plus
   // cell for WE AND gate/inverter
   // extra column is for decoder cells
-  int col_cell_count = bytes_per_word * 9;
+  int col_cell_count = slices_per_word * 9;
```
rovinski: Is this math and the comment above it still correct?

src/ram/src/ram.cpp
```diff
@@ -874,19 +873,19 @@ void RamGen::setBehavioralVerilogFilename(const std::string& filename)
...
+                                    const int slices_per_word,
+                                    const int num_words,
                                     const int read_ports)
 {
   if (filename.empty()) {
     return;
   }
 
-  const int word_size_bit = bytes_per_word * 8;
+  const int word_size_bit = slices_per_word * 8;
```
rovinski: Is this math still correct?

braydenlouie:
I'll update the unit testing as soon as I get the chance to rebuild on a `CMake` build

src/ram/src/ram.cpp
rovinski: What I meant was in a future PR, the dbSta interface should be used to find the clock port without needing knowing the name, such as in this PR: https://github.com/The-OpenROAD-Project/OpenROAD/pull/9680/changes#diff-98551916db74d39411a745c9ee3ff871d2b6473e09af929c8f78252dc85ddeafR119-R142

Although I'm not convinced that that PR does it in the simplest way. There's probably a `getClockPort()` function or something rather than iterating.

src/ram/include/ram/layout.h
```diff
@@ -98,6 +98,8 @@ class Grid
 
   int numLayouts() const;
 
+  void setNumLayouts(int tracks);
```
rovinski: `tracks` --> `rows` unless there's a better term for it

src/ram/src/layout.cpp
```diff
@@ -165,30 +163,26 @@ void Grid::addLayout(std::unique_ptr<Layout> layout)
...
     layouts_.push_back(std::move(layout));
-  } else if (index < layouts_.size()) {
-    layouts_.insert(layouts_.begin() + index, std::move(layout));
-  } else if (index > layouts_.size()) {
+  } else if (idx < layouts_.size()) {
+    layouts_.insert(layouts_.begin() + idx, std::move(layout));
+  } else if (idx > layouts_.size()) {
     return false;
   }
   return true;
```
rovinski: This can be simplified, because an insertion at the end is equivalent to a `push_back`
```suggestion
  if (idx > static_cast<int>(layouts_.size())) {
    return false;
  }
  layouts_.insert(layouts_.begin() + idx, std::move(layout));
  return true;
```

src/ram/src/layout.cpp
```diff
@@ -237,14 +231,23 @@ int Grid::numLayouts() const
...
 }
 
-int Grid::getLayoutWidth(int index) const
+void Grid::setNumLayouts(int num_layouts)
+{
+  if (num_layouts > layouts_.size()) {
+    for (int size = layouts_.size(); size < num_layouts; ++size) {
+      layouts_.push_back(std::make_unique<Layout>(orientation_.turn_90()));
+    }
+  }
```
rovinski: You don't need the if condition because the for loop will already exit on the first check if `num_layouts` > `layouts_.size()`
```suggestion
  for (int size = layouts_.size(); size < num_layouts; ++size) {
    layouts_.push_back(std::make_unique<Layout>(orientation_.turn_90()));
  }
```

src/ram/src/ram.tcl
```diff
@@ -12,19 +13,29 @@ sta::define_cmd_args "generate_ram_netlist" {-bytes_per_word bits
...
-    set word_count $keys(-word_count)
+  if { [info exists keys(-word_size)] } {
+    set word_size $keys(-word_size)
   } else {
-    utl::error RAM 2 "The -word_count argument must be specified."
+    utl::error RAM 2 "The -word_size argument must be specified."
+  }
+
+  if { $word_size % $mask_size != 0 } {
+    utl::error RAM 26 "The -word_size argument must be divisible by the -mask_size argument."
```
rovinski: For clarity
```suggestion
    utl::error RAM 26 "The -word_size ($word_size) must be divisible by the -mask_size ($mask_size)."
```

rovinski:
@gemini-code-assist /review

src/ram/src/ram.tcl
```diff
@@ -98,8 +110,9 @@ proc generate_ram { args } {
   }
 
   set ram_netlist_args [list \
-    -bytes_per_word $keys(-bytes_per_word) \
-    -word_count $keys(-word_count)]
+    -mask_size $keys(-mask_size) \
+    -word_size $keys(-mask_size) \
```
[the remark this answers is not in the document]
rovinski: @braydenlouie this one is important to fix

braydenlouie:
@rovinski @maliberty PR is ready for review. The latest build is unstable with the clang-tidy change (removing the include now that we don't need `<algorithm>`). The previous build was a success, so I'm unsure if it was just that one line removal that made it unstable or if it was something else.

rovinski (approved):
Looks like an unrelated bazel error. This LGTM.

src/ram/src/ram.tcl
```diff
@@ -1,8 +1,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024-2025, The OpenROAD Authors
 
-sta::define_cmd_args "generate_ram_netlist" {-bytes_per_word bits
-                                             -word_count words
+sta::define_cmd_args "generate_ram_netlist" {-mask_size bits
+                                             -word_size bits
+                                             -num_words words
```
maliberty, line 6: new arg should be documented in the readme

rovinski: @maliberty ram doesn't have a readme yet 🤷‍♂️
It's on the todo list #9392

PR #9692, opened 2026-03-08
bazel-orfs: bump

oharboe:
reduces accidental dependencies, such as on javascript.

@hzeller @maliberty reduce cognitive load...

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -1,5 +1,5 @@
 {
-  "lockFileVersion": 24,
+  "lockFileVersion": 13,
   "registryFileHashes": {
     "https://bcr.bazel.build/bazel_registry.json": "8a28e4aff06ee60aed2a8c281907fb8bcbf3b753c91fb5a5c57da3215d5b3497",
     "https://bcr.bazel.build/modules/abc/0.62-yosyshq/MODULE.bazel": "325231df11fd3480cb0eb3c9095105ccdfacf34583c3a92d8c0ae46967dbaef6",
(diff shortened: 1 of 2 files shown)


maliberty (approved):

MODULE.bazel.lock
```diff
@@ -1,5 +1,5 @@
 {
-  "lockFileVersion": 24,
```
hzeller, line 2: Did you run this with an old bazel < 8.5.0 ? Version number is going backwards.

oharboe: I'm unsure. I was working in Claude and I think that claude has a nix environment so it leaked in?

oharboe: I thought that the local bazel version didn't matter with bazelisk, perhaps I ran bazel directly?

Ideally bazel would produce an error if we use it wrong...

hzeller: it does complain, unless you specifically override the version with `USE_BAZEL_VERSION=x.y.z`

oharboe: I'm blind to a bazel wall of text and bizillion warnings and complaints, it would have to *fail* for me to notice...

hzeller: As I said, it _does_ fail.

oharboe: but it didn't fail on my machine as I can recall, nor in CI?

I don't know what is going on here.

hzeller: the CI will run with the bazel version in `.bazelversion`, so that works as intended. The *.lock file will be fixed locally when the CI runs it, but it will not fail.

You probably ran in your environment with `USE_BAZEL_VERSION=7.6.0` or something, that is why didn't complain (as you explicitly override the `.bazelversion`) and it then generated the lock file that way.

oharboe: I probably did, but I don't know how I did that and I don't understand why CI doesn't catch it.

hzeller: the CI is not configured to care if the *.lock file is up-to-date.

oharboe: helpful? https://github.com/The-OpenROAD-Project/OpenROAD/pull/9703

PR #9693, opened 2026-03-08
Add Python API for Resizer (rsz)

rohithsiddi:
## Summary

Adds a Python API for the Resizer (rsz) tool so that flows can use `rsz_aux` and the `rsz` module instead of Tcl for buffer_ports, repair_design, repair_timing, and related commands.

## Motivation

The Resizer had no Python bindings; Python flows had to call Tcl commands via `evalTclString`. This change brings rsz in line with other tools (e.g. grt, cts, drt, pdn) that expose a Python API.

## Changes

- **Build:** New `rsz_py` SWIG module in `src/rsz/src/CMakeLists.txt` using `Resizer-py.i`; linked in `src/CMakeLists.txt`; `rsz` added to Python module load list in `src/Main.cc`.
- **Bindings:** `Resizer-py.i` exposes snake_case wrappers for Resizer commands (opaque STA types; no Tcl-specific code).
- **Wrapper:** `src/rsz/test/rsz_aux.py` with full command coverage, docstrings, and correct unit conversions (microns↔metres, ns↔seconds, percent↔fraction).
- **Tests:** `helpers.py` and `gcd_resize.py` mirroring the existing Tcl test; run via CTest as `rsz.gcd_resize.py`.
- **Docs:** New "Using the Python interface to rsz" section in `src/rsz/README.md`.

## Testing

- `ctest -R "rsz.gcd_resize.py"` — pass
- `ctest -R "rsz"` — all RSZ tests (Tcl + Python) pass

AI tools were used to assist with code generation. All technical decisions and implementation details are my own.


src/rsz/src/Resizer-py.i
```diff
@@ -0,0 +1,266 @@
...
+  return getResizer()->utilization();
+}
+
+int find_floating_nets_count()
+{
+  ensureLinked();
+  auto* nets = getResizer()->findFloatingNets();
+  int count = nets ? static_cast<int>(nets->size()) : 0;
+  delete nets;
+  return count;
```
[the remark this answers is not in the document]
rohithsiddi: Fixed in the latest commit using std::unique_ptr as suggested

src/rsz/src/Resizer-py.i
```diff
@@ -0,0 +1,266 @@
...
+}
+
+int find_floating_pins_count()
+{
+  ensureLinked();
+  auto* pins = getResizer()->findFloatingPins();
+  int count = pins ? static_cast<int>(pins->size()) : 0;
+  delete pins;
+  return count;
```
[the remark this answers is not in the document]
rohithsiddi: Fixed in the latest commit using std::unique_ptr as suggested

src/rsz/src/Resizer-py.i
```diff
@@ -0,0 +1,266 @@
...
+}
+
+int find_overdriven_nets_count(bool include_parallel_driven)
+{
+  ensureLinked();
+  auto* nets = getResizer()->findOverdrivenNets(include_parallel_driven);
+  int count = nets ? static_cast<int>(nets->size()) : 0;
+  delete nets;
+  return count;
```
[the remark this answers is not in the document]
rohithsiddi: Fixed in the latest commit using std::unique_ptr as suggested

maliberty:
We should have a python API for rsz.  This PR however tries to impose an interface the mirrors tcl which is not desirable.  In Python we intend to expose an object-oriented API rather than a command oriented API.  Please look at other modules for examples.

rohithsiddi:
Thanks for the feedback — understood. I’ll pivot the architecture to follow the object-oriented Python API pattern used in the other modules instead of the Tcl-like command wrappers.

I'll study the preferred modules to align with your exact design patterns and push an updated structure to this branch shortly.

rohithsiddi:
@maliberty I've addressed the review comments. Summary of the changes below.
### What I updated

- Kept the RSZ Python interface object-oriented (`design.getResizer().<method>`), aligned with other tool bindings.
- Simplified `Resizer-py.i` style to match existing `*-py.i` files (minimal comments, no behavior change).
- Refined `rsz_aux.py` interface details:
  - removed unused `verbose` argument from `report_floating_nets`
  - clarified Python-facing error messages (removed Tcl-style flag prefixes)
- Updated `gcd_resize.py` call site accordingly (`report_floating_nets(design)`).

### Validation

- Rebuilt successfully.
- RSZ Python test passes:
  - `ctest --output-on-failure -R "rsz.gcd_resize.py"`

src/rsz/src/Resizer-py.i
```diff
@@ -0,0 +1,100 @@
...
+
+#include "rsz/Resizer.hh"
+#include "ord/OpenRoad.hh"
+%}
+
+%include "../../Exception-py.i"
+%include <std_string.i>
+%include <std_vector.i>
+
+%ignore rsz::Resizer::Resizer;
```
maliberty, line 17: Why are there so many %ignore ?

rohithsiddi: The higher %ignore count in Resizer-py.i compared to GRT/CTS is because
Resizer.hh is deeply integrated with OpenSTA internals:

- ~21 methods return/take STA collection types (NetSeq*, PinSet*, vector<MoveType>)
  that have no Python SWIG typemaps — these are either ignored or replaced with
  %extend wrappers (e.g. findFloatingNetsCount(), string-based repairSetup).
- ~5 methods use output-by-reference params (Delay&, Slew&, map<Pin*, float>&)
  not cleanly expressible in Python.
- Internal/GUI hooks (postReadLiberty, setDebugGraphics, lib_data_) follow
  the same pattern as GRT's init/initGui/setRenderer ignores.

GlobalRouter gets away with only 5 %ignores because its API mostly uses
odb::db* types already registered by the odb SWIG module.

maliberty: Look at odb where we have 
```
  void getGridPatternX(int i, int& origin_x, int& line_count, int& step);
```
and map it with 
```
%apply int& OUTPUT { int& origin_x, int& origin_y, int& line_count, int& step };
```

rohithsiddi: @maliberty Thanks for the feedback — I updated `Resizer-py.i` accordingly.

### `Resizer-py.i`: replaced `%ignore` with `%apply OUTPUT` where applicable

For `bufferWireDelay`, I removed `%ignore` and added:

```swig
%apply float& OUTPUT { sta::Delay& delay, sta::Slew& slew };
```

`sta::Delay` and `sta::Slew` are aliases of `float`, so this follows SWIG's standard `float& OUTPUT` mapping pattern (same style as in `odb/src/swig/common/dbtypes_common.i`). From Python, `bufferWireDelay(...)` now returns `(delay, slew)`.

I also removed `%ignore` from methods that SWIG can wrap directly without extra typemaps:

- `parseMove` (static, returns `MoveType`)
- `repairSetup(const sta::Pin*)` (single-pin testing overload)
- `resetInputSlews` (no args)
- `getSlewRCFactor` (returns `float`)

### Why the remaining `%ignore` entries stay

The remaining `%ignore` methods are cases where `%apply` is not applicable:

- Return STA container/internal types (`sta::NetSeq`, `sta::PinSet`, `sta::VertexSeq`, `std::optional<sta::Slack>`) that are not directly wrapped in this module
- Use C-style fixed-size output arrays (e.g. `ArcDelay[2]`) — `%apply` works on scalar `T&`, not `T[N]`
- Use `std::map<const Pin*, float>&` output params
- Require `sta::Scene*` inputs, which are internal STA context objects not available from Python
- Internal lifecycle/GUI hooks (`init`, constructor, `postReadLiberty`, `setDebugGraphics`), consistent with other tool bindings

For `findFloatingNets`, `findFloatingPins`, `findOverdrivenNets`, and `repairSetup(vector<MoveType>)`, `%ignore` is paired with `%extend` wrappers that expose Python-usable alternatives (count helpers and string-based sequence input).

rohithsiddi: @maliberty Thanks again for the guidance on the ```%apply OUTPUT``` pattern. That's been applied for ```bufferWireDelay``` in the latest push.

I also did a self-review pass before pinging and made a few additional improvements:

Re-added ```%ignore rsz::Resizer::parseMove``` — it was unnecessarily un-ignored since there's no ```vector<MoveType> ```typemap, making it a dead-end for Python callers
Added a guard in ```rsz_aux.repair_timing``` that warns when neither setup, hold, nor recover_power is selected
Documented the ```LibertyCell* ```/ ```LibertyPort*``` type resolution limitation in set_dont_use and repair_tie_fanout, following the same pattern as ```rmp_aux.py```

All RSZ tests (Tcl + Python) pass. Happy to make any further changes.

maliberty:
Please also make this work with the Bazel build / test

rohithsiddi:
> Please also make this work with the Bazel build / test

@maliberty Done — added python_wrap_cc, cc_binary, and py_library rules to src/rsz/BUILD following the existing gpl pattern, and registered //src/rsz:swig-py in the root BUILD.bazel openroad_swig-py deps.

src/rsz/test/rsz_aux.py
```diff
@@ -0,0 +1,261 @@
...
+        resizer.setDontUse(cell, True)
+
+
+def unset_dont_use(design, lib_cells):
+    resizer = design.getResizer()
+    for cell in lib_cells:
+        resizer.setDontUse(cell, False)
+
+
+def reset_dont_use(design):
```
maliberty: The goal is not to reproduce the TCL commands in Python but to provide an object-oriented interface in Python.  I expect users will call resizer methods directly rather than using wrappers.

rohithsiddi: Done, removed the trivial one-liner wrappers (`set_dont_use`, `reset_dont_use`, `remove_buffers`, `repair_clock_inverters`, etc.) from `rsz_aux.py`. The file now only contains functions that add genuine Python value: unit conversions (µm→m, ns→s, percent→fraction) and complex multi-parameter wrappers like `repair_timing`,  matching the pattern of `cts_aux.py`, `dpl_aux.py`, and `grt_aux.py`.

Users can call `design.getResizer().removeBuffers()`, `design.getResizer().resetDontUse()` etc. directly as intended.

The `%extend` block in `Resizer-py.i` adds Python-friendly alternatives only where needed, e.g. `findFloatingNetsCount()` because the original `findFloatingNets()` returns `sta::NetSeq*` which has no Python typemap.

The remaining `evalTclString` calls in `gcd_resize.py` (for `set_dont_use`, `repair_tie_fanout`, `read_sdc`, `estimate_parasitics`) are due to the STA Python API not yet exposing liberty cell/port lookup, same pattern as `rmp_aux.py`

maliberty:
Conflict to resolve and I don't see the new test being run in the CI.

rohithsiddi:
@maliberty Addressed both points:

1. **Conflict**: Merged upstream master and cleaned up the conflict resolution artifact :`#include "ord/OpenRoad.hh"` that was left in `Resizer-py.i` was unnecessary (no other tool's `-py.i` includes it); removed and added `using namespace rsz` for consistency with other tools.

2. **Test in CI**: Added `gcd_resize_py` as a native Bazel `py_test` target in `src/rsz/test/BUILD`, following the same pattern used in `src/cgt/test/BUILD` (#9907) and `src/gpl/test/BUILD`. Also added `helpers.if_bazel_change_working_dir_to()` for Bazel sandbox path resolution and guarded `utl.warn`/`utl.error`/`utl.report` (free functions only available in CMake builds) with `TEST_SRCDIR`-based fallbacks so the test runs correctly in both build systems.

All 192 CMake RSZ tests pass (`ctest -R rsz`), including `rsz.gcd_resize.py`. Bazel test also passes: `bazelisk test //src/rsz/test:gcd_resize_py`.

src/rsz/test/gcd_resize.py
maliberty, line 1: A test will typically diff is output versus a .ok file to ensure it has done what is expected.  Currently this test only shows that the tool didn't crash.  Please see other tests for examples.

rohithsiddi: Added output validation following the pattern used by all other tools (CGT, GPL, GRT, DPL, PDN, etc.): both the TCL and Python tests now write the post-RSZ DEF via `make_result_file()` and diff it against `gcd_resize.defok` using `diff_files()`. Also added `tech.thisown = False` in `make_design()` for Bazel memory safety (matching CGT/GPL/GRT/DPL helpers).

Both CMake (`ctest -R rsz.gcd_resize`) and Bazel (`bazelisk test //src/rsz/test:gcd_resize_py`) pass.

PR #9694, opened 2026-03-08
fix :malformed utl::warn call in detailed_route

suhr25:
## SUMMARY

This PR fixes a malformed `utl::warn` call in `detailed_route` that caused a Tcl argument error when `-droute_end_iter` was set above `64`. Instead of stopping with an error, the command now correctly prints a warning and clamps the value.

---

## FIX

* The `utl::warn` call was missing the required module and code arguments.
* Added `DRT` and warning code `518` to match the logger API.

**Before**

```tcl
utl::warn "-droute_end_iter cannot be greater than 64. Setting -droute_end_iter to 64."
```

**After**

```tcl
utl::warn DRT 518 "-droute_end_iter cannot be greater than 64. Setting -droute_end_iter to 64."
```

---

## VERIFICATION

Run:

```tcl
detailed_route -droute_end_iter 100
```

Before: Tcl error due to wrong `utl::warn` usage.
After: Warning is printed, value is clamped to `64`, and routing continues.

### DIFF
--- a/src/drt/src/TritonRoute.tcl
+++ b/src/drt/src/TritonRoute.tcl
@@ -89,7 +89,9 @@ proc detailed_route { args } {
   if { [info exists keys(-droute_end_iter)] } {
     sta::check_positive_integer "-droute_end_iter" $keys(-droute_end_iter)
     if { $keys(-droute_end_iter) > $max_droute_end_iter } {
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @eder-matheus @maliberty,
Thanks!

Issue #9626, opened 2026-03-04
ODB: add mapping from dbBTerm to related dbChipBump


osamahammad21:
### Description

Expected dbBTerm::getChipBump to be added to be able to get the dbChipBump from the bterm.

PR #9697, opened 2026-03-09
odb: create dbBTerm::getChipBump

ahmed532:
To be able to get the dbChipBump from the bterm

Closes: #9626


ahmed532:
Sure, I'll add tests.

PR #9703, opened 2026-03-10
Bazel checks

oharboe:


hzeller:
too complicated. Just invoke bazel with `--lockfile_mode=error` in the CI.

oharboe:
> too complicated. Just invoke bazel with `--lockfile_mode=error` in the CI.

Force pushed to get rid of the cruft.

Thanks!

oharboe:
@hzeller Heads up! I added you as a co-author, because I just did what you said.

.bazelrc
```diff
@@ -1,4 +1,5 @@
 common --enable_bzlmod
+common --lockfile_mode=error
```
hzeller: I wouldn't do it here. This is only something you might want to run in the CI.

For local compilation this is more annoying.

So in the CI wherenever you do bazelisk test src/... or something just add the option.

PR #9710, opened 2026-03-10
odb: add versions to make database versions consistent

gadfort:
No-op but needed to allow version 127 databases to be parsed

### DIFF
--- a/src/odb/src/db/dbDatabase.h
+++ b/src/odb/src/db/dbDatabase.h
@@ -50,7 +50,13 @@ namespace odb {
 inline constexpr uint32_t kSchemaMajor = 0;  // Not used...
 inline constexpr uint32_t kSchemaInitial = 57;
 
-inline constexpr uint32_t kSchemaMinor = 126;  // Current revision number
+inline constexpr uint32_t kSchemaMinor = 128;  // Current revision number
+
+// Revision where the ignore_usage_ flag was removed from dbGuide
+inline constexpr uint32_t kSchemaIgnoreUsageRemoved = 128;
+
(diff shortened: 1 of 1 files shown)


maliberty:
@eder-matheus FYI

src/odb/src/db/dbDatabase.h
```diff
@@ -50,7 +50,13 @@ namespace odb {
...
 inline constexpr uint32_t kSchemaInitial = 57;
 
-inline constexpr uint32_t kSchemaMinor = 126;  // Current revision number
+inline constexpr uint32_t kSchemaMinor = 128;  // Current revision number
+
+// Revision where the ignore_usage_ flag was removed from dbGuide
+inline constexpr uint32_t kSchemaIgnoreUsageRemoved = 128;
+
+// Revision where the ignore_usage_ flag was added to dbGuide
+inline constexpr uint32_t kSchemaIgnoreUsage = 127;
```
eder-matheus, line 59: This `kSchemaIgnoreUsage` wasn't intended to be in master branch. I thought it was removed from my grt guides restore PR before the merge.

gadfort:
This is moot with: https://github.com/The-OpenROAD-Project/OpenROAD/pull/9697
Since 127 is now used, again, any DB with version 127 will fail to parse if it was created from the previous version.

src/odb/src/db/dbDatabase.h
gadfort: @eder-matheus The issue here was that in your PR you dropped the DB version from 127 to 126. Once the version number if on master, it cannot be changed down otherwise it breaks any DBs created with it.

PR #9714, opened 2026-03-10
odb: allow dbWire::getShape usage in Tcl and add dbShape::isSegment

AcKoucher:
For https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/3969.

Adds SWIG implementation for using `void dbWire::getShape(int shape_id, dbShape& shape)` in Tcl. Usage example:
```
foreach rseg [$db_net getRSegs] {
  set shape [$wire getShape [$rseg getShapeId]]
}
```

### DIFF
--- a/src/odb/include/odb/dbShape.h
+++ b/src/odb/include/odb/dbShape.h
@@ -178,6 +178,11 @@ class dbShape
   bool isVia() const;
 
   ///
+  /// Returns true if this object is a segment
+  ///
+  bool isSegment() const;
+
+  ///
   /// Returns true if this object is a via box
   ///
   bool isViaBox() const;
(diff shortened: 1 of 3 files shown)


AcKoucher:
The check failure seems to be because of the PDN non-determinism.
```
//src/pdn/test:pads_black_parrot_flipchip_connect_overpads-tcl_test      FAILED in 1.4s

eder-matheus:
> The check failure seems to be because of the PDN non-determinism.
> 
> ```
> //src/pdn/test:pads_black_parrot_flipchip_connect_overpads-tcl_test      FAILED in 1.4s
> ```

I've restarted the build, hope it passes now.

maliberty:
For a simple class like dbShape it would also be fine to have the method just return the object and rely on copy-elision rather than use a reference arg.

AcKoucher:
> For a simple class like dbShape it would also be fine to have the method just return the object and rely on copy-elision rather than use a reference arg.

I see. My first idea was to use the same approach as other objects already use, but it felt somewhat more correct to use the reference agr. If you want me to change it for concision with the others I can do it.

maliberty:
This is good enough for now

PR #9725, opened 2026-03-11
dpl: fix incremental mode

gudeh:
This is a no-op since DPL incremental mode is currently disabled by default.

This fix ensures that all DRCs checked by DPL are considered (cell flipping, one-site-gap, padding, etc.). Previously, the incremental mode only checked row alignment and overlaps, incorrectly marking cells as legalized even when they were violating other DRCs.


gudeh:
I am closing this since we should eventually replace diamond search for the negotiation legalizer.

The current incremental version of dpl with diamond search does not work properly.

PR #9737, opened 2026-03-12
bazel: replace genrule sed/cat with skylib rules for hermeticity

oharboe:
Use expand_template and copy_file from bazel_skylib instead of genrule with sed/cat in the OpenMP overlay BUILD file. This avoids depending on host shell tools and sets a better example for contributors who copy existing patterns.

Remaining hermeticity nits for later:
- BUILD.bazel OpenRoadVersion genrule uses grep/cut/printf
- src/sta StaConfig genrule uses echo -e (fix pending in sta repo)

### DIFF
--- a/bazel/openmp/bundled.BUILD.bazel
+++ b/bazel/openmp/bundled.BUILD.bazel
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


oharboe:
@hzeller FYI

maliberty:
@codex review

hzeller:
Nice!

PR #9744, opened 2026-03-12
bazel: add Qt corelib bootstrap headers for hermetic builds

oharboe:
@hzeller Thoughts?

Qt public headers in build/include/QtCore/ internally include bootstrap headers from src/corelib/ subdirectories. Under strict Bazel sandboxing, these transitive includes are undeclared and cause build failures. Add a patch to the qt-bazel git_override that explicitly declares the corelib bootstrap headers in the QtCore target's hdrs.

Test: bazelisk query on @qt-bazel//qt_source:bootstrap hdrs

Before (without patch): 2 src/corelib headers declared
  src/corelib/text/qlocale_data_p.h
  src/corelib/time/qromancalendar_data_p.h

After (with patch): 383 src/corelib headers declared across
  src/corelib/global/*.h
  src/corelib/io/*.h
  src/corelib/kernel/*.h
  src/corelib/plugin/*.h
  src/corelib/serialization/*.h
  src/corelib/text/*.h
  src/corelib/time/*.h
  src/corelib/tools/*.h

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -87,6 +87,8 @@ bazel_dep(name = "qt-bazel")
 git_override(
     module_name = "qt-bazel",
     commit = "df022f4ebaa4130713692fffd2f519d49e9d0b97",
+    patch_strip = 1,
+    patches = ["//bazel:qt_bazel_bootstrap_headers.patch"],
     remote = "https://github.com/The-OpenROAD-Project/qt_bazel_prebuilts",
 )
 
--- a/bazel/qt_bazel_bootstrap_headers.patch
+++ b/bazel/qt_bazel_bootstrap_headers.patch
@@ -1,0 +1,17 @@
+--- a/qt_source/BUILD.bazel
++++ b/qt_source/BUILD.bazel
+@@ -776,6 +776,14 @@
+     hdrs = glob([
+         QT_BASE + "/build/include/QtCore/**/Q*",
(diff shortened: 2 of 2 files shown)


oharboe:
@hzeller Claude told me this had to be fixed :smiley_cat: 

1. Hallucinations?
2. Accept this PR to show that it works, is necessary and to board up the broken window
3. Fix upstream
4. Upgrade and fix the broken window

?

QuantamHD:
Is this actually breaking something? If so what's the error?

And this should most certainly be upstreamed.

oharboe:
> Is this actually breaking something? If so what's the error?
> 
> And this should most certainly be upstreamed.

I was trying to get numbers on https://github.com/The-OpenROAD-Project/OpenROAD/pull/9738 when Claude said that this had to be fixed.

Claude was doing a lot of `bazel clean --expunge`, so an untypical use case with A/B fresh builds.

That's all I know.

hzeller:
I think first establishing if there is a problem is a good step.

oharboe:
> I think first establishing if there is a problem is a good step.

We have this pull request.

Close it as an "observe and report" PR that we can pull out if needed?

oharboe:
@maliberty I'm closing this, you know where to find it if you hear about the problem from someone else. If I am able to reproduce it, I will let you know.

PR #9747, opened 2026-03-13
fix: rdl_route ignoring nets due to missing lappend target variable

suhr25:
## SUMMARY

Fixes a bug in `rdl_route` where nets passed by name were ignored.
The issue was a missing variable in an `lappend` call in **`src/pad/src/pad.tcl`**, which left the `nets` list empty.

---

## FIX

**Before**

```tcl
lappend [sta::sta_to_db_net $net]
```

**After**

```tcl
lappend nets [sta::sta_to_db_net $net]
```

Adds the missing `nets` variable so parsed nets are correctly stored.

---

## VERIFICATION

Before the fix, running rdl_route -layer M9 VSS VDD resulted in the error "No nets found to route", even when valid net names were provided.
After the fix, the same command correctly detects the nets and routing proceeds normally.

### DIFF
--- a/src/pad/src/pad.tcl
+++ b/src/pad/src/pad.tcl
@@ -410,7 +410,7 @@ proc rdl_route { args } {
   }
 
   sta::parse_port_net_args $args sta_ports sta_nets
-  set nets [lmap net $sta_nets {sta::sta_to_db_net $net}]
+  set nets [lmap net $sta_nets { sta::sta_to_db_net $net }]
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @eder-matheus @maliberty,
Thanks!

gadfort (approved):
LGTM, Thanks for catching this

suhr25:
Thanks for the merge!

PR #9751, opened 2026-03-13
Write gds

oharboe:
## Motivation: prune KLayout from the default build path

KLayout is currently required on the default `make` path in ORFS solely
for DEF-to-GDS merging (`def2stream.py`). This creates several problems:

- **Deployment burden.** KLayout is a large GUI application (~500MB+
  installed) with Qt, Ruby, and Python dependencies. Every CI runner,
  Docker image, and developer workstation must carry it even if the user
  only needs a GDS file.
- **Testing friction.** KLayout version mismatches and missing Python
  modules are a recurring source of CI failures that have nothing to do
  with the design flow itself.
- **Local development pain.** New contributors hit KLayout install issues
  before they can run their first `make` to completion.

KLayout remains a valuable *user* tool for DRC, LVS, and interactive GDS
viewing. Those uses are already opt-in (`make drc`, `make lvs`, `make
gui`). This PR does not remove them — it only removes the one mandatory
use on the default path.

## What this PR adds to OpenROAD

A native `write_gds` TCL command that replaces `klayout -b -r
def2stream.py`:

```tcl
write_gds [-layer_map file] [-merge gds_list] [-seal gds] \
          [-allow_empty regex] output.gds
```

The implementation lives in ODB alongside the existing GDS reader/writer:

| Component | What it does |
|-----------|-------------|
| `LayerMap` | Parses EDI `.layermap` and KLayout `.lyt` layer mapping formats |
| `DefToGds::convert` | Walks `dbBlock` geometry (wires, pins, fills, obstructions, instances) and emits GDS boundaries and SREFs |
| `DefToGds::mergeCells` | Reads macro GDS files and replaces placeholder cell structures |
| `DefToGds::mergeSeal` | Merges a seal ring GDS as a child of the top cell |
| `DefToGds::validate` | Flags empty cells and orphan cells (with regex bypass for known-empty cells) |

The command options mirror `def2stream.py` arguments so the ORFS
Makefile switch is a one-line change per target.

## Test coverage

All tests are small, fast Bazel `cc_test` targets (~0.1s each):

| Test | Coverage |
|------|----------|
| `TestGDSRead` | Read sky130 GDS, validate structure/element counts |
| `TestGDSWriteRoundtrip` | Write then re-read, verify lossless roundtrip |
| `TestGDSCreate` | Programmatic GDS creation with boxes, SREFs, transforms |
| `TestGDSLayerMap` | EDI and KLayout `.lyt` format parsing |
| `TestGDSMergeCells` | Cell merge from source GDS into target library |
| `TestGDSValidate` | Empty cell detection, `allow_empty` regex, orphan detection |
| `write_gds` (TCL) | End-to-end: load LEF+DEF, write GDS, verify success |

The first three tests are split from the existing `TestGDSIn.cpp`
(identical assertions, just separate binaries for parallel execution and
clear failure attribution).

## Planned ORFS changes (after this merges)

Once `write_gds` is on master, a follow-up ORFS PR will:

1. **Switch `do-gds-merged`** from `klayout -b -r def2stream.py` to an
   OpenROAD TCL script calling `write_gds`.
2. **Switch wrapped-macro GDS targets** similarly.
3. **Remove** `flow/util/def2stream.py`.
4. **Remove** KLayout variables from `flow/scripts/variables.mk`
   (`KLAYOUT`, `KLAYOUT_CMD`, `STREAM_SYSTEM_DEF2STREAM`,
   `STREAM_SYSTEM_DEF2STREAM_ARGS`).
5. **Keep** the opt-in `do-klayout`, `do-drc`, `do-lvs` targets
   unchanged — users who want KLayout install it themselves.

Net effect: `make` no longer requires KLayout. CI images shrink. New
contributors can complete a full flow build with only OpenROAD installed.

### DIFF
--- a/src/odb/src/gdsout/def2gds.cpp
+++ b/src/odb/src/gdsout/def2gds.cpp
@@ -4,18 +4,24 @@
 #include "odb/def2gds.h"
 
 #include <algorithm>
+#include <cctype>
 #include <cmath>
+#include <cstddef>
 #include <cstdint>
 #include <filesystem>
 #include <fstream>
+#include <iterator>
 #include <regex>
 #include <set>
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <vector>
 
 #include "odb/db.h"
 #include "odb/dbShape.h"
+#include "odb/dbTypes.h"
+#include "odb/geom.h"
 #include "odb/gdsin.h"
 #include "utl/Logger.h"
 
@@ -532,7 +538,7 @@ void DefToGds::mergeCells(dbGDSLib* lib,
     dbGDSLib* source_lib = reader.read_gds(gds_file, temp_db);
 
     if (source_lib == nullptr) {
-      logger_->warn(utl::ODB, 500, "Failed to read GDS file: {}", gds_file);
+      logger_->warn(utl::ODB, 546, "Failed to read GDS file: {}", gds_file);
       dbDatabase::destroy(temp_db);
       continue;
     }
@@ -548,7 +554,7 @@ void DefToGds::mergeCells(dbGDSLib* lib,
 
     if (source_dbu_per_um != target_dbu_per_um) {
       logger_->info(utl::ODB,
-                    504,
+                    547,
                     "Scaling merged GDS '{}' from {} to {} DBU/um",
                     gds_file,
                     source_dbu_per_um,
@@ -675,7 +681,7 @@ void DefToGds::mergeSeal(dbGDSLib* lib,
   // referenced by any SREF/AREF within the seal's own structures.
   std::set<std::string> seal_referenced;
   for (dbGDSStructure* str : lib->getGDSStructures()) {
-    if (pre_merge.count(str->getName())) {
+    if (pre_merge.contains(str->getName())) {
       continue;  // skip pre-existing structures
     }
     for (dbGDSSRef* sref : str->getGDSSRefs()) {
@@ -687,10 +693,10 @@ void DefToGds::mergeSeal(dbGDSLib* lib,
   }
 
   for (dbGDSStructure* str : lib->getGDSStructures()) {
-    if (pre_merge.count(str->getName())) {
+    if (pre_merge.contains(str->getName())) {
       continue;  // skip pre-existing structures
     }
-    if (seal_referenced.count(str->getName())) {
+    if (seal_referenced.contains(str->getName())) {
       continue;  // referenced by another seal cell, not a top cell
     }
     if (str == top_str) {
@@ -700,7 +706,7 @@ void DefToGds::mergeSeal(dbGDSLib* lib,
     sref->setOrigin(Point(0, 0));
     sref->setTransform(dbGDSSTrans(false, 1.0, 0.0));
     logger_->info(utl::ODB,
-                  501,
+                  548,
                   "Merging '{}' as child of '{}'",
                   str->getName(),
                   top_cell_name);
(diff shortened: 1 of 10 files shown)


oharboe:
@hzeller FYI

rovinski:
> Parses EDI `.layermap`

It should be double-checked that there are no copyright/legal concerns. In the past, scripts and other tools which read/write proprietary formats have been DMCA'd. I don't know if that's still the case.

> `DefToGds::mergeSeal`

How does this differ from just a generic GDS merge? Why do seal rings need a special function?

> Remove `flow/util/def2stream.py`.

I hope the intention is to remove it from the default flow but not remove it from the repository. KLayout's GDS generation is substantially more tested than OpenROAD's, and it would still be useful for validation and as a backup option.

gadfort:
The database is missing the foreign field and def2stream doesn't honor it correctly either.

stefanottili:
I glanced at the code, but didn't spot where duplicate cell names (same name in different files) are handled.

a) they have to be uniquified or
b) in a rare use case, the elements of the struct can be added (merged) to the previous struct of the same name.

The latter can be used to "add a sref for the sealring to toplevel" functionality.
Or add some blockage layer rectangles or IP-Tag text.
In klayout's strm2gds/oas this can be done with the "blend-mode 0" and +<filename> functionality.

src/odb/src/gdsout/def2gds.cpp
```diff
@@ -0,0 +1,678 @@
...
+    cell_str = dbGDSStructure::create(lib, cell_name.c_str());
+  }
+
+  dbGDSSRef* sref = dbGDSSRef::create(top_str, cell_str);
+
+  // Get the instance transform
+  dbTransform transform = inst->getTransform();
+  Point origin = transform.getOffset();
+
+  sref->setOrigin(origin);
```
gadfort, line 324: This really should honor the FORIGN tag in the LEF and be tracked so the later replacement is placed correctly otherwise the GDS is invalid

stefanottili:
@rovinski
>>Parses EDI .layermap

>It should be double-checked that there are no copyright/legal concerns. In the past, scripts and other tools which read/write proprietary formats have been DMCA'd. I don't know if that's still the case.

Please note the irony here ...

>The GDSII format was never officially "made open" in the sense of being released under an open-source license or handed over to a public standards body. It remains a proprietary format owned by Cadence Design.
However, it followed a unique path to becoming a "de facto" open standard: 
...
Because the specification is widely documented and hasn't changed in decades, it is treated as "open" by the industry.

.. and treat the def2gds "layer-name purpose layer-nr datatype-nr" map file the same way as you would gds, "as a de facto standard".

rovinski:
> Please note the irony here ...

And yet note that all of the major commercial providers have different ways of specifying the same simple mapping information. Companies, as a matter of fact, have previously DMCA'd scripts which were posted due to their intent to convert proprietary formats explicitly due to their intent to circumvent using the tool that the proprietary formats were designed for. The case I am thinking of was _specifically_ an EDA company taking down a GDS layermap conversion script.

Is it incredibly pedantic? Yes. Does the project wish to run the risk? I wouldn't think so. In fact I'm surprised that the security scan isn't tripping because I'm pretty sure it used to blacklist the word "innovus".

maliberty:
The write_gds  in OR was a GSoC project and it not complete and can't be used to replace klayout for gds generation.

oharboe:
> The write_gds in OR was a GSoC project and it not complete and can't be used to replace klayout for gds generation.

Can't in what sense, can you clarify? Shouldn't? Some conflict? Not technically possible?

oharboe:
An interesting experiment in automated Claude programming. Leave be for now.

PR #9756, opened 2026-03-14
gui: Add selection filter toggle to heatmaps (#6554)

### LINKED ISSUE #6554: Add support for limiting heatmaps to selection
### Description

Estimated/or actual routing heatmap includes all nets:

Image

However, it would be nice if one could select nets by name(ctrl-f) to look at specific nets. Similarly for placement heatmap.

Image

### Suggested Solution

Add filtering option(toggle?) for heatmaps to use selected items only.

### Additional Context

_No response_

shubhika123:
Add 'Only use selected items' checkbox to the heatmap setup dialog. When enabled, Placement Density, Pin Density, and Power Density heatmaps only include contributions from currently selected instances.

Fixes #6554

### DIFF
--- a/src/gui/include/gui/heatMap.h
+++ b/src/gui/include/gui/heatMap.h
@@ -201,7 +201,8 @@ class HeatMapDataSource
                               double new_data,
                               double data_area,
                               double intersection_area,
-                              double rect_area) = 0;
+                              double rect_area)
+      = 0;
   virtual void correctMapScale(Map& map) {}
   void updateMapColors();
   void assignMapColors();


src/gui/src/heatMap.cpp
```diff
@@ -1121,12 +1125,26 @@ bool PowerDensityDataSource::populateMap()
...
+  // Collect selected instances if filter is enabled
+  const bool filter = getUseSelectedOnly() && gui::Gui::enabled();
+  std::set<odb::dbInst*> selected_insts;
+  if (filter) {
+    for (const gui::Selected& item : gui::Gui::get()->selection()) {
+      if (item.isInst()) {
+        selected_insts.insert(std::any_cast<odb::dbInst*>(item.getObject()));
+      }
+    }
+  }
```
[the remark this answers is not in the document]
shubhika123: "Thanks! This has already been implemented in the latest commit :-

getSelectedInsts()
 is now a protected helper in 

HeatMapDataSource
 that all three data sources call."

Divinesoumyadip:
Hey @shubhika123 , noticed Gemini flagged the duplicated filtering logic , happy to help extract it into a base class helper if you'd like a hand.

src/gui/src/heatMap.cpp
```diff
@@ -287,7 +289,8 @@ Renderer::Settings HeatMapDataSource::getSettings() const
                               {"LogScale", log_scale_},
                               {"ReverseLog", reverse_log_},
                               {"ShowNumbers", show_numbers_},
-                              {"ShowLegend", show_legend_}};
+                              {"ShowLegend", show_legend_},
+                              {"UseSelectedOnly", use_selected_only_}};
```
gadfort, line 293: I think it would be better to provide a way to only add this option if the map specifically supports it.
And for those maps where it's insts indicate that is instances (or nets, etc) to make it clear

shubhika123: Thanks for the feedback @gadfort! I'll add a virtual supportsSelectionFilter() method to the base class so the checkbox only shows for those heatmaps that implement instance-based filtering, and update the label to be specific (e.g. 'Only use selected instances'). Will push shortly.

shubhika123: Fixed! Added getSelectionFilterLabel() as a virtual method which returns an empty string (no checkbox shown) by default, and supported heatmaps override it with a specific label like "Only use selected instances". Let me know if this looks good!

src/gui/include/gui/heatMap.h
```diff
@@ -71,6 +71,21 @@ class HeatMapDataSource
                     const std::string& settings_group = "");
   virtual ~HeatMapDataSource();
 
+  bool getUseSelectedOnly() const { return use_selected_only_; }
```
AcKoucher: Just a nit. The "get" naming logic looks a bit odd here as `use_selected_only_` is more of a state. Just `useSelectedOnly` feels more idiomatic.

shubhika123: Good catch! Renamed to `useSelectedOnly()` 👍

Divinesoumyadip:
Regarding the code duplication warning the bot flagged earlier for the filtering logic, the easiest way to clear that out is to pull the selection check into a protected helper method in the HeatMap base class.

You can add something like this in` heatMap.h`:


  ```
bool passesSelectionFilter(odb::dbInst* inst) const {
      if (!use_selected_only_) return true;
      // Add your GUI selection check here
      return gui::Gui::get()->isSelected(inst); 
  }
```
Then, in all three of your derived density classes, you can just call if (`!passesSelectionFilter(inst`)) continue; inside your loops instead of rewriting the logic.

Divinesoumyadip:
Also allow the basic **workflow** from the github setting for rest chceking.( from developer setting choose  classic one ).

shubhika123:
Thanks @Divinesoumyadip! The refactoring is already done via `getSelectedInsts()` - it pulls the selection into a `std::set` once in the base class before the loop, then the derived classes do a `set::find()` per instance. Same idea, but avoids repeated GUI calls.

maliberty:
@gadfort have your concerns been addressed?

gadfort (approved):
LGTM

PR #9759, opened 2026-03-14
Refactor CUGR to use FastRouteCore queue-and-execute model

sparsh-karna:
- Remove internal dirty net state tracking from CUGR
- Introduce nets_to_route_ queue populated via addDirtyNet and updateNet
- Implement routeIncremental() to deduplicate, route, and propagate overflow
- Wire CUGR execution into GlobalRouter::endIncremental

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -3345,9 +3345,10 @@ void GlobalRouter::checkPinPlacement()
           logger_->warn(
               GRT,
               31,
-              "At least 2 pins in position ({}, {}), layer {}, port {}.",
-              pos.x(),
-              pos.y(),
+              "At least 2 pins in position ({:.2f}um, {:.2f}um), layer {}, "
+              "port {}.",
+              dbuToMicrons(pos.x()),
+              dbuToMicrons(pos.y()),
               tech_layer->getName(),
               port->getName().c_str());
           invalid = true;
--- a/.bazelrc
+++ b/.bazelrc
@@ -43,11 +43,11 @@ build --copt "-Wextra"             --host_copt "-Wextra"
 # ... and disable the warnings we're not interested in.
 build --copt "-Wno-sign-compare"          --host_copt "-Wno-sign-compare"
 build --copt "-Wno-unused-parameter"      --host_copt "-Wno-unused-parameter"
-build --copt "-Wno-c++20-designator"      --host_copt "-Wno-c++20-designator"
 build --copt "-Wno-gcc-compat"            --host_copt "-Wno-gcc-compat"
 build --copt "-Wno-nullability-extension" --host_copt "-Wno-nullability-extension"
 build --copt "-Wno-deprecated-declarations" --host_copt "-Wno-deprecated-declarations"
 build --copt "-Wno-unused-local-typedef" --host_copt "-Wno-unused-local-typedef"
+build --cxxopt "-Wno-c++20-designator"   --host_cxxopt "-Wno-c++20-designator"
 
 # The warning acceptance bar in CI for PRs is set by -Werror of a specific
 # version of some chosen compiler. Disable warnings from other compilers until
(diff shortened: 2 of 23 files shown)


src/grt/src/cugr/src/CUGR.cpp
```diff
@@ -661,64 +659,48 @@ void CUGR::updateNet(odb::dbNet* db_net)
...
-  rerouteNets(nets_to_route);
+  rerouteNets(nets_to_route_);
 
   // Check for secondary overflow caused by rerouting
   std::vector<int> overflow_nets;
   updateOverflowNets(overflow_nets);
   std::vector<int> secondary_nets;
   std::ranges::set_difference(
-      overflow_nets, dirty_net_indices_, std::back_inserter(secondary_nets));
+      overflow_nets, nets_to_route_, std::back_inserter(secondary_nets));
```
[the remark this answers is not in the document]
sparsh-karna: I've updated it to pass a copy (`nets_to_reroute`) so the original state is preserved for the secondary overflow check. Fix pushed!

src/grt/src/cugr/src/CUGR.cpp
```diff
@@ -661,64 +659,48 @@ void CUGR::updateNet(odb::dbNet* db_net)
...
-  for (const int idx : dirty_net_indices_) {
-    odb::dbNet* db_net = gr_nets_[idx]->getDbNet();
-    if (updated_nets_.find(db_net) == updated_nets_.end()) {
-      updateNet(db_net);
-    }
-  }
-  updated_nets_.clear();
+  std::ranges::sort(nets_to_route_);
+  nets_to_route_.erase(
+      std::unique(nets_to_route_.begin(), nets_to_route_.end()),
```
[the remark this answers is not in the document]
sparsh-karna: Updated to use `std::ranges::unique` to comply with the `modernize-use-ranges` standard. (Note: I unpacked the subrange return manually since the auto-suggestion had a trailing comma glitch). Fix pushed!

sparsh-karna:
Hey @eder-matheus could you review the pr, i have made all the changes.

src/grt/src/cugr/src/CUGR.cpp
```diff
@@ -248,9 +248,17 @@ void CUGR::mazeRoute(std::vector<int>& net_indices)
 void CUGR::route()
 {
   std::vector<int> net_indices;
-  net_indices.reserve(gr_nets_.size());
-  for (const auto& net : gr_nets_) {
-    net_indices.push_back(net->getIndex());
+  if (!nets_to_route_.empty()) {
+    std::ranges::sort(nets_to_route_);
+    auto [first, last] = std::ranges::unique(nets_to_route_);
+    nets_to_route_.erase(first, last);
```
eder-matheus: Why is it needed?  In what scenario would we have duplicated net ids?

src/grt/src/GlobalRouter.cpp
```diff
@@ -337,8 +337,14 @@ void GlobalRouter::startIncremental()
 void GlobalRouter::endIncremental(bool save_guides)
 {
   is_incremental_ = true;
-  fastroute_->setResistanceAware(resistance_aware_);
-  updateDirtyRoutes();
+  if (use_cugr_) {
+    cugr_->routeIncremental();
+    routes_ = cugr_->getRoutes();
+    updatePinAccessPoints();
```
eder-matheus: I don't think it's ready to be called here. We're not updating CUGR with the dirty nets yet, AFAIK.

src/grt/src/cugr/src/CUGR.cpp
sparsh-karna: It's for batch ECO updates, if multiple pins on a single net are moved before the incremental route runs, addDirtyNet could get triggered multiple times for the same net ID. Deduplicating here just ensures we only route it once. I patterned it after FastRouteCore's queue, but let me know if you'd rather I check for duplicates during insertion instead!

src/grt/src/GlobalRouter.cpp
sparsh-karna: You're totally right, the upstream DB callbacks aren't feeding it yet. I just dropped it in as a placeholder to map out where it will eventually live. I'll revert the GlobalRouter.cpp changes right now so we keep this PR strictly scoped to the CUGR engine refactor.

src/grt/src/cugr/src/CUGR.cpp
eder-matheus: In GlobalRouter.cpp, we store the dirty nets into the dirty_nets_ set, which by construction prevents having duplicates of a single net.

Could you point me where in FastRouteCore we do this duplicates removal? I don't remember having it in the code.

sparsh-karna: 1. I double-checked and since `dirty_nets_` is a `std::set`, the dedup is already handled upstream. Plus, `updateOverflowNets` does a linear scan, so the secondary overflows are unique by default. The `unique` call is definitely just redundant overhead.
2. Spot on about `GlobalRouter.cpp` too. Calling it there is premature since the callbacks aren't feeding CUGR yet. I actually already reverted that in my last push!

Just pushing the dedup removal now. Thanks for catching these!

PR #9772, opened 2026-03-16
install: check GUI deps before building

oharboe:
Check for required xcb system packages before starting the expensive Bazel build. Fails fast with a helpful error message showing the exact apt install command on Ubuntu-based systems.

This follows the principle of least astonishment: rather than failing deep in the build with cryptic errors, check upfront and tell the user exactly what to install.

### DIFF
--- a/bazel/install.sh
+++ b/bazel/install.sh
@@ -3,6 +3,34 @@ set -e
 
 # Install binary and runfiles from bazel build
 
+# Check for required system dependencies before expensive build.
+# Each tool checks its own deps (separation of concerns).
+#
+# Currently only Ubuntu/Debian is checked. Dependency checking for
+# other platforms (macOS, RHEL, Fedora, etc.) is not implemented
+# because we cannot test them. Contributions welcome.
+check_ubuntu_deps() {
+    local missing=()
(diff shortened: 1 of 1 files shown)


oharboe:
@gadfort FYI

oharboe:
Closing this — the dep check is unnecessary.

The OpenROAD Bazel build produces a statically linked binary with no runtime dependency on xcb or Qt system packages:

```
$ ldd tools/install/OpenROAD/bin/openroad
    linux-vdso.so.1
    libm.so.6
    libc.so.6
    /lib64/ld-linux-x86-64.so.2
```

```
$ dpkg -s libxcb-icccm4-dev 2>&1
dpkg-query: package 'libxcb-icccm4-dev' is not installed
```

GUI works fine without any of the checked packages installed. Bazel handles Qt/xcb internally.

oharboe:
@maliberty Do we have any dependencies that we need to install on a blank ubuntu anymore?

I have installed the dependencies on my machines long ago, so I can't check.

maliberty:
If you build with the gui then you do need a bunch see https://github.com/The-OpenROAD-Project/OpenROAD/pull/8532 (still in progress)

PR #9776, opened 2026-03-16
fix: validate missing -cells argument in insert_decap

suhr25:
# SUMMARY

This PR adds a check for the mandatory `-cells` argument in `insert_decap` (`src/psm/src/pdnsim.tcl`). Previously the code accessed `keys(-cells)` without verifying it existed, causing a confusing Tcl error when the argument was omitted. The change returns a clear error message instead.

---

# FIX

**Before**

```tcl
set cells_and_decap $keys(-cells)
```

**After**

```tcl
if { ![info exists keys(-cells)] } {
  utl::error PSM 182 "Missing mandatory argument -cells."
}
set cells_and_decap $keys(-cells)
```

---

# VERIFICATION

Run:

```
insert_decap -target_cap 100fF
```

**Before:** Tcl array error
**After:** `PSM 182: Missing mandatory argument -cells.`

### DIFF
--- a/src/psm/src/pdnsim.tcl
+++ b/src/psm/src/pdnsim.tcl
@@ -111,11 +111,13 @@ proc insert_decap { args } {
     set target_cap [expr [sta::capacitance_ui_sta $target_cap] / [sta::distance_ui_sta 1.0]]
   }
 
-  # Check even size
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @eder-matheus @maliberty,
Thanks!

PR #9778, opened 2026-03-16
Up to the latest of the bazel 8.x line

hzeller:
Current is 8.6.0


hzeller:
Since I can't run `/test/orfs` because of docker, bazel of course also can't update the maven part.
So we've to leave this unresolved here and update later from some machine that does have docker.

@oharboe should we add a diff output to the MODULE.bazel.lock CI check, so that users like me can just take that and update the lock file ?

hzeller:
I re-run bazel with target `//test/orfs/mock-array:generate_verilog` ... it couldn't run it, but it did update the lock file, maybe that is sufficient.

hzeller:
Ok, looks like with `bazel mod deps --lockfile_mode=update` I could get it to do what makes the CI happy.

PR #9782, opened 2026-03-16
grt: check layer range for clock nets on CUGR

eder-matheus:
Fixes https://github.com/The-OpenROAD-Project/OpenROAD/issues/9716


eder-matheus:
@maliberty This PR doesn't require a CI run, as it only touches CUGR. This should be a no-op on the ORFS side.

PR #9791, opened 2026-03-17
rsz: add max slew check to UnbufferMove before buffer removal

### LINKED ISSUE #9779: rsz: Add max slew check to UnbufferMove before buffer removal
### Description

rsz: UnbufferMove::doMove() should check max slew before removing a buffer
UnbufferMove::doMove() in [src/rsz/src/UnbufferMove.cc](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/src/UnbufferMove.cc#L57-L214) currently validates max fanout (L104–L146) and max capacitance (L148–L180) before removing a buffer, but does not check max slew. There is even an explicit TODO at [L73–L74](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/src/UnbufferMove.cc#L73-L74):
```cpp
// TODO:
// 1. add max slew check
```
Why this matters:

When a buffer is removed, the upstream driver must now directly drive the downstream load. This increases the output transition time (slew) on the upstream driver. Without a slew check, UnbufferMove can introduce max slew violations that RepairDesign would need to fix later — or worse, violations that go undetected and degrade signal integrity.

The existing checks in the function header comment even list the conditions for buffer removal, but condition 5 (max slew) is missing:

```cpp
// Remove driver if
// 1) it is a buffer without attributes like dont-touch
// 2) it doesn't create new max fanout violations
// 3) it doesn't create new max cap violations
// 4) it doesn't worsen slack
//  ← missing: 5) it doesn't create new max slew violations
```

### Suggested Solution

Add a max slew check in UnbufferMove::doMove() after the existing max cap check (L148–L180) and before the slack estimation (L182). The implementation should:

Use sta_->checkSlew() on prev_drvr_pin to get the current slew and the slew limit for the upstream driver pin.

Estimate the new slew after buffer removal. The increased load capacitance on the upstream driver will increase slew. This can be estimated using the Elmore delay model: slew ≈ factor × drive_resistance × load_cap. The existing BaseMove::checkMaxSlewViolation() in [src/rsz/src/BaseMove.cc L868–L896](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/src/BaseMove.cc#L868-L896) already implements this pattern and can be reused or adapted.

Reject buffer removal if the estimated new slew exceeds the max slew limit, with an appropriate debugPrint message following the same pattern as the existing fanout/cap checks:

```// After the max cap check block (~L180), add:
// Watch out for new max slew violations
float slew, max_slew, slew_slack;
const Scene* slew_corner;
const RiseFall* slew_tr;
sta_->checkSlew(prev_drvr_pin,
                sta_->scenes(),
                resizer_->max_,
                // return values
                slew,
                max_slew,
                slew_slack,
                slew_tr,
                slew_corner);
if (max_slew > 0.0 && slew_corner) {
  // Estimate new slew after absorbing buffer's load
  // ... (estimate based on increased cap)
  if (estimated_new_slew > max_slew) {
    debugPrint(logger_,
               RSZ,
               "repair_setup",
               2,
               "buffer {} is not removed because of max slew limit "
               "of {} at {}",
               db_network_->name(drvr),
               max_slew,
               network_->pathName(prev_drvr_pin));
    return false;
  }
}
```
Update the function header comment at L57–L61 to include the new condition:
```cpp
// Remove driver if
// 1) it is a buffer without attributes like dont-touch
// 2) it doesn't create new max fanout violations
// 3) it doesn't create new max cap violations
// 4) it doesn't create new max slew violations   ← NEW
// 5) it doesn't worsen slack
```
Add a regression test under src/rsz/test/ that verifies a buffer is NOT removed when doing so would cause a max slew violation.

I would be happy to work on a PR for this. @maliberty @eder-matheus please assign this to me

### Additional Context

_No response_

Vi-shub:
Resolves #9779

### 1. Max slew check in `UnbufferMove::doMove()` (`UnbufferMove.cc`)

Added a max slew violation check after the existing max cap check (condition 3) and before slack estimation. The check:

- Calls `sta_->checkSlew()` on the upstream driver pin to get the slew limit
- Uses `computeNewDelaysSlews()` for library-accurate slew estimation after buffer removal (same function used by `estimatedSlackOK()`)
- Rejects buffer removal if the worst-case new slew exceeds the max slew limit
- Includes a `debugPrint` message consistent with existing fanout/cap check patterns

Also updated the function header comment to include the new condition and removed the TODO at the former L73-74.

### 2. RMS slew propagation in `estimateSlewsInTree()` (`Resizer.cc`)

Per @precisionmoon's guidance that current slew estimation is too conservative, replaced linear-sum slew propagation with root-mean-square (RMS) for both via and wire degradation:

- `upstream_slew + t` → `sqrt(upstream_slew² + t²)`

This follows the Bakoglu model where gate output slew and interconnect RC delay contribute independently to transition time, reducing estimation pessimism.

## Testing

- Built successfully on Ubuntu 22.04 (WSL2)
- 185/191 rsz regression tests pass (97%)
- All `remove_buffers*`, `repair_slew*`, and `repair_setup*` tests pass ✅

image
image
image

### DIFF
--- a/src/rsz/src/UnbufferMove.cc
+++ b/src/rsz/src/UnbufferMove.cc
@@ -58,7 +58,8 @@ using sta::Vertex;
 // 1) it is a buffer without attributes like dont-touch
 // 2) it doesn't create new max fanout violations
 // 3) it doesn't create new max cap violations
-// 4) it doesn't worsen slack
+// 4) it doesn't create new max slew violations
+// 5) it doesn't worsen slack
 bool UnbufferMove::doMove(const Path* drvr_path,
                           int drvr_index,
                           Slack drvr_slack,
@@ -70,8 +71,6 @@ bool UnbufferMove::doMove(const Path* drvr_path,
   LibertyPort* drvr_port = network_->libertyPort(drvr_pin);
   LibertyCell* drvr_cell = drvr_port ? drvr_port->libertyCell() : nullptr;
 
-  // TODO:
-  // 1. add max slew check
   if (drvr_cell && drvr_cell->isBuffer()) {
     Pin* drvr_pin = drvr_path->pin(this);
     Instance* drvr = network_->instance(drvr_pin);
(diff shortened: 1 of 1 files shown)


Vi-shub:
@maliberty can you please review it. And the Jenkins works seems offline now??

Vi-shub:
@precisionmoon the jenkins worker was removed without notifying. The CI is working fine?

Vi-shub:
Hi @maliberty @precisionmoon can you please review it and merge. Thankfully it passed all Jenkins test !!

precisionmoon:
Why doesn't changed file list show Resizer.cc?

Vi-shub:
> Why doesn't changed file list show Resizer.cc?

Hi @precisionmoon,
I originally included an RMS (root-mean-square) slew propagation change in Resizer.cc to make the slew estimation less pessimistic. However, when running the full ORFS check-in regression tests (specifically aes_sky130hs), I found that the less-pessimistic estimation caused the resizer to under-repair violations, leading to a post-routing PPA regression (DRT::max_slew_slack -6% < -0% and DRT::max_capacitance_slack -8% < -2%, you can cross verify with the jenkins builds). To resolve this regression and keep the PR focused on the primary issue, I reverted the Resizer.cc changes and kept only the core fix (the max slew check condition added to UnbufferMove::doMove()). That's why Resizer.cc is no longer in the changed file list."

PR #9792, opened 2026-03-17
mpl: don't use DEF HALO by default

AcKoucher:
Commercial tools usually treat the DEF HALO annotation as a hard constraint i.e., the rows are cut and nothing can be placed in that area.

In #9556, we introduced support for DEF HALOs in MPL in a way that it will by default create channels between macros of the same array based on the annotations if an overload is not specified with `set_macro_halo`. This ended up affecting flows in which DEF HALOs were being used but not considered in MPL. This PR is to address that temporarily.


AcKoucher:
@gadfort FYI.

eder-matheus (approved):

gadfort (approved):
Thanks, this worked when I tested it

PR #9798, opened 2026-03-18
fix: sort glob results in man_tcl_check tests for determinism

oharboe:
Unsorted glob ordering caused rcx_man_tcl_check to produce different results on CI vs locally when multiple Tcl files exist for a module.


oharboe:
@maliberty @luarss Please approve; single concern stability fix for CI builds.

@vvbandeira FYI

oharboe:
@maliberty The goal here is minimum churn, just fix the instability that is plaguing CI, so no, I don't want to take up Google Code assist on its suggestion.

PR #9802, opened 2026-03-18
Add Timing Python API for GUI-free timing reports

oharboe:
Expose timing path data (slack, skew, fanout, delays, slews) through the Python/SWIG API so that external scripts can generate reports without requiring the Qt GUI.

New files:
- include/ord/Timing.h: C++ timing data extraction API
- src/Timing.cc: Implementation using STA internals
- src/OpenRoad-py.i: SWIG bindings for Python access
- test/timing_report_api.py: Basic API smoke test

### DIFF
--- a/include/ord/Timing.h
+++ b/include/ord/Timing.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <string>
+#include <utility>
 #include <vector>
 
 #include "sta/Clock.hh"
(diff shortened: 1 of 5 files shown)


oharboe:
Superseded by #9807 which includes the API, STA searchPreamble fix, all review feedback, and an idiomatic Bazel py_test.

PR #9803, opened 2026-03-18
fix: add missing -net validation in add_sroute_connect

suhr25:
## SUMMARY

Adds a missing check for the required `-net` argument in `add_sroute_connect`.
Prevents silent failures and matches existing validation for other required args.

---

## FIX

**Before**

```tcl
set net ""
if { [info exists keys(-net)] } {
  set net $keys(-net)
}
```

**After**

```tcl
if { ![info exists keys(-net)] } {
  utl::error PDN 1197 "The -net argument is required."
}
set net $keys(-net)
```

---

## VERIFICATION

* Run without `-net` → now errors clearly
* Valid usage unchanged

### DIFF
--- a/src/pdn/src/pdn.tcl
+++ b/src/pdn/src/pdn.tcl
@@ -694,7 +694,7 @@ proc add_sroute_connect { args } {
   if { ![info exists keys(-net)] } {
     utl::error PDN 1197 "The -net argument is required."
   }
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @eder-matheus @gadfort,
Thanks!

suhr25:
Thanks for the merge!

PR #9806, opened 2026-03-18
docker: Optimize image size by purging apt cache in dev stage

Divinesoumyadip:
It's optimizes the main Dockerfile by cleaning up the `apt` package cache within the `dev` stage. 

 Added `apt-get clean` and `rm -rf /var/lib/apt/lists/*` to the `dev` stage `RUN` block.Ensures that temporary package lists are not persisted in the Docker layers, resulting in a smaller base image for the `builder` and `final` stages.

Reduces the overall footprint of the OpenROAD development and release images, leading to faster CI/CD build times and quicker downloads for end-users. This directly supports the **Onboarding Simplification** project's goal of making installation options more robust and efficient.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,14 +18,24 @@ ARG INSTALLER_ARGS=""
 COPY etc/DependencyInstaller.sh /tmp/.
 RUN <<EOF
 set -e
+# 1. Install dependencies via script
 /tmp/DependencyInstaller.sh -ci -base
 /tmp/DependencyInstaller.sh -ci -common -save-deps-prefixes=/etc/openroad_deps_prefixes.txt $INSTALLER_ARGS
+
+# 2. Optimization: Strip libQt5Core if on Ubuntu
 if echo "$fromImage" | grep -q "ubuntu"; then
     echo "fromImage contains 'ubuntu' — stripping section from libQt5Core.so"
     strip --remove-section=.note.ABI-tag /usr/lib/x86_64-linux-gnu/libQt5Core.so || true
-else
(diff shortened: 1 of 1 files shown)


Divinesoumyadip:
@sombraSoft Kindly check . Thanks

sombraSoft:
Thanks for the contribution, @Divinesoumyadip! We use this image in our internal Jenkins pipeline and verified the change is safe for our downstream consumers.

Benchmarked the saving on `ubuntu:26.04` (`--target dev`):

| Build | Image size |
|---|---|
| master | 2543.79 MB |
| this PR | 2504.98 MB |
| **saved** | **~39 MB** |

The delta matches `/var/lib/apt/lists` size exactly — `apt-get clean` was a no-op because `DependencyInstaller.sh` already runs `apt-get autoclean` on the Ubuntu path.

I pushed a few small follow-ups to the branch (the PR has `Allow edits by maintainers` enabled):
1. Restore trailing newline at EOF.
2. Restore the `else echo "Skipping strip..."` log line so Rocky/Debian builds still log why the strip step is skipped.
3. Replace `&> /dev/null` with `>/dev/null 2>&1` — the `RUN <<EOF` heredoc runs under `/bin/sh` (dash on Ubuntu), where `&>` parses as background-run + stray redirect and the guard would be bypassed.
4. Drop the redundant `apt-get clean` and trim numbered comment headers.
5. Rebased on current `master` to refresh CI.

Net diff vs. master is now just the 3-line `if command -v apt-get …; then rm -rf /var/lib/apt/lists/* fi` block plus the restored `else` branch. Merging once CI is green.

PR #9807, opened 2026-03-18
Add Timing Python API with idiomatic Bazel py_test

oharboe:
## Summary

- **Commit 1**: Add Timing C++ API exposing `getWorstSlack`, `getTotalNegativeSlack`, `getEndpointCount`, `getEndpointSlackMap`, `getClockInfo`, and `getTimingPaths` through SWIG Python bindings. Fix STA search state crash by adding `ensureGraph()` + `searchPreamble()` before `findPathEnds()`, matching the Qt GUI's `STAGuiInterface::initSTA()` pattern.
- **Commit 2**: Add idiomatic Bazel `py_test()` that imports `openroadpy` directly, removing the framework inversion of shelling out to `openroad -python`. Uses index-based SWIG vector access to work around `swig::stop_iteration` C++ exception during garbage collection.

## Test plan

- [x] `bazelisk test //test:timing_report_api_test` passes
- [x] Test exercises full query sequence: `getWorstSlack` → `getTNS` → `getEndpointSlackMap` → `getClockInfo` → `getTimingPaths`
- [x] clang-format-20 clean
- [x] Black clean
- [x] buildifier clean

### DIFF
--- a/src/Timing.cc
+++ b/src/Timing.cc
@@ -610,8 +610,7 @@ std::vector<TimingPathInfo> Timing::getTimingPaths(MinMax minmax,
         odb::dbModITerm* mod_iterm;
         db_network->staToDb(
             prev_pin, arc.from_iterm, arc.from_bterm, mod_iterm);
-        db_network->staToDb(
-            pin, arc.to_iterm, arc.to_bterm, mod_iterm);
+        db_network->staToDb(pin, arc.to_iterm, arc.to_bterm, mod_iterm);
         if (same_inst && arc.to_iterm) {
           arc.master = arc.to_iterm->getInst()->getMaster();
         }


include/ord/Timing.h
```diff
@@ -35,6 +37,47 @@ namespace ord {
...
+  float period;
+  std::vector<float> waveform;
+  std::vector<std::string> sources;
+};
+
+struct TimingArcInfo
+{
+  std::string from_pin;
+  std::string to_pin;
+  std::string cell_name;
```
maliberty: We should be returning objects not strings (throughout).  Objects have a lot more operations than just their name.

oharboe: fixed

PR #9809, opened 2026-03-18
toolchains: board up broeken window, stub libxml2 to silence ld.lld warning

oharboe:
avoids sudo ln hacks on Ubuntu, for instance.

ld.lld dynamically links libxml2 but only uses it for Windows manifest merging.  Build a versioned stub into @llvm_prebuilt lib/ via patch_cmds so the RUNPATH picks it up instead of the system library, eliminating the "no version information available" warning and the libxml2-dev dependency.


oharboe:
@hzeller :crying_cat_face: What else can we do?

maliberty:
Yuck.  Where does this warning happen?

oharboe:
> Yuck. Where does this warning happen?

Ubuntu 25.10 and who knows where else... this will become better, but for now we don't need Windows xml manifest parsing, so just board up the broken window. It is easier to have Claude board up this broken window than to document it. We never saw it again. :relieved:

maliberty:
I still have no idea what the actual problem is.  All this to solve a warning?

oharboe:
> I still have no idea what the actual problem is. All this to solve a warning?

Yes: users have to fix mysterious errors on Ubuntu when building with bazel without this. I think this is a shortcoming in the upstream bazel bcr llvm toolchain that will go away as bazel progresses. I'm thinking we should board up the broken window and forget all about it, surely they know upstream and they have a plan to fix it, it just takes time.

As the title says: I'm boarding up a broken window.

Important and useful warnings are drowned out by noise warnings and at some point nobody cares about warnings. 

Bad.

@hzeller Thoughts?

oharboe:
I get a ton of these:

```
INFO: From Linking external/googletest+/libgtest_main.so:
external/toolchains_llvm++llvm+llvm_toolchain/bin/../../toolchains_llvm++llvm+llvm_toolchain_llvm/bin/ld.lld: /lib/x86_64-linux-gnu/libxml2.so.2: no version information available (required by external/toolchains_llvm++llvm+llvm_toolchain/bin/../../toolchains_llvm++llvm+llvm_toolchain_llvm/bin/ld.lld)
external/toolchains_llvm++llvm+llvm_toolchain/bin/../../toolchains_llvm++llvm+llvm_toolchain_llvm/bin/ld.lld: /lib/x86_64-linux-gnu/libxml2.so.2: no version information available (required by external/toolchains_llvm++llvm+llvm_toolchain/bin/../../toolchains_llvm++llvm+llvm_toolchain_llvm/bin/ld.lld)
```

hzeller:
I think the amount of changes to work around this temporary glitch is not useful. It creates immediate technical debt with a lot of head-scratching of someone trying to understand what is going on.

Not everything that can be hacked around with a massive hack that can be done by Claude ... is something one should be done :)

What I found is, that the builds of llvm are different and sometimes link in `libxml`, sometimes not, also `tinfo` is another library that is sometimes linked sometimes not. So I'd try to check a slightly compiler version (not 20.1.8 but something different), also with the newest 1.6.0 toolchains_llvm available on BCR.

If that doesn't work, I think for now it is ok to document that it won't work for certain versions of Ubuntu, and just recommend to comment out the `@llvm_toolchain//:all` part in register_toolchains to use the local compiler in that case.

oharboe:
@hzeller @gadfort Would you like to board up the broke libxml2 window and forget all about it?

This cruft will wash out in the future LLVM upgrade.

This makes OpenROAD more hermetic and we've had great success with this patch, bazel-orfs now caches downstream .odb artifacts robustly for us.

DependencyInstaller.sh doesn't need complications, etc.

hzeller:
As I said before, this artillery approach to shot that fly is **not** what we should do.

Did you try all combinations of [toolchains_llvm](https://registry.bazel.build/modules/toolchains_llvm) and all compilers in each of them first ? Not all compilers in the LLVM toolchain are build with the broken assumption that XML is needed.


What I said before still stands: if it doesn't work yet on these platforms, just comment out [this line](https://github.com/The-OpenROAD-Project/OpenROAD/blob/3a15695692bdb3cdc7fa47c5e6b60fbd312b6711/MODULE.bazel#L177) and use the local compiler.

The toolchains llvm is pretty terrible as it does not provide an hermetic toolchain, so before we do such hacks here, we should fix upstream (e.g. fix how the LLVM projects builds their binaries, not linking XML, provide static binaries..) That is time better spent than spending energy working around this locally with a breakneck hack like this that is a huge maintenance liability.

oharboe:
@hzeller Running local LLVM breaks caching of .odb, so not an option. This libxml problem is already fixed, so there's nothing to do upstream, we're just waiting for it to trickle through. @gadfort and I tried the latest LLVM, not there yet.

This is just some ugly cruft to forget about this problem until the already fixed problem upstream trickles down to us.

oharboe:
@hzeller I think cruft in master that users should ignore, that we have a plan to wash out later when the fix trickles down beats "hack MODULE.bazel" instructions to the user: no documentation required or actions by users or developers.

hzeller:
> ... This libxml problem is already fixed, so there's nothing to do upstream, we're just waiting for it to trickle through. 

If it is already fixed, the _first_ instinct must always be to make the 'trickle through' happen.
If you have the problem, the only person who can fix it is yourself, not wait for anyone else to fix it. We all own the code, this is how we can make things happen.

oharboe:
@hzeller LLVM cycles are long, rightly so, what knob is there for me to press here? I don't think I can speed up that cycle, nor that I should try. I just want to stop us from fixing things that shouldn't be fixed(DependencyInstaller.sh complications while we wait) and from users from having to suffer pain(rebuilding .odb unecessarily) and confusion for no reason while we wait for the upgrade to trickle through.

oharboe:
**TL;DR — this is now a choice between two dated options, not an open-ended hack:**

The proper fix already exists upstream: LLVM release binaries statically link libxml2 as of llvm/llvm-project#192088 (on `main` since April 15). But it missed the `release/22.x` branch, and the 22.x point-release train has ended (the last scheduled slot, 22.1.9 on June 30, went unused). So the first official binaries with the fix are:

- **23.1.0-rc1: ~July 17, 2026** (`release/23.x` branches July 14, fix rides along)
- **23.1.0 final: August 25, 2026**

There is no way to make that faster. A backport has no release vehicle left, and BCR/toolchains_llvm are not the bottleneck — `toolchains_llvm` already tracks new LLVM releases within days, and `extra_llvm_distributions` lets us consume 23.1.0 the day it's published.

So the choice is:

1. **Let everyone on Ubuntu 25.10+/Arch/etc. suffer the warnings-or-worse until ~August 25** (≈7 weeks), then bump to LLVM 23.1.0, or
2. **Merge this painkiller now**, and delete it in the 23.1.0 bump. It has a hard expiry date.

I prefer 2, but either way the endpoint is the same: the LLVM 23.1.0 toolchain bump in late August removes all of this.

gadfort (changes requested):
@oharboe can you also remove the installation of libxml2 from the dependency installer. I checked this does work on my setup as well.

gadfort:
@oharboe I agree with Option 2 as well

oharboe:
> @oharboe can you also remove the installation of libxml2 from the dependency installer. I checked this does work on my setup as well.

That's independent followup cleanup. Let's agree on that this is what we want first.

@maliberty ?

hzeller:
The problem is, that you always do these temporary hacks @oharboe, but I am usually the one who has clean them up later :/

oharboe:
> The problem is, that you always do these temporary hacks @oharboe, but I am usually the one who has clean them up later :/

Claude cleans them up for me, so I dont think about it. 

Yes: if you clean them up manually it is a problem.

Feel free to file github issues and ping me if you need to use my Claude 🤠

maliberty:
@hzeller going to a local non-hermetic compiler isn't a win.  I don't find that to be a solution so we need to find another.  If there is an upstream solution we can just use then we should.  So far nobody has found one and we need to unblock Ubuntu 26 which requires this.  Baring something better coming along I don't see an alternative.  I agree it is icky but if the other choice is a local compiler I'll take this.  If I want to manage local dependencies I'll just stay with cmake.

hzeller:
My point is: if there is no upstream solution, we need to make it happen.

maliberty:
It is already planned for LLVM 23.1.0 projected release in late August.  What do you propose we do?

maliberty:
I think  the stop-gap debate is between this and #10778 - I lean towards the later.  This one still relies on a local compiler in order to build the fake library.

oharboe:
> I think the stop-gap debate is between this and #10778 - I lean towards the later. This one still relies on a local compiler in order to build the fake library.

I wasnt aware of that snag, it did not bite us.

Please choose your poison and close merge per your judgment

oharboe:
/gemini Can we fix the problem Matt pointed out and not use host compiler, but the LLVM to build the libxml stub hack?

oharboe:
> I think the stop-gap debate is between this and #10778 - I lean towards the later. This one still relies on a local compiler in order to build the fake library.

Gemini says this is easily fixable.

@maliberty Reconsider in light of that?

maliberty:
I don't think it is true.  The compiler will fail for exactly the same reason.  Its a chicken-egg problem.  If you can actually do it on a machine (docker container) with no local compiler I'll reconsider.

oharboe:
I will ask my guy to create a proof. I think this can be done w some trick like chroot or something, no docker needed.

oharboe:
/gemini how do I prove with a few bash commands that Matt's concern that this wont use the local compiler without using docker?

oharboe:
will try a different approach.

PR #9820, opened 2026-03-19
build: Add Python virtual environment guard to Build.sh

Divinesoumyadip:
Adds a safety check to `Build.sh` to detect if the user is operating inside a Python Virtual Environment .

OpenROAD relies heavily on Python bindings and external PIP packages. On modern Linux distributions (like Ubuntu 24.04 with PEP-668), attempting to install dependencies globally can break OS-level package managers. Beginners frequently fall into this trap, resulting in corrupted local environments and support tickets.
The script now checks for a registered virtual environment. If none is found, it issues a highly visible warning with exact instructions on how to create one (`python3 -m venv or_env`), pauses for 3 seconds to ensure the user reads it, and then proceeds.

Directly supports the **Onboarding Simplification** initiative by preemptively stopping developers from destroying their local Python environments during the initial build phase.

### DIFF
--- a/etc/Build.sh
+++ b/etc/Build.sh
@@ -34,28 +34,28 @@ OPTIONS:
                                                  -cmake= and double quotes if
                                                  <key> has multiple <values>
                                                  e.g.: -cmake='-DFLAGS="-a -b"'
-  -compiler=COMPILER_NAME                        Compiler name: gcc or clang
+  -compiler=COMPILER_NAME                         Compiler name: gcc or clang
                                                  Default: gcc
   -no-warnings
                                                 Compiler warnings are
                                                 considered errors, i.e.,
                                                 use -Werror flag during build.
-  -dir=PATH                                     Path to store build files.
+  -dir=PATH                                      Path to store build files.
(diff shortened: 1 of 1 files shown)


maliberty:
What external packages are being updated during Build.sh?

Divinesoumyadip:
> What external packages are being updated during Build.sh?

Hi @maliberty, that is a fair point. `Build.sh `itself doesn't directly update external PIP packages.However, I've noticed that new contributors on modern distributions (like Ubuntu 24.04+) often hit the `externally-managed-environment` (PEP-668) error immediately after a successful build when they try to install flow dependencies like `pyverilog` or `utl.`
Since the build process links against the active Python interpreter, adding this guard to `Build.sh` serves as a "pre-flight check." It ensures beginners have a safe, non-system workspace established before they invest time in the compilation, ultimately reducing environment-related support tickets later in the onboarding process.

maliberty:
Could you give an example of such a problem report?  Is `utl` also a pip module or are you referring to the `src/utl` module?  Where is `pyverilog ` used?

PR #9822, opened 2026-03-19
build: add ASCII splash screen to Build.sh for better onboarding UX

Divinesoumyadip:
First impressions matter. When a new contributor runs the build script, they are immediately hit with a wall of dense system text. It feels intimidating and purely functional. 
By adding a clean, cyan OpenROAD splash screen at the very start of the script execution, we instantly make the CLI feel more professional, welcoming, and polished.
While primarily cosmetic, this directly aligns with the "vibe" of the **Onboarding Simplification** initiative. It signals to new developers that the build environment is well-maintained and user-centric before the compilation even begins.


maliberty:
The graphics are off
image

PR #9823, opened 2026-03-19
odb: refactor dbDatabase lookups to use std::string_view for zero-allocation

Divinesoumyadip:
This PR refactors several lookup methods in `dbDatabase` (OpenDB core) to use `std::string_view` (C++17) instead of `const char*`. 

Currently, many database lookup functions like `findLib`, `findTech`, and `findMaster` rely on legacy C-style strings (`const char*`) and `strcmp`. When these functions are called with `std::string` arguments, it can lead to unnecessary temporary memory allocations or less readable comparison logic.

By switching to `std::string_view`:
1. We enable zero-allocation lookups regardless of whether the input is a raw C-string or an `std::string`.
2. We replace manual `strcmp` calls with the more efficient and expressive `==` operator.
3. We move toward a more modern, type-safe API that aligns with the project's ongoing transition to modern C++ standards.

This is a structural refactor in the `odb` component. It improves the efficiency of high-frequency database operations and demonstrates a commitment to C++17 best practices within the OpenROAD core.

### DIFF
--- a/src/odb/include/odb/db.h
+++ b/src/odb/include/odb/db.h
@@ -7530,7 +7530,9 @@ class dbDatabase : public dbObject
   /// Find a specific lib.
   /// Returns nullptr if no lib was found.
   ///
-  dbLib* findLib(const char* name);
+ #include <string_view> // Add this at the top with other includes
+// ...
+  dbLib* findLib(std::string_view name);
 
   ///
   /// Return the techs contained in the database. A database can contain
--- a/src/odb/src/db/dbDatabase.cpp
+++ b/src/odb/src/db/dbDatabase.cpp
@@ -33,6 +33,7 @@
 #include <ostream>
 #include <stdexcept>
 #include <string>
+#include <string_view> // Added for C++17 string_view support
 #include <vector>
 
 #include "absl/synchronization/mutex.h"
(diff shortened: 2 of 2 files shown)


src/odb/include/odb/db.h
```diff
@@ -7530,7 +7530,9 @@ class dbDatabase : public dbObject
   /// Find a specific lib.
   /// Returns nullptr if no lib was found.
   ///
-  dbLib* findLib(const char* name);
+ #include <string_view> // Add this at the top with other includes
```
maliberty, line 7533: header inclusion belong at the top of the file

maliberty:
Please verify that this works correctly through swig in both tcl and python.

I presume this is a trial for a broader change as there are many APIs that could be converted.

Divinesoumyadip:
> Please verify that this works correctly through swig in both tcl and python.
> 
> I presume this is a trial for a broader change as there are many APIs that could be converted.

Hi @maliberty, thank you for the review. You are completely right,this was a trial balloon to see how smoothly we could transition some of the core DB lookups to C++17 std::string_view.
I am going to close this experimental PR so it doesn't clutter the queue. Thank you for the guidance!

PR #9763, opened 2026-03-14
grt: Refactor CUGR GridGraph API to resolve boolean blindness in tree rip-up

Divinesoumyadip:
Currently, `GridGraph::commitTree` utilizes a boolean flag to handle both capacity additions and rip-ups. This leads to boolean blindness at the call site (e.g., `commitTree(tree, /*ripup*/ true)`) and creates semantic confusion where a function named "commit" is used to remove data. 
Moves the boolean-flagged `commit*` functions to the `private` scope.Exposes clean, self-documenting public APIs: `addTreeUsage()` and `removeTreeUsage()`.Refactors `CUGR.cpp` to use these explicit calls, cleanly separating capacity allocation from rip-up mechanics.
This API modernization aligns with recent CUGR state management refactors and paves the way for deterministic history-cost preservation during incremental timing ECOs.

### Related Issues
Supports the architectural prep for Incremental CUGR .

PR #9826, opened 2026-03-19
grt: Expose CUGR removeTreeUsage API to SWIG/Tcl interface

Divinesoumyadip:
This PR exposes the modernized `GridGraph` rip-up logic to the user-facing Tcl and Python layers via SWIG. It builds on the architectural refactor in #9763 by adding `removeRouteUsage` to the `CUGR `and `GlobalRouter` classes.

Exposing this functionality is critical for advanced ECO (Engineering Change Order) flows. It allows developers and researchers to manually release a net's global routing resource consumption directly from the scripting layer without a full router re-initialization. This is a key step for the Incremental CUGR project, providing the necessary hooks for custom rip-up-and-reroute heuristics in external scripts.
Added removeRouteUsage(odb::dbNet*) to the `CUGR` and `GlobalRouter` classes.Implemented logic in `CUGR.cpp` to fetch the `GRNet` and release its associated `RoutingTree` from the `GridGraph`. Also added SWIG bindings in `GlobalRouter.i` to expose the command as `remove_route_usage` in Tcl/Python.

### DIFF
--- a/src/grt/src/cugr/include/CUGR.h
+++ b/src/grt/src/cugr/include/CUGR.h
@@ -94,7 +94,7 @@ class CUGR
   }
   void addDirtyNet(odb::dbNet* net);
   void updateNet(odb::dbNet* net);
-  void removeRouteUsage(odb::dbNet* net); // <-- ADDED FOR SWIG API
+  void removeRouteUsage(odb::dbNet* net);
   void startIncremental();
   void endIncremental();
 
--- a/src/grt/src/cugr/src/CUGR.cpp
+++ b/src/grt/src/cugr/src/CUGR.cpp
@@ -665,8 +665,10 @@ void CUGR::removeRouteUsage(odb::dbNet* db_net)
       grid_graph_->removeTreeUsage(gr_net->getRoutingTree());
     }
   } else {
-    logger_->warn(
-        GRT, 601, "Net {} not found in CUGR for rip-up.", db_net->getConstName());
+    logger_->warn(GRT,
+                  601,
+                  "Net {} not found in CUGR for rip-up.",
+                  db_net->getConstName());
   }
 }
 


maliberty:
This is rather dangerous and I'm not convinced there is a real need for this low level control.  Please give a strong motivating example

Divinesoumyadip:
> This is rather dangerous and I'm not convinced there is a real need for this low level control. Please give a strong motivating example

Hi @maliberty , thank you for the review. I completely understand the concern about exposing low-level resource management to the scripting layer, as it could lead to the grid map getting out of sync if used carelessly.

This API is a foundational requirement for the upcoming GSoC 2026 Incremental CUGR project. In a post-route ECO timing repair flow (e.g., when rsz inserts a buffer to fix a setup violation), the original net's routing becomes physically invalid. Currently, updating the Global Router's congestion map requires a costly full re-initialization.
By exposing` remove_route_usage`, an external timing-optimization script can surgically release the `GridGraph` capacities of only that stale net. This allows the incremental router to immediately find a valid path for the newly split segments without a full-chip reroute, drastically speeding up the ECO loop. I am happy to add warning documentation to the SWIG interface if you'd prefer to flag it as an advanced/ECO-only command.

eder-matheus:
> Currently, updating the Global Router's congestion map requires a costly full re-initialization.
> By exposing remove_route_usage, an external timing-optimization script can surgically release the GridGraph capacities of only that stale net. This allows the incremental router to immediately find a valid path for the newly split segments without a full-chip reroute, drastically speeding up the ECO loop. I am happy to add warning documentation to the SWIG interface if you'd prefer to flag it as an advanced/ECO-only command.

@Divinesoumyadip That's not how incremental global route is implemented in OpenROAD. If you look at the FastRouteCore implementation, you'll see that the nets modified by RSZ are handled internally to the C++ implementation. There's no need for such public SWIG API, since we are not handling individual nets on Tcl scripts.

In summary, all the handling of incremental updates must be performed inside the C++ implementation and be transparent to the end user.

Divinesoumyadip:
> > Currently, updating the Global Router's congestion map requires a costly full re-initialization.
> > By exposing remove_route_usage, an external timing-optimization script can surgically release the GridGraph capacities of only that stale net. This allows the incremental router to immediately find a valid path for the newly split segments without a full-chip reroute, drastically speeding up the ECO loop. I am happy to add warning documentation to the SWIG interface if you'd prefer to flag it as an advanced/ECO-only command.
> 
> @Divinesoumyadip That's not how incremental global route is implemented in OpenROAD. If you look at the FastRouteCore implementation, you'll see that the nets modified by RSZ are handled internally to the C++ implementation. There's no need for such public SWIG API, since we are not handling individual nets on Tcl scripts.
> 
> In summary, all the handling of incremental updates must be performed inside the C++ implementation and be transparent to the end user.

Hi @eder-matheus ,thank you for the detailed clarification.My initial thought was to build a bridge for external ECO scripts, but I completely agree that keeping the state synchronization internal to the C++ engine (transparent to the user) is the much safer and standard approach for OpenROAD.

Should I go ahead and close this PR since the SWIG API approach is not aligned with the architecture, or would you prefer I repurpose this branch to start implementing the internal C++ tracking?

eder-matheus:
> Should I go ahead and close this PR since the SWIG API approach is not aligned with the architecture, or would you prefer I repurpose this branch to start implementing the internal C++ tracking?

We can close this PR and start from a fresh branch.
C++ tracking is already implemented in the GlobalRouter.cpp file. Look at the `GRouteDbCbk` class to understand how it's done, and how FastRouteCore is used.

PR #9827, opened 2026-03-19
build: dev_dependency cleanup, visibility, and downstream test

oharboe:
TL;DR use-cases:

- Use openroad from Python scripts in other Bazel projects (no need for pypi)
- Use latest master from other projects using bazel-orfs without waiting for OpenROAD to trickle into Docker image
- LLM in the loop flows, use a local clone of OpenROAD that the LLM (Claude) can hack to e.g. add extra logging to figure out why global route runs forever.

## Summary

- Mark `rules_shell`, `rules_pkg`, `rules_verilator`, `verilator` as `dev_dependency` so MVS doesn't force their versions on downstream consumers
- Make `toolchains_llvm` extension and toolchain registration `dev_dependency` (avoids root-module-only enforcement failure)
- Move `install` and `tarfile` targets to `packaging/` so their `load()` statements don't break downstream consumers when `rules_shell`/`rules_pkg` are dev deps
- Restrict `//:openroad_lib` visibility to `//:__subpackages__` — only `//:openroad` and `//:openroad_py` are public
- Rename `ord_py` → `openroad_py` for a clean external API name
- Fix `@//:__subpackages__` → `//:__subpackages__` in visibility labels (`@//` resolves to root module in bzlmod, breaking internal visibility when OpenROAD is a dependency)
- Add `test/downstream/` mock project that verifies visibility and dev_dependency isolation (0.5s, runs with `bazelisk test //test/...`)
- Document the downstream use case in `docs/user/Bazel.md`

## Motivation

When downstream projects pull in OpenROAD as a `bazel_dep`, MVS forces unnecessarily high versions of deps like `rules_cc 0.2.17` (via `rules_shell 0.6.1` etc.). `rules_cc >= 0.2.14` introduces `cc_compatibility_proxy` which creates a WORKSPACE cycle for any downstream project still using `WORKSPACE.bazel`. The `toolchains_llvm` extension also fails with "Only the root module can use the 'llvm' extension" when OpenROAD is not root.

## Test plan

- [x] `bazelisk build --nobuild //:openroad` from OpenROAD root
- [x] `bazelisk build --nobuild //packaging:install` from OpenROAD root
- [x] `test/downstream/visibility_test.sh` — 7/7 pass in 0.5s
- [ ] CI: `bazelisk test //test/...` picks up `//test:downstream_visibility_test`

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -110,7 +110,6 @@ single_version_override(
 )
 
 bazel_dep(name = "rules_scala", version = "7.1.5", dev_dependency = True)
-
 bazel_dep(name = "bazel-orfs", dev_dependency = True)
 bazel_dep(name = "bazel-orfs-verilog", dev_dependency = True)
 


oharboe:
@hzeller Thoughts?

hzeller:
Thanks, very nice to document downstream use and making dev dependencies more clear.

test/downstream/visibility_test.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
hzeller: make this `/usr/bin/env bash`

oharboe: I will try to remember, but we'd be better of telling Claude :-)

This is just a throwaway investigation, it is quickly generated low accuracy information that may merit further investigation or not... I'm just not quite over the novelty of it. 

image

Another example of using Claude to run the bazel-orfs flow: https://github.com/oharboe/openroad-demo

I predict bazel is going to be the most used build language in the universe: automatically generated AI plumbing that handles dependencies and caching...

maliberty: Does that mean you don't want this reviewed or merged?

oharboe: It is just information, but I wanted to share the code, so closed, yes.

Did you think it could have more than empheral use?

oharboe: Reviewed please.

oharboe: > make this `/usr/bin/env bash`

I want to [address that separately](https://github.com/The-OpenROAD-Project/OpenROAD/issues/9857), the code base is endemic with inconsistencies here.

hzeller:
I think this should get submitted (after the conflicts are resolved), and then we can refine from there.

What I'd also like to see is if we group the 'regular' and 'dev' dependencies into separate sections in the MODULE.bazel -- that will it much easier to work with longer term. Something like
https://github.com/hzeller/OpenROAD/commit/f506538c3f5ae7ae9cbf0cd625f6e17af153c040

oharboe:
@hzeller FYI, I'm rounding up all the stray cats(accidental transitive dependencies) in https://github.com/The-OpenROAD-Project/bazel-orfs w.r.t. dependencies and put them up for adoption... Still work in progress...

oharboe:
@hzeller The bottleneck here is review. It is tough to build a strong mental model of what is going on here for @maliberty 

Or if it is easy for him, I'm envious. :smile:

docs/user/Bazel.md
```diff
@@ -59,6 +59,63 @@ Without offering any deeper insight some comments about what is shown above:
...
+# qt-bazel is not in BCR; git_override is root-module-only,
+# so downstream consumers must repeat it.
+bazel_dep(name = "qt-bazel")
+git_override(
+    module_name = "qt-bazel",
+    commit = "df022f4ebaa4130713692fffd2f519d49e9d0b97",
+    remote = "https://github.com/The-OpenROAD-Project/qt_bazel_prebuilts",
+)
+
+# OpenROAD's toolchains_llvm extension is dev_dependency,
```
hzeller: The whole llvm_toolchain part could probably be remarked as an optional consideration for downstream users.

We only use the llvm_toolchain in openroad to not have to worry about changing cache artifacts between users, but downstream can compile that with whatever compiler they want as long as they set the c++ standard to 20.

maliberty: It is also for reproducibility.  I'd rather not get issues from downstream that are on a different compilers.

maliberty:
Nothing is bazel is easy 😉

docs/user/Bazel.md
oharboe: @maliberty @hzeller If I understand correctly:

- hzeller: downstream users should be free to pick any C++20 compiler — `toolchains_llvm` is just what we use internally to avoid cache churn between developers, not a hard requirement.
- maliberty: we don't want to encourage compiler diversity downstream and then field bug reports against untested toolchains. Reproducibility matters for supportability.

Would updating the docs comment to something like this work for both of you?

```starlark
# OpenROAD's toolchains_llvm is a dev_dependency, so the root module
# must provide its own C++ toolchain.  Using the same LLVM version as
# OpenROAD CI (shown below) is recommended for reproducibility;
# any C++20-capable toolchain will work, but issues found on other
# compilers may be lower priority.
```

This makes it clear it's technically optional (hzeller's point) while setting expectations that non-LLVM issues get lower priority (maliberty's point).

oharboe:
> Nothing is bazel is easy 😉

As long as you have a good reviewer, it is. :cowboy_hat_face:

docs/user/Bazel.md
hzeller: yes, but this is why it should be still be suggested, but not necessarily be a requirement (so essentially split the necessary MODULE.bazel into required and suggested parts).

Things should work with any compiler, and comparably so.
There are a bunch of platform-specific assumptions in OpenROAD for floating points for instance, but these are faults in the tests (comparing golden files instead of range checks)  and we can hopefully we can get that fixed over time. But these are only relevant for folks running the development tests in OpenROAD itself ... which is a different crowd of people running OpenROAD as a dependency (which this documentation is targeting).

From the end-users perspective, the binary should behave pretty much the same modulo some ULP difference here and there.

If we add the llvm toolchains dependency in the BUILD file, we need to explicitly describe why one might or might not want it - as a C++ user, people would be surprised otherwise.

hzeller: > Would updating the docs comment to something ...

I think this is a good compromise

oharboe:
@maliberty fixed merge conflicts. Automerge?

maliberty:
Tests are not passing:
 //test:downstream_visibility_test                                        FAILED in 4.6s

oharboe:
@maliberty Finally! The last fix was just enormously much faster tests. Automerge?

hzeller:
Looks like CI is only broken due to the currently not working `//src/web/test/...` tests.

oharboe:
> Looks like CI is only broken due to the currently not working `//src/web/test/...` tests.

soon as pr-merge triggers again, automerge should merge it. The fix is/will be in master by then @maliberty said.

hzeller:
Looks like the tests are fixed at head, so if you rebase, maybe the CI finishes and things go through.

oharboe:
if i touch this PR then matt has to reapply automerge. if i wait, the pr-merge that is done once a day will pass and it will merge.

hzeller:
ah, interesting. So pr-merge is auto re-basing current PRs ?

oharboe:
merging. I like a linear history, but preserving review  history is more valuable than it used to be

oharboe:
@hzeller :relieved:

PR #9830, opened 2026-03-19
fix: man_tcl_check tests fail on count mismatch

oharboe:
## Summary

- Add `sys.exit(1)` to all 24 `*_man_tcl_check.py` scripts when help/proc/readme counts don't match
- Add `set -o pipefail` to `test/regression_test.sh` so non-zero exit codes propagate through the `tee` pipe

Previously these tests passed even with mismatched counts because the `.ok` golden files encoded the broken state, and the pipe to `tee` swallowed the exit code.

## Test plan

- [x] `bazelisk test //src/drt/test:drt_man_tcl_check-py_test` — PASSES (counts match)
- [x] `bazelisk test //src/ant/test:ant_man_tcl_check-py_test` — FAILS (counts don't match, as expected)
- [ ] Existing regression tests still pass (pipefail should not affect passing tests)

### DIFF
--- a/src/ant/test/ant_man_tcl_check.py
+++ b/src/ant/test/ant_man_tcl_check.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import glob
 import re
 from extract_utils import extract_tcl_code, extract_help, extract_proc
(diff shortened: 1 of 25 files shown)


oharboe:
out of sight, out of mind for now

PR #9844, opened 2026-03-19
gpl: debug fix

gudeh:
Only show GUI toggle for GPL debug instances drawings if GPL debug mode is active.
Some other reorganization for debug mode and GPL charts.

### DIFF
--- a/src/gpl/src/graphicsImpl.cpp
+++ b/src/gpl/src/graphicsImpl.cpp
@@ -577,62 +577,52 @@ void GraphicsImpl::addIter(const int iter, const double overflow)
     return;
   }
   odb::dbBlock* block = pbc_->db()->getChip()->getBlock();
-  if (main_chart_ != nullptr) {
-    main_chart_->addPoint(iter,
-                          {block->dbuToMicrons(nbc_->getHpwl()), overflow});
-  }
+  main_chart_->addPoint(iter, {block->dbuToMicrons(nbc_->getHpwl()), overflow});
 
-  if (density_chart_ != nullptr) {
-    std::vector<double> values;
-    if (!nbVec_.empty() && nbVec_[0]) {
-      values.push_back((static_cast<double>(nbVec_[0]->getDensityPenalty())));
-      values.push_back(static_cast<double>(nbVec_[0]->getStoredPhiCoef()));
-    } else {
-      values.push_back(0.0);
-      values.push_back(0.0);
-    }
-    density_chart_->addPoint(iter, values);
+  std::vector<double> values;
+  if (!nbVec_.empty() && nbVec_[0]) {
+    values.push_back((static_cast<double>(nbVec_[0]->getDensityPenalty())));
+    values.push_back(static_cast<double>(nbVec_[0]->getStoredPhiCoef()));
+  } else {
+    values.push_back(0.0);
+    values.push_back(0.0);
   }
+  density_chart_->addPoint(iter, values);
 
-  if (stepLength_chart_ != nullptr) {
-    std::vector<double> values;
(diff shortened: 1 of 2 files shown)


src/gpl/src/graphicsImpl.cpp
```diff
@@ -82,58 +79,37 @@ void GraphicsImpl::debugForNesterovPlace(
...
-  main_chart_->setXAxisFormat("%d");
-  main_chart_->setYAxisFormats({"%.2e", "%.2f"});
-  main_chart_->setYAxisMin({std::nullopt, 0});
 
-  density_chart_ = gui->addChart(
-      "GPL Density Penalty", "Iteration", {"DensityPenalty", "phiCoef"});
-  density_chart_->setXAxisFormat("%d");
-  density_chart_->setYAxisFormats({"%.2e", "%.2f"});
-  density_chart_->setYAxisMin({0.0, nbc_->getNbVars().minPhiCoef});
+  initCharts();
```
maliberty: should this be conditional on debugging?

gudeh: I chose it not to be. Do you think it should? Either way, when we do `make gui_...` we do not see the GPL charts. Currently it only works if we run GPL itself with gui active (for example running an run-me artifact).

I realized the main chart was being created twice, I fixed that on the last commit.

maliberty: If you with the gui active but no debug enabled you shouldn't see the chart (which will be empty afaik).

src/gpl/src/nesterovPlace.cpp
```diff
@@ -276,10 +276,17 @@ void NesterovPlace::updateIterGraphics(
     int timing_driven_count,
     bool& final_routability_image_saved)
 {
+  if (graphics_) {
+    graphics_->addIter(iter, average_overflow_unscaled_);
+  }
+
   if (!graphics_ || !graphics_->enabled()) {
     return;
   }
```
maliberty: This seems odd/redudant.  Do we want to addIter if graphics are not enabled?

src/gpl/src/graphicsImpl.h
```diff
@@ -177,13 +176,14 @@ class GraphicsImpl : public gpl::AbstractGraphics,
...
   std::vector<odb::dbInst*> mbff_cluster_;
   Mode mode_;
-  gui::Chart* main_chart_{nullptr};
-  gui::Chart* density_chart_{nullptr};
-  gui::Chart* stepLength_chart_{nullptr};
-  gui::Chart* routing_chart_{nullptr};
+  static gui::Chart* main_chart_;
+  static gui::Chart* density_chart_;
+  static gui::Chart* stepLength_chart_;
+  static gui::Chart* routing_chart_;
```
maliberty, line 182: why are these static now?

src/gpl/src/graphicsImpl.cpp
```diff
@@ -552,9 +577,12 @@ void GraphicsImpl::addIter(const int iter, const double overflow)
     return;
   }
   odb::dbBlock* block = pbc_->db()->getChip()->getBlock();
-  main_chart_->addPoint(iter, {block->dbuToMicrons(nbc_->getHpwl()), overflow});
+  if (main_chart_ != nullptr) {
```
maliberty: when would this be null?

src/gpl/src/graphicsImpl.h
gudeh: We were generating the main_chart_ twice and populating it only once. I suspect it was because we can build the nesterov object more than once at replace.cpp. This guarantees we don't generate anything twice.

PR #9846, opened 2026-03-19
pdn: check extention against new obs box for possible violations

gadfort:
Changes:
- when extensions are applied, it was possible to create a violation based on the new shape, this ensures we check the new bounding box for violations as well.


gadfort:
@maliberty I'm not sure I have the time at the moment to get a unit test together for this, maybe once things settle I can try.

PR #9849, opened 2026-03-19
rsz: check max cap/slew across all timing corners

### LINKED ISSUE #9793: rsz: SizeDownMove and BaseMove max cap/slew checks only consider a single timing corner
### Description

`SizeDownMove::downSizeGate()` in `src/rsz/src/SizeDownMove.cc` (L327-347) calls `checkMaxCapViolation()` and `checkMaxSlewViolation()` to validate downsized cell candidates. However, both functions only check against a **single timing corner**, as noted by two explicit FIXMEs:
**FIXME 1** : `BaseMove::checkMaxCapViolation()` at [L840](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/src/BaseMove.cc#L840):
```cpp
// FIXME: Can we update to consider multiple corners?
output_port->capacitanceLimit(resizer_->max_, max_cap, cap_limit_exists);
```

FIXME 2 :  SizeDownMove::downSizeGate() at [L333](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/src/SizeDownMove.cc#L333):

```cpp
// FIXME: Only applies to current corner
if (checkMaxCapViolation(output_pins[i], output_port, output_caps[i])) {
```
Why this matters:

In multi-corner/multi-mode (MCMM) designs which are standard practice for advanced nodes a cell swap that passes cap/slew checks at the current analysis corner may violate limits at another corner. For example:

A slow corner may have a different (tighter) max_cap limit
Different PVT conditions change drive resistance, affecting slew
Without multi-corner checking, SizeDownMove can introduce violations that repair_design must later fix or worse, violations that slip through entirely if repair doesn't re-check all corners.

### Suggested Solution

Update checkMaxCapViolation() and checkMaxSlewViolation() in BaseMove.cc to iterate over all timing corners:

1. checkMaxCapViolation() : Replace the single capacitanceLimit() call with a loop over all scenes:

```cpp
bool BaseMove::checkMaxCapViolation(const sta::Pin* output_pin,
                                     sta::LibertyPort* output_port,
                                     float output_cap)
{
  // Check across all corners instead of just one
  for (const sta::Scene* scene : *sta_->scenes()) {
    float max_cap;
    bool cap_limit_exists;
    sta_->findCapLimit(output_port, scene, resizer_->max_,
                       max_cap, cap_limit_exists);
    if (cap_limit_exists && max_cap > 0.0 && output_cap > max_cap) {
      // ... existing debugPrint ...
      return true;
    }
  }
  return false;
}
```
2. checkMaxSlewViolation() : Similarly update to use sta_->findSlewLimit() across all corners.

3. SizeDownMove::downSizeGate() : Update the output cap computation to also check across corners (currently uses a single scene parameter).

The multi-corner pattern already exists in RepairDesign::checkSlew() ([L1233](https://github.com/The-OpenROAD-Project/OpenROAD/blob/master/src/rsz/src/RepairDesign.cc#L1233)) which calls sta_->checkSlew() across sta_->scenes() this can serve as a reference implementation.

I would like to work on a PR for this. @precisionmoon @maliberty @eder-matheus  please assign this to me.

### Additional Context

Impact: QoR regression in MCMM designs unchecked violations in non-active corners may persist through the flow.

Vi-shub:
Fixes #9793
### Problem
`BaseMove::checkMaxCapViolation()` and [checkMaxSlewViolation()](cci:1://file:///c:/Users/smsha/Desktop/OpenROAD/src/rsz/src/BaseMove.cc:869:0-899:1) only checked against a **single timing corner**, as noted by two FIXMEs:
- `BaseMove.cc:L840`  `// FIXME: Can we update to consider multiple corners?`
- `SizeDownMove.cc:L333`  `// FIXME: Only applies to current corner`
In multi-corner/multi-mode (MCMM) designs, a cell swap passing checks at the current corner may violate limits at another corner (e.g., slow corner has different `max_cap`).
### Changes
**`BaseMove::checkMaxCapViolation()`** replaced single-corner `output_port->capacitanceLimit()` with `sta_->checkCapacitance(pin, sta_->scenes(), ...)` which queries across all corners and returns the worst violation. Follows the pattern used by `RepairDesign::needRepairCap()`.
**`BaseMove::checkMaxSlewViolation()`** replaced single `scene` parameter with iteration over `sta_->scenes()`, checking the slew limit at every corner. Removed the `scene` parameter from the function signature.
**[SizeDownMove.cc](cci:7://file:///c:/Users/smsha/Desktop/OpenROAD/src/rsz/src/SizeDownMove.cc:0:0-0:0)** removed both FIXME comments, updated the call site.
### Testing
All existing [rsz](cci:1://file:///c:/Users/smsha/Desktop/OpenROAD/src/rsz/src/Resizer.cc:227:0-236:1) regression tests pass. The change is conservative (more checking = more restrictive), so no golden file updates needed.

### DIFF
--- a/src/rsz/src/BaseMove.cc
+++ b/src/rsz/src/BaseMove.cc
@@ -836,11 +836,11 @@ bool BaseMove::checkMaxCapViolation(const sta::Pin* output_pin,
 {
   // Check capacitance limit across all corners (not just one).
   float cap1, max_cap1, cap_slack1;
-  const sta::Scene* corner1;
+  const sta::Scene* scene1;
   const sta::RiseFall* tr1;
   sta_->checkCapacitance(
       output_pin, sta_->scenes(), resizer_->max_,
-      cap1, max_cap1, cap_slack1, tr1, corner1);
+      cap1, max_cap1, cap_slack1, tr1, scene1);
 
   debugPrint(logger_,
              RSZ,
@@ -851,7 +851,7 @@ bool BaseMove::checkMaxCapViolation(const sta::Pin* output_pin,
              max_cap1,
              output_cap);
 
-  if (max_cap1 > 0.0 && corner1 && output_cap > max_cap1) {
+  if (max_cap1 > 0.0 && scene1 && output_cap > max_cap1) {
     debugPrint(logger_,
                RSZ,
                "opt_moves",
@@ -875,12 +875,12 @@ bool BaseMove::checkMaxSlewViolation(const sta::Pin* output_pin,
   float output_res = output_port->driveResistance();
   float output_slew = output_slew_factor * output_res * output_cap;
 
-  // Check slew limit across all corners (not just one).
-  for (const sta::Scene* corner : sta_->scenes()) {
+  // Check slew limit across all scenes (not just one).
+  for (const sta::Scene* scene : sta_->scenes()) {
     float max_slew;
     bool slew_limit_exists;
     sta_->findSlewLimit(
-        output_port, corner, resizer_->max_, max_slew, slew_limit_exists);
+        output_port, scene, resizer_->max_, max_slew, slew_limit_exists);
 
     if (slew_limit_exists && output_slew > max_slew) {
       debugPrint(logger_,
--- a/src/rsz/src/BaseMove.cc
+++ b/src/rsz/src/BaseMove.cc
@@ -827,18 +827,20 @@ vector<const sta::Pin*> BaseMove::getOutputPins(const sta::Instance* inst)
       outputs.push_back(pin);
     }
   }
-
   return outputs;
 }
 
 bool BaseMove::checkMaxCapViolation(const sta::Pin* output_pin,
-                                    sta::LibertyPort* output_port,
-                                    float output_cap)
+                                     sta::LibertyPort* output_port,
+                                     float output_cap)
 {
-  float max_cap;
-  bool cap_limit_exists;
-  // FIXME: Can we update to consider multiple corners?
-  output_port->capacitanceLimit(resizer_->max_, max_cap, cap_limit_exists);
+  // Check capacitance limit across all corners (not just one).
+  float cap1, max_cap1, cap_slack1;
+  const sta::Scene* corner1;
+  const sta::RiseFall* tr1;
+  sta_->checkCapacitance(
+      output_pin, sta_->scenes(), resizer_->max_,
+      cap1, max_cap1, cap_slack1, tr1, corner1);
 
   debugPrint(logger_,
              RSZ,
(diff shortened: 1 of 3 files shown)


Vi-shub:
Hi @precisionmoon can you please review this. Thanks for giving time!!

precisionmoon:
Have you run ORFS check-in tests?

Vi-shub:
Hi @precisionmoon .Yes, I have run the ORFS check-in tests locally (via ./test/regression flow) and they all pass without any PPA regressions. Since this change uniformly makes the checkMaxCapViolation and checkMaxSlewViolation checks more restrictive by considering all corners rather than a single default corner, it avoids cell down-sizings that would violate limits in other corners and does not negatively impact the flow.

src/rsz/src/BaseMove.cc
```diff
@@ -858,38 +860,40 @@ bool BaseMove::checkMaxCapViolation(const sta::Pin* output_pin,
...
-               "opt_moves",
-               2,
-               "  skip based on max slew {} gate={} slew={} max_slew={}",
-               network_->pathName(output_pin),
-               output_port->libertyCell()->name(),
-               output_slew,
-               max_slew);
-    return true;
+  // Check slew limit across all corners (not just one).
+  for (const sta::Scene* corner : sta_->scenes()) {
```
precisionmoon: scenes are a combination of corners and modes.   We want to iterate through scene, not corner name.

src/rsz/src/BaseMove.cc
```diff
@@ -858,38 +860,40 @@ bool BaseMove::checkMaxCapViolation(const sta::Pin* output_pin,
...
+                 "opt_moves",
+                 2,
+                 "  skip based on max slew {} gate={} slew={} max_slew={}",
+                 network_->pathName(output_pin),
+                 output_port->libertyCell()->name(),
+                 output_slew,
+                 max_slew);
+      return true;
+    }
   }
```
precisionmoon, line 897: Also, why not use Sta::checkSlew() API similar to Sta::checkCapacitance()?

Vi-shub:
Hi @precisionmoon thanks for the review. I have updated the code so the variable name is scene instead of corner.

Regarding Sta::checkSlew(): We cannot use Sta::checkSlew() here because we are evaluating a hypothetical cell swap for a new candidate cell (output_port). If we call Sta::checkSlew(output_pin) or Sta::checkCapacitance(output_pin), OpenSTA queries the current network graph. This evaluates against the limits of the existing cell on that pin, which evaluates incorrectly if our new candidate cell has a different max capacitance or drive resistance. 
(Is there a recommended way to query a LibertyPort's max capacitance across all scenes?) I have been searching through but I guess needed your help :)

Issue #9858, opened 2026-03-20
Feature Request: Migrate Buildifier to Bazel Lint Umbrella


oharboe:
If we agree on this, then we can convert this into PR at a click of a button.

## Description

Buildifier (format + lint for BUILD, .bzl, and MODULE.bazel files) currently
runs as a standalone GitHub Actions workflow
(`.github/workflows/buildifier.yaml`) that downloads its own binary and
caches it independently. This should be migrated to the Bazel lint umbrella
(`//:lint_test` and `//:fix_lint`) for consistency with how TCL linting
already works.

### Current state

The GitHub Actions workflow does two things:

```bash
# Format check
./buildifier -r -mode=check -lint=off .

# Lint check
./buildifier -r -lint=warn .
```

The tool version (`v8.2.1`) is pinned in the workflow YAML, not in Bazel's
dependency management. A `.buildifier.json` config already exists at the
repo root.

### Problem

1. **Cannot run locally** — developers must push and wait for CI to discover
   buildifier violations. This violates
   [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment):
   `bazelisk run //:fix_lint` should fix everything it can and report the
   rest, but currently it knows nothing about BUILD files.

2. **Version drift** — the buildifier version is pinned in GitHub Actions
   YAML, separate from Bazel's hermetic dependency management
   (`bazel/requirements.in`, `MODULE.bazel`). It can drift out of sync
   with what developers run locally (if they run it at all).

3. **Inconsistent pattern** — TCL linting already uses the Bazel lint
   umbrella. Every additional language that stays outside the umbrella is
   a special case that new contributors have to discover separately.

### Proposed targets

Per the [naming convention](LintTargets.md#target-naming-convention):

| Target | What it does |
|--------|-------------|
| `//:lint_bzl_test` | `buildifier -r -mode=check -lint=warn .` (read-only) |
| `//:fmt_bzl_test` | `buildifier -r -mode=check -lint=off .` (format only) |
| `//:tidy_bzl` | `buildifier -r -mode=fix -lint=fix .` (auto-fix) |

Wire into umbrellas:
- Add `lint_bzl_test` and `fmt_bzl_test` to `//:lint_test`
- Add buildifier fix step to `bazel/fix_lint.sh`

## Suggested Solution

1. Add buildifier as a Bazel-managed tool dependency (it's already available
   as a `@buildifier` Bazel module)
[...]

PR #9865, opened 2026-03-21
bazel: Migrate Buildifier to Bazel Lint Umbrella (#9858)

Divinesoumyadip:
Fixes  Issue #9858

This migrates Buildifier into the Bazel lint umbrella, following the same pattern already established for TCL linting. Previously, Buildifier only ran as a standalone GitHub Actions workflow, meaning developers had to push to CI just to discover formatting violations in BUILD and .bzl files.
The change adds `buildifier_prebuilt` (v8.2.1) as a Bzlmod dev dependency and introduces three new shell scripts  `bzl_fmt_test.sh`, `bzl_lint_test.sh`, and `bzl_tidy.sh` , mirroring the existing TCL lint pattern exactly. The two read-only check targets (`fmt_bzl_test` and` lint_bzl_test`) are wired into the `//:lint_test` umbrella, and `fix_lint` has been updated to run both `tidy_bzl` and `tidy_tcl` so developers get a single command that auto-fixes both Starlark and TCL files locally. The old `.github/workflows/buildifier.yaml `workflow has been retired. Both new tests pass locally.


bazel/bzl_tidy.sh
```diff
@@ -0,0 +1,12 @@
...
+BUILDIFIER_BIN=$(realpath "$1")
+
+if [ -z "$BUILD_WORKSPACE_DIRECTORY" ]; then
+  echo "Error: This script must be run via 'bazelisk run'"
+  exit 1
+fi
+
+cd "$BUILD_WORKSPACE_DIRECTORY"
+"$BUILDIFIER_BIN" -r -mode=fix -lint=fix .
+echo "Buildifier tidy complete!"
```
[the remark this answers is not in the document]
Divinesoumyadip: Thanks for the suggestion. The `git ls-files` approach fails in Bazel's sandboxed test environment with` fatal: not a git repository` (exit 128) since the sandbox doesn't have access to the` .git` directory. The `tags = ["local"]` has been added for consistency with the TCL pattern, and the` -r` recursive flag on buildifier correctly discovers all Starlark files without needing git. Both tests pass locally.

Divinesoumyadip:
@maliberty all CI checks including Jenkins are passing. Would appreciate a review when you get a chance!

oharboe:
Nice! I have not reviewed it, but just a note on buildifier versions and operating systems. I am not sure why the MODULE.bazel.lock is in git, but I think it is a security issue, users can tell bazel to check for supply chain attacks.

I wonder if it is possible to make the lint check ignore non security concerns in .lock diffs. Would require some unittests in python also under CI tests as such a script cant be checked on any specific dev machine and would need to be curated to encode all these headaches. The upside to making this script is reduced cognitive load of contributors.

oharboe:
I usually just point Claude at the PR and ask it to create new commits to address concerns in the PR, which it happily does. 

Except when Gemmini and Claude disagrees, in which case I ask Claude to give a snarky explanation why Gemmini is hallucinating with a record of tests or evidence that Gemmini is wrong.

Divinesoumyadip:
> Nice! I have not reviewed it, but just a note on buildifier versions and operating systems. I am not sure why the MODULE.bazel.lock is in git, but I think it is a security issue, users can tell bazel to check for supply chain attacks.
> 
> I wonder if it is possible to make the lint check ignore non security concerns in .lock diffs. Would require some unittests in python also under CI tests as such a script cant be checked on any specific dev machine and would need to be curated to encode all these headaches. The upside to making this script is reduced cognitive load of contributors.

Thanks @oharboe! The `MODULE.bazel.lock` is standard Bzlmod practice for reproducible builds and supply chain verification. The lock diff noise in lint is a good point i think.

bazel/bzl_fmt_test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
```
luarss:
```suggestion
set -euo pipefail
```

bazel/bzl_fmt_test.sh
```diff
@@ -0,0 +1,7 @@
...
+BUILDIFIER_BIN=$(realpath "$1")
```
luarss: is `realpath` portable across all of our supported platforms?

bazel/bzl_fmt_test.sh
```diff
@@ -0,0 +1,7 @@
...
+if [ -n "${BUILD_WORKSPACE_DIRECTORY:-}" ]; then
```
luarss: Refer to how other `.sh` locates WORKSPACE, let's keep things consistent.

luarss (changes requested):

Divinesoumyadip:
@luarss I've addressed all three suggestions in the latest commits.

Please let me know if anything else needs changing.

PR #9867, opened 2026-03-21
etc: fix macOS dependency installer for googletest and pyqt

Divinesoumyadip:
Fixes three bugs in the macOS dependency installer that caused build failures for new contributors on macOS.
`googletest `was missing from the Homebrew package list which caused CMake to fail with `Could NOT find GTest` during the build. Adding `googletest` to the brew install line fixes this directly. The `pyqt5` formula name is also incorrect ,the valid Homebrew formula is pyqt and using the wrong name causes a silent failure during dependency installation leaving GUI support broken on macOS. Finally` --break-system-packages` has been added to the `pip3 install click` command which is required on macOS with Python 3.12+ where pip enforces system package protection by default and fails without this flag.
Related Issues: Addresses macOS build failures reported by contributors building locally on Apple Silicon.

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -938,7 +938,8 @@ EOF
     fi
     log "Install darwin base packages using homebrew (-base or -all)"
     _execute "Installing Homebrew packages..." brew install bison boost bzip2 cmake eigen flex fmt groff googletest libomp or-tools pandoc pkg-config pyqt python spdlog tcl-tk zlib swig yaml-cpp
-    _execute "Installing Python click..." pip3 install click --break-system-packages
+    _execute "Installing pipx..." brew install pipx
+    _execute "Installing Python click..." pipx install click
     _execute "Linking libomp..." brew link --force libomp
     _execute "Installing lemon-graph..." brew install The-OpenROAD-Project/lemon-graph/lemon-graph
 }


etc/DependencyInstaller.sh
```diff
@@ -937,8 +937,28 @@ EOF
...
+        pandoc \
+        pkg-config \
+        pyqt \
+        python \
+        spdlog \
+        swig \
+        tcl-tk \
+        yaml-cpp \
+        zlib
+    _execute "Installing Python click..." pip3 install click --break-system-packages
```
luarss: Not sure about `--break-system-packages`, seems risky. Any alternatives?

Divinesoumyadip: > Not sure about `--break-system-packages`, seems risky. Any alternatives?

Thanks for the feedback! You're right, `--break-system-packages` is a blunt approach. I can replace it with pipx install click instead , that's cleaner for Homebrew environments. Will update the PR.

Divinesoumyadip:
> clang-tidy review says "All clean, LGTM! 👍"

Replaced pip3 install click `--break-system-packages` with `pipx install click` cleaner for Homebrew environments. Also added a macOS fallback for `nproc `using `sysctl -n hw.logicalcpu`.

@luarss Kindly check and give me review .

etc/DependencyInstaller.sh
```diff
@@ -937,8 +941,28 @@ EOF
...
+        pandoc \
+        pkg-config \
+        pyqt \
+        python \
+        spdlog \
+        swig \
+        tcl-tk \
+        yaml-cpp \
+        zlib
+    _execute "Installing Python click..." pipx install click
```
luarss: whats `pipx`? does it come installed with system python?

etc/DependencyInstaller.sh
```diff
@@ -937,8 +941,28 @@ EOF
         exit 1
     fi
     log "Install darwin base packages using homebrew (-base or -all)"
-    _execute "Installing Homebrew packages..." brew install bison boost bzip2 cmake eigen flex fmt groff libomp or-tools pandoc pkg-config pyqt5 python spdlog tcl-tk zlib swig yaml-cpp
-    _execute "Installing Python click..." pip3 install click
+    _execute "Installing Homebrew packages..." brew install \
```
luarss: you can revert this. I think gemini's suggestion need not be followed for this line. Packages can be on a single line

etc/DependencyInstaller.sh
```diff
@@ -16,7 +16,11 @@ set -euo pipefail
 PREFIX=""
 CI="no"
 SAVE_DEPS_PREFIXES=""
-NUM_THREADS=$(nproc)
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
luarss: revert this as well please. Keep the PR clean, focused on one objective

etc/DependencyInstaller.sh
Divinesoumyadip: > whats `pipx`? does it come installed with system python?

`pipx` is not always pre-installed. I can use `pip3 install click` inside a venv instead, or add` brew install pipx` before the pip call. Which do you prefer?

luarss: Not sure which one will have the same global `python3-click` effect. Can you investigate?

Divinesoumyadip: > Not sure which one will have the same global `python3-click` effect. Can you investigate?

I investigated  so brew install click installs a Kubernetes CLI tool, not Python's click library. It's not available as a Homebrew formula. The cleanest global solution is brew install pipx && pipx install click also, pipx is available via Homebrew and installs click in an isolated but globally accessible environment. Want me to update with this?

etc/DependencyInstaller.sh
Divinesoumyadip: > revert this as well please. Keep the PR clean, focused on one objective

Reverted the multiline formatting and nproc changes . PR is now focused only on the `googletest `and `pyqt `fixes. Also updated click installation to use brew install `pipx` &&` pipx install` click since python-click is not available as a Homebrew formula (brew install click installs a Kubernetes tool, not Python's click).

etc/DependencyInstaller.sh
luarss: Yes please.

src/grt/src/cugr/src/CUGR.cpp
```diff
@@ -229,7 +228,9 @@ void CUGR::mazeRoute(std::vector<int>& net_indices)
     maze_route.constructSparsifiedGraph(wire_cost_view, grid);
     maze_route.run();
     std::shared_ptr<SteinerTreeNode> tree = maze_route.getSteinerTree();
-    assert(tree != nullptr);
+    if (tree == nullptr) {
```
luarss: why is this diff here?

Divinesoumyadip:
@maliberty kindly review when you get a chance. luarss has approved. Jenkins pr-merge failure is a pre-existing CI issue unrelated to this change.

PR #9871, opened 2026-03-21
refactor:remove stale packages from dependency list

### LINKED ISSUE #9868: Check potential unused deps
### Description

`python3-click, qimgv, default-jdk, lsb-release` seems to be stale on dep list.

### Suggested Solution

_No response_

### Additional Context

_No response_

alokkumardalei-wq:
### What does this PR do?
Fixes #9868.
This PR cleans up the dependency installer scripts by removing a few outdated packages that OpenROAD doesn't actually use anymore, exactly as reported in #9868.

Specifically, it removes:
- `python3-click`
- `qimgv`
- `default-jdk`
- `lsb-release`
Before removing them, I ran a full search across the entire OpenROAD codebase and verified that there is zero source code or scripts currently relying on these packages. 

## Potential Impact
Removing these will help speed up the Docker builds and save local developers from downloading hundreds of megabytes of unnecessary software (especially the Java Development Kit!).

### DIFF
--- a/etc/DependencyInstaller.sh
+++ b/etc/DependencyInstaller.sh
@@ -982,7 +982,7 @@ _install_ci_packages() {
     log "Install CI dependencies (-ci)"
     _execute "Updating package lists..." apt-get -y update
     _execute "Installing CI packages..." apt-get -y install --no-install-recommends \
-        apt-transport-https ca-certificates curl gnupg jq parallel \
+        apt-transport-https ca-certificates curl gnupg jq lsb-release parallel \
         python3 python3-pandas python3-pip software-properties-common \
         time unzip zip
 


sombraSoft:
Actually the only reason I think we would need `default-jdk` is for the -ci option in which it installs `bazelisk` which needs the sdk.
@alokkumardalei-wq, if you rather move the `default-jdk` package install to the `_install_ci_packages` function that would make more sense.

maliberty:
Why does bazelisk need a jdk?

sombraSoft:
not actually bazelisk but bazel does! if the goal of this script is to setup bazel too we need jdk, no?

maliberty:
Why does bazel need a jdk?

alokkumardalei-wq:
Hello @sombraSoft thank you for the feedback but I think @maliberty asked the right question since I think (I might be wrong here) modern versions of Bazel actually bundle their own internal hermetic JRE inside the binary for execution. Since OpenROAD is a C++/TCL codebase and doesn't compile Java targets, we do not need to install the heavy `default-jdk` system package at all, not even for CI I think. It can be safely completely removed as a stale dependency.

Please let me know if I missed something here.
Thank you!

sombraSoft:
Well, I guess I couldn't expect less from Bazel 😅
Of course they would self contain their own runtime dependencies - apparently bazel builds itself.
Good find!

etc/DependencyInstaller.sh
```diff
@@ -982,9 +982,9 @@ _install_ci_packages() {
     log "Install CI dependencies (-ci)"
     _execute "Updating package lists..." apt-get -y update
     _execute "Installing CI packages..." apt-get -y install --no-install-recommends \
-        apt-transport-https ca-certificates curl default-jdk gnupg python3 \
-        python3-pip python3-pandas jq lsb-release parallel \
```
vvbandeira, line 964: Please keep `python3-pip lsb-release` they are required for some CI tools

maliberty:
The request to keep lsb-release needs to be addressed.

PR #9874, opened 2026-03-22
mpl: consolidate cluster type booleans into ClusterType enum

### LINKED ISSUE #9463: mpl: update cluster types
### Description

Clusters can be of 3 types nowadays (standard cell, mixed and hard macro) while having many `isSomething` methods to cover IOs, pins and specialized macro structures.
```
  bool is_cluster_of_unplaced_io_pins_{false};
  bool is_cluster_of_unconstrained_io_pins_{false};
  bool is_io_pad_cluster_{false};
  bool is_io_bundle_{false};
  bool is_array_of_interconnected_macros_{false};
  bool is_macro_array_{false};
```

### Suggested Solution

Consolidate the many implicit cluster types under the ClusterType enum instad of relying on `isSomething` methods.

### Additional Context

_No response_

alokkumardalei-wq:
## What does this PR do?
Fixes #9463
Consolidate the many implicit cluster types under the ClusterType enum instead of relying on isSomething methods, as suggested in #9463.
The new enum class ClusterType has 9 values: StdCell, Macro, Mixed, ConstrainedIOs, UnconstrainedIOs, IOBundle, PAD, MacroArray, InterconnectedMacrosArray.

## Impact
This change improves type safety (scoped enum class vs plain enum), reduces the number of member variables in Cluster, and makes cluster categorization explicit and centralized.

### DIFF
--- a/src/mpl/src/graphics.cpp
+++ b/src/mpl/src/graphics.cpp
@@ -178,14 +178,14 @@ void Graphics::fetchSoftAndHard(Cluster* parent,
 
   for (auto& child : parent->getChildren()) {
     switch (child->getClusterType()) {
-      case HardMacroCluster: {
+      case ClusterType::Macro: {
         std::vector<mpl::HardMacro*> hard_macros = child->getHardMacros();
         for (HardMacro* hard_macro : hard_macros) {
           hard.push_back(*hard_macro);
         }
         break;
       }
-      case StdCellCluster: {
+      case ClusterType::StdCell: {
         if (child->isLeaf()) {
           soft.push_back(*child->getSoftMacro());
         } else {
@@ -193,10 +193,12 @@ void Graphics::fetchSoftAndHard(Cluster* parent,
         }
         break;
       }
-      case MixedCluster: {
+      case ClusterType::Mixed: {
         fetchSoftAndHard(child.get(), hard, soft, outlines, (level + 1));
         break;
       }
+      default:
+        break;
     }
   }
 }
@@ -682,9 +684,9 @@ void Graphics::setSoftMacroBrush(gui::Painter& painter,
     return;
   }
 
-  if (soft_macro.getCluster()->getClusterType() == StdCellCluster) {
+  if (soft_macro.getCluster()->getClusterType() == ClusterType::StdCell) {
     painter.setBrush(gui::Painter::kDarkBlue);
-  } else if (soft_macro.getCluster()->getClusterType() == HardMacroCluster) {
+  } else if (soft_macro.getCluster()->getClusterType() == ClusterType::Macro) {
     // dark red
     painter.setBrush(gui::Painter::Color(0x80, 0x00, 0x00, 150));
   } else {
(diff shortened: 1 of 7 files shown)


src/mpl/src/graphics.cpp
```diff
@@ -178,25 +178,27 @@ void Graphics::fetchSoftAndHard(Cluster* parent,
...
         }
         break;
       }
-      case MixedCluster: {
+      case ClusterType::Mixed: {
         fetchSoftAndHard(child.get(), hard, soft, outlines, (level + 1));
         break;
       }
+      default:
+        break;
```
AcKoucher: Don't add a default case that way the compiler will complain if any value is uncovered.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -1651,20 +1651,23 @@ bool HierRTLMP::validUtilization(
...
       }
-      case MixedCluster: {
+      case ClusterType::Mixed: {
         const TilingList& tilings = cluster->getTilings();
         mixed_cluster_macro_area += tilings.front().area();
         mixed_cluster_std_cell_area += cluster->getStdCellArea();
+        break;
       }
+      default:
+        break;
```
AcKoucher: Ditto.

src/mpl/src/hier_rtlmp.cpp
```diff
@@ -2674,20 +2677,23 @@ bool Pusher::designHasSingleCentralizedMacroArray()
...
         // we use the area of its SoftMacro abstraction, because the
         // Cluster::getArea() will give us the actual std cell area
         // of the instances from that cluster.
         if (child->getSoftMacro()->getArea() != 0) {
           return false;
         }
+        break;
       }
+      default:
+        break;
```
AcKoucher: Ditto.

src/mpl/src/object.h
```diff
@@ -73,11 +73,17 @@ using BundledNetList = std::vector<BundledNet>;
...
+  StdCell,
+  Macro,
+  Mixed,
+  ConstrainedIOs,
+  UnconstrainedIOs,
+  IOBundle,
+  PAD,
+  MacroArray,
+  InterconnectedMacrosArray
 };
```
AcKoucher: This belongs to the `Cluster` class and should be called `Type`. That way, we'll have something like `if (cluster->getType() == Cluster::Type::Macro)` externally and `type_ == Macro` internally which is both more readable and concise.

AcKoucher (changes requested):
@alokkumardalei-wq I appretiate you working on this.

I think you tried rebasing your branch and something went wrong - note that there are changes in `dbSta`.

Many regression tests are failing and the BAZEL mock-array flow as well (see the head/merge check pipelines in the PR). This PR should not change any results. Note that, with the changes here, all the "custom" clusters that were handled with bools will need to be explicitly handled, so I suspect the results differences may come from that.

AcKoucher:
@alokkumardalei-wq I see you're pushing changes. Please, let me know when this is ready for another review.

PR #9884, opened 2026-03-23
Bump bazel-orfs to 272cc07, add bazel-orfs-verilog dep

oharboe:
bazel-orfs moved generate.bzl and verilog.bzl into the bazel-orfs-verilog submodule.

bazel-orfs-verilog will eventually be replaced by some functionality in bazel bcr, this is part of putting the bazel-orfs-verilog stray kitten in bazel-orfs up for adoption.

@MrAMS FYI


oharboe:
@maliberty I think it would good to have this on the books before https://github.com/The-OpenROAD-Project/OpenROAD/pull/9827 

Just to make any triage, what could possibly go wrong?, easier....

PR #9909, opened 2026-03-24
test: add power reporting test for sta standalone and openroad

oharboe:
Add a minimal test/orfs/power module demonstrating how to write a Tcl power reporting script that works with both standalone OpenSTA and OpenROAD. Without loading technology LEF files, OpenROAD fails with "ORD-2010 no technology has been read" at link_design. The read_lefs proc conditionally loads LEFs when running in OpenROAD.

Also expose gcd targets to test/orfs subpackages via default_visibility.

https://github.com/oharboe/openroad-demo/blob/main/ideas/per-command-openroad-binaries.md

### DIFF
--- a/test/orfs/power/ok.sh
+++ b/test/orfs/power/ok.sh
@@ -1,8 +1,4 @@
 #!/usr/bin/env bash
 #
 # Check if contents of $1 is equal to "OK"
-set -e
-if [ "$(cat "$1")" == "OK" ]; then
-  exit 0
-fi
-exit 1
+[ "$(cat "$1")" == "OK" ]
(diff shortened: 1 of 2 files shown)


maliberty:
Are we running all of gcd just to verify `if { [info exists ::env(TECH_LEF)] } {` ?

oharboe:
> Are we running all of gcd just to verify `if { [info exists ::env(TECH_LEF)] } {` ?

No. It is already being built, so we are using it: free.

oharboe:
out of sight, out of mind for now. I may have an oportunity to revisit later. We ran into slang name mangling issues(reported) + uniqification in hierarhical synthesis that have to be sorted out before we can switch from sta standalone to using openroad+loading lef files to do power reporting. :sweat:

PR #9910, opened 2026-03-24
Standalone Per-Command OpenROAD Binaries

oharboe:
17 standalone `cc_binary` targets — each calls the C++ API directly.

No TCL interpreter, no framework, no SWIG, no `InitOpenRoad()`.

## The Debug Cycle Problem

Editing one line in `src/drt/src/FlexDR.cpp` today requires relinking all 37 modules (3,788 source files). Then debugging goes through TCL.

```
# Today: edit → rebuild everything → debug via TCL
bazelisk build //:openroad                                    # 3,788 source files
cat > /tmp/script.tcl <<< "read_db crash.odb; detailed_route -verbose 1"
gdb --args bazel-bin/openroad -exit /tmp/script.tcl

# With this PR: edit → rebuild only drt → debug directly
bazelisk build //test/orfs/openroad:openroad-detailed_route   # 1,165 source files
gdb --args bazel-bin/.../openroad-detailed_route --read_db crash.odb -verbose 1
```

Incremental rebuild after editing one `.cpp`: **~10 seconds**.

## Available Binaries

Naming follows the `git-{cmd}` convention (principle of least astonishment).

| Binary | Source files | vs monolithic |
|--------|-------------|---------------|
| `openroad-detailed_placement` | 1,095 | 3.5x fewer |
| `openroad-check_placement` | 1,095 | 3.5x fewer |
| `openroad-filler_placement` | 1,095 | 3.5x fewer |
| `openroad-optimize_mirroring` | 1,095 | 3.5x fewer |
| `openroad-detailed_route` | 1,165 | 3.2x fewer |
| `openroad-check_antennas` | 1,061 | 3.6x fewer |
| `openroad-extract_parasitics` | 1,102 | 3.4x fewer |
| `openroad-tapcell` | 1,059 | 3.6x fewer |
| `openroad-density_fill` | 1,132 | 3.3x fewer |
| `openroad-global_route` | 1,160 | 3.3x fewer |
| `openroad-repair_design` | 1,184 | 3.2x fewer |
| `openroad-global_placement` | 1,755 | 2.2x fewer |
| `openroad-clock_tree_synthesis` | 1,193 | 3.2x fewer |
| `openroad-init_floorplan` | 1,066 | 3.6x fewer |
| `openroad-place_pins` | 1,137 | 3.3x fewer |
| `openroad-macro_placement` | 1,632 | 2.3x fewer |
| `openroad-analyze_power_grid` | 1,211 | 3.1x fewer |
| *monolithic openroad* | *3,788* | *baseline* |

## Correctness

Integration test (`integration_test.py`) runs each command two ways —
standalone binary and monolithic `openroad -exit` — then diffs the
output ODB:

- `openroad-detailed_placement`: **bit-for-bit identical** to monolithic
- `openroad-optimize_mirroring`: **bit-for-bit identical** to monolithic
- `openroad-detailed_route`: both produce valid routed ODB (thread-dependent ordering)
- `openroad-global_route`: both produce valid routed ODB (thread-dependent ordering)

## Key Finding: `dbSta(nullptr, db, &logger)` works

The `Tcl_Interp*` parameter is optional — `initVars` already checks
`if (tcl_interp)` before calling `setTclInterp()`. This unlocks all
sta-dependent commands for standalone use without any upstream changes.

## Known Workarounds

All workarounds are local to `test/orfs/openroad/`. Zero churn in
other modules.

**`_exit(0)` in sta-dependent binaries:** `sta::ReportTcl::~ReportTcl()`
calls `Tcl_UnstackChannel()` on nullptr `Tcl_Interp`, segfaulting during
cleanup. Routing/placement completes correctly and output is flushed
before exit. Proper fix: make `ReportTcl` destructor null-safe.

**No-op `AbstractGraphicsFactory` in detailed_route:**
`TritonRoute::initDesign()` calls `graphics_factory_->reset()`
unconditionally. Proper fix: null-check in `initGraphics()`.

## Bug Reports Become Trivial

```
Steps to reproduce:
1. Download crash.odb
2. bazelisk run //test/orfs/openroad:openroad-detailed_route -- --read_db crash.odb
```

No environment variables. No TCL script. Files + one command.
Copy-paste directly into `gdb --args`.

### DIFF
--- a/test/orfs/openroad/analyze_power_grid_main.cc
+++ b/test/orfs/openroad/analyze_power_grid_main.cc
@@ -11,6 +11,7 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <stdexcept>
 #include <string>
 #include <vector>
 
@@ -60,40 +61,45 @@ void usage(const char* prog)
 
 bool parse_args(int argc, char* argv[], Options& opts)
 {
-  for (int i = 1; i < argc; i++) {
-    std::string arg = argv[i];
-    auto next = [&]() -> std::string {
-      if (i + 1 >= argc) {
-        std::cerr << "Error: " << arg << " requires an argument\n";
-        std::exit(1);
-      }
-      return argv[++i];
-    };
+  try {
+    for (int i = 1; i < argc; i++) {
+      std::string arg = argv[i];
+      auto next = [&]() -> std::string {
+        if (i + 1 >= argc) {
+          throw std::invalid_argument(std::string(arg)
+                                      + " requires an argument");
+        }
+        return argv[++i];
+      };
 
-    if (arg == "--read_db") {
-      opts.read_db = next();
-    } else if (arg == "--read_lef") {
-      opts.read_lef.push_back(next());
-    } else if (arg == "--read_def") {
-      opts.read_def = next();
-    } else if (arg == "-net") {
-      opts.net = next();
-    } else if (arg == "-voltage_file") {
-      opts.voltage_file = next();
-    } else if (arg == "-em_file") {
-      opts.em_file = next();
-    } else if (arg == "-error_file") {
-      opts.error_file = next();
-    } else if (arg == "-enable_em") {
-      opts.enable_em = true;
-    } else if (arg == "-help" || arg == "--help" || arg == "-h") {
-      usage(argv[0]);
-      std::exit(0);
-    } else {
-      std::cerr << "Error: unknown argument: " << arg << "\n";
-      usage(argv[0]);
-      return false;
+      if (arg == "--read_db") {
+        opts.read_db = next();
+      } else if (arg == "--read_lef") {
+        opts.read_lef.push_back(next());
+      } else if (arg == "--read_def") {
+        opts.read_def = next();
+      } else if (arg == "-net") {
+        opts.net = next();
+      } else if (arg == "-voltage_file") {
+        opts.voltage_file = next();
+      } else if (arg == "-em_file") {
+        opts.em_file = next();
+      } else if (arg == "-error_file") {
+        opts.error_file = next();
+      } else if (arg == "-enable_em") {
+        opts.enable_em = true;
+      } else if (arg == "-help" || arg == "--help" || arg == "-h") {
+        usage(argv[0]);
+        std::exit(0);
+      } else {
+        std::cerr << "Error: unknown argument: " << arg << "\n";
+        usage(argv[0]);
+        return false;
+      }
     }
+  } catch (const std::invalid_argument& e) {
+    std::cerr << "Error: " << e.what() << '\n';
+    return false;
   }
 
   if (opts.read_db.empty()) {
--- a/test/orfs/openroad/check_antennas_main.cc
+++ b/test/orfs/openroad/check_antennas_main.cc
@@ -10,6 +10,7 @@
 #include <cstdlib>
 #include <fstream>
 #include <iostream>
+#include <stdexcept>
 #include <string>
 #include <vector>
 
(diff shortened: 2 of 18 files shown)


oharboe:
@maliberty @osamahammad21 @hzeller 


# Per-Command OpenROAD Binaries: Debug Cycle Improvement

## Compilation Units

| Target | Source files | vs monolithic |
|--------|-------------|---------------|
| `detailed_placement` | 1,095 | **3.5x fewer** |
| `check_placement` | 1,095 | **3.5x fewer** |
| `filler_placement` | 1,095 | **3.5x fewer** |
| `optimize_mirroring` | 1,095 | **3.5x fewer** |
| `detailed_route` | 1,165 | **3.2x fewer** |
| monolithic `openroad` | 3,788 | baseline |

The ~1,100 shared baseline is `odb` + `sta` (core libraries every module
needs). The monolithic binary adds **2,600+ files** from 34 modules you
don't need when debugging one command.

## Debug Build Cycle

```
                          monolithic    standalone
                          openroad      detailed_route
                          ─────────     ──────────────
Source files              3,788         1,165
Initial debug build       ~15 min       ~2 min
Incremental rebuild       relink all    ~10 sec
gdb invocation            indirect      direct
                          (via TCL)     (C++ main)
```

## The Workflow

```bash
# 1. Edit the code
vim src/drt/src/FlexDR.cpp

# 2. Rebuild (seconds)
bazelisk build //test/orfs/openroad:detailed_route

# 3. Debug directly
gdb --args bazel-bin/test/orfs/openroad/detailed_route \
  --read_db grouted.odb -verbose 1

# 4. Repeat from step 1
```

## Before vs After

**Before** (monolithic):
```bash
bazelisk build //:openroad          # relinks 37 modules
cat > /tmp/script.tcl << 'EOF'      # write a TCL script
read_db grouted.odb
detailed_route -verbose 1
EOF
gdb --args bazel-bin/openroad -exit /tmp/script.tcl
```

**After** (standalone):
```bash
bazelisk build //test/orfs/openroad:detailed_route   # links 5 modules
gdb --args bazel-bin/.../detailed_route --read_db grouted.odb -verbose 1
```

## Test Results

| Binary | Test design | Result |
|--------|-------------|--------|
| `detailed_placement` | gcd Nangate45 | 294 cells, 0 failures |
| `check_placement` | gcd Nangate45 | Violations detected / clean pass |
| `filler_placement` | gcd Nangate45 | 2,092 fillers placed |
| `optimize_mirroring` | gcd Nangate45 | 134 mirrored, -1% HPWL |
| `detailed_route` | gcd ASAP7 | 20s, 3,623 vias, exit 0 |

oharboe:
@hzeller Thoughts?

oharboe:
@maliberty FYI I found an unrelated pr-head failure https://github.com/The-OpenROAD-Project/OpenROAD/issues/9912

oharboe:
@hzeller Thoughts on reworking the integration tests so that they don't have to build openroad to iterate on a specific problem but use a much narrower binary. I'm rebuilding all the time here. Different machines, folders, etc. 

ccache isn't good enough for bazel :cry:

hzeller:
I think having small binaries for some aspects sounds like a good idea, in partiuclar for quick turn around times. But of course it proliferates the number of little binaries, and then we still need one monolithic one in the end for easier end-user experience. Maybe we can pull some bazel tricks to avoid being repetetive in doing so, but can have a list-comprehension that builds a bunch of `cc_binary()`-s ?

I have not found that the final binary links particularly slowly though, maybe less than a second for the full `:openroad` binary ?

OTOH re-building a whole bunch of *.cc files if there is one changed, because they all happen to be in one big `cc_library()` is a current problem that should be addressed by making libraries smaller.

The other issue is: the tests we are talking about are all integration tests, which is why they need the binary, but actually are meant to test a small functionality.
If the tests are not using the binary, but are written in c++ and linking only the functionality and test exactly that with a `cc_test()`, this would improve the situation also a whole lot. The tcl tests could still be there, but the easier reasoning and super-fast turn-around times then can be done with the `cc_tests()`. Right now, we have only 64 `cc_tests()` and over a thousand tcl/python tests, that are integration tests essentially. I think that ratio should be the opposite.

So making more binaries and making them smaller is a good stop-gap, but we need to get to a point where we have full and reliable test coverage with `cc_test()`s, _in particular_ as this will help to confidently modify and refactor code (I see you use Claude more these days ... AI tools are better if they have a good test harness. Right now, there is essentially none).

So I think

  * very short term: investigating making small targeted binaries but with using bazel tricks sounds good.
  * but also starting immediately: breaking libraries into smaller chunks. Currently we have 10000 LoC per library, that should be more 400-800 LoC (one header, one *.cc file, rarely more).
  * Test most functionality with `cc_test()`, making sure we have full lfunctional and branch coverage. Use the tcl/python tests mostly for integration (load a file, do something, emit) as and additional end-to-end testing, but not load-bearing for functionality. Looking at a source file, abstracting,  figuring out what good things to test and writing tests for it is an excellent use of AI tools.
  
Other than that, I like to have the following in my `~/.bazelrc` or `user.bazelrc` for improved caching and faster build times.

```
# mkdir this directory first
build --disk_cache=/path/to/my/home/directory/.bazel-disk-cache --experimental_disk_cache_gc_max_age=30d
common --force_pic  # reducing times a file is built; see https://github.com/google/xls/issues/1936
```

oharboe:
Thank you! Will absorb this, but it takes time...

Right now, why isn't `common --force_pic ` on in OpenROAD?

hzeller:
It might (or might not) have a tiny impact on performance, even if the final binary is mostly statically linked. So this is why it probably makes sense to measure impact first.

oharboe:
My main problem currently is just the total cognitive load... Anything that makes things faster that I don't have to think about would reduce cognitive load enormously because it solves a) I don't have to think about it b) things go faster so I don't forget what I'm working on when I switch between tasks.

maliberty:
@hzeller I don't follow
> re-building a whole bunch of *.cc files if there is one changed, because they all happen to be in one big cc_library() is a current problem that should be addressed by making libraries smaller.

Only modified files should rebuild regardless of how the are organized into libraries.  The library itself will rebuild but ar takes near zero time.

oharboe:
I think something is wrong with caching and rebuilding, but I have no idea what, and I have nothing to feed Claude. Seems to rebuild *all the time*.

I think what is missing is some sort of "credhelper" capability where .bazelrc queries the environment for which disk-cache to use: zero configuration 100% secure.

Now there's 100% security that the user disables in frustration.

maliberty:
Are you going to do this for every command?  You would have thousands of binaries.

The number of source files is not an important metric, compile time is.  My experience matches @hzeller in that compile/link time in the incremental loop is not a particular bottleneck.

How would you pass arguments to these executables as these commands have switches?  How would you generate such from ORFS or the like?

Not too excited on first pass.

hzeller:
Using the cache (I use the local `--disk_cache`), I also can even `bazel clean --expunge` or check out a fresh OpenROAD in a separate directory and it rebuilds quickly due to the cache.
If that is not working, something else is wrong in the set-up.

oharboe:
@hzeller What is the solution here if I want a detailed router binary and that is not exposed as a binary by openroad repository. Would openroad have to expose the cc_library()'s to support my use-case of creating a binary on my end as the consumer of openroad?

compile times are very painful for me and I have no idea why everything is compiling all the time and having a much narrower binary would ease my pain.

I can see how OpenROAD developers, even if they can save build/link time by 50%, would prefer to have a simpler mental model with a single openroad binary that is familar. I'm also pretty sure a lot wouldn't switch from split binaries to a shared binary: momentum and cognitive load of churn...

hzeller:
Yes, having functionality specific cc_library()-ies would probably be good for this. These well-defined libraries then can have a visibility of public so that projects linking `@openroad` can access and use them (e.g. to build a specific binary)

oharboe:
I did create a list from history of the most often changing binaries in the list of standalone binaries I made here. It cut compilation time for those binaries in half for a full build and more for incremental builds. However, I think the distance between the full build of openroad and full/partial build of, say, detail_route will improve as things progresses on whittling down the giant modules, but I don't have any data to back up that hunch.

The question is whether someone working on detail_place/route, etc. would in fact use these standalone binaries or prefer openroad giant binary.

oharboe:
Well... there *are* consumers of standalone binaries other than openroad: C++ unit-tests. So everybody loves the speed, they just don't want the cognitive load of more a "detail_route" binary.

maliberty:
Not of interest to me.

PR #9923, opened 2026-03-24
sta: fix read_power_activities segfault when no design loaded

### LINKED ISSUE #5384: read_power_activities SEGFAULT
### Describe the bug

1. untar [crash.tar.gz](https://github.com/user-attachments/files/16243562/crash.tar.gz)
2. `./run-me-mock-array-asap7-base.sh`

Crash:

```
OpenROAD v2.0-14511-gedb4e7e37 
[INFO ODB-0227] LEF file: /home/oyvind/OpenROAD-flow-scripts/flow/platforms/asap7/lef/asap7_tech_1x_201209.lef, created 24 layers, 9 vias
[INFO ODB-0227] LEF file: /home/oyvind/OpenROAD-flow-scripts/flow/platforms/asap7/lef/asap7sc7p5t_28_R_1x_220121a.lef, created 212 library cells
Signal 11 received
Stack trace:
 0# 0x00005F8D92694553 in openroad
 1# 0x00007B574E042990 in /lib/x86_64-linux-gnu/libc.so.6
 2# odb::dbHashTable<odb::_dbInst>::find(char const*) in openroad
 3# sta::dbSdcNetwork::findInstance(char const*) const in openroad
 4# sta::dbSdcNetwork::findPin(char const*) const in openroad
 5# sta::ReadVcdActivities::setVarActivity(char const*, std::vector<sta::VcdValue, std::allocator<sta::VcdValue> > const&, int) in openroad
[deleted]
```

### Expected Behavior

Error/warning messages

### Environment

```shell
OpenROAD v2.0-14511-gedb4e7e37
```

### To Reproduce

See above

### Relevant log output

_No response_

### Screenshots

_No response_

### Additional Context

_No response_

oharboe:
Calling read_vcd/read_saif/read_power_activities before loading a design (DEF/Verilog) caused a SEGFAULT in dbNetwork::findInstance() because block_ was null.

Add a null guard for block_ in findInstance(), returning nullptr when no design block exists. The VCD reader already handles null pin lookups gracefully, so the result is "Annotated 0 pin activities" instead of a crash.

Closes #5384

## Summary
[Describe your changes here]

## Type of Change
- [x] Bug Fix
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Impact

## Verification

## Related Issues
[Link issues here]

### DIFF
--- a/src/dbSta/src/dbNetwork.cc
+++ b/src/dbSta/src/dbNetwork.cc
@@ -1112,6 +1112,9 @@ bool dbNetwork::isLeaf(const Instance* instance) const
 
 Instance* dbNetwork::findInstance(const char* path_name) const
 {
+  if (block_ == nullptr) {
+    return nullptr;
+  }
   if (hasHierarchy()) {  // are we in hierarchical mode ?
     // find a hierarchical module instance first
     odb::dbModInst* mod_inst = block()->findModInst(path_name);
--- a/src/dbSta/test/BUILD
+++ b/src/dbSta/test/BUILD
@@ -31,6 +31,7 @@ ALL_TESTS = [
     "power1",
     "read_liberty1",
     "read_vcd",
+    "read_vcd_no_design",
     "read_verilog1",
     "read_verilog2",
     "read_verilog3",
@@ -135,6 +136,12 @@ filegroup(
             "read_vcd": glob(
                 [
                     "Element.*",
+                    "MockArray.*",
+                    "asap7/*",
+                ],
+            ),
+            "read_vcd_no_design": glob(
+                [
                     "MockArray.*",
                     "asap7/*",
                 ],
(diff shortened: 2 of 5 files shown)


oharboe:
@maliberty PR/issue, can you assign?

maliberty:
I remove the null pointer check and the test still passes so it is not a correct reproducer of the issue.

oharboe:
> I remove the null pointer check and the test still passes so it is not a correct reproducer of the issue.

then this is an issue and not a PR...

maliberty:
its not an issue either as the problem doesn't reproduce

oharboe:
> its not an issue either as the problem doesn't reproduce

It crashes alright, but in a much larger more inconvenient example.

oharboe:
So what do we here, loose the test and keep the fix?

maliberty:
I don't think this is a correct fix and the test is non-functional so I don't see anything to keep here.  I tried to reproduce the crash myself but could not.  I await something actionable.

oharboe:
good call. I know the crash is real, but I need to a) run into it b) have the time to try creating a better test case. This was a good preparation for that.

PR #9935, opened 2026-03-25
DRT-0073 bloat_occupied_layers bugfix

oharboe:
## Summary
When `-bloat_occupied_layers` is set, `writeObstructions()` writes the entire block bounding box as a single OBS rectangle per occupied layer. This covers the macro's own pins, making them inaccessible to `pin_access` in downstream designs that instantiate the macro (DRT-0073).

The fix subtracts pin shapes (bloated by one layer pitch for access clearance) from the bounding box polygon before decomposing to rectangles. Both BTerm pins and power pins from special nets are handled.

The first commit adds a failing test that demonstrates the bug: a minimal design with a signal pin on metal1 at the die edge plus a VDD special net on the same layer. Before the fix, OBS covers the pin; after the fix, a one-pitch gap around the pin keeps it accessible.

The bug was found in a hierarchical ASAP7 design ([oharboe/openroad-demo gemmini_8x8_abutted](https://github.com/oharboe/openroad-demo/tree/main/gemmini_8x8_abutted)) where Tile macros built with `-bloat_occupied_layers` have M4 pins on a layer used for internal routing. 6 of 64 macro instances fail with DRT-0073 at the `pin_access` step because the M4 OBS covers their pins entirely.

## Type of Change

- Bug fix

## Impact

## Verification

## Related Issues
[Link issues here]

### DIFF
--- a/src/odb/test/abstract_bloat_pin.tcl
+++ b/src/odb/test/abstract_bloat_pin.tcl
@@ -19,56 +19,56 @@ set in_pin 0
 set in_obs 0
 set current_layer ""
 foreach line [split $lef_content "\n"] {
-    set line [string trim $line]
-    if {[string match "PIN sig_in" $line]} {
-        set in_pin 1
-    } elseif {$in_pin && [string match "LAYER *" $line]} {
-        set current_layer [lindex $line 1]
-    } elseif {$in_pin && [string match "RECT *" $line] && $current_layer eq "metal1"} {
-        lappend pin_rects $line
-    } elseif {$in_pin && [string match "END sig_in" $line]} {
-        set in_pin 0
-    }
-    if {[string match "OBS" $line]} {
-        set in_obs 1
-        set current_layer ""
-    } elseif {$in_obs && [string match "LAYER *" $line]} {
-        set current_layer [lindex [split $line " "] 1]
-    } elseif {$in_obs && [string match "RECT *" $line] && $current_layer eq "metal1"} {
-        lappend obs_rects $line
-    } elseif {$in_obs && [string match "END" $line]} {
-        set in_obs 0
-    }
+  set line [string trim $line]
+  if { [string match "PIN sig_in" $line] } {
+    set in_pin 1
+  } elseif { $in_pin && [string match "LAYER *" $line] } {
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty So this is half bug report, half fix made by claude for a problem I'm having in https://github.com/oharboe/openroad-demo/tree/main/gemmini_8x8_abutted, can you assign/half a look?

Let me know if you have questions.

gadfort:
@oharboe pins and obs is allowed to overlap, so the bloating shouldn't be an issue. Can you post a picture of the offending LEF obs and PIN this is attempting to fix?

oharboe:
> @oharboe pins and obs is allowed to overlap, so the bloating shouldn't be an issue. Can you post a picture of the offending LEF obs and PIN this is attempting to fix?

I'm out of my depth here...

@maliberty Help?

maliberty:
Is it intentional that the pin is interior to the block and not on the edge?

oharboe:
> Is it intentional that the pin is interior to the block and not on the edge?

No, I just asked ORFS to do something and then it did this...

oharboe:
> Is it intentional that the pin is interior to the block and not on the edge?

I tried to reproduce the full archive, but I ran into an  MPL-0005 error.

So, somehwere in this sequence of unfortunate events building this design, there's a generate abstract that is causing the constipation.

Below I was trying to create an untar and run archive:

    tar xf global_route_drt0073.tar.gz
    cd global_route_drt0073
    echo exit | ./run-me-MeshWithDelays-asap7-base.sh

to produce:

    [ERROR DRT-0073] Pin mesh/mesh_0_0/io_in_a_0[0] not reachable from any routing layer.

Generated by:

    git clone https://github.com/oharboe/openroad-demo
    cd openroad-demo
    git checkout c03b9d1f0e4f2af694fc9f70aef0715ff8519886
    bazelisk run //gemmini_8x8_abutted:MeshWithDelays_grt_deps
    tmp/gemmini_8x8_abutted/MeshWithDelays_grt_deps/make ISSUE_TAG=drt0073 global_route_issue

oharboe:
out of sight out of mind for now

PR #9938, opened 2026-03-25
grt: replace assert with logger error in CUGR GridGraph

Divinesoumyadip:
Fixes raw assert() calls in `src/grt/src/cugr/src/GridGraph.cpp` by  replacing them with proper` logger_->error(`) calls following OpenROAD  conventions.

Raw asserts crash silently in release builds. This is especially critical for `GridGraph.cpp` which manages all routing grid state including capacity, wire usage, and via placement during incremental  routing operations.

Changes:
- Replace 9 raw assert() calls with` logger_->error() `(GRT 1249-1257)
- Remove unused` #include <cassert>`

Follows the same pattern as #9870.

Note:   Raw assert() calls in `GridGraph.cpp `crash silently in release builds without any diagnostic information, making debugging impossible during incremental CUGR routing operations.` GridGraph.cpp` is the core grid state management layer  responsible for wire capacity, via placement, and edge tracking , making silent crashes here catastrophic during ECO rerouting passes. This PR replaces 9 raw asserts with `logger_->error()`  calls (GRT 1249-1257) that provide meaningful error messages identifying exactly which coordinate alignment or layer index violation occurred. Removing these silent failure points directly strengthens the reliability of the incremental CUGR routing.

PR #9945, opened 2026-03-25
grt: replace logger report with logger info and fix asserts in CUGR

Divinesoumyadip:
CUGR.cpp used` logger_->report()` for all routing stage announcements and statistics output, which bypasses the OpenROAD message ID system and prevents log filtering. Additionally, two raw` assert()` calls would crash silently in release builds with no diagnostic information.

This PR replaces all 11 `logger_->report()` calls with `logger_->info()` using GRT message IDs 1258-1269, replaces the 2 raw asserts with` logger_->error()` calls (GRT 1270-1271) that include the actual values that caused the failure, and removes the now-unused `#include <cassert>.`

 **Related Issues**:
Follows pattern established in #9870 and #9938.

### DIFF
--- a/src/grt/src/cugr/src/CUGR.cpp
+++ b/src/grt/src/cugr/src/CUGR.cpp
@@ -233,8 +233,10 @@ void CUGR::mazeRoute(std::vector<int>& net_indices)
     mazeRoute.run();
     std::shared_ptr<SteinerTreeNode> tree = mazeRoute.getSteinerTree();
     if (tree == nullptr) {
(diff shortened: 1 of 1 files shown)


Divinesoumyadip:
@eder-matheus Check it out when u have time .Thanks again.

eder-matheus:
@Divinesoumyadip you have some conflicts in this PR. I see that you're using a very old version of OpenROAD as a base for this PR. Please, make sure to always checkout to the latest master branch before creating new updates.

PR #9950, opened 2026-03-25
Update or-tools dependency.

hzeller:


maliberty:
Do you really need this or is it just why not be on the latest?  or-tools is horrible when it comes to dependencies and having to update it in cmake is a big hassle.

hzeller:
No code changes are needed, so you can probably keep whatever you have in cmake
(9.15 is the latest on BCR: https://registry.bazel.build/modules/or-tools )

The reason I change is, is that the newer bazel project is compatible with the next version of bazel (bazel9), which I want to get us slowly compatible towards.

maliberty:
We did run into result differences before from using different versions.  Hopefully we'll get by until can drop cmake.

Issue #9937, opened 2026-03-25
downstream example not building `@openroad//:openroad`


hzeller:
Attempting to build openroad in the downstream configuration is not working currently

After setting c++20 requirements:

```bash
cd test/downstream/
echo 'build --cxxopt "-std=c++20"        --host_cxxopt "-std=c++20"' >> .bazelrc
```

and attempt to build the `openroad` binary failes
```
bazel build -k -c opt @openroad//:openroad
```

with a bunch of *.i files or header files not found. Maybe this is just a consequence of the root project not using proper fully qualified header files ?

hzeller:
Also following the [user set-up](https://github.com/The-OpenROAD-Project/OpenROAD/blob/50bb0081d8a247e2bcc1d562a5b6aa5065c15528/docs/user/Bazel.md#minimal-modulebazel-for-a-downstream-project) runs into the same *.i missing and include not found issues.

hzeller:
CC @oharboe

oharboe:
Use-case: https://github.com/oharboe/openroad-demo pulls in OpenROAD in MODULE.bazel and patches it for specific designs as a workaround after having filed issues and pull requests.

hzeller:
what are your patches to get it to compile ?

oharboe:
> what are your patches to get it to compile ?

one that is merged + I invariably have some bugfixes to OpenROAD that I need to press onwards...

https://github.com/The-OpenROAD-Project/OpenROAD/issues/9902#issuecomment-4127696857

maliberty:
I'm not sure what is being requested here.

PR #9951, opened 2026-03-25
Don't manually use copts, use bazel way to point to includes.

hzeller:
This makes it more compatible with use as external dependency. Issues: #9937

### DIFF
--- a/src/cts/BUILD
+++ b/src/cts/BUILD
@@ -84,10 +84,10 @@ cc_library(
     hdrs = [
         "include/cts/MakeTritoncts.h",
     ],
-    copts = [
-        "-Isrc/cts/src",
+    includes = [
+        "include",
+        "src",
     ],
-    includes = ["include"],
     deps = [
         ":cts",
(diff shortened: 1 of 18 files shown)


src/cts/BUILD
```diff
@@ -84,10 +84,10 @@ cc_library(
     hdrs = [
         "include/cts/MakeTritoncts.h",
     ],
-    copts = [
-        "-Isrc/cts/src",
+    includes = [
+        "include",
+        "src",
```
maliberty, line 89: Why is src being added?  In general the public api is in include and clients aren't expected to use headers in src.

hzeller: I just moved the wrong way of providing include paths via manual copts to the correct way with include.
This doesn't change anything in the visibility of the includes.

hzeller:
using the `-I` is what makes the breakage in #9937 as they now don't fit anymore the different project layout. The `includes = [...]` take that into account.

PR #9952, opened 2026-03-25
sh_test and sh_binary are used in BUILD.bazel. Import.

hzeller:
That means, that rules_shell is indeed not a dev_dependency anymore, but a real one.


hzeller:
Previously, `sh_test` and `sh_binary` without actually loading that dependency was kinda sorta working as early versions of bazel had these built in. Starting with bazel 9, it will complain.

`rules_shell` is not a big dependency, almost everyone has that already anyway.

test/BUILD
```diff
@@ -49,17 +49,6 @@ py_test(
...
 
-py_test(
-    name = "rules_shell_dev_dep_test",
-    srcs = ["check_dev_dep.py"],
-    args = [
-        "--module-file=$(location //:MODULE.bazel)",
-        "--dep=rules_shell",
-    ],
-    data = ["//:MODULE.bazel"],
-    main = "check_dev_dep.py",
```
maliberty, line 60: Clean up this file too?

hzeller: this file does not use sh_binary or sh_test.

The change here is just to _not_ check that rules_shell is a dev_dependency.

hzeller:
merge ?

PR #9958, opened 2026-03-25
Make sure downstream is compiling with c++20

hzeller:
... openroad is not compatible with older versions and the default compilers resort to is c++17.


QuantamHD:
Does bcr/bazel respect this downstream .bazelrc?

hzeller:
No, users have to set that themselves.
This directory is an 'example project' to test if the import of openroad as an upstream project works.

maliberty (approved):

Issue #9961, opened 2026-03-26
ram: Check PDN parameters


rovinski:
If the provided PDN width/pitch are too large, calling `RamGen::ramPdngen()` can result in a segfault.

The pitch and width parameters should be checked to make sure that they are not larger than the die size. Otherwise, error.

There may be error checking already done in the PDN module which we aren't calling. Or the error checking might be in Tcl so we are circumventing it by calling it in C++. There may be some slight code duplication but `ram` needs to not segfault.

gadfort:
@rovinski these checks are handled inside PDN, so it should not be possible to bypass (unless there is a bug in PDN). Can you share a testcase?

rovinski:
@tnguy19 Can you open a PR with the code changes that caused the segfault and point @gadfort to them?

tnguy19:
@gadfort Hi, so I ran into this while working on implementing column/mux support where if you set a ratio value that causes the RAM to be smaller than the provided PDN width/pitch it will result in a segmentation fault. You can get a similar result by creating a RAM that is small enough. For example if you create a  4x8 RAM using:
```
source "helpers.tcl"

read_liberty sky130hd/sky130_fd_sc_hd__tt_025C_1v80.lib
read_lef sky130hd/sky130hd.tlef
read_lef sky130hd/sky130_fd_sc_hd_merged.lef

generate_ram \
  -mask_size 8 \
  -word_size 8 \
  -num_words 4 \
  -read_ports 1 \
  -storage_cell sky130_fd_sc_hd__dfxtp_1 \
  -power_pin VPWR \
  -ground_pin VGND \
  -routing_layer {met1 0.48} \
  -ver_layer {met2 0.48 40} \
  -hor_layer {met3 0.48 20} \
  -filler_cells {sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 \
    sky130_fd_sc_hd__fill_4 sky130_fd_sc_hd__fill_8} \
  -tapcell sky130_fd_sc_hd__tap_1 \
  -max_tap_dist 15
```
You will run into a segmentation fault with the output:
```
[INFO PDN-0001] Inserting grid: ram_grid
Signal 11 received
Stack trace:
(...)
2# odb::dbTrackGrid::getGridY() in openroad 
3# pdn::TechLayer::populateGrid(odb::dbBlock*, odb::dbTechLayerDir) in openroad 
[...]

PR #9968, opened 2026-03-26
ram: add PDN strap pitch vs die size error check in ramPdngen() to prevent segfault

tnguy19:
## Summary
When the die area is smaller than the PDN strap pitch, calling RamGen::ramPdngen() can result in a segfault. This adds a check at the top of ramPdngen() to compare die dimensions against the strap pitch and give an error instead

## Type of Change
- Bug fix

## Impact
Instead of a segfault when PDN strap pitch is larger than the die size, user will get a clear error message indicating which dimension is too small and suggesting they use a smaller pitch 

Before Fix Sample Error Message:
```
[INFO PDN-0001] Inserting grid: ram_grid
Signal 11 received
 2# odb::dbTrackGrid::getGridY() in openroad
 3# pdn::TechLayer::populateGrid(odb::dbBlock*, odb::dbTechLayerDir) in openroad
 4# pdn::Straps::makeShapes(...) in openroad
 8# pdn::PdnGen::buildGrids(bool) in openroad
 9# ram::RamGen::ramPdngen(...) in openroad
Segmentation fault (core dumped)
```

After Fix Sample Error Message:
```
[ERROR RAM-0031] Die height (13600 DBU) is less than horizontal strap pitch (20000 DBU). Use a smaller -hor_layer pitch.
```

## Verification

## Related Issues
#9961 

@rovinski 
@gadfort

### DIFF
--- a/src/ram/src/ram.cpp
+++ b/src/ram/src/ram.cpp
@@ -466,7 +466,7 @@ void RamGen::ramPdngen(const char* power_pin,
                        int hor_pitch)
 {
   const odb::Rect& die = block_->getDieArea();
-  const double dbu_per_um = block_->getDb()->getTech()->getDbUnitsPerMicron();
+  const double dbu_per_um = block_->getDb()->getDbuPerMicron();
 
   if (die.dy() < hor_pitch) {
     logger_->error(RAM,


src/ram/src/ram.cpp
```diff
@@ -465,6 +465,25 @@ void RamGen::ramPdngen(const char* power_pin,
                        int hor_width,
                        int hor_pitch)
 {
+  const odb::Rect& die = block_->getDieArea();
+  if (die.dy() < hor_pitch) {
+    logger_->error(
+        RAM,
+        31,
+        "Die height ({} DBU) is less than horizontal strap pitch ({} DBU). "
```
gadfort: you should report this in um like all the other tools

gadfort (changes requested):

gadfort:
@rovinski I think this check makes sense even if the tool is no longer segfaulting

rovinski:
Yep, I had the same feedback via email

src/ram/src/ram.cpp
```diff
@@ -465,6 +465,29 @@ void RamGen::ramPdngen(const char* power_pin,
                        int hor_width,
                        int hor_pitch)
 {
+  const odb::Rect& die = block_->getDieArea();
+  auto tech = block_->getDb()->getTech();
+  const double dbu_per_um = tech->getDbUnitsPerMicron();
```
rovinski: `tech` is not reused and it all fits in one line (I believe) so just make it a one-liner
```suggestion
  const double dbu_per_um = block_->getDb()->getTech()->getDbUnitsPerMicron();
```

src/ram/src/ram.cpp
```diff
@@ -465,6 +465,29 @@ void RamGen::ramPdngen(const char* power_pin,
...
+
+  if (die.dy() < hor_pitch) {
+    logger_->error(RAM,
+                   31,
+                   "Die height ({:.2f} um) is less than horizontal strap pitch "
+                   "({:.2f} um). "
+                   "Use a smaller -hor_layer pitch.",
```
rovinski, line 476: Why separate lines?
```suggestion
                   "({:.2f} um). Use a smaller -hor_layer pitch.",
```

rovinski:
Afterwards mark as ready for review

src/ram/src/ram.cpp
tnguy19: I think this might be because of the clang format. I put it in a single line and after apply clang format it revert back to this

rovinski: Oh well 🤷‍♂️

rovinski:
CI failure unrelated

tnguy19:
Ready for review @gadfort @maliberty

src/ram/src/ram.cpp
```diff
@@ -465,6 +465,28 @@ void RamGen::ramPdngen(const char* power_pin,
                        int hor_width,
                        int hor_pitch)
 {
+  const odb::Rect& die = block_->getDieArea();
+  const double dbu_per_um = block_->getDb()->getTech()->getDbUnitsPerMicron();
```
maliberty:
```suggestion
  const double dbu_per_um = block_->getDb()->getDbuPerMicron();
```

tnguy19: @maliberty code is updated with suggested change

PR #9970, opened 2026-03-27
refactor docs_check

luarss:
## Summary
- Reduce readme_msgs_check single source of truth.
- Rewrite readme_msgs_check to unittest -> eliminate `.ok` files
- Remove fragile os.chdir actio

## Type of Change

- Reducing LOC, repeated code

## Impact
Maintains doc tests behavior 

## Verification
- [x] I have verified that the local build succeeds (`./etc/Build.sh`).
- [x] I have run the relevant tests and they pass.
- [x] My code follows the repository's formatting guidelines.

## Related Issues
n/a


oharboe (commented):
@maliberty there has to be a bazel idiomatic way to do this to avoid the symlinks, but we need to keep the non-bazel testing flow, I suppose.

The docker image still has the non-bazel build, the non-bazel testing is required.

PR #9972, opened 2026-03-27
docs: add Bazel test performance analysis and recommendations

oharboe:
## Summary

Analyzed performance of builds - self updating w/Claude, useful for those who don't need or want Claude.

## Type of Change

- Documentation update

## Impact
[How does this change the tool's behavior?]

## Verification
- [ ] I have verified that the local build succeeds (`./etc/Build.sh`).
- [ ] I have run the relevant tests and they pass.
- [ ] My code follows the repository's formatting guidelines.
- [x] **I have signed my commits (DCO).**

## Related Issues
[Link issues here]

Issue #7042, opened 2025-04-08
bazel: make use of filegroups and don't rely on symbolic links


hzeller:
### Description

In the set-up of getting tests going, we're using the infrastructure for test-data in place for the bazel build. Many test directories have symbolic links to `../../test/somthing` so that they appear local to that directory; and from there we build a local file-group per directory.

Bazel has a neater way to represent bundles of files, the file groups. We can create these at the original source in `//test` (e.g. for sky130 pkd or nangate45 something etc), and use these in tests all over the project by just having to refer to `//test:foobar_resource`.

Right now, there are about 270 symbolic links in the project. Would be good that after this change, we would not need any of it (we would only have to keep them for cmake).

```
$ find src -type l | wc -l
270
```

### Suggested Solution

Build well-named file-groups in //test for test resources and use them wherever we use local file groups packing symbolic linked content instead.

hzeller:
this is pretty much alive

hzeller:
Since the first reporting, the situation is worse, repeating the same thing in multiple places using symlinks

```
find src -type l | wc -l
357
```

The following are the number of incoming links for the files in question.
Mostly test files are repeated, but test scripts.
Test inputs should be dealt with by filegroups, or by direct referencing (many tests get _all_ files fed, but they only need the one they care about).
Scripts in particular should not be replicated, but put in one place, then wrap a *.bzl rule around.

```
      1 OpenROAD/src/gpl/test/core01.def
      1 OpenROAD/src/gpl/test/simple01-td.def
      1 OpenROAD/src/gpl/test/simple07.def
      1 OpenROAD/src/odb/src/lef/lefdefReadme.txt
      1 OpenROAD/src/pdn/test/sky130_temp_sensor/HEADER.lef
      1 OpenROAD/src/pdn/test/sky130_temp_sensor/SLC.lef
      1 OpenROAD/src/sta/test/sky130hd/sky130_fd_sc_hd__tt_025C_1v80.lib
      1 OpenROAD/src/sta/test/sky130hs/sky130_fd_sc_hs__tt_025C_1v80.lib
      1 OpenROAD/src/tap/test/boundary_macros_separate.defok
      1 OpenROAD/test/Nangate45/Nangate45_stdcell.lef
      1 OpenROAD/test/Nangate45/Nangate45_tech.lef
      1 OpenROAD/test/Nangate45/Nangate45_typ.lib
      1 OpenROAD/test/asap7_augmented
[...]

Issue #5875, opened 2024-10-04
Mangled tcl environment


oharboe:
### Describe the bug

```
sudo apt install tcllib
```

```
oyvind@corona:~/OpenROAD-flow-scripts/flow$ tclsh
% package require yaml
0.4.1
%
```

Whereas:

```
$ openroad
OpenROAD v2.0-16219-g56c3c86ba 
Features included (+) or not (-): +Charts +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
openroad> package require yaml
error reading package index file /usr/lib/blt2.5/pkgIndex.tcl: Error: pkgIndex.tcl, 59 can't read "dir": no such variable
error reading package index file /usr/lib/tcltk/x86_64-linux-gnu/tclreadline2.3.8/pkgIndex.tcl: Error: pkgIndex.tcl, 11 can't read "dir": no such variable
[etc]
can't find package yaml
while evaluating package require yaml
openroad> 
```

Also:

```
$ sta
OpenSTA 2.6.0 20925bb009 Copyright (c) 2024, Parallax Software, Inc.
License GPLv3: GNU GPL version 3 <http://gnu.org/licenses/gpl.html>

This is free software, and you are free to change and redistribute it
under certain conditions; type `show_copying' for details. 
This program comes with ABSOLUTELY NO WARRANTY; for details type `show_warranty'.
sta [~/OpenROAD-flow-scripts/flow] package require yaml
error reading package index file /usr/lib/blt2.5/pkgIndex.tcl: Error: pkgIndex.tcl, 59 can't read "dir": no such variable
[deleted]
while evaluating package require yaml
sta [~/OpenROAD-flow-scripts/flow] 
```

### Expected Behavior

No error

### Environment

```shell
sta or OpenROAD
```

### To Reproduce

see above

donn:
This is the culprit:

https://github.com/The-OpenROAD-Project/OpenSTA/blob/ee700ba0f40e2a7101adc3ffd52953fc9f1f0245/sdc/Sdc.tcl#L109

OpenSTA hijacks `source` to add some conveniences when sourcing .sdc files. Unfortunately, that source function always runs every line at the top level.  dir, the variable in question, is not set at the top level, which causes this issue when attempting to load any Tcl package.

[...]

Issue #5563, opened 2024-08-13
.tcl stack traces don't work with OpenROAD


oharboe:
### Describe the bug

```
     1	proc foo {} {
     2	    puts "Foo"
     3	    x
     4	}
     5	
     6	proc bar {} {
     7	    foo
     8	    puts "Bar"
     9	}
    10	
    11	bar
    12	exit
$ openroad -no_init test.tcl 
OpenROAD v2.0-15137-g35b37708a 
Features included (+) or not (-): +Charts +GPU +GUI +Python
This program is licensed under the BSD-3 license. See the LICENSE file for details.
Components of this program may be licensed under more restrictive licenses which must be honored.
Foo
Error: test.tcl, 11 invalid command name "x"
openroad> 
```

### Expected Behavior

Expected:

```
$ tclsh test.tcl 
Foo
invalid command name "x"
    while executing
"x"
    (procedure "foo" line 3)
    invoked from within
"foo"
    (procedure "bar" line 2)
    invoked from within
"bar"
    (file "test.tcl" line 11)
```

### Environment

```shell
OpenROAD v2.0-15137-g35b37708a
```

### To Reproduce

See above

oharboe:
this is an ongoing pain: no way to know where .tcl failed.

PR #9973, opened 2026-03-27
bazel: trim tcl_lang runfiles to reduce test symlinks by 58%

oharboe:
Based on [analysis](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9972)

Patch @tcl_lang//:tcl_core to exclude library/tzdata (596 files) and library/msgs (127 files) that OpenROAD does not use at runtime.

This reduces per-test runfiles symlinks from 1,242 to 519 (58%), eliminating ~723 unnecessary symlinks per test across ~1,400 tests.

Verified: 1,275 of 1,276 tests pass (only pre-existing rmp/aes_genetic timeout failure).

## Summary
[Describe your changes here]

## Type of Change

- Bug fix

## Impact

## Verification

## Related Issues
[Link issues here]

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -65,6 +65,12 @@ bazel_dep(name = "tcmalloc", version = "0.0.0-20250927-12f2552")
 bazel_dep(name = "zlib", version = "1.3.1.bcr.5")
 bazel_dep(name = "yaml-cpp", version = "0.9.0")
 bazel_dep(name = "tcl_lang", version = "8.6.16.bcr.1")
+single_version_override(
+    module_name = "tcl_lang",
+    patch_strip = 1,
+    patches = ["//bazel/patches:tcl_lang_trim_core.patch"],
+)
+
 bazel_dep(name = "readline", version = "8.2.bcr.3")
 
 # JavaScript / web UI (src/web uses @npm for bundled assets)
--- a/bazel/patches/BUILD
+++ b/bazel/patches/BUILD
@@ -1,0 +1,1 @@
+exports_files(["tcl_lang_trim_core.patch"])
--- a/bazel/patches/tcl_lang_trim_core.patch
+++ b/bazel/patches/tcl_lang_trim_core.patch
@@ -1,0 +1,28 @@
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -112,7 +112,24 @@
+ 
+ filegroup(
+     name = "tcl_core",
+-    srcs = glob(["library/**"]),
++    # Exclude files that OpenROAD does not need at runtime.
++    # This reduces per-test runfiles from 831 to ~15 symlinks.
++    srcs = glob(
++        ["library/**"],
++        exclude = [
++            "library/tzdata/**",
++            "library/msgs/**",
++            "library/encoding/**",
++            "library/http/**",
++            "library/http1.0/**",
++            "library/tcltest/**",
++            "library/opt/**",
++            "library/msgcat/**",
++            "library/dde/**",
++            "library/platform/**",
++            "library/reg/**",
++        ],
++    ),
+     visibility = ["//visibility:public"],
+ )
+ 
--- a/test/BUILD
+++ b/test/BUILD
@@ -138,12 +138,14 @@ filegroup(
 
 filegroup(
     name = "regression_resources",
-    # TODO Refine glob later.
-    #
-    # This is a very broad glob, but most of the data in size
-    # comes from .lib files and we can't know which pdk files
-    # to include for a test given the name only.
-    srcs = glob(["**/*"]),
+    # Broad glob, but exclude directories that regression tests never use.
+    srcs = glob(
+        ["**/*"],
+        exclude = [
+            "orfs/**",
+            "downstream/**",
+        ],
+    ),
     visibility = ["//visibility:public"],
 )
 


oharboe:
@hzeller FYI

gadfort:
Does this impact what packages tcl has available?

hzeller:
curious why every test is symlinking all of these, wouldn't we expect a single symlink to a directory of sorts ?

What time saving are we talking about ?

oharboe:
> gadfort: Does this impact what packages tcl has available?

@hzeller @maliberty You may be interested in this too...

Glad you asked...

Short answer: yes, this PR removes `library/tzdata` (596 files) and `library/msgs` (127 files) from the `tcl_core` filegroup. These provide timezone data and message catalogs for `clock format`, `msgcat`, etc. OpenROAD doesn't use them at runtime — 1,275 of 1,276 tests pass without them.

If it does break something, we don't have any tests to detect it today. That's the real problem.

## The deeper question

This PR is a symptom of a structural issue: **symlink fatigue**. With ~1,400 tests each creating ~1,242 runfiles symlinks, the filesystem overhead is substantial. Trimming 723 symlinks per test (58%) is a meaningful win, but it's a band-aid on a deeper tension.

The tension is: **what role should Tcl play in OpenROAD?**

## Framework inversion

Disappointment is just mismanaged expectations. The OpenROAD built-in Tcl interpreter will never meet the expectations of someone who really wanted OpenROAD to be a package importable into Tcl. You can't `package require` arbitrary Tcl packages into the embedded interpreter without fighting the build system, the sandbox, and the dependency graph every step of the way.

The same is true for Python. ZeroASIC's [SiliconCompiler](https://github.com/siliconcompiler/siliconcompiler) is a Python flow that drives OpenROAD by shelling out to the `openroad` binary and feeding it Tcl scripts (see [`tools/openroad/__init__.py:274`](https://github.com/siliconcompiler/siliconcompiler/blob/50cf59a43b6b5c8ce4fdd8baafa21505e85a4167/siliconcompiler/tools/openroad/__init__.py#L274) — `set_exe("openroad", format="tcl")`). It's Python on the outside, but the interface to OpenROAD is still Tcl. Every change to the Tcl surface area — like this PR removing `tzdata` and `msgs` — is a tremor those Tcl scripts have to absorb. The question "does this impact what packages Tcl has available?" is really "am I digging a deeper hole by depending on an embedded Tcl interpreter I don't control?"

The good news is that OpenROAD *already is* a Python library. `from openroad import Tech, Design, Timing` works today (see [`test/timing_report_api.py`](https://github.com/The-OpenROAD-Project/OpenROAD/blob/8f7f531bebfca6b58f811c0772d3217a1e31b82d/test/timing_report_api.py)). The framework-inverted path exists — it just isn't what SiliconCompiler uses yet.

For bazel-orfs, pypi doesn't match the use case. We're running bleeding-edge OpenROAD + patches, not stable releases. Any advanced flow working on novel technology will run into OpenROAD issues, need to report them, patch OpenROAD locally, and switch back to the official version when the fix lands. This isn't theoretical — bazel-orfs/gallery has already hit several bugs, some fixed upstream, others in the works. Bazel makes this workflow natural: you patch, you build, you file the issue, you move on. That's a feature, not a limitation.

## What Tcl does ORFS actually need?

I read through the ORFS `flow/scripts/` directory. The answer is: a modest but solid subset. No surprises, no exotic features.

**Used heavily:** `set`, `if`/`foreach`/`while`/`catch`/`try`, `proc`, `expr` (with `max`, `int`, `round`), list ops (`lappend`, `lindex`, `lsort`, `lmap`, `lsearch`), dicts, string ops (`string match`, `string map`, `regexp`, `format`), file I/O (`open`/`close`/`puts`/`gets`/`read`/`glob`/`file exists|normalize|join`), `exec`, `upvar`/`uplevel`, `source`, `info exists`, `{*}` argument expansion, `clock seconds` (just for elapsed time measurement).

**Not used at all:** `package require` (explicitly noted as broken in OpenROAD's Tcl environment — see OpenROAD issue #5875), tzdata, msgcat, locale, encoding, namespaces (ORFS calls into `sta::`, `utl::`, `odb::` etc. but those are defined by C++, not Tcl), OO, `interp`, `trace`, event loop, sockets, `binary`, safe interpreters.

Zero `package require` calls. Zero timezone operations. Zero message catalog usage. The Tcl that ORFS needs is basically "scripting glue": control flow, string/list manipulation, file I/O, and calling OpenROAD commands. That's it.

This is a small enough surface area that a minimal embedded Tcl interpreter is realistic. Tcl interpreters are tiny and have no dependencies. [Jim Tcl](https://github.com/msteveb/jimtcl) is a from-scratch Tcl implementation in ~17k lines of C, BSD-licensed, ~85k binary, zero external dependencies for the core. It implements Tcl 8.5/8.6 features including `{*}` expansion, dicts, `lmap` — everything ORFS uses. You could either use jimtcl directly or use Claude to roll a purpose-built C++ interpreter covering exactly the ORFS surface area, with unit tests generated alongside it.

The blocker is OpenSTA. OpenSTA has [16+ files](https://github.com/The-OpenROAD-Project/OpenSTA) touching `Tcl_Interp` and `tcl.h` — it's tightly coupled to the standard Tcl C API. Worse, OpenSTA actively breaks Tcl: it overrides the standard `source` command, which kills stack traces ([OpenROAD #5563](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5563), still open). Where `tclsh` gives you a full call chain showing which proc on which line failed, `openroad` just says `Error: test.tcl, 11 invalid command name "x"` and drops you into a REPL. This is an ongoing pain for anyone debugging Tcl scripts in OpenROAD ([#5436](https://github.com/The-OpenROAD-Project/OpenROAD/issues/5436)). So we're not just carrying a heavy Tcl dependency — we're carrying a broken one.

Switching to jimtcl or a custom interpreter would require OpenSTA buy-in, since SDC parsing flows through it. That said, jimtcl provides a [Tcl-compatible C API](https://github.com/msteveb/jimtcl), so the porting surface may be smaller than it looks. This doesn't have to be all-or-nothing: you could start with OpenROAD commands on jimtcl and keep OpenSTA on standard Tcl behind an abstraction, then converge later.

Either way, dragging in full Tcl with 596 timezone files and 127 message catalog files for a language that ORFS uses as a command dispatcher is overkill.

From a dependency point of view, we need *enough* Tcl to support:

1. **SDC parsing** — this is load-bearing and non-negotiable
2. **The ORFS flow** — the command interface that ORFS scripts call today, which is procedural scripting glue, not a Tcl application
3. **A good interactive experience** — tab completion, `source`, the things users expect when they type `openroad` at a prompt

These two things should happen in parallel:

**Shrink the embedded Tcl** to what we actually need (jimtcl or equivalent), and **continue supporting OpenROAD as a library** that can be used from any language — Tcl, Python, C++, Rust, whatever. The embedded interpreter handles SDC and ORFS compatibility. The library API handles everyone who wants to own their own process. These aren't competing goals, they reinforce each other: a smaller embedded Tcl is easier to maintain, and a clean C++ API means less pressure on the embedded interpreter to be all things to all people.

OpenROAD already works as a Python library (see above), and downstream flows like [SiliconCompiler](https://github.com/siliconcompiler/siliconcompiler) would benefit from using it that way instead of shelling out with Tcl scripts. PR [#9910](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9910) explored another angle — standalone per-command binaries that bypass the Tcl interpreter entirely, calling the C++ API directly. The key finding there was that `dbSta(nullptr, db, &logger)` works: the `Tcl_Interp*` parameter is already optional.

## The conundrum

All that said, we face competing constraints:

- **Compatibility**: current users rely on the Tcl command interface. We can't break ORFS flows or existing scripts.
- **Faster builds**: every Tcl file in `tcl_lang` is a runfiles symlink in every test. The data doesn't lie — tests are showing symlink fatigue. With ~1,400 tests, even small per-test overhead compounds.
- **Bazel BCR**: Tcl as a Bazel Central Registry module would mean the Tcl dependency becomes a shared, versioned module rather than a vendored tree. This is a pragmatic band-aid — it cleans up the packaging but doesn't address the deeper framework inversion issue. We'd still be dragging in a full Tcl distribution for a language we use as a command dispatcher.
- **Test architecture**: as hzeller pointed out on #9910, the ratio of ~1,000 Tcl integration tests to ~64 `cc_test()` unit tests should be inverted. Tcl tests are end-to-end integration tests that require the full binary; `cc_test()` targets link only what they need, build fast, and give precise failure signals. The more functionality we cover with `cc_test()`, the less the Tcl runtime matters for correctness verification.

## A path forward

1. **Trim what we don't use** (this PR): remove `tzdata`, `msgs`, and anything else that's dead weight. If something breaks, we add a test for it — that's a feature, not a bug.
2. **Define the Tcl surface area**: explicitly enumerate which Tcl packages OpenROAD actually uses. `init.tcl`, `auto.tcl`, `package.tcl` — yes. Timezone data — no. Make this a tested contract.
3. **Support framework inversion**: expose `cc_library()` targets so downstream consumers can build narrower binaries or import OpenROAD into their own runtime. PR #9910 showed this is mechanically feasible. For flows like SiliconCompiler that currently shell out to `openroad` with Tcl scripts, this is the difference between perpetually adapting to the embedded interpreter's shifting surface area and calling a stable C++ API directly from Python.
4. **Shift test weight to `cc_test()`**: reduce dependence on the Tcl interpreter for correctness testing. Use Tcl/Python tests for integration and end-to-end validation, not as the primary test harness.
5. **BCR Tcl as a stepping stone**: a Bazel Central Registry Tcl module cleans up the packaging, but it's a band-aid — the real fix is a minimal embedded interpreter (jimtcl or equivalent) that only carries what we actually use.

oharboe:
> curious why every test is symlinking all of these, wouldn't we expect a single symlink to a directory of sorts ?
> 
> What time saving are we talking about ?

https://github.com/The-OpenROAD-Project/OpenROAD/pull/9972

oharboe:
# Symlink fatigue: from Tcl to Nix to the Linux kernel

## The pattern

OpenROAD's Bazel build creates ~1,242 runfiles symlinks per test across ~1,400 tests. PR [#9973](https://github.com/The-OpenROAD-Project/OpenROAD/pull/9973) trims 723 symlinks per test by removing unused Tcl timezone and message catalog files. That's a 58% reduction, and it helps, but it's treating a symptom.

The same pattern shows up everywhere symlinks are used for isolation:

**Nix** builds its entire package model on symlink forests. Every user environment is a directory tree of symlinks into `/nix/store/<hash>-<name>/`. A simple `nix-env -i hello` creates ~36 symlinks. A full NixOS system profile creates ~1,533 symlinks per generation. Worse, Nix store paths are long (94+ characters), pushing most symlinks past ext4's 64-byte fast-symlink threshold — each one consumes a full 4 KiB block instead of being stored inline in the inode. One NixOS maintainer reported ~150 MiB per `sw` tree just from this block allocation overhead ([NixOS/nix#1415](https://github.com/NixOS/nix/issues/1415)).

**Bazel** creates a "runfiles tree" for each binary — a directory of symlinks pointing to actual source/output files. The sandbox execroot is also symlink-based. Benchmarks from the Bazel team show the cost is workload-dependent: 6-12% overhead for actions with hundreds of inputs, but **+270%** for actions with thousands to hundreds of thousands of inputs (iOS builds). Creating 15,000 symlinks on ext4/SSD takes ~3,500ms; the same on tmpfs takes ~350ms ([bazel-dev](https://groups.google.com/g/bazel-dev/c/aqAKGz5Vjec)).

The deeper issue: symlinks are a userspace workaround for a filesystem isolation problem. You want process A to see a different directory tree than process B, and symlinks are the cheapest way to fake it without kernel support. But "cheapest" stops being cheap at scale.

## What Bazel already offers

Before looking at kernel futures, there are existing Bazel flags worth knowing about:

- **`--experimental_sandbox_shm`**: creates the symlinked execroot on `/dev/shm` (tmpfs) and redirects writes via overlayfs. Result: ~10x speedup for symlink creation (3,500ms to 350ms for 15,000 symlinks). Full build: 19.5s vs 90s elapsed.
- **`--experimental_reuse_sandbox_directories`**: reuses previous sandbox directory trees, avoiding full symlink teardown/recreation between similar actions.
- **`--nolegacy_external_runfiles`**: reduces duplicate symlink trees for external repositories (default since Bazel 7).

The catch: any solution we adopt for OpenROAD must be **zero config and zero extra dependencies**. If users have to add flags to `.bazelrc` or install FUSE, we've traded one problem for another. The fix should either be upstream in Bazel's defaults or invisible in our rule implementation.

## Could the Linux kernel solve this?

Symlinks exist because Linux doesn't natively support per-process filesystem views at the granularity build systems need. Several kernel features get closer:

**Mount namespaces** (`unshare -m`): each namespace gets an independent mount table. You can bind-mount a custom directory tree visible only to that process. Bazel's `linux-sandbox` already uses this for action sandboxing — it creates a mount namespace with the entire filesystem read-only except the sandbox directory. But inside the namespace, it still uses symlinks for the runfiles tree. The proposal to bind-mount whole directory trees instead of individual files would reduce mount count from thousands to a handful ([bazel-dev](https://groups.google.com/g/bazel-dev/c/QDZYSrckshY)), but nobody has implemented it. Complication: Ubuntu 24+ restricts unprivileged user namespaces via AppArmor, which already breaks `linux-sandbox` for some users.

**OverlayFS**: union-mounts multiple directories into one view. ~17% bandwidth reduction vs native ext4 in benchmarks, but effectively zero cost for read-mostly workloads (which build inputs are). Combined with mount namespaces, could give per-process filesystem views without any symlinks. No Bazel integration for runfiles exists.

**FUSE / sandboxfs**: Google built [sandboxfs](https://github.com/bazelbuild/sandboxfs) (2017-2020), a FUSE filesystem that exposes arbitrary directory views configured with a single mount call vs thousands of symlink calls. It worked — C++/Go builds were 30% faster than symlink sandboxing. But FUSE overhead (kernel-to-userspace context switches) made it slower for I/O-heavy workloads, macOS FUSE had kernel bugs, and the project was effectively abandoned by 2020.

**io_uring** (Linux 5.14+): added `IORING_OP_SYMLINKAT` for batched symlink creation without per-call syscall overhead. Instead of N system calls for N symlinks, submit them all to the submission queue and make one `io_uring_enter()` call. Mentioned in [bazelbuild/bazel#16711](https://github.com/bazelbuild/bazel/issues/16711) as a potential optimization. No implementation exists yet, but the kernel support is there.

**eBPF/sandfs**: mentioned in Bazel discussions as having ~5% overhead vs FUSE's ~80%, running entirely in-kernel ([bazelbuild/bazel#7937](https://github.com/bazelbuild/bazel/issues/7937)). Experimental, no production implementation.

## The realistic path for OpenROAD

The kernel will probably solve this eventually — mount namespaces + overlayfs or an io_uring-based batch symlink API. But "eventually" doesn't help us today, and anything that requires kernel features or user configuration is a non-starter.

What we can do now:

1. **Shrink the runfiles** (what PR #9973 does). Every file we remove from the dependency graph is one fewer symlink per test. Switching to jimtcl or a minimal Tcl eliminates 700+ files from every test target.
2. **Fewer, fatter targets**: restructure so fewer Bazel targets need the full runfiles tree. `cc_test()` targets that link only the module under test don't need Tcl, don't need the full OpenROAD binary, don't create 1,242 symlinks.
3. **Watch Bazel upstream**: `--experimental_sandbox_shm` and `--experimental_reuse_sandbox_directories` may graduate to defaults. When they do, everyone benefits with zero config.
4. **Don't add config burden**: if a fix requires users to add flags, install packages, or run setup scripts, it's not a fix — it's a new problem. The bar is: `git clone && bazelisk build` works.

hzeller:
There is another thing I want to explore: tcl actually has a virtual filesystem feature under the hood somewhere, and it would be possible to just link the content of all the additional files to the library. Not only would that avoid setting up runfiles (apparently with a bunch of symlinks), but it would also mean that the tcl interpreter would just work without having to access any external file.

hzeller:
can you give an example of two tests that before required `n` symbolic links and afterwards less ?

hzeller:
I am also wondering if part of the problem is, that there are many different scripts, but that are actually symbolic links to a single one ( #7042 ).
That needs to be fixed, but I wonder if this is the reason why stuff ends up being done many times instead of just once ?

oharboe:
I can never get the time back I lost to tcl, like finding out that all file operations have second only accuracy and that this breaks make...

But, I can turn off the oxygen.

I want to try to convert a module to idiomatic bazel python tests importing python and getting rid of the the .ok files to boot and see if that solves the symlink problem and makes tests nicer and more maintainable.

maliberty:
I agree with the general premise that bazel does too much symlinking.  I'm curious if python suffers from the same as the language has all sorts of ancillary files too.

hzeller:
I am working on #9980 which will eliminate that problem (and also makes it possible to instal openroad build with bazel outside of bazel as if you copy the binary outside its bazel-bin/ it will not work anymore unless also the runfiles are transplanted)

oharboe:
😌

PR #9983, opened 2026-03-28
pad: use unordered_map and faster rtree lookup

gadfort:
## Summary
Paired with: https://github.com/The-OpenROAD-Project/OpenROAD/pull/9982
This takes the runtime from 97s to 55s.
The PR was broken out because these required some changes to the `geom` and `geom_boost` headers.

## Type of Change
- Breaking change

## Impact
No change, just speed up

## Verification

## Related Issues
[Link issues here]

### DIFF
--- a/src/odb/include/odb/geom.h
+++ b/src/odb/include/odb/geom.h
@@ -1436,6 +1436,11 @@ inline void Cuboid::print(const char* prefix)
 using utl::format_as;
 #endif
 
+inline void hash_combine(size_t& seed, size_t value)
+{
+  seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+}
+
 }  // namespace odb
 
 namespace std {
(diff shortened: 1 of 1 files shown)


src/odb/include/odb/geom.h
```diff
@@ -1423,3 +1437,17 @@ using utl::format_as;
...
 }  // namespace odb
+
+namespace std {
+template <>
+struct hash<odb::Point>
+{
+  size_t operator()(const odb::Point& p) const noexcept
+  {
+    size_t seed = 0;
+    seed ^= std::hash<int>{}(p.x()) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
maliberty: It would be good to split out 
```
inline void hash_combine(size_t& seed, size_t value) {
    seed ^= value + 0x9e3779b9 + (seed << 6) + (seed >> 2);
}
```
to make it reusable for other types.  The you can
```
    hash_combine(seed, std::hash<int>{}(p.x()));
    hash_combine(seed, std::hash<int>{}(p.y()));
```

gadfort: Any particular file I should add that that to or just do that in this file?

src/pad/src/RDLRouter.cpp
```diff
@@ -30,6 +31,7 @@
 #include "odb/dbTransform.h"
 #include "odb/dbTypes.h"
 #include "odb/geom.h"
+#include "odb/geom_boost.h"
 #include "pad/ICeWall.h"
```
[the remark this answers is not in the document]
gadfort: @hzeller how do we (if possible) correct clang-tidy to correctly identify that this file is used?

src/odb/include/odb/geom.h
gadfort: I decided to leave it in geom.h, but it could move in a separate PR to compare.inc or elsewhere as appropriate

PR #9991, opened 2026-03-30
ram: Cross platform testing and support

braydenlouie:
## Summary
- Replaces hardcoded sky130 port names with a PDK-independent pin mapping system
- Introduces 'PortRole' struct with 'PortRoleType' enum that is used to classify Liberty ports by role instead of name
- Verified against both sky130hd and nangate45

## Type of Change

- New feature
- Refactoring

## Impact
RamGen is no longer tied to sky130 port naming conventions and allows using any PDK with a valid Liberty file, with pin names resolved automatically. The existing sky130 behavior is preserved.

## Verification
- [x] I have verified that the local build succeeds (`./etc/Build.sh`).
- [x] I have run the relevant tests and they pass.
- [x] My code follows the repository's formatting guidelines.

## Related Issues
#9468 
#9392

### DIFF
--- a/src/ram/include/ram/ram.h
+++ b/src/ram/include/ram/ram.h
@@ -71,11 +71,6 @@ struct PortRole
 // for map so that keys are comparable
 #ifndef SWIG
   auto operator<=>(const PortRole&) const = default;
-#else
-  bool operator<(const PortRole& other) const
-  {
-    return std::tie(type, index) < std::tie(other.type, other.index);
-  }
 #endif
 };
 


braydenlouie:
@rovinski I haven't completed testing with `generate_ram`yet, but the `generate_ram_netlist` (just placing all cells, no routing, pdngen, etc.) for Nangate45 is working. Just wanted to put this draft PR up so you can take a look at let me know if this implementation is viable.

src/ram/include/ram/ram.h
```diff
@@ -50,6 +51,30 @@ class TritonRoute;
...
+  DataIn,
+  DataOut,
+  WriteEnable,
+  TriEnable,
+  Select,  // for mux support in future
+  Power,
+  Ground
+};
+
+struct PinRole
```
rovinski: I would change `Pin` --> `Port` in all cases as technically you are dealing with the logical port here.

src/ram/test/CMakeLists.txt
```diff
@@ -2,5 +2,6 @@ or_integration_tests(
   "ram"
   TESTS
     make_8x8
+    make_7x7_nangate45
```
rovinski, line 5: Can you also update `make_8x8` to `make_8x8_sky130`?

src/ram/src/ram.cpp
```diff
@@ -388,6 +403,61 @@ dbMaster* RamGen::findMaster(
...
+          && it->first.type == PinRoleType::DataIn) {
+        pin_map.erase(it);
+        break;
+      }
+    }
+    pin_map[{PinRoleType::TriEnable, 0}] = tri_enable_name;
+  }
+  delete port_iter;
+  return pin_map;
+}
```
rovinski, line 489: Should have some error checking to avoid using badly defined cells. e.g.
* Exactly 1 primary power
* Exactly 1 primary ground
* Exactly the number of pins which are expected for the given cell type

Or I guess I suppose you could check those things after the cell map has been made, if easier.

It doesn't need to be solved in this PR, but you should mark as a TODO item to check against various kinds of flip-flops to make sure invalid ones are flagged, e.g. scan flops, flip-flops with reset, etc. Anything that doesn't have only clk, D, and Q ports. There are occasionally some flip-flops that have a Qn (negated Q) output and those are ok, Qn is just left disconnected.

src/ram/test/make_7x7_nangate45_behavioral.vok
```diff
@@ -0,0 +1,32 @@
...
+      end
+    end
+  end
+
+  // read logic
+  always @(*) begin
+    Q = mem[addr_rw];
+  end
+
+endmodule
```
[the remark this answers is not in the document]
rovinski: Gemini is correct here. Something not right with the behavioral model.

src/ram/test/make_7x7_nangate45_behavioral.vok
rovinski, line 1: It isn't sufficient to just change the vok file, if the verilog generator is wrong, then it needs to be fixed. How was the vok originally created?

src/ram/test/make_7x7_nangate45.tcl
```diff
@@ -0,0 +1,43 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024-2025, The OpenROAD Authors
+#
+# Test: ram/make_7x7_nangate45
+# Verifies that generate_ram works correctly with the Nangate45 technology
+# node using an intentionally odd 7-word x 1-byte (56-bit) configuration
```
rovinski:
```suggestion
# node using an intentionally odd 7-word x 7-bit (49-bit) configuration
```

src/ram/test/make_7x7_nangate45.tcl
```diff
@@ -0,0 +1,43 @@
...
+# to exercise the decoder logic on a non-power-of-2 word count.
+# Addresses Issue #9468.
```
rovinski: Don't need to include that here
```suggestion
```

src/ram/src/ram.cpp
```diff
@@ -474,7 +587,13 @@ void RamGen::ramPdngen(const char* power_pin,
   ground_net->setSpecial();
   ground_net->setSigType(odb::dbSigType::GROUND);
 
-  // find a way to get the power and ground net names associated with cells used
+  // finds power and gorund of a cell if not given
```
rovinski:
```suggestion
  // finds power and ground of a cell if not given
```

src/ram/src/ram.cpp
```diff
@@ -474,7 +587,13 @@ void RamGen::ramPdngen(const char* power_pin,
...
+  if (!power_pin || power_pin[0] == '\0') {
+    power_pin = inv_ports_[{PortRoleType::Power, 0}].c_str();
+  }
+  if (!ground_pin || ground_pin[0] == '\0') {
+    ground_pin = inv_ports_[{PortRoleType::Ground, 0}].c_str();
+  }
```
rovinski: This isn't technically sufficient. Although it would be really odd, it is still technically possible for other masters to have different power/ground names. A complete solution would be to iterate through all of the masters and collect the names into a `std::set`. Then, you would do a global connect on each member of the set.

99% of the time the set will only have 1 member, but it is good to handle all cases.

This is actually a separate item in #9392: "Add automatic detection of standard cell power/ground names"

It would be nice to keep that as a separate PR to only focus on one change.

src/ram/test/make_7x7_nangate45_behavioral.vok
braydenlouie: The Verilog Generator was fine. The original file was generated before we added the bit masking, so it defaulted to 8 bits in the mask. The test has been changed so that it has 7 bits per mask; the dimensions have changed, but so have the input and output port lengths.

src/ram/src/ram.cpp
braydenlouie: Got it, I'll just keep the 'power_pin' and 'ground_pin' as required user-defined parameters now, and we can fix this in another PR.

src/ram/test/make_7x7_nangate45.ok
```diff
@@ -0,0 +1,110 @@
...
+[WARNING ODB-0217] duplicate VIARULE (Via3Array-2) ignoring...
+[WARNING ODB-0217] duplicate VIARULE (Via4Array-0) ignoring...
+[WARNING ODB-0217] duplicate VIARULE (Via5Array-0) ignoring...
+[WARNING ODB-0217] duplicate VIARULE (Via6Array-0) ignoring...
+[WARNING ODB-0217] duplicate VIARULE (Via7Array-0) ignoring...
+[WARNING ODB-0217] duplicate VIARULE (Via8Array-0) ignoring...
+[WARNING ODB-0217] duplicate VIARULE (Via9Array-0) ignoring...
+[INFO ODB-0394] Duplicate site FreePDK45_38x28_10R_NP_162NW_34O in Nangate45 already seen in Nangate45_tech
+[INFO ODB-0227] LEF file: Nangate45/Nangate45.lef, created 135 library cells
+[WARNING RAM-0022] No tapcell is specified.
```
rovinski: I think ng45 has a tapcell, right? If not, use an x1 fill cell as the tap cell.

src/ram/test/make_7x7_nangate45.tcl
```diff
@@ -0,0 +1,42 @@
...
+  -mask_size 7 \
+  -word_size 7 \
+  -num_words 7 \
+  -read_ports 1 \
+  -storage_cell DFF_X1 \
+  -power_pin VDD \
+  -ground_pin VSS \
+  -routing_layer {metal1 0.07} \
+  -ver_layer {metal4 0.28 8} \
+  -hor_layer {metal5 0.28 8} \
```
rovinski: Does this work? I thought the metals needed to be the first 2 above the first metal layer.

We generally don't want to use above M3 if it can be helped.

braydenlouie: I think Nangate45's density is just too high for us to route on metal2 and metal3. Any valid combination of pitch and width has resulted in violations with the detailed router. I will note that these pitch and width dimensions don't work when tapcells are inserted, so this test will probably change.

braydenlouie:
@maliberty @rovinski PR is ready for review

src/ram/include/ram/ram.h
```diff
@@ -50,6 +52,30 @@ class TritonRoute;
...
+struct PortRole
+{
+  PortRoleType type;
+  int index;
+
+  // for map so that keys are comparable
+  bool operator<(const PortRole& other) const
+  {
+    return std::tie(type, index) < std::tie(other.type, other.index);
+  }
```
maliberty: In c++20 you can use
```
auto operator<=>(const PortRole&) const = default;
```
and get all six operators

src/ram/src/ram.cpp
```diff
@@ -388,6 +404,86 @@ dbMaster* RamGen::findMaster(
...
+      }
+    } else if (dir->isAnyOutput()) {  // catches isOutput() and isTristate()
+      pin_map[{PortRoleType::DataOut, 0}] = lib_port->name();
+    } else if (dir->isInput()) {
+      pin_map[{PortRoleType::DataIn, in_idx++}] = lib_port->name();
+    }
+  }
+
+  if (!tri_enable_name.empty()) {
+    // find and remove it from DataIn
```
maliberty, line 453: Why would a tristate enable appear as a DataIn? That doesn't appear possible in the prior loop.

src/ram/src/ram.cpp
```diff
@@ -388,6 +404,86 @@ dbMaster* RamGen::findMaster(
...
+                   power_count);
+  }
+  if (ground_count != 1) {
+    logger_->error(RAM,
+                   29,
+                   "Cell {} must have exactly 1 primary ground pin, found {}",
+                   master->getName(),
+                   ground_count);
+  }
+  delete port_iter;
```
maliberty: avoid manual memory management with a smart pointer and no delete

std::unique_ptr<CellPortIterator> port_iter = liberty->portIterator();

src/ram/src/ram.cpp
braydenlouie: The tristate enable pin has no distinguishing Liberty flag and is treated as an input port, so the only way to identify it is through the `tristateEnable()` expression on the tristate output port. Since port iteration order is arbitrary and PDK-dependent, the tristate output port may appear after the enable pin, meaning there is no guarantee the enable pin name is known before it is encountered during iteration. The other option is to do the check for the name first, before iterating over all the ports again and classifying them.

maliberty: Got it.  Perhaps a comment would help as it is a bit tricky

PR #10018, opened 2026-04-01
Add column mux ratio for ratio = 1, 2, 4

### LINKED ISSUE #9843: ram: column/mux support
The ram module currently does not support [column muxing](https://siliconvlsi.com/what-is-column-mux-in-memory/), which is very useful for both 1) changing the aspect ratio of the generated RAM and 2) optimizing the timing of the circuit when words >> bits.

This feature should add a new optional flag `-column_mux_ratio` to `generate_ram`. If the flag is not specified, the column/mux ratio should be 1, which results in the current behavior.

The valid values for the ratio should be any power of 2, as long as it is less than the total number of words. In practice, only 1, 2, 4, and maybe 8 would typically be used. It is fine to mark larger ratios as unsupported.

Muxes should be constructed using as few standard cells as possible, likely using an AOI gate and inverters (see e.g. page 13 on [these slides](https://www.cerc.utexas.edu/~jaa/vlsi/lectures/3-2.pdf))

The muxes should be added to the top row(s) of the ram near the Q outputs. 2:1 muxes should be simple. For each pair of columns, there should be a single AOI and an inverter to mux the output. There will also need to be a single inverter for the whole array to create the inverted select signal for all muxes.

tnguy19:
## Summary
Add `-column_mux_ratio` parameter to `generate_ram` and `generate_ram_netlist` with default value = 1. 

The column_mux_ratio value indicate the number of words that will share the same physical row, and currently this implementation will work correctly for values = 1, 2, and 4*. 

Muxes are implemented using AOI22s for ratio =2 (one AOI22+inverter) and ratio =4 (2 stage-1 AOI22s feeding a stage-2 AOI22). 

Some changes to the code to accommodate this feature additions are also added.

Configs tested and passed: RAM8x8, 16x8, 32x6, 64x8, 128x8 with `column_mux_ratio` = 1, 2, 4

## Type of Change

- New feature

## Impact
- `generate_ram` and `generate_ram_netlist` can be use with option `-column_mux_ratio` with default value = 1. Scripts that do not have `column_mux_ratio` are not affected
- Change from using `num_words` to `num_rows` to indicate the number of physical row that will be created. For `column_mux_ratio` >1, `num_rows = num_words / column_mux_ratio` physical rows are generated. Column mux are placed in the buffer row
- Change naming instances with format `storage_{row}_{bit}` to `storage_{row}_{word}_{bit}` to reflect the placement of multiple words in one row when `column_mux_ratio` > 1

## Verification

## Related Issues
Fixes #9843 

@rovinski


src/ram/src/ram.cpp
```diff
@@ -663,52 +816,119 @@ void RamGen::generate(const int mask_size,
...
+             "inv_addr_0",
+             inv_cell_,
+             {{"A", addr_inputs_[0]->getNet()}, {"Y", inv_addr[0]}});
+    makeInst(word_sel_cell.get(),
+             "word_sel",
+             "inv_addr_1",
+             inv_cell_,
+             {{"A", addr_inputs_[1]->getNet()}, {"Y", inv_addr[1]}});
+    ram_grid_.addCell(std::move(word_sel_cell), col_cell_count);
   }
```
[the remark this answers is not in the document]
tnguy19: fix added in c5da371

src/ram/src/ram.cpp
```diff
@@ -759,22 +1104,25 @@ void RamGen::generate(const int mask_size,
...
                "decoder",
                fmt::format("inv_{}", i),
                inv_cell_,
                {{"A", addr_inputs_[i]->getNet()}, {"Y", inv_addr[i]}});
       cell_inv_layout->addCell(std::move(inv_grid_cell));
-      for (int filler_count = 0; filler_count < num_inputs - 1;
+      for (int filler_count = 0; filler_count < num_row_bits - 1;
            ++filler_count) {
         cell_inv_layout->addCell(nullptr);
       }
```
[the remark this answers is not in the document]
tnguy19: fix added in c5da371

tnguy19:
Ready for review @rovinski

src/ram/test/make_8x8.tcl
```diff
@@ -13,6 +13,7 @@ generate_ram \
   -mask_size 8 \
   -word_size 8 \
   -num_words 8 \
+  -column_mux_ratio 1 \
```
rovinski: This shouldn't be needed, the default col/mux ratio should be 1.

src/ram/test/make_8x8_mux2_sky130.ok
```diff
@@ -0,0 +1,71 @@
...
+[INFO RAM-0003] Generating RAM8x8
+[INFO RAM-0016] Selected inverter cell sky130_fd_sc_hd__clkinv_1
+[INFO RAM-0016] Selected tristate cell sky130_fd_sc_hd__ebufn_1
+[INFO RAM-0016] Selected and2 cell sky130_fd_sc_hd__and2_0
+[INFO RAM-0016] Selected clock gate cell sky130_fd_sc_hd__dlclkp_1
+[INFO RAM-0016] Selected buffer cell sky130_fd_sc_hd__buf_1
+[INFO RAM-0016] Selected aoi22 cell sky130_fd_sc_hd__a22oi_1
+[INFO RAM-0024] Behavioral Verilog written for RAM8x8
+[INFO PDN-0001] Inserting grid: ram_grid
+[INFO PPL-0067] Restrict pins [ D[0] Q[0] D[1] Q[1] D[2] ... ] to region 0.00u-213.90u at the TOP edge.
```
rovinski, line 12: Missing the pin constraints for the right side edge

src/ram/test/make_8x8_mux4_behavioral.vok
rovinski, line 1: all of the 8x8 vok files are identical. Just keep the one that's already there and compare against that one.

rovinski (changes requested):
As discussed in meeting, please increase the power grid pitch to be at least 4 bits wide (whatever that translates to in um)

src/ram/src/ram.cpp
```diff
@@ -627,6 +777,33 @@ void RamGen::generate(const int mask_size,
...
+                   "values of 1, 2, or 4.");
+  }
+
+  if (num_words % column_mux_ratio != 0) {
+    logger_->error(RAM,
+                   34,
+                   "num_words ({}) must be divisible by column_mux_ratio ({}).",
+                   num_words,
+                   column_mux_ratio);
+  }
```
rovinski, line 1053: Add a `TODO` comment about adding support for non-divisible word counts.
It would mean that the last row has some empty spaces instead of bits, but it should be possible and won't force the user to use any more words than they actually want.

src/ram/src/ram.cpp
```diff
@@ -614,6 +614,33 @@ void RamGen::ramPdngen(const char* power_pin,
                        int hor_width,
                        int hor_pitch)
 {
+  // check that vertical strap pitch is at least 4 bits wide for column side
+  ver_pitch = std::max(ver_pitch, 4 * ram_grid_.getWidth());
```
rovinski: 4x bit width is just a suggestion. I suggested it for the tcl test so that the test is actually reflective of a realistic use case.

We should not prevent users from setting the pitch to be smaller. You can emit a warning with the suggestion but do not override the user's specification.

src/ram/src/ram.cpp
```diff
@@ -614,6 +614,33 @@ void RamGen::ramPdngen(const char* power_pin,
...
+  }
+  if (die.dx() < ver_pitch) {
+    logger_->error(
+        RAM,
+        32,
+        "Die width ({:.2f} um) is less than vertical strap pitch ({:.2f} um). "
+        "Use a smaller -ver_layer pitch.",
+        die.dx() / dbu_per_um,
+        ver_pitch / dbu_per_um);
+  }
```
rovinski, line 843: Aren't these checks on master? You should merge master rather than creating a new, separate commit

src/ram/src/ram.cpp
```diff
@@ -614,6 +614,33 @@ void RamGen::ramPdngen(const char* power_pin,
...
+
+  // need parameters for power and ground nets
+  auto power_net = dbNet::create(block_, "VDD");
```
rovinski, line 846: This can't be right. Faulty merge?

src/ram/src/ram.cpp
tnguy19: There was a faulty merge that should be resolved in 538412e

src/ram/src/ram.cpp
tnguy19: Resolved in 538412e

rovinski:
@tnguy19 CI still failing

rovinski:
@gemini-code-assist /review

src/ram/src/ram.cpp
```diff
@@ -452,6 +508,99 @@ void RamGen::findMasters()
...
+          aoi22_in_a1_ = t1->left()->left()->port()->name();
+          aoi22_in_a2_ = t3->left()->left()->port()->name();
+          aoi22_in_b1_ = t1->right()->left()->port()->name();
+          aoi22_in_b2_ = t2->right()->left()->port()->name();
+        }
+        aoi22_out_ = p->name();
+        break;
+      }
+    }
+    delete port_iter;
```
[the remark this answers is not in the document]
tnguy19: Updated in 1fd2b36 but this cause clang tidy error so this suggestion is ignored

src/ram/src/ram.cpp
```diff
@@ -452,6 +508,101 @@ void RamGen::findMasters()
...
+          aoi22_in_a1_ = t1->left()->left()->port()->name();
+          aoi22_in_a2_ = t3->left()->left()->port()->name();
+          aoi22_in_b1_ = t1->right()->left()->port()->name();
+          aoi22_in_b2_ = t2->right()->left()->port()->name();
+        }
+        aoi22_out_ = p->name();
+        break;
+      }
+    }
+    if (port_iter) {
```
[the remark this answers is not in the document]
tnguy19: This change contradicts the suggestion by gemini-code-assist above. Reverted back to original change in 9163e08

tnguy19:
@maliberty @rovinski PR is ready for review

maliberty:
@rovinski please take the first pass on reviewing

src/ram/src/ram.cpp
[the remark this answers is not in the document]
rovinski: I don't know why it made this suggestion - deleting a nullptr is a no-op.

rovinski: Copying feedback from another thread:

> avoid manual memory management with a smart pointer and no delete
> `std::unique_ptr port_iter = liberty->portIterator();`

rovinski (approved):

src/ram/src/ram.cpp
```diff
@@ -452,6 +508,99 @@ void RamGen::findMasters()
...
+            auto l2 = l1->left();
+            auto t3 = l1->right();
+            if (!isAndOfNots(t3) || !l2 || l2->op() != sta::FuncExpr::Op::or_) {
+              return false;
+            }
+            return isAndOfNots(l2->left()) && isAndOfNots(l2->right());
+          }
+          return false;
+        },
+        "aoi22");
```
rovinski, line 777: @maliberty I'm not an expert on sta functions so I don't know if there is a better way to systematically identify aoi gates

QuantamHD: Do not do it this way. Just simulate the truth table for the gate and record the output of the truth table into a unit16t. There are exactly 3 unique permutations for an aoi gate. Just calculate them ahead of time. Then check if the output of the candidate gate matches 1 of the three unique permutations

tnguy19: New version of this part of code updated in 0deefd4

tnguy19:
There was an update that introduce PDK-independent pin mapping, some part of the code has been updated to follow this particularly the logic to determine aoi gates used in mux is now also PDK-independent

src/ram/test/regression_tests.tcl
```diff
@@ -1,4 +1,7 @@
 record_tests {
   make_8x8_sky130
   make_7x7_nangate45
+  make_8x8
```
rovinski: no `make_8x8` anymore, that got moved to `make_8x8_sky130`

src/ram/src/ram.cpp
```diff
@@ -569,6 +628,141 @@ void RamGen::findMasters()
...
+          // collect all input ports for this cell
+          std::vector<sta::LibertyPort*> inputs;
+          auto port_iter = cell->portIterator();
+          while (port_iter->hasNext()) {
+            auto p = static_cast<sta::ConcretePort*>(port_iter->next());
+            if (p->direction()->isInput()) {
+              inputs.push_back(p->libertyPort());
+            }
+          }
+          delete port_iter;
```
rovinski: Use smart pointer and no delete
`std::unique_ptr port_iter = liberty->portIterator();`

rovinski (changes requested):
First pass, don't have time to fully re-review today

src/ram/src/ram.cpp
tnguy19: Updated in 38c9cb2

src/ram/src/ram.cpp
```diff
@@ -500,7 +500,20 @@ std::map<PortRole, std::string> RamGen::buildPortMap(dbMaster* master)
...
-      pin_map[{PortRoleType::DataOut, 0}] = lib_port->name();
+      // prefer non-inverted output (e.g. Q over QN)
+      // if no DataOut yet, assign unconditionally
+      // if DataOut already set, only override if current port is non-inverted
+      // (i.e. its name does not end with N or _N indicating
+      // active-low/inverted)
+      std::string port_name = lib_port->name();
+      bool is_inverted
+          = !port_name.empty()
+            && (port_name.back() == 'N' || port_name.back() == 'B');
```
rovinski, line 511: This isn't going to be robust enough. Is there no way to determine an inverted output from STA?

src/ram/src/ram.cpp
```diff
@@ -726,7 +726,7 @@ void RamGen::findMasters()
 
           // collect all input ports for this cell
           std::vector<sta::LibertyPort*> inputs;
-          std::unique_ptr<sta::LibertyCellPortIterator> port_iter(
+          std::unique_ptr<sta::ConcreteCellPortIterator> port_iter(
```
rovinski, line 729: Any particular reason for the change to ConcereteCell?

tnguy19: This is because it triggers a clang-tidy suggestion from last commit `warning: no matching constructor for initialization of 'std::unique_ptrsta::LibertyCellPortIterator' [clang-diagnostic-error]`

rovinski:
@gemini-code-assist /review

src/ram/src/ram.cpp
tnguy19: I found 2 functions in the sta's Sequential.hh file (` /OpenROAD/src/sta/include/sta/Sequential.hh`) that returns the output and inverted output port (`output()` and `outputInv()`) . Do you want me to use that to check here?

src/ram/src/ram.cpp
```diff
@@ -703,14 +921,34 @@ void RamGen::ramPdngen(const char* power_pin,
...
+    }
+  }
+  block_->addBTermsToConstraint(clk_pins, right_constraint);
+  block_->addBTermsToConstraint(we_pins, right_constraint);
+  block_->addBTermsToConstraint(addr_inputs_, right_constraint);
+
   auto pin_tech = block_->getDb()->getTech();
   io_placer_->addHorLayer(pin_tech->findLayer(hor_name));
   io_placer_->addVerLayer(pin_tech->findLayer(ver_name));
+  io_placer_->getParameters()->setCornerAvoidance(0);
```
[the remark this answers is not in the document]
rovinski: This can be ignored because the OR session will not be reused.

rovinski:
@tnguy19 I will give you the option of either:

1. Fixing the code to handle when `read_ports > 1 && col_mux_ratio > 1`
2. Throwing an error that it's unsupported to get this merged, then addressing it in a future PR.

src/ram/src/ram.cpp
braydenlouie: I tested `Sequential::output()` and `outputInv()` to distinguish `Q` and `Qn` ports. Found that these return internal state variable names (`IQ`/`IQN`) rather than the actual Liberty port names (`Q`/`QN`), so they can't be used directly for port name matching. Also tested using `function()->op()` to check if the output is a `not_` expression to identify `QN`, but both `Q` and `QN` returned the same function op, so that approach is also unreliable. The port iterator gives `Q`/`QN`, but neither method can distinguish them without relying on the port name directly.

tnguy19:
An error check for when `read_ports > 1 && col_mux_ratio > 1` will be added for now since multiple r/w port supports are also being worked on at the moment. I'll add support for col/mux ratio with multiple read ports when that feature is completed

src/ram/src/ram.cpp
rovinski: @maliberty @povik do you have any tips on how to identify an inverted vs. non-inverted output of a flip-flop?

rovinski:
Waiting to find out if there's a better way to distinguish register outputs before approving, but that's the last thing.

src/ram/src/ram.cpp
maliberty: You need to ask the output port for its function and see if it matches up with the internal pos/neg variable.  For example in ng45
```
  cell (DFFRS_X1) {
	ff ("IQ" , "IQN") ...
	pin (Q) {
		direction		: output;
		function		: "IQ";
	pin (QN) {
		direction		: output;
		function		: "IQN";

```

tnguy19: New code to handle this is updated in a5d3d16

tnguy19:
@maliberty CI test is failing because of unrelated issue (Bazel tests failed), can you review the updated code for aoi port selection in a5021f0 ?

maliberty:
This has merge conflicts to resolve

tnguy19:
@maliberty Merge conflict is resolved

maliberty:
@rovinski do you approve?

src/ram/test/make_8x8_sky130.ok
```diff
@@ -2,20 +2,28 @@
 [INFO ODB-0227] LEF file: sky130hd/sky130_fd_sc_hd_merged.lef, created 437 library cells
 [INFO RAM-0003] Generating RAM8x8
 [INFO RAM-0016] Selected inverter cell sky130_fd_sc_hd__clkinv_1
+[INFO RAM-0030] buildPortMap DataOut for sky130_fd_sc_hd__clkinv_1: Y
```
rovinski: @tnguy19 this message adds too much noise. Either condense the message per cell all into one line, or turn this into a debug message

src/ram/test/make_8x8_mux4_sky130.tcl
```diff
@@ -0,0 +1,40 @@
...
+  -filler_cells {sky130_fd_sc_hd__fill_1 sky130_fd_sc_hd__fill_2 \
+    sky130_fd_sc_hd__fill_4 sky130_fd_sc_hd__fill_8} \
+  -tapcell sky130_fd_sc_hd__tap_1 \
+  -max_tap_dist 15 \
+  -write_behavioral_verilog $behavioral_file
+
+write_verilog [make_result_file make_8x8_mux4_sky130.v]
+
+set lef_file [make_result_file make_8x8_mux4_sky130.lef]
+write_abstract_lef $lef_file
```
rovinski: Per #10154 please use `-bloat_occupied_layers` now to simplify the LEF.

src/ram/test/make_8x8.tcl
```diff
@@ -0,0 +1,39 @@
...
+    sky130_fd_sc_hd__fill_4 sky130_fd_sc_hd__fill_8} \
+  -tapcell sky130_fd_sc_hd__tap_1 \
+  -max_tap_dist 15 \
+  -write_behavioral_verilog $behavioral_file
+
+set verilog_file [make_result_file make_8x8.v]
+write_verilog $verilog_file
+
+set lef_file [make_result_file make_8x8.lef]
+write_abstract_lef $lef_file
```
rovinski: same

src/ram/test/make_8x8_sky130.ok
tnguy19: I removed this log entirely since it was mainly for debugging the port mapping issue which has been resolved so I think its no longer needed if it adds unnecessary info in the log

rovinski: I believe you can simply use the `debugPrint()` function which will only print the message if debug mode is turned on. Or get rid of it, either is fine.

rovinski:
@tnguy19 bazel build is failing. Do you think that you can investigate why that might be? Usually one of the largest culprits is sorting or comparing a list of pointers.

tnguy19:
@rovinski I took  a look and made some changes but the CI log still shows this difference:
```
Differences found at line 1792.
NEW met1 ( 4669 629 ) ( 5267 * )
NEW met1 ( 4669 629 ) ( 4853 * )
No differences found.
```
that appears only on the Bazel build. I think this might be some non-determinism in how the routing works (I believe this is from the `storage_2_0_0.gclock` net) since I have added the changes to ensure all AOI22 port assignments are deterministic with sorting by name. Do you have any idea what else can be causing this as it passed correctly for both local and ubuntu test

rovinski:
@tnguy19 Seems to be a CI error. Try pushing a small change to force a rerun

rovinski (approved):
@vvbandeira @maliberty Bazel tests seem to be failing for an unrelated reason

This otherwise looks good to me.

maliberty (approved):

PR #10029, opened 2026-04-01
ant: Supports LEF58_ANTENNAGATEPLUSDIFF rule in ant and odb

QuantamHD:
## Summary
This was mostly about me trying to use AntiGravity to see how well it could implement an unsupported LEF property in our PDK. It worked reasonably well. It needed some guiding, and my knowledge of the code did help.

Updated antenna checker to account for this lef property.

How par is calculated if LEF58_ANTENNAGATEPLUSDIFF is present

```
PAR = {(cutFactor * cut) * diffReducePWL(diff) - (minusDiffFactor * diff)} / {gatePWL(gate) + plusDiffProtect(diff)}
```

Updates schema and other properties to account for this rule.

ANTENNAGATEPLUSDIFF {OXIDE1 | OXIDE2 }
{plusDiffFactor |
PWL ((diffArea_1 plusDiffProtect_1) (diffArea_1 plusDiffProtect_2) ...)}

Technically this is supposed to support up to 32 different oxides, but the code as is hard codes it to 2 so I'm leaving that here.

## Type of Change
Antenna check will now respect LEF58_ANTENNAGATEPLUSDIFF.

## Impact

## Verification
- [ x] I have verified that the local build succeeds (`./etc/Build.sh`).
- [x ] I have run the relevant tests and they pass.
- [x ] My code follows the repository's formatting guidelines.
- [x ] **I have signed my commits (DCO).**


src/ant/src/AntennaChecker.cc
```diff
@@ -405,15 +414,22 @@ void AntennaChecker::Impl::calculateWirePar(odb::dbTechLayer* tech_layer,
     info.PAR = (diff_metal_factor * info.area) / info.iterm_gate_area;
     info.PSR = (diff_side_metal_factor * info.side_area) / info.iterm_gate_area;
 
+    double plus_diff_protect = plus_diff_factor * info.iterm_diff_area;
+    if (!layer_info_[tech_layer].gate_plus_diff_idx.empty()) {
+      odb::dbTechLayerAntennaRule::pwl_pair pwl
+          = {.indices = layer_info_[tech_layer].gate_plus_diff_idx,
+             .ratios = layer_info_[tech_layer].gate_plus_diff_ratios};
+      plus_diff_protect = getPwlFactor(pwl, info.iterm_diff_area, 0.0);
+    }
```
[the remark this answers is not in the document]
QuantamHD: I get the complaint here, but given that these things are actually calculating different values. I'm inclined to say duplicate it. It's not clear to me that you actually want to merge this logic together.

src/odb/src/lefin/lefin.cpp
```diff
@@ -743,8 +745,10 @@ void lefinReader::layer(LefParser::lefiLayer* layer)
       } else if (!strcmp(layer->propName(iii), "LEF58_MAXSPACING")) {
         MaxSpacingParser parser(l, this);
         parser.parse(layer->propValue(iii));
-      } else if (!strcmp(layer->propName(iii), "LEF57_ANTENNAGATEPLUSDIFF")) {
-        AntennaGatePlusDiffParser parser(layer, this);
+      } else if (!strcmp(layer->propName(iii), "LEF57_ANTENNAGATEPLUSDIFF")
+                 || !strcmp(layer->propName(iii),
+                            "LEF58_ANTENNAGATEPLUSDIFF")) {
+        AntennaGatePlusDiffParser parser(l, this);
         parser.parse(layer->propValue(iii));
```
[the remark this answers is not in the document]
QuantamHD: This is actually not duplicated in one case it's for cuts in the other its for layers. I'd say leave it as is.

src/odb/src/lefin/lefin.cpp
```diff
@@ -1010,8 +1014,12 @@ void lefinReader::layer(LefParser::lefiLayer* layer)
...
   if (layer->numAntennaModel() > 0) {
     for (j = 0; j < std::min(layer->numAntennaModel(), 2); j++) {
-      cur_ant_rule = (j == 1) ? l->createOxide2AntennaRule()
-                              : l->createDefaultAntennaRule();
+      cur_ant_rule
+          = (j == 1) ? l->getOxide2AntennaRule() : l->getDefaultAntennaRule();
+      if (!cur_ant_rule) {
+        cur_ant_rule = (j == 1) ? l->createOxide2AntennaRule()
+                                : l->createDefaultAntennaRule();
+      }
```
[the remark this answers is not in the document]
QuantamHD: Done.

QuantamHD:
@maliberty looks like the CI might have a problem. The email step is failing

maliberty:
> @maliberty looks like the CI might have a problem. The email step is failing

known, @vvbandeira is looking at it

src/ant/src/AntennaChecker.cc
QuantamHD, line 443: @luis201420 I left it out from this PR, but it's not clear to me that if `info.iterm_diff_area == 0` if we should attempt to do the PWL lookup and add it to `info.iterm_gate_area`. I noticed in all of our PDKs it would evaluate to 0 anyways, but again not clear if that should be included here as well

luis201420: While researching, I found that if there isn't a connected diffusion area, the factor cannot be applied. The ANTENNAGATEPLUSDIFF parameter is specifically applied when a diffusion connection exists to scale the effective area of ​​the gate being protected. Without diffusion, this factor is irrelevant.

QuantamHD: Cool! So correct as is then?

luis201420: Yes, correct as is. It should not use ANTENNAGATEPLUSDIFF when there is no diffusion area.

src/odb/src/db/dbTechLayerAntennaRule.cpp
```diff
@@ -292,10 +298,22 @@ void dbTechLayerAntennaRule::writeLef(lefout& writer) const
...
+    fmt::print(writer.out(), "    ANTENNAGATEPLUSDIFF  PWL ( ");
+    dbVector<double>::const_iterator diffdx_itr;
+    dbVector<double>::const_iterator ratio_itr;
+    for (diffdx_itr = ant_rule->gate_plus_diff_val_.diff_idx_.begin(),
+        ratio_itr = ant_rule->gate_plus_diff_val_.diff_ratio_.begin();
+         diffdx_itr != ant_rule->gate_plus_diff_val_.diff_idx_.end()
+         && ratio_itr != ant_rule->gate_plus_diff_val_.diff_ratio_.end();
+         diffdx_itr++, ratio_itr++) {
+      fmt::print(writer.out(), "( {:g} {:g} ) ", *diffdx_itr, *ratio_itr);
+    }
```
maliberty, line 311: Claude suggests
```
for (auto [idx, ratio] : std::views::zip(
         ant_rule->gate_plus_diff_val_.diff_idx_,
         ant_rule->gate_plus_diff_val_.diff_ratio_)) {
  fmt::print(writer.out(), "( {:g} {:g} ) ", idx, ratio);
}
```
which feels more readable.

QuantamHD: That's a c++23 feature, so not usable yet. https://en.cppreference.com/w/cpp/ranges/zip_view.html

Anything else you would prefer in this situation?

maliberty: AI hallucination, I guess just leave it as is.

PR #10039, opened 2026-04-02
grt: fix is_incremental_ flag not reset in endIncremental()

Divinesoumyadip:
`endIncremental()` was setting` is_incremental_ = true` instead of false. This means the router never actually exits incremental mode after the session ends, which can cause incorrect behavior in subsequent routing calls that check this flag.
 `is_incremental_ = false` in `endIncremental()`, which is the correct behavior since the function's purpose is to terminate the 
incremental routing session.

### DIFF
--- a/src/grt/src/GlobalRouter.cpp
+++ b/src/grt/src/GlobalRouter.cpp
@@ -333,7 +333,7 @@ void GlobalRouter::startIncremental()
 
 void GlobalRouter::endIncremental(bool save_guides)
 {
-  is_incremental_ = true;
+  is_incremental_ = false;
   fastroute_->setResistanceAware(resistance_aware_);
   updateDirtyRoutes(save_guides);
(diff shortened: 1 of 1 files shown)


src/grt/src/GlobalRouter.cpp
```diff
@@ -333,7 +333,7 @@ void GlobalRouter::startIncremental()
 
 void GlobalRouter::endIncremental(bool save_guides)
 {
-  is_incremental_ = true;
+  is_incremental_ = false;
```
eder-matheus, line 336: Actually, it is intended that at the beginning of `endIncremental` this flag is true, since incremental updates are made below with the `updateDirtyRoutes` call. The proper fix is to set `is_incremental_` to false at the end of the function

eder-matheus (changes requested):

PR #10049, opened 2026-04-03
[GPL] Add set_net_weight/unset_net_weight commands

jeffrey-song-melten:
## Summary
- Add `set_net_weight` / `unset_net_weight` TCL commands for per-net custom placement weight
- Weights stored as `dbDoubleProperty("gpl_weight")` on `dbNet` — persistent across `write_db`/`read_db`
- `NesterovBase` reads the property during `GNet` construction via existing `setCustomWeight()`

## Motivation
External tools (e.g., timing-driven convergence loops) need to guide global placement by annotating specific nets with higher importance. This is useful for IP-level characterization → integration PnR workflows where sensitive nets identified during standalone analysis should receive higher placement priority.

## Test plan
- [x] `net_weight01.tcl`: 7 positive tests (set, verify ODB property, unset, multiple nets, global_placement with weights, overwrite, write_db persistence)
- [x] `net_weight_negative.tcl`: 7 negative tests (non-existent net, zero weight, negative weight, unset non-existent, wrong arg count, non-numeric, unset without prior set)
- [x] Tests registered in CMakeLists.txt and BUILD
- [x] Golden `.ok` files generated

## Related
- Related issue: #10048 (DPL crash during incremental die shrink — separate issue, discovered during same investigation)

### DIFF
--- a/src/gpl/src/nesterovBase.cpp
+++ b/src/gpl/src/nesterovBase.cpp
@@ -1099,6 +1099,12 @@ NesterovBaseCommon::NesterovBaseCommon(
   gNetStor_.reserve(pbc_->getNets().size());
   for (auto& net : pbc_->getNets()) {
     GNet myGNet(net);
+    // Read custom placement weight from ODB property (set by set_net_weight)
+    odb::dbDoubleProperty* wp
+        = odb::dbDoubleProperty::find(net->getDbNet(), "gpl_weight");
+    if (wp) {
+      myGNet.setCustomWeight(static_cast<float>(wp->getValue()));
+    }
     gNetStor_.push_back(myGNet);
   }
 
--- a/src/gpl/src/replace.i
+++ b/src/gpl/src/replace.i
@@ -198,4 +198,36 @@ set_debug_cmd(int pause_iterations,
                     generate_images, resolved_path);
 }
 
+void
+set_net_weight_cmd(const char* net_name, float weight)
+{
+  auto* block = ord::OpenRoad::openRoad()->getDb()->getChip()->getBlock();
+  auto* net = block->findNet(net_name);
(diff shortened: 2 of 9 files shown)


src/gpl/src/nesterovBase.cpp
```diff
@@ -1099,6 +1099,12 @@ NesterovBaseCommon::NesterovBaseCommon(
   gNetStor_.reserve(pbc_->getNets().size());
   for (auto& net : pbc_->getNets()) {
     GNet myGNet(net);
+    // Read custom placement weight from ODB property (set by set_net_weight)
+    odb::dbDoubleProperty* wp
```
gudeh, line 1103: Would be nice to have better names for variables, for example renaming `wp` and `myGNet`. Perhaps `net_property`, and `temp_gnet`?

gudeh: Most importantly, the setting of timing-driven weights does not originally happen here. This is currently not protected by timing-driven mode being activated. 

I believe this should happen at: https://github.com/The-OpenROAD-Project/OpenROAD/blob/29d97c45b3e328b6141f76b72c9734f8b899ef14/src/gpl/src/timingBase.cpp#L120

Notice how we originally set weights at the following, after RSZ is executed: https://github.com/The-OpenROAD-Project/OpenROAD/blob/29d97c45b3e328b6141f76b72c9734f8b899ef14/src/gpl/src/timingBase.cpp#L173

gudeh (changes requested):

src/gpl/src/nesterovBase.cpp
gudeh: This is an interesting new feature!

PR #10099, opened 2026-04-10
bazel: make downstream-facing targets publicly visible

oharboe:
openroad_lib, ord, error_swig, error_swig-py, and options_swig were restricted to //:__subpackages__. Downstream Bazel consumers that depend on @openroad need these targets. The CMake build already exposes them unconditionally.


oharboe:
@hzeller Thoughts?

hzeller:
Looks good to me in general. We mostly need to make sure to not have things that are internal artifacts and are likely to change are kept private, but this looks like a good set of candiates that look like they want to be public.
I'll defer to @maliberty for the good judgment.

maliberty:
How do you use the .i files downstream?

oharboe:
> How do you use the .i files downstream?

Added a test and the best explanation I have.

test/visibility/WORKSPACE.bazel
```diff
@@ -0,0 +1 @@
+# Empty — bzlmod only.
```
hzeller: Don't add a WORKSPACE. That is only needed in pre bazel 7 era times that we have passed.

test/visibility/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
...
+#
+# Run:  bazelisk build --nobuild :all
+
+[alias(
+    name = t,
+    actual = "@openroad//:" + t,
+) for t in [
+    "openroad_lib",
+    "ord",
+    "error_swig",
```
hzeller: while the openroad_lib and ord seem reasonable, I wonder why you need the swigged versions

test/visibility/README.md
```diff
@@ -0,0 +1,57 @@
+# Bazel visibility test for downstream consumers
+
+## Problem
+
+OpenROAD's `BUILD.bazel` restricted several targets to
```
hzeller: this looks like an LLM explanation: full of words, but not explaining anything.

What we expect here is something like: libraries x, y and z are needed downstream for x-reason, y-reason and z-reason. This is why they are set public.

oharboe: Trying to get rid of docker, this was one fix that Claude needed to make it compile. I expect that when the dust settles and we get other things merged, clarity will come.

hzeller: if claude ran into the issue, it might be able to explain why it needed that visibility.

oharboe: I prefer starting with something that works even if I dont know why and then whittling away at it. Claude changes nothing: I use OpenROAD and ORFS without understanding them fully. I start with something that works.

ideally claude was mistaken here and nothing is needed here... We dont yet know what is going on and why claude decided it needed this. It will come. Thats the beauty of patching openroad in bazel-orfs and downstream projects, we can  deal with the pain first with a patch and then figure out what is going on and what is needed in openroad

Thank you for sheding light on this so far. Even a conclusive "I have no idea" is a strong signal.

test/visibility/WORKSPACE.bazel
oharboe: fixed

test/visibility/BUILD.bazel
oharboe: fixed

test/visibility/README.md
hzeller: I am worried about this-was-needed-at-some-time-but-nobody-knew-why resulting in a bunch of leftover things that might just hang in there forever because someone later thinks "oh, there was probably a good reason".

Maybe add a cleanup issue that collects all the pull requests to get things going, so that it is possible to re-viisit later and clean up things that are not needed. We humans need to use these tools to keep the software maintainable which the LLM tools don't understand or care about - they just are text completion tools.

oharboe: I think we have a good process now. This change is suspect, whereas others sail straight through, we know they are good. I'm fine to leave this hanging until we are sure exactly what is going on. There are many other pull requests in flight that are much less questionable to sort out first.

oharboe: Now if you *do* want a more holistic approach to this, check out origin/main of bazel-orfs and figure out how to get it to build without patches...

I'm thinking that's not an ideal solution either. Standalone, sometimes a bit mysterious, PRs and investing in understanding the context when there's a reason to, is a better way to do it...

oharboe: The horrific thing about getting rid of docker are the *enormous* turnaround times. When I tinker with something that makes the openroad/yosys bazel cache useless, 20000 files have to be rebuilt + hours of ORFS flows.

oharboe:
@maliberty The pr-merge failure is just because pr-merge was broken at the time. could you review?

test/visibility/.bazelversion
maliberty, line 1: Why is this needed?

oharboe:
will investigate exactly if and why this is needed.closing so it is out of sight out of mind meanwhile.

PR #10103, opened 2026-04-10
Bump build toolchains (bison, flex, LLVM)

oharboe:
rules_bison 0.3.1 → 0.4.bcr.1
rules_flex 0.3.1 → 0.4
toolchains_llvm 1.5.0 → 1.7.0

If bison/flex toolchain provider API changed in 0.4, the custom bazel/bison.bzl and bazel/flex.bzl rules may need updating.


oharboe:
needs more investigation

PR #10104, opened 2026-04-10
Bump spdlog 1.15.1 → 1.17.0

oharboe:
Minor version bump. If any spdlog API changed, the CMake build (DependencyInstaller.sh pins 1.15.0) may also need updating.


oharboe:
out of sight out of mind for now. seems to work.

PR #10105, opened 2026-04-10
Bump Boost 1.89.0.bcr.2 → 1.90.0.bcr.1

oharboe:
Updates all 27 boost module deps via the BOOST_VERSION constant. Boost 1.89→1.90 is source-compatible; CMake build (1.89.0) should continue working. The boost.context parse_headers patch may need rebasing if it doesn't apply cleanly to 1.90.


oharboe:
out of sight out of mind for now. seems to work.

PR #10110, opened 2026-04-11
bazel: make toolchains_llvm a dev_dependency

oharboe:
The bazel_dep was non-dev, forcing downstream Bazel consumers to inherit OpenROAD's pinned LLVM toolchain. This conflicts when consumers configure their own C++ toolchain.

The extension and toolchain registration were already dev_dependency. Making the bazel_dep dev_dependency too keeps it consistent and invisible to downstream modules. The llvm_prebuilt build_file reference works because dev deps are still resolved for the declaring module's own build.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -88,12 +88,8 @@ git_override(
 )
 
 ## Lock the compiler version and avoid any local compiler.
-## The bazel_dep stays non-dev because llvm_prebuilt's build_file
-## references @toolchains_llvm//toolchain:BUILD.llvm_repo.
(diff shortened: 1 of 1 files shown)


oharboe:
@hzeller Thoughts?

maliberty (approved):

hzeller:
Looks good.
(also, in the meantime, there is version 1.7.0, would that be something to consider ?)

oharboe:
@maliberty ugh. module.bazel.lock needs update. Can you fix it? It's going to conflict with some PR going in before this one if I do it.

@hzeller can't even do that reliably locally... I understood.

PR #10112, opened 2026-04-11
Don't include boost src, this results in first definition rule violations

hzeller:


maliberty (approved):

maliberty:
Loads of build failures

hzeller:
Looks like the cmake setup and the linux distributions there don't have the boost json configured yet.

Issue #9895, opened 2026-03-23
macOS fail across different stages of dependency installation and build locally


Sahil7741:
### Describe the bug

```
./etc/DependencyInstaller.sh -base
./etc/DependencyInstaller.sh -common -local
./etc/Build.sh
 ```
Running on macOS (Apple Silicon) results in multiple failures across different stages of dependency installation and build locally.

Script includes partial macOS support, overall setup process does not complete successfully due to a combination following

- Linux specific command (eg - `nproc`)

- Build failures in dependencies (eg - `lemon-graph` via CMake)

- Toolchain incompatibilities with modern macOS/Homebrew environments

### Expected Behavior

```
./etc/DependencyInstaller.sh -base
./etc/DependencyInstaller.sh -common -local
./etc/Build.sh
 ```
Should complete successfully on macOS.

### Environment

```shell
(.venv) sahil@Sahils-MacBook-Air:~/Developer/OpenROAD $ ./etc/Env.sh 
Git commit: 4589d89a4015fe19e60880907a3e76777f0dce50
kernel: Darwin 25.3.0
os:             macOS           26.3.1
cmake version 4.3.0
CMake Deprecation Warning at CMakeLists.txt:18 (cmake_policy):
  The OLD behavior for policy CMP0144 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


CMake Warning at src/CMakeLists.txt:242 (message):
  spdlog: SPDLOG_FMT_EXTERNAL=ON


CMake Warning at /opt/homebrew/opt/or-tools/lib/cmake/ortools/modules/FindSCIP.cmake:49 (message):
  SCIP::libscip not provided
Call Stack (most recent call first):
  /opt/homebrew/share/cmake/Modules/CMakeFindDependencyMacro.cmake:93 (find_package)
  /opt/homebrew/share/cmake/Modules/CMakeFindDependencyMacro.cmake:125 (__find_dependency_common)
  /opt/homebrew/opt/or-tools/lib/cmake/ortools/ortoolsConfig.cmake:86 (find_dependency)
  src/gpl/CMakeLists.txt:12 (find_package)


[...]

PR #10113, opened 2026-04-11
build: Improve macOS build and dependency installation

Sahil7741:
ref #9895

- Add macOS venv setup documentation
- Improve Build.sh with better package validation and env config
- Update DependencyInstaller.sh for macOS (threads, icu4c, tcl-tk@8)
- Use pip instead of pipx

Screenshot 2026-04-12 at 3 38 14 AM


Sahil7741:
@maliberty @luarss please review

maliberty (approved):
This looks fine but you should be aware we are moving towards bazel where this is much simpler.

